PR #171, opened 2012-08-06
Add export_curves module. It is usefull for exporting selected subtrees ...

felipesanches:
...to DXF files. Only 2D content can be exported with this module.

This branch adds an export_curves command for exporting a sub-tree composed of only 2d primitives.

For instance, one can have something like:

module cuts_for_acrylic_sheet(radius){
//algoritmically define placing of screw holes, and all other necessary cuts
difference(){
square([50, 50]);
translate([10,15]) circle(r=radius);
translate([40,15]) circle(r=radius);
translate([10,35]) circle(r=radius);
translate([40,35]) circle(r=radius);
}

//then you can preview the sheet of 2mm laser-cut acrylic like this:
color([0.4, 0.4, 0.4, 0.7])
linear_extrude(height=2)
cuts_for_acrylic_sheet(radius = m3_diameter/2);

//And you can export the vector curves for sending to the laser cutter this way:
export_curves("cuts_for_acrylic_sheet_using_M3.dxf")
cuts_for_acrylic_sheet(radius = m3_diameter/2);

export_curves("cuts_for_acrylic_sheet_using_M4.dxf")
cuts_for_acrylic_sheet(radius = m4_diameter/2);

The nice thing about being able to export subtrees is that you can i.e. design several parts of a machine made of laser-cut parts and export each vector file from the appropriate portions of the SCAD code.


tcurdt:
This would be super useful.

MichaelAtOz:
Side effects, optimisation issues?

```
export("file1.stl") something();
export("file2.stl") union() { cube() import("file1.stl");
export("file1.stl") sphere();
```

If that was sequential (adding the time concept to OpenSCAD) I could handle doing such, but are we killing future parallel processing? 
Do do we add a programmed `sequential() { a(); b(); }` construct to force sequencing when needed?

I have previously argued for something similar, part() IIRC, to allow production of multiple objects in one program.

mickvav:
I'll leave it here. Looks like a long story... https://github.com/openscad/openscad/pull/1534

t-paul:
Having `export()` as side effect is problematic, e.g. see discussion in #1534 mentioned above. The special module for just one use case also would make things hard to maintain and evolve.

PR #448, opened 2013-07-30
using polylines for DXF export

felipesanches:
Everytime I send DXF files generated by openscad to our laser cutting service provider, they complain that the vectors are "all broken apart". I ended up figuring out that it means the current DXF exporter generates several independent line segments. 

So this patch uses DXF polilynes instead, which makes it much nicer to do further editing of the generated DXF vectors.

happy hacking,
Felipe Sanches
metamaquina.com.br

### DIFF
--- a/src/export.cc
+++ b/src/export.cc
@@ -138,6 +138,7 @@ void export_dxf(CGAL_Nef_polyhedron *root_N, std::ostream &output)
 {
 	setlocale(LC_NUMERIC, "C"); // Ensure radix is . (not ,) in output
 	// Some importers (e.g. Inkscape) needs a BLOCKS section to be present
+
 	output << "  0\n"
 				 <<	"SECTION\n"
 				 <<	"  2\n"
@@ -149,9 +150,12 @@ void export_dxf(CGAL_Nef_polyhedron *root_N, std::ostream &output)
 				 << "  2\n"
 				 << "ENTITIES\n";
 
+  double prev_x, prev_y;
 	DxfData *dd =root_N->convertToDxfData();
 	for (size_t i=0; i<dd->paths.size(); i++)
 	{
+    bool had_a_previous_polyline = false;
+    bool first_point = true;
 		for (size_t j=1; j<dd->paths[i].indices.size(); j++) {
 			const Vector2d &p1 = dd->points[dd->paths[i].indices[j-1]];
 			const Vector2d &p2 = dd->points[dd->paths[i].indices[j]];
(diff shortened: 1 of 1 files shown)


matthijskooijman:
I just ran into this problem as well - because the segments are cut apart, our visicut/laos powered lasercutter jumps all over the place trying to cut them one by one...

In case it helps, openjscad.org seems to generate polyline DXF files that import into inkscape and visicut just fine, though I haven't got autocad to test.

This openjscad program:

```
function main() { 
 return circle(100);
}
```

Generates this DXF file: https://gist.github.com/matthijskooijman/c71338b4a5feaf091ab4

Anyone with autocad that can confirm it still loads?

RudolfCardinal:
Since OpenSCAD-2019.05-x86_64 still exports lots of chopped-up lines, but has a perfectly good SVG export, attached is a quick standalone workaround in Python 3 to convert SVG output to DXF using LWPOLYLINE, via the "svg.path" and "ezdxf" libraries. This obviates the need to use Inkscape as an intermediate (in my case, Inkscape was struggling).

[svg_to_dxf.zip](https://github.com/openscad/openscad/files/3634995/svg_to_dxf.zip)

t-paul:
Closing, abandoned long time ago.

wheybags:
Since this comes up in google searches still, I'll add the solution I settled on here.
I tried @RudolfCardinal's script, to export as svg then convert svg->dxf, and it worked pretty well but the resulting drawing was vertically flipped. Using inkscape to convert from SVG  worked great for me instead, with a command line like the following:
`inkscape --export-filename=export_side.dxf export_side.svg --export-extension=org.ekips.output.dxf_outlines`

The only catch is that the `--export-filename` command was only recently added to inkscape, so you will need a copy of a development version. I used their bleeding edge ppa here: https://inkscape.org/release/inkscape-master/gnulinux/ubuntu/ppa/dl/
This will obviously not be a problemif you're reading this far in the future :D

I would also like to point out, that it would still be really nice to have a proper solution integrated into openscad directly.

naggie:
I tried the `dxf_outline` inkscape extension -- I had to use it independently as the ` --export-extension` flag is apparently no longer available with inkscape 1.0 or something.

This can be accomplished by cloning the inkscape extensions repository and trying something like:

```
python3 ~/extensions/dxf_outlines.py --output=plate.dxf plate.svg
```

However, whilst this produces a DXF (from an OpenSCAD 2021.1 SVG) the DXF seems to be a collection of unconnected lines instead of a nice polyline.

What I ended up doing was a view of my part with `mirror([0,1])` applied such that I could use @RudolfCardinal 's `svg_to_dxf.py` without the resultant vertical-flip encountered by @wheybags .

I just wanted to say thanks, @RudolfCardinal -- I now have a DXF file that I _think_ I can get manufactured! Any idea what's causing the flip? I assume it's just 2 different co-ordinate systems.

PR #450, opened 2013-08-04
Loop extensions and other goodies

devlaam:
Before this fork gets completely out of hand, i think it is good to issue this pull-request. Besides the loop() en loop_extrude() additions, there where some others too, like box() and position(). Furthermore, i extended the center modifier. Documentation can be found in the doc/Loop.pdf. All these additions do no break existing openSCAD scripts. I true hope this add value to the nice work of the people who created openSCAD.


kintel:

PR #599, opened 2014-01-12
cylinder/circle radius modes

brianolson:
adds $rm={INNER_RADIUS,MIDPOINT_RADIUS,OUTER_RADIUS} to cylinder and circle. This adjusts the facets so that the radius is inside the facets, straddling the facets, or outside the facets (the radius lies on the vertexes) (the current current behavior and default behavior).

Adds example025.scad demonstrating this feature, and reference images rendering that as tests.

There may be a glitch in my linux box's OpenGL drivers and the reference images may need to be re-generated to be pixel-perfect.


eadz:
Hi all,

I thought I would give a new user's perspective on this PR. I'm probably a pretty typical new OpenSCAD user: programmer, new to CAD, new to 3d printing, using OpenSCAD for 3d printing.

I printed out my very first object the other day, I designed it myself in OpenSCAD. The process was easy and fun, but the screw hole was too small. 

I had no idea what was wrong, I assumed my design was incorrect. I came across this PR by chance ( browsing through the PRs ), and I would like to offer my opinion. 

Judging from the gallery page, 3d printing is a very common use for OpenSCAD. I don't think that as someone who wants to make parts fit together, I should have to use extra functions or external libraries. 

This feature would be more useful than the center=true/false feature IMHO. 

@MichaelAtOz .. yes I just drilled out my holes too! 

I would say 
 - yes to a global default ( personally I would use middle - accurate volume )
 - yes to a ternary option of ???=INNER|MIDDLE|OUTER

It is a shame that the original maintainer cannot continue. I don't even know what programming language OpenSCAD is written in. I have however updated the wiki to warn users about this issue. 

If there is a 'recommended' workaround, and easy examples to achieve INNER/MIDDLE/OUTER that would be fantastic. Does anybody have any suggestions in this regard?

mrvn:
I had the same problem with hole sizes as everyone else and I would very much like to see this natively in openscad. reading through all the comments I have a few of my own:

1) Having 2 ways to draw a circle (cylinder, sphere) is a bad idea. If you have a circle primitive then fix it. If you have a circle module then remove the primitive. (please do the former).

2) People stumble across this problems because their holes get too small. Holes are made using difference(). That suggests to me that the difference() function should toggle between inner and outer circle. That way positive spaces will use inner and negative space use outer. That also speaks towards a $ variable I think.

3) there are several ways you can want something in between inner and outer: 
    a) midpoint between inner and outer, gives a good approximation of the circle
    b) same circumference, important for pulley systems where you want to advance the string an exact amount
    c) same area, think about making a measuring cup

hairykiwi:
I'm another relatively new user of OpenSCAD, with a reasonable amount of traditional CAD experience.

In much the same way other mature CAD software enable polygon's to be defined as either inscribed or circumscribed, I like @GilesBathgate's suggestion, albeit in the direct opposite form:
`circumscribed = true | false`
\- this for example, is effectively how MoI3D implements it.

The only other rationale for my preference being opposite Giles' is the result of playing with the idea of extending the ternary option to be a percentage variable instead - in this scenario, the relative size of the percentage better suggests what happens to the size of the polygon created:

- `circumscribed = false` would internally translate to inscribed / `0(%)` : the polygon's vertices lay on - or '0% outside' the specified defining diameter,
- `"mid"` or `"middle"` would translate to `50(%)`,
- `circumscribed = true` would translate to `100(%)`: the polygon's vertices lay 'as much as possible' - or 100% outside the defining diameter, and
- any intermediate percentage would provide absolute granularity.

Implemented as such, any of the following would be valid inputs:
`circumscribed = true | false | mid`,
`circumscribed = 0,`,
`circumscribed = 41.93708`,
`circumscribed = 100`.

BTW, special thanks @brianolson for your perseverance with this issue - I really hope something comes of it.

GilesBathgate:
Wow, this was a fun discussion from yesteryear. Thanks for bringing it back up @hairykiwi. I like your suggestion, it certainly seems very clean to simply have a single parameter, and a value between 0%-100% to determine whether its the inner circle outer circle, or midcircle. 

Although rather than using percentages its traditional to have the value normalised between 0.0 and 1.0, with 0.5 being 50%, I think that would make the true and false syntax easier to implement since true=1, and false=0.  There is no simple way to allow the ```circumscribed = mid``` syntax you propose though since mid is not currently a valid token in openscad.

hairykiwi:
Ah, normalised values vs percentages; of course - now you're talking Giles!

The normalised values approach is so self-explanatory it makes the very concept of a 'mid' token appear complete and utter excess baggage.

jakob:
As an OpenScad user, can you please add this feature to the core language? 

I absolutely don't care what you call it. It doesn't matter.

It's not only useful for radial holes, but also for pockets for hex nuts. I don't use OpenSCAD very often, so every time I want to add a hole, I have to get a piece of paper, start drawing, just to come up with the cos() formula again.

Please just make this a part of OpenSCAD.

TakeItAndRun:
Please have a look at this blogpost.:
http://hydraraptor.blogspot.de/2011/02/polyholes.html

Nophead came up with the module polyholes() to size holes correctly.


2016-11-14 10:34 GMT+01:00 Jakob Egger <notifications@github.com>:

GilesBathgate:
@TakeItAndRun I think pockets for hex nuts is the more appropriate use case. However @jakob there is nothing stopping you from creating a library that translates from in-radius to out-radius dimensions, with a given number of sides.

jakob:
Thanks for your suggestions, but as I said, I really wish this feature was built into OpenSCAD.

I came to this issue by looking at the documentation how to make correctly sized holes with cylinder, only to discover that cylinder() doesn't support it. Then I discovered that there was a patch submitted for this functionality almost two years ago, but it's in limbo due to excessive bikeshedding.

Could someone please resolve this? Either accept the patch, or reject it. If you reject it, just close this issue and update the docs to state that cylinder() always makes polygons that are smaller than the true circle.

Also, further up someone said that you need to multiply the radius with cos(360/$fn) to get the correct size holes. This is not correct. That will make the holes even smaller. So just for the record, this is how I make a hole that has a diameter of at least 20 units:

    n = 6;
    cylinder(h=5, d=20/cos(180/n), $fn=n);

Maybe that bit could be included in the docs?

GilesBathgate:
@jakob The correct formula is `a = r * cos( pi / n)` because openscad takes values in degrees not radians this becomes `a = r * cos(180 / n )` The formula above (which were mine) are probably mistakes from translating from c++ code which was using tau instead of pi sorry. I think you should be fine using the formula for diameter in place of radius as you have done. These formula are from here: https://en.wikipedia.org/wiki/Apothem

smurfix:
Can we either close this PR or adapt it to current OpenSCAD?

As to its usefulness, IMHO inner/mid/outer … constants of 1,2,3 … well … I'd rather use 0/0.5/1 as constants that feed directly into the math, that allows me to use 0.9 as "outer circle but needs a snug fit" or whatever.

Erudition:
> As to its usefulness, IMHO inner/mid/outer … constants of 1,2,3 … well … I'd rather use 0/0.5/1 as constants that feed directly into the math, that allows me to use 0.9 as "outer circle but needs a snug fit" or whatever.

That's a great solution! A single value on the unit interval [0..1] where 0 is the smaller, and 1 is the larger. You can default to 0 and people can think of it as a boolean, or use a middle value (such as 0.5) to scale it between the two! Good compromise, super useful.

ioquatix:
This is such an awesome idea can we please have it!? :)

t-paul:
As the long discussion shows, it's unlikely to be implemented in the core language. I think the suggestion would be to have this implemented as user space feature. We can collect those type of features and eventually have some "stdlib" for that. This may need some additional help for maintaining it though.

coryrc:
> As the long discussion shows, it's unlikely to be implemented in the core language. I think the suggestion would be to have this implemented as user space feature. We can collect those type of features and eventually have some "stdlib" for that. This may need some additional help for maintaining it though.

For anybody coming along now, BOSL2 implements it in [cyl()](https://github.com/BelfrySCAD/BOSL2/wiki/shapes3d.scad#functionmodule-cyl) (3d) and [ellipse()](https://github.com/BelfrySCAD/BOSL2/wiki/shapes2d.scad#functionmodule-ellipse) (2d) with the `circum=` argument.

jordanbrown0:
We have MCAD, included with the installation, but it needs a lot of TLC.

coryrc:
> We have MCAD, included with the installation, but it needs a lot of TLC.

[It is not](https://packages.debian.org/sid/amd64/openscad/filelist).

Additionally, IMO it's not great to suggest an untested, undocumented, incoherent, and unmaintained library when BOSL2 exists.

jordanbrown0:
OK, so MCAD is not in the **Debian** distribution of OpenSCAD.  I don't have ready access to a Linux system to check the openscad.org Linux distribution, but I did just check and it's still there in the openscad.org Windows distribution.

I **did** say that it needed a lot of TLC.

BOSL2 is great.  BOSL2 is clearly better.  But it's also not bundled with OpenSCAD, and the authors are not interested in having it bundled.  I was responding to the "stdlib" comment.  (Which, sigh, I didn't realize was being quoted from a much older comment.)  If somebody wanted to come up with  a "stdlib", MCAD wouldn't be a silly place to start.

thehans:
@coryrc It is its own package https://packages.debian.org/sid/amd64/openscad-mcad/filelist

coryrc:
> If somebody wanted to come up with a "stdlib", MCAD wouldn't be a silly place to start.

I guess we'll just have to disagree on this point.

PR #621, opened 2014-01-31
Importcenter

donbright:
I am not sure if you will want this, but I thought I'd submit it as a possible solution for the "how do we import STL that have strange offsets" question that was discussed a few weeks ago by Emmett Lallish and others. 

This will add a 'center=true/false' option to the import() function, just like exists for sphere() and cube(). The imported STL will be 'centered' at the origin. in other words, half of the object will be on one side of the origin, and half on the other... true for all three dimensions x y z. 

It adds a 'translate' function to 'polyset' b/c it seems the easiest fastest way to do the transformation. ( i suppose inserting a transform() node above the import() might be another way, but ... it seemed complicated )


kintel:
Not sure why this was closed. Admittedly it needed some more work, but the center parameter sounded like a decent idea.

agodinhost:
+1 for the center idea into the stl import.

donbright:
i guess i wasnt happy with it? :)

also sometimes i like to tidy up, clean up a little bit.  

Looking at it now, ... i think its more complicated than just getting a bounding box and translating by half.  Thinking back, i'm guessing maybe i wanted to have it translate using CGAL's number type instead of just using Polyset's double.

For example say you have a 1 dimensional CAD program called "LineCAD". The hottest CAD program for all your 1-dimensional CAD needs. Your number type allows numbers between -10 and 10. 11 is out of bounds. If you add 1 to 10 it will either wrap around to -10 or throw an out of range exception or maybe call itself infinity, depending on how your machine deals with numbers. 

Now say you have an object with coordinates -8 and 7. You get a bounding box, and take it's width... it's 15. You can't represent 15 in your number type. It decides the width is infinity. So all of a sudden you have a box that is infinitely wide, and you try to translate it by infinity divided by two to make it move to the origin. Oh what a headache.

PR #659, opened 2014-02-17
allow reading .scad files from standard input

8p0:
This patch loads the root source code form standard input if "-" is
given as file name. In this case handle_dep() is disabled for the root
module but still executed for all included files.


kintel:
The pull request in #1440 looks cleaner. Closing this.

PR #658, opened 2014-02-17
Improve command line help

8p0:
Generate the cmdline() help text programmatically using boost::program_options. This way the help text is never out of sync with the command line options.

Requires C++ (pull request #657)

PR #660, opened 2014-02-17
Allow writing to stdout, refractor openscad.cc on the way

8p0:
These patches refractor openscad.cc. They include pull requests #657, #658 and #659. On top of them they
- allow writing output to standart ouput
- allow to specify either an --action or and -o[utput] to run command line mode
- write to temporary files commit them after all operations succeeded; this prevents garbled output files and is required for proper use in Makefiles
- remove a lot of redundant code


kintel:

PR #674, opened 2014-02-26
Jawsreader bug fix

donbright:
Fix a bug where the menus are unusable for users who have screenreaders because qtaccessibility hasn't been included. Per bug report by Robert Jaquiss to mailing list.

### DIFF
--- a/src/MainWindow.ui
+++ b/src/MainWindow.ui
@@ -335,8 +335,6 @@
     <addaction name="separator"/>
     <addaction name="editActionHide"/>
     <addaction name="viewActionHide"/>
-    <addaction name="viewActionFocusEditor"/>
-    <addaction name="viewActionFocusConsole"/>
    </widget>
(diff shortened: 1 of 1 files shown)


hzulla:
Hey @donbright - over there at Sonic Pi, there is some need for developers who know about screen readers and how to make a Qt application more accessible. Would you guys want to share some of your insights?

t-paul:
@hzulla Qt4 proved to be quite problematic, essentially that went nowhere. Qt5 has totally new support (which is included by default), still I did not have too much success when I had a look some time ago. QScintilla did not seem to have much support for accessibility back then either.

I'd like to have a look at this topic again, as I think it's important, but time is quite limited so that might take a while.

donbright:
@hzulla 

i am sorry but i never solved this problem. QT has some problem way down deep in the library with the way it interacts with the event system on that makes certain uses of Editor widgets not work properly with screenreaders.

Issue #265, opened 2013-01-30
Feature request: allow center= to accept more than a boolean


cbrinker:
I find myself spending a lot of my modeling time doing something akin to this:

  translate([0,0,10/2]) cylinder(r=5, h=10, center=true); 

Which suffers if I need to adjust the height of the cylinder to 20 I have to remember to update the translate. It would be nice to accomplish the same with something akin to:

  cylinder(r=5, h=10, center=[1,1,0])

Does this already exist and I didn't catch it in the docs?

Thanks,
Chris
##

Salamandar:
Up ;)
I was gonna open an issue when I saw this one. This would considerably simplify a LOT of openSCAD files.

ranguli:
@kintel Should we move this to the [ideas for development](https://github.com/openscad/openscad/wiki/Ideas-for-Development-Tasks) section of the wiki if it's under consideration, or close the issue otherwise?

kintel:
@ranguli A writeup on the wiki would indeed be appreciated. Not sure if that warrants closing the issue as the discussion here may be worth keeping, but at least it would anchor the feature request there.

ranguli:
@kintel Perhaps a "Long-term" or "Reviewed" label might help to clean up some of the older issues that have been examined and aren't ready to be closed but aren't currently being worked on?

HWXLR8:
I will pay cold hard cash to see this implemented. Name your price.

jordanbrown0:
See https://github.com/BelfrySCAD/BOSL2/wiki/attachments.scad for a library that does this and much more.

pipatron:
> See https://github.com/BelfrySCAD/BOSL2/wiki/attachments.scad for a library that does this and much more.

Agreed, can't imagine doing anything complex in OpenSCAD without BOSL2. Looking at the speed of which new features are implemented into core OpenSCAD, this won't happen in the language itself.

gsohler:
Yes,  BOSL2 is a great library, but it's based on a limited tool., So BOSL2
is , too
OpenSCAD would be so much better if it provided access to the generated
vertices, But it doesn't.

BOSL2 made it possible to circumvent  this limitation by introducing VNF.
VNF is basically vertex and triangle information of your solid. but you can
only have it when created by BOSL2, you would never get from an
arbitrary OpenSCAD solid.
I dare to say, that my fork is moving way faster. You can easily access the
vertices of any solid, you can have 1D, 2D or 3D objects and you can store
[...]

PR #753, opened 2014-04-13
Add ability to center cubes only on selected axes.

FueCarlos:
For example, you can say:

```
  cube(5, center=[true, true, false]);
```

to center the cube on the X and Y axes, but not the Z axis.

You can still use:

```
  cube(5, center=true);
```

to center on X, Y and Z.

Also works for squares.  In theory cylinders can use the format, but only the Z axis matters in keeping with the original functionality.

Testing done:
Manual tests on cube and cylinder.
New test case in cube-tests.


t-paul:
Closing this now as there's still no final conclusion of what the API should look like (see #265).

PR #793, opened 2014-05-13
Rendering error with concave polyhedron faces 

OskarLinde:
This patch fixes a rendering error with concave polyhedron faces. Concave faces are now triangulated if ENABLE_CGAL.

TODO:
* [ ] Port or rewrite to use new tessellation code


kintel:
We now have a better tessellator, see PolysetUtils and GeometryUtils.
We also need to tessellate before sending stuff to CGAL due to CGAL being extremely picky about planarity, so perhaps tessellating early is a decent option. Needs some more investigation.

kintel:
Fixed by tessellating just before rendering.

PR #804, opened 2014-05-22
Detatchrenderwindow

shaina7837:
I have made a new window named renderWindow and a QAction in View Menu. Whenever we will trigger New Render Window action in View menu, a new QGLView (rendering window) will open separately and get detached from previous MainWindow. And when we will close the new window. The widget will come back to its  MainWindow.


kintel:
This was too hard to get right. If we want this functionality we need to find a better way of solving it.

PR #820, opened 2014-06-03
Speed up minkowski() a bit

OskarLinde:
Will need some testing


michalelektryk:
> 
> 
> This piece of code
> 
> ```
> for (int i = 0; i < 3; i++) {
>     if (a[i] < b[i]) return true;
>     else if (a[i] == b[i]) continue;
>     return false;
> }
> ```
> 
> can be written shorter:
> 
> ```
> for (int i = 0; i < 3; i++) {
>     if (a[i] != b[i]) return a[i] < b[i];
> }
> ```
> 
> Hope this could help for something :-)

it is not an good idea. you're slowing down computation by removing continue statement

Issue #350, opened 2013-05-07
Lazy union (aka. no implicit union)


kintel:
Don't perform a union in the following cases:
- Separate root objects
- `for` loops: A `for` loop will create a list of separate object passed on to the parent node (or the root)
- `children()`. A `children()` statement will pass on the list of children to the parent node

See also https://github.com/openscad/openscad/wiki/OEP2:-Implicit-Unions.

The code lives in https://github.com/openscad/openscad/tree/issue350

This feature is currently experimental. To enabled it, _lazy unions_ must be enabled in Preferences or the cmd-line (`--enable=lazy-union`).

Exporting top-level object to STL/OFF/AMF/DXF/SVG will attempt to keep the objects separate as far as the used file format supports it.

TODO:
- [x] Basic implementation
- [x] Tag as experimental feature (to be able to merge)
- [x] Test csg output (top level, `for()` and `children()`)
- [x] Test `intersection() for()`
- [x] Test `hull() for()`
- [x] Test nested for test
- [x] Test simple top-level objects
- [x] Test `minkowski() for()`
- [x] Test `difference() for()`
- [x] Test `children()`
- [x] STL export
- [x] OFF export
- [x] AMF export
- [x] DXF export
- [x] SVG export
- [x] Deprecate `intersection_for`: To combine this with the experimental feature, we need to defer builtin initialization until after preferences has been read => no need to combine it with experimental for now, deprecate `intersection_for` after we move out of the experimental stage
- [ ] Document the behavior of each file format for some slicer engines
- [ ] Consider the unwrap operator: https://github.com/openscad/openscad/wiki/OEP4%3A-Unwrap-Operator
- [ ] Evaluate compatibility with OpenSCAD2: https://github.com/doug-moen/openscad2/blob/master/rfc/Functions.md#fixing-the-module-composability-problem
- [ ] Test AMF import of separate objects.

This has been discussed a bit, and is interesting for a number of reasons:
- Speed: In many cases, most (half?) of the time and memory used for compiling an OpenSCAD design is consumed by the final union of the top-level objects (see e.g. #261 and #342)
- Necessity: Slicer software (Slic3r in particular, others unknown) supports slicing overlapping objects as it performs an implicit union of all submeshes. 
- Multi-material 3D printing: We could, at the top-level, specify different colors for different top-level objects. When exporting as separate meshes, these colors could be mapped to materials by the slicer software. Combined with amf or obj export, this could make multi-material prints a lot more convenient.
[...]

PR #834, opened 2014-06-15
SVG Import

t-paul:
SVG Importer based on libxml, eigen3 and boost. This is supposed to become a library later, but right now it's simply integrated into the OpenSCAD build system.

The test suite currently has examples from the SVG-Spec, the W3C Test Suite and a collection of images from the OpenClipArt library. The results are published on http://files.openscad.org/svglibtest/ for discussion how specific cases should be handled.

ToDo:
- [x] Drag&Drop should create an import() statement
- [x] Remove the new width/height parameter to import.
- [x] Make experimental
- [ ] Clarify how to handle intersection of multiple imported paths
- [x] Test on Linux
- [x] Test on MacOS
- [ ] Test on Windows
- [x] Add export-import tests
- [ ] Investigate single export-import test failure
- [ ] Test lazy union export (see #350)
- [ ] Look into applying DPI when importing px or unitless designs

Might be considered as well:
- [ ] Viewport clipping -> I don't think this is important (@kintel)


adius:
What is the future of this one?

kintel:
It needs more love.
I could be possible to add it as an experimental feature while we iron out the remaining issues.

adius:
I think experimental feature sounds like a good idea!

kintel:
@t-paul I brushed a little dust off this one, and added an experimental check. I feel it might make sense to merge to master, to get some more external input on this.

t-paul:
Cool, that would unblock the AMF-Import too :-).

kintel:
..and the export-import test now proves that importing exported svg's yield the same result :)

kintel:
Old SVG test files are here: https://github.com/openscad/svg-tests

Neon22:
This is my test file - I use it in the inkscape to openscad extension. It has nested paths and loops.
[test_inkscape_generate.zip](https://github.com/openscad/openscad/files/372601/test_inkscape_generate.zip)
http://www.thingiverse.com/thing:1065500

kintel:
@Neon22 Thanks, that's helpful. This has been merged to master, so it will be available in the next snapshot build. Perhaps you'd like to test it :) Which OS are you using?

Neon22:
windoze 7.

t-paul:
Nightly builds on OBS are updated, Windows builds are next on my list... `libxml2` is included in MXE and I think I tested Windows builds some time ago, so hopefully no build issues.

kintel:
@t-paul Related to Windows, I just noticed this: http://www.appveyor.com/docs/installed-software

-> They have msys2 installed, so it might be possible to bootstrap a Windows build on that service

kintel:
@Neon22 See http://files.openscad.org/snapshots/ ```OpenSCAD-2016.07.20-x86-*```

Neon22:
Trying this on 64bit zip (exe does not show up anywhere).
Using:
```
import("C:/Users/user3/Downloads/OpenSCAD-2016.07.20-x86-64/openscad-2016.07.20/test_inkscape_generate.svg");
```
but getting format not supported error:
```ERROR: Unsupported file format while trying to import file '"../../test_inkscape_generate.svg"' ```

What's the proper command - any options ?

kintel:
@Neon22 It's still experimental, so you have to enable it in Preferences->Features

Neon22:
ahh brilliant thanks....

Neon22:
Tested:
**Primitives:**
- rect single, multiple - OK
  - does pay attention to rounding but ignores the width of the strokes.
  - fail if rect is hollow - see it as a solid.(note below)
- circle single, multiple - OK
  - does not use stroke width, always solid
- arc - Fail
  - a circle is converted to arc if not complete. very common
- text - Fail
- path - single, nested, multiple - OK
  - stroke is ignored unless path is open then it is used for width.
  - so an unfilled path (outline) appears solid.

**Note:**
1. IMHO Generally acceptable for text to fail. Should add prominent note advice to convert text to path before exporting from app.
2. Arc should not fail.
3. Would be good to have flag to ignore thickness of stroke or not... currently ignored unless an open path.
4. if primitive has no fill but has stroke - should be a closed path of stroke thickness - but currently shows solid. I.e. fill/stroke state ignored everywhere except unclosed path.
5. Files with an item that fails to load - seem to not return successful primitives. Should probably convert all they can and fail silently if one item fails. This may be an anomaly - a file that initially failed to load would load after a different file had succeeded. so may be some instability with initial conditions.

Zip attached of files.
[openscad-svg_tests.zip](https://github.com/openscad/openscad/files/377423/openscad-svg_tests.zip)

**Final note.**
In my inkscape exporter I exported all as polygon lists with linear extrude pattern and height primitives for user control.
Use cases that should be handled IMHO:
- how high to extrude the resulting imported svg - needs to be a parameter as not derivable from svg file.
- whether to use stroke width or not
- for closed paths - interpret as though open (see my jigsaw pattern example for why user might want to do that to the surrounding closed path).
- how to hand the result to a module to do more with it - is it a polygon list.
- check my code (link above) to see what I created in openscad

IF we can't separate the many internal primitives into sep objects then user wil have to manually sep and load several files. In which case its absolutely critical that they are not repositioned/rescaled relative to each other.

Neon22:
Tested: (in general working very well except for Arcs and fill/stroke state. Top marks)
**Primitives:**
- rect single, multiple - OK
  - does pay attention to rounding but ignores the width of the strokes.
  - fail if rect is hollow - see it as a solid.(note below)
- circle single, multiple - OK
  - does not use stroke width, always solid
- arc - Fail
  - a circle is converted to arc if not complete. very common
- text - Fail
- path - single, nested, multiple - OK
  - stroke is ignored unless path is open then it is used for width.
  - so an unfilled path (outline) appears solid.

**Note:**
1. IMHO Generally acceptable for text to fail. Should add prominent note advice to convert text to path before exporting from app.
2. Arc should not fail.
3. Would be good to have flag to ignore thickness of stroke or not... currently ignored unless an open path.
4. if primitive has no fill but has stroke - should be a closed path of stroke thickness - but currently shows solid. I.e. fill/stroke state ignored everywhere except unclosed path.
5. Files with an item that fails to load - seem to not return successful primitives. Should probably convert all they can and fail silently if one item fails. This may be an anomaly - a file that initially failed to load would load after a different file had succeeded. so may be some instability with initial conditions.

Zip attached of files.
[openscad-svg_tests.zip](https://github.com/openscad/openscad/files/377423/openscad-svg_tests.zip)

**Final note.**
In my inkscape exporter I exported all as polygon lists with linear extrude pattern and height primitives for user control.
Use cases that should be handled IMHO:
- how high to extrude the resulting imported svg - needs to be a parameter as not derivable from svg file.
- whether to use stroke width or not
- for closed paths - interpret as though open (see my jigsaw pattern example for why user might want to do that to the surrounding closed path).
- how to hand the result to a module to do more with it - is it a polygon list.
- check my code (link above) to see what I created in openscad

IF we can't separate the many internal primitives into sep objects then user wil have to manually sep and load several files. In which case its absolutely critical that they are not repositioned/rescaled relative to each other.

kintel:
@Neon22 Thanks, that's very helpful. There are a number of things we need to sort out. Let's continue this discussion in #1740.

kintel:
@Neon22 Btw., what was wrong with the arc? It looks correct to me (renders as a 'pacman' in OpenSCAD)

Neon22:
I saw nothing at all. could be this initial condition I was referring to. On initial start it would not load the ..generate.svg nor the arc test but I had loaded rects and circles before trying to load the arc. but the arc failed. I saw nothing.
later I coudl load in 05-complex which is same as generate

PR #836, opened 2014-06-15
AMF Import

t-paul:
This is based on the SVG Import (mainly as it's also using libxml2 for parsing), so it should be only considered after #834 is merged.

The import supports reading multiple objects which are currently joined into a single object (that might be changed later when #350 is available).

No support for curved triangles.

ToDo:
- [x] Wait for #834 to be merged
- [x] Verify how to handle multi-volume objects
- [x] Add support for reading zipped files
- [x] Enable libzip for macOS build
- [x] Test lazy union export (see #350) -> defer to later
- [x] Add "export-import" tests, similar to how we handle other bidirectional formats, e.g. ```add_cmdline_test(offpngtest EXE ${PYTHON_EXECUTABLE} SCRIPT ${CMAKE_SOURCE_DIR}/export_import_pngtest.py ARGS --openscad=${OPENSCAD_BINPATH} --format=OFF --render EXPECTEDDIR monotonepngtest SUFFIX png FILES ${EXPORT3D_TEST_FILES})```

Note: reading zipped files uses libzip (http://www.nih.at/libzip/) which seems to be available on all platforms. In addition the support is optional, if ```ENABLE_LIBZIP``` is not defined, then only plain text AMF files can be imported.


t-paul:
All of the `offpngtest` and `offcgaltest` cases are disabled by the Bug tag. So what to add for AMF then? We are currently not able to export non-manifold AMF files, so none of those test would have any chance to work. Any ideas what test would make sense then?

kintel:
I need to look more closely at the state of  CMakeLists.txt.
The original purpose of the export-import tests was to test that exported files can be re-imported yielding the same result. Looks like we accidentally disabled some of those tests.
As a start, we really only need to test a naive example, then start probing for larger parts of the feature set we use from each file format when exporting.

t-paul:
Yep, I guess it would be ideal to have a list of base-line tests for the import/export (without special issues like non-manifold and so on) and add special stuff as needed. That will help for the 3MF import/export too.

kintel:
-> #1822

donbright:
i dont recall that export/import ever worked properly

kintel:
@donbright We've been using the export/import for a while. Seems to work, but there are fun corner cases to deal with

kintel:
I believe this is now good to go from my side

t-paul:
Ok from my side too.

Issue #844, opened 2014-06-23
Simply extending "echo" to write text files, allows exporting BOM/configuration values


CarlosGS:
I think "echo" needs to be able to write values to text files. This would be very useful in order to automatically generate BOMs and debug measurements and information to files.

But the implication that excites me more is the fact that it would open the possibility to write other ".scad" files:
After reading all the discussion involving "defining global variables from local scopes", "retrieving values from modules", etc, I think adding the capability to write text files could provide a good solution to situations that require such tasks. The point is that using "echo" we could write configuration ".scad" files in order to update variables in the main code automatically, thus being able to incorporate results from the inside of sub-tree calls.
It would not be the cleanest solution, but probably one of the simplest ways to implement this kind of functionality into OpenScad with minimal effort.

Example:
echo("openscad rocks",tofile="test.txt",append=false) // append determines whether or not the file is overwritten or the text appended to the end. tofile=undef means the default behavior for echo
The minor backward incompatibility would be giving a value to the "tofile" argument unknowingly.

I am no expert but may give a try and implement this, as it is quite simple. But I want to know your thoughts first in order to get the best solution possible and promote standardization.

Thanks,
Carlos

PR #845, opened 2014-06-23
Added file writing functionality to echo

CarlosGS:
This implements file writing functionality to the echo function. Example:

echo2file_demo.scad

```
echo("test");
echo(tofile="echo2file_demo_autogenerated.txt","test1\n",append=false);
echo(tofile="echo2file_demo_autogenerated.txt","test2\n");
echo(tofile="echo2file_demo_autogenerated.txt","\n");

a = 42;
echo(tofile="echo2file_demo_autogenerated.txt",str("a=",42,";"));
```

Console output

```
ECHO: "test"
WRITING TO FILE echo2file_demo_autogenerated.txt
WRITING: test1

APPENDING TO FILE echo2file_demo_autogenerated.txt
WRITING: test2

APPENDING TO FILE echo2file_demo_autogenerated.txt
WRITING: 

APPENDING TO FILE echo2file_demo_autogenerated.txt
WRITING: a=42;
```

echo2file_demo_autogenerated.txt

```
test1
test2

a=42;
```

Original message ( https://github.com/openscad/openscad/issues/844 )
I think "echo" needs to be able to write values to text files. This would be very useful in order to automatically generate BOMs and debug measurements and information to files.

But the implication that excites me more is the fact that it would open the possibility to write other ".scad" files:
After reading all the discussion involving "defining global variables from local scopes", "retrieving values from modules", etc, I think adding the capability to write text files could provide a good solution to situations that require such tasks. The point is that using "echo" we could write configuration ".scad" files in order to update variables in the main code automatically, thus being able to incorporate results from the inside of sub-tree calls.
It would not be the cleanest solution, but probably one of the simplest ways to implement this kind of functionality into OpenScad with minimal effort.


jgrjgr:
is there any movement on this one? even if the output is random in nature it would still be incredibly usefull to have

t-paul:
I think it's still not clear how to make this a stable feature, generating files somewhere in the processing is colliding with the descriptive nature of scad scripts, even more so than the existing echo(). It's going to create lots of confusion as it looks like it's imperative programming, but in fact it isn't.

Especially for BOM generation or meta data in general (e.g. slicer hints for 3d printing), I think a descriptive solution will be a much better fit.

jgrjgr:
ok, so is this a feature that might find its way into the mainstream version at some point in the future when the nitty gritty bits are worked out? ,  I'm just trying to get a idea on weather its worth investing a bit of time into modifying my slic3r post processor and releasing that so we can generate BOMs from scad files when we generate objects.

nophead:
I make structured BOMs for Mendel90 using echo and a Python script
<https://github.com/nophead/Mendel90/blob/master/bom.py>. It just runs
OpenScad from the command line to capture the ECHO text. It didn't need any
changes to OpenScad.

jgrjgr:
interesting script,  i'll probably modify the python script you have there or write my own in something that compiles to a standalone executable

runsun:
If you are using Linux, I've got a bash script that might help:

openscad <scad_file> -o dump.stl -D "ISDEMO=true" 2> <output_file>; (find
<output_file> -type f -print0 | xargs -0 sed -i 's/ECHO: "//g'); (find
<output_file> -type f -print0 | xargs -0 sed -i 's/"$//g')

It has 3 parts, joined with ";" :

(1) Run <scad_file> and save to <output_file>
(2) Remove "ECHO: "
(3) Remove trailing "

It can be modified easily and saved in a sh file (say, ops_dump.sh)and run
like:

  ./ops_dump.sh <scad_file> <output_file>





On Fri, Apr 24, 2015 at 1:31 AM, jgrjgr <notifications@github.com> wrote:

> interesting script, i'll probably modify the python script you have there
> or write my own in something that compiles to a standalone executable
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/openscad/openscad/pull/845#issuecomment-95817979>.
>



-- 
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Runsun Pan, PhD
[Fax] 832-534-3213
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

runsun:
Just wrote up a short python to dump echo output to a file. Works pretty
well on my linux.

############################## runscad.py
import sys, os, re

if len(sys.argv)<3:
   print '''
# runscad.py
#
# Not enough arguments. Required 3 arguments:
#
#    python runscad.py <test_file.scad> <command(s)> <output_html_file>
#
# For example:
#
#    python runscad.py myprog.scad "FORHTML=true;MODE=10;run()" tests.html
'''
else:

   cmd = 'openscad %s -o dump.stl -D "%s" 2> %s'% tuple(sys.argv[1:])
   os.system( cmd )

   lines = open( html, 'r').readlines()

   # Remove the starting 'ECHO: "' and trailing " on each line
   ptn = re.compile(r'ECHO: \"|"$')
   for i,line in enumerate(lines):
        lines[i] = re.sub( ptn,'', line)
        print lines[i]

   # Save to file. Exclude the last line, which reads:
   # "Current top level object is empty."
   open( html, 'w').write( '\n'.join( lines[:-1] ) )
##############################


On Tue, Apr 28, 2015 at 2:25 PM, Runsun Pan <runsun@gmail.com> wrote:

> If you are using Linux, I've got a bash script that might help:
>
> openscad <scad_file> -o dump.stl -D "ISDEMO=true" 2> <output_file>; (find
> <output_file> -type f -print0 | xargs -0 sed -i 's/ECHO: "//g'); (find
> <output_file> -type f -print0 | xargs -0 sed -i 's/"$//g')
>
> It has 3 parts, joined with ";" :
>
> (1) Run <scad_file> and save to <output_file>
> (2) Remove "ECHO: "
> (3) Remove trailing "
>
> It can be modified easily and saved in a sh file (say, ops_dump.sh)and run
> like:
>
>   ./ops_dump.sh <scad_file> <output_file>
>
>
>
>
>
> On Fri, Apr 24, 2015 at 1:31 AM, jgrjgr <notifications@github.com> wrote:
>
>> interesting script, i'll probably modify the python script you have there
>> or write my own in something that compiles to a standalone executable
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/openscad/openscad/pull/845#issuecomment-95817979>.
>>
>
>
>
> --
> ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> Runsun Pan, PhD
> [Fax] 832-534-3213
> ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
>



-- 
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Runsun Pan, PhD
[Fax] 832-534-3213
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

jgrjgr:
@nophead is your python script working with the latest version of openscad ?  for me on windows the latest version doesn't output anything on the command-line so even piping the output to a file won't work anymore

nophead:
I have never tried it with anything later than 2013.06 because later versions don't give the same view when rendering and don't generate the same STLs.

Having said that OpenScad should still work from the command line. Are you sure you are running openscad.com, not openscad.exe?

jgrjgr:
i think i worked out half the problem, my batch files were trying to run;

 "C:\Program Files (x86)\OpenSCAD\openscad.com"

which just produces and bad command or file name error,  i took the "(x86)" out as it's still a valid path as i had two copies of openscad installed on this machine

i took both copies of openscad off my machine and reinstalled the latest version to 

"C:\Program Files\OpenSCAD\" 

now the com file and exe file work and produce the echo output but none of the debug info that it used to, it still spits out warnings though like the manifold error

kintel:
@jgrjgr Looks like you have two issues
1) problems with spaces or other characters in the path
2) debug output

Specifically, what type of debug output are you missing? OpenSCAD shouldn't output debug output for release builds, but some debug messages might have snuck into some earlier builds.

t-paul:
Closing, there's still to many open questions and issues to add even more side-effects to `echo()`. Some of the use cases mentioned are even problematic as writing files on evaluation that are fed back into the source would cause loops with automatic design reload. For most if not all cases there should be other options.

PR #888, opened 2014-07-30
Ctrl+Alt+A to toggle animation mode

droidmunkey:
Quickly Toggle Animation


kintel:

PR #890, opened 2014-07-30
Added Menu Option to Reverse Animation, step forward, step back

droidmunkey:
Added Menu Option to Reverse Animation
Added Menu option + shortcut keys to step forward and backwards through animation


kintel:

Issue #937, opened 2014-09-05
Get transformation matrix of given place


Gregwar:
Hello,
First, thanks to all dev/contributors for this amazing and powerful tool!
We'd like to have a way to know the transformation matrix at certain points of the scad
For instance:

``` scad
rotate([45,0,0]) {
    translate([10,5,0]) {
        rotate([20,0,-40]) {
            // What transform matrix is used here?
            cube(20);
        }
    }
}
```

The goal is to make a tool to build robot joints separately in SCAD, and then have a way to automatically know the 3D model of it programmatically

A suggestion would b having a special command to do that, like `show_matrix()` or so, that would output the transform matrix to the text output

I can do a PR for that
##

clothbot:
@kintel - How easy would it be to implement a 'c.set_variable("$matrix",...)' around about here (next to $children):

https://github.com/openscad/openscad/blob/master/src/module.cc#L194

...that calculates the transformation matrix to the module-local origin by traversing/calculating back down the module_stack?

I'm calling it "$matrix" because "TransformModule::instantiate" in transform.cc assigns to node->matrix.

Andrew.

kintel:
I don't think that would work, as that would make the module instance dependent on its parent.
$children is only dependent on the arguments.

clothbot:
Thinking out loud, could it instead work by "pushing" $matrix forward, into children via TransformModule::instantiate:

https://github.com/openscad/openscad/blob/master/src/transform.cc#L56

kintel:
Pushing would work. We'd have to explicitly push it though, or at least find a way of determining which module instances actually use the matrix. Otherwise, the entire caching scheme would break.

clothbot:
Somewhere round about here:

https://github.com/openscad/openscad/blob/master/src/transform.cc#L180

...is where I'd try iterating over instantiatednodes and assign "this->$matrix * matrix" to each instantiatednode's 'c.set_variable' list.

Devil's in the details though. ;-)
.

clothbot:
...or maybe that should be "matrix * this->$matrix", since you want to do things like map global-origin up to local coordinate space.

clothbot:
Now that I've had a chance to poke at it, seems like some not-insignificant rearchitecting would be required to get this to work.

[...]

PR #952, opened 2014-09-15
Adding marker() nodes

kintel:
From @Gregwar (original pull request #939)

This introduces marker() nodes.

Markers are preserved during the CSG rendering, the contents is evaluated just like echo so you can pass it variables etc.

The advantage is that you can "mark" some parts of your code, and then access your markers in the CSG, simple to parse, to get for instance the locate coordinate system at a certain point from the origin of the part

See #937

TODO:
- [x] Make marker() experimental
- [ ] Add tests

To move out of experimental, we need some discussion on this feature, as well as documentation, but we can handle that in a separate pull request

### DIFF
--- a/openscad.pro
+++ b/openscad.pro
@@ -262,6 +262,7 @@ HEADERS += src/typedefs.h \
            src/transformnode.h \
            src/colornode.h \
            src/rendernode.h \
+           src/markernode.h \
            src/textnode.h \
            src/openscad.h \
            src/handle_dep.h \
@@ -344,6 +345,7 @@ SOURCES += src/version_check.cc \
            src/surface.cc \
            src/control.cc \
            src/render.cc \
+           src/marker.cc \
            src/text.cc \
            src/dxfdata.cc \
            src/dxfdim.cc \
--- a/src/control.cc
+++ b/src/control.cc
@@ -31,6 +31,7 @@
 #include "modcontext.h"
 #include "builtin.h"
 #include "printutils.h"
+#include "markernode.h"
 #include <sstream>
 #include "mathc99.h"
 
@@ -45,6 +46,7 @@ public: // types
 		CHILD,
 		CHILDREN,
 		ECHO,
+        MARKER,
 		ASSIGN,
 		FOR,
 		INT_FOR,
@@ -257,10 +259,12 @@ AbstractNode *ControlModule::instantiate(const Context* /*ctx*/, const ModuleIns
 	else
 		node = new AbstractNode(inst);
 
-	if (type == ECHO)
+	if (type == ECHO || type == MARKER)
 	{
 		std::stringstream msg;
-		msg << "ECHO: ";
+        if (type == ECHO) {
+    		msg << "ECHO: ";
+        }
 		for (size_t i = 0; i < inst->arguments.size(); i++) {
 			if (i > 0) msg << ", ";
 			if (!evalctx->getArgName(i).empty()) msg << evalctx->getArgName(i) << " = ";
@@ -271,7 +275,13 @@ AbstractNode *ControlModule::instantiate(const Context* /*ctx*/, const ModuleIns
 				msg << val.toString();
 			}
 		}
-		PRINTB("%s", msg.str());
+        if (type == ECHO) {
+    		PRINTB("%s", msg.str());
+        } else {
+            MarkerNode *markernode = new MarkerNode(inst);
+            markernode->value = msg.str();
+            node = markernode;
+        }
 	}
 
 	if (type == ASSIGN)
@@ -311,6 +321,7 @@ void register_builtin_control()
 	Builtins::init("child", new ControlModule(ControlModule::CHILD));
 	Builtins::init("children", new ControlModule(ControlModule::CHILDREN));
 	Builtins::init("echo", new ControlModule(ControlModule::ECHO));
+	Builtins::init("marker", new ControlModule(ControlModule::MARKER));
 	Builtins::init("assign", new ControlModule(ControlModule::ASSIGN));
 	Builtins::init("for", new ControlModule(ControlModule::FOR));
 	Builtins::init("intersection_for", new ControlModule(ControlModule::INT_FOR));
--- a/src/control.cc
+++ b/src/control.cc
@@ -55,7 +55,9 @@ public: // types
 public: // methods
 	ControlModule(Type type)
 		: type(type)
-	{ }
+	{
+		if (type == MARKER) this->feature = &Feature::ExperimentalMarkerModule;
+	}
 
 	virtual AbstractNode *instantiate(const Context *ctx, const ModuleInstantiation *inst, const EvalContext *evalctx) const;
 
--- a/src/feature.cc
+++ b/src/feature.cc
@@ -19,6 +19,7 @@ Feature::list_t Feature::feature_list;
  * context.
  */
 const Feature Feature::ExperimentalTextModule("text", "Enable the <code>text()</code> module.");
+const Feature Feature::ExperimentalMarkerModule("marker", "Enable the <code>marker()</code> module.");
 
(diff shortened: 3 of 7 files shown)


Gregwar:
Any news on this?

kintel:
We're currently finalizing a release. New experimental functionality will be managed post-release.
Are you still using this, in production?
Did you connect with @laird to see how they could use the same / similar functionality?

Gregwar:
I am using it in development, with my own build
I could fix the PR

kintel:
@Gregwar Updating the pull request with something kept in sync with master would indeed be helpful.

Gregwar:
I've updated our fork to be in sync with your master:
https://github.com/openscad/openscad/compare/master...RhobanProject:markers?expand=1
(Sorry for the space/tab glitch)

Gregwar:
Hello
Any news on this ?

Gregwar:
Still no news on this
Ping @kintel ?

Gregwar:
(I can give you a patch without conflicts)

doug-moen:
Sorry I'm late to the discussion. I just want to mention how CSG tree metadata works in OpenSCAD2. No new syntax is required. Instead, the CSG tree is extended to include variable definitions: top level definitions, and those within brace brackets. This allows you to attach arbitrary name-value metadata pairs to any node.

For example, consider this script:

sz=10;
cube(sz);
translate([sz,sz,0]) {
    label="translated cube";
    cube(sz);
}

In the latest release version, the CSG tree looks like this:

group() {
	cube(size = [10, 10, 10], center = false);
	multmatrix([[1, 0, 0, 10], [0, 1, 0, 10], [0, 0, 1, 0], [0, 0, 0, 1]]) {
		cube(size = [10, 10, 10], center = false);
	}
}

In OpenSCAD2, we extend the CSG tree by including variable definitions. Top level definitions are placed in the top level group, while variable definitions inside { } brackets are placed in interior nodes. But assign() and let() nodes are not included in the CSG tree. It would look something like this:

group() {
	sz = 10;
	cube(size = [10, 10, 10], center = false);
	multmatrix([[1, 0, 0, 10], [0, 1, 0, 10], [0, 0, 1, 0], [0, 0, 0, 1]]) {
		label = "translated cube";
		cube(size = [10, 10, 10], center = false);
	}
}

Gregwar:
@doug-moen this is not the case on `master`, is it?

What would become:
```
abc = 12+34;
```

And:
```
abc = 12;
abc = 34;
```
?

doug-moen:
No, this has been proposed but not implemented. Marius has requested that
other feature requests be considered in the context of OpenSCAD2, which is
why I'm supplying this information.

The CSG tree will contain the final value of variables after evaluation is
complete. The purpose is to add metadata to the tree. The definition a=2+2
would add the metadata a=4 to the tree, just as marker(2+2) would add
marker (4) to the tree.

Gregwar:
Ok, but what would add:

```
meta="info 1"
meta="info 2"
```

?

kintel:
It takes time to properly implement such functionality, as we also need to maintain this in the future. OpenSCAD is in need of a major language overhaul, something @doug-moen has spent significant time thinking about. The missing link is time to actually implement it, so in the mean time we try to make sure that new features are in-line with the current OpenSCAD2 proposal, or spawns productive discussions around the OpenSCAD2 proposal so it can be adjusted to fit real-world use-cases better.

Marker nodes is an example of model meta-data which can be used in a subsequent export operation to be made accessible by external software. 
There was also a long discussion on the mailing list on a related topic, an export() function: http://forum.openscad.org/export-function-tt15428.html

..to answer your question: In OpenSCAD2 it would be disallowed to override a variable.

Gregwar:
@kintel ok, in this case the "variable" way is maybe not the best way. I personnaly think this is odd compared with having a special node. (Maybe `marker()` should be `meta()`)

doug-moen:
If we added this feature to OpenSCAD right now without any other OpenSCAD2
related changes, then the result would be meta="info 2".

It's whatever the final value of 'meta' is after the script is evaluated.
You can get the same answer by running this script:

meta="info 1";
meta="info 2";
echo(meta);

This will ECHO: "info 2"

Another way to answer your question is that within a node, each metadata
tag is unique.

Gregwar:
Ok, this is arguable

Gregwar:
In the meantime, I am already using marker nodes from my own custom OpenSCAD build in a project, here is a screencast of the current version of the editor:

https://www.youtube.com/watch?v=GItIgJM7jrE

It automatically parses SCAD files to generate the wizard you can see. There is comments on these files that explain what it contains and markers to describe where are the anchor points to connect other parts (the blue cylinders you can see I'm clicking on)

The "generate" button runs OpenSCAD (which is used as backend) to re-compile files with new parameters. There is a cache system to keep all the already generated files and be faster.

Gregwar:
(This is a typical example of use case for metadata in the CSG file)

BTW, I am parsing the CSG with custom code, maybe we could imagine having a library that allow an easy crawl of metadatas with associated transform matrixes :-)

kintel:
@Gregwar I'd like to extract the parser+AST of OpenSCAD (or at least .csg files) into a separate library.

doug-moen:
Thanks for the video link, Grégoire. I also looked at the github project
and the metabot.cc web site, and this is awesome!

Gregwar:
@doug-moen thanks, I'm working on the next version of Metabot currently, the "Studio" (and physics simulation) are works for my thesis, but it could be a good thing to imagine allowing people designing robots from provided generic & parametric parts

We could imagine that one day another tool will exist, compliant with OpenSCAD to deal with assemblies

nophead:
Why does it need to be another tool? OpenScad can do complex assemblies
with hundreds of parts.

doug-moen:
@nophead: watch Gregwar's video to see what kind of GUI he has in mind for working with assemblies. https://www.youtube.com/watch?v=GItIgJM7jrE

nophead:
Yes its a great tool for people that want to configure modular robots with
a GUI, something quite specialist. The whole point of OpenScad is you can
do everything in code, including making assemblies, and it's completely
general.

Gregwar:
OK, And what about that:
https://github.com/openscad/openscad/wiki/Meta-Data-Use-Cases

This is just a RFC, right?
What is the roadmap for such features?

Gregwar:
Actually, what if every nodes that would begin with "@" would be kept in the CSG, and behaves exactly as my marker() nodes currently does?

The only thing that I could dream of then would be a "json" output that explain the annotations for the meta use case.

Let's explain an example:

```
@Module(name="something", description="something cool")
module something() {
    cube([10,10,10]);
    translate([10,0,0]) {
     @Point(name="An important point");
    }
}
```

Then, I could either compile it with:

    openscad something.scad -o something.json

Wich would give me something like (obj is the concerned object):

```
[
{name:"Module", args: {description: "something cool"}, obj: {type: "module", name: "something"}},
{name:"Point", args: {name: "An important point"}, obj: null}
]
```

And also compile the csg:

    openscad something.scad -o something.csg

```
group() {
        @Module(description="something cool")
        cube(size = [10, 10, 10], center = false);
        multmatrix([[1, 0, 0, 10], [0, 1, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]) {
            @Point(name="An important point");
        }
}
```

kintel:
@Gregwar I'm trying to pull together the various use-cases similar to this into a common location. There are several overlapping ideas and proposals which could solve both this, the meta-data as well as the "multiple export" feature wishes. Question is if it makes sense to cover all of that with one mechanism. I'll likely open a new issue for tracking the larger issue.

Gregwar:
Ok, would be indeed cool to have a common issue that provide an overview for this

kintel:
This never materialized, closing.

PR #1048, opened 2014-12-04
Headless OpenSCAD / CMake

t-paul:
This pull request is only for tracking and discussing changes to make OpenSCAD easier to use in headless environments (similar to the openscad_nogui binary that is used for the test suite).

Update (@kintel): I did a bit of a refactor of this, trying to unify how we search for libraries and generally getting the GUI build to work as well.


kintel:
@t-paul I know this is old, but we could commit this to a `cmake` folder and gradually tinker with it in the master branch. It shouldn't interfere with anything else. That would also make it easier to gradually refactor some of the tests build system.

kintel:
Merged this so we have a testable starting point. Will open a new issue to track improvements

PR #1050, opened 2014-12-04
Add fecho function.

t-paul:
Replacement for #754 which was auto-closed when removing the obsolete "unstable" branch.


t-paul:
Implemented via #1830.

PR #1094, opened 2014-12-21
Use QtConcurrentRun to initialize FontConfig in a separate thread.

t-paul:


kintel:
See #1194 - whenever the font cache is initialized from a Render operation (F6), this causes us to bring up the dialog from another thread. Since Qt GUI isn't threadsafe, this causes the crash.

PR #1110, opened 2014-12-26
Add trace() node to convert images to vector graphic using potrace.

t-paul:
Prototype for a `trace()` module that can import images (PNG) and convert them to 2D vector graphics.

Library used is potrace: http://potrace.sourceforge.net/

Implementation is prepared to handle building with a missing library. In that case a dummy geometry is generated (just a border with the size of the bitmap) and a warning is shown in the console.

TODO:
- [ ] Decide how the API should look like, currently it's `trace(file = "img.png", threshold = 0.5);`
- [ ] Better `$fn` handling, currently the default creates too much vertices
- [x] Make experimental
- [x] Add potrace to OS X dependency build
- [x] Test OS X snapshot build
- [x] Build on Windows: Probably means adding an MXE package for potrace
- [ ] More tests
- [ ] Test against trace2scad: http://aggregate.org/MAKE/TRACE2SCAD/
- [ ] Support more image formats: As this also affects surface() it should be a separate issue
- [ ] Support caching. Basic caching done, but looks like some stuff still isn't cached.

openscad-potrace-1


atartanian:
this is an awesome idea. It seems like your current way of handling the trace operation is pretty good, but might be worth including some of the other arguments that potrace allows:

 -z, --turnpolicy policy    - how to resolve ambiguities in path decomposition
 -t, --turdsize n           - suppress speckles of up to this size (default 2)
 -a, --alphamax n           - corner threshold parameter (default 1)
 -n, --longcurve            - turn off curve optimization
 -O, --opttolerance n       - curve optimization tolerance (default 0.2)

atartanian:
turdsize (though i wouldn't call the argument that :stuck_out_tongue_closed_eyes: ) would be really useful in your example above with the parrot

also, --longcurve and --opttolerance also seem like they would be useful parameters to expose

MichaelAtOz:
> -O, --opttolerance n - curve optimization tolerance (default 0.2)

Please, no case sensitive syntax...

kintel:
@atartanian Do you have any examples of practical potrace usage which utilize the types of arguments you suggested?

atartanian:
@kintel I haven't used potrace extensively, but from reading the documentation and working with other bitmap to vector tools, I believe some of those options will be useful to have access to.

the most important would be:
-t, --turdsize n - suppress speckles of up to this size (default 2)
-O, --opttolerance n - curve optimization tolerance (default 0.2)

turdsize allows the user to reject very small paths which are common if you are coming from a photograph or any other image that has a lot of noise. the example above with the parrot would be helped a lot by being able to reject the very small holes on the head, back, and tail.

opttolerance allow the user to make a trade-off between complexity and precision. this is usually a very important value to be able to tweak when trying to get a really good trace of a bitmap in my experience. this comes into play with images that are noisy or with shapes that have a lot of fine detail. In noisy images, you would want to turn the opttolerance up to smooth out the curve so that you don't get lots of stairstepping. With an image with fine detail, you would want to turn it down to avoid smoothing too much of the detail away.

mast76:
I hope this feature is not forgotten. ^up vote^

t-paul:
Nope, not forgotten but on hold due to lack of time.

PR #1116, opened 2014-12-27
Add optional scale markers to large axes

TDeagan:
While add-on rulers can be extremely useful, there are many times that a quick scaling marker would be handy. This feature generates tics every 10 units, driven off the log10 of the l variable defining axes length.  As you zoom in or out, the tics automatically re-scale themselves.  Every tenth tic is slightly larger.  A menu item is added to the view menu to enable or disable the feature.

I experimented with menu driven scaling (.01, .1, 1, 10, etc.), but using small increments when zoomed out brought the app to it's knees.  Adding a visual indicator of the current scaling might be nice, but I'm not sure where to put it (possibly below the small axes,) since it changes while zooming and would clutter the console.

Additionally, a toolbar icon could be added.

image

Zoom sequence:
image
image
image
image


t-paul:
This is superseded by #1131, so we can close it, right?

TDeagan:
> This is superseded by #1131, so we can close it, right?

That's correct, I probably should've closed this one but I wasn't sure of the protocol folks have worked out (sorry!)  How does it work best for this project, originator closes dead/superseded or the admins do it?

kintel:
It tends to be best that the originator does it, as you'd know best where it's at.

PR #1131, opened 2015-01-02
Scalemarkers numbered

TDeagan:
This pull request represents a set of changes to add numbered scale markers as discussed at https://github.com/openscad/openscad/pull/1116

The scale marking is user select-able using a menu item in the View menu.

The markers change proportionally based on the length of the axes using a log function to provide units in powers of ten at all zoom levels.

The markers currently in code provide a single arm to one side of the axis, additional arms are present but commented in the code.

The numbers are generated by GL_LINES, GL_LINE_STRIPs or GL_LINE_LOOPs as appropriate.  Their sizing remains consistent regardless of scale. 

screen shots:
image
image

Here are some downsides (though not necessarily issues, just artifacts):

when the model is flush to the axis, it gets harder to read:
image

When you aren't in the 'upright' orientation, the characters (like any .scad based text) are reversed (seen from below):
image

The tics are reliant on the axes' length. So when they give out at some zoom level, so do the scale markings:
image


kintel:
I already merged this, but found a small bug:
If your design is not centered in the origin and you zoom into it, the scale markers eventually disappear since you assume that the center of focus is the origin.

TDeagan:
> I already merged this, but found a small bug:
> If your design is not centered in the origin and you zoom into it, the scale markers eventually disappear since you assume that the center of focus is the origin.

Frak!  I thought sure I tested that.  I must have gotten complacent thinking I was mimicking the GLView::showAxes setup.  Basing my scaling on ```cam.projection == Camera::PERSPECTIVE ? cam.viewer_distance : cam.height;``` is the problem.

Sigh, there goes my fantasy of faking an understanding of how the camera stuff programmatically works.  Sorry about that!!!  I'l dig in and get a sufficient understanding.

Issue #1039, opened 2014-11-30
Set mousecursor when rotating or dragging in viewport


Scalpel78:
When the mouse is used to rotate or drag the view, the cursor should be changed to display a suitable cursortype.
##

Scalpel78:
I like the rotate icon you've made :+1:

PR #1132, opened 2015-01-02
Change cursor

TDeagan:
This pull request is related to GUI feature request 'Set mousecursor when rotating or dragging in viewport' #1039.

A cross cursor is provided as via QApplication::setOverrideCursor for translation and a rotate icon for rotation in the MouseMoveEvent

A public domain rotate icon is added.  Tested on Windows 8.1 and Debian Jessie (I don't have access to OS X at the moment.)

The setOverrideCursor pushes the cursor state onto a stack and pops the state at release.  Testing demonstrated edge cases where the user could go form rotate to translate (or vice versa) directly and push two changes onto the stack.  This required counting the number of pushes and executing the appropriate number of pops in the MouseReleaseEvent.


kintel:
It works :)
@TDeagan Did you stumble across a zoom icon while at it?

TDeagan:
Should be easy enough to make white with black trim versions of the ones use on the toolbar (plus and minus in magnifying glass).  I'll whip some up and figure out where the hooks are.

TDeagan:
So I can find the place in QGLView.cc that responds to the shift-right and shift-middle zooms:
```
 // Right button pans in the xz plane
      // Middle button pans in the xy plane
      // Shift-right and Shift-middle zooms
      if ((QApplication::keyboardModifiers() & Qt::ShiftModifier) != 0) {
        cam.viewer_distance += (GLdouble)dy;
```

But I'm finding, on at least the computer I'm on, that I can initiate zoom with a touchpad gesture (which creates it's own special cursor) and with the middle button.  I read above that the middle button should pan, not zoom, so that's a little weird (I checked my mouse config and middle is mapped to middle.)

I'm not sure if this behavior is different on different boxen (I'll try some other boxes at work tomorrow, but I'm pretty sure the one I'm on here at home is the only three button system I have.)  I also don't have any actual mice, but I'm willing to guess that a mousewheel would also zoom.

Does the middle button pan for everyone else?  Reading through the QMouse docs (http://qt-project.org/doc/qt-4.8/qmouseevent.html) it appears that the mousewheel is a different class, the QWheelEvent Class
(http://qt-project.org/doc/qt-4.8/qwheelevent.html) and the touchpad is yet another, the QTouchEvent Class (http://qt-project.org/doc/qt-4.8/qtouchevent.html) so it appears feasible to add them.  Hpwever, I'm not sure if doing so merits a larger discussion about their potential and use for possibly more than just displaying a zoom cursor. Or maybe that's a good start.  I don't know.

So.... here are a couple options I'd like to get some feedback on:
*  Option 1 is to just add a Zoom Cursor, e.g. 
zoom_cursor to the shift middle and shift right button events

* Option 2 is to add in the QWheelEvent and QTouchEvent classes to the QGLView code and try to respond to all mouse/touchpad zoom events.

Option 1 is very easy, option 2 is a better overall experience assuming all aspects are code-able without truly unnatural acts.

MichaelAtOz:
I have a 3 button (plus 2 on the side) wheel mouse also with wheel left/right scroll ability. W7/64 bit.
Left - rotates
Right - translates
Middle - Nothing
Vertical scroll - zoom in/out
Horizontal scroll - zoom in/out (would be good to do something else with this, Orthogonal/Persective maybe?)
Both side buttons - translate.

I'd suggest Option 1, then Option 2 as another issue with more discussion. But who am I...

kintel:
@TDeagan I think the zoom cursor in MeshLab has potential: https://github.com/starseeker/meshlab/blob/master/src/meshlab/images/cursors/plain_zoom.png

(..but without the normal mouse cursor)

kintel:
This got stalled. @TDeagan if you're still around and interested in reviving this, it should be possible to make it work without too many more changes.

PR #1142, opened 2015-01-07
Show error/warning status in viewport.

t-paul:
Error:
openscad-error-1

Warning(s):
openscad-warning-1

The link opens the console window (in case it's not already open) and closes the message box.


Scalpel78:
The font is too large on Windows. Also, the text doesn't wrap, so in my example screenshot I cannot make the viewport any smaller. Tested on 2014.01.08 for Win64.

image

t-paul:
see #1149

PR #1144, opened 2015-01-07
Change license for example files.

t-paul:
- Change the examples written by me to CC0
- Include CC0 license text

### DIFF
--- a/examples/Advanced/offset.scad
+++ b/examples/Advanced/offset.scad
@@ -1,16 +1,4 @@
 // offset.scad - Example for offset() usage in OpenSCAD
-//
-// Written in 2014 by Torsten Paul <Torsten.Paul@gmx.de>
-//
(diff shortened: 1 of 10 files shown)


hroncok:
So now we have:
 * CC0 examples by Paul
 * CC BY-SA examples (Advanced/fractal.scad)
 * CC - GNU GPL (in fact GNU GPLv2+) examples (Shapes/sphere.scad and Shapes/chopped_blocks.scad)
 * no license (assuming the same license as OpenSCAD use applies) examples (multiple)

Any chance for simplification?

t-paul:
Yes, the plan is to get all licensed as CC0.

hroncok:
Great.

kintel:
@hroncok Just pushed a fix - only fractal is left. Waiting to hear from the authors.

Issue #1149, opened 2015-01-09
GUI issues


t-paul:
- [x] Some icons are not working with a dark theme
- [x] Welcome screen is unreadable in dark theme
- [x] The font for the info window should only use bold so it reacts to user settings

![he7f4f6](https://cloud.githubusercontent.com/assets/1330241/5674608/c3dd79f0-97ad-11e4-9825-757ae5ff8088.png)

![snapshot1](https://cloud.githubusercontent.com/assets/1330241/5681985/327ae0ea-981d-11e4-8e96-5ae9fcd0b4c4.png)

MichaelAtOz:
Should the extra warning ONLY show when the console is hidden?

t-paul:
No, I don't think it should depend on the console visibility. There might be other ways like hide it after some time or even make it configurable.

t-paul:
See #1150...

MichaelAtOz:
> There might be other ways like hide it after some time or even make it configurable.

Perhaps when there is activity in the editor window?

But configurable would be good, I usually have the console open so it would be superfluous for me.

PR #1109, opened 2014-12-25
Toolbar icons

shaina7837:
New icons as per requested in https://github.com/openscad/openscad/issues/1056 are added.

Issue #1076, opened 2014-12-16
RAM and OS version in Library Info is wrong in Windows 8.1


Scalpel78:
In the Library Info I have this line:
System information: Windows **8 (v6.2)** 8 CPUs **3.91 GB RAM**

Correct info should be: Windows **8.1 (v6.3)** 8 CPUs **16GB RAM**

The Windows version number error is most likely because the GetVersionEx has been changed in Windows 8.1 - see http://msdn.microsoft.com/en-us/library/windows/desktop/dn481241(v=vs.85).aspx

For the RAM I'm unsure of the cause. If I run the following code in a separate Visual Studio C++ project I get 15.9087. The code is very similar to that OpenSCAD uses in PlatformUtil-win.cc, but without the PlatformUtils::toMemorySizeString call.

	MEMORYSTATUSEX memoryinfo;
	memoryinfo.dwLength = sizeof(memoryinfo);
	if (GlobalMemoryStatusEx(&memoryinfo) != 0) {
		cout << "RAM:" << (float)memoryinfo.ullTotalPhys / (1024 * 1024 * 1024) << endl;
	}

The 3.91 GB OpenSCAD reports could be a 32-bit upper limit, but the application isn't running in 32-bit mode. It's running as a regular 64-bit app.

Scalpel78:
I can confirm that the RAM size is correct now.

Windows version is still reported as standard Windows 8 (v6.2).

t-paul:
Looks like the VersionHelper.h is not available on MinGW, but I found http://hg.m1xa.com/findversion which uses the same underlying API to detect versions up to 8.1 and it's MIT licensed.
This reports now ```Windows 7 SP1 (v6.1.1.0) 2 CPUs 3.00 GB RAM``` on my VM which matches the ```HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion``` info.

@Scalpel78 I've pushed the change to https://github.com/openscad/openscad/tree/win-version-detect - can you test this on your Win8.1 system?

Scalpel78:
Thanks - I didn't get the development environment up and running on my machine, so either we'll have to wait until I get it working (might take a long time), or I can test this if the change is merged into the master branch and a new fileset is made.

kintel:
Snapshots will come soon :)
[...]

PR #1150, opened 2015-01-09
Gui updates

t-paul:
This is fixing the issues from #1149 and includes the new Toolbar icons from #1109.

- [ ] reconsider look of the render icon
- [ ] reconsider look of the reset-view icon

snapshot2


MichaelAtOz:
I think the size of the compile error message is a bit over the top compared to all the other GUI items
It also resizes the edit window to make room when the overall window is small. Also grows the window if needed. I'm not sure if there are still people out there with small monitors?? But I occasionally use small OpenSCAD windows side-by-side.

Also, is it possible to ignore DEPRECATED for this purpose? There is little you can do about it and it stuffs up the display for every compile.

warning

t-paul:
The font should be fixed already.

MichaelAtOz:
For info that was 2015.01.08.

t-paul:
It's work in progress. I just closed issue #1149 as the pull request now exists here.

t-paul:
Snapshots 2015.01.10 are online. Note that those are including both the ```gui-updates``` (#1150) and the ```win-version-detect``` (#1076) changes which are not yet merged...

Please check the reported windows version too. The change is supposed to detect win8.1 but of cause it should still correctly report older Windows version as well.

MichaelAtOz:
1. Windows version appears to be correct for my W7. Would there be any need to report Windows 32 bit v's 64 bit? (it does show that gcc is 64 bit)

2. The compile warning message is better, but the close icon needs work:
compile icon

3. The new splash screen background is better but it conflicts with the two buttons:
splash screen

Scalpel78:
My splash screen looks slightly better than Michaels.

image

But how do you get the dark background inside the main application? I can set the viewport and the editor to dark themes, but not the application itself.

image

MichaelAtOz:
> But how do you get the dark background inside the main application? I can set the viewport and the editor to dark themes, but not the application itself.

I suspect that is Windows Display Personalisation. Tho I think the OP is Linux or Mac.

t-paul:
For a 32bit build running on a 64bit system, it will include [WOW64](http://en.wikipedia.org/wiki/WoW64) in the version line (which has nothing to do with the game :smiley:).

PR #1151, opened 2015-01-10
Add source code reformat option using astyle library.

t-paul:
See Artistic Style, A Free, Fast, and Small Automatic Formatter.
http://astyle.sourceforge.net/

This is completely optional, if the library is not found at `qmake` time, the functionality is disabled.

It's using the C++ formatting, so the behavior is not perfect, but it should be still useful to reformat complete files.

TODO:
- [ ] Add to macosx-build-dependencies.sh
- [ ] Make sure it works with MXE
- [ ] Make sure it works with MSYS2


kintel:
We probably need to be able to control the formatting. Not sure how configurable astyle is.
Did you look into Uncrustify - looks like they've got more languages implemented: http://uncrustify.sourceforge.net

t-paul:
With the ongoing work on a scad specific QScintilla lexer, we might be able to base the formatting on this code instead of astyle.

I think I've skipped uncrustify back then as I did not find a description of how to use it as a library (without spending much time on searching; the debian package seems to only ship a binary). Calling an external program seemed to be more trouble than it's worth for OpenSCAD.

PR #1168, opened 2015-01-18
Create a new command called "convex"

BonyJordan:
I always find myself constructing fairly simple polyhedra which are not quite cubes and not quite pyramids, but are always convex.  There's no obvious way to use hull() with a list of points, and it seems horrible to put epsilon sized cubes there.  Meanwhile, polyhedron() works, but it's kind of a pain to figure out the faces each time.  So, to simplify my life, I have created a command "convex", which takes an array of 3D-points and produces the convex hull of those points.

### DIFF
--- a/src/cgalutils.cc
+++ b/src/cgalutils.cc
@@ -148,6 +148,11 @@ namespace CGALUtils {
 			}
 		}
 
+		return applyHull(points, result);
+	}
+
+	bool applyHull(const std::list<K::Point_3> &points, PolySet &result)
+	{
 		if (points.size() <= 3) return false;
 
 		// Apply hull
--- a/src/cgalutils.h
+++ b/src/cgalutils.h
@@ -13,6 +13,7 @@ typedef std::vector<PolygonK> PolyholeK;
 
 namespace CGALUtils {
 	bool applyHull(const Geometry::ChildList &children, PolySet &P);
+	bool applyHull(const std::list<K::Point_3>& points, PolySet &P);
 	CGAL_Nef_polyhedron *applyOperator(const Geometry::ChildList &children, OpenSCADOperator op);
 	//FIXME: Old, can be removed:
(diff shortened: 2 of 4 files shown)


t-paul:
Nice, that sounds useful.

Maybe that could be integrated into polyhedron if it has no faces defined?

Note that primitives.cc can't depend on CGAL, this causes the build error in the test suite (See Travis: https://travis-ci.org/openscad/openscad/builds/47464722).

BonyJordan:
Your polyhedron idea was excellent, I implemented that right away.  I hope I have fixed primitives.cc correctly, but I cannot easily replicate the test script on my mac.

BonyJordan:
I see.. we can't build csgtexttest.  But is it okay to make polyherdron() a non-primitive?  I could cut my own hull algorithm I suppose.

t-paul:
I hope @kintel has an idea how to best untangle this.

kintel:
I'd like to avoid too many compile-time conditions, so I think the best solution would be to move LeafNode::createGeometry() to GeometryEvaluator::visit() for all leaf nodes.

BonyJordan:
I can put the code from LeafNode::createGeometry() into GeometryEvaluator::visit(),  but since LeafNode is currently opaque, I'd have to add the shape-type information into LeafNode, which sort of kills its ability to be a nice virtual super class for not just the sphere/cube stuff but also the importing stuff.  Is that what you guys want?

kintel:
Perhaps it needs some more thinking. The idea behind LeafNode was to use if for nodes which can create their own geometry. Separating out CGAL was done for other reasons (compile time issues, and API encapsulation of a huge library). There are other LeafNodes which need tools for geometry creation too; e.g. surface(), text(), the future trace(). Perhaps we just need to sit down and rethink this and refactor it accordingly.

In any case, I think it's too late to include this in the upcoming release, so we don't need to stress to get this fixed.

gadgetmind:
Sorry to resurrect this - I wasn't sure whether to comment here or create a new forum/mail thread. 

Is this feature now dead? It's something I've often wanted (without knowing it!) and have ended up creating small cubes, which is less than ideal.

PR #1173, opened 2015-01-22
español

b4zz4:
new languge spanish


b4zz4:
all spanish
----
todo el español

kintel:
@b4zz4 Could you add your name and contact info to the language file? https://github.com/openscad/openscad/blob/master/locale/es.po

..just so we can notify you when something seems to be out of sync.

PS: The current master is a Release Candidate and we plan to release in a few days, so a final check that all is ok would be cool :)

kintel:
wops, your name is there, sorry. I got confused by Don Bright also having his name there.

PR #1179, opened 2015-01-25
Mac build failing on line 908-909

clothbot:
Looks like lines 908-909 need to be reversed for my Mac OS X build to work on my latest git pull.

Error message is:

--snip--

./locale/ru.po:908: missing 'msgstr' section
./locale/ru.po:910:7: syntax error
msgfmt: found 2 fatal errors
error running msgfmt
make: *** [OpenSCAD.app/Contents/MacOS/OpenSCAD] Error 1

--end-snip--

Andrew.


t-paul:
Thanks, looks like we need to find a way to catch those cases on Travis too...

PR #1189, opened 2015-01-28
Updated the Czech translation

hroncok:
It is 99 % finished. But I have a few concerns:

What is the meaning here and where is the text displayed? https://github.com/openscad/openscad/blob/master/src/ProgressWidget.ui#L14
https://github.com/openscad/openscad/blob/master/src/Preferences.ui#L1393


kintel:
Removed the unused titles

Issue #1251, opened 2015-03-09
Crash on exit while processing


kintel:
Trying to quit openscad while something is processing tends to cause a crash.
This should be fixed by properly cleaning up after ourselves and killing existing jobs on exit (or rather on closing of a mainwindow). In cases where processing is done in a separate thread, we might want to (unsafely) end the thread.

nomike:
I submitted #6750 which fixes this. If a rendering is happening while the user closes the main window, the render is aborted, preventing the crash.

PR #1273, opened 2015-03-20
Better handling of shared resources when closing main windows

kintel:
This should fix/improve #1251

Update: Still crashes on long previews. We prob. need to cancel the preview and postpone closing the window until the cancel happens.


kintel:
Closing as abandoned effort

PR #1290, opened 2015-03-30
Add support for interactive axis-aligned clipping planes in OpenGL window

emogenet:

The slider moves the clipping plane along the axis and spans the length of
the bounding box of the scene.

full scene is visible.


t-paul:
Nice feature.

There seems to be an issue with the test suite not rendering anything, maybe some initialization problem? (Some info about the test suite is in https://github.com/openscad/openscad/blob/master/doc/testing.txt)

emogenet:
Yeah, clipMode not initialized in the right place.
I have a fix. I'll make another pull request.

PR #1291, opened 2015-03-31
Add slider and radio buttons to UI to interactively control an OpenGL axis-aligned clipping plane

emogenet:

The slider moves the clipping plane along the axis and spans the length of the
bounding box of the scene.

full scene is visible.

While the slider moves, the current position of the clipping plan is displayed
in HUD mode in the 3D scene.

Also useful to get an idea of what the 3D printer will see when slicing
(clip on Z axis).


kintel:
The reason why the build failed is that you forgot to add glQuickText.cpp to test/CMakeLists.txt

PR #1292, opened 2015-03-31
Gl code split

ankush1995:


kintel:
One question: what is the binary file "src/.polyset.cc.swp" ?

ankush1995:
It's just a swap file created by vim in Linux OS. It is of no use here. By mistake I have uploaded it.

Issue #496, opened 2013-09-29
OpenCSG difference does not work if you view from inside the subtrahend object.


nophead:
If you subtract a large cube from something it only works if the viewpoint is outside the cube. E.g.

```
difference() {
    cube(20, center = true);
    cube(1000, 1000, 1000);
}
```

It seems like the large cube gets culled because the front of it is in front of the clip plane but the back is still relevant to the difference.
##

kintel:
This does behave as expected; don't think there is much actionable here

PR #1293, opened 2015-03-31
Add slider and radio buttons to UI to interactively control an OpenGL axis-aligned clipping plane

emogenet:

The slider moves the clipping plane along the axis and spans the length of the
bounding box of the scene.

full scene is visible.

While the slider moves, the current position of the clipping plan is displayed
in HUD mode in the 3D scene.

Also useful to get an idea of what the 3D printer will see when slicing
(clip on Z axis).


kintel:
There are some challenges in integrating this:
* Using the glClipPlane approach is not compatible with previews. OpenCSG is not aware of the clipping planes and will not render the preview correctly
* The text rendering also works only in render mode, not preview mode
* Using a glClipPlane just clips the surface mesh. The objects designed are solid, and when clipping I would expect the inside of the object to be solid, not hollow.
 
The first two are possible to work around by enabling this feature only for render mode. I'm not too fond of that way of doing it though. Alternatively, work on getting OpenCSG to work with clipping planes (which would also address #496).
A different approach altogether would be to use OpenCSG to clip instead of using a clipping plane.

emogenet:
For some reason, my pull requests don't update automatically when I
push to my repo ...  I'm going to send a new pull request than implements
a better version.

Issue #1287, opened 2015-03-29
openscad fails to load scad files with non-ascii file names from file explorer


1i7:
Hello, OpenSCAD 2015.03, Windows 8

I have колесо.scad file, double-click in file explorer to open in OpenSCAD, OpenSCAD opens, but file does not load, OpenSCAD shows error message in console:

Failed to open file ������.scad: No such file or directory

Same file opens from File/Open menu dialog

kintel:
We'll handle this in #1296

jordanbrown0:
Fingers crossed that this will get fixed with issue #113 / PR #6136 .

jordanbrown0:
Works in 2026.04.26.

Issue #1286, opened 2015-03-29
import dxf does not work with non-ascii file names


1i7:
Hello, Windows 8, OpenSCAD 1015.03

import fails to load dxf files with non-ascii file name.

this will load dxf file to workspace:
import(file = "C:/gear/gear.dxf");

this will not load same dxf file with different name to workspace:
import(file = "C:/gear/колесо.dxf");

this seems to be regression, at least in 2013.06 running on the same system same file loads fine.

kintel:
This is a problem in multiple places. Looks like some refactoring gone wrong.
I'll look into it.
Suggestion:
* All filenames should be utf-8 std::string or QString internally
* Conversions between std::string and boost::filesystem::path should use an utf-8 codec
* Any usage of fopen and iostreams are broken by design under Windows, and we need to use wide char variants (or a wrapper library)
* We need manual and automated tests for all code paths (GUI and file operations)

kintel:
We'll handle this in #1296

jordanbrown0:
See also #113; this one might well be fixed when that one is fixed.

jordanbrown0:
Works in 2026.04.26.

Issue #113, opened 2012-05-13
non ascii filenames cause errors on Windows


donbright:
Note: Attempted solved by #1296, but that PR needs a new owner with some Windows foo.

from chrysn from the mailing list:

Prepare:

```
$ mkdir schön
$ echo "include &lt;b.scad>;" > schön/relative.scad
$ echo "include <../schön/b.scad>;" > schön/not_that_relative.scad
$ echo "sphere();" > schön/b.scad
```

Fails, printing wrongly formatted paths in the console
`$ openscad schön/relative.scad`
Fails too, printing wrongly formatted paths
`$ openscad schön/not_that_relative.scad`

In the directory:
`$ cd schön`

Fails, printing correctly formatted paths:
`$ openscad relative.scad`

WORKS:
`$ openscad not_that_relative.scad`

Fails, printing only the filename:
`$ openscad -o test.dxf relative.scad`

Fails, printing only the filename:
`$ openscad -o test.dxf not_that_relative.scad`

The whole thing behaves slightly differently depending on the LANG env. variable, e.g. LANG=en_GB.UTF-8 vs. LANG=C

Same problem happens also with customizer data file as reported in #2453.

##

ocastelao:
Not working in Windows 7 64 bits: "Can't open file "C:/Users/�scar/Documents/OpenSCAD/ViE.stl" for export"

t-paul:
@ocastelao Which places did you try? Command line, Open via Menu entry, Open from the scad file, ...?

ocastelao:
I can create and save .scad files, but i can't export to .stl files using the STL buttom of the graphic interface, because the "�" character.

t-paul:
@ocastelao Could you please have a look at the [2015-06-14 snapshots](http://www.openscad.org/downloads.html#snapshots)? Those include the (proposed) fix for the export.

ocastelao:
I have just tried OpenSCAD 2015.06.14 64 bits, and STL export now works well... Thank you!

Sastas:
Still a problem. I tried the last release. Cannot export STL file to "G:/3D nyomtató/Modellek/" If I try export anywhere else to G(or other drives) what contains only ascii, works well.

kintel:
@Sastas Which version are you using?

Sastas:
2015.03. I downloading the latest update now, and Write here the resoult.

Sastas:
[...]

PR #1296, opened 2015-04-02
#113 utf-8 tests

kintel:
This pull request addresses various non-ascii filename issues, as discussed in #113, #1286, #1287 

TODO:
* [ ] Improve test cases
* [ ] Perform tests
* [ ] Fix launching screen issue
* [ ] Fix include<> issue
* [ ] Fix import issue (#1286)
* [ ] Fix double-click issue (#1287)


MichaelAtOz:
For posterity [Boost Path Name Portability Guide](https://www.boost.org/doc/libs/1_69_0/libs/filesystem/doc/portability_guide.htm) may be useful.

thehans:
I think this is getting close to being usable.  Just a few notes so far.

I think stat calls are still a problem so StatCache may need to be rewritten, either to 
1) use `boost::filesystem::exists()` + `boost::filesystem::last_write_time()` (not great that we'll need two calls that probably do similar things under the hood)
2) possibly change `StatCache::stat` to accept a `boost::filesystem::path` instead of `std::string`, then I think we could call `path::generic_wstring` and `wstat` specifically for windows ( `#ifdef _WIN32 ...`).

Another note about StatCache is that I'm not convinced that the cache is even being useful because it seems that its only good for 0.190 **milliseconds** unless I'm misinterpreting the code here:
https://github.com/openscad/openscad/blob/a30dbed0e18398d86821edd8cc3b8a46c6633a30/src/StatCache.cc#L36
I talked a little with @nophead about it here: https://github.com/openscad/openscad/pull/2790

The last thing I'm not sure about is how the `PlatformUtils::ensureStdIO` interacts with nowide.  We may need to attach the c++ cout/cerr streams separately from stdout/stder, I still need to research that some more.  
Also maybe a good idea to just move completely over to stream implementation, which would mean replacing any occurences of `printf`, `fprintf(stderr,...` with equivalent `nowide::cout`, `nowide::cerr` calls.

kintel:
I had hoped that NoWide would have made  it into Boost by now, but it looks stalled although it pass review: https://lists.boost.org/boost-announce/2017/06/0516.php

thehans:
@kintel wrote:
> I had hoped that NoWide would have made it into Boost by now, but it looks stalled although it pass review: https://lists.boost.org/boost-announce/2017/06/0516.php

I just noticed that NoWide is listed as a New Library in the latest 1.73 release(April 28th 2020):
https://www.boost.org/users/history/version_1_73_0.html
https://www.boost.org/doc/libs/1_73_0/libs/nowide/doc/html/index.html

Also there's been some activity on their github lately, specifically they just created a PR which addresses the feature request I made about supporting `stat`. :tada: 

I guess now we just wait a couple more years for it to be standard on Ubuntu etc.  :wink:

Issue #891, opened 2014-07-31
extend animation export options


t-paul:
- Change animation panel to dockable window and decouple show/hide of the window from the actual animation logic
- Add more export options
  - Animated GIF - [test file](http://files.openscad.org/video/test.gif) / [optimized file](http://files.openscad.org/video/test-optimized.gif)
  - WebM (VP8) - [test file](http://files.openscad.org/video/test.webm)
  - MP4 - [test file](http://files.openscad.org/video/test.mp4)
- Suggestions from comment in #889
  - save step size and fps (probably should go in the .scad file)
  - buttons in animation widget to play/pause fwd/rewind
  - image export - auto uncheck after a full loop
  - image export - option to do full render of images instead of F5/Preview

Prototype screenshot
![openscad-animation-window](https://cloud.githubusercontent.com/assets/1330241/3770896/d1998cec-18ef-11e4-94c2-53f924c7afec.png)
##

jordanbrown0:
Animation has been made a dockable window some time in the last couple of years, so dropped that part from the title.

PR #1300, opened 2015-04-05
Interactive Clipping Planes

emogenet:
Re-implemented clipping planes to take comments from https://github.com/openscad/openscad/pull/1293 into account.
1. Add XYZ radio buttons to select interactive clipping plane in UI
2. Add slider to interactively move clipping plane in UI
3. Clipping plane is implemented using OpenCSG when in OpenCSG rendering mode
4. Clipping plane is implemented with glClipPlane when in CGAL rendering mode
5. Added a method to display quick and dirty in-scene anti-aliased fonts. Currently
    used to display clipping plane, but could be used to display others things in the
    3D view, e.g. axes, dimensions, etc

This is not 100% perfect yet: in CGAL rendering mode, the clipped walls aren't
rendered. Will add this later if I find the time, but the feature is very usable at this point.


kintel:
Thanks for the update. Some comments:
* We need a way to show/hide the GUI
* The CSG clipping object causes rendering artifact as it seems to be a bit oversized. We might need to adjust the camera clipping planes to take the new negative object into account.
* The 2D text stopped working (only tested on Mac). It just displays a black rectangle per character. Since we already have a Qt GUI, perhaps displaying the text in the GUI is a cleaner solution?
* The clipping plane color needs to be synced with the color scheme, or perhaps just use the negative color
* The clipping plane uses only the positive halfspace, while the default camera looks down the negative Y  axis. Not sure what to do about that.

Sorry for a lot of requests, but once we add features, we have to maintain them..

znort987:
Couple of remarks/questions:

1. Regarding show/hide the GUI : what do you have in mind exactly ? Some sort of collapsable widget ? Treeview ? Let me know what you want in more detail, happy to build it.

2. The 2D text works fine on the various boxen I've tested on. (windoze and a couple of buntus with various graphics cards as well as Mesa). I'll give it a try on a Mac if I can manage get a hold of one of these things. It might be that my code fail to properly initialize something in the OpenGL context which happens to be off by default on Apple's implementation. A QT display is fine. In fact, I'd like to have a input/output textfield so you can enter the clipping value numerically. Still, I'd want glQuickText to work properly, because I'm planning to add in-scene dimension annotations (e.g. measurements), and for those, Qt won't cut the mustard.

3. Re. syncing with color scheme / negative color. No idea what those are. Where can I get the color to paint the clip walls from in the code ?

4, Last remark about positive halfspace. Not sure what you mean here. Do you mean that for Y, the clipping plane clips the wrong way ? That's easy enough to fix. Although I'd argue that the default view is weird. But either way, it's an easy fix, let me know what you want.

emogenet:
Couple of remarks/questions:

    Regarding show/hide the GUI : what do you have in mind exactly ? Some sort of collapsable widget ? Treeview ? Let me know what you want in more detail, happy to build it.

    The 2D text works fine on the various boxen I've tested on. (windoze and a couple of buntus with various graphics cards as well as Mesa). I'll give it a try on a Mac if I can manage get a hold of one of these things. It might be that my code fail to properly initialize something in the OpenGL context which happens to be off by default on Apple's implementation. A QT display is fine. In fact, I'd like to have a input/output textfield so you can enter the clipping value numerically. Still, I'd want glQuickText to work properly, because I'm planning to add in-scene dimension annotations (e.g. measurements), and for those, Qt won't cut the mustard.

    Re. syncing with color scheme / negative color. No idea what those are. Where can I get the color to paint the clip walls from in the code ?

4, Last remark about positive halfspace. Not sure what you mean here. Do you mean that for Y, the clipping plane clips the wrong way ? That's easy enough to fix. Although I'd argue that the default view is weird. But either way, it's an easy fix, let me know what you want.

t-paul:
For the GUI, I'm thinking the use of dock-able windows could be the way to go. Like for the (still incomplete) change to the animation input https://github.com/openscad/openscad/issues/891.

The menu entries for show/hide probably need to get their own top level menu ("Windows"?) later.

I'm not sure what the issue with the pull request is. I never tried with `master`, but when using a separate branch in the local repository, github did always pick up additional changes. Maybe this feature is blocked for `master`?

kintel:
For how to use colors, see e.g. https://github.com/openscad/openscad/blob/master/src/OpenCSGRenderer.cc#L123

MichaelAtOz:
> The menu entries for show/hide probably need to get their own top level menu ("Windows"?) later.

Perhaps a 'View' menu, Get rid of the word 'Hide' i.e just Editor, Console, Toolbars with a tick or no tick (or a cross if doable?), Then the View Edges, View Axis etc could be moved to the same View main menu. The existing View is too big, particularly where shortcuts are concerned. ??

emogenet:
Closing this as the original repo for this was removed and recreated.
A proper branch was created in the new repo to track this.
The new pull request that replaces this one is here: https://github.com/openscad/openscad/pull/1424

PR #1312, opened 2015-04-13
Comment out search term not found WARNING

clothbot:
Returning empty result should be sufficient for detecting a no-match condition.

See forum discussion http://forum.openscad.org/Suppressing-quot-search-term-not-found-quot-warning-in-search-td12359.html


hroncok:
Why don't just delete this?

Or, for backwards compatibility, add nowarn=True argument?

t-paul:
I'd vote for deleting. In my view adding another parameter is a bigger problem for future compatibility than just to remove the warning.

Hmm, looks like Travis got stuck uploading the test results again... but the search tests cases need updating
```
30 - echotest_search-tests (Failed)
31 - echotest_search-tests-unicode (Failed)
```

kintel:
I agree; delete and update test case results

PR #1313, opened 2015-04-13
Removed deprecated assign statement

clothbot:


kintel:
The test results needs to be updated to reflect this change.

kintel:
Yah, travis really makes things easier. Just wish they had a Windows and Mac version as well :)

hroncok:
They actually have Mac version, but it is currently full. Stay tuned. http://docs.travis-ci.com/user/multi-os/

t-paul:
I think I've read that Mac Travis statement already a year ago, so I guess it's not high on the prio list (or maybe just too expensive due to Apple licensing policies?).

Issue #1230, opened 2015-02-26
The same file can be opened multiple times


Feuermurmel:
This may have specifically implemented as a feature, so forgive me if that's the case. If a file is opened (either via _File_ > _Open…_ or by opening the file from the Finder) that has already been opened, a new window for the same fie is opened. I would have expected that this would just bring the existing window for that file to the front.

kintel:
TODO: Define correct behavior for each platform

MichaelAtOz:
If this is implemented there should be a way to open the same file in more than one window/and/or/instance. I do this from time to time.

kintel:
Right, I guess a common use-case is to render different views of the same design.

MichaelAtOz:
Yes, with auto-reload, it works nicely.

Feuermurmel:
@MichaelAtOz I didn't think about that! I agree that this is useful and I know a lot of programs that offer that functionality. A common implementation is to provide a command along the lines of "New Window for \<document name\>" in the _View_ menu. Maybe even add a button to the toolbar.

Feuermurmel:
@MichaelAtOz I didn't think about that! I agree that this is useful and I know a lot of programs that offer that functionality. A common implementation is to provide a command along the lines of "New Window for <document name>" in the _View_ menu. Maybe even add a button to the toolbar.

Feuermurmel:
@MichaelAtOz I didn't think about that! I agree that this is useful and I know a lot of programs that offer that functionality. A common implementation is to provide a command along the lines of "New Window for <document name>" in the _View_ menu. Maybe even add a button to the toolbar.

Feuermurmel:
@MichaelAtOz I didn't think about that! I agree that this is useful and I know a lot of programs that offer that functionality. A common implementation is to provide a command along the lines of "New Window for <document name>" in the _View_ menu. Maybe even add a button to the toolbar.

jotpandher:
Done.

jordanbrown0:
I think the multiple views use case is better addressed through an explicit multiple views mechanism, ref #2385.

PR #1323, opened 2015-04-23
solve issue 1230

jotpandher:
The same file can be opened multiple times #1230 
Done.


kintel:
As discussed in #1230, we didn't agree on how this should work. Multiple views of the same file is a common workflow.

kintel:
Closing as this was part of a GSoC application and no progress was made in a while

PR #1339, opened 2015-05-12
Add "Export to Clipboard" to copy the current 3d view to the system clipboard.

t-paul:
Tested only on Linux so far (Copy to LibreOffice Writer works).

TODO:
- [x] test on Mac OS
- [x] test on Windows

### DIFF
--- a/src/MainWindow.h
+++ b/src/MainWindow.h
@@ -196,7 +196,7 @@ private slots:
 	void actionExportSVG();
 	void actionExportCSG();
 	void actionExportImage();
-	void actionExportToClipboard();
+	void actionCopyViewport();
 	void actionFlushCaches();
 
 public:
--- a/src/MainWindow.h
+++ b/src/MainWindow.h
@@ -196,6 +196,7 @@ private slots:
 	void actionExportSVG();
 	void actionExportCSG();
 	void actionExportImage();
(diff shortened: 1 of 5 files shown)


kintel:
Tested on OS X - export to clipboard and paste to imgur.com works!
Some comments:
* Perhaps this belongs in some other menu, with a convenient shortcut?
* Are there any other tools doing similar things?

MichaelAtOz:
'Copy Viewport' in Edit menu? With Ctrl-Shift-C (as a modifier of std copy shortcut)?

t-paul:
Sounds good to me, patch updated. Tested on Windows, copy&paste into Write document works.

t-paul:
Windows snapshots (2015-05-16) are updated based on current master + this patch.

MichaelAtOz:
Works nicely, W7/64, paste to Word, paste to Infranview (save as .jpg).

Anything else from Master need testing?

t-paul:
Master has currently only some additional fixes, most notably @kintel fixed #1325.

Otherwise, I'm uploading a special preview version for #1110 to toy with the proposed `trace()` module which is not yet in master (as `OpenSCAD-2015.05.17-x86-{32,64}_potrace.zip`, feature needs to be enabled in preferences).

Issue #505, opened 2013-10-17
Quiet mode for command line


benhowes:
I'm using openscad with makefiles from the command line a fair bit and it would certainly be handy to be able to turn off everything other than errors with a -q, or put them on stdout so I can pipe them to /dev/null or something.

I searched through the issues and didn't find this as one, sorry if there is.

Thanks!
##

ChrisVesters:
Is this still revelant?

benhowes:
It looks like #1343 closes this.

PR #1343, opened 2015-05-13
Implement quiet mode

ArchimedesPi:
Implements suggestion in #505. Doesn't print anything except error messages when triggered.
The global quiet state is held in `OpenSCAD::quiet`, similar to how debug is handled.
Quiet mode is triggered by **`-q`** or **`--quiet`**.


hroncok:
Nice. Does it print warnings?

ArchimedesPi:
@hroncok Nope. Should it?
The big problem with this though, is that since we don't have seperate warning, error, and info `PRINT` macros, when a multiline ERROR or WARNING gets printed, with ERROR or WARNING prepended on only one line, only the line starting with ERROR or WARNING gets printed.
I've heard some talk about reimplementing *all* the logging stuff with boost::log, which could solve a bunch of these problems.

hroncok:
It would be really cool to have various verbosity levels with boost::log. I
use similar mechanism with Python logging and it is quite useful.

ArchimedesPi:
Then you could add an optional parameter to --quiet, like --quiet=WARNING.
That would mean hooking in the debug stuff though, for optimal ease-of-use.

MichaelAtOz:
`-q` & `--quiet` needs to be added to `-h` help text. I check against 2017.04.05.

PR #1351, opened 2015-05-22
Add ray casting on model from command line

emogenet:
This adds a new command line option that casts one ray on the model and
outputs the result on stdout as:

<triangleId> <distance from ray origin> <intersection point> <normal at intersection point>

This is useful for doing after the fact measurements of the model.

It is also very useful to implement regression testing on complex parametric models.

### DIFF
--- a/src/export.cc
+++ b/src/export.cc
@@ -101,6 +101,177 @@ void exportFile(const class Geometry *root_geom, std::ostream &output, FileForma
 	}
 }
 
+struct Intersection {
+
+    uint64_t tId;
+    CGAL::Point_3<CGAL_Kernel3> point;
+    CGAL::Vector_3<CGAL_Kernel3> normal;
+
+    Intersection(
+        uint64_t                     _tId,
+        CGAL::Point_3<CGAL_Kernel3>  _point,
+        CGAL::Vector_3<CGAL_Kernel3> _normal
+    )   :   tId(_tId),
+            point(_point),
+            normal(_normal)
+    {
+    }
+};
+
+struct IntersectionSorter {
+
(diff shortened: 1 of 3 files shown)


emogenet:
Example use:

./openscad --raycast=0,0,0,1,1,1 examples/Basics/CSG.scad 
ECHO: version = [2015, 5, 22]
CGAL Cache insert: union(){cube(size=[15,15,15],center=true (1142976 bytes)
CGAL Cache insert: difference(){cube(size=[15,15,15],center (700656 bytes)
CGAL Cache insert: multmatrix([[1,0,0,-24],[0,1,0,0],[0,0,1 (1142976 bytes)
CGAL Cache insert: intersection(){cube(size=[15,15,15],cent (690640 bytes)
CGAL Cache insert: multmatrix([[1,0,0,24],[0,1,0,0],[0,0,1, (700656 bytes)
CGAL Cache insert: group(){multmatrix([[1,0,0,-24],[0,1,0,0 (2534096 bytes)
0 1574 9.95813967898173224569 5.74933462428799568755 5.74933462428799568755 5.74933462428799568755 0.64446303864007825624 0.58027736788500949849 0.49794133002517720898

donbright:
how much work would it be to add on a PNG image dumper and turn this into a raytracer? (we link lodepng.cpp already)

emogenet:
@donbright : this is really not the purpose of this patch: the purpose of this patch is to probe models for geometric extent along a particular line (to e.g. help in further modeling).

In particular, this patch uses CGAL to cast rays on the model, and CGAL emphasizes exactitude over speed for its ray intersection routines.

It would be - I think - *extremely* slow to render a PNG image with this.

kintel:

Issue #1362, opened 2015-06-04
remove OPENSCAD_TESTING build flag


chrysn:
the OPENSCAD_TESTING flag currently tweaks openscad's behavior to give reproducible results in unit tests. since 4d18db6 was merged, its use has been greatly reduced, but the remaining uses are:
- disabling building of the QT gui
- dropping the user library path
- normalizing ECHO output (removing timestamp=, normalizing floats)

as unit tests are nowadays run through python scripts, the normalization can happen there. dropping the user library path could be handled with a command line option (or people should just not have breaking things in their library when they run unit tests), and QT can stay (because then, we wouldn't need to build openscad_nogui for testing any more at all).

i can offer a patch that does all the normalization in test_cmdline_tool.py. then, we can drop the OPENSCAD_TESTING ifdefs in the code, and stop building openscad_nogui for the tests (and use the regular openscad instead).

this is a prerequisite for #1361.
##

chrysn:
a patch series is available on http://anonscm.debian.org/git/collab-maint/openscad.git in the chrysn/drop-openscadtesting branch -- can you work from that / review that or should i fork openscad on github and send a github internal pull request?

t-paul:
Only getting an empty page at that link....

t-paul:
Ahh, ok, that only works as git clone url.

chrysn:
as noted on irc, my patches erroneously removed the ENABLE_EXPERIMENTAL flag (which is still useful to build some binaries in tests/); moreover, those recently failed to build because the build system and the binaries rely on being built without QT (previously implicit with OPENSCAD_TESTING, now explicit with OPENSCAD_NOGUI). my branch has been updated to fix that.

@t-paul, there is also a web version available of the branch at http://anonscm.debian.org/cgit/collab-maint/openscad.git/log/?h=chrysn/drop-openscadtesting .

kintel:
@chrysn If you push your patches to github and d a pull request, it would simplify reviewing and applying this.

rewolff:
chrysn, fyi, I have had a need for the "-nogui" version. having it built/tested may help keep it in working order.

PR #1363, opened 2015-06-07
remove OPENSCAD_TESTING build flag

chrysn:
This solves issue #1362.

### DIFF
--- a/openscad.pro
+++ b/openscad.pro
@@ -208,11 +208,17 @@ CONFIG += fontconfig
 CONFIG += gettext
 
 #Uncomment the following line to enable the QScintilla editor
-CONFIG += scintilla
+!nogui {
+  CONFIG += scintilla
+}
 
 # Make experimental features available
 experimental {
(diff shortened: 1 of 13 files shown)


kintel:
I experience two test failures related to fonts using this branch:
http://files.openscad.org/tests/mac_64-bit_nvidia-geforce-gt-750m_eddj_report.html

chrysn:
the font failures happen on travis as well ([see build #1732](https://travis-ci.org/openscad/openscad/builds/65850760)) -- they just never happened on my systems because i'm always using system fonts on debian, and thus they're installed on the build machines.

my general take on non-system-default paths (be it PATH, openscad library or font path) is that all tools (even testing ones) should respect the system defaults and the environment, and only the tools that know that we're in a build situation will export appropriate overrides via the environment; i'll try to fix it along that line.

chrysn:
the font failures are resolved now, things build fine.

Issue #1355, opened 2015-05-29
openscad_nogui requires -lX11


chrysn:
building the test suite failed on debian sid with:

    Linking CXX executable openscad_nogui
    /usr/bin/cmake -E cmake_link_script CMakeFiles/openscad_nogui.dir/link.txt --verbose=1
    /usr/bin/c++   -g -O2 -fPIE -fstack-protector-strong -Wformat -Werror=format-security -D_FORTIFY_SOURCE=2  -fno-strict-aliasing -DEIGEN_DONT_ALIGN -O2 -g    -fPIE -pie -Wl,-z,relro -Wl,-z,now CMakeFiles/openscad_nogui.dir/home/chrysn/debian_packaging/openscad/openscad/src/openscad.cc.o  -o openscad_nogui -rdynamic libtests-offscreen.a libtests-cgal.a -lGLEW -lopencsg libtests-common.a libtests-core.a -lGLU -lGL -lglib-2.0 -lfontconfig -lfreetype -lharfbuzz -lboost_program_options -lboost_filesystem -lboost_regex -lCGAL -lboost_thread -lboost_system -lpthread -lgmp -lmpfr 
    /usr/bin/ld: libtests-offscreen.a(OffscreenContextGLX.cc.o): undefined reference to symbol 'XGetErrorText'
    //usr/lib/x86_64-linux-gnu/libX11.so.6: error adding symbols: DSO missing from command line
    collect2: error: ld returned 1 exit status

i managed to get it to build by adding `LINK_FLAGS "-lX11"` to the `set_target_properties(openscad_gui PROPERTIES...` line in tests/CMakeLists.txt as a workaround, but that will likely not work with other operating systems, and should be better set up where `OffscreenContextGLX.cc` is selected in the platform specific settings.

donbright:
i installed debian sid and it won't even get that far. 

```
[ 69%] Building CXX object CMakeFiles/tests-cgal.dir/home/don/openscad/src/cgalutils.cc.o
c++: internal compiler error: Killed (program cc1plus)
Please submit a full bug report,
with preprocessed source if appropriate.
See <file:///usr/share/doc/gcc-4.9/README.Bugs> for instructions.
CMakeFiles/tests-cgal.dir/build.make:284: recipe for target 'CMakeFiles/tests-cgal.dir/home/don/openscad/src/cgalutils.cc.o' failed
[...]

PR #1365, opened 2015-06-10
Issue1355

donbright:
Fix X11 build issues  on Debian Sid mentioned in Issue #1355 reported by @chrysn

Also many fixes to build system and splitting of CGALutils to deal with gcc crashing on machines with less than 2 GB of RAM

Tested on

machine AMD64:
Debian Sid Unstable
Fedora 22 
Ubuntu 15.04
Ubuntu 14.04
Debian 7
Debian 8
openSUSE 13.2

machine ppc64:
Fedora 18


kintel:
Looks good to me (but I see it mostly from an OS X perspective).
@chrysn @teepee Could you sanity check that this fits with your workflows?

kintel:
Thanks don!

donbright:
thanks

PR #1366, opened 2015-06-11
Create multilingual appdata file using itstool

hroncok:
- updated appdata file to a newer syntax
 - created an english appdata template
 - the strings are extracted during updatepot function in translation-update.sh
 - the multilingual file is created at the end of updatemo in the same script
 - itstool needed when running make (added this information to README)

Note: I did not update the scripts for dependencies obtaining, don't know what's the proper way of dealing with them.

Please sanity check this and let me know what you think.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,7 @@ testdata/scad/misc/use-tests.scad
 /lexer_lex.cpp
 /parser_yacc.cpp
 /OpenSCAD.app
+/openscad.appdata.xml
 /openscad.pro.user
 /openscad
 /locale/*/*/*.mo
--- a/contrib/appdata.its
+++ b/contrib/appdata.its
@@ -1,0 +1,9 @@
+<!-- Copyright 2013 Richard Hughes <richard@hughsie.com> -->
+<its:rules
+  xmlns:its="http://www.w3.org/2005/11/its"
+  version="1.0">
+  <its:translateRule translate="no" selector="/component"/>
+  <its:translateRule translate="yes"
+    selector="/component/summary |
+              /component/description"/>
+</its:rules>
--- a/openscad.appdata.xml.in
+++ b/openscad.appdata.xml.in
@@ -1,7 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<application>
- <id type="desktop">openscad.desktop</id>
- <licence>CC0</licence>
+<component type="desktop">
+ <id>openscad.desktop</id>
+ <metadata_license>CC0-1.0</metadata_license>
+ <project_license>GPL-2.0+ and CC0-1.0</project_license>
+ <name>OpenSCAD</name>
  <summary>The Programmers Solid 3D CAD Modeller</summary>
  <description>
   <p>OpenSCAD is a software for creating solid 3D CAD models. Unlike most free software for creating 3D models (such as Blender) it does not focus on the artistic aspects of 3D modelling but instead on the CAD aspects. Thus it might be the application you are looking for when you are planning to create 3D models of machine parts but pretty sure is not what you are looking for when you are more interested in creating computer-animated movies.</p>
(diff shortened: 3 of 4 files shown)


donbright:
This appears to only being used to modify the appdata.xml file so it does not affect the building of the openscad binary itself if the itstool is missing. 

I think it would be nice if it more gracefully handled the situation where 'itstools' is not installed on the system. The current patch i tested died during 'make' in translation-update.sh with an error message. 

```/home/donb/openscad/scripts/translation-make.sh
Compiling language files...
msgfmt -c -v -o ./locale/fr/LC_MESSAGES/openscad.mo ./locale/fr.po
268 translated messages.
268 translated messages.
msgfmt -c -v -o ./locale/de/LC_MESSAGES/openscad.mo ./locale/de.po
268 translated messages.
msgfmt -c -v -o ./locale/cs/LC_MESSAGES/openscad.mo ./locale/cs.po
266 translated messages, 2 untranslated messages.
msgfmt -c -v -o ./locale/es/LC_MESSAGES/openscad.mo ./locale/es.po
217 translated messages, 41 fuzzy translations, 8 untranslated messages.
./scripts/translation-update.sh: line 93: itstool: command not found

```

itstools itself it has three main dependencies

1. python
2. python-lxml2 module
3. libxslt

I have been trying to test on a machine without root access and these dependencies are very difficult to build and install. For example python-lxml2 module from ftp://xmlsoft.org/libxml2/python/ does not  compile properly ( libxml.c:2360:28: error: dereferencing pointer to incomplete type) - tried two different versions.

hroncok:
So maybe, when it's not installed, just produce a warning and keep going (just copy the .in file). Because as I think of this now, users who compile openscad themselves do not need the appdata file, only package maintainers do. Sounds reasonable.

kintel:
@hroncok I agree with your last comment. Could you update your patch to handle that situation?

One question: Since appdata.xml is only used for the Linux build, could just ignore that for Windows and Mac builds, to avoid unnecessary warnings?

hroncok:
Of couse, I'll do it.

hroncok:
How about this?

t-paul:
Travis seems to have some issues with the script (I think it did also have issues with the `{a,b}` style stuff before)
```
./scripts/translation-update.sh: 101: [: Linux: unexpected operator
```

Maybe the check for the tool should be at the beginning to also cover the other call in https://github.com/openscad/openscad/pull/1366/files#diff-08362abd4275966580f5227f3c421b32R29

Instead of moving the files around, I think it would be more robust to copy temporarily and remove the copies after the itstool call. Right now if one of the steps fails, the files are gone.

hroncok:
> Maybe the check for the tool should be at the beginning to also cover the other call in https://github.com/openscad/openscad/pull/1366/files#diff-08362abd4275966580f5227f3c421b32R29

Well, when creating the pot files, the tool has to be present, and I guess only you (i.e. core contributors) do that, not the users who compile openscad. The messages from the xml file should be in the .pot you push in git, otherwise it makes no sense to have the command in there.

> Instead of moving the files around, I think it would be more robust to copy temporarily and remove the copies after the itstool call. Right now if one of the steps fails, the files are gone.

OK. BTW the bug is being resolved, so this might go away soon.

hroncok:
Ok, I don't understand why the following code fails on Travis:

      if [ x"$(uname -s)" == x"Linux" ]; then
       echo "itstool missing, won't apply translations to openscad.appdata.xml"
      fi

t-paul:
Hmm, the script is using `#/bin/sh` so it's probably using `dash` which is very strict regarding the syntax for comparing strings... http://stackoverflow.com/questions/1089813/bash-dash-and-string-comparison

hroncok:
@kintel please remove on of the `=` as said by @t-paul

t-paul:
Looks like that was indeed the issue... Travis now says:
```
217 translated messages, 41 fuzzy translations, 8 untranslated messages.
itstool missing, won't apply translations to openscad.appdata.xml
```

hroncok:
Cool, thanks.

donbright:
I am getting ungraceful make fails on solaris

217 translated messages, 41 fuzzy translations, 8 untranslated messages.
./scripts/translation-update.sh: itstool: not found

PR #1367, opened 2015-06-13
Use QSaveFile (if available) to write designs to disk. (fixes #1350)

### LINKED ISSUE #1350: Saving file when HD is full ends up in data loss
So, I just experienced horror here. Thankfully OpenSCAD and I do backups, but could be worse.

So, have the file on a partition which is almost full. Save file. You will notice error, and OpenSCAD now reloads the file, and I now have a truncated file loaded with no way of getting back the original.

Would be good if it didn't reload the file in the editor if it fails saving.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/16352898-saving-file-when-hd-is-full-ends-up-in-data-loss?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

t-paul:
Always check the status of the QTextStream to catch problems like disk full.

In addition QSaveFile can ensure the old file is not destroyed as it's using a temporary file which is committed at the end by just renaming it to the output file name. Unfortunately this is only available with Qt 5.1.

I guess this needs some cross-platform testing. On Linux it did not simply catch the device full problem as advertised. Instead it happily committed a zero byte length file. This is now caught by checking the QTextStream status.

### DIFF
--- a/src/MainWindow.h
+++ b/src/MainWindow.h
@@ -15,6 +15,7 @@
 #include <QMutex>
 #include <QSet>
 #include <QTime>
+#include <QIODevice>
 
 enum export_type_e {
 	EXPORT_TYPE_UNKNOWN,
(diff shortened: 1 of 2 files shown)


kintel:
Works fine on Mac.

PR #1380, opened 2015-07-05
Add asc()/ord() builtin function.

t-paul:
Suggestion for a builtin `asc()` function to complement the existing `chr()`.
- Accepts single or multiple strings as input, internally everything is handled as just a list of characters
- Empty string(s)
  - `echo(asc(""));` => `0`
- One single character will produce a number as output
  - `echo(asc("←"));` => `8592`
- Multiple characters as input will produce a list of numbers
  - `echo(asc("abc"));` => `[97, 98, 99]`
  - `echo(asc("ab", "c"));` => `[97, 98, 99]`
- Reverses the `chr()` call
  - `i = "test"; o = chr(asc(i)); echo(i == o);` => `true`
- Invalid input will produce a warning and is skipped
- In case of no valid arguments at all, `undef` is returned

TODO:
- [ ] Verify tests
- [ ] Docs
- [ ] Cheat sheet


MichaelAtOz:
Perhaps empty string should return zero? That is the typical representation of [NUL](https://en.wikipedia.org/wiki/Null_character) including in Unicode UTF-8.

t-paul:
The 0 byte is currently not really supported (`chr(0)` returns an empty string) which is mainly a decision to prevent building strings that contain an inline 0 byte.
Returning 0 for `asc("")` would match that, so I guess that might be a good idea.

doug-moen:
I suggest that asc() may not a good name for the function. I suggest ucode(), which is also part of my OpenSCAD2 proposal.

The name asc() is short for ASCII, and originated with the BASIC programming language, which was tied to the ASCII character encoding. Microsoft Visual Basic has an asc() function which returns the byte value 0-255 for the first byte in the character string. In Basic, a string is an array of bytes, and the language provides direct access to those bytes.

OpenSCAD strings work differently from Basic strings, and to avoid confusion, the function names should reflect the model. An OpenSCAD string is an array of Unicode code points. A code point doesn't fit into a single byte, and we don't provide any way to directly access the byte representation of a string. So I propose the name ucode(), to make it clear that it returns a unicode code point value, and not a byte value (0-255) or an ASCII code value (0-127).

t-paul:
I see no issues in changing the name to `ucode()`. The question regarding the OpenSCAD2 proposal was more about the actual parameter and result handling (e.g. returning a list of codepoints if the string has a length > 1, allowing multiple string parameters, ...).

doug-moen:
The behaviour of the proposed asc() function is more complex and inconsistent than I expected.

What I would expect, based on the asc() function in Microsoft basic, is a function that takes a single argument, which is a non-empty character string. And this returns the value of the first code point in the string, ignoring subsequent code points. If anything else is passed as the argument list, then the result is undef.
* asc("a") => 97
* asc("abc") => 97
* asc() => undef
* asc("") => undef
* asc(3.1416) => undef
* asc("foo", "bar") => undef

Alternatively, if the intent is to convert a character string into an array of code point values, then let's make this work consistently:
* asc("") => []
* asc("a") => [97]
* asc("abc") => [97,98,99]

In either case, I don't see the rationale in supporting more than one argument. We already have a str() function for converting a list of values into a character string. You can quite easy write asc(str(a,b,c)). It adds needless complexity, with no clear benefit, for asc() to subsume all of the functionality of str().

doug-moen:
t-paul said: "The question regarding the OpenSCAD2 proposal was more about the actual parameter and result handling (e.g. returning a list of codepoints if the string has a length > 1, allowing multiple string parameters, ...)."

In OpenSCAD2,
* ucode("a") converts a single character or 'code point' to the integer value.
* [for (c=str) ucode(c)] converts a string into a list of code point values.

The second expression relies on an OpenSCAD extension to the 'for' loop, which supports iteration over the characters in a string.

In OpenSCAD2, ucode() should report an error if you pass an illegal argument list, rather than returning undef. When a run-time error is reported, evaluation of the OpenSCAD script is terminated, and no model is produced.

doug-moen:
The OpenSCAD2 design esthetic is called [Composable Building Blocks](https://github.com/doug-moen/openscad2/blob/master/rfc/Composable_Building_Blocks.md).

It is based on the idea of simple, orthogonal operations. Each operation has a single job, and they don't overlap. These operations can be freely composed together to solve complex problems.

In this context,
* ucode(ch) is a simple operation that converts a character to an integer
* list comprehensions are used to build a list. To convert a string to a list of integers, you compose ucode() with a list comprehension: [for (c=str) ucode(c)]
* str() converts a sequence of values into a string. It can be composed with the other two operations:  [for (c=str(a,b,c)) ucode(c)]

When I designed OpenSCAD2, my principle aim was to eliminate all of the barriers which prevent concepts from being composed together. Being able to pass functions as arguments to functions is one example, using 'for' to iterate over the characters in a string is another example.

When you design a programming language based on the principles of orthogonality and composability, the resulting language is both simpler and more powerful.

kintel:
Restarting this.

I quickly looked at 4 "randomly chosen" languages:
* Javascript: fromCharCode() fromCharCode()
* Ruby, Python, Haskell: chr() ord()

-> Not sure if this is a pattern, but the obvious question is: Should we simply call this function `ord()` ?

doug-moen:
ord() comes from Pascal, it means cast a value belonging to an "ordinal type" to an integer (aka the value's ordinal number). So it works with characters, booleans and enums. In Haskell and Ruby, it's the same.

Python doesn't have a character type, and `ord` aborts if the argument is not a string of length 1. So it doesn't work with booleans.

Since we don't have a character type, we have strings instead, why not design the API around this fact, rather than using a distorted version of the Pascal API. Here's an alternative to consider:
* code_to_str([146,157,157]) => "foo"
* str_to_code("foo") => [146,157,157]

kintel:
@doug-moen Thanks for that piece of input!

While I agree with your sentiment, I'm a bit hesitant to introducing another way of naming such functions for OpenSCAD (at least pre-OpenSCAD2). ..in particular because we already have `chr()` which has a similar counterpart in a few other languages.
Hence, I think it would make sense to use `ord()` to name an (as exact as possible) inverse of `chr()`.

`ord()` over `asc()` simply because it has more momentum in popular programming languages today.

..but I think we agree to the point that this is merely a question of naming said functions, so we still have time for more counter suggestions :)

MichaelAtOz:
You could always depricate `chr()` - is it really has ASCII associations, at least for me.
As Unicode uses the terms Character and Code Point, you could use
``` 
aString = uchar(numbers);
aVector = ucode(string);
```

Or if you want to keep `chr()` then `code()`? But I like the above...

t-paul:
Followup in #2509...

Issue #1399, opened 2015-07-24
Refactor interface to CSG engine 


bobc:
This relates to the general objective "Implement prototypes that could be tested inside the OpenSCAD framework" of https://github.com/openscad/openscad/wiki/Project:-Survey-of-CSG-algorithms 

This is a fairly major change, so I open this as a "Request For Comments"

Goal: allow a backend CSG engine to be chosen at compile time, or none. The command line and GUI should still run as far as possible without a CSG selected.

Rationale: with a cleaner interface, it will be easier to implement prototypes. I have a prototype which uses the Carve CSG library, but it is rather hacked.

*Comment : there is already partial support for this with "ENABLE_CGAL" flag, but it has decayed*

Proposal :  
- Create a new interface layer called "CSGIF" (better names welcome!)
- Create new class "CSGIF_Polyhedron" to replace CGAL_Nef_polyhedron
- single include file "CSGIF.h" selects between no CSG, CGAL or other (eg. carve).
- Functions which are specific to a CSG engine should be collected together, this includes all the CSG ops, export functions specific to CSG engine, and OpenGL rendering.

There are some functions which aren't CGAL dependent which should be made generic, e.g. CGALCache. There are also some parts that "deep dive" into CGAL structures, these need to be genericised.

There are some functions in CGAL which may not be in other CSG engines, e.g. Minkowski, hull. These would need to be implemented in the CSGIF layer, I suggest flags such as ENABLE_HULL etc if the functions are not available.

Plan:
- I will create a csg_refactor branch in my repo, and make a PR for review. 

I am quite open to questions, comments. Anything from coding style to architecture. I am quite motivated to get this done right!

oysteinkrog:
I'm curious; what are the advantages of being able to use other CSG libraries (such as carve)?

bobc:
CGAL gives good results but is relatively slow. On a quick test I did it was 12 times slower than Carve. The Holy Grail is a backend that is fast AND gives good results, but first we need to evaluate them in practical use with OpenSCAD programs.

My end goal is actually to add support for multiple volumes with different color or material properties. To do that will require more CSG operations, so a faster backend is a must.

Faster rendering and support for color are two frequently requested features.

bobc:
I have been mulling over this PR, and I think that although it works, it is not a very elegant solution. 

[...]

PR #1401, opened 2015-07-26
Refactor interface to CSG backend

bobc:
Fix for #1399 

Please review and comment!


t-paul:
Could you check the test suite? It's currently not compiling.

For a short description see https://github.com/openscad/openscad/blob/master/doc/testing.txt

bobc:
Ok, I think I fixed the compile problems, just running the tests now. My laptop is rather slow :(

kintel:
I fully agree with the topic of this.
Did anyone test and review this yet? I'll eventually get to it, but currently quite a bit behind due to vacation.

donbright:
Some documentation would be helpful, for example what does "CSGIF" stand for? "CSGIF_polyhedron" is a mouthful to say.  I was only getting used to "CGAL Nef Polyhedron" after 4 years of staring at it. Nef comes from Walter Nef who invented the mathematical model of polyhedra that CGAL is using in the 1970s... CGAL stands for Computational Geometry Algorithms Library, which of course is the CSG backend currently being used by OpenSCAD.

bobc:
CSGIF is "CSG Interface", an interface layer between OpenSCAD and the CSG backend. I would call it just CSG_polyhedron, but the prefix "csg" is already used to refer to OpenCSG which would cause confusion.

Some design notes would be a good idea, I will put something together. I made some comments in #1399, they would have been better made here. I think an obvious improvement would be to create a class based interface for the CSG interface.

donbright:
Thanks for the link i had missed that.

hmm... I only see two files that use "CSG___" for OpenCSG, CSGTermEvaluator.h / .cc

In my humble opinion, it would be a good idea to consider the possibility of just changing that to OCSGTermEvaluator and give primary namespace of "CSG_blah blah" to your system, for example CSG_polyhedron. 

Anything else in there that is OpenCSG should named something like OpenCSG or OCSG and let your system takeover the CSG name.  (since OpenCSG never actually did CSG in the first place). Just my two cents.

Also.... if you are renaming the class itself, can you rename the files too? CGAL_Nef_polyhedron.cc can become CSG_polyhedron.cc? 

Thanks

kintel:
See #1399 - I think this needs more design work.

Issue #1425, opened 2015-08-20
Animate Filename Generation - Duplicate/Missing Filenames


DBramley:
Due to the 5 decimal place rounding used in the calculation of $t, filenames generated using the formula, round(Steps*$t), are not contiguous.

Taking two Steps values as examples:

1) Steps=996, expected output 996 frames.

1000 frames are saved but 4 of these frames are overwritten by subsequent frames with the same name. There are no obvious gaps in the output, but 4 frame files are in fact missing.

The required time increment per step is 1/996=0.001004... but this is effectively rounded to 5dp so it becomes 0.00100, which is smaller than the required time increment. When step 126 is reached, $t=0+(126-1) * 0.00100=0.125 so the file name is 0.125 * 996=124.5 which becomes "frame00125". However at the next step, 127, $t=0+(127-1) * 0.00100=0.126 so the file name is 0.126 * 996=125.496 which becomes "frame00125" again, overwritting the previous frame.

2)Steps=1005, expected output 1005 frames.

1000 frames are saved, but there are 5 gaps in the sequence of file names generated.  These are obvious on inspection of the output.

The required time increment per step is 1/1005=0.000995... but this is effectively rounded to 5dp so it becomes 0.00100, which is largerer than the required time increment. When step 100 is reached, $t=0+(100-1) * 0.00100=0.099 so the file name is 0.099 * 1005=99.495 which becomes "frame00099". However at the next step, 101, $t=0+(101-1) * 0.00100=0.1 so the file name is 0.1 * 1005=100.5 which becomes "frame00101", so file name "frame00100" is skipped.

Further details can be found here on the OpenSCAD Forum:
  http://forum.openscad.org/Bug-in-Animate-tp13528p13539.html

I believe, as doug.moen suggested, that this issue could be solved by holding the current step as an integer, Step,  $t is then calculated from this as Step/Steps, incrementing from 0/Steps to (Steps-1)/Steps. The filenames can simply be generated using Step.

kintel:
@DBramley If you have a dev setup, could you give #1431 a spin?
If not, tell us which OS you're on and we might be able to get a build going.

DBramley:
I do not have a development setup.  I am running under Windows 7 64bit. Very happy to test if a compiled build can be supplied.
Thanks.

t-paul:
[...]

PR #1431, opened 2015-09-01
Use integers for underlying frame counter. Fixes rounding issue #1425

kintel:
Could use some testing

### DIFF
--- a/src/MainWindow.h
+++ b/src/MainWindow.h
@@ -36,7 +36,8 @@ public:
 	class Preferences *prefs;
 
 	QTimer *animate_timer;
-	double tval, fps, fsteps;
+	int anim_step;
+	int anim_numsteps;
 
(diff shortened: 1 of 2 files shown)


donbright:
i was able to confirm Trygon's bug from the forum url http://forum.openscad.org/Bug-in-Animate-td13528.html#a13539 and confirm your fix on ubuntu 14.04 amd64 using Trygon's code:

```
color("black") text(str("$t = ",$t));
translate([0,-20,0]) color("black") text(str("$t * 1128 = ",$t*1128));
translate([0,-40,0]) color("black")
    text(str("Filename: frame00000 + ",round($t*1128))); 
```

Bug (master) - Frame00128.png was missing, filenames ended at Frame01128.png

Fix (issue1425) - Frame00128.png existed, filenames ended at Frame01127.png

PR #1432, opened 2015-09-05
Fix #1407 by storing vector values as heap objects (ValuePtr) rather than stack objects(Value)

### LINKED ISSUE #1407: Recursive module crash
I'm using linux and the following code makes openscad crash for me:

rec();

module rec(a=1)
{
    rec([a,10,10]);    
}

I tried many of the older recursion issues, and they were all detected properly, this one is not. 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25443257-recursive-module-crash?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

kintel:
All tests pass, but needs further testing.

One question: Are we _ever_ modifying a Value object? If so, this will break spectacularly at some point.
=> No, Values are truly immutable, so we should be safe.

### DIFF
--- a/src/Preferences.cc
+++ b/src/Preferences.cc
@@ -691,8 +691,8 @@ void Preferences::initComboBox(QComboBox *comboBox, const Settings::SettingsEntr
 	comboBox->clear();
 	Value::VectorType vector = entry.range().toVector();
 	for (Value::VectorType::iterator it = vector.begin();it != vector.end();it++) {
-		QString val = QString::fromStdString((*it)[0].toString());
-		std::string text((*it)[1].toString());
+		QString val = QString::fromStdString((*it)[0]->toString());
+		std::string text((*it)[1]->toString());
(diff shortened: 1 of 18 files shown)


donbright:
I can confirm the bug on ubuntu 14.04, 
also can confirm that the patch fixes the bug, 
also can confirm the patch has the same number of test-passes as the current master.
  
BTW on ubuntu 14.04 currently only 1187 / 1208 tests pass (98%) . (ctest -C All)

do you want to add a regression test for this bug? i can push a commit to this branch...

kintel:
Please commit to this branch if you have anything. Still need to give this some testing in the field, so it will take a few days to merge properly.

Issue #615, opened 2014-01-30
Allow command line to take a .scad file from stdin. [$5]


BenFrantzDale:
Instead of 

```
$ echo "cube([1,2,3])" > /tmp/foo.scad
$ openscad -o foo.stl /tmp/foo.scad
```

it would be nice to be able to do

```
$ echo "cube([1,2,3])" | openscad -o foo.stl
```

See also: #616.
##

ArchimedesPi:
````
$ echo "cube([1,2,3])" | openscad -o foo.stl -
````
is a more standard syntax.

**Comment by MichaelAtOz** [comment]
> is a more standard syntax.

For *nix

**Comment by runsun** [comment]
Using a blank.scad file which is blank, OpenSCAD is already capable of doing such a thing:

>>> openscad blank.scad -D "cube([2,3,4]);" -o dump.png

**Comment by littleyoda** [comment]
No need for a empty file. /dev/null works.
`openscad /dev/null -D "cube([2,3,4]);" -o dump.png`

**Comment by t-paul** [comment]
Thanks @aconz2, change is merged, bounty can be claimed.

**Comment by funrollloops** [comment]
You can also use `/dev/stdin` on Linux:
```bash
echo "cube(10);" | openscad /dev/stdin -o cube.png
```

PR #1440, opened 2015-09-23
read from STDIN when input filename is `-`

aconz2:
This may need further work as I'm not sure what the desired behavior in lines 402-405 is.

Currently compiles and tested with simple example.

References #615 

TODO:
- [ ] Add test case to the test framework
- [ ] Deal with `top_ctx.setDocumentPath()`
- [ ] Add documentation

### DIFF
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -45,6 +45,7 @@
 
 #include <string>
 #include <vector>
+#include <istream>
 #include <fstream>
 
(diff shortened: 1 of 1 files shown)


kintel:
Line 402-405, assuming you're referencing the code ending with ```top_ctx.setDocumentPath(...)```, is to set the path used when including other files. For stdin, I guess we should set this to the cwd.

PR #1462, opened 2015-10-23
Enable compilation without OpenCSG.

GilesBathgate:
This is just a patch so that the ENABLE_OPENCSG compiler define works again, should someone want to compile without OpenCSG for whatever reason.


donbright:
Tests OK on ubuntu 14.04

PR #1463, opened 2015-10-23
QOpenGLWidget

GilesBathgate:
Since Qt 5.4 the QOpenGLWidget seems to be an almost drop in replacement for QGLWidget. The new widget is much more responsive. For a test (on Qt5) with the old QGLWidget try:

sphere(10,$fn=100);

Even with the QTBUG-39370 workaround in place the Gui is Laggy. However with QOpenGLWidget the gui is responsive again.


t-paul:
Nice, I'll have a look how that behaves on my Debian system. The slowdown seemed to be most noticeable on Linux.

GilesBathgate:
Ideally all the gl functions should be called through the new cross platform interface:
http://doc.qt.io/qt-5/qopenglwidget.html#opengl-function-calls-headers-and-qopenglfunctions
However this would take significantly more changes in respect to OGL_Helper.h

kintel:
@GilesBathgate Did you ever figure out if Qt supports mixing their own OpenGL interface with GLEW?

GilesBathgate:
@kintel 

On GNU/Linux Qt is linked against the OpenGL implementation directly, so in that case QOpenGLWidget is interoperable with glu/glew/glut etc. 

This situation on Windows is that Microsoft ostracised the OpenGL standard, and push their own competing standard, by providing only OpenGL 1.1 function calls in its opengl32.dll, and forbid any GPU vendor with replacing that DLL with a driver-provided one exposing more functionality. Vendors then worked around that absurd limitation by using a small piece of the GL standard which allows for extensions to be exposed and function entry points of extensions to be resolved at runtime.

By default on windows since Qt 5.4, Qt does not link against opengl32.dll, or glu32.dll, so if you want to use native glu, you need to at least add ```LIBS += -lopengl32 -lglu32``` to the .pro file, and also ```QCoreApplication::setAttribute(Qt::AA_UseDesktopOpenGL);``` should be called early inside main.c
However even after doing so I could not get it to work.

The other option is to port all the GL calls to go through a versioned QOpenGLFunctions wrapper (such as QOpenGLFunctions_3_2) but this means either dropping glu/glew/glut or coming up with a clever hack to make those libraries wrap QOpenGLFunctions. 

For now, the old QGLWidget works on windows, so it could just fall back to using that.

I don't know what the situation regarding OpenGL is on macosx, I presume similar to GNU/Linux

Regards
Giles

kintel:
@GilesBathgate Thanks for clarifying. Based on that, my understanding is:

* GLEW requires us to link agains OpenGL, resolving GL functions by calling the linked entry points to the library.
* Qt decided to do this dynamically, by actually manually locating and opening the DLL, taking out the OpenGL problem from the app developer's linker: https://github.com/qtproject/qtbase/blob/d8f41bdc2efa36dd9e2f80e23d7cb4874c019452/src/plugins/platforms/windows/qwindowsglcontext.cpp#L159

..the result being that GLEW doesn't necessarily know how to access the GL functions.
Projects like glbinding (https://github.com/cginternals/glbinding) do smth. similar as Qt.

Needs more digging, I just wanted to share my thoughts..

GilesBathgate:
@kintel I am experimenting with something similar to this: http://pastebin.com/SW7KQnNB

only on windows the syntax has to be like this for some reason:

```
typedef float GLclampf;
typedef unsigned int GLbitfield;

extern "C" {

void _imp__glClear(GLbitfield mask)
{
}
...
```

GilesBathgate:
@kintel  This blog article describes the decision Qt Team made more succinctly https://blog.qt.io/blog/2014/11/27/qt-weekly-21-dynamic-opengl-implementation-loading-in-qt-5-4/

donbright:
to summarize the link for future reference... it appears there are at least four possible GL backends for Qt5  on Windows... 

1. graphics-vendor supplied opengl drivers (nvidia,amd), called as "desktop opengl"

2. the ANGLE system which runtime translates OpenGL to DirectX and bypasses graphics-vendor OpenGL drivers completely in favor of graphics-vendor DirectX drivers. (used by firefox and chrome for WebGL). 

3. the ANGLE +WARP system, same as 2, but output to DirectX software rasterizer, Microsoft WARP

4. Mesa3D GL software rasterizer (which of course bypasses all graphics drivers). It has several different engines available, the fastest is LLVMPIPE.

GilesBathgate:
@donbright For my project, option 1) doesn't seem to work. When I start-up the application the whole UI is a black window?! 

I have however managed to solve the problem by creating a GLWrapper DLL. I have now published this as a project on Github https://github.com/GilesBathgate/glw It needs to be extended to support all of the OpenGL1.0 api calls (perhaps even OpenGL1.1) since openscad uses many more calls via glew/glu etc. But it does work in principle.

GilesBathgate:
This might also be relevant to the cause:

https://www.kdab.com/opengl-in-qt-5-1-part-1/

GilesBathgate:
@t-paul There are a number of outstanding issues with this approach.

1) It still doesn't work for my project (despite thinking it did, it only worked on the windows environment on which it was compiled)

2) OpenCSG requires OpenGL2.0, glwrapper currently only wraps the QOpenGLFunctions_1_0.
Granted that GLEW loads the extensions as needed, but I don't know how well this will play with Qt.

3) No noticeable performance speedup was noticed by @t-paul on his system, so this may only be a worthwhile investigation for future compatibility if and when Qt drop QGLWidget.

kintel:
Anyway, I think we should merge this patch with one change: Add a compile-time macro to turn this _on_. That would enable us to tinker with this in master and gradually enable it per platform.

kintel:
@GilesBathgate Won't your code partially skipping```running_under_wine``` fail to build under Windows? What's the purpose of that code?

kintel:
I see some rendering differences. Any idea what could cause this?

QOpenGLWidget:
screen shot 2015-12-02 at 13 21 40 pm

QGLWidget:
screen shot 2015-12-02 at 13 21 42 pm

GilesBathgate:
@kintel regarding ```running_under_wine``` it seems to be hard-coded to false. Historic reasons?

```c++
+#if !(QT_VERSION >= 0x050400)
 static bool running_under_wine = false;
+#endif
```

As for the rendering differences, my only guess is that its something different in the way QOpenGLWidget might be handling something different in the lighting code called in the InitialiseGL() function?

GilesBathgate:
@kintel ```running_under_wine``` is set by init(), don't know how I missed that before...either way, QOpenGLWidget has no such method ```SwapBuffers()```

GilesBathgate:
@kintel I have fixed problem with the reference to ```running_under_wine``` in the init() function.

PR #1469, opened 2015-10-30
new feature - interactive adjust of numerical value in source code

josefpavlik:
Added feature to allow adjusting numerical value in the source code and see the effect in real time.

When the cursor is on a number, you can increment or decrement it with Alt-Up or Alt-Down keys
and see the change of model in real time.
The Alt-Left and Alt-Right keys navigates to left or right within the number and allow you to change
the increment value. Navigating to right over the number, new decimal digits are added automatically.

When the update of preview is slow and you change the number quickly, the frame rate is limited, however when you stop to increment, the actual model is shown as soon as possible.


t-paul:
Very cool. Works fine for me (on Linux).

josefpavlik:
note - works only with qscintilla editor and I've tested it only on Linux. Hopelessly my mac is old, no way to compile it for snow leopard...

t-paul:
Maybe the feature should be suppressed when the cursor is not placed at a number? Currently it inserts a 1 with `ALT+UP` if there's no number at all.

josefpavlik:
Yes, the idea was that the feature works only on  a number, seems that there is a little bug. I'll fix it.

josefpavlik:
I've add the check if we are on number before incrementing it.
Also I've add new feature - when you press the alt-backspace (undo) immediately after interactive modify of number, without releasing the alt key, the undo invokes refresh of preview of the model. However, if you release the alt key, the alt-backspace works normally, without invoking the preview.

kintel:
@josefpavlik Looks good. I fixed a mac issue - see your pull requests.
Also, I'm not sure about the alt-backspace function. Alt-Backspace isn't generally bound to Undo - that's just a convention on certain platforms. Usually, undo is Ctrl-Z.

josefpavlik:
the alt-backspace is assigned to undo on Ubuntu. The only change that I've made is, that after pressing the alt-backpace (immediately after alt-arrow), the refresh of preview is invoked. Changing the number and seeing the real time preview is natural, that when you strike backspace and the number under cursor changes, that the preview is refreshed. However if the alt-backspace is not bound to undo but is bound to anything else (or not bound at all), the only change is, that the preview is refreshed. 
When you press alt-backspace without using alt-arrow feature immediately before, nothing changes respect original program.

kintel:
The default keybinding on Mac uses Alt-Backspace to delete the previous word, so this feature needs to be aware of the current keybinding layout. Of course existing Alt-<arrows> are also already bound.
Perhaps we need to take a closer look at the keyboard layout situation.

josefpavlik:
so in this case, when you press alt-backspace immediately after alt-key, the previous word will be deleted and the preview of the model will be updated..

kintel:
@josefpavlik That's correct. ..so I guess the right fix is to not look for alt-backspace, but look for the key-combo bound to undo, and use that one to trigger a refresh.

josefpavlik:
@kintel I've rewrote the undo/redo preview handling and I've improved the undo during interactive mode.

Undo now undoes all consecutive arrows during the same alt session.  I.e. press alt, arrow, arrow, arrow arrow, depress alt will be undoed with one only ctrl-z or alt-backspace. Any key other than arrow breaks the group. Depress alt, press alt is sufficient to break the group.

When you invoke undo immediately after alt-arrows, preview will be automatically invoked after the undo is executed. You can make more undo or redo, the preview will be refreshed, but when you press any other key than undo or redo, the auto preview will be disabled.

Does not matter the key assignment, this should work for any operating system and works also with alternative assignment, so works either with ctrl-z and alt-backspace on Linux. I'm testing the assigned command, no some fixed key combination.

kintel:
@josefpavlik Looks great - thanks for hanging in there!

josefpavlik:
Thank you for the comments and for the merging.

t-paul:
Very nice, I'll try to find a way for putting that into the documentation. A mini screencast would be cool, but I'm not sure Wikibooks supports that.

t-paul:
What about http://files.openscad.org/tmp/OpenSCAD-interactive-number-adjustment.webm ?

josefpavlik:
nice video

Issue #1191, opened 2015-01-28
Add angle arguments to rotate_extrude()


clothbot:
Enhancement request, prompted by discussion http://forum.openscad.org/How-to-create-an-cylindrical-arc-td11257.html

Please add angle arguments to rotate_extrude() to support partial-arc extrusions:

Hypothetical usage:

module hollow_square(offset=10,a=45,w=5,th=1) translate([offset,0]) rotate(a) difference() {
  square(w,center=true);
  square(w-2*th,center=true);
}

// rotate_extrude from a=0 to a=30
rotate_extrude(a=30) hollow_square();
// rotate_extrude from a1=45 to a2=135
rotate_extrude(a1=45,a2=135) hollow_square();

kintel:
See also http://forum.openscad.org/partial-rotate-extrude-tt2695.html

We really only need one angle argument as it's trivial to add a rotate(). This also helps with orthogonality.

foulowl:
I would love to have this feature as well. Thank you!

thehans:
This can be closed since the merge of PR #1478

Issue #1282, opened 2015-03-24
rotate_extrude() of 2D object in the negative X half plane results in an inside-out object


kintel:
While this is not a common workflow, people tend to get into this situation. This should be easy enough to fix. Testing needed.

TakeItAndRun:
There are two solutions:
1) ignore everything less than x<0
2) reflect the negative half plane into the positive half plane and rotate_extrude
I'd prefer 2).

kintel:
I think 2) makes sense. Of course keeping the current limitation that objects crossing the Y axis are not going to make OpenSCAD happy.

atartanian:
i can think of another solution, which would be to clip the polygon along the y axis, flip it across the y axis and then union it with the part of the shape that is on the +x side. this results in the same shape you get if you have an object crossing the y axis, but without the inverted faces

atartanian:
which is basically #2 but with cleaning up the polygon before rotating it

thehans:
This is handled in the recently merged PR #1478 and can be closed.

PR #1475, opened 2015-11-10
Angle parameter for rotate_extrude

thehans:
This is a proposed patch for Issue #1191 

It works correctly as far as I can tell, but this is my first dive into openscad code.

### DIFF
--- a/src/GeometryEvaluator.cc
+++ b/src/GeometryEvaluator.cc
@@ -792,6 +792,26 @@ static Geometry *rotatePolygon(const RotateExtrudeNode &node, const Polygon2d &p
 {
 	PolySet *ps = new PolySet(3);
 	ps->setConvexity(node.convexity);
+	if (node.angle <= 0) return ps;
+
+	if (node.angle != 360) {
+		PolySet *ps_start = poly.tessellate(); // starting face
+		Transform3d rot(Eigen::AngleAxisd(M_PI/2, Vector3d::UnitX()));
+		ps_start->transform(rot);
+		// Flip vertex ordering for starting polygon
+		BOOST_FOREACH(Polygon &p, ps_start->polygons) {
+			std::reverse(p.begin(), p.end());
+		}
+		ps->append(*ps_start);
+		delete ps_start;
+
+		Polygon2d end_poly(poly);
+		PolySet *ps_end = end_poly.tessellate();
+		Transform3d rot2(Eigen::AngleAxisd(-node.angle*M_PI/180, Vector3d::UnitZ()) * Eigen::AngleAxisd(M_PI/2, Vector3d::UnitX()));
+		ps_end->transform(rot2);
+		ps->append(*ps_end);
+		delete ps_end;
+	}
 
 	BOOST_FOREACH(const Outline2d &o, poly.outlines()) {
 		double min_x = 0;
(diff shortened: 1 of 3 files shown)


kintel:
Thanks!
There are a few things needed from here, primarily tests and documentation.
See https://github.com/openscad/openscad/blob/master/doc/testing.txt and https://github.com/openscad/openscad/blob/master/doc/hacking.txt

I'll come back in a while and add a checklist to the pull request, so we can get it all merged.
If you want, we can temporarily merge this into a feature branch if you don't want to manage pull requests no this feature yourself.

thehans:
I noticed it failed pretty much every rotate_extrude test.  I haven't looked into how the tests work yet, but I'm guessing it doesn't like that I flipped the starting angle.

Lines 835, 839, 841 would have had -M_PI/2 in them, i made it positive.  My reason for this was that I noticed it was starting the arc on the negative side of the x axis rather than the positive.

Also I wasn't sure for partial angles, which direction to rotate.  Currently it goes clockwise when viewed from the top(linear_extrude with twist seems to do this also), but counterclockwise from X+ is how I've always pictured angles in cartesian grids.

kintel:
I think we should rotate counter-clockwise, starting from the positive X axis.
Some parts of OpenSCAD violate some default convention, but let's fix those later.

thehans:
I changed the default rotation to counter-clockwise as discussed.  

I noticed Issue #1282 while I was at it, and addressed that as well.  Geometries in the negative X side now start their arc from that side.

The code also now allow for negative angle inputs, which arc from the X axis in the clockwise direction.
Here is a quick test I used to verify faces for all combinations of positive/negative inputs.

```
module face(x) {
  translate([x,0]) difference() {
    square(5,center=true);
    circle(2);
  }
}
rotate_extrude(angle=45) face(10);
rotate_extrude(angle=45) face(-10);
rotate_extrude(angle=-45) face(20);
rotate_extrude(angle=-45) face(-20);
rotate_extrude() face(30);
rotate_extrude() face(-40);
```

thehans:
OK, so I am more confident now that the code is performing correctly.  

There are two types of test errors to deal with.  

1) The image tests fail because as mentioned before they are starting the rotation from the +X axis.  This is really apparent in the $fn=3 case for example, or other odd numbers.  Should I just commit my new generated output pngs into the tests/regression path?

2) The dump tests fail since there is a new parameter in the text.  One thing that is still possibly open to change is the ordering of the parameters.  Teepee had mentioned in irc that it might be worthwhile to position angle as the first param for convenience, but this would technically break backwards compatibility.  Though i noticed the whole file parameter thing is deprecated anyways.  Even disregarding the dxf file parameters, I currently have angle placed after convexity.

Here is what my code is outputting at the moment.  What do you think should be done?

    // example with file parameter
    -rotate_extrude(file = "open-polyline.dxf", layer = "", origin = [0, 0], scale = 1, convexity = 2, $fn = 0, $fa = 12, $fs = 2);
    +rotate_extrude(file = "open-polyline.dxf", layer = "", origin = [0, 0], scale = 1, convexity = 2, angle = 360, $fn = 0, $fa = 12, $fs = 2);

    // example without file parameter
    -rotate_extrude(convexity = 3, $fn = 0, $fa = 12, $fs = 2) {
    +rotate_extrude(convexity = 3, angle = 360, $fn = 0, $fa = 12, $fs = 2) {

Additionally there are two tests that failed locally which are omitted from travis CI
411 - opencsgtest_rotate_extrude-tests (Failed)
This has the above mentioned difference of starting on +X axis

465 - opencsgtest_issue1165 (Failed)
this has some z fighting artifacts unrelated to rotate_extrude, which i'm assuming is just difference in graphics cards or something.

kintel:
Moved to #1478 to facilitate collaboration

PR #1478, opened 2015-11-12
Angle parameter for rotate_extrude

kintel:
See original PR at #1475.
This is a proposed patch for Issue #1191.
TODO:
* [x] Positional parameter for angle? -> No immediate need
* [x] Verify that failing tests can be recreated due to rotation starting point
* [x] Verify that opencsgtest_issue1165 is harmless
* [x] General test cases
* [x] Test 0 angle
* [x] Test negative angle
* [x] Test inf/-inf angles
* [x] Test NaN angle
* [x] Documentation. Rudimentary doc: https://en.wikibooks.org/w/index.php?title=OpenSCAD_User_Manual/2D_to_3D_Extrusion&stable=0#Rotate_Extrude
* [x] Cheat sheet
* [x] Optional: Updated example


thehans:
All the rotate_extrude image based tests are based on this in this location:
`openscad/testdata/scad/3D/features/rotate_extrude-tests.scad`
Since geometries entirely on the -X side also start their rotation from that side, this is equivalent to how the previous code worked and what the tests expect.  To show this I added mirrror([1,0]) to the child of each rotate_extrude call in this scad file, and reran the tests, and all the image based ones passed(besides issue1165 test).
[rotate_extrude-tests.scad.txt](https://github.com/openscad/openscad/files/32901/rotate_extrude-tests.scad.txt)

Below is the output for opencsgtest_issue1165, which doesn't call rotate_extrude at all.  I'm assuming Travis server generates a similarly non-passing output and that's why its disabled anyways.
issue1165-actual

Finally is a test file I've used which covers pretty much every case for rotate_extrude with angles, and its resulting image from my local build[test.scad.txt](https://github.com/openscad/openscad/files/32900/test.scad.txt)
screenshot from 2015-11-12 07 23 23

kintel:
@thehans I think that some people may exploit the current rotate_extrude() behavior in ways which may break their designs when using your implementation.
e.g., someone may have decided to create a hollow triangular tube this way:
``` rotate_extrude($fn=3) translate([2,0]) square([1,5]); ```

If the tube suddenly change direction, this could mess up their design.
I would suggest that we keep the existing behavior if ```angle``` is not specified, but use the more mathematically correct one when given an ```angle```.

Other ideas?

kintel:
We should probably disable ```opencsgtest_issue1165``` due to platform dependent output.

thehans:
OK I switched it back to previous behavior for angle==360.

I'm holding off on updating the dump test expected results since I'm not sure what you want in regards to parameter ordering

currently it goes after convexity, which goes after file dxf params.
copy from previous PR comment:

    // example with file parameter
    -rotate_extrude(file = "open-polyline.dxf", layer = "", origin = [0, 0], scale = 1, convexity = 2, $fn = 0, $fa = 12, $fs = 2);
    +rotate_extrude(file = "open-polyline.dxf", layer = "", origin = [0, 0], scale = 1, convexity = 2, angle = 360, $fn = 0, $fa = 12, $fs = 2);

    // example without file parameter
    -rotate_extrude(convexity = 3, $fn = 0, $fa = 12, $fs = 2) {
    +rotate_extrude(convexity = 3, angle = 360, $fn = 0, $fa = 12, $fs = 2) {

t-paul:
The order in the CSG file output is not very important. As all parameters are named, the order is not used to assign the values. (I probably would put `angle` as first parameter as it's the most important for the user, but that's just a matter of taste).

The order would become important if we would want to support the short-cut `rotate_extrude(30)` instead of `rotate_extrude(angle = 30)`. As currently `file` is the first positional parameter (and we can't change that without breaking backward compatibility, even though `file` should be deprecated for some time now), that would need some sort of hack similar to what `linear_extrude()` does in https://github.com/openscad/openscad/blob/7060b2582796e194b269a1f829f2f60ba992cd87/src/linearextrude.cc#L84 

I guess `angle` should not part of the `AssignmentList` value at all, so by default it's not possible to give it as positional parameter, only as named. If we decide to add the hack/workaround it could still be set if it's not defined by name and the first positional parameter is a number.

kintel:
I agree with @t-paul: Making the angle argument positional could be interesting but not necessary. Having it first in the argument list makes sense, for readability as it's the primarily interesting parameter.

thehans:
All the checkboxes are ticked in my mind except for documentation and cheat sheet, not sure how those are handled.

kintel:
@thehans Documentation is updating the wikibooks docs (https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/2D_to_3D_Extrusion#Rotate_Extrude), cheatsheet is in git: https://github.com/openscad/openscad.github.com/tree/master/cheatsheet

thehans:
@kintel Won't it be confusing to document this feature on the wiki before its available in any release though?  Should that just wait until later?  I don't know what the release schedule is like.

kintel:
@thehans Usually, we tag such feature in the doc, see e.g. https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/List_Comprehensions.
If we haven't released a version yet, we tag it with the milestone containing the functionality; in this case "2016.XX".

We don't currently have a good solution for the cheatsheet. It would be cool to have a dev vs. release version online. If we don't update cheatsheet, we should at least create an issue or put a note somewhere to make sure we remember to update it before next release.

kintel:
Thanks @thehans !

renschler:
re @thehans 's comment above. As a new openscad user I found this confusing. I tried using the angle parameter and it took me a while (about 20 min playing around with stuff until i found this issue) to realize that the documentation was describing *future* functionality, not existing functionality.

PR #1485, opened 2015-11-13
Support alpha channel for transparent backgrounds

thehans:
This allows color-schemes to be specified with an alpha channel, and for exported png's to include that alpha data.

Currently all the color scheme values are 6 hex chars for the r,g,b bytes.  It now supports optional 4th byte.  So background is made transparent like so: 
`    "background" :         "#00000000",`

This makes it possible to export animations with transparencies for creating nice gifs or apngs.

Like this:
openscad_gif3
If image is viewed on another html page with different color background, that will show through.


kintel:
Looks good! I'll add a few TODO items.

donbright:
it's very cool.... but does it interact OK with the test suite? 

i remember choosing to avoid Alpha in the offscreen PNG export (see comment) because of possible issues with portability.... although of course i could have been wrong.

thehans:
>Setting alpha to 0 causes black background in the GUI. We should probably communicate the intended background color to Qt

In my example I set the entire color to transparent black, but setting to other colors with 00 alpha actually still works for transparent png export, and still displays them in the gui as if they were solid, normal color.  I can set to #ffffff00 for white background, etc.  Does this behave differently on your computer?

The background, or even the model colors can also be partial transparent.  It all works nicely.

>Include one built-in colorscheme with transparent background.

Any preferences/ideas for colors to use in a new built-in color scheme?

> it's very cool.... but does it interact OK with the test suite?

don, all tests have passed for me locally and on the CI server without issue.  I don't have a mac to test on though.

kintel:
@thehans I used "#f8f8f800" as background and it resulted in black. This might be related to different management of GL-based widget in Qt on different platforms.

kintel:
In terms of built-in color schemes; I don't have any preferences, but as this might be come a useful feature, I would tend towards basing it on the default color scheme - just for recognition purposes.

donbright:
OOOOh i got it to work from command line... really interesting.  I modified color-schemes/render/"tomorrow night" .json to be         "background" :         "#f8f8f800",
 
     echo "cube();" >> x.scad
    ./openscad x.scad --colorscheme="Tomorrow Night" -o x.png


delme

However when I do 'file / export as image' from the GUI the transparency is not preserved? It's just a flat color as displayed in the render window. f8f8f8. (however, on the plus side, i dont see the all-black that Marius is seeing). 

I am on linux ubuntu 14 using qt5 .

kintel:
Further testing on Mac: Neither cmd-line nor Save as Image works. In both cases I get the background color without alpha. GUI still gives black background if alpha is 0. Any other value for alpha gives the specified background RGB color, as if alpha wasn't there.

thehans:
Embarrassingly, I didn't even know about the File ->"Save as Image..." feature.
I can look into that one.  I had been testing mainly with Animate -> "Dump Pictures" feature. (Though I did also test command line -o flag with explicit colorscheme).

Marius, for your black background issues, this is only on Mac?   In your Help -> Library Info does it show alpha enabed?  "RGBA (8888)",  On Linux, prior to my commit it was showing me "RGBA (8880)".  That's what the openscad/src/openscad.cc edit addressed.
I noticed the windows release showed RGBA (8888) already though.

kintel:
@thehans I've only tested on Mac. Both the GUI and cmd-line shows RGBA (8888).

I think that the blackness problem is related to widget compositing. Since "Save as Image" renders the RGB color correctly, that color somehow made it into the widget. Perhaps there is either a black background shining through or somehow the widget getting separately cleared with black.

In terms of alpha not being exported, that sounds like a different issue. perhaps Qt doesn't preserve alpha when reading from the widget's framebuffer.

Perhaps this is easier debugged in a stand-alone Qt-OpenGL example app..

kintel:
..and: the written png file doesn't even have an alpha channel.

nophead:
The old bug with alpha, which I think was fixed by not propergating the
alpha to the png, was that transparent objects in a scene that showed what
was behind when on the screen were written as transparent in the png. That
meant they showed as checkerboard in an image viewer.

thehans:
Hrm, now that you mention in, the test outputs (for example some of the thrown together ones with transparent objects), are showing checkerboard through.  I guess the tests don't really compare transparency between the expected and actual since everything still passes.

What about going through the image buffer between grabbing the fram and writing to PNG and altering the pixels in a loop.  If the pixel R,G,B channels don't match the background color then set the alpha to 1?

nophead:
Well conceivable you could have something that wasn't background with the
same colour as the background. Would be odd but it isn't an algorithm that
works in every case.

thehans:
OK how about this.  Pass over the pixels, if alpha != 0 then set to 1.  This would mean background can either be fully opaque or fully transparent, no partials.  I consider this an acceptable compromise.  I would mostly use it for animated GIFs that only have 1bit alpha anyways.

nophead:
The issue I think is the partially transparent object that are not
background. They seem to need the alpha channel removing before export. I
don't really understand why. When an object is drawn with alpha I would
expect it to be blended with the existing contents of the frame buffer but
I wouldn't expect the result in the frame buffer to have an alpha value.
However the original bug seemed to imply that frame buffer ended up with
alpha. That should only be the case when the background has alpha in my
naive view.

thehans:
@nophead Yes, the semi-transparent objects, by definition have a non-zero alpha.  Therefore forcing any pixel with non-zero alpha to a full alpha value after grabbing the frame is how I work around the issue.  This fix assumes that if you want transparent background, you will set it to 0 alpha.  That is the only value that will be output as transparent.

There are already many issues with object transparency such as order of rendering that are their own independent issues that I don't quite feel qualified to address.  I don't know enough to explain *why* the objects make the PNG semi-transparent exactly, this is the best I can do to address it "in post" though.

I just checked in the fix by the way, so try it out if you like.  The tests passed on my computer, I expect they will pass on the travis CI when that finishes. (Also, small note that the 2nd to last commit I made was to make the test server actually fail on images with alpha channel differences, so I know its actually testing for them).

nophead:
Semi transparent objects should be drawn into the frame buffer with with
alpha enabled but the resulting pixels should be a blend of the object's
colour and whatever is behind. That should result in a solid colour unless
the background is transparent. It seems that opencsg is rendering them
incorrectly if it is leaving the alpha value not equal to one.

thehans:
OK, more progress, I decided to really look into the blending.  I have reverted my "workaround" changes.

OpenGL blending has always confused me but this little demo tool really helped me understand a bit better:
http://www.andersriggelsen.dk/glblendfunc.php

The blend setting that was being was this:  
`glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);` in GLView.cc
This works fine for blending RGB values based on alpha, and is the common way to handle transparency, but it leaves the alpha channel messed up.

What is needed is to be able to process the colors and alpha channel separately:
`glBlendFuncSeparate(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA, GL_SRC_ALPHA, GL_DST_ALPHA);`
This way the background's alpha doesn't get overwritten.

I've been testing this, and it seems to work fairly well, except for the case where a semi-transparent object(such as a '#' highlighted one) is involved with a difference operation.  I haven't tracked it down yet, but it seems that the rendering code is sort of overriding the blending function in this case.

It seems to only affect the cut faces.

screenshot from 2015-11-17 09 48 19

thehans:
If an opaque object is cut by a semi-transparent one, then only the cut faces have incorrect alpha.

However if an object being cut is semi-transparent itself, then all of its faces appear to be affected.

screenshot from 2015-11-17 10 05 29

thehans:
After some discussion in IRC, we determined that problem stems from OpenCSG which uses the alpha channel to compute visibility.  This is an issue in OpenCSG 1.3.2 using the default SCS algorithm.

Upgrading to OpenCSG 1.4.0 adds support for an alternate method that does not alter alpha, as long as the graphics card supports `GL_ARB_fragment_program`.  All tests pass using this latest version.

Unfortunately Travis CI is still on 1.3.2 so these tests won't pass there.  Not sure if we can update the test server with the latest OpenCSG or what the best solution might be.
It would still require people like me whose distro is still on 1.3.2 to build and install their own updated 1.4.0.

For reference I'm on Linux Mint 17.2, which uses Ubuntu 14.04 repos.

kintel:
I believe we get OpenCSG from the PPA: https://launchpad.net/~openscad/+archive/ubuntu/releases
Perhaps we could convince @hyperair or @chrysn to update it?

chrysn:
opencsg has been in debian and ubuntu for quite some time, i'd consider the opencsg package in the ppa more of a relic from before it was in the distributions. is upgrading the ci server to wily (current stable ubuntu) not an option? if you do need it backported: which ubuntu version are you running on?

kintel:
The CI server (Travis) is running Ubuntu 12.04. They currently have 14.04 in beta, so hopefully we can switch to that one eventually. For the 12.04 I thought we were still using the PPA.

thehans:
Even 14.04 still has OpenCSG 1.3.2 in its repos.  So it would still require a PPA or something to get 1.4.0  

http://packages.ubuntu.com/trusty/libs/libopencsg1

kintel:
@thehans Do you think we should disable this feature for OpenCSG < 1.4 ?

thehans:
I guess it would probably be best to disable, not sure what the best way to handle the version detection.

By the way, the fix for Issue #1495 would partially address this on 1.3.2

Since that will make the cut faces be drawn opaque even though opencsg would have cleared the alpha channel on those parts.  

However, in its still a problem in cases where a highlighted object is cut into.
screenshot from 2015-11-19 11 26 40

thehans:
OK, I made the png alpha export conditional based on OPENCSG_VERSION.  Tests are passing again on Travis CI.

However I don't know how I would run a test conditionally (the colorscheme-transparent test that I would need to add for this feature) based on the library version.

Also without the fix for Issue #1495, the expected result for transparent background would look like this:
screenshot from 2015-11-19 16 18 33

I could include the fix in this branch, but I figured it might be better to separate the issues.

kintel:
To run a test conditionally, I think I would try to extract the OpenCSG version during cmake time, and disable the test if OpenCSG < 1.4

thehans:
Test case for transparent colorscheme has been added.  It will fail on the test server since alpha is disabled on OpenCSG < 1.4  

I'm don't know cmake well enough(or at all) to make it disable a test based on library version.

I altered the imagemagick calls to fill transparent image parts with a magenta (#FF00FF) background color before performing the difference operation.

If transparent background is not filled in and imagemagick is simply set to `-alpha On`, comparing an image containing transparencies against *itself* results in a failure.  Here is an example, the right image is a result of differencing the left image with itself.
screenshot from 2015-12-03 18 33 11

thehans:
Looks like that last change completely broke image comparison on the test server, which is using "old" imagemagick.

My local machine has imagemagick version `8:6.7.7.10-6ubuntu3` and handles it fine.

Not sure yet whether it doesn't like the `-alpha Remove` setting or the `-background magenta` setting, or both.

kintel:
FYI: The test server runs Ubuntu 12.04

thehans:
OK I took the hint and set up a 12.04 VM.  Found another imagemagick approach that does the same thing but appears to be more backwards compatible.  All tests are now passing on the test server except the transparent one, as expected.

thehans:
Changed the blend func again for alpha.  This time I am quite confident it is correct.  

Now using:
`glBlendFuncSeparate(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA, GL_ONE, GL_ONE_MINUS_SRC_ALPHA);`

This time I actually derived the equation for setting the blending factors on the alpha components.  Below is my thought process.

Lets say every object has a transmittance value "T" that tells you what percentage of light is able to pass through the object.  Light<sub>in</sub> * T = Light<sub>out</sub>

Now lets say we have a surface1 and surface2, light will pass through both.  The total transmittance is equal to the product of individual transmittances.
Light<sub>out</sub> = Light<sub>in</sub> * T<sub>total</sub> = Light<sub>in</sub> * T<sub>1</sub> * T<sub>2</sub>

Now, since we are working with alpha(opacity) instead of transmittance, lets substitue that in.
A = 1 - T
T<sub>total</sub> = T<sub>1</sub> * T<sub>2</sub>
A<sub>total</sub> = 1 - T<sub>total</sub> = 1 - (T<sub>1</sub> * T<sub>2</sub>) = 1 - ((1-A<sub>1</sub>)(1-A<sub>2</sub>)) = 1 - (1 - A<sub>1</sub> - A<sub>2</sub> + A<sub>1</sub>A<sub>2</sub>) = A<sub>1</sub> + A<sub>2</sub> - A<sub>1</sub>A<sub>2</sub>

Now instead of calling them surface 1 and 2, lets call then dst (the background existing in the framebuffer), and src (the new object being drawn on top)
A<sub>total</sub> = A<sub>src</sub> + A<sub>dst</sub> - A<sub>src</sub>A<sub>dst</sub>
or the same equation written slightly differently:
A<sub>total</sub> = A<sub>src</sub> * 1 + A<sub>dst</sub> * (1 - A<sub>src</sub>)
Now we finally see the blending factors of "1" and "1-A<sub>src</sub>" which the src and dst alpha values are scaled by respectively, and which are represented in the two last parameters of glBlendFuncSeparate.

s-ol:
Whats the status on this?

kintel:
@thehans Feel like trying to revive this? It's a bit tricky to get it tested and validated on all platforms, but we should be able to deselect Qt4 which may make it easier.

thehans:
@s-ol, @kintel I'll try to take a look in the next few days.  

Would be nice to wrap up my oldest still-open PR.

whosawhatsis:
I'd also REALLY love to have this working.

thehans:
I think the previous tests I did on this were all using the old `QGLWidget` which handles alpha differently than `QOpenGLWidget` used by Qt5.4+

My biggest problem now is that there's been a bug in ```QOpenGLWidget::grabFrameBuffer``` which has been ignoring alpha since the widget was introduced.  

It seems that the issue has been fixed for Qt 5.12 and up, but my dev machine is using Linux Mint 19.2 (based on Ubuntu 18.04), which uses Qt 5.9.5 still.

The fix can be seen here:
https://codereview.qt-project.org/c/qt/qtbase/+/235928
And its associated bug (which I just commented in earlier):
https://bugreports.qt.io/browse/QTBUG-55245#

I started thinking maybe I could write some wrapper for pre-5.12 Qt versions, which overrides grabFrameBuffer to implement the above fix, but its doing all the work inside some QOpenGLWidget**Private** class so I don't know anymore if wrapping it in such a way and overriding is even possible.

thehans:
Some more details on the Qt5 bug, its only an issue when exporting from GUI, since command line should use `OffscreenView` instead of Qt widgets for grabbing frames.

So Travis and Appveyor *tests* should be fine (as soon as I fix the expected results, unrelated to current problem discussion).

So the question becomes: which of all our various builds are using sufficient Qt5 versions for this to work?  I tried to document all the ones I'm aware of:

- [X] (GOOD) [Circle CI MXE](https://github.com/openscad/mxe/blob/master/src/qtbase.mk) Looks to be using 5.13
- [X] (GOOD) CircleCI AppImage: I downloaded a recent build and saw `Qt version: 5.12.3` in Library Info.
- [X] (GOOD) flatpack: t-paul says this is on 5.13.2
- [ ] (PROBLEM) Snap package is stuck on Qt 5.5.1 as far as I can tell.
    - There is [preliminary support](https://github.com/snapcore/snapcraft/releases/tag/3.9) for `core20`, which I'm guessing should have new enough Qt packages that we could test, until that is marked stable (presumably around April if snapcraft is roughly on schedule with ubuntu release?)
    - @t-paul found "one post suggesting it's possible to just dump downloaded *.deb files into the build env", so that could be another option, by using [this PPA](https://launchpad.net/~beineri/+archive/ubuntu/opt-qt-5.12.6-xenial)
    - I asked in #snapcraft IRC and they mentioned this [ongoing proposal/discussion to improve alternative package sources](https://forum.snapcraft.io/t/proposal-additional-package-sources/2199/4)
- [ ] (PROBLEM) OBS builds:  There are **many** of these for different distros, I don't know the various Qt version details besides 18.04 still using Qt 5.9.5.
    - I don't think there's a good way to solve this for all supported distros besides documenting that GUI-based image exports won't work for those built on Qt < 5.12.  
    - Command line exports should still be usable at least, regardless of Qt version.

Not sure if I missed anything there.  Let me know if anyone notices something, and I can update this comment with new info.

herbertp:
A brief chat with t-paul clarified that 'simply' disabling transparency on older Qt versions (< 5.12) for the GUI would allow this to be merged because "it's a feature useful enough to have it".

Irev-Dev:
Hey,
Just leaving a comment expressing my interest in transparent backgrounds.

Use-case is very specific, but here it is all the same.

We use the image previews on [CadHub](https://cadhub.xyz/dev-ide/openScad#fetch_text_v1=https%3A%2F%2Fraw.githubusercontent.com%2FGliaX%2FStethoscope%2Fmaster%2Fsource_files%2Fscad%2Fstethoscope_y-piece.scad), and having transparent backgrounds would allow us to control more of our theme in the browser where the rest of our theming is done.

praetp:
any update on this ?

tmolteno:
Also really keen to see this done. OpenSCAD is fantastic for generating diagrams for physics lectures, and transparency would be wonderful.

whosawhatsis:
Would still really like to see this happen.

chrysn:
Not stale: AFAICT, the issue still persists.

t-paul:
Keeping the PR open does not help much, moved to feature request #6648 (which will never auto-close). Code is still here if someone takes on the topic.

Issue #1496, opened 2015-11-19
Transparent objects render faces out of order


thehans:
When viewing a transparent object, the preview has artifacts due to the faces being drawn in uncontrolled order.

This shows up as a strange seam in cylinders for example:
![screenshot from 2015-11-19 12 00 58](https://cloud.githubusercontent.com/assets/252233/11279311/4ebeee38-8eb5-11e5-9d31-9a5a75056de3.png)

It looks much better if the back faces are drawn before the front faces.  This can be accomplished with `glEnable(GL_CULL_FACE);` and `glCullFace` calls.  Drawing the object with the front culled first, then again with the rear culled.
![screenshot from 2015-11-19 13 05 34](https://cloud.githubusercontent.com/assets/252233/11281057/716d441c-8ebe-11e5-8c4c-d1fb60add0e9.png)

The downside to this is that it could possibly make preview rendering slower.  I haven't found any noticeable slowdowns based on my subjective interactive testing, but if there are any suggested performance tests, I would be willing to try them.

To mitigate some of this potential slowdown the rear faces can be only drawn when color alpha < 1.  And only draw the front faces when when not rendering cutouts.
##

kintel:
I'm slightly concerned about performance degradation.
Otoh, rendering lots of transparent objects probably isn't a major workflow.

thehans:
Fixed by #1497

PR #1497, opened 2015-11-19
Render transparent objects rear faces before front faces

thehans:
for issue #1496 

TODO:
- [x] Consider performance impact


thehans:
Its currently a little difficult to directly measure rendering performance.  I'm thinking it might be good to have some performance test functionality built in, where enabling it would render frames as quickly as possible, and spit out an FPS count.  

This might make sense to put the FPS readout as an added feature to the animation settings, since it doesn't necessarily always hit the FPS that you set it to.  
And then maybe a checkbox next to FPS input box for "Max" which would completely disable any delay between frames.

Also needed is some particularly gnarly test case that gives the renderer something to chew on.

Thoughts?

kintel:
See #1561 - to support higlighting unexpected backfaces, we render _front_ faces first, then backfaces.
I think the #1561 workflow is more important, but I wanted to raise that conflict here, in case someone else have other arguments.

thehans:
@kintel I rebased this on current master and fixed conflicts.  In doing this I cleaned up the behavior of my code so that opaque object rendering should be unchanged.  I think the changes needed for #1561 can go within that conditional branch for handling opaque objects without any problem.

I also think it is a harmless performance enhancement to cull the front faces of negative shapes, so I did that, which, should hopefully make up for any performance impact of doing two passes on transparent shapes.

src/OpenCSGRenderer.cc
```diff
@@ -123,10 +123,22 @@ void OpenCSGRenderer::renderCSGProducts(const CSGProducts &products, GLint *shad
 				colormode = ColorMode::MATERIAL;
 			}
 			
-			setColor(colormode, c.data(), shaderinfo);
 			glPushMatrix();
 			glMultMatrixd(csgobj.leaf->matrix.data());
-			render_surface(csgobj.leaf->geom, csgmode, csgobj.leaf->matrix, shaderinfo);
+			
+			if (setColor(colormode, c.data(), shaderinfo)) {
```
kintel: I think it would be cleaner to move the alpha check out of `setColor()` and do it manually here.

thehans: The problem is that setColor doesn't unconditionally set the color that you give it from this scope. It depends if the object is highlighted, etc.  Putting the check here would mean basically duplicating all the code of setColor.   This function previously had void return type, I figured a boolean return was the simplest way to get that info out of the function.  Would it be preferrable to return the full Color4f basecol instead?  Then check the alpha of that here?

kintel: Yes, that's a cleaner solution.
Better would be to e.g. split out those calculations from `setColor()`, but that's probably out of scope for this PR.

kintel:
Humm, somehow these tests fail on the Mac CI:
```
The following tests FAILED:
	322 - cgalpngtest_polyhedron-nonplanar-tests (Failed)
	643 - csgpngtest_polyhedron-nonplanar-tests (Failed)
	895 - monotonepngtest_polyhedron-nonplanar-tests (Failed)
	918 - cgalstlcgalpngtest_polyhedron-nonplanar-tests (Failed)
```

Any ideas?
I'll have to open the PR in an OpenSCAD-owned branch to get the test result upload from the CI to work..

kintel:
@thehans It looks like the test failures are unrelated to this PR: Master also fails now. I suspect homebrew update on TravisCI from 4.9 to 4.11 to be the issue -> we should fix that (CGAL 4.11 support) separately.

Issue #1495, opened 2015-11-19
Difference with highlighted object makes cut faces transparent


thehans:
When subtracting highlighted geometries:  X-#Y, the new faces cut into X are given the color of Y.  This ends up putting transparent faces on otherwise solid colored objects if Y is highlighted.

After some discussion in IRC kintel suggests a solution of changing this line:
https://github.com/openscad/openscad/blob/master/src/OpenCSGRenderer.cc#L110
from
`if (j_obj.flag & CSGTerm::FLAG_HIGHLIGHT) {`
to
`if (highlight && j_obj.flag & CSGTerm::FLAG_HIGHLIGHT) {`

This will makes it so the faces cut by highlighted objects stay the same color as they would be when not highlighted.

This will change the expected test results for all colorscheme logo tests and a couple other highlight related tests.

Before
![logo-expected](https://cloud.githubusercontent.com/assets/252233/11276657/884ed9fe-8ea8-11e5-8821-6989db4b9251.png)
After
![download](https://cloud.githubusercontent.com/assets/252233/11276773/094553bc-8ea9-11e5-81c0-986219c45f33.png)

One side effect seems to be that faces cut into a highlighted object (#X - Y) will be drawn first with non-highlighted color, then drawn over with highlighted.

thehans:
One case that this still doesn't address is X-Y where Y is set to alpha < 1 using the color function and not highlight.  This will still make a transparent face in a solid colored object.  I'm thinking it would be good to also force alpha=1 when drawing cutouts.

This can be done by inserting 
`if (colormode == COLORMODE_CUTOUT) basecol[3] = 1;`
before line 77 here:
https://github.com/openscad/openscad/blob/master/src/renderer.cc#L77

nophead:
If the cut face is normally coloured by the object making the cut it seems
logical it should be made transparent if it is cut by a transparent object.
However it has always seemed a little odd to me that cut faces are anything
different from the solid they are part off.

thehans:
Ideally I think it should work this way, for a difference X-Y
if Y is colored then the cut faces get the color of Y, but the alpha of X
else if only X is colored the cut faces get the color of X, and the alpha of X
else the cut faces get default cut color from colorscheme

Highlighting Y should make Y transparent and red, but not affect the color of faces it cuts.

[...]

PR #1498, opened 2015-11-19
Avoid transparencies on cut faces.

thehans:
Proposed fix for Issue #1495


kintel:
One issue: If you cut into a transparent object using transparent cutout, we effectively negate that transparency. That would break the following case:
```
cube(30, center=true);
color("green", alpha=0.5) translate([0,-15,0]) difference() {
    cube([10,4,10], center=true);
    translate([0,-2,0]) sphere(3);
}
```

screen shot 2015-11-19 at 22 32 49 pm

kintel:
There's still one issue left:
screen shot 2015-12-02 at 11 39 07 am

The problem is that highlighted objects subtracted from other highlighted objects should still have the highlighted material. To detect this, I think we should change our rendering code to _not_ render highlighted objects during the normal rendering pass, _unless_ we're subtracting from a not-highlighted positive. Needs more tinkering, and perhaps a better test case.

PR #1500, opened 2015-11-19
New feature - Anaglyph stereo view of the model

josefpavlik:
New feature added - 3d view of the model using special red-cyan glasses.
https://en.wikipedia.org/wiki/Anaglyph_3D
anaglyph2

The feature can be activated pressing ALT-S (like Stereo Vision) or clicking on the toggle button under the preview window.
When the feature is activated, the model view is automatically switched to black and white mode enhancing the 3d view experience.

The eye distance can be adjusted with mouse wheel holding the ctrl key.


kintel:
While stereo viewing is cool, I think that anaglyph rendering is just the tip of the iceberg. I expect there to be a range of other viewing modes interesting for this purpose; oculus rift, google glass, morpheus, polarized glasses, active 3D glasses...

I think we should rather refactor such functionality to be independent on the type of rendering performed.
I also think that this is best done after refactoring out 3D rendering code to use a more suited OpenGL library than our home-grown rendering code, see https://github.com/openscad/openscad/wiki/Project%3A-Improve-OpenGL-rendering

josefpavlik:
@kintel You are right about the 'tip of the iceberg'. However the red-cyan anaglyph is the simplest 3d view mode. Either for the use, as for the implementation. Anyone can use is, there is only need of cheap glasses that comes with many DVD. You have not need of special monitor or expensive glasses.

The implementation is very simple, so there is minor possibility of introduce a bug.

However I've read the article about the refactoring of the rendering code. It's interesting, but hopelessly I have'nt sufficient experience with opengl and however I have not time for doing a challenging task as it. However, you are right that the rendering code should be improved.

donbright:
i was impressed by how little you needed to change to accomplish this. seems like more of the code is modifying the UI than modifying the engine. just curious what is "BW Mode"?

josefpavlik:
@donbright Work on anaglyph stereo view - 10%. User interface for put it on/off - 40%. The black-and-white mode - 50% of the work.
BW mode is black and white mode - the colormap is forced to grayscale when you put on the stereo view to enhance the 3d view. The anaglyph is better on black and white image.

donbright:
thanks... very cool.

josefpavlik:
I've made another fix. When the model is colored (using color() function), Now the colors are converted to gray during the stereo view mode. Before only default colors were converted do gray.

josefpavlik:
@kintel 
I've port the patch to latest version. Please include it in mainstream. 2 years ago you have wrote that this patch is nice, but that the entire rendering code should be refactored using another libraries. But after 2 years we have the same code. So I think that you can add my feature to allow people to using it and when after another years entire code will be refactored, I can try to adapt it to new code.

shalnoff:
would be great to see this feature in the next release

koendv:
Josef Pavlik,

I'd like to thank you for this patch, it's been a nice source of inspiration.

https://github.com/koendv/openscad-raspberrypi

regards,

koen

audunmg:
This would be very useful, it is a really nice QOL addition, and I constantly miss it.

koendv:
Thank you. I'm happy you like it. If you look in "Releases", you can also download versions for Windows and intel linux.

The 3D vision works, but:
- only anaglyph 3D is supported (green/red glasses)
- the 3D vision is calculated on the cpu of the pc, instead of in the video card.
This can be improved.

FreeCAD is an open source cad program, with 3D vision. Like OpenSCAD, FreeCAD runs on Windows, Mac and Linux.
In FreeCAD the 3D image is calculated in the video card, and various 3D modes are supported (anaglyph, side-by-side, interleaved, ...)

So if I had to begin again, I would take the 3D OpenGL from FreeCAD and add it to OpenSCAD OpenGL.

But, as you say, what determines success is getting the code included in the OpenSCAD git.

best,

koen

kintel:

PR #1518, opened 2015-12-12
Change OpenCSGRenderer to use opengl display lists

obliviand:
This is a very quick fix I implemented for a large design I'm working on to speed up rendering in preview.  I also recently cleaned up memory leaks including the "use" command memory leak in the parser.

TODO:
- [ ] Look into the memory leak fixes (are some of them redundant?)
- [ ] Coding style updated / clarification
- [ ] Refactor or add FIXME to static_cast
- [ ] Port to ThrowntogetherRenderer
- [ ] Optional: Cherry pick fixes not related to display lists to master


kintel:
@obliviand I'm working on a larger refactoring which touches on some of this stuff (see https://github.com/openscad/openscad/pull/1527). It could make sense to rebase your code on top of that branch.

Also, in the longer run I think using VBOs is a better solution that display lists, but that requires more deep OpenGL refactoring. Have you had any thoughts in that direction? Ping @thehans (who has been tinkering with those ideas a bit).

obliviand:
@kintel Just merged csgterm-refactor.  opencsgtest_highlight-and-background-modifier test is failing, but looking at the expected image, the code appears to render correctly.  The expected test file has a tan block on the left sphere that should be transparent red highlighted.  Not sure why the expected test file changed.  I won't go changing the expected results, so please have a look.

As for VBOs, I looked at using them initially but had some issues that were not solvable as quickly as the display lists.  Other than major refactoring, as you point out, the only other way would be to use VBOs in place of display lists.  The upper limit of performance for preview would still be related to the multiple for loops in the rendering routine.  Even if you load all of the respective vertices in VBOs, they still must be iterated through.  Each OpenCSGPrim has to have a VBO like the display list, then loop through the OpenCSGPrim::render method for the primitives list.  I'm not sure you will get better performance over display lists with VBOs unless all of the work done by the for loops and OpenCSG is offloaded to the GPU.

kintel:
@obliviand Thanks - I'll take a look. To clarify: The main goal of moving to VBOs won't be increased performance over display lists, but compatibility with newer OpenGL, including GLES, where display lists are deprecated or not supported at all.

obliviand:
@kintel I'll have a look at moving all the renderers to VBO.  I'll open another branch since it probably will need refactoring.  As it happens, I'm hitting the upper limit on preview performance again.

kintel:
Preview performance is also bound by the fact that we need to render using multiple render passes, e.g. potentially rendering each negative object many times. This is an inherent issue with the CSG rendering algorithm which would need to be addressed somewhere else (e.g. through algorithmic optimization or by e.g. opportunistic caching of sub results in preview mode).

TakeItAndRun:
Hi Jason,

as I understand it your are trying to speed up rendering in preview.

So I assume that you have a quite complicated thing modeled in OpenSCAD.

There is the way to optimize code that you may or may not be aware:

Doining a lot of differenece() with objects that are overlapping is time
consuming.
So if you create a tube with a lot of round holes in it, you can do that by
subtracting cylinders that start at the origin (and overlap there).
Or you can subtract little cylinders that are placed the the radisus of the
tube.

The later way gives you an increase of speed of about 20% for F5 and F6, as
well.

Now, I checked also the times between redraws: the later way give faster
re-draws by about 12%.

Maybe this is a way to speed  up your designs.


(I came across this problem first, here:
http://kitwallace.tumblr.com/post/78764510201/diamond-cuts)

Here is the code I used to test this:

h=100;
dh=2;
a=10;
d=1;
e=0.01;
rc=1;
hc=3*a;
$fn=1000;
n=20;

module tube(){
    difference(){
        cylinder(h,a,a,true);
        cylinder(h+2*e,a-d,a-d,true);
    }
}

module kross1(){
    for(i=[360/n:360/n:360])
    rotate([90,0,i])cylinder(hc,rc,rc,true);
}
module kross2(){
    for(i=[360/n:360/n:360])
    rotate([90,0,i])translate([0,0,a-d/2])cylinder(d+2*e,rc,rc,true);
}

module var1(){
    difference(){
        tube();
        for(i=[-h/2:dh:h/2])
            translate([0,0,i])children();
    }
}

var1()kross2();

// time between re-draws (F5)
// var1: 3.67 sec
// var2: 3.25 sec


2015-12-20 21:24 GMT+01:00 Marius Kintel <notifications@github.com>:

kintel:
@obliviand Some comments/questions:
* Is using a trailing underscore in member names your way of signifying private instance members? (e.g. OpenCSGRenderer::root_products_). While I enjoy seeing more ways of improving code readability, I prefer consistency. We try to standardize on using "this->" to signify that we're accessing instance members.
* I noticed that you didn't update ThrowntogetherRenderer to use display lists. Did you just await that until the OpenCSG code is good?
* You mixed in a bunch of unrelated fixes (coding style, memory leaks). In general I'd like those to be separated out - makes review a lot easier. ..but not critical right now.
* Instead of const_cast<>, I think I'd prefer tagging the _*built_ members as immutable. Any comment on that?

obliviand:
@kintel
* Yes, and it's a habit at this point.  Also, "this->" gets long winded and you can end up with hard to find mistakes if you forget to add "this->" since it's not required. https://google.github.io/styleguide/cppguide.html.  If you have a similar style guide, I will follow that in the future, but I didn't see one when I looked.
* I rarely use thrown together mode.  I did the change initially just to make OpenCSG mode usable.  I can go back and have a look at that as well.
* Like I said above, the coding style is more of a habit.  The memory leaks commits were based on compiling with the gcc -fsanitize=address option.  Sometimes it points you in the wrong direction, but I think the remainder was still valid and related to the OpenCSGRenderer changes.  I thought I had made the memory leak commits separate, but in the same branch?
* Do you mean make them mutable with the mutable keyword? It would probably be better to refactor since there should be a first pass state initialization from the compiled program CSG products to create the opengl state and cache the display ids/vao,vbos (non-const).  After that the rendering steps can still be const for the draw method when the user is interacting with the opengl view.  This would remove the built state variables, but it would probably duplicate some amount of code between the draw and renderCSGProducts methods.

kintel:
We should clarify coding style. We could use a linter to help enforce it. A fair bit of the coding style is in flux due to various contributors having different styles and I fear that a reformatting would introduce a lot of disruptive conflicts in open branches.

I think the leak commits were separate. I might cherry-pick them to master separately, depending on what it takes to merge the rest.

Yes, I was referring to the mutable keyword. I agree that refactoring is better.

kintel:
@obliviand Did you see my pull request?

kintel:
@obliviand Another question: I didn't quite understand your memory leak fixes in https://github.com/obliviand/openscad/commit/eec67eca77639fd7a8bc5c138f3cd0624e08da87

You're attempting to reset() a bunch of shared_ptrs stored in STL containers. However, when the STL containers get destructed, the reference count of the shared_ptr objects should automatically be decreased. Are you trying to fix something else that I've overlooked?

obliviand:
@kintel.  You can ignore the GeometryCache.cc changes, they were from the gcc -fsanitize pointing to GeometryCache as the leak location.

kintel:
@obliviand The memory leak fixes in ~CSGTreeEvaluator looks equally redundant.

obliviand:
@kintel  Yes. I forgot they were in there as part of the gcc -fsanitize troubleshooting.  The only memory leak that was left should have been the use command in the parser.  Also, to clarify, your refactoring of the csgterm code eliminated the root cause of the leaks for the display lists implementation that I fixed a different way under the old code.

obliviand:
@kintel Did you see my question about using glVertexPointer, etc. as an intermediate step for the renderers?  If so I will implement based on your changes.

kintel:
@obliviand No, I didn't catch that question..

kintel:
..and another question: Why do you set all the freed strings to 0 afterwards? Just to catch potential bugs in bison accidentally reusing one of those values? To be consistent, that should also be done for deleted objects, but it looks to me that the standard way of doing this is to just free() and forget.

obliviand:
@kintel Zeroing pointers prevents use after free type issues.  It probably would be safer to be consistent and add them after the deletes as well.

kintel:
@obliviand About gl*Pointer() - I didn't quite understand the question. Using e.g. glVertexAttribPointer to specify vertex arrays should be consistent with OpenGL 2.0 ->, and it, to my knowledge, not deprecated in any subsequent OpenGL version. So yes, that could be a good way forward as it would allow moving to VBOs easier later on while keeping immediate mode rendering for as long as it's required by OpenCSG.

obliviand:
@kintel Please take a look at https://github.com/obliviand/opencsg-vbo-example.  It's a simple concept implementation based on the OpenCSG example program but using VAO/VBOs and some of the polyset code and transform from OpenSCAD.  You should be able to drop it in the OpenCSG source directory example and compile.  Some notes:
* It uses the default shaders in the OpenGL drivers for the respective GPU (i.e. does not work with custom shader programs emulating a fixed pipeline).  OpenCSG appears dependent on OpenGL immediate calls information not translated to custom shader programs by the OpenGL drivers.
* I've only run it on Linux.
* The example supports 3.0 with VBOs only and >= 3.1 with VAO/VBOs.
* I've run it on Intel (OpenGL 3.0) and nVidia GPU (OpenGL 4.5), no ATI.
* Added menu item to switch between using Display List and VAO/VBOs.  Crashes on Intel when switching back from VBO to Display Lists.  NVidia appears to work without crashing.
* On both GPU's I have, VBOs run ~57% slower than display lists with this code based on the frame rate.  Best I can tell from research is that optimization is happening in GPU drivers for Display Lists and most of the VBO performance hit is from cache misses when binding objects.  Without rewriting OpenCSG, I'm not sure how to get the same performance from the VAO/VBO implementation across difference GPU vendors.  NVidia apparently has extensions like GL_NV_vertex_buffer_unified_memory to help speed this back up, but that doesn't help other GPUs.

kintel:
Thanks, I'll look at it.
FYI: The csgterm-refactor branch has been merged to master.

kintel:
Had a look at this on OS X, but it uses GLUT. GLUT on OS X is deprecated (and only supports OpenGL <=2.1). FreeGLUT isn't even ported. OpenGL isn't very fun to play with multi-platform these days..

kintel:
Got it to work on OpenGL 2.1. I get similar performance results as you. Looking at the trace, I cannot see any performance penalty when binding objects, but there's a function call overhead for VBO which, together with glMatrix calls make up almost 30% of the time spent drawing small objects.

kintel:
..and a comment about rewriting OpenCSG: OpenCSG is pretty old and has a bunch of code for handling the fixed function pipeline. The only remnant of this left in the code path we use is (to my knowledge) that they use glColor() to specify ID materials. Since we can use a custom shader for all the depth buffer magic in OpenCSG, we could easily rewrite this to use a shader uniform. Question is more if it makes sense to rewrite anyway, to use modern OpenGL all the way. The only tricky part in OpenCSG (IMO) is the occlusion query stuff used to optimize subtraction sequences.

obliviand:
When I looked through the code OpenCSG appeared to still be using glAlphaFunc and glDepthFunc which I could not find a shader variable that captured the state of these calls.  I wrote a fixed pipline emulation shader to try to keep compatibility wiht the OpenCSG internals and removed it when I could not find these state variables as a GLSL pass through to the shader.  Also, nothing was drawing through the OpenCSG pipeline with the custom shader.  Only the second pass object draw worked if I removed the glDepthFunc calls.  My guess is that the built in opengl driver shaders have access to those states from the immediate mode calls but I could not find any formal documentation for GLSL allowing access to state from those old calls for backward compatibility.

From what I read, the VBO function call (if you are talking about glBindBuffer/glVertexPointer/glNormalPointer for OpenGL 2.1) overhead is the GPU trying to find the recent buffer bound.

It would make sense to rewrite with modern OpenGL if it will be faster than the current implementation.  Even though deprecated calls are being used, the display lists are still supported on my OpenGL 4.5 GPU.  Display lists are obviously faster than immediate mode on the GPUs I have.  VBO/VAOs not so much in mixed mode, so it would have to be a pure VBO shader CSG implementation to be faster.  It appears for a rewrite to be faster, rather than looping through multiple objects on the CPU every time the object is drawn, a single vertex buffer per top level object would be needed along with a shader and relevant state data per vertex.  Easier said than done obviously.

kintel:
glDepthFunc() shouldn't be deprecated. glAlphaFunc() should be replaced by a shader uniform and a fragment discard test. Thanks for catching that - it slipped through my net.
FYI: The function overhead I measured wasn't specific to VBO functions, glMatrix* had the same overhead, so I suspect this is simply normal function overhead when calling into the GL library. I measured on Mac, which tends to have higher overhead than necessary though.

In any case, a pure VBO version is quite some work, so I agree that going for display lists is a much lower hanging fruit right now. Thanks for hanging in there and investigating.

obliviand:
@kintel I decided to work on a vbo preview renderer for a few weeks in the new year.  Please take a look https://github.com/obliviand/openscad.git opengl-vbo-renderer branch when you get a chance.

Some of the first steps were re-factoring by moving polyset-gl.cc OpenGL functions into Renderer base class in renderer.cc.  After that it opened the door to do a standalone vbo based renderer without impacting existing renderers for now.

It's based on the current master, so best I can tell from the tests, the same tests that I see fail on this branch fail on master right now.

I left the other renderers alone and added code to enable/disable the new renderer via the opencsg preferences.  The new renderer is only included if you build with CONFIG+=experimental.

It's not perfect yet, but may allow someone to build on it an get to a pure shader csg implementation.
The edge shader code is copied into this class with changes to support color values being included in a vbo.

Tested on intel gpu: Noticeably uses main ram on large assemblies with multiple parts, but its fast on redraw.
Tested on nvidia gpu with nouveau driver: Uses gpu ram, redraw is slow like old render method.
Tested on nvidia gpu with binary driver: Uses gpu ram, redraw is fast.

kintel:
@obliviand Yay! Would you mind opening a new pull request, or force-pushing to this branch?

obliviand:
@kintel Probably shouldn't yet.  I found a bad regression after I decided to upload yesterday (of course!).  The following code shows the error:

```
difference()
{
    cube([20,20,5],center=true);
    difference(){
    cube([10,10,10],center=true);
    cube([5,5,20],center=true);
    }
}
```

I have a large model with multiple parts including scad parts and imported STL parts.  The STL all appear to be fine.  The scad parts render fine if the parts don't have a similar nested differences in them.  The ones that do have nested differences show errors similar to the below image.  Not sure if it's getting messed up in the OpenCSG::render call or trying to do the follow on glDepthFunc(GL_EQUAL) product draw all at once.  I thought I captured all of the vertex, attributes and ordering upfront so that redrawing it later should come out the same as the OpenCSGRenderer, but apparently not...

This image is the nested difference error from the above code:

screenshot from 2017-02-08 18-16-04

This image is with the old OpenCSGRenderer enabled with the above code:

screenshot from 2017-02-08 18-16-49

obliviand:
@kintel Ok, fixed the nested difference regression, and pushed to my repository.  It seems you can't have a vbo bound when calling into OpenCSG.  Also, reworked the product_vertex_sets to maintain draw consistency and reduce the number of drawCSGProducts calls required to do root/background/highlight operations.

I'll initiate a pull request from the new branch.  Still experimental, so I'm sure there will be other regressions I haven't found yet.

obliviand:
Pull request submitted as #1935 for new vbo based renderer.

PR #1536, opened 2016-01-05
Extended List-Comprehension

t-paul:
This extends the list comprehensions by some of the features from the OpenSCAD2 RFC.

* Generalize to allow multiple LC expressions in a vector expression
* Add `if`/`else`
* Add `each`

TODO:
* [x] @kintel: Code review
* [x] Look into the "if-if-else" conflict
* [x] Tag as experimental
* [x] test experimental stuff (@kintel)

Move to separate PR for accepting this as a stable feature:
* Evaluate the 'each' syntax
* Cheat sheet
* Documentation
* Optional: Add or update example

See https://github.com/doug-moen/openscad2/blob/master/rfc/Generators.md


kintel:
Something to think about: ```[ if (a) if (b) c else d ]```
Currently, this binds the else to the nearest if, which is common. However, what if I want to bind it to the first if? We currently don't have any ways of specifying that in the syntax. In C you're use curly braces to specify the scope but we don't have that. Should we allow parentheses for this purpose (e.g. ```[ if (a) (if (b) c) else d ]```.
Perhaps something for the mailing list.

t-paul:
It seems the `loop` macro in lisp solves that by using an optional `end`.

t-paul:
Hmm, using `(` `)` looks a bit strange but would prevent introducing another `end` keyword.

With the (optional) `end` it would look like:
```
echo([ for (a = [0 : 3]) if (a < 2) if (a < 1) ["+", a] end else ["-", a] ]);
ECHO: [["+", 0], ["-", 2], ["-", 3]]
```
vs. automatic binding to the nearest `if` without the `end`
```
echo([ for (a = [0 : 3]) if (a < 2) if (a < 1) ["+", a] else ["-", a] ]);
ECHO: [["+", 0], ["-", 1]]
```

I guess I still should read up on shift/reduce conflicts to refresh my mind...

Or maybe @doug-moen has an even better suggestion?

doug-moen:
In OpenSCAD2, within a list comprehension or expression, you use (if (cond) expr else expr) or (if (cond) expr) to force grouping.

I even wrote a conflict free Bison grammar for this. I use %nonassoc twice to resolve the dangling else ambiguity, and I do not use %left or %right. I didn't publish it since it's slightly out of sync with the actual language proposal on github, and I never got around to updating it. I don't like the official OpenSCAD grammar because, with the shift/reduce conflict and all of the %left/%right directives, I don't understand what it means. Too scary. So I wrote another grammar from scratch that, for me, is easier to understand, and also easier to modify (since no conflicts are allowed).

kintel:
@doug-moen Would it be productive to replace the existing grammar with your new one, and retrofit the old syntax there instead? There might be a testing challenge, but I believe we have good test coverage of the parser.

t-paul:
Would it make sense to look at some tool chain that handles utf-8 natively? The currently needed workarounds are quite gross and from what I'm reading `flex` is still not able to handle utf-8.

doug-moen:
@kintel Yes, if I strip out a bunch of productions that only make sense in OpenSCAD2, then I can produce a grammar that could be useable in OpenSCAD. Note that my grammar is "looser" than the original OpenSCAD grammar and permits some constructs that are supposed to be flagged later as errors by a semantic analyzer pass. My grammar has 'if' expressions, 'for' expressions, 'let' expressions, and 'each' expressions. The code that restricts certain expressions to inside a list literal (like 'for' and 'each') is in the semantic analyzer. In OpenSCAD2, "if (expr) expr else expr" is allowed as an expression, as an alternative to "expr ? expr : expr", but that could be disallowed the same way that "for" is disallowed if you prefer.

doug-moen:
Torsten: "Would it make sense to look at some tool chain that handles utf-8
natively? The currently needed workarounds are quite gross and from what
I'm reading flex is still not able to handle utf-8."

I suggest writing the lexer in C++ using Boost.Spirit.Lex. Use the ICU
library (for character classification) and the Boost unicode iterators to
handle Unicode correctly. It should be straightforward to support Unicode
identifiers this way.

Issue #1217, opened 2015-02-15
Sequential assignment of default module arguments


daprice:
It would be helpful if default arguments for modules could be defined dyamically and sequentially, so each definition could use the value of any previous arguments (similar to how the proposed `let()` function works). For example, in the following code, the assignment of `b` should be able to use the value of `a`:
```
module myModule(a, b=a) {
	echo(b);
}

myModule(a=1);
```

If this was implemented, the echoed output would be `1`.

However, in the current nightly this just causes a warning: `Ignoring unknown variable 'a'.`

MichaelAtOz:
Then

s=0.01; // smidgen

module m(params,s=s*2,s2=s*4) {
  echo(s,s2);
}

Would not work anymore.

MichaelAtOz:
oops markdown mishap, make that `s=s*2,s2=s*4`

kintel:
I agree with @MichaelAtOz - this would break compatibility of such a corner case.
We do plan to break such compatibility in the future, so that parameter lists behave the same way in all context. When we do, we will provide either a compatibility mode or an upgrade option for old scripts.
That's a bit of work, so we'll probably switch it all at the same time.

doug-moen:
@kintel said "so that parameter lists behave the same way in all contexts"

I can see the use case for default values in function and module definitions. But in function and module calls, it's less clear that we want "sequential assignment".

Here's some code that was just posted to the forum:
```
function merge(left, right, kp, kl=0, kr=0, _acc=[]) =
    kl >= len(left) && kr>= len(right) ? 
        _acc:
        merge( left  = left ,
               right = right ,
               kp    = kp,
               kl  = next_k(kl, kr, len(left),  len(right), kp, left[kl], right[kr]),
               kr  = next_k(kr, kl, len(right), len(left),  kp, right[kr], left[kl]),
               _acc  = concat(  _acc, 
                                next_entry(kr, kl, len(left), len(right), kp, left[kl], right[kr] )
                             )
             );
```
Look at the recursive call to merge(), which sets the kl= and kr= parameters based on the values of kl and kr in the parent invocation merge(). This code would be harder to write if function calls were changed to use sequential assignment for keyword arguments.

kintel:
@doug-moen Point taken. Any ideas of a pattern to follow, or how to communicate this behavior clearly?

doug-moen:
[...]

PR #1587, opened 2016-02-21
Add echo() and assert()

### LINKED ISSUE #381: an assert function forcing compilation to quit
With large scripts and many parameters certain combinations of parameters may not generate a usable result. Currently there is no way to quit compilation and return a string describing the problem.

I propose an assert function with a syntax like : assert(<test if true>, <return string>);

If the <test if true> value is false, the function will abort generation and return the <return string> value. The return string value can be constructed like the values for echo.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/378323-an-assert-function-forcing-compilation-to-quit?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

t-paul:
Fixes #381

**TODO**:
- [x] fix `toHtmlEscaped` which is only available in Qt5. See [```Qt::escape(src)```](https://doc.qt.io/qt-4.8/qt.html#escape)
- [ ] Create new issue for (Documentation, Cheat sheet and example)
- [x] Code review (@kintel)
   - [x] General & tests. -> Some minor cleanup possible.
   - [x] Evaluate stack trace proposal. -> I believe this can be cleanly applied later -> separate issue
   - [x] Evaluate ```message=``` proposal -> Consider rewriting assert() -> assert(condition=, message=) (i.e. single condition). Might need more discussion
- [ ] Sequential assignment in echo() breaks backwards compatibility. Someone might rely on the old behavior. How to deal with that? See [echotest_echo-tests](https://github.com/openscad/openscad/blob/master/testdata/scad/misc/echo-tests.scad), which fails with ```--enable=echo```.
- [ ] Merge with master (e.g. resolve the new type of "function calls", perhaps in a new supertype)

This adds `assert()` as expression and statement and `echo()` as expression. In addition the `echo()` statement is aligned with the new behavior of using sequential assignment and return child modules.

echo expression:
```
a = 3;
b = 6;
t6 = echo(c = 2, d = c + 9) a*b*c*d;
echo(t6 = t6);

// ECHO: c = 2, d = 11, 396
// ECHO: t6 = 396
```

assert expression:
```
a = 10;
b = 20;
v = assert(a < 20 && b < 20, "Test! <html>&</html>") a + b;

// ERROR: Assertion '((a < 20) && (b < 20))' failed: "Test! <html>&</html>"

a = 8;
b = 18;
v = assert(condition = a < 20 && b < 20) a + b;
echo(v);

// ECHO: 26
```

assert statement:
```
assert(false);

// ERROR: Assertion 'false' failed

a = 8;
b = 6;
assert(a < 10 && b < 8) { cube(a); sphere(b); }

// produces both cube and sphere as long as both assertion expressions are true
```


src/control.cc
```diff
@@ -159,6 +160,13 @@ AbstractNode* ControlModule::getChild(const ValuePtr &value, const EvalContext*
 	return modulectx->getChild(n)->evaluate(modulectx);
 }
 
+static void applyChildren(const ModuleInstantiation *inst, AbstractNode *node, Context &c)
+{
+	inst->scope.apply(c);
+	std::vector<AbstractNode *> instantiatednodes = inst->instantiateChildren(&c);
+	node->children.insert(node->children.end(), instantiatednodes.begin(), instantiatednodes.end());
```
kintel: since applyChildren() is always called on a fresh node, it's not necessary to copy, we can just do like the GroupNode: ```	node->children = inst->instantiateChildren(&c);```

src/evalcontext.cc
```diff
@@ -43,14 +43,66 @@ ModuleInstantiation *EvalContext::getChild(size_t i) const
...
+	}
+}
+
+std::ostream &operator<<(std::ostream &stream, const EvalContext &ec)
+{
+	for (size_t i = 0; i < ec.numArgs(); i++) {
+		if (i > 0) stream << ", ";
+		if (!ec.getArgName(i).empty()) stream << ec.getArgName(i) << " = ";
+		ValuePtr val = ec.getArgValue(i);
+		if (val->type() == Value::STRING) {
```
kintel: Couldn't we just use Value::operator<< here?

src/expr.cc
```diff
@@ -463,45 +463,127 @@ void ExpressionMember::print(std::ostream &stream) const
...
-	return this->first->evaluate(&c);
+	ValuePtr result = this->first ? this->first->evaluate(&c) : ValuePtr::undefined;
+
+	std::stringstream msg;
+	EvalContext echo_context(&c, this->call_arguments);
+	msg << "ECHO: " << echo_context;
+
+	if (this->first) {
+		if (echo_context.numArgs()) msg << ", ";
+		if (result->type() == Value::STRING) {
```
kintel: Couldn't we just use Value::operator<< here?

t-paul: Not as-is if we want the same output as the statement `echo()`. The `operator<<` uses `QuotedString` which converts `TAB` to `\t` and such.

kintel: ah, good point. thx.

t-paul:
Updated `assert` to use `assert(condition, message)`.

dcousens:
LGTM, will test soon :+1:

greyltc:
Is there anything I can do to help with getting this PR accepted and merged? I'd love to have this feature.

kintel:
@greyltc The sequential assignment issue needs to be resolved, which I guess needs some discussion. Apart from that, help with documentation/cheat sheet and an example would be cool.

doug-moen:
@kintel said "The sequential assignment issue needs to be resolved"

My view is that echo() should not have sequential assignment semantics.
 1. It's not backward compatible.
 2. It's not the right design. In general, function & module calls with labeled parameters should not introduce a new scope, they shouldn't have sequential assignment semantics.
  * It's confusing to developers who have used other programming languages with labeled parameters, since no other language works this way.
  * It gets in the way, it's a barrier to programming. There are legitimate reasons you might have local variables or parameters with the same name as labels used in a function call. It's annoying and inconvenient if those variables are shadowed and made inaccessible (and silently substituted) within a function call. I argue this case in more detail in #1217.

t-paul:
It would simply behave like `let()`. Is that considered wrong too?

t-paul:
As the pull request is pretty much dead due to the massive conflicts, I've extracted the `assert()` part to a new pull request #1827.

I'll wait with `echo()` until it's clarified.

nophead:
Since echo in functions is for debugging, it seems wrong to me that it can
have side effects. On the other hand let is there to assign variables.

kintel:
@nophead has a good point. Are there any obvious uses for this feature?

doug-moen:
What @nophead said. You should be able to comment out an echo(...) without it changing the meaning of the program, but that's not safe if `echo(args)expr` binds names within `expr`.

MichaelAtOz:
-1 let() style.
+1 get echo() expression into dev snapshot for testing

t-paul:
There's nothing to get in, so far it's only listing what is not wanted (actually we may want to move the discussion to some new issue as that pull request is dead regarding the code attached).

There's no need that `echo()` behaves exactly like `let()`. That comment related more to the general statement that sequential assignments are evil, yet we introduced those and there was discussion to use it in other places too. So that leaves me confused.

As for the side effect, that's certainly a point that makes sense. To get forward, maybe we should leave `echo()` as it is module only and have a new function/module that has a nicer interface.

nophead:
Why can't echo in a function work just the same as the statement version?

People have been desperate for this for ages. Several times I have
translated OpenScad functions into Python to debug them. Like many things
in OpenScad the discussion goes round in circles and features that people
have already implemented never get included. There was an echof years ago.
Not sure why it wasn't just called echo, but it worked. If a simple debug
function can't be agreed upon I can't see anything else moving forward.

t-paul:
Using the `echo() expression` style, there would be a nice way to just insert the debug statement into an existing expression. Assuming "normal" style, there's no good way to give it an obvious return value, just defining a random parameter name like `echo(a = "hallo", return = 3);` to be the expression value seems ugly and might even cause issues for the implementation. Always returning 0 or undef also means to find a neutral solution that does not break the original expression defeating the point of easy debugging a bit.

kintel:
It sounds like `echo() expression` without any side effects would be a good middle way for this.

As @t-paul alluded towards, the old `fecho` suggestion was a very, early external hack, which wasn't even close to be merged, but which eventually triggered @t-paul to build a better one himself.

t-paul:
See #1830 for updated `echo()` code, closing this one.

thehans:
Could assert be made to handle multiple expressions, comma separated? 

This code passes without any error, is that intended?  ```assert(true, false);```

krichter722:
> Could assert be made to handle multiple expressions, comma separated?
>
> This code passes without any error, is that intended? assert(true, false);

This can be handled with `assert(true && false)` or `assert(true); assert(false);` and thus isn't compatible with KISS.

Does OpenSCAD have an `assert` statement, now, or not? https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/Other_Language_Features#assert says yes, source build of openscad-2015.03-1-882-ge990ac49e (tag update wouldn't hurt) says `WARNING: Ignoring unknown module 'assert'. `, i.e. no. This discussion above is slightly confuse and the number and complexity of references to pull requests difficult. `echo` prints text, `let` is a scope modifier and `assert` is a do-or-die compile or runtime conditional check. All three have no intersection of functionality and should be implemented as intuitive as possible.

kintel:
@krichter722 It's still experimental and must be enabled from Preferences. See here for the TODO list for next step: https://github.com/openscad/openscad/issues/1860

MichaelAtOz:
> Does OpenSCAD have an assert statement, now, or not?

As the wiki mentions: "[Note: Requires version 2017.01 Experimental Build]"
It is in snapshots, I'm using 2017.01.20 and it's there.

PR #1589, opened 2016-02-25
Correction of example OLD / 23

angerangel:
Example 23 needs library **fonts.cad**, but the path was incorrect. I just added library file in example folder and corrected the path. This way it will work with any OS.


kintel:
The original version should be correct as we're shipping MCAD with OpenSCAD and it should be in the default library path. What issue did you have with this?

angerangel:
the original code uses:
`use <MCAD/fonts.scad>`

but the last version of OpenScad, on linux, doesn't permit to open examples. The open examples button is grey.
So I downloaded the examples in folder:
`/home/max/.local/share/OpenSCAD/libraries/examples/OLD/`
and this way I need to add also the library in the current folder to make work the example

t-paul:
@angerangel Which distribution is that? In some cases (e.g. Debian/Ubuntu) the MCAD library is a separate package.

angerangel:
Ok, you are right. I use Linux Mint (debian based) and the package **openscad-mcad** is a separate package. I installed it and suddenly the examples appeared and they are working correctly.
Thank you

t-paul:
Cool, thanks for reporting back. So I'll close the issue then...

Issue #649, opened 2014-02-13
Unify export GUI [$50]


kintel:
Idea to simplify the GUI, reduce mouse clicks and to reduce duplicated code:
- Create one common export dialog with a keyboard shortcut
- Have an option for choosing file format (either through a GUI element or by filename)
- Remember the folder and settings to make the next export operation smoother
- Look at #648 and see if we can add a checkbox for opening the resulting file after export
- Look into whether exporting for 3D printing (octoprint or external service) fits here
- Support for setting export parameters per file format
  - STL: ascii / binary
  - SVG: units, stroke width, fill color, margins (#3270)
  - PDF: color, ruler on/off, meta data (date, openscad version) (#3266), 

- Optional: Configure the toolbar icon to a user-defined export shortcut, instead of just "Export STL"
- Optional: Configure the proposed export hotkey (F7) to a user-defined export shortcut
- Optional: Add echo output (suggested in scope of the PDF export PR #3266).
##

t-paul:
_Idea from #1386 by @kwendenarmo - moved here to collect export enhancements in on place_


When designing complex parts, it is common to divide them into modules. This option is really great! But when you have to export each to print, for example, must be appointed each of these parts as separate files.
It would be great that when only a single module has rendered, the default name for the export file is that of that module.
This would save a lot of time and could eliminate errors as overwrite files by mistake.

MichaelAtOz:
> I use XP mainly and it is as I described.

@nophead you have upgraded haven't you?

nophead:
Yes I use Win7 now. I still don't like the fact that OpenSCAD has started remembering the last folder I saved in. I always want it to start where my source file is. I will probably have to fork OpenSCAD and put it back the way it was.

MichaelAtOz:
>  I will probably have to [fork](https://en.wikipedia.org/wiki/Fork_%28software_development%29) OpenSCAD and put it back the way it was.

I note IIRC you have had a similar perspective on some other issue recently.

It would be a shame to have what I call the fork of death occurring.

As a minimum, you could have a registry (or what ever the Linux/OS X/Qt equivalent is) setting that, at one setting does it your way, and the other the 'other' way.

I realise that, Firefox's about:config for example, can get out of control, but it may be a good approach.

??

nophead:
[...]

PR #1600, opened 2016-03-10
Restore width & height parameters and use mm in SVG header

### LINKED ISSUE #1597: SVG export no longer includes width or height
As of commit ab338a2c20bf7e94317d43a93e1c013fc531c629, where the export formats were split into separate files, openscad is no longer adding the width or height to exported SVGs.

As mentioned in this forum thread http://forum.openscad.org/SVG-Export-td12216.html ,  If SVGs are exported using openscad, other software will generally assume the SVGs are in pixels (which they normally aren't for laser cutter work - SVGs are commonly preferred for lasers as they are nice closed paths).  

Previously I was able to work around this issue by opening the SVG in a text editor and adding 'mm' to the end of the units in the width and height tag.  This causes all the SVG software to assume all other measurements are in mm.

Any chance of getting this back, and can we reopen the discussion on adding a "Default SVG export units" option somewhere? (I should probably open a separate issue for that?)

t-paul:
Fixes  #1597.

This also sets the export unit to mm which seems to be a bit more useful than the default pixel units.
Ultimately this (and some additional options) should be user configurable (see #649).

### DIFF
--- a/src/export_svg.cc
+++ b/src/export_svg.cc
@@ -78,7 +78,8 @@ void export_svg(const shared_ptr<const Geometry> &geom, std::ostream &output)
 	output
 		<< "<?xml version=\"1.0\" standalone=\"no\"?>\n"
 		<< "<!DOCTYPE svg PUBLIC \"-//W3C//DTD SVG 1.1//EN\" \"http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd\">\n"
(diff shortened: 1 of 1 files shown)


Narann:
lol, I just get why my svg exported files didn't works. Now I simply add `mm` to `width` and `height` and that's perfect!

2015.03-1

PR #1622, opened 2016-04-13
Updated spanish translation

javialamo:
Some corrections to the spanish translation

### DIFF
--- a/locale/es.po
+++ b/locale/es.po
@@ -1,20 +1,20 @@
 # Spanish translations for OpenSCAD package.
 # This file is distributed under the same license as the OpenSCAD package.
-# bazza <bazza@riseup.net>, 2015.
+# <javialamo+github@gmail.com>, 2016.
 #
 msgid ""
 msgstr ""
 "Project-Id-Version: OpenSCAD 2015.03\n"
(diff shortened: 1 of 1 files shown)


kintel:
Thanks!

RedSoxFan04:
I recently tried OpenScad and noticed that most of it is still in Spanish. Is it a bug in the code? Or has it always been like this? Do you need a translator?

t-paul:
Translation issue on MacOS - ref #1813.

kintel:
@RedSoxFan04 Could you try the latest development snapshots?

RedSoxFan04:
I just tried the latest development snapshot, and you're right, the bug has been squashed. When I opened OpenSCAD, the whole program was in English this time. 

It just took me an extra 30 seconds or so to get it to open because MacOS Sierra has this annoying issue where it will only let you install software from the Mac App Store and "Identified developers", but you can always override by clicking "Open Anyway) in the Security & Privacy section of System Preferences.

I'm glad the language bug has been taken care of.

donbright:
see also https://github.com/openscad/openscad/issues/1818 and https://github.com/openscad/openscad/issues/880

PR #1626, opened 2016-04-20
enable/disable folding added

jasleenkaur:
Enable/Disable option for Code Folding  is added in Preferences.


kintel:

Issue #1488, opened 2015-11-15
Saving during compile cancels text output


thehans:
During a long compile I often remember to save via Ctrl-S, but after doing this there are no further compile messages logged to console.
##

kintel:
This is probably due to the hack we use to manage the console output. This could really need a rewrite, perhaps use (or take ideas from) some logging framework? See setCurrentOutput() and clearCurrentOutput in MainWindow (https://github.com/openscad/openscad/blob/master/src/mainwin.cc#L2760)

thehans:
I confirm that this has been fixed.  Closing.

PR #1627, opened 2016-04-20
Do not clear handler after save

anna-marina:
Fixes issue #1488. Now saving action doesn't break log output in window.


kintel:
thx!

Issue #1187, opened 2015-01-28
When Line wrap indentation Style is set to Same, Size option should be disabled


hroncok:
When **Line wrap indentation** **Style** is set to **Same**, the **Size** option should be disabled (greyed out), because it changes no behavior.
##

t-paul:
Closed via #1634.

hroncok:
Thanks.

PR #1634, opened 2016-04-28
Disable spinbox when it has no effect. Fix issue #1187

anna-marina:
How it looks
screenshot_2016-04-28_21-17-06
screenshot_2016-04-28_21-16-38


kintel:
thx!

PR #1637, opened 2016-05-04
Fix issue 214. When a syntax error is reported, we report the source …

doug-moen:
…file name

and the line number within the source file where the error occurred.
This works even if the syntax error is within an included file.

The parse() function, and the places where it is called, have been changed
as necessary to enforce the convention that the second argument is the source
filename, and not the name of the parent directory containing the source file.
Previously, different calls to parse() made different assumptions about
the meaning of the second argument, which probably means include and import
were broken in some cases.

### DIFF
--- a/src/ModuleCache.cc
+++ b/src/ModuleCache.cc
@@ -105,8 +105,8 @@ bool ModuleCache::evaluate(const std::string &filename, FileModule *&module)
 		
 		FileModule *oldmodule = lib_mod;
 		
-		std::string pathname = boosty::stringy(fs::path(filename).parent_path());
-		lib_mod = dynamic_cast<FileModule*>(parse(textbuf.str().c_str(), pathname.c_str(), false));
+        fs::path pathname = fs::path(filename);
+		lib_mod = dynamic_cast<FileModule*>(parse(textbuf.str().c_str(), pathname, false));
 		PRINTDB("  compiled module: %p", lib_mod);
 		
 		// We defer deletion so we can ensure that the new module won't
--- a/src/lexer.l
+++ b/src/lexer.l
@@ -54,7 +54,7 @@ int lexerget_lineno(void);
 static void yyunput(int, char*) __attribute__((unused));
 #endif
 extern const char *parser_input_buffer;
-extern std::string parser_source_path;
+extern fs::path parser_sourcefile;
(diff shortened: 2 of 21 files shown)


kintel:
See the Travis tests - somehow this didn't build on the test server (Ubuntu 12.04). I didn't look at the details.

doug-moen:
Syntax errors now look like this:
```
ERROR: Parser error in file "/res/doug/src/openscad/testdata/scad/bugs/syntax-error-in-line-2.scad", line 2: syntax error

Can't parse file '/home/doug/res/src/openscad/tests/../testdata/scad/bugs/issue214.scad'!
```
(In the above test, the syntax error is in an included file, and the error location is reported correctly.)

I haven't published a unit test for this change. I tried writing an echotest, but the output contains an absolute pathname. Abstracting away the absolute pathname requires a significant change to the test framework, which may be more controversial than the actual change that I'm trying to get approved.

kintel:
I agree that unit tests for this is a bit tricky, so let's save that for later.

If you're interested in digging:
We do have filtering of paths for "file =" parameters. We could extend this to also handle absolute paths in error messages. See https://github.com/openscad/openscad/blob/master/tests/test_cmdline_tool.py#L97

kintel:
In terms of tests: Could you add a few manual test-cases with a README on how to validate?
We do this for some stuff we don't have automated tests for, see https://github.com/openscad/openscad/tree/master/testdata/modulecache-tests

kintel:
@doug-moen I did a quick test, but somehow I cannot get includes to work at all with this PR. It fails with ```WARNING: Can't open include file ...```

I didn't yet look at your code.
As mentioned above, a few manual test cases covering this would make this easier to look into.

doug-moen:
@kintel: Thanks for the heads up. All of my tests so far (and all the automated unit tests, for that matter) have been using the command line, and in that context, `include` works fine. It turns out that `include` is only broken in the interactive GUI case. So I'll fix that.

Also, yes, I realize I need to publish a guide to manual testing before you can accept changes.

doug-moen:
Added 8 manual tests of `include`, `use` and the new functionality. They all pass for both the CLI and the GUI.

kintel:
thx - looks good!

PR #1638, opened 2016-05-05
Backspace unindents feature

anna-marina:
Add backspace unindents option for QScintilla editor
screenshot_2016-05-05_19-36-22


kintel:
thx again :)

PR #1651, opened 2016-05-28
Parameteric window

amarjeetkapoor1:


kintel:
This is now handled in #1751

PR #1653, opened 2016-05-29
Highlight object with source under-cursor, pick object with ctrl-click select its source

mnottale:
This branch adds two feature that help matching the source code and the opencsg rendering:
- If enabled as a show option, the nodes whose source is under the cursor in editor are highlighted
- When ctrl-clicking on the rendering, the source code of the node being clicked is selected

### DIFF
--- a/src/CSGTreeEvaluator.cc
+++ b/src/CSGTreeEvaluator.cc
@@ -131,8 +131,17 @@ void CSGTreeEvaluator::applyToChildren(State &state, const AbstractNode &node, O
 		}
 	}
 	if (t1) {
+	  Location loc = node.modinst->getLocation();
+	  bool cursor_incl =
+	    (cursor_line > loc.first_line
+	      || (cursor_line == loc.first_line
+	        && cursor_column >= loc.first_column))
+	    &&
+	    (cursor_line < loc.last_line
+	      || (cursor_line == loc.last_line
+	        && cursor_column <= loc.last_column));
 		if (node.modinst->isBackground()) t1->setBackground(true);
-		if (node.modinst->isHighlight()) t1->setHighlight(true);
+		if (node.modinst->isHighlight() || cursor_incl) t1->setHighlight(true);
 	}
 	this->stored_term[node.index()] = t1;
 }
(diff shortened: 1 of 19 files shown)


kintel:
Thanks for the pull request - this is functionality that has been on the radar for a while.
A few questions/comments:
* I'm working on a AST node refactoring, see #1641. One of the ideas is to store a source code reference in the top-level ASTNode class, thus offering source code references for all AST nodes, not just moduleinstantiations. I'll look into rebasing your PR on top of my branch. You may want to take a look at the same.
* I notice that you're using "ID color" picking. This might be a bit sensitive to certain rendering techniques/artifacts. Last I looked into this, I preferred picking directly in object space in the CSG tree, but I should look at this again. Did you test this function using various rendering effects, like transparency modes?

Otherwise, this looks relatively clean. The stack fix might warrant a separate PR, as it could be fast-tracked.

mnottale:
Regarding source code location, @$, @1, @2 etc can be used everywhere in the parser, and provide start/end line and column for each element of the rule. It doesn't handle includes though.

About color picking: when in picking mode the opencsgrenderer disables lighting, and replace the setColor call by a glColor(id1,id2,id3,1) so it should be robust.

doug-moen:
The parser does accurately track file name and line number information
while parsing an included file. If your code doesn't handle includes,
should it be updated, or does my change fix the problem? This was committed
3 days ago:

https://github.com/openscad/openscad/commit/0cbef206ef0395ee314978868f1515563f2577eb

mnottale:
@doug-moen: Thanks for the hint, I updated my branch to ignore locations not from the main parsed file (since nothing in the code care about them at this point)

kintel:
I didn't manage to get this to work properly. the highlighting seems a bit off; I'm getting the wrong object highlighted. The source code index is also one character off it seems. The 3D highlighting also works a bit spuriously. That last might be due to a missing redraw after picking.

I haven't had time to dig into this to find the cause of this, but does any of this ring a bell?

kintel:
@mnottale FYI: I've added a Location object to AST nodes in my _ast-refactor_ branch (#1641). As that is probably closer to being merged than this one, it would make sense to rebase on top of _ast-refactor_, to avoid future merge conflicts.

kintel:

This feature eventually got added independently from this PR though.

PR #1654, opened 2016-05-30
Handle RLIMIT_STACK set to infinity.

mnottale:
GNU Make 3.81 on ubuntu 14.04 sets RLIMIT_STACK to infinity, which propagates to the commands being run. This breaks openscad as shown by the "recursion-test-function.scad" test, since (I'm guessing) it uses stack limits to detect infinit recursion.


kintel:
Thanks!

Does anyone else find i curious that Ubuntu 14.04 ships with an ancient version of GNU Make? V3.82 fixed this bug in 2010, but somehow, Ubuntu decided not to package it?

PR #1669, opened 2016-06-18
DO NOT MERGE: GSOC2016

t-paul:
The pull request is currently used only for running the tests on Travis.


src/Assignment.h
```diff
@@ -13,7 +13,7 @@
 
 class Annotation;
 
-typedef std::map<const std::string, Annotation *> AnnotationMap;
+typedef std::map<std::string, Annotation *> AnnotationMap;
```
amarjeetkapoor1: @kintel Why was const removed?

t-paul:
All code is in #1751, so we can close this, right?

kintel:
Yep

Issue #1504, opened 2015-11-24
Add way to compute model volume


Harvie:
Since OpenSCAD is meant for making solid manifold models, i guess there is way to compute volume of such model. I am not sure how hard it's to compute volume of complex objects (like boolean intersection of sphere, cube and cone), but there's probably at least some numeric way for doing this. It doesn't have to be perfect (i know that eg. circle in OpenSCAD is in fact the polygon, so it can't be 100% accurate). But such feature can be quite handy if you will manage to achieve some level of accuracy suitable for practical uses.

Everybody knows how to calculate volume of cube, but calculating more complex shapes can be tricky and i think that OpenSCAD is really perfect tool for such case.

There should be two volumes calculated...
1.) volume of used material (cavities/bubbles buried inside model are subtracted from volume = can be used for calculating weight of model if you know density of used material).

2.) volume of model ignoring the cavities/bubbles = exact volume of water that will be displaced when fully submerged (water will not fill internal cavities).
##

DrDub:
Stochastic computation of volumes via sampling should be doable and it is not that different that solving numerically the closed forms for the shapes.

That'll take care of (1). (2) might need a different method, maybe similar to raytracing.

nophead:
IIRC it can be calculated exactly for a convex polyhedron by tessellating the facets with triangles and joining each triangle vertex to an arbitrary point in the centre to form many tetrahedra, for which there is a formula for the volume. A concave polyhedron can be decomposed into multiple convex ones.

kintel:
@nophead Your first solution also works for concave polyhedrons. If the faces are planar, tessellating them isn't even necessary.

MichaelPFrey:
related to/duplicate of #433

jordanbrown0:
Duplicate of #433.

PR #1388, opened 2015-07-14
Probe

blobule:
A while ago I posted about adding a feature to measure geometry dimensions (see issue #301).
Time has passed, openscad improved, and I still need this feature. So I re-implemented it for the current openscad version.

It adds a new function, probe(), which will analyse the geometry of its first children, set some variables about its bounding box, and then evaluate the remaining children with those variables available. The variables are:
bbempty : true if the geometry is empty
bbmin and bbmax : the minimum and maximum coordinates of the bounding box
bbsize and bbcenter : the size and center of the bounding box

For example, you can do this:
probe() {
    sphere(r=10);
    echo(bbsize);
}
and you shoud see the size (near 10x10x10) displayed. To get the geometry, you have to repeat it:
probe() {
   sphere();
   sphere();
}
So to center only in X the object, simply do this:
probe() {
   object();
   translate([-bbcenter[0],0,0]) object();
}
You can impose fast computation using $exact=false, which will turn off CGAL. In all cases, it must be understood that this function has a computational cost. This is why I call it probe(), to suggest it will do some work, rather than bounds(), for example.

To display the bounding of an object, you can do this:
module see_bbox() {
    probe()  {
        children();
        if( !bbempty ) {
            %translate(bbcenter) cube(bbsize,center=true);
            children(); // display the geometry itself
        }
    }
}

If used sparingly, this function can help do things that are impossible right now because they require to estimate dimensions on complex geometry, which can be mathematically intractable.

Here are a few example of what you can do easily using probe(). It can align objects bounding boxes in many different ways. It makes attaching objects together very easy also.
probe_bbox

It makes "probing" objects possibile: you can intersect with a "probe" and obtain specific measures that can be used to place other objects. For example:

piston

Also, you can do more complex things, like search for complex intersections that are very hard to compute in any other way. For example:
touch

Technically, the probe() function is located in control.cc. It evaluates the geometry just like render, and can process polyset as well as nef_polyhedron representations.

This is the first pull request I make... I have no idea if its going to work or if this is the proper procedure to submit for openscad. In any case, I truly enjoy openscad and I think it really needs that feature. I hope it can be useful.

I will post documentation, examples and tests on request.

Issue #586, opened 2013-12-31
Resize and centering


elalish:
The new Resize command is great, especially for importing STLs, but it doesn't cover the other common problem besides unknown units: unknown origin. I propose adding a "center" argument to resize, which if true, also centers the bounding box of the child object on the origin. 

I think it might also be helpful to have a function called "get_center" that takes any child object and returns the 3-vector of the center of its bounding box. 
##

stemark:
It's not that hard to approximate the center if the piece is at all symmetrical.
Just mirror the piece and then mess with the offset until the two images align.

If you really want to get close then you can difference() them. 


module piece(){
    import("file.stl");
}

off=[-10,-10];

%translate(off)
    platform();

mirror()
%translate(off)
    platform();

jordanbrown0:
PR #4478 will let you (slowly, for complex assemblies) retrieve geometry information - like resize, but making the information available to userspace.

Issue #301, opened 2013-03-10
Functions that can query a given 3D shape and provide size or location info


filipmu:
All openscad functions take parameters as input and produce shapes.  It would be useful to have functions that take shapes as input and produce parameters.  This would be useful for aligning shapes.  Some examples are max/min dimensions on each axis, center of gravity, etc.  There are a number of uses of this feature: 1-since all shapes in openscad are polyhedrons, spheres and cylinders do not render exactly as specified.  So there is no way to exactly align shapes based only on input parameters. 2-To align two objects created by separate modules, the user needs to know the inner workings of each module to calculate how to align them.  3- Application of complex transformations like Minkowski and Convex Hull or many CSG combinations create shapes with very complex boundaries that can't be easily calculated.
##

lordofhyphens:
#1155 might be a useful step towards a solution, especially if you allow for iterative compilation.

MatthiasWM:
RFC: The `probe()` solution seems to have too many side effects and non-obvious variable names. I would like to suggest the following interface:

The `union()`/etc. functions could have optional arguments that work in a "call by reference" way, for example:
```
module someGeometry() { ... }

// render the bounding box of two random geometries
my_bbox = [ [0,0,0], [0,0,0]];
// store the bounding box data in my_bbox as an array of two vectors
union(query_boundingbox=my_bbox) {
  someGeometry();
  someOtherGeometry()
}
%translate (my_bbox[0]) cube(my_bbox[1]);
```

Positives:
* no change in syntax
* caching works
* `union()` call can provide many more attributes
* query many bounding boxes without collision of variable names

Negatives:
* `query_boundingbox=my_bbox` my be surprising; users may expect the left side to be the one that receives a value; we could reverse the order (not logical) or add a syntax, i.e. `boundingbox->my_bbox`
* `union()` returns a geometry that will be rendered; we need an additional method 'query()' that performs a 'union()' call without returning any geometry

tcurdt:
Indeed `query_boundingbox=my_bbox` is surprising and not quite so obvious.
I have no idea about the internals but I assume something like this is not an option?

    bbox = union() {
      ..
    }

MatthiasWM:
I have tried `bbox = union() { ... }`, but it would require a change in syntax, possibly failing existing scripts. A few alternatives that are syntax compatible, or at least not entirely disruptive:
```
myBBox = [ [0, 0, 0], [0, 0, 0] ]; // not needed, just illustrating the returned format
[...]

PR #1713, opened 2016-06-26
Probe (bounding box, volume, centroid)

blobule:
A  new attempt at making a valid pull request for probe() functionality.


blobule:
Hmmm.... seems that this attempt passed the comlete build, but failed testing for the 3 examples I added (examples/Probe/...).
I did generate the "expected" png results and added them (tests/regression/probe*).
The tests are passing without problem here (ctest ...).

Any suggestion on how to proceed to get these examples to pass?

t-paul:
Did you add all the files? Easiest solution is to use `TEST_GENERATE=1` as mentioned in the docs.

I'm getting:
```
probe_volume_1
Error: test 'dumptest-examples' is missing expected output in /data/OpenSCAD/openscad-master/tests/regression/dumptest-examples/probe_volume_1-expected.csg
 actual image: /data/OpenSCAD/openscad-master/tests/dumptest-examples-output/probe_volume_1-actual.csg

 expected image: /data/OpenSCAD/openscad-master/tests/regression/dumptest-examples/probe_volume_1-expected.csg
```

blobule:
I see...
In the docs/testing.txt, it says use "TEST_GENERATE=1 ctest -C Examples -R your_example".
This will not generate files under regression/dumptest-examples/, but only files under regression/throwntogethertest/,  regression/opencsgtest/, regression/cgalpngtest/.

running "TEST_GENERATE=1 ctest -R probe_bbox" (without -C Examples) provided the missing file under regression/dumptest-examples/

Maybe docs/testing.txt could be adjusted to reflect this?
Thanks for the help. I will resubmit the pull request.

blobule:
In an attempt to answer the need for the computation of bounding box, volume and centroid, here is my proposal for a new command: probe().

See previous discussions about this: pull request #1388 , issue #1504 , issue #586, issue #301, and in the forum at http://forum.openscad.org/Volume-and-Center-of-mass-td15421.html

probe() will analyse the geometry of its first children, then define preset variables (empty, bbsize, bbcenter, bbmin, bbmax, volume, centroid) which can be used during the evaluation of its subsequent children. The variable only exists for the children of probe(). The first children is never added to the render tree. The volume/centroid is only computed on demand, with parameter volume=true.

The bounding box is always fast to compute. The volume can be fast or slow. It relies on a nef-polyhedron decomposition into convex shapes, so the more non-complex is your shape....

To print the volume of a geometry, just do
```
probe(volume=true) {
    someGeometry();
    echo("bounding box size is ",bbsize, " and volume is ",volume);
}
```
You could illustrate the bounding box of any geometry with this module:
```
module showBBox() {
    probe() {
        children(); // what is probed
        children(); // display the geometry itself
        color("purple",0.3) translate(bbcenter) cube(bbsize,center=true); // the bounding box
    }
}
showBBox() someGeometry();
```
Obviously, it make it extremely easy to center/resize/align any geometry. For example, this will center the geometry according to the bounding box:
```
module centerize() {
    probe() {
        children();
        translate(-bbcenter) children();
    }
}
centerize() someGeometry();
```
It is possible to go much further and align two geometries on any axis and with any anchor point. This is ideal to put place text, and to manage stl imports.

Using recursion, probe() makes it possible to do the following things:
- find the correct parameter to get a geometry with a specified volume
- find the correct parameter to locate the intersection of complex geometries
- optimize a model to get it to "stand up" (i.e. make sure the centroid is above its base)

Anyway, this pull request is not very complete at the moment. More examples are needed and, of course, documentation. However, I hope it can be tried so we can discuss this feature and see how if it is a worthwhile addition to openscad.

Yona-Appletree:
Thanks for creating the pull request. I think it's an essential feature, however, I'm not convinced this is the right way of going about it. See [my comment](https://github.com/openscad/openscad/issues/301#issuecomment-141852823) on #301 about bringing geometry back into variable space with `render()`.

This method only exposes some useful variables about the model, not the actual geometry. Additionally, it can only be invoked from module space. I would be happier with an option to probe that exposes the geometry in a probe_geometry special variable or similar. But it would still not be callable from functions (e.g. no function centerOf(points)), which is unfortunate. 

That all being said, if the maintainers would prefer this over a full merging of function / module space (without a complete language overhaul), I'd much prefer this over where we are now.

Patola:
When is this feature added to OpenSCAD? There is a huge crowd waiting for it. It is desperately needed. Hurry
(Don't take this too serious, and btw, amazing work, @blobule )

mvelasco93:
No plans on adding this?

blobule:
Hi,

Sorry, I got tired of asking and justifying why measuring bounding box,
centroid and volume is a good thing in a 3D parametric CSG system like
Openscad.

If there is renewed interest, I'll gladly update my fork of openscad
and send a new pull request. But I need assurance that I won't do all
this to be told "we don't want it".

Sincerely,

Sébastien

Le lundi 18 décembre 2017 à 02:54 +0000, Made Velasco a écrit :

kaklik:
Hi! I want this feature. I need it for precise computation of aerodynamics models. I think it could be the demonstrable application of this function.  It is because of the surface of models is the complex function itself the function is approximated by polygons.  But the volume, center of gravity and other related parameters need to be known precisely.

blobule:
Hi,

Le jeudi 03 mai 2018 à 14:04 -0700, Jakub Kákona a écrit :
> Hi! I want this feature. I need it for precise computation of
> aerodynamics models. I think it could be the demonstrable application
> of this function. It is because of the surface of models is the
> complex function itself the function is approximated by polygons. But
> the volume, center of gravity and other related parameters need to be
> known precisely.

Thanks for your interests in this feature. I also think it is a must
for Openscad. If the maintainers of openscad show interest, I'll send a
new pull request and we can integrate it.

Otherwise, I put it in a lua-based version of openscad that I made,
just for fun, checking how the lua syntax can be used to emulate
openscad...


Sebastien

rjp:
I would also like this feature - enough that I'm considering building my own version of OpenSCAD just to include this patch. Baffling that you can't measure geometry in a parametric system.

blobule:
Hi everyone,

On Tue, 2019-10-01 at 01:25 -0700, Michael wrote:
> The issue is that it would require a render, as for preview the
> geometry is not calculated.

As I am the one who implemented this feature a while ago. I would like
to reafirm that features such as probe have a deep impact on what can
be modelled in openscad, so it should be considered for inclusion, even
if a render is needed to compute it.

We all know that openscad has no way of positionning geometry relative
to other geometry, and this is a severe handicap.

The first example that comes to mind is the import of external STL
geometry, which is impossible to automatically center or scale.
Some shy attempts have been made to tackle this...  for example text()
has alignment options, but there is still no way to know what is the
bounding box of some text string. This is not acceptable when you
consider that the string might be a parameter, which can change. After
all, openscad is advertized as ideal to make parametrized models...

I am sorry for my pessimistic tone, but my probe() pull request was
rejected years ago, and I will only update and resubmit if I am told it
is a desired addition to openscad. By the way, functions like probe(),
volume(), centerOfMass() and render() could be highligted in red, just
to illustrate the fact that they are more computationaly expensive (I
would make minkowski red also...). Such functions, even if they are
computationaly expensive, are very very powerful. Reducing a language
expressive power for the sake of a fast preview is, in my opinion, not
a good compromise and hurts the language in the long run.

In the mean time I implemented my own openscad "clone" with lua as a
base language and two dedicaded modules: "nef" for rendering and
"opencsg" for previewing. Works very well, and it has boundingBox(),
volume(), centerOfMass(), ... and you can use +-*/ for CSG operations
:-)

Sorry for the long message. Its long because I truly appreciate they
"openscad way of modelling", and care about improving it.

Sincerety,

Sébastien

nophead:
Yes I can't see why your code can't be merged because it doesn't make render any slower, it just makes it more useful. People already use render in preview and know the consequences. For 2D objects it is negligible so could easily work for text.

t-paul:
Related forum post: http://forum.openscad.org/Difference-between-modelling-with-Openscad-and-Freecad-tp27412p27453.html

martin-frbg:
Another vote for resurrecting and merging this. I just spent a few hours with Blender re-coloring a complex geometry that openscad would only export as STL, and I know the issue is simply that it contains some edge-sharing polyhedra. As I understand the proposed functionality, it would make it trivial to scale these around their centroid so that they no longer touch (while still keeping this impression in a 3d-printed model)

CaptainFalcon92:
Hello people. Recently i come in the need of calculating the volume of an object made in openscad. 
First i would like to thanks and congratulate all the peoples involved in making openscad, what a great tool ! It has been years since this pull request opened, but still i think having such powerful functions could really bring openscad to an other level.
Although i would like to get volume from CLI (command or flag), i sincerely think this topic deserve to be rebirth.
Maybe you guys could work on such a powerful feature together and find the sweet spot ? :)
I'm unfortunately too ignorant in this program sources and internal to be of any help, but encouragement.
I wish this find it's way in, as is, or maybe rebuilt differently idk..

Meanwhile, do someone have an advice on how i could compute a part's volume ? Thanks for reading.
Best regards.

blobule:
Hi!

As we know, the pull request was rejected on the basis that volume (and
center of mass and bounding box) require the equivalent of "render()"
to perform the computation. This was considered a danger to make
openscad slower than it already is...

My view on this is that when you really need volume/center of
mass/bounding box, you dont care to wait for a render.

So if I get a clear "we wont reject a pull request for volume/center of
mass/bounding box", i'll update the code and make a new pull request
immediatly.

Sincerely,

Sébastien
DIRO/FAS/Université de Montréal

kintel:
Is it even possible to calculate the center of the bounding box without performing a render?

blobule:
Hi Marius,

On Sat, 2020-02-08 at 18:36 -0800, Marius Kintel wrote:
> Is it even possible to calculate the center of the bounding box
> without performing a render?

Usually, any computation which requires accurate knowledge of the shape
will require a (slow) render.

However, now that you mention it, bounding box is a special case... By
rendering the object using OpenCSG (the fast "preview" mode) offscreen
from three directions (X, Y, and Z axis, respectively), with
orthographic cameras, it is perfectly possible to get the XYZ bounding
box of an object... But bounding box is not a very accurate measure. At
least it would be computed fast.

One very useful of fast bounding box computation would be to detect if
an object is empty. This makes it possible to detect when parts
overlap, for example, and adjust recursively positions until objects
appear to be in contact (some kind of "slide object X in direction Y
until it rests on another object").

This needs testing.... If I can find some time, I'll check it out. Mybe
we could start with those fast operations, then do the slow ones.

Sebastien

jordiblasco:
Hi @kintel,

I'm really happy with OpenSCAD but this is kind of a mandatory feature for several engineering and physics projects.

I'm currently rendering the STL and using Numpy-STL the center of gravity
https://numpy-stl.readthedocs.io/en/latest/usage.html#evaluating-mesh-properties-volume-center-of-gravity-inertia

To be honest, I would support to have the implementation developed by @blobule as it is. You don't need to calculate the centre of gravity every time you make a small change, in the same way, you don't need to render the whole model after every single small variation. What about an opt-in option?

I'm kind to get involved in the development in order to make this happen. Actually, I'm working on a model that has different materials with different densities. So, it could be nice to extend this implementation to density-aware structures.

Thanks, @blobule for your contribution!

Jordi

t-paul:
There is no disagreement regarding the feature.

The question is how to fit this into the language in a way that is not based on magic variables. There were a number of suggestions in the discussions, some are linked above. It should not be needed to call a big black box function with lots of opt-in features. The better strategy would be features that behave the same but can be combined to be more powerful ([Composable Building Blocks](https://github.com/doug-moen/openscad2/blob/master/rfc/Composable_Building_Blocks.md)). 

Also an important requirement for this type of features is to have enough tests for all cases to prevent accidental changes. The test suite is an important part of making sure OpenSCAD can evolve at all.

blobule:
Hello everyone,

After seeing recent requests for a feature providing shape information
(bounding box, center of mass, volume, ...), which I implemented
experimentaly as "probe()" a while ago, I am interested in making these
feature work properly in openscad.

The main problem is this: how to you return a scalar (or vector) value
computed on a shape?
Openscad offers functions, which do not take a geometry as parameter,
and geometry operators, which must return geometry, not scalars.

One solution, which is not great I admit, is magic variables...

probe() geometry(); // does render() if needed and set magic variables
echo(boundingBox);
echo(centerOfMass);
echo(volume):

May I suggest another approach?
We could rely on module to "name" the geometry:

// Any module without parameters is actually "naming" this geometry.
// Name this geometry "test"
module test() { geometry(); }

echo( test.volume )
echo( test.boundingbox )
echo( test.centerOfMass )

When you call a module without (), then a render() is be performed on
that module called without parameters. The resulting geometry would be
used to compute some value which is returned.


So... magic variables? functions on named geometries?
Any other way?

Sébastien

t-paul:
There is a discussion based on the OpenSCAD2 proposal in #3087 / #3088.

blobule:
Openscad2 is a future project, right?
So I assume that there is still no interest in this feature for the current openscad?

t-paul:
No, it's a path to an improved language, not a coding project. Quite a number of things suggested there are already integrated into the dev snapshots and more to come step by step. Most importantly function literals are working, making functions first class citizens. It also serves as guidance to go for things like composability instead of creating monsters like `search()`.

soychicka:
I'll chime in and lend my support for integration of even a slow feature that requires render() to retrieve bounds of a given 3D component.

Heck, make it disabled unless you explicitly enable in preferences or require a flag at application launch...
 
For non-random generative applications that don't have frequently changing parameters - esp. when I'm integrating objects into a complex assembly from STL or other libraries that may change when I change local parameters (i.e. if I'm using Customizer), at least let me call `inspect()` (my suggestion) to return the current coordinates, bounding box and center of the object of interest, PLEASE?  Maybe store the most recent value in customizer instead of some invisible magic variable?

Meantime, to compensate for the absence of this functionality, I implemented my own OO ruby-based command line tool to compile to openscad (and allow for the use of variables to reference identical objects that get mirrored across axes, etc).

I'd strongly suggest reconsidering this prior to implementation of Openscad2... require manual enabling, go ahead and make it more efficient later on - but this is crazy-essential for a lot of us.

The time consumed by implementing workarounds and to maintain them as we update our language version is undoubtably >> the time it would take to do a low-fidelity render($fn=5) call.

(If pentaradial symmetry is good enough for echinoderms, it's good enough for returning bounding box dimensions in most use cases 🤓)

t-paul:
Again, OpenSCAD2 is __NOT__ some kind of future project or version which could be blocking other things, it's pretty much the opposite. It's a guideline how to integrate new features without breaking backward compatibility. So if at all, it exists to help integrating features like this one.

The argument "make it better later" does not work for this kind of changes. Once a language feature is released it's hard to change. If people would like to see that feature, I'd suggest helping with the design by discussing ways for integration.

blobule:
Hi,I was told a few times, and I understand, that there are no great ways to integrate geometry information (bounding box, volume, centroid and more) into the openscad language, openscad2 or not.I proposed a few approaches, all rejected. So anyone has any suggestion on how to return computed information on geometry in openscad without breaking compatibility? How about creating a new module info() which returns not geometry but numeric results and could be used in expressions?V=info("volume") sphere();Openscad needs a bridge from geometry to expressions... Can openscad2 offer a guideline for this?Le 11 août 2020 11 h 41 a.m., Torsten Paul <notifications@github.com> a écrit :
Again, OpenSCAD2 is NOT some kind of future project or version which could be blocking other things, it's pretty much the opposite. It's a guideline how to integrate new features without breaking backward compatibility. So if at all, it exists to help integrating features like this one.
The argument "make it better later" does not work for this kind of changes. Once a language feature is released it's hard to change. If people would like to see that feature, I'd suggest helping with the design by discussing ways for integration.

peta:
I cannot understand how the work and engagement of @blobule in order to provide a solution to the need of probably most typical OpenSCAD users, can be blocked and deferred for such a long time now, in favor of some pointless academic theories. 

Do you want to have a dinosaur that will die out in a few years, or a living and evolving species that, even when not evolution's perfect product, will change and adapt to actual *real world* challenges?

When you cannot (after all the years) settle on merging contributions like blobules to the main code base, then at least provide an extension mechanism that allows easy *plug-ins* of user language extensions.

combs:
This'd be really nice to have as a user.

davrandom:
In the meantime it is 2022 and (afaik) there is still no way to calculate a volume in OpenSCAD. Even if the code is there.

t-paul:
More useful feedback would be looking at #3956.

blobule:
Hi, and happy new year to all OpenSCAD enthousiasts.

Le 2022-01-01 à 12 h 29, davrandom a écrit :
>
> In the meantime it is 2022 and (afaik) there is still no way to 
> calculate a volume in OpenSCAD. Even if the code is there.
>
> Message ID: ***@***.***>
>
Yes... my pull request was in 2016...  time flies :-)

I will not submit a new pull request for now, as I feel that consensus 
has not been reached about how to incorporate geometric measures in 
OpenSCAD.

However, if anyone want to use the feature, I made a fork with the 
probe() feature, with examples : 
https://github.com/blobule/openscad/tree/probe 
<https://github.com/blobule/openscad/tree/probe> (make sure you use the 
*probe* branch ).

openscad examples

The probe() module does the following:

  * compute the bounding box. Great for resizing/aligning
    anything,including imported STL.
  * identifies empty or non empty geometry (to detect contact between
    objects)
  * optionally compute volume and center of mass

Probe makes it possible to perform a binary search for making two 
objects touch each other, for example.

You must remember that probe() can' t compute anything without first 
doing a "render", so you will lose some speed on the F5 preview. 
However, the cache works fine, so its not that bad. In any case, probe 
provides info that is usually worth waiting for... just make sure you 
use it sparingly, only when needed.

I welcome feedback on this, and I will be happy to discuss how to 
properly integrate this into OpenSCAD permanently, if the community 
wants to go that way.

Best regards,

Sébastien

amatulic:
I'm **less** interested in a `probe()` command and **more** interested in having volume, bounding box, and centroid information output with all the rest of the statistics one already gets when rendering. You know, the part that you see in the console window that looks like this:
````
   Simple:        yes
   Vertices:      491
   Halfedges:    1658
   Edges:         829
   Halffacets:    684
   Facets:        342
   Volumes:         3
Rendering finished.
````
Just a few additional staistics can be added to that list. I don't care if the volume, centroid, etc. are for all the parts being rendered at once. If I want to see it for a single part, I can just suppress the others.

**Comment by jordanbrown0** [comment]
PR#4478 provides this feature.

**Comment by amatulic** [comment]
@jordanbrown0 I don't see anything in PR #4478 about volumes or centroids, only bounding boxes.

**Comment by jordanbrown0** [comment]
Ah, sorry.  I missed / forgot the "volume" part, and confused "centroid" with "center".

But it's at least a step closer.

I believe there are techniques that can be used to derive the volume from the list of triangles, but I am not enough of a mathematician to be sure.
E.g.: https://mathworld.wolfram.com/PolyhedronVolume.html

**Comment by jordanbrown0** [comment]
Applying that algorithm directly to the output from 4478's render() is awkward because it wants triangles and 4478's render() can yield non-triangles.  As the Wolfram page says, you can triangulate arbitrary polygons, but that's tedious.

Well... I tried a cheesy triangulation algorithm against a couple of simple shapes, and give or take the fact that a single-object model gets the sign of the volume wrong, it looks right.  (I suspect that the reason that it gets the sign wrong on single-object models is that CGAL doesn't actually get involved, and the non-CGAL result is in some way backwards.)

But again I'm not really enough of a mathematician to be sure that what I'm doing is right.

Here's that program.  Of course most people don't have a 4478 build and can't run it, but you could build your own faces array.

```
g = {{ cube(10); translate([20,0,0]) cube(10); }};
r = render(g);

function sum(a, i=0) = i >= len(a) ? 0 : a[i] + sum(a, i+1);
function triangulate(a) =
    len(a) < 4
    ? [ a ]
    : [
        each triangulate([
            a[0],
            for (i=[2:len(a)-1]) a[i]
        ]),
        [a[0], a[1], a[2]]
    ];

points = r.objects[0].points;

faces = [
    for (f = r.objects[0].faces) [
        for (pnum = f) points[pnum]
    ]
];

tfaces = [
    for (f = faces) each triangulate(f)
];

function curl(faces) = 1/6*sum([
    for (f = faces) f[0] * cross(f[1]-f[0], f[2]-f[0])
]);

echo(curl(tfaces));
```
yields:
```
ECHO: 2000
```

jordanbrown0:
https://mathworld.wolfram.com/PolyhedronCentroid.html
gives a sort of similar function for centroid, but unfortunately I don't understand all of the notation it uses.  I've asked for help.

jordanbrown0:
Hmm.  That triangulation algorithm only works for convex polyhedra.  Still, there are other triangulation algorithms that actually work.

jordanbrown0:
> Hmm. That triangulation algorithm only works for convex polyhedra. 

Oops, polygons.

blobule:
Hi everyone, its nice  to  see  that  volume/centroid/bounding box  might  finally be added to  openscad.

In  probe(), in  order  to  compute the volume, you must first  decompose the  nef-polyhedra into a union  of  convex parts.  Then  you add  the volumes,  computed on each convex  shapes.

Issue #1647, opened 2016-05-15
Kill Boost Filesystem V2 workarounds


kintel:
We have a bunch of workaround for compiling against Boost Filesystem V2.
V2 was replaced with V3 as default in Boost 1.46. I don't believe we support building on any systems which use Such an old version of Boost any longer, so we can start killing off these workarounds.

See https://github.com/openscad/openscad/blob/master/src/boosty.h

@donbright I believe you're the original author of boosty.h. If you have any opinions, feel free to chip in.

donbright:
" in a few years, this file should be deleted as unnecessary."

The Prophecy has come to pass... the Wizard has arrived to heal the Codebase... all praises to the Open Source...

donbright:
just to be clear... yes, take it all out, remove it.. . . i would have done it myself by now but I was waiting to hear what @DrCodeWizzard  to do a pull request or make a comment...

DrCodeWizzard:
sorry I've been a bit busy with my regular work. I thought i already merged this in into the code base? Also while developing this change i realized that not all of boosty.h can be removed due to version of boost installed on the test rig.

kintel:
Yes, this was already merged from #1717. I just forgot to close this.

PR #1717, opened 2016-06-28
 #1647 remove workarounds for older versions of boost

DrCodeWizzard:
another config

another one

### DIFF
--- a/cgal/decompose.cpp
+++ b/cgal/decompose.cpp
@@ -1,4 +1,5 @@
 #include <boost/foreach.hpp>
+#include <boost/filesystem.hpp>
 #include <boost/regex.hpp>
 #include <sstream>
 #include <iostream>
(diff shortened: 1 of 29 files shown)


DrCodeWizzard:
any ideas why these tests are failing? On my local linux box they all pass.

t-paul:
Xvfb did not start, seems to happen now and then lately. I've re-triggered the Travis run.

DrCodeWizzard:
How do i re-trigger travis for future reference?

t-paul:
There's a restart button on the top right, but I don't know if this is tied to some access control. I suspect it's restricted to repository team members.

DrCodeWizzard:
how do i get this merged in?

kintel:
Could you write a sentence about what you did, and what's remaining (e.g. required boost version)?
-> then I can review it and merge.

DrCodeWizzard:
I have removed support for boost version earlier than 1.46. I have removed all calls to boosty::* methods from the project which have direct support in boost version 1.46 and greater. I have not moved minimum boost support requirements to 1.48 as that requires modifying travis configuration extensively. 

I have removed from boosty.h definitions of all methods which were required to support boost versions prior to boost v1.46

PR #1726, opened 2016-07-03
support of negative numbers in parameters in slider

amarjeetkapoor1:


amarjeetkapoor1:
@kintel Was there any specific reason that you added in progress label?

kintel:
@amarjeetkapoor1 It's an automated label from waffle.io: https://waffle.io/openscad/openscad
I've been testing using that, but haven't really gotten into it, as our development doesn't really fit scrum-style these days.

PR #1758, opened 2016-07-26
Dependency changes: New Qt check & Arch support for getting deps.

lewisjb:
- New way to check if Qt is installed
- Arch is now supported for getting dependencies

---

Previously `qt_sysver` was using `qmake -v` just to check if 5 or 4 was installed, so this PR changes it to extract the actual Qt version from that which simplifies the process, and works on the latest version. (The script wouldn't work for me as `QT_VERSION_STR` wasn't in `qglobal.h`.

I also added Arch support to `uni-get-dependencies.sh` as I use Arch and there wasn't any support for it previously


donbright:
this looks good to me, as long as the conflicts can get resolved.

lewisjb:
@donbright fixed

kintel:
Quick question: How does this relate to #1772 ?

lewisjb:
From the Qt side, both this PR and that one try to solve the problem that the Qt check doesn't work anymore.
I'd say merging both (after the merge conflicts) would give the best results as that one found the new location for the version string which helps when they don't have `qtchooser`, however I believe my solutions handles them having `qtchooser` much nicer (as instead of asking for the location, then finding the version string, you just ask it for the version string. Thus should be more efficient and won't have an issue if they move its location again or something like that).

If you merge theirs I can resolve the merge conflict on my side

kintel:
@lewisjb Sorry for the delay. I've merged #1772. If you could merge any conflicts with this PR, we can move forward with this as well.

lewisjb:
@kintel No problem! It appears to be failing the build, but I have no idea why, as the logs don't have regarding the script

t-paul:
The build failure on Linux should be fixed by #1819 I just merged. The macOS error seems to be caused by Travis environment update to 10.11.

lewisjb:
@t-paul Ahh right, cheers. Do I need to do anything else or is this ready to be merged?

kintel:
@lewisjb Is this good to merge? @t-paul feel free to merge if you deem it good.

lewisjb:
@kintel Yeah, it's good to merge

Issue #801, opened 2014-05-19
Minor. Edit/Undo does not grey-out when nothing to undo


MichaelAtOz:
So you don't know whether you have more to undo to get back to the original.
Many programs have it greyed out when nothing to undo.
Low priority.

kintel:
Fixed by #1779

PR #1762, opened 2016-08-03
Fix issue #801

onewhl:
Fixed issue #801. Made Undo button disabled when nothing to undo.
mainwindow


kintel:
Thanks for the pull request.

One remaining issue:
If I select all and remove all the text, I cannot undo that, since you disable the button when the source code is empty. Is there a better way of checking if undo should be available?

onewhl:
@kintel yes, you right, I will fix it.

kintel:
Replaced by #1779

Issue #962, opened 2014-09-29
Close editor button appears to belong to Search and replace toolbar


hroncok:
I've just closed the editor instead of the Search and replace toolbar. The Done button in the toolbar does not provide very good UX.

![scadfind](https://cloud.githubusercontent.com/assets/2401856/4437532/5ea7de08-479a-11e4-8fad-94b49e07e5f5.png)
##

MichaelAtOz:
I also fall for this, reopening.

PR #1767, opened 2016-08-11
Redesign find and replace panel

onewhl:
Current design of panel is not usability(see issue #962), so I suggest redesign.
How it looks:  
_Find..._
find

_Find and replace..._
findandreplace

### DIFF
--- a/src/MainWindow.ui
+++ b/src/MainWindow.ui
@@ -15,7 +15,16 @@
     <property name="spacing">
      <number>0</number>
     </property>
-    <property name="margin">
+    <property name="leftMargin">
+     <number>0</number>
+    </property>
+    <property name="topMargin">
+     <number>0</number>
+    </property>
+    <property name="rightMargin">
+     <number>0</number>
+    </property>
+    <property name="bottomMargin">
      <number>0</number>
     </property>
     <item>
@@ -30,7 +39,16 @@
        <property name="spacing">
         <number>0</number>
        </property>
-       <property name="margin">
+       <property name="leftMargin">
+        <number>0</number>
+       </property>
+       <property name="topMargin">
+        <number>0</number>
+       </property>
+       <property name="rightMargin">
+        <number>0</number>
+       </property>
+       <property name="bottomMargin">
         <number>0</number>
        </property>
        <item>
@@ -162,7 +180,16 @@
           <number>0</number>
          </property>
          <layout class="QHBoxLayout" name="horizontalLayoutAnimate">
-          <property name="margin">
+          <property name="leftMargin">
+           <number>0</number>
+          </property>
+          <property name="topMargin">
+           <number>0</number>
+          </property>
+          <property name="rightMargin">
+           <number>0</number>
+          </property>
+          <property name="bottomMargin">
            <number>0</number>
           </property>
           <item>
@@ -245,7 +272,7 @@
      <x>0</x>
      <y>0</y>
      <width>1118</width>
-     <height>22</height>
+     <height>19</height>
     </rect>
    </property>
    <widget class="QMenu" name="menu_File">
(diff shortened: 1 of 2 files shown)


kintel:
@hroncok @t-paul Any comments on this? It looks cleaner to me, so I'll merge unless you have any opinions.

t-paul:
The change does not really address the issue raised in #962, but I don't have a good idea how to solve that except maybe moving the whole find panel to the bottom of the editor.

Apart from that, I agree it looks a bit simpler, even though the texts take quite some screen space. So I'd vote for merging.

MichaelAtOz:
A line across could help (with the above), similarly with the console  (it also provides a visible anchor to drag) 
Should also do something similar to allow gui close of the animate pane.?
gui line

t-paul:
How the title bar is drawn, is up to the selected Qt theme, I don't know if there's a safe way to change things there without the risk of breaking other themes.
Maybe we could try to generate a very thin border around the title bar?

kintel:
Yep, this doesn't address #962, but it still a worthwhile usability upgrade.

t-paul:
Hmm, that broke the nightly builds with the UI file property `clearButtonEnabled`.

```
[  543s] In file included from src/MainWindow.h:6:0,
[  543s]                  from src/openscad.cc:572:
[  543s] objects/ui_MainWindow.h: In member function 'void Ui_MainWindow::setupUi(QMainWindow*)':
[  543s] objects/ui_MainWindow.h:640:25: error: 'class QLineEdit' has no member named 'setClearButtonEnabled'
[  543s]          findInputField->setClearButtonEnabled(true);
[  543s]                          ^
[  543s] objects/ui_MainWindow.h:661:28: error: 'class QLineEdit' has no member named 'setClearButtonEnabled'
[  543s]          replaceInputField->setClearButtonEnabled(true);
[  543s]                             ^
```

t-paul:
Ahh, yes: ["This property was introduced in Qt 5.2."](http://doc.qt.io/qt-5/qlineedit.html#clearButtonEnabled-prop)

I guess we can just remove it for now...

kipbits:
the old design was better(at least on mac). is it possible to change it back?

kipbits:
[#626](https://github.com/openscad/openscad/pull/626) / [#628](https://github.com/openscad/openscad/pull/628) had a better feel on mac. is it possible to switch back. 

also I was thinking a big improvement for find-replace would be to have a find-count, and highlighting of all found words in a different color from normal cursor highlight. I was in the process of trying to maybe do my own when i discovered this thread. 

the feature i am interested in is still not implemented...

kipbits:
I've reverted the design on my local machine(with a few minor changes); still working on adding improvements. I had some confusion with the [qscintilla](http://pyqt.sourceforge.net/Docs/QScintilla2/) editor vs the legacy([qtextedit](http://doc.qt.io/qt-4.8/qtextedit-members.html)) editor. It appears that all functionality is handled by [scintillaeditor.cpp](https://github.com/openscad/openscad/blob/master/src/scintillaeditor.cpp), and that [legacyeditor.cc](https://github.com/openscad/openscad/blob/master/src/legacyeditor.cc) no longer gets called on for anything? I would like to try and add find-count, and highlighting-of-all-found-matches. I think this can be done with [scintilla](http://www.scintilla.org/ScintillaDoc.html) using [Markers](http://www.scintilla.org/ScintillaDoc.html#Markers) and [Indicators](http://www.scintilla.org/ScintillaDoc.html#Indicators). Might take me abit of time so I'm posting this here, but if knowbody has any objections I'll see what i can do... expect a PR.

kintel:
@kipbits The idea is to eventually kill the old editor. We just wanted to make sure there aren't any showstopper bugs in the scintilla one. Having the old editor available makes it easy to test if a particular issue is related to the editor itself or of it's an issue with another component.

kipbits:
OK, I finally got around to working and making some changes here. I have highlighting-match-words working now, but i'm still working on the find-count display. Also, I made a few minor changes for sake of code clarity. I'm going to try and do it all in one big PR, unless find-count turns out to be too difficult
screen shot 2017-01-29 at 11 41 17 am

kipbits:
continue discussion: https://github.com/openscad/openscad/pull/1922

PR #1771, opened 2016-08-12
Fix replace function: when selected string is empty do not replace 

onewhl:
I found the bug. Follow the steps to repeat:

1) **Edit** -> **Find and replace...**
2) Enter **Search string** 
3) Enter **Replacement string**
4) Click **Replace**
_When you finish replacing all strings, go to 5)_
5) Click **Replace** again 
6) And again and again

You will see that replacement string added after ever click. 
It happened because selected empty string. Need check selected string is empty or not  and if yes do not replace


kintel:
Thx!

PR #1772, opened 2016-08-14
fix: Qt moved QT_VERSION_STR to qconfig.h

vesim987:


kintel:
@t-paul Does this look good to you?

PR #1778, opened 2016-08-20
Fix finding qscintilla2.prf on ubuntu.

CarloWood:
qscintilla2.prf is part of package libqt5scintilla2-dev
but not found without this patch, as it is installed in,
for example when using qt5, /usr/share/qt5/mkspecs/features.
However, since this path completely depends on which qmake
is in your PATH, one should really just use qmake (variables)
to find the prf files.

prf files are installed in
$(qmake -query QT_INSTALL_DATA)/mkspecs/features

For example,

sean:~>ls -l "$(qmake -query QT_INSTALL_DATA)/mkspecs/features"
total 4
-rw-r--r-- 1 root root 491 Jul 15  2014 qscintilla2.prf

sean:~>cd /usr/src/debian/qgroundcontrol
History changed from "/home/carlo" to "/"
Environment changed from "/" to "/usr/src/debian/qgroundcontrol"
sean:/usr/src/debian/qgroundcontrol>which qmake
/usr/src/qt-5/5.5/gcc_64/bin/qmake
sean:/usr/src/debian/qgroundcontrol>
  ls -l "$(qmake -query QT_INSTALL_DATA)/mkspecs/features"
total 420
drwxrwxr-x 2 carlo carlo  4096 Aug  2 22:24 android/
-rw-rw-r-- 1 carlo carlo    23 Oct 13  2015 build_pass.prf
-rw-rw-r-- 1 carlo carlo   362 Oct 13  2015 c++11.prf
... etc

### DIFF
--- a/scintilla.pri
+++ b/scintilla.pri
@@ -18,6 +18,16 @@ scintilla {
     }
   }
 
+  # The above looks in too specific paths, use QT_* variables to find where things really should be.
+  isEmpty(QSCILOADED) {
+      exists($$[QT_INSTALL_DATA]/mkspecs/features/qscintilla2.prf) {
+      include($$[QT_INSTALL_DATA]/mkspecs/features/qscintilla2.prf)
+      INCLUDEPATH = $$[QT_INSTALL_HEADERS] $$INCLUDEPATH
+      LIBS = -L$$[QT_INSTALL_LIBS] $$LIBS
+      QSCILOADED=yes
+    }
+  }
+
   # The qscintilla2.prf which ships with QScintilla is broken for Mac/Windows
   # debug builds, so we supply our own
   isEmpty(QSCILOADED) {


t-paul:
For which Ubuntu version is that needed? Our own qscintilla2.prf is just a fallback for older, broken packaging. If correctly installed, the prf file should be picked up by qmake without any additional setup. This is fixed in Debian for some time now, so I'd expect it to be in Ubuntu now too.

CarloWood:
On Sat, 20 Aug 2016 09:33:29 -0700
Torsten Paul <notifications@github.com> wrote:

> For which Ubuntu version is that needed? Our own qscintilla2.prf is
> just a fallback for older, broken packaging. If correctly installed,
> the prf file should be picked up by qmake without any additional
> setup. This is fixed in Debian for some time now, so I'd expect it to
> be in Ubuntu now too.
> 

>dpkg -L libqt5scintilla2-dev | grep prf
ii  libqt5scintilla2-dev           2.8.1-2ubuntu1       all                  Scintilla source code editing widget for Qt5, development files

>dpkg -L libqt5scintilla2-dev | grep prf
/usr/share/qt5/mkspecs/features/qscintilla2.prf

Was this path changed in later versions of this package?

Nevertheless, prf files MUST be installed into $$[QT_INSTALL_DATA]/mkspecs/features
so my solution seem more clean as it picks up the right qscintilla2.prf
independent of which qmake (install) you're currently using in your PATH.

I often use some development Qt that is installed in a non-default place,
which really should work (the configuration and compilation, not running the
result of course) as long as you use the corresponding qmake executable.

Using $$[QT_INSTALL_DATA] is therefore more correct than looking
in $someprefix/lib/x86_64-linux-gnu/qt5/mkspecs/features, and it should be
possible, even, to remove the block above the block of code that I added
(I left it in in order to be sure not to break existing trees
that depend on the use of OPENSCAD_LIBRARIES).

Or am I not understanding something?

Regards,
--- 
Carlo Wood <carlo@alinoe.com>

PS I'm running trusty, but that doesn't really seem relevant
   for the argument.

>apt-cache policy libqt5scintilla2-dev
libqt5scintilla2-dev:
  Installed: 2.8.1-2ubuntu1
  Candidate: 2.8.1-2ubuntu1
  Version table:
 *** 2.8.1-2ubuntu1 0
        500 http://nl.archive.ubuntu.com/ubuntu/ trusty-updates/main amd64 Packages
        100 /var/lib/dpkg/status
     2.8.1-2 0
        500 http://nl.archive.ubuntu.com/ubuntu/ trusty/main amd64 Packages

t-paul:
Yes, it's relevant, I've reported the bug against 2.8.1-4 in Debian. Trusty seems to have a still broken package.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=751531

kintel:
Is there anything actionable here?

t-paul:
I might be missing something, but it seems the patch would work around a (Debian/Ubuntu) packaging error which was fixed more than 2 years ago. We already have a workaround, which is probably not perfect as it's used for multiple situations, but I'm still not in favor of adding even another workaround if the only affected distribution is Trusty.

CarloWood:
I don't think it's a work around; it's just the Right Way to find and include that file. The commit only adds something, only does something when it wasn't found before, but that was in order to make it easier to accept this patch: this can impossibly break old behavior because it only does something when the current method fails. However, if it was up to me I'd remove everything above it and just use this one method to find the correct qscintilla2.prf

probonopd:
Just to let you know that the problem persists on Ubuntu 14.04 trusty.

t-paul:
As the link above shows, the bug was reported a bit after release of Trusty, so nobody will fix that. The only chance would have been some backport of the fixed Debian package, but now it's probably also too late for that.

probonopd:
Please see https://github.com/openscad/openscad/issues/1849#issuecomment-321176869. Since Travis CI is using trusty by default it would be pretty important to have at least a proven workaround.

t-paul:
Maybe creating a symlink via `.travis.yml` would be enough?

probonopd:
Possibly, but from where to where?

t-paul:
Looking at my (now 3 years old... oh my...) bugreport

> The features file is installed in /usr/share/qt5/mkspecs/features/qscintilla2.prf
> but qmake seems to only read from /usr/lib/x86_64-linux-gnu/qt5/mkspecs/features.
> This causes qmake not to see the file with just "qmake CONFIG+=qscintilla2"

I think that would be
`sudo ln -s /usr/share/qt5/mkspecs/features/qscintilla2.prf /usr/lib/x86_64-linux-gnu/qt5/mkspecs/features/qscintilla2.prf`

As there was also an issue with the actual header file location, there might be some more symlinks needed.

probonopd:
Right, especially 

```
In file included from src/scintillaeditor.cpp:5:0:
src/scintillaeditor.h:8:32: fatal error: Qsci/qsciscintilla.h: No such file or directory
 #include <Qsci/qsciscintilla.h>
                                ^
compilation terminated.
make: *** [objects/scintillaeditor.o] Error 1
make: *** Waiting for unfinished jobs....
In file included from src/scadlexer.cpp:3:0:
src/scadlexer.h:5:29: fatal error: Qsci/qsciglobal.h: No such file or directory
 #include <Qsci/qsciglobal.h>
                             ^
compilation terminated.
make: *** [objects/scadlexer.o] Error 1
```

t-paul:
Found the part about the header files... I think the following should fix the includes:
`sudo ln -s /usr/include/qt5/Qsci /usr/include/x86_64-linux-gnu/qt5/`

PR #1779, opened 2016-08-20
Fix issue801. Made Undo button disabled when nothing to undo

onewhl:
How it looks:
mainwindow

I took into account a comment from last pull request and fixed it.


kintel:
Thx - works now.

Btw., for the future: There is no need to create a new pull request when you fix problems in an existing pull request. If you just push to the same branch, the pull request will update itself automatically

PR #1789, opened 2016-09-13
Polish translation

MarPiRK:
There seem to be a lot of strings without l10n activated. Haven't fixed that, because I didn't want to mix it with Polish translation, but if you want to I can do it in separate PR.

Also I updated pot file, because there were some strings not included in it.

### DIFF
--- a/locale/LINGUAS
+++ b/locale/LINGUAS
@@ -1,2 +1,2 @@
 # available languages
-fr ru de cs es
+fr ru de cs es pl
--- a/locale/pl.po
+++ b/locale/pl.po
@@ -1,37 +1,27 @@
-# #-#-#-#-#  openscad-tmp.pot (OpenSCAD 2016.09.13)  #-#-#-#-#
 # SOME DESCRIPTIVE TITLE.
 # Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
-# This file is distributed under the same license as the OpenSCAD package.
+# This file is distributed under the same license as the PACKAGE package.
 # FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
 #
-#, fuzzy
 msgid ""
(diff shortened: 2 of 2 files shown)


kintel:
@MarPiRK We might have forgotten to tag some strings. If you're willing to hunt, a pull request would be cool.

kintel:
thx!

t-paul:
Note that strings used only in console mode are currently not marked (e.g. the command line help output) as the localization is only tested in GUI so far and might need some fixes to work early in command line calls (also I think it's not fully decided if we even want to translate those). Otherwise what @kintel said... it's always a possibility that unmarked strings sneak in, so pointing those out helps.

MarPiRK:
You mean cli console, not render console?

t-paul:
Yes, CLI, the localization is initialized too late for quite a number of messages. I'm not sure, but I think we also skipped most of the error messages ending up in the console window as translated error messages can make troubleshooting more difficult (yeah, only a weak argument :-).

I guess the current state is like
* GUI - should be fully localizable (anything else is clearly a bug)
* GUI / Console window - only partially done, especially not sure about error messages
* CLI output - essentially not localized

MarPiRK:
Hmm, I see your point about troubleshooting. I think though that it may be easier for user to find error in code by himself. Maybe it could be good compromise to make console error translated by default, so it would be easier for user to understand them and add little checkbox which would change console language, so that user can post it somewhere seeking for help.

t-paul:
It did not mean to say "we must not translate the error messages", it's just the current state we decided quite some time ago when we added support for localization as it made things a bit easier at the time. The (small) problems caused by translated error messages could be solved also by other means, like error codes or something similar.

MarPiRK:
Yeah, I just thought about error codes. Do you suggest anything specific?

PR #1790, opened 2016-09-14
dependency helper scripted adapted for KDE Neon distribution

yeus:
install dependencies on KDE Neon (current version).


t-paul:
Thanks! What's the name schema for that distribution? Would it make sense to maybe just check for `KDE neon 5.` to cover all 5.x releases?

donbright:
im testing kdeneon in a vm, needed this to test. i can add the sub-version thing

PR #1794, opened 2016-09-14
Fix Russian translation

SlesarevIlya:
File with a Russian translation has been corrected in accordance with the rules of the Russian language.


kintel:
thx!

Issue #1064, opened 2014-12-11
GUI Splash Screen - Usability?


MichaelAtOz:
A few times now I have miss-used the Open button on the new splash screen when wanting to opening recent files. 
I click on the file on top of the recent list, then click the Open button, because visually it is right next to the file. It could perhaps happen with examples too.

Do we really need the separate Open Recent and Open Examples buttons? Just, "IF (file or example has highlight) open highlighted item ELSE display File-Open Dialogue".

Obviously I will get used to it, but for new users that may be better, KISS.
##

jordanbrown0:
> it's not like you can highlight a bunch of files and then open them all with the open buttons so why should there be a separate open action.

There's a decent argument that you should be able to do exactly that.

PR #1795, opened 2016-09-16
Proposal for a decision of issue #1064

SlesarevIlya:
Hello, I made a design option for launchscreen, adding a checkbox with a choice of startup options.
Please view and comment it.

Accordingly, in an embodiment of the buttons it is possible to open the double-click the file.
openscad1
In an embodiment without buttons the file is opened with a single click.
openscad2

### DIFF
--- a/locale/ru.po
+++ b/locale/ru.po
@@ -967,7 +967,7 @@ msgstr "Включить перетаскивание Редактора и Ко
 
 #: objects/ui_Preferences.h:1161
 msgid "Show Welcome Screen"
-msgstr "Показать экран преветствия"
+msgstr "Показать экран приветствия"
 
 #: objects/ui_Preferences.h:1162
 msgid "Enable user interface localization (requires restart of OpenSCAD)"
@@ -1031,7 +1031,7 @@ msgid ""
 "Warning: You may experience OpenCSG rendering errors.\n"
 "\n"
 msgstr ""
-"Предупреждение: Возможны ошибки рендринга OpenCSG.\n"
+"Предупреждение: Возможны ошибки рендеринга OpenCSG.\n"
 "\n"
 
 #: src/QGLView.cc:132
--- a/src/Preferences.cc
+++ b/src/Preferences.cc
@@ -166,6 +166,7 @@ void Preferences::init() {
 	this->defaultmap["advanced/reorderWindows"] = true;
 	this->defaultmap["launcher/showOnStartup"] = true;
 	this->defaultmap["advanced/localization"] = true;
+    this->defaultmap["launcher/showOnMethodOpen"] = true;
 
 	// Toolbar
 	QActionGroup *group = new QActionGroup(this);
@@ -477,6 +478,12 @@ void Preferences::on_launcherBox_toggled(bool state)
 {
 	QSettings settings;
  	settings.setValue("launcher/showOnStartup", state);	
+}
+
+void Preferences::on_launcherBox2_toggled(bool state)
+{
+    QSettings settings;
+    settings.setValue("launcher/showOnMethodOpen", state);
 }
 
 void Preferences::on_checkBoxShowWarningsIn3dView_toggled(bool val)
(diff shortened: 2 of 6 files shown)


kintel:
Thanks for your pull request.

I think that for this issue, we're looking to make the UI more user friendly. I think that adding more checkboxes is a bit of a step in the wrong direction.
Good UI design isn't trivial, so such things tends to take a bit of time to figure out properly..

Pointedstick:
Indeed. More checkboxes is generally the wrong answer.

SlesarevIlya:
Okay, i understand you, thanks.

donbright:
you know, i'm not sure i really understand this. 

is there more description of what "one click the file" means? thanks

MichaelAtOz:
> is there more description of what "one click the file" means? thanks

Single click on a leaf of the recent file list or the example tree, opens that file, v's double clicking it as now.

(I'm ambivalent on the subject)

t-paul:
It's about issue #1064. I don't have any problem with the current layout, but also don't see an issue with changing it a bit.
Although what's been discussed so far (here and in #1064) does not feel like an improvement to me, just different.

kintel:
Closing this as this doesn't improve usability. I'm still interested in an improvement - make it cleaner, more intuitive, perhaps fewer buttons.

donbright:
oh my,  i missed the title. ok. thanks for explaining.

it seems like if you took this and got rid of the check box, it would basically be like @atartanian 's suggestion on #1064 which seems to make sense to me.

atartanian:
+1 for just making the checked behavior the only way it works. there is just no need for those "open" buttons, the filenames can provide direct access, which is more intuitive and reduces the complexity of the UI

t-paul:
I'm against using a way to select the files that's going against the normal UI behavior, e.g. using single clicks where the default is double click.

MichaelAtOz:
> I'm against using a way to select the files that's going against the normal UI behavior, e.g. using single clicks where the default is double click.

+1. That is web browsing UI, not application UI.

I'm not suggesting this, but to get consistent UI with a single click would need to change the file/example names to links like those below, with the underline when hovered behaviour (example from Control-panel/Power-options);
link ui

t-paul:
That type of links is fine in general, we have it already for the warnings. And we hopefully get it sometimes in the log window pointing to errors.

In case of list boxes or the tree for the examples, that's not easy to solve though. I don't see how to convert this to web style without losing the start dialog feel. It would make sense if we had some "start" document like some IDEs do, but right now that's not really possible with the single file text editor.

donbright:
Even if this particular issue (splash screen) is kept as "desktop ui" style for now, eventually in the future some kid will change it to web style. like how mac OSX changed the way mouse scrolls because they know in the future everything will be touchscreen.... so too in the future everything will be fingers single-tapping on icons.  

Web is future of ui. qt is moving to qml which is basically a flavor of javascript. all the kids are growing up are learning to code ui in javascript. touching with fingers, swiping, and all that. 

OpenSCAD's own file/examples/ menu is one click, ubuntu dash is one click, ubuntu side bar is one click, Chrome startup "dialer" is one click. Android is one click, iOS is one click. Desktop PC market is dying, phone market is rising.  most common "open file" operation people do these days is people opening a podcast episode or a netflix video or a tweet or an instagram post or a youtube video. They are opening a "document", it just so happens to be an audio or video document. All of this is single click. We are the last desktop generation. When we fade away, desktop ui will go with us. We will keep the flame a live for a while but the new generations have their own ideas. 

https://blog.codinghorror.com/double-click-must-die/

MichaelAtOz:
Don, where are you making these comments? It doesn't show under the 'view it on GitHub' link.

 

  _____

MichaelAtOz:
I think the point should be that things like Qt & operating system APIs should translate program calls to match target UI. e.g. a file/open dialog, and make the UI suit the locale. The worst outcome is an application which does not match the behaviour of the platform it is running on.

Issue #1798, opened 2016-09-18
Can't generate PNG file without an X server


l0b0:
From an [automated build](https://travis-ci.org/l0b0/graphics/builds/160819837#L230-L235):

```
/usr/bin/openscad --projection=ortho --camera=0,0,0.5,90,0,0,3 -o chisel/equilateral_plug_example.png chisel/equilateral_plug_example.scad
Compiling design (CSG Products normalization)...
Normalized CSG tree has 2 elements
Compiling background (1 CSG Trees)...
Unable to open a connection to the X server
Can't create OpenGL OffscreenView. Code: -1.
```

I tried the same in a text mode console, and got the same result. This creates a zero-length file and returns with a *zero exit code.*

```
$ openscad --info
OpenSCAD Version: 2015.03-3
System information: Linux 4.7.3-2-ARCH #1 SMP PREEMPT Thu Sep 8 09:44:02 CEST 2016 x86_64 Arch Linux 4 CPUs 3.78 GB RAM
Compiler: GCC "6.1.1 20160802" 64bit
Boost version: 1_61
Eigen version: 3.2.9
CGAL version, kernels: 4.8.1, Cartesian<Gmpq>, Extended_cartesian<Gmpq>, Epeck
OpenCSG version: OpenCSG 1.4.0
Qt version: 4.8.7
QScintilla version: 2.9.2
MingW build: No
GLib version: 2.48.2
Application Path: /usr/bin
Documents Path: /home/username/.local/share
Resource Path: /usr/share/openscad
User Library Path: /home/username/.local/share/OpenSCAD/libraries
User Config Path: /home/username/.config/OpenSCAD
Backup Path: /home/username/.local/share/OpenSCAD/backups
OPENSCADPATH: <not set>
OpenSCAD library path:
  /home/username/.local/share/OpenSCAD/libraries
  /usr/share/openscad/libraries

OPENSCAD_FONT_PATH: <not set>
OpenSCAD font path:


GLEW version: 2.0.0
OpenGL Version: 3.0 Mesa 12.0.3
GL Renderer: Mesa DRI Intel(R) Sandybridge Mobile 
GL Vendor: Intel Open Source Technology Center
RGBA(8888), depth(24), stencil(8)
GL_ARB_framebuffer_object: yes
GL_EXT_framebuffer_object: yes
GL_EXT_packed_depth_stencil: yes
GL context creator: GLX
PNG generator: lodepng
GLX version: 1.4
[...]

PR #1799, opened 2016-09-18
Indicate failure to create offscreen buffer for PNG export with non-zero exit status.

t-paul:
Mentioned in #1798.


donbright:
tested good on ubuntu 16

Issue #1331, opened 2015-05-01
Add support for import / export 3MF format


clothbot:
Creating this issue to gather info and interest around the new 3MF file format:

* [What is 3MF](http://www.3mf.io/what-is-3mf/)
* [3MF Specification](http://www.3mf.io/what-is-3mf/3mf-specification/)

Cautiously optimistic that the standard will be as open as claimed...

t-paul:
Yeah, https://xkcd.com/927/ :-). But then it's not really new, it's just the already existing Microsoft format. I guess it remains to be seen what "The source code will be on Github and cross-platform code is in development." means. At least the licensing requirements for the source code sound reasonable and Microsoft does not try to destroy evil open source anymore (not so sure about some other consortium members) ;-).

clothbot:
Straw(man) poll over on [the mailing list](http://forum.openscad.org/New-3MF-file-format-td12525.html) is leaning toward wait'n'see.  (to put it politely... ;-)

clothbot:
As per the discussion on the mailing list, closing this ticket for the time being.

Near-future development activity needs to focus on multi-material and colour/texture support in the already implemented, basic AMF format support.

t-paul:
This may finally happen... #1802 :-).

PR #1802, opened 2016-09-20
3mf Import/Export

t-paul:
This adds import and export for the [3MF file format](http://www.3mf.io/) using [lib3mf](https://github.com/3MFConsortium/lib3mf).

See also #1331.

ToDo:
- [x] Build on Travis Ubuntu 12.04 and 14.04
- [x] Build on Travis OSX / homebrew (using openscad/homebrew-tap repo)
- [x] Mac release build (macosx_build_dependencies.sh)
- [x] Update MXE build (currently stuck, it builds on MXE but using the library crashes)
- [x] Add patch to MXE build ([MinGW issue](https://sourceforge.net/p/mingw-w64/bugs/565/) - wcstof crashes with endPtr == nullptr)
- [x] The build system should fail (or at least warn) if lib3mf wasn't found at qmake time.
- [x] Add "export-import" tests (see #1822), similar to how we handle other bidirectional formats, e.g. `add_cmdline_test(offpngtest EXE ${PYTHON_EXECUTABLE} SCRIPT ${CMAKE_SOURCE_DIR}/export_import_pngtest.py ARGS --openscad=${OPENSCAD_BINPATH} --format=OFF --render EXPECTEDDIR monotonepngtest SUFFIX png FILES ${EXPORT3D_TEST_FILES})`
- [x] Update macOS build script
- [x] Look into macOS install_name


kintel:
The lib3mf build should work for Mac now. I wouldn't call what they've got a "build system", but it will work as long as they don't change it :/

t-paul:
Yeah, it's just some generated file from an IDE or so. At least they accepted the minimal wrapper Makefile to allow installation (https://github.com/3MFConsortium/lib3mf/pull/9) so building on OBS, MXE and homebrew is ok-ish.

t-paul:
Ok, Windows/MinGW issues are now fixed. The `wcstof()` bug in MinGW is fixed, but until this is going to find it's way into MXE, I've added a workaround to the `lib3mf` MXE build.
Export file creation needed a special case, to enforce binary stream format, otherwise `lib3mf` failed due to `CR` -> `CRLF` conversion.

Not sure yet, why the `precise` Travis build now fails to install dependencies.

t-paul:
For some reason the `sudo` calls in the external `travis-before-install.sh` script did work for `trusty` but not in `precise`. I've changed config to call the whole script via `sudo`, which works for both.

I'm going to rebase this to get rid of the Travis testing stuff, so please update the branch if it's checked out somewhere...

t-paul:
I'll have a look if it's possible to give some kind of final status output listing what optional features are enabled.

t-paul:
Something like https://github.com/openscad/openscad/blob/inputdriver3/info.pri from the `inputdriver3` branch.

Used simply by adding `CONFIG+=info`

kintel:
hm, I guess the challenge is that we're using qmake both to build dependable binaries as to build development binaries with potentially partial configs.

I'd say that we should make sure that the "dependable binaries" part is solid first.
..then make it more convenient to perform partial builds.

I guess there might be different requirements for various distro builds?

rogersachan:
@t-paul Any update? Can you build a Win32 binary of your PR so I can test?

scripts/macosx-build-dependencies.sh
```diff
@@ -53,6 +53,7 @@ PACKAGES=(
     "libxml2 2.9.7"
     "fontconfig 2.12.4"
     "hidapi 0.7.0"
+    "lib3mf a466df47231c02298dde295adb6075b0fc649eba"
```
kintel: I see there's a bunch of recent activity on lib3mf - should we bump this to a more recent version?

features/lib3mf.prf
```diff
@@ -0,0 +1,49 @@
...
+# 3. system's standard include paths
+
+OPENSCAD_LIBRARIES_DIR = $$(OPENSCAD_LIBRARIES)
+
+!isEmpty(LIB3MF_INCLUDEPATH) {
+  LIB3MF_CFLAGS = -I$$LIB3MF_INCLUDEPATH -I$$LIB3MF_INCLUDEPATH/Model/COM
+} else {
+  !isEmpty(OPENSCAD_LIBRARIES_DIR) {
+    exists($$OPENSCAD_LIBRARIES_DIR/include/lib3mf/Model/COM/NMR_DLLInterfaces.h) {
+      LIB3MF_CFLAGS = -I$$OPENSCAD_LIBRARIES_DIR/include/lib3mf -I$$OPENSCAD_LIBRARIES_DIR/include/lib3mf/Model/COM
```
kintel: It seems a bit odd to add some subpath under 3mf as our include path. Couldn't we just `#include <lib3mf/Model/COM/NMR_DLLInterfaces.h>` in our code?

scripts/macosx-build-dependencies.sh
```diff
@@ -766,6 +767,25 @@ build_hidapi()
...
+  if [ ! -f $version.tar.gz ]; then
+    curl -LO https://github.com/3MFConsortium/lib3mf/archive/$version.tar.gz
+  fi
+  tar xzf $version.tar.gz
+  cd lib3mf-$version
+  CXXFLAGS="$CXXSTDFLAGS" make CC=cc CXX=c++ LD=c++ LDFLAGS="$LDSTDFLAGS" -j"$NUMCPU" -C Project/Lib3MFGCC
+  make -j"$NUMCPU" -C Project/Lib3MFGCC install INSTALL_ROOT=$DEPLOYDIR
+#  install_name_tool -id @rpath/libCGAL.dylib $DEPLOYDIR/lib/libCGAL.dylib
+#  install_name_tool -id @rpath/libCGAL_Core.dylib $DEPLOYDIR/lib/libCGAL_Core.dylib
+#  install_name_tool -change libCGAL.11.dylib @rpath/libCGAL.dylib $DEPLOYDIR/lib/libCGAL_Core.dylib
```
kintel: Note to self: Have a look at this

features/lib3mf.prf
t-paul: They had lots of internal references that need the multiple `-I` flags. Maybe that is fixed now and just the main `.../lib3mf` is enough.

t-paul: Yep, that's fine now, I changed our code too, so the Model/COM deep link is not needed anymore.

kintel:
@t-paul Looks like Info.plist.in is used by the cmake build system

t-paul:
Then we always edited the wrong one `i.e. not the Info.plist.in` in the past?

kintel:
I guess nobody builds binaries with cmake yet, so that file may be lagging behind.

PR #1804, opened 2016-09-23
File names changed to QScintilla_gpl-2.9.2.tar.gz

clothbot:
Looks like the project decided to change QScintilla-gpl-* to QScintilla_gpl-* in version 2.9.2 and 2.9.3.


kintel:
thx!

PR #1805, opened 2016-09-25
Green Refactoring: Rewrote Annotation to work directly on an Expression object

kintel:
@amarjeetkapoor1 @t-paul 
Could you take a look at this patch?

During code review, I struggled with understanding the Annotation
class, but I eventually realized that it's just storing an Expression
object, but has a pretty convoluted way of retrieving it. To make this
simpler and more readable, I've refactored the Annotation class down
to the bare minimum.

All tests should still be green.

I might have misunderstood something, so please comment if there is something I've left out.
@amarjeetkapoor1 If this looks OK, could just merge it into the gsoc2016-refactored branch.

### DIFF
--- a/src/Assignment.h
+++ b/src/Assignment.h
@@ -41,25 +41,15 @@ typedef std::vector<Assignment> AssignmentList;
        
 class Annotation
 {
-protected:
-    Annotation(const std::string name, const AssignmentList assignments, const AssignmentList args);
-    
 public:
+    Annotation(const std::string &name, shared_ptr<class Expression> expr);
     virtual ~Annotation();
 
     std::string dump() const;
-    const std::string & get_name();
-    virtual ValuePtr evaluate(class Context *ctx, const std::string& var) const;
-    
-    Annotation& operator=(const Annotation&);
+    const std::string &getName() const;
+    virtual ValuePtr evaluate(class Context *ctx) const;
 
-    static const Annotation * create(const std::string name, const AssignmentList assignments);
-
-//private:
-    std::string name;
-    /** Defines the names and values parsed from the script */
-    AssignmentList assignments;
 private:
-    /** Defines the names of the positional parameters */
-    AssignmentList args;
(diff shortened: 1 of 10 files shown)


t-paul:
The extended functionality remains from the initial prototype which had a much wider scope (e.g. cover various different annotations like https://github.com/openscad/openscad/wiki/Meta-Data-Use-Cases).

If you want to go for the simpler solution, I guess that's fine. It seems the native syntax discussion might still need some time, so we probably can extend later again.

kintel:
That would explain some things :)
Since there are no other traces of the old prototype in this code, I _think_ it makes sense to keep the slim version for now.

rimmartin:
Any chance annotations will make it to the master branch?

I was looking to add annotations for cam instructions

kintel:
This has been in master for a long time, but perhaps it doesn't do what you think it does. This is only related to the Customizer.

rimmartin:
hmmm, testing with version 2017.06.03.nightly. Guess I was thinking more in terms of the link above to
https://github.com/openscad/openscad/wiki/Meta-Data-Use-Cases

```
@Description(text = "Primary with bearing race")
module primaryDisk() {
}
```
In my case if the parser would at least skip over then I was experimenting with instead of drawing threads, adding annotations for cam. For example

```
@CAMTable(tool-number=5)
cylinder(h=bearingH-bearingH1+20, r1=bearingN2/2, r2=bearingN2/2, $fa=1, center=false);
```

This way instead of graphically displaying threads it would be the cylinder hole before tapping and the annotation would provide choice of tapping tool http://linuxcnc.org/docs/2.7/html/gcode/g-code.html#gcode:g10-l1 from a cnc enabled machine configuration of cutters and taps.

Just exploring the possibility.

Also I've been capturing the scad language as a domain specific language by developing an antlr v4 grammar https://github.com/antlr/antlr4 for scad

kintel:
The extended case hasn't been implemented. We slimmed down the scope of the customizer to make it realistic to develop and merge during a summer.

I'd be interested in more info about your antlr grammar. Perhaps the mailing list would be a good place to chat about that. ..or IRC.

rimmartin:
@kintel ok, good.  I'll run the antlr v4 grammar for scad through the examples and continue getting it right. I'll join the mailing list and by next weekend make the grammar in it's current form available.

Issue #1765, opened 2016-08-04
echo() escaped character & HTML conflict


MichaelAtOz:
1000 words;
![escape and html](https://cloud.githubusercontent.com/assets/1972961/17421241/acc1b756-5aeb-11e6-84a7-c353195ca861.JPG)
Some `escaped characters` (newline in this case, not tab) prevent `HTML` from being displayed.
Conversely, lines 11 &12, `HTML` prevents `escaped tab` from being displayed.

I haven't tried other escape sequences.

In OpenSCAD 2015.3-2 and 2016.07.20 on Win7.

t-paul:
Someone on IRC asked about this issue, so I guess it would make sense to also define what the expectation is. I never really expected the console to render HTML. It might be useful in some cases but is both risky in the long run (do we maybe want to support loading scripts directly from the web?) and sort-of conflicts with command line usage where HTML is not directly usable.

As mid-term solution, maybe split console (= for messages generated by the app) and output (= messages generated by the scripts)?

doug-moen:
My expectation is that I can use echo for debugging, and for printing the values of variables, including string variables. This means that echo should print strings literally, and not interpret them as HTML. The current behaviour should be considered a bug.

If we want to support HTML output to the console, then we should create a new function for that. The small number of people who want this will also want the ability to print to the console without having "ECHO:" preprended to the output.

MichaelAtOz:
The current behavior (baring bug in this report) is not a bug, and is useful. It is built into Qt, so eliminating it would presumable be more work.
During debug it is very useful to format output.
I seem to recall someone having quite extensive use of html, I think it was runsun  [e.g. BOM - forum](http://forum.openscad.org/OpenSCAD-and-Woodworking-td13966.html)

MichaelAtOz:
> but is both risky in the long run (do we maybe want to support loading scripts directly from the web?) and sort-of conflicts with command line usage where HTML is not directly usable.

It is a [subset](http://doc.qt.io/qt-5/richtext-html-subset.html) of html

bulgakovk:
I noticed that If I use only one tab symbol - HTML still renderes correct but tab replaces by empty symbol.
![2016-09-13 22-21-41](https://cloud.githubusercontent.com/assets/8983736/18488108/b3b307d4-7a00-11e6-8d0d-c5f7bb598d06.png)

HTML with two tabs still works nice but all tab symbols replace by only one space.
[...]

PR #1812, opened 2016-09-30
Issue1765

bulgakovk:
Solution for #1765

### DIFF
--- a/src/control.cc
+++ b/src/control.cc
@@ -263,23 +263,23 @@ AbstractNode *ControlModule::instantiate(const Context* /*ctx*/, const ModuleIns
 		node = new GroupNode(inst);
 		std::stringstream msg;
 		msg << "ECHO: ";
+		static boost::regex tagRegEx(".*<[^>]+>.*");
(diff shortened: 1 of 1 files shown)


t-paul:
I'm not sure this is a good idea, it will break other cases as there's no safe way to distinguish between HTML and normal text. Right now that's just happens by random interpretation in the console text widget (which might even depend on the Qt version used), which is indeed not good, but trying to fight against that is probably not going to work in all cases.

Examples:
```
echo("text: <hallo>\n");
ECHO: "text: <hallo><br/>"
```
```
echo("text: <b>\n");
ECHO: "text: 
"
More output is now rendered in bold (which is an additional different issue)
```

kintel:
If we want to do any Qt-specific formatting, we should probably do that in the component which handles Qt output: https://github.com/openscad/openscad/blob/master/src/mainwin.cc#L2760

In terms of using special formatting for logging, perhaps look at how people solve the mess of using escape codes in terminal output vs. log files.

We need a better solution

Issue #1800, opened 2016-09-18
Windows 10 is not reported correctly in library info


t-paul:
It seems the version detection is broken with Win10 again (reported in #1409).

> I'm running Win10 anniversary edition, which is incorrectly reported in the lib info window as WIn 8:
> System information: Windows 8 (v6.2.0.0) 4 CPUs 5.93 GB RAM

We used https://bitbucket.org/AnyCPU/findversion to detect the Win8 versions that are not reported by the old (and now deprecated) interface but this fails on Win10.

donbright:
sorry about that, fixed now

kintel:
Fixed by #1817

PR #1817, opened 2016-10-05
Wininfo fix 2016 for issue #1800

donbright:
quick fix for issue #1800

also simplify code for vast majority of machines and get rid of extra stuff for old machines (the exact version of windows 2002 server, or whatever, can be found by looking up referenced MSDN 724833 article instead of us keeping 50 lines of code for machines that are running OSes that are beyond their official end of life)

https://msdn.microsoft.com/en-us/library/windows/desktop/ms724833(v=vs.85).aspx


kintel:
It looks like ```scripts/winmanifest.xml``` is missing..

donbright:
sorry about that, fixed now (travis cl is some mac thing that is unrelated to this)

PR #1830, opened 2016-10-27
Add echo() expression.

t-paul:
Based on discussion in #1587 

```
function f1(x) = [ for (a = x) echo() a ];
echo("f1(v) = ", f1(v));

ECHO: 2
ECHO: 5
ECHO: 7
ECHO: "f1(v) = ", [2, 5, 7]

function f2(x, i = 0) = echo(i) len(x) > i ? x[i] + f2(x, i + 1) : 0;
echo("f2(v) = ", f2(v));

ECHO: 3, 0
ECHO: 2, 7
ECHO: 1, 12
ECHO: 0, 14
ECHO: "f2(v) = ", 14
```


nophead:
Seems odd to me that echo print the expression that comes after it, rather than just what is inside its parenthesis. Very confusing with the recursion as it shows i counting down, but it actually counts up. Presumably this is because it doesn't know the value of the expression until it has fully recursed.

MichaelAtOz:
> Presumably this is because it doesn't know the value of the expression until it has fully recursed.

Yes, so it won't be able to debug infinite recursion.
It could be confusing, but perhaps, if there parameters, they get output immediately, and if there are no parameters then the trailing expression is output.

```
a=3; b=5;
x = echo(a=a) a*b; // echo: a=3
y= echo() a*b; // echo: 15
```

kintel:
@doug-moen Any opinions on this?

nophead:
Yes that seems more logical.

MichaelAtOz:
That way you could do both down recursion, and up recursion ie with the results;

```
function f2(x, i = 0) = echo() len(x) > i ? echo(str("x[",i,"]=",x[i])) x[i] + f2(x, i + 1) : 0;
echo("f2(v) = ", f2(v));
```

t-paul:
Yes, that feature was mainly the reason for printing the expression value. Assuming we drop that and always print parameters before evaluating the expression, the simplest way to still print the evaluation result I could think of so far (untested, but should work I believe) would be something like
```
let(result = a * b) echo(result) result
```
Anyone seeing an easier solution?

I don't really like having subtle differences in behavior depending on the number of parameters, trying to explain that clearly in the documentation seems difficult.

Also, we _could_ print values when unwinding from a stack overflow, but that would still produce the reverse order.

Anyway, so if we have a way to print the actual expression value, should we go for what @nophead suggested in the first comment an restrict to _only_ print parameters _before_ evaluation and just return the value of the expression?

nophead:
You could have a bazaar syntax line echo(=) to echo the following expression, but I am not sure I like it.

E.g.
function test(a,b) = echo(a = a, b = b,  "result:" =) a * b;

t-paul:
So that would work too (when just printing parameters before evaluation):
```
v=[4,7,9];
function result(x) = echo(result = x) x;
function f2(x, i = 0) = result(len(x) > i ? echo(str("x[",i,"]=",x[i])) x[i] + f2(x, i + 1) : 0);
echo("f2(v) = ", f2(v));

ECHO: "x[0]=4"
ECHO: "x[1]=7"
ECHO: "x[2]=9"
ECHO: result = 0
ECHO: result = 9
ECHO: result = 16
ECHO: result = 20
ECHO: "f2(v) = ", 20
```

nophead:
That looks like a neater solution.

doug-moen:
+1 to the solution that Torsten just pushed.

I want `echo(args)body` to print args *before* the body is evaluated. For example, I have a function that infinite loops or crashes on some particular argument, I want to print the argument *before* this happens.

I don't want `echo` to print the result of evaluating the body, because that's usually not what I need, and also that forces printing to not occur until after the body is evaluated, which is undesirable. I don't feel the need for special magic syntax for this either, I'm perfectly happy to request this explicitly by writing `let(x=body)echo(args)x` if the need ever came up, as Torsten previously suggested.

kintel:
I agree to the current state with one caveat:
If you want to print the result of a (potentially recursive) expression, you have to either evaluate twice:
```function f(x) = echo(g(x)) g(x);```

..or rewrite using let():
```function f(x) = let(tmp=g(x)) echo(tmp) tmp```

My current suggestion would be:
```echo([expr, ...]) [expr]``` - prints its arguments and returns its chained expression. If no chained expression is given, returns its last passed argument.

..so the above two examples would be reduced to:
```function f(x) = echo(g(x));```

It is, of course possible to add this helper in user-space:
```
function echoAndReturn(expr) = let(tmp=expr) echo(expr) expr;
function f(x) = echoAndReturn(g(x));```
```

Note: What I suggest is a pretty clean extension to the current state, so it's not necessary to solve it before merging this.

**Comment by doug-moen** [comment]
@kintel, Your proposed extension makes `echo(...)` into an expression, with a different interpretation than `echo(...)` in a statement context. That gets weird when we consider OpenSCAD2, where statements and expressions are unified, and `cube(10)` becomes an expression.

For example, what does this mean in OpenSCAD2: `echo(cube(10));`?
If we use the current statement interpretation, then a shape is echoed and nothing else happens. If we use the proposed new expression interpretation, then the cube value will be inserted into the model.

Since your proposal doesn't actually add any new power, I suggest we think about this more before acting.

**Comment by nophead** [comment]
> If you want to print the result of a (potentially recursive) expression, you have to either evaluate twice:
function f(x) = echo(g(x)) g(x);

>..or rewrite using let():
function f(x) = let(tmp=g(x)) echo(tmp) tmp

No you just need @t-paul's helper function:
```
function result(x) = echo(result = x) x;
```
Then you just wrap your result in result(). It gets evaluate once, passed to result() which prints it and returns it.

kintel:
@doug-moen I agree - this needs to make sense also on the statement level in the future. I withdraw my suggestion :)

Issue #1838, opened 2016-10-29
Build error on OS X/macOS >= 10.11 and/or Xcode 8


dmopalmer:
When running macosx-build-dependencies.sh with the -f flag, it fails to build qt with the problem described in [this discussion](http://stackoverflow.com/questions/39492617/xcode-8-error-project-error-xcode-not-set-up-properly-even-though-the-licens).

The symptom error message is `Xcode not set up properly. You may need to confirm the license agreement by running /usr/bin/xcodebuild.`

The fix is, apparently, in ../libraries/src/qt-everywhere-opensource-src-5.7/qtbase/mkspecs/features/mac/default_pre.prf

```
    # Make sure Xcode is set up properly
+    isEmpty($$list($$system("/usr/bin/xcrun -find xcodebuild 2>/dev/null"))): \
-    isEmpty($$list($$system("/usr/bin/xcrun -find xcrun 2>/dev/null"))): \
        error("Xcode not set up properly. You may need to confirm the license agreement by running /usr/bin/xcodebuild.")
```

(manually copied/pasted/modified: this is not an automatic diff suitable for a patchfile)

I haven't actually implemented and/or tested the fix because I don't know the openscad build system, where to put patches, etc.

kintel:
@dmopalmer is there an upstream bug report for this that we could track?

dmopalmer:
On the qt site:
https://bugreports.qt.io/browse/QTBUG-55649
https://bugreports.qt.io/browse/QTBUG-56004
note that it is fixed in 5.7.1

kintel:
thx - we could patch our build system in the interim.
If you want to have a crack at it, this is how we did it last time:
https://github.com/openscad/openscad/commit/e15c36f53fa2ecdbcef189c5e4a93d08cdb33fc3#diff-1b921c0a239f34c540d19ba7197c4e7b

dmopalmer:
I am so far from being able to successfully compile openscad that I can't expect to reach a good endpoint, therefore I would not be able to enter a validated pull request.

(Next problem, my MacPorts-installed qt is conflicting with the one installed by anaconda.)

kintel:
I don't expect there to be other problems than qt, so it shouldn't be that hard to get it to work. I use that script all the time, but I'm still trying to get an Xcode 8 system up and running..

dmopalmer:
I just submitted a pull request.  Two files had to be patched, vs the singe file of QTBUG-56004

[...]

PR #1847, opened 2016-10-30
Patch to allow building QScintilla with Xcode 8

kintel:
See #1838 

Fix from: Homebrew/homebrew-core#5877

upstream discussion: https://www.riverbankcomputing.com/pipermail/qscintilla/2016-October/thread.html#1168


dmopalmer:
And now it founders on
```
configure: error:
*** GLib requires a 64 bit type. You might want to consider
*** using the GNU C compiler.
```

kintel:
@dmopalmer Please discuss such things in #1838

PR #1880, opened 2016-12-03
Fix for angle argument in rotate_extrude

LemurPhil:
rotate_extrude is ignoring the angle argument


t-paul:
It's not ignored, that line defines the positional arguments (vs. the named arguments). I think it was intentionally not added there.

The change would allow `rotate_extrude("file", "layer", [0, 0], 20);` with 20 being the angle.

Due to the deprecated/legacy arguments that would need to be in front, it's probably not that useful to have as positional parameter and putting it first would break the (the deprecated) usage with the file name first.

LemurPhil:
Agreed, my apologies.

PR #1663, opened 2016-06-06
Rename to RELEASE_NOTES.md

jakemckown:
This file used Markdown syntax but did not have the `.md` extension. It previewed as plain text.

TODO: Rename RELEASE_NOTES references in scripts and docs
- [ ] hacking.txt
- [ ] scripts/makereleasenotes.sh

PR #1889, opened 2016-12-11
rename RELEASE_NOTES to RELEASE_NOTES.md to reflect markdown content

lostapathy:
This completes the work started on #1663

Renaming RELEASE_NOTES to RELEASE_NOTES.md so that github will show it formatted rather than as plain text.


lostapathy:
Apologies, I missed the status change on this.

I'm a little lost as to what's not right yet, but if you can provide some guidance I'm happy to help.  The only thing I could find tying RELEASE_NOTES to appcast.xml is scripts/makereleasenotes.sh and I already update that in the PR.  It seems to work fine from my testing.  Please advise.

kintel:
You're right, I didn't review your changes carefully enough.

PR #1894, opened 2016-12-21
Updating the years on the copyright notice and might as well get ahead of us

TomK32:
Just noticed the year was still 2015 in openscad.


kintel:
Thx - I guess no Christmas-time release this year is a good thing :)

PR #1912, opened 2017-01-08
Added keyword children

### LINKED ISSUE #1908: Keyword "children" is not highlighted when using QSintilla Editor
Suggest solution: add the keyword "children" to keywordSet[3] in scadlexer.cpp

redfast00:
Should fix #1908


t-paul:
Thanks, don't worry about the reported test failure. That's an independent issue when running on macOS.

Issue #1237, opened 2015-03-05
Automatic Reload: Slowdown with large # of dependencies


kintel:
From [mailing list](http://forum.openscad.org/2015-02-05-less-responsive-tt11580.html)

When Automatic Reload and Compile is turned on, we regularly scan for changes in dependencies. If the dependency tree is large, this causes a noticable slowdown - typically laggy rendering.
Some ideas:
- If files are included in multiple places, don't check them twice. This can be done by e.g. traversing the dependencies as a DAG before scanning for changes.
- Use platform tools for listening for filesystem changes rather than polling manually. See e.g. [QFileSystemWatcher](https://doc.qt.io/qt-5/qfilesystemwatcher.html)

Problem even more critical when the project is on a network share (#1297).
##

nophead:
If you currently do check files more than once and recursively then I am sure fixing that alone will be enough to fix the problem. It's akin to the problem with min and max being very slow in recursive functions. The number of files polled will explode exponentially in a project like Mendel90.

Smartlug:
Same issue and even blocking when too long... windows state "not responding" in the title, disable window and when re-enabled... reload again...

t-paul:
@Smartlug is that a public project somewhere which could be used to check improvements against?

nophead:
@t-paul,
   You can use Mendel90. If you open scad/main.scad it pans and scrolls reasonably quickly with auto reload off and becomes almost totally unresponsive with it on. Older versions don't slow down as much but also don't spot all the nested dependencies either. The initial F5 does take about 15 minutes though.

t-paul:
@nophead yeah, I know that's the case reported on the mailing list. Having some additional test cases is always good as there could be other cases than the suspected multiple include of the same config file. E.g. with the whole model on a network drive, things get even more tricky.

nophead:
It's the fact that the config is included in every file and itself includes a lot of other files. I suspect I could make a skeleton of that with empty files and see the same slowdown since it is only checking mod times, not reading them, so the contents are irrelevant.

I think having the files on a network will be slow anyway because I don't think the stats get cached. Not sure if you can register for changes to non-local files either.

I once wrote a program that needed to monitor dozens of locally networked files. IIRC I just did them one at a time in the idle loop, so it didn't impact the responsiveness of the GUI, but still did them as fast as it could. Not quite the same as this though because it only had to respond to changes when it was actively being interacted with.

nophead:
[...]

PR #1926, opened 2017-01-31
Slowdown with large # of dependencies #1237

nophead:
Stat results are cached for the duration of a dependency scan.

TODO:
* [x] Code review
* [x] Run through module cache tests (https://github.com/openscad/openscad/blob/master/testdata/modulecache-tests/README.txt)


kintel:
Looks good at first glance. Am I correctly seeing that this only caches the `stat()` call, and doesn't try to cache subtrees being accessed more than once?

nophead:
Not sure what you mean by subtrees. What I found was that with a project with 115 files stat was being called more than 11000 times on the same files over and over again. Now StatCache::stat is called just as many times but it only calls ::stat once for each of the 115 files.

One thing I am unsure about is the change I made here: https://github.com/nophead/openscad/blob/d201970b139341881486a1abe510dc01b3feb78a/src/FileModule.cc#L89

FileModule::include_modified() was computing the full path using find_valid_path() and that is expensive as it checks for the file existing. But IncludeFile already has the full path when it is registered, so I changed to using that.

src/StatCache.cc
```diff
@@ -0,0 +1,65 @@
...
+#include "StatCache.h"
+#include "printutils.h"
+
+#include <sys/stat.h>
+#include <string>
+#include <unordered_map>
+
+typedef std::unordered_map<std::string, struct stat> StatMap;
+
+static StatMap statMap;
```
kintel: Since we can have multiple MainWindow instances in the same OpenSCAD process, it doesn't sound right to have a global cache and let each MainWindow clear the cache at will. A per-document cache sounds like the better design here. I _believe_ the worst case would just be bad performance though, so if it's a lot of work to fix, just leave a fat FIXME instead.

MichaelAtOz: Are you talking about just this cache or geometry caches?

kintel: Just this cache, as it's designed to expire for each compilation. The geometry caches are reusable across MainWindows and persist between compilations.

nophead: The life of the cache is just while each window runs includesChanged() and handleDependencies(), so having a cache per window would not make any difference. As it stands now each window will scan just the files it uses. If those files overlap with the set from another window then they will get scanned twice but at slightly different times.

To get a further speed up with multiple Windows then the cached results need to be reused. That means it has to be a global cache but it should be cleared less often. I.e. about once per 200ms as that is how often each window checks for changes.

I could remove the clear function and put a timestamp for each entry in the cache. When it is more than 200ms old I could consider it stale and re-run the stat. That way each unique file would be scanned every 200ms and the results shared between windows.

BTW, why has the Travis build got stuck at the osx build?

nophead:
Hmm,
   Auto reload doesn't seem to work with multiple documents open. Only one window updates automatically. This is before my changes.

nophead:
Auto reload works with multiple windows if the loaded file changes but it doesn't work when a dependency changes. In my test I have the top level file of a machine in one window and just the z axis in another window. I then change a small component used by the z axis and it updates in the machine view but doesn't change the z axis view. I.e. there are two versions of the z axis, one out of date.

In an attempt to debug this I added a PRINTB statement to the timer callback and just managed to confuse myself.

```
void MainWindow::checkAutoReload()
{
    PRINTB("checkAutoReload %s %08X", this->fileName.toLocal8Bit().constData() % this);
	if (!this->fileName.isEmpty()) {
		actionReloadRenderPreview();
	}
}
```
This sometimes prints the name of the main file in the z axis console and the name of the z axis file in the main console but mostly prints both of them in the msys2 console that I launched openscad.exe from.

Why would PRINTB go to the wrong desitnation? mainwin.cc has lots of PRINTB statements and they presumably all go to the correct console.

nophead:
I should add that if I open the files in the opposite order then the last one opened is the one that doesn't work. If I close the first one it the second one starts working.

nophead:
Turning Automatic Reload off is sufficient to make it work in the other Window. Turning it back on stops it working again.

nophead:
To answer my own question, setCurrentOutput(); needs to be called before PRINTB to get it to go to the right console.

src/StatCache.cc
kintel: I'm not concerned with optimization opportunities by sharing the cache between windows.
My concern was that the automatic reload for two windows using an overlapping file set could run at approximately the same time, causing both to use the cache during the same traversal, thus invalidating the cache at unfortunate points in time. This could happen since the dependency checking is driven by timers, and the event loop may be processed during a checking operations (by calling `QCoreApplication::processEvents()` causing more than one MainWindow to process dependencies simultaneously).

Looking closely at the code, I _think_ this should never happen, but the logic for managing this is very convoluted (through flags for blocking `processEvents()` as well as global GUI lock - could really benefit from a cleanup).
..so what you've written should work for now, but would need a refactoring at the same time as the mess mentioned above.

kintel: The Travis OS X build isn't too reliable due to tight resource restrictions on their free tier servers. My plan is to improve that by running a subset of the tests to reduce our resource usage.

kintel:
Two comments on the multiple window issue:
* We had to instate a global GUI lock to manage resource usage. That will block other windows from automatically reloading. Could use some love.
* The `setCurrentOutput()` scheme is a bit messy as well, and may cause output to be temporarily directed to the wrong place, especially for functions triggered asynchronously. Could also need rethinking.

src/StatCache.cc
nophead: I am assuming the GUI is event driven, not multithreaded. I.e. I am assuming if one window calls handleDependencies() then that function will run to completion and not be pre-empted by another thread even if another window's timer fires.

I have updated my cache code and removed the explicit clear, so the results are now shared between windows.

I think I know why the second window doesn't update: Because my windows are using the same modules the first window will call updateDependencies and that will update all the modules and return true. When the second window calls updateDependencies they will already be updated, so it will return false. Hence the second window doesn't think it needs to redraw.

nophead: I am thinking the solution is to return the latest modified time instead of a bool from updateDependencies() and includesChanged() and the windows keep the last results to compare with. Quite a big change though and not related to this change.

kintel: Your assumption holds for now, but it only takes one `processEvents()` call to break it. `processEvents()` is generally done in PRINT to allow output during time consuming functions.

nophead:
I am thinking that fixing #1928 will need all the same tests as this so I should roll that change into this PR to avoid it having to be tested twice. What do you think?

kintel:
I say let's keep them separate. For #1928 we probably need to extend the manual tests a bit.

nophead:
OK, so should I do a new PR for #1928 but based on this one, rather than master, as it touches exactly the same code places. If I did it on master the two PRs would have merge conflicts with each other.

I must disclose that I know very little git foo and the last time I wrote C++ was 2011.

kintel:
yep, creating a new branch based on this one is probably the cleanest. I would fix up the indentation on this branch first to avoid further merging issues.

nophead:
I think I have fixed the indentation. It's hard to tell with tabs as they
mean different things in different editors. I only ever use spaces for my
own projects.

kintel:
I'll look at it. Using tabs was an early decision which I don't agree with and I want to switch to spaces, but there are so many open branches that I fear it will cause chaos.
Ideas are welcome - they key is consistency.

kintel:
@nophead FYI: You had some line ending issues on the new files, which I fixed. That's typically due to a broken local git config, or due to a mix of DOS, cygwin, msys2 environments without a good line ending strategy.

nophead:
Yes the problem with line endings is because msys2 has its own version of git that doesn't use my gitconfig. Sorry about that.

Issue #1928, opened 2017-02-01
Automatic Reload and Preview doesn't work properly with multiple documents open. 


nophead:
If the documents share modules the first window will see the changes and update but the second window doesn't see the changes because includesChanged() and handleDependencies() only return true once for each set of file changes.

nophead:
To state the bug more concisely: If two windows are viewing files which share a common subtree of dependencies then only one widow will update when any of those dependencies change. Changing non-shared files will update the correct window.

kintel:
Fixed by #1929

PR #1929, opened 2017-02-03
Fix for issue #1928 Automatic Reload and Preview doesn't work properl…

nophead:
…y with multiple documents open.

includesChanged() and handleDependencies() now return the latest modification time instead of a bool changed flag.
It is then the responsibility of the client to decide if a change has occurred.

TODO:
* [x] Code review
* [x] Add tests 1 to 3 below (or equivalent) to the manual module cache tests
* [x] Run through tests


kintel:
Thanks for looking into this!

For completeness: Do you have a way of testing that this actually fixes the mentioned problem, i.e. a failing test that succeeds with this patch?

nophead:
I just open multiple windows looking at the same project. One is the entire
project and the other a sub assembly. Before this change changing any file
would only update the first widow opened. The second never updates.

With the new code I checked that changing an include used by both makes
them both update, similarly a used file. I also checked changing files not
used by one window don't make it redraw.

kintel:
I've verified #1928 with a few basic tests, but I'm unable to get the correct result with your branch.
I haven't had the chance to look at your code yet, but see below for my failing tests:
For all three tests below, only the lower-level subtree is automatically updated when a file is changed externally. Top-level file eventually updates if I keep moving the camera around for a bit (same as for the master branch).

**Test 1 (include):**

issue1928-include.scad:
```
include <issue1928-included.scad>
sphere(10);
```
issue1928-included.scad:
```
translate([20,0,0]) cube(20);
```

**Test 2 (use):**

issue1928-use.scad:
```
include <issue1928-used.scad>
sphere(10);
MyModule();
```

issue1928-used.scad:
```
module MyModule() {
     translate([-20,0,0]) cube(12);
}
MyModule();
```

**Test 3 (use and include):**

issue1928-include-use.scad:
```
include <issue1928-included.scad>
use <issue1928-used.scad>

sphere(10);
MyModule();
```

nophead:
>only the lower-level subtree is automatically updated when a file is changed externally.

How do you know the subtree has updated if the window does not redraw the top level with the changes?

I wonder if debug modes behaves differently because no amount of mouse moving caused an update for me before my changes. Why would they, I didn't see any code that says update when the mouse moves? Only one window updated because includesChanged and handleDependencies could only return true once for any single change to a common subtree. I don't see how a mouse move could provoke another return of true.

Your comments about print statements potentially processing other events has me a bit worried because none of this code is re-entrant, and wasn't before my changes. In release mode, which is all I have tried, the GUI lock seems to make the windows update sequentially and prevent re-entering the dependency code. But I don't think there are any active print statements in the dependency code in release mode.

I will see if I can build a debug version and report back.

nophead:
What I found with test1 was that if I touch issue1928-included.scad then a window displaying issue1928-include.scad redraws but a window displaying issue1928-included.scad does not. However if I make an actual change to the file, for example make the cube bigger, then both windows update.

nophead:
Those three tests work fine for me in release mode. I can have all five files open and changing any of them updates the windows affected. I think when touching a file the updates are too fast to see, but making an actual change always results in it being shown.

nophead:
I made a debug version by building QT from source. Deleting all the openscad\objects, running qmake and mingw32-make.exe. It compiled all the files with -DDEBUG and made openscad.exe in the debug directory. Running that doesn't print any debug information in the openscad console, or the shell console. What am I doing wrong?

kintel:
To build a debug version: `qmake CONFIG+=debug`

kintel:
..and to clarify my sentence from last night:
"only the lower-level subtree is automatically updated when a file is changed "

I meant: Only the window with the included/used file open is automatically updated when that file is changed.

..so for Test 1:
* Open both files in separate windows
* edit `issue1928-included.scad` in an external editor

=> The windows with ` issue1928-included.scad` opened is updated, but no the other window.

kintel:
In terms of print, event handling and reentrance: Yes it's a bit scary. The print system could use some refactoring.

We have two measures in place to limit reentrance: A global GUI lock and the `procevents` member in MainWindow. procevents should be kept false most of the time, but needs to be turned on during known long-running processes to make sure the output window gets refreshed. It's easy to mess that up..

kintel:
..and you won't need a debug version of Qt itself. If you're using glibc, I think debug and release versions should be interoperable, unlike msvcrt.

nophead:
I built a debug version of QT because it said I needed to here:
https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/Building_on_Microsoft_Windows#Building_Debug_Version

I found I needed to add --debug=all to the command line to get debug
output. This line
<https://github.com/openscad/openscad/blob/master/src/FileModule.cc#L137>
then constantly prints a module address. Is it correct it constantly prints
or have a broken that?

> The windows with issue1928-included.scad opened is updated, but no the
other window.

That is odd because all three tests all work correctly in debug or release
for me on Win 7.

kintel:
Ah, right DLL naming schemes causes different heaps to be mixed in the same executable. I don't do any Windows development, so I don't have those specific issues on top of my mind : /

The `--debug` cmd-line option is to turn on extra debug output. Some places, like FileModule, outputs quite a lot of it. That's normal - usually I just specify the actual output I want, e.g. `--debug=ModuleCache`.

I'll try to read through your code and do some additional tests today.
Testing on a Linux VM is sometimes also helpful to sort out any platform-specific issues.

I've only been testing on Mac btw.

nophead:
Updating the window of the file that has changed is the only code I haven't touched and uses the original stat, so if that is all that is working then none of my changes are working for you. However I haven't changed the StatCache code and you have tested that before.

I really can't think of anything platform dependent. If you can force a window to update with mouse movements I think the event handling must be not handling timer events correctly. But the old code used them in exactly the same way.

Does it detect dependencies changing correctly with one window open?

kintel:
I think something broke in the master branch - I'll look into it and get back to this.

kintel:
Testing master shows that this commit breaks the auto-reload for me: fc78e0ae33392a34c928aa5ec810700b246cc9f9

nophead:
Do you mean that particular commit or the PR it was part of?

I thought you signed off that PR by running all the manual tests on it? Now
you say it breaks auto reload? Auto reload never worked with multiple
windows and that PR did nothing to fix that. The new one does though.

kintel:
That particular commit broke auto-reload for a single window when included or used dependencies changed.
Apparently, I tested this before that last commit was added. If I back out just that commit, the master branch works as it should.

nophead:
But that commit only affects the test build. Without it the test code does
not build. It has no effect at all on the main build.

nophead:
Ah I see you edited the commit reference.

If that PR breaks it then I suspect clock() does not work on your platform. Is this on a VM or a real platform?

nophead:
The reason it does not work is that clock() returns the time in the current process, not real time. That explains why moving the mouse updates eventually because you will burn some CPU time. Not sure why it works on Windows now!

I am struggling to find a platform independent time with sufficient resolution to resolve 190ms. I normally write my own with conditionals for Linux and Windows but have nevr used OSX. Does OSX have the Posix version of ctime?

kintel:
OS X is Unix and it should be fully Posix compliant. I haven't used the new C++ chrono things myself. 
There is also Boost date_time: http://www.boost.org/doc/libs/1_55_0/doc/html/date_time/posix_time.html

kintel:
Btw., if it's too annoying to make work we can probably kill the cache validity extension stuff for now - it's not critical to make this work.

nophead:
I have re-implemented the cache timestamp with ftime(). That works on Win7 and compiles on Linux, so should work at least as well on that platform and presumably the same on OSX.

astdumptest_include-tests and astdumptest_use-tests fail. Are these fails anything to do with this?

kintel:
Thx.
The astdumptest stuff is unrelated.

Issue #1566, opened 2016-01-24
preview of 2D thickness needs to be improved


MichaelAtOz:
Like #1202 preview of 2D thickness needs to be improved.
Current 2D objects can be confused with 3D objects.
![2d preview height](https://cloud.githubusercontent.com/assets/1972961/12533602/51332f14-c289-11e5-9c48-b8e07ee79c7d.JPG)
So a thickness like 0.001 would be better.
Questionable weather it should sit at Z=0 or centered. I think Z=0.
Also debatable whether 2D preview should show the same colour as 2D rendered objects.

nophead:
2D preview should look as close as possible to 2D render, otherwise it isn't a preview.

doug-moen:
2D preview and 2D render look different, and I agree they should look the
same.

2D objects can't be exported as STL, because they aren't 3D. The
Z-thickness of a 2D object in preview/render is just a presentation issue,
it doesn't affect export. So the question becomes, what's the best way to
present 2D objects in the preview/render views?

I think Michael's ideas about presentation are okay. Give 2D objects a very
small Z-thickness (if zero thickness isn't feasible for some reason). They
should sit at Z=0, not at Z=-1 as in the current render view. Giving them a
distinct colour also makes sense.

nophead:
Yes the F6 view is nice because it looks completely different, so you can't
mistake it for 3D. Can preview be the same, i.e. green with red edges?

F5 isn't currently at Z = -1, it's actually Z = -0.1, which is weird.

kintel:
@nophead See #1202

GilesBathgate:
1. Preview (F5) mode gives 2D primitives a "thickness" but the following shows that its not necessary.

```c#
polyhedron([[0,0,0],[0,10,0],[10,10,0],[10,0,0]],[[0,1,2,3]]);

translate([10,10,0])square(10);
```

2. In Render (F6) mode you can get the square to display by commenting out the polyhedron, or the polyhedron to display by commenting out the square, but never both.

doug-moen:
@GilesBathgate: that call to polyhedron() is bad. It returns a non-manifold
mesh which will cause CGAL assertions if you try to operate on it further.
There has been discussion elsewhere about fixing polyhedron to report bad
arguments.

But yes, its good to point out that preview is already capable of
displaying 2D objects with no thickness.

GilesBathgate:
@doug-moen What assertions?

Example 1)
```c#
hull(){
[...]

PR #1933, opened 2017-02-08
#1566 Render 2D previews much thinner than 1 mm

kintel:
Needs testing before merge!

Next step would be to deal with the z fighting - there are a few ways to fix that, but that may require refactoring some of the rendering code.
Keep in mind that we also have the Throwntogether mode which renders backfaces in different colors.

### DIFF
--- a/src/polyset-gl.cc
+++ b/src/polyset-gl.cc
@@ -82,8 +82,8 @@ void PolySet::render_surface(Renderer::csgmode_e csgmode, const Transform3d &m, 
 	}
 #endif /* ENABLE_OPENCSG */
 	if (this->dim == 2) {
-		// Render 2D objects 1mm thick, but differences slightly larger
-		double zbase = 1 + ((csgmode & CSGMODE_DIFFERENCE_FLAG) ? 0.1 : 0);
+		// Render 2D objects quite thin, but differences slightly larger
+		double zbase = 0.001 + ((csgmode & CSGMODE_DIFFERENCE_FLAG) ? 0.001 : 0);
 		glBegin(GL_TRIANGLES);
 
 		// Render top+bottom
@@ -211,8 +211,8 @@ void PolySet::render_edges(Renderer::csgmode_e csgmode) const
 			}
 		}
 		else {
-			// Render 2D objects 1mm thick, but differences slightly larger
-			double zbase = 1 + ((csgmode & CSGMODE_DIFFERENCE_FLAG) ? 0.1 : 0);
+		// Render 2D objects quite thin, but differences slightly larger
+			double zbase = 0.001 + ((csgmode & CSGMODE_DIFFERENCE_FLAG) ? 0.001 : 0);
 
 			for (const Outline2d &o : polygon.outlines()) {
 				// Render top+bottom outlines


RonaldoCMP:
I have not found any explanation why 2D objects could not be previewed with 0 thickness as it is rendered. It seems that 2D thickness is misleading not only the user. The preview and render of the following code are different and are not void as they should:

<code>
 intersection(){
      rotate([30,0,0]) square(20);
      rotate([25,0,0]) square(20);
}
</code>

2D objects seem to be processed as they are 3D objects by preview.

kintel:
@RonaldoCMP The previewer doesn't distinguish between 2D and 3D objects. We simply convert 2D objects into thin 3D ones. We _can_ writing code the handle 2D preview properly, it's mostly a matter of work, and careful testing. This could be a good task for a new developer who has experience with or wants to learn about 3D rendering.

This pull request is to provide a minimal starting point to see what issues arises, and perhaps spawn motivation for helping or testing.

RonaldoCMP:
2017-02-08 13:17 GMT-02:00 Marius Kintel <notifications@github.com>:

> @RonaldoCMP <https://github.com/RonaldoCMP> The previewer doesn't
> distinguish between 2D and 3D objects. We simply convert 2D objects into
> thin 3D ones.
>
Ok, but why the render of my example is a rectangle whose dimensions come
from nowhere?

kintel:
It's because the preview engine resolves both the intersection and transformation after we've given it thin 3D objects. Internally, your `square(20)` is converted to the equivalent of `translate([0,0,-0.5]) cube([20,20,1])`. Try substituting those and you should get the same behavior.

nerdfever:
Until there's a proper fix to render zero-thickness 2D objects, can we get a warning?

The user should be warned that in preview, 2D text of thickness less than n (n currently is 1) will be rendered as being n thick.

nophead:
But that would give a warning on any project using 2D objects. Anybody
using 2D objects should know how they get rendered. It might be a surprise
the first time but you soon learn.

RonaldoCMP:
Nope.
```
difference(){ 
    intersection(){ 
        rotate([30,0,0]) 
            square(20); 
        rotate([25,0,0]) 
            square(20); 
    } 
    projection()
        intersection(){ 
            rotate([30,0,0]) 
                translate([0,0,-0.5]) cube([20,20,1]); 
            rotate([25,0,0]) 
                translate([0,0,-0.5]) cube([20,20,1]); 
        } 
}
```
render2d

kintel:
@RonaldoCMP oh, I misunderstood. For render (as opposed to preview), the story is different: We convert the 3D transformation into a 2D one, see https://github.com/openscad/openscad/blob/master/src/GeometryEvaluator.cc#L573

This does give interesting results if applying 3D transformations to 2D objects.

As part of the effort on this topics, we could more clearly define what should happen when applying transformations to 2D objects. As the transformation modules themselves doesn't know what shapes are passed as parameters, we might need to take extra care to implement properly.

wolf105:
I have an issue with this request, as this request is not well thought through. The whole discussion so far has been based upon the (unspoken) assumption that software needs to be adjusted to match human thinking, including human idiocy. 
The idiocy I am referring to here is taking for granted that a 2D object not only can, but actually does exist in a 3D world. As far as my understanding of reality goes, that assumption is false and far from true. 
2D, in my understanding at least, is an abstraction that makes it convenient to draw on a **flat** surface, and fails as soon as the concept of **flat surface** is abandoned, and the idiocy referred to above lies in taking an abstraction for the real thing. An abstraction is useful - no doubt about that - but only as long as the limitations of its connection with reality are not obscured. And that is where this request is failing.
My suggestion would be to instruct the compiler to accept 2D commands only into a 3D environment after the 2D shape has been extruded into a 3D shape - this way the user programmer is forced to think about what he/she is doing and be aware of the underlying implementation. Automatic extrusion, with whatever default you may choose, is going to create these issues again and again. 

The true underlying cause for this issue is obscurantism - the tendency of lifting the user away from reality and into a dream world. The ability to directly place 2D objects in a 3D environment is a bug - a bug not in the implementation of the software, but a bug in the thinking behind it.
wolf

nerdfever:
@wolf105 One of us misunderstands the issue (could be me).

The issue as I understand it is how to preview render a 2D object *after* it has been extruded into 3D.

Currently OpenSCAD preview (F5) renders text as 1 unit thick even if it has been extruded to a smaller thickness (0.5 units, for example). 

The result is a misleading preview.

The full F6 render works fine (but is of course much slower and so less useful for debugging).

See this thread for background: http://forum.openscad.org/text-2D-or-3D-td12217i20.html

(That thread is what led me to this issue on Git.)

RonaldoCMP:
@wolf105 You are messing things up. It is true that our discussion doesn't
use the formal language of mathematics. But it is absolutely clear to
everybody here that 2D and 3D spaces are distinct. This does not prevent us
from representing 2D objects in 3D spaces via isomorphisms. May be you are
not aware of it but any 2D vector space  is isomorphic to any plane in any
3D vector space. A very similar abstraction allows the user (you call
idiot) to see  a representation of the 3D space in a 2D screen. Who is
promoting the obscurantism?


2017-02-08 20:59 GMT-02:00 wolf105 <notifications@github.com>:

nophead:
>Currently OpenSCAD preview (F5) renders text as 1 unit thick even if it
has been extruded to a smaller thickness (0.5 units, for example).

No that isn't true. Once you extrude a 2D shape it becomes 3D and is
previewed with the correct thickness.

linear_extrude(height = 0.0001)
   text("A");

Produces a very thin A, not one 1 unit high.

kintel:
@wolf105 This issue is about iteratively improving software. Anything not related to that is probably better discussed on the mailing list, perhaps in context of OpenSCAD2 or smth.

thehans:
I suspect a thickness this small was not previously implemented due to z-fighting issues.  

I have not tested this patch directly, but even in OpenSCAD versions without this patch, you can test the results of this sort of change by abusing the scale operation (you can scale the z axis independently, even for "zero thickness" 2d parts).  

Try the following example, and see if it looks acceptable to you (I personally don't think so)
```
scale([1,1,0.001]) difference() {
  square(10);
  circle(10);
}
```
From a little testing and playing around I've done, I think using a thickness of 0.1 would be an improvement over the 1mm thickness that is currently used while having reasonable resistance to z-fighting.  Much smaller than that and z-fighting becomes very noticeable very fast.

Even at 0.1 its easy enough to induce z-fighting issues if you create a large enough 2d object and then zoom the camera out until the artifacts occur.

Try this next example, "zoom to fit", then orbit the camera around, you will probably see artifacts.  I do anyways.
```
scale([1,1,0.1]) difference() {
  square(200);
  circle(200);
}
```

nophead:
Since the 2D operations are now pretty fast perhaps all 2D geometry should
be evaluated before preview. OpenCSG seems to have no problem displaying
thin objects but evaluating the 2D ops in 3D with thin operands causes Z
fighting as well as unexpected results.

nophead:
Or perhaps the negative shapes should still be 1 unit high.

kintel:
making negative shapes orders of magnitude thicker might indeed help.

I like the idea of evaluating 2D shapes. That's a bit of a separate investigation into performance (think animation) and color handling.

thehans:
Yeah I really like this idea of positive and negative shapes having different z scales, this does seem to fix the problem in my "scale abusing" tests at least. 

And its probably beneficial to go even larger than 1 for the negative shapes.

I did some more testing and, and it works out roughly that you can comfortably zoom to fit an object of magnitude equal to the difference in magnitude between the negative and positive shape.

So in this example, objects can be comfortably rendered up to a size of 1e(2 - (-3)) == 1e5
```
difference() {
 scale([1,1,1e-3]) square(1e5);
 scale([1,1,1e2]) circle(1e5);
}
```

MichaelAtOz:
> And its probably beneficial to go even larger than 1 for the negative shapes.

Presumably `#` and `%` modifiers make the shape positive for preview purposes?

e.g.
```
difference() {
  square(10);
  #circle(10);
}
```

wolf105:
Let me respond here, comment by comment.
to @nerdfever: On a practical side, my issue is two-fold: 
the minor one is(and it is so minor I can live with it, but I suspect many aspirant OpenSCAD users wont be able to): How do you view a shape, edge-on, if it is very thin? Because it is so thin, you cannot view it, it is invisible. Use polyhedron() with a points and faces list that reflects just one facet, and you'll see what I mean.

The second issue is far more insidious, as it centers about rounding, and it is one I cannot live with at all. And because English, like any other human language I am reasonably familiar with, automatically rounds (and you as a user of this language may not even be aware of it), this issue cannot be discussed at the required level of accuracy in plain English. It can be discussed properly only using a language that does not permit implicit rounding. This means using Mathematics or a programming language with a compiler strong enough to point out to the programmer where his/her thinking went wrong. C is not such a language. For example: In front of every C function the programmer needs to declare the type of the function's output. Since the function's name encodes a pointer to the function's result, and the type declaration how to interprete the sequence of bits that represent the result, it is easy to introduce bugs via inappropriately interpreting memory contents.

Let me demonstrate what I mean with rounding within a language. @RonaldoCMP writes in this thread "... This does not prevent us from representing 2D objects in 3D spaces via isomorphisms. May be you are not aware of it but any 2D vector space is isomorphic to any plane in any 3D vector space..." I am very well aware that this statement has been considered to be true for almost 2000 years. Read up the history of the 'parallel axiom' if you want to get more background. Today, anybody reasonably deeply knowledgeable in geometry knows that this statement is false. If I may jump the details, the thinking error that @RonaldoCMP fell for, as did Euclid 2000 years before him, is that on the surface of a cube, it is well possible to draw a triangle whose sum of inner angles is 180 degrees. But on the surface of a sphere, this is not possible, as the sum of angles there is always greater than 180 degrees. And if you draw a triangle on the surface of a hyperboloid, the sum of angles is always less than 180 degrees. And since the spaces I have exampled differ in their metric, they cannot be isomorphic either. 
To quote Wikipedia:
"In mathematics, an isomorphism (from the Ancient Greek: ἴσος isos "equal", and μορφή morphe "form" or "shape") is a homomorphism or morphism (i.e. a mathematical mapping) that admits an inverse.[note 1] Two mathematical objects are isomorphic if an isomorphism exists between them. An automorphism is an isomorphism whose source and target coincide. The interest of isomorphisms lies in the fact that two isomorphic objects cannot be distinguished by using only the properties used to define morphisms; thus isomorphic objects may be considered the same as long as one considers only these properties and their consequences." 
And this is the rounding reference: If you include 'Sum of Angles' as an inherent part of the properties of a space or surface, a triangle on a sphere is not isomorphic with a triangle on the surface of a cube. But if you exclude that requirement, @RonaldoCMP is right. It is possible to remove 'Sum of Angles' from the list of properties, but does that reflect the world we live in? 
Take a GPS. If you want one that provides you with a position on earth within an accuracy of a couple of kilometers, you are permitted to assume that the sum of angles of a triangle is 180 degree. But if you want accuracy better than a meter or so, you have to build into your software that the sum of angles is less that 180 degrees. By how much is determined by the strength of the gravity field between your GPS and the satellites that provide it with signals. Or, at the atomic end of the scale, if you want to understand what works you have to include spin into the list of properties that matter has. If you don't, all particles obey Maxwell statistics, Fermi-Dirac and Bose-Einstein statistics need not be considered. In other words, the only matter that can exist is an ideal gas, solids and liquids are not possible. If you do, the real world is not isomorphic. If you don't, you don't have humans to argue whether isomorphism is for real or not.

But let's stick closer to OpenSCAD, and let me give examples of where rounding affects the user, even if his/her scale is limited to what a small 3D printer can produce.

The first example is about z-fighting. If you use this code to create two intersecting cubes
rotate(1,[1,0,0]) cube(10,center=true);
color("red")  rotate(-1,[1,0,0]) cube(10,center=true);
and move the camera around it, you'll see, once you have chosen an appropriate viewing angle, that the intersection of the two cubes is saw-tooth shaped, and that the size and shape of the saw-tooth depends upon the viewing angle, and not on the object viewed. In other words, what you see is an expression of the rounding choices that OpenCSG, the library OpenSCAD uses for Preview (F5), makes. Since CGAL, the library OpenSCAD uses to render, does not do any rounding, Render (F6) is not affected, and you can do boolean operations with it.
difference()
{cube(10);
rotate(1,[1,0,0]) cube(10,center=true);
color("red")  rotate(-1,[1,0,0]) cube(10,center=true);}
But Preview (F5) fails.

The second example involves "degenerate triangles". It is one of the many problems issue945e.stl has. Consider the two triangles
rounding
ABC and DFE. The dots in the picture represent possible positions that the vertices A,B,C,D,E,F, can take. The positions between the dots are forbidden because of rounding. The distances between the vertices are not to scale, so as to promote understanding of the issues involved in fixing "degenerated triangles". ABC and DFE are the same triangles, shifted sidewards, except that vertex E has been rounded downwards, vertex C upwards from the true position which is supposed to lie exactly on the average of the y-coordinate of the two vertices and is not shown in the picture. 
CGAL uses the concept of half-edges and half-spaces, which is to say that edges and spaces are ordered entities. In the pictured example, CGAL treats half-edge AB as different from half-edge BA, the difference being that AB runs from left to right, BA from right to left. In CGAL's view, the triangle composed from half-edges AB, BC and CA is inside-in, as the order in which one travels around the triangle from A to A, via B and C, is counter-clockwise. In the view of polyhedron(), by the way, travelling around a face counter-clockwise makes that face outside-in. In triangle DFE, orientation is the other way around, for CGAL the triangle is inside-out, and for polyhedron() it is outside-in. 
The difference between ABC and DFE is just in how the y-coordinate of vertex E was rounded. If triangles have an altitude as low as ABC and DFE have, rounding has the implication that the orientation of either ABC or DFE are not acceptable to CGAL for boolean operations. That is one of the consequences the use of order has, and it means that by conversion from CGAL's arbitrary precision number format to the finite precision format that OpenSCAD, and .stl files, uses, "degenerate triangles" are unavoidable and will happen. But it also means that "degenerate triangle" is a cover-up for poor rounding strategies. The presence of "degenerate triangles" has another consequence, as far as I currently can see, but cannot currently prove. It spawns other "degenerate triangles" in an attempt to fix the orientation issues inherent in exiting from CGAL number format to OpenSCAD number format, and that worsens the issues around "non-manifold". "Non-manifold" is, like "degenerate triangle", a term designed to distract from the real issue, the real issue being the order in which vertices are presented to the tesselator. If I get the order right in my attempts do deal with "degenerate triangles", any of the test-files I have been presented with so far are acceptable for boolean operations. But I haven't figured out an algorithm that can do it as well, an algorithm that stands up to what I can throw at it myself, or what other users have encountered already, and is presented to me for throwing at it.

My interest in doing this kind of work is in dramatically speeding up OpenSCAD's rendering times. I have already shown that 10000 fold speed increases can be achieved, but the code I use is buggy, in that it e.g. does not gracefully handle self-intersections, and thus its output cannot be guaranteed to be acceptable to CGAL. Since "non-manifold" is one of the error messages I receive, investigating its origin led to "degenerate triangles". And solving the issue of "degenerate triangles" has led to a strategy by which I expect to be able to achieve my goal. Whether I get there including OpenSCAD, and thus make the speed gains available to others as well, or whether I have to exclude OpenSCRAP, will depend upon the responses of the OpenSCAD community, in particular their willingness to adhere to Nikolaus Kopernicus' maxime: 'A theory is only worthwile to the degree it reflects the real world.' And to me that means 'OpenSCAD turns to OpenSCRAP if it does not meet my needs for accuracy, speed and predictability'.

Predictability is at stake here. The availability of 80 bit numbers means I need not worry about accuracy. Speed I can inject myself. But predictability arises from the compiler, and thus depends upon others. If they tilt the scales towards convenience at the cost of predictability, I can only withdraw and leave the community to fend for itself, since usability has been destroyed. 

I can only hope that my above examples have instilled a sense of the consequences of what to expect if this proposal goes through. With a skillful use of words, it is possible to void all I have said. But it does not change the facts, it only hides them, and they bite later. Ask Marius Kintel about special variables . . .
wolf

kintel:
This is obviously a too complex topic for a pull request. I suggest we continue this discussion on the mailing list.

PR #1938, opened 2017-02-12
redo find-replace highlights and word count fix MCAD

kipbits:
redo find-replace highlights and word count fix MCAD

PR #1922, opened 2017-01-30
FindReplace: Highlight found match words & find-count

kipbits:
These changes are to implement highlighting of found matches for
enhanced find&replace functionality.

Highlighting was implemented by designating a new indicator number 9 to
represent all find-related-indicators, and are updated through the
addition of the following functions&variables:
ScintillaEditor::resetFindIndicators(
MainWindow::hideFind(
MainWindow::showFind(
MainWindow::showFindAndReplace(
& variable:
ScintillaEditor.findIndicatorNumber

A few other changes were made to current class member names to prevent
confusion with the new functionality, and to improve readability:
MainWindow::find(
   ->   MainWindow::showFind(
MainWindow::findAndReplace(
   ->   MainWindow::showFindAndReplace(
ScintillaEditor.indicatorNumber
   ->   ScintillaEditor.errorindicatorNumber

I also made a few adjustments to MainWindow.ui I tried to go for a
compromise of the original legacy design, and what changes have been
made to the design, I hope this aesthetic change is accepted along with
the find-highlight-functionality as 1 PR.

Also Note, resetFindIndicators( returns an int. this is a find count
number, and I still have plans to display a find count in my next PR.

Thankyou

PR #1947, opened 2017-02-14
Find&Replace: Highlight match-words & Display find-count (FINAL)

kipbits:
These changes are to implement highlighting of found matches for
enhanced find&replace functionality.

TODO:
* [x] Code review

Highlighting was implemented by designating a new indicator number 9 to
represent all find-related-indicators, and are updated through the
addition of the following functions&variables:
ScintillaEditor::resetFindIndicators(
MainWindow::hideFind(
MainWindow::showFind(
MainWindow::showFindAndReplace(
& variable:
ScintillaEditor.findIndicatorNumber

A few other changes were made to current class member names to prevent
confusion with the new functionality, and to improve readability:
MainWindow::find(
   ->   MainWindow::showFind(
MainWindow::findAndReplace(
   ->   MainWindow::showFindAndReplace(
ScintillaEditor.indicatorNumber
   ->   ScintillaEditor.errorindicatorNumber

I also made a few adjustments to MainWindow.ui I tried to go for a
compromise of the original legacy design, and what changes have been
made to the design, I hope this astetic change is accepted along with
the find-highlight-functionality as 1 PR.

Also Note, resetFindIndicators( returns an int. this is a find count
number, and I still have plans to display a find count in my next PR.

Thankyou

Addition: this PR also include working find-count display

there was some confusion getting MCAD configured properly
reference PR's:
#1922
#1938


kipbits:
ok try again. each time i do this it seems to mess-up the repo for the previous PR
I'm did this one with a combination of terminal and Github desktop
ok, I followed README
    git clone git://github.com/openscad/openscad.git
    cd openscad
    git submodule update --init

then changed .git/config
	[remote "origin"]
		url = https://github.com/kipbits/openscad.git

fingers crossed....

kipbits:
2 success and 1 overrun, what is the hold up?

kintel:
Don't worry about the overrun - it's just the Mac VM running out of time.
I'll review this as soon as I have some time.

src/scintillaeditor.cpp
```diff
@@ -540,27 +571,16 @@ void ScintillaEditor::replaceSelectedText(const QString &newText)
     if ((qsci->selectedText() != newText)&&(qsci->hasSelectedText())) qsci->replaceSelectedText(newText);
 }
 
-void ScintillaEditor::replaceAll(const QString &findText, const QString &replaceText)
+void ScintillaEditor::replaceAll(const QString &findText, const QString &replaceText)//QScintilla bug no longer an issue
```
kintel: Why is the QScintilla bug no longer an issue?

src/MainWindow.h
```diff
@@ -158,9 +158,10 @@ private slots:
 
 private slots:
 	void selectFindType(int);
-	void find();
-	void findString(QString);
-	void findAndReplace();
+    void hideFind();
```
kintel: Would you mind using the same indentation style as the existing code (single tab indent)? I'm aware that this isn't the optimal solution, but consistency is important.
This applies to all changes.

kipbits: there are other inconsistencies like this already in the file. should i change these too?

src/scintillaeditor.cpp
kipbits: I'm having trouble reproducing this bug in qscinitilla2.9.3. I suppose i could try and add back in a case for "#if QSCINTILLA_VERSION == 0x020903"? my memory is a bit fuzzy on this but i seem to remember selectAll() creating an issue where focus was no longer on the findreplace pannel after the hitting the button, and all my text was still selected. don't know if this is a real issue, but changed it. I suppose i can change it back, or add a seperate compiler condition. let me know what you want

src/MainWindow.h
kintel: Feel free to update any other inconsistencies - I tend to only point out the ones I can see in diffs.

src/scintillaeditor.cpp
kintel: If you're uncertain I'd say put it back the way it was. I'm afraid of reintroducing an old bug on systems with an older version of QScintilla.

Unless you feel like installing QScintilla 2.7, reproduce the bug, and check that it's fixed in the new code (which I totally understand not being worth doing)..

kipbits:
will get to it soon

kintel:
Thx - looks good!

t-paul:
Looks like it causes build issues on Debian8 and OpenSUSE 13.2 ([OBS](https://build.opensuse.org/package/show/home:t-paul/OpenSCAD)), so maybe the version check is slightly off?

```
[ 150s] src/scintillaeditor.cpp: In constructor 'ScintillaEditor::ScintillaEditor(QWidget*)':
[ 150s] src/scintillaeditor.cpp:144:24: error: 'FullBoxIndicator' is not a member of 'QsciScintilla'
[ 150s] qsci->indicatorDefine(QsciScintilla::FullBoxIndicator , findIndicatorNumber); 
```

kipbits:
perhaps it's your version of Qscistintilla? could try changing the indicator style although i kind of liked FullBox

t-paul:
Yes, but I can't change the QScintilla versions of existing distributions, so we need to find a way to at least disable usage of newer features for those.

Windows/MXE build also fails (According to [mxe.cc](http://mxe.cc/#packages) this has Qt 5.8.0 and QScintilla 2.8.4):

```
../src/scintillaeditor.cpp: In constructor 'ScintillaEditor::ScintillaEditor(QWidget*)':
../src/scintillaeditor.cpp:169:24: error: 'FullBoxIndicator' is not a member of 'QsciScintilla'
  qsci->indicatorDefine(QsciScintilla::FullBoxIndicator , findIndicatorNumber); 
                        ^
../src/scintillaeditor.cpp: In member function 'virtual bool ScintillaEditor::eventFilter(QObject*, QEvent*)':
../src/scintillaeditor.cpp:784:50: error: 'CtrlModifier' is not a member of 'Qt'
   if ((ke->modifiers() & ~Qt::KeypadModifier) == Qt::CtrlModifier|Qt::AltModifier) {
                                                  ^
../src/scintillaeditor.cpp:785:49: error: 'Meta' is not a member of 'Qt'
    QKeyEvent* event = new QKeyEvent(ke->type(), Qt::Meta, ke->text());
```

kintel:
FullBoxIndicator was added in 2.9 - not sure what our minimum requirement is (e.g. what version run on Ubuntu 12.04?), but we definitely cannot use this as it is. At the minimum, a compile-time version check is needed.

t-paul:
Hmm, `FullBoxIndicator` might be a bit nicer for some cases, but even with the dark color scheme (Tomorrow Night) and large font (Noto Moto@18) it does not seem to warrant extra code or more complicated build. In the light color mode I personally use the difference is almost impossible to see.

I guess a simple `#define` to switch all the indicators would be fine though.

Ubuntu 12.04 (which has 2.6 I think) was kicked off of OBS, so I guess we might drop that and have 14.04 as oldest LTS?

Debian8 has `libqscintilla2-11-2.8.4+dfsg-1`, OpenSUSE 13.2 has `libqscintilla2-11-2.8.2-2.1.8`.

Comparison against `RoundBoxIndicator` we already use:
screenshot from 2017-03-11 10-25-56

kintel:
Travis still uses Ubuntu 12.04, but their 14.04 version has been stable for a while, so perhaps we can finally drop 12.04 support.

kipbits:
you should be able to fix this by switching to roundboxindicator. anyway i broke finger the other day so i'm going to take a little vacation from coding

kipbits:
just fyi(not sure how much else i have to contribute here, but will tell my horror story). I'm having complications from surgery and the whole thing. i'll probably be gone longer than originally anticipated. anyway just wanted to warn people about going to teaching hospitals(university associated). Also make sure you have a clear understanding with your anesthesiologist. most importantly don't rush into any quickly scheduled(next day) surgery.

Good luck all

...and keep your hands in a safe position when closing the door behind you

MichaelAtOz:
Ouch!
I noticed the find count in the snapshot, handy, thanks.

Good luck.

Issue #1940, opened 2017-02-12
Customizer crashes when trying to add a preset


SparkyCola:
Hi, I just tested the customizer on this tiny test: 

`/* [Test] */`
`//test`
`x = 5;`
`cube(x);`

and get the following error when on the console when I hit the "+"-key and enter a name (nothing fancy, only lower-case ascii without any spaces):

`openscad-nightly: /usr/include/boost/optional/detail/optional_reference_spec.hpp:108: T& boost::optional<T&>::get() const [with T = boost::property_tree::basic_ptree<std::__cxx11::basic_string<char>, std::__cxx11::basic_string<char> >]: Assertion `ptr_' failed.
Aborted (core dumped)`

I'm using version 2017.02.11.nightly (git e990ac4) on Ubuntu 16.10 64-bit.

kintel:
Fixed by #1948

PR #1948, opened 2017-02-14
Solves issue1940: Adds check for parameterSets object in json

amarjeetkapoor1:
Solves issue #1940 by adding check for parameterSets object in JSON object and create paramaterSets object if not present.


kintel:
I still have problems creating a new parameter set from scratch. It doesn't crash, but doesn't do anything either.
Not sure if that's related - I cannot reproduce the original crash either..

amarjeetkapoor1:
Everything is working fine on my system. I am being able to create parameter set from scratch and there is also no crash.

kintel:
@amarjeetkapoor1 OK, I'll test some more.

kintel:
I found some other issue, but I'll open new github issue for those

Issue #1956, opened 2017-02-21
Parse error caused by an included file does not recompile when fixed


nophead:
FileModule::parse() returns null when there is a syntax error. This causes lib_mod to be set to null here: https://github.com/openscad/openscad/blob/master/src/ModuleCache.cc#L117 and so this line: https://github.com/openscad/openscad/blob/master/src/ModuleCache.cc#L74 never checks the includes again.

I think this has always been the case but I have only noticed it after my changes.

nophead:
The same problem occurs in winmain.cc. Parse errors cause the root module to be deleted but the is where its include files are stored and they can cause parse errors.

I think the solution is to store the dependency graph in its own data structure.

nophead:
I think I can fix it by storing the last good module pointer in the cache entry and the window and use that for the checking includes. Whenever there is a good parse delete the old module and set it to the new one.

kintel:
@nophead To clarify that I understand this correctly: Is this only a problem with automatic reload, or are you also seeing this when manually triggering a preview using F4 or F5 ?

nophead:
The problem in mainwin.cc only affects auto reload because and you can force recompilation with F5 or F4. The effect of the same problem in the ModuleCache.cc is worse because you can't force a used module to be recompiled, unless you touch it, or clear the module cache, because it stops looking at the include times.

oysteinkrog:
I can confirm this problem. I have always just had to manually reload the window.

nophead:
Fixed by #1958

PR #1958, opened 2017-02-22
Issue 1956

nophead:
Fix for issue #1956 

TODO:
* [x] Code review
* [x] Add or modify test case in https://github.com/openscad/openscad/tree/master/testdata/modulecache-tests


kintel:
One issue: If I start out with a file with errors, neither the auto-reload nor manual F5 will trigger a recompile upon fixing the error.
See [README.txt](https://github.com/nophead/openscad/blob/b8c3880fdefd03a94193f2701168fa6c6698def0/testdata/modulecache-tests/README.txt) test 18 and 19.

nophead:
I am looking at how to make the parser return the dependencies even when parsing fails. This line: https://github.com/nophead/openscad/blob/master/src/parser.y#L648 looks like a memory leak to me. I.e. it creates a new module but doesn't return it so it looks like it isn't ever deleted, just left dangling when parse is called again.

nophead:
Hmm, seems like there are other memory leaks as repeatedly pressing F5 causes the working set to increase, with or without a syntax error.

nophead:
Works with includes that start off broken and then get fixed now.

src/mainwin.cc
```diff
@@ -732,8 +733,9 @@ void MainWindow::updateReorderMode(bool reorderMode)
 
 MainWindow::~MainWindow()
 {
-	if (root_module) delete root_module;
-	if (root_node) delete root_node;
+	delete root_module;
+	delete last_good_module;
```
nophead: Yes that was a change you made in minor style tweaks, not me. It is fixed in my latest version. last_good_module is renamed parsed_module and it is deleted but root_node isn't because if it is non null it is guaranteed to be the same.

src/StatCache.cc
```diff
@@ -59,7 +59,7 @@ int StatCache::stat(const char *path, struct stat *st)
 		}
 		statMap.erase(iter);                            // Remove stale entry
 	}
-	CacheEntry entry{};                               // Make a new entry
+	CacheEntry entry;                                 // Make a new entry
```
nophead: During a compilation you get a warning for every member that isn't explicitly initialised. Daft because it is guaranteed to be zero.

The structure doesn't need initialised because if stat fails it isn't used, so I removed the initialisers to get rid of the warnings.

src/mainwin.cc
kintel: My change shouldn't change behavior; deleting an object doesn't modify the pointer. This was merely a coding style update.

src/StatCache.cc
kintel: Which compiler gives this warning? To my knowledge, this is explicitly how structs are supposed to be zero-initialized in C++11.

nophead: Both mingw64 in the msys2 on Windows 64 build and whatever Travis uses for the Linux builds. 

I don't know C++11 because I left my job writing C++ in 2011 to make Mendel90 kits! I used to initialise structs with = {0} but that gives exactly the same warnings with these compilers.

kintel: thx - that's an old gcc with not-yet-full C++ support. Anyway, I agree that initializers are not strictly necessary here.

Issue #1961, opened 2017-02-24
Warnings and error messages only shown five times


nophead:
This line: https://github.com/openscad/openscad/blob/4a0ac7b2fa6e8ca390e3b95e1c2d2541cf9d77af/src/printutils.cc#L62 prevents the same message being printed more than five times consecutively. The bug is that the ring buffer is not reset at the start of compilations, so a single error or warning in a project will only be shown for the first five compilations.

See http://forum.openscad.org/Strange-behaviour-with-assert-tt20545.html

kintel:
Fixed by #1962

PR #1962, opened 2017-02-24
Fix for #1961: Warnings and errors only shown five times

nophead:
lastmessages now cleared at the start of the compilation.


kintel:
Thanks - good catch!

Issue #149, opened 2012-07-19
Feature Request: Make CGAL/OpenCSG render type available to the code. [$100]


daid:
When making my objects in OpenSCAD. I use the OpenCSG render for a "quick render" and CGAL for export, as most people will do.

Now, I want my circles to be pretty detailed for the CGAL export, so I have a  line "$fs=0.4;$fa=5;" at the top. However, this slows down the OpenCSG render a lot. So I comment it out while I am doing quick refreshes with OpenCSG.

Now it would be nice if there was some special variable that was set so you know what type of rendering is used, and allows you to make different code paths depending on that variable.
##

nophead:
>Perhaps an environment array of some sort would make sense, e.g. set $ENV["rendertype"] to "Preview" or "Render" ?

Seems complicated because it looks like an associative array and OpenSCAD doesn't have them. It is also ugly, so I would end up assigning it to a $ variable anyway. Doing so lets it be overridden in a subtree, which might be useful.

 >Does a program want to distinguish between F6 (via menu/cmd-line) or via render()?

Not in my use case no.  I use F5 to draw an assembly view and F6 to lay out the STLs for printing. My assembly view has render()s in all the top level objects, just before I colour them.

>Also variable for GUI v's cmd-line??

Not needed because you can make your own with -D.

MichaelAtOz:
Given the potential for other environmental values, as per #380, what about a function?
You can still do a $ variable,
`$preview=system("preview");`

[Extensible](https://en.wikipedia.org/wiki/Extensibility).

I agree, "preview" should be true within render(), there may be a use for also knowing you are within a render() on top of that, but that's a separate issue.

My use thoughts first went to different $fn etc resolution depending on what state you are in.

nophead:
In my mind $variables are extensible, so I added one more :)

I don't see how one more is a problem. Even if somebody already had used $preview it will simply override the default, so it is fully backwards compatible.

I am not sure why you want to know you are inside render(). You can always put $variables in your code to affect subtrees. This PR solves a different problem. My files are always an assembly view, so it never makes sense to F6 them. To see the printed parts laid out for printing I need to change one line and then press F6. If I accidentally press F6 CGAL goes away for ever trying to make a union of an entire assembly. Now I have

if($preview)
    display_assembly();
else {
    display_box_stl();

    translate([0, box_width + 2, 0])
        display_box_lid_stl();
}

F5 gives this:

[...]

PR #1963, opened 2017-02-25
Solution for #149: $preview variable added.

nophead:
$preview is true for an OpenCSG preview, i.e F5, and false for F6.
Also set when the command line is used to generate a png with OpenCSG.

TODO:
* [x] Code review
* [x] Add test cases producing all reasonable output formats
* [x] Consider/test cmd-line override
* [x] Consider/test source code override
* [x] Consider/test dynamic scope overrides


t-paul:
I've mentioned in other places why I think creating more `$` special variables it not a good way forward, that's something @kintel will need to decide...

Still the feature as such would be a very useful thing to have, so assuming it's the way to go, I have 2 questions
- Could we make this `boolean`? It feels strange to have it as number.
- Should we make it an error to assign a new value via script (which would capture the current `-D` implementation too I think)?

doug-moen:
I agree with @t-paul, it ought to be a boolean variable, `true` and `false`.

nophead:
Yes it could be a boolean but to convert to two numbers for something like
$fn you need the ternary operator so 0, 1 is actually more useful because
you can make two numbers with a formula, as in C.

Anyway perhaps I am old fashioned, happy to change if people prefer.

I think it needs to be changeable with -D because you cant know what value
people need when generating csg and echo files. In my case I would want it
true but it's a bit odd when no preview is being generated for it to
automatically be true and other people might need it to be false depending
on what they want the csg to represent.

As for more $variables I just don't see the problem. They are simple,
natural and powerful. A function taking a string to select unrelated return
values is ugly IMHO. Something that looks like an associative array is ugly
as well and opens a big can of worms because the language doesn't have
them. Neither can be overridden for a subtree or from the command line
either.

doug-moen:
I agree it should be a $ variable, for consistency with the existing
pattern (all of the other "system variables" begin with "$").

I think there needs to be a more convenient way to convert a boolean to 0
or 1 in an arithmetic context. Every language provides this. I personally
use a function called `bit` for this:
   function bit(b) = b ? 1 : 0;
I think a basic function like this should be built in. Either that, or the
boolean values should be automatically converted to 0 or 1 in an arithmetic
context, as happens in Python and Javascript. Of course, we should not be
trying to fix this problem in your PR request.

kintel:
I'm a bit uneasy about introducing more special variables. However, realistically speaking, we're not likely to get a lot more resources to be able to design and implement better ways of solving this.
I'll add a few TODO items and mull it over for a bit.

nophead:
> I'm a bit uneasy about introducing more special variables. 

Nobody has said what problem more of them causes. The fact they are prefixed with $ means they are very unlikely to clash with user variables and even if they do the user definition would override.

They are looked up separately, so don't slow down normal variable access. They are stored in an unordered map meaning search has average constant time complexity, so I don't think speed is an issue.

In a scripting language that aims to be simple, familiar and flexible they seem to be appropriate. They behave just like OS environment variables in shell scripts, including the dynamic scope. Not OO, but then OpenSCAD isn't an OO language. Grouping them into a system object is also artificial.

nophead:
Regarding testing command line, source code and dynamic overrides: as these are only set in the top level context they behave just the same as the other $ variables, so all these things work as expected.

nophead:
As an aside I notice that you put even trivial functions in the .cc file and only include prototypes in the .h file, for example the Value class constructors. When I was writing C++ for embedded platforms I always put one line functions in the .h so that they got inlined and that gave a massive speed increase and usually small code size reduction as well. Is this still the case, or are modern compilers able to inline across files?

doug-moen:
"are modern compilers able to inline across files?"

Yes. Gcc has a -flto flag for this. But it's not enabled automatically, eg
it's not part of -O3. The documentation for it is quite extensive:
apparently, you need to know what you are doing before you enable it. I've
never worked on a project that uses it.

nophead:
Seems like it is still a good idea to put simple functions in the class declaration then. Things like constructors and accessors. Anything that is likely to be shorter than a function call.  Inlining makes a huge difference with value type classes.

kintel:
In some cases, hiding the implementation in the .cc files can yield significant compile-time optimizations. This is especially the case with a bit (too) tightly coupled codebases like OpenSCAD's when there are frequent changes in lower-level components.
It should be a lot more stable now, so we can probably move more of the implementations back into the header files where that feels reasonable. We'd be happy to accept separate PR's for such green refactoring.

nophead:
>can yield significant compile-time optimizations

Do you mean reduction in compile time or code optimisations performed at compile time. I can't see how hiding information from the compiler gets better code but I can see it will speed up compilation itself.

I am amazed at how slowly MingW compiles OpenSCAD. I was compiling much bigger projects with hundreds of source files much faster 15 years ago. That was with gcc and Visual Studio on PCs not as powerful as I have now. 

However, the source files were generated by a UML tool I wrote, so they had exactly the correct minimum include files in the .h and .cpp files to satisfy the dependencies that were visible in the class diagrams.

kintel:
Reduction in compile time - typically when recompiling after changing the implementation of a common component.

kintel:
@nophead I'm uneasy about more special variables as this is not a very thought-through feature. I think I'll merge it though. It's useful and has little damage potential :)
Would you mind adding a a test cases, to test that $preview is set correctly for the main situations (preview/throwntogether vs. render vs. csg output) ?

(If you have issues getting the test framework set up, that would be great to get that sorted out as it's a core part of how we work and having contributors write and run tests significantly reduces our workload.)

nophead:
I will have ago but I currently don't know anything about the test framework. Is there any documentation?

kintel:
@nophead Docs are not perfect: https://github.com/openscad/openscad/blob/master/doc/testing.txt
If we could extend this with how to do in using the MSYS2 build that would be cool.

nophead:
So I installed cmake, python2, ImageMagick and diff under MSYS2 using pacman. When I run cmake . I get 
```
$ cmake .
-- Building for: NMake Makefiles
-- The C compiler identification is unknown
-- The CXX compiler identification is unknown
CMake Error at CMakeLists.txt:44 (project):
  The CMAKE_C_COMPILER:

    cl

  is not a full path and was not found in the PATH.

  To use the NMake generator with Visual C++, cmake must be run from a shell
  that can use the compiler cl from the command line.  This environment is
  unable to invoke the cl compiler.  To fix this problem, run cmake from the
  Visual Studio Command Prompt (vcvarsall.bat).

  Tell CMake where to find the compiler by setting either the environment
  variable "CC" or the CMake cache entry CMAKE_C_COMPILER to the full path to
  the compiler, or to the compiler name if it is in the PATH.


CMake Error at CMakeLists.txt:44 (project):
  The CMAKE_CXX_COMPILER:

    cl

  is not a full path and was not found in the PATH.

  To use the NMake generator with Visual C++, cmake must be run from a shell
  that can use the compiler cl from the command line.  This environment is
  unable to invoke the cl compiler.  To fix this problem, run cmake from the
  Visual Studio Command Prompt (vcvarsall.bat).

  Tell CMake where to find the compiler by setting either the environment
  variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
  to the compiler, or to the compiler name if it is in the PATH.


-- Configuring incomplete, errors occurred!
See also "C:/msys64/home/ChrisP/openscad/tests/CMakeFiles/CMakeOutput.log".
See also "C:/msys64/home/ChrisP/openscad/tests/CMakeFiles/CMakeError.log".

```

So it looks like cmake is deciding to make a native Windows version of the tests to run under Visual Studio. Am I correct in thinking I need to fool cmake to think it is making a Linux version?

kintel:
It looks like cmake tries to find the Visual Studio compiler, while we want it to use the mingw compiler. I would think this is a pretty common issue. I guess setting the `CMAKE_CXX_COMPILER` variable to the correct compiler executable would be a good start.

kintel:
..or perhaps `cmake -G "MSYS Makefiles" .`.
See https://stackoverflow.com/questions/37365355/cmake-msys-makefiles-generator-missing

nophead:
Thanks. I got cmake to run to completion. 
mingw32-make.exe compiles lots of files but falls over as follows:
```
/openscad/src/CGALRenderer.cc.obj -c /C/msys64/home/ChrisP/openscad/src/CGALRenderer.cc
In file included from C:/msys64/mingw64/x86_64-w64-mingw32/include/windows.h:71:0,
                 from C:/msys64/home/ChrisP/openscad/src/OGL_helper.h:47,
                 from C:/msys64/home/ChrisP/openscad/src/CGAL_OGL_Polyhedron.h:33,
                 from C:/msys64/home/ChrisP/openscad/src/CGALRenderer.cc:39:
C:/msys64/mingw64/x86_64-w64-mingw32/include/wingdi.h:3353:68: error: 'WINBOOL Polygon(HDC, const POINT*, int)' redeclared as different kind of symbol
   WINGDIAPI WINBOOL WINAPI Polygon(HDC hdc,CONST POINT *apt,int cpt);
                                                                    ^
In file included from C:/msys64/home/ChrisP/openscad/src/polyset.h:6:0,
                 from C:/msys64/home/ChrisP/openscad/src/CGALRenderer.cc:34:
C:/msys64/home/ChrisP/openscad/src/GeometryUtils.h:6:31: note: previous declaration 'typedef class std::vector<Eigen::Matrix<double, 3, 1> > Polygon'
 typedef std::vector<Vector3d> Polygon;
                               ^~~~~~~
mingw32-make[2]: *** [CMakeFiles/tests-offscreen.dir/build.make:255: CMakeFiles/tests-offscreen.dir/C_/msys64/home/ChrisP/openscad/src/CGALRenderer.cc.obj] Error 1
mingw32-make[1]: *** [CMakeFiles/Makefile2:107: CMakeFiles/tests-offscreen.dir/all] Error 2
mingw32-make: *** [Makefile:95: all] Error 2
```

It seems that Polygon is defined as a function in windows.h and that clashes with a type definition in GeometryUtils.h. I am wondering how CGALRenderer.cc manages to compile in the OpenSCAD build but not in the tests.

nophead:
Turns out I need to add -DNOGDI to the command line.

kintel:
That sounds reasonable. I guess that proves that you're the first one to build the test suite on Windows native :)

nophead:
It seems so as I also had to add -D_USE_MATH_DEFINES to get M_PI and -DUNICODE to get wide characters. 

The tests all compile and run now. There are 37 failures though but they seem to be only three or four causes. The first one I am looking at is the red outline around complex 2D objects like fonts and the tree drawn by recursion goes wrong. I haven't noticed it with any of my own 2D objects but it certainly doesn't work even with pre-compiled Windows snapshots. The extra lies jump around when the image is rotated.

module_recursion

nophead:
It seems to go wrong when it tries to draw a polygon with more than 1200 vertices in F6 or thrown together but not F5.
```
circle(r = 50, $fn = 1201);
```
Goes wrong but two circles with $fn=1200 are OK.

kintel:
Cool!

Your screenshot definitely looks like a bug, but let's deal with the test failures separately from this issue. We already have a few test failures which trigger on certain Linux distributions that we had to disable as we couldn't find the cause of the errors.

nophead:
Do you want me to open a separate issue for each type of test failure or a general one for failures on Windows?

kintel:
I think separate ones will make them easier to fix.

nophead:
It seems to be an Intel 4000 graphics driver bug because it works on two Win7 laptops I tried. There is mention of it here: https://www.opengl.org/discussion_boards/showthread.php/182428-GL_LINE_LOOP-drawing-incorrectly

nophead:
It seems to be an Intel 4000 graphics driver bug because it works on two Win7 laptops I tried. There is mention of it here: https://www.opengl.org/discussion_boards/showthread.php/182428-GL_LINE_LOOP-drawing-incorrectly

nophead:
I updated my graphics driver and that fixed the 2D outline problem.

nophead:
Can anybody explain why the OSX tests passed on the commit where I had forgotten the check in the test results? The other two builds failed all the new tests as expected.

Ironically the OSX failed when I did check them in because it timed out. So it gives exactly the opposite result to the other two builds.

kintel:
@nophead I think this is an artifact of how Travis executes pull requests; they simply clone the HEAD of your branch. If, for some reason, one of their runners is slow to start, different runners may clone different versions of your branch if you pushed new commits during this period.
The Mac runners on Travis CI are particularly slow to start, probably due to lots of traffic on limited hardware.

nophead:
Yes I think I did push the new commit while Travis was running but it labels the results with the commit number, so it shouldn't grab the head as it is totally misleading.

Anyway it shows that tests pass on all three builds when they don't time out so can this be merged now?

kintel:
Thanks for hanging in there and getting the tests to run on Windows!

nophead:
Thanks. I guess I should claim my own bounty now.

PR #1967, opened 2017-03-08
Editor enhancements, first commit

buttim:


kintel:
@t-paul How similar is this to the auto-complete work you did a while back?

src/scintillaeditor.cpp
```diff
@@ -644,6 +742,14 @@ bool ScintillaEditor::eventFilter(QObject* obj, QEvent *e)
 		 || e->type()==QEvent::KeyRelease
 		 ) {
 		QKeyEvent *ke = static_cast<QKeyEvent*>(e);
+		//printf("KEYPRESS. MODIFIERS: %08x\n",ke->modifiers());
+#ifdef _WIN32
+		if ((ke->modifiers() & ~Qt::KeypadModifier) == Qt::CtrlModifier|Qt::AltModifier) {
+			QKeyEvent* event = new QKeyEvent(ke->type(), Qt::Meta, ke->text());
```
t-paul: What's this change doing? It does not compile on MXE due to both Qt::CtrlModifier and Qt::Meta not being defined.

buttim: I think this could be an issue with Qt5. It did compile on Qt4. It tries to resolve a conflict on Windows with italian keyboards. To close a curly bracket you have to press Shift AltGr and a 4 symbols key (+ ] * }). The hotkey detection of Qt triggers the menu voice I associated with Edit/Toggle all folds (Ctrl+Shift+*). That's because AltGr gets somehow translated to AltGr or right Alt. Not even sure the snippet did in effect solve the problem. We might simply choose another hotkey but that would be a nuisance for people using SciTE (from wich the hotkey is borrowed)

kintel:

Issue #1815, opened 2016-10-02
Warn or produce error if multiple Root Modifiers in source


thehans:
Often times during development and debugging of a script I will use the root modifier "!" to single out a component.  Sometimes I add another "!" to a different component and forget to remove the last one.  In the case of multiple root modifiers, the behavior is undefined.

It would be nice if OpenSCAD would warn the user if it finds more than one of the "!" modifiers and points the user to the lines so they can be found quicker.

kintel:
Perfect task for new contributors :)

nophead:
It would be more useful if it rendered all of them if there is more than one. It would be obvious if one was left in by mistake but it also allows more control over what is displayed.

amarjeetkapoor1:
@kintel 
I gave it a shot as I want to understand more about OpenSCAD language.
Just added following at https://github.com/openscad/openscad/blob/master/src/localscope.cc#L49
 Thinking that this would output desired result but I didn't 
Here isRootModifier takes count of children having 
```
if(modinst->tag_root && isRootModifier>1){
    PRINTB("WARNING: Root Modifier Added At Line%d", modinst->location().firstLine());
}
```
Other option that I thought and It worked is add following in parser. It worked. If its right way then I would send PR.
 
```
module_instantiation:
          '!' module_instantiation
            {
                $$ = $2;
                if ($$) $$->tag_root = true;
                if(isRootModifier){
                    PRINTB("WARNING: Root Modifier Added At Line%d",LOC(@$).firstLine());
                }
                isRootModifier=true;
            }
```

kintel:
@amarjeetkapoor1 Look here: https://github.com/openscad/openscad/blob/master/src/node.cc#L94

That's the code responsible for finding the first root-tagged node.
I think searching for more root nodes would belong there.

@nophead's suggestion also sounds interesting, although that's a bit more work as we'd have to create a new temporary container for the tagged nodes.

We could start by showing a warning and perhaps extend it to show all root nodes later. One idea would be for ```find_root_tag()``` to return a list of nodes and display the warning if the list length is >1.

t-paul:
The warning is probably still needed even with multiple roots for cases where one of the root nodes is a sub-tree of another, e.g.
```
!difference() {
    !cube();
    sphere();
}
```
We could still show both, but I'm not sure that's a good idea.

MichaelAtOz:
I suggest the message shows the `!` ie
```
"WARNING: Root Modifier (!) Added At Line%d"
[...]

PR #1981, opened 2017-03-31
Print warning  with multiple roots

amarjeetkapoor1:
In reference to issue #1815.  

It prints warning for all the root modifiers added in the program with line number.

### DIFF
--- a/src/node.cc
+++ b/src/node.cc
@@ -29,6 +29,7 @@
 #include "ModuleInstantiation.h"
 #include "progress.h"
 #include "stl-utils.h"
+#include "printutils.h"
 
 #include <iostream>
 #include <algorithm>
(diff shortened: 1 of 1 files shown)


kintel:
We should probably only print the warning if there are more than _one_ root modifier.

src/node.cc
```diff
@@ -93,10 +94,26 @@ std::ostream &operator<<(std::ostream &stream, const AbstractNode &node)
...
 {
-  for(auto v : n->children) {
-    if (v->modinst->tag_root) return v;
-    if (auto vroot = find_root_tag(v)) return vroot;
-  }
-  return NULL;
-}
+	std::vector<AbstractNode*> rootTags;
+
+	std::function <void (AbstractNode *n)> find_root_tags = [&] (AbstractNode *n)
```
kintel: You should be able to use `auto` here instead of declaring it using std::function

amarjeetkapoor1: Okay.

By the way, I was thinking of restricting the capture list in which case auto will not work.

```
 std::function <void (AbstractNode *n)> find_root_tags =  find_root_tags = [&rootTags, &find_root_tags] (AbstractNode *n)

```

amarjeetkapoor1: Or Is it okay like this only?

kintel: Sorry, you're right. auto won't work as this is a recursive function. I'll merge it as is.

Issue #901, opened 2014-08-05
linear_extrude() with slices should short-circuit the faceting of twisted extrusions when facets between slices align


hyperair:
Sometime between 7952d24 and fde73a4, the faceting behaviour of twisted linear extrusions appear to have changed. When rendering https://github.com/hyperair/MCAD/blob/eeff4ee/fasteners/threads.scad, which attempts to calculate a suitable $fa that will generate a round number of slices to align the facets between each slice, 7952d24 (2014.05) renders them with very clean, straight facets, while fde73a4 (2014.08) renders them with a huge amount of facets.

Here is a screenshot of how it looked with the previous version...
![screenshot5](https://cloud.githubusercontent.com/assets/88355/3806491/9822fa28-1c4d-11e4-9265-73242e990464.png)

and here's how it looks with the current version:
![screenshot8](https://cloud.githubusercontent.com/assets/88355/3806486/7c8a6ada-1c4d-11e4-8b1a-019b533fb33c.png)

A zoomed in view:
![screenshot7](https://cloud.githubusercontent.com/assets/88355/3806493/9ebc4272-1c4d-11e4-81d3-07a7a52b62ad.png)
##

vector76:
I know I'm late to the party but I agree with TakeItAndRun that the subdivision of the non-planar quadrilaterals is occurring along the longer diagonal when it might make sense for it to occur along the shorter diagonal.

This is all that's needed to exhibit the behavior:
`linear_extrude(height=5, twist=30, slices=2) circle(r=10, $fa=30);`

![linear_rotate_tesselation](https://cloud.githubusercontent.com/assets/955138/14269561/616905fc-faab-11e5-872b-6687ffa0a0cf.png)

It appears from the code (within add_slice()) that "splitfirst" controls which of the two subdivisions of the parallelogram occurs.  Either (prev1, curr2, curr1) and (curr2, prev1, prev2), or (prev1, prev2, curr1) and (prev2, curr2, and curr1).

[...]

PR #1987, opened 2017-04-09
Add circulate= to linear_extrude for better screw threads

jepler:
I have been considering the problem reported in issue #901 for some time.  This PR adds a parameter `circulate=` to `linear_extrude`, which is the original request in that issue.  As you can see in the attached image, this improves the appearance of a screw thread.

This patch probably needs documentation and an image test before it's suitable for inclusion, not to mention healthy discussion of whether it can even be used safely in enough circumstances to warrant it.  However, it does improve some screw threads quite a bit!

xtry

The (fictional) thread was generated by this openscad program:
````
function frompolar(r, th) = [ cos(th) * r, sin(th) * r ];
function r(th) = 
    let(th = th % 60)
    (th < 10) ? 18 + th / 4 :
    (th < 20) ? 20.5 :
    (th < 30) ? 18 + (30 - th) / 4 :
    18;

module shape() {
    polygon([for(th=[0:1:360]) frompolar(r(th), th)]);
}

TH = 1440;
SL = 1 + TH;
HH = 144;
rotate([0,90,0])
translate([-24,0,0]) 
linear_extrude(twist=TH, height=HH, slices=SL, center=true) shape();
rotate([0,90,0])
translate([ 24,0,0]) 
linear_extrude(circulate=1, twist=TH, height=HH, slices=SL, center=true) shape();
````

### DIFF
--- a/src/GeometryEvaluator.cc
+++ b/src/GeometryEvaluator.cc
@@ -624,11 +624,23 @@ static void translate_PolySet(PolySet &ps, const Vector3d &translation)
 	}
 }
 
+namespace
+{
+template<class T>
+const typename T::value_type &modulo_index(const T &container, typename T::size_type idx) {
+	ssize_t size = container.size();
+	idx %= size;
+	if(idx < 0) idx += size;
+	return container.at(idx);
+}
+}
+
 static void add_slice(PolySet *ps, const Polygon2d &poly, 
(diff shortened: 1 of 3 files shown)


jepler:
I broke something in this while cleaning up before making the PR.  I'll resubmit when I figure out what it was.  Thanks for your time.

PR #1973, opened 2017-03-27
Extrude and minor tweaks

mhermier:
Hi,
I would like to propose these some small changes set.
There are minor changes to offer some tools to simplify some printing code, and a big change that add a generalized extrude. The last one is based on rotate_extrude, but use a list of transformation matrix to describe the extrusion path. Since it is a general extrude, I think it really should be named 'extrude'.
It is not a perfect tool, and might generate some broken volumes. But I consider it an advanced command like 'polygon' or 'polyhedron' which can also generate bad volumes. So it should not be a big deal.

PR #1989, opened 2017-04-10
 Add circulate= to linear_extrude for better screw threads

jepler:
I have been considering the problem reported in issue #901 for some time.  This PR adds a parameter `circulate=` to `linear_extrude`, which is the original request in that issue.  As you can see in the attached image, this improves the appearance of a screw thread.

This patch probably needs documentation and an image test before it's suitable for inclusion, not to mention healthy discussion of whether it can even be used safely in enough circumstances to warrant it.  However, it does improve some screw threads quite a bit!

The (fictional) thread below was generated by this openscad program:
````
function frompolar(r, th) = [ cos(th) * r, sin(th) * r ];
function r(th) = 
    let(th = th % 120)
    (th < 32) ? 18 + th / 8 :
    (th < 64) ? 22 :
    (th < 96) ? 18 + (96 - th) / 8 :
    18;

function points() = [for(th=[0:8:359]) frompolar(r(th), th)];

module shape() {
    polygon(points());
}

TH = 720;
SL = 1 + TH/8;
HH = 96;

translate([-24,0,0]) 
linear_extrude(twist=TH, height=HH, slices=SL, center=true) shape();

translate([ 24,0,0]) 
linear_extrude(circulate=1, twist=TH, height=HH, slices=SL, center=true) shape();

translate([ 72,0,0]) 
linear_extrude(circulate=-1, twist=-TH, height=HH, slices=SL, center=true) shape();
````

The leftmost solid is the "original" linear_extrude, showing the issue #901 problem clearly; the middle duplicates it with an appropriate `circulate` parameter, and the third shows that `circulate` can be used for both signs of `twist`.

xtry

### DIFF
--- a/src/GeometryEvaluator.cc
+++ b/src/GeometryEvaluator.cc
@@ -624,11 +624,23 @@ static void translate_PolySet(PolySet &ps, const Vector3d &translation)
 	}
 }
 
+namespace
+{
+template<class T>
+const typename T::value_type &modulo_index(const T &container, typename T::size_type idx) {
+	ssize_t size = container.size();
+	idx %= size;
+	if(idx < 0) idx += size;
+	return container.at(idx);
+}
+}
+
 static void add_slice(PolySet *ps, const Polygon2d &poly, 
 											double rot1, double rot2, 
 											double h1, double h2, 
 											const Vector2d &scale1,
-											const Vector2d &scale2)
+											const Vector2d &scale2,
+											long int circulate1, long int circulate2)
 {
 	Eigen::Affine2d trans1(Eigen::Scaling(scale1) * Eigen::Rotation2D<double>(-rot1*M_PI/180));
 	Eigen::Affine2d trans2(Eigen::Scaling(scale2) * Eigen::Rotation2D<double>(-rot2*M_PI/180));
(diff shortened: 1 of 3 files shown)


jepler:
.. the first commit in this PR is not critical, it is more relevant to investigating other splitting strategies.  As such, it could be dropped without consequence beyond fixing up the second commit to apply again.

mhermier:
It sounds to me that to fix a behavior which is just an artifact of tool is used to generate the form. And I would not expect anything else from the linear_extrude command than the first form.
That said I'm really not against a "simplify" tool, that would try to reduce the triangle/polygon count trying to produce second form from first form, when invoked in a way like:

`simplify ($epsilon = 0) linear_extrude(twist=TH, height=HH, slices=SL, center=true) shape();`

Anyway to me "linear_extrude" is in all the case the wrong tool to generate threads. I mean you need to do *excessive* math to generate the correct plane for linear_extrude, while it is easier to describe the thread form in the perpendicular plane. This is also why I made https://github.com/openscad/openscad/pull/1973 . With it, one can create custom helix/thread commands on top of it and easily produce lighter thread complex forms, like the type of approximation you want to achieve with your change.

PR #1992, opened 2017-04-14
Handle unclosed start conditions. Fixes #1890

### LINKED ISSUE #1890: Critical error for unclosed block comment
TODO:
* [x] Make sure customizer doesn't crash
* [ ] Make the OpenSCAD output a parser error for unclosed `/*`
* [ ] Make the OpenSCAD output a parser error for unclosed `"`

When this code is previewed:
`/*`

OpenSCAD pops up this alert:
screenshot from 2016-12-11 19 38 40

The application no longer previews or renders after fixing the error and adding geometry.

Ideally this error would be notified through the console, and handled such that OpenSCAD would continue to function.

kintel:
TODO:

* [x] Add tests
* [x] Test with customizer

### DIFF
--- a/src/lexer.l
+++ b/src/lexer.l
@@ -153,6 +153,43 @@ use[ \t\r\n]*"<"	{ BEGIN(cond_use); }
 <<EOF>>         { parsererror("Unterminated use statement"); yyterminate(); }
 }
 
+\"			{ BEGIN(cond_string); stringcontents.clear(); }
+<cond_string>{
+\\n			{ stringcontents += '\n'; }
+\\t			{ stringcontents += '\t'; }
+\\r			{ stringcontents += '\r'; }
+\\\\			{ stringcontents += '\\'; }
+\\\"			{ stringcontents += '"'; }
+{UNICODE}               { parser_error_pos -= strlen(lexertext) - 1; stringcontents += lexertext; }
+\\x[0-7]{H}             { unsigned long i = strtoul(lexertext + 2, NULL, 16); stringcontents += (i == 0 ? ' ' : (unsigned char)(i & 0xff)); }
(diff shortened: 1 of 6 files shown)


amarjeetkapoor1:
It working for use and include but not for `"` and I even tested after replicating the code for `/* ` which was missing but it didn't worked for both cases.

Plus I also tried by changing the position of line in the given block.

Is it working on your system?

kintel:
I did some light testing and it looked ok. If you find anything that doesn't work, please here so we can add it to test cases.

amarjeetkapoor1:
Simple test case for which it didn't work:

Test Case 1:

`"`

Test Case 2:
`/*`

amarjeetkapoor1:
I have made some changes and now it's working.

Earlier this was causing the problem and the unclosed comments and string were not getting detected.
`
src/lexer.l:222: multiple <<EOF>> rules for start condition cond_string
src/lexer.l:233: multiple <<EOF>> rules for start condition cond_lcomment
`
But still unable to get why was this message was getting displayed.

kintel:
@amarjeetkapoor1 I've added test cases. Could you check that the customizer behaves correctly now, as described in #1890?

amarjeetkapoor1:
Changes related to customizer have been made before this. To counter this problem. So, customizer was behaving correctly and is behaving correctly.

Just one more thing. Does it seems okay to have two error messages?

```
ERROR: Parser error in file "/home/amarjeet/mountPoint/projects/openscad/examples/Parametric/sign.scad", line 58: Unterminated string 
ERROR: Parser error in file "/home/amarjeet/mountPoint/projects/openscad/examples/Parametric/sign.scad", line 58: syntax error 
ERROR: Compilation failed! 
```

kintel:
Regarding the two parser errors: That's suboptimal. I couldn't figure out how to suppress the generic "syntax" error. Any ideas?

amarjeetkapoor1:
No idea at present.

Issue #910, opened 2014-08-17
OFF export issues


kintel:
The current OFF export requires us to create a CGAL Polyhedron. This causes some corner cases to fail exporting to OFF which works for STL (https://github.com/openscad/openscad/blob/master/src/export.cc#L266)
- [x] Implement direct export of PolySet (#3883)
- [x] Enable conversion from Nef polyhedron to PolySet (see #337)
##

Issue #909, opened 2014-08-17
AMF export issues


kintel:
The current AMF export forces a creation of a Nef polyhedron even though that's unnecessary (https://github.com/openscad/openscad/blob/master/src/export.cc#L292).
- [ ] Implement direct export of PolySet
- [ ] Enable conversion from Nef polyhedron to PolySet (see #337)
##

doug-moen:
I gather that the problem with creating a Nef polyhedron is that we get an error message if the mesh is not 2-manifold. Issue #910 says "The current OFF export requires us to create a CGAL Polyhedron. This causes some corner cases to fail exporting to OFF which works for STL ", and #1621 says that AMF export is less reliable.

I am wondering if these corner cases that fail are the same as the corner cases which are not permitted by the AMF standard. If so, then maybe it isn't a bad thing if we refuse to create an invalid AMF file. Here's what the standard says:

6.6 Restrictions on Geometry—All geometry shall comply with the following restrictions:

1. Every triangle shall have exactly three different vertices.
2. Triangles may not intersect or overlap except at their common edges or common vertices.
3. Volumes shall enclose a closed space with nonzero volume.
4. Volumes may not overlap.
5. Every vertex shall be referenced by at least three triangles.
6. Every pair of vertices shall be referenced by zero or two triangles per volume.
7. No two vertices can have identical coordinates.
8. The outward direction of triangles that share an edge in the same volume must be consistent.

#1621, first comment, gives a model that fails at AMF export, but this model has self-intersection (restriction 2 above).

nophead:
Those rules should be applied to STLs as well as AMF because although STL
does not have those rules it should do. They don't look hard to code either
so could be applied to polyset.

kintel:
AMF does indeed have a strict non-manifold requirement, so it would make sense to respect those.
STL, otoh, doesn't, so such checks should be optional.
I'm believe that e.g. 3MF specifically allows such vertex or edge self-intersections, so what we should allow really depends on the downstream processes.

As mentioned in #1621, performing these checks ourselves is probably a better solution rather than relying on partially available functionality in CGAL.

kintel:
Closing as AMF export is deprecated

PR #1214, opened 2015-02-12
Don't merge: Testing SVG icon support

kintel:
@t-paul Could you give this a spin? Just testing if SVG icons will work as that might fix some resolution issues. Works on Mac.

PR #2002, opened 2017-04-25
Add wavefront obj and x3d export

tu500:
TODO:
* [ ] Code review
* [ ] Add a few test cases
* [ ] This PR exhibits the same export issues as AMF and OFF. If merged, we should create new issues tracking this for OBJ and X3D as well, referencing the two mentioned issues: #909 and #910


kintel:
@donbright Didn't you explore OBJ export/import a while back? Could you take a quick look at how this compares to your explorations?

Info.plist
```diff
@@ -93,6 +93,26 @@
...
 		</dict>
+		<dict>
+			<key>CFBundleTypeExtensions</key>
+			<array>
+				<string>x3d</string>
+			</array>
+			<key>CFBundleTypeName</key>
+			<string>X3D file</string>
+			<key>CFBundleTypeRole</key>
+			<string>Viewer</string>
```
kintel: We cannot import X3D files, so this entire block should be removed. Same for OBJ.

src/MainWindow.ui
```diff
@@ -1288,6 +1290,24 @@
     <string>Export as &amp;AMF...</string>
    </property>
   </action>
+  <action name="fileActionExportX3D">
+   <property name="icon">
+    <iconset resource="../openscad.qrc">
+     <normaloff>:/images/X3D.png</normaloff>:/images/X3D.png</iconset>
```
kintel: This image resource is missing. Same for `OBJ.png`

src/export_indexed_triangles.cc
```diff
@@ -0,0 +1,128 @@
+/*
+ *	OpenSCAD (www.openscad.org)
+ *	Copyright (C) 2009-2011 Clifford Wolf <clifford@clifford.at> and
+ *													Marius Kintel <marius@kintel.net>
```
kintel: Indentation glitch

src/export_indexed_triangles.cc
```diff
@@ -0,0 +1,128 @@
...
+export_indexed_triangles(const PolySet &ps)
+{
+	PolySet triangulated(3);
+	PolysetUtils::tessellate_faces(ps, triangulated);
+
+	std::vector<export_vec3> vecs;
+	std::vector<export_indexed_triangle> tris;
+
+	for(const auto &p : triangulated.polygons) {
+		assert(p.size() == 3); // STL only allows triangles
```
kintel: Comment is misleading

src/export_indexed_triangles.h
```diff
@@ -0,0 +1,35 @@
+#pragma once
+
+#include <vector>
+#include <utility>
+
+#include "polyset.h"
+
+
+struct export_vec3 {
```
kintel: Why not use `Eigen::Vector3d` as we do other places, see `linalg.h`

src/export_indexed_triangles.h
```diff
@@ -0,0 +1,35 @@
...
+	double x, y, z;
+
+	bool operator==(const export_vec3 &rhs) const {
+		return x == rhs.x && y == rhs.y && z == rhs.z;
+	}
+};
+
+struct export_indexed_triangle {
```
kintel: We could probably use `IndexedTriangle` here. See `GeometryUtils.h`

src/export_indexed_triangles.cc
```diff
@@ -0,0 +1,128 @@
...
+#include "polyset.h"
+#include "polyset-utils.h"
+
+#ifdef ENABLE_CGAL
+#include "CGAL_Nef_polyhedron.h"
+#include "cgal.h"
+#include "cgalutils.h"
+
+
+static size_t insert_vector(std::vector<export_vec3> &vecs, const export_vec3 &vec)
```
kintel: This is not very performant. A better idea could be to use `Reindexer` for this; see 'Reindexer.h'.
`Reindexer` is used multiple other places in the codebase.

src/export_obj.cc
```diff
@@ -0,0 +1,123 @@
...
+static void append_obj(const shared_ptr<const Geometry> &geom, std::ostream &output)
+{
+	if (const CGAL_Nef_polyhedron *N = dynamic_cast<const CGAL_Nef_polyhedron *>(geom.get())) {
+		append_obj(*N, output);
+	}
+	else if (const PolySet *ps = dynamic_cast<const PolySet *>(geom.get())) {
+		auto r = export_indexed_triangles(*ps);
+		append_obj(r.first, r.second, output);
+	}
+	// else if (const Polygon2d *poly = dynamic_cast<const Polygon2d *>(geom.get())) {
```
kintel: Any reasons why this was commented out? It looks resonable.

src/MainWindow.ui
tu500: Do you have a way to generate new icons, a script? Where do the existing come from?

src/export_obj.cc
tu500: Yes it is, I just got confused about the capabilities of obj files, sorry.

tu500:
Considering test cases, I don't know enough about the code base to generate good test cases and I didn't find any tests for the other exports. If output is deterministic one could do something like

```sh
echo 'cube(10);' > test.scad
openscad -o test.x3d test.scad
diff test.x3d expected.x3d
```

but that doesn't seem right.

Regarding the export issues you mentioned: I do have an overload for PolySets. Do you mean something else by that?

src/MainWindow.ui
kintel: They are hand-drawn. There are some SVGs buried in a PR somewhere (#1214). I don't think icons are necessary unless these icons make it into a toolbar.

kintel:
Tests: It's tricky to test without implementing the corresponding import functionality. I think our current tests rely on bidirectional export/import. A naive check assuming deterministic test output for a very simple testfile is better than no test at all. Note: All our tests are black-box regression tests, so they are pretty easy to implement. See here: https://github.com/openscad/openscad/blob/master/doc/testing.txt

kintel:
Export issues: The challenge is that these exports, as far as I can tell, convert directly from a Nef_3 polyhedron into a CGAL Polyhedron, then into the correct file format. This conversion fails in some corner cases. If you look at the e.g. STL export vs. AMF export, you can see that the code path is quite different. The mentioned issues are tracking this discrepancy. I'm not expecting you to fix that, but we should add issues so we know to deal with it later.

..of course, if you decide to fix it that would be awesome :)

tu500:
Ok, this might be getting off topic. As far as I understand it the STL code converts Nef polyhedrons to PolySets and exports those. There is also a code path which converts Nef polyhedrons to CGAL polyhedrons and exports those, but this one is unused. The AMF code converts PolySets to Nef polyhedrons, those to CGAL polyhedrons and then exports them.

The point I wanted to make was that I copied the STL workflow.

tu500:
Also, thanks for the review and especially for explicitly pointing to the right headers. I redid the code and addressed your comments. I'll look into test code and icons.

MichaelAtOz:
@tu500 What did you create the icons with?

tu500:
@MichaelAtOz As @kintel mentioned there is a PR with SVG icons #1214.

tu500:
I should mention I added these icons, together with a conversion script, into my svg-icons branch. @kintel If the icons are ok, maybe you want to merge those into the PR, for reference.

Also, the new icons look minimally different than the old ones. If this is a problem I would suggest recreating all icons from the svg version.

src/export_x3d.cc
```diff
@@ -0,0 +1,150 @@
...
+static void append_x3d(const shared_ptr<const Geometry> &geom, std::ostream &output)
+{
+	if (const CGAL_Nef_polyhedron *N = dynamic_cast<const CGAL_Nef_polyhedron *>(geom.get())) {
+		append_x3d(*N, output);
+	}
+	else if (const PolySet *ps = dynamic_cast<const PolySet *>(geom.get())) {
+		auto r = export_indexed_triangles(*ps);
+		append_x3d(r.first, r.second, output);
+	}
+	// else if (const Polygon2d *poly = dynamic_cast<const Polygon2d *>(geom.get())) {
```
kintel: This check should be re-enabled.

src/export_obj.cc
kintel: I think this should simply be re-enabled then.

kintel:
@donbright You worked on smth. similar a while back. I feels like this could be merged early and deal with import as a separate task. Opinions?

kintel:
@tu500 ..and there's a merge conflict with `openscad.cc`..

jgwinner:
I found a need for OBJ or GLTF export, and found this open ticket ...

Looks like it was a conflicted merge? What's the  status?

I was trying to build a parametric object that could be exported to GLTF or OBJ, for inclusion in a Mixed Reality Extension (Node.JS). 

        == John ==

t-paul:
I guess the current status would be abandoned, but it might not be too complicated to revive it. As listed in the task list, some test cases would be needed too (basically integration with the existing test framework).

kintel:
Closing as stalled. OBJ import/export has since been merged

PR #2003, opened 2017-04-25
Fixes to allow MSYS2 build of the test suite.

nophead:


tests/CMakeLists.txt
```diff
@@ -259,9 +263,9 @@ endif()
 #
 message(STATUS "Using C++11")
   if (NOT CMAKE_COMPILER_IS_GNUCXX OR GCC_VERSION VERSION_GREATER 4.6)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++11")
```
kintel: Is this necessary? We don't use GNU C++ extensions, so `-std=c++11` should be the correct flag.

nophead: gnu++11 is used when building OpenSCAD itself so I made the tests the same. It appears to enable the constants in math.h
```
#if !defined(__STRICT_ANSI__) || defined(_POSIX_C_SOURCE) || defined(_POSIX_SOURCE) || defined(_XOPEN_SOURCE) || defined(_GNU_SOURCE) || defined(_BSD_SOURCE) || defined(_USE_MATH_DEFINES)
#define M_E		2.7182818284590452354
#define M_LOG2E		1.4426950408889634074
#define M_LOG10E	0.43429448190325182765
#define M_LN2		0.69314718055994530942
#define M_LN10		2.30258509299404568402
#define M_PI		3.14159265358979323846
#define M_PI_2		1.57079632679489661923
#define M_PI_4		0.78539816339744830962
#define M_1_PI		0.31830988618379067154
#define M_2_PI		0.63661977236758134308
#define M_2_SQRTPI	1.12837916709551257390
#define M_SQRT2		1.41421356237309504880
#define M_SQRT1_2	0.70710678118654752440
#endif
```

Without it being set _USE_MATH_DEFINES needs to be set to get the constants. This was done in a couple of files only. OpenSCAD compiled because gnu++11 was set. The tests didn't on Windows.

I can change them both to c++11 and add -D_USE_MATH_DEFINES instead if you prefer.

kintel: Ah, qmake might insert something there which overrides our attempt to specify `-std=c++11`.
Let's leave it like you suggested.
Could you add a comment in the CMakeFiel that gnu++11 is required to get math constants defined, so we remember this in the future?

nophead: Yes can do. 

The odd thing is that with c++11 you only need D_USE_MATH_DEFINES on Windows. On Linux one of the other switches must turn the constants on. It doesn't make sense to me that math.h behaves differently on different systems when c++11 is selected.

kintel: It's because M_PI is defined by POSIX, and Windows isn't POSIX compliant.

PR #2004, opened 2017-04-25
Fixed version going wrong on Windows when the day is less than 10.

nophead:
Now strips leading zeros instead of appending .0 to prevent integers being interpreted as octal.

### DIFF
--- a/version.pri
+++ b/version.pri
@@ -19,10 +19,10 @@ isEmpty(VERSION) {
 
     QDATE = $$_DATE_
     QDATE_SPLIT = $$split(QDATE)
-    QDAY = $$member(QDATE_SPLIT,2)
+    QDAY = $$num_add($$member(QDATE_SPLIT,2))
     
     !isEmpty(TEST1) { 
-      contains( QDAY, $$DATE_Z ) { 
+      equals( QDAY, $$num_add($$DATE_Z) ) {
         # message("Assuming YYYY/MM/DD format")
         VERSION_YEAR = $$DATE_X 
         VERSION_MONTH = $$DATE_Y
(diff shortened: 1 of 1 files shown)


kintel:
@nophead I might have to revert this; `num_add` requires Qt-5.8.
A soft fix could be to require Qt-5.8 to build on MSYS2 for now.
Any ideas?

t-paul:
MXE also has Qt5.8, so if it just affects the Windows build, we might be ok. I'll kick a Windows snapshot build to see what happens...

nophead:
So is there any other way to convert a string to a number in qmake? What an awful language!

Which platforms don't have Qt5.8?

kintel:
Various Linux distros, MacPort and homebrew are not on Qt 5.8.
I didn't yet upgrade the Mac binaries, but that's fixable.

t-paul:
Getting a build error on MXE mentioning `enums.h`. Wasn't there a fix like `#undef DIFFERENCE` somewhere? Looks like that hits here too.

kintel:
I seem to remember that there's a better way of working around WIN32 global namespace pollution. I'll see what I can google up.

kintel:
I think `WIN32_LEAN_AND_MEAN` was what I was thinking about: See e.g. https://stackoverflow.com/questions/11040133/what-does-defining-win32-lean-and-mean-exclude-exactly

Before attempting to do that, it would be interesting to know where `winuser.h` is included. If we could isolate that, it would be a better design, IMO.

nophead:
@t-paul yes #1996 is needed to build on Windows since the prefixes were removed from the enums.

winuser.h is included in \msys64\mingw64\x86_64-w64-mingw32\include\windows.h
windows.h is in included in:
findversion.h
OffscreenContextWGL.cc
OGL_helper.h
PlatformUtils-win.cc
QGLView.cc

The file that fails to compile is mainwin.cc

nophead:
WIN32_LEAN_AND_MEAN does not prevent winuser.h from being included.

nophead:
@t-paul, sorry I meant  #2013, it is the antidote to #1996.

kintel:
@nophead Thanks for the info, let's continue the Windows discussion in #2013.

nophead:
Should be fixed by #2022 so that num_add is only needed on Windows.

PR #2013, opened 2017-04-28
Fix for #1996 DIFFERENCE clashes with winuser.h

nophead:
C++11 might prefer scoped enums but they don't play well with #define.

### DIFF
--- a/src/enums.h
+++ b/src/enums.h
@@ -1,4 +1,5 @@
 #pragma once
+#undef DIFFERENCE //#defined in winuser.h
 
 enum class OpenSCADOperator {
 	UNION, 


kintel:
Comment from #2004:

I seem to remember that there's a better way of working around WIN32 global namespace pollution. I'll see what I can google up:

I think WIN32_LEAN_AND_MEAN was what I was thinking about: See e.g. https://stackoverflow.com/questions/11040133/what-does-defining-win32-lean-and-mean-exclude-exactly

Before attempting to do that, it would be interesting to know where winuser.h is included. If we could isolate that, it would be a better design, IMO.

kintel:
Looking at the code, I cannot see where `windows.h` would be included from `mainwin.cc`, given the list of files your provided earlier.
What I'd like to do is to see if we can isolate this, as there should be no need to include `windows.h` from our own headers.

Btw., is the Windows build instructions on Wikibooks in a working state? (in case I find motivation to install Windows and give it a shot).

nophead:
The include route is via a generated file and then via QGLWidget.

In file included from C:/msys64/mingw64/x86_64-w64-mingw32/include/windows.h:72:0,
                 from C:/msys64/mingw64/include/QtCore/qt_windows.h:64,
                 from C:/msys64/mingw64/include/QtGui/qopengl.h:49,
                 from ../../../mingw64/include/QtOpenGL/qgl.h:45,
                 from ../../../mingw64/include/QtOpenGL/QGLWidget:1,
                 from src/QGLView.h:9,
                 from objects/ui_MainWindow.h:35,
                 from src/MainWindow.h:6,
                 from src/mainwin.cc:31:

So although there are several explicit includes of windows.h the one that breaks enums is implicit.

The MSYS2 Windows instructions work for OpenSCAD. Some notes are need adding for the test build.

kintel:
@nophead Unrelated to this particular issue, I noticed that it's including `QGLWidget` instead of `QOpenGLWidget`. `QOpenGLWidget` is the Qt5 replacement for `QGLWidget`, but needs to be enabled specifically due to some early compatibility issues, see https://github.com/openscad/openscad/blob/master/opengl.pri

Some testing on Windows would be cool if you feel like it.

nophead:
I will have a look later. What problems do I need to look out for?

Another way to fix this issue would be to not use block capitals for enum names. Then they won't clash with #defines that are usually upper case by convention.

nophead:
I have it running with QOpenGLWidget and haven't noticed anything being different.

nophead:
Apart from this in the terminal!
```
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
Error !!!!!!!!!!!!!!!!!!!!!!!
```

t-paul:
That's some lazy error/assertion handling inside CGAL.

nophead:
OK, I did have a bunch of CGAL errors when I accidentally pressed F6 on a full machine assembly.

kintel:
I think the real solution would be to hide the QGLWidget/QOpenGLWidget interface from our QGLView class, but that's a bit out of scope.

How about this middle ground solution?
* Add a file `undef_windows.h`, containing all the namespace pollution we want to undo for the whole project. Possibly just `DIFFERENCE` for now.
* Include this file inside an `#if defined(Q_OS_WIN)`, anywhere we include `windows.h` from a header file. Possibly just `QGLView.h` for now.

nophead:
It seems overly complex to me compared to an #undef DIFFERENCE or renaming the enum to Difference.

You can only fix name pollution by macros with an undef_windows.h. And I don't see how undef_widows.h is any better than my original fix that has a comment that it clashes with winuser.h. If you had a C symbol clash, which is probably likely, the you would need to encapsulate the windows / OpenGL dependency anyway.

Also DIFFERENCE is pretty much a special case as it is stuck in the middle of a bunch of macros with RT_ prefix and is used to make a couple of them. Most Windows macros are prefixed, so it looks like a mistake to expose DIFFERENCE because it seems to be just an internal value.

If you avoid block capitals then you are much less likely to clash with a macro and enums don't have all uppercase names by convention.

So what I am saying is this seems to be such a one off special case it doesn't need a more general solution.
The real issue is that there is no Travis CI build for Windows. If there was as soon as new symbol was defined the clash would be immediately obvious and could simply be avoided by using a different name.

kintel:
It's just a little bit more complex than undef'ing one symbol inline, but I still think it's a better solution:
* This is really a bug in Qt, exposing an underlying Win32 symbol
* ..which is a bug in Win32 polluting the global namespace with non-prefixed symbols

Win32 has numerous such issues, which we don't want to care about in client code. I'm sure other platforms has similar issues and as an app developer I don't really want to be aware of those. Putting it all in a place which attempts to repair Win32 bugs makes sense to me, and include those files as close to the offending code as possible.

If you find it to be unnecessarily work right now, I'll just merge the patch and clean up later.

nophead:
I think it isn't worth making a better solution unless the problem crops up again. This one was very bad luck with several bugs aligning. 

Can we add a Windows MXE build to Travis CI? Does it cost anything or is it completely free for OS projects?

t-paul:
Yes, having MXE on Travis too would be nice, but that needs quite some setup. We can't just build MXE as that takes way too long (total build on our server for both 32bit and 64bit is about 8 hours IIRC). It might be possible if we can deposit the build env somewhere, not sure how to configure the builders though.

kintel:
I'll merge this and leave further cleanup for later

kintel:
There might be free CI services who offer native Windows environments. With the MSYS2 build capable of running tests, this is suddenly an option.

PR #2021, opened 2017-05-01
Fixed a plethora of compiler warnings.

nophead:
Should be no actual changes.

A suspicious one is https://github.com/nophead/openscad/blob/85ac94a2f3e27f7936091dd0efba67a6098d5f76/src/libsvg/shape.cc#L28. Why have an empty copy constructor?

### DIFF
--- a/src/comment.cpp
+++ b/src/comment.cpp
@@ -303,15 +303,15 @@ void CommentParser::collectParameters(const char *fulltext, FileModule *root_mod
 		}
 
 		// Look for the group to which the given assignment belong
-		for (const auto &groupInfo : groupList) {
-			if (groupInfo.lineNo >= firstLine) {
-				//creating node for description
-				shared_ptr<Expression> expr(new Literal(ValuePtr(groupInfo.commentString)));
-				annotationList->push_back(Annotation("Group", expr));
-				break;
-			}
-		}
+		int i=0;
+		for (; (unsigned)i < groupList.size() && groupList[i].lineNo<firstLine;i++);
+		i--;
 
+		if (i >= 0) {
+			//creating node for description
+			shared_ptr<Expression> expr(new Literal(ValuePtr(groupList[i].commentString)));
+			annotationList->push_back(Annotation("Group", expr));
+		}
 		assignment.addAnnotations(annotationList);
 	}
 }
--- a/src/comment.cpp
+++ b/src/comment.cpp
@@ -304,7 +304,7 @@ void CommentParser::collectParameters(const char *fulltext, FileModule *root_mod
 
 		// Look for the group to which the given assignment belong
 		int i=0;
-		for (;i<groupList.size() && groupList[i].lineNo<firstLine;i++);
+		for (; (unsigned)i < groupList.size() && groupList[i].lineNo<firstLine;i++);
 		i--;
 
 		if (i >= 0) {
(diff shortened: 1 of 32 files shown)


kintel:
Thanks a lot! I'll have a look later.

kintel:
@t-paul Any comments on that `shape` copy constructor? I'm not sure what the roles of parent/children is after a file has been read..

src/value.cc
```diff
@@ -694,7 +694,7 @@ class bracket_visitor : public boost::static_visitor<Value>
     const auto i = convert_to_uint32(idx);
     if (i < str.size()) {
 			//Ensure character (not byte) index is inside the character/glyph array
-			if (i < g_utf8_strlen(str.c_str(), str.size()))	{
+			if ((int)i < g_utf8_strlen(str.c_str(), str.size()))	{
```
kintel: `g_utf8_strlen` is guaranteed to return a positive integer, so rather wrap that return value in a `static_cast`.

t-paul:
I don't think the shape copy constructors are used right now, so it's probably quite incomplete. The plan was to produce a data structure similar to the OpenSCAD AST which can be evaluated after reading the file.

src/comment.cpp
```diff
@@ -304,7 +304,7 @@ void CommentParser::collectParameters(const char *fulltext, FileModule *root_mod
 
 		// Look for the group to which the given assignment belong
 		int i=0;
-		for (;i<groupList.size() && groupList[i].lineNo<firstLine;i++);
+		for (; (unsigned)i < groupList.size() && groupList[i].lineNo<firstLine;i++);
```
nophead: I assume this is part of the customiser, which I have never used, so I don't really understand what this does. Your code looks more correct but I can't convince myself it does the same as the original because of the --i. So I think perhaps it either fixes a bug or creates a new one.

nophead: Indeed several customiser tests fail so it looks like this isn't equivalent.

amarjeetkapoor1: I think its failing just because of given statement
`if (groupInfo.lineNo >= firstLine) `

It should be:
`if (groupInfo.lineNo < firstLine) `

nophead: The original code finds the last group with lineNo less than firstLine. Marius' version finds the first group with lineNo >= firstLine, which is out by one. Your suggestion would find the first group with lineNo < firstLine.

To do it Marius style I think the loop needs to iterate backwards but I don't know how you do that with the C++00x version of for.

amarjeetkapoor1: Following code passed all the tests related to customzier:

`#include <boost/range/adaptor/reversed.hpp>`

``` 
for (const auto &groupInfo :boost::adaptors::reverse(groupList)){
            if (groupInfo.lineNo < firstLine) {
            //creating node for description
            shared_ptr<Expression> expr(new Literal(ValuePtr(groupInfo.commentString)));
            annotationList->push_back(Annotation("Group", expr));
            break;
        }
    }
```

kintel:
Let's leave the customizer code alone - this was to get rid of warnings after all. I guess that was a bit too proactive of a fix.

src/comment.cpp
kintel: @amarjeetkapoor1 You can apply this on your own branch if you want. We just merged this without as it was supposed to be some quick fixes to remove warning, not smth. that requires a conversation..

PR #2022, opened 2017-05-01
Now only uses num_add on Windows builds.

nophead:
Partially reverted #2004 so version numbers are back to floats to avoid the leading zeros.


kintel:
Should this simply revert the old commit or also introduce the check for Windows? I might be mislead by the commit message..

nophead:
This PR removes the three num_add calls that were outside the win32 switch. So on the the other platforms it will work with older qmake. It still needs QT5.8 for Windows builds.

No need to revert the old PR, just add this on top.

Issue #2024, opened 2017-05-03
STL files missing from depencency file when output is PNG


nophead:
handle_dep() is called from import_stl(). When a png file is being generated the dependency file is written before the STL file gets imported, so the entry goes missing.

include and use call handle_dep() from the parser, so get added much earlier. import isn't handled by the parser as it is just a module name. I am guessing moving the handle_dep() call to ImportModule::instantiate would fix the problem.

kintel:
In general, handle_dep() should probably move out of the parser. We'd need to refactor include and use into propers AST nodes. That's on my radar, just hasn't happened yet.

kintel:
..which means that the dependency step could be one single AST traversal.

nophead:
The dependency information for include and use is also tracked in FileModule, which leads me to another bug. Changes to imported files don't trigger a reload.

nophead:
Is there a way to get hold of the parent FileModule from ImportModule::instantiate?

If there is I could add a new list of import dependencies to FileModule so that changes update correctly and I could also then remove all the handle_dep calls and just traverse the FileModule dependency tree to create the deps file.

nophead:
Fixed by #2025

PR #2025, opened 2017-05-04
Fixed imports and surface files missing from dependencies.

nophead:
Fix for #2024.

Dependencies can now be written for all file types, e.g. echo because they are added at module instantiation time rather than when the file is read to create geometry.


nophead:
Looks like the precise build has been broken by a Travis update yesterday.

PR #2028, opened 2017-05-11
1917: QSettings is now cached

DrCodeWizzard:


DrCodeWizzard:
Just a heads up this code reduces the number of lstat operations to about 1800. I suspect this is due to sync on write of values. If do not do that and instead rely on QT periodically syncing (there is a background thread in charge of this), we can probably lower the number of accesses to ~100. Thoughts?

PR #2030, opened 2017-05-11
1917: QSettings is now cached

DrCodeWizzard:
QSettings is now cached. This reduces the number of file accesses that QSettings performs. Writes to persistent storage are done periodically by QSettings in background.

Note: This also intentionally disables the test run on Travis/Ubuntu 12.04.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,11 +8,6 @@ matrix:
       env: DIST="trusty"
       sudo: required
       cache: apt
-    - os: linux
-      compiler: gcc
-      dist: precise
-      env: DIST="precise"
-      cache: apt
     - os: osx
       compiler: clang
       env: DIST="osx"
--- a/openscad.pro
+++ b/openscad.pro
@@ -351,8 +351,8 @@ HEADERS += src/version_check.h \
            src/parameter/parametertext.h \
            src/parameter/parametervector.h \
            src/parameter/groupwidget.h \
-           src/parameter/parameterset.h\
-           src/QWordSearchField.h
+           src/parameter/parameterset.h \
+           src/QWordSearchField.h \
            src/QSettingsCached.h
 
 SOURCES += \
@@ -423,9 +423,9 @@ SOURCES += \
            src/ModuleCache.cc \
            src/GeometryCache.cc \
            src/Tree.cc \
-	   src/DrawingCallback.cc \
-	   src/FreetypeRenderer.cc \
-	   src/FontCache.cc \
+	       src/DrawingCallback.cc \
+	       src/FreetypeRenderer.cc \
+	       src/FontCache.cc \
            \
            src/settings.cc \
            src/rendersettings.cc \
--- a/openscad.pro
+++ b/openscad.pro
@@ -353,6 +353,7 @@ HEADERS += src/version_check.h \
            src/parameter/groupwidget.h \
            src/parameter/parameterset.h\
            src/QWordSearchField.h
+           src/QSettingsCached.h
 
 SOURCES += \
            src/libsvg/libsvg.cc \
@@ -492,7 +493,9 @@ SOURCES += \
            src/parameter/groupwidget.cpp \
            src/parameter/parameterset.cpp \
            src/parameter/parametervirtualwidget.cpp\
-           src/QWordSearchField.cc
+           src/QWordSearchField.cc\
+           \
+           src/QSettingsCached.cc
 
 
 # ClipperLib
--- a/src/Dock.cc
+++ b/src/Dock.cc
@@ -1,5 +1,5 @@
 #include <iostream>
-#include <QSettings>
+#include "QSettingsCached.h"
 
 #include "Dock.h"
 
@@ -13,7 +13,7 @@ Dock::~Dock()
 
 void Dock::setVisible(bool visible)
 {
-	QSettings settings;
+	QSettingsCached settings;
 	settings.setValue(configKey, !visible);
 	if (action != nullptr) {
 		action->setChecked(!visible);
--- a/src/Preferences.cc
+++ b/src/Preferences.cc
@@ -29,8 +29,8 @@
 #include <QMessageBox>
 #include <QFontDatabase>
 #include <QKeyEvent>
-#include <QSettings>
 #include <QStatusBar>
+#include <QSettings>
 #include <boost/algorithm/string.hpp>
 #include "GeometryCache.h"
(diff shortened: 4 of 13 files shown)


t-paul:
Hum, disabling whole test suites is a bit naughty. That said, we were talking about dropping support for 12.04 now that Travis has newer versions available and also most people have finally moved on.

DrCodeWizzard:
Yes, However the travis build was failing to pull boost... So after chatting with Marius on irc I went ahead and disabled the build...

t-paul:
Ok, cool. So when this is merged, 12.04 will be officially dropped (while strictly speaking that's only about Travis, it's very likely other changes making it incompatible with 12.04 will sneak in soon). I'll make a note in the description above...

kintel:
We may just merge in the 12.04-related commit separately. I think it's about time to drop that support.

kintel:
Looks good - thx!

donbright:
This std::mutex stuff breaks mingw cross build (at least on the romilly machine)



```

../src/QSettingsCached.h:46:21: error: 'mutex' in namespace 'std' does not name a type
         static std::mutex ctor_mutex;
                     ^
../src/QSettingsCached.h: In constructor 'QSettingsCached::QSettingsCached()':
../src/QSettingsCached.h:16:33: error: 'mutex' is not a member of 'std'
                 std::lock_guard<std::mutex> lock{ctor_mutex};
                                 ^
../src/QSettingsCached.h:16:33: error: 'mutex' is not a member of 'std'
../src/QSettingsCached.h:16:43: error: template argument 1 is invalid
                 std::lock_guard<std::mutex> lock{ctor_mutex};

```

see also 

https://stackoverflow.com/questions/36083003/error-mutex-does-not-name-a-type

https://stackoverflow.com/questions/14191566/c-mutex-in-namespace-std-does-not-name-a-type

https://github.com/bitcoin/bitcoin/issues/8511

log here:

[nightbuild.log.txt](https://github.com/openscad/openscad/files/1210170/nightbuild.log.txt)

kintel:
@donbright Two questions:
1) Is this only failing for the 32-bit build? 
2) Are you using the latest MXE, i.e. one that comes with a mingw that supports the std::thread library?

I'd like to be able to use C++11 proper, but if this stays an issue, we could temporarily work around it by rewriting the a Qt mutex since this code is Qt-dependent anyway.

DrCodeWizzard:
Alternatively I could replace the QSettings heap object with a static class
variable... That would get rid of the need for a mutex.

Konrad

donbright:
@kintel i think its 64bit and 32bit , as far as mxe version im not exactly sure....  @t-paul might know more about the version of mxe on the romilly build machine. he has a stable branch created there just for building openscad binaries. (actually there are two different versions of mxe under /opt ... could become a diskspace issue some day) 

i will try  to put an updated version of mxe under /opt/mxe_rolling and see if it makes a difference. will take a day or two to rebuild the packages. 

@DrCodeWizzard  i vote we try to update mxe first and see if it works

donbright:
i updated mxe to their git master under /opt/mxe_rolling and it still doesnt build, but its for a different reason. 

apparently mxe updated CGAL version ,and newer CGAL is using std::mutex too. 

build log here:

[x.txt](https://github.com/openscad/openscad/files/1222042/x.txt)

cgal info indicating they are using std::mutex

http://doc.cgal.org/latest/Manual/installation.html

not quite sure how to fix, will take some digging. off the top of my head, maybe mxe has an older c++ compiler / library... maybe CGAL has to have some special flags supplied.. . . even if @DrCodeWizzard eliminated mutex from OpenSCAD itself we still have to deal with the CGAL issue.

kintel:
There's a hint here: https://github.com/mxe/mxe/issues/1411

..e.g. do we have the `i686-w64-mingw32.static.posix-g++` executable in our MXE build?

donbright:
i will try that. 

`openscad@romilly:/opt/mxe_rolling$ make -j2 MXE_TARGETS='x86_64-w64-mingw32.stat
ic.posix i686-w64-mingw32.static.posix' qtbase qscintilla2 mpfr eigen opencsg cgal nsis glib libxml2 freetype fontconfig harfbuzz`

will probably take overnight to build. 

i might be  misunderstand the CGAL docs but it looks like they may have some build flag that will tell them to use boost mutexes and stuff instead of std::mutex,, but maybe MXE is not doing that when it configures CGAL? anyways. if CGAL is going to default in the future to std::mutex i guess we need to jump on that train eventually.

 XXXX.mingw32.static.posix seems to be unofficial as well,  its not on the main mxe.cc page.

donbright:
@kintel thanks your solution worked, at least as far as building goes.

i pushed the fix to the default mxe build script 9b03788..f837fc5

@t-paul i am not sure of the best way to update /opt/mxe on romilly... i don't want to mess up the openscad branch you have there. . .  but basically the change is to just change MXE_TARGETS to add the '.posix' at the end and rebuild the whole mxe tree.

t-paul:
The branch was both preserving the local changes from romilly and including stuff that's not yet upstream in MXE. I've updated the branches, dropping the local changes as those do not seem to be applicable anymore (CGAL is updated, OpenCSG is updated locally, but should have all the changes, MXE_TARGETS are coming from the build script).

This is now building...

t-paul:
Looks like MXE build is though now (both 32bit and 64bit)...
```
real    905m38.569s
user    800m0.060s
sys     78m30.716s
```

Next up, OpenSCAD build...

t-paul:
Looking good, at least the build did succeed, thanks @donbright
```
sending incremental file list
OpenSCAD-2017.08.30-x86-32-Installer.exe
     16,476,392 100%  124.46MB/s    0:00:00 (xfr#1, to-chk=3/48)
OpenSCAD-2017.08.30-x86-32.zip
     16,500,950 100%   58.50MB/s    0:00:00 (xfr#2, to-chk=2/48)
OpenSCAD-2017.08.30-x86-64-Installer.exe
     16,637,629 100%   34.42MB/s    0:00:00 (xfr#3, to-chk=1/48)
OpenSCAD-2017.08.30-x86-64.zip
     16,632,164 100%   18.38MB/s    0:00:00 (xfr#4, to-chk=0/48)

sent 66,266,235 bytes  received 92 bytes  7,796,038.47 bytes/sec
total size is 729,395,443  speedup is 11.01

real    42m44.868s
user    39m37.852s
sys     2m57.088s
```

PR #2035, opened 2017-05-20
Laser cutter SVG output

jmacarthur:
This adds an option to make the SVG export more useful for laser cutters. The first option is to add different colours to interior and exterior polygons, which then allows laser cutters to cut interior polygons first. The second allows general offsetting of all polygons to account for the width of the laser beam (kerf). 

In order to allow different colours, the 'laser output' option has to output SVGs with separate <path> elements for each polygon, rather than one path for all output, as the current one does. This has also raised a slight issue with the preferences system; since the preferences can't be accessed without Qt, I've had to #ifdef out the code which uses them, so this new export option won't work in headless builds, or in the tests.

Apologies if the code is a bit rough; I haven't done c++ for a while. Tested on Ubuntu 16.04.2.

### DIFF
--- a/src/Preferences.cc
+++ b/src/Preferences.cc
@@ -196,13 +196,13 @@ void Preferences::init() {
 
   // Advanced pane	
 	QValidator *intValidator = new QIntValidator(this);
-	QValidator *floatValidator = new QDoubleValidator(this);
+	QValidator *doubleValidator = new QDoubleValidator(this);
 #ifdef ENABLE_CGAL
 	this->cgalCacheSizeEdit->setValidator(intValidator);
 #endif
 	this->polysetCacheSizeEdit->setValidator(intValidator);
 	this->opencsgLimitEdit->setValidator(intValidator);
-	this->laserOffset->setValidator(floatValidator);
+	this->laserOffset->setValidator(doubleValidator);
 
 	initComboBox(this->comboBoxIndentUsing, Settings::Settings::indentStyle);
 	initComboBox(this->comboBoxLineWrap, Settings::Settings::lineWrap);
--- a/src/Preferences.cc
+++ b/src/Preferences.cc
@@ -168,6 +168,8 @@ void Preferences::init() {
 	this->defaultmap["advanced/reorderWindows"] = true;
 	this->defaultmap["launcher/showOnStartup"] = true;
 	this->defaultmap["advanced/localization"] = true;
+	this->defaultmap["advanced/laserColors"] = false;
+	this->defaultmap["advanced/laserOffset"] = 0.0;
 
 	// Toolbar
 	QActionGroup *group = new QActionGroup(this);
@@ -193,12 +195,14 @@ void Preferences::init() {
 	this->defaultmap["3dview/colorscheme"] = "Cornfield";
 
   // Advanced pane	
-	QValidator *validator = new QIntValidator(this);
+	QValidator *intValidator = new QIntValidator(this);
+	QValidator *floatValidator = new QDoubleValidator(this);
 #ifdef ENABLE_CGAL
-	this->cgalCacheSizeEdit->setValidator(validator);
+	this->cgalCacheSizeEdit->setValidator(intValidator);
 #endif
-	this->polysetCacheSizeEdit->setValidator(validator);
-	this->opencsgLimitEdit->setValidator(validator);
+	this->polysetCacheSizeEdit->setValidator(intValidator);
+	this->opencsgLimitEdit->setValidator(intValidator);
+	this->laserOffset->setValidator(floatValidator);
 
 	initComboBox(this->comboBoxIndentUsing, Settings::Settings::indentStyle);
 	initComboBox(this->comboBoxLineWrap, Settings::Settings::lineWrap);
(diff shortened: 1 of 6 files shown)


kintel:
Thanks for the pull request. I have a few comments/questions about this:

This is really a CAM extension to OpenSCAD, and I feel that it doesn't really belong in OpenSCAD itself, but should rather be a separate tool or a plugin or smth. (except we don't currently have a plugin API).
Also, I would expect most people using a laser cutter to also use the CAM software that comes with it. Such software should handle both cutting order and kerf compensation based on the model/configuration/type of the laser cutter in question.
Is the functionality you added a workaround for a particular laser cutter vendor?

**If** we decide to merge this anyway, here are some comments on the code:
* Turning this on and off doesn't belong in Preferences, but rather in an export option or CAM option. Cmd-line options can also be used in headless mode.
* The colors are hardcoded to red/green. Are these default colors used by your laser cutter vendor? 
* This isn't specific to SVG export and should be made general to all 2D exports (currently DXF and SVG).

jmacarthur:
Hi, I understand that, it really is a CAM thing. The problem is that I'm not often in control of the CAM software; I send SVG files to laser cutting services who use their own software. Most of these services do mainly decorative work rather than mechanical, so don't see the importance of kerf adjustment.

Red and green aren't magic colours; it only matters that there are different colours for interior and exterior lines. Visicut (the tool I use when I do have direct access to a laser cutter) allows ordering of cuts by line colour but doesn't have any pre-assigned order.

This could be done by a post-processing script on the SVG; it just turned out to be much easier to do this inside OpenSCAD as it already has the Clipper library. So, I will have a think about what I can do about the options, and will raise another request if I can get something a bit more suitable for OpenSCAD.

Thanks for taking time to have a look through it.

kintel:
A separate cmd-line tool might work too. We could potentially add this in a contrib folder or smth. if you'd like to benefit from existing infrastructure in the OpenSCAD repo.

nophead:
2D kerf compensation is easy inside OpenSCAD using offset.

It would be handy to be able to control the SVG colours though. There is already a PR to do that but it doesn't allow the inside edges to be different from outside.

MichaelAtOz:
+1, whenever I look into laser cutting, or CNC, with OpenSCAD in mind as the geometry source, it all looks too hard{1}. 
There appears to be a need somewhere between CAD & CAM, the CAM wants coloured lines (and arcs for efficient cuts) an 
presumably due to the use of cheap CAM 
[{1}](https://spin.atomicobject.com/2015/07/24/atoms-bits-programming-openscad-laser-cutting/)
"The SVG can’t be printed directly because my laser only cuts hairline (0 width) vectors. You can see in the SVG above that OpenSCAD exported vectors with 0.5pt and filled polygons. This needs to be cleaned up in a vector editing program. I used Corel Draw, but Inkscape and other vector editors work fine.

One oddity of OpenSCAD is that it creates unitless geometry. I treated all dimensions as mm, but when importing the SVG into Corel, the sizes were wrong. I just selected the piece and manually set the dimensions of the wheel. If your vector editor doesn’t support setting accurate physical dimensions, you’ll have a very hard time getting accurate cuts."

MichaelAtOz:
oops, premature post, having some Web problems, I'll edit it when it will let me....

MichaelAtOz:
+1, whenever I look into laser cutting, or CNC, with OpenSCAD in mind as the geometry source, it all looks too hard{1}. Deferring my investment in this area... :(

There appears to be a need somewhere between CAD & CAM, the CAM wants coloured lines (and arcs for efficient cuts) and certain line widths. The Epilog manual {2} says for AutoCAD it 'prints' the AutoCAD drawing, and the AutoCAD lines should be coloured/sized as appropriate.

So this is similar to the colour 3D printing requirement. It needs support in the CAD side to specify the inputs to the CAM side.

This is presumably due to the use of cheap CAM on consumer level devices, and historical printer emulation due to CAD S/W not being CAM aware, I understand laser movement was just emulating a plotter originally. (I'd hope things like rocket building use a better CAM...)
 
So I think there is a need to be able to specify 'line' width and colour in OpenSCAD, somehow. What would you think if you had to manually edit every STL file before feeding it to the slicer?


[{1}](https://spin.atomicobject.com/2015/07/24/atoms-bits-programming-openscad-laser-cutting/)

"The SVG can’t be printed directly because my laser only cuts hairline (0 width) vectors. You can see in the SVG above that OpenSCAD exported vectors with 0.5pt and filled polygons. This needs to be cleaned up in a vector editing program. I used Corel Draw, but Inkscape and other vector editors work fine.

One oddity of OpenSCAD is that it creates unitless geometry. I treated all dimensions as mm, but when importing the SVG into Corel, the sizes were wrong. I just selected the piece and manually set the dimensions of the wheel. If your vector editor doesn’t support setting accurate physical dimensions, you’ll have a very hard time getting accurate cuts."


[Epilog Manual](https://www.epiloglaser.com/assets/downloads/manuals/legend-manual-web.pdf) Section 12
"In the AutoCAD print window, set the pen colors that you 
are using to .001” (.0254 mm).
The AutoCAD default for all pens is .010 and a line weight 
this large disables vector cutting. If the line weight is not 
changed to .001 inch, vector cutting will not be possible.
When the laser receives a job where the line weights are 
too large, the laser system will just beep when you attempt 
to run the job."

MichaelAtOz:
[here](http://www.cam-service.com/en/cam-system-cagila/take-a-tour/) is an overview of a more sophisticated CNC CAM process, it still requires line colours for 'special applications'.

Issue #1840, opened 2016-10-29
Customizer: Don't scan for parameters in included files


kintel:
Fix handling of includes, those should not be scanned for parameters (see [forum post](http://forum.openscad.org/GSoC-OpenSCAD-project-td18102i20.html#a18761))

(moved from #1781)

amarjeetkapoor1:
Any idea on how to solve this issue. Only solution that is coming to my mind is that we keep the name of file like the location is stored for each node (might be usefull for long run) or simply store a boolean to tell whether its from from included file or not to make customizer behave like that of Thingiverse. Otherwise, The behavior is somewhat correct (50%) as it's meant to behave like if part of main file.

MichaelPFrey:
Currently, parameters from includes get either no or the wrong description:

![error](https://user-images.githubusercontent.com/24962768/36644903-2c1cb240-1a61-11e8-977b-f95b3cb1826a.png)

tricky situation

amarjeetkapoor1:
#2044 Could solve this BUT there is a problem attached.
Which would be solved easily by increasing stack size that wouldn't be a viable solution but not good one and another one would require some extra work.

PR #2044, opened 2017-05-28
Don't scan for parameters in included files 

amarjeetkapoor1:
In reference to issue #1840 .
It adds the file name to the location of the node and also doesn't scan for parameters in included files.

### DIFF
--- a/src/AST.cc
+++ b/src/AST.cc
@@ -1,3 +1,3 @@
 #include "AST.h"
 
-const Location Location::NONE(0, 0, 0, 0);
+const Location Location::NONE(0, 0, 0, 0, "");
--- a/src/AST.h
+++ b/src/AST.h
@@ -1,11 +1,16 @@
 #pragma once
+#include<string>
 
 class Location {
+
 public:
-	Location(int firstLine, int firstCol, int lastLine, int lastCol)
-		: first_line(firstLine), first_col(firstCol), last_line(lastLine), last_col(lastCol) {
+	Location(int firstLine, int firstCol, int lastLine, int lastCol,
+			const std::string &filename)
+		: first_line(firstLine), first_col(firstCol), last_line(lastLine),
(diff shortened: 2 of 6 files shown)


amarjeetkapoor1:
@kintel This PR is failing because of error "memory exhausted" is the case of test related rand.scad.

And the **solution** is to define `#define YYINITDEPTH ` in parser with some value greater than 200 
as what I found it that there is no relocation of stack happening in the parser and the upper limit is 200 not the 20000 as defined by 
`#define YYMAXDEPTH 20000`

So, should I define #define YYINITDEPTH in the parser?

https://stackoverflow.com/questions/20861945/how-to-increase-stack-size-in-bison-and-solve-memory-exhausted

amarjeetkapoor1:
Another way is to define `# define YYLTYPE_IS_TRIVIAL 1` in parser although this also worked but not sure should we do that in general.

kintel:
Defining YYINITDEPTH would make sense _if_ this is actually the issue.
To my understanding, this would only happen you have more than 200 unshifted symbols on the bison stack. That tends to indicate that something is very wrong (e.g. bug in the parser rules), but I haven't looked that the details.

amarjeetkapoor1:
I checked the code of the main branch with YYMAXDEPTH=200 (i.e. no reallocation) and got the memory exhausted error. 

Above error is indeed due to a lot of unshifted symbols on the bison stack which went unnoticed due to the reallocation of the stack but this code is stopping the reallocation which would only be enabled either by defining  `YYLTYPE_IS_TRIVIAL` or by defining our own `yyoverflow` which will be a non-trivial task.

and how would we find the bug in parser rules which is causes memory exhaust?

MichaelPFrey:
Closed/obsolet by #2322?

t-paul:
Yes, should be covered by #2322, tested with a simple case and that worked ok.

PR #2049, opened 2017-06-03
Update progress as we process 2D items

DrCodeWizzard:
Update progress bar as we process 2D items. 
To test open openscad and run the following:
for(o = [0:10:10000]){
translate([o,0])
circle(r=10, center=true);
}


MichaelAtOz:
Also eventually needs testing with multi-thread enabled.

DrCodeWizzard:
yeah I have not tested it with multithreading enabled (not sure how to flip that switch right now). I would need to see how the progress update is implemented to see what changes would need to be made to make the progress consistent and race condition free...

MichaelAtOz:
Well the 3D one, same mechanism,  works multi-threaded (seems to), I just left the above as a reminder for final checking.

DrCodeWizzard:
ok :-) What's the magic switch for a test?

Issue #1209, opened 2015-02-08
Progress bar slightly too big, changes status bar height


MichaelAtOz:
Looks odd. Status bar grows in height while progress underway, then shrinks back down.
It seems to be the bar height, not the n/1000 text or stop icon.

ChrisVesters:
I looked into it, turns out that the status bar has height 23, while the progress bar has height 26.
It is not possible to shrink the progress bar, as this is fixed by Qt. The only option seems to be to make a bit more room for the status bar and increase it to 26, such that it can hold the progress bar.

ChrisVesters:
An alternative way is to set the max height of the statusbar to 23.
I have taken screenshots from before and after to see how the status bar looks. Only some whitespace before and after the actual progress bar is lost, so this is a viable option.

![fixed](https://cloud.githubusercontent.com/assets/4047497/26697954/39aef8c6-4714-11e7-98ca-012493c6ae29.png)
![original](https://cloud.githubusercontent.com/assets/4047497/26697955/39b6e43c-4714-11e7-94f1-e61c371f29fe.png)

MichaelAtOz:
Looks good. I always like seeing a bit more usable screen real estate squeezed out, excessive white space is a pet hate. Just look at this page!

MichaelAtOz:
Fixed by #2050.
Thanks Chris. Claim the bounty.

ChrisVesters:
@kintel This issue can be closed after you do some checks yourself if you want.
Is there a procedure such that it is easier to close the issue?

MichaelAtOz:
Seems Windows version is different.

Using 2018.01.06 these show the height difference.
![pre f6](https://user-images.githubusercontent.com/1972961/36363422-df91a42a-1534-11e8-9b81-549718a2881f.JPG) ![during f6](https://user-images.githubusercontent.com/1972961/36363426-e78ac7c4-1534-11e8-960f-97b9e090f740.JPG)

MichaelAtOz:
I wonder if my Windows settings are at play?
i.e. 
![windows colours settings](https://user-images.githubusercontent.com/1972961/36363657-9d7aea54-1536-11e8-952f-8daf9c4ba430.png)
Tho there is no 'status bar' option...

nophead:
What version of Windows is that, XP? It looks completely different for me on Win7.

I would add a picture but Github seems to have lost the ability at the moment.

MichaelAtOz:
[...]

PR #2050, opened 2017-06-03
Issue 1209 - Fix status bar growing when showing progress bar.

ChrisVesters:
- Set max height of status bar.
- Issue #1209

### DIFF
--- a/src/MainWindow.ui
+++ b/src/MainWindow.ui
@@ -420,7 +420,14 @@
    <addaction name="menu_View"/>
    <addaction name="menuHelp"/>
   </widget>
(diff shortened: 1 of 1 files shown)


MichaelAtOz:
@kintel This should be in the snapshots?? 

If so it hasn't worked for Windows, 2018.01.06 the status bar still grows & shrinks.

This is pre F6:
pre f6

This is during F6:
during f6

kintel:
This should be in the snapshots. The status bar has a maximum size of `23`. Perhaps 23 isn't the correct number across all environments? If you reopen the original issue and post your screenshots, that would be helpful.

Issue #2019, opened 2017-05-01
CMake build system


kintel:
In #1048, we merged a rough first attempt at a CMake build system.
We're not close to be able to use CMake in production, but we'll track improvements ideas here:
* [x] Make it work on Linux
* [ ] Make it work on (any) Linux box
* [ ] Make it work with MXE
* [ ] Integrate test framework (e.g. using a subdir)
* [ ] Be able to build releases (all platforms)
* [ ] Be able to build development snapshots (all platforms)
* [ ] Option for headless (no Qt) build

thehans:
Regarding linux, I tried to run cmake ./ from a fresh openscad clone and got an error about my libxml version.

This is on Linux Mint 18.1 (based on Ubuntu 16.04)

[cmake.txt](https://github.com/openscad/openscad/files/972278/cmake.txt)

kintel:
@thehans I've downgraded the libxml2 requirement to 2.9.

FYI: The version requirements are currently not validated against our actual requirement, so feel free to report or fix such issues. Testing on Linux is very appreciated!

thehans:
Ok, the next errors I encountered were

```
CMake Error at CMakeLists.txt:381 (add_executable):
  Cannot find source file:

    src/MainWindow.cc

  Tried extensions .c .C .c++ .cc .cpp .cxx .m .M .mm .h .hh .h++ .hm .hpp
  .hxx .in .txx
```
and then 
```
CMake Error at CMakeLists.txt:381 (add_executable):
Cannot find source file:

    src/AboutDialog.cc

  Tried extensions .c .C .c++ .cc .cpp .cxx .m .M .mm .h .hh .h++ .hm .hpp
  .hxx .in .txx
```

After changing those to .h in CMakeLists.txt, cmake exits without error.  Log attached.
[cmake_log.txt](https://github.com/openscad/openscad/files/1149970/cmake_log.txt)

Then running `make`, I get this error during linking (full log also attached below):
```
[100%] Linking CXX executable OpenSCAD
/usr/bin/ld: CMakeFiles/OpenSCAD.dir/src/OffscreenContextGLX.cc.o: undefined reference to symbol 'XGetErrorText'
//usr/lib/x86_64-linux-gnu/libX11.so.6: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
CMakeFiles/OpenSCAD.dir/build.make:4493: recipe for target 'OpenSCAD' failed
[...]

PR #2068, opened 2017-07-15
Fix cmake missing file errors

thehans:
Fixes the following cmake errors encountered on linux:

CMake Error at CMakeLists.txt:381 (add_executable):
  Cannot find source file:

    src/MainWindow.cc

  Tried extensions .c .C .c++ .cc .cpp .cxx .m .M .mm .h .hh .h++ .hm .hpp
  .hxx .in .txx

CMake Error at CMakeLists.txt:381 (add_executable):
Cannot find source file:

    src/AboutDialog.cc

  Tried extensions .c .C .c++ .cc .cpp .cxx .m .M .mm .h .hh .h++ .hm .hpp
  .hxx .in .txx

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -313,10 +313,9 @@ set(OFFSCREEN_SOURCES
   src/OpenCSGRenderer.cc)
 
 set(GUI_SOURCES
-  src/MainWindow.h
+  src/mainwin.cc
   src/OpenSCADApp.cc
   src/WindowManager.cc
-  src/AboutDialog.h
   src/Preferences.cc
   src/FontListDialog.cc
   src/FontListTableView.cc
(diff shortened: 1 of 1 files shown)


thehans:
Related to #2019

CMakeLists.txt
```diff
@@ -313,10 +313,10 @@ set(OFFSCREEN_SOURCES
   src/OpenCSGRenderer.cc)
 
 set(GUI_SOURCES
-  src/MainWindow.cc
+  src/MainWindow.h
```
kintel: Should be `src/mainwin.cc`

CMakeLists.txt
```diff
@@ -313,10 +313,10 @@ set(OFFSCREEN_SOURCES
...
   src/OpenSCADApp.cc
   src/WindowManager.cc
-  src/AboutDialog.cc
+  src/AboutDialog.h
```
kintel: We shouldn't add headers here, but create a separate cmake variable, e.g. `GUI_HEADERS` and add it there.
Not adding headers separately was an oversight on my part, as I forgot that we have some header-only code.

thehans:
@kintel I commited changes based on your comments, does that look like what you expected?

Issue #2077, opened 2017-07-31
Feature Request: add "offset" as option to linear_extrude


mrkmg:
Add the ability to define an "offset" to the linear_extrude. For non-symmetrical shapes, it is currently not possible to create a smooth transition to a slightly smaller size. Scale will scale from origin, which only works for symmetrical object.

mrkmg:
I am not well versed in C/C++ (Minor experience in school). I was looking to see if this is something I could implement. I do not think I have enough understand of how this was written or how to deal with 3d geometry to work on this.

I started by looking at https://github.com/openscad/openscad/blob/master/src/linearextrude.cc to see if that is where I would start work on that. Looking though, and I have no idea what is going on there. Any pointers on where I would even start?

I am willing to learn to make this a reality. It would make many of my creations much better!

thehans:
That code just handles the parsing of the parameters, so you'd need to add support for this "offset" parameter there.  But the real work on the geometry happens here:
https://github.com/openscad/openscad/blob/b48b525adeeaa5a320c5d81ec575a9ae83cf0987/src/GeometryEvaluator.cc#L678-L730

thehans:
I think the difficult part with this is going to be figuring out which vertices in one layer connect with vertices in the next, since the number of vertices may change as rounded corners/chamfers are added.

mrkmg:
I will give it a go. 

My naive idea would be to connect from the smaller side to the larger side, and just search for the closest vertex. I have gone through in my head and it sounds like it could be a good starting point.

Now, I need to go peak at the offset code, and work from there.

TLC123:
A working offset feature could possibly be further exploited to generate rounded end caps of extrusion.
for simple shape i don't se any major issues.
One hard nut probably is complex shapes with small islands and holes that might vanish in an offset before
 the last level is reached how is those cases handled.

mrkmg:
Okay. So a couple hours in and I am making some progress.

I have a question though on how an Outline2d actually represents an outline.

I have the following scad file:
```
linear_extrude(10)
square([10, 10]);
```

I added in some debugs, and am looping through the outline passed into `static Geometry *extrudePolygon(const LinearExtrudeNode &node, const Polygon2d &poly)`;

specifically, I am doing the following:
```
std::vector<Outline2d> outlines = poly.outlines();
for (unsigned int j = 0; j < outlines.size(); j++) {
    for (unsigned int k = 0; k < outlines[j].size(); k++) {
[...]

PR #2079, opened 2017-08-02
Linear extrude offset- WIP

mrkmg:
So, after a couple days of work, I have the following demo:

image

See the issue #2077 for details on my progress, and plenty of self doubt.

As you can see, it's not perfect. Also, my c++ sucks. A lot. Some guidance on how to continue on this and maybe fix the bugs with vertices which do not translate correctly, or how to make the c++ better would be greatly appreciated.


mrkmg:
So the issues I am having:

- When I tessellate the cut of the top and bottom layers, some triangles end up having their vertices on the same edge. This leads to the bottom example, where you get "flat" chunks which hang out.
- When a feature disappears, like a small hole or island, you end up with flat chunks.

I was thinking of trying to detect these flat chunks, and just removing them. The other idea was using the slices that we already have for scaling. That would allow the islands or holes to potentially fill up partially. It would be up to the user to choose how many slices.

mrkmg:
So, I implemented the slices feature. I am getting a problem with rendering some of the models with positive offets, and still having the problem with flat XY faces which are only on one side. Overall though I think I am making good progress.

I will still need help with the c++ specific stuff like deleting references and clearing memory. I am not versed in how to write proper c++

mrkmg:
Here is my latest test:

```
$fn=12;

linear_extrude(10, offset = -2, slices = 5)
union() {
    difference () {
        square([20, 20]);
        circle(6);
        
        translate([10, 10])
        circle(5);
    }
    translate([20, 20])
    circle(5);
}
```


image

mrkmg:
More progress, I have tried a large variety of shapes. So far, it seems to be working. I am getting hopeful!

image

As you can see, it can close holes elegantly. I know some of the ways I am doing things is "hacky" and probably against some sort of rules in C++. Also, when it comes to the pointers, I am totally just guessing when I should be using * or & or nothing, completely from when the compiler yells at me.

mrkmg:
Having an issue when rendering, and I have no idea what this means:

ERROR: CGAL error in CGAL_Nef_polyhedron3(): CGAL ERROR: assertion violation! Expr: e->incident_sface() != SFace_const_handle() File: /usr/include/CGAL/Nef_S2/SM_const_decorator.h Line: 329 

Can anyone help me on this? At least tell me what an incident_sface is so I have an idea of what to try. So far, trial and error has been working for me.

nophead:
I think it means you have a self intersection.

mrkmg:
So, I have fixed almost all the issues I can find, even with pretty complex objects. I have resolved the intersecting faces issue, tested using the result of linear_extrude with offset on unions, differences, and intersections. It does appear to be working as expected.

Like I have said a couple times, the code quality is obviously pretty horrible. If someone has some time to mentor me or give some suggestions, I would be more than willing to learn and make the changes needed to get this pull ready.

Here is another example of my lastest updates.

image

```
$fn=30;

linear_extrude(5, offset = -2, slices = 5)
union() {
    difference () {
        square([20, 20]);
        circle(6);
        
        translate([10, 10])
        circle(5);
    }
    translate([20, 20])
    circle(5);
}



translate([35, 0])
linear_extrude(5, offset = -2, slices = 1)
circle(6);

translate([35, 20])
linear_extrude(5, offset = 2, slices = 1)
circle(6);

translate([0, -20])
linear_extrude(5, offset = -.25, slices = 1)
text("ABC123");

translate([5, -40, 0])
union() {
    difference() {
        linear_extrude(5)
        translate([-5, -5])
        square([20, 20]);

        linear_extrude(5, offset = 4, slices = 4)
        square([10, 10]);

    }
    
    linear_extrude(5, offset = -1)
    square([5, 5]);
    translate([5, 0])
    linear_extrude(5, offset = -1)
    square([5, 5]);
    translate([0, 5])
    linear_extrude(5, offset = -1)
    square([5, 5]);
    translate([5, 5])
    linear_extrude(5, offset = -1)
    square([5, 5]);
}


translate([60, 20, 5])
rotate([180, 0, 0])
linear_extrude(5, offset = 5, slices = 20)
difference() {
    circle(10);
    translate([2, 0])
    circle(3);
    translate([-2, 0])
    circle(3);
}
```

mrkmg:
TODO:

1) Currently all offsets are miter. Seeing as this is using a single parameter (offset), instead of the 3 (delta, r, chamfer) that the offset module uses, I am unsure how to make it similar. Maybe having offset, offset_delta, and offset_chamfer?
2) Clean up code.  It's garbage.
3) If you create an object, lets say circle(5), and then offset to -5 with only 1 slice so the entire top layer disappears, then the result is only the bottom layer, aka a flat single face object. This is obviously no good.
4) More testing. Aside from the issue above, everything else I tried does compile and render. I have tried about 20 or so shapes.

kintel:
Thanks for taking a stab at this. I just had a glance at your code - here are my initial comments:

First of all, don't worry about code right now; that's easy to clean up. The primary challenge with this feature is primarily to design the API and then to define corner case behavior.

With the added `offset` parameter i feel we're really stretching the original intent behind `linear_extrude()` quite far. Especially considering the `offset()` module analogy and the fact that your offset implementation disables the other parameters to `linear_extrude()` (e.g. `twist`, `scale`).

This would be significantly cleaner to implement if we could somehow let pass the actual `offset()` module to `linear_extrude()`, something which still requires significant refactoring to pull off.

I do agree that, if we want to introduce an offset parameter, it should ideally support the same offset mechanisms as the `offset()` node. This will, of course open for challenges like how to map outlines with varying number of vertices.

Intuitively, it sounds a bit like this is similar to the `skin()` function in https://github.com/openscad/list-comprehension-demos#extrusionscad. A native implementation of `skin()` as a module could perhaps do all of what this offset parameter does. To make that work properly, we probably need to be able to unroll `for()` loops (aka. "no implicit union"; #350) to be able to generate skinnable slices.

mrkmg:
If the idea that having offset as a parameter of linear_extrude is going too far, then I would assume that having scale as one is also. If we had a skin function, or the ability to pass a "module" as a "parameter" as you stated then the same logic could be applied to scale. I can not think of a case where scale differs from offset in relation to linear_extrude given the ability to use one of the other methods you proposed for offset. We could pass the scale() module just the same as the offset() module, or just scale and skin.

I do believe I can get the twist function to work. I have not figured it out yet, but I am becoming more confident and since my last post come up with an idea. This would only leave the scale function, which I also think could be implemented.

I'll admit though, that I do not know what the long term plans are. I do not mean the following as a critism, and I obviously think this project is amazing, but it does seem like development is quite slow, and there are only a few maintainers. If those features you spoke of are too far out, and require huge rework, then I think this is a good way to accomplish many tasks that currently are very difficult or require a large amount of work.

If we do not want to extend linear_extrude too far, maybe introducing another top level module, like offset_extrude, which only takes (height, r, delta, chamfer). That would solve a couple problems, like blocking parameters from linear_extrude, and also gives us the ability to access all the offset types.

nophead:
I really think this is needs to be implemented without having to wait for module parameters and other big changes which never seem to happen. Too many very useful pull requests just get ignored. It is better to have a lot of parameters to linear_extrude than not have this facility. It is needed to be able to make laser cut stamps as you want to taper thin details like text outwards to make them stronger. 

It is much more useful than scale. I can't think why you would need that with offset.

mrkmg:
I do not think it can replace scale completely, as it functions completely different. I do also see the issue of having scale and twist not working if offset is passed in. In response to that, I would like to be able to implement both scale and twist parameters alongside of offset.

At the end of the day, I have a *mostly* working implementation, and if we do not want to overload linear_extrude I think a fair compromise would be to have another function like offset_extrude. That way we can implement all the different offset types, and not worry about touching linear_extrude.

I have not yet tried to add a new module from the c++ side. It was much easier to modify what was already there. Anyways, I am open to either method, but agree with @nophead that this could be very useful to many people and the models they are trying to create. I know there are a few models I have that I have used very awkward and slow work-arounds that this would solve (which is why I have devoted so much time to this issue already).

I am also not opposed to helping with other issues once I become more comfortable with the codebase. Being a programmer, OpenSCAD has been my favorite modelling program so far. None of the other Open Source alternatives are as easy to  or fun to use.

mrkmg:
So,

I have made a new module, offset_extrude which mimics the offset module with slices and height added. 

But, I am not having an issue with the preview. When I do an intersection or difference with the result of offset_extrude, preview does not work. Nothing shows up. Everything looks correct when I render though. I am not sure what I would have to do get preview to work.

What would cause render to work, but preview not to?

Here is an example (use the latest commit) which recreates the problem:

```
    intersection() {
        offset_extrude(3, r = 2)
        circle(2);
        
        linear_extrude(3)
        square([5, 5]);
        
    }
```

mrkmg:
I think my preview may have something to do with the back of the faces. I am not specifically rendering the back of the faces, and I have a feeling I should be. I get that idea from looking at the way that linear_extrude works. It seems to add 2 or 4 polygons per face.

I am looking into this more, but not sure how I would accomplish this with how I am rendering everything. Hmm, more learning!

mrkmg:
I am not sure why, but if I manually perform append_poly and add_vertex instead of appending the whole polyset preview starts to work. Must be something with how rendering in preview works versus render.

I also reverted everything related to linear_extrude.

Couple issues still. If you have a 0 delta, you get no edges. Also, extremely small sides lead to the same issue. Looking at work arounds now.

kintel:
I think `offset_extrude()` may be a decent trade-off.

kintel:
One initial review comment: It would be helpful if you follow the same coding style as the remaining code. As a start, the main thing to keep in mind is that we currently use tabs, not spaces, to indent code.

kintel:
I've added a bunch of TODO items in the initial description. As long as we tag this feature as experimental, we can merge this PR early and manage further cleanup as time permits. It would be nice to have a pretty good idea of the API and decent test coverage before doing that + that coding style should be somewhat coherent.

mrkmg:
I will work on the todo over the next week!

MichaelAtOz:
@nophead said
> Too many very useful pull requests just get ignored. It is better to have a lot of parameters to linear_extrude than not have this facility. It is needed to be able to make laser cut stamps as you want to taper thin details like text outwards to make them stronger.

I understand laser cut has a tapered v shaped beam, but how do you influence the taper from a design perspective? I thought laser cutting used 2D for path generation. Or are you suggesting visual simulation of the cut in 3D?

doug-moen:
With the right software and controller board, a laser cutter is capable of
making variable depth cuts into soft material like wood or inkstamp
material.
[image: Inline images 1]

nophead:
I ordered a Glowforge 3D laser printer about two years ago but haven't received it yet. It can modulate the beam intensity to produce variable depth engraving so you can make shallow 3D objects with no overhangs. See https://community.glowforge.com/t/dovetail-joint/9946

mrkmg:
I have not forgot about this! I am just slammed on other projects currently. Will be coming back to this hopefully over the holidays.

mrkmg:
I have done some initial code cleanup. Anything else I need to do cleanup wise. I converted everything to tabs and renamed all variables to be snake-case.

I also merged master, so it's not mergeable again.

mrkmg:
So, getting back in this and I found a couple "bugs", or models which do not extrude well. Specifically,

```
offset_extrude(height, r = -2)
difference() {
    circle(10);
    
    translate([0, 10, 0])
    circle(10);
}
```

The edges are absolutely awful. I am working on figuring out a work around for this case.

mrkmg:
I have found the root cause of *most* of the issues I can find.

Occasionally, the faces of a plane and its offset are "sliced" in a way that some surfaces do not "touch" both sides. Attached in an example.

edge

As you can see in this example, the polys are set in a way that the "red" section will all map to the bottom face.

Is there a way to force a "face" to have all their polygons attach to both sides? The polygons I generate to make the edges will always be "donuts", and if we could force the polys to have points that correlate on both sides, then I think we can deal with almost any shape.

kintel:
I have trouble understanding what you mean from that figure..
It sounds like your issue is correlating vertices between the original and the offset shape. Since these two shapes may have a different number of vertices, your initial attempt to do this is to build a polygon with a hole, triangulate this polygon, and use the resulting triangles to define connectivity between the top and bottom of the extruded shape.
While this is an interesting way of doing this, you would, as you point out, require a special type of triangulation which guarantees a clean "strip" of triangles.
I think the most appropriate solution is to build this triangle strip manually instead of using `tessellate()` to do this.

mrkmg:
That is what I kind of assumed I would have to do. Much more learning then I see. 

My first attempt at this was manually building the triangles. The problem I had was not understanding how the triangles work, and issues with "front" and "back" faces.

mrkmg:
I feared that I would have to do that. I am still on am very shaky on how to properly build triangles from scratch.

I guess more learning is coming my way!

kintel:
The main thing to keep in mind when building triangles is the winding order. When building quads, making a good choice of how to split the quad into two triangles is also important.
The existing code should be a good starting point for that.

Hopping onto #openscad on IRC freenode sometimes helps for interactive discussions.

TLC123:
Seams to me like tessellation is a good place to start.
It solves  internal island pretty robust. 
Your solution just needs to flip edges until no triangle has same z in all points. 

until no more changes
   find flat triangel
      does it share edge with a sloping neighbour ? flip edge with it



> I have found the root cause of _most_ of the issues I can find.
> 
> Occasionally, the faces of a plane and its offset are "sliced" in a way that some surfaces do not "touch" both sides. Attached in an example.
> 
> edge
> 
> As you can see in this example, the polys are set in a way that the "red" section will all map to the bottom face.
> 
> Is there a way to force a "face" to have all their polygons attach to both sides? The polygons I generate to make the edges will always be "donuts", and if we could force the polys to have points that correlate on both sides, then I think we can deal with almost any shape.
> 
> This is what I "want" it to be:
> 


> edge - want

mrkmg:
@TLC123 To be honest, this is WAY out of my league. I tried for days and days to understand how to tessellate myself, and after accomplishing nothing I gave up. I know, not the best way to be. I started working on this because I wanted it for a model I was making to 3d print. I accomplished what I needed with the first version of the PR of this, and figured I would try to get this ready for others to use. But, alas, I was unable to accomplish that.

At this point, I have not worked on this in like 2 years, and I am not sure I would be able to achieve any more success now than I did 2 years ago... At least not without some major learning which I just do not have time to do.

mrkmg:
@TLC123 I have figured out what you meant, and implemented it, and it WORKED.

I have pushed up the current state of where I am at. I am putting together a comprehensive test suite of objects now, so I can verify this does not fail on any of them.

@kintel, can you remove the stalled tag, as I am working on this actively.

Thanks for everyones help, especially those over in the IRC.

mrkmg:
Here is an example of all my test shapes so far. Seems to be working for everything I throw at it.

All

I think I am at the point to start going through the stages listed in the first comment.

As stated above (and by looking at my GH profile), I am not a c++ developer, and almost all my experience in c++ has been educational in nature. I am sure my PR will need someone with a c++ eye to go through it.

I am willing to make any changes (obviously), but I would need some guidance.

As for documentation, I can start on that asap.

mrkmg:
From the original checklist I have added an example, and added regression tests. 

Whats is the next step for me?

thehans:
Hi, I've just checked out your branch and did some testing to see if I could break it. :wink: 

I made this script which is meant to be run as an animation so it can iterate over a wide range of values (not useful as a "ctest" case, just manual/visual verification)
```
$fs = 1;
$fa = 1;
// Enable View -> Animate  
// Set  FPS: 10, Steps: 100
// View in F12 Thrown together mode

// positive offset
translate([0,10,0]) mirror([0,0,1]) 
  offset_extrude(5, r=15*$t, slices=5) 
    text("ABC", halign="center");

// negative offset
translate([0,-10,0]) 
  offset_extrude(5, r=-15*$t, slices=5) 
    text("ABC", halign="center", valign="center");
```
The first issue is where r=0, it should act like a normal linear_extrude, but looks like only outputs the top/bottom faces with no connection in-between.  I think that would be fairly simple to fix.

The more involved issue though is the inconsistency of how the new/deleted loops become merged as the animation evolves.  For increasingly positive offsets the letter "C is one kind of special case, as it creates a new inner loop when the ends initially touch, then eventually the loops disappears as it closes in on itself.
I know the artifacts visible in the animation are mostly fixable by specifying a large enough value for slices... but I'm wondering if we can figure out a way to detect offsets/heights at which certain events happen (addition/removal of outlines or maybe just vertices in general).  Maybe via a binary search through intermediate offset values.  I haven't yet analyzed the code for how you are handling the vertex merges like this, but I think it would be possible to create an implementation that removes the need to specify a number of "slices" altogether, and just automatically/optimally handles these positions.

I hope this suggestion isn't too daunting or discouraging (maybe it would be better considered as a possible future enhancement if so), as I can see this new feature being very useful!  
For example, in creating fillets fairly efficiently via a stack of extrusions with cos and sin for offset/height multipliers. I made an [example script](https://gist.github.com/thehans/b47ab7077c862361eb5d8f095448b2d4) recently which does this, but without offset_extrude it relies on hull, and so only it works on individual convex shapes.

nophead:
Note that even the current linear_extrude will create self intersections when extruding a thin rectangle with a twist unless slices is set high enough.

mrkmg:
@thehans 

Thank you for that test. I never thought of using animate to check out many iterations. Great idea.

I am working on the fix for the 0 offset now. Basically going to emulate how LinearExtrude works with 0 twist and 0 scale.

As for the inconsistency with appearing/disappearing features, I am aware of that. I have performed a numerous number of tests which show this off, and tried to think of ways to address it. The "algorithm" is far from bulletproof. To be honest, I think I have taken this further than my skills already, and at this point I am doing lots of guess and check. Implementing a "smart" system which detects feature changes seems like a whole league beyond my understanding. I am 100% open to help though! 

For uses, I actually wrote a very similar library to yours a while back, but without using hull and just many small slices (https://www.thingiverse.com/thing:2461392/). If you are 3d printing, it does not matter if it is "smooth", as you can just set your slice height equal to your layer depth. This idea came from my dissatisfaction with the end result though.

Finally, I guess I want to ask. How much can we leave to the user to deal with? Obviously we can not address every case of odd user input. Are the current flaws acceptable? What needs to be "fixed" or addressed before we can move to the experimental stage, and then release?

Thank you!

thehans:
I've pasted a suggested update to your test case below, which adds some missing permutations of positive, zero, and negative offset of different types, and added some concavity to the square shape so effects of different types of negative offset can be seen:
```
$fn=40;
size = 15;
off = size / 8;

offset_types(size*2)
  angular();

translate([0,0,size*5]) offset_types(size*2)
  curved();

module offset_types(x) {
  for(i = [-1,0,1]) { // negative and positive offsets
    translate([i*x, size *-2])
      offset_extrude(size, r = i * off) 
        children();
    translate([i*x, size * 0])
      offset_extrude(size, delta = i * off, chamfer = true)
        children();
    translate([i*x, size * 2])
      offset_extrude(size, delta = i * off, chamfer = false)
        children();
  }
}

module curved() {
  difference() {
    circle(size / 2);
    translate([0, -size / 2]) circle(size / 2);
  }
}

module angular() {
  difference() {
    square(size, center = true);
    rotate(-135) square(size);
  }
}
```
I left the curved profile in there, stacked vertically.  But maybe its enough to only use the angular version since I think it exercises the functions well enough, displays the difference between offset types more clearly, and would result in a less zoomed-out/cluttered test image.

> How much can we leave to the user to deal with? Obviously we can not address every case of odd user input. Are the current flaws acceptable?

As @nophead points out, linear_extrude already has limitations when not given enough slices (which actually results in arguably worse self-intersecting geometry compared to mostly aesthetic limitations of offset_extrude), so I may just be overly optimistic to want perfect automatically-deduced slices.

I have some more ideas how that might be implemented, but will probably require modifying or re-implementing clipper's offset functions.  I'll play around and see if I can implement something on top of your work based on those ideas, but at this point I guess I'd vote to *not* let that hold back pushing your changes through as experimental.

src/offsetextrude.cc
```diff
@@ -0,0 +1,146 @@
...
+    bool isRadius = this->join_type == ClipperLib::jtRound;
+    auto var = isRadius ? "(r = " : "(delta = ";
+
+    stream  << this->name() << var << std::dec << this->delta;
+    if (!isRadius) {
+        stream << ", chamfer = " << (this->chamfer ? "true" : "false");
+    }
+    stream << ", height = " << this->height
+           << ", slices = " << this->slices
+           << ", $fn = " << this->fn
```
thehans: missing output for center parameter
```
           << ", center = " << (this->center ? "true" : "false")
```

mrkmg: This has been fixed

mrkmg: This has been fixed

mrkmg:
@thehans any luck with detecting optimal slices?

thehans:
@mrkmg I made a little bit of progress and I think I have an overall plan that should work, but still needs quite a bit of development, requiring re-implementing clipper's offset functionality as I suspected before.  I've been focusing on various other improvements lately, and haven't gotten back to this one just yet, but I still plan to continue the efforts soonish.

TLC123:
Safe Slice height guesstimate could be little less than half of the smalkest feature dimensions. Minimum feature dimensions could probably be determined reliably by the shortest internal edge of the triangulated main shape.

RonaldoCMP:
The operations of rotation (twist), scale and offset may not all pairwise commutative. So, an extra care should be taken designing their order in linear_extrude. In the current state, linear_extrude with scale and twist is not a twist of a scaled extrusion when a non-uniform scale is prescribed. That leads to weird unexpected results. 

All three operations seems to be pairwise commutative when uniform scale is in play. The non-uniform scale is the devil.

RonaldoCMP:
A correction: offset and scale are commutable only if scale=1.

TLC123:
I thought this feature was a done deal. Really sad to see all that fine work done just to fizzle out on the finish line.

khumarahn:
I'd be very interested in the linear extrusion with offset. What is the current state of this? Is it usable, even if not on all shapes, in @mrkmg branch?

mrkmg:
It works for pretty much every shape. There are some shapes that the result is not perfect, but I have not found any which are invalid.

Obviously, it will need to re-worked to make it work against the latest version.

In regards to @TLC123 , I would love to see this in a release version, but as I stated all the way back when I started this c++ is not my strong suite, and particularly contributing to projects like this are way outside my wheelhouse. I am 100% open to working with someone on this, to get it ready for a testing release or w/e but without some specific help I don't think any more progress will be made on it.

khumarahn:
@mrkmg I tried to compile your branch by changing repo in gentoo ebuild for live openscad, but compilation failed. I will try to compile it again soon.

TLC123:
> Conflicting files
> CMakeLists.txt
> src/highlighter.cc

Is these the only remaining issues?

mrkmg:
@TLC123 It's been a long while since I have worked on this. I will try in the next week or so get it to compile again and see if there is any outstanding issues I am not thinking of.

richfelker:
Great to hear there's progress being made on this again. I've been looking forward to it for a long time.

mrkmg:
@TLC123, I have rebased on the latest master, and fixed all the issues I could find. All tests are passing, and the build is working on all platforms I can build for here. I went through all the code also and cleaned up to try and match the existing style.

@khumarahn, The build is working on my two dev machines (One Manjaro, one ubuntu). I would say try the latest commit. If it still doesn't work, then maybe try the latest master and make sure that builds on your machine.

Overall, I think this is a decent position for someone to review it. Looking at the original "Todo List", I believe we can start going through it:

- [x] Stage 1: Coding style fixes *(may need more input on this from someone more familiar with the project)*
- [ ]  Stage 1: Code review *(need another one)*
- [ ]  Stage 1: Tag as experimental
- [x]  Stage 2: Add test cases covering the main features and obvious corner cases *(I have a few tests, which try all the cases I could think of)*
- [ ]  Stage 2: Add to wikibooks docs *(not sure how to do)*
- [ ]  Stage 2: Add to cheat sheet *(not sure how to do)*
- [x]  Stage 2: Add an example accessible from the Examples menu

khumarahn:
Similar with
```
offset_extrude(5, r=-2.51) square(5,5);
```
there is no extrusion. Maybe this is how it is supposed to work... I expected to see a pointy pyramid, like if I chamfered a cube.

mrkmg:
@khumarahn , you are offsetting too much, so features dissapear, which is why it looks like that. Try increasing the slices.

The following look pretty good to me:

```
offset_extrude(5, r = -2.51, slices = 40)
square(5);

translate([10, 0, 0])
offset_extrude(0.5, r=-0.5, slices = 12)
text("u");
```

To sum up why you see those results: Your reducing the size of the model to nothing at points, and by default offset_extrude only does one offset. If it goes to nothing, there is nothing to extrude up to (your square example). In the text example, something similar is happening, but only part of the "u" disappears, so when it extrudes up, it finds the nearest vertex to connect to.

Example

I have actually tried to address this with "smart" slices, but I am just not clever enough to come up with any sort of solution that is even remotely consistent, or frankly even effective. I figured at this point, some documentation around the function can explain these potential pitfalls. I see it very similar to $fn and $fa and trying to get the smoothness you want.

khumarahn:
@mrkmg Thanks! Using slices does make sense. I see that I should have looked at the examples before trying to do something :-)

khumarahn:
I am trying to do more chamfered text. For some reason,
```
union() {
    // bottom part
    translate([0,0,1.01])
        offset_extrude(height=-1, r=1)
            text("Lorem imsum.");
    // top part
    translate([0,0,1]) 
        offset_extrude(height=1, r=-0.25)
            text("Lorem imsum.");
}
```
gives an error `UI-WARNING: Object may not be a valid 2-manifold and may need repair! location unknown `. Blender highlights a lot of non-manifold edges (see the screenshot). There should not be any tangent faces in the union, this is quite strange.
screenshot-20201114@131237

khumarahn:
To the above: the text can be shortened to `s.` to still keep the error.

khumarahn:
non-manifold edges for the single letter `s`:
```
union() {
    // bottom part
    translate([0,0,1.01])
        offset_extrude(height=-1, r=1)
            text("s");
    // top part
    translate([0,0,1]) 
        offset_extrude(height=1, r=-0.25)
            text("s");
}
```
screenshot-20201114@131915

khumarahn:
Sorry for the spam, but here is an even simpler non-manifold case:
```
offset_extrude(height=-1, r=1) text("s");
```
Here openscad does not complain, but I can see non-manifold edges in blender:
screenshot-20201114@132241

It is not clear if the offset_extrude is working as designed.

nophead:
Looks like it is inflated enough to self intersect at the base. That won't work in OpenSCAD.

khumarahn:
It seems to work correctly with heavily intersecting stuff like
screenshot-20201114@133107
So I imagined it would work on `s` too...

nophead:
Does it work when you union it with a unit cube()? OpenSCAD will draw non-manifold polyhedra but will not operate on them.

nophead:
That example is actually separate letters unioned together rather than a self intersecting shape. Separate objects can be unioned but the S can't can't be unioned with itself.

khumarahn:
union with a cube works correctly (and blender does not show non-manifold edges).
screenshot-20201114@133758

khumarahn:
in the `abc` example individual letter also self-intersect...

mrkmg:
@khumarahn again this is due to not enough slices. If we look at just the bottom part, I can see some really weird connections at just one slice. Specifically, there ends up being parts on the "top" of the bottom piece that are REALLY thin.

only 1 slice

My guess is that because its "connecting" with a near zero width intersection, that's why it ends up being invalid. I increased the slices on the bottom part to 5, and got this much cleaner result. for the bottom portion.

with more slices

All together, looks pretty good once the slices were increased.

complete

I think this feedback in super valuable, and shows off what the primary issue with this feature will be. I think I need to go back to the drawing board on slices. Maybe the default should be more than 1 slice? Or maybe this can still be dealt with in documentation? Either way, @khumarahn has shown that this will be a pain point, and possibly lead people to thinking this feature is broken right out of the gate.

Also, I looked at the "s" example, and I was unable to get an error about non-manifold. Was that error in blender or OpenSCAD? I have never used blender, so I would need some instructions on how to recreate the error.

khumarahn:
For the single `s` the error was in blender, and it happened even with 10 slices...

mrkmg:
@khumarahn I am not sure how to address this issue. I defer to one the more experienced OpenSCAD devs.

khumarahn:
@mrkmg Yeap this thing looks rather complicated.

I can do nearly the same with minkowski and it works well but is very very slow.

TLC123:
> For the single `s` the error was in blender, and it happened even with 10 slices...

If it successfully unions in OpenSCAD (F6) the generation is solid.
Issue is likely to be in some of the many flaws of export.
What format did you use to transfer to Blender?

TLC123:
>  I think at the end of the day, offset_extrude makes it really easy for the user to create non-manifold or invalid models. I welcome any input on fixing these issues if its possible within OpenSCAD and they should be working as designed.

No worse than polyhedron then. As for function its to little surprise that offset_extrude mimics the exact same limitations as plain Offset().  Shapes of varying width vill never end in a single sharp ridge, but neither was that expected as you can not Offset() a shape down to its _Straight skeleton_.  This excellent function enables overlaying user modules of extrusion with chamfer, fillets and other user defined profiles. Proficient warnings in the manual should be sufficient.

khumarahn:
> If it successfully unions in OpenSCAD (F6) the generation is solid.
> Issue is likely to be in some of the many flaws of export.
> What format did you use to transfer to Blender?

I used stl. (Is there a better format for exporting from openscad?)

Also blender complains about non-manifold geometry, but I can't detect visually what is wrong.

khumarahn:
ah, I think I understand a bit about blender's problem. 
screenshot-20201115@102812
See the edge on the screenshot. It is shared between 4 faces, so it is a non-manifold edge in blender's perception, even if for a human eye the geometry looks all right. I wonder if such edges are a problem for openscad.

nophead:
For CGAL objects need to be perfectly manifold, so four edges shared would
throw an exception. However they might be distinct in the infinite
precision number system CGAL uses and only get merged when converted to
Polysets or when exported as floats.

TLC123:
> For the single `s` the error was in blender, and it happened even with 10 slices...

For inward offsets,  a safe estimation for number of slices is: 
The offset per slice should be less than half of the width of thinnest feature of you shape.

Similar for outward offsets, but from  shape+ offset.

@mrkmg  An initial micro slice guarantees manifoldness but is not  an elegant solution that i would suggest.
I recon function is good "as is" for now, sure with room for future automatic analytical optimizations.

khumarahn:
Just an observation: `offset(r=1) text("s");` does not generate the edge which is shared between 4 faces (on stl export) in `offset_extrude(height=-1, r=1) text("s");`:
screenshot-20201115@110846

nophead:
Probably because 2D operations are done with the Clipper library that is
quite robust at removing self intersections. You can clean up imported svgs
by using a very small offset so they are passed through Clipper.

khumarahn:
> You can clean up imported svgs by using a very small offset so they are passed through Clipper.
That is quite useful! thanks for the advice!

mrkmg:
I just pushed a fix for a couple memory leaks (discovered using valgrind) and added a test to make sure slices don't break (I broke slices and the previous tests did not catch the issue).

Next, I think I need to write some docs. **How/Where do I write docs?**

richfelker:
I don't want this to block progress, but if you're doing the offset with a straight skeleton representation, it should be possible to compute the Z displacements at which next change of intersection will happen, which in turn should make it possible to work independently of the user specifying a desired number of slices.

mrkmg:
> I don't want this to block progress, but if you're doing the offset with a straight skeleton representation, it should be possible to compute the Z displacements at which next change of intersection will happen, which in turn should make it possible to work independently of the user specifying a desired number of slices.

I did try to implement a feature like that, but I never got anything to work. With some direction, I would be willing to try again. I do believe it's very usable as it is now though.

khumarahn:
> doing the offset with a straight skeleton representation, it should be possible to compute the Z displacements

@richfelker I recently learned about the straight skeleton representation, and that cgal has it implemented... It should be, eventually, possible to implement the perfect negative offset, independent of the number of slices. Or make the offset with slices following a curve to imitate a fillet. Actually, this seems very straightforward on the cgal side, but I am complete noob in openscad, so maybe it is hard here...

Is there anything like straight skeleton for a perfect positive offset?

khumarahn:
> Is there anything like straight skeleton for a perfect positive offset?

Maybe a straight skeleton of the shape' complement. Hmmmm.

khumarahn:
Hey. I have been fiddling with cgal. I took the polygon the the letter `s` with which I had problems, built a straight skeleton (inner and outer) and constructed faces of offset extrusion:
screenshot-20201116@202718

It works really well and needs no slices, although I don't know how to add top and bottom faces. Also I don't know how to make a chamfer for outer extrusion. And how to read svg.

mrkmg:
@khumarahn can you push your implementation up to a repo? I am VERY curious. Wondering if I could integrate your method with mine.

khumarahn:
@mrkmg

I guess I should set up a repo... All I did so far is this little piece of modified example in cgal:
```
template<class K>
void print_off_straight_skeleton( CGAL::Straight_skeleton_2<K> const& ss )
{
    typedef CGAL::Straight_skeleton_2<K> Ss ;

    typedef typename Ss::Vertex_const_handle     Vertex_const_handle ;
    typedef typename Ss::Vertex_const_iterator   Vertex_const_iterator ;
    typedef typename Ss::Halfedge_const_handle   Halfedge_const_handle ;
    typedef typename Ss::Halfedge_const_iterator Halfedge_const_iterator ;
    typedef typename Ss::Face_const_handle       Face_const_handle ;
    typedef typename Ss::Face_const_iterator     Face_const_iterator ;

    // count vertices for the floor (i.e. skeleton vertices)
    // and build an index for them
    int floor_vertices_number = 0;
    std::vector<int> floor_vertices_index(ss.size_of_vertices(), 0);
    for ( Vertex_const_iterator i = ss.vertices_begin(); i != ss.vertices_end(); ++i ) {
        if (i->is_skeleton()) {
            int d = std::distance(ss.vertices_begin(), i);
            floor_vertices_index[d] = ss.size_of_vertices() + floor_vertices_number;
            floor_vertices_number ++;
        }
    }

    // count bisector edges 
    int bisector_number = 0;
    for ( Halfedge_const_iterator i = ss.halfedges_begin(); i != ss.halfedges_end(); ++i )
    {
        if (i->is_bisector()) {
            bisector_number++;
        }
    }

    std::cout << "OFF\n";
    std::cout << ss.size_of_vertices() + floor_vertices_number
        << " " << ss.size_of_faces() * 2
        << " " << ss.size_of_halfedges() + bisector_number
        << "\n";

    for ( Vertex_const_iterator i = ss.vertices_begin(); i != ss.vertices_end(); ++i )
    {
        std::cout << i->point().x() << " " << i->point().y() << " " << i->time()
            << std::endl;
    }
    for ( Vertex_const_iterator i = ss.vertices_begin(); i != ss.vertices_end(); ++i )
    {
        if (i->is_skeleton()) {
            std::cout << i->point().x() << " " << i->point().y() << " " << 0.0
                << std::endl;
        }
    }

    for ( Face_const_iterator i = ss.faces_begin(); i != ss.faces_end(); ++i )
    {
        int vertex_count = 0;
        Halfedge_const_handle h0 = i->halfedge();
        Halfedge_const_handle h = h0;
        do {
            h = h->next();
            vertex_count++;
        } while (h != h0);
        
        // roof
        std::cout << vertex_count;
        do {
            h = h->next();
            std::cout << " " << std::distance(ss.vertices_begin(), h->vertex());
        } while (h != h0);
        std::cout << "\n";

        // floor
        std::cout << vertex_count;
        do {
            h = h->next();
            if (h->vertex()->is_skeleton()) {
                std::cout << " " << floor_vertices_index[std::distance(ss.vertices_begin(), h->vertex())];
            } else {
                std::cout << " " << std::distance(ss.vertices_begin(), h->vertex());
            }
        } while (h != h0);
        std::cout << "\n";
    }
}
```

khumarahn:
It generates an OFF file which I can open in blender... But for some reason not in openscad. Let me look into setting up a repo with all the stuff to run my example.

khumarahn:
I created a repo: https://github.com/khumarahn/cgal-straight-skeleton

khumarahn:
So far I can generate solid (yay!) objects like this:
screenshot-20201116@221154

These are "chamfers" at 45 degrees (can be scaled and cut to be useful). This should work for polygons with holes too. I generate them as OFF files which I can open in blender but for some reason not in openscad, and I have hard time figuring out why because openscad does not even give an error message.

mrkmg:
I have no clue what's going on there lol. I thought about a possible way to tackle this problem. Working on an implementation now to see if there is any merit to the idea.

khumarahn:
I did a bit more exercise in cgal:
screenshot-20201118@001342
screenshot-20201118@001832
I split the faces of what I had before into slices, thinking that if I assign different height to vertices in different slices I can get something like a fillet.

I don't know what it is I got, but it is not a fillet or anything like it. Well, at least I learned how to subdivide polygons into strips :-)

khumarahn:
Ah, this is more like a fillet.
screenshot-20201118@002716

mrkmg:
@khumarahn, can your code deal with disappearing features gracefully? For example, see this model which my solution looks pretty awful for:

```
offset_extrude(2, delta = -1, slices = 5)
union() {
    square(3.5, true);
    rotate([0, 0, 45])
    square(3, true);
}
```

I have put probably 10 hours into trying things to make this look better. I have done more reading on finding 3d line/plane intersections than pretty much anything else  when working on this, and I am still no closer :-(

example1

khumarahn:
@mrkmg I think it works fine, not quite sure what you mean by disappearing features.
screenshot-20201118@012641

khumarahn:
Things like this we can build directly from the straight skeleton, in what feels like 20 lines of code:
screenshot-20201118@013009

mrkmg:
@khumarahn, does straight skeleton have the ability to deal with the other modes of offsetting? The offset used in the Clipper lib allows for 3 different ways to offset; by radius, by delta, or delta with chamfer.

Looking at the examples you have shown so far seem to all be "delta.

khumarahn:
I am not sure if the straight skeleton can be used for other kinds of offsetting. Maybe not.... Also the method in cgal it works for simple polygons with holes, so we would need extra boolean work for unions of disjoint polygons.

But I don't know much: I am a noob with zero experience in cgal or openscad, I only learned about the straight skeleton a couple of days ago. I have seen a number of scientific papers about using straight skeleton for offsets and chamfers, but I did not really read them.

mrkmg:
Hmmm. I think the value of using radius outweighs the occasional poor looking result, but I will leave that decision up the maintainers of OpenSCAD. Maybe we use the current clipper based method for radius and chamfer, but the straight skeleton method for delta offsets.

TLC123:
> Things like this we can build directly from the straight skeleton, in what feels like 20 lines of code:
> screenshot-20201118@013009

I note that  your experiments all end in sharp points or ridges. Can you demonstrate how your method looks with smaller offsets . 
This important for inserting release angle in mold making. For example a 5x5 square extrude 5 offset .1  ?

khumarahn:
> Hmmm. I think the value of using radius outweighs the occasional poor looking result, but I will leave that decision up the maintainers of OpenSCAD. Maybe we use the current clipper based method for radius and chamfer, but the straight skeleton method for delta offsets.

I would be happy to see the inner 45 degrees pyramid in openscad (like https://github.com/openscad/openscad/pull/2079#issuecomment-729318091). It is a simple operation, but if scaled, truncated or subtracted, it can give lots of very good looking chamfers. And we don't need anything fancy for it - its structure is fully described by the straight skeleton, we just need to read it, and with cgal it took me less than a page of code.

khumarahn:
> Hmmm. I think the value of using radius outweighs the occasional poor looking result, but I will leave that decision up the maintainers of OpenSCAD. Maybe we use the current clipper based method for radius and chamfer, but the straight skeleton method for delta offsets.

@mrkmg I am fine with having to do slices to reduce the artifacts. I started fiddling with straight skeleton when I could not use your code because it gave non-manifold geometry (at least on export from openscad, so the problem may be elsewhere, but in the end it is still not usable).

P.S.: I missed you suggestion to go into discord. I am up for it.

khumarahn:
I learned when the straight skeleton may give potentially undesired results for inside offsets:
screenshot-20201118@114551
Looking at [Skeletal Structures for Modeling Complex Chamfers and Fillets](https://www.palfrader.org/research/2018/Held,%20Palfrader%20-%20Skeletal%20Structures%20for%20Modeling%20Complex%20Chamfers%20and%20Fillets,%20CADC%202018%20[HePa18b]%20-%20draft.pdf)
screenshot-20201118@114753
it may be better to use the contours from a voronoi diagram instead. (Which is a bit more complicated because they may be circles as well as lines.)

TLC123:
Ok how about design this as expandable parameters.
That way a version based on clipper offset could be launched soon. 

offset_extrude(5, r = -2/delta =6, method = clipper/skeleton/voronoi )

khumarahn:
Maybe also have the pyramid for straight skeleton / voronoi skeleton separately? Say `straight_skeleton_pyramid() square(3);` would generate a pyramid with 45 degrees slopes which can later be cut or scaled to purpose, and used internally to generate offsets.

With voronoi, the generation of offsets is more technical than with the straight skeleton. I looked a bit at the interface in cgal - it would take quite some time to understand what is going on. The idea is simple and this is of course doable, but may be technical.

khumarahn:
I just thought... 
screenshot-20201118@155231

Still using the straight skeleton, it is very easy to get "round" corners if we smoothen the corners of the base polygon. This leads to a change of shape of the base polygon, but it can be very small if the corners are turned into (piecewise linear) arcs of very small radius. Like if we apply an infinitesimal clipper's `offset` to the original polygon.

The advantage of this method is that it is easy to implement (I mean for connected polygons and polygons with holes; for disjoint shapes we would have to also use boolean). And for things like text it gives perfectly satisfactory results.

richfelker:
Editing the base polygon is undesirable because it makes it not fit with existing extrusions of the same polygon. There are lots of applications where you want the base of one `offset_extrude` to match up perfectly with the base of another `linear_extrude` or `rotate_extrude`.

You can use straight skeletons to do rounded corners and stuff (IIRC WP even has some examples), and I think it's pretty much equivalent to what you said, but applying the change only to slices with nonzero offset and leaving the base polygon alone.

TLC123:
@khumarahn 
As @richfelker mentions: It make most sense that offset_extrude behaves like offset so part can match up.
But  i much support the idea of, and see the use-fullness of a "pyramidal extrusion". But implemented as a separate feature.


> Say `straight_skeleton_pyramid() square(3);` 
Alternative?
roof ( ) square(3);
hip_roof ( ) square(3);
roof_extrude () square(3);

khumarahn:
screenshot-20201118@173553
screenshot-20201118@173827
I did a crude algorithm which smoothens corners of a polygon, and applied the pyramid extrusion. Just to see what it looks like. It does give nicely rounded offsets.

I agree with @richfelker that is it better to keep the polygon as it is... This would require a deeper understanding of cgal internals though.

@TLC123 any working interface is fine with me... Also I have the pyramid with slices, where I can assign any height to each slice, using a function like `z -> z^2` or whatever (like in the wavy examples before).  This may be useful too, if there is a way to pass a function.

mrkmg:
I think we are getting off track, and trying to do too much. This PR has been open for like 2-3 years now, in a state which is mostly functional, and of great use in TONS of circumstances. I personally have used my build for 50+ models, all 3d printed. I know it's not perfect, but we could sit here for another 2-3 years thinking of cases it does not perform perfectly for. 

My proposal is this: let's refine what I have already, using the clipper method, get this merged and ready for launch, then discuss further enhancements in another Issue/PR. I really like the skeleton based approach, and think it could be a useful, but I do not understand polygons/3d space stuff well enough to integrate it. I just wanted this feature for my needs, and got this solution to work.

I don't want to come off brash, but I have refactored this PR 3 or 4 times now to bring it up to date with the current master, and resolved nearly all invalid models which I considered sane. I am fine continuing to use this feature all by myself, on my own personal build. If we don't cross the finish line this time, I likely won't come back to this and will just continue to use my version for my own needs.

Again, I am not trying to be rude or anything, I just do not have the drive to continue working on this beyond what I have already put into it. I love OpenSCAD, and use it all the time for my 3D printing models, but my development skills are mostly web based. I don't use c++, or 3d space polys/etc, professionally or personally. My work is mostly web-based C# and JS.

I hope y'all understand my position.

Again to reiterate from the original TODO list, I think the next step is to make documentation around what I already have. I just need direction on where to put that documentation.

TLC123:
>   I just need direction on where to put that documentation.

 Im shooting from my ass here but wouldnt it just be to edit the wikibook. Starting
 in https://en.wikibooks.org/wiki/OpenSCAD_User_Manual#Work_in_progress  and moving it up when merged.
Pretty much just copy paste the basics off linear_extrude. But there is my opinion no obligation to make it anywhere as verbose to check that box.


```
Offset Extrude
Offset Extrude is a operation that takes a 2D object as input and generates a 3D object as a result. See Linear Extrusion for more.
 
Offset Extrude provides a offset parameter so that the object can be expanded or contracted over the extent of the extrusion, allowing extrusions to be flared inward or outward along the extrusion. See Offset for more information

Usage
offset_extrude(height , r=offset,   slices = 20 ) {...}

Mesh Refinement
The slices parameter defines the number of intermediate points along the Z axis of the extrusion. 
```

t-paul:
@TLC123 yep, that would be the way to go.

If there are known limitations, it would be good to mention those too.

mrkmg:
Working on documentation now

khumarahn:
@mrkmg I understand your pain very well :-)

Sorry for all my spam. I came here thinking that your code will be useful to my project, and only started looking for ways to do offset extrusion differently when it did not work....

I get plenty of invalid geometry errors on export in the simplest cases, most notably in the letter `s` as above:
screenshot-20201114@132241
That edge which has four adjacent faces, even if it is a merge of two distinct (but coinciding) edges which happens only on export, should not be there in the first place: it is not a part of the offset of the base polygon. I'd say that things like this need investigation before this PR merged.

I, myself, do not have any knowledge of the internals of openscad, and as a user I only wrote very basic stuff. I would like a version of offset extrusion which I can use, but it would be hard to implement for openscad without help.

mrkmg:
I would contend that the letter s is not a simple case, as the offset is causing an overlap, which is causing the weird issue you are getting. If you offset a little less there, by lets say .9 instead of one, which prevents the overlap it should produce a good, manifold model.

richfelker:
> I think we are getting off track, and trying to do too much. This PR has been open for like 2-3 years now, ...

Agree 100%. That's what I meant by not wanting the ideas I suggested to delay/derail this. My motivation in posting them was just to keep in mind whether the current proposal is compatible with that kind of future improvement without significantly altering the interpretation of existing `.scad` files. Not to suggest your work should be redone with different algorithms.

richfelker:
As for the manifold errors/invalid geometry, I'd actually *rather* they remain errors now (restrictions on how you can use the new feature) than produce working output that's "wrong", since the former is fixable in the future without breaking backwards compatibility and the latter isn't (someone will be depending on whatever the first "wrong"-but-working output is).

mrkmg:
I have created a draft of the documentation.

https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/WIP/Offset_Extrude

I plan to make some screenshots, and add some examples of how to use it.

richfelker:
The documentation looks mostly good, but it doesn't clearly explain what the `r`/`delta` value is. I assume that's the offset *at the final `height`*, and linear-interpolated up to there. But it's probably also different for `center=true` - in that case does it reach the full offset in one direction (or both?) or does it go halfway in both directions?

thehans:
@richfelker the center parameter should be analogous to `linear_extrude`'s center parameter: its simply translating the result downwards (centered about the Z axis), but has no other effect on the height or shape of the model.

mrkmg:
@richfelker I will expand the definitions there. I kinda of left the explanation of r, delta, chamfer to the already existing offset module. 

And yes, that is how the center works. I will expand that definition as well.

adrianVmariano:
So this is very interesting to see, as I have implemented this function (or something like it) in user space.  I'm curious about two things:  Can I specify a different offset for each slice?  This is crucial because I mainly use it to create rounded edges.  If the offset has to be the same for each slice it's very limited.  And secondly, what is the final method for aligning vertices when the offset changes the vertex count?  It looks like this has been an issue.  In my code the offset tracks where the vertices came from, so the alignment is always "correct".  But this may not be possible depending on how offset works. 

My user space implementation is in the BOSL2 library and examples of its behavior can be seen at this wiki:

https://github.com/revarbat/BOSL2/wiki/rounding.scad#offset_sweep

mrkmg:
You can use userspace and a for-loop to offset each piece you need. At this point, you cannot provide offset per slice inside offset_extrude.

t-paul:
My suggestion would be to try getting this merged with the current feature set. A great next step would be to allow passing a function for offset calculation like
```
offset_extrude(90, r = function(h) 2 * sin(8 * h)) circle(5);
```
I'm not yet sure how challenging that will be, but it would be the natural extension of the recent changes to support higher order functions.

richfelker:
OpenSCAD does not have first-class functions so there's no way to pass a function to a module. You could design something where you pass an array of z-displacement, offset pairs, but this isn't necessary; instead just union a sequence of the existing piecewise-linear `offset_extrude` module instances.

t-paul:
No need for workarounds, function literals exist for about a year in the development snapshot (still as experimental feature): https://github.com/openscad/openscad/pull/3077 - nice library around that feature by @thehans: https://github.com/thehans/funcutils

adrianVmariano:
> OpenSCAD does not have first-class functions so there's no way to pass a function to a module. You could design something where you pass an array of z-displacement, offset pairs, but this isn't necessary; instead just union a sequence of the existing piecewise-linear `offset_extrude` module instances.

If you do this naively you're going to have a sequence of pieces that share a face, and taking the union of those generally produces problems in OpenSCAD, so it's not so simple.

richfelker:
In my experience, that's actually not a problem with exact matches of 2D extrusion faces.

mrkmg:
@t-paul, the functional parameters were not available when I started this. I would be willing to support this. Can you point me to an example of how another module deals with this?

I assume I would need to add something to the `OffsetExtrudeModule`, and then the `GeometryEvaluator`. If you can point me towards an existing implementation I bet I can muddle through from there.

khumarahn:
Hi All again,

I almost stopped spamming here, but before I go:
screenshot-20201119@133024
screenshot-20201119@133043

I implemented a "roof" function, with syntax suggestion of @TLC123. It alone gives nice and consistent chamfers. Blender does not find non-manifold problems.

I created a PR and will continue in https://github.com/openscad/openscad/pull/3486 This PR is very elementary as of now, I only had to write a page of code, most of which is type declarations. It is incomplete though... to complete it I need some advice on hanldling of 2d polygons, which I don't know how to handle. I will expand in my PR.

khumarahn:
Here is what I don't know how to do. The input to the algorithm is `Polygon2d poly`. Then `poly.outlines()` contains outlines of all polygons in the shape, but does not seem to contain information of which outlines are holes and which are not.

I need to break this into an array of disjoint polygons with holes.

richfelker:
Is this still moving forward? If not, what's it blocked on?

bmuessig:
Good job, @khumarahn!

richfelker:
Ping. What's still blocking this? Is there anything I can do to help it move forward?

larsrc:
Also curious to hear about the progress here. I've been wanting it several times, what prompted me to find the bug was wanting to put a sharper edge on https://www.thingiverse.com/thing:3962122/files. I hope this goes through, it would be extremely useful!

your-friend-alice:
I'd love to see this finished and merged!

one (very slow) workaround if you don't mind being restricted to a radius outline (and not delta or miter):

this would be more precise and possibly cut the time in half if minkowski allowed 2d objects (instead of very thin 3d objects)

```
minkowski() {
    linear_extrude(0.1) text("test");
    cylinder(d1=4, d2=0, h=4);
}
```

mrkmg:
Unless someone else wants to take this over, I recommend closing this. After years I have lost interest to work in this project. The method used here is sub-par, and while it does produce good results for many many geometries, there are still many it will bork on.

I do not have the skill or know how to take it to the finish line.

If anyone wants to give it a try, I would be willing to jump on a voice chat or something to explain what I was trying to achieve.

underhood:
sad to hear that but understandable.... spent a long time trying to achieve sth. like this, in the end, went with not so straightforward solution using `minkovski`
I do believe however that something like this should be very easy to do for users (as it would be with your patch).
That being said take into account I am OpenSCAD newbie.

t-paul:
@khumarahn would you be interested in combining the `roof()` work with this `offset_extrude()`? Earlier discussions seemed to indicate the straight skeleton data might help with the tricky issues @mrkmg came across.

khumarahn:
> @khumarahn would you be interested in combining the `roof()` work with this `offset_extrude()`? Earlier discussions seemed to indicate the straight skeleton data might help with the tricky issues @mrkmg came across.

I'm up for it. Though I don't exactly understand how you'd do that. At present, a significant part of functionality of `offset_extrude` can be done with `roof` and a few unions/intersections. (@underhood this is merged into the experimental branch, and there is a chance it may help you.)

khumarahn:
> Tried using `roof` suggested here and then `difference` but that produced suboptimal results.

@underhood can you give more details, what exactly are you trying to do?

Issue #1870, opened 2016-11-23
dependency missing in documentation/scripts libzip-dev


ert485:
I needed to run " sudo apt install libzip-dev " before running qmake. Error was:

> qmake openscad.pro
> Project MESSAGE: If you're building a development binary, consider adding CONFIG+=experimental
> Project MESSAGE: Using QOpenGLWidget
> Project MESSAGE: Using C++11
> Package libzip was not found in the pkg-config search path.
> Perhaps you should add the directory containing `libzip.pc'
> to the PKG_CONFIG_PATH environment variable
> No package 'libzip' found
> Package libzip was not found in the pkg-config search path.
> Perhaps you should add the directory containing `libzip.pc'
> to the PKG_CONFIG_PATH environment variable
> No package 'libzip' found

t-paul:
It's optional, but yes, it should be added to the documentation...

ert485:
Ah I was under the impression that it failed. Did I mention I'm new to this haha

kintel:
A nicer warning message wouldn't hurt either, e.g. "Optional libzip library not found - reading compressed AMF files disabled"

t-paul:
The warning is from pkg-config, not sure it's possible to change that without hacks (especially with qmake). I do want to add the overview output at the end though.

thehans:
Just came across this same issue, overview output would be nice as its really unclear that it apparently succeeded.

thehans:
I think it wouldn't hurt to add libzip-dev to scripts/uni-get-dependencies.sh, just to avoid this message.

PR #2086, opened 2017-08-16
add missing debian dependency: libzip-dev

hoijui:
This was found on Ubuntu 16.04.3 LTS.

If i remember right, this dependency system is deprecated anyway, but.. maybe it will help someone until it is actually replaced, and gain some visibility as a pull request.


thehans:
Relevant to #1870

hoijui:
:-)
thank you!

Issue #2087, opened 2017-08-18
OSX dependency build instructions are a bit unclear


thepelkus-too:
I feel silly for making this mistake, but when I saw the *From Source* option numbered 1 in the README and then saw *Homebrew* also numbered 1, I thought that both were required instead of realizing that they were alternatives to each other. It would be helpful to clarify that you only need one of the three options for making sure dependencies are installed.

I've got a commit with the suggested changes at https://github.com/thepelkus-too/openscad/commit/f368d75839464c9eb3c325c69492dc32b0471117; I'd be happy to open a PR if you agree that they would be useful.

kintel:
PR would be nice!

PR #2088, opened 2017-08-18
Make the dependency installation options clearer

thepelkus-too:
Make sure that the different options are enumerated correctly, and that the text preceding them notes that only one of the three steps is necessary. Addresses #2087


kintel:
thx!

Issue #2091, opened 2017-08-19
[TODO] Wireframe Width


deckar01:
Add a preference for customizing the wireframe line thickness.

- [x] [MVP] Use a 1px line thickness for wireframes
- [ ] [PR] Add a preference for setting the line thickness for wireframes
- [x] :cake: Calculate the default thickness based on the display density

deckar01:
I would like to get a hard coded value of 1px working in my local build since I have been using wireframes heavily when checking if union operations had any precision problems. The chunky lines make it hard to spot errors without zooming in closely.

I would be open to finishing out the TODO with a preference so I can start getting my feet wet with this code base.

kintel:
1px may also be a better default value

deckar01:
There are comments already in `Polyhedron::draw()` for making the lines and nodes smaller. After comparing a high density display to an standard display I realized that there is not a one size fits all pixel value. The current default actually looks OK on a retina display, but on an old Air it is way too thick.

I added a 3rd item to the description for calculating the default using the display density.

deckar01:
I changed the line width to 1px and the point size to 3px, then multiplied them by `QScreen::devicePixelRatio()`.

Retina:

<img width="1562" alt="screen shot 2017-08-19 at 9 47 46 pm" src="https://user-images.githubusercontent.com/3108007/29491696-b67d8ab4-8528-11e7-82e4-0580027c1a6b.png">

Air:

![screen shot 2017-08-19 at 9 47 33 pm](https://user-images.githubusercontent.com/3108007/29491697-be1b1304-8528-11e7-8525-1d08acd0b8a3.png)

I think this looks good on x1 and x2 devices, and it appears to match the weight of the axis lines.

It looks like I should probably be using `QGLView::getDPI()` since it is behind a version flag, but I'm not quite sure how to use it in `OGL_helper.h` yet.

deckar01:
I tried passing the DPI from `GLView` down to `Polyhedron::draw()`, but got stuck at `polyset->render_*()`. I suspect `polyset` contains a collection of `Polyhedron`, but relevant implementation was external.

It was a lot simpler to just probe the `QScreen` in `OGL::Polyhedron::get_dpi()` than try to pass it through the stack anyway. Does this sound like an acceptable solution?

There are a few other places that are using hard coded line widths instead of DPI adjusted values. I might investigate those and create new issues for noticeably thick lines on standard displays.

MichaelAtOz:
[...]

PR #2096, opened 2017-08-20
DPI Normalized Wireframe Rendering

deckar01:
Use the screen's DPI when calculating the line width and point size of wireframes.

See #2091 

<details>
<summary>1x DPI Wireframe</summary>

screen shot 2017-08-19 at 9 47 33 pm

> Note: The "DPR" value in this screenshot was for debugging and is not included in the PR.
</details>

<details>
<summary>2x DPI Wireframe</summary>

screen shot 2017-08-19 at 9 47 33 pm

> Note: The "DPR" value in this screenshot was for debugging and is not included in the PR.
</details>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,4 @@ src/*.org
 src/*.old
 src/*.debug
 /Testing
+.DS_Store
--- a/src/OGL_helper.h
+++ b/src/OGL_helper.h
@@ -27,7 +27,7 @@
 #include "system-gl.h"
 #include <cstdlib>
 #include <QScreen>
-#include <QGuiApplication>
+#include <QApplication>
 
 // Overridden in CGAL_renderer
 /*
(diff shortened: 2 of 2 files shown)


kintel:
Looks reasonable. Thx!

deckar01:
It looks like the Travis build is failing due to `#include <QScreen>`. I don't fully understand why the QT versions are so different between Travis and my local machine. I apologize for the inconvenience.

A backwards compatible namespace could be used to provide `QScreen`, but this process is a bit opaque to me. I think a safer approach might be to put the includes behind the same QT version check that `get_dpi()` uses to make the dependency more explicit.

kintel:
I forgot to review this properly; `OGL_helper.h` needs to be independent of Qt; we only use Qt for the UI, not for the underlying engine. ..so we need another mechanism for passing the DPI into the renderer.

I'll revert the commit for now.

deckar01:
The renderer should probably encapsulate the DPI logic. The problem is that it doesn't encapsulate the actual draw operations. It's kind of weird that a renderer class delegates to a geometry class to draw itself.

It seems more natural to have `Renderer::draw(Polyhedron polyhedron)` than to have `Polyhedron::draw(float dpi)`.

I'm not sure if that is a flaw in the rendering pipeline of this library or a quirk of the CGAL geometry hierarchy.

kintel:
There is a lot of room for refactoring of our renderer. I agree that some of the logic is inverted.
In terms of `OGL_helpers`, this is a helper class from CGAL itself that we had to slightly modify. I would either try to minimize how much we touch that one, or rewrite it completely to fit our needs.

PR #2097, opened 2017-08-21
Revert "DPI Normalized Wireframe Rendering"

kintel:
Reverts openscad/openscad#2096 due to Qt dependency issues.


deckar01:
I though that seemed too easy. 😛

PR #2098, opened 2017-08-21
CMake: require QT version 5.4 (minimum) [fix]

hoijui:
At least because `QOpenGLWidget` was added only in that version:
http://doc.qt.io/qt-5/qopenglwidget.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,7 +92,9 @@ if(NOT HEADLESS)
     set(CMAKE_POSITION_INDEPENDENT_CODE ON)
   endif()
 
-  add_definitions(-DUSE_QOPENGLWIDGET)
+  if ("${Qt5_VERSION}" VERSION_GREATER_EQUAL "5.4")
+    add_definitions(-DUSE_QOPENGLWIDGET)
+  endif()
 
   find_package(Qt5QScintilla 2.8.0 REQUIRED QUIET)
   message(STATUS "QScintilla: ${QT5QSCINTILLA_VERSION_STRING}")


kintel:
@t-paul @donbright Does this mess up any Linux distro compatibility to your knowledge?

hoijui:
i found this only, because i am on Trisquel linux, which is old (based on Ubuntu 14.04). it comes with cmake that is also way too old (version 2.8.something). so i manually downloaded and installed latest cmake. and then i figured out that my distro has a too old QT (5.2.1 or so it was).
i imagine, that systems with sufficiently new cmake also come with sufficiently new QT.
this is just a fail-fast mechanism, so one does not have to go through all the hassle and find out why this particular include is not found.

hoijui:
ahaa...
i can also try to incorporate tha tinstead then, sounds like it would
make more sense.. and i cna alos test it on my system.

2017-08-21 16:19 GMT+02:00, Marius Kintel <notifications@github.com>:

donbright:
Yes i think this might throw some people off because the old qmake system was very forgiving about detecting qt4 or qt5 and working either way automatically without the user having to think about it. 

Off the top of my head the easiest way to find which versions of QT ship with various linux distros is to look at distrowatch.com, i have studied a few below, looks like most switched within the last 2 years or so. 

https://distrowatch.com/table.php?distribution=mint (>=18.2)

https://distrowatch.com/table.php?distribution=fedora  (>=22)

https://distrowatch.com/table.php?distribution=ubuntu (>=15.10)

https://distrowatch.com/table.php?distribution=freebsd (>=11.1)

https://distrowatch.com/table.php?distribution=netbsd (>=6)

https://distrowatch.com/table.php?distribution=centos (**fail**, they are still on qt4)

https://distrowatch.com/table.php?distribution=slackware (**fail**, they are still on qt4)

https://distrowatch.com/table.php?distribution=mageia (>=6)

https://distrowatch.com/table.php?distribution=manjaro (>=15.2)

https://distrowatch.com/table.php?distribution=solus (>=1.2.1)

https://distrowatch.com/table.php?distribution=opensuse (>=42.1)

https://distrowatch.com/table.php?distribution=elementary (>=0.4)

By the way, this is one situation where the Nix package system might in the future be the best solution,  if they get their OpenGL driver situation sorted out ( https://github.com/NixOS/nixpkgs/issues/9415 ) I have built openscad with qt5 on ubuntu 14 and 12 using nix-shell and it was incredibly simple. It took a lot less lines than this text im writing right now. Only problem is the GL driver itself.

kintel:
One way of solving this could be to declare our CMake build system to _not_ support legacy systems and require Qt 5.4 for anyone wanting to use it.

hoijui:
i changed the pull request.
it now does the same as qmake:
using the -DUSE_QOPENGLWIDGET define only for QT >= 5.4

i also added a fix commit required to make the CMake generated binary not error out at runtime because of "bad rounding".

the old code - failing completely with QT < 5.4 - is available under tag `cmakeRequireQt54` on my repo now.

kintel:
Could you take out your last commit? Those things are platform/compiler dependent and probably belongs to a separate pull request.

hoijui:
ouhh yeah, makes sense.. done!

kintel:
thx!

PR #2099, opened 2017-08-22
[WPI] DPI Normalized Wireframe Rendering

deckar01:
Use the screen's DPI when calculating the line width and point size of wireframes.

See #2091 

This improves upon #2096 by passing the DPI from `MainWindow::qglview::getDPI()` to `CGALRenderer::setDPI()` to `Polyhedron::set_dpi()` to address the concerns raised in https://github.com/openscad/openscad/pull/2096#issuecomment-323648424.

In response to https://github.com/openscad/openscad/pull/2096#issuecomment-323868456 I opted to "minimize how much we touch [OGL_helpers]", because "rewriting it completely" seems disproportionate to the scope of the issue (and my familiarity with the code base :wink:).

<details>
<summary>1x DPI Wireframe</summary>

screen shot 2017-08-19 at 9 47 33 pm

> Note: The "DPR" value in this screenshot was for debugging and is not included in the PR.
</details>

<details>
<summary>2x DPI Wireframe</summary>

screen shot 2017-08-19 at 9 47 33 pm

> Note: The "DPR" value in this screenshot was for debugging and is not included in the PR.
</details>

### DIFF
--- a/src/CGALRenderer.cc
+++ b/src/CGALRenderer.cc
@@ -44,6 +44,7 @@
 
 CGALRenderer::CGALRenderer(shared_ptr<const class Geometry> geom)
 {
+	this->dpi = 1;
 	if (auto ps = dynamic_pointer_cast<const PolySet>(geom)) {
 		assert(ps->getDimension() == 3);
 		// We need to tessellate here, in case the generated PolySet contains concave polygons
@@ -66,6 +67,11 @@ CGALRenderer::CGALRenderer(shared_ptr<const class Geometry> geom)
 
 CGALRenderer::~CGALRenderer()
 {
+}
+
+void CGALRenderer::setDPI(float dpi)
+{
+	this->dpi = dpi;
 }
 
 shared_ptr<class CGAL_OGL_Polyhedron> CGALRenderer::getPolyhedron() const
--- a/src/CGALRenderer.cc
+++ b/src/CGALRenderer.cc
@@ -87,6 +87,7 @@ void CGALRenderer::buildPolyhedron() const
 	CGAL::OGL::Nef3_Converter<CGAL_Nef_polyhedron3>::convert_to_OGLPolyhedron(*this->N->p3, this->polyhedron.get());
 	// CGAL_NEF3_MARKED_FACET_COLOR <- CGAL_FACE_BACK_COLOR
 	// CGAL_NEF3_UNMARKED_FACET_COLOR <- CGAL_FACE_FRONT_COLOR
+	this->polyhedron->set_dpi(this->dpi);
 	this->polyhedron->init();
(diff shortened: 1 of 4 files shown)


kintel:
Better :)

Two questions:
* What should we do with offscreen-rendering? https://github.com/openscad/openscad/blob/master/src/export_png.cc#L32
* How should we handle preview rendering with visible edges?

deckar01:
My intuition is that since offscreen rendering uses qglview's dimensions then it should use its DPI also. It would be nice if it had a pop up like photoshop that made the dimensions and DPI configurable. It looks like this would need to be added, because it is defaulting to 1 right now. I should search for any other uses of `CGALRenderer` that need the same consideration.

I think all rendered lines should be normalized (several already were), but maybe not in this PR. Maybe `CGALRenderer::setDPI()` should be pushed down to `Renderer` now in preparation of those changes though.

kintel:
By offscreen rendering, I was referring to [OffscreenView](https://github.com/openscad/openscad/blob/master/src/OffscreenView.h), which doesn't use Qt at all.

deckar01:
The `export_png*()` functions are used by the Qt window and the CLI.

Since the camera size is based on the Qt window's size, I think it makes sense to use the Qt window's DPI. From a UX perspective this operation feels like an export of the render window, so my expectation is that the PNG output will exactly match the render view.

Since the camera size can be configured in the CLI, then the CLI should probably have an option to configure DPI as well. Most of the ground work for handling the DPI in `export_png*()` will already be laid, so it shouldn't be too much extra work to give the CLI control over this setting.

kintel:
Agreed - sounds like a decent solution. I think we can ignore the CLI for now (use default DPI) as we don't have a wireframe export option.

deckar01:
I was wrong about `MainWindow` calling `export_png*()`. It actually grab's the frame directly from `qglview` buffer. We get the desired behavior for free, because `MainWindow` is already rendering a DPI corrected wireframe to the `qglview`. Sorry for the confusion, I should have just tested the PNG export before grepping for the function usage (early in the morning 😴). It is already working the way you suggested.

https://github.com/openscad/openscad/blob/25cde91/src/mainwin.cc#L2263

Regarding my earlier comment "`CGALRenderer::setDPI()` should be pushed down to `Renderer`", does this sound like a desirable change? I don't see any harm in letting the DPI default to 1 in all the renderers if they are not using it yet. When the preview renderer's edge's get normalized it will probably need to happen, so doing it now would save a little bit of future churn.

kintel:
Ref. the "dashed line" issue for 1px lines in preview mode; should we default this to 2px?

We could probably render wireframes better in preview but that's likely more work.

PR #2107, opened 2017-09-10
Fix MacOS First Run

### LINKED ISSUE #2089: OSX Snapshots Fail to Open
Every time I download a new developer snapshot on OSX it fails to launch from finder. I have tried allowing the unidentified developer in the security settings, but it just doesn't open.

The work around is always to launch it from the command line:

```sh
open -a /Applications/OpenSCAD
```

Once I do that I can open it from finder/spotlight just fine.

---

**Edit:**

Here are the screenshots of the security checks that are failing:

1. Double click on OpenSCAD.app
screen shot 2017-08-19 at 12 12 02 pm

2. System Preferences > Security & Privacy > General > Open Anyway
screen shot 2017-08-19 at 12 12 16 pm

3. Open
screen shot 2017-08-19 at 12 12 24 pm

Nothing seems to happen and the [log entry](https://github.com/openscad/openscad/issues/2089#issuecomment-323501003) mentioned below occurs.

deckar01:
Handle the `-psn_*` argument that Finder on OSX sends to apps.

> It looks like Finder appends a process ID argument.
> 
> https://stackoverflow.com/questions/10242115/os-x-strange-psn-command-line-parameter-when-launched-from-finder
> 
> FreeSCAD had the same problem and posted a workaround in this thread.
> 
> https://forum.freecadweb.org/viewtopic.php?t=4447

https://github.com/openscad/openscad/issues/2089#issuecomment-328371812

Fixes #2089.

### DIFF
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -784,6 +784,16 @@ int gui(const vector<string> &inputFiles, const fs::path &original_path, int arg
 }
 #endif // OPENSCAD_QTGUI
 
+std::pair<string, string> customSyntax(const string& s)
+{
+#if defined(Q_OS_MACX)
+	if (s.find("-psn_") == 0)
+		return {"psn", s.substr(5)};
+#endif
+
+	return {};
+}
+
 int main(int argc, char **argv)
 {
 	int rc = 0;
@@ -832,6 +842,9 @@ int main(int argc, char **argv)
 		("d,d", po::value<string>(), "deps-file")
 		("m,m", po::value<string>(), "makefile")
 		("D,D", po::value<vector<string>>(), "var=val")
+#ifdef Q_OS_MACX
+		("psn", po::value<string>(), "process serial number")
+#endif
 #ifdef ENABLE_EXPERIMENTAL
 		("enable", po::value<vector<string>>(), "enable experimental features")
 #endif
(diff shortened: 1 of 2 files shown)


src/openscad.cc
```diff
@@ -784,6 +784,16 @@ int gui(const vector<string> &inputFiles, const fs::path &original_path, int arg
...
 #endif // OPENSCAD_QTGUI
 
+std::pair<string, string> customSyntax(const string& s)
+{
+#if defined(Q_OS_MACX)
+	if (s.find("-psn_") == 0)
+		return make_pair(string("psn"), s.substr(5));
+#endif
+
+  return make_pair(string(), string());
```
kintel: Tiny suggestion: `return {};` should do the job here.

src/openscad.cc
```diff
@@ -784,6 +784,16 @@ int gui(const vector<string> &inputFiles, const fs::path &original_path, int arg
 }
 #endif // OPENSCAD_QTGUI
 
+std::pair<string, string> customSyntax(const string& s)
+{
+#if defined(Q_OS_MACX)
+	if (s.find("-psn_") == 0)
+		return make_pair(string("psn"), s.substr(5));
```
kintel: Tiny suggestion: `return {"psn", s.substr(5)};` should do the job here.

deckar01: I didn't actually know what `make_pair` did, I was just following the example. I assumed it was some kind of special struct. 😆

src/openscad.cc
deckar01: Thanks for your insight.

src/openscad.cc
kintel: `make_pair()` simply constructs a pair using ADL (argument-dependent lookup). C++11 is capable of initializing this directly based on the return type.

PR #2092, opened 2017-08-20
Added --view option #2032

kintel:
Added --view option taking a comma-separated list of options as parameter.
This should solve most of #2032.

Example:
`openscad test.scad --view axes,scaleMarkers -o out.png`

TODO:
* [ ] Consider further low hanging fruits in this domain
* [ ] Tests
* [ ] Doc

PR #2116, opened 2017-09-28
Line art export

cgspeck:
Thank you so much for this incredibly useful program. 

This branch adds a colour scheme (`LineArt`) and a switch (`--showedges`, which must be used with `--render` and a png output file) which allows you to output your models in grey-scale on a white background with each object's edges showing.

I hacked this together because I needed to export my models in high resolution so that I could generate scaled paper plans from them.

It gives the CLI the ability to generate something approaching [AutoCad's Flatshot](http://blog.ketiv.com/autocad-2016-hit-me-with-your-flatshot) (with obscured lines hidden) function.

@kintel this may conflict with the work you have done in #2092, though I note your  branch doesn't envisage offering `showedges` as one of the options (and there is also the question about a colour scheme).

Please let me know if you would consider accepting this PR, and what additional work would be required to make this happen.

Alternately I'd be happy to extend #2092 if you indicate what changes you'd like to see.


kintel:
Thanks for your PR! What you've implemented definitely sounds like one of the "low hanging fruits" I  referred to in #2092. I have a few suggestions:
* Could you base this PR on #2092, and add `showedges` as a view parameter rather than a top-level parameter?
* Could you easily make this work without the `--render` option? i.e. have similar behavior for previews and "throwntogether" rendering mode?
* Are you aware that we support user-space color schemes, so this doesn't need to go into the main app. If you still think it's a generally useful scheme, it should probably be PR'ed separately

cgspeck:
Thanks, I have added showEdges to your grid-export branch and opened a PR #2119 against that.

PR #2119, opened 2017-10-05
Adding showedges option

cgspeck:
Add an additional view option, has been rebased off master.


src/GLView.cc
```diff
@@ -171,6 +171,8 @@ void GLView::paintGL()
 	// mark the scale along the axis lines
 	if (showaxes && showscale) GLView::showScalemarkers(axescolor);
 
+	if (showedges) GLView::showedges = true;
```
kintel: What does this line do?

cgspeck: Thanks, removed :-)

kintel:
Merged - thx!

Issue #1860, opened 2016-11-08
Accept assert() as stable


kintel:
TODO:
- [x] Documentation
- [x] Cheat Sheet
- [x] Examples
- [x] Remove from experimental
- [x] Update syntax highlighting (scadlexer.cpp and highlighter.cc)

MichaelPFrey:
I would like to add
- [ ] update syntax highlighting
- [ ] scadlexer.cpp
- [ ] highlighter.cc
to the list

I am working on the documentation:

https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/Other_Language_Features#assert

t-paul:
Good point, thanks. Added to he main list.

MichaelPFrey:
color-schemes/editor/visualstudio.json has its own keyword list (unlike all the other color schemes).
Those keywords are effective (as in they are applied)

I do not know which of this two alternatives is more sensible:
- Remove the Keyword List from color-schemes/editor/visualstudio.json (prevents further inconsistency, but I do not know if there are side-effects and/other usages of this file)
- Update the Keyword List in color-schemes/editor/visualstudio.json (please check all keywords, for e.g "children" is also missing)

MichaelPFrey:
In OpenSCAD, I can overwrite build in functions. Example:
function sin() = true;
echo (sin(1));

but I can not overwrite assertion:
function assert() = true;
echo (assert(1));

t-paul:
Added new issue #2506 as followup on the function override comment. I think that's not just specific to `assert()`.

PR #2122, opened 2017-10-11
#1860 Updated syntax highlighting for assert (scadlexer.cpp, highlighter.cc and visualstudio.json)

MichaelPFrey:


MichaelAtOz:
oooh - I'll have to get the SpaceNav out and dust it off... thanks.

MichaelPFrey:
I am new to using git, so I merged master with inputdriver3 instead of the otherway around... upps...

To be clear: I do not have a 3D-Mouse, I am using gamepads. But given that the driver structure is using the joystick API, most of what I do should be of use to a 3D Mouse.

Issue #999, opened 2014-10-30
Issues with small scale polygons due to resolution limit / clean-up when calling Clipper


t-paul:
Reported on the mailing list: http://forum.openscad.org/Performance-regression-2014-10-02-tp9989p9994.html

There is an anomaly with $fn, where increasing this value actually decreases it.

```
linear_extrude(height=1)
difference() {
    square([1,1]);
    translate([0.5, 0]) circle(r=0.25, $fn=300); // more coarse
    translate([1, 0.5]) circle(r=0.25, $fn=30);
}
```

Looks like this only happens for small values. With the same model scaled up 10 times, the issue does not occur.

![openscad-fn-issue](https://cloud.githubusercontent.com/assets/1330241/4849978/da7e4542-6065-11e4-996f-eba72d0d0bdd.png)
##

dlaw:
I am struggling to work around this issue. I use openscad to model keys for mechanical locks, which are relatively small and precise. It is natural to use units of inches for consistency with industry documentation. I take a Minkowski sum with a 0.029 inch radius circle with `$fn=360` or thereabouts. An error of 0.001 inches is significant.

I am happy to take a stab at providing a patch, but I would like to hear your thoughts about what the most elegant / correct solution would be. A flag to disable vertex cleanup for individual operations? Or a way to locally or globally alter the magic number 0.001 in src/clipper-utils.cc, a la `$fn` and friends?


More details, if you're interested:

I had to change my code from
```
  intersection() {
    minkowski() {
      circle(0.029, $fn = 360);
      polygon(...);
    }
    square(...);
  }
```
to
```
  scale(.001)
  intersection() {
    minkowski() {
      scale(1000)
      circle(0.029, $fn = 360);
      scale(1000)
      polygon(...);
    }
    scale(1000)
    square(...);
  }
```
which is tremendously inelegant.

Here are some photos showing the error introduced. Each display window is one inch wide.
![bad](https://cloud.githubusercontent.com/assets/606010/9393900/e9cfc5fa-4752-11e5-8ea0-2ce46b2a13a5.png)
![good](https://cloud.githubusercontent.com/assets/606010/9393901/e9d31d36-4752-11e5-9708-91483fdaa165.png)

PraiseSatan:
[...]

PR #2123, opened 2017-10-13
Use default clipper cleaning distance

thehans:
To minimize negative effects of clipper's flawed cleaning algorithm.  If distance not specified, clipper uses 1.415 ~= sqrt(2), the length of one diagonal.  http://www.angusj.com/delphi/clipper/documentation/Docs/Units/ClipperLib/Functions/CleanPolygons.htm

Proposed improvement for Issue #999


kintel:
I think it's worth trying this, but we should add some test cases that exercise this a bit. There are some examples in #999 that could be used as a starting point.

FYI: Failing testcases usually live in https://github.com/openscad/openscad/tree/master/testdata/scad/bugs and are added here: https://github.com/openscad/openscad/blob/master/tests/CMakeLists.txt#L1481

They are not run by default, but can be run using `ctest -C Bugs`. Once an issue is fixed, we tend to move the testcase to https://github.com/openscad/openscad/tree/master/testdata/scad/3D/issues and from then on leave it running as a normal regression test.

thehans:
@kintel The failing test was added in PR #2197.

This PR has been updated to move the new test into passing `issues` folder, and has passed successfully. I think it should be all good to merge.

kintel:
@thehans Has a conflict against master now..
I'll have a look at this; do some more extensive manual testing.

kintel:
OK, I tested against all reported duplicates of this issue. Looks good - I think the current test case covers what we need to test.

-> if you fix the conflict we can merge this.

thehans:
Ah, completely did not see the conflict.  After a brief complete mangling of this PR branch I think I have it back in order now.

PR #2128, opened 2017-10-14
Updated syntax highlighting for assert (scadlexer.cpp, highlighter.cc and visualstudio.json) #2122

MichaelPFrey:
see also #2122


t-paul:
Retriggered Travis which failed due to connection problem with OBS...

PR #2131, opened 2017-10-15
fixing "off by one error" when using /dev/input/js0

MichaelPFrey:
add button 0 and button 9 (/dev/input/js0 numbers the first button 0. button 9 is to have an even number of buttons)

renumber the axis starting from 0 to confirm with low-level tools like jstest-gtk

(maybe we should make the numbering OS dependent?)

### DIFF
--- a/src/Preferences.cc
+++ b/src/Preferences.cc
@@ -221,6 +221,7 @@ void Preferences::init() {
         initComboBox(this->comboBoxRotationY, Settings::Settings::inputRotateY);
         initComboBox(this->comboBoxRotationZ, Settings::Settings::inputRotateZ);
         initComboBox(this->comboBoxZoom, Settings::Settings::inputZoom);
+        initComboBox(this->comboBoxButton0, Settings::Settings::inputButton0);
         initComboBox(this->comboBoxButton1, Settings::Settings::inputButton1);
         initComboBox(this->comboBoxButton2, Settings::Settings::inputButton2);
         initComboBox(this->comboBoxButton3, Settings::Settings::inputButton3);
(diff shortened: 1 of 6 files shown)


t-paul:
I'm not yet sure what's the best way for that. It's certainly at least dependent on the driver. I was thinking we will (as mid-term goal) need some device - and maybe also OS - dependent configuration screen.

As first step we should be ok to make it usable for people to gather feedback for various input devices.

MichaelPFrey:
Thank you. To give a bit of context:

I was trying to show the current axis value and the currently pressed button in the preference GUI of openscad. That would make the numbering/naming/... irrelevant, as the user could see which button he/she is pressing and which axis he/she is manipulating.
The issue with that idea was, that a joystick can fire a very high update rate. Given that QT is a very powerful, high level framework, it can not really handle that update rate. The result is a glitching and crashing UI. I could of course buffer the inputs, but this would add a lot of complexity.
Then I also have a controller with a deadzone. The more I thought about it, the clearer it got to me, that we should push users to setup and adjust their input device proper in their operating system/drivers, so that we can keep openscad as lean as possible. Given that: It is useful, when we number things consistent to those tools. (with the issue beeing, that Windows starts with 1)

Eitherway: Starting from zero should be easier, as the driver can implement 0 as a dummy to get its numbers consistent. Starting with 1 was an issue, as I could not use the first button on my gamepads, as linux numbered them as 0.

(Off course, in the long run we should generate the UI from code, allowing any number and numbering, but I would prefer to get a minimum viable product in to the experimental build of openscad, then to wait for perfection)

MichaelAtOz:
When I was testing @t-paul previous version, on Win7/64, I had an issue where the SpaceMouse in its neutral state would leave the OpenSCAD display rotating very slowly. One of the calibration requirements will be to measure the actual reading in the home/neutral position and establish a new baseline. That was without the SpaceMouse driver IIRC, so there were no adjustments that could be done.

Issue #2020, opened 2017-05-01
Feature Tweak - Line marker numbering


MichaelAtOz:
I seem to frequently have the scale marker numbers not showing on the visible preview window.
For example from this:
![scale small](https://cloud.githubusercontent.com/assets/1972961/25571294/92dd764a-2e71-11e7-9084-7b0836b9852e.PNG)
One scroll wheel notch gives me this:
![scale one step](https://cloud.githubusercontent.com/assets/1972961/25571298/a2ddc41e-2e71-11e7-8bf6-abce3d8a0ed0.PNG)

One possibility would be to have an intermediate stage where every 2nd mark is numbered. 

[$15 Bounty](https://www.bountysource.com/issues/44706972-feature-tweak-line-marker-numbering)

DMCShep:
Screenshot 1 is at the closer zoom where every 10 units are labeled. Screenshot 2 is at the further zoom where every 100 units are labeled (you'd see the 100 labels if you panned a bit). Every ten marks shows a number. Took some screenshots of a guitar nut I just made demonstrating this.

Zoomed in to label every ten:
![10 zoom](https://user-images.githubusercontent.com/20283733/29479034-f97ee682-843d-11e7-978c-aa8dda0769e1.png)

One click out to label every hundred:
![100 zoom](https://user-images.githubusercontent.com/20283733/29479045-0e03868a-843e-11e7-92f3-e4e267900ef5.png)

Panning to show said label:
![100 zoom shift](https://user-images.githubusercontent.com/20283733/29479054-17638464-843e-11e7-84ac-d6d285c59edc.png)

MichaelAtOz:
@DMCShep was that to just confirm the behaviour?

DMCShep:
Oops, thought the bug was the lack of any label, rather than it not being visible in the viewport.

In any case, detection of when the marks need to change scale should probably be independent of when the labels need to change scale. Wouldn't hurt to consider labels at every 10, 20, 50, 100, etc. units either.

MichaelAtOz:
This is a feature request. Could maybe a bug, but basically it was not helpful & I wanted to spend some bounty money on an annoying thing.

thehans:
I like the improved labeling scheme from @justbuchanan , but I have a suggestion to improve it further.

So the issue was mainly the size of the "jump" that the labels scale by was too large.

Originally it was a single jump * 10 for every decimal power.

[...]

PR #2132, opened 2017-10-17
Show minor tick labels when major tick labels are invisible. #2020

justbuchanan:
This pull request addresses issue #2020.

It can sometimes be difficult to tell what units the tick marks represent depending on how zoomed in you are relative to the current axes scale. The axes change "units" every time the viewport changes in size by a factor of 10. When you scroll out just past the point where the units change again, you usually can't see any tick labels at all. I added additional labels every 2 minor tick marks that are visible when very few (sometimes 0) major tick marks are visible. I think this makes it a bit easier to tell what the denominations are in those situations.

See the code comment for how to determine whether to show these additional labels given the current zoom level. The threshold is adjustable, but I think it's at a reasonable value now.

Here are some screenshots to show change in behavior.

Current behavior zoomed in:
master-zoomed-in

Current behavior zoomed out:
master-zoomed-out

New behavior zoomed in (labels every 2 minor ticks):
pr-zoomed-in

New behavior zoomed out (no minor tick labels):
pr-zoomed-out

### DIFF
--- a/src/GLView.cc
+++ b/src/GLView.cc
@@ -506,42 +506,59 @@ void GLView::showCrosshairs()
 
 void GLView::showScalemarkers(const Color4f &col)
 {
-	// Add scale tics on large axes
+	// Add scale ticks on large axes
 	auto l = cam.zoomValue();
 	glLineWidth(this->getDPI());
 	glColor3f(col[0], col[1], col[2]);
 
-	// determine the log value to provide proportional tics
-	auto log_l = static_cast<int>(log10(l));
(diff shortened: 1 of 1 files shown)


t-paul:
Yes, looks much nicer. Only the center looks a bit strange, is that 3 times zero for all the axes overlapping?

image

justbuchanan:
Good catch, I forgot to exclude 0 :). It should be fixed now.

kintel:
@t-paul Feel free to merge this if your concern was addressed

Issue #2045, opened 2017-05-29
Feature request: Add filename to console for export


MichaelAtOz:
ATM the console shows e.g.
```
...
Rendering finished.
STL export finished.
CSG export finished.
etc.
```
For Shapeways, different materials have different Design Guidelines, such as Minimum Wall Thickness, Clearance etc.
Thus often I need to export multiple STLs, with a change to a relevant variable to suit a particular material.
I do this by including something as a postfix to the filename.
This can get confusing. I often switch windows while the STL is produced.
So knowing the last filename will inform of where I was up to.

So something like `STL export finished. some_file_name.stl`
or, to preserve the existing message, in case some poeple process it.
```
Exporting some_file_name.stl
STL export finished.
```

A possible related issue is whether the console should be cleared at each compile/preview/render.
I think it would be good to be able to scroll back beyond the last operation.
This probably needs discussion??

t-paul:
Looks like this is already implemented. Only exception was CSG which is now covered by #2872.

MichaelAtOz:
> Only exception was CSG which is now covered by #2872.

And Export/Image, which I [mentioned](https://github.com/openscad/openscad/issues/2845#issuecomment-470369146).

t-paul:
Hmm, I skipped that as it's not a normal export and did not even have any info, but we can add that too...

t-paul:
#2875

MichaelAtOz:
Thanks.

PR #2133, opened 2017-10-17
Add filename to console export message. #2045

justbuchanan:


src/mainwin.cc
```diff
@@ -2183,7 +2183,8 @@ void MainWindow::actionExport(FileFormat format, const char *type_name, const ch
 	exportFileByName(this->root_geom, format,
 		export_filename.toLocal8Bit().constData(),
 		export_filename.toUtf8());
-	PRINTB("%s export finished.", type_name);
+	PRINTB("%s export finished: %s",
+		type_name % export_filename.toLocal8Bit().constData());
```
kintel: Since we're printing to the GUI console, should this be `toUtf8()` ?

justbuchanan: Yes, I think you're right. I just changed it. Both seem to work the same on my system (modern Linux), but maybe on other systems Qt wants utf-8 even if the local codec is something else?

Note that there's some inconsist use of the two in this file (maybe others). For example there's a `PRINTB` in `MainWindow::actionExportCSG()` that uses `toLocal8Bit()`.

kintel: To my understanding, local 8 bit encoding is for performing system calls outside of Qt. Qt should prefer its own encoding for all string uses. OpenSCAD APIs prefer `std::strings` where possible and UTF-8 otherwise. To my knowledge Linux uses utf-8 as its native 8-bit encoding, while e.g. Windows may use some internal codepage stuff.

kintel: There may be some inconsistency. If you find any, feel free to fix and PR :)

justbuchanan: Using utf-8 for most everything internally, then local 8-bit for system interactions makes sense. Btw, I see that as of Qt 5, there's a [QString::toStdString()](http://doc.qt.io/qt-5/qstring.html#toStdString) method. This might be best to use in most places - it does use toUtf8() behind the scenes though.

kintel: I agree that `toStdString()` would be preferable once we kill Qt4 support.

Issue #2100, opened 2017-08-24
Accept HTML Color code


kuldeepdhaka:
Please accept \#RRGGBB (and \#RRGGBBAA will not be a big either if the non alpha is implemented)

People (including me) who have worked with the HTML color code will find it very handy.

I dont think this is supported because nothing mentioned in documentation
https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/Transformations#color

and if i try to use, result in
```
WARNING: Color name "#b5a642" unknown. Please see
WARNING: http://en.wikipedia.org/wiki/Web_colors
```

MichaelAtOz:
Thanks, I needed a challenge.
[rgb.txt](https://github.com/openscad/openscad/files/1250754/rgb.txt)

Corona688:
This is a terrific idea.

justbuchanan:
Closed as of #2134.

PR #2134, opened 2017-10-17
Add support for specifying colors with hex color codes. #2100

justbuchanan:
This allows for colors to be specified in "#rrggbb" or "#rrggbbaa" format. QColor has a builtin way of doing this, so it's fairly straight-forward. Note that QColor uses the "#aarrggbb" format for the version with alpha. I re-arranged the input before feeding it to QColor because the "#rrggbbaa" format seems more common, but using the native Qt format would also be an option.


thehans:
If its intended be web-compatible syntax, it would be nice to also support shorthand (3 or 4 digit) hexadecimal notation:
#RGB[A]
Where each shorthand hex digit is just duplicated internally to form the actual value.

https://drafts.csswg.org/css-color/#typedef-hex-color

kintel:
Unfortunately, `src/color.cc` is part of the core and cannot depend on Qt.

justbuchanan:
@thehans good point, I added support for those formats too.

@kintel I didn't realize that, but makes sense. I rewrote it using standard C++ (and some boost). It ended up being easier to add the additional formats this way too because QColor doesn't parse those from what I can tell.

justbuchanan:
Btw, is there a good place to add a couple test cases for this?

src/color.cc
```diff
@@ -213,6 +213,47 @@ ColorModule::~ColorModule()
...
+
+	std::array<int,4> rgba;
+	rgba[3] = channel_max; // default alpha channel to 100%
+
+	for (unsigned i = 0; i < (hex.size() - 1) / stride; i++) {
+		const std::string chunk = hex.substr(1 + i*stride, stride);
+
+		// validate - only [0-9a-zA-Z] are allowed
+		for (char c : chunk) {
+			if (!std::isdigit(c) && !('a' <= c && c <= 'z') && !('A' <= c && c <= 'Z')) return boost::none;
```
thehans: Hexadecimal digits should be [0-9a-fA-F]

justbuchanan: Oops, just fixed it :)

kintel:
@justbuchanan We don't currently have any unit tests, but a regression test for this would make sense. See https://github.com/openscad/openscad/blob/master/doc/testing.txt

Smth. based on one of these two tests should work:
https://github.com/openscad/openscad/blob/master/testdata/scad/3D/features/color-tests.scad
https://github.com/openscad/openscad/blob/master/testdata/scad/3D/features/testcolornames.scad

I wouldn't test anything exhaustive, just try to cover typical failure modes like different RGB values, with/without alpha, different casing. ..and perhaps failure modes like color parsing errors.

justbuchanan:
Cool, I just added some examples to testcolornames.scad and updated the expected files. Let me know if there's anything I missed.

src/color.cc
```diff
@@ -213,6 +213,47 @@ ColorModule::~ColorModule()
...
+	const bool long_syntax = hex.size() == 7 || hex.size() == 9;
+	if (!short_syntax && !long_syntax) return boost::none;
+
+	if (hex[0] != '#') return boost::none;
+
+	// number of characters per color channel
+	const int stride = short_syntax ? 1 : 2;
+	const float channel_max = short_syntax ? 15 : 255;
+
+	std::array<int,4> rgba;
```
kintel: For a small improvement in readability, what if you made this a `Color4f`, and in the end just returned `rgba / channel_max`?

src/color.cc
```diff
@@ -213,6 +213,47 @@ ColorModule::~ColorModule()
...
+	rgba[3] = channel_max; // default alpha channel to 100%
+
+	for (unsigned i = 0; i < (hex.size() - 1) / stride; i++) {
+		const std::string chunk = hex.substr(1 + i*stride, stride);
+
+		// validate - only [0-9a-fA-F] are allowed
+		for (char c : chunk) {
+			if (!std::isdigit(c) && !('a' <= c && c <= 'f') && !('A' <= c && c <= 'F')) return boost::none;
```
kintel: Could you use `isxdigit(c)` instead?

..or maybe even smht. like `if (!std::all_of(std::cbegin(chunk), std::cend(chunk), std::isxdigit)) return boost::none;`

src/color.cc
```diff
@@ -236,8 +277,14 @@ AbstractNode *ColorModule::instantiate(const Context *ctx, const ModuleInstantia
 		if (webcolors.find(colorname) != webcolors.end())	{
 			node->color = webcolors.at(colorname);
 		} else {
-			PRINTB_NOCACHE("WARNING: Color name \"%s\" unknown. Please see", colorname);
-			PRINT_NOCACHE("WARNING: http://en.wikipedia.org/wiki/Web_colors");
+			// Try parsing it as a hex color such as "#rrggbb".
+			boost::optional<Color4f> hexColor = parse_hex_color(colorname);
```
kintel: `const auto hexColor`

src/color.cc
justbuchanan: Cool, I didn't see that `Color4f` was just an `Eigen::Vector`.

src/color.cc
justbuchanan: Thanks, I hadn't seen `isxdigit` before, that's a lot cleaner.

I switched to `std::all_of`, but unfortunately `isxdigit` requires a cast from `char`, so I don't think there's a way to use it there without wrapping it in a lambda (see "Notes" [here](http://en.cppreference.com/w/cpp/string/byte/isxdigit)). Let me know if you see a better way to do that.

kintel: Makes sense. Readability is a primary goal, so let's keep it easy.

PR #2136, opened 2017-10-18
Cache homebrew downloads on travis-ci

justbuchanan:
I noticed that a lot of osx builds were failing on travis-ci due to timeouts. A large part of this is installing packages - a recent build took almost 15 minutes updating and installing packages with homebrew.

I added homebrew's download directory to travis's cache list, which will hopefully speed up package installation.

I haven't tested this yet, but opening this pull request will start it running. After the first run on travis, pushing a new commit should start another build that uses the previously-filled cache. That should tell us if it saves any/much time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -95,3 +95,4 @@ addons:
     build_command: make
     branch_pattern: coverity_scan
 
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ matrix:
       cache:
         directories:
           - "$HOME/Library/Caches/Homebrew"
+          - "$(brew --cache)"
 
 notifications:
   irc:
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,9 @@ matrix:
       dist: trusty
       env: DIST="trusty"
       sudo: required
-      cache: apt
+      cache:
+        - apt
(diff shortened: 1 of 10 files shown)


justbuchanan:
This does improve osx build times, but isn't a big enough change to prevent them from timing out.

The osx builds seem to usually timeout somewhere during the main build of openscad. Here's an example: https://travis-ci.org/openscad/openscad/jobs/289387095

With the homebrew download cache, it makes it past the main compile, builds the tests, and typically times out some time during the testing phase. Here are two examples:
* https://travis-ci.org/openscad/openscad/jobs/289564950
* https://travis-ci.org/openscad/openscad/jobs/290065774

kintel:
I don't really see any improvement in build times. This PR enables caching of homebrew installs. Looking at the timing for the homebrew part of the build, your two optimized builds look like this:
* 962.05 sec
* 979.38 sec

Looking at past successful builds:
* 905.85
* 477.57
* 737.70
* 759.35
* 500.18
* 487.61

-> it feels like something isn't really being cached, or that downloading the cache takes longer than a good, normal build..

justbuchanan:
Yes, you're right, I was looking at the wrong numbers. This doesn't seem to really be helping at all, so I'll close it.

Issue #2109, opened 2017-09-12
Failed to open non ASCII file in Win10 (Russian name of scad file)


prokator:
On open name became corrupted (like failed encoding) and file is not shown. On other hand - initial file is available (on file system)
version: OpenSCAD version 2017.01.20 (git 59df0d1)

kintel:
This sounds like a duplicate of #113.
We have some work in progress in #1296, I'm just uncertain about the right solution, and we don't have many Windows dev resources.

jordanbrown0:
I have no idea why I closed this as a dup of #4831; the two appear to be unrelated.

However, closing it as a dup of **something** was probably appropriate; I did a bunch of work on supporting non-ASCII file names, and today I can't duplicate a problem with opening a Russian-name file.

PR #2138, opened 2017-10-18
Add ccache to cmake. #2109

justbuchanan:
This caches compiled object files, which can significantly speed up partial re-builds. This is most useful when switching between git branches. Normally, any changed files and dependencies would need to be re-compiled each time you switch branches. With ccache, any file that was previously compiled with the same inputs can be re-used from the cache.

I added a check so that it will only use ccache if it is installed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,14 @@ matrix:
       dist: trusty
       env: DIST="trusty"
       sudo: required
-      cache: apt
+      cache:
+        - apt
+        - ccache
     - os: osx
       compiler: clang
       env: DIST="osx"
+      cache:
+        - ccache
 
 notifications:
   irc:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -128,6 +128,15 @@ add_definitions(-DENABLE_EXPERIMENTAL)
 # Stack size 8MB; github issue 116
 add_definitions(-DSTACKSIZE=8388608)
 
+# Setup ccache (if available) to speed up recompiles. It's especially useful
+# when switching back and forth between branches where large numbers of files
+# would otherwise need to be re-compiled each time.
+find_program(CCACHE_PATH ccache)
+if (CCACHE_PATH)
+  set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ${CCACHE_PATH})
+  set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK ${CCACHE_PATH})
+endif()
+
 #
 # Version
 #
--- a/openscad.pro
+++ b/openscad.pro
@@ -147,6 +147,8 @@ netbsd* {
   QMAKE_CXXFLAGS_WARN_ON += -Wno-sign-compare
 }
 
+!lessThan(QT_VERSION, 5.9): CONFIG += ccache
+
 CONFIG(skip-version-check) {
   # force the use of outdated libraries
   DEFINES += OPENSCAD_SKIP_VERSION_CHECK
--- a/scripts/macosx-build-homebrew.sh
+++ b/scripts/macosx-build-homebrew.sh
@@ -44,7 +44,7 @@ $TAP tap openscad/homebrew-tap
 # FIXME: We used to require unlinking boost, but doing so also causes us to lose boost.
 # Disabling until we can figure out why we unlinked in the first place
 # brew unlink boost
-for formula in eigen boost cgal glew glib opencsg freetype libzip libxml2 fontconfig harfbuzz qt5 qscintilla2 imagemagick; do
+for formula in eigen boost cgal glew glib opencsg freetype libzip libxml2 fontconfig harfbuzz qt5 qscintilla2 imagemagick ccache; do
   log "Installing formula $formula"
   brew ls --versions $formula
   time brew install $formula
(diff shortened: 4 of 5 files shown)


kintel:
Will this clear the ccache cache on `make clean`? If not, what's the process for clearing the cache (to be able to perform a full rebuild for cases where ccache fails or you change any environment not detected by ccache)?

kintel:
Oh, btw., running ccache on TravisCI could be really interesting.

justbuchanan:
Ccache stores things in `~/.ccache`, so clearing that directory will reset the cache. `make clean` doesn't have any effect on ccache. In theory, ccache records all relevant environment bits for each compilation, so it *shouldn't* be susceptible to the types of situations where `make clean` would fix a build issue. In practice, it's possible you may need to reset ccache, but I haven't encountered that yet after using ccache for several projects.

I think TravisCI would see a pretty large speed improvement with ccache. The only thing we'd need to do after this PR is add the ccache directory to travis's cache list.

kintel:
I guess my question is more like: Does ccache figure out what compiler you're using and correctly wipe the cache if you `apt-get upgrade` your compiler?, and does it know what environment variables influence the build?

justbuchanan:
Yes, it records the compiler that was used and all of the parameters passed to it. Their website has a more complete list of what goes in the hash that identifies each object file: https://ccache.samba.org/manual.html#_common_hashed_information.

Ccache is invoked by `make` just before the compiler. Where `make` would have previously done something like `gcc src/main.cc -o build/main.o`, it instead invokes `ccache gcc src/main.cc -o build/main.o`. In this way, ccache can record compiler info and all relevant parameters, which it hashes together to identify the object file. Then ccache invokes the compiler itself, passing on the parameters.

justbuchanan:
I tried enabling this for TravisCi and it seems to make a fairly large difference.

Before:
image

After:
image


The speedup won't be quite as big for commits that actually change code though.

kintel:
Note: TravisCI builds OpenSCAD using qmake. cmake is only used to build some smaller test utilities. Not sure how to enable ccache for the qmake build, but that should have a more significant impact.

justbuchanan:
TravisCI actually sets up ccache differently when you add `ccache: travis` to the config. If you look in the build logs, you'll see:
* export PATH=/usr/lib/ccache:$PATH
* Check for working C compiler: /usr/lib/ccache/gcc -- works
* Check for working CXX compiler: /usr/lib/ccache/g++ -- works

This applies ccache to the qmake setup, independent of the changes I made to the CMakeLists.txt file.

Btw, it looks like the test cmake setup is separate from the main one. Should I also add ccache to tests/CMakeLists.txt?

kintel:
Yes, it wouldn't hurt adding ccache to tests/CMakeLists.txt, for consistency.

kintel:
I'm happy with this as is, but one final question: Is there an easy way for me to enable ccache locally for the qmake build system?

kintel:
Humm, I guess for qmake I need to override some env. variables to trick qmake into using ccache as a compiler..

justbuchanan:
I just tried and the export command I showed from travis didn't work for me either... I tried doing `export PATH="/usr/lib/ccache/bin:$PATH"` instead and it worked for me. A quick test you can do is `which c++`, which should give you `/usr/lib/ccache/bin/c++`. I'm not sure why the travis command didn't include the `/bin` at the end of the path.

kintel:
On macOS, qmake appears to be too smart and doesn't look in the PATH for the compiler. This is probably why TravisCI has their own hand-tuned system containing ccache rather than installing it using homebrew.

justbuchanan:
Recent versions of qmake have a ccache option builtin: https://codereview.qt-project.org/#/c/196345/.

If there's an easy way to check the version, we could turn it on when it's available. Aside from the version-checking, it looks as simple as `CONFIG += ccache`.

kintel:
Nice - wasn't aware of that; perhaps time to upgrade Qt..

kintel:
It should be possible to do (untested):
```
!lessThan(QT_VERSION, 5.9): CONFIG += ccache
```

justbuchanan:
Cool, I tried that out and it works well. I also tested with the version set to a higher number than I have, and it worked fine and didn't use ccache.

kintel:
ok, let's try this. If anything goes wrong we can add a flag to disable/enable it.

kintel:
@justbuchanan Could you have a look at this build failure: https://travis-ci.org/openscad/openscad/jobs/293137464

-> somehow the Mac build still times out. It technically succeeds, but somehow still times out. Any ideas?

justbuchanan:
Hmm, that's strange. I see "could not download cache" in the travis logs. Maybe it's a failure on their end? I'll open a test PR to trigger another build and see if it has better luck.

image

donbright:
sorry i am a broken record on this stuff but it needs to be said. 

in my opinion and experience, a tradeoff between reliability of build vs speed of build  -> my vote is always going to be with reliability. the time we save building is not worth the time lost to debugging errors caused by not having a clean build. one of these speedups is 20 minutes out of 70 minutes.

we should be asking, why does it take 70 minutes to build in the first place? there is too much unnecessary code, too much duplicated code, too much complicated code that could be simplified. 

my two cents (yes i realize i am outvoted but that is no reason, not to vote!)

Issue #1853, opened 2016-11-04
STL export, wrong facet normal values [$50]


indazoo:
Hi

I experienced hole errors in netfabb and meshlab when loading the STL files exported by OpenScad.
First I thought the error must lie between my ears but in this case it seems not to be the case.

In a STL file, for each triangle there are three vertexes and for these a norm vector.
After repairing my STL file in Netfabb I realised the norm vector changed for a lot of values between these two STL files. The vertexes stayed the same.
So below you find some OpenScad code from which I believe does calculate the norm vector correctly.
I took the triangle vertexes from OpenScad's ascii export, used them as points in OpenScad and then calculated the normal vectors.
The norm vectors found in OpenScad's STL file were not(!) corresponding with the calculated  vertexes. The netfabb output showed way better acurracy. The differences are quite significant.
I used version 2016.10.04 (git 7e0935d)  and first 2015.04.25 (git 2b33922) .

```
case1_pt1 = [2.40086, -3.19936, 10];
case1_pt2 = [2.40086, -3.19936, 9.96858];
case1_pt3 = [2.45734, -3.15223, 10];
OpenScad_pt_1_stl_export = [0.640593, -0.76788, 0];
Netfab_corrected_pt_1 = [0.640692, -0.767798, 0.000000];
echo("***************");
echo("**  CASE 1   **");
echo("OpenScad norm vector (ascii stl export) = ",  OpenScad_pt_1_stl_export);
echo("Calculated from pts = ", triangle_normal_stl(case1_pt1, case1_pt2, case1_pt3) );
echo("Netfabb repaired norm vector (ascii stl export)= ", Netfab_corrected_pt_1);

case2_pt1 = [2.21668, -3.32328, 10];
case2_pt2 = [2.05917, -3.42926, 9.9189];
case2_pt3 = [2.40086, -3.19936, 10];
OpenScad_pt_2_stl_export = [0.558244, -0.829677, -7.70895e-016];
Netfab_corrected_pt_2 = [0.558230, -0.829686, 0.000039];
echo("****************");
echo("**  CASE 2    **");
echo("OpenScad norm vector (ascii stl export) = ",  OpenScad_pt_2_stl_export);
echo("Calculated from pts = ", triangle_normal_stl(case2_pt1, case2_pt2, case2_pt3) );
echo("Netfabb repaired norm vector (ascii stl export) = ", Netfab_corrected_pt_2);

function triangle_normal_stl(p1, p2, p3) =
[...]

Issue #420, opened 2013-06-26
STL and AMF output should have a canonicalization option


donbright:
nop head has to run stl through a 'canonicalization' script that sorts the triangles. 

this should be an option available in openscad itself. 

https://github.com/nophead/Mendel90/blob/master/c14n_stl.py
##

t-paul:
Added AMF too. As it's XML it might be useful for that too.

thehans:
I'm looking into fixing this, but I'm a little confused about the export code.  It looks like STL export converts CGAL_Nef_polyhedron to Polyset before exporting:
https://github.com/openscad/openscad/blob/master/src/export_stl.cc#L153-L156

While AMF export does the opposite, converting Polysets to CGAL_Nef_polyhedron before exporting:
https://github.com/openscad/openscad/blob/master/src/export_amf.cc#L162-L165

I don't understand the reasoning, or if either option is particularly better than the other, but it seems like they should be changed so that both follow the same general export procedure if possible?

kintel:
The STL approach should be significantly more robust when exporting designs that are not technically manifold. The AMF, OFF etc. exporters didn't get that facelift, so they're still using the method resulting in the least amount of code.
Getting all exporters to do the same thing sounds like the correct approach.

GilesBathgate:
@kintel @thehans It seems like the main difference between `tessellatePolygonWithHoles` and `convert_to_Polyhedron` is that one uses a Constrained_delaunay_triangulation and the other uses just a plain Constrained_triangulation. The former is better as it produces fewer sliver triangles.

As an aside I think you could have just swapped in the delaunay one here: https://github.com/openscad/openscad/blob/master/src/CGAL_Nef3_workaround.h#L82

BTW I am looking at this approach, which is supposed to be much faster and uses delaunay  (If I can get it to work correctly)
https://github.com/GilesBathgate/cgal/commit/580486a1f9700d10d104e396312593683c605640

kintel:
IIRC, using CGAL's delaunay triangulator only works if we're using rationals, and even then, we can only extract the resulting triangles if the design is perfectly manifold. This caused a bunch of issues, hence the introduction of a separate triangulator.
@GilesBathgate I believe you opted for keeping rationals as late in the process as possible to accommodate this, including experimenting with rational representation of STL/OBJ files?

GilesBathgate:
>triangulator only works if we're using rationals
[...]

Issue #337, opened 2013-04-24
export directly to stl from nef polyhedron w/o manifold check [$10]


donbright:
per mailing list 4/24/13

export directly to stl from nef polyhedron w/o manifold check

throw warning if object is not manifold, offer option to export anyway

(note to the user they may need to repair STL with netfab etc)

TODO:
- [x] Implement conversion code from Nef polyhedron directly to PolySet (done by @donbright)
- [x] Add some representative export-import tests (after #816 is merged)
- [x] Consider adding a geometry sanity check (-> separate issue #925)
##

doug-moen:
The following model is self intersecting. It produces no warning when exporting to STL, and the resulting STL file is ambiguous (different results from Slic3r and Simplify3d, according to Ronaldo). The same model produces a fatal error message in the console when you export to AMF.

p = [ [0,0], [7,0], [7,7], [3,7], [3,3], [10,3], [10,10], [0,10] ];
linear_extrude(2) polygon(p);

According to this issue, there should be a warning. Should I create a new issue or reopen this one?

kintel:
I think a new issue would make this easer to track.

PR #2140, opened 2017-10-18
Calculate STL normals from string formatted vertex data for consistency

thehans:
for #1853

### DIFF
--- a/src/export_stl.cc
+++ b/src/export_stl.cc
@@ -51,6 +51,8 @@ static void append_stl(const PolySet &ps, std::ostream &output)
 		stream.str("");
 		stream << p[2][0] << " " << p[2][1] << " " << p[2][2];
 		std::string vs3 = stream.str();
+		
+		
 		if (vs1 != vs2 && vs1 != vs3 && vs2 != vs3) {
 			// The above condition ensures that there are 3 distinct vertices, but
 			// they may be collinear. If they are, the unit normal is meaningless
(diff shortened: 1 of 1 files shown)


kintel:
Looks good!
Is there a way you could write a single test case that confirms that the output matches the input?
(not to validate that your code is correct, but to catch failed attempts to refactor this in the future)

thehans:
Ok, I'm looking into creating a test, but the workflow for this test is unlike any existing tests.  They all basically expect to compare two pngs at the end.

I guess I'm looking at creating another script similar to 
 ```export_import_pngtest.py``` , call it ```export_import_stltest.py``` and then figuring how to work that into the rest of the ctest framework

Any tips or suggestions?

kintel:
I think this should be doable with the current python script by doing smth. like this:
```
add_cmdline_test(mytest EXE ${OPENSCAD_BINPATH} ARGS -o SUFFIX stl FILES ${MY_TEST_FILES})
```

The script should pick up on the stl suffix and compare the STL files.

thehans:
@kintel that would require me to push an "expected" STL to github right?  The problem is because of #420.  I don't think the ordering of the triangles will be guaranteed across platforms.

The test procedure which I've done manually so far is:
1. Open a `issue1853.scad` file, and do an STL export.  (this will count as our "expected" file later)
2. Import that STL and export to STL again.
3. diff the two STL files

That seems to work ok since its all done from the same build. 

In the comments of the tests/CMakeLists.txt here, it lists the different cmdline test types: https://github.com/openscad/openscad/blob/master/tests/CMakeLists.txt#L1612-L1628
A couple are STL related, but these ultimately compare against png files at the end:
```
# o stlpngtest: Export to STL, Re-import and render to PNG (--render)
# o stlcgalpngtest: Export to STL, Re-import and render to PNG (--render=cgal)
```
So that's why I'm thinking there would need to be a new test type, with its own .py script:
`# o stlstltest: Export to STL, Re-import and export to STL again, compare with first export.`

Or maybe I should just say this is blocked by #420 and start working on that.

kintel:
ok, sounds fair. If it's too much work, time is probably best spend elsewhere.

MichaelAtOz:
I seem to recall @nophead had a method that sorted the STL so they could be compared.

thehans:
@MichaelAtOz Yes, it's mentioned in the initial comment of #420 :stuck_out_tongue: 

And to save you an extra click, the direct link:
https://github.com/nophead/Mendel90/blob/master/c14n_stl.py

I think the ultimate solution for #420 should be rewritten to be handled by the c++ stl export code, but maybe this python script would be a good addition to the test suite in this case.

kintel:
@thehans Re-reading this PR. Consistent normal output is only implemented for PolySet exports, not for CGAL Nef3 export. Is this just an oversight or was this done on purpose?

thehans:
@kintel I don't think the function `static void append_stl(const CGAL_Polyhedron &P, std::ostream &output)` is ever called with the current code and I guess I feel like it should be deleted but wasn't going to bother with it in this PR.  All the code paths end up converting to polyset if its not already.  

Is there a reason to keep this always true conditional or any of its `else` block code ?
https://github.com/thehans/openscad/blob/f2a42bc3a3f29a9c52e34bd7e693c7bdbd8c5f4e/src/export_stl.cc#L177-L185
because I think the only call to the CGAL_Polyhedron method is in the unused else branch.

kintel:
Good point. The always-true `if` was introduced when fixing #337 a couple of years back. I guess we could kill the always-false branch now.

PR #2142, opened 2017-10-20
Remove mixed tab characters from python source

thehans:
whitespace-only changes.

.py source files had tabs mixed with spaces, all tabs removed and indented to the appropriate multiple of 4 spaces


kintel:
Thx. My goal is to do this for C++ code as well. I was hoping to get a `clang-format` config to manage the style to allow this to run smoother, but that's going to take a while..

PR #2144, opened 2017-10-22
Test build should find same packages as main build

thehans:
Also whitespace changes: removed some mixed tabs, made indentation more consistent.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -366,29 +366,6 @@ if (UNIX)
   endforeach()
 endif()
 
-# OpenCSG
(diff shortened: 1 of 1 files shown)


tests/CMakeLists.txt
```diff
@@ -875,6 +875,7 @@ target_link_libraries(tests-offscreen ${OFFSCREEN_CTX_LIBRARIES})
 #
 # csgtexttest
 #
+find_package(OpenCSG REQUIRED)
```
kintel: When adding this, we should remove the manual CSG finding code above as I assume they end up doing the same thing.

PR #2147, opened 2017-10-25
Add clang-format script

justbuchanan:
I'm not sure how interested you all are in having a set code format, but I've seen it discussed on a few issues on this project and have personally found these tools to be really useful, especially on projects with many contributors. Having consistent formatting across files improves readability and the tools available, such as clang-format, make this really easy.

This PR adds a basic script for running clang-format on this project's code. By default, it only re-formats files that differ from the master branch, which prevents spurious diffs from showing up in irrelevant files. It also provides a `--all` flag that will instead reformat *all* .h and .cc files in the repo.

### config

clang-format reads from the `.clang-format` file for formatting options. I set this to default to the Google style for most things, but made a couple minor changes based on the current code style. Please check out the clang-format docs site and see if there are other options that would better suit the current/desired format: https://clang.llvm.org/docs/ClangFormatStyleOptions.html.

### try it out!

(Save your current work first, then...) Run `./scripts/beautify.sh --all` to reformat everything, then `git diff` to see what changes were made. If you see changes that you dislike, there may be config options (linked above) to change them.

### usage

Initially, you'll want to reformat the whole repository. This is easy to do, but has some downsides:
* lots of extra diffs with open pull requests
  * these are reduced if you run the formatter on each PR branch before attempting to merge, but the extra diffs still make it harder for git to figure out what happened in some situations.
* Sometimes the auto-formatted version of the code is worse than the original.
  * Code comments may need to be re-wrapped - sometimes the auto-formatter wrapping does weird things
  * On occasion, you may need to add `// clang-format off` and `// clang-format on` comments before and after, respectively, some code pieces that you'd prefer manual formatting for.

After the initial formatting, you only need to re-run the formatter on files as they change. In practice, this means encouraging pull-request submitters to run the formatting script and/or running it before merging the pull requests.

Here's a diff showing an initial formatting run on the whole repo. There are a lot of diffs, but most of them are just whitespace. https://github.com/justbuchanan/openscad/commit/02d505ccbad72771603a37c6f15264592e184fd9.

TODO
- [ ] Add note to contributing guide
- [ ] Allow running from within the `scripts` dir. Needs to `cd` to the root directory of the repo before running

### DIFF
--- a/scripts/beautify.sh
+++ b/scripts/beautify.sh
@@ -22,7 +22,7 @@ function reformat_all() {
 DIFFBASE="origin/master"
 function reformat_changed() {
     ANCESTOR=$(git merge-base HEAD "$DIFFBASE")
-    FILES=$(git --no-pager diff --name-only $ANCESTOR | grep -E "\.h|\.hpp|\.cc|\.cpp" | $FILTER_CMD)
+    FILES=$(git --no-pager diff --name-only $ANCESTOR | grep -E "\.(h|hpp|cc|cpp)" | $FILTER_CMD)
     if [ $? -ne 0 ]; then
         echo "No files to format, exiting..."
     else
--- a/scripts/beautify.sh
+++ b/scripts/beautify.sh
@@ -7,14 +7,14 @@
 FORMAT_CMD="clang-format -i -style=file"
 
 function reformat_all() {
-    find . -name "*.h" -o -name "*.cc" | xargs $FORMAT_CMD
+    find . -name "*.h" -o -name "*.hpp" -o -name "*.cc" -o -name "*.cpp" | xargs $FORMAT_CMD
 }
 
 # reformat files that differ from master.
 DIFFBASE="origin/master"
 function reformat_changed() {
     ANCESTOR=$(git merge-base HEAD "$DIFFBASE")
-    FILES=$(git --no-pager diff --name-only $ANCESTOR | grep -E "\.cc|\.h")
+    FILES=$(git --no-pager diff --name-only $ANCESTOR | grep -E "\.h|\.hpp|\.cc|\.cpp")
     if [ $? -ne 0 ]; then
         echo "No files to format, exiting..."
     else
--- a/scripts/beautify.sh
+++ b/scripts/beautify.sh
@@ -4,13 +4,15 @@
 
 # note: with the -style=file option, clang-format reads the config from ./.clang-format
 # See here for the full list of supported options: https://clang.llvm.org/docs/ClangFormatStyleOptions.html
-FORMAT_CMD="clang-format -i -style=file"
+FORMAT_CMD_CLANG_FORMAT="clang-format -i -style=file"
+FORMAT_CMD_UNCRUSTIFY="uncrustify -c .uncrustify.cfg --no-backup"
+FORMAT_CMD=$FORMAT_CMD_CLANG_FORMAT
 
 # Filter out any files that shouldn't be auto-formatted.
 # note: -v flag inverts selection - this tells grep to *filter out* anything
 #       that matches the pattern. For testing, you can remove the -v to ssee
 #       which files would have been excluded.
-FILTER_CMD="grep -v -E objects|src/polyclipping|src/CGAL_Nef3_workaround.h|src/CGAL_workaround_Mark_bounded_volumes.h|src/convex_hull_3_bugfix.h|src/OGL_helper.h"
+FILTER_CMD="grep -v -E objects|src/libsvg|src/libtess2|src/polyclipping|src/CGAL_Nef3_workaround.h|src/CGAL_workaround_Mark_bounded_volumes.h|src/convex_hull_3_bugfix.h|src/OGL_helper.h"
 
 function reformat_all() {
     find src -name "*.h" -o -name "*.hpp" -o -name "*.cc" -o -name "*.cpp" \
@@ -32,7 +34,10 @@ function reformat_changed() {
 }
 
 # main
-if [[ "$#" -eq 1 && "$1" == "--all" ]]; then
+if [ "`echo $* | grep uncrust`" ]; then
+    FORMAT_CMD=$FORMAT_CMD_UNCRUSTIFY
+fi
+if [ "`echo $* | grep -- --all`" ]; then
     echo "Reformatting all files..."
     reformat_all
 else
--- a/scripts/beautify.sh
+++ b/scripts/beautify.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+# Reformat C++ code using clang-format
+
+# note: with the -style=file option, clang-format reads the config from ./.clang-format
+# See here for the full list of supported options: https://clang.llvm.org/docs/ClangFormatStyleOptions.html
+FORMAT_CMD="clang-format -i -style=file"
+
+function reformat_all() {
+    find . -name "*.h" -o -name "*.cc" | xargs $FORMAT_CMD
+}
+
+# reformat files that differ from master.
+DIFFBASE="origin/master"
+function reformat_changed() {
+    ANCESTOR=$(git merge-base HEAD "$DIFFBASE")
+    FILES=$(git --no-pager diff --name-only $ANCESTOR | grep -E "\.cc|\.h")
+    if [ $? -ne 0 ]; then
+        echo "No files to format, exiting..."
+    else
+        echo -e "Reformatting files:\n$FILES"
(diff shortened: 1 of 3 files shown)


kintel:
Nice work!
Last I looked at this, I attempted to establish a clang-format file that resulted in the smallest possible diff after running clang-format on the entire codebase. This would still be a pretty large diff, but would allow for us to get clang-format introduced without touching every single line of code.

..then the next step could be making the big tabs->spaces switch, plus adopting other best practices that we may have overlooked or deferred in the name of consistency.

I agree that, ideally, we should be able to run clang-format on all code. We can wrap particularly tricky pieces of code with `// clang-format off`  `// clang-format on`.

Any comments on the idea of minimizing the first diff?

justbuchanan:
Thanks!

Do you still have your old clang-format config or do you remember anything in particular that was in there?

I adjusted the config to use tabs instead, which brings the format-everything [diff](https://github.com/justbuchanan/openscad/commit/7ec4a4802ff4cfebb03b3918fb0754fda7bfe157) down from ~31k lines changed to ~20k lines. Still pretty massive, but better. Other than that I don't have a whole lot of ideas for making the diff smaller.

Btw, I discovered a nifty website for interactively trying out config options - it makes it very easy to see what all the options do: https://zed0.co.uk/clang-format-configurator/.

kintel:
This is how far I got:
```

**Comment by kintel** [comment]
clang-format is missing a lot of options; it feels like it's designed to accommodate the codebases of a few large players. ..so slowly moving to a major coding style should make tools work better for us.

**Comment by kintel** [comment]
Anyway, if it too much work for too little gain let's just bite the bullet and make the switch.

**Comment by justbuchanan** [comment]
Cool, I tried that out that config and it brings the diff down to ~13k lines.

Agreed, it would be nice if it was more configurable. Some other formatters have more options (astyle, uncrustify), but clang-format generally seems to a better job (from personal experience) and is under more active development.

File: scripts/beautify.sh [hunk 1]
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+# Reformat C++ code using clang-format
+
+# note: with the -style=file option, clang-format reads the config from ./.clang-format
+# See here for the full list of supported options: https://clang.llvm.org/docs/ClangFormatStyleOptions.html
+FORMAT_CMD="clang-format -i -style=file"
+
+function reformat_all() {
+    find . -name "*.h" -o -name "*.cc" | xargs $FORMAT_CMD
```
kintel: We might want to add "*.hpp", "*.cpp" to this.
Also note that some code originates elsewhere and keeping the diff small is probably more valuable than coding style (e.g. `polyclipping/`, `CGAL_Nef3_workaround.h`, `CGAL_workaround_Mark_bounded_volumes.h`, `convex_hull_3_bugfix.h`, `OGL_helper.h`

kintel:
Also note that clang-format only looks at whitespace, so it cannot deal with other style choices (like allowing if statements without curly braces)

scripts/beautify.sh
justbuchanan: Good point, I went ahead and added those extensions.

The simplest way to exclude files if there are just a few would probably be to add `// clang-format off`/on comments them. I could also add a file blacklist to the script, but it would add a bit of complexity to get it to work with both `reformat_all` and `reformat_changed`.

kintel: The CGAL files could easily get `// clang-format *` added, but the polyclipping ones are verbatimely copied from upstream. Could we ignore that entire folder somehow?

kintel: Alternatively, we could move the folder out of the `src/` into a new `ext/` folder or smth.

justbuchanan: I ended up finding a relatively simple way to exclude a list of file patterns, so I think we can do that instead of having to modify or move those files.

scripts/beautify.sh
```diff
@@ -0,0 +1,44 @@
...
+# note: -v flag inverts selection - this tells grep to *filter out* anything
+#       that matches the pattern. For testing, you can remove the -v to ssee
+#       which files would have been excluded.
+FILTER_CMD="grep -v -E objects|src/polyclipping|src/CGAL_Nef3_workaround.h|src/CGAL_workaround_Mark_bounded_volumes.h|src/convex_hull_3_bugfix.h|src/OGL_helper.h"
+
+# C++ file extensions
+SRC_PATTERN=".*\.(h|hpp|cc|cpp)"
+
+function reformat_all() {
+    find . -regextype posix-extended -regex "$SRC_PATTERN" \
```
kintel: `-regextype` is a GNU extension and won't work on BSD-based systems like macOS.

justbuchanan: Darn, that's unfortunate... Now that you mention it, I think I've hit this issue before on my mac

kintel: `find -E` appears to have the equivalent function on BSD.

justbuchanan: I changed it back to the way I had it before.

justbuchanan: Unfortunately, it doesn't look like `-E` works on gnu `find` :/

scripts/beautify.sh
```diff
@@ -0,0 +1,41 @@
...
+FORMAT_CMD="clang-format -i -style=file"
+
+# Filter out any files that shouldn't be auto-formatted.
+# note: -v flag inverts selection - this tells grep to *filter out* anything
+#       that matches the pattern. For testing, you can remove the -v to ssee
+#       which files would have been excluded.
+FILTER_CMD="grep -v -E objects|src/polyclipping|src/CGAL_Nef3_workaround.h|src/CGAL_workaround_Mark_bounded_volumes.h|src/convex_hull_3_bugfix.h|src/OGL_helper.h"
+
+function reformat_all() {
+    find . -name "*.h" -o -name "*.hpp" -o -name "*.cc" -o -name "*.cpp" \
```
kintel: Perhaps limit the find to `src` for now to avoid it picking up too much other stuff

kintel:
Hum, unfortunately the clang-configurator you linked to doesn't support configuring `BraceWrapping`, which we use quite a bit.

justbuchanan:
Ahh, that is unfortunate.

You can get a similar edit-refresh going on your computer if you run [filewatcher](https://github.com/thomasfl/filewatcher) and open up the config file and a source file in the project. Something like `filewatcher .clang-format "clang-format -style=file -i src/somefile.cc"`, will re-format whenever you save the config file.

kintel:
Looking at this some more, I wonder if `uncrustify` (https://github.com/uncrustify/uncrustify) could be a better way of iteratively cleaning up and eventually changing our coding style.
We could start with a minimal config like:
```
indent_with_tabs=2
indent_columns=output_tab_size
indent_class=true
indent_namespace=true

nl_func_leave_one_liners=true

sp_angle_shift=remove
sp_permit_cpp11_shift=true
```

..and expand to a full config once we clean up existing code.
I've played around with a config which I could submit, but it's not complete.

Once we're in a better place, clang-format may be a good way to help getting our coding style more similar to other large code-bases.

justbuchanan:
Yes, that looks like a good option. The diff is definitely a lot smaller with your uncrustify config than with clang-format.

What are your thoughts on the pros/cons of doing this incrementally vs doing all re-formatting at once? (assuming you have a desired end-state in mind)

It seems to me that the main downside of any reformatting is the merge conflicts that you create for open pull requests. Any line of code that has been formatted on master after a given PR branch has branched off will be a conflict if the PR also touches that line. However, you can generally resolve most of these issues by also running the formatter on the PR branch before attempting to merge.

If you were to re-format everything all at once, you only have to consider the existing changes that have branched off from master. If you were to do it in two stages (just making up a number), then you need to consider the merge-conflicts you'll have to resolve now with the first formatting run and also the merge conflicts later on when you apply the second stage of formatting. I think in total, you'll end up having to resolve more merge conflicts doing this in steps vs doing it all at once.

If you're more comfortable doing this incrementally, I think that would be a fine option too.

kintel:
Doing it incrementally is mostly to be able to digest and review the changes. One option could be to create a feature branch, and gradually turn on and commit the incremental changes, but make the merge to master as one large PR. After that, I agree that running the script on any PR before merging would probably be required for existing PRs.

kintel:
I've created a branch `coding-style`: https://github.com/openscad/openscad/tree/coding-style
..and opened a PR against master from that branch.

Let's just push or PR changes to the coding-style branch until we feel enough ground has been covered to merge to master.

kintel:
Superseded by #2171

PR #2148, opened 2017-10-26
Make Inputdriver5 an experimental Feature

MichaelPFrey:
Added the inputdriver to the list of experimental features.
CONFIG+=experimental as we explicitly work on an experimental feature.
Add code to stop the drivers when the features gets disabled during run time.
Hide the UI when the feature is disabled.

I think it makes sense tobe able to disable the inputdriver(s), even when the feature becomes one day stable.

### DIFF
--- a/openscad.pro
+++ b/openscad.pro
@@ -1,4 +1,3 @@
-CONFIG+=experimental
 # Environment variables which can be set to specify library locations:
 # MPIRDIR
 # MPFRDIR
--- a/openscad.pro
+++ b/openscad.pro
@@ -1,3 +1,4 @@
+CONFIG+=experimental
 # Environment variables which can be set to specify library locations:
 # MPIRDIR
 # MPFRDIR
--- a/src/feature.cc
+++ b/src/feature.cc
@@ -27,7 +27,7 @@ const Feature Feature::ExperimentalForCExpression("lc-for-c", "Enable C-style <c
 const Feature Feature::ExperimentalAmfImport("amf-import", "Enable AMF import.");
 const Feature Feature::ExperimentalSvgImport("svg-import", "Enable SVG import.");
 const Feature Feature::ExperimentalCustomizer("customizer", "Enable Customizer");
-
+const Feature Feature::ExperimentalInputDriver("input drive", "Enable input drivers (requires restart)");
 
 Feature::Feature(const std::string &name, const std::string &description)
 	: enabled(false), name(name), description(description)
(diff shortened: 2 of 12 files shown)


src/feature.cc
```diff
@@ -27,7 +27,7 @@ const Feature Feature::ExperimentalForCExpression("lc-for-c", "Enable C-style <c
 const Feature Feature::ExperimentalAmfImport("amf-import", "Enable AMF import.");
 const Feature Feature::ExperimentalSvgImport("svg-import", "Enable SVG import.");
 const Feature Feature::ExperimentalCustomizer("customizer", "Enable Customizer");
-
+const Feature Feature::ExperimentalInputDriver("input-drive", "Enable input drivers (requires restart)");
```
kintel: Should this be "input-driver" ?

src/input/InputDriverManager.cc
```diff
@@ -129,6 +129,18 @@ std::string InputDriverManager::listDrivers()
     return stream.str();
 }
 
+void InputDriverManager::closeDrivers()
+{
+	stopRequest = true;
+	timer->stop();
+	InputEventMapper::instance()->stop();
+
+    for (drivers_t::iterator it = drivers.begin();it != drivers.end();it++) {
```
kintel: C++11 range iterators could be used here: `for (auto driver : drivers) {`.

Also: indentation mismatch (in multiple places in this PR)

openscad.pro
```diff
@@ -1,3 +1,4 @@
+CONFIG+=experimental
```
kintel: This shouldn't be the default

src/feature.cc
MichaelPFrey: typo - thank you for catching it

openscad.pro
MichaelPFrey: Reverted, but I am kind of split on this. Yes, this should not/never come into the master branch, we agree on that.
On the other hand: This is a feature branch were we are working on an experimental feature.
Someone that is checking out an experimental feature branch may not know that on top of checking out the experimental branch has to explicitly add +CONFIG+=experimental
(even when knowing it, it is an extra step)
I guess the input-driver is kind of a special feature branch

src/input/InputDriverManager.cc
MichaelPFrey: changed for loops to c++11 for loop
indentation matched to preexisting code

openscad.pro
kintel: Good point. Perhaps @t-paul could comment on this as we need a mechanism for remembering to revert such things before merging to master.

kintel:
@t-paul This is also yours - code was reviewed.

PR #2149, opened 2017-10-26
using for loops instead of copy&paste code

MichaelPFrey:
I think it is easier in the long run, when we use for-loops where possible instead of copy&paste code.
Offcourse, there are no input devices with an unlimited number of buttons, but the buttons on for e.g. a gamepad add up quickly.

"getSettingEntryByName" will also become infrastructure for things like axis calibration, where we need to save and restore values for currently nine axis. "getSettingEntryByName" allows the - as shown - to access the settings kind of like an array.

### DIFF
--- a/src/Preferences.cc
+++ b/src/Preferences.cc
@@ -224,16 +224,15 @@ void Preferences::init() {
         initComboBox(this->comboBoxRotationY, Settings::Settings::inputRotateY);
         initComboBox(this->comboBoxRotationZ, Settings::Settings::inputRotateZ);
         initComboBox(this->comboBoxZoom, Settings::Settings::inputZoom);
-        initComboBox(this->comboBoxButton0, Settings::Settings::inputButton0);
-        initComboBox(this->comboBoxButton1, Settings::Settings::inputButton1);
-        initComboBox(this->comboBoxButton2, Settings::Settings::inputButton2);
-        initComboBox(this->comboBoxButton3, Settings::Settings::inputButton3);
-        initComboBox(this->comboBoxButton4, Settings::Settings::inputButton4);
-        initComboBox(this->comboBoxButton5, Settings::Settings::inputButton5);
-        initComboBox(this->comboBoxButton6, Settings::Settings::inputButton6);
-        initComboBox(this->comboBoxButton7, Settings::Settings::inputButton7);
-        initComboBox(this->comboBoxButton8, Settings::Settings::inputButton8);
-        initComboBox(this->comboBoxButton9, Settings::Settings::inputButton9);
+
+        for (int i = 0; i < 10; i++ ){ 
+			std::string s = std::to_string(i);
+            QComboBox* box = this->centralwidget->findChild<QComboBox *>(QString::fromStdString("comboBoxButton"+s));
+			Settings::SettingsEntry* ent = Settings::Settings::inst()->getSettingEntryByName("button" +s );
+			if(box != 0 && ent != nullptr){
+				initComboBox(box,*ent);
+			}
+		}
 
 	SettingsReader settingsReader;
 	Settings::Settings::inst()->visit(settingsReader);
--- a/src/Preferences.cc
+++ b/src/Preferences.cc
@@ -226,13 +226,13 @@ void Preferences::init() {
         initComboBox(this->comboBoxZoom, Settings::Settings::inputZoom);
 
         for (int i = 0; i < 10; i++ ){ 
-			std::string s = std::to_string(i);
+            std::string s = std::to_string(i);
             QComboBox* box = this->centralwidget->findChild<QComboBox *>(QString::fromStdString("comboBoxButton"+s));
-			Settings::SettingsEntry* ent = Settings::Settings::inst()->getSettingEntryByName("button" +s );
-			if(box != 0 && ent != nullptr){
-				initComboBox(box,*ent);
-			}
-		}
+            Settings::SettingsEntry* ent = Settings::Settings::inst()->getSettingEntryByName("button" +s );
+            if(box && ent){
+                initComboBox(box,*ent);
+            }
+        }
 
 	SettingsReader settingsReader;
 	Settings::Settings::inst()->visit(settingsReader);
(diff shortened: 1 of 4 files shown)


src/settings.cc
```diff
@@ -202,6 +202,16 @@ void Settings::visit(SettingsVisitor& visitor)
 	}
 }
 
+SettingsEntry* Settings::getSettingEntryByName(std::string name)
+{
+	for (std::list<SettingsEntry *>::iterator it = entries.begin();it != entries.end();it++) {
```
justbuchanan: Since C++11, you can also do something like `for (SettingsEntry* entry : entries) {` and avoid having to deal with iterators directly.

kintel: Agreed - prefer C++11 for all new code.

kintel: Going down the C++11 route, this would be even better: `for (auto entry : entries) {`

src/settings.cc
```diff
@@ -202,6 +202,16 @@ void Settings::visit(SettingsVisitor& visitor)
...
+		if(((*it)->name().compare(name))== 0){
```
kintel: You can use the `==` operator. Assuming the for loop gets fixed: `if (entry->name() == name) return entry;`

src/settings.h
```diff
@@ -76,6 +76,7 @@ class Settings
     static Settings *inst(bool erase = false);
 
     void visit(class SettingsVisitor& visitor);
+    SettingsEntry* getSettingEntryByName(std::string name);
```
kintel: `const std::string &name`

src/settings.cc
MichaelPFrey: I try to match the preexisting code style, in this case the visit function.
Well: changed for my function, preparing a pull-request for the visit function in the master branch.
(changing the visit function in a feature branch would not make much sense)

src/input/InputEventMapper.cc
```diff
@@ -149,16 +149,11 @@ int InputEventMapper::parseSettingValue(const std::string val)
...
-    actions[4] = QString(s->get(Settings::Settings::inputButton4).toString().c_str());
-    actions[5] = QString(s->get(Settings::Settings::inputButton5).toString().c_str());
-    actions[6] = QString(s->get(Settings::Settings::inputButton6).toString().c_str());
-    actions[7] = QString(s->get(Settings::Settings::inputButton7).toString().c_str());
-    actions[8] = QString(s->get(Settings::Settings::inputButton8).toString().c_str());
-    actions[9] = QString(s->get(Settings::Settings::inputButton9).toString().c_str());
+    for (int i = 0; i < 10; i++ ){
+		std::string is = std::to_string(i);
+		Settings::SettingsEntry* ent =s->getSettingEntryByName("button" +is);
+		actions[i] =(s->get(*ent).toString().c_str());
```
kintel: This may be a bit nicer: `actions[i] = QString::fromStdString(s->get(*ent).toString())`

src/Preferences.cc
```diff
@@ -224,16 +224,15 @@ void Preferences::init() {
...
-        initComboBox(this->comboBoxButton4, Settings::Settings::inputButton4);
-        initComboBox(this->comboBoxButton5, Settings::Settings::inputButton5);
-        initComboBox(this->comboBoxButton6, Settings::Settings::inputButton6);
-        initComboBox(this->comboBoxButton7, Settings::Settings::inputButton7);
-        initComboBox(this->comboBoxButton8, Settings::Settings::inputButton8);
-        initComboBox(this->comboBoxButton9, Settings::Settings::inputButton9);
+
+        for (int i = 0; i < 10; i++ ){ 
+			std::string s = std::to_string(i);
+            QComboBox* box = this->centralwidget->findChild<QComboBox *>(QString::fromStdString("comboBoxButton"+s));
```
kintel: Indentation looks a bit off here

src/Preferences.cc
```diff
@@ -224,16 +224,15 @@ void Preferences::init() {
...
+			Settings::SettingsEntry* ent = Settings::Settings::inst()->getSettingEntryByName("button" +s );
+			if(box != 0 && ent != nullptr){
```
kintel: s/0/nullptr/ .. or simply `if (bot && ent)`

src/settings.cc
kintel: Yup, sometimes it's easier to follow existing style, sometimes it's easier to update all of it once touching a file. We made the switch to C++11 a while back, but there is still some pieces of code lagging behind.

src/Preferences.cc
MichaelPFrey: This is a mix of tab and space indentation.
I now changed the settings in text editor.
Fixed indentation commited.

src/Preferences.cc
MichaelPFrey: Thank you. I am coming from a Javascript background - that left some paranoia when checking for zero and null.
Change commited.

src/Preferences.cc
kintel: Note: We actually use tab indentation. Existing spaces could be cruft left from the previous committer.

(we'll eventually make the switch to spaces; work in progress)

src/input/InputEventMapper.cc
MichaelPFrey: Do you really want to use a QT-Functions where we are not interacting with the UI/QT?
Sure, using the input-driver without UI would not make sense, but still: The InputEventMapper is low-level code that could be used independent of GUI framework beeing used.

kintel: In general not, but in this case, `actions` is an array of `QString`:
```
    QString actions[10];
```

MichaelPFrey: understood, thank you.

src/settings.h
MichaelPFrey: changed

kintel:
@t-paul This is yours - the code has been reviewed.

src/Preferences.cc
MichaelPFrey: thank you for the heads up. Note: We are here in a feature branch.

I would like to suggest, that we leave "code cosmetics" "as is" for now in this feature branch.
I agree that matching the indentation with the master branch and things like that have to be a requirements for the pull request from the feature branch into the master branch, but currently I would prefer focusing on code quality and features.

kintel: Fair enough. I'll leave detail to @t-paul who currently "owns" the feature branch.

t-paul: So far I've seen git handling whitespace only changes quite well. That said, I agree with @MichaelPFrey, it's probably still better to do style cleanup separately.

PR #2150, opened 2017-10-27
Test PR to debug osx builds on travis

justbuchanan:
See #2138

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,3 +262,4 @@ If you had problems compiling from source, raise a new issue in the
 
 This site and it's subpages can also be helpful:
 http://en.wikibooks.org/wiki/OpenSCAD_User_Manual/Building_OpenSCAD_from_Sources
+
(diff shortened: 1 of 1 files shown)


justbuchanan:
I think I have this figured out. The reason that the master branch build wasn't cached was because there wasn't a cache available yet on the master branch and travis doesn't re-use caches from PR branches (even the PR that the merged commits came from).

If you look at the 3 commits on this PR branch, the first two were started at around the same time and neither one used a cache. This is because there wasn't a master branch cache available and this was a fresh pull request. The third commit was started later and used the cache created by the first commit's build and ran very quickly.

The master branch cache won't be populated until a build on the master branch succeeds, which is less likely to happen without the cache... sort of a chicken-and-egg problem. Once the first cache-less build passes though, it should work for subsequent builds.

kintel:
ok, I finally got a master build to succeed at 48m9s - I'll trigger a new one and see if it works.

justbuchanan:
Success! The next build on master was cached and ran in ~18 minutes.

https://travis-ci.org/openscad/openscad/jobs/293757332

PR #2151, opened 2017-10-27
c++11 style for-loop

MichaelPFrey:
See pull request #2149

### DIFF
--- a/src/settings.cc
+++ b/src/settings.cc
@@ -89,9 +89,9 @@ Settings::~Settings()
 
 void Settings::visit(SettingsVisitor& visitor)
 {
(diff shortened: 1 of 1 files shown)


src/settings.cc
```diff
@@ -89,9 +89,9 @@ Settings::~Settings()
 
 void Settings::visit(SettingsVisitor& visitor)
 {
-	for (std::list<SettingsEntry *>::iterator it = entries.begin();it != entries.end();it++) {
```
kintel: indentation mismatch

MichaelPFrey: fixed - the file was using a mix of four space and tab indentation

PR #2152, opened 2017-10-27
modifing the MXE download link

delpiro:
the file mxe can't be download with the old download link


kintel:
@t-paul We're using our own MXE for a good reason, right?

delpiro:
git clone git://github.com/openscad/mxe.git if you try this it is not working, I changed it 
git clone https://github.com/openscad/mxe.git
and it worked, please check if the download link is working or not.

t-paul:
Yes, MXE is now rolling update, so without fully automated builds it's could break at inconvenient times. Also there's some minimal patches that are not upstream yet.

Using the `https` link in the scripts makes sense, that's probably more compatible nowadays and it's usually not meant to push back changes via that clone.

kintel:
@delpiro Your patch is currently empty..

PR #2155, opened 2017-10-29
use constants for array sizes in the inputdriver

MichaelPFrey:
I think we agree that using constant for array sizes is better then magic numbers.
How ever: I am uncertain how and where in openscad constants shall be declared.
I have no problem to change it how ever you like it.

Note that I lowered the number of axis intentionally from 10 to 9.
The reason is, that the UI currently only supports axis 0..8.
Nine axis has also the benefit, that I can arrange them in the UI in a nice 3x3 grid

This pull request is kind of both a question and a preparation for further code.
(I need these constants for trim and deadzone correction. Will also increase the number of buttons at some point.)

### DIFF
--- a/src/input/InputEventMapper.cc
+++ b/src/input/InputEventMapper.cc
@@ -37,10 +37,10 @@ InputEventMapper::InputEventMapper()
 {
     stopRequest=false;
 
-    for (int a = 0;a < MAX_AXIS;a++) {
+    for (int a = 0;a < max_axis;a++) {
         axisValue[a] = 0;
     }
-    for (int a = 0;a < MAX_BUTTONS;a++) {
+    for (int a = 0;a < max_buttons;a++) {
         button_state[a]=false;
         button_state_last[a]=false;
     }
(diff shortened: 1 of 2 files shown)


t-paul:
Considering the slow move to slightly more modern C++, for new code we probably want to use `static const int` as those can be scoped in classes or namespaces and such.

Things like dead zone may need to be a configuration option later though, e.g. I had no issue with the current fixed settings where @MichaelAtOz did see the model drifting in center position.

MichaelPFrey:
understood and I fully agree.
changed "define" to "const static int".

I have the code for deadzone already prepared, but I prefer to get an consent on "infrastructure" before moving forward. I need the deadzone and trim to correct my gamepad, which has both issues.

t-paul:
Sounds good. My view is that we should get some reasonable baseline working on the 3 platforms to get it merged. I suspect there's going to be more feedback after that. I'll post some pointers in the main places, maybe that will get people interested in trying it already with the branch snapshot.

I'll try to find some time to detail my point of view for merging to master as base for discussion (in the main issue). The dead zone seems to be one point that would be needed or a number of people/devices.

PR #2156, opened 2017-10-29
trim and deadzone for inputdriver 5

MichaelPFrey:
trim and deadzone for inputdriver 5.
The UI is allready prepared, but I would like to get feedback on how I store the values for the trim and deadzone.
It is kind of tricky, that the ValueType NUMBER is actually a double, but beeing used as an int.

### DIFF
--- a/src/Preferences.cc
+++ b/src/Preferences.cc
@@ -964,6 +964,10 @@ void Preferences::updateButtonState(int nr, bool pressed) const{
 	label->setStyleSheet(Style);
 }
 
+void Preferences::AxesChanged(int nr, double val) const{
+	//will follow
+}
+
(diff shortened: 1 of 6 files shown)


t-paul:
I don't see a big issue yet regarding the double vs. int as the user visible part is defined by the UI. Tiny offsets in actual storage as double should not matter much regarding the actual calculation. I'd vote for just using as it is and see how that goes. If really needed we can also consider some changes in the settings code.

MichaelPFrey:
Thank you.

I can leave with the int/double workaround do, I just wanted to leave a comment for further developers that this is an artifact, not a careful design decision. (as opposed to some very specific oddities in the source code, that look like artifacts, but are actually a brilliant implementation)

PR #2160, opened 2017-10-30
hidapi_decode_axis1 using the InputEventAxis

MichaelPFrey:
This allows the use of the axis mapping.

Note: This was tested with an SpaceExplorer which reports values between -350 and +350.
This is in conflict to the comment just above in the code:
> // Values are in the range -10..10 at min. speed and -2595..2595 at max. speed.

Maybe we need to split hidapi_decode_axis again for different mice reporting different ranges or the range is somehow OS dependent. Unfortunately, I can neither cross compile nor use a different mice.

### DIFF
--- a/src/input/HidApiInputDriver.cc
+++ b/src/input/HidApiInputDriver.cc
@@ -81,8 +81,6 @@ void HidApiInputDriver::run()
  */
 void HidApiInputDriver::hidapi_decode_axis1(const unsigned char *buf, unsigned int len)
 {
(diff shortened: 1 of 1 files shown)


t-paul:
IIRC this is actually a setting which is available via the official driver, with 350 probably being the default. As we currently have no means to set that, we may need to allow that as config too for people using their 3d controllers also with the proprietary driver. But I guess that's a matter of testing with various input devices.

PR #2161, opened 2017-10-30
Inputdriver5 hidapi update

MichaelPFrey:
Inputdriver5 hidapi update from https://github.com/signal11/hidapi

added the settings required for HIDAPI under linux into the openscad.pro file. (commented out as we can not ask our users to run/test openscad with root privileges just for 3D mice support)

Note that the current /src/hidpapi/hid.c file should better be in the path /src/hidpapi/windows/hid.c, but I do not know which uncommited code is coded to the current path.

### DIFF
--- a/openscad.pro
+++ b/openscad.pro
@@ -539,6 +539,15 @@ unix:!macx {
   SOURCES += src/input/DBusInputDriver.cc
 }
 
+#unix:!macx {
+#DEFINES += ENABLE_HIDAPI
+#HEADERS += src/input/HidApiInputDriver.h
+#SOURCES += src/input/HidApiInputDriver.cc
+#HEADERS += src/hidapi/hidapi.h
+#SOURCES += src/hidapi/hid.c 
+#LIBS += `pkg-config libudev --libs` -lrt
+#}
+
 unix:!macx {
   DEFINES += ENABLE_JOYSTICK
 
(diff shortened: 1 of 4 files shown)


t-paul:
I'm not sure we want to include Linux HIDAPI, that should be available as `libhidapi-dev` (on Debian).
Otherwise the update is a good idea, the included one is quite old. I think at this time it only was a single file for both Windows and MacOS.
There can't be any other uses than this branch, so it's basically free to shuffle around as needed.

MichaelPFrey:
Thank you - I will look into it.
Having a dependency sounds better the including more and more files.

I think that it would be helpful to have a "simple" way for developer to use the HIDAPI under linux, even if we never plan on shipping HIDAPI support under linux.

I am careful with shuffling files, as I do not have a cross compile toolchain, so I can not check whether or not it breaks something.

t-paul:
Yes, having easy dev support for HIDAPI on Linux too is quite useful. Looking at `hidapi.pri`, it should enable for Linux if `pkg-config --silence-errors --cflags hidapi-libusb` returns something (which it does on my system: `-I/usr/include/hidapi`).

MichaelPFrey:
my approach in the pull-request does not make sense as discussed with t-paul.
Discussing the build-chain is would be a separate issue, but is currently not necessary for me.

PR #2168, opened 2017-10-31
Inputdriver5 UI improved

MichaelPFrey:
- display of current axis reading
- trim and deadzone setable via UI
- split the UI for button and axis
- increased number of buttons


t-paul:
Nice. I'll trigger a new Windows build...

PR #2172, opened 2017-11-01
customizer: try to minimize spacing

MichaelPFrey:
bildschirmfoto von 2017-11-01 08-52-20
bildschirmfoto von 2017-11-01 08-52-58


MichaelAtOz:
Hallelujah, someone who understands the value of screen real estate..., especially vertical space. It is a good first attempt...I encourage further savageness... ;)

kintel:
Thanks - looks real good!

PR #2175, opened 2017-11-02
Coding style part1

kintel:
This is as small of a patch I can meaningfully produce.
It still does some aggressive indentation some places, but I don't know if that can be dealt with; at least all indenting should be auto-generate now.

Generated by `./scripts/beautify.sh --all --uncrustify`

### DIFF
--- a/src/parsersettings.h
+++ b/src/parsersettings.h
@@ -13,6 +13,6 @@ extern int parser_error_pos;
 void parser_init();
 
 fs::path search_libs(const fs::path &localpath);
-fs::path find_valid_path(const fs::path &sourcepath, 
-                         const fs::path &localpath, 
-                         const std::vector<std::string> *openfilenames = nullptr);
+fs::path find_valid_path(const fs::path &sourcepath,
+												 const fs::path &localpath,
+												 const std::vector<std::string> *openfilenames = nullptr);
--- a/.uncrustify.cfg
+++ b/.uncrustify.cfg
@@ -3,8 +3,9 @@ input_tab_size=2
 output_tab_size=2
 indent_columns=output_tab_size
 indent_class=true
-indent_namespace=false
+indent_namespace=true
 indent_extern=true
+indent_align_assign=false
 
 nl_func_leave_one_liners=true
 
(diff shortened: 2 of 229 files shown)


kintel:
@justbuchanan

src/parsersettings.h
```diff
@@ -13,6 +13,6 @@ extern int parser_error_pos;
 void parser_init();
 
 fs::path search_libs(const fs::path &localpath);
-fs::path find_valid_path(const fs::path &sourcepath, 
-                         const fs::path &localpath, 
-                         const std::vector<std::string> *openfilenames = nullptr);
+fs::path find_valid_path(const fs::path &sourcepath,
+												 const fs::path &localpath,
+												 const std::vector<std::string> *openfilenames = nullptr);
```
justbuchanan: This indent seems a little large.

justbuchanan:
Looks good! I like that uncrustify has an "ignore" option so you can direct it to only fix certain types of style mistakes.

I scanned through the diff and overall it looks good. I noticed one minor indenting issue in src/parsersettings.h (and possibly other places) where it added a large indent before function parameter declarations. Not a big deal, but maybe there's a related option to adjust?

thehans:
I don't have anything to add about specific settings, but I saw this and thought it might be pretty handy if you are trying to tweak very specific indenting options: http://universalindent.sourceforge.net/features.php

src/parsersettings.h
kintel: It's because the github web viewer default to a tab size of 8, while i use 2.

kintel: ..which is one of the reasons why we shouldn't have used tab for indenting in the first place : /

kintel: https://github.com/openscad/openscad/blob/e1c1ba37fa717f882cc44a7eac70be9a83a1dd7e/src/parsersettings.h?ts=2

kintel:
@justbuchanan Unfortunately, the ignore option doesn't work very well, I tried to ignore what I could to keep the patch small. If you see anything I missed I'd be interested.

kintel:
@thehans Universalindent looks slightly out of maintenance. Both uncrustify and clang-format are actively developed tools; we just need to establish a good config.

PR #2178, opened 2017-11-04
Coding style part2

kintel:
* Added spacing rules for `&` and `*`, operators, assignment and comma
* Always add a newline before class braces
* if-else without braces should be one-liners


kintel:
@justbuchanan

PR #2179, opened 2017-11-05
Coding style part3

kintel:
* Template angle spacing
* Operator spacing rules
* Function paren space rules
* brace newline and spacing rules

### DIFF
--- a/.uncrustify.cfg
+++ b/.uncrustify.cfg
@@ -56,4 +56,8 @@ sp_after_operator=remove
 sp_after_operator_sym=remove
 sp_after_operator_sym_empty=remove
 
+sp_func_proto_paren=remove
+sp_func_def_paren=remove
+sp_func_call_paren=remove
+
 align_right_cmt_span=0
--- a/.uncrustify.cfg
+++ b/.uncrustify.cfg
@@ -52,4 +52,8 @@ sp_angle_word=force
 sp_angle_shift=remove
 sp_permit_cpp11_shift=true
 
+sp_after_operator=remove
+sp_after_operator_sym=remove
+sp_after_operator_sym_empty=remove
+
 align_right_cmt_span=0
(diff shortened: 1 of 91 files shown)


kintel:
@justbuchanan

src/CGALCache.h
```diff
@@ -8,7 +8,7 @@
 class CGALCache
 {
 public:
-	CGALCache(size_t limit = 100*1024*1024);
+	CGALCache(size_t limit = 100 *1024 *1024);
```
justbuchanan: It's up to you, but do you think it would make more sense if there were spaces on both sides of the operators?

src/CGAL_Nef_polyhedron.cc
```diff
@@ -94,20 +94,20 @@ PolySet *CGAL_Nef_polyhedron::convertToPolyset() const
 #endif
 
 void CGAL_Nef_polyhedron::resize(const Vector3d &newsize,
```
justbuchanan: It didn't change in this PR, but is this function signature right? It looks like it's missing a closing parentheses and perhaps more args? (there's a trailing comma)

src/CGALCache.h
kintel: I saw that, I think this is a bug in uncrustify where it mistakes this to be a pointer as it's inside a function declaration. Couldn't find my way around it.

src/CGAL_Nef_polyhedron.cc
kintel: This is just because you're viewing github diffs with a large tab size. The rest of the function parameters are indented to match the parenthesis.

src/CGALCache.h
justbuchanan: Ahh ok. After I commented on this line, I scrolled down further and saw that in most places it had spaces on both sides as expected.

src/CGAL_Nef_polyhedron.cc
kintel: -> https://github.com/openscad/openscad/pull/2179/files?ts=2#diff-b626954cb7b549ae78b89800cf925ba7L96

justbuchanan: Oops, my mistake, I did the same thing the other day...

Issue #2105, opened 2017-09-09
File Loading Error in 2017 windows snapshots


runsun:
Errors when loading files in different window snapshots (including all 2017 versions). 
(Windows 10 standard edition, 64 bit)

Upon loading, a popup window says:

Popup up Critical Error: A critical error was caught. The application may have become unstable: vector:: _MN_range_check: __n(which is 0) >= this -> size() (which is 0)

This is lethal. After it, every file, even a file containing as simple as cube([1,2,3]); failed to draw.

It seems to be happening on more complicated files (like files that import large files) but I am not sure.

This error was introduced since OpenSCAD-2016.11.11-x86-64-Installer.exe

Snapshots tested:

2015.03-2 (lastest release): OK
OpenSCAD-2015.10.17-x86-64-Installer.exe : OK
OpenSCAD-2016.01.26-x86-64-Installer.exe : OK
OpenSCAD-2016.02.01-x86-64-Installer.exe : OK
OpenSCAD-2016.02.09-x86-64-Installer.exe : OK
OpenSCAD-2016.04.06-x86-64-Installer.exe : OK
OpenSCAD-2016.05.12-x86-64-Installer.exe : OK
OpenSCAD-2016.07.20-x86-64-Installer.exe : OK
OpenSCAD-2016.09.18-x86-64-Installer.exe : OK
OpenSCAD-2016.10.04-x86-64-Installer.exe : OK
OpenSCAD-2016.11.11-x86-64-Installer.exe : ERROR
OpenSCAD-2017.01.20-x86-64-Installer.exe : ERROR 
OpenSCAD-2017.08.30-x86-64-Installer.exe : ERROR

t-paul:
Hmm, that might be related to the Customizer. Can you reproduce that with some specific file? If yes, does it happen with the Customizer disabled (in Preferences) and could you share one of those files? So far I could not reproduce the issue.

runsun:
1) The files tested is a lib that imports more than 10 files and total having more than 21,000 lines. By itself, it runs ok.  The error appeared when it is imported by another file that is 2000 lines by itself. So I am suspecting file size problem. 

WIll try to make files of minimal size that still produced the error.
 
2) No error if Customizer turned off.

t-paul:
So maybe it's worth having a go with the full big bundle of files, if those can be shared somehow. For geometry it's often extremely helpful to minimize the data to search in. In this case there's a chance, it's not so important.

runsun:
[...]

PR #2183, opened 2017-11-06
BUGFIX: customizer crashes with empty vector, refractored checkVectorWidget() method

MichaelPFrey:
How to reproduce:
  * open OpenSCAD
  * enable customizer
  * enter
    * vector0=[];
  * press F5
crash

The refactoring is not necessary for the fix, but it makes sense to do it as that was the reason why I found that bug.


t-paul:
Thanks, I think that fixes #2105.

PR #2191, opened 2017-11-11
comment parser refractored

MichaelPFrey:
The current comment parser tries and manages to parse things, it really shouldn't:
```
a=2; // [[1,2,3],[4,5,6],[9,5,4,2,3,1]]
echo(a);
d=4; // [[1,2,3]:a,50:d]
echo(d);
```
Looking into the details, due to the recursive nature, I can imagine that internally, the parse can generate a lot of interesting data-structures (which also makes unclear to a human reading the parser, what the parser is actually supposed to do).

The restructured code should be easier to read.

### DIFF
--- a/src/comment.cpp
+++ b/src/comment.cpp
@@ -286,7 +286,7 @@ void CommentParser::collectParameters(const char *fulltext, FileModule *root_mod
  
 		// Extracting the parameter comment
 		std::string comment = getComment(std::string(fulltext), firstLine);
-		// getting the node for parameter annnotataion
+		// getting the node for parameter annotation
 		shared_ptr<Expression> params = CommentParser::parser(comment.c_str());
 		if (!params) {
 			params = shared_ptr<Expression>(new Literal(ValuePtr(std::string(""))));
--- a/src/comment_parser.y
+++ b/src/comment_parser.y
@@ -1,125 +1,133 @@
+%glr-parser
+
 %{
-    #include <sstream>
-    #include <string.h>
-    #include "Assignment.h"
-    #include "expression.h"
-    #include "printutils.h"
-    #include "value.h" 
-    #include "comment.h" 
-    void yyerror(const char *);
-    int comment_lexerlex(void);
-    int comment_parserlex(void);
-    extern void comment_lexer_scan_string ( const char *str );
-    Expression *params;
+	#include <sstream>
+	#include <string.h>
+	#include "Assignment.h"
+	#include "expression.h"
+	#include "printutils.h"
+	#include "value.h" 
+	#include "comment.h" 
+	void yyerror(const char *);
+	int comment_lexerlex(void);
+	int comment_parserlex(void);
+	extern void comment_lexer_scan_string ( const char *str );
+	Expression *params;
 %}
 %union {
-    char *text;
-    char ch;
-    double num;
-    class Vector *vec;
-    class Expression *expr;
+	char *text;
+	char ch;
+	double num;
+	class Vector *vec;
+	class Expression *expr;
 };
 
 
 %token<num> NUM
-
 %token<text> WORD
+
 %type <text> word
 %type <expr> expr
+%type <expr> num
+%type <expr> value
+%type <vec> values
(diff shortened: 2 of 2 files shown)


kintel:
I haven't looked at the code, but this seems to change behavior causing tests to fail

MichaelPFrey:
Thank you - I will look into it.
I have a suspicion what it could be.
Will working on the UI, I found an undocumented feature that my implementation did not support.

**Question in general:**
Do you agree with the general idea of making the comment_parser "more restricted"/"less powerful"?
The current parser is nice, as it can create a lot of powerful constructs for future use.
On the other hand, the current UI does not expect does constructs, resulting in unexpected behavior.
Now we could adjust either the UI code or the parser.

t-paul:
That's a difficult question actually. Personally I'm thinking the comment solution should stay as it is, as it basically mirrors what the Thingiverse customizer allows. Adding features here may then cause confusion why that does not work in Thingiverse.
Mid term, I'd prefer to have a fully integrated set of annotation (or whatever we want to call those) in the normal language. That would be the "native" customizer solution which could support any additional features we can come up with, and also include the parameter/value export for use by external customizer UIs (who knows, maybe even our own web based customizer?).

@kintel what's your view on that?

kintel:
I agree with @t-paul that this solution should aim to be as close to Thingiverse's solution as we can. If your patch helps going in that direction, I say let's apply it once tests pass.

t-paul:
Oh, just to clarify the "stay as it is" was only meant feature wise. Any improvement (readability, error handling, ...) on the comment parser is totally welcome of cause.

MichaelPFrey:
Thank you for the hint with the testing.
I now realized, that the comment parser is not only used for the customizer, but also the AST generation used for testing.
Reading other issues, there is also the idea to create a reformat feature based on AST.
Meaning: It is necessary to fully parse the comments, so that the comment can be recreated from the AST-Tree.
Therefor, my approach of simplifying the parser to the minimum required for the customizer is incorrect

t-paul:
I suspect this separate parser will stay for a while, so this patch still seems like a good idea if it's not too complicated to fix the test failure. Looking at the failures, at least some cases are numbers returned as strings.

MichaelPFrey:
It is a rabbits hole, but who am I telling that?
The first offender I found was one like this:
 a = true;//parameter
I specifically disabled the parsing of single words. as we currently to not use that for the customizer.
An other offender is here, just with a number:
 x1=1; //12

Going by patters, I would also have to handle:
 a = true;//this parameter number 1
meaning a recursion for words.
No big deal, but specifically the kind of recursive definition that I did not have in mind, as the data parsed and collected is not used in the UI.

Going to the bigger picture, the current parser has no issue with that:
 a=2; // [[1,2,3],[4,5,6],[9,5,4,2,3,1]]
Turning it into a (presumably) valid AST structure.
Originally, I planned on rejecting such inputs as simply malformed, simply failing.
No big deal. But knowing that the AST tree is used for regression testing and might get other uses in the future (e.g. reformating), I have handle that "somehow".

Note that I had to add
 %glr-parser
as by syntax,
 [0:10]
could be either a range or a labeled combo box for numbers, with just one value.
 [1:10,2:20]
I do not mind enabling glr-parser for short strings parsed for the costumizer, but for a general purpose comment parser? I really have to think about that one.

The next thing that crossed my mind was future proving the comment parser/customizer. Meaning making the parser "ignore" unknown expressions like
 r1=1; //[1:12]  show_if(sphere)
 n1=1; //[1:12]  spinbox
 r2=1; //show_if(sphere) [1:12] 
 n2=1; //spinbox [1:12] 
With my approach, that will be hard.
With the current approach, that is just tricky.

I also did not know at the time I started reworking the parser, that is parser is derived from the main parser definition. That is a plus when someone is working on both. as the comment parser is currently just a stripped down version of the main parser. (while mine is mostly rewritten, forcing people to learn two clearly distinct parsers)

A lot of things to keep in mind with no simple hot fix.
I will think about that and then create a new pull request when I found an adequate solution.

Issue #899, opened 2014-08-03
Failing csgpngtest_text-font-direction-tests on travis


t-paul:
That might be caused by not selecting the correct font...

kintel:
It seems clear that this was a freetype bug. Closing

PR #2198, opened 2017-11-19
Passing tests moved out of bugs folder, into issues.

thehans:
All 4 related issues are closed, fixed already.


thehans:
Well, these all passed on my system, but looks like they all still fail on Travis.  I guess they're still in the bugs directory for a reason.  Shall I just close this?

kintel:
It would be nice to understand why this fails
Somehow the automated html test report upload started failing at some point..

kintel:
ah, right, uploads don't work for forked PRs for security reasons.

kintel:
See #2199 when it builds - you should be able to see exactly what fails

thehans:
In Issue #899 you closed this with a comment that it was a bug in freetype.  So I guess it was fixed upstream, but the Travis is still on older version.   I figured this one might fail since Travis is on an older Ubuntu repo than me, but was worth trying.

As for the others, no idea until report is done.

thehans:
OK, so it looks like the #899 issue is actually that the expected images are from an overhead 2D viewpoint, but the actual image is from 3D viewpoint.  I thought I might need to put it in a `testdata/scad/2D/issues` folder, but there wasn't one, so I had added it to `/testdata/scad/3D/issues` instead.  I guess I need to modify `/tests/CMakeLists.txt` a bit to properly handle 2D issues.

The other ones are all missing throwntogether expected images.  It seems that "bugs" do not have throwntogether tests run on them, but "issues" do.   I noticed I had to do this with PR #2123 also.
@kintel should the "bugs" be changed to also test throwntogether?

kintel:
We could change bugs to also test throwntogether. No real reason not to, probably just something we forgot. It's likely easy to add, just a bit annoying to go back and add those expected images.

thehans:
@kintel I've tried adding ISSUES_2D_FILES and running it along with other 2D tests, but it is still failing with the wrong camera angle on my end.  I guess I don't know what determines this camera angle for the 2D tests.

kintel:
Thanks a lot of doing the dirty work!

PR #2203, opened 2017-11-21
Inputdriver5: preferences debugged and cleaned up

MichaelPFrey:
using a stepping between 0 and 1 as indicator for double is a clean and simple solution with some welcome benefits.

Zoom Gain Limit risen from 9.99 to 99.9.


MichaelPFrey:
feel free to squash the commits

PR #2207, opened 2017-11-30
Fix: customizer should not save an empty json file when the user is using "save as"

MichaelPFrey:


kintel:
One question: If there already was a JSON file, but someone removed the last parameter set, what happens then?

MichaelPFrey:
parameterWidget->writeFileIfNotEmpty gets only ever called by MainWindow::actionSaveAs()

If the someone removes the last parameter set of a json file, an essentially empty json file is saved.
This handled by a different method. I consider that the consistent/expected behaviour, as I do not want that OpenSCAD starts deleting files.

MichaelPFrey:
To be honest: There is one corner case. Let us say there are three files in a folder:
   * a.scad
   * b.scad
    * b.json
When the user would open a.scad and then "save as" "b.scad", the b.json files will not be overwritten (with essentially no content). Someone might see that as an issue - but that would be a separate issue that is affecting less users then the issue beeing fixed.

kintel:
This sounds good for now

PR #2211, opened 2017-12-08
csgmode cleanup

thehans:


thehans:
Looks like the travis mac build hit a time limit at the very end, but the log shows that all tests passed.

Issue #2206, opened 2017-11-30
cheatsheet or part of documentation should be part of app to work in offline mode [$25]


jtoy:
I use the app offline often and when I click on any of the help items, it takes me online which i cant go online. I think at least some basic stuff like the cheatsheet should be part of the app so people can use the app without having to be online.

##

donbright:
i find this request very interesting. would you be willing to tell more about your use case? what kind of setting and what kind of hardware / operating system? 

i think it is not too difficult, qt includes html rendering that we probably use somewhere already

jtoy:
The usecase is having developers be able to use the app in a fully self contained way. I am  offline often and when I use go to use the app I cannot look up any documentation because it just links to online. Not everyone has a 24/7 connection to the internet.

t-paul:
There's also a script that scrapes the Wiki manual and creates a standalone version. But there's no good idea yet how to integrate that. We can't scrape on every build.

thehans:
@t-paul Can you clarify why we can't?  Do you mean because it would make the build susceptible to vandalism?  If that's the issue, is it possible to "tag" a revision of the wiki which is known good?

t-paul:
Even just OBS is running about 40 builds for every check-in. I don't want to be blamed for DDOS-ing Wikibooks if we have multiple check-ins.
Well, OBS might not allow network access in the first place, so instead of scraping every time it probably would just fail. Not sure what's a good solution, maybe some kind of staging in a separate repo?

donbright:
yeah i didnt want to deal with the wikibook, thats too much, and too big. this is just for the minimal cheatsheet. given the cheatsheet itself, a reasonably dedicated user could figure out most of the stuff in the wikibook.

thehans:
BTW #2177, while not the same issue, would more or less address the root of the problem in my opinion.

MichaelAtOz:
$25 Bounty added.

ixil:
https://github.com/ixil/openscad_offliner So I was bored sometime last month - then forgot about it - then remembered and so here is this? 
it may be wip - I can't actually recall whether I finished this or not, let me know

t-paul:
Implemented via #3853.

@abshk-jr Looks like we missed this issue with the bounty attached, feel free to claim it.

[...]

PR #2213, opened 2017-12-12
provide minimal cheatsheet documentation, for offline + broken systems

donbright:
this is a proposed fix for issue #2206 .. if the user is offline and loads the cheatsheet, it will auto detect this fact and load an offline copy of the Cheatsheet. 

it works by storing the cheatsheet inside the qt resources. it uses QT network functions to detect network connection. if the system is offline, then openscad unpacks the cheatsheet from its resources and saves it under the platform specific Documents/OpenSCAD  folder and opens it with the local web browser.

it would be 'cleaner' to make this cheatsheet file a separate part of the installation process... but this design is so that the user will be guaranteed to have basic minimal documentation, even on the most broken system. as long as they have the openscad binary, they can still have basic cheatsheet reference material.


kintel:
The packaging looks good!
My concern with this is that we're duplicating content from the website repo, so this is bound to get out of sync. A minor additional point is that the cheat sheet links to wikibooks, which is unlikely to be available when offline.

To remedy the first, could we add the cheatsheet as a subrepo or smth. (e.g. pull it at build time), to avoid having to maintain two copies?

MichaelAtOz:
I'm wondering if a screen grab of the cheatsheet as a graphic would be simpler?
Give it a hotkey to toggle a detached overlay window, movable, remembering its location, if you have a big screen you could park it off to one side & open/close it quickly for reference.
??
Surely the build system could capture a new image when the source changes??

donbright:
@michaelatoz  yes that is true... but that would involve a web rendering engine grabbing the html and rendering it to an image file at build time. i feel like thats a lot to depend on for a build system. 

the synchronization issue is also a problem. it will take some thinking. 

i also do not like this 'detect network connect' thing i wrote, that stuff never works right. 

i also do not like depending on a web browser.

Issue #1645, opened 2016-05-14
Create persistent cache feature


MarPiRK:
It would be great to add persistent cache, so when openscad is run in shell all cache is shared between separate runs. It'd be very useful regarding #1632.
Also I'd like to announce, that I'm planning to create CMake scripts supporting compiling OpenSCAD files and there it would be even more useful, especially when project has shared modules.
It'd be nice if there was shell argument with directory, where cache would be saved – it'd ease my work with CMake script.

t-paul:
See also related discussion in #1437.

kintel:
See also https://github.com/openscad/openscad/wiki/Ideas-for-Development-Tasks#persistant-caching

MarPiRK:
I have idea how to do it. There may be used some sort of protocol buffer (like [Cap'n Proto](https://capnproto.org/), which I'm familiar with), which would manage entire cache.

I'd use protobuf for three reasons:

1. It's fairly easy to create whatever, platform independent cache structure we'd like to have.
2. There would be possibility, within single library, to implement single process memory cache (like it's done currently), local file cache (what was proposed in this issue) and remote cache (via RPC calls), what would enable distributed compiling over network.
3. Multiple processes could use the same cache simultaneously without any kind of collisions and conflicts.

t-paul:
It would be nice to use something where we don't need to write the server part ourselves (e.g. like [memcached](https://memcached.org/)). The page you linked mentions sandstorm.io, so is there also some server side implementation that could be used (and installed independently) by just implementing the client side in OpenSCAD? I've seen the talk about [sandstorm.io](https://sandstorm.io/) from Asheesh Laroia at DebConf15 which was interesting.

MarPiRK:
In Cap'n Proto server side is application specific, so there is no universal server. I also thought about some kind of NoSQL database (like serverless [UnQLite](https://unqlite.org/)), but serialization-deserialization part would need to be handled in OpenSCAD code. Cap'n Proto handles this process automagically (and platform independent) basing on schemes, which are conveniently translated into C++ headers (look [there](https://capnproto.org/cxx.html)).

MarPiRK:
Or we can use Cap'n Proto just for serialization of data structures (like Geometry, CGAL), which would be saved in memcached.

MarPiRK:
[...]

PR #2214, opened 2017-12-12
CGAL cache persistance

rmie:
First take on implementing a persistance layer for the CGAL cache. The implementation is based on a memory mapped file that can be shared and used by between several OpenSCAD processes simultaneously.

Objects inserted into the memory cache are added to the file cache as well and are moved to the top of the LRU list whenever requested from either of both. Objects are removed in LRU order once memory allocation for new objects fails, as long as needed. To avoid a race condition between "contains"/"get" and cache removals from other processes, "contains" adds them back into the memory cache.

Because of memory mapping, cache sizes are limited to about 2-3GB on 32bit systems (is this still relevant?), but virtually unlimited on 64bit systems.

As of now, the size of the cache file can't be changed. Growing it should be easy, shrinking might be difficult.

### DIFF
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -55,6 +55,7 @@
 #ifdef ENABLE_CGAL
 #include "CGAL_Nef_polyhedron.h"
 #include "cgalutils.h"
+#include "CGALCache.h"
 #endif
 
 #include "csgnode.h"
@@ -836,6 +837,8 @@ int main(int argc, char **argv)
 		("colorscheme", po::value<string>(), "colorscheme")
 		("debug", po::value<string>(), "special debug info")
 		("quiet,q", "quiet mode (don't print anything *except* errors)")
+		("cachefile", po::value<string>(), "")
+		("cachefilesize", po::value<unsigned int>()->implicit_value(1024), "size in MiB")
 		("o,o", po::value<string>(), "out-file")
 		("p,p", po::value<string>(), "parameter file")
 		("P,P", po::value<string>(), "parameter set")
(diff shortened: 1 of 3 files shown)


rmie:
Just seen that it breaks the build for osx, try to fix it.

src/openscad.cc
```diff
@@ -836,6 +837,8 @@ int main(int argc, char **argv)
 		("colorscheme", po::value<string>(), "colorscheme")
 		("debug", po::value<string>(), "special debug info")
 		("quiet,q", "quiet mode (don't print anything *except* errors)")
+		("cachefile", po::value<string>(), "")
+		("cachefilesize", po::value<unsigned int>()->implicit_value(1024), "size in MiB")
```
kintel: use default_value to make vm["cachefilesize"].as<unsigned int>() work

rmie:
thx for the feedback.

I have a few general questions. 

Is it worthwhile at all? I can't find an open issue, but have seen some discussion about such a feature on the mailing list in the past, so I guess yes.

The general approach? using file backed shared memory seems like overkill, and actually wasn't my first approach. The first implementation was just simply read/write the memory cache on startup/shutdown, turned out to be way too expensive. The second try, selective read on startup, complete rewrite on shutdown is fast(er), but entries from intermediary operations are quickly lost. That's how I ended up with a (exclusive) memory mapped file, which still had issues with reusing objects from libraries (e.g. the same 20 tooth bevel gear used in several scad files) and concurrency (having the same scad file open in several windows). That's how I ended up with shared memory and a global cache that can easily be in the order > 100GB (disk is cheap) 

TODO
- [ ] make this experimental
- [ ] Preferences/GUI for cachefile and size: this largely depends on the way forward, global vs. per scad file cache. IMHO, a huge global cache file is a very efficient way.
- [ ] cache invalidation

BTW: If you are wondering about my motivation to drive this forward, I'm working on a plugin to render OpenSCAD files from Cura and keep export definitions and slicer settings as comments in OpenSCAD files.
https://community.ultimaker.com/topic/20031-cura-openscad-integration/?tab=comments#comment-19996

t-paul:
Looks like another "+1" for actual [meta data](https://github.com/openscad/openscad/wiki/Meta-Data-Use-Cases).

kintel:
@rmie The feature request issue for this is #1645

kintel:
I think this is perfectly worthwhile - it's just a large enough chunk of code that I need to review it with a clear mind, so please give me some time.

One thing to keep in mind: GUI preferences usually don't affect cmd-line usage, so just make sure that the two don't disagree on things like cache sizes.

kintel:
I'll add questions here as they show up:
* Since we have multiple caches (e.g. CGALCache, GeometryCache), did you think about allowing a single cache file to back them all? There is heavy lifting that's not necessarily done by CGAL, plus we may at some point decide to use other libraries to process certain types of data.

Issue #2195, opened 2017-11-14
tan(90) is defined


thehans:
Run this code
`echo(tan(90));`

Result:  `1.63312e+16`
Expected result: ~~`undef`~~ edit: `nan` probably would be more appropriate

donbright:
this matches C++ because tan () is taking radians as input rather than the number 90. 


```
don@serebryanya:~/src/openscad/binmaster$ cat t.cc
#include <cmath>
#include <iostream>
#include <iomanip>
main() {
std::cout << tan(M_PI*90.0/180.0)<< "\n";
std::cout << std::setprecision(25) << tan(M_PI*90.0/180.0) << "\n";
}
don@serebryanya:~/src/openscad/binmaster$ g++ t.cc
don@serebryanya:~/src/openscad/binmaster$ ./a.out 
1.63312e+16
16331239353195370

```


Basically we are dealing with the difference between ideal "tan" function and "tan" as it exists in a computer, which is using approximations of an infinite process. 

I think the answer may lie somewhere within Niven's Theorem. Niven had some interesting things to say about when Sine has rational outputs aka value. There are only three angles, 0 30 and 90, producing 0, 1, 1/2. That's it. There are no other inputs to sine that will produce rational output.

Now those angles, 0 30 90, are rational numbers but if you convert to radians, they are 0, pi/6, pi/2, but pi is an irrational number with an infinite number of digits. So you can't get the rational results you want, if your system is based on radian input. There is no rational radian that is equivalent to "30 degrees". In other words, radians and degrees are not equivalent if one is working with a finite number of digits, which one always is inside of a computer. 

Soooo lets say you built a computer trigonometry system that was built on the idea of Sine having angles for input. Then you could give it exactly 30 degrees, or 90 degrees, and get a perfect results of  1/2 or 1 as output. That would be interesting, but it would also mean you re-implemented an entire mathematical system for the benefit of two special input numbers (and their multiples), out of several billion.

Now OpenSCAD actually does use angles for input.... and it converts them to radians. But what if OpenSCAD were to treat those "special angles" as "special numbers" for each trig function, and skip over C++ math functions and return the rational value? For example sin(30)=.5 instead of 0.4999999999999999444888488?

[...]

PR #2215, opened 2017-12-14
Added new test and fixes for trigonometric functions

thehans:


thehans:
This addresses #2195 and a bunch of similar/related issues that were discovered as the test was being written.

nophead:
1/0 gives inf and -1/0 gives -inf. I think tan(90) and tan(-90) should do the same rather than nan.

doug-moen:
+1 to what @nophead said. tan(90) should be inf so that cotangent calculations do not break. See my comment in  #2195 for more details.

kintel:
Looks good to me from a code perspective, but I'm not a heavy user of these functions.
I'll merge unless someone raises concerns.

PR #2218, opened 2017-12-15
avoid caching of trivially different/identical meshes

rmie:
In general, the CGAL memory cache contains multiple copies of the same mesh. The reason for this are trivially differences in the cache key (the geometry), that are caused by eg. implicit unions or modifications that aren't changing the mesh at all (eg. color or render).

For example (maybe exaggerated, but to make the point): 

module dstar(r = 30, R = 50) {
    difference() {
        sphere(r = R);
        rotate([0, -30, -45]) translate([R + 0.7*r, 0, 0]) sphere(r = r);
    }
}

for( tx = [-60, 60], ty = [-60, 60]) 
    color([0.5 + tx/120, 0.5 + ty/120, 0]) 
        render() translate([tx, ty, 0]) dstar();

This results in 14 copies of the same, or only trivially different mesh in the cache [1]

This patch prevents duplicate copies by, comparing size and key (substring) with objects already in the cache. I haven't observed any noticeable negative effect on computation time so far.

[1] pre patch output (14 copies cached)
CGAL Cache insert: difference(){sphere($fn=0,$fa=12,$fs=2,r (1183344 bytes)
CGAL Cache insert: group(){difference(){sphere($fn=0,$fa=12 (1183344 bytes)
CGAL Cache insert: multmatrix([[1,0,0,-60],[0,1,0,-60],[0,0 (1183344 bytes)
CGAL Cache insert: render(convexity=1){multmatrix([[1,0,0,- (1183344 bytes)
CGAL Cache hit: group(){difference(){sphere($fn=0,$fa=12 (1183344 bytes)
CGAL Cache insert: multmatrix([[1,0,0,-60],[0,1,0,60],[0,0, (1183344 bytes)
CGAL Cache insert: render(convexity=1){multmatrix([[1,0,0,- (1183344 bytes)
CGAL Cache hit: group(){difference(){sphere($fn=0,$fa=12 (1183344 bytes)
CGAL Cache insert: multmatrix([[1,0,0,60],[0,1,0,-60],[0,0, (1183344 bytes)
CGAL Cache insert: render(convexity=1){multmatrix([[1,0,0,6 (1183344 bytes)
CGAL Cache hit: group(){difference(){sphere($fn=0,$fa=12 (1183344 bytes)
CGAL Cache insert: multmatrix([[1,0,0,60],[0,1,0,60],[0,0,1 (1183344 bytes)
CGAL Cache insert: render(convexity=1){multmatrix([[1,0,0,6 (1183344 bytes)
CGAL Cache insert: color([0,0,0,1]){render(convexity=1){mul (1183344 bytes)
CGAL Cache insert: color([0,1,0,1]){render(convexity=1){mul (1183344 bytes)
CGAL Cache insert: color([1,0,0,1]){render(convexity=1){mul (1183344 bytes)
CGAL Cache insert: color([1,1,0,1]){render(convexity=1){mul (1183344 bytes)
CGAL Cache insert: group(){color([0,0,0,1]){render(convexit (4733112 bytes)

[2] post patch output (1 copy cached)
CGAL Cache insert: difference(){sphere($fn=0,$fa=12,$fs=2,r (1160912 bytes)
CGAL Cache trivial: group(){difference(){sphere($fn=0,$fa=12 (1160912 bytes)
CGAL Cache trivial: multmatrix([[1,0,0,-60],[0,1,0,-60],[0,0 (1160912 bytes)
CGAL Cache trivial: render(convexity=1){multmatrix([[1,0,0,- (1160912 bytes)
CGAL Cache hit: difference(){sphere($fn=0,$fa=12,$fs=2,r (1160912 bytes)
CGAL Cache trivial: group(){difference(){sphere($fn=0,$fa=12 (1160912 bytes)
CGAL Cache trivial: multmatrix([[1,0,0,-60],[0,1,0,60],[0,0, (1160912 bytes)
CGAL Cache trivial: render(convexity=1){multmatrix([[1,0,0,- (1160912 bytes)
CGAL Cache hit: difference(){sphere($fn=0,$fa=12,$fs=2,r (1160912 bytes)
CGAL Cache trivial: group(){difference(){sphere($fn=0,$fa=12 (1160912 bytes)
CGAL Cache trivial: multmatrix([[1,0,0,60],[0,1,0,-60],[0,0, (1160912 bytes)
CGAL Cache trivial: render(convexity=1){multmatrix([[1,0,0,6 (1160912 bytes)
CGAL Cache hit: difference(){sphere($fn=0,$fa=12,$fs=2,r (1160912 bytes)
CGAL Cache trivial: group(){difference(){sphere($fn=0,$fa=12 (1160912 bytes)
CGAL Cache trivial: multmatrix([[1,0,0,60],[0,1,0,60],[0,0,1 (1160912 bytes)
CGAL Cache trivial: render(convexity=1){multmatrix([[1,0,0,6 (1160912 bytes)
CGAL Cache trivial: color([0,0,0,1]){render(convexity=1){mul (1160912 bytes)
CGAL Cache trivial: color([0,1,0,1]){render(convexity=1){mul (1160912 bytes)
CGAL Cache trivial: color([1,0,0,1]){render(convexity=1){mul (1160912 bytes)
CGAL Cache trivial: color([1,1,0,1]){render(convexity=1){mul (1160912 bytes)
CGAL Cache insert: group(){color([0,0,0,1]){render(convexit (4643384 bytes)

### DIFF
--- a/src/CGALCache.cc
+++ b/src/CGALCache.cc
@@ -19,10 +19,21 @@ shared_ptr<const CGAL_Nef_polyhedron> CGALCache::get(const std::string &id) cons
 
 bool CGALCache::insert(const std::string &id, const shared_ptr<const CGAL_Nef_polyhedron> &N)
 {
-	auto inserted = this->cache.insert(id, new cache_entry(N), N ? N->memsize() : 0);
+	size_t size = (N ? N->memsize() : 0);
+	for(const auto& entry : cache) {
+		// leaking the details of cache::Node is a bit ugly
+		// TODO provide an iterator with access to cache<Key, T>
+		if (entry.second.t->N->memsize() == size &&
+			id.find(entry.first) != std::string::npos) {
+				PRINTB("CGAL Cache trivial: %s (%d bytes)", id.substr(0, 40) % size);
+				return false;
+		}
+	}
+
+	auto inserted = this->cache.insert(id, new cache_entry(N), size);
 #ifdef DEBUG
-	if (inserted) PRINTB("CGAL Cache insert: %s (%d bytes)", id.substr(0, 40) % (N ? N->memsize() : 0));
-	else PRINTB("CGAL Cache insert failed: %s (%d bytes)", id.substr(0, 40) % (N ? N->memsize() : 0));
+	if (inserted) PRINTB("CGAL Cache insert: %s (%d bytes)", id.substr(0, 40) % size);
+	else PRINTB("CGAL Cache insert failed: %s (%d bytes)", id.substr(0, 40) % size);
 #endif
 	return inserted;
 }
--- a/src/cache.h
+++ b/src/cache.h
@@ -59,6 +59,7 @@ class Cache
 	};
 	typedef typename std::unordered_map<Key, Node> map_type;
 	typedef typename map_type::iterator iterator_type;
+	typedef typename map_type::const_iterator const_iterator_type;
 	typedef typename map_type::value_type value_type;
 
 	std::unordered_map<Key, Node> hash;
@@ -117,6 +118,9 @@ public:
 
 	bool remove(const Key &key);
 	T *take(const Key &key);
+
+	const_iterator_type begin() const {return hash.cbegin(); }
+	const_iterator_type end() const {return hash.cend(); }
 
 private:
 	void trim(int m);


kintel:
I see what you're trying to do, and this makes a lot of sense when seen in the context of persistent caching. The implementation feels a like a bit of a hack. I wonder if some of this logic could be moved to the client of the cache rather than being managed by the cache itself. Also, iterating over the entire cache feels a bit wrong.

I'll thing about this a bit more. If you have any ideas for how to make this cleaner that would be awesome.

thehans:
I made a slight edit to the debug print statements just to be sure what the full cache key looks like, and this is the result of the example scad from @rmie on a build from master :
```
Saved backup file: /home/hans/.local/share/OpenSCAD/backups/unsaved-backup-hXM29087.scad
CGAL Cache insert: difference(){sphere($fn=0,$fa=12,$fs=2,r=50);multmatrix([[0.612372,0.707107,-0.353553,0],[-0.612372,0.707107,0.353553,0],[0.5,-2.77556e-17,0.866025,0],[0,0,0,1]]){multmatrix([[1,0,0,71],[0,1,0,0],[0,0,1,0],[0,0,0,1]]){sphere($fn=0,$fa=12,$fs=2,r=30);}}} (1205776 bytes)
CGAL Cache insert: group(){difference(){sphere($fn=0,$fa=12,$fs=2,r=50);multmatrix([[0.612372,0.707107,-0.353553,0],[-0.612372,0.707107,0.353553,0],[0.5,-2.77556e-17,0.866025,0],[0,0,0,1]]){multmatrix([[1,0,0,71],[0,1,0,0],[0,0,1,0],[0,0,0,1]]){sphere($fn=0,$fa=12,$fs=2,r=30);}}}} (1205776 bytes)
CGAL Cache insert: multmatrix([[1,0,0,-60],[0,1,0,-60],[0,0,1,0],[0,0,0,1]]){group(){difference(){sphere($fn=0,$fa=12,$fs=2,r=50);multmatrix([[0.612372,0.707107,-0.353553,0],[-0.612372,0.707107,0.353553,0],[0.5,-2.77556e-17,0.866025,0],[0,0,0,1]]){multmatrix([[1,0,0,71],[0,1,0,0],[0,0,1,0],[0,0,0,1]]){sphere($fn=0,$fa=12,$fs=2,r=30);}}}}} (1205776 bytes)
CGAL Cache hit: group(){difference(){sphere($fn=0,$fa=12 (1205776 bytes)
CGAL Cache insert: multmatrix([[1,0,0,-60],[0,1,0,60],[0,0,1,0],[0,0,0,1]]){group(){difference(){sphere($fn=0,$fa=12,$fs=2,r=50);multmatrix([[0.612372,0.707107,-0.353553,0],[-0.612372,0.707107,0.353553,0],[0.5,-2.77556e-17,0.866025,0],[0,0,0,1]]){multmatrix([[1,0,0,71],[0,1,0,0],[0,0,1,0],[0,0,0,1]]){sphere($fn=0,$fa=12,$fs=2,r=30);}}}}} (1205776 bytes)
CGAL Cache hit: group(){difference(){sphere($fn=0,$fa=12 (1205776 bytes)
CGAL Cache insert: multmatrix([[1,0,0,60],[0,1,0,-60],[0,0,1,0],[0,0,0,1]]){group(){difference(){sphere($fn=0,$fa=12,$fs=2,r=50);multmatrix([[0.612372,0.707107,-0.353553,0],[-0.612372,0.707107,0.353553,0],[0.5,-2.77556e-17,0.866025,0],[0,0,0,1]]){multmatrix([[1,0,0,71],[0,1,0,0],[0,0,1,0],[0,0,0,1]]){sphere($fn=0,$fa=12,$fs=2,r=30);}}}}} (1205776 bytes)
CGAL Cache hit: group(){difference(){sphere($fn=0,$fa=12 (1205776 bytes)
CGAL Cache insert: multmatrix([[1,0,0,60],[0,1,0,60],[0,0,1,0],[0,0,0,1]]){group(){difference(){sphere($fn=0,$fa=12,$fs=2,r=50);multmatrix([[0.612372,0.707107,-0.353553,0],[-0.612372,0.707107,0.353553,0],[0.5,-2.77556e-17,0.866025,0],[0,0,0,1]]){multmatrix([[1,0,0,71],[0,1,0,0],[0,0,1,0],[0,0,0,1]]){sphere($fn=0,$fa=12,$fs=2,r=30);}}}}} (1205776 bytes)
Geometries in cache: 8
Geometry cache size in bytes: 553024
CGAL Polyhedrons in cache: 6
CGAL cache size in bytes: 7234656
Normalized CSG tree has 4 elements
Compile and preview finished.
Total rendering time: 0 hours, 0 minutes, 8 seconds
```

I think that if many of the sections of those string keys are not relevant to the cache (they are applied after pulling the geometry from cache?) then they should be stripped from the "key" before insertion/lookup, so that each unique geometry has only one unique key.  Then you shouldn't need to loop through for every insertion, right?

thehans:
Also the `group(){...}`  with just a single node (difference) inside, I would think is extraneous?  Is the group related to being returned from a module?  Can we simplify away any group of length == 1 at some earlier point in node/tree generation?

kintel:
@thehans What this PR does is to skip caching of any node if a substring of the node key is already cached and happen to be of the same size. This effectively drops colors, transforms and groups with single children. It does cause us to recalculate said nodes, but these operations tend to be really cheap compared to CGAL unions.

With a smarter client side, some of @rmie's optimizations wouldn't be necessary - question is how to best design and implement this properly.

rmie:
The proper place for this optimization, IMHO, would be the parser. There, it should be simple to decide if the modification of a given node in the AST is 'trivial' (e.g. color, render, translate, group, etc.). I have no extensive knowledge about lex/yacc (only used once in my live), but I suspect this to be none trivial.

I agree that it is hack-ish, but caches by itself are already hacks :)
Finally caching is about correctness (never return wrong/outdated results) and efficiency (have as many hits as possible). The fist is true in this case and leveraging knowledge about the entries is a common strategy to increase efficiency.    

You are concerned about iterating all other entries, from my observations, the list is usually short (<100 entries) and each step boils down to two pointer dereferences and one integer comparison. The string comparison only happens if there is a match in object size.

kintel:
@rmie Thanks for the feedback. We already have a bunch of hacks, and adding a small one on top shouldn't be a big deal. I'm just trying to encourage people to take a second pass over some of these suggested hacks in an attempt to reduce the amount of added technical debt. If you don't have any further ideas, I'd be happy to accept this patch with a few modifications:
* Document the main chunk of code explaining the hackiness and where this functionality "should ideally" live
* If possible, find way of not exposing the internals of the cache nodes.

src/CGALCache.cc
```diff
@@ -19,10 +19,21 @@ shared_ptr<const CGAL_Nef_polyhedron> CGALCache::get(const std::string &id) cons
...
 bool CGALCache::insert(const std::string &id, const shared_ptr<const CGAL_Nef_polyhedron> &N)
 {
-	auto inserted = this->cache.insert(id, new cache_entry(N), N ? N->memsize() : 0);
+	size_t size = (N ? N->memsize() : 0);
+	for(const auto& entry : cache) {
+		// leaking the details of cache::Node is a bit ugly
+		// TODO provide an iterator with access to cache<Key, T>
+		if (entry.second.t->N->memsize() == size &&
+			id.find(entry.first) != std::string::npos) {
+				PRINTB("CGAL Cache trivial: %s (%d bytes)", id.substr(0, 40) % size);
```
kintel: This should be wrapped in an `#ifdef DEBUG`.

..or we could just switch to using `PRINTDB` for all debug prints in this file so it can be switched at run-time.

PR #2221, opened 2017-12-17
Travis macos build speed tweaks

thehans:
The macos build often fails due to exceedingt travis-ci's job time limit (50min).  This is an attempt to speed up those jobs somewhat and reduce the probability of timeout failures.

-Cached homebrew, this seems to reduce the homebrew step by roughly 2-3minutes.
-Also added some timing info to homebrew script for "debugging" build times
-Tweaked parallel job count for ctest.  -j4 performed faster than -j8 in my testing.  There are only 2 cores available to us anyways ( https://docs.travis-ci.com/user/reference/overview/ ), running too many threads might cause slowdown due to memory swapping or just

### DIFF
--- a/scripts/travis-ci.sh
+++ b/scripts/travis-ci.sh
@@ -20,8 +20,15 @@ travis_finish() {
   echo -en "\ntravis_time:end:$travis_timer_id:start=$travis_start_time,finish=$travis_end_time,duration=$duration\r\033[0m"
 }
 
+PARALLEL=-j2
+if [[ "$DIST" == "trusty" ]]; then
+    PARALLEL_CTEST=-j1
+else
+    PARALLEL_CTEST=-j2
+fi
+
 travis_start qmake "Building OpenSCAD using qmake"
-qmake CONFIG+=experimental CONFIG+=nogui && make -j2
+qmake CONFIG+=experimental CONFIG+=nogui && make $PARALLEL
 travis_finish qmake
 
 travis_start cmake "Building tests using cmake"
@@ -32,19 +39,13 @@ if [[ $? != 0 ]]; then
   echo "Error configuring test suite"
   exit 1
 fi
-make -j2
+make $PARALLEL
 if [[ $? != 0 ]]; then
   echo "Error building test suite"
   exit 1
 fi
 
 travis_finish cmake
-
-if [[ "$DIST" == "trusty" ]]; then
-    PARALLEL=-j1
-else
-    PARALLEL=-j8
-fi
 
 travis_start ctest "Running tests using ctest"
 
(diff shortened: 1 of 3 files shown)


kintel:
Looks good. One thing that puzzles me though is when the ccache cache is used when when not. We've had ccache enabled for a while, but a lot of builds still seem to rebuild everything..

scripts/travis-ci.sh
```diff
@@ -20,8 +20,15 @@ travis_finish() {
   echo -en "\ntravis_time:end:$travis_timer_id:start=$travis_start_time,finish=$travis_end_time,duration=$duration\r\033[0m"
 }
 
+PARALLEL=-j2
+if [[ "$DIST" == "trusty" ]]; then
+    PARALLEL_CTEST=-j1
+else
+    PARALLEL_CTEST=-j4
```
justbuchanan: You may be able to use `nproc` to lookup the number of cores instead of having to hard-code it. Something like `PARALLEL_CTEST="-j$(nproc)"`.

thehans:
@kintel yeah I'm also puzzled by that.  I have been trying to look for any pattern for why it runs when it does but I don't understand ccache very well yet.  

By the way I had been testing on travis on my own fork, so you can see how the build times break down in those recent builds: https://travis-ci.org/thehans/openscad/pull_requests 
It feels like it generally runs faster than the openscad's "fork" for some reason.  That might just be something like luck of starting builds at times when travis-ci.org has had a lower load.

scripts/travis-ci.sh
kintel: This is mostly tuned for memory consumption on virtual hosts, so it's hard to find a good number

thehans: @kintel, by the way do you know what the issue was that led to keeping -j1 for trusty?  
https://github.com/openscad/openscad/commit/a0d92731292da184c7490ab606c24e518841e8e4

t-paul said something to the effect that it had caused some intermittent test failures?

kintel: I think there was something related to Xvnc - multiple X11 environment causing some weird errors. Could also have been GL drivers.

thehans: @justbuchanan It can help in some cases to set more jobs than cores, to help make up wasted cpu time when jobs are blocking on i/o.  so using the exact # of cores is not necessarily the goal.  but i feel like the j8 might have been excessive.  j4 seemed more reasonable and (in my sample size of 1 build) at -j4, the ctest portion of the macos job did run fastest compared to other builds at -j2 and -j8

I tried to benchmark these to some degree, but it is admittedly difficult to measure with the kind of variation in build times that we see from travis-ci.org 

Also something that may be of interest is the: "Backlog macOS Builds for Open Source projects" chart on this page:  https://www.traviscistatus.com/
You can get an idea how overloaded their mac servers are throughout the day/week. This mostly affects how long you have to wait before the mac build even boots/begins, sometimes 30min up to an hour or so if the backlog is high.  But I'm not sure if there is any high correlation between this backlog and the actual job run times, once they actually have begun.

PR #2228, opened 2017-12-20
Don't build unused libraries

thehans:


thehans:
Before: https://travis-ci.org/openscad/openscad/jobs/319269959#L1629
Building tests using cmake 695.92s

After: https://travis-ci.org/openscad/openscad/jobs/319297154#L1941
Building tests using cmake 375.99s

About 320s (~5.5 minutes) faster

PR #2232, opened 2017-12-28
Minor refactoring and removal of unused code

kintel:
Related to #2140 

@thehans Since you touched this code last, would you mind giving it a lookover?


thehans:
@kintel  Looks good to me, no issues that I can see.

PR #2234, opened 2017-12-29
Add QGamepad based input driver.

t-paul:


openscad.pro
```diff
@@ -554,6 +554,15 @@ unix:!macx {
   SOURCES += src/input/JoystickInputDriver.cc
 }
 
+contains(QT_VERSION, ^5\\..*) {
```
kintel: You can use `!lessThan(QT_MAJOR_VERSION, 5)` to be more future proof

src/input/QGamepadInputDriver.h
```diff
@@ -0,0 +1,46 @@
...
+public:
+    QGamepadInputDriver();
+    virtual ~QGamepadInputDriver();
+    virtual void run();
+    virtual bool open();
+    virtual void close();
+
+    virtual const std::string & get_name() const;
+private:
+    QGamepad *gamepad;
```
kintel: We could protect this object with a `unique_ptr` to ensure that this object gets deleted in all cases.

src/input/QGamepadInputDriver.h
```diff
@@ -0,0 +1,46 @@
...
+    char name[1024];
+    volatile bool stopRequest;
```
kintel: What does this member variable do?

src/input/QGamepadInputDriver.h
```diff
@@ -0,0 +1,46 @@
...
+    QGamepadInputDriver();
+    virtual ~QGamepadInputDriver();
+    virtual void run();
+    virtual bool open();
+    virtual void close();
+
+    virtual const std::string & get_name() const;
+private:
+    QGamepad *gamepad;
+    char name[1024];
```
kintel: This looks unused

src/input/QGamepadInputDriver.h
MichaelPFrey: stopRequest looks like an artefact from JoystickInputDriver.

The JoystickInputDriver uses the bool stopRequest to stop its main while loop.
As the while loop is in a different thread [the while loop is in the run method], I used volatile to make sure that the compiler would not start doing funny business while optimizing.

Given that the run methode of this driver is empty, there should be no need for this volatile bool.

t-paul: Yep, both variables are just left-overs, I'm just going to remove those...

PR #2238, opened 2017-12-30
listDrivers() regardless whether SCINTILLA_EDITOR is used or not

MichaelPFrey:
i do see no reason for the "#ifdef USE_SCINTILLA_EDITOR" in context of the InputDriverManager.

### DIFF
--- a/src/LibraryInfo.cc
+++ b/src/LibraryInfo.cc
@@ -3,8 +3,8 @@
 #include <vector>
 #ifdef USE_SCINTILLA_EDITOR
 #include <Qsci/qsciglobal.h>
-#include "input/InputDriverManager.h"
 #endif
+#include "input/InputDriverManager.h"
 
 #include "version_check.h"
 #include "PlatformUtils.h"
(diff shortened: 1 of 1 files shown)


t-paul:
Hmm, we have to revert that for now. This breaks building the test suite as this is not compiling with Qt. I guess the misuse of the SCINTILLA define was just too sneaky.
As there is not much use to have the input drivers without Qt, we probably can the this separately to Qt availability.

MichaelPFrey:
Interesting and no problem.

But: This rises the question, if this is the right place to list the drivers.
Listing the drivers also became somewhat redundant with #2242 and will become even more obsolete, when we start adding further status outputs for each driver.

Then again: For the moment, that is a valid workaround.

t-paul:
I think it's good to list things in one place. We just need to make it more obvious that this file is valid for both GUI (with Qt) and command line (without Qt) builds.
That separation is not all that clear, especially with the different build systems. Maybe that will also improve a bit once cmake for the main build is ready. But that probably will take some time still...

PR #2242, opened 2017-12-31
Add option to enable/disable specific drivers

MichaelPFrey:
I want to explain some ideas:
All drivers are off by default to avoid any potential problems.
This is kind of silly, as the user now has to:
- add CONFIG+=experimental to build it
- enable the inputdrivermanager as an experimental feature
- enable a specific driver
and restart openscad for the last two step each.
How ever: in the long run (when the feature is considered stable), the first two steps will become obsolete.

I specifically want to show all drivers, even when not supported/build.
bildschirmfoto von 2017-12-31 18-53-45
This is for the benefit of for e.g. Developers. I might use different builds - I do not want one build change the config to disable a driver, just because the current build does not have that driver build in.
(It goes similarly for experimental builds on Windows)
It also shows "clearly" in a screenshot, that a specific version has a specific driver in general foreseen, but not in this build. This will be interesting once we get question from power- and end-users, not just developers.
I added a tool tip to explain, why some off the drivers are disabled. ("This driver was not enabled during build time and is thus not available.")

What I am currently thinking about, is how I could stop a driver when the user wants to disable it during run time. (Having to restart OpenSCAD to stop a driver is counterintuitive)

Feel free to suggest alternate ideas.
Feel free to give feedback.
Feel free ultimately squash all my commits during merging.


t-paul:
Nice, I think that's a useful (maybe even needed?) feature for normal use as well.

Maybe the manager could be used to retrieve the possible options as list so changing the preferences would not be needed if we add other driver. But having just the checkboxes is fine too.

MichaelPFrey:
It is clearly a needed feature, as we never know for certain if a driver could cause a crash on a specific system (that also being one of the reason for off by default). Also, currently the DriverManager might be able to handle multiple drivers, but the UI clearly is not designed for that (axis 1 is axis 1, regardless of which input driver it came from).
Some drivers are also potentially conflicting - e.g. QGamepad and Joystick (and we need both - QGamepad works for things like the XBox360 controller, but does not like cheap no-name controllers). 

Asking the InputManager for the drivers is a good idea, that would also open up the possibility to start and stop drivers during run time. But this would mean a lot changes to the infrastructure.
(Also: We currently only generate little of the GUI by code)
I also like the idea of showing all driver even when they are not build (otherwise we might get more question "why do I not have/see driver...?").

An other idea to get access to the drivers from the UI is to make the drivers into Singeltons. Few users will have more then one or two (the second beeing DBUS) running, let alone have the need do run two instances of the same driver.
With the Singelton Design Pattern we could request the instance from the class.

Having the driver description in the UI code is something I currently mind.
I would prefer to have the description in the source code of the driver.
(which off-course only works when the driver can be build)

MichaelPFrey:
This is about right for now. It works good and I am happy with the code as in changing as little as possible.
Now with the QGamepad driver and the Joystick Driver we also have a need for this feature.
We can off-course revisit it later.

Feel free to squash or merge the commits.

PR #2250, opened 2018-01-05
Add easy build script

SConaway:


scripts/build.sh
```diff
@@ -0,0 +1,41 @@
+#! /usr/bin/env bash
+
```
kintel: Could you add some usage notes here, as well as some info about who should use this script?
..perhaps with a warning that this will install stuff globally on your system using sudo.

scripts/build.sh
```diff
@@ -0,0 +1,41 @@
...
+  fi
+}
+
+
+# pull MCAD
+if ! [ -x "$(command -v git)" ]; then
+  echo 'Error: git is not installed.' >&2
+  exit 1
+fi
+cd ..
```
kintel: It looks like this script must be run using the `scripts/` folder as CWD, but then cd's out of that folder.
This isn't really a very common behavior. Please consider another strategy for running this, e.g. `SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"`

scripts/build.sh
```diff
@@ -0,0 +1,41 @@
...
+  echo Running on macOS
+
+  if  [ -x "$(command -v brew)" ]; then
+    echo 'Looks like Homebrew is installed. Using Homebrew.' >&2
+    ./scripts/macosx-build-homebrew.sh
+
+    qmake openscad.pro
+    make
+    sudo make install
+  fi
```
kintel: An else statement would be useful here.

SConaway:
So I just somehow found that I never finished this PR. I'll be finishing it soon. Sorry

SConaway:
There are the requested changes. Unfortunately, I'm unable to test on my machine atm, so I'm not sure how well it works.

PR #2252, opened 2018-01-05
Inputdriver5 ignore mousewheel on the combo boxes

MichaelPFrey:
A lot of GUI elements in OpenSCAD are in Scroll Areas.
This causes a conflict, as some elements within the Scroll Areas are also reacting to the mousewheel.
Especially in the settings, where the user might spend a considerable amount of time to get it just right, it is annoying when simply scrowling down unintentionally changes various settings.

for reference:
   * https://stackoverflow.com/questions/3241830/qt-how-to-disable-mouse-scrolling-of-qcombobox
   * https://stackoverflow.com/questions/5821802/qspinbox-inside-a-qscrollarea-how-to-prevent-spin-box-from-stealing-focus-when
   * http://doc.qt.io/archives/qt-4.8/qobject.html#installEventFilter

### DIFF
--- a/openscad.pro
+++ b/openscad.pro
@@ -363,7 +363,8 @@ HEADERS += src/version_check.h \
            src/input/InputEventMapper.h \
            src/input/InputDriverManager.h \
            src/input/AxisConfigWidget.h \
-           src/input/ButtonConfigWidget.h
+           src/input/ButtonConfigWidget.h \
+           src/input/WheelIgnorer.h
 
 SOURCES += \
            src/libsvg/libsvg.cc \
@@ -512,7 +513,8 @@ SOURCES += \
            src/input/InputEventMapper.cc \
            src/input/InputDriverManager.cc \
            src/input/AxisConfigWidget.cc \
-           src/input/ButtonConfigWidget.cc
+           src/input/ButtonConfigWidget.cc \
+           src/input/WheelIgnorer.cc
 
 # CGAL
 HEADERS += src/ext/CGAL/convex_hull_3_bugfix.h \
--- a/src/input/ButtonConfigWidget.cc
+++ b/src/input/ButtonConfigWidget.cc
@@ -30,6 +30,7 @@
 #include "QSettingsCached.h"
 #include "input/InputDriverManager.h"
 #include "SettingsWriter.h"
+#include "WheelIgnorer.h"
 
 ButtonConfigWidget::ButtonConfigWidget(QWidget *parent) : QWidget(parent)
 {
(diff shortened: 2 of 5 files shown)


src/input/ButtonConfigWidget.cc
```diff
@@ -61,6 +62,12 @@ void ButtonConfigWidget::init() {
 			initComboBox(box,*ent);
 		}
 	}
+
+	WheelIgnorer *wheelIgnorer = new WheelIgnorer();
```
kintel: This object should probably receive a parent so it gets deleted at some point.
You can also `auto` this for less repetition: `auto wheelIgnorer = new WheelIgnorer();`

src/input/ButtonConfigWidget.cc
```diff
@@ -61,6 +62,12 @@ void ButtonConfigWidget::init() {
...
+	QList<QComboBox *> widgets = this->findChildren<QComboBox *>();
```
kintel: This is another place where `auto` is nice: `auto widgets = this->findChildren<QComboBox *>();`

src/input/ButtonConfigWidget.cc
```diff
@@ -61,6 +62,12 @@ void ButtonConfigWidget::init() {
...
+	foreach (QComboBox* b, widgets) {
```
kintel: Not sure if `foreach` can take an auto argument, but I assume it can: `foreach (auto b, widgets) {`.
..or just use C++: `for (auto b : widgets) {`

openscad.pro
```diff
@@ -512,7 +513,8 @@ SOURCES += \
            src/input/InputEventMapper.cc \
            src/input/InputDriverManager.cc \
            src/input/AxisConfigWidget.cc \
-           src/input/ButtonConfigWidget.cc
+           src/input/ButtonConfigWidget.cc \
+           src/input/WheelIgnorer.cc
```
kintel: At some point, adding the inputdriver stuff to the top-level `CMakeLists.txt` would be nice. May be out of scope for this particular PR though.

src/input/ButtonConfigWidget.cc
MichaelPFrey: using auto also allowed me to name the variable "comboBoxes" instead of widgets, without repeating the same word three times. I start to see the benefits of auto.

src/input/ButtonConfigWidget.cc
MichaelPFrey: Thank you, that makes sense. I now set the ButtonConfigWidget and AxisConfigWidget as parents respectively.

kintel: On this note, you should be able to pass this to the `WheelIgnorer` contructor; that's the common way of assigning parents in Qt. Adding a constructor also makes it clear that this is how the object is supposed to be used (e.g. you could require an argument to be passed).

src/input/ButtonConfigWidget.cc
kintel: Another side effect of this is that you don't even need to store the widgets as a temporary variable, you could iterate directly; `for (auto comboBox : this->findChildren<QComboBox *>())  {..`

PR #2256, opened 2018-01-07
Add full git URL for MCAD to .gitmodules

lostapathy:
This existing .gitmodules file uses a relative path, rather than a full URL.  When a user forks the openscad repo and tries to `git submodule update --init`, it tries to download from `git://github.com/<user>/MCAD.git` rather than the upstream MCAD library.

This is probably not desirable behavior, since most new OpenSCAD contributors won't have (or want) their own fork of MCAD.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "libraries/MCAD"]
 	path = libraries/MCAD
-	url = ../MCAD.git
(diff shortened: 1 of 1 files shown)


kintel:
@hyperair What's your opinion on this?

hyperair:
@kintel probably better to have it this way. Users who have their own forks can use `git remote add` on the submodule anyway.

Issue #1754, opened 2016-07-25
Remove unimplemented CGAL nodes


t-paul:
The code has `glide` and `subdiv` nodes defined but both hit an assert while geometry evaluation, crashing OpenSCAD.

kintel:
Yep, these were never implemented, and I agree that we should kill them.

thehans:
Fixed by #2258

PR #2258, opened 2018-01-08
Issue #1754

thehans:


thehans:
Accidentally transposed some numbers in commits but this is for 
https://github.com/openscad/openscad/issues/1754

PR #2260, opened 2018-01-09
inputdriver5: create the "axisValues" using a for loop and snprintf

MichaelPFrey:
Using a for loop and snprintf has the following benefits:
- we can easily adjust the number of axis
- the translation is simpler (only two strings to translate instead of 18)

The variable names come from the QCombox, where we ultimately use this values.
http://doc.qt.io/qt-5/qcombobox.html#addItem-1

snprintf is part of c++11 http://www.cplusplus.com/reference/cstdio/snprintf/
ngettext (https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html) does not apply, as we are not dealing with plural, but an enumeration.

### DIFF
--- a/src/settings.cc
+++ b/src/settings.cc
@@ -71,16 +71,14 @@ static Value values(std::string s1, std::string s1disp, std::string s2, std::str
 static Value axisValues() {
 	Value::VectorType v;
 	v += ValuePtr(value("None", _("None")));
-	char userData [5];
-	char text [40];
 
 	for (int i = 0; i < InputEventMapper::getMaxAxis(); i++ ){
-		snprintf(userData, 5, "+%d", i+1);
-		snprintf(text, 40,_("Axis %d"), i);
+		auto userData = (boost::format("+%d") % (i+1)).str();
+		auto text = (boost::format(_("Axis %d")) % i).str();
 		v += ValuePtr(value(userData, text));
 
-		snprintf(userData, 5, "-%d", i+1);
-		snprintf(text, 40, _("Axis %d (inverted)"), i);
+		userData = (boost::format("-%d") % (i+1)).str();
+		text = (boost::format(_("Axis %d (inverted)")) % i).str();
 		v += ValuePtr(value(userData, text));
 	}
 	return v;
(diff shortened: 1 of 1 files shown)


src/settings.cc
```diff
@@ -70,24 +71,18 @@ static Value values(std::string s1, std::string s1disp, std::string s2, std::str
...
-	v += ValuePtr(value("-8", _("Axis 7 (inverted)")));
-	v += ValuePtr(value("+9", _("Axis 8")));
-	v += ValuePtr(value("-9", _("Axis 8 (inverted)")));
+	char userData [5];
+	char text [40];
+
+	for (int i = 0; i < InputEventMapper::getMaxAxis(); i++ ){
+		snprintf(userData, 5, "+%d", i+1);
+		snprintf(text, 40,_("Axis %d"), i);
+		v += ValuePtr(value(userData, text));
```
kintel: Instead of using a helper function `value()` to create a 2-element vector, you should be able to do this directly (untested): `v += ValuePtr({userData, text});`. If this doesn't work, we should consider adding it.

Similarly for the other helper functions; we should try to move them into Value/ValuePtr.

src/settings.cc
```diff
@@ -70,24 +71,18 @@ static Value values(std::string s1, std::string s1disp, std::string s2, std::str
...
-	v += ValuePtr(value("+8", _("Axis 7")));
-	v += ValuePtr(value("-8", _("Axis 7 (inverted)")));
-	v += ValuePtr(value("+9", _("Axis 8")));
-	v += ValuePtr(value("-9", _("Axis 8 (inverted)")));
+	char userData [5];
+	char text [40];
+
+	for (int i = 0; i < InputEventMapper::getMaxAxis(); i++ ){
+		snprintf(userData, 5, "+%d", i+1);
+		snprintf(text, 40,_("Axis %d"), i);
```
kintel: I think both these lines could benefit from using a higher level string formatter. We use boost::format() other places, e.g.:
```
auto userData = boost::format("+%d") % (i+1);
auto text = boost::format(_("Axis %d")) % i;
```

Same below.

MichaelPFrey: Ah, that makes sense. Not having to deal with buffer sizes makes things a lot safer and easier.
Thank you.

src/settings.cc
MichaelPFrey: I have to look into that later, something is confusing me how the 2-element vector is currently constructed.

Issue #1433, opened 2015-09-06
Feature request: 3D mouse support [$50 awarded]


greysAcademicCode:
It would be great to see support for [a 3D mouse](http://www.3dconnexion.com/products/spacemouse.html) added.

Maybe some effort could be borrowed from the FreeCAD project for this.
[This seems to be the patch](https://github.com/FreeCAD/FreeCAD/commit/012e50c524967d952371267683fc096ccd1eb76e) that added 3D mouse support to FreeCAD.
##

[$0 Bounty](https://www.bountysource.com/issues/26486922-feature-request-3d-mouse-support)
Another bounty added to #2135.

t-paul:
The https://github.com/openscad/openscad/tree/inputdriver2 branch has the (incomplete) prototype based on some code from the mailing list. It's basically working on all supported platforms, but still needs lots of work to tune the behavior and provide a reasonable configuration GUI.

greysAcademicCode:
Great!!
I can't get that branch to build though:
```
make: *** No rule to make target 'org.openscad.OpenSCAD.xml', needed by 'openscad_adaptor.h'.  Stop.
```
Can you offer any advice?

t-paul:
Whops, that file was hidden by .gitignore. Please pull and try compiling again...

greysAcademicCode:
It works!
By default, the inputs are not mapped in an intuitive manner (for me), but I assume you know that.
Are the settings in the input tab of the preferences connected to anything?

t-paul:
I'm not yet sure about how to do the preferences, I guess there should be at least some meta data from the device to tell the number of axis/buttons and so on. Also we probably want to remember defaults for different devices. In the end, the UI will probably be a much more complex thing compared to the actual device event handling :-).

The current defaults work for my Space Mouse Wireless with the Joystick driver (I think without any jsdev axis remapping), it's quite likely those are different with other drivers:
* Translation: Axis1 / Axis 2 (inverted) / Axis 3 (inverted)
* Rotation: Aixis 4 / Axis 5 (inverted) / Axis 6 (inverted)

MichaelAtOz:
@greysAcademicCode some more info may be helpful.
t-paul said

> It includes some initial support for mapping events via configuration, but there is no GUI yet.
>
The default config should be added when running the new version the first time. Right now it's basically the axis number with +/- defining the
direction:
>
[HKEY_CURRENT_USER\Software\OpenSCAD\OpenSCAD\InputDriver\axis]
[...]

PR #2261, opened 2018-01-09
Inputdriver5 automatically generate the actions for the button comboxes

MichaelPFrey:
I do not want to maintain the action list, so I would like to automate it:
bildschirmfoto von 2018-01-09 05-09-24
I do not want to maintain the translation of the action list, so I would just reuse the translation of the menu:
bildschirmfoto von 2018-01-09 05-08-58

Please merge  #2254 first - I will think about how to integrate it here.
Now, #2255 becomes more important as the action list is generated dynamically.
The code is currently a bodge job,  I am locking into improve it.


MichaelAtOz:
Are you going to manually handle left/right/middle clicks or something tricky?

MichaelPFrey:
@MichaelAtOz this for  #2130 which is in turn for #1433
In short: This is work on the GUI for the 3D-Mice/Joystick driver.

But your question is valid: I better change the title of this pull request and I should add "use more meaning full icons" to my to do list.

MichaelPFrey:
So, now I am happy with the look and coding. Feel free to merge/squash this pull request.

bildschirmfoto von 2018-01-30 04-40-31

t-paul:
Awesome, that looks very nice.

PR #2264, opened 2018-01-16
Inputdriver5: backend code for inputdriver configuration

MichaelPFrey:
* back-end code for setting the joystick number
* spaceNavDominantAxisOnly option

### DIFF
--- a/src/input/JoystickInputDriver.cc
+++ b/src/input/JoystickInputDriver.cc
@@ -64,8 +64,8 @@ bool JoystickInputDriver::open()
 {
     stopRequest = false;
 
-    char* path=const_cast<char*> (("/dev/input/js" + nr).c_str());
-    fd = ::open(path, O_RDONLY);
+    auto path = boost::format("/dev/input/js%d") % this->nr;
+    this->fd = ::open(path.str().c_str(), O_RDONLY);
 
     if (fd < 0) {
         return false;
@@ -98,5 +98,5 @@ const std::string & JoystickInputDriver::get_name() const
 
 void JoystickInputDriver::setJoystickNr(std::string jnr)
 {
-    nr=jnr;
+    this->nr=jnr;
 }
--- a/src/input/JoystickInputDriver.cc
+++ b/src/input/JoystickInputDriver.cc
@@ -64,7 +64,9 @@ bool JoystickInputDriver::open()
 {
     stopRequest = false;
 
-    fd = ::open("/dev/input/js0", O_RDONLY);
+    char* path=const_cast<char*> (("/dev/input/js" + nr).c_str());
+    fd = ::open(path, O_RDONLY);
+
     if (fd < 0) {
         return false;
     }
@@ -92,4 +94,9 @@ const std::string & JoystickInputDriver::get_name() const
 {
     static std::string name = "JoystickInputDriver";
     return name;
+}
+
+void JoystickInputDriver::setJoystickNr(std::string jnr)
+{
+    nr=jnr;
 }
--- a/src/input/JoystickInputDriver.h
+++ b/src/input/JoystickInputDriver.h
@@ -35,11 +35,13 @@ public:
     void run() override;
     bool open() override;
     void close() override;
-
     const std::string & get_name() const  override;
+    void setJoystickNr(std::string jnr);
+
 private:
     int fd;
     int version;
+    std::string nr="0";
     unsigned char axes;
     unsigned char buttons;
     char name[1024];
--- a/src/input/JoystickInputDriver.h
+++ b/src/input/JoystickInputDriver.h
@@ -41,7 +41,7 @@ public:
 private:
     int fd;
     int version;
-    std::string nr="0";
+    std::string nr{"0"};
     unsigned char axes;
     unsigned char buttons;
     char name[1024];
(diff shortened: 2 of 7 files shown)


src/input/JoystickInputDriver.cc
```diff
@@ -93,3 +95,8 @@ const std::string & JoystickInputDriver::get_name() const
     static std::string name = "JoystickInputDriver";
     return name;
 }
+
+void JoystickInputDriver::setJoystickNr(std::string jnr)
+{
+    nr=jnr;
```
kintel: Prefix member access with `this->` to make it more clear what's being changed.
Also (Out of scope for this PR), more descriptive variable names would make code more readable by future readers.

src/input/JoystickInputDriver.h
```diff
@@ -35,11 +35,13 @@ class JoystickInputDriver : public InputDriver
...
     bool open() override;
     void close() override;
-
     const std::string & get_name() const  override;
+    void setJoystickNr(std::string jnr);
+
 private:
     int fd;
     int version;
+    std::string nr="0";
```
kintel: Prefer uniform initialization:
`std::string nr{"0"};`

src/input/SpaceNavInputDriver.h
```diff
@@ -38,9 +38,11 @@ class SpaceNavInputDriver : public InputDriver
     void run() override;
     bool open() override;
     void close() override;
+    void setDominantAxisOnly(bool var);
 
     const std::string & get_name() const override;
 
 private:
     bool spnav_input();
+    bool DominantAxisOnly=true;
```
kintel: Member variables should start with a lower-case character. Also, uniform initialization is preferable:
`bool dominantAxisOnly{true};`

src/openscad.cc
```diff
@@ -791,14 +792,22 @@ int gui(vector<string> &inputFiles, const fs::path &original_path, int argc, cha
     app.connect(&app, SIGNAL(lastWindowClosed()), &app, SLOT(releaseQSettingsCached()));
 	app.connect(&app, SIGNAL(lastWindowClosed()), &app, SLOT(quit()));
 	if (Feature::ExperimentalInputDriver.is_enabled()) {
+		Settings::Settings *s = Settings::Settings::inst();
```
kintel: You could `auto` this.

src/openscad.cc
```diff
@@ -791,14 +792,22 @@ int gui(vector<string> &inputFiles, const fs::path &original_path, int argc, cha
...
 #ifdef ENABLE_HIDAPI
         InputDriverManager::instance()->registerDriver(new HidApiInputDriver());
 #endif
 #ifdef ENABLE_SPNAV
-        InputDriverManager::instance()->registerDriver(new SpaceNavInputDriver());
+        SpaceNavInputDriver spaceNav = new SpaceNavInputDriver();
```
kintel: This looks odd; should this be a pointer?
It could be fixed by `auto` though :)

src/input/JoystickInputDriver.cc
```diff
@@ -64,7 +64,9 @@ bool JoystickInputDriver::open()
 {
     stopRequest = false;
 
-    fd = ::open("/dev/input/js0", O_RDONLY);
+    char* path=const_cast<char*> (("/dev/input/js" + nr).c_str());
```
kintel: This confuses me. Are there implementations of `open()` that doesn't take a `const char *`?
As we use `boost::format` other places, you could do that here as well; smth. like this:
```
auto path = boost::format("/dev/input/js %d") % nr;
this->fd = ::open(path.c_str(), O_RDONLY);
```

src/openscad.cc
MichaelPFrey: thank you, that is a mistake. I currently have not setup my build system to build the spaceNav driver.

src/input/SpaceNavInputDriver.cc
```diff
@@ -115,8 +115,12 @@ void SpaceNavInputDriver::close()
 
 }
 
+void SpaceNavInputDriver::setDominantAxisOnly(bool var){
+    DominantAxisOnly = var;
```
kintel: Capitalization typo, also use `this->` for clarity

src/input/JoystickInputDriver.cc
MichaelPFrey: thank you for your suggestion.
changed to boost::format

src/input/JoystickInputDriver.cc
MichaelPFrey: changed - thank you for your suggestion.

src/input/JoystickInputDriver.h
MichaelPFrey: thank you - I did not know that syntax.

t-paul:
Hmm, not sure if I made a mistake when merging, but it seems the driver specific enable/disable setting has no effect anymore.

MichaelPFrey:
I will look into it. The thing that springs to my mind:
The driver specific enable/disable is mend to require a restart of openscad.
Drivers can not be enabled/disabled "on the fly".

MichaelPFrey:
I found the issue. Fix is in progress.

PR #2267, opened 2018-01-21
DBus interface debugged

MichaelPFrey:
This rework is based on the car example from QT.

I have a simple implementation of the other side of the interface.
I will clean it up and then upload to github as a reference implementation.

### DIFF
--- a/org.openscad.OpenSCAD.xml
+++ b/org.openscad.OpenSCAD.xml
@@ -2,37 +2,37 @@
 "http://www.freedesktop.org/standards/dbus/1.0/introspect.dtd">
 <node>
   <interface name="org.openscad.OpenSCAD">
-    <signal name="zoom">
-      <arg name="zoom" type="d" direction="out"/>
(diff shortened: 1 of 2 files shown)


MichaelPFrey:
https://github.com/MichaelPFrey/openscad-dbus-reference

Next step is making something useful using this Interface and a BBC Microbit.

MichaelAtOz:
So this is a IPC mechanism to drive the OpenSCAD UI?

MichaelPFrey:
Yes, D-Bus is a inter-process communication mechanism that can be used to drive the OpenSCAD UI.
D-Bus is linux only.

I have to say: This is an awesome way to write a custom driver.
https://github.com/MichaelPFrey/openscad-dbus-microbit
(the BBC microbit has an accelerometer, an electronic compass and two buttons)

t-paul:
Awesome, I just got a Micro:bit too...

PR #2269, opened 2018-01-23
Mac High Sierra 10.13.2 build

wilsoc5:
I had three issues building OpenScad on my mac. 

The first issue related to the identification of using libc++ vs libstdc++. Boost was build with libc++ but the qmake profile was unable to identify it properly. 

I also had an issue where STACKSIZE was undefined for some reason. Added it to the DEFINES section in the qmake project and everything worked. 

Finally, clang was getting confused with EventFilter. It kept looking for a function as opposed to the class: 
```
src/OpenSCADApp.cc:16:31: error: cannot initialize a new value of type 'EventFilter' (aka 'bool (*)(void *, long *)') with an rvalue of type
      'OpenSCADApp *'
        this->installEventFilter(new EventFilter(this));
                                     ^           ~~~~
1 error generated.
```
Renaming the EventFilter class fixed this particular issue.

### DIFF
--- a/c++11.pri
+++ b/c++11.pri
@@ -3,7 +3,8 @@ macx {
   # We attempt to auto-detect it by inspecting Boost
   dirs = $${BOOSTDIR} $${QMAKE_LIBDIR}
   for(dir, dirs) {
-    system(grep -q __112basic_string $${dir}/libboost_thread* >& /dev/null) {
+    #system(grep -q __112basic_string $${dir}/libboost_thread* >& /dev/null) {
+    system(otool -L /usr/local/lib/libboost_thread*  | grep libc++ >& /dev/null ) {
       message("Using libc++11")
       CONFIG += libc++
     }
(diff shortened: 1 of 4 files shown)


c++11.pri
```diff
@@ -3,7 +3,8 @@ macx {
   # We attempt to auto-detect it by inspecting Boost
   dirs = $${BOOSTDIR} $${QMAKE_LIBDIR}
   for(dir, dirs) {
-    system(grep -q __112basic_string $${dir}/libboost_thread* >& /dev/null) {
+    #system(grep -q __112basic_string $${dir}/libboost_thread* >& /dev/null) {
+    system(otool -L /usr/local/lib/libboost_thread*  | grep libc++ >& /dev/null ) {
```
kintel: This searches in `/usr/local/lib`, which is not necessarily where boost is located.

PR #2279, opened 2018-02-02
inputdriver5, dbus: methode to get the name of supported actions

MichaelPFrey:
usage/example see https://github.com/MichaelPFrey/openscad-dbus-reference/pull/1

I am not to sure about the data type.
I could use array or QStringList, but QString maps to String which should be easy to use in any programming language. Using newline and splitting on it is also simple and easy.

### DIFF
--- a/org.openscad.OpenSCAD.xml
+++ b/org.openscad.OpenSCAD.xml
@@ -34,5 +34,8 @@
     <method name="buttonPress">
       <arg name="idx" type="u" direction="in"/>
     </method>
+    <method name="actions">
+      <arg name="names" type="s" direction="out"/>
+    </method>
   </interface>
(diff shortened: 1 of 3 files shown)


t-paul:
I usually prefer the more specific way. Especially newlines can be annoying across the different platforms. So maybe going for an actual string list would be better?

MichaelPFrey:
Changed from "New Line separated String" to "QStringList".

DBus is Linux only, so multi platform and especially windows are are non concern.

As I understand after looking it up:
QStringList will actually be send as a String Array over the DBus,
So, other progammers not using QT do not have to deal with a QT datastructur.

MichaelPFrey:
d-feet

PR #2281, opened 2018-02-04
inputdriver 5: expose the camera state via dbus driver

MichaelPFrey:


t-paul:
@kintel what would be the best way regarding camera. Right now there's still different interpretations regarding display and internal use.

Do we want to match the `$vpX` values maybe? We may need getZoom() then in addition.

MichaelPFrey:
I saw the issue with the display camera value not matching the value read via dbus.
But one thing to note: getRotation and getTranslation should match rotateTo and translateTo.

Users of the dbus interface should try to avoid read-modifiy-write operations due to the interprocess communication overhead, but I can think of use cases, where it would make sense.
(e.g. taking the camera control away from the user and then reseting it to where it was before)
So, any math we apply to getters has to reversed by the setters.

MichaelPFrey:
an other thing to note:
Future input drivers might start out as "dbus plugins", before being integrated into openscad.
It is so much easier to develop, when openscad can run continuously and only a small wrapper program has to be recompiled each time.

Where I am getting at: The dbus api has to be somewhat consistent with the internal driver structure, so that porting a driver can be relatively straight forward.

kintel:
I think one challenge is that the camera system in OpenSCAD is a bit half baked, and once we refactor it into a more standard way of managing cameras, this API may have to change with it.
..but that's not a big deal as long as this is experimental functionality.

MichaelPFrey:
Interesting. but even if this feature would be stable: Given Openscads release cycle, we might as well allow use breaking changes on the Input Driver between releases. The DBus driver would then simply announce as a new version of the API on the dbus.
I also presume that a reworked camera system would be a feature branch, allowing us to work on it in parallel.

kintel:
Agreed, camera refactoring is something that could be done in parallel, just wanted to mention it so we don't spend too much time working around odd camera behavior :)

PR #2282, opened 2018-02-04
inputdriver 5: expose rotate2 via dbus 

MichaelPFrey:
this is building on #2280 (which are the first two commits in this pull request)

Naming it rotate2 is not ideal, but I am unable to come up with a better one.

The name rotate2 comes from the method of the same name that was unused in qglview.

Naming it relative rotate would be confusing with the relative flag of the normal rotate function.

On the other-hand: Rotate2 could be confused with RotateTo which is not ideal either.

### DIFF
--- a/org.openscad.OpenSCAD.xml
+++ b/org.openscad.OpenSCAD.xml
@@ -1,6 +1,6 @@
 <!DOCTYPE node PUBLIC "-//freedesktop//DTD D-BUS Object Introspection 1.0//EN"
 "http://www.freedesktop.org/standards/dbus/1.0/introspect.dtd">
-<node>
+<node  xmlns:tp="http://telepathy.freedesktop.org/wiki/DbusSpec#extensions-v0http://telepathy.freedesktop.org/wiki/DbusSpec#extensions-v0">
   <interface name="org.openscad.OpenSCAD">
     <method name="zoom">
       <arg name="zoom" type="d" direction="in"/>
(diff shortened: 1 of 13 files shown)


t-paul:
As it's still all experimental, I guess we are still able to change things (a.k.a. not-yet-stable-interface). So maybe we can introduce some more logical names in the interface, regardless of what's used internally.

MichaelPFrey:
I am open to suggestions. Hhmm... the description of rotate2 is
>     // This vector describes the rotation.
>     // The direction of the vector is the angle around which to rotate, and
>     // the length of the vector is the angle by which to rotate
RotateByVector ?

kintel:
Using the length of the vector an an angle is a bit odd. Any particular reason for this?
I would try to make it more explicit.

MichaelPFrey:
You are asking a good question. The rotate2 method in qglview is currently actually unused, so... no idea where this method was coming from and what the intention of it was. The comment is also directly lifted from there.
Using it on a joystick feels like a natural control.

Note that the current rotate function even when set to "relative" is kind of awkward.
To explain what I mean: open the candleStand example (any version of openSCAD - it does not need to be from this feature branch), (candle holders looking up) and rotate it around its axis.
Then turn it upside down (candle holder looking down) and turn it around its axis. The rotation is inverted.
There is no issue with that as long as I use the mouse to rotate the object, but my intuition blocks when I use a 3D mice or gamepad and axis get inverted for no apparent reason.

PR #2283, opened 2018-02-04
Fix build on Fedora 27

piffio:
In order to properly build OpenSCAD on Fedora 27 a couple of fixes
are required:

* Check for QT VERSION in arch-specific header files
* Install lizip-devel and ccache as build dependencies


t-paul:
Looks like fedora took a different path for that, on Debian I still see `qconfig.h`, but that change looks fully compatible with that too.

PR #2291, opened 2018-02-10
 rewrite of createGroup using regex 

MichaelPFrey:
see this pull request as a formal question whether iterating over characters or regex is preferred to handle strings in a situation like this.

I would like to add a feature to the customizer and would like to get the coding style down, before writing new code.

### DIFF
--- a/src/comment.cpp
+++ b/src/comment.cpp
@@ -4,7 +4,10 @@
 #include <string>
 #include <vector>
 #include <boost/range/adaptor/reversed.hpp>
-#include <regex>
+// gcc 4.8 and earlier have issues with std::regex see
+// #2291 and https://stackoverflow.com/questions/12530406/is-gcc-4-8-or-earlier-buggy-about-regular-expressions
+// therefor, we use boost::regex
+#include <boost/regex.hpp>
 
 struct GroupInfo {
 	std::string commentString;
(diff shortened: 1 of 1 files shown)


MichaelPFrey:
1072 to 1077 failed
most relevant should be
>  Start 1072: customizertest_group
> 1072/1099 Test 1072: customizertest_group ...................................***Failed    0.24 sec

Oddly, does test seem to pass on my machine:
group
My GCC is
```
gcc (Ubuntu 7.2.0-8ubuntu3) 7.2.0
Copyright (C) 2017 Free Software Foundation, Inc.
```

while travis is using
```
gcc (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4
Copyright (C) 2013 Free Software Foundation, Inc.
```

Looking trough the code of openSCAD, I see that currently `#include <boost/regex.hpp>` is used by other parts of the program. Now that I see that it is obvious - I will adapt my code.

On an other note: The tests are missing some test cases in my opinion:
Multiple sets of square brackets:
`/*[Properties] [of] [support] */`
and arbitrary text outside of those square brackets:
`/*[Properties] [of] [support] show_if(lamp) */`

But extending those test cases is a separate task.

kintel:
@MichaelPFrey If it's easy to port from `boost::regex` to C++11 `std::regex`, we should do that. The boost code is probably just left behind from pre-C++11 times.

MichaelPFrey:
I do not understand, as we have conflicting goals.
One goal is to move toward C++11.
Then we have Travis using gcc 4.8.4, that has known issues with C++11 style regex:
https://stackoverflow.com/questions/12530406/is-gcc-4-8-or-earlier-buggy-about-regular-expressions

I see two ways: Either I make my code working for gcc 4,8 (using boost) or someone changes Travis to use a more recent gcc.

For me, both ways work. All I want is to get rid of handling individual characters.

kintel:
Wops, I wasn't aware of that gcc issue. Let's use boost then, and perhaps leave a comment describing why so that the next person reading this code doesn't try to port to C++11..

MichaelPFrey:
No problem. Porting of this simple regex expression is/was straight forward.
Now lets see if Travis is happy.

PR #2295, opened 2018-02-11
costumizer: Parameter widget refraction

MichaelPFrey:
My aim is to improve the readability of the code, before changing datastructers and adding features.


src/parameter/ParameterWidget.cc
```diff
@@ -68,6 +68,7 @@ ParameterWidget::~ParameterWidget()
 {
 }
 
+//delets the currently selected/active Parameter set
```
kintel: s/delets/deletes/

PR #2300, opened 2018-02-13
Add render color theme Cornfield Dark

piit79:
Version of the trademark Cornfield render color theme with dark background

### DIFF
--- a/color-schemes/render/cornfield-dark.json
+++ b/color-schemes/render/cornfield-dark.json
@@ -1,0 +1,19 @@
+{
+    "name" : "Cornfield Dark",
+    "index" : 1050,
+    "show-in-gui" : true,
+
+    "colors" : {
+        "background" :         "#000000",
+        "axes-color" :         "#e5e5e5",
+        "opencsg-face-front" : "#f9d72c",
+        "opencsg-face-back" :  "#9dcb51",
+        "cgal-face-front" :    "#f9d72c",
+        "cgal-face-back" :     "#9dcb51",
+        "cgal-face-2d" :       "#00bf99",
+        "cgal-edge-front" :    "#ffec5e",
+        "cgal-edge-back" :     "#abd856",
+        "cgal-edge-2d" :       "#ff0000",
+        "crosshair" :          "#f0f0f0"
+    }
+}


MichaelPFrey:
For those that want a picture for reference:
bildschirmfoto von 2018-02-15 20-08-20

piit79:
Thanks @MichaelPFrey! Should have done that myself. That's exactly how I run OpenSCAD, just with Monaco font ;)

MichaelPFrey:
Just of interest: Do mind the white/gray borders? QT has support for stylesheets.
I made a simple hack job using http://tech-artists.org/t/release-qt-dark-orange-stylesheet/2287#post14381
bildschirmfoto von 2018-02-15 20-35-20
It would not be too hard to make user selectable stylesheets possible.
The hack job is just one line, the real implementation will have some overhead, but the most complicated thing will be to find the right place in the preference dialog.

piit79:
I don't have that problem - I'm using Breeze Dark system theme in KDE Neon and it all matches nicely :slightly_smiling_face:
openscad-dark
By the way, noticed the code in your screenshot - is that some kind of tutorial/example? I need to look into the `offset()` operation and how it makes rounding corners easier. It's quite painful the way I do it. I sometimes use `minkowski()` and in simpler situations just add quarter discs manually...

kintel:
This is the same as Cornfield, just with black background and white axes, right?
The black background is a bit harsh against the rest of the gray backgrounds; did you test out making the background similar to the editor?

MichaelPFrey:
@piit79 For the screenshots I used the examples that ship with OpenSCAD.

piit79:
Good point, @kintel ... Truth is, without much thought, I just used the scheme "Tomorrow Night" as the base (as that's what I was using) and copied over the model colours from the internal "Cornfield" theme. Didn't realize the background is pure black (`#000000`) :slightly_smiling_face:
It does look better with the background from "Monokai":
openscad-dark-grey
Trouble is, this only looks good with the "Monokai" editor theme - it looks similarly out of place with other dark themes with different backgrounds. The pure black looks a bit more neutral maybe.
Not sure what to do now. What do others think? I will personally be using a version with the "Monokai" background to match the editor. I could create a version for each dark theme - like "Cornfield Monokai", "Cornfield Solarized Dark" etc. However, I've got a feeling this would not go well with the developers :wink:
The best solution would be a checkbox "Use editor background color". But that would require quite a bit more work. I'll try to look into it.
Any input much appreciated.

MichaelAtOz:
I prefer backgrounds to be solid colours, mainly white or black, so in this case black. It helps when you export images and want to e.g. change the background to transparent.

kintel:
Btw., are everyone aware that color schemes can be user-installed ?

kintel:
Having a separate (GUI and cmd-line) option for background color might be the right solution, especially considering the work towards supporting transparent backgrounds for PNG export: #1485

MichaelAtOz:
@piit79 & @MichaelPFrey I don't suppose you'd be OK with changing "axis-color" & "crosshair" to "#ffffff"?
To go with another theme Cornfield White, where "axis-color" & "crosshair" is "#000000" & "background" is "#ffffff". 
Basically makes the two inverse monochrome for the background+decorations, with cornfield object colours. Makes post processing of exported images a little easier.
Yes, I know I can make my own & have done so.
I think these binary versions would be handy for everyone in the distribution.

kintel:
@MichaelAtOz What sort of post-processing are you doing? Are there any deeper issues that we could resolve to help with that?

MichaelAtOz:
@kintel I suppose not so much 'processing' in the full sense.
As I mentioned before, changing backgrounds to transparent, using Gimp,
changing colours, lightening/darkening, cropping, resizing.
Some of that can be done via command-line and/or purposeful sizing/position in the GUI before export.
As I mentioned I can make my own scheme, but having black & white bits is good because they don't get caught up as much when dealing with shades. For example if the axis letters or scale-marker was a shade of grey, they would match the grays in the shading - see below.
I mainly proposed the above to have two mono schemes for others who may try to do similar.

This is around graphics for T-shirts and the like. 
Images on a white/light T v's a black/darker one, need finessing, usually the shading from OpenSCAD is too harsh.

I had though of asking for lighting control..._you asked_...

For example: **(yes people we are going to have Merch)**
The first image (IIRC - I'm still picking up where I left off) is as-exported - good for dark T's, the second, for a white T cut back the dark shading. 
(right-click open-in-new-tab to see them properly, first has white text)
wbt-bbt letters tp cm logo t-shirts v3 2864x3600
wbt-bbt letters tp cm logo t-shirts v3 2864x3600 lighter2 wb

kintel:
@MichaelAtOz Would rendering onto a transparent background make your post-processing easier? (aka. #1485)

Issue #2303, opened 2018-02-18
High memory consumption for module recursion


thehans:
I have a test module that creates nothing but a deeply nested bunch of group() nodes.
Moderate levels of recursion result in exploding memory usage.

```
module recur(i) {
  if (i > 0) recur(i-1);
}

recur(400);

// Memory usage for various i values, using F5 preview
// 0    29.0MB
// 100  34.2MB
// 200  78.5MB
// 300  190.2MB
// 400  442.6MB
// 500  1.0GB
// 600  1.3GB
// 700  2.0GB
// 800  2.9GB
// 900  4.0GB
// 1000 5.4GB
// Each tested from a freshly opened instance of openscad
```
Memory usage seems to increase quadratically with the depth of recursion.

I tried to do some profiling and came up with this, which appears to show that most of the time is being spent in NodeDumper::visit

![screenshot from 2018-02-17 18-34-02](https://user-images.githubusercontent.com/252233/36347028-469d3950-1412-11e8-9aa2-6708a5c33439.png)

I don't know the Node traversing code well, but should it even be calling NodeDumper for a basic F5 preview?

kintel:
Thanks for looking into this.
NodeDumper is typically used to generate cache keys for the module cache.

thehans:
OK I thought maybe NodeDumper was only meant for exporting .csg file.

It seems like *some* cache here is not being utilized efficiently/correctly, but I still don't know in what way.

I tried some more profiling.  This is including the recent string building optimizations, and I lowered the recursion depth in the scad, and increased perf record_size to max because I think the stack frames were getting truncated in the previous screenshot.  In this flame graph you can see the stack all the way to the bottom for most of the samples:
![screenshot from 2018-02-18 12-43-47](https://user-images.githubusercontent.com/252233/36355474-6b0659a6-14a9-11e8-83ed-d88083916831.png)

So in mainwin.cc:121 there is the call to Tree::getString which is what kicks off the whole NodeDumper traversal, and it has this FIXME comment:
```
			// Dump the tree (to initialize caches).
			// FIXME: We shouldn't really need to do this explicitly..
			this->tree.getString(*this->root_node);
```

Any insights?

thehans:
I still barely understand what I'm doing with `perf` here, and ideally would be recording something more specific to memory like malloc calls or something.

kintel:
[...]

PR #2304, opened 2018-02-18
Optimize NodeDumper string building

thehans:
doesn't use any less memory, but runs recur(1000) in 4s instead of 8s

### DIFF
--- a/src/nodedumper.cc
+++ b/src/nodedumper.cc
@@ -39,33 +39,35 @@ void NodeDumper::handleIndent(const State &state)
 	including braces and indentation.
 	All children are assumed to be cached already.
  */
-void NodeDumper::dumpChildBlock(const AbstractNode &node, std::stringstream &dump)
+void NodeDumper::dumpChildBlock(const AbstractNode &node, std::stringstream &dump) const
 {
-	if (!this->visitedchildren[node.index()].empty()) {
+	const auto &it = this->visitedchildren.find(node.index());
+
+	if (it == this->visitedchildren.end() || it->second.empty()) {
+		dump << ";";
+	} else {
 		dump << " {\n";
 		dumpChildren(node, dump);
 		dump << this->currindent << "}";
 	}
-	else {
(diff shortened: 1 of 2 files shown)


thehans:
Related to Issue #2303

src/nodedumper.cc
```diff
@@ -39,24 +39,21 @@ void NodeDumper::handleIndent(const State &state)
...
 	}
-	return dump.str();
 }
 
-std::string NodeDumper::dumpChildren(const AbstractNode &node)
+void NodeDumper::dumpChildren(const AbstractNode &node, std::stringstream &dump)
 {
-	std::stringstream dump;
+	bool empty = true;
 	for (auto child : this->visitedchildren[node.index()]) {
```
kintel: We could optimize this a tiny step further and remove the need for that empty variable.
This will also make it possible to make this into a const function. Similar for `dumpChildBlock()`.

```
  const auto &it = this->visitedchildren.find(node.index());
  if (it == this->visitedchildren.end()) return;
```

src/nodedumper.cc
```diff
@@ -39,35 +39,33 @@ void NodeDumper::handleIndent(const State &state)
...
-	if (!this->visitedchildren[node.index()].empty()) {
+	const auto &it = this->visitedchildren.find(node.index());
+
+	if (it == this->visitedchildren.end() || it->second.empty()) {
+		dump << ";";
+	} else {
 		dump << " {\n";
-		const auto &chstr = dumpChildren(node);
-		if (!chstr.empty()) dump << chstr << "\n";
+		dumpChildren(node, dump);
```
kintel: I think I'll merge this as is, but this constructs makes me think that we should manage stream operators better.
A quickfix could be to return the stream reference from `dumpChildren()`, making this construct look more natural, e.g:
```
  dump << " {\n" 
       << dumpChildren(node, dump)
       << this->currindent
       << "}";
```

PR #2308, opened 2018-02-19
Add indent option, only indent when necessary

thehans:
Still does not address memory usage when dumping with indentation, but its a step forward.


thehans:
experiencing git difficulties.  didn't have my branch up to date when i created this

kintel:
Btw., thanks again for taking a pro-active role with this. It really increases our bandwidth!

thehans:
I was thinking another way of handling the indentation level, would be to change `NodeDumper::handleIndent` so that instead of it constructing `this->currindent` string, it would just keep a indent counter.   Then in the functions that write indents to the stringstream,  do a `'\t'`character fill using that counter.  But I wasn't sure if that would actually be an improvement.

kintel:
An indent counter sounds like a decent optimization. It would still just be an implementation detail, and make the constructor choice of indentation string an easy add-on.

thehans:
I can't picture how an indent string constructor would be usable though, because the indent level changes as it traverses?

kintel:
In case we'd like to use spaces vs. tabs in some contexts, and since the "no indent" automatically falls out of that by specifying a zero-length string.
This would be a NodeDumper constructor argument, as this visitor is only constructed once.

thehans:
Ah, I see, yeah that makes sense.  The string represents one single indentation. not overall.

thehans:
It was pretty straightforward to support string so I went ahead and did that.
As an example of trying out other options, I used two spaces `"  "` as the indent string, only in the case of menu action for Design -> Display CSG Tree...

So the popup window results won't be char-for-char identical to CSG file export, not sure if anyone cares.
I can change that back to `"\t"` if wanted.

src/nodedumper.cc
```diff
@@ -26,11 +26,10 @@ bool NodeDumper::isCached(const AbstractNode &node) const
 void NodeDumper::handleIndent(const State &state)
 {
 	if (state.isPrefix()) {
-		this->currindent += "\t";
+		this->currindent++;
 	}
 	else if (state.isPostfix()) {
-		this->currindent.erase((this->currindent.length() >= 1) ? 
-													 this->currindent.length() - 1 : 0);
+		this->currindent = this->currindent > 0 ? this->currindent - 1 : 0;
```
kintel: Now that I read this code, I don't think `currindent` can ever become negative, so I think `this->currindent--;` should be sufficient

Issue #2103, opened 2017-09-07
File Loading Warning in 2017 windows snapshots


runsun:
File loading warning in 2017 window snapshots:
(Windows 10 standard edition, 64 bit)

An error message about "Parameter Set" shows up in pink in the console for a sec then disappears. Supposed we are loading "c:/myOpenScad/file1.scad", the error message is : 

``
ERROR: Cannot open Parameter Set:'c:/myOpenScad/file1.json': c:/myOpenScad/file1.json: cannot open file 
Loaded design: c:/myOpenScad/file1.scad
``

That is, when loading `whatever.scad`, it complains it can't find `whatever.json`. 

This appears to be a non-lethal error and everything seems to proceed as usual. 

This error was introduced since OpenSCAD-2017.01.20-x86-64-Installer.exe 

Here are some snapshots I tested: 

2015.03-2 (latest release)
OpenSCAD-2015.10.17-x86-64-Installer.exe 
OpenSCAD-2016.01.26-x86-64-Installer.exe   
OpenSCAD-2016.02.01-x86-64-Installer.exe 
OpenSCAD-2016.02.09-x86-64-Installer.exe  
OpenSCAD-2016.04.06-x86-64-Installer.exe  
OpenSCAD-2016.05.12-x86-64-Installer.exe  
OpenSCAD-2016.07.20-x86-64-Installer.exe  
OpenSCAD-2016.09.18-x86-64-Installer.exe  
OpenSCAD-2016.10.04-x86-64-Installer.exe   
OpenSCAD-2016.11.11-x86-64-Installer.exe   
OpenSCAD-2017.01.20-x86-64-Installer.exe : Warning 
OpenSCAD-2017.08.30-x86-64-Installer.exe : Warning

t-paul:
Can you please create separate issues for different errors?

t-paul:
@kintel I guess that would need to go into the "next release" bucket?

runsun:
As you guys would expect, no more warning after turning off the Customizer

MichaelPFrey:
isn´t this issue a duplicate of #1927?

kintel:
Closing as duplicate of #1927

PR #2311, opened 2018-02-20
costumizer: improved file handling

### LINKED ISSUE #2307: Costumizer: replacing .scad with .json
While reading the code of the customizer, I found this:
`this->jsonFile = scadFile.replace(".scad", ".json").toStdString();`
which has the implication, that when I have a file called "this.scad" with a path like this:
`/home/userName/myProjects/files.scad/this.scad`
the json file ends up in the folder
`/home/userName/myProjects/files.json/this.json`
That is actually a cool hidden feature.

But thinking deeper, this has to be a bug/issue: Maybe someone is using a username like students.scad or mu.scadet, resulting in OpenSCAD trying to write in someone elses userprofile...

So, this has to be a bug. If no one disagrees, I would write a fix for that issue.
Most robust would be to remove the file ending from the scad file and then to add .json.
(Keeping in mind, that the file ending can be any capitalization of .sCAd, none, ...)

If someone thinks that this is a feature, we need a very clear specification to workaround issues like  usernames randomly containing .scad.

MichaelPFrey:
* move the writeable check from "readParameterSet" to "readFile"
   * This also fixes an issue with empty (0 bytes) json files: Empty json files cause an exception, causing readParameterSet to return false, without actually checking if the file would be writable.
* checking if the file exists, before trying to check for readability or writeabilty
*central setFile function with replace_extension instead of replace (closes #2307)

### DIFF
--- a/src/parameter/ParameterWidget.cc
+++ b/src/parameter/ParameterWidget.cc
@@ -115,17 +115,13 @@ void ParameterWidget::readFile(QString scadFile)
 	setFile(scadFile);
 	bool exists = boost::filesystem::exists(this->jsonFile);
 	bool writeable = false;
+	bool readable = false;
 
 	if(exists){
-		bool readable = readParameterSet(this->jsonFile);
+		readable = readParameterSet(this->jsonFile);
 
-		//check whether file is writeable
-		std::fstream file;
-		file.open(this->jsonFile, std::ios::app);
-		if (file.is_open()) {
-			file.close();
-			writeable = true;
-		}
+		//check whether file is writeable or not
+		if (std::fstream(this->jsonFile, std::ios::app)) writeable = true;
 	}
 
 	if(writeable || !exists){
--- a/src/parameter/ParameterWidget.cc
+++ b/src/parameter/ParameterWidget.cc
@@ -104,16 +104,26 @@ void ParameterWidget::onSetSaveButton()
 void ParameterWidget::readFile(QString scadFile)
 {
 	this->jsonFile = scadFile.replace(".scad", ".json").toStdString();
-	bool readable=readParameterSet(this->jsonFile);
-	if(readable){
+	bool writeable = false;
+
+	bool readable = readParameterSet(this->jsonFile);
+
+	//check whether file is writeable
+	std::fstream file;
+	file.open(filename, std::ios::app);
+	if (file.is_open()) {
+		file.close();
+		writeable = true;
+	}
+
+	if(writeable){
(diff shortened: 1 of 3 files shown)


src/parameter/ParameterWidget.cc
```diff
@@ -101,19 +105,37 @@ void ParameterWidget::onSetSaveButton()
...
+
+	if(exists){
+		bool readable = readParameterSet(this->jsonFile);
+
+		//check whether file is writeable
+		std::fstream file;
+		file.open(this->jsonFile, std::ios::app);
+		if (file.is_open()) {
+			file.close();
+			writeable = true;
```
kintel: I think these 5 lines of code could be simplified to smth. like this:
```
if (std::stream{this->jsonFile, std::ios::app}) writeable = true;
```

kintel:
Looks great! I made a suggestion to simplify some code..

src/parameter/ParameterWidget.cc
MichaelPFrey: if (std::fstream(this->jsonFile, std::ios::app)) writeable = true;

works, as in:
* it detects, when a json file is read only
* if the json file is not readonly, I can add a new parameter set and save
but I wonder how.

this is using the [constructor of fstream](http://www.cplusplus.com/reference/fstream/fstream/open/). It creates the fstream on the stack. Ah: I found the hint I needed: [RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization), the enclosing scope beeing the brackets of the "if".

Neat :-)

kintel: I was a bit terse in my comment - glad you got it :)
Don't hesitate to ask me to clarify. Hard to gauge what detail level to add..

MichaelPFrey: I have the hints I need, thank you. I am always a bit confused where C++ and QT are on the continuum between C and Java (for both of which I have college education). Doesn´t help that a lot of modern C++/QT text books skim over memory management and Stackoverflow tends to use C solutions for C++ problems.

kintel: These are good reads which cover modern C++:
* A tour of C++ (https://www.amazon.com/Tour-C-Depth/dp/0321958314)
* Effective Modern C++ (e.g. https://edisciplinas.usp.br/pluginfile.php/1995323/mod_resource/content/1/Effective%20Modern%20C%2B%2B%202014.pdf)

MichaelPFrey:
that could have coincidentally have fixed #2103
(hard to test - I was  not looking for that issue)

PR #2313, opened 2018-02-20
Rewrite NodeCache to store single string with indices

thehans:
Still WIP, trying to debug a segfault
For Issue #2303

### DIFF
--- a/src/CGALCache.h
+++ b/src/CGALCache.h
@@ -12,9 +12,9 @@ public:
 
 	static CGALCache *instance() { if (!inst) inst = new CGALCache; return inst; }
 
-	bool contains(const size_t id) const { return this->cache.contains(id); }
-	shared_ptr<const class CGAL_Nef_polyhedron> get(const size_t id) const;
-	bool insert(const size_t id, const shared_ptr<const CGAL_Nef_polyhedron> &N);
+	bool contains(const std::string &id) const { return this->cache.contains(id); }
+	shared_ptr<const class CGAL_Nef_polyhedron> get(const std::string &id) const;
+	bool insert(const std::string &id, const shared_ptr<const CGAL_Nef_polyhedron> &N);
 	size_t maxSize() const;
 	void setMaxSize(size_t limit);
 	void clear();
@@ -30,5 +30,5 @@ private:
 		~cache_entry() { }
 	};
 
-	Cache<size_t, cache_entry> cache;
+	Cache<std::string, cache_entry> cache;
 };
--- a/src/CGALCache.h
+++ b/src/CGALCache.h
@@ -12,9 +12,9 @@ public:
 
 	static CGALCache *instance() { if (!inst) inst = new CGALCache; return inst; }
 
-	bool contains(const std::string &id) const { return this->cache.contains(id); }
-	shared_ptr<const class CGAL_Nef_polyhedron> get(const std::string &id) const;
-	bool insert(const std::string &id, const shared_ptr<const CGAL_Nef_polyhedron> &N);
+	bool contains(const size_t id) const { return this->cache.contains(id); }
+	shared_ptr<const class CGAL_Nef_polyhedron> get(const size_t id) const;
+	bool insert(const size_t id, const shared_ptr<const CGAL_Nef_polyhedron> &N);
 	size_t maxSize() const;
 	void setMaxSize(size_t limit);
 	void clear();
@@ -30,5 +30,5 @@ private:
 		~cache_entry() { }
 	};
 
-	Cache<std::string, cache_entry> cache;
+	Cache<size_t, cache_entry> cache;
 };
--- a/src/Tree.cc
+++ b/src/Tree.cc
@@ -4,14 +4,12 @@
 
 #include <assert.h>
 #include <algorithm>
-#include <functional>
 #include <sstream>
 #include <tuple>
 
 Tree::~Tree()
 {
 	this->nodecachemap.clear();
-	this->nodeidmap.clear();
 }
 
 /*!
@@ -44,43 +42,24 @@ const std::string Tree::getString(const AbstractNode &node, const std::string &i
 	is stripped for whitespace. Especially indentation whitespace is important to
 	strip to enable cache hits for equivalent nodes from different scopes.
 */
-size_t Tree::getId(const AbstractNode &node) const
+const std::string Tree::getIdString(const AbstractNode &node) const
 {
 	assert(this->root_node);
 	const std::string indent = "";
 	const bool idString = true;
 	const bool idPrefix = false;
 
-	auto i = node.index();
-	auto result = nodeidmap.find(i);
-	if (result == nodeidmap.end()) {
-		
-		// Retrieve a nodecache given a tuple of NodeDumper() options  
-		NodeCache &nodecache = this->nodecachemap[make_tuple(indent,idString,idPrefix)];
-		
-		if (!nodecache.contains(node)) {
-			nodecache.clear();
-			NodeDumper dumper(nodecache, this->root_node, indent, idString, idPrefix);
-			dumper.traverse(*this->root_node);
-			assert(nodecache.contains(*this->root_node) &&
-						 "NodeDumper failed to create id cache");
-		}
-		
-		const std::string &nodeidstr = nodecache[node];
-		size_t idhash = std::hash<std::string>{}(nodeidstr);
-		#if DEBUG
-			const size_t HALFLENGTH = 40;
-			if (nodeidstr.size() > HALFLENGTH*2) {
-				PRINTDB("nodeidmap insert {%1%,%2$X} => \"%3% ... %4%\"", i % idhash % nodeidstr.substr(0,HALFLENGTH) % nodeidstr.substr(nodeidstr.size()-HALFLENGTH,HALFLENGTH) );
-			} else {
-				PRINTDB("nodeidmap insert {%1%,%2$X} => \"%3%\"}", i % idhash % nodeidstr );
-			}
-		#endif
-		
-		return nodeidmap[i] = idhash;
-	} else {
-		return result->second;
+	// Retrieve a nodecache given a tuple of NodeDumper constructor options
+	NodeCache &nodecache = this->nodecachemap[make_tuple(indent,idString,idPrefix)];
+
+	if (!nodecache.contains(node)) {
+		nodecache.clear();
+		NodeDumper dumper(nodecache, this->root_node, indent, idString, idPrefix);
+		dumper.traverse(*this->root_node);
+		assert(nodecache.contains(*this->root_node) &&
+					 "NodeDumper failed to create id cache");
 	}
+	return nodecache[node];
 }
 
 /*!
@@ -90,5 +69,4 @@ void Tree::setRoot(const AbstractNode *root)
 {
 	this->root_node = root; 
 	this->nodecachemap.clear();
-	this->nodeidmap.clear();
 }
--- a/src/Tree.cc
+++ b/src/Tree.cc
@@ -44,7 +44,7 @@ const std::string Tree::getString(const AbstractNode &node, const std::string &i
 	is stripped for whitespace. Especially indentation whitespace is important to
 	strip to enable cache hits for equivalent nodes from different scopes.
 */
-const size_t Tree::getId(const AbstractNode &node) const
+size_t Tree::getId(const AbstractNode &node) const
 {
 	assert(this->root_node);
 	const std::string indent = "";
--- a/src/Tree.cc
+++ b/src/Tree.cc
@@ -24,11 +24,11 @@ const std::string Tree::getString(const AbstractNode &node, const std::string &i
 	bool idString = false;
 	bool idPrefix = false;
 
-	// Retrieve a nodecache given a tuple of NodeDumper() options  
+	// Retrieve a nodecache given a tuple of NodeDumper constructor options
 	NodeCache &nodecache = this->nodecachemap[std::make_tuple(indent,idString,idPrefix)];
 
 	if (!nodecache.contains(node)) {
-		NodeDumper dumper(nodecache, indent, idString, idPrefix);
+		NodeDumper dumper(nodecache, this->root_node, indent, idString, idPrefix);
 		dumper.traverse(*this->root_node);
 		assert(nodecache.contains(*this->root_node) &&
 					 "NodeDumper failed to create a cache");
@@ -60,7 +60,7 @@ size_t Tree::getId(const AbstractNode &node) const
 		
 		if (!nodecache.contains(node)) {
 			nodecache.clear();
-			NodeDumper dumper(nodecache, indent, idString, idPrefix);
+			NodeDumper dumper(nodecache, this->root_node, indent, idString, idPrefix);
 			dumper.traverse(*this->root_node);
 			assert(nodecache.contains(*this->root_node) &&
 						 "NodeDumper failed to create id cache");
--- a/src/Tree.cc
+++ b/src/Tree.cc
@@ -71,9 +71,9 @@ size_t Tree::getId(const AbstractNode &node) const
 		#if DEBUG
 			const size_t HALFLENGTH = 40;
 			if (nodeidstr.size() > HALFLENGTH*2) {
-				PRINTB("nodeidmap insert {%1%,%2$X} => \"%3% ... %4%\"", i % idhash % nodeidstr.substr(0,HALFLENGTH) % nodeidstr.substr(nodeidstr.size()-HALFLENGTH,HALFLENGTH) );
+				PRINTDB("nodeidmap insert {%1%,%2$X} => \"%3% ... %4%\"", i % idhash % nodeidstr.substr(0,HALFLENGTH) % nodeidstr.substr(nodeidstr.size()-HALFLENGTH,HALFLENGTH) );
 			} else {
-				PRINTB("nodeidmap insert {%1%,%2$X} => \"%3%\"}", i % idhash % nodeidstr );
+				PRINTDB("nodeidmap insert {%1%,%2$X} => \"%3%\"}", i % idhash % nodeidstr );
 			}
 		#endif
 		
--- a/src/Tree.cc
+++ b/src/Tree.cc
@@ -9,26 +9,25 @@
 
 Tree::~Tree()
 {
-	this->nodecache.clear();
-	this->nodeidcache.clear();
+	this->nodecachemap.clear();
 }
 
 /*!
 	Returns the cached string representation of the subtree rooted by \a node.
 	If node is not cached, the cache will be rebuilt.
 */
-const std::string &Tree::getString(const AbstractNode &node, const std::string &indent) const
+const std::string Tree::getString(const AbstractNode &node, const std::string &indent) const
 {
 	assert(this->root_node);
-	if (!this->nodecache.contains(node)) {
-		this->nodecache.clear();
-		this->nodeidcache.clear();
-		NodeDumper dumper(this->nodecache, indent, false);
+	NodeCache &nodecache = this->nodecachemap[indent];
+
+	if (!nodecache.contains(node)) {
+		NodeDumper dumper(nodecache, indent, false);
 		dumper.traverse(*this->root_node);
-		assert(this->nodecache.contains(*this->root_node) &&
+		assert(nodecache.contains(*this->root_node) &&
 					 "NodeDumper failed to create a cache");
 	}
-	return this->nodecache[node];
+	return nodecache[node];
 }
 
 /*!
@@ -39,25 +38,33 @@ const std::string &Tree::getString(const AbstractNode &node, const std::string &
 	is stripped for whitespace. Especially indentation whitespace is important to
 	strip to enable cache hits for equivalent nodes from different scopes.
 */
-const std::string &Tree::getIdString(const AbstractNode &node) const
+const std::string Tree::getIdString(const AbstractNode &node) const
 {
 	assert(this->root_node);
+	std::string indent = "";
+	NodeCache &nodecache = this->nodecachemap[indent];
 
-	if (!this->nodeidcache.contains(node)) {
-		const std::string &nodestr = getString(node, "");
-		const boost::regex re("[^\\s\\\"]+|\\\"(?:[^\\\"\\\\]|\\\\.)*\\\"");
-		std::stringstream sstream;
-		boost::sregex_token_iterator i(nodestr.begin(), nodestr.end(), re, 0);
-		std::copy(i, boost::sregex_token_iterator(), std::ostream_iterator<std::string>(sstream));
-
-		const auto &result = this->nodeidcache.insert(node, sstream.str());
-		PRINTDB("Id Cache MISS: %s", result);
-		return result;
-	} else {
-		const auto & result = this->nodeidcache[node];
-		PRINTDB("Id Cache HIT:  %s", result);
-		return result;
+	if (!nodecache.contains(node)) {
+		nodecache.clear();
+		NodeDumper dumper(nodecache, "", false);
+		dumper.traverse(*this->root_node);
+		assert(nodecache.contains(*this->root_node) &&
+					 "NodeDumper failed to create id cache");
 	}
+
+	// FIXME: this is a HACK just to get things to compile and run
+	// Should not run regex on every call
+	// I think ideally we can have nodedumper take a parameter to generate pre-formatted idstring style
+	const std::string &nodestr = nodecache[node];
+	//                     [^\s\"]+|\"(?:[^\"\\]|\\.)*\"
+	const boost::regex re("[^\\s\\\"]+|\\\"(?:[^\\\"\\\\]|\\\\.)*\\\"");
+	std::stringstream sstream;
+	boost::sregex_token_iterator i(nodestr.begin(), nodestr.end(), re, 0);
+	std::copy(i, boost::sregex_token_iterator(), std::ostream_iterator<std::string>(sstream));
+
+	return sstream.str();
+
+
 }
 
 /*!
@@ -66,5 +73,5 @@ const std::string &Tree::getIdString(const AbstractNode &node) const
 void Tree::setRoot(const AbstractNode *root)
 {
 	this->root_node = root; 
-	this->nodecache.clear();
+	this->nodecachemap.clear();
 }
--- a/src/Tree.cc
+++ b/src/Tree.cc
@@ -4,12 +4,14 @@
 
 #include <assert.h>
 #include <algorithm>
+#include <functional>
 #include <sstream>
 #include <tuple>
 
 Tree::~Tree()
 {
 	this->nodecachemap.clear();
+	this->nodeidmap.clear();
 }
 
 /*!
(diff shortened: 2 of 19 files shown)


thehans:
OK, I got the segfaults worked out.  I still don't quite have the formatting correct on NodeDumper string, Its putting a bunch of semicolons at the end, maybe shouldn't have completely removed all the child handling code.  I think id strings are still really broken, but some of the tests are passing now.

thehans:
Getting closer to 100% passing

thehans:
I have a few issues/questions remaining:
The main issue, which I haven't found a solution to is the failing tests that have to do with root modifiers.
The problem seems to be with running `Tree::getString` or `Tree::getId` when tree.root_node is not the "absolute root", so it never gets a call to `NodeDumper::visit(State &state, const RootNode &node)`, 
and doesn't properly init the cache string(happens in postFix of RootNode handling).
I'm not sure the best way to handle this.
Perhaps `NodeDumper::visit(State &state, const AbstractNode &node)` code could somehow check if it is processing the root of the Tree, and then duplicate the cache initializing code from `NodeDumper::visit(State &state, const RootNode &node)` in that case.

A question relating to highlight/background modifiers.  I think its already being handled correctly, but not 100% sure: if node has a highlight or background modifier directly on itself, that modifier is **not** supposed to be included in its string / id string, right?

I also changed the code to keep just hash of the node id strings, so the CGALCache and CSGTextCache key sizes are kept reasonable.
However `CSGTextCache` and `CSGTextRenderer` (used only by test suite `csgtexttest.cc`) seem totally redundant now.  I'm thinking both of them can be removed entirely and replace their usage in `csgtexttest.cc` with a simple call to `Tree.getString()`?

Lastly I am looking for input on how I can test that CGALCache is working 100% as expected now that so much internal has changed.

I'm pretty excited about these changes, I think its going to have a really big impact on memory consumption for complex models, and possibly some tangible speed improvement too. (it has already made at least an order of magnitude noticeable speed increase with heavy recursion like testing my L_systems.scad, but I think its likely to see some benefit to more typical use cases as well).

kintel:
Thanks for taking this on. I'll review your code and see if I can come up with some ideas on the remaining topics .

thehans:
I think the root modifier issue is fixed, tests are passing now.

It looks like CGALCache is duplicating something, but I need to debug some more.  I see two inserts for the same exact size.  Seems like one has a group() around it and the other doesn't.  Which seems like some inconsistency about whether or not the root node is output. 

This also raises a point about the csg output which I think can be simplified by not outputing group nodes when they have less than 2 children.  However because its not known ahead of time if the children were also group nodes that were optimized out, then the traversal needs to change or be cached in yet another way.  Like put another map in NodeDumper to give each node a stringstream, then those get combined in postFix.
I think most importantly doing this on id strings could possibly result in more matches, but also I'd personally like to see the same done for plain csg output, though this will require a change to all the expected csg test outputs.

I think I'll spend some time adding more debug logging around nodecache etc. next.

thehans:
OK I looked into the CGAL cache duplicates and that seems to be existing behavior.  See also PR #2218

I made a debug build from before any of my cache-related changes (25806364a27e15ca1765d048418542810e5e11f0)
and tried ran the example code from PR #2218
```
module dstar(r = 30, R = 50) {
difference() {
sphere(r = R);
rotate([0, -30, -45]) translate([R + 0.7*r, 0, 0]) sphere(r = r);
}
}

for( tx = [-60, 60], ty = [-60, 60])
color([0.5 + tx/120, 0.5 + ty/120, 0])
render() translate([tx, ty, 0]) dstar();
```

The baseline outputs 
```
Saved backup file: /home/hans/.local/share/OpenSCAD/backups/unsaved-backup-hXM23737.scad
Rendering Polygon Mesh using CGAL...
CGAL Cache insert: difference(){sphere($fn=0,$fa=12,$fs=2,r (1205776 bytes)
CGAL Cache insert: group(){difference(){sphere($fn=0,$fa=12 (1205776 bytes)
CGAL Cache insert: multmatrix([[1,0,0,-60],[0,1,0,-60],[0,0 (1205776 bytes)
CGAL Cache insert: render(convexity=1){multmatrix([[1,0,0,- (1205776 bytes)
CGAL Cache hit: group(){difference(){sphere($fn=0,$fa=12 (1205776 bytes)
CGAL Cache insert: multmatrix([[1,0,0,-60],[0,1,0,60],[0,0, (1205776 bytes)
CGAL Cache insert: render(convexity=1){multmatrix([[1,0,0,- (1205776 bytes)
CGAL Cache hit: group(){difference(){sphere($fn=0,$fa=12 (1205776 bytes)
CGAL Cache insert: multmatrix([[1,0,0,60],[0,1,0,-60],[0,0, (1205776 bytes)
CGAL Cache insert: render(convexity=1){multmatrix([[1,0,0,6 (1205776 bytes)
CGAL Cache hit: group(){difference(){sphere($fn=0,$fa=12 (1205776 bytes)
CGAL Cache insert: multmatrix([[1,0,0,60],[0,1,0,60],[0,0,1 (1205776 bytes)
CGAL Cache insert: render(convexity=1){multmatrix([[1,0,0,6 (1205776 bytes)
CGAL Cache insert: color([0,0,0,1]){render(convexity=1){mul (1205776 bytes)
CGAL Cache insert: color([0,1,0,1]){render(convexity=1){mul (1205776 bytes)
CGAL Cache insert: color([1,0,0,1]){render(convexity=1){mul (1205776 bytes)
CGAL Cache insert: color([1,1,0,1]){render(convexity=1){mul (1205776 bytes)
CGAL Cache insert: group(){color([0,0,0,1]){render(convexit (4822816 bytes)
Geometries in cache: 4
Geometry cache size in bytes: 248288
CGAL Polyhedrons in cache: 15
CGAL cache size in bytes: 21703680
Total rendering time: 0 hours, 0 minutes, 19 seconds
   Simple:        yes
   Vertices:     2120
   Halfedges:   11192
   Edges:        5596
   Halffacets:   6968
   Facets:       3484
   Volumes:         5
Rendering finished.
```

Its a little more work to piece together the messages after these changes(because I can't access the id string in a debug print directly from CGALCache now), but I manually combined corresponding lines so the id string can be seen
```
CGAL Cache insert: BC4C6AC87182B9B2 (1205776 bytes) "difference(){sphere($fn=0,$fa=12,$fs=2,r ... 1]]){sphere($fn=0,$fa=12,$fs=2,r=30);}}}"
CGAL Cache insert: 6416B5E6D5FA1EE8 (1205776 bytes) "group(){difference(){sphere($fn=0,$fa=12 ... ]]){sphere($fn=0,$fa=12,$fs=2,r=30);}}}}"
CGAL Cache insert: C6975546A9F7CF58 (1205776 bytes) "multmatrix([[1,0,0,-60],[0,1,0,-60],[0,0 ... ]){sphere($fn=0,$fa=12,$fs=2,r=30);}}}}}"
CGAL Cache insert: FFCC8D8A22F4A58B (1205776 bytes) "render(convexity=1){multmatrix([[1,0,0,- ... ){sphere($fn=0,$fa=12,$fs=2,r=30);}}}}}}"
CGAL Cache hit: 6416B5E6D5FA1EE8 (1205776 bytes) "group(){difference(){sphere($fn=0,$fa=12 ... ]]){sphere($fn=0,$fa=12,$fs=2,r=30);}}}}"
CGAL Cache insert: 300848EE6D7CE5D8 (1205776 bytes) "multmatrix([[1,0,0,-60],[0,1,0,60],[0,0, ... ]){sphere($fn=0,$fa=12,$fs=2,r=30);}}}}}"
CGAL Cache insert: 9C0847995E3EEB10 (1205776 bytes) "render(convexity=1){multmatrix([[1,0,0,- ... ){sphere($fn=0,$fa=12,$fs=2,r=30);}}}}}}"
CGAL Cache hit: 6416B5E6D5FA1EE8 (1205776 bytes) "group(){difference(){sphere($fn=0,$fa=12 ... ]]){sphere($fn=0,$fa=12,$fs=2,r=30);}}}}"
CGAL Cache insert: F411E9FA3DA60482 (1205776 bytes) "multmatrix([[1,0,0,60],[0,1,0,-60],[0,0, ... ]){sphere($fn=0,$fa=12,$fs=2,r=30);}}}}}"
CGAL Cache insert: C165ED48A0604B9A (1205776 bytes) "render(convexity=1){multmatrix([[1,0,0,6 ... ){sphere($fn=0,$fa=12,$fs=2,r=30);}}}}}}"
CGAL Cache hit: 6416B5E6D5FA1EE8 (1205776 bytes) "group(){difference(){sphere($fn=0,$fa=12 ... ]]){sphere($fn=0,$fa=12,$fs=2,r=30);}}}}"
CGAL Cache insert: 7D7B081A1A5890A6 (1205776 bytes) "multmatrix([[1,0,0,60],[0,1,0,60],[0,0,1 ... ]){sphere($fn=0,$fa=12,$fs=2,r=30);}}}}}"
CGAL Cache insert: 9387C156432EDB89 (1205776 bytes) "render(convexity=1){multmatrix([[1,0,0,6 ... ){sphere($fn=0,$fa=12,$fs=2,r=30);}}}}}}"
CGAL Cache insert: 5DDCF2E8D11695A4 (1205776 bytes) "color([0,0,0,1]){render(convexity=1){mul ... {sphere($fn=0,$fa=12,$fs=2,r=30);}}}}}}}"
CGAL Cache insert: 3FAF76D6A72AEB13 (1205776 bytes) "color([0,1,0,1]){render(convexity=1){mul ... {sphere($fn=0,$fa=12,$fs=2,r=30);}}}}}}}"
CGAL Cache insert: DCC07388256E4142 (1205776 bytes) "color([1,0,0,1]){render(convexity=1){mul ... {sphere($fn=0,$fa=12,$fs=2,r=30);}}}}}}}"
CGAL Cache insert: 1DFB3CB29065F043 (1205776 bytes) "color([1,1,0,1]){render(convexity=1){mul ... {sphere($fn=0,$fa=12,$fs=2,r=30);}}}}}}}"
CGAL Cache insert: 25C9B157D57CD0A7 (4822816 bytes) "group(){color([0,0,0,1]){render(convexit ... sphere($fn=0,$fa=12,$fs=2,r=30);}}}}}}}}"
```

The full logs from each console window are also attached.

[newcache.log](https://github.com/openscad/openscad/files/1746822/newcache.log)
[baseline_cache.log](https://github.com/openscad/openscad/files/1746823/baseline_cache.log)

src/nodecache.h
```diff
@@ -1,52 +1,67 @@
...
-	}
+    NodeCache() { }
+    virtual ~NodeCache() { cache.clear(); }
+
+    bool contains(const AbstractNode &node) const {
+        auto result = cache.find(node.index()); 
+        return result != cache.end() && result->second.second >= 0L;
+    }
+
+    const std::string operator[](const AbstractNode &node) const {
```
kintel: The const qualifier for the return value doesn't add anything as it's passed by value anyway.

src/CGALCache.h
```diff
@@ -12,9 +12,9 @@ class CGALCache
 
 	static CGALCache *instance() { if (!inst) inst = new CGALCache; return inst; }
 
-	bool contains(const std::string &id) const { return this->cache.contains(id); }
-	shared_ptr<const class CGAL_Nef_polyhedron> get(const std::string &id) const;
-	bool insert(const std::string &id, const shared_ptr<const CGAL_Nef_polyhedron> &N);
+	bool contains(const size_t id) const { return this->cache.contains(id); }
```
kintel: I find this slightly scary. Hashed containers are usually designed such that they gracefully handle hash collisions. In this case, it looks like we'd potentially do significant damage in case of a hash collision.

src/Tree.cc
```diff
@@ -4,31 +4,36 @@
...
 	assert(this->root_node);
-	if (!this->nodecache.contains(node)) {
-		this->nodecache.clear();
-		this->nodeidcache.clear();
-		NodeDumper dumper(this->nodecache, indent, false);
+	bool idString = false;
+	bool idPrefix = false;
+
+	// Retrieve a nodecache given a tuple of NodeDumper constructor options
+	NodeCache &nodecache = this->nodecachemap[std::make_tuple(indent,idString,idPrefix)];
```
kintel: I _think_ you should be able to just do `{indent,idString,idPrefix}`

src/nodedumper.cc
```diff
@@ -17,80 +18,88 @@
...
+			
+			const boost::regex re("[^\\s\\\"]+|\\\"(?:[^\\\"\\\\]|\\\\.)*\\\"");
+			std::stringstream namestream;
+			namestream << node;
+			std::string name = namestream.str();
+			boost::sregex_token_iterator it(name.begin(), name.end(), re, 0);
+			std::copy(it, boost::sregex_token_iterator(), std::ostream_iterator<std::string>(dump));
+		
+			if (node.getChildren().size() > 0) 
+				dump << "{";
```
kintel: Prefer enclosing curly braces if a statement spans multiple lines

src/nodecache.h
```diff
@@ -1,52 +1,67 @@
...
+    }
+
+    void insert_end(const size_t nodeidx, const long endindex) {
+        auto result = cache.find(nodeidx);
+        assert(result != cache.end() && "end index inserted before start");
+        auto indexpair = result->second;
+        assert(indexpair.second == -1L && "end index inserted twice");
+        cache[nodeidx] = std::make_pair(indexpair.first, endindex);
+#if DEBUG
+        PRINTB("NodeCache insert {%i,[%d:%d]}", nodeidx % indexpair.first % endindex );
```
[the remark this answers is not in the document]
thehans: Good idea, this can get very spammy.

src/CGALCache.h
thehans: Hmm, good point, it looks like std::hash I'm using in `Tree.cc` is not particularly good about collisions https://stackoverflow.com/questions/7968674/unexpected-collision-with-stdhash
especially because according to that link its not required to hash the entire string.
Do you think a different hash algorithm exists that is robust enough to avoid collision on its own or it should be handled explicitly in some other way.   I'm not sure the best approach

src/Tree.cc
thehans: compiler doesn't like it
```C++
/home/hans/projects/openscad/src/Tree.cc: In member function ‘const string Tree::getString(const AbstractNode&, const string&) const’:
/home/hans/projects/openscad/src/Tree.cc:28:70: error: converting to ‘std::map<std::tuple<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool, bool>, NodeCache>::key_type {aka std::tuple<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool, bool>}’ from initializer list would use explicit constructor ‘constexpr std::tuple< <template-parameter-1-1> >::tuple(_UElements&& ...) [with _UElements = {const std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&, bool&, bool&}; <template-parameter-2-2> = void; _Elements = {std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool, bool}]’
  NodeCache &nodecache = this->nodecachemap[{indent,idString,idPrefix}];
```

src/nodecache.h
```diff
@@ -1,52 +1,67 @@
...
+
+    void clear() {
+        cache.clear();
+        root_string = "";
+    }
 
 private:
-  std::vector<shared_ptr<std::string>> cache;
-	std::string nullvalue;
+    std::map<size_t, std::pair<long,long>> cache;
```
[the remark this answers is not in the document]
thehans: I think the only case where we don't have entries for all nodes is when root modifier is used to process a sub Tree.  I'll change to unordered_map

src/nodecache.h
```diff
@@ -1,52 +1,67 @@
...
+        auto result = cache.find(node.index());
+        assert(result != cache.end() && "nodecache miss");
+        auto start = result->second.first;
+        auto end = result->second.second;
+        return root_string.substr(start, end-start);
+    }
+
+    void insert_start(const size_t nodeidx, const long startindex) {
+        assert(cache.find(nodeidx) == cache.end() && "start index inserted twice");
+        cache[nodeidx] = std::make_pair(startindex, -1L);
```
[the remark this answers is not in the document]
thehans: Couldn't get this change to compile.

src/CGALCache.h
thehans: I guess CGALCache is not going to be storing anywhere near **all** the node keys, so maybe its best that I just revert to using getIdString as before.

thehans: The CSGTextCache on the other hand does keep all node, I think.  So if I went back to strings, I still feel that should be rewritten.

thehans:
I've been thinking more about the idea of optimizing away group nodes with <= 1 non-empty children.  And I'm realizing that in order for it to be possible I need to bring back the whole `visitedchildren` concept and related code that I tore out of NodeDumper.  Doh!

src/nodecache.h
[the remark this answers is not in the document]
kintel: Right, `emplace()` is templated so it won't be able to do argument-dependent lookup.
FYI: Any of these would work:
```
        this->cache.insert({nodeidx, {startindex, -1L}});
        this->cache[nodeidx] = {startindex, -1L};
```

src/Tree.cc
kintel: That's curious.. Don't think we need to dig deeper for now though :)

thehans:
I reverted the code that implemented the hash of id strings, so we don't have to worry about collisions.  I think this PR is in a good position to be merged now.

I still want to try getting something to work regarding removing unnecessary group nodes.  But maybe that's best handled in another PR.  Merging this now should fix #2317 

I'll leave it up to you if you would rather merge #2318 (reverting #2303) or this PR as an immediate solution to #2307.

kintel:
Let's aim to merge this instead of reverting. It's acceptable to break master for a short amount of time.
If you have any ides for how we could implement automated tests for this, I'd be interested.

src/CGALCache.h
kintel: Just to follow up on this even though it seems fixed now: Using hashes as keys is always a bad idea. The key needs to be unique. This is what hash maps do; store all items with the same hash in the same bucket and reverting to full key comparison only for hash collisions.
If you're worried about the time taken to compute the hash from a key, the solution is probably to lazily compute the hash and cache it. ...in case the same object is processed many times during a traversal.

src/nodecache.h
```diff
@@ -13,55 +13,50 @@
...
-    virtual ~NodeCache() { cache.clear(); }
+    virtual ~NodeCache() { }
 
     bool contains(const AbstractNode &node) const {
-        auto result = cache.find(node.index()); 
-        return result != cache.end() && result->second.second >= 0L;
+        auto result = this->cache.find(node.index()); 
+        return result != this->cache.end() && 
+            result->second.second >= 0L && 
+            this->rootString.size() >= result->second.second;
```
kintel: What does that last check do? Will we ever generate indices that are beyond the size of the full string?

src/nodecache.h
```diff
@@ -13,55 +13,50 @@
...
 
     void insert_end(const size_t nodeidx, const long endindex) {
-        auto result = cache.find(nodeidx);
-        assert(result != cache.end() && "end index inserted before start");
-        auto indexpair = result->second;
+        // throws std::out_of_range on miss
+        auto indexpair = this->cache.at(nodeidx); 
         assert(indexpair.second == -1L && "end index inserted twice");
-        cache[nodeidx] = std::make_pair(indexpair.first, endindex);
+        this->cache[nodeidx] = std::make_pair(indexpair.first, endindex);
```
kintel: Since you opted for `emplace()` in the `insert_start()` method, we should probably `emplace()` here as well

src/nodecache.h
thehans: Its mainly an added protection in case this method is called before NodeDumper finishes traversing, because the rootString value isn't set until the very end.
An empty string check would also do roughly the same thing.

src/nodecache.h
thehans: As I understand it, emplace is for adding a key where there isn't one yet.  

From http://en.cppreference.com/w/cpp/container/map/emplace
>Inserts a new element into the container constructed in-place with the given args if there is no element with the key in the container.

When we insert the end index, we are updating the value(second part of the pair) for an existing key, so its a little different.

src/nodecache.h
kintel: Good point

src/nodecache.h
kintel: You’re right

kintel:
@thehans Out of curiosity, did you identify any scenarios where you could benefit from optimizing away unnecessary group nodes?

thehans:
If the code is calling to modules, with `if`/`else`, `for`,`echo` or basically any of the control structures in control.cc these all get group nodes that end up in the cache id.

Here is a maybe a bit contrived example, but the `echo` statement within an `if` will change the node id key and mess with cache even though it has nothing to do with geometry.  
```OpenSCAD
$fn=60; // moderate complexity to test run time

breakCache = true;

// breakCache = false; // F6 render takes 9s
// *** Flush caches in between tests ***
// breakCache = true;  // F6 render takes 15s


module moon(say=false) {
  // expensive difference that would ideally be cached
  difference() {
    sphere();
    translate([0,-1.3,1.3]) {
      sphere();
      // troublesome extra group nodes that don't affect geometry, but change cache id
      if (say) echo("That's no moon!");
    }
  }
}

moon();
translate([4,0,0]) moon(breakCache);

```

Also the [L_systems.scad](https://gist.github.com/thehans/a1494db8046a58832e2ebb10a5908a66) script I was using to test this, even at iteration of 1: `dragon_curve(n=1, tree=true);` results in an overwhelming number of group nodes.  
The actual geometry of the result is just **2** `square()`s, but CGAL shows:
Geometries in cache: **39**
Geometry cache size in bytes: 6016.

And the CSG looks like this:
```OpenSCAD
group() {
  group() {
    group() {
      group();
      group() {
        group() {
          group() {
            group() {
              group();
            }
            group() {
              group() {
                group() {
                  group() {
                    group() {
                      group() {
                        group() {
                          group() {
                            group() {
                              group() {
                                multmatrix([[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]) {
                                  group();
                                  multmatrix([[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]) {
                                    multmatrix([[1, 0, 0, 0], [0, 1, 0, -0.2], [0, 0, 1, 0], [0, 0, 0, 1]]) {
                                      square(size = [1, 0.4], center = false);
                                    }
                                  }
                                }
                              }
                            }
                          }
                          group() {
                            group() {
                              group() {
                                group() {
                                  group() {
                                    group() {
                                      group() {
                                        group() {
                                          group() {
                                            group() {
                                              group() {
                                                group() {
                                                  group() {
                                                    group() {
                                                      multmatrix([[1, 0, 0, 1], [0, 1, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]) {
                                                        group();
                                                        multmatrix([[2.22045e-16, 1, 0, 0], [-1, 2.22045e-16, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]) {
                                                          multmatrix([[1, 0, 0, 0], [0, 1, 0, -0.2], [0, 0, 1, 0], [0, 0, 0, 1]]) {
                                                            square(size = [1, 0.4], center = false);
                                                          }
                                                        }
                                                      }
                                                    }
                                                  }
                                                }
                                                group() {
                                                  group() {
                                                    group();
                                                  }
                                                }
                                              }
                                            }
                                          }
                                        }
                                        group() {
                                          group() {
                                            group();
                                          }
                                        }
                                      }
                                    }
                                  }
                                }
                                group() {
                                  group() {
                                    group();
                                  }
                                }
                              }
                            }
                          }
                        }
                      }
                    }
                  }
                  group() {
                    group() {
                      group();
                    }
                  }
                }
              }
            }
          }
        }
      }
      group();
    }
  }
}

```

MichaelAtOz:
I haven't followed the details, but has anyone put on their multi-threaded hat in regard to this?

kintel:
@MichaelAtOz This is all independent of the multi-threading stuff. This is all about efficiently generating keys for the various caches, allowing us to identify separate pieces of a design that results in equivalent geometries.

kintel:
@thehans Thanks. Glancing at your examples, my intuition is that the issue is that empty _sibling_ group nodes, or trees of empty group nodes could be optimized away.

MichaelAtOz:
Yes I know, just wanted to make sure it wasn't going to increase contention when multi-threaded.

kintel:
All good - thanks for keeping an eye on such discussions!

kintel:
@thehans ..so if you were to implement special handling of empty group nodes, would that improve this situation? e.g.
```
Response NodeDumper::visit(State &state, const GroupNode &node)
{
         if (state.isPrefix() && node.getChildren().size() == 0) return Response::PruneTraversal;
         else return visit(state, static_cast<const AbstractNode&>(node));
}
```

thehans:
Ooh, yes I think something like that could work very well.  It still needs to handle inserting the start and end indices.

Also I would change it to also return ContinueTraversal if `getChildren.size() == 1`
I think for a leaf node, Continue and Prune would be equivalent.

Here is a more simplified example of cache duplication, every group that you wrap around the square creates another GeometryCache entry.  So this would have 3 entries:
```
group() {
  group() { 
    square();
  }
}
```
That's what the case of children.size()==1  is intended to handle.

thehans:
Here's my take on it that looks seems to work, and helps a bit with cache duplication
```C++
Response NodeDumper::visit(State &state, const GroupNode &node)
{
	if (node.getChildren().size() <= 1) {
		if (state.isPrefix()) {
			this->cache.insertStart(node.index(), this->dumpstream.tellp());
		} else if (state.isPostfix()) {
			this->cache.insertEnd(node.index(), this->dumpstream.tellp());
		}
		return Response::ContinueTraversal;
	} 
	return visit(state, static_cast<const AbstractNode&>(node));
}
```

This handles the simple example above, wrapping group nodes around a single element, so the square only takes one GeometryCache entry.

However if you complicate it a little bit, you can cause it to miss again
```
group() {
  group() {
    square();
    group();
  }
}
```
The second group has two children, but not after the empty `group();` child is omitted from the dump.  But it doesn't know that during `state.isPrefix()`

Comparing this with the other above iteration 1 example from L_system.scad.  The "2 squares" now uses 14 cache entries instead of 39.  So still not flawless, but a definite improvement.

Here's the csg output from the iteration 1 with this change
```
group() {
  group() {
    group() {
      group() {
        multmatrix([[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]) {
          multmatrix([[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]) {
            multmatrix([[1, 0, 0, 0], [0, 1, 0, -0.2], [0, 0, 1, 0], [0, 0, 0, 1]]) {
              square(size = [1, 0.4], center = false);
            }
          }
        }
        group() {
          group() {
            group() {
              multmatrix([[1, 0, 0, 1], [0, 1, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]) {
                multmatrix([[2.22045e-16, 1, 0, 0], [-1, 2.22045e-16, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]) {
                  multmatrix([[1, 0, 0, 0], [0, 1, 0, -0.2], [0, 0, 1, 0], [0, 0, 0, 1]]) {
                    square(size = [1, 0.4], center = false);
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

Another thing I'm noticing from that CSG is some extra identity matrix multiplications.  I think those are from some `rotate` calls in scad that depend on an angle variable which is often 0.

It might be worthwhile some kind of similar override for TransformNode's that skips past itself if it detects the identity matrix input.

thehans:
@kintel Another approach that might work well in combination with the above is to be more selective about which node are even *eligible* to be cached by GeometryEvaluator.  So maybe GeometryEvaluator would say: this is a group with a single child, its cache would be the same as its child's cache, no need to add another entry?

nophead:
The redundant group nodes make the CSG output almost unreadable. It would be good to get rid of them or not create them in the first place if that is possible.

kintel:
This gets messy pretty fast. Perhaps a cleaner way of thinking about this could be to process the entire node tree and prune/rewrite no-op nodes. That's a separate discussion though.

PR #2319, opened 2018-02-25
Customizer UI Improvment

MichaelPFrey:
* only enable the customizer UI when ready 
* descriptionLoD "Description Only" (hiding the variable name when a description is set) 
* set a flag when we have an unreadable json file 
* ask before overwriting unreadable JSON-Files

see also #2316

### DIFF
--- a/src/parameter/ParameterWidget.cc
+++ b/src/parameter/ParameterWidget.cc
@@ -406,7 +406,7 @@ void ParameterWidget::writeParameterSets()
 		bool ok = true;
 		QMessageBox msgBox;
 		msgBox.setText(_("Saving presets"));
-		msgBox.setInformativeText( _("A json file was found, but was unreadble. Do you want to overwrite it?"));
+		msgBox.setInformativeText(QString(_("%1 was found, but was unreadble. Do you want to overwrite %1?")).arg(QString::fromStdString(jsonFile)));
 		msgBox.setStandardButtons(QMessageBox::Save | QMessageBox::Cancel);
 		msgBox.setDefaultButton(QMessageBox::Cancel);
 		int ret = msgBox.exec();
--- a/src/parameter/ParameterWidget.cc
+++ b/src/parameter/ParameterWidget.cc
@@ -107,6 +107,7 @@ void ParameterWidget::onSetSaveButton()
 }
 
 void ParameterWidget::setFile(QString scadFile){
+	this->unreadableFileExists = false; // we can not know it, but we do not want to report a problem when there is none
 	boost::filesystem::path p = scadFile.toStdString();
 	this->jsonFile = p.replace_extension(".json").string();
 }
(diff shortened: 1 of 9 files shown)


src/parameter/ParameterWidget.cc
```diff
@@ -388,6 +396,24 @@ void ParameterWidget::updateParameterSet(std::string setName)
...
 	}
 }
+
+void ParameterWidget::writeParameterSets()
+{
+	if(this->unreadableFileExists){
+		bool ok = true;
+		QMessageBox msgBox;
+		msgBox.setText(_("Saving presets"));
+		msgBox.setInformativeText(QString(_("%1 was found, but was unreadble. Do you want to overwrite %1?")).arg(QString::fromStdString(jsonFile)));
```
kintel: `this->jsonFile`

src/parameter/ParameterWidget.cc
```diff
@@ -388,6 +396,24 @@ void ParameterWidget::updateParameterSet(std::string setName)
...
 		}
-		writeParameterSet(this->jsonFile);
+		writeParameterSets();
 	}
 }
+
+void ParameterWidget::writeParameterSets()
+{
+	if(this->unreadableFileExists){
+		bool ok = true;
```
kintel: Looks like `ok` is never used

src/parameter/ParameterWidget.cc
```diff
@@ -388,6 +396,24 @@ void ParameterWidget::updateParameterSet(std::string setName)
...
+		QMessageBox msgBox;
+		msgBox.setText(_("Saving presets"));
+		msgBox.setInformativeText(QString(_("%1 was found, but was unreadble. Do you want to overwrite %1?")).arg(QString::fromStdString(jsonFile)));
+		msgBox.setStandardButtons(QMessageBox::Save | QMessageBox::Cancel);
+		msgBox.setDefaultButton(QMessageBox::Cancel);
+		int ret = msgBox.exec();
+
+		if (ret == QMessageBox::Cancel) {
+			return;
+		}
```
kintel: You could probably make the above a tiny bit cleaner by doing:
`if (msgBox.exec() == QMessageBox::Cancel) return;`

src/parameter/parameterslider.cpp
```diff
@@ -13,14 +13,21 @@ ParameterSlider::ParameterSlider(ParameterObject *parameterobject, int showDescr
 	connect(slider, SIGNAL(sliderReleased()), this, SLOT(onReleased()));
 	connect(slider, SIGNAL(valueChanged(int)), this, SLOT(onSliderChanged(int)));
 	connect(doubleSpinBox, SIGNAL(valueChanged(double)), this, SLOT(onSpinBoxChanged(double)));
-	if (showDescription == 0) {
+	if (showDescription == 0 || showDescription == 3) {
```
kintel: This may be a good refactoring for later: Use an enum instead of the magic `showDescription` integer

MichaelPFrey: entered into #2316

MichaelPFrey:
thank you for the review - changed as suggested.

PR #2321, opened 2018-02-26
Feature addition: Option to raise the window upon auto-reload

jnweiger:
When auto reload happens, the main window is raise()d to the front.
This is very useful with generating the .scad file from another application that
would cover most of the screen space. I use openscad together with inkscape
and inkscape likes to be in front most of the time...
An option is added to the Preferences advanced page to disable that new feature. 
I suggest to enable by default.

Also added scripts/docker-build.sh -- a quick wrapper to build on linux
bringing all prerequisites into a docker container to not pollute the build host.
Read and adjust the script before use, some things are hardcoded there.
It works with a git checkout outside the container, so that all object files and
the binary are easily accessible.
It shortcuts to the next needed step, but is quite slow doing so.

### DIFF
--- a/locale/de.po
+++ b/locale/de.po
@@ -1051,6 +1051,10 @@ msgstr "Startbildschirm anzeigen"
 msgid "Enable user interface localization (requires restart of OpenSCAD)"
 msgstr "Lokalisierung der GUI einschalten (benötigt Neustart von OpenSCAD)"
 
+#: objects/ui_Preferences.h:1184
+msgid "Bring window to front after automatic reload"
+msgstr "Fenster nach vorne bringen nach automatischem Neuladen"
+
 #: objects/ui_ProgressWidget.h:72
 msgid "%v / %m"
 msgstr "%v / %m"
--- a/locale/openscad.pot
+++ b/locale/openscad.pot
@@ -1028,6 +1028,10 @@ msgstr ""
 msgid "Enable user interface localization (requires restart of OpenSCAD)"
 msgstr ""
 
+#: objects/ui_Preferences.h:1184
+msgid "Bring window to front after automatic reload"
+msgstr ""
+
 #: objects/ui_ProgressWidget.h:72
 msgid "%v / %m"
 msgstr ""
--- a/scripts/docker-build.sh
+++ b/scripts/docker-build.sh
@@ -1,0 +1,46 @@
+#! /bin/bash
+#
+# openscad-docker-build.sh -- a simple script to compile openscad using a docker container.
+#
+# (C) 2018 jnweiger@gmail.com
+# Distribute under GPL-2.0 or ask.
+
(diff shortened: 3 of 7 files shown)


t-paul:
I can see how it could be useful in some workflows. I'm not sure we should change the default behaviour though. It's good to have that option, but forcing the new behavior onto all people feel unnecessary.

Is `raise()` the correct call? The documentation sounds like `activateWindow()` might be needed?

jnweiger:
Thanks @kintel and @t-paul for the review!
Feel free to chose default off. Fine with me. I had chosen default  on, to make the reload more noticeable. Sometimes the change is so minimal that one has doubt if the reload triggered. Seeing the window popping forward is a nice-to-have additional indicator (but not more).

I am using only `raise()`, so that the focus is **not** 'stolen'. If we call `activateWindow()` we also tell the WindowManager to transfer mouse and keyboard focus to the openSCAD window. That is more intrusive might be considered bad UX. (At least I would prefer it to switch focus myself ...)

Other code example in openSCAD https://github.com/openscad/openscad/blob/master/src/mainwin.cc#L271:
When we show() the properties window, we call all three: show(), raise() and activateWindow()  to make sure that focus is transfered from the main openSCAD window to the Properties window. Here my feeling is that it is perfectly right to use `activateWindow()` - but this topic is still a bit muddy to me.

MichaelAtOz:
> transfer mouse and keyboard focus to the openSCAD window. That is more intrusive might be considered bad UX.  (At least I would prefer it to switch focus myself ...)

+1, that really annoys me.... 
When using an external editor & auto-reload, you don't want focus stolen from the editor.

kintel:
@jnweiger Sounds like `raise()` is the correct solution. In terms of the default flag, It's easier if you change it in this PR, then we can simply merge it directly.

jnweiger:
Changed the PR as requested.

jnweiger:
Thanks!

jnweiger:
Is there a release planned in the near future? I wonder if we still have a chance to get into Ubuntu 18.04... Currently Ubuntu users are stuck with 2015.3 even on lates Ubuntu 17.10 ...

MichaelPFrey:
@jnweiger you might want to take a look on https://github.com/openscad/openscad/projects/1 and also note, how long features tend to stick with the experimental status.

Things move slowly around here.

kintel:
@jnweiger We're continuously trying to pin down a reasonable feature set for the next release, but we're only a few people doing spare-time coding. If there is anything you'd like to contribute, Michael's list is good. Moving experimental features into stable ones would be a great place to start.

jnweiger:
Yeah, I saw the project board already. Do not worry about ubuntu 17.10, the debian pull deadline is already over. Same over here, I have only two hands and so many projects to juggle :-)

I also discovered the openscad-nightly packages in https://build.opensuse.org/package/show/home:t-paul/OpenSCAD -- all platforms covered. Way cool! Keep up the good work. I prefer good over fast anyway!

PR #2322, opened 2018-02-26
Add the filename into the AST Tree

### LINKED ISSUE #1840: Customizer: Don't scan for parameters in included files

MichaelPFrey:
originally _Don't scan for parameters in included files without changing YYLTYPE [WIP]_
#2044 rebased to current master and changed, so that we do not need to change YYLTYPE.

closes #1840

Work in Progress.


src/AST.h
```diff
@@ -1,32 +1,39 @@
...
 
 
 	static const Location NONE;
-;
 private:
 	int first_line;
 	int first_col;
 	int last_line;
 	int last_col;
+	std::string filename;
```
kintel: It feels a bit intense to store the filename for each AST node. What  you're really using this for is to determine if two assignments belong to the same file.
I've been contemplating ways of managing assignments in a better way, which could help towards this. My idea thus far is to parse included files into separate `FileModule` instances, and resolve the actual assignments during evaluation. This would also make OpenSCAD's special assignment rules more explicit and isolated in our source code. See #1743 

The above PR may take time to push through though, so we may want to implement your solution in the interim.

MichaelPFrey: I could have used a flag called "included" to get the same result for just now.

I picked the filename, as it seems more future prove in regards to general error handling.
(e.g. in which include did the assertion fail?)

Also, that specific line is from #2044, where the argument was about failing tests, not the general idea.

I see two ways:
* we can close this pull request and I create a new one based  on this with a "included flag"
or
*  I change the std::string to an int, storing the association from int to string in a look up table

I have personal attachment to the string solution specifically, if you have bigger plans for the long run, I do not want to get in your way. Well, the more I talk: the included flag sounds sensible.

kintel: Keep in mind that we may have to handle reassignment in a smart way. Haven't thought about that yet..
Nevertheless, I think the string solution is still the better one as it's really explicit and should be easy to refactor later.

t-paul: Could we handle that via a shared pointer created in the parser? I guess having the filename around would also help with error messages.

MichaelPFrey: shared pointer sounds perfect. 32/64 bits more per AST node is cheap for the benefits in error reporting we get in knowing the file name.

But why specially created in the parser?
For e.g. FileModule could (and should) normalize the paths and could then provide use with a pointer.

I will look into it.

MichaelPFrey:
ToDo list:
   - [  ] FileModule:
      - [  ] setFilename should in my opinion get the logic to create absolute path (currently, getFilename() gives different results depending on when it is called)
      - [  ]  remove getFullpath() (when setFilename handles getting the absolute path, this methode becomes obsolet)
   - [  ] ast
      - [  ]  std::string filename; -> shared pointer to save memory
     - [  ] #include<string> accidentally included multiple times 
   - [  ] ast
      - [  ]  clean up the messy path handling (see FileModule)
   - [  ]  create a list of issues (e.g. related to error reporting), where this (knowing the file name of an ast-node) could be key to resolve the issue

src/AST.h
kintel: @MichaelPFrey Going down that route, would a pointer to the owning `FileModule` be even better, to make the connection explicit? Would probably have to be a `weak_ptr` or smth. to avoid circular references.

MichaelPFrey:
please carefully review src/comment_parser.y
It looks like a novice mistake, but I do not know if my patch contains a novice mistake or the current code.
Fact is, that I had memory management issue, that I could resolve with this patch.

src/lexer.l
```diff
@@ -298,7 +298,7 @@ void includefile()
   std::string fullname = fullpath.generic_string();
 
   filepath.clear();
-  filename_stack.push_back(fullpath);
+  filename_stack.push_back(std::make_shared<fs::path>(fs::path(fullpath.generic_string())));
```
kintel: Can't you just pass fullpath directly? `std::make_shared<fs::path>(fullpath)`

src/AST.h
```diff
@@ -1,20 +1,22 @@
...
 
 class Location {
 
 public:
 	Location(int firstLine, int firstCol, int lastLine, int lastCol,
-			const std::string &filename)
+			std::shared_ptr<fs::path> path)
 		: first_line(firstLine), first_col(firstCol), last_line(lastLine),
-		last_col(lastCol), filename(filename) {
+		last_col(lastCol), path(path) {
```
kintel: Tiny optimization: `path(std::move(path))` - that saves us one extra reference count inc/dec

src/parser.y
```diff
@@ -635,7 +635,7 @@ bool parse(FileModule *&module, const char *text, const std::string &filename, i
   lexerin = NULL;
   parser_error_pos = -1;
   parser_input_buffer = text;
-  parser_sourcefile = path;
+  parser_sourcefile = std::make_shared<fs::path>(fs::absolute(fs::path(filename)));
```
kintel: We already have a local variable `path`, which could be passed directly to `make_shared()`

kintel:
It looks like the original code has a memory leak every time the parser fails, which is fixed in your code. Not sure if that was what you were hinting towards..

kintel:
Note: I approved this, but I'm not sure if you wanted to do more work before merging..

MichaelPFrey:
Thank you for your review.
I will follow your suggestions and then test.
And then wait a few days, so that I can read the changes with a fresh mind.

Some of the code is from hunting down Memory errors, where I presumed that my code was at fault, not the comment parser. (now it is obvious: the more shared pointer we use, the more likely it becomes that we discover other issues with memory management)

The crazy thing is always, when the release build crashes, but the debug does not, so I kind of wanted to use travis to test for other hidden memory issues.

src/lexer.l
MichaelPFrey: yes we can - I just got paranoid because I got memory access errors, but those where caused by the comment parser that is also using shared pointers.

src/AST.h
MichaelPFrey: Interesting idea - yeah that makes sense.
The constructor is its own context with no other code accessing path.
The caller of location already has to handle the shared pointer correctly and the location class as well.

kintel: FYI: Effective Modern C++ has a discussion on this topic:
“Item 41: Consider pass by value for copyable parameters that are cheap to move and always copied.”

MichaelPFrey:
feel free to merge.

Using the file names for other purposes (e.g. error messages for assertions) should be its own pull request.

Given that this is a pretty "clean" as in a minimal change implementation I can not really add or remove code.

The crash I saw while developing was presumably from a compiler hickup (presumable to related to Location:None), that I resolved by changing src/comment_parser.y and thus forcing openscad to recompile the related files. So the change on src/parser.y is unnecessary (using the current file or mine does not make a notable difference during ctest and normal UI operation - both pass automated and manual testing).

PR #2324, opened 2018-03-01
Translation of customizer in general and german specifically

MichaelPFrey:


scripts/generate-potfiles.sh
```diff
@@ -20,3 +20,21 @@ for src in src/*.h src/*.cc src/*.cpp src/*.mm
 do
 	echo $src
 done
+
+for ui in src/parameter/*.ui
```
kintel: Instead of copying this block, could you rather change the original loop to `for ui in $(find src -name "*.ui")` ?

MichaelPFrey: It is not as simple as that, please also note line 8 and line 26:
```
UI="${ui#src/}"
UI="${ui#src/parameter/}"
```
looks like I am tomorrow reading up on bash scripting. There must be a clean solution. (which in turn would help when the inputdriver is merged into main as it would automatically pick up the new files)

kintel: Right.. Perhaps we could just locate `ui_*.h` in the `objects` folder rather than trying to create the filename based on `.ui` files.

PR #2327, opened 2018-03-03
Cleanup and optimatzion

MichaelPFrey:
optimizing getting the lower case file extension during use registration  -> we should not need to create a copy

optimizing: we only need one EmptyStringLiteral as we can share the pointer to it.
I use a static variable in the method as I do not want to elevate the deceleration to the class level.

initializing static variable  in parametervirtualwidget.cpp

### DIFF
--- a/src/comment.cpp
+++ b/src/comment.cpp
@@ -259,6 +259,8 @@ static GroupList collectGroups(const std::string &fulltext)
 */
 void CommentParser::collectParameters(const char *fulltext, FileModule *root_module)
 {
+	static auto EmptyStringLiteral(std::make_shared<Literal>(ValuePtr(std::string(""))));
+
 	// Get all groups of parameters
 	GroupList groupList = collectGroups(std::string(fulltext));
 	int parseTill=getLineToStop(fulltext);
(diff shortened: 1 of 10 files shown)


MichaelPFrey:
Travis still unhappy - well - I will be giving it a try to tomorrow again. I think I try moving the EmptyStringLiteral into an other scope next.

kintel:
It feels like the Travis issue is unrelated; it only fail on Mac, and message look a bit like this - it looks like there was a Python2 to Python3 upgrade happening somewhere:
```
File "/Users/travis/build/openscad/openscad/tests/test_pretty_print.py", line 74
    print 'problem writing to',filename
                             ^
SyntaxError: Missing parentheses in call to 'print'. Did you mean print(print 'problem writing to',filename)?
```

t-paul:
Can you merge in the test script updates from `master`. That should hopefully fix the errors related to python3.

src/comment.cpp
```diff
@@ -278,7 +280,7 @@ void CommentParser::collectParameters(const char *fulltext, FileModule *root_mod
 		// getting the node for parameter annotation
 		shared_ptr<Expression> params = CommentParser::parser(comment.c_str());
 		if (!params) {
-			params = shared_ptr<Expression>(new Literal(ValuePtr(std::string(""))));
+			params = EmptyStringLiteral;
```
kintel: @MichaelPFrey Do you know why we assign an empty string to the annotation in cases where we don't find any annotation comments? I have a patch that removes this and regenerates the tests, but I'm not sure if it would break any UI stuff. See also #2776

PR #2328, opened 2018-03-03
improvent handling of changed customizer preset

MichaelPFrey:
*  if current preset is changed, ask user if he wants to save the current preset 
* indicate the the current preset is changed with an asterisks
see also #2316
* move common code to parametervirtualwidget.cpp
* set the parent for parametewidgets
* showDescription/DescriptionLoD (Level of Detail)
   * unify the naming
   * turn into enum

Todo:
- [x] merge #2330 with master, merge master with this
- [x] reduce scope of the variables
- [x] Testing
- [x] when precisely has _valueChanged_ to be changed? -> covered by testing
- [x] write a short warning, that comboBoxPreset->setCurrentIndex(lastComboboxIndex); can cause a recursion
-  ~~disable the save button when there is nothing to save?~~ -> NO. User might want to save when for e.g. parameters where added in the SCAD file, that are not yet present in the json file

### DIFF
--- a/src/parameter/ParameterWidget.cc
+++ b/src/parameter/ParameterWidget.cc
@@ -473,9 +473,9 @@ void ParameterWidget::updateParameterSet(std::string setName)
 		}
 		setMgr->addParameterSet(setName, iroot);
 		const QString s(QString::fromStdString(setName));
-		if (this->comboBoxPreset->findText(s) == -1) {
+		if (this->comboBoxPreset->findData(s) == -1) {
 			this->comboBoxPreset->addItem(s, QVariant(s));
-			this->comboBoxPreset->setCurrentIndex(this->comboBoxPreset->findText(s));
+			this->comboBoxPreset->setCurrentIndex(this->comboBoxPreset->findData(s));
 		}
 		writeParameterSets();
 	}
--- a/src/parameter/ParameterWidget.cc
+++ b/src/parameter/ParameterWidget.cc
@@ -70,8 +70,8 @@ void ParameterWidget::resetParameter()
 {
 	if(this->valueChanged){
 		QMessageBox msgBox;
-		msgBox.setText(_("changes on current preset not saved"));
-		msgBox.setInformativeText(
+		msgBox.setWindowTitle(_("changes on current preset not saved"));
+		msgBox.setText(
 			QString(_("The current preset %1 contains changes, but is not saved yet. Do you really want to reset the preset and lose your changes?"))
 			.arg(comboBoxPreset->itemData(this->comboBoxPreset->currentIndex()).toString()));
 		msgBox.setStandardButtons(QMessageBox::Ok | QMessageBox::Cancel);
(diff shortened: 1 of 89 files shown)


src/parameter/parametervirtualwidget.h
```diff
@@ -5,6 +5,8 @@
 #include "ui_ParameterEntryWidget.h"
 #include "parameterobject.h"
 
+enum descLoD {ShowDetails,Inline,HideDetails,DescOnly};
```
kintel: Prefer `enum class`, to get some extra type safety.
..and Capitalize the first letter for good measure.

src/parameter/parametervirtualwidget.h
```diff
@@ -13,7 +15,7 @@ class ParameterVirtualWidget : public QWidget, public Ui::ParameterEntryWidget
 	ParameterObject *object;
 
 public:
-	ParameterVirtualWidget(QWidget *parent = nullptr);
+	ParameterVirtualWidget(QWidget *parent,ParameterObject *parameterobject, int descriptionLoD);
```
kintel: This `int` should really be the enum you just defined.

src/parameter/parametervirtualwidget.cpp
```diff
@@ -11,6 +12,24 @@ ParameterVirtualWidget::ParameterVirtualWidget(QWidget *parent) : QWidget(parent
 	policy.setHorizontalStretch(0);
 	policy.setVerticalStretch(0);
 	this->setSizePolicy(policy);
+
+	this->object = parameterobject;
```
kintel: As this is a pure assignment, we can clarify this by instead adding it to the initialization list for this constructor.

src/parameter/parameterset.h
```diff
@@ -25,6 +25,6 @@ class ParameterSet
 	bool readParameterSet(const std::string &filename);
 	void writeParameterSet(const std::string &filename);
 	void applyParameterSet(class FileModule *fileModule, const std::string &setName);
-	const bool isEmpty(){return root.empty();}
+	bool isEmpty() const{return root.empty();};
```
kintel: stray semicolon.
Some spacing wouldn't hurt either; `bool isEmpty() const { return root.empty(); }`

src/parameter/ParameterWidget.cc
```diff
@@ -188,8 +207,33 @@ void ParameterWidget::setComboBoxPresetForSet()
 	}
 }
 
+//set selection
 void ParameterWidget::onSetChanged(int idx)
 {
+	static int lastComboboxIndex(0);
```
kintel: Not sure I understand the use of `static` here.
e.g. what happens if you have two OpenSCAD windows open with two different files and two customizers?

src/parameter/parametervirtualwidget.h
MichaelPFrey: good idea. It results in just one awkward line, where I have to static_cast comboBoxDetails->currentIndex() - but that´s fine as static_cast is precisely the right indicator for whats going on.

src/parameter/ParameterWidget.cc
MichaelPFrey: opps... it feels wrong to have a member variable that is only supposed to be used by one function, but I guess a member variable is the best solution after all.

src/parameter/ParameterWidget.cc
```diff
@@ -422,8 +478,8 @@ void ParameterWidget::writeParameterSets()
 {
 	if(this->unreadableFileExists){
 		QMessageBox msgBox;
-		msgBox.setText(_("Saving presets"));
-		msgBox.setInformativeText(QString(_("%1 was found, but was unreadble. Do you want to overwrite %1?")).arg(QString::fromStdString(this->jsonFile)));
+		msgBox.setWindowTitle(_("Saving presets"));
+		msgBox.setText(QString(_("%1 was found, but was unreadble. Do you want to overwrite %1?")).arg(QString::fromStdString(this->jsonFile)));
```
kintel: Typo: s/unreadble/unreadable/

PR #2330, opened 2018-03-04
Customizer inheritence

MichaelPFrey:
- resolving the multi inheritance in openscad, making the data structure/sharing more explicit.
- this-> to access members
-...

### DIFF
--- a/src/parameter/parameterextractor.cpp
+++ b/src/parameter/parameterextractor.cpp
@@ -37,8 +37,7 @@ void ParameterExtractor::setParameters(const FileModule* module,entry_map_t& ent
     const ValuePtr defaultValue = assignment.expr->evaluate(&ctx);
     if (defaultValue->type() == Value::ValueType::UNDEFINED) continue;
 
-    ParameterObject *entryObject = new ParameterObject();
-    entryObject->setAssignment(&ctx, &assignment, defaultValue);
+    ParameterObject *entryObject = new ParameterObject(&ctx, &assignment, defaultValue);
 
     //check whether object exist or not previously
     if (entries.find(assignment.name) == entries.end()) {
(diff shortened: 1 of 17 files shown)


src/parameter/parameterobject.cpp
```diff
@@ -4,8 +4,30 @@
 #include "modcontext.h"
 #include "annotation.h"
 
-ParameterObject::ParameterObject() : focus(false)
+ParameterObject::ParameterObject(Context *ctx, const Assignment *assignment, const ValuePtr defaultValue) : focus(false)
```
kintel: Since `assignment` isn't allowed to be NULL, I'd prefer a `const &` here; makes it a bit more clear how it's being used.

src/parameter/parameterset.h
```diff
@@ -25,4 +25,6 @@ class ParameterSet
 	bool readParameterSet(const std::string &filename);
 	void writeParameterSet(const std::string &filename);
 	void applyParameterSet(class FileModule *fileModule, const std::string &setName);
+	bool isEmpty(){return root.empty();}
```
kintel: Should be a `const` method

src/parameter/parameterextractor.cpp
```diff
@@ -37,33 +37,27 @@ void ParameterExtractor::setParameters(const FileModule* module, bool rebuildPar
     const ValuePtr defaultValue = assignment.expr->evaluate(&ctx);
     if (defaultValue->type() == Value::ValueType::UNDEFINED) continue;
 
-    ParameterObject *entryObject = new ParameterObject();
-    entryObject->setAssignment(&ctx, &assignment, defaultValue);
+    ParameterObject *entryObject = new ParameterObject(&ctx, &assignment, defaultValue);
```
kintel: Is there a way we could make this e.g. a `shared_ptr`, to get rid of some error prone memory management?

src/parameter/parameterobject.cpp
MichaelPFrey: sounds good - I by the way have now 2nd and 3rd Edition of Bjarne Stroustrups "The C++ Programming Language"-  thank you for the recommendation. I know that there is a 4th edition, but the basics like this are still the same.

src/parameter/parameterextractor.cpp
MichaelPFrey: good idea - that would resolve both memory leaks (caused by not deleting) and race conditions (caused by deleting in one UI event, while an other event is waiting to be processed).

added to #2316 as this is not straight forward.

PR #2333, opened 2018-03-07
Update link downloads & documentation (proposal)

kant:
Link to image via coverity needs an update (https://scan.coverity.com/projects/2510). Plus minor fixes (proposals)


kintel:
Not sure if coverity is alive any longer. @t-paul ?

Issue #2185, opened 2017-11-07
customizer mouse scrolling behavior


MichaelPFrey:
The customizer is using a scrollarea, meaning that when a lot of parameters are on display (or vertical space is at a premium), you have to scroll trough list.

Given that the spinboxes (incl. vectors) are sensitive to scrolling as well, there is little space to move.
With a slider or combox on screen, there is only the scrollbar left, where the user can scroll trough the entries.

When a slider is scrolled to maximum, the mouse scroll now affects the scroll area by scrolling up.
When a slider is scrolled to minimum, the mouse scroll now affects the scroll area by scrolling down.
In eithercase, the scroll area should remain in position.

Any ideas on how to solve that?

this issue originates from #1845

MichaelPFrey:
Some of my ideas:
  * fine tuning
    * fully disable the scrolling on the slider (-> the use can still scroll in the spin box)
    * fully disable scrolling on the comboxes 
    * right align the vector input
  * disable the scrolling in the widget, while the scrollarea is scrolling
  * disable the scrolling in the scrollarea, while the widgets are scrolling

MichaelAtOz:
I don't use any apps which have a similar layout.
Anyone come across this before in other apps?

MichaelPFrey:
Good observation. I mean, the layout (input elements like slider and combox in scroll area) is not uncommon, any website has to deal with that.

What is uncommon, that the input elements react to the scroll wheel.
We could simply disable the scroll wheel for most input elements and the user would not even realize that something is missing.
(yet: I do not want to do that - it really encourages users that think coding or 3d modeling is not theirs to use OpenSCAD. It lowers the bar of entry by giving these user imitate feedback with even using the keyboard)

As the scrollbar on the side only appears when there is not enough vertical space:
Discussing the general UI with a non-OpenSCAD user, a suggestion was to use tabs instead of folding boxes. Using tabs would also bring OpenSCAD more in line with http://customizer.makerbot.com/docs (and also give the Global Tab a reason for existing)

MichaelPFrey:
Working on a related problem in the preferences of the inputdriver (#2252), I found this: https://stackoverflow.com/questions/5821802/qspinbox-inside-a-qscrollarea-how-to-prevent-spin-box-from-stealing-focus-when

[...]

PR #2338, opened 2018-03-11
FileModule::getFullpath(): fixing issue under windows with backslash

MichaelPFrey:
path / path under windows causes boost to use backslash.
The file paths within openscad are normalized to normal slash.

Issue was discovered by  MichaelAtOz  while trying to test a new build for #2185

this fix was tested under windows 10 using https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/Building_on_Microsoft_Windows for build instructions.

see also #2322

### DIFF
--- a/src/FileModule.cc
+++ b/src/FileModule.cc
@@ -194,7 +194,7 @@ const std::string FileModule::getFullpath() const {
 	if(fs::path(this->filename).is_absolute()){
 		return this->filename;
 	}else if(!this->path.empty()){
-		return (this->path + "/" + this->filename);
+		return (fs::path(this->path) / fs::path(this->filename)).generic_string();
 	}else{
 		return "";
 	}
--- a/src/FileModule.cc
+++ b/src/FileModule.cc
@@ -194,7 +194,7 @@ const std::string FileModule::getFullpath() const {
 	if(fs::path(this->filename).is_absolute()){
 		return this->filename;
 	}else if(!this->path.empty()){
-		return (fs::path(this->path) / fs::path(this->filename)).string();
+		return (this->path + "/" + this->filename);
 	}else{
 		return "";
 	}


MichaelAtOz:
`"/"` and losing `fs::path` seems optimistic, there have been lots of problems with this in the past.
I'm no expert, but wouldn't `generic_string()` be better?

kintel:
See also https://github.com/openscad/openscad/blob/master/src/value.cc#L60
Not sure if we can share any code, but similar ideas there.

MichaelPFrey:
@MichaelAtOz Perfect hint - yes, you are right.
OpenSCAD is typically using generic_string.

kintel:
@MichaelPFrey Would this have been caught by the test framework, had we run it under Windows?

MichaelPFrey:
@kintel This issue specifically: I do not think that the automated tests under windows would have been able the catch that, due to a quirk of the testing and/or the implementation of file module.

Taking a look at where the issue takes effect and the only place where getFullpath is actually used:
```
if(firstLine>=parseTill || (
	assignment.location().fileName() != "" &&
	assignment.location().fileName() != root_module->getFilename() &&
	assignment.location().fileName() != root_module->getFullpath()
	)) {
	continue;
}
```
Line 3 is - as fare as I remember - for the automated testing.
Line 4  is - as fare as I remember - for the GUI.

assignment.location().fileName() is always a full path, so that is easy.

With the GUI, getFilename() only returns the filename, so I had to write getFullpath(), that assembles path and filename into a fullpath.

But: For some reason, during automated testing getFilename() does deliver the full path (meaning that filename of file module contains the full path).

kintel:
FYI: The test framework simply run OpenSCAD in cmd-line mode, so this should be reproducible.
Anyway, no big deal, just nice to know if we could improve test coverage.

Issue #1910, opened 2017-01-07
when an assertion fails, openscad should output the file name, when the assertion failed in an included or used file


MichaelPFrey:
**Observation:**
When an assertion fails in an included file, OpenSCAD outputs the correct line number, but not file where the assertion failed.

**Issue:**
This makes it hard to trace down in where (as in within which file) the assertion failed.
This issue makes it hard to write libraries that relay on assertions.

**Suggested solution:**
when an assertion fails, openscad should output the file name, when the assertion failed in an included or used file

**Additional suggestion:**
Write a test case, where an assertion in an included file fails.

**Workaround:**
if(invalid condition){
  echo("<font color='red'>Error: Description of the error</font>"); 
  assert(false);
}

**Related:**
related to #1828 but this issue should be simpler to implement
related to #1860 but in my opinion not blocking

PR #2341, opened 2018-03-13
writing the filename where the assertion failed

MichaelPFrey:
failed in 

based on #2322 closing #1910

testing is an issue

### DIFF
--- a/src/AST.h
+++ b/src/AST.h
@@ -17,7 +17,7 @@ public:
 	}
 
 	std::string fileName() const { return path ? path->generic_string() : ""; }
-	std::shared_ptr<fs::path> filePath() const { return path; }
+	fs::path filePath() const { return *path; }
 	int firstLine() const { return first_line; }
 	int firstColumn() const { return first_col; }
 	int lastLine() const { return last_line; }
--- a/src/AST.h
+++ b/src/AST.h
@@ -17,6 +17,7 @@ public:
 	}
 
 	std::string fileName() const { return path ? path->generic_string() : ""; }
+	std::shared_ptr<fs::path> filePath() const { return path; }
 	int firstLine() const { return first_line; }
 	int firstColumn() const { return first_col; }
 	int lastLine() const { return last_line; }
--- a/src/FileModule.cc
+++ b/src/FileModule.cc
@@ -32,9 +32,10 @@
 #include "modcontext.h"
 #include "parsersettings.h"
 #include "StatCache.h"
-
+#include "evalcontext.h"
 #include <boost/algorithm/string.hpp>
 #include <boost/filesystem.hpp>
+#include "boost-utils.h"
 namespace fs = boost::filesystem;
 #include "FontCache.h"
 #include <sys/stat.h>
@@ -179,6 +180,16 @@ AbstractNode *FileModule::instantiateWithFileContext(FileContext *ctx, const Mod
 		// FIXME: Set document path to the path of the module
 		auto instantiatednodes = this->scope.instantiateChildren(ctx);
 		node->children.insert(node->children.end(), instantiatednodes.begin(), instantiatednodes.end());
+	}
+	catch (AssertionFailedException &e) {
+		auto docPath = boost::filesystem::path( ctx->documentPath() );
+		auto uncPath = boostfs_uncomplete((*(e.loc.filePath())), docPath);
+
+		std::stringstream msg;
+		msg << e.what();
+		msg << " failed in file " << uncPath.generic_string() <<",";
+		msg << " line " << e.loc.firstLine();
+		PRINT(msg.str());
 	}
 	catch (EvaluationException &e) {
 		PRINT(e.what());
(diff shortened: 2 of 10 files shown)


kintel:
Nice!

How hard would it be to output a path relative to the main file? I believe we do that other places, I can hunt that down later if you don't find it.

MichaelPFrey:
hhmmm... interesting idea. I planned on using the the absolute path, but had to switch to the filename because of the automated testing. Using paths relative to the main file resolves those issue.

boostfs_uncomplete() sounds like the right method.
Now I just need two paths.

Context contains a filename and the parent Context.

I also just realized, that I should test with both include and use.

I will look into implementing and then testing it.

kintel:
Sounds great! Let me know if you need any input.

MichaelPFrey:
The context works for include,
include
but not for use:
use

When mixing include and use, it would get weird.
So, that approach does not work (at least without further modifications.
I am still looking into it.

MichaelPFrey:
I could try to get an instance of MainWindow and ask for the root_module - but that approach is UI-Mode only.

I could argue, that we can use the fullpath in Console, as console outputs are noisy by nature - but then the test fails.

Either-way: Bypassing the current datastructures  to get the path of the root module seems more logical then, then to modify the current datastructur (e.g. Context).

kintel:
Since you only have a few tests, we could use a workaround for making the tests work by substituting the absolute path during cmake time. We already do this for some tests, see https://github.com/openscad/openscad/blob/master/tests/CMakeLists.txt#L1113

..and then just open a new issue to output relative filenames.

MichaelPFrey:
I found the solution :-) The exception handler knows the path I need, so I just had to modify exception and exception handler.

I just need to clean up the code and may adjust the test cases - but that is straight forward.

MichaelPFrey:
The tests look and are awkward:
`ERROR: Assertion 'false' failed in file ../../../../../testdata/scad/functions/assert-expression-fail1-test.scad, line 1`
but that is a separate issue. If you insert this code
```
		std::cout << docPath.generic_string() << "\n";
		std::cout << (*(e.loc.filePath())).generic_string() << "\n";
		std::cout << "----------"  << "\n";
		std::cout.flush();
```
into the exception handler, you see that path are awkward:
```
/home/michael/openscad/tests/../testdata/scad/functions
/home/michael/openscad/tests/../testdata/scad/functions/assert-expression-fail1-test.scad
```
I could try to canonicalize the paths before uncompleting them - but that is the issue: In an exception handler, I should not resort to call methods that throws exceptions. The canonicalization should happen before we even try to read a file.

src/AST.h
```diff
@@ -17,6 +17,7 @@ class Location {
 	}
 
 	std::string fileName() const { return path ? path->generic_string() : ""; }
+	std::shared_ptr<fs::path> filePath() const { return path; }
```
kintel: Did you intend to return a writable path here? At first glance it looks like this should really just return an `fs::path` by value.

src/FileModule.cc
```diff
@@ -180,6 +181,16 @@ AbstractNode *FileModule::instantiateWithFileContext(FileContext *ctx, const Mod
 		auto instantiatednodes = this->scope.instantiateChildren(ctx);
 		node->children.insert(node->children.end(), instantiatednodes.begin(), instantiatednodes.end());
 	}
+	catch (AssertionFailedException &e) {
+		auto docPath = boost::filesystem::path( ctx->documentPath() );
```
kintel: Coding style: No space after `(` or before `)`. We'll start enforcing this soon.

src/FileModule.cc
```diff
@@ -180,6 +181,16 @@ AbstractNode *FileModule::instantiateWithFileContext(FileContext *ctx, const Mod
...
+		auto uncPath = boostfs_uncomplete((*(e.loc.filePath())), docPath);
+
+		std::stringstream msg;
+		msg << e.what();
+		msg << " failed in file " << uncPath.generic_string() <<",";
+		msg << " line " << e.loc.firstLine();
+		PRINT(msg.str());
```
kintel: Using `PRINTB` may make this look nicer by being able to drop the `stringstream`. I'll let you decide what works best.

src/AST.h
MichaelPFrey: Oopps... that is an oversight on my part. The object return should off course not be writable.

src/FileModule.cc
MichaelPFrey: I see your point, but... that could go wrong if not done properly. Let us unwind it: msg << e.what(); is the "old"/current message inc. "ERROR  Assertion " followed by the expression... and the expression could contain %s.

I could off-course move the whole message building into the exception handler and use %s for including the expression... but going down that rabbit hole ends in rewriting all error messages incl. making them translatable. So, that sounds like rainy holiday project. I´ll make my self a note.

MichaelPFrey:
I think it is good to merge.

I think I will revisit the testing at a latter point in time and once that is fixed, we can add some more specific test cases. But as a feature for the UI: I am happy so far.

During the weekend, I will make my self a windows build.

src/FileModule.cc
kintel: Not sure I understand that concern. This is not a bit deal, but wouldn't this work?
```
PRINTB("%s failed in file %s, line %d", e.what(), uncPath.generic_string(), e.loc.firstLine());
```

MichaelPFrey: Ah - now I see, I will do that.

Issue #2342, opened 2018-03-14
Error Memory exhausted


edwardramg:
Good morning,

I have a problem with an Open Scad file of 250,000 lines of code:
translate([0.00000000000000, 0.00000000000000, 1])
cube([0.01,0.01,0.01]);
translate([0.00000000000000, 0.02004008016032, 1])
cube([0.01,0.01,0.01]);

... Until 250,000 lines:
translate([10.00000000000000, 10.00000000000000, 1])
cube([0.01,0.01,0.01]);

When I open it with OpenScad, I get the following error:
ERROR: Parser error in line 39982: memory exhausted 
ERROR: Compilation failed! 

Any ideas on how to solve it?

Thank you,

Edward

thehans:
What version of OpenSCAD are you running?

Can you attach a a single .scad file that reproduces the problem?

I did a quick google search and found this entry for "memory exhausted": https://www.gnu.org/software/bison/manual/html_node/Memory-Exhausted.html
which links to
https://www.gnu.org/software/bison/manual/html_node/Recursion.html#Recursion
With the takeaway being:
> Any kind of sequence can be defined using either left recursion or right recursion, but you should always use left recursion, because it can parse a sequence of any number of elements with bounded stack space.

I wonder if there are any cases of right recursion in openscad parser code that can be changed to left recursion as the bison docs recommend?
I haven't looked very deep at how the parser works yet.

edwardramg:
Thank you, I'll try your proposal and I'll let you know.

Saludos cordiales,

Atte... 

!Mario Eduardo Ramos García

De: Hans L [mailto:notifications@github.com]
Enviado el: miércoles, 14 de marzo de 2018 02:19 p. m.
Para: openscad/openscad <openscad@noreply.github.com>
CC: Mario Eduardo Ramos Garcia <mramos@imp.mx>; Author <author@noreply.github.com>
Asunto: Re: [openscad/openscad] Error Memory exhausted (#2342)


If you really want to display hundreds of thousands of cubes like this I would recommend putting the translation points into one large list, and using a for loop over the list to translate, that would at least cut down the file size somewhat.

points = [

 [0,0,1],[1,0,1], //... etc.

];

for (p = points) translate(p)

  cube([...]);

edwardramg:
Thank you for your help, the code of my scad file is already working properly, with the code suggestion you gave me.


Best regards

Atte... 

!Mario Eduardo Ramos García

[...]

Issue #1991, opened 2017-04-10
ERROR: Parser error in file .... memory exhausted 


oysteinkrog:
Hi,

I'm having some trouble with a memory exhaustion problem in the parser, it seems like I'm hitting a hardcoded limit :/

I suspect this is because I've organized my design in a very modular fashion.
In particular I've landed on a design pattern where I have "data" files that contain arrays of object definitions, e.g. bearings, screw threads etc. The problem is that I include these data files in many of my sub-modules/files, and when I in turn try to use these modules the underlying files are probably "included" several times.
Is there any kind of cache for included files?

If anyone wants to take a look my project is here: https://github.com/oysteinkrog/kmax1/
It's easy to reproduce the problem by opening main.scad and reverting the last commit ("temp fix").

Regards
Øystein

thehans:
This parser error should be fixed by #2343 which was included in the latest stable release of OpenSCAD.  
@oysteinkrog Give it a try if you like, feel free to re-open if you still experience any issues.  

Closing.

PR #2343, opened 2018-03-15
Use left recursion for "input" in parser

thehans:
This is a first pass at addressing issue #2342 by changing `input` to use left recursion as recommended by bison docs: https://www.gnu.org/software/bison/manual/html_node/Recursion.html#Recursion 

This change seems to allow openscad to get past the compile stage for very large scad files without a "memory exhausted" error from bison.  However it seems the issue is passed further down the chain because (with eg. 40k lines input, 20k elements) it now segfaults in `CSGTreeNormalizer::normalizePass` from I'm guessing too much recursion, at a stack depth of about 18,000 
https://github.com/thehans/openscad/blob/issue2342/src/CSGTreeNormalizer.cc#L100

Not sure if there's some better way to handle `normalizePass` that would allow it to support higher element counts, should it be recurring this much for a flat file input of many translated cubes?


kintel:
The CSG tree is a binary tree, built using that recursive process. There is room for optimization, but that particular process can be tricky to get right. We have decent test coverage for that though, as I’ve broken that code multiple times already.

thehans:
I did some work to rewrite `CSGTreeNormalizer::normalizePass` to be iterative, which was pretty tricky and the resulting code is not as elegant as I would like.  

I also found that some stack limits would be reached in `CSGProducts::import` and rewrote that as iterative; that one at least is straightforward.  

I am attaching some of the test scad files based on a 500,000 line scad example from issue #2342 , truncated at various lengths for testing:
[ManyCubes_scad.zip](https://github.com/openscad/openscad/files/1853528/ManyCubes_scad.zip)
In these examples there is one element for every 2 lines of .scad code.
I hesitate to add these as actual tests just due to how large of a commit it would be.

Before changes, OpenSCAD would fail to preview just 40,000 lines.  Now it seems to handle up to around 400,000 before things get iffy.   Testing this high requires increasing the default csg element limit in settings.   Previewing around 400kloc a segfault occurs due to stack limits in the way that shared_ptr gets reset recursively in the tree, coming from the code that resets the rootnode in `CSGTreeNormalizer::normalize`
```C++
this->rootnode.reset();
```

Another test I tried was creating large number of elements in a loop.  This does not test stress the parser in the same way, but the normalization issues are the same.  This makes 100k elements for preview testing:
```
xcount = 100;
ycount = 100;
zcount = 10;
totalcount = xcount*ycount*zcount;
echo("total elements:", totalcount);
for(x=[1:xcount],y=[1:ycount],z=[1:zcount]) 
  translate([x,y,z]) cube(0.5);
```

I feel like 100k elements is a fair goal to be capable of displaying, since that's the default limit in preferences/settings anyways.

MichaelPFrey:
just an idea for a parser test case: you could write a phyton script that generates a SCAD file.

kintel:
We already do generate test files, although just using simple substitution in cmake, but it's conceptually no different: https://github.com/openscad/openscad/blob/master/tests/CMakeLists.txt#L1113

oysteinkrog:
I think I ran into this problem in #1991. My code is a complex model (parametric 3d printer), so a much more realistic codebase than #2342. I can test windows binaries if needed.

thehans:
@Kintel I finally added a parser test case generated from python.  Do you think this is ok to merge now?

kintel:
Thanks - I'll try to review this soon. The unrolling of the recursive algorithm scares me a bit, but I think we can mostly trust the tests.

PR #2346, opened 2018-03-19
customizer: add support for text max. length

MichaelPFrey:
It would make sense tobe able to limit the string length, so I implemented just that:

test1 = "123"; //[3]
~~test2 = "1234"; //[0:4]~~
~~test3 = "12345"; //[0:1:5]~~


kintel:
It's slightly misleading to allow using a range when the minimum is not respected. Stepsize is also not very useful (nor being used).

MichaelPFrey:
I see your point.

The "support" for [0:4] and [0:1:5] is based on our internal data structure which are for `for (i = [start:step:end]) {}`. The customizer on the otherhand is based on the thingiverse customizer, where the [max] and we support that in our customizer for compatibility (so we might as well use it here).

PR #2354, opened 2018-03-27
Customizer: Bugfix: Do not save hidden variables

MichaelPFrey:
see  #2316

### DIFF
--- a/src/parameter/ParameterWidget.cc
+++ b/src/parameter/ParameterWidget.cc
@@ -406,7 +406,10 @@ void ParameterWidget::updateParameterSet(std::string setName)
 	if (!setName.empty()) {
 		pt::ptree iroot;
 		for (entry_map_t::iterator it = entries.begin(); it != entries.end(); it++) {
-			iroot.put(it->first,it->second->value->toString());
+			std::string groupName=entries[it->first]->groupName;
+			if (groupName != "Hidden") {
+				iroot.put(it->first,it->second->value->toString());
+			}
 		}
 		setMgr->addParameterSet(setName, iroot);
 		const QString s(QString::fromStdString(setName));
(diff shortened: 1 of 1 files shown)


src/parameter/ParameterWidget.cc
```diff
@@ -406,7 +406,12 @@ void ParameterWidget::updateParameterSet(std::string setName)
 	if (!setName.empty()) {
 		pt::ptree iroot;
 		for (entry_map_t::iterator it = entries.begin(); it != entries.end(); it++) {
-			iroot.put(it->first,it->second->value->toString());
+			std::string VariableName = it->first;
+			std::string VariableValue = it->second->value->toString();
+			std::string groupName = entries[VariableName]->groupName;
```
kintel: Isn't this the same as `it->second->groupName`?
Could you rewrite the entire loop as smth. like this?
```
for (const auto &entry : entries) {
  if (second->groupName != "Hidden") {
    const auto &VariableName = entry.first;
    const auto &VariableValue = entry.second->value->toString();
    iroot.put(VariableName, VariableValue);
  }
}
```

kintel: Just a small suggestion to slightly increase the readability of that code. The surrounding code could need some love as well, but that can be tackled later :)

MichaelPFrey: yes, it->second->groupName works.
yes, the for loop works and reads much nicer.
I am pushing the changes.

I agree that the readability can be improved much the same way in other places. It is just kind of tricky to find the balance of adding features, refactoring (on a larger scale) and increasing readability (on a local scope). Currently, I am working on Groups only beeing shown under a condition (which also helps understanding the bigger picture)

PR #2355, opened 2018-03-27
Limit number of frames painted 

thehans:
This change is intended to workaround redundant calls to `QGLView::paintGL()`.  There are many possible events and function that result in a paint call.  

Below is an example script for manually testing the responsiveness of orbiting the camera view (left click drag).  I get about 1fps with these settings on the nightly OBS build, and about 2fps with my build; feels 2x faster.
```
$fn = 100;
for(x=[1:10],y=[1:10],z=[1:10])
  translate((1+$t)*[x,y,z])
    sphere(d=1);
```

One issue I've noticed (why I've set "DO NOT MERGE" for now) is that these changes don't seem to play nicely with the animation toolbar.  For example setting FPS 100 ans steps 100, it flickers through a lot of frames that only show background.  This may only be a problem when Animation toolbar FPS is greater than the `MAX_FPS` define in `QGLView.h`.  Should the Animation toolbar FPS have an upper limit imposed?

And another side effect is that the using Alt-Up/Down arrow repeatedly on the value in this script seems to add about almost a second delay to the last frame: 
```
cube(1);
```

Besides these two regressions which I'm still trying to understand, I think overall that viewing complex geometries is about twice as fast.

### DIFF
--- a/src/QGLView.cc
+++ b/src/QGLView.cc
@@ -58,6 +58,7 @@ QGLView::QGLView(QWidget *parent) :
 	QGLWidget(parent)
 #endif
 {
+  installEventFilter(this);
   init();
 }
 
@@ -71,7 +72,9 @@ void QGLView::init()
 
   this->mouse_drag_active = false;
   this->statusLabel = nullptr;
+
   setMouseTracking(true);
+
   #ifdef USE_QOPENGLWIDGET
     setUpdateBehavior(QOpenGLWidget::PartialUpdate);
   #endif
@@ -108,10 +111,6 @@ void QGLView::initializeGL()
     fprintf(stderr, "GLEW Error: %s\n", glewGetErrorString(err));
   }
   GLView::initializeGL();
-
-  frameTimer = new QTimer(this);
-  QObject::connect(frameTimer, SIGNAL(timeout()), this, SLOT(renderFrame()));
-  frameTimer->start(1000 / MAX_FPS);
 }
 
 std::string QGLView::getRendererInfo() const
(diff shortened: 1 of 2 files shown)


thehans:
I've been testing the Preview performance particularly. The above sphere example is not recommended to be F6 Rendered.

thehans:
I am trying a few different things still.  I think ideally we can have the OpenGL rendering happening in its own `QThread` that is decoupled from the GUI event loop(and ditch the QTimer that I initially tried).  But I'm still having trouble fully understanding how to implement threading properly with `QOpenGLWidget`.

Here's the relevant section from the docs:  http://doc.qt.io/qt-5/qopenglwidget.html#threading
> When a thread is done updating the framebuffer, call update() on the GUI/main thread to schedule composition.

I don't know specifically how to do this.

Also not really understanding the majority of the following paragraph:
> Extra care has to be taken to avoid using the framebuffer when the GUI/main thread is performing compositing. The signals aboutToCompose() and frameSwapped() will be emitted when the composition is starting and ending. They are emitted on the GUI/main thread. This means that by using a direct connection aboutToCompose() can block the GUI/main thread until the worker thread has finished its rendering. After that, the worker thread must perform no further rendering until the frameSwapped() signal is emitted. If this is not acceptable, the worker thread has to implement a double buffering mechanism. This involves drawing using an alternative render target, that is fully controlled by the thread, e.g. an additional framebuffer object, and blitting to the QOpenGLWidget's framebuffer at a suitable time.

thehans:
OK, I just realized there is a threaded `QOpenGLWidget` example:
https://github.com/qt/qtbase/blob/5.11/examples/opengl/threadedqopenglwidget/glwidget.cpp

I'll see if I can adapt this to the OpenSCAD code

thehans:
So the example from Qt uses `QOpenGLFunctions` and `QOpenGLContext` to pass context between threads, and these classes are not compatible with GLEW library that OpenSCAD is currently using.  Its not clear to me the differences in how GLEW handles the context/threading and if it requires some special handling to be on a separate thread and be compatible with `QOpenGLWidget` (or if possible at all).  

I'm wondering if there is any objection to the idea of converting OpenSCAD away from using GLEW and instead use Qt's `QOpengGLFunctions` and context handling.  I think the big difference would be that OpenSCAD's `Renderer` classes would then depend on Qt.  
Would that cause problems for headless versions?

kintel:
Since the headless version doesn't use Qt at all, that would probably not fly.
I haven't looked into this, but if we do all OpenGL rendering in a separate thread, how would that be incompatible with GLEW? Can't we just promise to only call into GLEW from either QOpenGLWidget (in GUI mode) or the main thread in cmd-line mode?

thehans:
The problem is that I need to transfer the OpenGL context to the thread that is doing the rendering.  
Its still a little unclear but I guess that `QOpenGLWidget` is automatically creating the context internally.

The Qt example of multithreaded rendering does something like this to move the context from `QOpenGLWidget` to another thread:
```C++
this->context()->moveToThread(renderthread)
```

This would require me to `#include <QOpenGLContext>`, but doing so gives a compiler warning from Qt that its not compatible with GLEW:
```
In file included from /usr/include/x86_64-linux-gnu/qt5/QtGui/QOpenGLContext:1:0,
                 from /home/hans/projects/openscad/src/QGLView.h:12,
                 from /home/hans/projects/openscad/src/QGLView.cc:28:
/usr/include/x86_64-linux-gnu/qt5/QtGui/qopenglcontext.h:49:2: warning: #warning qopenglfunctions.h is not compatible
 with GLEW, GLEW defines will be undefined [-Wcpp]
 #warning qopenglfunctions.h is not compatible with GLEW, GLEW defines will be undefined
  ^
/usr/include/x86_64-linux-gnu/qt5/QtGui/qopenglcontext.h:50:2: warning: #warning To use GLEW with Qt, do not include 
<qopengl.h> or <QOpenGLFunctions> after glew.h [-Wcpp]
 #warning To use GLEW with Qt, do not include <qopengl.h> or <QOpenGLFunctions> after glew.h
  ^
```

thehans:
I pushed my latest broken code just to give an idea of what I'm trying to do.  It still gives compiler warnings about QtOpenGLContext/GLEW incompatibility and segfaults immediately on `glewInit()`
I'm probably doing dozens of things wrong here and I've going back and forth testing various approaches so the code is a little scatterbrained.

I tried to roughly follow the multithreaded example from Qt(I linked in a comment above), although I omitted or changed parts of code related to multiple contexts.  I think OpenSCAD should only ever have one context whereas the example renders to multiple windows simultaneously.

thehans:
I started a topic on Qt forum asking for some help:
https://forum.qt.io/topic/89458/possible-to-use-qopenglwidget-with-glew-and-render-outside-of-main-thread

Still discussing there, but one thing that was suggested was that headless mode could use "offscreen QPA"
http://doc.qt.io/qt-5/qpa.html
Not sure what this would entail exactly but seems like its not feasible until OpenSCAD can drop support for Qt4 builds.

thehans:
This was something issue I was chasing that was apparently cause by my Qt version at the time (5.5.1)
I can no longer reproduce any extra frame draws with my current build using Qt 5.9.5 on Linux Mint 19

PR #2356, opened 2018-03-28
 customizer: modernizing the for loops

MichaelPFrey:


src/parameter/ParameterWidget.cc
```diff
@@ -252,22 +252,23 @@ void ParameterWidget::connectWidget()
 		}
 	}
 	cleanScrollArea();
-	for (std::vector<std::string>::iterator it = groupPos.begin(); it != groupPos.end(); it++) {
-		if(groupMap.find(*it)!=groupMap.end()){
+
+	for (const auto &it : groupPos) {
```
kintel: `it` is no longer an iterator, it's a string; change to `groupName` ?

src/parameter/ParameterWidget.cc
```diff
@@ -292,9 +293,9 @@ void ParameterWidget::rebuildGroupMap(){
 			it++;
 		}
 	}
-	for (group_map::iterator it = groupMap.begin(); it != groupMap.end(); it++){
-		it->second.parameterVector.clear();
-		it->second.inList=false;
+	for (auto &it : groupMap) {
```
kintel: `it` isn't an iterator. This time it's a "group map element". Unfortunately this is still a bit annoying until C++17 hits, so just renaming from `it` to `elem` or smth. will make this clearer.

src/parameter/parametercombobox.cpp
```diff
@@ -55,19 +55,20 @@ void ParameterComboBox::setValue()
...
 	const Value::VectorType& vec = object->values->toVector();
-	for (Value::VectorType::const_iterator it = vec.begin(); it != vec.end(); it++)	{
-		if ((*it)->toVector().size() > 1) {
-			comboBox->addItem(QString::fromStdString((*it)->toVector()[1]->toString()),
-												QVariant(QString::fromStdString((*it)->toVector()[0]->toString())));
-		}
-		else {
-			comboBox->addItem(QString::fromStdString((*it)->toString()),
-												QVariant(QString::fromStdString((*it)->toString())));
+	for (const auto &it : vec) {
```
kintel: s/it/value/ ?

src/parameter/parametercombobox.cpp
```diff
@@ -55,19 +55,20 @@ void ParameterComboBox::setValue()
...
+		QString text, data;
+		if ((*it).toVector().size() > 1) {
```
kintel: I find it a bit cleaner to just to `it->toVector()`. Same for a few instances of this below

PR #2359, opened 2018-03-29
Customizer: show_if for groups

MichaelPFrey:
Example SCAD Files:
```
sphereF=true;
boxF=true;

/*[sphere] show_if(sphereF)*/
r = 10;
/*[box] show_if(boxF)*/
b = 10;
/*[cylinder] show_if(cylinderF)*/
r2 = 1;

if(sphereF) sphere(r);
if(boxF) translate([r,0,0])cube(b);
```

```
show = "box + sphere"; //["box + sphere", "box", "sphere"]

/*[sphere] show_if(sphereF)*/
r = 10;
/*[box] show_if(boxF)*/
b = 10;
/*[cylinder] show_if(cylinderF)*/
r2 = 1;

/*[hidden]*/    
sphereF = (show=="box + sphere") || (show=="sphere");
boxF =    (show=="box + sphere") || (show=="box");

if(sphereF) sphere(r);
if(boxF) translate([r,0,0])cube(b);
```

### DIFF
--- a/src/AST.cc
+++ b/src/AST.cc
@@ -15,3 +15,21 @@ std::string ASTNode::dump(const std::string &indent) const
 	print(stream, indent);
 	return stream.str();
 }
+
+void ASTNode::addAnnotations(AnnotationList *annotations)
+{
+	for (auto &annotation : *annotations) {
+		this->annotations.insert({annotation.getName(), &annotation});
+	}
+}
+
+bool ASTNode::hasAnnotations() const
+{
+	return !annotations.empty();
+}
+
+const Annotation * ASTNode::annotation(const std::string &name) const
+{
+	auto found = annotations.find(name);
+	return found == annotations.end() ? nullptr : found->second;
+}
--- a/src/AST.h
+++ b/src/AST.h
@@ -2,6 +2,7 @@
 
 #include <string>
 #include <memory.h>
+#include <annotation.h>
 #include <boost/filesystem.hpp>
 namespace fs = boost::filesystem;
 
@@ -22,7 +23,6 @@ public:
 	int firstColumn() const { return first_col; }
 	int lastLine() const { return last_line; }
 	int lastColumn() const { return last_col; }
-
 
 	static const Location NONE;
 private:
@@ -45,8 +45,13 @@ public:
 	const Location &location() const { return loc; }
 	void setLocation(const Location &loc) { this->loc = loc; }
 
+	virtual void addAnnotations(AnnotationList *annotations);
+	virtual bool hasAnnotations() const;
+	virtual const Annotation *annotation(const std::string &name) const;
+
 protected:
 	Location loc;
+	AnnotationMap annotations;
 };
 
 std::ostream &operator<<(std::ostream &stream, const ASTNode &ast);
--- a/src/Assignment.cc
+++ b/src/Assignment.cc
@@ -29,25 +29,6 @@
 #include "expression.h"
 #include <sstream>
 
-void Assignment::addAnnotations(AnnotationList *annotations)
-{
-	for (auto &annotation : *annotations) {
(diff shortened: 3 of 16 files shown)


MichaelPFrey:
This would be an exclusive feature.
It is not supported by thingiverse.

t-paul:
I'm still thinking it would be useful to have the comments as compatible with Thingiverse as possible.
The proposal for a native customizer markup included also that the actual meta data is parsed along the script not scraped from comments. Some examples are listed in the wiki - https://github.com/openscad/openscad/wiki/Meta-Data-Use-Cases

MichaelPFrey:
Currently, the Thingiverse customizer is not working, so I can not test. I presume that incompatible syntax/features simply get ignored (similarly to how our customizer reacts).

Using our own meta data syntax would break compatibility of SCAD files. (a new syntax yields errors in old releases)
Expanding the comment syntax is "compatible" as in it gets ignored when not support (but it still provides context information to a person editing SCAD files)

Our own meta data syntax would presumably coexist with the comment style for support of thingiverse files.

Presuming that we do not want to have two UI and Back end codes, the new meta data style would presumably generate the same or similar data-structures as the current comment passed parser. So, any extension to the comment based syntax could be reused for the meta data style syntax.

t-paul:
Yeah, the fixing seems to be still work in progress. We talked with one of the admins on IRC, but I don't know what the current state is. I tried earlier today and the "it's not working" marker is gone, but it's still producing the "can't load library" error.

The main problem I see with extensions in the comment syntax is that it's confusing for the user when trying to test stuff locally and then only part of the features work on Thingiverse.

The native syntax would not be compatible with older versions of OpenSCAD of cause, but that's the case with all new language features. When using new features, the new OpenSCAD version is required. I think that's not too surprising for users. We do try to not break older scripts with new versions as that is very annyoing. E.g. there's a number of things that are deprecated for quite some time but still work while producing a warning.

As for the UI, yes, that's supposed to work with just one data structure which is already called `Annotation` as that was coming from the native prototype I did quite some time ago :-).

The benefit of using the normal parser would also help making the `show_if()` more powerful as that could be an annotation parameter featuring full expression logic, like `visible = sphere_selected && radius > 6` although the scoping might be a tiny challenge to get it most useful.

MichaelPFrey:
Take a look how I implemented the conditions: show_if() as implemented  is using a boolean from the context of the root_module. So, that actually works:
```
/*[sphere] show_if(sphereF)*/
r = 10;
/*[hidden]*/    
sphereF = (show=="box + sphere") || (show=="sphere");
```
For someone new to openSCAD it might consider it strange that the assignment is after the first usage of the variable in the script - but within OpenSCAD that is "by nature" correct.
I was thinking about parsing the condition - but that would open up a can of worms.

Using a variable look up on to the root module context is pretty elegant as the user has the full power of OpenSCAD.

The draw back of this implementation is off-course, that the condition is only evaluated after a compile.

MichaelPFrey:
@kintel What do you prefer? Expanding the comment syntax for now or adding a new meta data syntax? (Note: expanding the comment syntax does not mean we can not add a meta data syntax latter - but the more feature we add in the comment syntax the less need will be for a native syntax)

kintel:
I've always viewed the comment syntax as a temporary workaround to help transitioning into a better design meta-data format. The big question is when someone will have time to work on the latter, so the current comment-based syntax may end up having more permanence than initially expected.

I find it slightly scary to start adding more parsing logic to the comment syntax.
If you find this syntax really helpful, I suggest we keep it, but take note that we should reconsider this when making the next stable release; i.e. make it easy to disable.

MichaelPFrey:
The argument with parser complexity is convincing for me. Every hack for it self maybe reasonable, but it stacks up - and that on top of a comment parser that it self is has room to improve. So, I close that pull request for now and may revisit it later.

PR #2366, opened 2018-04-02
Qt fixes

kintel:
* Out greaterThan() check didn't work with 2-digit minor version numbers (e.g. Qt 5.10)
* qmake doesn't like spaces in resource filenames


t-paul:
Is there a way to make ccache optional? So far it's not been a huge deal, but we force that on all build systems / distribution packagers (or I just don't know how to disable, `CONFIG-=ccache` did not work).

kintel:
@t-paul How about this?

We introduce a separate file, e.g. `local.pri`. This would be an optional file describing local build settings.
Then we could do (in common.pri):
```
exists(local.pri): include(local.pri)
```

..and in anybody's local.pri:
```
CONFIG -= ccache
```

Is that smth. you'd be happy about as a builder of packages?
We could probably go further and building some common presets as well.

kintel:
@t-paul See #2367

PR #2367, opened 2018-04-03
qmake revamp

kintel:
* Moved qmake config features to `features/*.prf`
* Introduced `defaults.pri` for standard config defaults
* Introduced the optional `local.pri` for local build env overrides
* Depends on https://github.com/openscad/openscad/pull/2366


kintel:
@t-paul Could you have a look at this and see if that improves your ccache workflow (and in general looks like a good idea)?

t-paul:
Merged into #2452.

Issue #2377, opened 2018-04-15
OPENSCADPATH relative to opened .scad file (for project-specific libraries)


sadr0b0t:
Hello,

I want to have project-specific libraries in one dir in the same dir where .scad file is located.

For example, we have project1 with files:

project1/model.scad 
project1/modules/lib1/lib1.scad
project1/modules/lib2/lib2.scad

I want to import lib1 and lib2 libraries transparently without providing modules subdir name like:

~~~scad
use <lib1/lib1.scad>
use <lib2/lib2.scad>
~~~

currently I can use full relative path including "modules" dir name like:
~~~scad
use <modules/lib1/lib1.scad>
use <modules/lib2/lib2.scad>
~~~
But this looks dirty.

I have tried to use OPENSCADPATH env variable with relative path value to achieve that but it gives only partial result.

If I run openscad from console from project dir like this:
~~~
export OPENSCADPATH=modules
cd /home/user/projects/project1
openscad model.scad
~~~

I can see relevant library path in the build info dialog:
>OPENSCADPATH: modules
OpenSCAD library path:
/home/user/projects/project1/modules

and can include modules without providing "modules" dir name (this is exactly what I want).

But if I run openscad by clicking on the model.scad file in file manager, importing library without "modules" dir name willl NOT work

add to .profile or global env varialbles
~~~
export OPENSCADPATH=modules
~~~

Then logout/login (on Linux) and open .scad file from file manager

>OPENSCADPATH: modules
    OpenSCAD library path:
    /home/user/modules

So, OpenSCAD will not use OPENSCADPATH value relative to current project dir in this case, it will attach it to $HOME dir value at startup, so "include <lib1/lib1.scad>" without "modules" will not work.

The use-case here is to have nodejs-like module dependency management.

There are at least 2 projects
nodejs npm-based: https://github.com/RobertFach/Openscad-Modules
ruby gems-based: https://github.com/lostapathy/scad_bundler#draft-conventions

The 1st one suggests to import libraries with full relative path like "use <node_modules/lib1/lib1.scad>" (which looks quite dirty to me), the 2nd one suggest to run openscad with script that sets OPENSCADPATH (which is quite inconvenient).

So my request here is:
- either hardcode project-related library dir name "modules"/"scad_modules"/"lib"/"libraries" or any other, so OpenSCAD would also search modules inside this dir if it exists in current project (like nodejs does with "node_modules" dir, for example)
- or allow OPENSCADPATH to be relative to currently opened scad file

[...]

PR #2384, opened 2018-04-20
Scad modules: load project-specific libraries recursively in nodejs-like way

sadr0b0t:
Create pull-request from this ticket:
https://github.com/openscad/openscad/issues/2377

1. Allow relative library paths in OPENSCADPATH
2. Search for included library file recursively starting from source .scad file dir+libdir to fs root+libdir when library dir is set as relative path
3. Add 'scad_modules' relative lib dir path to libraries paths for project-specific libraries by default

In a project like
project1/model.scad
project1/modules/lib1/lib1.scad
project1/modules/lib2/lib2.scad

one will be able to import lib1 and lib2 like:

~~~openscad
use <lib1/lib1.scad>
use <lib2/lib2.scad>
~~~

if OPENSCADPATH is set to 'modules' (or contains it among multiple values)
~~~
OPENSCADPATH=modules
~~~

A bit more complex usecase:

Imagine, we still have project
project1/model.scad
project1/modules/lib1/lib1.scad
project1/modules/lib2/lib2.scad

model.scad references lib1.scad like above
~~~openscad
use <lib1/lib1.scad>
~~~

and lib1.scad, in turn, references lib2.scad also in a similar way:
~~~openscad
use <lib2/lib2.scad>
~~~

This is common case if we use nodejs-like modules which depend on each other and refer to each other while being loaded to the same plain 'modules' dir.

But with current OpenSCAD library load strategy this will NOT work, because while being included from lib1.scad, lib2.scad should be located in the same dir as lib1.scad or in a subdir or somewhere in system libraries.

https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/Libraries
> In the case of a library file itself having use <...> or include <...> the directory of the library .scad file is the 'calling' file, i.e. when looking for libraries within a library, it does not check the directory of the top level .scad file.

Now change library search strategy:

once lib2.scad is included from lib1.scad like:
~~~openscad
use <lib2/lib2.scad>
~~~

first, try to search in relatively to lib1.scad dir in:
/home/user/project1/modules/lib1/modules/lib2/lib2.scad (NO)
if not found, then do the same for parent dir:
/home/user/project1/modules/modules/lib2/lib2.scad (NO)
and recursively until we reach root or find library:
/home/user/project1/modules/lib2/lib2.scad (YES)

in case, we did not find it above:
/home/user/modules/lib2/lib2.scad (NO)
/home//modules/lib2/lib2.scad (NO)
/modules/lib2/lib2.scad (NO)
no parent here -> library not found at all

In this case lib2.scad would be finally found in
project1/modules/lib2/lib2.scad

on the same level with lib1. This means, that different modules can be loaded in the same dir as main project dependencies and reference each other mostly by name.

This patch should not affect library dirs with absolute path in any way (so should not break any old setups), and libs with relative paths did not work anyway.

Also, add 'scad_modules' relative path to bult-in library dirs by default, so it could be used for openscad nodejs-style project-related modules (like "node_modules" in NodeJS), and the user will not have to add it OPENSCADPATH each time.

So, projects like:
project1/model.scad
project1/scad_modules/lib1/lib1.scad
project1/scad_modules/lib2/lib2.scad

will not have to set OPENSCADPATH to 'scad_modules' by default.

### DIFF
--- a/src/parsersettings.cc
+++ b/src/parsersettings.cc
@@ -97,7 +97,7 @@ void parser_init()
 		std::string sep = PlatformUtils::pathSeparatorChar();
 		typedef boost::split_iterator<std::string::iterator> string_split_iterator;
 		for (string_split_iterator it = boost::make_split_iterator(paths, boost::first_finder(sep, boost::is_iequal())); it != string_split_iterator(); ++it) {
-			add_librarydir(fs::absolute(fs::path(boost::copy_range<std::string>(*it))).string());
+			add_librarydir(fs::path(boost::copy_range<std::string>(*it)).string());
 		}
 	}
 
--- a/src/parsersettings.cc
+++ b/src/parsersettings.cc
@@ -132,4 +132,6 @@ void parser_init()
 	add_librarydir(PlatformUtils::userLibraryPath());
 
 	add_librarydir(fs::absolute(PlatformUtils::resourcePath("libraries")).string());
+	
+	add_librarydir("scad_modules");
 }
--- a/src/parsersettings.cc
+++ b/src/parsersettings.cc
@@ -14,15 +14,43 @@ static void add_librarydir(const std::string &libdir)
 }
 
 /*!
+	Searches lib 'localpath' in 'libdirpath' relatevely to current dir 'sourcepath',
+	then, if not found, in 'libdirpath' relatively to 'sourcepath' parent dir,
+	then relatevely to parent-parent dir etc until reaches file system root.
+*/
+static fs::path search_libs_recursively(const fs::path &localpath, const fs::path &sourcepath, const fs::path &libdirpath) {
+	if(sourcepath.empty()) {
+		return fs::path();
+	} else {
+		fs::path usepath = sourcepath / libdirpath / localpath;
+		//PRINTB("INFO: search lib: %s", usepath);
+		if (fs::exists(usepath) && !fs::is_directory(usepath)) {
+			return usepath;
+		} else {
+			return search_libs_recursively(localpath, sourcepath.parent_path(), libdirpath);
+		}
+	}
+}
+
+/*!
 	Searces for the given file in library paths and returns the full path if found.
 	Returns an empty path if file cannot be found or filename is a directory.
 */
-fs::path search_libs(const fs::path &localpath)
+fs::path search_libs(const fs::path &localpath, const fs::path &sourcepath)
 {
 	for(const auto &dir : librarypath) {
-		fs::path usepath = fs::path(dir) / localpath;
-		if (fs::exists(usepath) && !fs::is_directory(usepath)) {
-			return usepath.string();
+		fs::path libdirpath = fs::path(dir);
+		if(libdirpath.is_absolute()) {
+			fs::path usepath = libdirpath / localpath;
+			if (fs::exists(usepath) && !fs::is_directory(usepath)) {
+				return usepath;
+			}
+		} else {
+			// relative path: search lib recursively
+			fs::path usepath = search_libs_recursively(localpath, sourcepath, libdirpath);
+			if (!usepath.empty()) {
+				return usepath;
+			}
 		}
 	}
 	return fs::path();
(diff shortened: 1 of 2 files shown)


tsingi:
I have my scad set up like this...

.../projects
.../projects/lib.scad
.../projects/mything/
.../projects/mything/mything.scad

Where mything.scad imports "../lib.scad"
I do it this way so that when I change lib.scad it affects all projects that link it rather than a local copy.
I very much like doing it that way.

Still propagating the libraries in a sub directory as you suggest is much more straightforward for packaging.  

Any ideas on how to resolve that?  Off the top of my head I can see organizing the code using soft links so that it appears to the project that myproject/lib/lib.scad is in the project tree but in reality they are all pointing to the same file.  But this won't work on file systems that don't support soft links so that's a bad road.

That model appears to make this development style of file organization a no go.  Am I missing something?

sadr0b0t:
Hello,

1. This patch will not break your current setup

2. With this patch you will be able to arrange your setup like this:

.../projects
.../projects/scad_modules/lib1.scad
.../projects/mything1/mything1.scad
.../projects/mything2/mything2.scad

Both mything1.scad and mything2.scad will include lib1 like this:
~~~scad
include <lib1.scad>
~~~

The only benefit here is not to have ".." when include lib1 like:
~~~scad
include <.../lib1.scad>: 
~~~

This might give a bit more flexibility at some cases, but, actually, if you often edit both lib1.scad and mything1.scad and mything2.scad during the same dev cycle, you can consider them as parts of the same multiple-files monolithic project and your setup with ".." links can be a better solution in this case.

3. The approach with "scad_modules" and project-specific libraries would show its power when you deal with 3rdparty libraries, multiple 3rdparty libraries, which depend one on another, and when you need different stable versions of the same library in different projects.

So, your project depends on 3 3pty libraries: lib1,scad, lib2.scad and lib3.scad. You are not going to edit them during your dev cycle (probably, only some casual updates when the newer version is released). So, you just drop them to "scad_modules" dir in your project and reference them as "lib1/lib1.scad", "lib2/lib2.scad", "lib3/lib3.scad". Those libraries can even depend one on another, like: "lib1.scad" uses "lib2.scad" and "lib2.scad" uses "lib3.scad". They will also reference each other directly like "lib1/lib1.scad", "lib2/lib2.scad", "lib3/lib3.scad" without "scad_modules" or "..".

project1/mything1.scad
project1/mything2.scad
project1/scad_modules/lib1/lib1.scad
project1/scad_modules/lib2/lib2.scad
project1/scad_modules/lib3/lib3.scad

4. It will show max more power, when used with package manager and central module registry.

You just execute in your project dir in the console:
~~~
spm install lib1
~~~

And both lib1.scad, lib2.scad (because it is used by lib1) and lib3.scad (because it is used by lib2) all go to "scad_modules" dir automatically and you can start using them as described above. You can specify specific version of lib1 for different projects, update to recent versions of libraries with single command, specify dep library names in project manifest file, publish your own modules to the central registry, so they can be installed in the same easy way by other people etc etc.

(spm - "scad package manager" - an hypothetical opescad package manager, for example, alias to npm)

There are at least 2 projects which try to adopt package managers from NodeJS or Ruby to OpenSCAD at a very early stage (and they both suffer from missing "scad_modules" concept in OpenSCAD itself):
nodejs npm-based: https://github.com/RobertFach/Openscad-Modules
ruby gems-based: https://github.com/lostapathy/scad_bundler#draft-conventions

TomConlin:
allowing `import <foo>`  where foo is in/relative to the same dir  (which now seem to be renamed `use <foo>`)  to resolve again would be good to stop breaking backwards compatibility for those of us who do not revisit our old projects that often.

HKalbasi:
What is status of this? I think modern package management is a real demand and would revolutionize the openscad community. Is the problem with the concept of package manager, Or with the Implementation? Or is it just ignored?

t-paul:
Some kind of package management is quite high on the wishlist, but there's simply nobody who has time to work on that (as far as I'm aware). I think it would need also some discussion about the concept.

I was hoping we can go with some existing package manager (`entropic` sounded promising, but it looks pretty dead now, there's some prototypes using other language (e.g. ruby) package managers, but I'm not sure people would be happy to just push all things into that infrastructure and/or if those fit openscad well enough.

PR #2390, opened 2018-04-29
Allow each to work on scalar values (fixes #1851).

### LINKED ISSUE #1851: "each each" causes a crash
From the [mailing list](http://forum.openscad.org/each-each-td18863.html)
Example:
```
a = [ for (a=1) each each a];
```

-> Should we support `each each` ?

As a side effect this also causes a crash:
```
a=[each for(i=[1,2]) i];
```

-> This should be a syntax or run-time error

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/38873415-each-each-causes-a-crash?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

t-paul:
Assumption is that [each x] should just be a shortcut of [for(i=x) i].
This also enables safe use of [each each x].


t-paul:
It allows unwrapping multiple levels of nested vectors. The test case has some examples.

Issue #2375, opened 2018-04-12
Noise when rendering complete #featurerequest [$75]


dorkmo:
I think it'd be nice if there was an optional microwave beep when a rendering is done. 

thanks for the great software

##

MichaelAtOz:
+1.
Surely it should be the sound of trumpets and angels singing, or something...user configurable, maybe one of the built in sounds, 'Fax Error' (Windows) is probably not used much any more.

JustVic:
I made a try to implement it in pulrequest #2393.

MichaelAtOz:
Note for future, when this is expanded to file based, we probably want to keep one built-in sound in case users do strange things to the installation.

MichaelAtOz:
I was thinking about what size budget we should target for sound files.
So I checked past distributions, then got carried away:
```
(64bit unless noted - this is the Program Files size) 
Version		Size-MB		Mainly Due To 
2013.06		23.1
2015.03-1\32	30.0		.exe, fonts (~4MB)
2015.03-2\64	29.7
2016.06.08	29.6
2016.07.20	30.8 		.exe
2016.09.18	31.1		.exe
2016.10.04\32	31.1		-
2017.01.20	31.4		.exe
2017.08.31	34.5		.exe
2017.10.28	34.7		.exe
2018.01.06	34.5		.exe
2018.04.01	34.6		~
2018.05.05	39.0		.exe (presumably sound, note .wav file is 0.345MB)

```

So presumably we can afford a small number of MB, say low target of 2MB initially???
That would allows 2 big sounds and a few mediums and some more tinys.
And I can resample some sounds to squeeze them in if required.

MichaelAtOz:
> And I can resample some sounds to squeeze them in if required.

I got 1 800K+ file (at the unusual 96000 rate) down to ~10K with little noticeable sound difference BTW.

Lenbok:
I think that rather than adding a special-purpose feature like this, it would be much more useful to provide a general purpose hook that could run any user specified command. If my rendering is taking ages, I'm likely to have actually wandered out of earshot of the computer. With a general hook I could send a notification to my phone (e.g. via pushover/pushbullet).

MichaelAtOz:
Thanks Len, feel free to add another feature request, that may be something the community wants too.
This issue has a Bounty attached, and work has been done, so this should complete as specified.

JustVic:
Can this issue be closed or Do We need to do something else?

t-paul:
[...]

PR #2393, opened 2018-05-03
Play sound notification on render complete

JustVic:
Trying to implement sound notification feature issue #2375. I test it on my linux. Need to be tested on Windows and Mac. Also maybe needed  to add libcanberra to linux dependencies not sure. Say if I need to change something to make it better.


kintel:
I agree with @t-paul. Also, with QSound we could easily make the sound configurable through Preferences.

Next, someone will ask for Desktop Notifications, but let's cross that bridge later :)

MichaelAtOz:
> Desktop Notifications

Network alerts, emails, sms, Iot wall mounted indicator, IFTTT...

MichaelAtOz:
Looks like all platforms support WAVE. On the hunt for Open Source Waves...

JustVic:
Ok I will rewrite it with Qsound

t-paul:
@JustVic thanks. I think we should keep it simple. If the sound is too long, it's always possible to cut it in some audio software like Audacity.

So I'd suggest to have the scope as
- no stop or other control, just play file as is
- just a single bundled wav file for now (we can make that configurable later)
- option to enable/disable playing the sound (already implemented in the pull request)

Is that ok with the Bounty Backers of #2375?

MichaelAtOz:
Sorry, I should have added my thought to #2393.

> Is that ok with the Bounty Backers of #2375?

Well one is Anonymous the other is me, so add:
Confirm WAVE file works on the three platforms,
Assuming so, make it WAVE only (to possibly be expanded in other change),
(? @t-paul do we need to specify file location/name - presumably subdirectory in install dir for later extension)

t-paul:
Good point, there are 2 options to include the file that I can see (as we are talking one small file for now). One is to use Qt ressources which would essentially compile it in, the other is to put it in some installed folder. An installed file would have the benefit that the user could just overwrite while we have no config to change the sound.

Both solutions are ok, I think. I personally would slightly prefer the separate file in the resource folder indicated by `PlatformUtils::resourceBasePath()` (e.g. used in [scintillaeditor.cpp](https://github.com/openscad/openscad/blob/master/src/scintillaeditor.cpp#L406) to lookup color scheme files).

t-paul:
Note to self... Debian dev package is `qtmultimedia5-dev`

JustVic:
@MichaelAtOz Ok I rewrote to using wav file.
@t-paul I understand your points. Trying to make things simple.
I rewrote it with QSound. It looks like working on my linux, maybe need some testing. I use this wav sound https://freesound.org/people/original_sound/sounds/366102/, but I don't know where to add attribution (credits to sound), currently i add it in sounds/SOUND_CREDITS file. The sound is not perfect maybe, I'll look for some better sound a little more time, you can suggest better sounds.

t-paul:
Windows builds...
[OpenSCAD-2018.05.05-x86-32_sound-notification-Installer.exe](http://files.openscad.org/snapshots/OpenSCAD-2018.05.05-x86-32_sound-notification-Installer.exe)
[OpenSCAD-2018.05.05-x86-32_sound-notification.zip](http://files.openscad.org/snapshots/OpenSCAD-2018.05.05-x86-32_sound-notification.zip)
[OpenSCAD-2018.05.05-x86-64_sound-notification-Installer.exe](http://files.openscad.org/snapshots/OpenSCAD-2018.05.05-x86-64_sound-notification-Installer.exe)
[OpenSCAD-2018.05.05-x86-64_sound-notification.zip](http://files.openscad.org/snapshots/OpenSCAD-2018.05.05-x86-64_sound-notification.zip)

JustVic:
@t-paul I added qtmultimedia5-dev to .tarvis.yml.

t-paul:
Great, Travis is green now. Tested on Linux (Debian/Testing), looks good.

MichaelAtOz:
Works nicely on Windows 7/64. Thanks.

>  I don't know where to add attribution (credits to sound), currently i add it in sounds/SOUND_CREDITS file. 

@t-paul The sounds directory didn't make it to the distribution.
Presumably the Help/About text is the place for attribution? Remember it will be an expanded list eventually tho.
We sh/c/ould use a CC0 (public domain), but even then I lean toward attribution too??

> The sound is not perfect maybe, I'll look for some better sound a little more time, you can suggest better sounds.

It's OK in case we don't find a better one before long.

t-paul:
Yes, Help/About is probably the best place for CC-BY stuff. Tracking sources in the credits file even for CC0 is useful, at least for distributions that like to verify licenses of all files.

JustVic:
I moved sounds credits to AboutDialog. But I noticed warning "QCoreApplication::postEvent: Unexpected null receiver" it appears in terminal while exit from openscad. Currently I have not figured out why this appears.

t-paul:
The warning might be https://bugreports.qt.io/browse/QTBUG-66910

kintel:
Looks great!
I find the chosen sound to be a bit "tentative"; it almost sounds a bit like one of those warning sounds indicating that something is not quite ok.

JustVic:
@t-paul compiled with Qt version 5.7.1 does not have this bug, and no warnings are visible. so I think it is this bug in Qt 5.10.1 https://bugreports.qt.io/browse/QTBUG-66910, thanks for the hint. It does not affect on programm work for me.

JustVic:
maybe it also needs to add qtmultimedia to all platforms dependencies scripts. I can try to add for Arch linux and Debian(apt).

Lenbok:
Why don't you just make it run a user-supplied command on completion, rather than baking in this multimedia stuff. Then we can choose alternative notification methods (such as sending a message to our phone if we are out of earshot).

t-paul:
Ok, fine, lets list a couple of questions, I'll put my answers alongside, maybe you can provide your view.

- Is the original issue asking for a specific feature (#2375) useful?
> I never really missed that myself but it feels like a useful thing to have
- Will this prevent extensions like other notifications to be integrated later?
> No, not at all
- Will more complicated ways of notification (like calling external programs - which is an option I personally don't like) make this solution obsolete so it needs to be removed
> No, as it's the easiest to configure it's still useful even with more powerful options available
- Is BountySource a problem?
> I don't think so, the original poster thought that this feature is worth a bit of money, someone else agreed and increased the bounty. I don't know if this pull request would exist without the bounty, but I also don't think it matters. It's awsome that @JustVic contributed some time to get the feature implemented, accepting the requests for some changes and helping to get things built correctly. From my side this is ready to be merged (hence the review approved mark), waiting for the ok from @kintel on the MacOS side. Once the pull request is merged, we have a nice little feature in OpenSCAD, the bounty is earned and hopefully everyone who contributed to the bounty is happy to use the new feature. So while things can get complicated once money is involved, I totally can't see that here in this specific case.

Lenbok:
I'm pretty much in agreement with you, and that's a much better answer than "This issue has a Bounty attached, and work has been done" as a deciding criteria for pull requests being accepted. Just because a bounty is attached shouldn't mean that the PR should get a free pass compared to other issues and pull requests. 

I wouldn't use an audio notification myself, but having been struggling with some extremely slow openscad scripts over the weekend, could see the appeal of some other notification method (I've used this ability in all of octoprint, slic3r, and kisslicer).

Cheers!

kintel:
@Lenbok @MichaelAtOz @t-paul My only remaining question is whether a "cleaner" sound is more appropriate? The current sound feels a bit like a warning, IMO.

t-paul:
I think the current sound is ok. For me it does fit a "complete" signal and the length (time wise) seems quite appropriate. There are other good ones in #2394 - I guess we can have a look later and add a couple more for selection, or if you really prefer a different one for the first go, swapping the file (and attribution) is a quick thing to do :-).

MichaelAtOz:
@kintel I agree re the current sound.

There are plenty to choose from, I think it has narrowed down to (in #2394) C. Powerup/success.wav, 0.9s 78K, or 'OpenSCAD DTMF4' (11K), the earlier DTMFs are too high/grating IMO. (I loaded the DTMFs to freesound so they can be attributed easily)

Note that the sound is embedded in the exe, so should be small, see [here](https://github.com/openscad/openscad/issues/2375#issuecomment-386926320).

Over to you blue leader.

Also note, prior to an OpenSCAD release, I think this should be extended to allow sound files, so users can customise it. That was the basis of #2394.
(@t-paul @dorkmo)

JustVic:
@Lenbok  If you want I can try to make variant with user shell command on complete, but I think it must be separate issue. And I currently I don't know how it will work on Windows.

MichaelAtOz:
> how it will work on Windows.

This may help, how openscad.com spawns openscad.exe [Windows,](https://github.com/openscad/openscad/blob/master/winconsole/winconsole.c)

t-paul:
I agree that other types of notification should be handled via a different issue. Running external programs would ideally go though Qt as we are talking GUI feature (notably making it unavailable for command line usage too, same as with the current sound output - which is a good thing). That would at least solve the cross platform topic partially.

One thing to consider starting that is that it will be much more complicated to get that done, as there are quite some additional cases to solve (handle blocking external calls, handle lots of calls in short time, ...).

So if someone really wants that feature, please create a new issue for that, maybe including a specific example use case. However, I would vote for not implementing it just because we can.

MichaelAtOz:
> I changed sound to Powerup/success.wav

@kintel r u happy with that?
If so this should be ready for merge, or do you want further tests? I can't think of automated tests.

MichaelAtOz:
BTW - not for this change, but is there a test of the cmd line help text?

t-paul:
I don't think there's something for the help text. What would that check?

Lenbok:
@JustVic Thanks for the offer, I'd put it pretty low down the wish list though, given there would be several additional complexities to do it nicely from the sounds.

kintel:
Thx!

t-paul:
@MichaelAtOz what did you mean with the TODO? AFAICS the about text is updated, so should be all set. Except maybe mentioning the new feature in the manual.

Also as it's closed now, I guess we can finish the off-topic discussion :)....

Checking if the command line help is unchanged does not feel very useful as it's currently a separate text, so that would not just change without being specifically touched. I guess the useful check would be if it *not* changed when modifying options. I can't think of a test case to cover that. It probably would need the code to actually use a single source option description (and that would make a test case unnecessary).

donbright:
most computer systems, since the original macintosh in 1984, have sounds already bundled with the OS that are designed purposely to alert users to various things. the users are able to enable/disable this at the system level.

MichaelAtOz:
I imagine the work to integrate the Windows, Mac & Linux*n OS APIs used to select which 'beep' is non trivial.

dorkmo:
its kind of a miracle you can ask a program to play a wav file and itll just work on any operating system

donbright:
this broke the windows cross-build on my box.. havent quite figured out how to get MXE to use qt multimedia. anyone?

JustVic:
@donbright maybe it just needs add qtmultimedia to  openscad/scripts/mingw-x-build-dependencies.sh:

if [ "`echo $* | grep download`" ]; then
  PACKAGES='download-mpfr download-eigen download-opencsg download-cgal download-qtbase **download-qtmultimedia** download-glib download-libxml2 download-freetype download-fontconfig download-harfbuzz'
 else
  PACKAGES='qtbase **qtmultimedia** qscintilla2 mpfr eigen opencsg cgal glib libxml2 freetype fontconfig harfbuzz'
 fi
else
 MXE_TARGETS='i686-w64-mingw32.static.posix'
 if [ "`echo $* | grep download`" ]; then
  PACKAGES='download-mpfr download-eigen download-opencsg download-cgal download-qtbase **download-qtmultimedia** download-nsis download-glib download-libxml2 download-freetype download-fontconfig download-harfbuzz'
 else
  PACKAGES='qtbase **qtmultimedia** qscintilla2 mpfr eigen opencsg cgal nsis glib libxml2 freetype fontconfig harfbuzz'
 fi
fi

JustVic:
I did't try windows cross-build myself. It also needs add qtmultimedia to ./scripts/release-common.sh too. I will try to build it and see what happens...

nophead:
The sound does not play correctly on my Win7 machine. It seems to be split into three sections with short pauses in between.

MichaelAtOz:
It plays fine on mine.
Was your system loaded? Sound is async and scheduled as a normal priority process.

MichaelAtOz:
Also check you sound driver version??

nophead:
It does it with just OpenSCAD open but only when I have two documents open in multiple document mode, one of them being very big (my entire parts library, which is sluggish to draw in preview mode). With that open OpenSCAD is using about 15% CPU on a four core system.

This is an MSYS2 build.

The driver appears to be a generic Microsoft one called "High Definition Audio Device". It says it is up to date but is is from 2010.

I can play other sounds OK with OpenSCAD open, for example system sounds and YouTube videos. It seems to be just OpenSCAD interrupting itself when it is drawing a complex scene.

If I run a second instance the sound is OK. It is only multiple document mode that is the problem.

nophead:
Turning off automatic reload on my library test fixes it as well. Seems like the sound needs the idle loop to be run periodically to keep it going.

MichaelAtOz:
How hard would it be to do the sound in another thread?

nophead:
CGAL runs in its own thread and then signals to the routine that shows the stats and plays the sound. It is probably easy to move the sound into the CGAL thread.

The CGAL thread confuses me because it doesn't behave like a thread on Windows. The GUI is locked up most of time while rendering it only seems to process events every few seconds, perhaps when the progress bar moves. You can't really use the GUI while it is rendering.

MichaelAtOz:
> The GUI is locked up most of time while rendering it only seems to process events every few seconds, perhaps when the progress bar moves. You can't really use the GUI while it is rendering.

Not how it happens for me.
GUI is usable.

> when I have two documents open in multiple document mode,

I'll try MDI tomorrow.

nophead:
My current machine is W7-64 and it does have an SSD but OpenSCAD has always behaved like this for me, not matter how big or small the project is. I.e. the GUI only responds to events after many seconds while doing F6. Possibly just when the progress bar updates.

nophead:
Actually scratch that. I can use the GUI in F6. It is during F5 it behaves as described.

MichaelAtOz:
Well, F5 is not a separate thread, I suspect few of us push F5 with large models, and AFAIK this is reflective of your graphics subsystem. On my older PCs with aged GPUs it is tedious, newer with IIRC GTX 1060Ti, it is snappy.
I wonder if Preferences/Advanced setting are relevant.
I have turn off rendering set to 1000000000. Can't remember why, possibly my large multi-thread testing.

MichaelAtOz:
and of course F5 doesn't play sounds...

nophead:
I wonder if F5 can be done in a thread in exactly the same way as F6. In my assembly views I render everything, so it spends most of time in CGAL anyway. When I do F6 it is very fast because it has already cached the result when it was making the assembly view.

MichaelAtOz:
Threading F5 may be good, noting that with MDI you want the GUI not to be stalled.
Also note with MDI you can only do one F5/6 at a time.
I don't think lengthy previews was considered when MDI was introduced.

Have you tried running non-MDI with separate invocations, ie multiple exe processes.
I suspect with your Auto reload/preview setup that would be better.

MichaelAtOz:
..tho they would be isolated caches, but assuming you have lots of RAM that would be good after initially populating it.

PR #2438, opened 2018-07-29
Improvements for the Appdata file

cosimoc:
Add OARS content rating and basic release information.
See also https://github.com/flathub/flathub/pull/510


t-paul:
Why all the `none` entries in the content rating? Looking at the [data generator](https://hughsie.github.io/oars/generate.html) that seems to be identical to just

```
  <content_rating type="oars-1.1" />
```

cosimoc:
Thanks for the review!
I had copied that from another project instead of using the generator, which seems to have been updated in the meantime to output the more compact representation.
I updated the PR to use that here as well.

t-paul:
Thanks, I could not find much more explanations about the values, but as OpenSCAD is pretty much an offline application at this point, that seems ok.

PR #2447, opened 2018-08-15
[WIP] Use more automated config for cross-building with mxe (née mingw-cross-env)

tonytheodore:
Basically removes manually specified options and mxe-specific sections in favour of qmake/pkg-config to set up the build.


t-paul:
Imported the changes into branch mxe-gcc7. Will continue in #2452 with additional changes for the build scripts.

Big thanks, @tonytheodore. I could successfully build via gcc7 now.

PR #2454, opened 2018-08-21
Adds "record" data type

hexane360:
I've added support for simple, untyped key-value stores. Currently, records are created like `{key: "value", e: [1,2,3,4], f: 5}`, and objects are accessed with dot syntax. Under the hood, I use std::unordered_map.

When coerced to a boolean, only the empty record returns false. The `+`, `-`, and `*` operators are overloaded to refer to set union, difference, and intersection respectively, with values taken from the second operand. Currently, there is no way to iterate over records, but it's worth looking into.

I plan to use this feature as a more readable and extensible alternative to long lists of kwargs. For example, take a module like this:
```
module example(args) {
    defaults = {
        radius: 5.0,
        height: 10.0,
        points: [[0,0], [1,0], [1,1], [0,1]]
    };
    settings = defaults + args;
    echo(settings.points);
}
example({radius: 8.0});
```
The record syntax also allows the use of more complicated expressions than function argument defaults.

Another obvious use is for basic data objects, beyond what makes sense to use arrays or xyz object access for:
```
connector = {h: 3.0, w: 8.0, l: 30.0};
flange = {h: 3.0, r: 30.0};
function volume(cube) = cube.h * cube.w * cube.l;
volume(connector);
```

### DIFF
--- a/src/expr.cc
+++ b/src/expr.cc
@@ -371,6 +371,45 @@ void Vector::print(std::ostream &stream, const std::string &indent) const
 	stream << "]";
 }
 
+Record::Record(const Location &loc) : Expression(loc)
+{
+}
+
+bool Record::isLiteral() const {
+    for(const auto &e : this->children) {
+        if (!e.second->isLiteral()){
+            return false;
+        }
+    }
+    return true;
+}
+
+ValuePtr Record::evaluate(const Context *context) const
+{
+	Value::RecordType rec;
+	for(const auto &e : this->children) {
+	    rec[e.first] = e.second->evaluate(context);
+	}
+	return ValuePtr(rec);
+}
+
+void Record::print(std::ostream &s, const std::string &indent) const
+{
+    s << '{';
+    bool first = true;
+    for (const auto &val : children) {
+      if (!first) {
+        s << ", ";
+      } else first = false;
+      s << val.first << ": " << *val.second;
+    }
+    s << "}";
+}
+
+void Record::add(const char *key, Expression *expr) {
+    children[std::string(key)] = shared_ptr<Expression>(expr);
+}
+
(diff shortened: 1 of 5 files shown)


nophead:
Very nice. It would certainly simplify a lot of my code. Things like:
```
washer_diameter(screw_washer(screw))
```
would become:
```
screw.washer.diameter
```

The only downside I see is it would make it difficult to add dictionaries to OpenSCAD as it has taken over the syntax normally used. The difference being the keys are runtime values rather than compile time names.

t-paul:
I'd suggest to change the syntax slightly to match the [Object Literals](https://github.com/doug-moen/openscad2/blob/master/rfc/Objects.md) - that way we could later extend it to support the other features described there.

donbright:
any regression tests / examples for this?

t-paul:
@hexane360 do you still have time working on this feature?

hexane360:
@t-paul Not as much as in August, :P 
This was more of a quick mockup of an idea I had. Object literals are definitely a more beautiful representation of the same idea. However, I'm not sure how the set operation overrides would mesh with object literals. Would it make more sense to be two completely different objects with different syntaxes?

t-paul:
I'm not sure there is a need for multiple syntaxes. Just the object literal alone will not give the full feature set, but I think together with the object customization, it's almost there.

With something like the [unwrap operator](https://github.com/openscad/openscad/wiki/OEP4%3A-Unwrap-Operator) it should pretty much allow everything from the set operations too.

kintel:
I believe this has been superseded by the internal object support used by textmetrics and JSON import

jordanbrown0:
Concur; the syntax is the same as in #4478, except that I haven't overloaded the arithmetic operators like this.  (But object comprehension can be used for most of those use cases.)

PR #2455, opened 2018-08-21
making updateFindIndicators side effect free

### LINKED ISSUE #2440: Edit/Find starts immediate search
I have a very annoying issue with the implementation of the Find command. Here are the steps

- in some file open the Find with CMD-F and
- type the text you want to search and it will find it (nice so far).
- Now close the Find with the button to the right ("Fertig" in German, probably "Ready" in English) and it will close the Find window.
- Place the cursor to some other position (e.g. on the next page of your document).
- Now when you press CMD-F it will jump right to the next occurrence of the previous Find.

This is very annoying (because I intend to search from the current cursor position). CMD-F should only open the Find window and not perform a find by itself. This should only start on a CMD-G. Or after pressing the Next button in the Find window.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/61783807-edit-find-starts-immediate-search?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

MichaelPFrey:
This is an attempt of making updateFindIndicators (currently resetFindIndicator) side effect free.
This should fix #2440

### DIFF
--- a/src/editor.h
+++ b/src/editor.h
@@ -19,7 +19,7 @@ public:
 	virtual QString toPlainText() = 0;
 	virtual QTextDocument *document(){QTextDocument *t = new QTextDocument; return t;}
 	virtual QString selectedText() = 0;
-	virtual int resetFindIndicators(const QString &findText, bool visibility = true) = 0;
+	virtual int updateFindIndicators(const QString &findText, bool visibility = true) = 0;
 	virtual bool find(const QString &, bool findNext = false, bool findBackwards = false) = 0;
 	virtual void replaceSelectedText(const QString &newText) = 0;
 	virtual void replaceAll(const QString &findText, const QString &replaceText) = 0;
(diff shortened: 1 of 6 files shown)


MichaelPFrey:
This implementation should be much cleaner and efficient.

I will do some testing and clean up, but I think this is pretty much the best strategy to make resetFindIndicator side effect free.

What I am also thinking about is to move `this->findInputField->setFindCount(...);	`  into updateFindIndicators, as updateFindIndicators is never called without updating the FindCount afterwards.

MichaelPFrey:
so, now I am pretty happy now.

Just for reference:
https://qscintilla.com/indicators/
http://pyqt.sourceforge.net/Docs/QScintilla2/classQsciScintillaBase.html

Issue #2251, opened 2018-01-05
macOS Build fails on High Sierra .2


SConaway:
OpenSCAD fails to build on 10.13.2. I am happy to provide any other necessary details. I'm spinning up a 10.13.2 VM to test.

Log:

````
$ ./scripts/macosx-build-homebrew.sh
Thu Jan  4 21:19:20 PST 2018: Listing homebrew configuration
HOMEBREW_VERSION: 1.4.2
ORIGIN: https://github.com/Homebrew/brew
HEAD: 4e08da95c34c27ba6d5ca5f80718fe2788337cdf
Last commit: 4 days ago
Core tap ORIGIN: https://github.com/Homebrew/homebrew-core
Core tap HEAD: a4e3600a5d38576036d79406344c5c34e1f4afbd
Core tap last commit: 8 hours ago
HOMEBREW_PREFIX: /usr/local
HOMEBREW_NO_ANALYTICS_THIS_RUN: 1
CPU: octa-core 64-bit haswell
Homebrew Ruby: 2.3.3 => /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/bin/ruby
Clang: 9.0 build 900
Git: 2.15.1 => /usr/local/bin/git
Curl: 7.54.0 => /usr/bin/curl
Perl: /usr/bin/perl
Python: /usr/local/opt/python/libexec/bin/python => /usr/local/Cellar/python/2.7.14_1/Frameworks/Python.framework/Versions/2.7/bin/python2.7
Ruby: /usr/local/bin/ruby => /usr/local/Cellar/ruby/2.4.3_1/bin/ruby
Java: 9.0.1
macOS: 10.13.2-x86_64
Xcode: 9.2
CLT: 9.2.0.0.1.1510905681
X11: 2.7.11 => /opt/X11

real	0m4.180s
user	0m1.481s
sys	0m1.945s
Thu Jan  4 21:19:24 PST 2018: Updating homebrew
Already up-to-date.

real	0m1.514s
user	0m0.696s
sys	0m0.581s
Thu Jan  4 21:19:29 PST 2018: Installing formula eigen
eigen 3.3.4
Warning: eigen 3.3.4 is already installed

real	0m1.660s
user	0m0.774s
sys	0m0.576s
Thu Jan  4 21:19:32 PST 2018: Installing formula boost
boost 1.66.0
Warning: boost 1.66.0 is already installed

real	0m1.627s
user	0m0.762s
sys	0m0.557s
[...]

Issue #2219, opened 2017-12-15
Unable to make the project: src/libsvg/libsvg.cc:7:10: fatal error: 'libxml/xmlreader.h' file not found


ViliusKraujutis:
I'm unable to `make` OpenSCAD. I'm getting this error:
```
src/libsvg/libsvg.cc:7:10: fatal error: 'libxml/xmlreader.h' file not found
```

<img width="562" alt="screen shot 2017-12-16 at 01 03 50" src="https://user-images.githubusercontent.com/1859636/34063787-01e6af7c-e1fd-11e7-9e9d-dd639da26a5a.png">

You can see the console log here: 
https://gist.github.com/ViliusKraujutis/7f51abffcc855110603bcdd945b3eb46

I'm on Mac OS High Sierra.
I had OpenSCAD installed before.

t-paul:
This needs `libxml2`. The dependency build should already fetch and build that.

ViliusKraujutis:
So looking [here](https://coderwall.com/p/t7obea/installing-libxml2-for-python-in-mac-os-x), I'll be able to install it with `brew` like this:

    brew install libxml2

Right?

ViliusKraujutis:
Anyways, after installing and linking, I'm still unable to `make` OpenSCAD application on Mac:
<img width="1183" alt="screen shot 2017-12-16 at 08 51 37" src="https://user-images.githubusercontent.com/1859636/34068075-75d83ae0-e23e-11e7-915f-8a990a1fceac.png">

To install `libxml2`, I've done like this:
<img width="663" alt="screen shot 2017-12-16 at 08 52 52" src="https://user-images.githubusercontent.com/1859636/34068084-96bfac8e-e23e-11e7-9dac-3d62dea8e9b7.png">

Any ideas what's wrong now and how to resolve this issue?

thehans:
Did you try looking [here](https://github.com/openscad/openscad/blob/master/README.md) first?

There is a section called "Building for Mac OS X".  If you follow those instructions does it still fail?

kintel:
The python stuff is a red herring - we don't use any python code to interface with libxml2.

The problem seems to be that libxml2 in Homebrew is "keg only", meaning it won't be installed in the normal location. ..so we need to tell the build system where to pick it up.
As Homebrew suggests during install, you could try to set the `PKG_CONFIG_PATH` env variable to `/usr/local/opt/libxml2/lib/pkgconfig`.

[...]

PR #2457, opened 2018-08-23
Homebrewhighs

donbright:
this is a fix for issue #2219 

basically homebrew is broken on two things. first, libxml requires pkg-config to look in homebrew,

second, homebrew's qscintilla2 is not found by current openscad pri / prf files.

this is repaired by altering the process so that homebrew will now need its own 'setenv' script, like the other builds,  and users will have to manually run it, instead of homebrew being automatically detected like before. it incorporates @kintel suggestion of setting PKG_CONFIG_PATH but now that is just part of the "standard process" of a homebrew build.

IMHO this is a little bit 'cleaner', for example there is no -I/usr/include anywhere in the Makefile so its hopefully not trying to pull in system versions of packages like libxml2.  and it  allows quicker patching of future homebrew issues, for example if new dependencies require new "PKG_CONFIG_PATH" alterations we can just update the setenv_homebrew.sh script

(that mpfr fix sneaked in there too, my bad, shouldnt hurt anything tho)


kintel:
@donbright Looks like this broke finding boost on homebrew. I haven't looked into the details.

t-paul:
I think it's the CGAL 4.12 update on Travis which broke that.

I hacked around that problem in master with https://github.com/openscad/openscad/commit/a24ecaec107206e1523a97910db4dab152bb578b - I don't have a working env to find a proper fix.

kintel:
@t-paul thx - I've merged with master -let's see...

kintel:
@donbright Looking at this more closely, it looks like the problem is that the new `source scripts/setenv_homebrew.sh` isn't run in the Travis environment.

t-paul:
Yeah, the issue I mentioned above did hit the test suite / cmake build, the qmake application build was ok.

donbright:
dang sorry, i forgot that travis did OSX now

donbright:
i added a fix or two for test build... 

this btw might also fix #2251  although i am on a slightly  newer high sierra IIRC

donbright:
ok so i patched a few things and now it builds on travis... and the tests pass (!)

interesting that on my actual machine a bunch of the tests fail ....

scintilla.pri
```diff
@@ -18,6 +18,13 @@ scintilla {
     }
   }
 
+  BREW=$$(OPENSCAD_HOMEBREW)
+  !isEmpty(BREW) {
+    include(/usr/local/opt/qscintilla2//data/mkspecs/features/qscintilla2.prf)
```
kintel: Below there's a comment that "qscintilla2.prf which ships with QScintilla is broken for Mac", but here we include it nevertheless.
The broken aspect appears to be that they try to link with `-lqscintilla2_qt5_debug` in debug mode, but don't ship that library.

Issue #2439, opened 2018-08-03
Customizer: Out of place tab if double allocation of a variable of this group


PRouzeau:
label [Customizer]
Bug:
If a variable is allocated a first time, then the same variable is allocated a second time within a customizer group, the group tab is moved to the position of the first allocation, whatever its original position is supposed to be. In principle, it is not normal to allocate a variable twice when using customizer, but it can be difficult to track it without a message. This bug may not need to be removed, provided a message about the double allocation is sent.

MichaelPFrey:
Code showing the issue:
```
a=5;
/* [Group 1] */
a=10;
b=10;
/* [Group 2] */
a=20;
```
screenshot:
![bildschirmfoto von 2018-08-25 08-09-02](https://user-images.githubusercontent.com/24962768/44615603-3971d400-a83e-11e8-8e68-94f4ea988864.png)
hmmm...

t-paul:
Reassingment should ideally be an error. Right now it's not even a warning. It's not as easy as it sounds as there are some usecases that historically mis-use that feature (e.g. the -D command line parameter).

This also seems to be a duplicate of #1770.

MichaelPFrey:
every assignment knows its location:
`const Location loc =assignment.location();`
which includes `firstLine()`. `ParameterExtractor::setParameters` gets the assignments in order of first assignment, but with the information of the last assignment. We could reorder the assignments by location. That would at least fix the "out of place" placement of the group. std::sort with a custom function could be used for that.

@t-paul good catch - I can see it in two ways: Fixing #1770 also solves #2439. But #2439 could be solved/worked around without resolving #1770, as I see some issues with the customizer code as well.
Well, I start working on #1770. I see away to deal with include and the -D command line parameter.

t-paul:
Right, good point. Let's leave this open then as separate topic to sync data and location of the assignments.

MichaelPFrey:
Testing my patch, I found `value-reassignment-tests.scad`. This test case has a similar root cause to this issue. But thinking about it, it shows how complicated rearranging the statements would get.

Issue #1770, opened 2016-08-12
Assignment of Variables


amarjeetkapoor1:
OpenSCAD supports that we can re-assign the same variable with some different value anytime in SCAD file but It uses only the value that is assigned at last everywhere in rest of the file. So, What is use of supporting Re-Assignment of variable as its create confusion. 

```
a=12;
echo(a);
a=45;
```

Above code should output 12 instead of 45.
It should be either that we use value of variable till it is Re-Assign or we should give error when variable is Re-Assigned.

Plus there are several other abnormalities related to assignment like 

```
a=14;
b=12;
a=b+12;
echo(a);
```

and

```
a=14;
a=a+12;
echo(a);
```

both are gives undef value.

There are also problem with storing location of assignment in ASTnode like:

```
a=12;
a=14;
```

As at present we support re-assignment. So, the value of `a` =14 for whole program. but location of ASTnode remains 1. It is not updated when `a` is re-assigned.

doug-moen:
I agree this is a bug.

OpenSCAD is a declarative language, which means that by design, it is also a "single assignment" language: each variable can be assigned exactly once, within a given scope. So a program like
```
a=12;
a=14;
```
*should* be illegal.

However, there are two problems: a lack of error reporting, and the `include` mechanism.

One problem is that OpenSCAD does not, in most cases, give error messages for illegal programs. We give error messages for parse errors, because bison doesn't give us any alternative. Ditto for CGAL errors. For illegal programs like the one above, it's less code to not check for and report an error, so we just keep running. I think that, to reduce confusion, we should print a meaningful error message and abort in cases like this.

The other problem is that the `include` mechanism allows the script that contains `include <S>` statement to override variables imported from `S`. So, if `S` defines variable `a`, then you can write
```
include <S>
a=16;
```
and override the definition of `a` within the code imported from `S`. This is a feature that people use.

The problem is that the `include` statement is currently implemented by the lexical analyzer, and we can't currently distinguish an assignment to `a` imported from `S` from an overriding assignment to `a` in the current script. Hence we don't have enough information in the parse tree to be able to detect and report an error where there are two assignments to `a` in the same scope.

@kintel has a branch where he is fixing the parse tree nodes to keep track of the source file in which they originated. Once that is integrated into the main branch, we'll have the necessary data structure so that we can fix this bug and report an error for duplicate assignments.

t-paul:
[...]

PR #2460, opened 2018-08-26
warn, when a variable from the root module is overwriten in the root module

MichaelPFrey:
addressing #2439 #1770.

This still allows the root module to overwrite variables from includes without warning (which is an indented feature) and includes to overwrite variables from the root module without warning (which we can discuss).
It also avoids warnings from issues within used and inlcuded files.

The command line option `-D ` is an issue with that approach.
In the ModulCache, the command line is simply appended to the rootmodule:
`		textbuf << "\n" << commandline_commands;`
making the patch fail `openscad-override_override`. I mean I could regenerate the test, but the warning is not really acceptable:
`+WARNING: a was assigned on line 2 but was overwritten on line 6` as the scad file has only 5 lines. the 6th line are the appended `commandline_commands`.

### DIFF
--- a/src/ModuleCache.cc
+++ b/src/ModuleCache.cc
@@ -101,7 +101,7 @@ time_t ModuleCache::evaluate(const std::string &filename, FileModule *&module)
 			}
 			textbuf << ifs.rdbuf();
 		}
-		textbuf << "\n" << commandline_commands;
+		textbuf << "\n\x03\n" << commandline_commands;
 		
 		print_messages_push();
 		
--- a/src/lexer.l
+++ b/src/lexer.l
@@ -104,7 +104,6 @@ std::vector<std::string> openfilenames;
 
 std::string filename;
 std::string filepath;
-
 %}
 
 %option yylineno
@@ -205,6 +204,8 @@ use[ \t\r\n]*"<"	{ BEGIN(cond_use); }
 	if (!YY_CURRENT_BUFFER)
 		yyterminate();
 }
+
+"\x03"		return TOK_EOT;
 
 "module"	return TOK_MODULE;
 "function"	return TOK_FUNCTION;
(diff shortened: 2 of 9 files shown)


MichaelPFrey:
bildschirmfoto von 2018-08-26 08-23-12

MichaelPFrey:
using a special marker (currently \x03 which is End of Text) as separator to the `commandline_commands` works. Unprintable characters should not normally appear in a normal SCAD File.

kintel:
This is indeed a common confusion, and a warning would be helpful.
I wonder if, instead of using a magic marker, we could parse the included files separately, and hence implement this test for all scopes. I'll think about it - this PR may still be good in the interim, as it may take a while to pull of the include-file-compile approach...

MichaelPFrey:
Note that I only need the magic marker for `commandline_commands`, as giving warnings for intentional overwrites on line numbers that do not exist would be silly. [this build](https://travis-ci.org/openscad/openscad/jobs/420683511) failed intentionally on test `1079 - openscad-override_override` - to pass it (by not outputting a silly warning), I added the magic marker. The alternate approach would have been to regenerate the test to include the silly a warning.

Handling command line parameters differently would be indeed the better approach then using a magic marker. The man page states:
> -D var=val
> This option can be used to assign constant values to OpenSCAD variables.
> ...
but looking at the code, any OpenSCAD expression would be valid. So, the current-way we handle `commandline_commands` can be seen as a bug in many ways. (But finding a better implementation is also very very tricky and some users might be exploiting the currently undocumented behavior, so I am careful in touching that)

Currently, I intentionally check only for duplicate assignments in the root module:
https://github.com/openscad/openscad/pull/2460/files#diff-ad30922032159ce4721ce96984d14937R212
`if(assignmentWarning && rootmodule->getFullpath()==LOC(@$).fileName()){`

I think that could be reworked to also work for includes and give out meaning full warning, but that being said:
- There are situation, where a user wants to overwrite included variables
- plenty of times, the user does not care when there is a multiple assignment issue within "used"/"included" code (and even if: the user can always check individual scad files)
- we have to add code to calculate relative files paths in order to not create very long warnings (I can borrow code from the assertion warnings)

So, that needs a lot more care and considerations, while duplicate assignments in the root module clearly deserve a warning (Our own test files having duplicate assignments being an indicator) which also does not require too much code.

PR #2465, opened 2018-08-28
Update mainwin.cc

t0b1as:
Had an error during compilation, was fixed by this change.


t-paul:
Thanks, checking the [docs](http://doc.qt.io/qt-5/qsound.html), this is the correct way to include `QSound`.

PR #2471, opened 2018-08-31
AMF bug fix for Windows

nophead:
No longer uses filsystem::path for XML element path.

### DIFF
--- a/src/import_amf.cc
+++ b/src/import_amf.cc
@@ -324,21 +324,26 @@ int AmfImporterZIP::close_callback(void *context)
 	return zip_fclose(importer->zipfile);
 }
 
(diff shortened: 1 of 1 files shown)


MichaelAtOz:
Reminder: This needs checking on linux/mac '/' related stuff has bitten in the past.

t-paul:
Travis covers both MacOS and Linux, so we should be good.

src/import_amf.cc
```diff
@@ -91,7 +90,7 @@ class AmfImporter {
 	virtual xmlTextReaderPtr createXmlReader(const char *filename);
 };
 
-AmfImporter::AmfImporter() : path("/")
+AmfImporter::AmfImporter() : xpath("")
```
kintel: strings are default-initialized to empty, so this is redundant

PR #2473, opened 2018-08-31
Added in render() test of $preview.

nophead:
Proves render() has no effect.


kintel:
render() does have an effect; the object will be rendered using CGAL, which, as a side effect, removes any specified color. It does not change the value of $preview mid-script.
Good test to have though.

Issue #2450, opened 2018-08-17
Customizer: Creation of a group [Reserved] for recorded data not shown to user


PRouzeau:
Tag: [Customizer] [New Feature]
Variables under the [Hidden] group are no longer recorded, which is fine.
However, you may want to record some variables without showing them to the user.
These special groups which may contains data for future development, data record version, language, unit coefficient,etc. and may be named either (or both)  [Reserved] or [Future]

t-paul:
Why would that need any special feature?

MichaelPFrey:
You mean recorded as in being stored in the .json file?

My concern is, that stored hidden variables can cause a confusing state that can not be seen nor be fixed within OpenSCAD.
Coming from that, I would only see a point in "read only variables", that are shown in the Customizer UI, stored in the JSON File, but can not be changed in the Customizer UI. But I do not know what "read only variable" is supposed to mean, as in: Does the SCAD File take priority or the JSON File? What if the user "needs to change" the read-only variable for some reason?

The question is still: What is the use case? I know that the use cases evolve once a feature exists, but having a specific use case in mind while developing makes the feature more consistent.

Keeping track which JSON File Version belongs to which SCAD File Version is best done by using a Versioning System (Git/SVN/Mercurial/...).

If someone shares a json file without the scad file: Well that is an issue and knowing the Version of the SCAD file that created the json File would help. But always sending scad & json file would be easier.

If someone upgrades the scad file, but uses an old json file:  Knowing the Version of the SCAD file that created the json File could allow the scad file to use some logic - but that gets messy quickly.
New features being off by default would be easier.

Last but not least, we need to keep the thingiverse customizer in mind.

PRouzeau:
Use case is for version info and other parameters which shall not yet be modified by users.
Example: On my last app, I created the tab named "_reserved-do not modify" (need to be the first due to the parameters involved) 
which contains:
*version number 
*language digram
*unit coefficient
*another unit coefficient
![_private_tab](https://user-images.githubusercontent.com/11719573/44881675-c094d100-acb0-11e8-9388-5d2ac1f81528.png)

t-paul:
Why have them in customizer then?

PRouzeau:
To be recorded in the dataset.
In fact, this is a provision for future evolution

t-paul:
I don't know what that means. Non-editable values are in the *.scad file, why would those need to be somewhere else too?

MichaelAtOz:
[...]

PR #2478, opened 2018-09-06
save hidden variables, but to not apply them (see #2450)

MichaelPFrey:
there are for different possible implementation for hidden variables:

1. hidden variables are stored and retrieved from the json file
1. hidden variables **NOT stored** in json file, but are retrieved from the json file
1.  hidden variables are stored in the json file, but are **NOT retrieved** from the json file
1. hidden variables are **neither** stored and  nor retrieved from the json file

Originally, the OpenSCAD customizer has behaved as in 1.

Currently, the OpenSCAD customizer behaves as in 2. The customizer applying variables from the hidden group when found in a json file does not make sense.

Implementing 3 would be a reasonable compromise. Hidden variables are not applied, but still stored. Meaning: If a once hidden variable becomes visible, it also becomes applicable. This allows a designer to use the hidden section for reserved variables, that become customizable (and assigend with a different default) in a future version, without breaking existing preset. 

Implementing 4 would be consequent.

This is the implementation of 3.

### DIFF
--- a/src/parameter/ParameterWidget.cc
+++ b/src/parameter/ParameterWidget.cc
@@ -459,11 +459,9 @@ void ParameterWidget::updateParameterSet(std::string setName)
 
 		pt::ptree iroot;
 		for (const auto &entry : entries) {
-			if (entry.second->groupName != "Hidden") {
-				const auto &VariableName = entry.first;
-				const auto &VariableValue = entry.second->value->toString();
-				iroot.put(VariableName, VariableValue);
-			}
+			const auto &VariableName = entry.first;
+			const auto &VariableValue = entry.second->value->toString();
+			iroot.put(VariableName, VariableValue);
 		}
 		setMgr->addParameterSet(setName, iroot);
 		const QString s(QString::fromStdString(setName));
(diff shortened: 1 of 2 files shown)


MichaelPFrey:
This allows a designer to use the hidden section of the customizer for "reserved" fields.

An example use case would be, that the designer plans on implementing a specific feature (e.g. Moutning holes), but as it is not implemented yet, he does not want to show it.

Using [Hidden], the designer can default it to none, until he implements that features.
Old presets read by a newer SCAD file that implements "mounting holes" will still use no mounting holes for the old presets, as that is what is stored in the json.
Meanwhile, it allows the scad file to have different default, like "GoPro Mounting holes", for newly created presets.

(Yes, this feature requires s a lot of discipline and care from the designer to fully utilize it. But it is not even a problem when the user does not know that this features even exists. The variables are only applied when they are visible, so the user can always fix mistakes later)

MichaelPFrey:
documentation updated:
- https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/WIP#[Hidden]
- https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/WIP#hidden_parameters

PR #2479, opened 2018-09-07
adding a "design default values" parameter set

MichaelPFrey:
* adding a "design default values" parameter set. 
   * Always reset to the "design default values" before applying any preset.
      -> this should prevent various issues like variables from one parameter set being transferred to an other parameter set that those not have the specific parameter has not set.

### DIFF
--- a/src/mainwin.cc
+++ b/src/mainwin.cc
@@ -909,7 +909,7 @@ void MainWindow::refreshDocument()
 			PRINTB("Loaded design '%s'.", this->fileName.toLocal8Bit().constData());
 			if (editor->toPlainText() != text) {
 				editor->setPlainText(text);
-				this->contentschanged = true;
+				setContentsChanged();
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


MichaelAtOz:
Is the default set something that needs to be saved to work?

MichaelPFrey:
MichaelAtOz:
> Is the default set something that needs to be saved to work?

It does not need to be saved.
It is just always there on the top of the list of presets [in the List index 0].

The implementation is simple: Set all entries to the default value, then apply the selected preset.

"design default values" [in the List index 0] (feel free to suggest a better name) is not a real preset.
So, it sets all entries to the default value, sees that it is preset index 0 and those nothing.

It pretty much behaves just like an other preset.

There are a few quirks with this implantation:
   * Changes to the scad source code apply to the "design default values" only after "Preview"
      * upps... testing that, I realize that it crashes when I change the scad source file
   * the "reset" button does not work for "design default values"  yet

And when I am already on it:
   * the "reset" button should have a tool tip explaining that it "resets the currently selected preset"
   * at a few comments to explain what each method does (the naming is getting confusing)

well, back to work

PR #2488, opened 2018-09-20
Lighting fix

nophead:
Fixed the vector camera lighting to be from above, rather than below, so that it now matches the GUI. Most of the tests images have been updated.


nophead:
Fails the throwntogethertest_issue1803 test on OSX only. Seems to be a z-fighting difference between my machine and the the Travis server. It doesn't seem to run that test on the trusty build.

nophead:
Superceeded.

PR #2489, opened 2018-09-20
Viewall without autocentre fix

nophead:
Gimbal viewall without autocenter (only available from the command line) didn't take the camera offset into account. 

Note this is based on my lighting fix PR because the test images depend on that.


nophead:
Fails on OSX for the same reason as #2488

nophead:
Superseded.

PR #2490, opened 2018-09-21
Unify views

nophead:
The GUI default view is now the same angle as the command line, i.e. towards the first quadrant.

The altitude angle has been rounded to 20 degrees to avoid fractions in the GUI so most of the test images are slightly different.

Negative 0 no longer appears on the GUI.


kintel:
One concern I have with this is that people may have purposefully created designs that look best from the default view angle, e.g. objects that have a more natural "front" and "back".
As an experienced designer, what's your take on that?

tests/CMakeLists.txt
```diff
@@ -1701,9 +1704,9 @@ add_failing_test(offfailedtest EXE ${PYTHON_EXECUTABLE} SCRIPT ${CMAKE_SOURCE_DI
 add_failing_test(parsererrors EXE ${PYTHON_EXECUTABLE} SCRIPT ${CMAKE_SOURCE_DIR}/shouldfail.py ARGS --openscad=${OPENSCAD_BINPATH} --retval=1 -o SUFFIX stl FILES ${FAILING_FILES})
 
 # Verify that test framework is paying attention to alpha channel, issue 1492
-add_cmdline_test(openscad-colorscheme-cornfield-alphafail EXE ${OPENSCAD_BINPATH} ARGS --colorscheme=Cornfield -o SUFFIX png FILES ${CMAKE_SOURCE_DIR}/../examples/Basics/logo.scad)
+#add_cmdline_test(openscad-colorscheme-cornfield-alphafail EXE ${OPENSCAD_BINPATH} ARGS --colorscheme=Cornfield -o SUFFIX png FILES ${CMAKE_SOURCE_DIR}/../examples/Basics/logo.scad)
 # The "expected image" supplied for this "alphafail" test has the alpha channel for all background pixels cleared (a==0), when they should be opaque (a==1) for this colorscheme.  so if test framework is functioning properly then the image comparison should fail
-set_property(TEST openscad-colorscheme-cornfield-alphafail_logo PROPERTY WILL_FAIL TRUE)
+#set_property(TEST openscad-colorscheme-cornfield-alphafail_logo PROPERTY WILL_FAIL TRUE)
```
kintel: Could you add a comment explaining why this was commented out so that the next reader of this code has a starting point?

src/Camera.cc
```diff
@@ -8,13 +8,13 @@ Camera::Camera(CameraType camtype) :
 	PRINTD("Camera()");
 
 	// gimbal cam values
-	object_trans << 0,0,0;
-	object_rot << 35,0,25;
+	object_trans << -0.0,-0.0,-0.0;
```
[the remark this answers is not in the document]
nophead: They get negated when displayed on the GUI. With the default view it used to show -0.00, -0.00, -0.00. Now it shows 0.00, 0.00, 0.00, which look nicer.

tests/CMakeLists.txt
nophead: Done.

nophead:
Yes I can see changing the default GUI view is controversial. Alternatively I could change the default command line to match the GUI. All the 3D test images then need to be updated, but they do anyway because of the lighting change and the fraction of a degree altitude change.

In general the tests still look valid when viewed the other way but it is obvious a lot of them have been laid out for first quadrant viewing. We could change the test command to override the default. Or we could change all the test scad files to move them to the second quadrant (big job).

My reason for changing the GUI is looking towards the second quadrant seems illogical to me.

I don't think it makes sense for the default command line image to be different from the default GUI view.

nophead:
Here is an example of how the tests change:
image
image

kintel:
I generally agree to what you’ve done, I’m just not certain what’s illogical about looking towards the second quadrant. If we’re changing this I’m not too worried about the tests, we could even clean that up subsequently as a new issue.

Do you know if other design tools have converged on a common default view angle?

nophead:
It is natural to either design around the origin, or design in the first quadrant. As you can see if you design in the first quadrant it looks wrong viewed towards the second.

I don't use any other form of CAD but I have these installed on my machine.

Blender seems to look towards the second quadrant .
image

And so does FreeCad:
image

And Simplyfy3D
image

Only RepetierHost thinks like me:
image

So maybe there is a convention of viewing the three axes from the right, rather than the left. I will restore the default and regenerate all the tests images.

nophead:
So I have updated my other PR, which supersedes this one, to make the GUI view unchanged. The command line default is now the same.

PR #2491, opened 2018-09-21
Camera unification

nophead:
The camera now always works in gimbal mode.  Command line vector parameters are converted to gimbal parameters.

This makes the code much shorter and simpler and easier to maintain. It fixed 3 bugs where the camera modes had diverged but I fixed them in the original code first to create the test images for this bigger change. See http://forum.openscad.org/Light-source-and-default-camera-position-different-on-command-line-td24447.html for context.

gimbalDefaultTranslate() has been replaced by accessors and mutators that work in user space units. This removes all the ad-hoc conversions and camera copying.


kintel:
Thanks a lot for digging into this. I'll review it at my earliest convenience.

nophead:
The only change in the GUI mode is removal of the negative zeros in the default translate.
The command line default view is now the same as the GUI. 
The command line gimbal mode is unchanged except for when view all is set without view center.
Vector mode views are unchanged except that the lighting is now from above instead of below.

src/Camera.h
```diff
@@ -59,8 +59,12 @@ class Camera
...
 protected:
-        // Perspective settings
+	// Perspective settings
 	double viewer_distance;
 	// Orthographic settings
 	double height; // world-space height of viewport
+
+private:
+	// force angle to be 0-360
+	static double wrap(double angle) { return fmodf(360 + angle, 360); }
```
kintel: This method doesn't really belong in this class. If you move the implementation of `setVp*()` out of the header file, this could be a local helper function (even in an anonymous namespace to avoid any potential namespace pollution).

src/Camera.h
```diff
@@ -23,22 +23,22 @@ projection, Perspective and Orthogonal.
...
-	enum class CameraType { NONE, GIMBAL, VECTOR } type;
 	enum class ProjectionType { ORTHOGONAL, PERSPECTIVE } projection;
-	Camera(CameraType camtype = CameraType::NONE);
+	Camera();
 	void setup(std::vector<double> params);
-	void gimbalDefaultTranslate();
 	void setProjection(ProjectionType type);
 	void zoom(int delta);
-	double zoomValue();
+	double zoomValue() const;
```
kintel: 👍

src/Camera.cc
```diff
@@ -105,22 +66,23 @@ void Camera::setProjection(ProjectionType type)
...
+double Camera::zoomValue() const
 {
 	return viewer_distance;
 }
 
-std::string Camera::statusText()
+std::string Camera::statusText() const
 {
+	auto vpt = getVpt();
+	auto vpr = getVpr();
```
kintel: You could 'const' both these for a bit extra readability and compile-time checks.
You could even make those methods return a `const &`, but that's just optimization which probably doesn't give us much here.

src/Camera.cc
```diff
@@ -25,70 +18,38 @@ Camera::Camera(CameraType camtype) :
...
+		setVpr(params[3], params[4], params[5]);
 		viewer_distance = params[6];
 	} else if (params.size() == 6) {
-		type = CameraType::VECTOR;
-		eye << params[0], params[1], params[2];
-		center << params[3], params[4], params[5];
+		Eigen::Vector3d eye(params[0], params[1], params[2]);
+		Eigen::Vector3d center(params[3], params[4], params[5]);
+		object_trans = -center;
+		auto dir = center - eye;
```
kintel: Technically, both eye, center and dir are `const`.

src/Camera.cc
nophead: OK

src/Camera.cc
nophead: OK

src/Camera.h
nophead: OK I have moved them but moving these one liners into the cpp file in the usual format adds another 25 lines of code that has to be kept in sync with the header. That is why I have always written one line accessors inline. How does a private function pollute a namespace?

kintel: Private functions don't pollute the namespace, they just becomes part of the class API, which is unnecessary for such implementation defaults.
For helper functions, you can avoid polluting the namespace by declaring them `static`, or put them in an anonymous namespace, which should be pretty much the same thing. (`static` functions were deprecated for a while during standardization, so I guess a bunch of people adopted anonymous namespaces early on, but I believe the difference just comes down to preference).

PR #2492, opened 2018-09-22
warn, when an assigment from the RootModule overwritten by an include

MichaelPFrey:
beforemainafter

I used the uncomplete code as used by assert, so I added an "assert(false);" into my test files.

### DIFF
--- a/doc/testing.txt
+++ b/doc/testing.txt
@@ -78,11 +78,12 @@ Adding a new test:
 2) if the test is non-obvious, create a human readable description as comments in the test (or in another file in the same directory in case the file isn't human readable)
 3) if a new test app was written, this must be added to tests/CMakeLists.txt
 4) Add the tests to the test apps for which you want them to run (in tests/CMakeLists.txt)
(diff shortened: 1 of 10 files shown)


nophead:
I now get this with an empty file that includes my library. It isn't very helpful because it doesn't specify the filename. There are hundreds of files, some used, some included and many of the have a length defined.

image

MichaelPFrey:
@nophead I have not done a lot of testing with "use", so I would expect the issue be something related to the use instruction. The solution will likely be just to disable the warning when a file is being "used". I am going to look into it.

nophead:
I used an editor that can search all the files for "length =" and show line numbers of matching lines and it turned out they were duplicate definitions inside modules in a used file. They were a valid bug as there should only have been one definition in each module but fortunately the last one was the correct one.

It is a useful warning but it needs to show the file name, not just the line number.

kintel:
Since this is already merged, could someone summarize this as a new issue?

nophead:
OK, #2513

PR #2494, opened 2018-09-23
Command line axes, scales and crosshairs

nophead:
With unified camera code there is no reason not to offer axes, scales and crosshairs when exporting PNG from the command line. This PR does that with --showaxes, -- showscale, --showcrosshair.

The automated tests don't detect one pixel wide lines so these need to be manually checked. Regression is unlikely because it used the same drawing code as the GUI and standard command line processing.


kintel:
@nophead I have an existing PR (#2092) which aims to solve pretty much the same thing. In that PR, I did experiment with using a general purpose `--view` cmd-line option taking various view flags as argument. What do you think of that idea? My initial use-case was to try to make it a bit more readable to pass a bunch of view flags.

nophead:
Yes that is a nicer way to do it. I did consider wrapping the three flags into a structure but didn't consider it worth it but if we add showEdges and the camera then I think that is the way to go. It seems to have been languishing for a long time though.

nophead:
They are all orthogonal except axes and scales. I made scales force axes on. I don't think you really need to test the other combinations as they don't really interact.

The problem I had with testing was the image comparison is not sensitive to notice axes.

PR #2495, opened 2018-09-23
add test case for undef inf nan

MichaelPFrey:
This tests may look a bit silly, but in C++ for e.g.
`nan = nan` evaluates false which - fun facts - also applies to OpenSCAD when you manage to compare nan.

It also shows an potential issue in csg. csg should not contain nan or inf.

The results are highly likely to change when someone is cleaning up the language, but in any case:
Making sure that heavily abusing the language does not result in a crash is good for a regression test.


MichaelPFrey:
thank you for the tip, I am looking into it

PR #2497, opened 2018-09-24
add test case for undef, inf and nan

MichaelPFrey:
squased #2495


MichaelPFrey:
That is bizarre
`circle(NotANumber);`

```
-circle($fn = 0, $fa = 12, $fs = 2, r = -nan);
+circle($fn = 0, $fa = 12, $fs = 2, r = nan);
```

digging deeper: http://rextester.com/l/cpp_online_compiler_gcc

```
//g++  5.4.0

#include <iostream>

int main()
{
    std::cout << (1.0/0)/(1.0/0);
}
```

results in `-nan`

```
//clang 3.8.0

#include <iostream>

int main()
{
    std::cout << (1.0/0)/(1.0/0);
}
```

results in `nan`

well: better not output nan, That we truly have nan is provable by nan == nan evaluating false.

PR #2499, opened 2018-09-26
Boost options_description now used to generate command line help.

nophead:
The command line help message is now programmatically generated from the  boost::options_description. The descriptions have been tidied up and extended to show the experimental features available and the color schemes available with the default marked by a star.

Unrecognised options give an error again, reversing @kintel's commit https://github.com/openscad/openscad/commit/752f35eab6c7b20d188bc36836e622665c8eba42. This assumes #2107 makes it no longer necessary.

This is an example of the error handling:
```
$ ./release/openscad.exe -fred
unrecognised option '-fred'

Usage: openscad.exe [options] file.scad
  -o [ --o ] arg        output specified file instead of running the GUI, the
                        file extension specifies the type: stl, off, amf, csg,
                        dxf, svg, png, echo, ast, term, nef3, nefdbg

  -D [ --D ] arg        var=val -pre-define variables
  -p [ --p ] arg        customizer parameter file
  -P [ --P ] arg        customizer parameter set
  --enable arg          enable experimental features: assert | echo | lc-each |
                        lc-else | lc-for-c | amf-import | svg-import |
                        customizer

  -h [ --help ]         print this help message and exit
  -v [ --version ]      print the version
  --info                print information about the build process

  --camera arg          camera parameters when exporting png:
                        =translate_x,y,z,rot_x,y,z,dist or
                        =eye_x,y,z,center_x,y,z
  --autocenter          adjust camera to look at object's center
  --viewall             adjust camera to fit object
  --imgsize arg         =width,height of exported png
  --render arg          for full geometry evaluation when exporting png
  --preview arg         [=throwntogether] -for ThrownTogether preview png
  --projection arg      =(o)rtho or (p)erspective when exporting png
  --csglimit arg        =n -stop rendering at n CSG elements when exporting png
  --colorscheme arg     =colorscheme: *Cornfield | Metallic | Sunset |
                        Starnight | BeforeDawn | Nature | DeepOcean | Solarized
                        | Tomorrow | Tomorrow Night | Monotone

  -d [ --d ] arg        deps_file -generate a dependency file for make
  -m [ --m ] arg        make_cmd -runs make_cmd file if file is missing
  -q [ --quiet ]        quiet mode (don't print anything *except* errors)
  --debug arg           special debug info
  -s [ --s ] arg        stl_file deprecated, use -o
  -x [ --x ] arg        dxf_file deprecated, use -o
```

### DIFF
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -767,7 +767,7 @@ int main(int argc, char **argv)
 	int rc = 0;
 	StackCheck::inst()->init();
 #ifdef OPENSCAD_QTGUI
-    QCoreApplication *app = new QCoreApplication(argc, argv);
+	QCoreApplication *app = new QCoreApplication(argc, argv);
 	PlatformUtils::registerApplicationPath(app->applicationDirPath().toLocal8Bit().constData());
 	delete app;
 #else
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -128,7 +128,7 @@ public:
 	}
 };
 
-static void help(const char *arg0, po::options_description &desc, bool failure = false)
+static void help(const char *arg0, const po::options_description &desc, bool failure = false)
 {
 	std::stringstream ss;
 	ss << desc;
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -767,7 +767,7 @@ int main(int argc, char **argv)
 	int rc = 0;
 	StackCheck::inst()->init();
 #ifdef OPENSCAD_QTGUI
-	{
+	{   // Need a dummy app instance to get the application path but it needs to be destroyed before the GUI is launched.
 		QCoreApplication app(argc, argv);
 		PlatformUtils::registerApplicationPath(app.applicationDirPath().toLocal8Bit().constData());
 	}
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -128,35 +128,12 @@ public:
 	}
 };
 
-static void help(const char *progname, bool failure = false)
+static void help(const char *arg0, po::options_description &desc, bool failure = false)
 {
-  int tablen = strlen(progname)+8;
-  char tabstr[tablen+1];
-  for (int i=0;i<tablen;i++) tabstr[i] = ' ';
-  tabstr[tablen] = '\0';
-
-	PRINTB("Usage: %1% [ -o output_file [ -d deps_file ] ]\\\n"
-         "%2%[ -m make_command ] [ -D var=val [..] ] \\\n"
-	 "%2%[ --help ] print this help message and exit \\\n"
-         "%2%[ --version ] [ --info ] \\\n"
-         "%2%[ --camera=translatex,y,z,rotx,y,z,dist | \\\n"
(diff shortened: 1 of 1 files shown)


MichaelPFrey:
for those that do not follow the mailing list: http://forum.openscad.org/Command-line-parsing-td24611.html

src/openscad.cc
```diff
@@ -817,37 +791,57 @@ int main(int argc, char **argv)
...
+#endif
+		("help,h", "print this help message and exit")
 		("version,v", "print the version")
-		("info", "print information about the building process")
-		("render", po::value<string>()->implicit_value(""), "if exporting a png image, do a full geometry evaluation")
-		("preview", po::value<string>()->implicit_value(""), "if exporting a png image, do an OpenCSG(default) or ThrownTogether preview")
-		("csglimit", po::value<unsigned int>(), "if exporting a png image, stop rendering at the given number of CSG elements")
-		("camera", po::value<string>(), "parameters for camera when exporting png")
-		("autocenter", "adjust camera to look at object center")
+		("info", "print information about the build process\n")
```
kintel: Is the extra `\n` an oversight?

src/openscad.cc
```diff
@@ -798,6 +766,12 @@ int main(int argc, char **argv)
 {
 	int rc = 0;
 	StackCheck::inst()->init();
+#ifdef OPENSCAD_QTGUI
+	const auto application_path = QCoreApplication::applicationDirPath().toLocal8Bit().constData();
```
kintel: Not sure what happened here. `QCoreApplication::applicationDirPath()` will return an empty string until a `QCoreApplication` has been instantiated. Note: It doesn't matter _where_ this object is instantiated, as Qt will keep a pointer to it. Excerpt from the Qt source code:
```
QString QCoreApplication::applicationDirPath()
{
    if (!self) {
        qWarning("QCoreApplication::applicationDirPath: Please instantiate the QApplication object first");
        return QString();
    }
   [...]
```

src/openscad.cc
nophead: No I deliberately separated all the png options.

src/openscad.cc
nophead: Oops, I must not have tested it after I changed it to a static call. I have reverted to instantiating an app and deleting it again.

kintel:
@nophead Is it possible to suppress some arguments? e.g.:
```
  --psn arg             process serial number
```

The above is just a way of swallowing an argument passed by the OS, not meant for users to pass.

nophead:
Yes it can be made a hidden option like input-file. I will do that.

MichaelAtOz:
> --debug arg           special debug info

Does this function on release and/or snapshot versions, or just development builds?

kintel:
It's meant to be a development tool, so it's not "officially" supported, but I cannot see any actual debug build check, so I guess it's available right now.
We may want to change that at some point, e.g. for performance reasons.

kintel:
FYI: the "arg special debug info" is simply a comma-separated list of module names, where a "module" is any source code filename without suffix.

PR #2504, opened 2018-10-06
Accept the echo() function as stable

t-paul:
- Add example for echo() in function/expression context.
- Remove experimental flag of echo() in expressions.


kintel:
I think this was, for all practical purposes, the work of @t-paul.

MichaelPFrey:
side note: We may want o adjust the test case for this example to also include the console/echo outputs, so that we really test the example. 

I know that the `echo(version=version());` would get in the way of a "predictable" echo output, but as it is not necessary for show casing the echo function, it could be removed.

Or we simply add a `color("blue") text("Look in the Console", halign = "center");` to have something on both the screen and in the csg file so that we have at least a basic "smoke test".

t-paul:
Hmm, good point, I thought that is enabled automatically... maybe it's disabled because of that `version()`?

MichaelPFrey:
the CMakeList enabling of experimental functions in context with #2499 is a more general issue, so I opened #2514

PR #2509, opened 2018-10-07
Builtin ord function

t-paul:
This continues #1380.

Updated following the discussion:
- Extend for() to handle loops over strings `[for (c = str) ord(c)]`
- [Simplify](https://github.com/openscad/openscad/pull/1380#issuecomment-160484232) the ord() implementation to only handle a single character

ToDo:
- [x] Verify tests
- [x] Docs
- [x] Cheat sheet
- [x] Update syntax highlighting


t-paul:
Manual: [`ord()`](https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/String_Functions#ord)

src/control.cc
```diff
@@ -93,6 +93,20 @@ void ControlModule::for_eval(AbstractNode &node, const ModuleInstantiation &inst
...
+                else if (it_values->type() == Value::ValueType::STRING) {
+			const std::string val = it_values->toString();
+			const char *ptr = val.c_str();
+
+			gchar outbuf[8];
+			const glong len = g_utf8_strlen(ptr, -1);
+			for (;*ptr && len > 0;ptr = g_utf8_next_char(ptr)) {
+				const gunichar ch = g_utf8_get_char(ptr);
+				const gint end = g_unichar_to_utf8(ch, outbuf);
+				outbuf[end] = 0;
```
kintel: I wonder how much overhead there is to this. This may just be premature optimization, but would smth. like this be more efficient? ..similarly for other occurrences of such conversion from utf-8 to unicode and back to utf-8
```
  do {
    auto next = g_utf8_next_char(ptr);
    c.set_variable(it_name, ValuePtr(std::string{ptr, next-ptr}));
    for_eval(node, inst, l+1, &c, evalctx);
    ptr = next;
  } while (ptr);
```

t-paul: I don't think that works, `std::string{ptr, next-ptr}` would copy the encoded representation in utf-8, not the code point value?

E.g. [Unicode Character 'GRINNING FACE' (U+1F600)](http://www.fileformat.info/info/unicode/char/1f600/index.htm) would generate `0xF0 0x9F 0x98 0x80 / 0xf09f9880) instead of `0x0001F600`.

kintel: ..but in the code above, don't we start with UTF8, convert one character to unicode, then convert back to UTF8? My suggestion was to skip the conversion and just copy the UTF-8 bytes directly.

t-paul: Ahh, right, the char value is not used here as with `ord()` but just converted back to string. Let me try that...

kintel: Looks great. In terms of safety, I'm not sure what will happen for invalid strings. calling `g_utf8_validate()` prior to running such loops may be recommended.

t-paul: Yep, although it's probably better to push the guaranties into the parser so we don't have to validate on every evaluation.

MichaelPFrey:
is ord added to the syntax highlighting?

t-paul:
Yep, it is now :smile:

kintel:
@t-paul Feel free to merge if you're happy with this

Issue #2513, opened 2018-10-08
Warnings of variables overritten in used files need to show the file name.


nophead:
```
WARNING: length was assigned on line 25 but was overwritten on line 34 
WARNING: length was assigned on line 72 but was overwritten on line 79 
```
When the lines are not in the main file the filename should be mentioned, otherwise it can be hard to find in a library of hundreds of files. This is especially true since this is a new warning so it appears in old projects, so likely isn't a file that has recently changed.

MichaelPFrey:
Thank you for the report.

I have a working patch that disables the duplicate assignment warning for "used" files.
The idea is a following: The user might "use" libraries that he has no interest in modifying and/or that a legacy code, so making warnings for code that is isolated from the current code is counter productive.
A library maintainer gets the warning, when he or she is editing the code of the library it self.

Even if one considers the warning to be useful "anyway": I can not create a nice relative path. Outputting "sometimes" just filenames and sometimes relative paths  (that are not always relative to the file currently being edited) with filename would be confusing.

nophead:
>A library maintainer gets the warning, when he or she is editing the code of the library it self.

Not in my case because I don't ever load the used file into OpenSCAD directly. It is always used by another file that instantiates the module, either a project or a test file. The way my library is structured object descriptions are lists that are included and the code to draw them is used. Using it on its own doesn't draw anything because the descriptions are in another file that uses it.

I think OpenSCAD can always get the full path of a file it uses so we could output that or also get the full path of the main file and then use that to calculate the minimum relative  path.

MichaelPFrey:
> Using it on its own doesn't draw anything because the descriptions are in another file that uses it.
I tend to write some short self testing code in libraries I want to use using use.
The self-test getting ignored when beeing used.

> I think OpenSCAD can always get the full path of a file it uses, so we could output that or also get the full path of the main file and then use that to calculate the minimum relative path.

In theory: Sure. Give me two paths and I can use boostfs_uncomplete to get a relative path.

In practice: Tricky.

One example on why it is tricky: We currently do not properly canonicalize the paths in the test framework, making the testframe work unable to truly test boosts uncomplete function in context of assert and duplicate assignments.

More specifically with used files: Currently, the parser only know the RootFile, not the mainFile (why should it? It does not need it currently)

We could of course change the datastructur, but I do not fully understand the datastructures involved. For e.g. we have a module cache and the user can have multiple openscad windows open.

[...]

PR #2516, opened 2018-10-10
do not warn for duplicate assigmentens in "used" files

MichaelPFrey:
change on the parser: distinguish between mainModule and included modules

### DIFF
--- a/src/parser.y
+++ b/src/parser.y
@@ -64,6 +64,7 @@ int lexerlex(void);
 
 std::stack<LocalScope *> scope_stack;
 FileModule *rootmodule;
+bool bMainModule;
 
 extern void lexerdestroy();
 extern FILE *lexerin;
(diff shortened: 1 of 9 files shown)


kintel:
Two questions:
* Are we covering enough cases where such warnings should be issued? (See discussion in #2513)
* This manages a subset of possible overwrites; could we clarify in a code comment what we're actually trying to achieve?

src/parser.y
```diff
@@ -661,10 +662,12 @@ void yyerror (char const *s)
          (*sourcefile()) % lexerget_lineno() % s);
 }
 
-bool parse(FileModule *&module, const char *text, const std::string &filename, int debug)
+bool parse(FileModule *&module, const char *text, const std::string &filename, int debug, bool mainModule)
 {
   fs::path path = fs::absolute(fs::path(filename));
   
+  bMainModule = mainModule;
```
kintel: Would it be cleaner to instead pass the `assignmentWarning` flag as an argument, to have one less piece of logic to manage?

src/parser.y
```diff
@@ -215,13 +216,13 @@ assignment:
                         auto prevFile = assignment.location().fileName();
                         auto currFile = LOC(@$).fileName();
                         
-                        if(assignmentWarning && prevFile==RootFile && currFile == RootFile){
+                        if(bMainModule && assignmentWarning && prevFile==RootFile && currFile == RootFile){
```
kintel: `assignmentWarning && prevFile==RootFile` is shared between the if and else branches. Pulling this out could make the code a bit more readable.

MichaelPFrey:
I see that as somewhat of a hotfix for the confusing duplicate assignment warning, when a variable is overwritten within a "use<>", not as a final solution covering all bases.

#2492 it self was as much a feature as it is testing the water.  #2513 shows support for a verbose warning system. Knowing that, I am willing to go the extra mile and try to cover more cases of duplicate assignments.

MichaelPFrey:
replaced by #2516 

Keeped for reference, as #2516 is a more complicated patch, based on this one.

PR #2519, opened 2018-10-13
Minor cleanup (C++11) of scintillaeditor

kintel:
No functional changes

### DIFF
--- a/src/scintillaeditor.cpp
+++ b/src/scintillaeditor.cpp
@@ -20,10 +20,10 @@ public:
 
 QsciScintilla::WrapMode SettingsConverter::toWrapMode(Value val)
 {
-	std::string v = val.toString();
-	if (v == "Char") {
+	auto strVal = val.toString();
+	if (strVal == "Char") {
 		return QsciScintilla::WrapCharacter;
-	} else if (v == "Word") {
+	} else if (strVal == "Word") {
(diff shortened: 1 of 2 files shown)


t-paul:
Hmm, the change in `QStringList ScintillaEditor::colorSchemes()` seems to have broken builds on some older distros (looks like that is with gcc-4.8). Isn't that version supposed to support c++11 fully?

kintel:
humm, the issue in question is `for (const auto& dirEntry : fs::directory_iterator{color_schemes})`.

It looks like the ADL lookup of `begin()/end()` for `directory_iterator` arguments fail. Not sure why. It could be a boost version issue.
Also, this is gcc-4.9 - it should have all of C++11.

I don't have that compiler handy, bug I'll try to reproduce it this week-end, unless someone beats me to it :)

t-paul:
Hmm, docs say range support was added in boost 1.51.0, it still fails with 1.55 though...

kintel:
yup, but range-based relies on ADL of begin()/end(), and for const iterators cbegin()/cend(), which I think were added in C++14, so it may depend on how well compilers turns off C++14 support in C++11 mode perhaps?

t-paul:
Hmm, so maybe `boost::make_iterator_range(fs::directory_iterator{color_schemes}, {})` for now, I guess.

Issue #594, opened 2014-01-06
glib2 is a giant library, only used for utf8, and should be dropped


brianolson:
glib2 is about half a million lines of source, and openscad is only using maybe 2000 lines of it for some utf8 library code. openscad should drop glib2 for a smaller library (probably one that just deals with unicode).

donbright:
can we close this now? it would be nice to have less dependencies, but i just built openscad on 7 different linux machines without much problem. The build scripts for linux have been updated to deal with all these new libraries. Unicode text rendering is part of internationalization and appears to be baked into the core of modern linux distributions. It has also found it's way into MSYS and MSYS2 and the MXE cross-builder (for Windows(TM)). BSD surely ain't too far behind. 

I noticed this issue is prompted by problems building on OSX. I also noticed in the makefiles that OSX seems to have lots of build problems over the different releases of XCode and OSX, but IMHO those should be "OSX build fail" issues not generalized build issues. If someone wants to donate me a Mac i can set up different OSX versions in a VM and fix OSX build problems :)

donbright:
interesting note for historical researchers: 

Microsoft is now giving free downloads to Windows 10 insider preview, including installing it in a VM, and keeping it for free after release date. 

Apple makes it illegal to run OSX on non-apple Hardware, and also makes it illegal to run OSX inside a VM unless it is on Apple hardware. This has also made it harder to build OSX cross compilers to run from within linux. (see the https://github.com/tpoechtrager/osxcross project on github).

t-paul:
Right, I'm closing this now. If there are specific build problems, we can open a new issue.

MichaelAtOz:
> Apple makes it illegal to run OSX on non-apple Hardware

Do they test for that?

donbright:
it's not relevant whether they test for it. if you are an open source developer you have to follow the IP law to the best of your ability because you are risking not just your own self but the other people working on the project, the project itself, the users, and the project's future.  

you cant ship a product with the instructions "hey.. ok.. to install our product, first step, break the law and dont get caught"

a good bad example is what happened to the ReactOS project

https://en.wikipedia.org/wiki/ReactOS#Internal_audit

there are many others.

MichaelAtOz:
I wasn't suggesting that, I was just curious.

donbright:
[...]

PR #2523, opened 2018-10-15
Replace glib with utfcpp

kintel:
This is just an experiment to see if replacing glib with [utfcpp](https://github.com/nemtrif/utfcpp) would result in nicer and still correct code.
Can be killed if it's a dead end.
As discussed in #594

### DIFF
--- a/src/AboutDialog.html
+++ b/src/AboutDialog.html
@@ -71,7 +71,6 @@ Please visit this link for a copy of the license: <a href="http://www.gnu.org/li
 <li><a href="http://qt-project.org">Qt Toolkit</a></li>
 <li><a href="http://www.riverbankcomputing.co.uk/software/qscintilla">QScintilla</a></li>
 <li><a href="http://www.boost.org">Boost</a></li>
-<li><a href="https://developer.gnome.org/glib">GLib</a></li>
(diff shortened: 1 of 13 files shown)


t-paul:
Hmm, it interface certainly looks a bit nicer, although there's a bit of improvements possible with the `glib` stuff too, right now it's just sprinkled everywhere where needed.

Anyway, I'll collect my personal pro/cons here :-)

_pro_
- Small header-only lib with C++ interface
- Included in most distros
- Seems to be actively maintained, small number of changes

_cons_
- not in MXE (probably easy to do though as header only)
- Distros seem to ship old version (e.g. Ubuntu 18.04/18.10 have 2.3.4 with 2.3.5 released 2016-11-06 / 2.3.6 just now released)
- Using a different library because of "glib is big and I don't like it" seems to be a quite weak argument as it does focus on a not very important fact of the used code

kintel:
Thanks for your input. I figured an actual implementation will help us evaluate this better, so let's let it soak for a while :)

kintel:
It seems like glib is getting more unwieldy for each release. We still manage to keep it somewhat in shape, but it would be nice to be able to rely on something slightly easier to build.

PR #2520, opened 2018-10-14
duplicate assigment warnings

MichaelPFrey:
this replaces #2516 based on the discussion  #2513

Issue #2468, opened 2018-08-29
openscad doesn't warn on the use of unknown $variables


l29ah:
Only on the regular, non-$-prefixed ones, i can see the warning `WARNING: Ignoring unknown variable 'blah'.`.

MichaelPFrey:
closed with #2524

PR #2524, opened 2018-10-16
 warn on the use of unknown $variables, closing #2468

MichaelPFrey:
- [x] wait for #2535 beeing merged to master
- [x] wait for #2543 beeing merged to master
- [x] wait for #2520 beeing merged to master
(this is a low priority change in my opinion - this also gives people time to prepare for the change)
- [x] merge master with this branch
- [x] update test cases
- [x] remove do not merge label


nophead:
Looks like I might have been relying on this bug. I thought that comparing a variable with undef not giving a warning was deliberate but it might just have been this bug.

It seems reasonable that you should be able to compare against undef to guard against using an undefined variable without the test triggering an undefined variable use warning.

MichaelPFrey:
Hhmmm... in the moment of the variable lookup we can not know that it is for an undef comparison.

I would suggest a separate "isUndef()" methode/function for just that.

Note that c++ has an isnan() function [1](https://en.cppreference.com/w/cpp/numeric/math/isnan) because of how strange Nan compares [2](https://en.wikipedia.org/wiki/NaN#Comparison_with_NaN). So going for a function for a special variable state is not unrepresented.

nophead:
Can I suggest isUndef is added before this is merged as I can't live with half a million warnings in all my projects. It takes a long time to spit them out. I thought repeated identical warnings where supposed to be suppressed?

nophead:
Can I suggest isUndef is added before this is merged as I can't live with half a million warnings in all my projects. It takes a long time to spit them out. I thought repeated identical warnings where supposed to be suppressed?

nophead:
Can I suggest isUndef is added before this is merged as I can't live with half a million warnings in all my projects. It takes a long time to spit them out. I thought repeated identical warnings where supposed to be suppressed?

nophead:
Can I suggest isUndef is added before this is merged as I can't live with half a million warnings in all my projects. It takes a long time to spit them out. I thought repeated identical warnings where supposed to be suppressed?

nophead:
Can I suggest isUndef is added before this is merged as I can't live with half a million warnings in all my projects. It takes a long time to spit them out. I thought repeated identical warnings where supposed to be suppressed?

nophead:
Can I suggest isUndef is added before this is merged as I can't live with half a million warnings in all my projects. It takes a long time to spit them out! I thought repeated identical warnings were supposed to be suppressed?

PR #2530, opened 2018-10-21
FreeBSD and Boost build fixes

### LINKED ISSUE #2529: Fails to build with Boost 1.69 (develop snapshot)
After boostorg/logic@23cd89d4c80f build fails. See [error log](https://ptpb.pw/aK-J).

```c++
src/CSGTreeEvaluator.cc:179:9: error: no viable conversion from 'boost::tribool' to 'bool'
                        bool convex = ps->convexValue();
                             ^        ~~~~~~~~~~~~~~~~~
```

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/65048990-fails-to-build-with-boost-1-69-develop-snapshot?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

kintel:
Should fix #2529

### DIFF
--- a/openscad.pro
+++ b/openscad.pro
@@ -575,7 +575,7 @@ unix:!macx {
   SOURCES += src/input/DBusInputDriver.cc
 }
 
-unix:!macx {
+linux: {
   DEFINES += ENABLE_JOYSTICK
 
   HEADERS += src/input/JoystickInputDriver.h
--- a/openscad.pro
+++ b/openscad.pro
@@ -565,7 +565,7 @@ HEADERS += src/ext/libtess2/Include/tesselator.h \
            src/ext/libtess2/Source/sweep.h \
            src/ext/libtess2/Source/tess.h
 
-unix:!macx {
+has_qt5:unix:!macx {
   QT += dbus
   DEFINES += ENABLE_DBUS
   DBUS_ADAPTORS += org.openscad.OpenSCAD.xml
--- a/src/CSGTreeEvaluator.cc
+++ b/src/CSGTreeEvaluator.cc
@@ -176,7 +176,7 @@ shared_ptr<CSGNode> CSGTreeEvaluator::evaluateCSGNodeFromGeometry(
 			auto ps = dynamic_pointer_cast<const PolySet>(geom);
 			// Since is_convex() doesn't handle non-planar faces, we need to tessellate
 			// also in the indeterminate state so we cannot just use a boolean comparison. See #1061
-			bool convex = ps->convexValue();
+			bool convex{ps->convexValue()};
 			if (ps && !convex) {
 				assert(ps->getDimension() == 3);
 				auto ps_tri = new PolySet(3, ps->convexValue());
--- a/src/FileModule.h
+++ b/src/FileModule.h
@@ -3,7 +3,7 @@
 #include <string>
 #include <unordered_map>
 #include <unordered_set>
-#include <time.h>
+#include <ctime>
 
 #include "module.h"
 #include "value.h"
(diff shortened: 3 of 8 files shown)


jbeich:
Maybe also replace ftime() (missing on [OpenBSD](https://man.openbsd.org/ftime), requires `-lcompat` on [DragonFly](https://leaf.dragonflybsd.org/cgi/web-man?command=ftime&section=3), [FreeBSD](https://man.freebsd.org/ftime/3), [NetBSD](http://netbsd.gw.com/cgi-bin/man-cgi?ftime+3+NetBSD-current)) with `std::chrono` (C++11). See [dowstream hack](https://github.com/freebsd/freebsd-ports/blob/1548b5448a3fa91810be061182d28c3cd070d772/cad/openscad-devel/files/patch-openscad.pro) fixing
```
objects/StatCache.o: In function `StatCache::stat(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, stat&)':
StatCache.cc:(.text+0x21e): undefined reference to `ftime'
StatCache.cc:(.text+0x445): undefined reference to `ftime'
c++: error: linker command failed with exit code 1 (use -v to see invocation)
```

jbeich:
179074dff8c23cbc0e651ce8463737df0006f4ca (2018-04-19) snapshot with the patches here on top [builds fine](https://ptpb.pw/530X) (Qt4) against Boost 1.69 (2018-10-20) on FreeBSD:
```
=> SHA256 Checksum OK for openscad/ce5646179848.patch.
=> SHA256 Checksum OK for openscad/9c269c759b20.patch.
[...]
===>  Applying FreeBSD patches for openscad-devel-2018.04.19_5
[...]
 0x0000000000000001 NEEDED               Shared library: [libboost_filesystem.so.1.69.0]
 0x0000000000000001 NEEDED               Shared library: [libboost_program_options.so.1.69.0]
 0x0000000000000001 NEEDED               Shared library: [libboost_regex.so.1.69.0]
 0x0000000000000001 NEEDED               Shared library: [libboost_system.so.1.69.0]
 0x0000000000000001 NEEDED               Shared library: [libboost_thread.so.1.69.0]
```

kintel:
@jbeich My last commit should remove the need for `-lcompat`.
Are there any other downstream patches we should be aware of?

kintel:
Good point about the Joystick driver. It's now enabled just for Linux.

jbeich:
b8160eac0cca (2018-10-22) snapshot [builds fine](https://ptpb.pw/Yuhx), tested via [downstream patch](https://bugs.freebsd.org/bugzilla/attachment.cgi?id=198564&action=diff).

hidapi isn't detected on FreeBSD by default because `hidapi-libusb` is only available on Linux. According to downstream packages, DragonFly [uses](https://github.com/DragonFlyBSD/DeltaPorts/blob/master/ports/comms/hidapi/dragonfly/patch-configure.ac) FreeBSD implementation while [NetBSD](https://github.com/NetBSD/pkgsrc/blob/trunk/comms/libhidapi/patches/patch-Makefile) and [OpenBSD](https://github.com/rfht/mystuff/blob/master/comms/libhidapi/patches/patch-configure_ac) use Linux one. If macOS uses `pkg-config` (via [Homebrew](https://github.com/Homebrew/homebrew-core/blob/master/Formula/pkg-config.rb)) then it probably doesn't need hardcoded `HIDAPI_LIBS`. 
https://github.com/openscad/openscad/blob/8912e648df2f84b42d7406f5d2a50218f72f8ab6/features/hidapi.prf#L29-L43
https://github.com/signal11/hidapi/bob/a6a622ffb680c55da0de787ff93b80280498330f/configure.ac#L222
https://github.com/signal11/hidapi/blob/a6a622ffb680c55da0de787ff93b80280498330f/libusb/Makefile.am#L4

t-paul:
HIDAPI is currently only really intended for Windows (and maybe MacOS?). On Linux it probably would work but require root, so it's not really a good option. I suspect BSD has similar restrictions.

Issue #1584, opened 2016-02-17
type test functions


doug-moen:
Let's add the following built-in functions for testing the type of a value:
- `is_bool(X)` returns `true` if X is true or false, otherwise it returns `false`
- `is_int(X)` returns `true` if X is an integer, otherwise `false`. Sample implementation: `X == floor(X)`
- `is_num(X)` returns `true` if X is a number, otherwise `false`. Note that NaN is Not a Number, so `is_num(0/0)` returns `false`.
- `is_string(X)` returns `true` if X is a string, otherwise `false`
- `is_list(X)` returns `true` if X is a list, otherwise `false`. If we implement Generalized Ranges #1583, then `is_list` will return `true` if X is a range.

These functions will be more robust than the current idioms that some people are using, like `abs(X)==undef` to test if X is a number. See the forum discussion http://forum.openscad.org/Determining-what-data-type-a-variable-is-holding-td16111.html

In issue #1574, it is proposed that OpenSCAD report errors if bad arguments are passed to functions or modules. This change may break code that passes a bad argument to a function to test the type of a value, but if we provide type test functions, then there is a correct way to write this kind of code.

An `is_range` function is not proposed, due to Generalized Ranges #1583.

t-paul:
What's the benefit of having a list of separate test functions instead of a `typeof()` that lots of languages use?

Is it good to have values that have no function returning `true`? With the proposed list of tests, it's not possible to distinguish ranges and NaN.

As there is no `int` in OpenSCAD, I would not pretend that type exists.

doug-moen:
A type is a set of values over which a set of operations is defined. This is an abstract concept that is independent of how the language is implemented. My goal is to provide a useful high level abstraction, not to expose low level implementation details.

The distinction between integers and real numbers is a standard part of mathematics, and the distinction is part of OpenSCAD. The "integer" type mainly arises in connection with lists. The index `i` in `X[i]` is an integer. The `len` function returns an integer. The vertex indices in the faces argument of `polyhedron` are integers. The convexity argument of polyhedron is an integer, and it is documented that way in the manual. So you can't say there are no integers in OpenSCAD.

These functions can be used to define other, more fine grained type predicates like this:
```
function is_vec3(X) = is_list(X) && len(X)==3 && is_num(X[0]) && is_num(X[1]) && is_num(X[2]);
```
That's the type of a point or vector in 3-space.

[...]

PR #2535, opened 2018-10-25
is_undef for checking for undef without a warning

MichaelPFrey:
this goes back to #2524 and #2468 

the ~~current~~ first implementation ~~is~~ was a hack, but it already shows how it could work and where the tweaks have to go.

The todo list beeing:
- [x] documentation
- [x] syntax highlighting
- [x]  more test cases
   - [x]  multiple parameters?
      - [x] isundef(1,2)
   - [x] "calculation"/more complex expression
      - [x] isundef(log(1,2,3))
      - [x] isundef(len("test","upps"))
      - [x] isundef(isundef())
- [x]  `src/evalcontext.h` // `const AssignmentList &eval_arguments;` shall remain private
- [x]  isundef -> is_undef
with now just the cheat sheet beeing open.


t-paul:
Maybe `is_undef()` matching the #1584 request for other check functions?

MichaelPFrey:
Documentation: (tobe copied to [wikibooks](https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/WIP))

is_undef accepts one parameter. If the parameter is undef, this function returns true. If the parameter is not undef, it returns false. When checking a variable (like `is_undef(a)`), it does the variable lookup silently, meaning that is_undef(a) does not cause `WARNING: Ignoring unknown variable 'a'. `

The alternative is code like this:
```
if(a==undef){
    //code goes here
}
```
or
```
b = (a==undef) ? true : false;
```
will cause `WARNING: Ignoring unknown variable 'a'. `.

is_undef also works for special variables, allowing for things like this:
```
exploded = is_undef($exploded) ? 0 : $exploded; // 1 for exploded view
```
----
For older openscad version, is_undef can be emulated with
`function is_undef ( a ) = (undef == a) ;`
which off-course causes warning(s), but requires no changes to code relaying on is_undef().

src/func.cc
```diff
@@ -924,6 +924,22 @@ ValuePtr builtin_cross(const Context *, const EvalContext *evalctx)
 	return ValuePtr(result);
 }
 
+ValuePtr builtin_is_undef(const Context *, const EvalContext *evalctx)
+{
+	if (evalctx->numArgs() == 1) {
+		const auto &arg =evalctx->getArgs()[0];
+		ValuePtr v;
+		if(const Lookup* lookup = dynamic_cast<const Lookup*> (arg.expr.get())){
```
kintel: Tip: You can use `dynamic_pointer_cast` instead of unpacking the shared_ptr

kintel:
Looks good. Unless it has already been done, we also need to make a reminder to update cheatsheet and docs before the next release.

src/func.cc
MichaelPFrey: That is much nicer. And there is no need to repeat the types, so auto simplifies it further.

MichaelPFrey:
Documentation will follow. I prefer to have a consent about the idea and have the implementation merged, before adding a feature to the documentation. Once its documented, people may start anticipating.

Issue #2392, opened 2018-05-02
Dot bug


plumbum:
Please look this code.

```
l = 61;
w = 44;
h = 78 + 7;

thick = 1;

difference() {
    cube([l+thick*2, w+thick*2, h+thick]);
    translate([thick, thick, thick])
       cube([l, w. h+thick*2]);
}
```
It is have errors, but not generate errors afer parsing.
Detail look string `cube([l, w. h+thick*2]);`. Here i place dot instead comma after `w`. It's not generate error, but not generate vaild model.

t-paul:
Yes, it's using Z = 0 as only 2 values are given in the vector. That will never generate a parse error.

I'll set it as "Discussion" if it maybe should give a warning.

nophead:
w. h is not a syntax error because this like p.x are allowed. It gives
undef.

MichaelPFrey:
So, to simplify the issue to its essential:
`cube([1,1]);` should give a warning, as the third value of the vector is missing (and is thus assumed tobe zero)

Looking at the source code, it gets tricky quickly.

Oddly enough, getVec3 specifically defaults the z-value, when a Vector with two elements is supplied [[1]](https://github.com/openscad/openscad/blob/master/src/value.cc#L375), but cube does not supply a default [[2]](https://github.com/openscad/openscad/blob/master/src/primitives.cc#L195), so 0.0 is used a default [[3]
](https://github.com/openscad/openscad/blob/master/src/value.h#L198)

This whole defaulting the Z-Value is coming from scale: [[4]](https://github.com/openscad/openscad/blob/master/src/transform.cc#L89)

Well: That is tricky, even before I try black box testing:
```
translate([-6,0,0])
cube([1]);

translate([-4,0,0])
cube([1,1]);

translate([-2,0,0])
cube([1,1,1]);

translate([0,0,0])
cube([true,true,true]);

translate([2,0,0])
cube([true,2,"3"]);

translate([4,0,0])
cube(["1","2","3"]);
```
Yes that needs some attention.

@plumbum Technically, this has nothing to do with the parser. At least the way our parser is setup, we can only detect this issue later in the process. Breaking the process always breaks backward compatibility, so a warning is more adequate.

nophead:
getVec3 is also used by translate and that should accept two elements without warning, not just scale.

t-paul:
[...]

PR #2537, opened 2018-10-27
make the default value of getVec3 explicit by overloading getVec3

MichaelPFrey:
this goes back to #2392 but actually fixes an undocumented issue, that also show cases why silently defaulting a value is hard to manage: https://github.com/openscad/openscad/blob/a1f1bd502ca633984ff5c7483df046bb63568211/src/func.cc#L847-L857

This function uses the return value of getVec3 (a boolean) to explicitly handle the case, when a Vec2 is supplied to it.
In this specific case, the result is the same, but the code is technically not doing it the way it is written.

In other cases - like https://github.com/openscad/openscad/blob/a1f1bd502ca633984ff5c7483df046bb63568211/src/mainwin.cc#L1783-L1784 - it might be important to know that/how/why the third parameter defaulted to 0.0 when not supplied.

I mean: I am not sure if defaulting the third parameter to 0.0 if not supplied is not technically a bug in it of it self in some contexts (for e.g. #2392), but some SCAD files in the wild may relay on that, so making this default explicit is either way the best compromise.


kintel:
..and we should probably merge with master to get the builds passing

t-paul:
I'm still not 100% sure what exactly broke but it seems to be related to some changes in the Windows API headers on AppVeyor. I've added a workaround which fixes compilation and hopefully has no further effect as it disables a bunch of stuff in headers we don't use anyway.

PR #2538, opened 2018-10-27
Fixed the include path on Windows

nophead:


t-paul:
Thanks, lets see what the other builds have to say :-)...

t-paul:
I guess that was to be expected, now the MXE build is broken. I'll check how we can have both.

nophead:
I can't see why this would break it, was it always broken?

t-paul:
Looks like it's actually using `../src` as it's building in `$SRCDIR/mingw32.static.posix`. So I guess `$$PWD/src` should solve it for both cases.

t-paul:
Meh, that generates `-I/openscad/features/src/hidapi`, [never just trust StackOverflow](http://geek-and-poke.com/geekandpoke/2016/11/27/good-questions) I guess... Qt docs state it's actually `$$_PRO_FILE_PWD_` ... off to another try.

t-paul:
Ok, works on MXE and *should* work on MSYS2 too.

nophead:
This works on MSYS2 ```  HIDAPI_CFLAGS = -I$$_PRO_FILE_PWD_/src/hidapi```.  Is that what you had in mind?

t-paul:
Yep, that is in master now and confirmed on MXE via the [CircleCI build](https://circleci.com/workflow-run/46f29f54-acb9-4fd9-b9df-76e5bec1c238). So if that works for you on MSYS2, we should be good.

PR #2540, opened 2018-10-28
AppVeyor integration

t-paul:
Created the PR so we have a place for discussion :grin:...


appveyor.yml
```diff
@@ -0,0 +1,40 @@
...
+
+image:
+  - Visual Studio 2017
+
+cache:
+  - C:\msys64 -> scripts/msys2-install-dependencies.sh
+
+init:
+  - echo "init"
+#  - ps: iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-rdp.ps1'))
```
kintel: Perhaps add a comment that this is for Remote Desktop debugging?

scripts/msys2-install-dependencies.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+#pacman --noconfirm --ask 20 -Sy
+#pacman --noconfirm --ask 20 -Su
```
kintel: What are these commented-out lines?

t-paul: That is for core / system update. I'm still not sure if it's required. At some point it seemed to have broken the installed cmake but it might just be all those installations in path breaking things.

appveyor.yml
t-paul: added comment

PR #2541, opened 2018-10-28
No longer clears the console.

nophead:
A blank line is printed instead to separate each compilation.

The console clearing was not consistent because it depended on whether the main file changed or just its dependencies. It also cleared duplicate assignment warnings and the start up messages prematurely.

### DIFF
--- a/src/mainwin.cc
+++ b/src/mainwin.cc
@@ -432,7 +432,7 @@ MainWindow::MainWindow(const QString &filename)
 
 	setCurrentOutput();
 
-	std::string helptitle = "OpenSCAD " + openscad_versionnumber +  "\nhttp://www.openscad.org\n\n";
+	std::string helptitle = "OpenSCAD " + openscad_versionnumber +  "\nhttp://www.openscad.org\n";
 	PRINT(helptitle);
 	PRINT(copyrighttext);
 	PRINT("");
@@ -1048,7 +1048,7 @@ void MainWindow::compile(bool reload, bool forcedone, bool rebuildParameterWidge
 		auto mtime = this->root_module->handleDependencies();
 		if (mtime > this->deps_mtime) {
 			this->deps_mtime = mtime;
-			PRINTB("Module cache size: %d modules", ModuleCache::instance()->size());
+			PRINTB("Used file cache size: %d modules", ModuleCache::instance()->size());
 			didcompile = true;
 		}
 	}
--- a/src/mainwin.cc
+++ b/src/mainwin.cc
@@ -2873,7 +2873,7 @@ void MainWindow::consoleOutput(const QString &msg)
 	}
 	else {
 		qmsg = msg;
-        qmsg.replace("\n","<br>");
+		qmsg.replace("\n","<br>");
 	}
 	auto c = this->console->textCursor();
 	c.movePosition(QTextCursor::End);
--- a/src/mainwin.cc
+++ b/src/mainwin.cc
@@ -1048,7 +1048,7 @@ void MainWindow::compile(bool reload, bool forcedone, bool rebuildParameterWidge
 		auto mtime = this->root_module->handleDependencies();
 		if (mtime > this->deps_mtime) {
 			this->deps_mtime = mtime;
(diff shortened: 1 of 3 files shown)


t-paul:
I think that's a good idea, but we may need some way to limit the amount in the buffer, e.g. in case an animation is running.

nophead:
Good point and it looks easy to do.

nophead:
Well not as straightforward as I thought. I needed to change QTextEdit to QPlainTextEdit. That handles HTML slightly differently. The net effect is that when you put HTML in echo statements its effects don't continue across lines. For example if you set bold it only lasts for the current echo, whereas before you could have <b> in one echo and </b> in another and anything in between would be bold.

There is an old bug where sometimes the coloured warnings leak and colour the rest of the console, which I assume would now be fixed but if people use HTML echos then I expect this change would break some code.

t-paul:
I'd vote for completely disabling HTML in console. I'm not sure if that was really intended to be there or just happenend due to Qt supporting it. It still would be used internally for the ERROR/WARNING markers though.

nophead:
Its documented in the wiki, so I think it's too late to remove it. I also think I have heard people mention on the mailing list that they use it.

t-paul:
It's still causing random errors with unexpected translations (e.g. #1765). Maybe we need a real "output window" with some more user controlled output, but I think user space HTML in the console window is a feature that will never really work and probably never was intended to. Documenting a bug might be an option sometimes, but I would still rate this as a bug not a feature.

nophead:
Actually I think it will fix that issue somewhat. It seems better behaved. Not sure how you can have a tab in HTML though.

nophead:
This is what it does with @MichaelAtOz's tests:

image

Basically newline works as expected because I translate it to ```<br>```. 
Tabs inhibit HTML interpretation, as before, because when I detect them I add the whole string as plain text. Tabs don't make sense in HTML, they would be ignored. 

Net result is this is a slight improvement because you can now mix ```\n``` with HTML and HTML can no escape from the echo and pollute the rest of the console.

t-paul:
Hmm...
```
echo("hallo: <b>a \u0021 c</b>");
echo("hallo: <b>a \u0009 c</b>");
```
grafik

nophead:
Yes but tabs and HTML has never worked and can't really.

MichaelAtOz:
Note I understand people have done extensive work with HTML in the console, I recall BOMs and a fancy documentation system. Would you like me to ask the forum?

I don't particularly care, but if it goes, for debugging you really something in echo() to highlight a line something like echo("\bText") - appropriating '\b' for bold, a cludge I know.

t-paul:
Yes, I agree, it's a separate issue which should not block the changes of this PR.

src/mainwin.cc
```diff
@@ -1955,7 +1955,7 @@ void MainWindow::actionRenderPreview(bool rebuildParameterWidget)
 
 void MainWindow::csgRender()
 {
-	if (this->root_node) compileCSG(!viewActionAnimate->isChecked());
+	if (this->root_node) compileCSG();
```
kintel: We used to disable event processes during animation, probably to reduce flicker.
I just noticed that compileCSG() now always processes events. Would that have negative effects on animations? @t-paul any ideas?
I know that this originated in a commit I made 11 months ago; this must have been an oversight..

t-paul: I'm not sure, I found that unused warning a couple of days ago too when looking at some build logs, but was not sure where it's coming from. I don't remember a specific discussion, so oversight is a possibility.

MichaelAtOz:
So how far back does the console go, is it ever cleared? 
Perhaps 'clear cache' should clear it too (as the old stats are not relevant), or a new menu item?
Perhaps some may want it to persist for a long time tho?

nophead:
It is an arbitrary 500 lines, can easily be changed. Not sure clearing it with the cache is right because I often want to see the time difference to render when the cache is clear.

A clear option on the right click menu would be good. QPlainTextEdit would need to be subclassed to extend the context menu to do that.

As there is now a blank line between each compilation it is easy to see the last one, which wasn't the case before.

Issue #1843, opened 2016-10-30
Add line number to Warnings


MichaelAtOz:
Any chance? 

```
WARNING: Ignoring unknown variable 'DysonV'. 
ECHO: "recessV8", i = 1, logo = false, dysonV = 8, 2.5
WARNING: Ignoring unknown variable 'DysonV'. 
ECHO: "recessV8", i = 1, logo = false, dysonV = 8, 2.5
```

kintel:
We're getting closer do being able to do this - we recently added location tracking of all expressions. We just don't have that information available where we output that warning, so we still need a bit of refactoring.

t-paul:
I guess this should be "add file name and line number" so it works for multi file designs too. I've added the line number to `assert()` as it is available already, we still need some way to determine the file name.

kintel:
Let's open a separate issue for filename as that's a slightly different task.

Ma-XX-oN:
Don't forget the Errors as well.

MichaelPFrey:
which Warnings and/or Errors would also benefit from the line number (at later a relative path)?
Best is a short snippet of code that triggers the warning/error.

nophead:
I think all errors and warnings should show line number and file. Otherwise, with a large library, it can be like trying to find a needle in haystack.

MichaelPFrey:
The problem with "all" is, that not all errors and warnings belong (meaningfully) to a specific line.


One example for an ERROR not belonging to user code:

https://github.com/openscad/openscad/blob/1bcfae4388c6b4674422e03fba88510260ceffe8/src/PlatformUtils.cc#L134-L150

One example where the ERROR should not be able to be triggered by user code: (meaning I can not write a regression test for it)

https://github.com/openscad/openscad/blob/1bcfae4388c6b4674422e03fba88510260ceffe8/src/context.cc#L128

One example where a call stack would be more useful then a line number:

https://github.com/openscad/openscad/blob/1bcfae4388c6b4674422e03fba88510260ceffe8/src/exceptions.h#L24-L28

MichaelPFrey:
here we have a few:
```
polygon(points=[[0,0],[100,0],[130,50],[30]]);

polyhedron(
  points=[ [10,10,0],[10,-10,0],[-10,-10,0],[-10,10,0],
           [0]  ],
  faces=[ [0,1,4],[1,2,4],[2,3,4],[3,0,4],
              [1,0,3],[2,1,3] ]
 );
```

[...]

PR #2543, opened 2018-10-31
add line number to "Ignoring unknown variable" Warning

MichaelPFrey:
this partially addressees  #1843

I would be trival to output the filename, however: The filename is not very use full on its own.
Outputting the path breakes the tests. Uncompleting the path requires the path of the main file - the patch required for that is partially in #2520, so I would postpone outputting the filename until #2520 is merged.

### DIFF
--- a/src/AST.cc
+++ b/src/AST.cc
@@ -3,6 +3,10 @@
 
 const Location Location::NONE(0, 0, 0, 0, nullptr);
 
+bool Location::isNone() const{
+	return (firstLine()==0 && firstColumn()==0 && lastLine()==0 && lastColumn()==0 && filePath()==nullptr);
+}
+
 std::ostream &operator<<(std::ostream &stream, const ASTNode &ast)
 {
 	ast.print(stream, "");
--- a/src/context.cc
+++ b/src/context.cc
@@ -122,7 +122,7 @@ void Context::apply_variables(const Context &other)
 	}
 }
 
-ValuePtr Context::lookup_variable(const std::string &name, bool silent) const
+ValuePtr Context::lookup_variable(const std::string &name, bool silent, const Location &loc) const
 {
 	if (!this->ctx_stack) {
 		PRINT("ERROR: Context had null stack in lookup_variable()!!");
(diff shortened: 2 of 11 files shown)


MichaelPFrey:
I will add some more test cases to get more ERRORs and WARNINGs and then try to output the line numbers.

MichaelPFrey:
bildschirmfoto von 2018-11-01 10-34-25
sadly, I could not get the regression test to show the error. It seams like the echo-test not creating the geometry, but when we do create geometry, we do not log/compare the echo.

MichaelPFrey:
bildschirmfoto von 2018-11-01 19-14-19

MichaelPFrey:
I am happy so far.

This does not fully cover #1843, but outputing the line number for imports is much harder and I am not yet sure of the full list of Errors and Warnings that benefit from Line Numbers.

The warning creation will be revised later to add the file name and uncompleted path.

src/AST.cc
```diff
@@ -3,6 +3,10 @@
 
 const Location Location::NONE(0, 0, 0, 0, nullptr);
 
+bool Location::isNone() const{
+	return (firstLine()==0 && firstColumn()==0 && lastLine()==0 && lastColumn()==0 && filePath()==nullptr);
```
kintel: To avoid duplication: Instead of hardcoding the magic values in `NONE`, could we get those values from the `NONE` instance?
..or perhaps even implement the `==` operator?

src/context.cc
```diff
@@ -144,24 +144,28 @@ ValuePtr Context::lookup_variable(const std::string &name, bool silent) const
...
 	if (this->parent) {
-		return this->parent->lookup_variable(name, silent);
+		return this->parent->lookup_variable(name, silent, loc);
 	}
 	if (!silent) {
-		PRINTB("WARNING: Ignoring unknown variable '%s'.", name);
+		if(loc.isNone()){
+			PRINTB("WARNING: Ignoring unknown variable '%s'.", name);
+		}else{
+			PRINTB("WARNING: Ignoring unknown variable '%s', line %d.", name% loc.firstLine());
```
kintel: Is there a nice way we could avoid this repetition? I fear this may result in the two strings getting out of sync at some point.

src/AST.cc
MichaelPFrey: implementing == sounds like an awesome idea - I can imagine that we will find other uses for != and == later.

src/context.cc
MichaelPFrey: Work in progress. I have a good idea.

In theory, everything that can cause an error should have a location associated.
I still want to specifically handle the "location none" case as a fail safe, particularly when I later want to access the path/file (which is a nullpointer for location none), but the warning/errors without a line number should be rarely seen.

On a side note: I assume that translating errors and warnings is not a mid-therm goal - concatenating strings is easy, but can create awkward sentences/construct to translate.

kintel:
@MichaelPFrey Could you try merging with master and see if the build errors get resolved?

MichaelPFrey:
Errors and warnings are everywhere in the test cases, so changing one warning affects off course multiple tests.

PR #2547, opened 2018-11-02
Better error message for Recursion Exception

MichaelPFrey:
For tracing a recursion, a call stack would be best, but this is (currently) out of reach.

So, let us see if something lower hanging could be potentially be useful.

### DIFF
--- a/tests/regression/echotest/recursion-test-function-expected.echo
+++ b/tests/regression/echotest/recursion-test-function-expected.echo
@@ -1,1 +1,1 @@
-ERROR: Recursion detected calling function 'crash'
+ERROR: Recursion detected calling function 'crash' in file ../../../../../testdata/scad/misc/recursion-test-function.scad, line 1
--- a/tests/regression/echotest/recursion-test-function-expected.echo
+++ b/tests/regression/echotest/recursion-test-function-expected.echo
@@ -1,1 +1,1 @@
-ERROR: Recursion detected calling function 'crash' in file ../../../../../testdata/scad/misc/recursion-test-function.scad, line 1
+ERROR: Recursion detected calling function 'crash' in file recursion-test-function.scad, line 1
--- a/src/FileModule.cc
+++ b/src/FileModule.cc
@@ -181,6 +181,12 @@ AbstractNode *FileModule::instantiateWithFileContext(FileContext *ctx, const Mod
 		auto instantiatednodes = this->scope.instantiateChildren(ctx);
 		node->children.insert(node->children.end(), instantiatednodes.begin(), instantiatednodes.end());
 	}
+	catch (RecursionException &e) {
+		const auto docPath = boost::filesystem::path(ctx->documentPath());
+		const auto uncPath = boostfs_uncomplete(e.loc.filePath(), docPath);
+
+		PRINTB("%s in file %s, line %d", e.what() % uncPath.generic_string() % e.loc.firstLine());
+	}
 	catch (AssertionFailedException &e) {
 		const auto docPath = boost::filesystem::path(ctx->documentPath());
 		const auto uncPath = boostfs_uncomplete(e.loc.filePath(), docPath);
--- a/src/UserModule.cc
+++ b/src/UserModule.cc
@@ -40,7 +40,7 @@ std::deque<std::string> UserModule::module_stack;
 AbstractNode *UserModule::instantiate(const Context *ctx, const ModuleInstantiation *inst, EvalContext *evalctx) const
 {
 	if (StackCheck::inst()->check()) {
-		throw RecursionException::create("module", inst->name());
+		throw RecursionException::create("module", inst->name(),loc);
 		return nullptr;
 	}
 
--- a/src/boost-utils.cc
+++ b/src/boost-utils.cc
@@ -101,7 +101,7 @@ boostfs_uncomplete(fs::path const p, fs::path const base)
 
 	// create absolute paths
 	fs::path abs_p = fs::absolute(boostfs_normalize(p));
-	fs::path abs_base = fs::absolute(base);
+	fs::path abs_base = fs::absolute(boostfs_normalize(base));
 
 	fs::path from_path, from_base, output;
 
--- a/src/exceptions.h
+++ b/src/exceptions.h
@@ -21,13 +21,16 @@ public:
 
 class RecursionException: public EvaluationException {
 public:
-	static RecursionException create(const char *recursiontype, const std::string &name) {
+	static RecursionException create(const char *recursiontype, const std::string &name, const Location &loc) {
 		std::stringstream out;
 		out << "ERROR: Recursion detected calling " << recursiontype << " '" << name << "'";
-		return RecursionException(out.str());
+		return RecursionException(out.str(), loc);
 	}
 	~RecursionException() throw() {}
 
+public:
+	Location loc;
+
 private:
-	RecursionException(const std::string &what_arg) : EvaluationException(what_arg) {}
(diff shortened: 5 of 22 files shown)


MichaelPFrey:
examples

nophead:
Nice. Does it also work for built in functions like max? Often in a real work recursive function the stack overflow gets detected in a built in function call by the recursive function. It is then very difficult to know where it was called from.

t-paul:
Yep, improvement to point out the place in the source code instead of things like the built-ins would be nice. A full call stack is often not very helpful for recursion anyway, but right now the result is a bit random.

I've renamed the issue as the term localization is often used in context of language translation.

MichaelPFrey:
@nophead it would be helpful to get some "real world" recursive code to test against.

(especially as our test framework has some code/cases not covered)

nophead:
Sorry I can't remember what project I was working on when I got recursion error in a built in function and it isn't something I can search for.

Here is a made up function that shows the problem though:
```
function recurse(x) = max(x, recurse(x));

x = recurse(1);

ERROR: Recursion detected calling function 'max' 

```

MichaelPFrey:
@nophead that code is a good base for me to work on.
The output is now:
```
function recurse(x) = max(x, recurse(x));

x = recurse(1);

ERROR: Recursion detected calling function 'max' in file openscad, line 1 
```
In that case, getting line 1 (and the filename) is very helpful.

MichaelPFrey:
Okay, the linux test is showing an issue:
```
The following tests FAILED:
	 75 - echotest_recursion-test-function3 (Failed)
	 76 - echotest_recursion-test-function4 (Failed)
```
Okay, the MacOSX test is showing something else:
```
The following tests FAILED:
	 75 - echotest_recursion-test-function3 (Failed)
```
and the result from appevor is confusing:
```
The following tests FAILED:
	 76 - echotest_recursion-test-function4 (Failed)
```

```
-ERROR: Recursion detected calling function 'max' in file ../../../../../testdata/scad/misc/recursion-test-function4.scad, line 2
+ERROR: Recursion detected calling function 'recurse' in file ../../../../../testdata/scad/misc/recursion-test-function4.scad, line 2

```

looking at the code

https://github.com/openscad/openscad/blob/467a706f5ffe99e41e3ddea5cc26dc3c9a3fa524/src/expr.cc#L427-L437

https://github.com/openscad/openscad/blob/467a706f5ffe99e41e3ddea5cc26dc3c9a3fa524/src/stackcheck.cc#L28-L31

Interesting - so which function is getting blamed depends on stack size and stack usage (meaning seemingly random for the user). Yeah, reliably getting a line number and file name is a major improvement. 

The limit for how often a function can call it self then also depends on available stack.

Well, I need to think about that.

nophead:
Yes I suspected that might the case. The error message is fine in both cases, it is just a problem for the regression test. It needs to accept either result.

Thanks for working on this, it makes a huge improvement when working on large projects.

kintel:
@MichaelPFrey One idea to make the recursion check more deterministic could be to move the stack check into each overloaded `Context::evaluate_function()`. If the function triggering the overflow was a builtin function (i.e. in `BuiltinContext`), we could go one step up the context stack and report the last user-defined function.

This could of course be used to build a stack trace later on.

tests/regression/echotest/recursion-test-function-expected.echo
```diff
@@ -1 +1 @@
-ERROR: Recursion detected calling function 'crash'
+ERROR: Recursion detected calling function 'crash' in file ../../../../../testdata/scad/misc/recursion-test-function.scad, line 1
```
kintel: Why is the path so long. Wouldn't the file path and doc path be the same folder?

MichaelPFrey: We have the same idiosyncrasy in the assertion tests:

see #2341 and specifically https://github.com/openscad/openscad/pull/2341#issuecomment-373976212

https://github.com/openscad/openscad/blob/467a706f5ffe99e41e3ddea5cc26dc3c9a3fa524/tests/regression/echotest/assert-expression-fail1-test-expected.echo#L1

Looking at it in detail, the issue is our uncomplete:

https://github.com/openscad/openscad/blob/467a706f5ffe99e41e3ddea5cc26dc3c9a3fa524/src/boost-utils.cc#L79

https://svn.boost.org/trac/boost/ticket/1976#comment:2

I am surprised - we literally use 10 year old code where the author is anonymous and call the code not well tested.

I think I found it:

https://github.com/openscad/openscad/blob/467a706f5ffe99e41e3ddea5cc26dc3c9a3fa524/src/boost-utils.cc#L103-L104

either normalize neither or both.

----

As of boost 1.60.0 we could use `lexically_relative(const path& p) const` but I do not know which version we want to support.

t-paul: Added some newer info to https://github.com/openscad/openscad/wiki/Target-Platforms

nophead:
I had a play with module recursion and it seems it always names a user module, never a built in. It isn't always the module that actually recurses though, as with functions. In general the line number will call it out but perhaps not always.

```
module a() children();
module b() children();
module c() children();
module d() children();
module e() children();

module recurse(x) {
    a() b() c() d()
        translate(x)
            e()
                rotate(x)
                    recurse();
}
    
recurse(1);

ERROR: Recursion detected calling module 'd' 
```

So it looks like the stack check is done in built in functions but not built in modules. Perhaps it would be better if it wasn't done in built in functions.

tests/regression/echotest/recursion-test-function-expected.echo
kintel: Btw., Travis still uses 14.04 Trusty in production. 16.04 Xenial is in beta - hopefully it will be stable soon.

kintel: @MichaelPFrey The reason this code is a bit weird is that we used to use Qt for this, and when porting to Boost, we lost some convenience methods and went hunting into the future for the missing function...

MichaelPFrey: I read the history of the source code. I was just surprised, but looking into it, it all makes sense.

Boost tends to be very conservative when it comes to path, as symlinks and  relatives paths (and a combination of those and many more) can be used for security exploits (bypassing path checks).

Our use case (that I am massively expanding - originally, uncomplete was originally just used once) has little security implication, as we use the uncompleted path just as a string outputed to the UI for user convenience. (without providing clickable links or "any of that". The user has to actively locate and open the file.)

We never use the uncomplete path to save a path or open a path or file.
We just display the uncompleted path.
So we are "safe" in the context of our use case, even when the method it self might have flaws.

Still as always: Using C++ standard library (C++17) or boost (1.60.0) "when possible" always simplifies things, tends to be safer etc., but using a workaround for the mean time is of course fine, as long as we keep the implications in mind and to upgrade when the times comes (which will be in many years from now).

t-paul: I updated the wiki, for now it's 14.04 then.

If it helps a lot, conditionally using boost 1.60 for very specific features might be ok so we can benefit on newer systems while still having a fallback for everything too old.

MichaelPFrey: That is an idea, but we need to maintain our uncomplete either way.
I would wait for conditionally using boost 1.60 until most errors and warnings output the filename and uncompleted path, so that we have a lot test cases for comparing boosts and our uncomplete.

kintel:
Reporting recursion only for user-space functions, we could try moving the check to `ModuleContext::evaluate_function()`.

kintel:
@MichaelPFrey Sure, that's what I expected, just wanted you to ack it :)

PR #2571, opened 2018-11-11
Coverity fixes

t-paul:


src/Camera.cc
```diff
@@ -3,7 +3,7 @@
 #include "printutils.h"
 
 Camera::Camera() :
-	projection(ProjectionType::PERSPECTIVE), fov(22.5), viewall(false)
+	projection(ProjectionType::PERSPECTIVE), fov(22.5), viewall(false), autocenter(false), height(0.0)
```
kintel: `height` doesn't seem to be used anywhere

src/transform.cc
```diff
@@ -152,8 +152,9 @@ AbstractNode *TransformModule::instantiate(const Context *ctx, const ModuleInsta
 	else if (this->type == transform_type_e::TRANSLATE)	{
 		auto v = c.lookup_variable("v");
 		Vector3d translatevec(0,0,0);
-		v->getVec3(translatevec[0], translatevec[1], translatevec[2], 0.0);
-		node->matrix.translate(translatevec);
+		if (v->getVec3(translatevec[0], translatevec[1], translatevec[2], 0.0)) {
+			node->matrix.translate(translatevec);
```
kintel: I believe the new code is correct, but are there cases where this will break existing designs?

I also wonder if we generally should warn/err when this happens. Perhaps food for a future issue.

src/Camera.cc
t-paul: Oh, oops, indeed. Removed now.

src/transform.cc
t-paul: I think the only case where this would "break" existing designs is things like
```
translate([10, 0, "ohno"]) cube();
```
where the toDouble() returns false. All other cases are not touching the return values at all so it's leaving all axis values at 0.

I'm not sure that behavior is worth protecting, it's quite random as it depends on the order the axis values are retrieved internally. A warning would be useful though.

MichaelPFrey: @kintel #2537 was a preparation for creating a warning for cases like  #2392. I agree that we should generally create a warning when we ignore parameters for some reason. Error is to strong in my opinion, you never know if someone has designed something that relies on silently ignoring errors.

The more we look around the getVec3 calls, the more interesting things we find
(e.g. check against infinity, 0, ...) that could be generalized and extended (NaN, -inf, ...).

I open #2575 to track the work in progress.

Issue #2578, opened 2018-11-13
Segfault on program start, linux mint 19


thehans:
I am getting segfault immediately on running the nightly, so I compiled from master branch with debug info and ran gdb:
```
(gdb) run
Starting program: /home/hans/projects/openscad_baseline/openscad
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7fffe7aa1700 (LWP 14332)]
[New Thread 0x7fffe55d3700 (LWP 14348)]

Thread 1 "openscad" received signal SIGSEGV, Segmentation fault.
0x00007ffff3756280 in QTimer::stop() () from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
(gdb) bt
#0  0x00007ffff3756280 in QTimer::stop() () from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#1  0x000055555593fbda in InputDriverManager::closeDrivers (this=0x555556b5b510) at src/input/InputDriverManager.cc:135
#2  0x000055555579a022 in Preferences::setupFeaturesPage (this=0x555556718c80) at src/Preferences.cc:329
#3  0x000055555579e00d in Preferences::create (colorSchemes=...) at src/Preferences.cc:801
#4  0x00005555558da83e in MainWindow::MainWindow (this=0x5555563badf0, filename=...) at src/mainwin.cc:202
#5  0x00005555558c04c4 in gui (inputFiles=std::vector of length 1, capacity 1 = {...}, original_path=..., argc=1, argv=0x7fffffffda68) at src/openscad.cc:760
#6  0x00005555558c3035 in main (argc=1, argv=0x7fffffffda68) at src/openscad.cc:1072
(gdb) thread 2
[Switching to thread 2 (Thread 0x7fffe7aa1700 (LWP 14332))]
#0  0x00007ffff1e22bf9 in __GI___poll (fds=0x7fffe7aa0928, nfds=1, timeout=-1) at ../sysdeps/unix/sysv/linux/poll.c:29
29      ../sysdeps/unix/sysv/linux/poll.c: No such file or directory.
(gdb) bt
#0  0x00007ffff1e22bf9 in __GI___poll (fds=0x7fffe7aa0928, nfds=1, timeout=-1) at ../sysdeps/unix/sysv/linux/poll.c:29
[...]

PR #2580, opened 2018-11-13
Fix segfault when input driver feature not enabled

thehans:


thehans:
For Issue #2578

src/input/InputDriverManager.cc
```diff
@@ -132,7 +132,9 @@ std::string InputDriverManager::listDrivers() const
 
 void InputDriverManager::closeDrivers()
 {
-    timer->stop();
+    if (timer != nullptr) {
+        timer->stop();
+    }
     InputEventMapper::instance()->stop();
```
t-paul: Thanks, I guess need to review the remaining part of this method too. Looks like InputEventMapper creates a QTimer in the constructor.

Issue #2575, opened 2018-11-12
create warning when vector pased to inbuilt module is invalid


MichaelPFrey:
Two examples:
`translate([10, 0, "ohno"]) cube([2,2]);`
Note that sometimes we want to default the second parameter.

To find such cases, the code can be searched for getVec3 and getVec2.

Partially, does warnings are already in place:
https://github.com/openscad/openscad/blob/3839d0ee1c66336a8fa05cb81a2f8195f0c2f299/src/primitives.cc#L505
https://github.com/openscad/openscad/blob/3839d0ee1c66336a8fa05cb81a2f8195f0c2f299/src/primitives.cc#L560-L561

Those can be used as templates.

Note that some times we do check if the values are infinity (sometimes not).
https://github.com/openscad/openscad/blob/3839d0ee1c66336a8fa05cb81a2f8195f0c2f299/src/primitives.cc#L559
Note that getVec2 has an ignoreInfinite option:
https://github.com/openscad/openscad/blob/3839d0ee1c66336a8fa05cb81a2f8195f0c2f299/src/value.cc#L354
https://github.com/openscad/openscad/blob/3839d0ee1c66336a8fa05cb81a2f8195f0c2f299/src/value.cc#L363-L365
Some times, only positive values are meaningful (see cube).

MichaelPFrey:
making progress:
![bildschirmfoto von 2018-11-13 19-05-25](https://user-images.githubusercontent.com/24962768/48434560-52718d80-e77a-11e8-9df4-da1b1bd68505.png)
offcourse beeing pased on  #2543, without line numbers it would be really tricky.
Regression testing will be tricky due to  #2566.

Note that not all getVec calls should get warnings. This one for e.g. belongs to an deprecated feature.
https://github.com/openscad/openscad/blob/6670a636bcad0e2f6db0ead24e7e3707eab56708/src/rotateextrude.cc#L82

MichaelAtOz:
`cube();` and `sphere();` etc which default to 1, are commonly used in testing. I doubt it is in usable code tho.

Is there any practical difference between a WARNING and an ERROR? The above looks like ERROR still compiles objects.

nophead:
cube(n) is supposed to work is not? And I thought the default value was 1 so cube(); should be valid.

MichaelPFrey:
> Is there any practical difference between a WARNING and an ERROR? 

Opps... yes they should be warnings, not erros. Thank you, I will fix that.

[...]

PR #2589, opened 2018-11-16
warning when the parameter could not be converted

MichaelPFrey:
bildschirmfoto von 2018-11-16 21-45-17

this partially addresses #2575

### DIFF
--- a/src/primitives.cc
+++ b/src/primitives.cc
@@ -197,7 +197,7 @@ AbstractNode *PrimitiveModule::instantiate(const Context *ctx, const ModuleInsta
 			converted |= size->getDouble(node->z);
 			converted |= size->getVec3(node->x, node->y, node->z);
 			if(!converted){
-				PRINTB("WARNING: Unable to convert CUBE size=%s to a number or a vec3 of numbers, %s", size->toString() % inst->location().toString());
+				PRINTB("WARNING: Unable to convert cube(size=%s, ...) parameter to a number or a vec3 of numbers, %s", size->toEchoString() % inst->location().toString());
 			}
 		}
 		if (center->type() == Value::ValueType::BOOL) {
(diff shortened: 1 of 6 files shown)


src/primitives.cc
```diff
@@ -190,10 +190,16 @@ AbstractNode *PrimitiveModule::instantiate(const Context *ctx, const ModuleInsta
...
-		size->getDouble(node->z);
-		size->getVec3(node->x, node->y, node->z);
+		if(size != ValuePtr::undefined){
+			bool converted=false;
+			converted |= size->getDouble(node->x);
+			converted |= size->getDouble(node->y);
+			converted |= size->getDouble(node->z);
+			converted |= size->getVec3(node->x, node->y, node->z);
+			if(!converted){
+				PRINTB("WARNING: Unable to convert CUBE size=%s to a number or a vec3 of numbers, %s", size->toString() % inst->location().toString());
```
kintel: I wonder if we should print `CUBE` in lower case here, perhaps even `cube(size=%s)`.
Same goes for other similar print statements.

kintel:
@nophead You tend to have valuable input on console output. Any concerns with this?

src/primitives.cc
MichaelPFrey: `cube` is fine by me. `cube(size=%s)` can get confusing, as we do not show the state of for e.g. `center`. Note that cube is the simplest case, other modules have a much more complicated signature.
We could use [Ellipsis](https://en.wikipedia.org/wiki/Ellipsis) to hint add the omission like so `cube(size=%s...)` or so `cube(size=%s, ...)`. Note that we have also have highly complicated built ins like:
`polyhedron(..., triangles = [[0,1,2,3],[5,4,3,2],[0,4,5,1],[0,3,4],[5,2]], ...);`.

Note that it gets ugly when strings come in to play (`cube("[1,2,3]");`) - but this is an issue either-way.

MichaelPFrey:
A screenshot showing why the line number is not always enough to deduce the issue: (partially #2392)
bildschirmfoto von 2018-11-19 21-23-18
```
l = 61;
w = 44;
h = 78 + 7;

thick = 1;

difference() {
    cube([l+thick*2, w+thick*2, h+thick]);
    translate([thick, thick, thick])
       cube([l, w. h+thick*2]);
}

array = [1,[1],[1,2],[1,2,3]];

for (i = [1:4]) {
    cube(array[i]);
}
```

MichaelPFrey:
bildschirmfoto von 2018-11-19 22-34-39

PR #2593, opened 2018-11-18
 output the filename when the error is in an included or used file 

MichaelPFrey:


src/AST.cc
```diff
@@ -23,6 +24,9 @@ bool Location::isNone() const{
 
 std::string Location::toString() const{
 	if(this->isNone()) return "location unkown";
+	if(mainfile && (*path)!=(*mainfile)){
+		return "in file "+boostfs_uncomplete((*path), (*mainfile).parent_path()).generic_string()+ ", "+"line " + std::to_string(this->firstLine());
```
kintel: I find all these parentheses hard to read; `(*path)` => `*path`
also, `(*mainfile).` => `mainfile->`

src/parser.y
```diff
@@ -211,12 +211,12 @@ assignment:
                 bool found = false;
                 for (auto &assignment : scope_stack.top()->assignments) {
                     if (assignment.name == $1) {
-                        auto mainFile = mainFilePath.string();
+                        auto mainFile = (*mainFilePath).string();
```
kintel: Prefer `->` operator.
Same thing in a few other places

src/AST.h
```diff
@@ -11,9 +11,9 @@ class Location {
 
 public:
 	Location(int firstLine, int firstCol, int lastLine, int lastCol,
-			std::shared_ptr<fs::path> path)
+			std::shared_ptr<fs::path> path,std::shared_ptr<fs::path> mainfile)
```
kintel: This concerns me a bit, but my brain is a bit cooked right now.
In "MDI mode", we may compile and cache a module which is used by different main files. Now that we encode the path of the main file into the AST itself, wouldn't we need to also maintain cache keys that includes the file path of the main file?

Instead of going there, are there other ways we can do this; e.g. can we resolve this at run-time in a way where we don't have to keep track of the main file when parsing a module?

MichaelPFrey: I see the problem, but I have no good idea how to resolve it (yet).

The only reason why I need the mainfile is because I need a "stable" output for the regression test, which I want to achieve by uncompleting the path. Alternatively, I could always just output the file name without path. But I want to avoid inconstancy - the assertion and recursion error can uncomplete the path and as they can, they should.

I am tempted to separate the command line output from the GUI output.

In the command line, I can use this code to uncomplete the paths. Ugly as the mainfile is not always the mainfile, but "it works". Hmm... but that assumes that we never implement a command line option with module cache for bulk file output.

In the GUI, I could output the full filename. Or I could  use HTML links - something like just showing the file name which is a link to the full path or has the full path as a tool tip. But currently - and for good reason - we use QT_HTML_ESCAPE to sanitize the warnings. When I find a way to bypass the QT_HTML_ESCAPE for part of the message, I might as well ask the mainwindow which file it has open and then uncomplete with that.

Tricky. Let me dwell on that.

MichaelPFrey: I have a rough idea how I can approach it. If it works, I will close this PR and open a new one.

PR #2594, opened 2018-11-19
Updated lodepng

kintel:


MichaelPFrey:
Currently, we do not output the lodepng version in the versioninformation dialog.
Maybe we want to add it?

kintel:
Since we bundle lodepng, we'll always know the version, but it may be good just for completion

t-paul:
Added now.

PR #2596, opened 2018-11-19
SVG updates

t-paul:
- Adds `center` for SVG, default = false (changing current behavior) / based on page width, so that might only work for "normal" cases like default page export from InkScape. Full support would probably take the viewport settings into account
- A couple of C++11 updates
- Preparation for text support, but that will not be active anytime soon

[Full run](http://files.openscad.org/svg-test/svg-0001.html) of all [OpenClipart](https://openclipart.org/) SVG files (from the Debian package) though the svg import.

ToDo / ToCheck:

- [x] Confirm license clarification as discussed (affecting only src/libsvg)
- [x] This adds `object_parallel_to_source` to the builds need to verify if we need some Qt4 solution -> seems to work ok with Qt4.8.7
- [x] Remove code compilation from test build (still building diffpng for mingw) - continue cleanup?


src/import.cc
```diff
@@ -128,6 +128,9 @@ AbstractNode *ImportModule::instantiate(const Context *ctx, const ModuleInstanti
 	node->origin_x = node->origin_y = 0;
 	origin->getVec2(node->origin_x, node->origin_y);
 
+	auto center = c.lookup_variable("center", true);
+	node->center = center->toBool();
```
MichaelPFrey: In other parts of the code we do a type check before converting:
https://github.com/openscad/openscad/blob/707043efe658e9bf41ea102fafad1c093c7fdbfd/src/primitives.cc#L199
https://github.com/openscad/openscad/blob/707043efe658e9bf41ea102fafad1c093c7fdbfd/src/primitives.cc#L210-L212
I am not sure what that implies, I just see a difference in handling the same variable.

src/importnode.h
```diff
@@ -27,7 +27,7 @@ class ImportNode : public LeafNode
 	std::string layername;
 	int convexity;
 	double fn, fs, fa;
-	double origin_x, origin_y, scale;
+	double origin_x, origin_y, center, scale;
```
MichaelPFrey: why is center a double and not a bool? 

Cross checking with other parts of openscad
https://github.com/openscad/openscad/blob/707043efe658e9bf41ea102fafad1c093c7fdbfd/src/linearextrudenode.h#L23
https://github.com/openscad/openscad/blob/707043efe658e9bf41ea102fafad1c093c7fdbfd/src/primitives.cc#L102
bool seams tobe what we usually use.

tests/CMakeLists.txt
```diff
@@ -1447,6 +1242,7 @@ experimental_tests(cgalpngtest_cube-with-hole
                    astdumptest_assert-expression-fail1-test
                    astdumptest_assert-expression-fail2-test
                    astdumptest_assert-expression-fail3-test
+		   echotest_issue1851-each-fail-on-scalar
```
MichaelPFrey: This appears to be an artifact caused by https://github.com/openscad/openscad/pull/2596/commits/9f6ba917f3e11115a1574b9d3dd1bfb44bc74147 (adding blank line after this line)  https://github.com/openscad/openscad/commit/83e1b6c0d68de2a7aaaaf3790ee9647b224dd581 (removing this line) and https://github.com/openscad/openscad/pull/2596/commits/f09849bf5aa0511822bf1d43daa5848d8ccfc138 (updating to current master)
```suggestion
```

src/import.cc
t-paul: No idea, might be some leftover from experimenting with things. Makes sense to have it using the same logic as in other places.

src/importnode.h
t-paul: see other comment :).

src/libsvg/util.cc
```diff
@@ -1,27 +1,95 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
 #include <boost/spirit/include/qi.hpp>
 
 #include "util.h"
 
+BOOST_FUSION_ADAPT_STRUCT(libsvg::length_struct, (double, number) (std::string, unit))
```
MichaelPFrey: according to travis, ubuntu trusty (14.04) seems to not like this:
```
src/libsvg/util.cc:29:50: error: expected identifier before ‘(’ token
 BOOST_FUSION_ADAPT_STRUCT(libsvg::length_struct, (double, number) (std::string, unit))
                                                  ^
src/libsvg/util.cc:29:59: error: ‘number’ has not been declared
 BOOST_FUSION_ADAPT_STRUCT(libsvg::length_struct, (double, number) (std::string, unit))
                                                           ^
```
Ubuntu 14.04 should use boost 1.54. boost 1.54 should have BOOST_FUSION_ADAPT_STRUCT
https://www.boost.org/doc/libs/1_54_0/libs/fusion/doc/html/fusion/adapted/adapt_struct.html

Maybe a missing include?
```
#include <boost/fusion/adapted/struct/adapt_struct.hpp>
#include <boost/fusion/include/adapt_struct.hpp>
```
The boost headers seams to include a lot of boost headers them self, but this change change with different versions.

t-paul: Yep, lets see if that helps.

t-paul:
Those remaining 2 check boxes were the 2 topics waiting for your approval :smile:. With that given, we are good to go.

PR #2602, opened 2018-11-24
updating the translation

MichaelPFrey:
in general and german specifically. While at it, I also want to address other small things related.


MichaelPFrey:
Okey, the rabbit hole is much deeper then I thought it would be.
This PR is just for reference and to see how the CI will react. I will redo all the steps.

Issue #2062, opened 2017-06-27
"View All" for empty preview sets camera at infinity


thehans:
Steps to reproduce:
1) Start openscad, with new, empty document.
2) Hit F5 to preview
3) Click the "View All" toolbar button

Observe bottom status bar now says "distance: inf"

Scroll wheel zoom no longer is able to affect the camera distance at this point.

This happens in both perspective and orthogonal views.

MichaelAtOz:
(well it is viewing _all_)

thehans:
This now freezes OpenSCAD and requires a force quit.

thehans:
I was looking at the various bugs I have reported and #2014 caught my eye as being the same issue(with OpenSCAD freezing now) as this one.

In that bug, @teepee pointed out that an infinite loop was occurring in GLView::showScalemarkers
So I tried turning off `Show Scale Markers` in the View menu, and the freezing no longer happens (of course camera distance still goes to infinity though)

thehans:
Fixed in PR #2610

Issue #1005, opened 2014-11-02
Enable preview of scenes with all objects in debug mode


kintel:
A scene where all objects are in debug mode won't preview.
e.g.:

```
   %sphere(10);
```

PR #2610, opened 2018-11-29
Fix "View All" for no geometry. Issue #2062

thehans:
If bounding box is empty, then set camera to default distance and centered at [0,0,0] (keeping any rotation).


MichaelPFrey:
```
The following tests FAILED:
	870 - throwntogethertest_issue1005 (Failed)
report saved:
 Testing/Temporary/win_8664_llvmpipe-_llvm-7.0,-256_vdfa_report.html
```
on all major plattforms
bildschirmfoto von 2018-12-01 14-47-01
this goes back #1005

Awesome - with this patch, the test case is now actually more useful then before. So:

`TEST_GENERATE=1 ctest -R throwntogethertest_issue1005`

should do the trick.

Issue #1857, opened 2016-11-06
Cursor-centric mousehweel zooming


thehans:
This is a feature request to alter the zoom behavior so that points under the mouse cursor stay under the mouse cursor when zooming.   

This is easier to explain and conceptualize when dealing with a strictly 2d application, such as html/js canvas, in which I have successfully implemented this feature before.  In the 2d case the cursor position always corresponds with a single point in "world" space.  The goal then, when cursor-centric zooming, is to keep that world space coordinate corresponding with the same screen-space coordinate during the zoom operation.

Zooming in this way makes it very convenient to zoom out, adjust cursor pos, and zoom in to a new area very quickly without the need to manually pan in between.

For 3d, things are more tricky, and I don't know a solution off the top of my head, since the cursor now represents a ray in 3d space.  I'm hoping maybe with some discussion an appropriate algorithm could be found.  I am fairly certain I have seen behavior like this in other 3d modeling applications.

thehans:
I just checked FreeCAD, and it does implement this sort of zooming.

nairaner:
If I am correct this is how zoom in SketchUp behaves

BenjamenMeyer:
@thehans perhaps a bug on some platforms?

I'm on Kubuntu 17.10 which provides version 2015.03-2 out-of-the-box and am finding it very difficult to work with things since I can only zoom in and out; I can't change where the focus of the zoom is located. So while trying to model large (14 foot long) work benches, buildings (16 feet by 30 feet) or walls (61 inches by 97 inches) it gets very difficult to see how components line up to make sure placement is correct.

May be I'm doing something wrong (hope so), but this bug report/feature request seems related.

thehans:
Not a bug, this is a feature request.  Yes the zoom currently works as you describe, no matter which platform.  It currently zooms in/out around the origin, or whatever point you have panned the camera to using right-click-drag.

BenjamenMeyer:
@thehans thanks; yes - I understood the original request to be; but from the responses got the impression it might have been viewed as a bug. Either way; I'd love it as the request is more intuitive than the right-click drag - I'll have to try that out. Thanks!

thehans:
Just worked out an implementation that seems to function well.  Just wondering if the old behavior should be kept as a configurable preference?
I also left the shift + right click zoom as is, since the cursor changes position as that is adjusted, it wouldn't make much sense.

thehans:
Closed by #2611

PR #2611, opened 2018-11-29
Implemented cursor-centered mouse wheel zoom.  

thehans:
Issue #1857

### DIFF
--- a/src/QGLView.cc
+++ b/src/QGLView.cc
@@ -312,12 +312,13 @@ bool QGLView::save(const char *filename)
 
 void QGLView::wheelEvent(QWheelEvent *event)
 {
+  auto pos = event->pos();
 #if QT_VERSION >= 0x050000
     int v = event->angleDelta().y();
 #else
     int v = event->delta();
 #endif
-    zoom(v, true);
+    zoomCursor(pos.x(), pos.y(), v);
 }
 
 void QGLView::ZoomIn(void)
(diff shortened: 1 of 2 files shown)


thehans:
Just a note on how I've implemented this.

I first solved the math for orthogonal projection since this was much easier to think about.  

First get the cursor coordinates in "screen space"(where [0,0] is the view center, using [this reference](https://www.scratchapixel.com/lessons/3d-basic-rendering/ray-tracing-generating-camera-rays/generating-camera-rays) for equations).
If we think about what happens when only the camera distance is changed, any object under this point ends up moving in screen space, being scaled by an amount equivalent to the (old_distance / new_distance).  To get only the difference(delta) in screen space, we subtract one from that ratio.  Now this screen space delta is scaled to match the world space, based on the orthogonal frustum size, and the camera is simply translated to correct for that perceived movement under the cursor.

I was originally thinking that after working out the orthogonal projection, I would need a separate, more complex solution for perspective projection.  However from my testing I was pleasantly surprised that it seems to match up close enough that I don't feel like any further corrections are necessary.

src/QGLView.cc
```diff
@@ -312,12 +312,13 @@ bool QGLView::save(const char *filename)
 
 void QGLView::wheelEvent(QWheelEvent *event)
 {
+  auto pos = event->pos();
```
kintel: const

src/QGLView.cc
```diff
@@ -336,6 +337,21 @@ void QGLView::zoom(double v, bool relative)
...
+  auto old_dist = cam.zoomValue();
+  this->cam.zoom(zoom, true);
+  auto dist = cam.zoomValue();
+  auto ratio = old_dist / dist - 1.0; 
+  // screen coordinates from -1 to 1
+  auto screen_x = 2.0 * (x + 0.5) / this->cam.pixel_width - 1.0;
+  auto screen_y = 1.0 - 2.0 * (y + 0.5) / this->cam.pixel_height;
+  auto height = dist * tan(cam.fov /2 * M_PI / 180);
+  auto mx = ratio*screen_x*(aspectratio*height);
+  auto mz = ratio*screen_y*height;
```
kintel: You can const all the local variables in this function. Good for readability as you're promising not to change the values.

src/QGLView.cc
thehans: Thanks for the review!  Sorry, didn't mean to ignore this so long, I went ahead and committed the added `const` to those functions.

PR #2615, opened 2018-12-01
output the filename to which a warning belongs

MichaelPFrey:
bildschirmfoto von 2018-12-01 13-22-05

### DIFF
--- a/src/printutils.cc
+++ b/src/printutils.cc
@@ -1,4 +1,5 @@
 #include "printutils.h"
+#include "boost-utils.h"
 #include <sstream>
 #include <stdio.h>
 #include <boost/algorithm/string.hpp>
@@ -14,6 +15,8 @@ std::string OpenSCAD::debug("");
 bool OpenSCAD::quiet = false;
 
 boost::circular_buffer<std::string> lastmessages(5);
+std::string mainFilename("");
+std::string currentFilename("");
 
 void set_output_handler(OutputHandlerFunc *newhandler, void *userdata)
 {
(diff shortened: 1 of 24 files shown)


src/printutils.cc
```diff
@@ -14,6 +15,8 @@ std::string OpenSCAD::debug("");
 bool OpenSCAD::quiet = false;
 
 boost::circular_buffer<std::string> lastmessages(5);
+std::string mainFilename("");
+std::string currentFilename("");
```
kintel: It's technically unnecessary to initialize std::string as it's empty by default

MichaelPFrey:
A bit of elobration on the idea:
The idea is, that the console/outputhandler has to belong to a window/console and the window/console can/has to know the mainfile it is responsible for.

The printutils are now used as a protocol handler of sorts.

I could output the filename before each warning. But: The "Suppress output after 5 equal ERROR or WARNING outputs" is now triggered much rear, as warnings now tend to differ by line number and/or content, so  outputing the filename for a group of warnings instead of for each warning helps to keep the overview.

In other do avoid people associating the filename with echo outputs, in gets the same yellow background color as the warnings. If any non warning is outputed, the filename is outputed again.
a100

Given that the filename looks like a bit of a header, so not outputing it over warnings from the mainfile would look strange. So:
a101

One issue that this currently has is, that we currently are not at point where all relevant warnings know their location. (see #1843) This can lead a user down the wrong rabbit hole, as the filename beeing outputed is the last filename that  was trasmitted.

nophead:
I think the limit of five warnings the same was to handle deprecations being mentioned over and over again, hiding other errors. Unlike other errors and warnings one might choose to keep them for backwards compatibility rather than fix them.

Perhaps the 5 time limit should go as they will now always differ by line number and instead deprecations should only be mentioned once per feature per compilation.

src/printutils.cc
kintel: Maintaining such state in printutils makes me a bit uneasy. I don't have a great suggestion for alternatives though.
If you don't find any other way, perhaps we could add a comment here for future readers to keep this in mind?

MichaelPFrey: Taking two steps back: I could try to pass the mainfile via context/eval context.

MichaelPFrey:
replaced with #2619, which is closer to #2593

PR #2617, opened 2018-12-02
some fixes & minor improvements (mostly CMake related)

hoijui:
I can now again compile with CMake and run OpenSCAD on my debian stable 64bit system.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,6 +50,7 @@ set(CMAKE_PREFIX_PATH "$ENV{OPENSCAD_LIBRARIES}")
 find_package(Eigen3 REQUIRED QUIET)
 message(STATUS "Eigen3: ${EIGEN3_VERSION_STRING}")
 include_directories(${EIGEN3_INCLUDE_DIRS})
+add_definitions(-DEIGEN_DONT_ALIGN)
 
 find_package(Boost 1.36 REQUIRED COMPONENTS filesystem system thread regex program_options QUIET)
 message(STATUS "Boost: ${Boost_MAJOR_VERSION}.${Boost_MINOR_VERSION}.${Boost_SUBMINOR_VERSION}")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,7 +21,7 @@ project(openscad)
 
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/objects")
 set(AUTOGEN_BUILD_DIR "${CMAKE_BINARY_DIR}/objects")
-add_custom_target(make-objects-directory ALL COMMAND ${CMAKE_COMMAND} -E make_directory ${AUTOGEN_BUILD_DIR})
+file(MAKE_DIRECTORY ${AUTOGEN_BUILD_DIR})
 
 # Default to Release build
 if(NOT CMAKE_BUILD_TYPE)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,6 +44,9 @@ if(APPLE)
   set(CMAKE_CXX_FLAGS "-stdlib=libc++")
 endif()
 
+if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -frounding-math")
+endif()
 add_definitions(-DSTACKSIZE=8388608 -D_REENTRANT)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/Modules/")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,6 +44,8 @@ if(APPLE)
   set(CMAKE_CXX_FLAGS "-stdlib=libc++")
 endif()
 
+add_definitions(-DSTACKSIZE=8388608 -D_REENTRANT)
+
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/Modules/")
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -44,22 +44,38 @@ if(APPLE)
...
 set(CMAKE_PREFIX_PATH "$ENV{OPENSCAD_LIBRARIES}")
 
 find_package(Eigen3 REQUIRED QUIET)
 message(STATUS "Eigen3: ${EIGEN3_VERSION_STRING}")
 include_directories(${EIGEN3_INCLUDE_DIRS})
-
-find_package(Boost 1.36 REQUIRED COMPONENTS filesystem system thread regex program_options QUIET)
-message(STATUS "Boost: ${Boost_MAJOR_VERSION}.${Boost_MINOR_VERSION}.${Boost_SUBMINOR_VERSION}")
-include_directories(${Boost_INCLUDE_DIRS})
+add_definitions(-DEIGEN_DONT_ALIGN)
```
kintel: What does this do? It's a deprecated flag to Eigen

CMakeLists.txt
```diff
@@ -44,22 +44,38 @@ if(APPLE)
...
 
 find_package(CGAL REQUIRED QUIET)
 message(STATUS "CGAL: ${CGAL_MAJOR_VERSION}.${CGAL_MINOR_VERSION}")
 include_directories(${CGAL_INCLUDE_DIRS})
 add_definitions(-DENABLE_CGAL)
 
+# NOTE FindCGAL.cmake resets the BOOOST_* and Boost_* variables,
```
kintel: typo s/BOOOST/BOOST/

CMakeLists.txt
```diff
@@ -44,22 +44,38 @@ if(APPLE)
...
+#   because it looks for Boost itsself, so it has to be called
```
kintel: typo: s/itsself/itself/

CMakeLists.txt
```diff
@@ -44,22 +44,38 @@ if(APPLE)
...
+set(BOOST_DIRECTLY_REQUIRED_LIBRARIES filesystem system thread regex program_options)
+find_package(Boost 1.36 REQUIRED COMPONENTS ${BOOST_DIRECTLY_REQUIRED_LIBRARIES} QUIET)
+message(STATUS "Boost: ${Boost_MAJOR_VERSION}.${Boost_MINOR_VERSION}.${Boost_SUBMINOR_VERSION}")
+include_directories(${Boost_INCLUDE_DIRS})
+
+# Check if boost dependency libraries have been found too
+list(LENGTH BOOST_DIRECTLY_REQUIRED_LIBRARIES BOOST_DIRECTLY_REQUIRED_LIBRARIES_LENGTH)
+list(LENGTH Boost_LIBRARIES Boost_LIBRARIES_LENGTH)
+if(Boost_LIBRARIES_LENGTH EQUAL BOOST_DIRECTLY_REQUIRED_LIBRARIES_LENGTH)
+  message(FATAL_ERROR "No dependent Boost libraries found. Your CMake (${CMAKE_VERSION}) version might be too old for the version of boost you are using (${Boost_MAJOR_VERSION}.${Boost_MINOR_VERSION}.${Boost_SUBMINOR_VERSION}). In that case, you should have recieved warnings above of the type: 'Imported targets not available for Boost version ${Boost_VERSION}'")
```
kintel: typo: s/recieved/received/

CMakeLists.txt
```diff
@@ -44,22 +44,38 @@ if(APPLE)
   set(CMAKE_CXX_FLAGS "-stdlib=libc++")
 endif()
 
+set(CMAKE_C_FLAGS ${CMAKE_C_FLAGS} "-m64 -pipe -O2 -Wall -W -fPIE")
+set(CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS} "-m64 -pipe -fno-strict-aliasing -frounding-math -std=c++0x -O2 -std=c++0x -Wall -W -Wno-unused-local-typedefs -fPIE")
```
kintel: These look like gcc flags and should be wrapped in an appropriate check.
Also, m64 should probably be enabled only for 64-bit builds

CMakeLists.txt
hoijui: I don't know what it does, but it is in the `Makefile` and was before not in the CMake build file. You think we should rather remove it from the Makefile?

CMakeLists.txt
hoijui: I fixed this as you suggested (and also all the typos you mentioned above)

CMakeLists.txt
```diff
@@ -44,22 +44,45 @@ if(APPLE)
   set(CMAKE_CXX_FLAGS "-stdlib=libc++")
 endif()
 
+if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -pipe -O2 -Wall -W -fPIE")
```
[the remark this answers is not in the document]
hoijui: I am pretty ignorant about all the flags, to be honest. ;)
How I came to this set of flags:
I compiled with CMake, and I got segfaults when starting openscad. Then I compiled with the qmake Makefile, and there were no problems; no segfaults. I then compared the actual flags used by the two Makefiles (qmake & CMake generated), and I added all the ones missing from the qmake Makefile to CMake. I compiled wiht CMake, and there were no more segfaults.

So... I would be fine to reduce this set of flags to a minimum, if you have a good guess about which flags are the essential ones here. My own guess would be pretty crude, and because one build takes at least 20min on my system, I'd rather not test through too many combinations of flags.
I am pretty sure that `-frounding-math` is essential though.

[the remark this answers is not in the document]
hoijui: I wrote this commit abou ta year ago. I now tested without any of the flags, and with just rounding math, and neither of the produced binaries gives me any segfaults anymore. ;-)
so.. I willl make a commit with `-frounding-math` for GCC and Clang, and forget about al lthe other flags.
thanks, and sorry for causing you all this work!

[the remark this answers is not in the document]
hoijui: done

Issue #380, opened 2013-05-31
Using the time and date and filename of the current openscad script from within the script.


ivoknutsel:
I'd like to be able to retrieve the file name of the current script from a variable, also the path, date and time would be handy.

Together with a clean echo #382 and assert #381 function the read only parameters would allow using OpenSCAD in a scripted environment in a more robust way.

I propose : 
$mainfilename - the filename of the main file currently being processed
$mainfilepath - the filepath of the main file currently being processed
$currentfilename - the filename of the file this statement is in
$currentfilepath - the filepath of the file this statement is in
$currentdate
$currenttime
##

MichaelPFrey:
would someone mind specifying
```
$currentdate
$currenttime
```
?

OpenSCAD script are interpreted in multiple stages.

Two extreme cases:
If you want to print just a date and time on a part, I would use the date and time of when the parser started.
So
```
echo($currenttime);
//lots of crazy calculation
echo($currenttime);
```
would result in outputting the exact same time.
This is important as
```
echo($currenttime);
echo($currentdate);
```
has to work over midnight. Think of 23:59:59 2019-12-02.

If you want to measure elapsed time, I have two things to say: OpenSCAD is not a general purpose programming language. Calculating time is easier with unix time stamps. Due to the multi-phase nature, it is hard to measure what you might want to measure. 

The next thing would be: What format you want date and time in?

Why specifically
```
$currentdate
$currenttime
```
and not just
```
$date
$time
``?

**Comment by MichaelPFrey** [comment]
```
$mainfilepath - the filepath of the main file currently being processed
$currentfilepath - the filepath of the file this statement is in
```
Do you want absolute or relative paths?
I personally always prefer relative paths as we can write stable tests for them.

MichaelAtOz:
> Perhaps initially exclude the text based bits (ddd, dddd, MMM, MMMM) and look at locale stuff later? 

Add 'AP', 'ap', 'A' & 'a' to the exclusions (ie ignored) as they are also localised. See below re 24 hours.

Also note Qt5  `QString QDate::toString(` [[and `QDateTime` varient]](http://doc.qt.io/qt-5/qdatetime.html#toString) says 'Formats without separators (e.g. "ddMM") are currently not supported', which I think is unfortunate - ruling out the ISO 8601 'basic format'.

Qt4 did not have that qualification, it may be a bug?? Should we work around this as the basic format is often more useful?

[...]

PR #2618, opened 2018-12-02
add line number to various warnings

MichaelPFrey:
going back to #1843 

- [x] wait for #2619
- [x]  merge with master
- [x]  add file name where we already have context
- [x] add location to: (low-hanging as we have loc in the method signature)
   * WARNING: Experimental builtin function '%s' is not enabled.
   * WARNING: Experimental builtin module '%s' is not enabled.
- [x] add location to: (low-hanging as we have ModuleInstantion in the method signature)
   * WARNING: $fs too small - clamping to %f
   * WARNING: $fa too small - clamping to %f
   * Sidenote: What happens when I throw  nan, inf, "string" add it? -> moved to #2620
- [x] add location to: (Abstract Node has a ModuleInstation, which has  a location)
   * WARNING: Mixing 2D and 3D objects is not supported.
   * WARNING: Ignoring radius variable '%s' as diameter '%s' is defined too.
   * WARNING: Ignoring 3D child object for 2D operation
   * [x]  use the correct path to create the relative path
- [x] update testcases


kintel:
I glanced at this and have an idea:
Every time we create an `EvalContext` instance, we have a `Location` available. Would it make sense to simply make the Location part of the EvalContext, and this simplify the various function signatures by not having to pass all those location around?

MichaelPFrey:
Good question.

I have modeled this code (for the functions) based on the code for modules.
So going either-way, we should use the change for gaining some consistency.

Diving deeper:

What springs to my mind is, that the loc object is easily visible and usable.

Having to get it from eval context would of course work. Given that a function may has many warnings, it might make sense to "unpack" the location from the eval context as the first statement of the function.

One of the drawbacks of passing the location explicitly is, that we get the unused parameter warning.
I can off-course just remove the parameter name to avoid the warning, but that means the the loc parameter is not necessarily "visible" when glancing over the code.

kintel:
We can always refactor this later. My main argument is that the "evaluation context" feels like a natural and intuitive place to store a location, so it shouldn't come as a surprise then we casually unpack the location from there; `PRINTB("...", evalctx->loc.toString()`

MichaelPFrey:
Looking back at the various PRs leading upto this: Excellent idea: I open a new issue to track the whole mini-project

MichaelPFrey:
some of the UI tests:
test1

MichaelAtOz:
Just putting on my multi-threaded thinking cap®: All of that happens prior to actually rendering, so no concurrency issues?

MichaelAtOz:
@MichaelPFrey 
Comment in #380 "$linenumber - the line number, in the file (ie $currentfilename), of the line containing this variable", perhaps consider whether this value is easily accessible with what you're up to here. Different debate on how/if we make the value available to code, but pointless unless the line number is readily available.

src/Tree.h
```diff
@@ -16,13 +16,16 @@ class Tree
...
 
 	const std::string getString(const AbstractNode &node, const std::string &indent) const;
 	const std::string getIdString(const AbstractNode &node) const;
+	const std::string getDocumentPath() const;
 
 private:
 	const AbstractNode *root_node;
 	// keep a separate nodecache per tuple of NodeDumper constructor parameters
 	mutable std::map<std::tuple<std::string, bool, bool>, NodeCache>  nodecachemap;
+	std::string document_path="";
```
kintel: Nitpick: Not necessary as std::string initializes itself to empty

src/primitives.cc
```diff
@@ -118,15 +118,16 @@ class PrimitiveNode : public LeafNode
  * @return radius value of type Value::ValueType::NUMBER or Value::ValueType::UNDEFINED if both
  *         variables are invalid or not set.
  */
-Value PrimitiveModule::lookup_radius(const Context &ctx, const std::string &diameter_var, const std::string &radius_var) const
+Value PrimitiveModule::lookup_radius(const Context &ctx, const ModuleInstantiation *inst, const std::string &diameter_var, const std::string &radius_var) const
```
kintel: Why are we passing a `ModuleInstantiation` here, when we usually pass a `Location` reference?

MichaelPFrey: good suggestion. This makes the indention obvious and limits accidential exposure of interfaces.

The only reason I  used ModuleInstantiation was, because I tend to copy paste the line number code.

PR #2619, opened 2018-12-02
output the filename to which a warning belongs using "context"

MichaelPFrey:


MichaelPFrey:
toString is not the best name for the function. the whole PR needs clean up. What I want so see is, if the tests pass and I want feedback whether or not the approach seams reasonable.

To be clear: The implementation is is a hack and beside our automated tests, it is not tested yet.

especially use could have surprises.

kintel:
I think this could work.
When storing a pointer to the `Context` object in a node, please keep the lifetime of that object in mind so you don't end up storing a pointer to a deleted object.

MichaelPFrey:
that was supprisingly straightforward.

src/AST.h
```diff
@@ -24,7 +24,7 @@ class Location {
 	int lastColumn() const { return last_col; }
 	bool isNone() const;
 
-	std::string toString() const;
+	std::string toRelativeString(std::string docPath) const;
```
kintel: We might just as well make this a `const std::string &`

src/context.cc
```diff
@@ -187,21 +187,21 @@ bool Context::has_local_variable(const std::string &name) const
  * @param what what is ignored
  * @param name name of the ignored object
  */
-static void print_ignore_warning(const char *what, const char *name, const Location &loc){
-	PRINTB("WARNING: Ignoring unknown %s '%s', %s.", what % name % loc.toString());
+static void print_ignore_warning(const char *what, const char *name, const Location &loc, const Context ctx){
```
kintel: The Context should be a `&`, but I wonder if it's maybe cleaner to pass the documentPath as a `const string &` instead.

src/primitives.cc
```diff
@@ -96,6 +96,7 @@ class PrimitiveNode : public LeafNode
 			return "unknown";
 		}
 	}
+	const Context *ctx;
```
kintel: As mentioned earlier: Are you certain that this object will survive the PrimitiveNode instance?

Issue #340, opened 2013-04-27
Issue warning on overloading variables?


kintel:
Alan Cox suggests on the mailing list that we should issue a warning on attempts to overload a variable, as it would shadow an existing definition and potentially cause unintended behavior.

Simple example which would in this case cause a warning. In more complex scenarios it would be harder to spot the problem (e.g. overriding variable in an outer/global scope):

```
module mymodule(myvar) {
  myvar = 3;
  sphere(myvar);
}

mymodule(6);
```
##

kintel:
@MichaelPFrey Does your warning work fully or partially fix this?

MichaelPFrey:
@kintel No, not really and I am kind of supprised how granular the scopes are.
My warning only triggers, when a variable is overwritten in the same scope.
The parameters of a function and the function body are different scopes.

To create a warning for
```
module mymodule(myvar) {
  myvar = 3;
  sphere(myvar);
}

mymodule(6);
```
I could implement a check one scope "below" the current one, but then I need to know what kind of scope it is (which has to be implemented) before doing that. Sounds tricky and iterating over two scopes does not sound smart.

Alternatively, we could to use the same scope for the function parameter and function body.
I do not know how hard that would be and if that would have unwanted side effects.

I will look into it.

MichaelAtOz:
> Alternatively, we could to use the same scope for the function parameter and function body.
I do not know how hard that would be and if that would have unwanted side effects.

Tread carefully.

```
module mymodule(myvar,myvar2=myvar) {
  myvar = 3;
  sphere(myvar);
  echo("mymodule: ",myvar=myvar,myvar2=myvar2); // 3, 9
}

myvar=9;
mymodule(6);
echo("main: ",myvar=myvar); // 9
```

MichaelPFrey:
To address this:
```
function test_function() = a;
b=test_function(a=3);
a=4;
echo(b);
```
we need separate scopes for the function parameter and function body - so that makes the decision easier.

MichaelPFrey:
closed with #2628?

In theory, it could be improved, but it is tricky to warn every possible issue, when mean while
```
function clamp(a) = a > 10 ? 10 : a < 0 ? 0: a;

module mytest(a=5){
    a=clamp(a);
    echo(a);
}

mytest(20);
```
is acceptable. Warning, when a parameter is overwritten by a literal is a decent compromise.

PR #2628, opened 2018-12-09
Issue warning when overwritting parameters passed to modul #340

MichaelPFrey:
partially addresses #340 

The main challenge is to get the relevant information to where I need them.


MichaelPFrey:
bildschirmfoto von 2018-12-09 20-57-53

bildschirmfoto von 2018-12-09 21-01-31

bildschirmfoto von 2018-12-09 21-04-02

kintel:
Looking at this again: This could end up being very noisy. Argument shadowing tends to be more of a coding style choice than a real warning.
Perhaps we should have this off by default and let people tune warnings using cmd-line flags or Preferences? ..although that's a fair bit more plumbing to set up..

MichaelPFrey:
I think I am very specif about the warning.

Declaring a parameter just to overwrite it in the module is non-sense:
```
module mytest(a=5){
    a=1;
    echo(a);
}

mytest(2);
```
no matter how I turn it:
```
module mytest(a=5){
    echo(a);
    a=1;
}
```
```
module mytest(a){
    echo(a);
    a=1;
}

mytest(2);
```
```
module mytest(a=1){
    a="test";
    echo(a);
}

mytest();
```

Note that this does not trigger a warning:
```
module mytest(a){
    if(a==2){
        a=1;
        echo(a);
    }
}
mytest(1);
mytest(2);
mytest(3);
```
as the if is its own scope.

Okay, this can be discussed:
```
module mytest(){
    a="test";
    echo(a);
    echo(b);
}

mytest(a=2,b=3);
```
```
WARNING: Module mytest: Parameter a is overwritten, in file openscad, line 2
ECHO: "test"
ECHO: 3
```
but especially when someone assumes that he can use undocumented parameters, the warning that this does not work might be helpful.

Feel free to throw me some test case, maybe I am overlooking a legitimate use case.

But I like the general direction. If someone manages to shadow an assignment in a way, that there is no scenario where the assignment can have any effect, then we should warn.

I would not want to give people the option to turn this on and off. People might want to write libraries. Together with #2630 (and building on top of it https://github.com/MichaelPFrey/openscad/tree/trace), warnings will become are more and more relevant tool. Giving the people options does not help, as libraries are written assuming one or the other.

MichaelAtOz:
I'd also like to point out that we (presuming) all do bad stuff when debugging, overriding a variable in the middle of chunks of code is one such thing. If that happens to be within a repeatedly called nest, think a large `for` in a recursive module, then [warning overload ](https://psychologydictionary.org/warning-overload/)will happen.
I'm wondering if a key code toggle may be appropriate, or some such?

MichaelAtOz:
noting that the preview panel message will/should still show:
warning in gui

MichaelPFrey:
> ```
> function clamp(a) = a > 10 ? 10 : a < 0 ? 0: a;
> 
> module mytest(a=5){
>     a=clamp(a);
>     echo(a);
> }
> 
> mytest(20);
> ```

Okay - I accept that as a legitimate use case.

But at the same time, this does get confusing quickly:
```
function clamp(a) = a > 10 ? 10 : a < 0 ? 0: a;

module mytest(a=5){
    echo(a);  
    a=clamp(a);
    echo(a);
}

mytest(20);
```

```
function clamp(a) = a > 10 ? 10 : a < 0 ? 0: a;

module mytest(a=5){
    echo(a);  
    a=clamp(echo(a) a);
    echo(a);
}

mytest(20);
```
The echo function being executed before the echo modules is what it is, but that we have two different `a` within the same scope, can result in interesting things:
```
function clamp(a) = a > 10 ? 10 : a < 0 ? 0: a;

module mytest(a=5){
    b=min(a,50);
    rotate(min(a,50))cube();
    rotate(b)cube();
    a=clamp(echo(a) a);
}

mytest(45);
```
do be fair, this has nothing to do with modules, as it is a simple scoping issue:
```
function clamp(a) = a > 10 ? 10 : a < 0 ? 0: a;

a=45;
if(true){
    b=min(a,50);
    rotate(min(a,50))cube();
    rotate(b)cube();
    a=clamp(echo(a) a);
}
```

Uff..

MichaelPFrey:
Regarding:

> [warning overload ](https://psychologydictionary.org/warning-overload/)

#2630 should help in that regard.

MichaelAtOz:
> do be fair, this has nothing to do with modules, as it is a simple scoping issue:

That can't be fixed until we 'fix' the -D cmdline problem that causes it.

kintel:
Cool, let's give this a shot :)

Issue #329, opened 2013-04-11
Option to treat warnings as errors when compiling and rendering a .scad file


brantmerryman:
Currently if you **include** or **use** a file inside of a .scad document, and that file is not found, a warning is generated and the object still renders.

Often this means that a missing include file can cause holes to be the wrong size.

To do:
Add a compiler flag to treat warnings as errors. This should be something similar to the gcc options:
     \* -Werror **treat all warnings as errors**
     \* -Werror= _specific error_ **treat a specific warning as an error**
     \* -Wno-error= _specific error_ **treat a specific warning as a warning even when -Werror is in effect**
We would like to at least cover the case of a missing file.

Add a way to specify this inside of the GUI, possibly as a preferences item: "treat warnings as errors".

**Nice to have, but might be too fancy** Add a way to modify the current compiler flags from code - similar to #pragma.
##

5solids:
By the way, does anybody know of any way to cause an error that stops the machinery?

For the time being, I could live with defining my own module and simply calling that in case I'am not happy with a configuration parameter or similar. For instance:

module exit(rc) {
    echo(str("RC: ",rc));
    // something that stop the compiler
}

if (foo == 23) {
    echo("ERROR: invalid foo value");
    exit(1);
}

Another alternative could be an assert() function.

Regarding checking if an include worked, it might be work around to add a kind of eye-catcher variable in the include file and to check its value after the import. But then again it would be great if we would have a way to programmatically stop the compile.

kintel:
For assert(), see #381

jordanbrown0:
+1.  It's quite annoying to make a mistake, have F5 immediately point out the mistake, but then have to wait for a known-bad re-render.

kintel:
@MichaelPFrey This could be a great potential add-on to all the great warning work

MichaelPFrey:
allowing a scad file to overwrite hardwarning "on the fly" would have to be at two or three parter, as the implementation has to take into account the different phases of parsing, evaluating and rendering the script.

Specifying which warnings are effected would require labeling all warnings first.

> From Triffid on the mailing list:
> 
> Greg Frost's extruders import STLs of his gears if they're present, for display purposes only. If they're absent, the warning is harmless. FIXME: Link to these models.
> 
> SO, we should have a way to treat it as an error, and have a way to get a mere warning if the import is only there for display purposes and isn't necessary for the actual output
> 
> perhaps %use or %import could trigger warning-only mode or something?

[...]

PR #2630, opened 2018-12-11
first draft for adresing issue #329

MichaelPFrey:
Todo list
   * [ ] do not throw int, use a proper exception instead
   * [ ]  add GUI option to enable this feature
   * [ ] add cmd option to enable this feature
   * [ ] allow the SCAD script to enable/disable strict warnings?
   * [ ] make sure all excution paths are covered
      * [ ] render 
      * [ ] preview
      * [ ] animation
      * [ ] customizer
      * [ ] ...
   * [ ] make all warnings single line
   * [ ] GUI Code: make a specific methode "exceptionCleanup" that deals with things like the UI lock
   * [ ]  change some of the warnings to errors (e.g. stack context)
   * the UI generating warnings is an issue

### DIFF
--- a/src/mainwin.cc
+++ b/src/mainwin.cc
@@ -1011,6 +1011,7 @@ void MainWindow::refreshDocument()
 */
 void MainWindow::compile(bool reload, bool forcedone, bool rebuildParameterWidget)
 {
+try{
 	bool shouldcompiletoplevel = false;
 	bool didcompile = false;
 
@@ -1087,6 +1088,9 @@ void MainWindow::compile(bool reload, bool forcedone, bool rebuildParameterWidge
 	}
 
 	compileDone(didcompile | forcedone);
+}catch(int){
+	GuiLocker::unlock();
+}
 }
 
 void MainWindow::waitAfterReload()
--- a/src/mainwin.cc
+++ b/src/mainwin.cc
@@ -1088,7 +1088,8 @@ try{
 	}
 
 	compileDone(didcompile | forcedone);
-}catch(int){
+}catch(HardWarningException){
+	PRINT("Execution aborted");
 	GuiLocker::unlock();
 }
 }
@@ -1150,6 +1151,7 @@ void MainWindow::updateCompileResult()
 
 void MainWindow::compileDone(bool didchange)
 {
+try{
 	const char *callslot;
 	if (didchange) {
 		updateTemporalVariables();
@@ -1163,6 +1165,10 @@ void MainWindow::compileDone(bool didchange)
 
 	this->procevents = false;
 	QMetaObject::invokeMethod(this, callslot);
+}catch(HardWarningException){
+	PRINT("Execution aborted");
+	GuiLocker::unlock();
+}
 }
 
 void MainWindow::compileEnded()
@@ -1239,6 +1245,7 @@ void MainWindow::instantiateRoot()
 */
 void MainWindow::compileCSG()
 {
+try{
 	assert(this->root_node);
 	PRINT("Compiling design (CSG Products generation)...");
 	this->processEvents();
@@ -1288,7 +1295,7 @@ void MainWindow::compileCSG()
 		}
 		else {
 			this->root_products.reset();
-			PRINT("WARNING: CSG normalization resulted in an empty tree");
+			PRINT("<b>WARNING: CSG normalization resulted in an empty tree</b>");
 			this->processEvents();
 		}
 	}
@@ -1326,8 +1333,8 @@ void MainWindow::compileCSG()
 	if (this->root_products &&
 			(this->root_products->size() >
 			 Preferences::inst()->getValue("advanced/openCSGLimit").toUInt())) {
-		PRINTB("WARNING: Normalized tree has %d elements!", this->root_products->size());
-		PRINT("WARNING: OpenCSG rendering has been disabled.");
+		PRINTB("<b>WARNING: Normalized tree has %d elements!</b>", this->root_products->size());
+		PRINT("<b>WARNING: OpenCSG rendering has been disabled.</b>");
 	}
 #ifdef ENABLE_OPENCSG
 	else {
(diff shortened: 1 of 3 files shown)


t-paul:
Nice. As for allowing the script to override, I think that could be a separate discussion. Having a UI (and maybe cmd line?) option is already a good feature.

MichaelPFrey:
bildschirmfoto von 2018-12-12 04-50-47

warn

```
michael@michael-ThinkCentre-M92p:~/openscad$ ./openscad --hardwarnings undef.scad -o test.png
WARNING: Unable to convert translate(undef) parameter to a vec3 or vec2 of numbers, in file undef.scad, line 5
michael@michael-ThinkCentre-M92p:~/openscad$ 
```

MichaelPFrey:
Animation Test:
bildschirmfoto von 2018-12-13 21-51-05
bildschirmfoto von 2018-12-13 21-50-58
Test Code:
```
a=sin($t*360)*2;
b=cos($t*360)*2;
color([a,b,0])
cube();
```

MichaelPFrey:
Customizer test
bildschirmfoto von 2018-12-13 22-05-41


Test Code:
```
a=0; //[-2:0.1:2]
b=0;  //[-2:0.1:2]
c="red"; //["blue","test",a]
rotate(a/b)
cube();

color(c)
sphere();
```

src/mainwin.cc
```diff
@@ -1027,83 +1027,89 @@ void MainWindow::refreshDocument()
...
+				if (!this->root_module) {
+					emit highlightError( parser_error_pos );
+				}
 			}
 		}
-	}
 
-	compileDone(didcompile | forcedone);
+		compileDone(didcompile | forcedone);
+	}catch(HardWarningException){
```
kintel: `const HardWarningException&`

..because we don't want to make a copy of an exception. and we may not want to even know if it's a concrete class.

src/mainwin.cc
```diff
@@ -1163,19 +1169,25 @@ void MainWindow::updateCompileResult()
...
+		}
+		else {
+			callslot = "compileEnded";
+		}
 
-	this->procevents = false;
-	QMetaObject::invokeMethod(this, callslot);
+		this->procevents = false;
+		QMetaObject::invokeMethod(this, callslot);
+	}catch(HardWarningException){
```
kintel: `const HardWarningException&`

src/mainwin.cc
```diff
@@ -1255,113 +1267,119 @@ void MainWindow::instantiateRoot()
...
 																														this->background_products);
-	PRINT("Compile and preview finished.");
-	int s = this->renderingTime.elapsed() / 1000;
-	PRINTB("Total rendering time: %d hours, %d minutes, %d seconds\n", (s / (60*60)) % ((s / 60) % 60) % (s % 60));
-	this->processEvents();
+		PRINT("Compile and preview finished.");
+		int s = this->renderingTime.elapsed() / 1000;
+		PRINTB("Total rendering time: %d hours, %d minutes, %d seconds\n", (s / (60*60)) % ((s / 60) % 60) % (s % 60));
+		this->processEvents();
+	}catch(HardWarningException){
```
kintel: `const HardWarningException&`

src/printutils.cc
```diff
@@ -71,6 +76,9 @@ void PRINT_NOCACHE(const std::string &msg)
 			outputhandler(msg, outputhandler_data);
 		}
 	}
+	if(stop && OpenSCAD::hardwarnings){
+		throw HardWarningException("");
```
kintel: It's probably good style to provide some information in the exception, like the error message?

src/mainwin.cc
```diff
@@ -1027,83 +1027,89 @@ void MainWindow::refreshDocument()
...
 			}
 		}
-	}
 
-	compileDone(didcompile | forcedone);
+		compileDone(didcompile | forcedone);
+	}catch(HardWarningException){
+		exceptionCleanup();
+	}
+	OpenSCAD::hardwarnings = false;
```
kintel: Why are we setting + resetting this flag in various methods? Shouldn't this be a constant only set from Preferences?

src/printutils.cc
```diff
@@ -53,6 +56,7 @@ void PRINT(const std::string &msg)
 
 void PRINT_NOCACHE(const std::string &msg)
 {
+	bool stop(false);
```
kintel: Nitpick: Use assignment `auto stop = false` or uniform initialization `bool stop{false}`.

src/mainwin.cc
MichaelPFrey: The problem is, that we generate various warnings in various places.
Specifically: Any `WARNING:` in mainwin.cc is problematic for the approach I took. (Throwing an exception on the first `WARNING:`)
Not all warnings have something to do with the SCAD File currently processed.
https://github.com/openscad/openscad/blob/15410be568ed79c172e293dc62655b287d5449a1/src/mainwin.cc#L1483
https://github.com/openscad/openscad/blob/15410be568ed79c172e293dc62655b287d5449a1/src/mainwin.cc#L1501
https://github.com/openscad/openscad/blob/15410be568ed79c172e293dc62655b287d5449a1/src/mainwin.cc#L1592
Not wanting to mess with those, I decided to simply only enable hardwarnings, when the code block is guarded by a try() catch. But now that I think about that, this is a bad idea. We do not fully lock down the UI when processing a design, so the user can open a new window. Yeah... there is no way around it. The UI-Warnings need to use a different magic word.

src/printutils.cc
MichaelPFrey: you are right. Under normal circumstances it is redundant, but when the exception slips trough the net, the message is at least outputted on standard output, allowing to search for the hole in the net.

kintel:
I'm still not sure this is the best way of doing this, but it's better than what we had, so we can think about better solutions later.

PR #2633, opened 2018-12-14
$linenumber and $currentfilename "special" variables,  mainfilename() function

MichaelPFrey:
- this partially addresses #380
- this addresses https://github.com/openscad/openscad/issues/380#issuecomment-20239422
- [x] update change log
- [x] update cheatsheet https://github.com/openscad/openscad.github.com/pull/68
- [ ] update cheatsheet https://github.com/openscad/openscad.github.com/pull/68
- [x] what happens on an overwrite attempt?
- [ ] update wikibooks


MichaelPFrey:
`$linenumber="test";` 
results in 
```
ERROR: Parser error in file "/home/michael/openscad/testdata/scad/misc", line 1: syntax error 
ERROR: Compilation failed!
```
that could be improved, but is fine by me for now.

MichaelPFrey:
includetest

nophead:
I see that these are not variables at all, more like macros or special constants, as they get replaced by literals during syntax analysis. Neat as there there are no runtime penalties after that in terms of lookup and function side effects and they can never be changed.

I was thinking $preview should be implemted the same way, but then I realised it would not be possible to set it on the command line, which is necessary when generating .echo and .csg files, so it needs to stay as a variable.

These should probably be called "special constants" when documented to make it clear they cannot be overriden and of course don't have dynamic scope. If they are just tokens then perhaps another prefix as suggested before. Even _linenumber_ or _LINENUMBER_ like C.

t-paul:
Do we even know *why* we implement that? All those feature requests are only listing a huge amount of wishes for variables without spelling out a single use case.

t-paul:
So I think if we start implementing all those query things, we need to establish some ground rules covering all of them. Otherwise that will end up in a mess. Some of the points are already listed in the issue (e.g. windows path '\' or '/').
The "force-read-only" is already a good direction, giving an error in that case is certainly the right thing to do.
Also the comment from @nophead is key that the feature should be identifyable by syntax and not sneek in different behaviour for dynamic scoped variables (that most people are not even aware of). As it's a read-only feature, there's also options to maybe use one or two funcitons to return an object that has the different values as properties (looking at the OpenSCAD2 proposal again).

MichaelPFrey:
To be clear: I mainly want to resolve the discussion and I am happy with any result, meaning anything from a merging this to closing all related issues as "won't fix" and anything in between.

One reason to implement it in the parser are things like this:
```
echo(concat($linenumber,
$linenumber,
$linenumber,

$linenumber

));

```
if linenumber would be a function, I could do fun things like this:
```
echo(
linenumber
(

)
);
```
and have no idea what the result is supposed to be.

We could change the $ to #, making it #LINENUMBER to make it look like a preprocessor directive. (which it kind of is)

One of the use cases for all those variables is debugging. Given that I am working on adding the linenumber and filename to most of the warnings, that use case should become less and less relevant. Custom errors that require filename and line number can be done by using assert().

> e.g. windows path '' or '/'
most modern operating systems use /. Our errors and warning use /. Windows accepts /.
The regression test should be operating system independent. So: I do not see an issue here.

MichaelAtOz:
> I do not mind when we agree to not implement this.

> Question is: To we even want to give the user access to the mainfilename?

If we do not, and also for the other related things, we then NEED to talk about; debugging, library version management and a whole lot of stuff I don't have time to wikipedia the current trendy term for.

That will then spiral down the OpenSCAD2 black hole, and be doomed to repeat this in another 3 years.

In the end [I don't really give a damn](https://www.youtube.com/watch?v=xLnTWxpTQt4). Because you CAN live without it, but that leads to stagnation.

t-paul:
I explained in the past and will repeat every time this comes up. _OpenSCAD2 is not blocking anything, there is no black hole, no stagnation due to that_.

The thing that is limiting progress is developer time and motivation. Pretty much all the language changes that were merged and activated lately are guided by the OpenSCAD2 proposal. I very much believe that while taking some time the discussion made them much better. The only chance we have for continued progress actually is to bring a more consistent behavior into the language. This is a difficult process as we try to not break existing stuff if it can be avoided while still introducing new things. There is #2454 as a great example. This is proposing an aswesome new feature which only needs some tiny changes to almost perfectly fit the OpenSCAD2 proposal (Yes, I actually tried). It's sitting there because nobody is working on it.

A very nice representation of that discussion (for a much larger scale problem, C++) can be heard in the first couple of minutes of CppCon 2018: “[Grill the Committee](https://www.youtube.com/watch?v=cH0nJPbMFAY)” in the opening statements from Titus Winters and Herb Sutter.

MichaelPFrey:
I close this, as I am not really happy with the implementation and I do not see a real use case currently.

I prefer improving the internal error reporting.

If someone has a good idea how to improve the errors and warnings, please open an issue.
If someone has a simple example, where the warning is bad, please open an issue.

If someone has a good use case for linenumber, currentfilename and/or mainfilename, please write and example and open an issue - I am happy to revisit this topic when I see a good use case.

Issue #2634, opened 2018-12-14
IDF (Intermediate Data Format) import


Knochi:
Hi,

I'm an electronics guy with little experience in C++. I would like to be able to import IDF Files. These is a common interchange format form eCAD to mCAD. 
IDF export has to files:
- .emn contains outline of PCB and positions of holes and parts. 
- .emp contains outline and height of parts

I did an import script for PCB in Blender several years ago and remember it wasn't that hard. Just a bunch of polygons, holes and extrusions.

Wikipedia Article has a few nice links https://en.wikipedia.org/wiki/Intermediate_Data_Format

MichaelPFrey:
I am a PCB design engineer my self and I know the IDF Fileformat.

One argument for IDF support is KiCAD.

But I do not think that adding the ability to import IDF is the right approach.
Writing a convert in any script language from IDF to CSG (a subset of the Openscad Language) would be trivial.
Using scripts allows the user to easily hack the script. For example, when you seriously use IDF as exchange format, companies start to map more detailed models that replace the oversimplified the models from the emp files.

Something that I had to deal with professionally, was crap IDF Files from a vendor. It was "easy" to fix, but it involved scripting to look at the various board outlines and manually editing both emn and emp file and it showed me that the designer had no idea how to use his own PCB design software. That is not uncommon. Using an import script instead of an import allows for a very flexible adaptation.

Fun fact: [that is my contribution the IDF wikipedia article](https://en.wikipedia.org/w/index.php?title=Intermediate_Data_Format&type=revision&diff=801038001&oldid=740882075)

t-paul:
Yes, it would be much better adding support for STEP, but that is a *huge* project and STEP is a massive beast even with existing libraries. I think I looked at this before (maybe due to some discussion on IRC, I don't remember the details) and the conclusion was similar.

That said, with the better solution (probably) way out and not much alternative, maybe it's worth a shot. There seems to be a pretty straight forward usecase where this could help. Creating 3D-printed enclosures for PCBs. If this can give a simple way to get the PCB outline transfered into OpenSCAD as placeholder it might be useful for a number of people even with the attached limitations.

kintel:
I didn't look at your code for arcs, but we have DXF arc code in OpenSCAD that you could look at:
https://github.com/openscad/openscad/blob/master/src/dxfdata.cc#L208

Knochi:
Thanks! That helped a lot... implemented the facets function as a openscad function and calculated the fragments..

```
[...]

PR #2637, opened 2018-12-19
Add IDFv3 import.

t-paul:
[IDF](https://en.wikipedia.org/wiki/Intermediate_Data_Format) import - [IDFv3 Specification](https://www.simplifiedsolutionsinc.com/images/idf_v30_spec.pdf)

This is currently importing as 2D and parsing the board outline and drill holes sections. For a screenshot, see [comment](https://github.com/openscad/openscad/issues/2634#issuecomment-448380532) in issue #2634.

IDFv3 is supported as export format in KiCAD so this could be used to export board outlines to OpenSCAD.

Implementation uses a simple Boost Qi parser as IDF is a line based text format.

ToDo:
- [ ] Add support for other sections in parser, most will likely be ignored in the import
- [ ] Find solution for drill holes on the board edge (e.g. mouse bites) not producing useful geometry


MichaelPFrey:
Very interesting approach and an implementation that is surprisingly clean.

A few comments:

When you consider importing the models as out of scope for now (which i totally would understand), then please do not add the emp extension:
`else if (ext == ".emp") actualtype = ImportType::IDF3;`
I know CAD packages that do not properly name the files and the user has to manually rename the files (Mentor board station calls the files _brd and _lib), but  users renaming the files the wrong way around is not our issue.

IDF is also a bit awkward as you need to read both the emp and emn file to get very rough component outlines. I see why that breaks the current mold.

That could be an issue `nl = qi::skip[+qi::char_("\r\n")];` as some CAD system will use just new line when running under linux/unix/sun solaris.

Out of interest: Why isn´t this marked as experimental?

MichaelPFrey:
Awesome. What about https://github.com/KiCad/kicad-source-mirror/tree/master/utils/idftools/idf_examples as sample files?

Issue #1574, opened 2016-01-31
Fails silently on circle();


lukas-buergi:
I wanted to add a 2d circle, so typed:
circle(radius=5);
and then spent at least 15 minutes, probably more, debugging why the hell my structure based on that circle was behaving in a strange way (radius is wrong, the parameter is just called r).
Seriously? I can hardly believe it, am I missing something? Why the hell is there no error? Is there some verbose or warning flag I should add? I couldn't find anything.

So based on the assumptions that it's really supposed to fail silently in that case: _please_ add some warning in cases like this. And I can't think of a use of default values for parameters like this, that there is a default value and so a circle is entered no matter what just makes mistakes like this harder to track.
##

kintel:
This could be generalized into: We should give warning on attempting to instantiate any module with module parameters which are not declared or used by the module in question.
The implementation for built-in modules are likely a bit different than for user-defined modules so that's probably two separate tasks.

Such warnings would be a lot more effective if displayed in context. I'll open a separate issue on context-sensitive warnings and errors as that requires parser changes.

doug-moen:
circle(radius=5) doesn't work. But OpenSCAD doesn't report errors, it
silently ignores them. Good luck if you need to debug something. Posting
your code to the forum is the workaround.

Try this: circle(r=5)

MichaelAtOz:
> This could be generalized into: We should give warning on attempting to instantiate any module with module parameters which are not declared or used by the module in question.

Noting that such use with $variables is valid to affect sub modules for example. May want that warning as a preference option, on by default for beginners.

Perhaps only warn when a default is encountered in addition to unreferenced parameter? Fits this use case. Or review the use of defaults, I can't see much use for them, and most code is unlikely to use them, I only do for debugging.

doug-moen:
"Perhaps only warn when a default is encountered in addition to
unreferenced parameter?"

This would not have addressed the problem when jpmendes tried to use the
angle= parameter of rotate_extrude, as documented in the wiki, and it was
silently ignored. (He wasn't using the unreleased DEV version of OpenSCAD.)
It would have been better to issue an error message in this case.

See:
http://forum.openscad.org/This-seems-to-be-ultra-basic-I-cannot-have-the-rotate-extrude-angle-parameter-working-tp15180.html

TakeItAndRun:
Maybe this would be something for the google summer of code:
Generate consistent error/warning messages for all (internal) calls.
So there are no more silent fails, at all.

[...]

PR #2638, opened 2018-12-21
check for unknown parameters

MichaelPFrey:
- [x] add line number and filename to warning
   - [x] Make switch to deactivate warning
      - [think about default?
       - [x] UI-Test
          - [x] toggle the option while running
          - [x] preferences gets saved 
       - [x] Console-Test
   - [x]  cross reference relevant issues #1574
   - [x]  finding issues in our tests
      - https://github.com/openscad/openscad/blob/40797a1431c0aa6aa1111f3d868f54500d193ff1/testdata/scad/misc/localfiles_dir/localfiles_module.scad#L3
      - https://github.com/openscad/openscad/blob/40797a1431c0aa6aa1111f3d868f54500d193ff1/testdata/scad/misc/variable-scope-tests.scad#L77-L81
      - https://github.com/openscad/MCAD/pull/46
      - https://github.com/openscad/openscad/tree/master/libraries
   - [x]  update change log
   - handling of special variables? -> just ignore them for now
   - warn when Too many positional arguments are passed
   - [x] add intentional test cases


MichaelPFrey:
#2638 
bildschirmfoto von 2018-12-21 14-04-54

```
circle(radius=5);
cube(1,2,3,4,5,6);
```

bildschirmfoto von 2018-12-21 15-07-58

hhmmm... warning once would be fine, so:

bildschirmfoto von 2018-12-21 15-21-52

MichaelAtOz:
Hmm, not sure about this.
Using unspecified parameters is used in the wild.
One example is write.scad which uses them extensively.
write.scad may be used less now that text() is here, but it would have screens full of warnings.

MichaelPFrey:
Thank you for the hint - having specific cases helps a lot to figure out the right way to go.

I plan on adding an option to disable this feature. (command line and GUI)
It is also possible to know, if it is an inbuilt module or a user module.
(The way it is implemented, it treads overwritten modules as user modules, so even that is accounted for).
So we can always keep the warnings on for builtins, even when the user disables it for user modules.

I assume that this is the official Write.scad:

https://www.thingiverse.com/thing:16193

Looking at write.scad: Okay...
```
//default settings
	center=false;
	h = 4;			 //mm letter height
```
use is a separate scope from the mainfile and the module write uses the h  from the main scope.
But it is also trival to fix, `module write(word){` becomes `module write(word,h=h,t=t,center=center){`. Yeah, I see why we can not break the legacy behavior, but at the same time: It is trivial to fix and  in my opinion, the module signature becomes a better documentation of the module by declaring all parameters.

Also note that this very mechanism allows to overwrite pi...

There is a reason for this warnings. Currently a typo in a parameter name does not get any feedback to the user, the user can overwrite values he better should not etc.

I think this feature is the right decision on the long run, but the transition will be tricky. Options are not always a good solution - you might want to have those warnings for your own code, but might need to turn it off for legacy projects, resulting in forgetting to turn it back on.

MichaelPFrey:
Funny side finding:
`import(undef,undef,1,[],[],"test.dxf","test");`

results in
```
DEPRECATED: filename= is deprecated. Please use file=
DEPRECATED: layername= is deprecated. Please use layer=
WARNING: Can't open DXF file '/home/test.dxf'.
Geometries in cache: 3
```
As filename and layername are appended to the list of arguments

Anyway, this will now become:
```
WARNING: Too many unnamed arguments supplied, in file openscad, line 1
ERROR: Unsupported file format while trying to import file '""'
Geometries in cache: 1
```

MichaelPFrey:
Command line test:
```
a = "Hallo Welt";
module test(){
    echo(a);
}

test();
test(a="Hello World");
```

bildschirmfoto von 2018-12-22 20-36-01
```
$ ./openscad HelloScope.scad --parametercheckoff -o test.png
ECHO: "Hallo Welt"
ECHO: "Hello World"
$ ./openscad HelloScope.scad -o test.png
ECHO: "Hallo Welt"
WARNING: variable a not specified as parameter, in file HelloScope.scad, line 7
ECHO: "Hello World"
```

src/Preferences.cc
```diff
@@ -160,6 +160,7 @@ void Preferences::init() {
 	this->defaultmap["advanced/localization"] = true;
 	this->defaultmap["advanced/autoReloadRaise"] = false;
 	this->defaultmap["advanced/enableSoundNotification"] = true;
+	this->defaultmap["advanced/enableParameterCheck"] = true;
```
kintel: One day we'll have to split up the "Advanced" tab, but that's a future adventure..

src/evalcontext.cc
```diff
@@ -9,8 +9,8 @@
 #include "exceptions.h"
 
 EvalContext::EvalContext(const Context *parent, 
-												 const AssignmentList &args, const class LocalScope *const scope)
-	: Context(parent), eval_arguments(args), scope(scope)
+												 const AssignmentList &args, const Location &loc, const class LocalScope *const scope)
+	: Context(parent), eval_arguments(args), loc(loc), scope(scope)
```
kintel: 👍

src/Preferences.cc
MichaelPFrey: good idea. At least grouping would be a good first step.
I will come back to this, likely when I add even more options related to errors and warnings.

src/evalcontext.cc
```diff
@@ -35,19 +35,36 @@ ValuePtr EvalContext::getArgValue(size_t i, const Context *ctx) const
...
 {
   AssignmentMap resolvedArgs;
   size_t posarg = 0;
+  bool tooManyWarned=false;
   // Iterate over positional args
   for (size_t i=0; i<this->numArgs(); i++) {
     const auto &name = this->getArgName(i); // name is optional
     const auto expr = this->getArgs()[i].expr.get();
     if (!name.empty()) {
+      if(name.at(0)!='$' && (OpenSCAD::parameterCheck || (not usermodule))){
```
kintel: It feels to me that the `OpenSCAD::parameterCheck` and `usermodule` check should be external to this function and it should rather be passed an single bool, e.g. `silent`. Does that make sense to you?
I believe we have another `silent` flag like that somewhere else.

src/render.cc
```diff
@@ -47,9 +47,10 @@ AbstractNode *RenderModule::instantiate(const Context *ctx, const ModuleInstanti
 	auto node = new RenderNode(inst);
 
 	AssignmentList args{Assignment("convexity")};
+	AssignmentList optargs;
 
 	Context c(ctx);
-	c.setVariables(args, evalctx);
+	c.setVariables(evalctx, args, optargs);
```
kintel: In other places, you're just relying on the default value of optargs. Why not here?

src/openscad.cc
```diff
@@ -920,6 +920,7 @@ int main(int argc, char **argv)
 		("d,d", po::value<string>(), "deps_file -generate a dependency file for make")
 		("m,m", po::value<string>(), "make_cmd -runs make_cmd file if file is missing")
 		("quiet,q", "quiet mode (don't print anything *except* errors)")
+		("parametercheckoff", "disables the parameter check")
```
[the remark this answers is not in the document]
MichaelPFrey: good idea. we could also add an option to disable all optional warnings.

Yes this would be the first optional warning and I assume that we will add more for legacy support.

`-Wundeclared-parameter` is awkward because of the capital letter and that we normally use `-` for the short parameter name and `--` for the long name.

I will dwell a bit on that.

src/evalcontext.cc
MichaelPFrey: we have a `silent` flag in the variable lookup.

Adding a silent flag to resolveArguments is fine by me, must of the time it is called indirectly by setVariables.

src/render.cc
MichaelPFrey: This is just an artifact from debugging/cleanup.

src/openscad.cc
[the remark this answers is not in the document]
kintel: I borrowed the syntax from gcc, but we could come up with whatever fits our current style.

src/evalcontext.cc
kintel: The idea would be to move any global state checks as far up in the call hierarchy as we can, so that lower-level functions are less dependent on global state and thus easier to test or refactor later.

src/evalcontext.cc
```diff
@@ -35,19 +35,36 @@ ValuePtr EvalContext::getArgValue(size_t i, const Context *ctx) const
...
+          PRINTB("WARNING: variable %s not specified as parameter, %s", name % this->loc.toRelativeString(this->documentPath()));
+        }
+      }
       resolvedArgs[name] = expr;
     }
     // If positional, find name of arg with this position
     else if (posarg < args.size()) resolvedArgs[args[posarg++].name] = expr;
+    else if (!tooManyWarned){
+      PRINTB("WARNING: Too many unnamed arguments supplied, %s", this->loc.toRelativeString(this->documentPath()));
+      tooManyWarned=true;
```
MichaelPFrey: hmm... as the bool is now call silent, this should also be conditional.
(despite the fact that supplying too many positional arguments is a separate issue from using unspecified parameters)

src/primitives.cc
```diff
@@ -223,6 +227,12 @@ AbstractNode *PrimitiveModule::instantiate(const Context *ctx, const ModuleInsta
 		const auto r = lookup_radius(c, inst->location(), "d", "r");
 		const auto r1 = lookup_radius(c, inst->location(), "d1", "r1");
 		const auto r2 = lookup_radius(c, inst->location(), "d2", "r2");
+		if(r.type() == Value::ValueType::NUMBER && 
+			(r1.type() == Value::ValueType::NUMBER || r2.type() == Value::ValueType::NUMBER)
+			){
+				PRINTB("WARNING: Cylinder parameters ambiguous, %s", inst->location().toRelativeString(ctx->documentPath()));
```
MichaelPFrey: needs a testscase

src/openscad.cc
```diff
@@ -845,6 +845,15 @@ std::string join(const Seq &seq, const std::string &sep, const ToString &toStrin
     return boost::algorithm::join(boost::adaptors::transform(seq, toString), sep);
 }
 
+void flagConvert(bool &flag, std::string str){
```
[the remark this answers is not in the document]
MichaelPFrey: I assumed that silently ignoring typos would be a good approach, but saying that out-loud, I see the issue with that approach.

An exceptions seams a bit excessive to me, unless we plan on introducing multiple flags, which saying it out-loud, we do.

MichaelPFrey:
```
a="test";
module test(){
    echo(a);
}

test();
test(a="overwrite");
```

bildschirmfoto von 2018-12-30 14-06-29

```
./openscad --check-parameters on -o 1.png para.scad
ECHO: "test"
WARNING: variable a not specified as parameter, in file para.scad, line 7
ECHO: "overwrite"

./openscad --check-parameters off -o 1.png para.scad
ECHO: "test"
ECHO: "overwrite"

./openscad --check-parameters 1 -o 1.png para.scad
ECHO: "test"
WARNING: variable a not specified as parameter, in file para.scad, line 7
ECHO: "overwrite"

./openscad --check-parameters 0 -o 1.png para.scad
ECHO: "test"
ECHO: "overwrite"

./openscad --check-parameters true -o 1.png para.scad
ECHO: "test"
WARNING: variable a not specified as parameter, in file para.scad, line 7
ECHO: "overwrite"

./openscad --check-parameters false -o 1.png para.scad
ECHO: "test"
ECHO: "overwrite"

./openscad --check-parameters test -o 1.png para.scad
Could not parse '--check-parameters test' as flag
ECHO: "test"
WARNING: variable a not specified as parameter, in file para.scad, line 7
ECHO: "overwrite"
```

src/openscad.cc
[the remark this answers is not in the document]
kintel: Actually, I think you can write this as a handler for boost program_options, and let boost deal with the exception. That's pretty easy, but you'll likely spend more time reading docs than writing code, so consider that optional :)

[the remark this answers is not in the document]
MichaelPFrey: I would prefer getting this merged, when you consider it acceptable in the mean time.

Getting the location into the eval context makes it easier to add the line number and filename to the remaining messages. (Looking back, I should have done that earlier and in a separate PR, but oh well - in hindsight every thing becomes much clearer)

The whole business of the flag to disable it is also kind of an after thought. I hope most people will leave the option on, but at the same time I know that I will add alot more warnings that will help new users, but might annoy current power users. So adding options to disable warnings for legacy design is a good compromise.

Issue #2640, opened 2018-12-22
Crash with --debug & output redirection


MichaelAtOz:
`"C:\Program Files\openscad-2018.12.08.ci706\openscad" --debug=all cmd-line.scad -o cmd-line.echo > cmd-line.txt`

Regardless of .scad code, `-debug=all` without redirection, or redirection without `debug=all` works, but both crashes.

MichaelAtOz:
Windows 7, multiple OpenSCAD versions, 2018.12.08 is the latest I have at hand.

MichaelPFrey:
Ubuntu 18.10:
```
./openscad --debug=all cmd-line.scad -o cmd-line.echo > cmd-line.txt
Debug on. --debug=all
Camera: Camera()
openscad: BuiltinContext:
Context: 0x7ffdc26f38b0 (0)  document path:   vars:    PI = 0x55a868229ce0    $vpr = 0x55a86820e610    $fs = 0x55a86820d860    $fn = 0x55a86820da90    $vpd = 0x55a86820e700    $fa = 0x55a86820d7f0    $t = 0x55a86820dec0    $preview = 0x55a868212b20    $vpt = 0x55a86820e520
```

MichaelPFrey:
https://github.com/openscad/openscad/blob/f2848902e8264f2fd068694741de6aaaec33f452/src/openscad.cc#L352-L354
https://github.com/openscad/openscad/blob/f2848902e8264f2fd068694741de6aaaec33f452/src/context.cc#L223-L256
`v.second` has to become `v.second->toEchoString();` which is better:
 ```
./openscad --debug=all cmd-line.scad -o cmd-line.echo > cmd-line.txt
Debug on. --debug=all
Camera: Camera()
openscad: BuiltinContext:
Context: 0x7ffdf53055c0 (0)  document path:   vars:    PI = 3.14159    $vpr = [0, 0, 0]    $fs = 2    $fn = 0    $vpd = 500    $fa = 12    $t = 0    $preview = false    $vpt = [0, 0, 0]
```
now we add a few new lines:
```
./openscad --debug=all cmd-line.scad -o cmd-line.echo > cmd-line.txt
Debug on. --debug=all
Camera: Camera()
openscad: BuiltinContext:
Context: 0x7ffc9a68bad0 (0)
  document path: 
  vars:    PI = 3.14159
    $vpr = [0, 0, 0]
    $fs = 2
    $fn = 0
    $vpd = 500
    $fa = 12
[...]

PR #2641, opened 2018-12-22
Context::dump debug

MichaelPFrey:
this is based on #2640

I do not understand what

` s << boost::format("Context: %p (%p)") % this % this->parent; `

is supposed to do. Currently, it does something like

`Context: 0x7ffc9a68bad0 (0)`

as the parent of the top_ctx is off course a `nullptr`, which is kind of messy.

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -225,31 +225,31 @@ std::string Context::dump(const AbstractModule *mod, const ModuleInstantiation *
 {
 	std::stringstream s;
 	if (inst) {
-		s << boost::format("ModuleContext %p (%p) for %s inst (%p)") % this % this->parent % inst->name() % inst;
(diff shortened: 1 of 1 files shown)


MichaelPFrey:
Okay - now I start to understand.

`s << boost::format("Context: %p (%p)\n") % this % this->parent;`

uses %p as in "hexadecimal output". I am not certain if %p on a pointer is allowed, but it seams to work.

Mean while:

`s << boost::format("    %s = %s") % v.first % v.second;`

v.first is a string, but v.second is not. So, either v.second has to be converted to a string (as suggested in this PR) or it should be %p not %s.

Could that be a right track? I do not have a windows machine to test my theory and given that we output addresses, I can not write a regression test.

kintel:
yup, `%p` is taken from `printf`'s pointer mode.
The second `%s` was probably a bug it should indeed be converted to a string

PR #2642, opened 2018-12-22
3D Printing Integration

printathing:
This pull requests adds 3d Printing capabilities to the OpenSCAD workflow.  I have tested it on an Ubuntu 18.04 machine with one set of compiler flags, successfully.  It should definitely be tested on other OS's and with different sets of compiler flags as well to get any bugs out.  

Required for merging as experimental:
* [x] Review


src/mainwin.cc
```diff
@@ -2013,6 +2017,164 @@ void MainWindow::csgRender()
...
+	{
+		PRINT("Unable to open exported stl file.");
+		return 0;
+	}
+	
+	//Convert it to base 64:
+	QString stlFileB64=file.readAll().toBase64();
+	
+	//Base 64-encoded file contents:
+	jsonInput.insert("file", stlFileB64.toLocal8Bit().constData());
```
t-paul: This seems to confuse older Qt (on Travis / Trusty). Base64 encoded string should be safe without that `.toLocal8Bit().constData()`.

t-paul:
Windows build looks good, download (CircleCI): [32bit](https://circleci.com/gh/openscad/openscad/765#artifacts) / [64bit](https://circleci.com/gh/openscad/openscad/764#artifacts)

src/mainwin.cc
```diff
@@ -2013,6 +2017,164 @@ void MainWindow::csgRender()
...
+	
+	char statusStr[250];
+	
+	if (statusCodeV.toInt()!=200)
+	{
+		setCurrentOutput();
+		
+		sprintf(statusStr, "API status code: %d", statusCodeV.toInt());
+		
+		PRINT(statusStr);
```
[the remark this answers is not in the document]
printathing: Thanks for letting me know about that!  That's much simpler.  I'll integrate PRINTB.

src/mainwin.cc
```diff
@@ -2241,21 +2403,14 @@ void MainWindow::actionCheckValidity()
...
-	if (GuiLocker::isLocked()) return;
-	GuiLocker lock;
-#ifdef ENABLE_CGAL
-	setCurrentOutput();
-
 	if (!this->root_geom) {
 		PRINT("ERROR: Nothing to export! Try rendering first (press F6).");
 		clearCurrentOutput();
-		return;
+		return 0;
```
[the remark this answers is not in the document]
printathing: Ok.  Converted to true/false.

src/mainwin.cc
printathing: Good catch!  :)  Removed.

src/mainwin.cc
```diff
@@ -2013,6 +2017,164 @@ void MainWindow::csgRender()
...
+
+	//Make sure we can export:
+	if (! canExport(dim))
+	{
+		PRINT("Cannot 3D Print due to errors.");
+		return;
+	}
+	
+	//Ccreate a temporary file name valid on all systems:
+	QString exportFilename=QString(std::tmpnam(tempStlFileName));
```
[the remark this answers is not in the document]
printathing: Ok.  I tried your suggestion, but because QTemporary file can only give a file name after the file is open, it ended up creating a lot more code than what I had before.  Here's the diff:

```
index 4fcb4db9..a8775eea 100755
--- a/src/mainwin.cc
+++ b/src/mainwin.cc
@@ -2042,10 +2042,17 @@ void MainWindow::action3DPrint()
  {
    PRINT("Cannot 3D Print due to errors.");
    return;
- }
+     }
  
- //Ccreate a temporary file name valid on all systems:
- QString exportFilename=QString(std::tmpnam(tempStlFileName));
+ //Create a temporary file name valid on all systems:
+ QTemporaryFile exportFile;
+ //Open the file so we can get the name:
+ if ( ! exportFile.open())
+ {
+   PRINT("Could not open temporary file.");
+   return;
+ }
+ QString exportFilename=exportFile.fileName();
  
  //Render the stl to a temporary file:
  exportFileByName(this->root_geom, FileFormat::STL,
@@ -2167,9 +2174,6 @@ bool MainWindow::uploadStlAndGetPartUrl(const QString & exportFilename, const QS
  //Put it in our output partUrl:
  partUrl.setUrl(partUrlStr);
  
-    //Remove the temporary file:
-    remove(exportFilename.toLocal8Bit().constData());
-    
  return true;
 }

```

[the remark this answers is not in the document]
printathing: So, I'm not entirely sure that in this specific case, QTemporaryFile is an improvement.  Thoughts?

t-paul:
@kintel, do we need to make this optional as it will not work with Qt4? All the offical builds are ok, but it would break FreeBSD currently.

printathing:
@t-paul What won't work with Qt4?  None of this is super fancy.  Imagine there's a way to code it to work with both qt versions?

src/mainwin.cc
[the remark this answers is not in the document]
t-paul: It will also remove the buffer `tempStlFileName` and I think the automatic resource handling is useful. In addition it's not a fair comparison as `std::tmpnam()` *may* return NULL even when given a buffer, so that (probably unlikely) state is not checked.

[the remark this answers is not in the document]
printathing: Good point, yeah the automatic resource handling is better than a manual remove().  Agreed.   Thanks!

t-paul:
No, I don't think we really want to actively support Qt4. It's pretty much the exception case and not relevant for the officially supported platforms. My personal view would be:
- make it usable on Qt4 (e.g. for FreeBSD), accepting maybe some missing features
- drop Qt4 support officially after  we have a new release (hopefully soon-ish)

We did the same for the DBus interface support. It will just be disabled when building against Qt4.

t-paul:
Oh, about *what won't work*... The JSON stuff ([QJsonDocument](http://doc.qt.io/qt-5/qjsondocument.html) / [QJsonObject](http://doc.qt.io/qt-5/qjsonobject.html)) was added in Qt5.0. We certainly don't want to add external dependencies only for Qt4 support.

src/MainWindow.ui
```diff
@@ -868,6 +871,18 @@
...
+  <action name="designAction3DPrint">
+   <property name="icon">
+    <iconset resource="../openscad.qrc">
+     <normaloff>:/images/3dprint-32.png</normaloff>:/images/3dprint-32.png</iconset>
+   </property>
+   <property name="text">
+    <string>&amp;3D Print</string>
+   </property>
+   <property name="shortcut">
+    <string>F7</string>
```
kintel: Let's not make a shortcut for this. It's not something you'll every do very often or casually, and it's too easy to mis-click when indenting to hit F6.

src/mainwin.cc
```diff
@@ -2241,21 +2405,14 @@ void MainWindow::actionCheckValidity()
 #endif /* ENABLE_CGAL */
 }
 
-#ifdef ENABLE_CGAL
-void MainWindow::actionExport(FileFormat format, const char *type_name, const char *suffix, unsigned int dim)
-#else
-	void MainWindow::actionExport(FileFormat, QString, QString, unsigned int)
-#endif
+//Returns if we can export (true) or not(false) (bool)
+//Separated into it's own function for re-use.f
```
kintel: Nitpick: Stray `f`

src/mainwin.cc
```diff
@@ -2013,6 +2017,166 @@ void MainWindow::csgRender()
...
+	
+	setCurrentOutput();
+	PRINT("3D Printing...");
+	
+	unsigned int dim = 3;
+	
+	QUrl partUrl;
+	
+	//Where we hold our temporary stl file name:
+	char tempStlFileName[L_tmpnam];
```
kintel: Not in use?

src/mainwin.cc
```diff
@@ -2013,6 +2017,166 @@ void MainWindow::csgRender()
...
+	
+	//Upload the file to the 3D Printing server and get the corresponding url to see it.
+	//The result is put in partUrl.
+	bool uploadWorked=uploadStlAndGetPartUrl(exportFilename, userFacingName, partUrl);
+	setCurrentOutput();
+
+	//Check the result:
+	if (! uploadWorked)
+	{
+		PRINT("An error occured while contacting the 3D Print API.");
```
kintel: Could we provide more information about what kind of error? e.g. HTTP status code + message?
A cleaner way of doing that could be to just throw an exception with this info instead of passing writable arguments.

src/mainwin.cc
```diff
@@ -2013,6 +2017,166 @@ void MainWindow::csgRender()
...
+		return;
+	}
+	
+	//PRINT("partUrl:");
+	//PRINT(partUrl);
```
kintel: Stray comments

src/mainwin.cc
```diff
@@ -2013,6 +2017,166 @@ void MainWindow::csgRender()
...
+//    exportFilename  - The path to the temporary file that has the stl export in it.
+//    userFacingName  - Then name we should give the file when it is uploaded for the order process.
+//Outputs:
+//    partUrl         - The resulting url to go to next to continue the order process.
+bool MainWindow::uploadStlAndGetPartUrl(const QString & exportFilename, const QString &userFacingName, QUrl &partUrl)
+{
+	setCurrentOutput();
+	
+	//Create a request:
+	QNetworkRequest request(QUrl("https://print.openscad.org/api/v1/part-upload/"));
```
kintel: Do we need a trailing slash here? I feel that's a bit of an uncommon style for such an endpoint

src/mainwin.cc
```diff
@@ -2013,6 +2017,166 @@ void MainWindow::csgRender()
...
+	
+	//Convert it to base 64:
+	QString stlFileB64=file.readAll().toBase64();
+	
+	//Base 64-encoded file contents:
+	jsonInput.insert("file", stlFileB64);
+	
+	////Pring the input json:
+	//PRINT("Sending this JSON:");
+	//PRINT(QString(QJsonDocument(jsonInput).toJson()).toLocal8Bit().constData());
```
kintel: Stray commented out code

src/mainwin.cc
```diff
@@ -2013,6 +2017,166 @@ void MainWindow::csgRender()
...
+	
+	//Create a network access manager:
+	QNetworkAccessManager nam;
+	
+	//Send the post:
+	QNetworkReply *reply = nam.post(request, QJsonDocument(jsonInput).toJson());
+
+	//Wait for the reply:
+	while(!reply->isFinished())
```
kintel: Not in love with this. We should at least add a way to cancel it if the reply doesn't finish in reasonable time.

Cancellation could be offered through the progress bar (but that may be a bit confusing) or a dialog box which could show all the info that currently gets dumped to the console.

src/mainwin.cc
```diff
@@ -2013,6 +2017,166 @@ void MainWindow::csgRender()
...
+	{
+		qApp->processEvents();
+	}
+	
+	QVariant statusCodeV = reply->attribute(QNetworkRequest::HttpStatusCodeAttribute);  
+	
+	//Clean up the reply object:
+	reply->deleteLater();
+	
+	char statusStr[250];
```
kintel: Unused?

src/mainwin.cc
```diff
@@ -2013,6 +2017,166 @@ void MainWindow::csgRender()
...
+	
+	if (statusCodeV.toInt()!=200)
```
kintel: Shouldn't this endpoint return a `201` response as you're creating a new resource?

src/mainwin.cc
```diff
@@ -2013,6 +2017,166 @@ void MainWindow::csgRender()
...
+	{
+		setCurrentOutput();
+		
+		PRINTB("API status code: %d", statusCodeV.toInt());
```
kintel: I feel it's a bit much console debug output going on. Such info isn't really interesting to end users unless something goes wrong when it may be needed for diagnostics

src/mainwin.cc
```diff
@@ -2013,6 +2017,166 @@ void MainWindow::csgRender()
...
+	}
+	
+	//Interpret the response as a json document:
+	QJsonDocument jsonOutDoc= QJsonDocument::fromJson(reply->readAll());
+	
+	//Get the corresponding json object:
+	QJsonObject jsonOutput = jsonOutDoc.object();
+	
+	////Print the whole document:
+	//PRINT(QString(jsonOutDoc.toJson()).toLocal8Bit().constData());
```
kintel: Stray commented out code

src/mainwin.cc
```diff
@@ -2013,6 +2017,166 @@ void MainWindow::csgRender()
...
+	
+	//Extract the cartUrl:
+	QString partUrlStr=jsonOutput.value("data").toObject().value("cartUrl").toString();
```
kintel: What happens if the returned object is malformed, for any reason, e.g. server bug?

src/mainwin.cc
printathing: Good catch, thank you! :)

src/mainwin.cc
printathing: Yes, thank you.  That's left-over from a change I did for t-paul.  removed.

src/mainwin.cc
printathing: Hi Marius.  I cleaned up most of the print statements I had in the code before I submitted the pull request.  The ones remaining I thought would be useful if I'm ever in here debugging some issue we have.  They print key things like the json text we're sending, the json text we get back, and the url the function returns. All of that is important at a very top level for debugging.  Can we keep it in the codebase for future use so I don't end up having to write it again when I'm debugging something?

src/mainwin.cc
printathing: LOL.  Yes, also left over from a change I did for t-paul.  Removed now.  Thanks!

src/mainwin.cc
printathing: I'll provide that suggestion to Matt, but that would be a breaking change to v1 of the interface and I'm anticipating he would say it's not quite to the threshold to roll an entire new API version for that.  The API currently returns 200 on success.

src/mainwin.cc
printathing: Ok.  Are you aware that this status code only shows if there's an error?  So 99% of the time they won't see it.  They'll only see it if something goes wrong and I figure it'll be nice to have the status code when they write me for customer support.

src/mainwin.cc
printathing: Yes, please see above for why I value keeping it in there.

src/mainwin.cc
printathing: Then, I assume the user gets an error when we try to go to a url that isn't valid.  What would you like to happen?

src/mainwin.cc
printathing: Oh.  I do provide information about the status code - that's in the function itself.  I agree this could be done in a more explicit way.  I'll look into how to do that.

printathing:
One thing about the while loop.  I believe it still handles GUI events in that loop, so it's not the same as just a sleep or anything.  Does that make it more comfortable for you?

src/mainwin.cc
printathing: Hmmm...  Yeah, I agree that we should have a better outcome if the API never responds.  I don't like the idea of poping up a dialog box or something that makes this feel like it's some bolted-on piece instead of part of the actual tool itself.  How about separating the code into lines that run before the api returns and lines that runs after the api returns, and hooking up some sort of callback to the finished signal?  http://doc.qt.io/qt-5/qnetworkaccessmanager.html#finished  I am not very adept with QT signals, but I assume they enable callbacks.  With that in place, at least, the rest of the software should still continue to work just as well while waiting for the reply.  If it never comes, no big deal.  If it comes 10 minutes later then the browser will pop up 10 minutes later, and they'll eventually figure out what happened.  Thoughts?  I'm up for editing things to make this better, I just want to agree on the direction before I dive in.

printathing: One challenge with the progress bar is I don't think we have access to a % completion at this level.

src/mainwin.cc
printathing: The documentation has a trailing slash:  https://printathing.docs.apiary.io/#reference/parts/part-uploads/upload-a-part

I'll ask Matt if that is correct.

src/mainwin.cc
kintel: For personal debug code, I usually just leave them in my private branch. Otherwise, the entire codebase would be littered with commented out debug code.
If the debug output is generally useful, you can use the PRINTDB macro, which will allow it to be compiled into the app and enabled at runtime.

src/mainwin.cc
kintel: I'm asking since an independent server implementation would likely want to use a more conventional approach.

src/mainwin.cc
kintel: A background task could work as well. Just make sure that all the allocated memory gets dealt with. ..and make sure that nobody tries to access that temporary file after the QTemporaryFile object falls of the stack (as the file gets deleted).

I agree that a percent-wise progress bar doesn't make much sense. That's usually an option for existing progress bars..

src/mainwin.cc
MichaelPFrey: `#ifdef DEBUG` or `#if 0 // debug` could be options for such debug code.
Both can be found in our code base.

t-paul: Yes, for generally useful debug code, the best option is `PRINTD()` and `PRINTDB()`:

https://github.com/openscad/openscad/blob/44643571800e79e1c031a807e64052508e6e1b1f/src/printutils.h#L40-L52

kintel:
About Qt4: I guess we need to either: a) Kill Qt4 support across the boards or b) disable this feature when built against Qt4. a) is probably the right thing to do, but probably out of scope for this PR, so I guess I'll suggest b) for now. I'll start a TODO list of items needing addressing before we can merge.

src/MainWindow.ui
MichaelPFrey: For some people (with complicated geometry) render is actually faster then preview. Sometimes you need to render the design to see how certain preview artifacts are resolved.

Accidental uploads could be a privacy/confidentiality concern to some people.
You could open a can of worms regarding copyright.

Calling it "3D Print" is also a bit misleading, when it send the data to the internet (instead of to the 3D printer software/slicer).

One side note: The inputdriver automatically creates the action list. We could expand our input drivers to make it possible to map free/reserverd keyboard buttons to actions.

t-paul: I missed the shortcut when looking though the pull request earlier, here's my thinking for completeness :-).....

I'd be happy with F7 provided the following possible future development
- We add a send to print service preferences pane which has some means to disable the feature.
- We add a local OctoPrint, realistically there's at least a similar number of people with their own printer who may want to send to their own OctoPrint and not to an external service.

>  If they do end up accidentally press the button that could be a good way for them to discover that this new functionality even exists and plant a little seed to maybe try it later. It shouldn't take more than 5 seconds of their time to close a browser window.

That gave me a pause actually. I think that is not a good way of discovering this type of features. Considering also that this is uploading user data to an external service, we may need to have some "first start" info (not just talking GDPR compliance but also considering IP issues).

MichaelAtOz: > Usually I don't render unless I'm about to print something. I use the preview for everything up until that point because it's an order of magnitude faster than rendering. 

I think that is unusual. Often you will want to render/export and examine the STL before you print, a few times.

> I don't think I've "accidentally" pressed a function key even once in the last year. The function row is way at the top 

You have one of them old style keyboards, not very hip...

> options re-implementing the natural F5 F6 F7 combo

Actually for me preview, render, export would be natural, maybe F8 for Print?

I have my print machine isolated, so like someone* else mentioned, Fn-key to export to a default location with my USB stick is good. * in nopheads filename/directory stuff.

This just shows how everyone has their UI/UX quirks. 
So until Michael's idea of mapping keys, perhaps leave out the shortcut ATM.  

-----

I would absolutely NOT want a browser to open if I accidentally hit a key. (REALLY HATE those S/W updates that think opening a browser so they can show me something - and meanwhile track me - is.....) anyway
I agree with tp "preferences pane which has some means to disable the feature." except "...means to **enable** the feature", ie I don't think the default setup should point to network externalities, that is something the user should enable with intent. By all means make it easy to setup, for example a pre-populated drop-down or such.

> We add a local OctoPrint

I have yet to play with Octoprint, but if it is vanilla enough, or has its config in standard locations this would be good to add to that drop-downn list.

src/mainwin.cc
MichaelAtOz: I presume the API will document which [Base64](https://en.wikipedia.org/wiki/Base64#Implementations_and_history)? QT happens to use [RFC 2045](http://doc.qt.io/archives/qt-4.8/qbytearray.html#toBase64).

src/mainwin.cc
printathing: Ok, how about if I accept 200 or 201?  Seems like there are some other 20x codes I don't fully understand yet so I'm more comfortable with keeping it more localized.

src/mainwin.cc
printathing: Ok.  Sounds good.  I'll do that. :)

t-paul:
I'm testing the network part. Observations so far:
- QtNetwork has no built-in timeouts at all, this would need to be implemented separately
- Progress bar is simple via callbacks of QtNetworkReply
- No safeguard against multiple button presses, this causes multiple printing uploads
- `$fn = 600; sphere(10);` produces a 68Mb temp file which generates a QJson error `QJson: Document too large to store in data structure` - after this message, the file content is just missing

src/mainwin.cc
printathing: Hm..... Let me see....

If "jsonOutput.value("data")" is not an object, .toObject returns an empty object:

http://doc.qt.io/qt-5/qjsonvalue.html#toObject-1

> QJsonObject QJsonValue::toObject() const
> This is an overloaded function.
> 
> Converts the value to an object and returns it.
> 
> If type() is not Object, the QJsonObject() will be returned.

Then the .value("cartUrl") will try be applied to the empty QJsonObject:

http://doc.qt.io/qt-5/qjsonobject.html#value

> QJsonValue QJsonObject::value(const QString &key) const
> Returns a QJsonValue representing the value for the key key.
> 
> The returned QJsonValue is QJsonValue::Undefined if the key does not exist.
> 
> See also QJsonValue and QJsonValue::isUndefined().

I tested this out a bit and in that case it seemed to hang, so I added another exception to cover this case.  Thanks!
`    //Start trying to extract the cartUrl:
    auto cartUrlValue=jsonOutput.value("data").toObject().value("cartUrl");

    if (cartUrlValue == QJsonValue::Undefined)
        throw std::runtime_error("Could not get data.cartUrl field from results.");
`

printathing:
> About Qt4: I guess we need to either: a) Kill Qt4 support across the boards or b) disable this feature when built against Qt4. a) is probably the right thing to do, but probably out of scope for this PR, so I guess I'll suggest b) for now. I'll start a TODO list of items needing addressing before we can merge.

Ok.  This seems a bit out of my depth, so hopefully someone else can take care of a or b.

src/MainWindow.ui
printathing: The export happens automatically in this code, so I think F7 is the natural "3D Print" shortcut, if we decide to have one.

t-paul:
I can have a look at that (pretty much just disabling things based on Qt version... Is it ok to just push small updates to this branch (Not sure it's possible as it's master)?

kintel:
About the size issue; I didn't think about that. Perhaps we need some UI to ask for permission to upload a massive file. There are more "correct" ways of doing this, like exporting as a stream to the network, but that probably requires a bit of refactoring and careful resource management.

src/MainWindow.ui
printathing: Adding entire dialog and such for enabling or disabling one specific feature that should just blend in with everything else doesn't really seem value-add to me.  We'd be adding complexity for something that MIGHT happen and will probably very rarely happen, and where the downside of it happening isn't really that big - a couple wasted seconds.  This isn't a "Send $1000 from my bank account to ISIS" hotkey, it's just a streamlined way to print your 3d object.  ;)  LOL  Regarding IP - Print a Thing is bound by their terms of service to keep all files confidential, and their suppliers are bound to confidentiality as well.

My preferences, in order from most to least:

1.  Leave the hotkey in there, as is, and use the tool like that ourselves during a testing phase.  See how often we accidentally hit F7, and if it ACTUALLY ends up being a problem, let's solve the problem then.
2.  Do this initial release without the F7 hotkey.  :(  Reserve F7 for this feature (nothing else should use this- is there a way to enforce this?) in the future.  Come back to this in 3 months when the whole thing seems a lot less scary.

printathing:
> * [ ]  Dialog for initial run (for enabling the feature)

What do you have in mind for this?  How about "Would you like to send your design to a 3D Printing service to be printed? *Ok* *Cancel*"?

t-paul:
Yes, pretty much. I'm a bit concerned about people just getting stuff uploaded without any warning whatsoever (and not being a lawyer I can't judge if that might be even illegal in EU, IP addresses do count as personal data already).

Basically just an info screen that will explain what's going to happen, where it's uploading to and the choice for the user to leave the feature disabled.

kintel:
Perhaps specifying the URL? "...to the 3D print service at `https://print.openscad.org`" ? As we're SSL'ing the connection there should at least be some minimal level of trust involved. ..and we can take record of the first-run permission using our preferences API.

t-paul:
Yep, that's what I'm thinking. I can do the coding part, but someone else needs to come up with the text :).

printathing:
> * `$fn = 600; sphere(10);` produces a 68Mb temp file which generates a QJson error `QJson: Document too large to store in data structure` - after this message, the file content is just missing

Regarding this.  The API has a maximum file size as well - I believe it's around 50MB.  So we're not losing any functionality by having other things break at such large file sizes.  At least we get a sensical error message. :)

t-paul:
No, we are not getting any message. The message is just dumped to stdout so not visible to the user normally. Problem is that json `insert()` does not really return a useful error in this case. Maybe it's possible to search for the `file` key after the `insert()` to see if it was added successfully?

printathing:
Regarding: networking

I agree that some more robust (and cancelable) solution would be nice.  Perhaps it's a dialog that says "Exporting file and uploading to https://print.openscad.org..." and has a "Cancel" button on it if it takes too long because you're on a cell phone connection.  Would that do it?  In that case we don't have to worry about the rest of the app being responsive - the user is then expecting to just hold their horses for a bit.  I'm just looking for the simplest thing that scratches the itch.

printathing:
> No, we are not getting any message. The message is just dumped to stdout so not visible to the user normally. Problem is that json `insert()` does not really return a useful error in this case. Maybe it's possible to search for the `file` key after the `insert()` to see if it was added successfully?

Thanks, t-paul!  I'll try to catch this case and throw a useful exception. :)

t-paul:
Ok, I have added the experimental flag and merged master to resolve the merge conflict. Lets see what Travis has to say...

src/feature.cc
```diff
@@ -25,6 +25,7 @@ const Feature Feature::Experimental3mfExport("3mf-export", "Enable 3MF export.")
 const Feature Feature::ExperimentalSvgImport("svg-import", "Enable SVG import.");
 const Feature Feature::ExperimentalCustomizer("customizer", "Enable Customizer");
 const Feature Feature::ExperimentalInputDriver("input-driver", "Enable input drivers (requires restart)");
+const Feature Feature::Experimental3dPrint("3d-print", "Design upload to Print a Thing service (requires restart)");
```
kintel: Does it require restart?

t-paul:
So tests are green, so anything else you'd like to see for an initial merge?

kintel:
Do we need Qt4 handling, or can that wait you think?

src/feature.cc
t-paul: Yes, for showing/hiding the icons (well, new file might be enough).

kintel: ok, doesn't hurt to mention it

t-paul:
We don't use Qt4 anywhere, I'm only aware of FreeBSD needing this currently, so I would not make that block merging. Easier to work in parallel after the merge :).

kintel:
See #2664 for the issue tracking remaining work

kintel:
As this was merged a bit too early, I've added a bunch of the remaining cleanup tasks to the new issue

t-paul:
To clarify the "experimental" thing...

When building with that code, you need to give `CONFIG+=experimental` to the `qmake` call and then enable the feature in the preferences tab.

src/mainwin.cc
```diff
@@ -2027,6 +2033,170 @@ void MainWindow::csgRender()
...
+	
+	//Clean up the reply object:
+	reply->deleteLater();
+    
+    char errorMessage[150];
+	
+    //If it wasn't successful:
+	if (! (statusCodeV.toInt()==200 or statusCodeV.toInt()==201 ))
+	{
+        sprintf(errorMessage, "An error occured while contacting the 3D Print API.  \nStatus code returned: %d", statusCodeV.toInt());
```
MichaelPFrey: Just stating, meaning I have no preference either way:
PRINTB internally uses boost::format.
boost::format uses a dynamic stream, so you do not need the char buffer `char errorMessage[150];`.

As this is part of the GUI, QString::sprintf() and QString::arg() would also be option, but as the error message is a char[], using QString would be a bit overcomplicated.

kintel: I added this to #2664.

src/MainWindow.ui
MichaelAtOz: Ctrl+p is traditionally to print the window content to a paper printer, while OpenSCAD doesn't currently print the editors content (or the console or preview pane for that matter) I don't think this standard function should be messed with (IMO). Unless a dialogue box is to be used, which could later have options to paper-print these things??

Perhaps Shift-Ctrl-p?

MichaelAtOz: > makes perfect sense

https://xkcd.com/1172/

For your use case. Not mine. 
No way do I do Preview -> Render -> 'send something unexamined to a printer - particularly if it going to cost me real money'.

Personally I would like F7 to be export to STL or better for the future Export Dialogue that has been discussed from time to time.

MHO.

printathing: Sending it to Print a Thing doesn't cost any money.  It just uploads it to a site where you can preview the part in 3D and then add to to your cart.  There are about 5 different steps where you can choose to stop what you're doing.  Pressing F7 spends nothing.

MichaelAtOz: You are viewing this solely from your perspective, that the only thing anyone would ever want to do after rendering is to upload it to your print service. Fare enough that it what you want to promote.

I'm saying there are other use cases which also make perfect sense. While I hope you all success, I believe that upload will always be a niche case. As such using such a prominent key is not merited when there are more holistic core functions which suit. IMHO. But I am not the king of UI.

printathing: I can see the case for F7=export.  But I would think that if that was going to happen it would have happened by now?  I think the reason might be because there are lots of different types of export, so it's hard to apply one shortcut to them.  stl, off, amf, etc.  I'd be ok with F7=some "select which filetype you'd like to export" dialog, and F8=3DPrint, if it comes to that.  I'm guessing such a dialog gets rid of the convenience that a "shortcut" normally offers, though.

But I still think this just looks polished and makes sense:

https://ibb.co/S61Pcd7

Ultimately, this might be one thing where two intelligent and well meaning people might still disagree.  Nothing against either of us.

MichaelAtOz: An Export dialogue box has been on the books for some time #649.

PR #2647, opened 2018-12-24
add StackCheck to rail recursion implementation, 

### LINKED ISSUE #2632: echo in rekursive function crashes OpenSCAD
The following endless recursion crashes at least 2018.11.20.ci571 because of echo I assume. 

```
a = rek(10); 
function rek(a) = echo(a) rek(a); 
```

MichaelPFrey:
as echo() seams to able to consume stack space.

This should fix #2632, but as stack is compiler and operating system  dependent, it should be tested under all major systems.

### DIFF
--- a/src/function.cc
+++ b/src/function.cc
@@ -27,6 +27,7 @@
 #include "function.h"
 #include "evalcontext.h"
 #include "expression.h"
+#include "stackcheck.h"
 
 AbstractFunction::~AbstractFunction()
 {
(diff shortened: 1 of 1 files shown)


t-paul:
Are you sure that makes a difference? There is no recursion in this case (unless there's a second one where the tail recursion elimination was not possible).
I have not debugged yet, but I'm pretty sure issue in #2632 is not detected as tail recursion, so it should never run this code path.

MichaelPFrey:
You are right, the code is not threaded as tail recursion.
But now I am confused that my builds do not crash. I know that they did - but now they don't. The nightly build 2018.12.08 still does. I am confused.

Issue #2066, opened 2017-07-06
Export to path with spaces


darksnakezero:
When using the export function under windows and you export to a path that contains spaces, nothing happens.

Export to a simple path without spaces does work.

"Save as" works as intended. If it is too complicated to fix, at least a warning\error would be nice.

MichaelAtOz:
Which version of OpenSCAD?
What type of export?
Can you give a specific example of the path & filename?
Did you use the dialogue box to select the directory, or did you type the directory/filename?

MichaelAtOz:
FIXME: [Export CSG](https://github.com/openscad/openscad/blob/a84da9a110b46543520f7e2e73e69b0f071c90db/src/mainwin.cc#L2228) needs to use [completeBaseName()](http://doc.qt.io/qt-4.8/qfileinfo.html#completeBaseName), not baseName().

MichaelAtOz:
I can't reproduce this with 2015.03. or a recent snapshot. Windows 7/64
@darksnakezero Also what version of Windows?

darksnakezero:
Windows: Windows 10 Pro 64 bit
Openscad: 2015.03

working path U:\output.svg
non-working: C:\Benutzer\L ö\Desktop\output.svg.

German version I use:
From the menu: Datei->"Speichern Unter" works
Datei->Exportieren->Select any option
Then "choose Filename and path" window opens, I select said path and it closes without doing the export.

English translations could be:
File ->"Save As" working
File -> Export-> Any option -> "choose filepath"   not working, filepath window closes without error, but doesnt save the file

MichaelAtOz:
a. Check the console for any error messages.
b. can you try it with a directory with a space, but no unicode chars.

darksnakezero:
a) No error in console
b)
Path "U:\a space\output.svg" working
Path "U:\örk\output.svg" NOT working
So the problem really is the 'ö'. It is not really needed to support those paths I guess, but I would expect some warning popup in that case :-)
Again, I am sorry that I didnt test enough before writing that bug report.

MichaelAtOz:
Reproduce unicode export issue.
Export as CSG works, others don't

https://github.com/openscad/openscad/blob/a84da9a110b46543520f7e2e73e69b0f071c90db/src/mainwin.cc#L2263 
Seems to be the way to go.

kintel:
[...]

Issue #223, opened 2012-11-28
openscad editor should use utf8 text encoding on windows


sadr0b0t:
Hi,
I'm using openscad on linux and windows with same model file with some comments in Russian (utf-8). On Linux editor encoding is UTF-8, on Windows - cp1251, so comments are corrupted - have to recode file each time. I think openscad should be UTF-8-only, all other encodings should stay in the past.

PR #2648, opened 2018-12-24
Exported files now default to the same folder as the file in the window.

nophead:
If the folder or filename are changed then that is remembered per window and file type until a New, Open or Save As.

New files that have not been saved default to exporting into the platform specific default Documents folder.


kintel:
LGTM. @t-paul Did you have anything else to add?

t-paul:
Nope, I did some testing on Linux which looked ok.

t-paul:
@nophead mentioned, which is something to consider
> PlatformUtils::documentsPath() seems a reasonable default for an un-saved file. The only problem I see is it is only a std::string, which I convert to a QString, so it probably has encoding issues if the path has non-ascii characters

Might be ok to merge and then just go via testing (mostly on Windows as file system encoding is mostly problematic there).

nophead:
I think OpenSCAD is generally broken on Windows with international characters. I don't even know how it is supposed to work as I have never done it in C++ because 8 bit ASCII is all I need / use for English.

Python3 handles it well by converting everything to 32 bit unicode before it gets to the Python level. I.e. command line args and environment variables are converted from something locale specific, latin-1 in my case.

MFC had wide characters last time I looked at it, which are only 16 bit and you have to use wmain instead of main, etc. OpenSCAD uses main with 8 bit characters, so I don't know if it can be made to work like that. Perhaps looking at the source to Python3 would give some clues.

Or perhaps QT handles it with QStrings. The problem is PlatformUtils doesn't use QT.

MichaelAtOz:
> I think OpenSCAD is generally broken on Windows with international characters

I don't think so, every so often someone reports an int'l bug and it gets fixed, so I assume it works.
When UTF was introduced with `text()` the GUI handles everything OK (ish?), I recall a flurry of file/path issues around that time with a global change to the 'proper' method. I'll poke into some history...

t-paul:
Yes, I believe it's pretty much the handover between Windows file system when read via boost functions to Qt (e.g. import issues like #1286). For `text()` there's no issue as it is by declaration UTF-8 based so we also expect the script content to be UTF-8.

MichaelAtOz:
History #223, (#2066) dupl. of #113, but [this](https://github.com/openscad/openscad/pull/1296#issue-32529786) seems to be the hot spot.

MichaelAtOz:
I'm no expert, but a brief look says QString & std::string both support unicode.
So I'd suggest testing before doing much as @t-paul suggested.

nophead:
> I don't think so, every so often someone reports an int'l bug and it gets fixed, so I assume it works.

This test has always failed on my system.

```1122 - openscad-nonascii_sfære (Failed)```

It is excluded from the test suite in the CI build, together with other tests that fail on Windows:

```C:\msys64\usr\bin\bash -lc "cd /c/projects/openscad/tests && /mingw64/bin/ctest.exe -j2 -E opencsgtest_issue1165|opencsgtest_issue1215|throwntogethertest_issue1089|throwntogethertest_issue1215|openscad-nonascii|import_3mf|3mfpngtest|3mfexport'"```

I think that if I had international characters in my environment variables PlatformUtils would go wrong because they would be coded as latin-1, not utf8. You can't go from 8 bit ascii to 32 bit unicode without knowing the encoding, which I think varies under Windows depending on what language you use.

Yes the source code is UTF-8 by definition but command line arguments, filenames and environment variables are not on Windows.

In this PR it hardly matters as it is only the edge case if you try to export a new file before naming it and your documents path has international characters.

**Comment by nophead** [comment]
In fact, not only do the tests not work but the test suite itself is broken again, after I had fixed it once by converting it to Python3, where this all gets easier.

**Comment by MichaelAtOz** [comment]
I looked up [getFolderPath()](https://docs.microsoft.com/en-us/dotnet/api/system.environment.getfolderpath?view=netframework-4.7.2) & [string (Win variety)](https://docs.microsoft.com/en-us/dotnet/api/system.string?view=netframework-4.7.2) & [Qstring](http://doc.qt.io/qt-5/qstring.html) & std::string (don't have the link at hand) which all seem to be unicode aware.

I did a bunch of Unicode file/path tests in the hazy-past, so it if is broken it is not that broken, or re-broken. 
> "There's a hole on my Unicode dear Lisa dear Lisa. So fix it her Henry dear Henry, we'll fix it dear Henry, fix it we will" not necessarily right now.

Sorry - long day...

**Comment by nophead** [comment]
std::string isn't really unicode aware. It can hold utf-8 but doesn't know where the glyph boundaries are. I.e. strlen() will report the length in bytes, not the number of glyphs.

See https://stackoverflow.com/questions/402283/stdwstring-vs-stdstring

**Comment by nophead** [comment]
It all depends on what boost fs::path does. https://github.com/openscad/openscad/blob/master/src/PlatformUtils-win.cc#L71

The comment:
```Normally the path encoding is system local and we don't want to force conversion to UTF-8.```

Implies it won't work because QString expects const char* that I pass it from c_str() to be utf-8. I think the path from PlatformUtils::userDocumentsPath() will be in the correct encoding to pass to a file system function but that isn't utf-8 on Windows.

**Comment by nophead** [comment]
Perhaps I should have used this: http://doc.qt.io/qt-5/qstring.html#fromLocal8Bit when converting to QString.

**Comment by MichaelAtOz** [comment]
Hence

> So I'd suggest testing before doing much as @t-paul suggested.

Where I mean, and suppose tp did, get it into master and suck-it-and-see {it 'aint gona cause Armageddon}.

**Comment by MichaelAtOz** [comment]
> Initializing a String
One way to initialize a QString is simply to pass a const char * to its constructor. For example, the following code creates a QString of size 5 containing the data "Hello":
QString str = "Hello";
QString converts the const char * data into Unicode using the fromUtf8() function.
In all of the QString functions that take const char * parameters, the const char * is interpreted as a classic C-style '\0'-terminated string encoded in UTF-8. It is legal for the const char * parameter to be 0.

I saw similar for std::string, anyway, it's merged, a new exe will be there, we can have a look, not something to add to seasons stress...

> It isn't easy to test because 

I think you have XP-[alidocious](https://www.merriam-webster.com/words-at-play/origin-supercalifragilisticexpialidocious), Windows has been Unicode for a while, anyway we will see, not something to stress about.

**Comment by nophead** [comment]
Yes Windows has supported Unicode for longer than I can remember but it does it with a duplicated set of wide character system calls and UTF-16. On the other hand Unix does it with UTF-8. So I think to handle both OpenSCAD should use the wide version of main on Windows to get the command line and environment in UTF16 and have platform specific wrappers that convert to UTF-8 to be compatible with Unix above the platform layer. All file operations should convert back to UTF-16 and use the wide filesystem functions on Windows.

Passing UTF-8 to QT seems OK. I am not sure how boost is supposed to work.

**Comment by MichaelAtOz** [comment]
Yep, broken in two parts, whatever caused the console message to be scrambled, and the export didn't work.
pr2648
(hence testing failed, but no major damage)
I'll look into boost.

**Comment by MichaelAtOz** [comment]
BTW all I did was copy a few unicode chars from Character Map & pasted into the new folder name in explorer.

**Comment by MichaelAtOz** [comment]
For comparison, I ran an earlier version pre nophead's changes.
I selected the most recent logo.scad in the welcome screen, which presumably was the unicode-path one I saved above.
```
Failed to open file C:/Users/MeB/Documents/3D-REPRAP/Things/My Things/tests/pr2648-unicode/New folder-ÆÇĤŴǄƔ/logo.scad: The system cannot find the path specified.
```
So it appears to be stored OK & console handles it but doesn't find the file, pre-version can't deal with the unicode path, but; File/Open, select the directory, click on logo.scad &
```
Loaded design 'C:/Users/MeB/Documents/3D-REPRAP/Things/My Things/tests/pr2648-unicode/New folder-��HW??/logo.scad'.
```
it loaded the file OK, but console is garbled. So there is more to fix.

I'll raise another issue later to track these additional unicode problems.

MichaelAtOz:
e.g. in pre version

> Can't open file "C:/Users/MeB/Documents/3D-REPRAP/Things/My Things/tests/pr2648-unicode/New folder-��HW??/logo.csg" for export

MichaelAtOz:
BTW 'Exported files now default to the same folder as the file in the window' for non-unicode works well so far.

nophead:
Somehow part of this change has been reverted and it no longer works on the current master. This line https://github.com/openscad/openscad/blame/master/src/mainwin.cc#L2525 git blame says was changed 11 years ago but it should have my change from 8 days ago. Is git broken?

t-paul:
Argh, there was some merge conflicts. Maybe I did something wrong. I'll have another look.

nophead:
How does it get erased from the blame though? Surely it should say somebody changed it back, not that I never changed it?

t-paul:
I guess because it ended up as it was before missing the changes from this PR otherwise it should have indicated my merge commit.

1) you changed to implement this and it was merged first
2) someone else changed the variable name from camel case to snake case
3) when trying to resolve the merge conflict I picked the wrong one and changed the case for of the variable

t-paul:
I think I have it corrected, just giving it some testing... should be just: `filename` -> `exportPath(suffix)` what's missing.

t-paul:
Ok, should be fixed now. Thanks for catching that.

Hoquiamite:
I probably have no business being here, but I think this is the only group that can solve my problem. I'm an old(80) programmer that worked for several decades with COBOL then another decade with C. Six years ago I got interested in 3D printing, and since I was accustomed to code, I took up designing my repair components using openSCAD and it has served me well. A while back, I suffered a stroke, and things have gotten difficult for me. A few weeks ago, while working in openSCAD, I fumble-fingered something, and buggered up my screen layout. While attempting to get it back to the original layout, I only made matters worse, and I think I've painted myself into a corner and locked myself out of the tools necessary to straighten it out. I tried to figure out where the configuration is stored, but could not figure it out. I then decided to uninstal then reinstall openSCAD, but when I opened the reinstalled the program, the layout was still buggered.

Where, pray tell, is the configuration stored and how can I get back to the standard layout?

Hoquiamite:
my email is claremlove@gmail.com

Hoquiamite:
DSCN0010

jordanbrown0:
I'll help @Hoquiamite via e-mail.

Issue #517, opened 2013-10-27
Rotate by 90° is not accurate


hroncok:
Maybe it is caused by bad approximation of pi? But rotating the object by 90 degrees does not rotate the object exactly by 90 degrees.

See this (Note: updated example 2018-12-07)
``` c
difference() {
  cube(10,true);
    
  rotate([0,90,0]) 
      cube(10,true);
}
```

I know that pressing F5 would generate what I call Schrodinger's walls, but pressing F6 should render empty scene. Instead, this is produced:
<img width="512" alt="screen shot 2018-12-07 at 2 33 54 pm" src="https://user-images.githubusercontent.com/186094/49668772-28d01d00-fa2d-11e8-88d3-ef6bd9cd6061.png">

The following code will produce no output, as expected:

``` c
difference() {
  cube(10,true);
  cube(10,true);
}
```
##

kintel:
This was fixed a good while ago, by making certain angles behave perfectly

nophead:
How was this fixed? The code here: https://github.com/openscad/openscad/blob/2dc786bd5998c7e555c403d54d503e39e86abfef/src/transform.cc#L111 still looks inaccurate but the example above does seem to work now, so I am confused.

The user mode trig functions are now accurate but I can't see where they are used in openscad.

kintel:
I remember there being lots of discussions and some work done, and the examples accompanying this ticket now works.

If someone can reproduce this with a newer OpenSCAD, feel free to reopen.

kintel:
Thanks for the updated test - I've updated the description with the new example.

nophead:
This is my latest user mode version, updated to take both an angle and a vector:
```
module rotate(a, v) 
{
    if(v == undef) {
        av = len(a) == undef ? [0, 0, a] : a;
        cx = cos(av[0]);
        cy = cos(av[1]);
        cz = cos(av[2]);
        sx = sin(av[0]);
        sy = sin(av[1]);
        sz = sin(av[2]);
        multmatrix([
          [ cy * cz, cz * sx * sy - cx * sz, cx * cz * sy + sx * sz, 0],
          [ cy * sz, cx * cz + sx * sy * sz,-cz * sx + cx * sy * sz, 0],
          [-sy,      cy * sx,                cx * cy,                0],
          [ 0,       0,                      0,                      1]
        ]) children();
    }
    else { // formula from wikipedia transposed.
        s = sin(a);
        c = cos(a);
        C = 1 - c;
[...]

PR #2656, opened 2018-12-27
Accurate rotations in degrees. Fix for #517.

nophead:


nophead:
The second part is a replacement for AngleAxisd if it was factored out and then it could be used three times for the first part as before, but it is much less efficient to do it that way. It would generate three matrices that need to be multiplied together and each one is more computation to make than the three angle version.

nophead:
I see there are uses of AngleAxis3d elsewhere, so yes it is worth factoring it out and using it for those.

nophead:
Something went wrong with the travis build.
```
Warning: Permanently added 'files.openscad.org,104.236.249.91' (ECDSA) to the list of known hosts.
Enter passphrase for key '/home/travis/.ssh/id_rsa': 
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

t-paul:
I think that is normal, the ssh keys are not available on pull requests for security reasons, so ssh gets stuck. We probably need to fix that by not calling the upload in the first place.

nophead:
It normally works though. Is it due to the IP address changing perhaps?

t-paul:
Not sure, but I guess as it only tries to upload on test-suite failure, it's not happening that often.

t-paul:
Hmm, yes, something strange going on. It's green on Ubuntu Trusty and MacOS but fails on Ubuntu Xenial.

nophead:
Actually one of tests that failed is a CSG difference like this:
```
-			multmatrix([[-1, -1.22465e-16, 0, 0], [1.22465e-16, -1, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]) {
+			multmatrix([[-1, 0, 0, 0], [0, -1, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]) {
```
So the new version is more accurate and different, as expected. The odd thing is why is this allowed on some platforms and not others?

nophead:
The other odd thing is the point it went wrong because I doubt the optimisations I did affect the numerical results at all. The differences I see when I regenerate the files here is that lots of multimatrix values very close to zero are now actually zero. But that will have been the case right at the beginning as that is the point of the mod. Any rotations by 90 degrees will have been affected from the start. Why did Xenial suddenly start noticing and the others not?

I could really do with seeing the test report that failed to upload!

nophead:
Looks like it isn't the small changes that are problem. Something else must go wrong on Xenial. It must compile differently.

nophead:
Hmm, not only is Xenial half as fast as Trusty it also seems to get auto wrong. Hard to know what it compiles to without seeing the results.

kintel:
Not even sure if Xenial support in Travis is out if beta.
To see the test failures, you'd have to open a PR from a branch in the OpenSCAD repo itself.
I'll see if I can set that up for you

t-paul:
I can import the branch into the main repo, that should make the test result upload work... Let me know if that still would help. Right now, travis is green \o/.

nophead:
Yes the problem is fixed by replacing a couple of autos with Vector3d, so it looks like a compiler bug on Xenial. I am assuming it simply got all the rotations wrong with the auto in place. Looking at the Eigen normalised function template I am not too surprised. I am guessing that is what failed. It must have deduced a type that was subscriptable though, so I have no idea what that would be if it wasn't Vector3d.

Anyway I don't think there is much point investigating further. I prefer types to auto as it makes the code easier to read IMHO.

kintel:
@nophead It's not a compiler bug, it's a subtle pitfall in Eigen, see https://eigen.tuxfamily.org/dox/TopicPitfalls.html

Basically, Eigen evaluates all expressions lazily, unless you specify how to store temporary results. If you use auto, you're responsible for not changing any values to which Eigen may hold a reference.

kintel:
Good question, Trusty is using a pretty old gcc, perhaps that compiler is not too well optimized. A few days in front of a VM would probably clarify that, but not sure if it's worth anybody's time : /

kintel:
An `Affine3d` object describes an affine transformation, not just rotation. In effect it's a 4x3 matrix + a n extra [0,0,0,1] row to pad it to 4x4, see https://eigen.tuxfamily.org/dox/classEigen_1_1Transform.html

nophead:
So in the original code it sets the rotation part of the transform to a 3 by 3 matrix here: https://github.com/openscad/openscad/blob/2dc786bd5998c7e555c403d54d503e39e86abfef/src/transform.cc#L140 but it simply assigns a 3 by 3 matrix to it here: https://github.com/openscad/openscad/blob/2dc786bd5998c7e555c403d54d503e39e86abfef/src/transform.cc#L152

And here it assigns a 4x4 matrix: https://github.com/openscad/openscad/blob/2dc786bd5998c7e555c403d54d503e39e86abfef/src/transform.cc#L209 so that is what I did because I translated user code that used multmatrix.

What I am asking is can I just create 3x3 rotation matrices and assign them to the transform. Should I assign it with .rotate() or just a straight assignment? Is there any difference? I am surprised it allows direct assignment of both 3x3 and 4x4 matrices.

kintel:
You should be able to assign from both a 3x3 or 4x4 matrix. Eigen will just call `makeAffine()` afterwards in case the 4x4 matrix doesn't describe an affine transformation, see https://eigen.tuxfamily.org/dox/classEigen_1_1Transform.html#title22

When in doubt, I'd choose the construct that makes the code itself best show your intent.

nophead:
Looks like the inverse trig functions need similar treatment as they don't yield exact degrees in most cases. I wonder if it should be special cases got the common angles or just a function that rounds values very close to an exact degree as the biggest error is only about 3e-14.

It would produce tiny dead spots in the curve, as presumable the special cases do the other way. That might effect numerical convergence in slope following algorithms.

I don't know why the standard functions are in radians on a computer. It makes sense for algebra but using a unit that has no exact representation on a computer is bonkers.

nophead:
For atan2 I check if the rounded degrees are within 3.0E-14. That doesn't work for asin and acos that get increasingly inaccurate at the flat bits. For asin, acos and atan I check if the nearest integer degree gives the same sin, cos or tan, if so I return that integer, else the non-rounded value.

Not very efficient but I don't think the inverse functions will get called all that often.

nophead:
So as usual I have got myself in a complete mess with git.

It said I had a conflict with master in func.cc because @MichaelPFrey changed it but really it wasn't a conflict because he changed different lines to me. In an attempt to resolve it I rebased my changes on the latest master and then fixed the conflict manually. That worked locally but when I tried to push I got this:

```
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
```

How can I be behind my own remote? Anyway I did what it suggested and did a pull which was obviously very wrong. That undid my conflict fixes so I had to resolve them again. Now my branch has all of Michael's changes in at was well because it is no longer rebased to the tip of master.

How do I get out of this mess and what should I have done to resolve conflicts with master?

nophead:
And the latest master does not run on my system:
$ ./release/openscad.exe
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid

t-paul:
I had that error too, but it was gone after `make clean && make`.

nophead:
Yes that fixed it thanks.

kintel:
git rebase combined with git merge often causes challenges like this. Here are a few tricks I use when things get messy (or to avoid mess):

* Keep your origin and upstream separate:
```
git remote add origin git@github.com:kintel/openscad
git remote add upstream git@github.com:openscad/openscad
```

* Know your commits and rebase onto a known target. e.g. to rebase your last 10 commits onto upstream master:
```
git fetch upstream
git rebase --onto upstream/master HEAD~10
```

* Simplify your history:
```
git rebase -i HEAD~10
<reorder and squash>
```

* Forget all commits, but leave working copy as is:
```
git reset HEAD~
git add <...>
git commit <...>
```

nophead:
After you have done all that can you push it to an existing branch on github? That seems to be my problem. I had a nice rebased version locally but it refuses to push it to the PR already open.  
```
hint: Updates were rejected because the tip of your current branch is behind its remote counterpart. 
```

Clearly untrue as it was the same but based on a later version of master and with the conflicts resolved. So it was ahead, not behind in my view. The push would have to sort of stamp over the remote version to rebase it, so I presume that is the problem.

Should I just scrap this and open a new one?

Is it better to merge the tip of master rather than rebasing on it?

PR #2669, opened 2019-01-01
Accurate rotations in degrees. Fix for #517.

nophead:
All internal trig operations now in degrees, except dxf ellipses!
Inverse trig functions now give integral degrees when round tripped.
E.g. asin(sin(x)) == x exactly when x is an integer.


kintel:
@nophead Is this ready for review, or are you still tinkering with it?

nophead:
Yes it is ready for review. 

Getting rid of the uses of M_PI doesn't seem to make as much difference as I thought it would. I think that is because of the 2D system being integers and the grid snapping, etc.  It is still impossible to line everything up.

thehans:
> All internal trig operations now in degrees, except dxf ellipses!
> Inverse trig functions now give integral degrees when round tripped.
> E.g. asin(sin(x)) == x exactly when x is an integer.

Nice, so can/should we update the trig inverse tests for exactness then?
https://github.com/openscad/openscad/blob/master/testdata/scad/functions/trig-tests.scad#L98-L102

nophead:
Yes, epsilon of 0 should work. I added a new test https://github.com/openscad/openscad/blob/master/testdata/scad/functions/inv_trig_tests.scad but I will update the old one and remove mine.

nophead:
Done via #2679.

Issue #2664, opened 2018-12-30
Accept 3D printing integration as a stable feature


kintel:
TODO:
* [x] Handle Qt4 builds
* [x] Resolve handling of async response (busy loop vs. dialog vs. background)
* [x] Dialog for initial run (for enabling the feature)
* [x] Managing upload size (too big, progress, cancel)
* [x] Manage maximum QJsonObject size
* [x] Decide on shortcut to access service
* [x] ~If possible, remove trailing slash in service endpoint URL (not possible, needs new API version)~
* [x] Clean up uses of `sprintf()` - use stringstream instead.
* [x] Go through and clean up debug output and error handling
* [x] Use SSL to access files.openscad.org resources
* [x] Move the toolbar icon to the right of and into the same group as "Export STL"
* [x] Generic error "Cannot 3D Print due to errors." appears on stdout/stderr. Can we just will this generic error message and replace it with a real, concrete error message in the OpenSCAD console?
* [ ] ~On initial 3D print, this error is printed to stderr: `QFSFileEngine::open: No file name specified`~ this comes from `QTextBrowser::setHtml()` so I'll call that a QT bug at this point
* [ ] Can we make the 3D Print service selection UI a bit more friendly? Right now it's a radio box with a bunch of HTML above, with no direct connection between the HTML content and the radio boxes. This is likely just a question of UI tweaks, so it could be deferred..

t-paul:
See #2688.

kintel:
In terms of shortcut, I like the suggestion to make `F7` STL export and `F8` 3D printing.
In the future we may implement toolbar editing to allow F7 to be bound to other export formats.

printathing:
> In terms of shortcut, I like the suggestion to make `F7` STL export and `F8` 3D printing.
> In the future we may implement toolbar editing to allow F7 to be bound to other export formats.

Agreed!

t-paul:
I'd leave the UI details till we got some feedback, everything else should be covered by #2708.

PR #2672, opened 2019-01-02
Add info dialog for printing export and add support for OctoPrint.

t-paul:
This adds a printing export dialog which is enabled by default and allows switching between upload to local printer or external print service.

Text is all dummy, needs to be replaced (PrintInitDialog.html).

grafik

Also this adds support for sending designs to a local OctoPrint using the REST API, including configuration for slicing and direct printing.

grafik

The ToDo list of #2664 pretty much applies here too, only closing the "initial run dialog".


printathing:
I'm nervous about embedding marketing content in an html file that's in the software itself.    Can we get rid of the lorum ipsum part and just have the link to to "https://printathing.com/#OpenSCAD"?  We might move the about, terms of service and privacy policy as well, so let's remove that as well rather than having dead links at some point. They can use the site navigation.

kintel:
We could also host a service descriptor file at www.openscad.org, similar to how we host the auto-update descriptor. That way we can update it asynchronously. All decisions so far was done to minimize the efforts needed to pull this off, but we may have underestimated how minimal it's possible to get.

printathing:
Can we start with just the link to our website and at a future time, if we think it's an issue, do a "load an extra file from the website"?  This whole effort is getting a lot of feature creep, IMHO.

src/PrintInitDialog.html
```diff
@@ -0,0 +1,46 @@
...
+  </style>
+</head>
+
+<body>
+
+<h1>Exporting Designs for 3D Printing</h1>
+
+<p>This feature allows direct upload of the current design to a 3D Printing Service.</p>
+
+<h2>Print a Thing</h2>
```
kintel: As briefly discussed, since the URL points at `print.openscad.org`, the service in questions depends on who manages that backend server, which isn't tied to the OpenSCAD binary. We should probably pull this info from our website.
..and if we're offline, we cannot use it anyway, so perhaps we need a check for that?

src/PrintInitDialog.ui
```diff
@@ -0,0 +1,120 @@
...
+      <string>Print service selection</string>
+     </property>
+     <property name="flat">
+      <bool>false</bool>
+     </property>
+     <layout class="QVBoxLayout" name="verticalLayout">
+      <item>
+       <widget class="QRadioButton" name="radioButtonPrintAThing">
+        <property name="text">
+         <string>Upload Design to Print a Thing</string>
```
kintel: s/Print a Thing/<something..print service?>/

src/PrintInitDialog.ui
```diff
@@ -0,0 +1,120 @@
...
+    <widget class="QGroupBox" name="groupBox">
+     <property name="title">
+      <string>Print service selection</string>
+     </property>
+     <property name="flat">
+      <bool>false</bool>
+     </property>
+     <layout class="QVBoxLayout" name="verticalLayout">
+      <item>
+       <widget class="QRadioButton" name="radioButtonPrintAThing">
```
kintel: s/radioButtonPrintAThing/radioButtonPrintService/ ?

src/OctoPrint.h
```diff
@@ -0,0 +1,73 @@
...
+#include "Network.h"
+
+class OctoPrint
+{
+public:
+	OctoPrint();
+	virtual ~OctoPrint();
+
+	const QString url() const;
+	const std::string api_key() const;
```
kintel: We should try to stick to camelCase. I know we're violating this in a bunch of places, so I'm happy to defer this to a later cleanup if it's annoying to deal with right now.

src/mainwin.cc
```diff
@@ -2062,35 +2064,157 @@ void MainWindow::csgRender()
...
+	}
+
+	exportFileByName(this->root_geom, exportFileFormat, exportFileName.toLocal8Bit().constData(), exportFileName.toUtf8());
+
+	OctoPrint octoPrint;
+	try {
+		this->progresswidget = new ProgressWidget(this);
+		connect(this->progresswidget, SIGNAL(requestShow()), this, SLOT(showProgress()));
+
+		const QString fileUrl = octoPrint.upload(new QFile(exportFileName), userFileName,
```
kintel: Is there a reason why you don't construct the QFile instance inside the `upload()` function, as you're handing over ownership to it anyway?
Just nice to make it easy to understand object ownership by keeping the owner closer to the point of creation.

src/Network.h
```diff
@@ -0,0 +1,128 @@
...
+class NetworkException: public std::exception
+{
+public:
+	NetworkException(const QNetworkReply::NetworkError& error, const QString& errorMessage) : error(error), errorMessage(errorMessage) { }
+	virtual ~NetworkException() {}
+
+	const QNetworkReply::NetworkError& getError() const { return error; }
+	const QString& getErrorMessage() const { return errorMessage; }
+
+	virtual const char* what() const throw()
```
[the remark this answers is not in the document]
t-paul: done

src/MainWindow.h
```diff
@@ -183,6 +183,8 @@ private slots:
 	void csgRender();
 	void csgReloadRender();
 	void action3DPrint();
+	void sendToOctoPrint();
+	void sendToPrintAThing();
```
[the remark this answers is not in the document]
t-paul: done

src/PrintInitDialog.html
t-paul: Sure, read info via network and if not available fall back to just local OctoPrint.

src/PrintInitDialog.ui
t-paul: done

src/OctoPrint.h
t-paul: done

src/mainwin.cc
t-paul: I guess I was thinking to change the export stuff so we don't need the temp file but that would have caused too much changes in the end.

changed

kintel: right, we could ideally provide a QIODevice or smth. if we wanted to stream any info.

tests/FindLib3MF.cmake
```diff
@@ -36,7 +36,7 @@ if ("${LIB3MF_LIBDIR}" STREQUAL "")
 endif()
 
 if (NOT ${LIB3MF_LIBDIR} STREQUAL "")
-  set(LIB3MF_LDFLAGS "-L${LIB3MF_LIBDIR}" "-l3MF")
+  set(LIB3MF_LDFLAGS "-L${LIB3MF_LIBDIR}" "-l3MF -lzip -lz")
```
kintel: Is this an unrelated fix, or accidental?

t-paul: Yep, build fix as I wanted to try exporting 3MF to OctoPrint and the new Debian package strips the packaged libs from the lib3mf dev package.

src/OctoPrint.cc
```diff
@@ -97,22 +97,23 @@ const std::vector<std::pair<const QString, const QString>> OctoPrint::get_profil
...
+const QString OctoPrint::upload(const QString exportFileName, const QString fileName, network_progress_func_t progress_func) const {
 
 	QHttpMultiPart *multiPart = new QHttpMultiPart(QHttpMultiPart::FormDataType);
 	QHttpPart filePart;
 	filePart.setHeader(QNetworkRequest::ContentDispositionHeader, QVariant{"form-data; name=\"file\"; filename=\"" + fileName + "\""});
 	filePart.setHeader(QNetworkRequest::ContentTypeHeader, QVariant{"application/octet-stream"});
 
+	QFile *file = new QFile(exportFileName);
 	file->open(QIODevice::ReadOnly);
 	file->setParent(multiPart);
```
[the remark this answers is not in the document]
t-paul: Ahh, nice. done.

Issue #2685, opened 2019-01-07
Use SSL for macOS autoupdate check


kintel:
Once #2684 is done, update Info.plist and macOS auto updater to use SSL.

PR #2689, opened 2019-01-07
Update to https links

t-paul:
This should take care of #2685 and also changes a number of other links that verified ok via https.

* [x] Review
* [x] Test on macOS

### DIFF
--- a/Info.plist
+++ b/Info.plist
@@ -114,18 +114,5 @@
 	<string>dsa_pub.pem</string>
 	<key>SUFeedURL</key>
 	<string>https://files.openscad.org/appcast.xml</string>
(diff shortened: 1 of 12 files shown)


Info.plist
```diff
@@ -113,7 +113,7 @@
 	<key>SUPublicDSAKeyFile</key>
 	<string>dsa_pub.pem</string>
 	<key>SUFeedURL</key>
-	<string>http://files.openscad.org/appcast.xml</string>
+	<string>https://files.openscad.org/appcast.xml</string>
```
kintel: We can probably also remove the `NSExceptionAllowsInsecureHTTPLoads` key as well. See history

PR #2693, opened 2019-01-10
option to disable mouse centric zoom

MichaelPFrey:
* http://forum.openscad.org/Turn-on-off-cursor-centered-mouse-wheel-zoom-td25184.html
   * #1857
   * #2611

Maybe it is not a good idea to potentially read the settings multiple times per second.

I personally prefer mouse centric zoom, but I have other issues with the 3D view, so this option is not in my interest, but adding options in general is in my interest.


MichaelPFrey:
okay, I add a member variable to `QGLView.h` and try to implement a signal/slot chain.

korpx:
Works great! Thanks!

Issue #2686, opened 2019-01-07
Snapping Polyset vertices to a grid can cause CGAL exceptions


nophead:
This code shows the problem with F5. Clearing the cache before F6 gives a correct result.
```
echo(version());
eps = 1/128;
layer_height = 0.25;

module hanging_hole(z, ir, h = 100, h2 = 100) {
    infill_angle =  45;
    render(convexity = 3) translate([0, 0, z]) {
        rotate(infill_angle)
                intersection() {
                    rotate(-infill_angle)
                        linear_extrude(height = 3 * layer_height, center = true)
                                //offset(-eps) 
                            children();

                    cube([100, 2 * ir, 2 * layer_height], center = true);
                }
        below = min(z + eps, h2);
        translate([0, 0, -below])
            linear_extrude(height = below)
                children();
    }
}

$fn = 3;
cube();
render()
    translate([20, 10])
        difference() {
            translate([0, 0, 5 / 2])
                cube([25, 25, 5], center = true);
            
            hanging_hole(z = 2, ir = 2.5)
                circle(r = 5 / cos(180 / $fn));
            
        }
```
```
ECHO: [2019, 1, 3]
Compiling design (CSG Products generation)...
ERROR: CGAL error in CGAL_Nef_polyhedron3(): CGAL ERROR: assertion violation! Expr: e_below != SHalfedge_handle() File: /mxe/usr/x86_64-w64-mingw32.static.posix/include/CGAL/Nef_3/SNC_FM_decorator.h Line: 426
```

It goes wrong because a rotated copy of a triangle stacked on top of itself has very small differences in the vertex coordinates. This leads to some very thin triangles when unioned. When these are snapped to a grid a thin triangle can become degenerate because all its points end up in a line and CGAL then rejects it. It can also get flipped as the winding order can change when the triangle is thin.

It works in F6 because the union is never converted to a Polyset. F5 caches a broken Polyset so F6 does not work after F5.

Polyset::quantize() needs to detect degenerate triangles but it can't simply remove them. It needs to stitch up the hole that is left. It also needs to fix the winding order in some cases as well. Possibly there are other topology breaking things quantization can do as well. It can certainly create self intersections by merging vertices that were previously distinct, so general mesh repair is needed.

MichaelAtOz:
Another example #3168

kintel:
Status as of 2024-03-24: F5 still broken when using CGAL. Works with Manifold.

MichaelAtOz:
[...]

PR #2694, opened 2019-01-10
Fix for #2686.

nophead:
Removes degenerate and very thin triangles caused by quantisation by flipping edges with non-degenerate neighbours sharing the long edge.

### DIFF
--- a/src/grid.h
+++ b/src/grid.h
@@ -110,16 +110,16 @@ public:
 		createGridVertex(v, key);
 		typename GridContainer::iterator iter = db.find(key);
 		if (iter == db.end()) {
-			float dist = 10.0f; // > max possible distance
+			float dist = 10.0f; // > max possible distance squared
 			for (int64_t jx = key[0] - 1; jx <= key[0] + 1; jx++) {
 				for (int64_t jy = key[1] - 1; jy <= key[1] + 1; jy++) {
 					for (int64_t jz = key[2] - 1; jz <= key[2] + 1; jz++) {
 						Vector3l k(jx, jy, jz);
 						typename GridContainer::iterator tmpiter = db.find(k);
 						if (tmpiter == db.end()) continue;
-						float d = sqrt((key-k).squaredNorm());
+						float d = (key-k).squaredNorm();
 						if (d < dist) {
-						  dist = d;
+							dist = d;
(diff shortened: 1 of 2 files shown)


nophead:
While writing a test for this I have run into a problem where the PolySet I receive before edge flipping is sometimes wrong and sometimes right, non-deterministically. Flushing the cache and doing F5 again can give one of two results and so it seems can starting it from scratch.

The difference is four triangles out of 88 go through the same vertices but differently. Throwntogether shows a purple shimmering triangle but I have checked there are no holes in the mesh. It seems to be folded back on itself.

image

image

I.e. it seems the tessellation is non-deterministic. I don't think it is anything to do with my change but this wouldn't have been seen before because of the CGAL error that would follow if I didn't do the edge flip.

After my edge flip CGAL is happy with it either way but it looks to have a knife edge when it goes wrong. I do one less edge flip in the wrong case because there is one less degenerate triangle.

image

Does this wring any bells with anybody?

kintel:
I cannot think about why any of this should be non-deterministic. This is where unit tests would be nice, as it's tricky to establish valid minimal test cases : /

kintel:
I cannot remember the exact code path out of CGAL. It probably far easier to inspect that stack :)
I would think that somebody quantized before calling that function, but I may be wrong.

nophead:
The PolySet that goes wrong has already been triangulated when PolySet::quantizeVertices() is called. Other PolySets before that are quads. Mind you it does already have degenerate triangles before the grid snap I think. 

Triangulating double coordinates in floats and then snapping to a grid seems to be asking for problems. 

I don't think many of the grid values fit in a float. The grid value is 2^-20 and a float mantissa 24 bits? So grid points with values more than 16 (give or take a power of two) can't be distinct in float. Hence why STL export generates degenerate triangles.

nophead:
Another thing that has become apparent to me is that if you don't flush the cache before doing F6 and then export an STL you get a completely different result, generally much more untidy. surely this is not correct. I think perhaps there should be two different caches for F5 and F6 results when they are stored in different formats. I.e. Polysets versus Nef_polhedra. F6 shouldn't use a cached PolySet in place of a Nef_polhedra. Possibly this only happens with render().

PR #2645, opened 2018-12-23
Stack trace for function and module calls

MichaelPFrey:
wait for #2630 as they have to work nicely with each other

PR #2697, opened 2019-01-11
range check for built in modules

MichaelPFrey:
The code is somewhat reduntant, but with "stop at first warning" and #2645, it makes sense to warn as early as possible.

As I my self I have code that relays on negative module parameter not creating geometry, that is why it is an option that is off by default. (Kind of tricky: For new users, it should be on by default, but for legacy code it should be off by default)

I open the PR to see the test results and diff.

Note that `isfinite()`  is not the same as `!isinf()` - not a number is always fun.


MichaelPFrey:
nan is really annoying, as depending on the compiler, it either gets represented as nan or -nan.
I could mess with that while outputting, but really: That would come at a runtime penalty for a case that really does not matter, as outputting nan is usually  a problem with the user-code. So a bodge in the compare somewhat seams like the right thing to do, but I am open for a better idea.

MichaelAtOz:
The wiki doesn't mention -nan so this seems like a good approach.

PR #2723, opened 2019-01-24
UI Tweaks to Print Service.

printathing:
This keeps the user from having to manually resize the 3D Print Service dialog box every time it opens.  It is now large enough to show the content.  It also gets rid of a redundant "Do not upload" option that essentially makes the "3D Print" button do nothing at all.  If the user wants to back out, they can always press the cancel button.  Lastly, I made some console output a little more user friendly (IMHO).

### DIFF
--- a/src/PrintInitDialog.cc
+++ b/src/PrintInitDialog.cc
@@ -33,7 +33,7 @@
 PrintInitDialog::PrintInitDialog()
 {
 	setupUi(this);
-	this->radioButtonCancel->setChecked(true);
+	this->radioButtonPrintService->setChecked(true);
 
 	const auto printService = PrintService::inst();
 	QFile html(":/src/PrintInitDialog.html");
--- a/src/PrintInitDialog.ui
+++ b/src/PrintInitDialog.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>493</width>
-    <height>379</height>
+    <width>1100</width>
+    <height>950</height>
    </rect>
   </property>
   <property name="windowTitle">
--- a/src/PrintInitDialog.ui
+++ b/src/PrintInitDialog.ui
@@ -100,13 +100,6 @@
         </property>
        </widget>
       </item>
-      <item>
-       <widget class="QRadioButton" name="radioButtonCancel">
-        <property name="text">
-         <string>Do not upload</string>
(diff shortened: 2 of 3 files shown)


t-paul:
The whole point of that button was to allow the user to disable the feature and have a default that does never accidentally upload files somewhere. I don't agree with the dialog defaulting to external upload.

printathing:
Ok.  We each have our own opinions.  No worries.

Here's mine, for whatever it's worth:  If they clicked that button and then
were like "wait, this isn't what I want to do," I would think they would
click cancel and not click the button again.  If they do, they'll get the
same dialog again and can click cancel again.  In order to "accidentally"
send files somewhere, they'd need to:


   1. Click the 3D Print Button
   2. Click Ok.

That's two steps already.  I think that's plenty.

>

t-paul:
This is not just about opinions. I'm in EU where [GDPR](https://en.wikipedia.org/wiki/General_Data_Protection_Regulation) is active law. While I have no idea how that applies to an open source software provided, I'm voting against taking chances here. It's not needed and can potentially harm the public view of OpenSCAD.

The argument about 2 buttons is pretty weak considering `F7` + `Enter` is now export to STL and `F8` + `Enter` would be "upload to some external service".

kintel:
Looking at this with fresh eyes, I feel the current UI has a few issues:
* Print-a-thing:
   * The print a thing logo is way too big for a dialog box like this.
   * It's not obvious that the logo is a link
* Octoprint
   * Default error message is confusing: `ERROR: Protocol "" is unknown`
   * There is no link to octoprint.org
* The dialog box HTML content is confusing and doesn't map well to the radio buttons
* Since local printing and purchasing are really separate choices, I don't think "remember this" make a whole lot of sense. Also: It's unintuitive to go into preferences to uncheck that choice.

One possible solution:
1. Kill "remember" and  "do not upload"
2. Re-style radio buttons to use normal buttons and an exclusive group box. We could potentially style the buttons with appropriate logos to make it even more obvious what's going on (although that's likely a copyright challenge, and configuration challenge).
3. Only show the service descriptor HTML for the selected choice.
4. No default. After testing this a bit I find default actions like this is a bit too opaque.
5. `OK` performs the action, `Cancel` cancels, nothing is remembered.

kintel:
Superseded by #2730

PR #2730, opened 2019-01-27
Cleaned up the 3D Print dialog

kintel:
As discussed in #2723:
* Kill "remember" and "do not upload"
* Re-style radio buttons to use normal buttons and an exclusive group box. 
* Only show the service descriptor HTML for the selected choice.
* No default. After testing this a bit I find default actions like this is a bit too opaque.
* OK performs the action, Cancel cancels, nothing is remembered.

Also added OctoPrint logo with link to octoprint.org.

TODO:
* [x] The HTML for both services could be cleaned up and made nicer. Perhaps a common style; same size logo plus text would make sense. Ideas are welcome. Actual HTML better!
* [x] OctoPrint logo copyright/trademark notice in all the right places


kintel:
@printathing I believe this should address most of your concerns as well.

kintel:
screenshots updated

kintel:
@foosel We're adding OctoPrint support to OpenSCAD. See screenshot above. Do you have any concerns or wishes in terms of how we manage your trademarked logo?

foosel:
> Do you have any concerns or wishes in terms of how we manage your trademarked logo?

No concerns at all and only one wish - a bit of a margin between the text and the logo. Or alternatively some padding around the logo. Purely for aesthetic reasons though 😄 If you tell me what size you need I can also provide you with a matching file.

kintel:
@foosel Thx! I've updated the screenshot above with a bit extra space. A custom sized logo would be awesome! The target width is 128 pixels.

foosel:
Looks great as is now actually, but if you still want a 128x128px one (with the logo vertically centered as it's not at a 1:1 ratio), [there you go](https://drive.google.com/file/d/1kjsAS_cGAsUcxPTTOpwI7bOlWhJW5dtX/view?usp=sharing).

printathing:
A couple notes.  Print a Thing does not have any hypens.  "Print a Thing".  Also, the url is printathing.com, and the link should go to https://printathing.com/?utm_source=openscad (but let's not display all of that).

kintel:
Should be updated in the latest json

Issue #2636, opened 2018-12-15
Unnecessarily slow Preview


cebderby:
I had been using the Nightly as the last prod release was (presumably still is as no re-release) flickering badly with a very large model.  However, during this year I've found Preview mode very slow, and have been ignoring the Nightly and using a 2018.02 AppImage I happened to have downloaded to try.  This version has 2 advantages - fast Preview and double-click to set view focus (rotation, zoom centre).  A later AppImage from 2018.04 is much more like the current Nightly.

For comparison, for models above a certain complexity, Preview rendering time is 4s for 2018.02 against an unpleasant/unusable 36s for Nightly.  If it matters I'm using Mint 18.3 MATE 64bit.

I've been able to reproduce the problem with a trivial .scad file, in which the complexity can be changed by altering a single value, and a loop then generates some arbitrary geometry.  Hopefully this works for others, the numbers may be different depending on CPU speed etc I guess.  See attached.

[progress_bar_36s_preview_time.scad.txt](https://github.com/openscad/openscad/files/2682670/progress_bar_36s_preview_time.scad.txt)

My conclusion is that with the 2018.02 version, rendering time was up to 1s for the model plus up to 3s for the progress bar.  And that for the current nightly, the same models take up to 2s for the model plus up to ~35s for the progress bar.  (The model rendering time probably scaling linearly with complexity, the progress bar time linearly with the amount of time or fraction it is displayed).

Clearly having the progress bar take 3x (old) up to 15x (latest) the actual activity needs addressing, however my recommendation is that the progress bar is never shown for Preview, it's just too fast without!   Presumably the progress bar for Render needs to be checked, in case it's wasting time there too.

Don't know why the newer version takes 2s rather than 1s for a big model, but that's secondary to the main issue.

If possible, it would be good if the old view is cleared from the screen as late as possible, so the new preview just appears directly after it, without a 'flash' of cleared screen between.  That way the effect of tweaking a value and re-Preview can be seen the best.

I'd also love to see the return of double-click to set view focus too, if that's still possible (or ctrl-click or anything similar if there's a problem with double-click).

t-paul:
Please share the infos from Help->Library info. I can't reproduce that at all on Debian/Testing with 2018.12.08.nightly (git c257783). With numitems = 9999 I still get instant preview. Also there is no flashing here. Maybe this is due to different graphics drivers.

Double click is a known issue but there's no general solution in sight yet. (#2466).

cebderby:
[...]

PR #2735, opened 2019-01-27
Throttle progress bar updates to 5 per second

thehans:
Addresses Issue #2636

### DIFF
--- a/src/MainWindow.h
+++ b/src/MainWindow.h
@@ -15,6 +15,7 @@
 #include "export.h"
 #include <vector>
 #include <QMutex>
+#include <QElapsedTimer>
 #include <QTime>
 #include <QIODevice>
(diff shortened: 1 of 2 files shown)


t-paul:
Did you measure this really helps? From my tests so far, the issue seems to be more that there are too many events created, so it would be more effective to have `progress_report_prep()` only tag nodes where it is useful to fire the callback.

t-paul:
Ahh, found the note in #2636. That's quite some difference.

thehans:
Yes I ended up putting the throttle around the contents of `report_func` since you noted that commenting out that function gave a speedup.

t-paul:
Ahh, I see. With that big difference, it's certainly nice to have.

thehans:
OK, changed :)

PR #2741, opened 2019-01-30
datetime() built in function

MichaelPFrey:
The main idea is, that user can use this to add date and time to the model. Having date and time in the model (and the prints) allows the user to know which version of the scad file was used to create a specific print out. 

A bit of a concern is, that it is the current date time. Users might  get the idea to measure time differences, not realizing that due to the multiphase nature of OpenSCAD, that is not as useful as one might think.

The test case is a bit awkward. It does some sanity checks, but  not an actual functional test.

### DIFF
--- a/color-schemes/editor/visualstudio.json
+++ b/color-schemes/editor/visualstudio.json
@@ -35,7 +35,7 @@
     },
     "keywords" : {
         "keyword-set1" : "if else let for module function true false undef include use",
-        "keyword-set2" : "abs sign rands min max sin cos asin acos tan atan atan2 round ceil floor pow sqrt exp len log ln str chr ord concat isundef lookup search version version_num norm cross parent_module dxf_dim dxf_cross",
+        "keyword-set2" : "abs sign rands min max sin cos asin acos tan atan atan2 round ceil floor pow sqrt exp len log ln str chr ord concat isundef lookup search version version_num norm cross parent_module dxf_dim dxf_cross datetime",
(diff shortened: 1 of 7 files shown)


t-paul:
I think we may want to postpone that a bit and use [Objects](https://github.com/doug-moen/openscad2/blob/master/rfc/Objects.md) instead of a vector. I believe it's possible to start with the properties part first (no geometry). I've toyed a bit with a modified version of #2454 - changed to use the object syntax. Reason for that is that I'd like the IDF3 import to return an object which allows access to the 2D layer while still giving access to the thickness parameter.

MichaelPFrey:
this syntax took version() as reference, but I would not mind to wait to use proper objects syntax, if that is a mid-therm goal. Looking at version: Would functions like those also be updated to return an object?

I was actually surprised how easy it was to implement datetime() - the most struggle was to come up with a somewhat useful test.

t-paul:
Hmm, updating existing functions is difficult as that would break any script using those.

Good point about the time measurement which is not making too much sense, or at least might not match the expectation of what is going on. Maybe the time returned should just be fixed for one specific preview/render.

MichaelPFrey:
I think the object(s) could be implemented that they can silently be converted to a vector. That way, version could return an object (values accessible via . e.g. .year, .month, .day), but the user can still access the data with [0], [1], [2]. (Javascript has a similar duality)

MichaelPFrey:
I close that for the mean time. Waiting for the record data type seams sensible to me.

MichaelPFrey:
see also #380

Issue #2744, opened 2019-01-30
localization causing massive slowdowns on Windows


thehans:
This is more followup from the OpenSCAD forum thread by @nophead, trying to diagnose the slowness of preview for this script: https://gist.github.com/nophead/fd4ca58b5f6b19dc25bb98f7edbae90f

I tried profiling the build from CircleCI in Windows, and found that most of the time is being spent in `msvcrt.dll!setlocale`  
![setlocale](https://user-images.githubusercontent.com/252233/52020340-8a08ef80-24b6-11e9-809e-5da7fd602549.png)
The build doesn't have debug symbols so I can't see much about the execution path within OpenSCAD code that is leading to this, but un-checking Preferences -> Advanced -> "Enable user interface localization" makes it comparably fast to the linux builds I have tested.

thehans:
I could really use some assistance getting a windows build with debug symbols enabled.  Can we get CircleCI to spit one out, maybe on a separate "windows_dev" branch or something?

nophead:
I have returned to a large project I last played with March 18. Compiling
seems to have got massively slower. When it has been previewed once I
expect F5 to take about 20 seconds to run through everything and decide it
hasn't changed. It seems to be taking more like 4 minutes now.

nophead:
Now down to 11 seconds, my fault, not OpenScad.

thehans:
I got a debug build on Windows and I can see that `setlocale` is being called by `std::__convert_from_v(int* const&, char*, int, char const*, ...) ()`, which apparently happens whenever a floating point number is output to an ostream.  

I also found this short discussion which seems relevant, although I have to admit I don't fully understand:
https://stackoverflow.com/questions/22753707/is-ostream-operator-in-libstdc-thread-hostile

Here are the first few lines of stacktrace which seem to be the most called: 
```
#0  0x00007ffb99c59c19 in setlocale () from C:\Windows\System32\msvcrt.dll
#1  0x0000000001d8b801 in std::__convert_from_v(int* const&, char*, int, char const*, ...) ()
#2  0x0000000001c9c8f4 in std::ostreambuf_iterator<char, std::char_traits<char> > std::num_put<char, std::ostreambuf_iterator<char, std::char_traits<char> > >::_M_insert_float<double>(std::ostreambuf_iterator<char, std::char_traits<char> >, std::ios_base&, char, char, double) const ()
[...]

PR #2738, opened 2019-01-29
Optimizations for large/many polyhedrons

thehans:
for issue #2737

PR #2754, opened 2019-02-03
double-conversion library

thehans:
Testing a branch on openscad repo so test uploads work

Issue #2744


kintel:
Looks like some unrelated stuff, like tessellation, snuck into this PR..

thehans:
Yes I based this branch off of some prior optimizations from #2738.  
This included the tesselation change, and some beginning of the `toStream` optimization, which I did some more work on in this branch also.  

My reason for this was that all the changes were working towards optimize the same [script/use-case](https://gist.github.com/nophead/fd4ca58b5f6b19dc25bb98f7edbae90f) and made it easier for me to profile performance differences with the other bottlenecks out of the way.  

I didn't intend to be sneaky about it.  Would you prefer I split out the tesselation change?  I feel like the `toStream` changes particularly was integral to making the most out of faster double to string conversions, so I think that would ideally be kept in this PR.

kintel:
I'll review this as is - I was just checking if it was a mistake..

MichaelAtOz:
> One small missing piece

I just looked at the commit titles and didn't see it, was it added to the help/library-info?

thehans:
@MichaelAtOz It has not been added to library info.  Unfortunately there is no version information provided in the headers or any other way that I can see to retrieve it.

PR #2757, opened 2019-02-04
warnings related getvec2, dxf_cross and dxf_dim

MichaelPFrey:
based on #2575 

- rotate_extrude
- linear_extrude
- lookup
- dxf_cross
- dxf_dim


src/dxfdim.cc
```diff
@@ -60,13 +61,24 @@ ValuePtr builtin_dxf_dim(const Context *ctx, const EvalContext *evalctx)
 		ValuePtr n = evalctx->getArgName(i);
 		ValuePtr v = evalctx->getArgValue(i);
 		if (evalctx->getArgName(i) == "file") {
-			filename = lookup_file(v->toString(), 
+			rawFilename = v->toString();
+			filename = lookup_file(rawFilename, 
 			evalctx->documentPath(), ctx->documentPath());
```
MichaelPFrey: add warning, when a named parameter is supplied more then once (applies to other places here as well)

MichaelPFrey: I move that to an other to do list

PR #2758, opened 2019-02-05
Update cmake main build to work with lib3mf, Printing dialog.

thehans:
Also clean up and relocate some duplicated .cmake files between `tests/` and `cmake/Modules/`


kintel:
Looks good! We should spin up a cmake CI somewhere..

PR #2761, opened 2019-02-05
add is_list builtin

MichaelPFrey:
partially addressing #1584, but mainly addressing the mailing list thread "A warning too far?":

> WARNING: length() parameter could not be converted, in file global.scad, line 104
> v = is_undef(len(d)) ? [0, 0, d] : d;
> I think this is the standard OpenSCAD idiom to test if a parameter is a vector or a scalar. Is there any other way to do it?
> I think it is sufficient that taking the length of a scalar gives undef and using undef other than in is_undef() gives a warning, so you can't do it accidentally.

### DIFF
--- a/src/func.cc
+++ b/src/func.cc
@@ -955,6 +955,22 @@ ValuePtr builtin_is_undef(const Context *ctx, const EvalContext *evalctx)
 	return ValuePtr::undefined;
 }
 
+ValuePtr builtin_is_list(const Context *ctx, const EvalContext *evalctx)
+{
+	if (evalctx->numArgs() == 1) {
+		ValuePtr v = evalctx->getArgValue(0);
+		if (v->type() == Value::ValueType::VECTOR){
+			return ValuePtr(true);
+		}else{
+			return ValuePtr(false);
+		}
+	}else{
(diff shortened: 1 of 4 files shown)


nophead:
This still isn't a complete replacement because len(x) is also not undef for a string .  So len(x) != undef really means is x index-able.

I think it was possible to differentiate between all the possible types without generating a warning before using some clever tricks.

The cheatsheet is going to get bigger!

MichaelPFrey:
> This still isn't a complete replacement because len(x) is also not undef for a string .

Good point - is_string will be next on the list.


> The cheatsheet is going to get bigger!

Yeah, that is one of the prices to pay for the new warnings. I think making things more formal will help newcomers and increase readability - but the biggest price is, that complex scripts and libraries need to be reworked.

nophead:
Yes certainly more explicit and readable but more long-winded for power users.

I always strive for the shortest code and smallest language to solve a problem, so I am a bit conflicted.  Language additions that shorten code are always welcome but ones that lengthen it less so.

kintel:
Some higher-level tests like `is_indexable()` could, and should, be implemented in user-space. A great way of doing this would be to ship a "standard library". That's a long-standing wishlist item though.

PR #2764, opened 2019-02-05
Testing cmake on CI servers

thehans:
Creating PR to initiate Travis build


thehans:
Also I haven't tried switching circle-ci over to cmake in this branch yet so that's why those are all still passing ;)  I was mainly working on Travis first, but got a little stuck.  

I'm not sure which flags are really important, or if I should just try to match them all with the qmake build.
These are the main ones from qmake that are not in cmake, which I'm not sure about;
```
-fno-strict-aliasing	
-fPIE
-pipe
-std=c++0x	
-std=c++11
```
Should qmake be setting both `c++0x` *and* `c++11`?

cmake is setting only `-std=gnu++11` instead of `-std=c++11`, don't know if that is practically any different.

thehans:
Well, with the latest merge from master, Trusty is inexplicably passing the AMF tests that I couldn't figure out before.  No idea which change fixed that.  Mac builds are still good with the exception of lib3mf which seems is being looked into right now.

I will start looking into the AppVeyor build next, waiting for that to finish now...

thehans:
Made some progress with AppVeyor, but I'm now stalled by a bug in msys2 CGAL package where the cmake files reference the build path, not install path:
https://github.com/msys2/MINGW-packages/issues/1670#issuecomment-471409582

I temporarily broke various Travis builds while making changes for AppVeyor.  They are all building again now, but MacOS is still not finding lib3mf with cmake, even with changes from master merged in.

I still need to start testing CircleCI with cmake.

kintel:
@thehans Do you think it makes sense to set up a ~nightly build of master for the cmake build? Building for each PR is probably a bit much.

thehans:
I'm a little confused by what you mean.  I guess my understanding was that we would replace qmake with cmake entirely (by merging this PR) once the cmake build was considered working across all platforms.

cmake should only be building for this single PR as far as I know.  Or do you mean building for each commit in this specific PR is a bit much?

kintel:
1. Replacing the build system is a good goal, but I don't think I dare doing that quite yet. Until then, having all these build fixes merged would be good.

2. As a bonus, once we've merged the build system fixes to master, making the two builds run side-by would be nice as well. In this case it would be a bit much to run them both for all future PRs. Making cmake a nightly master build would be a great start. Then, once we feel comfortable, we can switch and make cmake the main build, while perhaps maintaining qmake on the side for a bit.

thehans:
OK, yeah that sounds good.  Shall I just revert the CI config/script changes for now so we can consider merging this?

A nightly Travis cmake-specific build seems like a good idea (since that's the only CI server currently working).  I'm not sure how different build types are configured(nightly vs every PR) or how to keep them separate though.  That could be done in another PR though.

kintel:
Yep, please pull out those few script changes so we can get the rest merged.

thehans:
OK those changes are reverted.  There is just one change I kept in `.travis.yml` where I removed cmake from the packages it installs the Travis environment already comes with a newer version.

PR #2772, opened 2019-02-09
2015.03: Build fix for boost 1.69

hroncok:
Cherry-picked from 8fb64db99ce05da1c72efb8b2f572995fef2ea86


kintel:
Does this actually build against 2015.03 without C++11 support?

hroncok:
I don't really know. (You don't need to merge it if there are concerns like this, I just wanted to contribute the backport back fo other interested parties.)

t-paul:
With 2019.01-RC2 out, there's probably not going to be any patch release for the ancient branch.

hroncok:
I was thinking other distro packagers.

kintel:
Sounds fair, I'll merge this to the branch, but won't release a new patchlevel update.

hroncok:
Thanks.

t-paul:
I've not heard of any distro packaging not released branches though.

hroncok:
Well we just take patches mostly. And it's always easier if they apply directly.

Issue #2702, opened 2019-01-16
Accept SVG import as a stable feature


kintel:
TODO:
* [ ] Decide on and document a minimal subset to support
* [x] Add more tests
* [ ] Call out any isolated bugs as separate issues
* [x] Test drag&drop of SVG files (tested on Linux)
* [ ] Test auto-reload and compile of imported SVG files

Supported SVG features:
* [x] Single shapes
* [x] Multiple shapes will be unioned
* [ ] FIXME: Document DPI/unit settings
* [ ] FIXME: Document how a viewport is imported
* [ ] FIXME: Document how stroke width influences imported shape

See also #1740, #1742, #2383, #2736

MichaelPFrey:
I would like to ask, how you would like to handle errors/warnings within or related to libsvg.
(for e.g. file not found or parsing error)

Using printutils would go against the intention of it being standalone library. Leaves us with error codes or exceptions. Given that this is a c++ library and project, exceptions seam to be right mechanism.
Any thoughts? I have not yet looked deep into the code and I do not know the bigger picture, so I rather ask before starting to hack.

t-paul:
#2596 adds a number of tests, mostly based on the SVG spec example files and also activates import/export test cases. I think that should cover the core functionality (= importing SVG Path).

As for error reporting, I suppose exceptions are the way to go. However, I don't think we need too much with regard to detailed warnings/parse errors messages as I suspect SVG is rarely written by hand (even though the spec seems to assume that is the main use case for some reason).

Note that #2596 also clarifies this bunch of code to have a different license (which was intended from the start but never happened to end up stated in the files). It probably makes sense have the relatively invasive changes from this PR merged first (or putting smaller changes on top of this branch).

MichaelPFrey:
I do not want to add much error handling - but some like "file not found" and a generic "error during parsing" would make sense, as silently returning no geometry can cause hard to track down issues in bigger projects.

t-paul:
Yes, that sounds great.

t-paul:
closed via #2596

t-paul:
Oh, whops, the experimental flag was not part of the PR...

PR #2775, opened 2019-02-10
Accept SVG import as a stable feature

t-paul:
Changes needed for #2702.

### DIFF
--- a/src/import_svg.cc
+++ b/src/import_svg.cc
@@ -5,39 +5,65 @@
 #include "clipper-utils.h"
 #include "AST.h"
 
-Polygon2d *import_svg(const std::string &filename, const bool center, const Location &loc)
+Polygon2d *import_svg(const std::string &filename, const double dpi, const bool center, const Location &loc)
 {
 	try {
 		const libsvg::shapes_list_t *shapes = libsvg::libsvg_read_file(filename.c_str());
+
+		double scale = 1.0;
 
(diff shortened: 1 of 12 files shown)


src/import_svg.cc
```diff
@@ -5,39 +5,65 @@
...
 			PRINTD("SVG shape");
 			const auto page = dynamic_cast<libsvg::svgpage *>(shape_ptr.get());
+			double unit_scale = 1.0;
 			if (page) {
-				width = page->get_width();
-				height = page->get_height();
+				const libsvg::length_t h = page->get_height();
+				switch (h.unit) {
+				case libsvg::unit_t::UNDEFINED:
+				case libsvg::unit_t::NONE:
```
kintel: NONE doesn't look like it's used any longer

t-paul: It's used inside `parse_length()` -> `get_unit()` (in libsvg/util.cc) when parsing numbers that have no unit.

t-paul:
Yep, that will be a little bit trickier than usual as the current tests do auto-scale, so they can't pick up simple scale changes. But we certainly need to extend the unit handling and add test cases...

kintel:
We can always render a manually sized object next to it for comparison

t-paul:
Right, and having the test case include 2 different units so relative changes are picked up too.

Issue #2703, opened 2019-01-16
Accept Customizer as a stable feature


kintel:
TODO:
* [ ] Do some manual testing of the current feature set
* [x] Investigate unnecessary JSON file creation, and suggest a resolution
* [x] If necessary: Resolve JSON file creation
* [x] Remove experimental flag

kintel:
@MichaelPFrey You may have a better idea what's absolutely needed for this to become stable, or is it already good enough?. We have two issue with a lot of suggestions: #1781 and #2316, that may need some cleanup

MichaelPFrey:
Good question. I currently see no blocking issue. The main benefit of the experimental status is, that the feature can quickly be disabled when it makes problem. I have most experimental features on all the time, so I assume that the customizer now plays pretty nicely with the other subsystems. I'll do some manual testing over the weekend

On thing I think I observed is, is that the customizer is creating unnecessary json files again. But I am not sure and even if: This is not a big deal.

kintel:
Thanks. I think investigating the JSON issue makes sense.  I've updated the TODO list.

t-paul:
I think the JSON file creation issue is fixed by https://github.com/openscad/openscad/pull/2776/commits/1aeb0a2b51532072a8a320e45be42ad4da7646d3.

PR #2776, opened 2019-02-10
Accept Customizer as a stable feature.

t-paul:
Changes needed for #2703.


MichaelPFrey:
Awesome and thank you for the debugging.

The failing test look simple:
```
The following tests FAILED:
	1112 - customizertest_description (Failed)
	1113 - customizertest_parameter (Failed)
	1116 - customizertest_allexpressionscomment (Failed)
	1117 - customizertest_group (Failed)
	1118 - customizertest-first_setofparameter (Failed)
	1119 - customizertest-wrong_setofparameter (Failed)
	1120 - customizertest-incomplete_setofparameter (Failed)
	1121 - customizertest-imgset_setofparameter (Failed)
	1122 - customizertest-setNameWithDot_setofparameter (Failed)
```
looking at the build artefacts,
```
-//Parameter("")
 x = 1;
-//Parameter(12)
 x1 = 1;
-//Parameter(12)
 x2 = 1;
-//Parameter(12)
 x3 = 1;
-//Parameter(12)
 x4 = 1;
-//Parameter([1 : 12])
 x5 = 1;
-//Parameter([1 : 2 : 12])
```
this is caused by `Don't output annotations when dumping ` I am kind of split on that change. Yes, outputting this information has no "real world usage". But I do remember that those test were very helpful when working on the comment parser. As you can guess, doing manual UI tests when working on a parser is painful and it is almost impossible to verify for other developers if the change breakes something unintentionally somewhere.

t-paul:
Heh, with the output enabled, the AST tests were failing. I guess we want to regenerate the Customizer tests then. There's probably not much point in having that output right now.

t-paul:
I guess it would make the customizer test pretty useless without that output, so I'll regenerate the AST tests instead.

WARNING: rebase incoming once it's green!

tests/regression/astdumptest/allexpressions-expected.ast
```diff
@@ -1,15 +1,24 @@
+//Parameter("")
```
kintel: Do we really want to add empty annotations to every assignment, and dump it here?

t-paul: That's what I thought first too and removed that output, but that failed all the customizer tests. I think those would be rendered pretty useless with that simple change.
E.g. https://github.com/openscad/openscad/blob/master/tests/regression/customizertest/parameter-expected.ast
So I reverted that and just regenerated that base AST tests. I guess we need a better way of testing the customizer parser so we can get rid of the sometimes useless and also non-conforming (I mean export is different from import) test cases.

kintel: perhaps suppressing output of empty parameter annotations, and adjusting test cases accordingly would suffice?

t-paul: I don't see a good way that is not adding horrible code just to optimize for some nicer test output. It might make more sense to see why we are generating empty annotations in the first place. But that's something I'd rather not touch at RC2 time.

kintel: That's a fair assessment; we can fix this later (or sooner if Michael deems it an innocent fix)

Issue #2401, opened 2018-05-24
Preferences CGAL & Polyset cache size limit [$15]


MichaelAtOz:
[Forum](http://forum.openscad.org/CPU-and-RAM-Resources-Allocations-Increase-for-OpenSCAD-tp24031p24080.html) found that the max input is limited to signed int max, 2^31 bytes.

Having looked at the code, this may be a hard limit for 32 bit support. But I'm no expert on the C internals.

cache.h has `void setMaxCost(int m) { mx = m; trim(mx);` making that size_t or unsigned int may get another 2GB?

There is also an anomaly src/Preferences.cc used toULong, src/mainwin.cc uses toUInt

64bit should presumably handle more?

Also while there I noticed [maxCost=100](https://github.com/openscad/openscad/blob/b9de27fbf1e3c73243c5453ad18dbe106e24b2b7/src/cache.h#L97), I'm wondering with larger cache sizes, whether that needs to be examined?

doug-moen:
`size_t` is unsigned, 32 bits on a 32 bit system and 64 bits on a 64 bit system. Just change the type to `size_t` everywhere.

MichaelAtOz:
[QIntValidator](http://doc.qt.io/qt-5/qintvalidator.html#details), [here](https://github.com/openscad/openscad/blob/7e22a7da586617a29c1e7a3082e43e5d6656a1b2/src/Preferences.cc#L197), appears to be the limiter.

kintel:
I agree with @doug-moen.
When fixing this, perhaps it's quick to look into presenting that number in a more human readable format in Preferences (e.g. using a `MB` postfix) ?

peter64:
Hey guys I posted a $15 bounty for this. I'de be entirely content with just having the QIntValidator removed (if that is all it takes to get the cache limit unleashed). If it requires more work to get >4GB on x64 systems I am happy to forgo that if it would be overly troublesome. 
 
The last nice to have would be if this can get integrated back into the source so that it is included in the multithreaded build or windows as that is what I use. https://github.com/openscad/openscad/pull/2399
Specifically so it ends up in the next build associated with 2399. The one I currently use is https://339-1049088-gh.circle-artifacts.com/0/64-bit/OpenSCAD-2018.11.04.ci339-x86-64_PR2399.zip

Also if $15 seems like too low a bounty let me know and I can up it slightly.
[...]

PR #2778, opened 2019-02-10
Set cache size in MB, allow >2GB for 64bit

thehans:
For Issue #2401


MichaelAtOz:
`// 1TB for 64bit or 2GB for 32bit` Thanks.
Now I have something to aim for...shame the Threadripper boards top out at 128GB...

kintel:
Since this effectively "forgets" any previously set value, should we fall back to the old key?
Not sure it worth too many brain cycles though

thehans:
OK, I added some code to grab defaults from old settings if found.

Issue #2564, opened 2018-11-10
get ctest -C All in an representable state


MichaelPFrey:
#2490 did not regenerate all tests.
some tests just need to be regenerated: ( #2565)
   * issue13
   * issue13b
   * issue13c
   * issue584
   * issue945
   * issue945b
   * issue945c
   * issue945d
   * issue945e
   * issue945f

Some tests need to be readjusted:
   * issue1580-back-to-back
   * issue1580-import-back-to-back

some may need to be readjusted
   * issue791
   * issue1876
(depending on what part of the geometry is critical)

may need manual generation/copying files around:
   * issue591

But I think we generally tend to overlook `ctest -C All`, but it would be good to go trough them before the next release.

kintel:
Closed by #2783

PR #2783, opened 2019-02-12
Updating "ctest -C All" expected results

thehans:
For #2564


thehans:
All the existing `ctest -C All` tests should be in order now, however it looks like there are a few scad files in `testdata/scad/bugs` that just aren't in the list and don't get run.  I haven't gone through them one by one, but its possible that some may no longer be bugs, and should be moved to another directory.  I'm thinking it might be nice if `tests/CMakeLists.txt` could automatically create a list from all scad files in the various directories like this without explicitly naming every one, so we don't miss any in the future.

Another nice to have, but maybe a bit complicated, would be if the test could report scad files from `bugs` that *don't* trigger any failures, so we can see more easily when they are fixed.

kintel:
There are some old files lying around in `bugs/`. The current strategy is to move files from bugs/ to issues/ as regression tests once the bugs are fixed. This wasn't done for earlier bugs.

MichaelPFrey:
First of all: Thank you for giving it a try.

Now to the diffs:

` tests/regression/throwntogethertest/example006-expected.png` when I load example006 in the UI, press F5 followed by F12, I see those green cubes.

` tests/regression/opencsgtest/example006-expected.png` when I load example006 in the UI, press F5, then the rounded edge is green as in the old file, not yellow as in the new file.

In this two cases, UI and console do not show the same thing.

That is one the hard part of regenerating the tests: It is hard to decode what is a bug that the test framework is supposed to catch or just an artefact.

The other hard parts are tests regarding bugs.
The other hard part is to check the perspective, as it can be hard to guess if the key geometry is still shown.

thehans:
I did edit example006.scad to be a little more efficient to render, that's why those particular results are different.  That one was a pet peeve of mine.  It now uses hull of spheres to make the rounded cube shape now.  Plus the sphere corners are a little cleaner now if you look closely, without overlapping rotated spheres.  I feel like its a better example, though now it doesn't really qualify as an "Old" example anymore.  (it uses `is_list` now too)

Also if you are loading example from Example menu, it comes from installed location, not the local folder so that could also be the issue if you are testing on this branch already.

MichaelPFrey:
You are right - I overlooked that not only the image has changed but also the code that is used to generate it. My bad.

thehans:
No problem. I actually only noticed that the Example menu pulls from system files after I went to save my edits and found that I didn't have permission ;-)

Here are the other cases which I edited which affect output(to better show geometry after the change in viewpoint of #2490 ), but should still fail in the relevant way:
`testdata/scad/bugs/issue1580-back-to-back.scad`
`testdata/scad/bugs/issue1580-back-to-back2.scad`
`testdata/scad/bugs/issue1876.scad`

Issue #2787, opened 2019-02-15
stack usage for module calls


MichaelPFrey:
Test case:
```
module test(x){
    if(x>0){
        test(x-1);
    }
}
for(a=[1:10]){
    z=a*1000;
    echo(z);
    test(a*1000);
    echo("ok");
}
```
On Ubuntu 18.10 build with current master:
```
ECHO: 1000
ECHO: "ok"
ECHO: 2000
TRACE: called by 'test', in file recdetect.scad, line 3.
```
with  OpenSCAD 2019.01.23 "nightly"
```
ECHO: 1000
ECHO: "ok"
ECHO: 2000
ECHO: "ok"
ECHO: 3000
ERROR: Recursion detected calling module 'test' in file michael, line 1
```
On Ubuntu 18.10 build on the 2019.01.17 (I have no idea what this build truly is, most likely it was the feature branch I was working on on that day)
```
Compiling design (CSG Tree generation)...
ECHO: 1000
ECHO: "ok"
ECHO: 2000
ECHO: "ok"
ECHO: 3000
ERROR: Recursion detected calling module 'test' in file recdetect.scad, line 1
```
Note that you can not compare windows to linux, as the build infrastructur and stackusage betwhen windows and linux are different. I also should not compare my custom build against a nigthly build as it uses a different build infrastructur and runtime libraries

This goes back to https://github.com/openscad/openscad/issues/1443#issuecomment-462171597

Testing around, I suspect #2645
https://github.com/openscad/openscad/blob/ac1d89f2381f8d4b6eaf0492c8ba1666d7d7f393/src/UserModule.cc#L43-L44
https://github.com/openscad/openscad/blob/ac1d89f2381f8d4b6eaf0492c8ba1666d7d7f393/src/ModuleInstantiation.cc#L86

commenting out those lines, seams to do the trick. Off-course I do not want to do that. Luckly, we have a well documented pattern to solve that kind of issue:
https://github.com/openscad/openscad/blob/ac1d89f2381f8d4b6eaf0492c8ba1666d7d7f393/src/context.cc#L182-L194
I will work on a more formal test and then a fix during the weekend.

MichaelPFrey:
There is/was discussion on putting the whole evaluation in a separate thread.
That should as a side effect significantly increase the amount of stack we have during this phase.

jamessynge:
What kind of thread does the evaluation happen on currently? A UI/event loop thread?

kintel:
[...]

PR #2788, opened 2019-02-15
optimize stack usage

MichaelPFrey:


MichaelPFrey:
currently, I am interested in the windows and OSX CI builds. `__attribute__ ((noinline))` does reduce the stack footprint in this case at the cost of runtime. Optimizing for stack usage is of higher importance than runtime during error handling.

Debug code for stack usage in module calls:
```
module test(x){
    if(x>0){
        test(x-1);
    }
}
for(a=[1000:100:10000]){
    test(a);
    echo(a, "ok");
}
```
Debug code for stack usage in function calls:
```
function add_up_to(n) = ( n==0 ? 0 : n + add_up_to(n-1) );

for(a=[5000:100:200000]){
    sum=add_up_to(a);
    echo(a, "ok");
}
```
this goes back to #2787, I assume that the scope will increase.

MichaelPFrey:
noinline for print_ignore_warning is preventive. It looks like currently the compiler considers it to be better to not inline methods that are used more then once, but that could change later, so adding noinline is both a reminder for the compiler and developers what not to do.

src/ModuleInstantiation.cc
```diff
@@ -70,6 +70,17 @@ void IfElseModuleInstantiation::print(std::ostream &stream, const std::string &i
...
 }
 
+/**
+ * This is separated because PRINTB uses quite a lot of stack space
+ * and the method using it evaluate()
+ * is called often when recursive modules are evaluated.
+ * noinline is required, as we here specifically optimize for stack usage
+ * during normal operating, not runtime during error handling.
+*/
+static void __attribute__ ((noinline)) print_trace(const ModuleInstantiation *mod, const Context *ctx){
```
kintel: We could probably use C++11 syntax here: `[[gnu::nolinline]]`, instead of the non-standard `__attribute__`.
No big deal though, as this is likely compiler-specific anyway.

src/UserModule.cc
```diff
@@ -37,11 +37,15 @@
 
 std::deque<std::string> UserModule::module_stack;
 
+static void __attribute__ ((noinline)) print_err(std::string name, const Location &loc,const Context *ctx){
+	std::string locs = loc.toRelativeString(ctx->documentPath());
```
kintel: If your goal is to reduce stack space, could just just pass this directly to PRINTB?

src/ModuleInstantiation.cc
MichaelPFrey: I worry about the gnu part in the alternate syntax. __attribute__ ((noinline)) is mentioned in [Attributes in Clang](https://clang.llvm.org/docs/AttributeReference.html). On the otherhand, while googling I found how chromium deals with it: [compiler_specific.h](https://chromium.googlesource.com/chromium/src/base/+/refs/heads/master/compiler_specific.h#79). On one hand, I do not like the idea of preprocessor magic, on the other hand I like the nice and clean `NOINLINE`, that can be replaced by anything the compiler likes.

I'd like to dwell on that.

src/UserModule.cc
MichaelPFrey: I do not worry about stack usage and runtime during error handling.
My concern is only stack usage during normal runtime.

Having PRINTB in a method that stacks (pun intended), lets the compiler reserve stack space, even if the execution path for the PRINTB is never used. Having the same PRINTB in a noinline methode means, that we only have to use the stack space when the error message is build.

Even if I would be concerned about stack usage when this method is called (during error handling): It is often surprising what the compiler can optimize, so I would look into that before premature code tweaking.
That is the reason for the noinline in the first place is the automatic optimization - the automatic optimization of inlineing a method that is used once makes both the program smaller and faster at runtime - at a small stack penalty, that for this very specific use case is relevant. gcc is really amazing.

src/ModuleInstantiation.cc
t-paul: I believe the `__attribute__ ((noinline))` is GNU too, so it would not change much. I'm not sure if we should worry too much at this point as there is no MSVC build in sight and the amount of work needed for that is certainly much bigger than "just" fixing a the attribute (Yeah, I know, famous last words, but if that makes a MSVC pull request happen tomorrow, I'm happy to fix this :grinning: ).

kintel: Yeah, to clarify: C++11 standardizes _how_ to specify attributes (the `[[ attr ]]` syntax), but `nolinline` is still a non-standard attribute.

MichaelPFrey: clang should support [[gnu::noinline]] https://clang.llvm.org/docs/AttributeReference.html#gnu-inline so let us see if our specific compiler version are able to understand it.

MichaelPFrey:
clang does not seem happy:
```
src/ModuleInstantiation.cc:80:15: error: 'noinline' attribute cannot be applied to types
static void [[gnu::noinline]] print_trace(const ModuleInstantiation *mod, const Context *ctx){
              ^
1 error generated.
make: *** [objects/src/ModuleInstantiation.o] Error 1
make: *** Waiting for unfinished jobs....
src/expr.cc:440:15: error: 'noinline' attribute cannot be applied to types
static void [[gnu::noinline]]  print_err(const char *name, const Location &loc,const Context *ctx){
              ^
src/expr.cc:452:15: error: 'noinline' attribute cannot be applied to types
static void [[gnu::noinline]]  print_trace(const FunctionCall *val, const Context *ctx){
              ^
``
I will look into that in the next few days, I think using the preprocessor to use compiler specific attributes makes the most sense in the long run.

src/ModuleInstantiation.cc
thehans: > I believe the `__attribute__ ((noinline))` is GNU too, so it would not change much. I'm not sure if we should worry too much at this point as there is no MSVC build in sight and the amount of work needed for that is certainly much bigger than "just" fixing a the attribute (Yeah, I know, famous last words, but if that makes a MSVC pull request happen tomorrow, I'm happy to fix this 😀 ).

@t-paul 
I think the MSVC build is pretty much ready to merge, just not sure what the proper fix is for this as I've commented it out for now:
https://github.com/openscad/openscad/pull/2824/files#diff-f1f0fb4f6529133f175aad3e8b5917eb

MichaelPFrey: @thehans I suggest `#include "compiler_specific.h"` in order to replace  `__attribute__ ((noinline)) ` with `NOINLINE`. That way, it defaults to defining `NOINLINE` to nothing, which is at least a workaround. (and a preparation when we find a proper way to noinline for MSVC or any other compiler)

https://github.com/openscad/openscad/blob/27ce666e9c9e3f80bd8f022cf94198a816b4901a/src/compiler_specific.h#L27-L33

PR #2789, opened 2019-02-16
Getting tests from MXE build to work on Windows

thehans:
Been using docker on my own machine to produce windows binaries etc, and wanted to get the `scripts/release-common.sh` tests option working once again.

All the default tests but 1 (non-ascii) are working for me now.


kintel:
Nice one; this must be a frustrating part of the codebase to work in...

PR #2790, opened 2019-02-16
Added debug code to StatCache.

nophead:
Dummy PR to get a build.

### DIFF
--- a/src/StatCache.cc
+++ b/src/StatCache.cc
@@ -60,8 +60,10 @@ int stat(const std::string &path, struct ::stat &st)
 			st = iter->second.st;                        // Not stale yet so return it
 			return 0;
 		}
-		statMap.erase(iter);                            // Remove stale entry
+		//statMap.erase(iter);                            // Remove stale entry
 	}
+	else
+		PRINTD(path);
 	if (auto rv = ::stat(path.c_str(), &st)) return rv; // stat failed
 	statMap[path] = {st, millis_clock()};
 	return 0;
--- a/src/StatCache.cc
+++ b/src/StatCache.cc
@@ -33,7 +33,7 @@
 
 namespace {
 
-const float stale = 0.190;  // Maximum lifetime of a cache entry chosen to be shorter than the automatic reload poll time
+const float stale = 190;  // Maximum lifetime of a cache entry chosen to be shorter than the automatic reload poll time
 
 double millis_clock(void)
 {
--- a/src/StatCache.cc
+++ b/src/StatCache.cc
@@ -29,15 +29,17 @@
 
 #include <string>
 #include <unordered_map>
-#include <chrono>
+#include <sys/timeb.h>
 
 namespace {
 
-const float stale = 190;  // Maximum lifetime of a cache entry chosen to be shorter than the automatic reload poll time
+const float stale = 0.190;  // Maximum lifetime of a cache entry chosen to be shorter than the automatic reload poll time
 
 double millis_clock(void)
 {
-	return std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
+	struct timeb tb;
+	ftime(&tb);
+	return tb.time + double(tb.millitm) / 1000;
 }
 
 struct CacheEntry
@@ -60,11 +62,13 @@ int stat(const std::string &path, struct ::stat &st)
 			st = iter->second.st;                        // Not stale yet so return it
 			return 0;
 		}
-		//statMap.erase(iter);                            // Remove stale entry
 	}
 	else
 		PRINTD(path);
-	if (auto rv = ::stat(path.c_str(), &st)) return rv; // stat failed
+	if (auto rv = ::stat(path.c_str(), &st)) {            // stat failed
+		if(iter != statMap.end())
+			statMap.erase(iter);                          // Remove stale entry
+    }
 	statMap[path] = {st, millis_clock()};
 	return 0;
 }   


thehans:
Hey did you discover anything from your testing?  I was just looking at StatCache and noticed that `stale = 0.190` line which seems wrong if that's in milliseconds.

I'm looking into it for the unicode related changes... from the microsoft docs I've read, it doesn't look like we can rely on stat to work on windows if we supply utf8 encoded path strings, unless mingw-w64 is doing some magic to handle it already.  I might need to change it to use boost filesystem commands for best compatibility.

nophead:
No I find the chrono stuff completely incomprehensible so I want back to my simple version but I still have the problem. It takes 5 seconds to check the dependencies but only on one project where I rearranged some includes and uses.

I have been trying to debug it but gdb doesn't work properly in an MSYS2 terminal. What did you use to profile the polyhedron problem?

thehans:
I used [Windows Performance Recorder/Analyzer](https://docs.microsoft.com/en-us/windows-hardware/test/wpt/windows-performance-analyzer) which has a feature to do CPU [flame graphs](https://docs.microsoft.com/en-us/windows-hardware/test/wpt/graphs)

Getting a debug build was tricky, I ended up creating a new branch / PR #2747 with some changes to generate debug symbols and not strip them.  I used the artifacts from CircleCI from that PR.

The last thing about it was that WPA expects .pdb debug files which typically come from visual studio's compiler.  I found this project though that can sort of generate a pdb file from gcc's dwarf debug symbols: https://github.com/rainers/cv2pdb

Run that on opescad.exe and it moves the symbols out of the file and into a separate .pdb in the same path.  It wasn't 100% reliable in that it would mix up some of the function names, so take the results with a grain of salt.  I think I was also lucky in that the native windows calls were taking much of the time, which WPA has no problem resolving regardless of pdb.

Also you have to explicitly tell WPA to load symbols once you have the recording opened:
https://docs.microsoft.com/en-us/windows-hardware/test/wpt/load-symbols-or-configure-symbol-paths

nophead:
Thanks I might have a go at that.

If you have optimised code then it would be no wonder symbols get mixed up. A lot of short C++ functions like accessors and mutators can be binary identical between different classes, so they can share the object code and have the same address for different symbols. For example an accessor that returns the first class member will be the same code for all classes if it is the same size member, say 32 bits.

nophead:
I can probably just use your PR because it will have the same problem as the current master I think,

nophead:
Hmm, seems to be Win10 only. I expected it so say so but it looked like it was preparing to install Windows 10 on my Win7 machine.

I have a tiny Win10 box. Not sure it is capable of running OpenSCAD though.

thehans:
There should be earlier versions of performance toolkit that are compatible with Windows 7, maybe this link works? http://www.microsoft.com/en-US/download/details.aspx?id=39982
The system requirements at the bottom nclude win 7

Also I've read that in older versions the profiler tool was called xperf, not sure if that is the most current for win 7.

nophead:
The older version of the performance analyser doesn't seem to have flame graphs. Does that mean it won't be useful?

Did you build cv2pdb from source? It says:

> The binary package of cv2pdb contains an executable cv2pdb.exe, which
> should be copied somewhere accessible through your PATH environment
> variable.

But it only seems to be source code. 

Interestingly your PR #2747 only takes 3 seconds to check the dependencies, whereas later versions are more like 5.

thehans:
There is a link on their releases page to an AppVeyor build.  If I remember correcltly, I just downloaded that and used it:
 https://github.com/rainers/cv2pdb/releases

I prefer flame graphs as a nice visualization of where CPU spends it time, but you should still be able to drill down in to the stack and see how much time is spent in various parts of code without one.

nophead:
Thanks.

Well you are quite right about 0.190 should now be 190. What I don't understand is how I thought I had fixed it four months ago and how when I tested the old version it still seemed slow. Basically I have wasted days of my time due to somebody deciding to refactor this code and not testing it. The old version was simple and easy to understand. chrono is both longer, less efficient and has fooled the person that changed it and myself as it is so complex.

Possibly 190us was long enough when my library was a bit smaller. If the dependency check completes in that time it would work. If it goes just over then it will slow down massively as a miss will precipitate more misses. 

I also didn't realise I need a make clean to switch between release and debug builds. I am used to the object codes being kept separate with the executables.

Similarly my command line code got refactored into 'modern code' without testing and ended up broken. Again I don't think the code is any shorter once you format it in a readable way.  Why is an inline functional version with join and lambda considered better than an old fashioned for loop? I don't think it is any shorter or more efficient.

Now I have to work out why, after a make clean, the Compiling design phase seems to have slowed by several orders of magnitude. Basically I changed stale and it fixed the slow auto reload and it compiled my project in about 5 minutes as normal but it has some debug logging so I did make clean and now it is taking so long I will have some lunch as it looks like it will be hours!

nophead:
It's a HeisenBug. With stale set to 190 it only works if I also have a PRINTB as well.
```
int stat(const std::string &path, struct ::stat &st)
{
	auto iter = statMap.find(path);
	if (iter != statMap.end()) {                      // Have we got an entry for this file?
		if (millis_clock() - iter->second.timestamp < stale) {
			st = iter->second.st;                        // Not stale yet so return it
			return 0;
		}
		statMap.erase(iter);                            // Remove stale entry
                PRINTB("stale: %f",(millis_clock() - iter->second.timestamp));
	}
	if (auto rv = ::stat(path.c_str(), &st)) return rv; // stat failed
	statMap[path] = {st, millis_clock()};
	return 0;
}
```

I.e. without the PRINTB it draws a frame very 5 seconds. With it one or two frames a second. And even if I revert it to using ftime it still behaves the same.

thehans:
Hmm, makes me wonder if its possible that there is some race condition coming from different threads?

nophead:
Yes I am beginning to think that if the time it takes to process the dependencies gets as long as the period it polls them at then it stops updating the window. 

I got rid of the cache timeout and just cleared it all before the dependency check. It didn't change the behaviour at all. With a small project you can pan the screen but it pauses a short while every time it checks the dependencies. With my pathological project it goes busy for 5 seconds at a time unless I have a PRINTB. 

There is already a flat list of dependencies so I might try checking the times on those instead of trawling though the massive tree of tens of thousands of checks on the same files that get satisfied by the cache.

t-paul:
`PRINTB` indirectly triggers the Qt event processing at the end of `MainWindow::consoleOutput()`.

nophead:
Do QT events have a priority. I.e. if a timer event takes longer to process that its repetition period would that stop other events being processed?

thehans:
https://doc.qt.io/qt-5/qtimer.html#accuracy-and-timer-resolution
> All timer types may time out later than expected if the system is busy or unable to provide the requested accuracy. In such a case of timeout overrun, Qt will emit timeout() only once, even if multiple timeouts have expired, and then will resume the original interval.

also this
> As a special case, a QTimer with a timeout of 0 will time out as soon as all the events in the window system's event queue have been processed
It sounds like it's a basic queue where events are processed as FIFO.

So as I understand it, other events could be still added to the queue while a timer event is processing.  And the next timeout that is emitted will just get added to the end of the queue.

nophead:
I found that with a single file project that includes all of my library handleDependencies() takes 0.25s. Not good as it is called every 0.2s, but it is still reasonably responsive.

With my larger project that has abut 12 files that each include just the subset of my library that they need then handleDependencies() takes 6.5s and it only responds every 6.5s. Adding the PRINTB causes it to process events every 190ms during handleDependencies(), which is why it gets responsive again.

I think when my larger project included all my library in every file it worked, so something weird is going on. Even so 0.25 seconds is a bit clunky. I think I can solve it by just checking the flat list of dependencies and only call handleDependencies() when something has changed. However I need to find out why it is so slow when there are only 78 used files and 118 dependencies in total.

I am struggling with cv2pdb at the moment as it doesn't find the dll is needs from my VS 2017 install.

t-paul:
Would it make sense to scale the delay by number of files? Something like `0.15 + ln(nr-of-files) / 8` maybe. That would increase the delay to about 0.6s at 100 files.

nophead:
The basic problem is handleDependencies is being called recursively many thousands of times checking the times on the same files over and over again. So even with the times cached it still takes a crazy amount of time.

I haven't quite got my head round why it is so many, it took 21 minutes just to print them all! The simple solution is just to mark the modules it has already checked as there is no point in checking them more than once.

One of the first C programs I wrote in the earl 80s was make program that checked dependencies only once by making a graph of them. I can apply the same logic it here and removed the cache since it will only look at the time on each file once.

thehans:
I don't really understand the file/module caching code.  Is it mainly for "Automatic reload and preview" purposes?  The timers related to that seem to be a little messy.  There is `autoReloadTimer` and another `waitAfterReloadTimer` that are being started and stopped on various seemingly intermingling conditions.  

Also the `GuiLocker` class seems like it could possibly create issues.  I think at the very least `GuiLocker::unlock()` needs a `gui_locked>0` check otherwise it could overflow to max unsigned int.  That and/or any code which calls `GuiLocker::unlock()` should also check `GuiLocker::isLocked()` first.  
I see before all the calls to `void MainWindow::compile(...)` that its being locked in the calling function,  *except* in `void MainWindow::waitAfterReload()`?  Its also hard to follow that each lock and unlock happen in different functions.  

I'm not sure what its intended purpose is exactly, to make sure we don't start more than one compile at a time?  If so, maybe a proper std::mutex or QMutex is in order.

nophead:
The module cache is there so used modules don't need to be parsed over and over again when they are included in multiple places. Each time a use is encountered it is looked up in the cache and if present nothing needs to be done, other than recording the use.

I think the lock is required for MDI mode. I.e. so multiple windows don't modify the data structures at the same time. Although my recollection of event driven stuff is it avoids a lot of problems with concurrency.

Anyway the problem is caused by the tree of dependencies exploding whereas as graph of them is quiet small. I.e. it only has has many nodes as there are files, in this case 118. So to check if any have changed  it should be just a matter of calling stat 118 times, The problem is each file gets checked many tens or perhaps 100s of thousands of time because the tree is fully explored instead of the graph. 

By refactoring my library code I managed to get the time down to about 1 second but I am sure most of that time is recusing and checking the stat cache over and over again. There is no point checking used modules that have already be checked.  And there is no point checking includes that have already been checked.

If I make a graph of the dependencies each file only needs to be checked once.

thehans:
Seems like it shouldn't necessarily need a special graph implementation, and could probably be handled by a std::set or std::unordered_map of already-processed files.  
Maybe rework how ModuleCache and StatCache interact so that StatCache is handled privately inside ModuleCache or something like that.

nophead:
By graph I don't mean a special data structure, it would be built with std::unordered_map(s).

Currently each module has a flat list of every file it includes and I think every module it uses including those used via its includes, That means if I include a library file in each file of my project and it includes a whole bunch of files like screws.scad that uses screw.scad which includes global.scad then each module ends up with a dependency on a large number of includes and uses, If it was graph instead of a flat list each project file  would depend on one library file. That would depend on all the others. 

Yes one could have a list of already processed modules to cut short the tree exploration. It would mean each module would be processed only once but even so it would be checking the same included files over and over again. It should really only need to check the directly included and used files. With the StatCache I don't think it it touching the file system more than once per file. It is basically using 6 seconds of CPU time recursing around the tree looking up values in an unordered_maps an enormous number of times. It isn't very efficient because it looks up the module cache entry lots of time and makes cache keys, etc, and is only indirectly recursive.

With a graph it would only need to check each file once and then attach the times to the graph such that each node has the latest time of itself and its children. It would replace the StatCache. Each module would ask if any of its directly included or used files had changed and they would indicate yes if they had changed or any of their dependencies had changed.

nophead:
Well a quick bodge solution was very simple indeed.  I removed this line: https://github.com/openscad/openscad/blob/master/src/FileModule.cc#L159. That has the effect of making the tree walk only visit each module once. I added a function to clear them all at the start by looping through the module cache.

Although it is checking the times on the includes over and over again the StatCache means it only takes less than 10ms with my large project. So turning on auto reload doesn't cause any noticeable slowdown at all. Not only that but the initial compilation and F5, etc are about 6 seconds faster. And the command line should also be faster by the same amount.

Although a graph solution would be even faster it isn't currently worth the effort. I will clean it up and make a PR. There is another issue in that the module paths are a mixture of / and \ depending how they are included, so I think it might get duplicate module cache entries in some cases. That should be easy to fix.

thehans:
Nice.  
For the mixed slashes, I think we just gotta make sure that `boost::filesystem::path::generic_string` ([ref](https://www.boost.org/doc/libs/1_55_0/libs/filesystem/doc/reference.html#path-generic-format-observers)) is called before anything goes into cache.

nophead:
Indeed, specifically here: https://github.com/openscad/openscad/blob/d4d3504975f9bd08c09e10db665de3fc83542747/src/lexer.l#L148

PR #2796, opened 2019-02-17
RFC: New extrusion mechanism 'extrude_for' allows loop-based extrusion

nrdvana:
This is a new feature I've long wished for in OpenSCAD.  It allows the user
to specify an arbitrary number of 2D shapes positioned in 3D space, and
extrudes from one to the next in sequence.  The shapes must be identical in
outline count and vertex count, and may not cross the plane of the
neighboring shape.  This prevents the most obvious deformities that might
result in an invalid solid, though many other edge cases exist that could
result in invalid geometry.

openscad-extrude

This adds two new Nodes; a for-loop derivative called ExtrudeForNode, and
a node ExtrusionShapeNode that marks the transition from 2D to 3D.

This also includes a questionable design decision of a new Geometry called
Polygon2dIn3d, with a getDimension() of 23.  This special geometry is
a 2D polygon set (Polygon2d) paired with a Transform3d.  As TransformNodes
process this geometry they alter the matrix-half rather than modifying the
2D coordinates.  This way the 2D vertices remain un-touched, making it
easy to construct an extrusion across them.


nrdvana:
After discussion with @kintel and teepee on irc, we agreed it might be better if the for() loop were separate from the extrusion, requiring the lazy-union feature from issue #350.  After some more thought, I think it should be possible (and reasonably doable without breaking back-compat) to make a series of changes that allow the "extrusion_shape" to be implied:

```
extrude() for(...) {
    translate(...) rotate(...) square(...)
}
```

So, the square generates a 2D primitive, the transforms upgrade it to "Polygon2dIn3d", the for loop returns an array of these, and the extrude does the same operation as I just implemented.  This would mean that you don't need to use a for() loop at all:

```
extrude() {
   translate([0,0,-5]) circle(1);
   translate([0,0,5]) circle(1);
}
```
would be able to create a cylinder since there was never an implied union of the 2D shapes, and the 3D transform was preserved.

To maintain compatibility, the explicit set operations like union() difference() etc would flatten Polygon2dIn3d back to 2d even if they had had 3D transforms performed on them.  The shapes would only gain 3D status after passing through an extrusion.

This would also allow future upgrades to hull() that people might like:

```
hull() {
   translate([0,0,-5]) cylinder();
   translate([0,0,5]) cylinder();
}
```
I don't currently know enough about the external geometry solvers to know whether they allow a 2D shape to be part of a hull computation, but if so I think it should drastically reduce the computation time vs. the current tactic of supplying wafer-thin 3D solids.

thehans:
As far as I understand, hull only deals with/cares about the set of points in the geometry; doesn't matter how they are connected.  So if it could be passed "2din3d" geometries, this would be half the input set of points compared to thin wafers.

src/Polygon2d.h
```diff
@@ -41,3 +41,18 @@ class Polygon2d : public Geometry
...
+{
+public:
+	Polygon2dIn3d(const Polygon2dIn3d &p): Polygon2d(p), matrix(p.matrix) {}
+	Polygon2dIn3d(const Polygon2d &p, const Transform3d &m): Polygon2d(p), matrix(m) {}
+	Polygon2dIn3d(const Polygon2d &p): Polygon2d(p), matrix(Transform3d::Identity()) {}
+
+	size_t memsize() const override;
+	std::string dump() const override;
+	Geometry *copy() const override { return new Polygon2d(*this); }
+	unsigned int getDimension() const override { return 23; }
```
kintel: This is quite the hack. If we decide to go down this route, we should rather change dimension to be a type-safe enum.

nrdvana: I agree :-)  actually my latest idea is to make a patch that does "lazy 2D transforms" and then combined with your patch that does "lazy unions", this extrusion patch would be a lot simpler.  Also I accidentally just force-pushed this branch and didn't mean to.

kintel: Force-pushing to your own feature branches is not a problem; I tend to do the same to be able to clean up commits.

thehans:
By the way, speaking of hull, this reminds me of a little user module I made a while ago that I call consecutive_hull, which can be used similarly as a generalized extrusion.

```
module consecutive_hull() {
  for (i = [1:$children-1])
    hull() {
      children(i-1);
      children(i);
    }
}
```

Something like that which could take 2din3d input, should be able to account for different vertex counts, and even plane crossing of neighboring geometries, although with a large downside of a massive union needed at the end.

nrdvana:
@thehans Neat!  I also didn't know about the children() op, but that is essentially what I've been doing in all my projects where I join wafer-thin hulls.  The one thing it can't do is handle concave shapes, and it is also an order of magnitude slower than just extruding, especially for fine-detail curved surfaces.

t-paul:
I think that not all of #350 is required, specifically not the part to make the change visible to the scripts and the exports.

Right now the geometry evaluation forces every node to return a single object which forces the implicit union after every operation. I believe what could help here is to change this so the nodes can decide if the union actually needs to be computed. E.g. in `scale() for() cube()` the `for()` would collect the child nodes without doing the union and the `scale()` should be able to pass along the same list with every child node scaled.

I guess that does not exactly turn it into simple change but it might still make it a bit more manageable.

nrdvana:
I'm about half way done with an improved implementation, but I need help understanding a bit of the code.  At [GeometryEvaluator.cc:607-612](https://github.com/openscad/openscad/blob/2a8d499c28bcb24b711f650eb8aca209011a1398/src/GeometryEvaluator.cc#L607-L612) I see it checking whether the Polygon2d is sanitized, and if so, running it through the Clipper sanitize routine.  Other similar code just reverses each Outline vector.  Is there a reason it needs to go back into the Clipper library?  Also, looks like a bug, because in [clipper-utils.cc](https://github.com/openscad/openscad/blob/2a8d499c28bcb24b711f650eb8aca209011a1398/src/clipper-utils.cc#L19-L26) it checks the value of Polygon2d->isSanitized() and uses that as the ``keep_orientation`` flag, but at the moment the polygon was passed in the winding is backward but isSanitized() is still true.

If all I need to do is reverse the outline vectors, that's more efficient anyway.

kintel:
@nrdvana If you know that reversing outlines gives the correct result, it would indeed be more efficient. 

It's possible there's a bug in there somewhere; if you find a way of triggering it that would be awesome.

nrdvana:
I don’t know it yet ;-)  which was why I was looking for hints about what else clipper might be accomplishing there.  I have about 20 test failures even when I run the un-altered master branch, so I need to figure that out still.  Hints about which other components use the “sanitized” feature would also be helpful.

nrdvana:
Ok, new version ready for review!

openscad-extrude-2

This is implemented in 3 patches: the first adds hidden 3d tracking to Polygon2d.  The second implements a feature-guarded "extrude()" builtin, and the third applies a small hack (guarded by same feature) to allow for() loop to pass its children to the parent node.

nrdvana:
(and now that I actually do intend to force-push, github breaks and won't show the update.  Please see https://github.com/nrdvana/openscad/commits/extrude_for )

gadgetmind:
Does this mean intersection_for could eventually be deprecated as intersection() for (...) would do the same thing?

t-paul:
Yes, `intersection_for()` is only needed because of the current way implicit unions work internally.

nrdvana:
In fact, if you give my patch a try you can do that right now :-)

    intersection() for(..., union=false)
    hull() for( ..., union=false)

nophead:
Nice, I hope this gets merged.

kintel:
We'll get to this after the release. In the meantime, if it's made experimental, that could significantly speed up the time to merge

nrdvana:
It's fully guarded by ``Feature::ExperimentalExtrude``, now.

nrdvana:
After using my feature some more, I decided it needed to be more convenient.  I added some detection for polygons getting unexpectedly modified, and automatic sequence-order detection for if the slices are given "backward".  I also enabled the ability for slices to share an edge, i.e. extrude around an axis.

One additional feature I'd like to add is for a conflict of intersecting slices to automatically put all the slices into "debug mode" as if they had been prefixed with '#'.  I haven't been able to figure out how that happens, or if it is possible to do just form the return value of GeometryEvaluator::visit or if I need to modify things elsewhere.  Does this seem like a reasonable feature?  Any advice on how to accomplish that?

To clarify, the following emits an error due to intersecting layers:
```
  extrude() {
    rotate(90, [0,1,0]) square([20,20]);
    translate([ 1, 0, -1 ]) rotate(45, [0,0,1])
      rotate(90, [0,1,0]) square([20,20]);
  }
```
and I would like it to show output of
```
  group() {
    #rotate(90, [0,1,0]) square([20,20]);
    #translate([ 1, 0, -1 ]) rotate(45, [0,0,1])
      rotate(90, [0,1,0]) square([20,20]);
  }
```
but I tried replacing the visit() code of ``extrude`` with the visit() code of ``group`` and didn't get any output.

nrdvana:
Here's another exiting new possibility: really easy screw threads!

```
module m_screw(len=10, pitch=.8, diam=5, $fn=20) {
	height=.866*pitch;
	poly=[ // x tracks "height" and y tracks "pitch"
		[diam/2-height, 0 ],
		// This draws a sawtooth with as many iterations needed to cover the 'len'
		for (t=[0:ceil(len/pitch)]) each([
			[ diam/2             , pitch*t + 0 ],
			[ diam/2             , pitch*t + pitch* 2/16 ],
			[ diam/2 - height*5/8, pitch*t + pitch* 7/16 ],
			[ diam/2 - height*5/8, pitch*t + pitch*11/16 ],
		]),
		[diam/2-height, pitch*ceil(len/pitch) + pitch*11/16 ]
	];
	intersection() {
		union() {
			// Extrude each half of the screw separately, and let "Union" work its magic to
			// assemble them.  Can't be done in a single 360* extrusion because it would end
			// with two faces facing eachother on same plane.
			extrude(convexity=10) for(t=[0:ceil($fn/2)], union=false)
				translate([0,0,(t/$fn-1)*pitch]) rotate(t*360/$fn, [0,0,1]) rotate(90, [1,0,0]) polygon(points=poly);
			extrude(convexity=10) for(t=[ceil($fn/2):$fn], union=false)
				translate([0,0,(t/$fn-1)*pitch]) rotate(t*360/$fn, [0,0,1]) rotate(90, [1,0,0]) polygon(points=poly);

			// That extrusion needed a hollow center, else there would be polygons at the axis
			// of the extrusion that "went sideways" (all points along same vector)
			// So, fill the center with a cylinder.
			cylinder(r=diam/2-height*3/4, h=len);
		}

		// Clip the above to the exact requested length
		cylinder(r=diam,h=len);
	}
}
m_screw();
```

image

SebiTimeWaster:
this awesome! when will it be merged?

ali1234:
I really like the concept.

I found a bug while testing the branch. Here is the code:

```
$fn = 16;

module helix() {
    extrude(convexity=10) for(a = [0:180], union=false) {
        rotate([0, 0, a*4]) translate([10, 0, a/18])
            rotate([90, 0, 0]) children();
    }
}

module test() {
    difference() {
        helix() children();
        cube(10);
    }
}

// works
translate([15, 0, 0]) test() circle(1);

// should be identical but does not work
translate([-15, 0, 0]) test() mirror([0, 1]) circle(1);
```

Basically feeding it mirrored polygons makes it do weird stuff if you then take the difference. Either one alone seems to work okay.

nophead:
That seems like a bug in in mirror if it reverses the winding order, although winding order doesn't seem to matter with 2D  polygons.

ali1234:
Turns out the difference isn't needed to trigger the bug. This previews okay but renders incorrectly:

```
module test() {
    extrude(convexity=10) {
        children();
        translate([0, 0, 10]) children();
    }
}

// works
translate([15, 0, 0]) test() circle(10);

// should be identical but does not work
translate([-15, 0, 0]) test() mirror([0, 1]) circle(10);
```

nrdvana:
Thanks for the review!  I'm still around and happy to work out the bugs, though it'll be a few days before I can get back to this.

nrdvana:
@ali1234 Ok, I think I have it fixed.

@kintel The code in Polygon2d is getting a little hairy, but all in the spirit of strictly preserving the previous assumptions/behavior.  It could get a lot simpler if it was safe to assume that any time any code transformed the polygon by a negative determinant matrix that the winding order should automatically swap, regardless of the sanitization flag.  (otherwise, how can you know which outlines are positive space vs. cutouts?)  If that new behavior sounds good to you, I could work on a patch which makes that change, and rebase these patches to apply on top of that.

zumpchke:
Why wasn't this merged?

scrameta:
This looks like a really great feature. The code structure changed quite a lot since 2019 so I've made a refreshed branch and manually updated/merged it (off latest main). 
https://github.com/scrameta/openscad/tree/extrude_refresh
I don't know the codebase at all so not sure if I've used the correct methods but the examples from this thread do at least work.

nrdvana:
> I've made a refreshed branch and manually updated/merged it

@scrameta Thanks!  I've just been using my own compiled OpenSCAD for the last 5 years (can't live without extrude), and it'll be nice to get caught up with the current feature set again.

tcurdt:
unstale - just because there is bot now does not change the fact why it was opened

kintel:
60 days is probably too quick to make PRs stale, will reconsider the config -> changed to 120 days now.

nrdvana:
I think #5584 is the successor of this work.  The interpolate feature is a significant upgrade to my version, allowing differing numbers of vertices per slice.  It has also been merged into pythonscad.  I'm going to close this one; anyone following it please switch to watching #5584.

Issue #2786, opened 2019-02-12
Issues with SpaceMouce (etc) support


MichaelAtOz:
I have a few issues.

First pedantics, [Axis v's Axes](https://grammarist.com/homophones/axis-vs-axes/), should the preferences tab be Axes and the label 'Axes Mapping' & button 'Auto Trim Axes'?

Actual bugs to follow.

MichaelAtOz:
Not sure of the terminology, so 'Axis n' describes the detection of movement of the control knob, as reflected by the coloured bars in Preferences. I'll use 'Channel' to describe the pairing of the X,Y,Z columns with the rows (Translation/Viewport-rel-Translation/Rotation etc), the result of which causes the preview to do something. Let me know if there is a more appropriate term.
So the Channel Y+Translation moves the objects in the preview up/down when Mapped to an Axis.

There is a problem with at least one Channel, I have yet to try all Mappings, which I'll get to.

Y+Viewport-rel-translation (VRT) Channel is not responding to input.
When mapped to Axis-1, with Preferences open beside the Preview, and the Main/Preview Window active, I can see the coloured bar for Axis-1 change, but there is no change in the Preview.  
I have X+VRT mapped to Axis-0, which works, moving the view left/right.

If I reverse this, X-VRT=Axis-1, Y-VRT=Axis-0, then up/down on the control-knob moves left/right, but left/right on the control-knob does not move.

There is also seems to be a difference in responsiveness on some axes, I'll have to try to quantify that...

MichaelAtOz:
This is with HIDAPI on Windows 7/64 BTW, with 2019.01.05.ci1524.

MichaelPFrey:
> Y+Viewport-rel-translation (VRT) Channel is not responding to input.

You are in orthogonal view. Please change it to perspective to see what it does. Or look in the bottom left corner, where translate = changes. This is not a bug, this is a very specific feature. When you map zoom to one axis and Y+Viewport-rel-translation to an other while in perspective view, you should get the vertigo effect.
Most users will therefor use zoom as it works in both orthogonal and perspective. Before you wonder why you can map two axis to zoom: Game controllers have two shoulder buttons.

Yes, this has little to no real world use, but the inputdriver is about giving as much control to the user as possible.

MichaelPFrey:
Maybe I also need to add, that openscad internally considers Z tobe up and down.
X is left/right. Y is "into and outof the screen".

MichaelAtOz:
[...]

PR #2797, opened 2019-02-17
Axis -> Axes (grammer, translation)

MichaelPFrey:
going back to #2786

### DIFF
--- a/src/Preferences.ui
+++ b/src/Preferences.ui
@@ -1976,10 +1976,10 @@
      <normaloff>:/icons/prefsInputAxis.png</normaloff>:/icons/prefsInputAxis.png</iconset>
    </property>
    <property name="text">
-    <string>Axis</string>
+    <string>Axes</string>
    </property>
    <property name="toolTip">
-    <string>Input driver configuration and Axis mapping</string>
+    <string>Input driver configuration and Axes mapping</string>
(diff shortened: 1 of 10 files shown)


src/Preferences.ui
```diff
@@ -1976,10 +1976,10 @@
      <normaloff>:/icons/prefsInputAxis.png</normaloff>:/icons/prefsInputAxis.png</iconset>
    </property>
    <property name="text">
-    <string>Axis</string>
+    <string>Axes</string>
    </property>
    <property name="toolTip">
-    <string>Input driver configuration and Axis mapping</string>
+    <string>Input driver configuration and Axes mapping</string>
```
kintel: I don't claim to be an English native speaker, but I believe on usually does not use plural in English in cases like "axis mapping" and "axis setup". The others fixes sound good though.

MichaelPFrey: I think you are right, but I am unsure as well.
If it would be button, it would clearly be button mapping and button setup.
Axis/Axes are confusing words...

nophead:
> Axis/Axes are confusing words...

Yes axes is the plural of both axis and axe. But when talking about a
mapping the singular form is more correct. I.e you could talk about
character mapping, not characters mapping.

Issue #596, opened 2014-01-08
Passing children to modules that don't accept them doesn't issue warning


kwikius:
```
module Ring(outerD,innerD,thickness){
 epsilon = 0.1;
    difference(){
        cylinder(h = thickness, d=outerD , $fn= 20)  //;
// NO STATEMENT TERMINATOR ";" in above line      ^^^  BUT DOESNT FAIL COMPILATION
// No warning
// next statement is  ignored...
        translate([0,0,-epsilon]){
          cylinder(h = thickness + 2 \* epsilon, d=innerD, $fn= 20);
       }
   }
}

Ring(100,80,10);

/*
regards Andy

OpenSCAD is great BTW :)
*/

```

##

PR #2806, opened 2019-02-21
warning, when a builtin module without support for it has children

### LINKED ISSUE #2804: Syntax error not detected, silently ignoring some primitives
_OpenSCAD version 2019.01-RC2 (git 35858ea)_ but the error exists at least since OpenSCAD version 2015.03-2  (running linux)
Forgetting the semicolon after a primitive before another primitive produces no syntax error. Only the first primitive is generated and shown. I would expect the parser to fail as it makes debugging the typo much harder otherwise.
```
cube()        // misses the syntax error, only primitive shown
cylinder();   // does not show up
```
Here is a second example:
```
sphere()     // misses the syntax error, only primitive shown
cylinder()   // does not show up
cube();      // does not show up
```

MichaelPFrey:
fixes #2804, partially adresses #596 

Todo:
- add test case
- update changelog
- ...

In scope of this PR:
- builtin modules

Not in scope of this PR:
- user modules

bildschirmfoto von 2019-02-21 19-32-56
bildschirmfoto von 2019-02-21 19-33-45
bildschirmfoto von 2019-02-21 19-34-12

### DIFF
--- a/src/primitives.cc
+++ b/src/primitives.cc
@@ -148,7 +148,7 @@ AbstractNode *PrimitiveModule::instantiate(const Context *ctx, const ModuleInsta
 	AssignmentList args;
 	AssignmentList optargs;
 	if(inst->scope.hasChildren()){
-		PRINTB("WARNING: module does not support child modules, %s", inst->location().toRelativeString(ctx->documentPath()));
+		PRINTB("WARNING: module %s() does not support child modules, %s", node->name() % inst->location().toRelativeString(ctx->documentPath()));
 	}
 
 	switch (this->type) {
--- a/src/primitives.cc
+++ b/src/primitives.cc
@@ -148,7 +148,7 @@ AbstractNode *PrimitiveModule::instantiate(const Context *ctx, const ModuleInsta
 	AssignmentList args;
 	AssignmentList optargs;
 	if(inst->scope.hasChildren()){
-		PRINTB("WARNING: modul does not support child modules, %s", inst->location().toRelativeString(ctx->documentPath()));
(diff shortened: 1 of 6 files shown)


src/primitives.cc
```diff
@@ -147,6 +147,9 @@ AbstractNode *PrimitiveModule::instantiate(const Context *ctx, const ModuleInsta
 
 	AssignmentList args;
 	AssignmentList optargs;
+	if(inst->scope.hasChildren()){
+		PRINTB("WARNING: modul does not support child modules, %s", inst->location().toRelativeString(ctx->documentPath()));
```
thehans: Typo `modul` vs `module`

MichaelPFrey: thank you, will be fixed with the next push

MichaelPFrey:
it is surprisingly hard to get the module name. I have a primitive_type_e. We have a function that can turn primitive_type_e into a string, but capsuled in a different class:
https://github.com/openscad/openscad/blob/a30dbed0e18398d86821edd8cc3b8a46c6633a30/src/primitives.cc#L72
Offcourse code can be moved and refactored, but that can have implications.

I will think about it.

(The main reason why this patch works and is so small, that we solved the linenumber issue - "back in the day" outputting the module name would have been required and still be less useful then line number and filename)

t-paul:
Isn't that just `node->name()` at this point?

MichaelPFrey:
> Isn't that just `node->name()` at this point?

Silly me. I was looking for a member variable, not a variable declared on the first line of the method:

https://github.com/openscad/openscad/blob/57ed520b33affb4869b36eb646fea241440cd323/src/primitives.cc#L141-L143

Issue #2731, opened 2019-01-27
Auto reload does not work after stop on first warning


nophead:
For example, with this file:
```
echo(version());
circle(r = 5, d = 2);
```
I get
```
ECHO: [2019, 1, 11]
WARNING: Ignoring radius variable 'r' as diameter 'd' is defined too, in file warning.scad, line 3
Execution aborted
```

If I then remove d = 2 in an external editor it does not reload. If I manually reload then it will clear the warning and auto reload works again until I get another one.

On the other hand if I have an error, rather than a warning, then it works as normal.

nophead:
I think "Execution aborted" should have a blank line after it, like "Compilation failed!", and perhaps be "Compilation aborted!" for consistency as well. Otherwise it is hard to see where the new compilation starts. 

Parsing design (AST generation)...
Saved backup file: C:/Users/ChrisP/Documents/OpenSCAD/backups/unsaved-backup-JRJgEDZY.scad
ERROR: Parser error in file "C:\Program Files\OpenSCAD", line 13: syntax error 
ERROR: Compilation failed!

Parsing design (AST generation)...
Saved backup file: C:/Users/ChrisP/Documents/OpenSCAD/backups/unsaved-backup-JRJgEDZY.scad
Compiling design (CSG Tree generation)...
WARNING: Ignoring unknown variable 'v', in file ./, line 13.
Execution aborted  **<--- here**
Parsing design (AST generation)...
Saved backup file: C:/Users/ChrisP/Documents/OpenSCAD/backups/unsaved-backup-JRJgEDZY.scad
Compiling design (CSG Tree generation)...

nophead:
I think the same issue happens with compilation errors:
```
WARNING: Can't open library 'screw.scad'.
ERROR: Parser error in file "screws.scad", line 120: stop on first warning 
ERROR: Compilation failed!
```

No auto reload

PR #2812, opened 2019-02-24
Speed up handle dependencies

nophead:
Massive speed up in handleDepedencies by only checking each module once.
Fixed StatCache timeout being 1000 times too small since using chrono.
Made module cache paths generic to avoid mixed slashes on Windows.
Stopped double compilation when file is first loaded by initialising the includes time.
Made console buffer 10 times bigger.

### DIFF
--- a/src/StatCache.cc
+++ b/src/StatCache.cc
@@ -33,7 +33,7 @@
 
 namespace {
 
-const float stale = 0.190;  // Maximum lifetime of a cache entry chosen to be shorter than the automatic reload poll time
+const float stale = 190;  // Maximum lifetime of a cache entry chosen to be shorter than the automatic reload poll time
 
 double millis_clock(void)
 {
--- a/src/StatCache.cc
+++ b/src/StatCache.cc
@@ -33,7 +33,7 @@
 
 namespace {
 
-const float stale = 190;  // Maximum lifetime of a cache entry chosen to be shorter than the automatic reload poll time
+const float stale = 190;  // 190ms, maximum lifetime of a cache entry chosen to be shorter than the automatic reload poll time
 
 double millis_clock(void)
 {
--- a/src/FileModule.cc
+++ b/src/FileModule.cc
@@ -120,8 +120,9 @@ time_t FileModule::handleDependencies()
 			wasmissing = true;
 			auto fullpath = find_valid_path(this->path, filename);
 			if (!fullpath.empty()) {
-				updates.emplace_back(filename, fullpath.generic_string());
-				filename = fullpath.generic_string();
+				auto newfilename = fullpath.generic_string();
+				updates.emplace_back(filename, newfilename);
+				filename = newfilename;
 			}
 			else {
 				found = false;
@@ -129,7 +130,6 @@ time_t FileModule::handleDependencies()
 		}
 
 		if (found) {
-			auto wascached = ModuleCache::instance()->isCached(filename);
 			auto oldmodule = ModuleCache::instance()->lookup(filename);
 			FileModule *newmodule;
 			auto mtime = ModuleCache::instance()->evaluate(this->getFullpath(),filename, newmodule);
@@ -144,14 +144,13 @@ time_t FileModule::handleDependencies()
 				PRINTDB("  %s: %p", filename % oldmodule);
 			}
 			// Only print warning if we're not part of an automatic reload
-			if (!newmodule && !wascached && !wasmissing) {
+			if (!newmodule && !oldmodule && !wasmissing) {
 				PRINTB_NOCACHE("WARNING: Failed to compile library '%s'.", filename);
 			}
 		}
 	}
 
 	// Relative filenames which were located are reinserted as absolute filenames
-	typedef std::pair<std::string,std::string> stringpair;
 	for (const auto &files : updates) {
 		this->usedlibs.erase(files.first);
 		this->usedlibs.insert(files.second);
--- a/src/FileModule.cc
+++ b/src/FileModule.cc
@@ -99,9 +99,12 @@ time_t FileModule::include_modified(const IncludeFile &inc) const
 	Check if any dependencies have been modified and recompile them.
 	Returns true if anything was recompiled.
 */
-time_t FileModule::handleDependencies()
+time_t FileModule::handleDependencies(bool is_root)
 {
-	if (this->is_handling_dependencies) return 0;
+	if(is_root)
+		ModuleCache::clear_markers();
+	else
+		if (this->is_handling_dependencies) return 0;
 	this->is_handling_dependencies = true;
 
 	std::vector<std::pair<std::string,std::string>> updates;
@@ -155,7 +158,6 @@ time_t FileModule::handleDependencies()
 		this->usedlibs.erase(files.first);
 		this->usedlibs.insert(files.second);
 	}
-	this->is_handling_dependencies = false;
 	return latest;
 }
 
--- a/src/FileModule.cc
+++ b/src/FileModule.cc
@@ -115,12 +115,10 @@ time_t FileModule::handleDependencies(bool is_root)
 	time_t latest = 0;
 	for (auto filename : this->usedlibs) {
 
-		auto wasmissing = false;
 		auto found = true;
 
 		// Get an absolute filename for the module
 		if (!fs::path(filename).is_absolute()) {
-			wasmissing = true;
 			auto fullpath = find_valid_path(this->path, filename);
 			if (!fullpath.empty()) {
 				auto newfilename = fullpath.generic_string();
@@ -145,10 +143,6 @@ time_t FileModule::handleDependencies(bool is_root)
 			}
 			else {
 				PRINTDB("  %s: %p", filename % oldmodule);
-			}
-			// Only print warning if we're not part of an automatic reload
-			if (!newmodule && !oldmodule && !wasmissing) {
-				PRINTB_NOCACHE("WARNING: Failed to compile library '%s'.", filename);
 			}
 		}
 	}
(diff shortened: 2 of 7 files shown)


src/StatCache.cc
```diff
@@ -33,7 +33,7 @@
 
 namespace {
 
-const float stale = 0.190;  // Maximum lifetime of a cache entry chosen to be shorter than the automatic reload poll time
+const float stale = 190;  // Maximum lifetime of a cache entry chosen to be shorter than the automatic reload poll time
```
kintel: Perhaps mention "milliseconds" here, so that when we eventually move to C++14 we can use the typesafe time operator and just write `constexpr auto stale = 190ms;`,

kintel:
Looks good!.
Since you've technically changed some behavior, did you test for some of the less common corner cases?
I once wrote up a manual test procedure here; admittedly tedious to go through but I've found bugs that way before:
https://github.com/openscad/openscad/blob/master/testdata/modulecache-tests/README.txt

nophead:
No I tested it on my large library putting in syntax errors, removing files etc.

Some of those test are already broken but not by me see: https://github.com/openscad/openscad/issues/2731

In general the new error handling seems to be built on shaky ground because PRINT now can throw an exception but the code that uses it didn't expect that at the time it was written. How do we know objects are left in a consistent state?

I am a bit unsure about the warning I removed because the comment says it wasn't to be given during auto reload but it was. I had a file that was used by several others in my library. I put a syntax error in at and I got a nice error message from the parser and a trace but after that I got 5 warnings from auto reload. Each one said it was aborting but it didn't. Then I got a second burst of 5 and then it stopped. I am not sure if the original code didn't anticipate the same file used by multiple paths, or if it wasn't expecting to get interrupted. Without it if you press F5 without fixing the syntax error you just see missing functions and modules. No longer any mention of the syntax error or the the fact the module is missing.

kintel:
ok, I'll open a ticket to do more testing of module caching before the release

kintel:
#2813

nophead:
I am working through the tests at the moment. This is what I have so far.

Test1: pass
Test2: module address does not change but nevertheless it picks up changes.
Test3: pass
There is no Test4
Test5: pass
Test6: pass
Test7: pass
Test8: pass
Test9: pass
Test10: pass
```
WARNING: Can't open include file 'circularincludefirst.scad'.
ERROR: Parser error in file "D:/msys64\home\ChrisP\openscad\testdata\modulecache-tests\circularincludesecond.scad", line 1: stop on first warning 
ERROR: Compilation failed!```
Test11:pass except the error marker in the editor doesn't go away unless F5 is pressed.

**Comment by nophead** [comment]
Test12: pass same comment as Test11
Test13: F4 does not make the sphere appear but F5 does. F4 doesn't seem to trigger a compile unless the main file has changed.

**Comment by nophead** [comment]
Test13 fails because F4 is coded to do nothing if the top level file hasn't changed. Presumable that was a change since the tests were written. On the other hand, if it always did a compile then the sphere would disappear because it would get an error.

**Comment by nophead** [comment]
Test14: fails. 

When cascade2.sh is run:
```
Loaded design 'D:/msys64/home/ChrisP/openscad/testdata/modulecache-tests/cascadetest.scad'.
WARNING: Can't open include file 'cascade-B.scad'.
ERROR: Parser error in file "D:/msys64/home/ChrisP/openscad/testdata/modulecache-tests/cascadetest.scad", line 1: stop on first warning 
ERROR: Compilation failed!

WARNING: Can't open library 'cascade-D.scad'.
ERROR: Parser error in file "D:/msys64/home/ChrisP/openscad/testdata/modulecache-tests/cascadetest.scad", line 1: stop on first warning 
ERROR: Compilation failed!
```

**Comment by nophead** [comment]
Test15: pass apart from the WARNING that I removed.
Test16: pass
Test17: Opening the second and third files did not auto compile, had to use F5. After that the rest worked.
Test18: pass
Test19: pass. Messages a little different.
```
ERROR: Parser error in file "D:/msys64\home\ChrisP\openscad\testdata\modulecache-tests\error.scad", line 3: syntax error 
Used file cache size: 2 files
WARNING: Ignoring unknown module 'errmod', in file mainuseincludingerror.scad, line 2.
TRACE: called by 'errmod', in file mainuseincludingerror.scad, line 2.
```

**Comment by nophead** [comment]
Yes looks like a few regressions but not all due to this change. I will check the fails on a version before my change.

**Comment by kintel** [comment]
Thanks, that's very helpful. Regression do sneak in easily. We should find a way of testing this automatically, but it's a bit tricky to pull off

**Comment by kintel** [comment]
Btw., feel free to merge this when you feel ready. Since there are already regressions, we can manage all those separately from this PR

**Comment by nophead** [comment]
Some of these issues, like F4 appearing to do nothing seem to be that the console output is not going to the right place. If I quit the program then I see the missing file warnings in the shell console.

Another, perhaps related, issue I have noticed is if there is output streaming to the console when I quit then I get a segfault on exit.

**Comment by nophead** [comment]
What is the purpose of clearCurrentOutput() ? I get that each window needs to set its console before printing else it can go the wrong window in MDI mode but why clear it so it goes nowhere visible?

**Comment by nophead** [comment]
Test13:

Making clearCurrentOutput() do nothing means I get WARNING: Can't open library 'missing.scad' when I open the file before I press anything and with auto reload off, which is a bit unexpected. 

Pressing F4 does nothing but pressing F5 gives.

```
WARNING: Can't open library 'missing.scad'.
ERROR: Parser error in file "D:/msys64/home/ChrisP/openscad/testdata/modulecache-tests/usemissing.scad", line 1: stop on first warning 
ERROR: Compilation failed!
```
When I create the missing file then if I haven't pressed F5 then it does draw the sphere when I press F4. F5 is what breaks things because it kills auto reload and it seems it kills F4 as well. Not too unexpected because F4 is almost the same as manually running auto reload.

kintel:
clearCurrentOutput() is a hack to try to ensure that the correct MainWindow instance gets the console output for messages related to the open document. I don't think anyone has attempted to refactor that yet as there is a bunch of tricky pieces of control flow in MainWindow, so it's easy to break stuff.

..but the F4 stuff does sound like a relatively harmless regression.

The crash sounds like #1251

nophead:
The cascade test failure is important to me as I think it was instigated for my workflow. Looking at the current code I don't understand how it ever worked and Github is lousy at showing file history.

The sprinkling of setCurrentOutput() and clearCurrentOutput)_ all over the place seems ugly and error prone. Clearing it means any missing set sends the message down a black hole. If they are all removed apart from when the window is closed the worst that will happen is it goes to the wrong window in MDI mode. I have actually seen that. Perhaps a slightly neater solution is to instantiate an object at the start of any handler that prints. It constructor does setCurrentOutput() and its destructor does clearCurrentOutput. That at least ensures they are always matched and halves the number of lines.

As I understand it the printing is done from event handlers of the current MainWindow instance and the output needs to go to the console belonging to that window.

Is is possible to get a pointer to that in the print function? Perhaps with QApplication::activeWindow() and then call parentWidget() until the result can be dynamically casted to MainWindow?

Or perhaps post a message to the active window that gets handled my MainWindow?

t-paul:
On Linux it's going to command line, but it's still pretty useless. We do track the active window already for the input driver stuff, so it might be possible to re-use that.

nophead:
Hmm, probably not because it isn't always the active one in terms of focus that we want. It is the one currently processing a timer in some cases. Perhaps QApplication::activeWindow() isn't that.

t-paul:
Right, actually we "just" want it tied to the window owning the script. For everything processing related, it probably could be handled via the root module.

nophead:
This seems to be the breaking change: https://github.com/openscad/openscad/blame/master/src/mainwin.cc#L895

It compiles the top level document when it is opened but it is after clearCurrentOutput(), so the output goes to the console. Previously I don't think it does anything when you load a document until you do F4, F5 or have auto reload and compile turned on.

Presumably it was for the customiser. Is it deliberately hiding the output?

I think the issue it causes is since it has compiled the top level already F4 doesn't do anything but as the output is hidden you never see the error.

I made a version with a scope object that ensures clearCurrentOutput is called at the end and then the output is visible and it makes more sense because you see that there is an error. It then makes sense that F4 doesn't do anything until you change the file.

Is it correct that the top level document is now parsed when it is opened. Should it only be if the customiser is enabled?

t-paul:
Hmm, yes, it's possible that it's trying to hide the output as it's just collecting the variables.

nophead:
As an experiment I tried removing all the setCurrentOutput and clearCurrentOuput calls. Instead I put a single call the setCurrentOutput in MainWindow::event() and a call to clearCurrentOutput in the destructor.

It nearly worked but if I do things in two windows at the same time some output goes missing. Very odd as there it should always be set to one window or the other or if it was null the command console. I don't know how it can completely vanish.

It got me thinking about threads. If CGAL is run on another thread and it outputs messages isn't its output going to get misdirected in MDI mode if another window calls setCurrentOutput() and clearCurrentOutput()? Seems like it needs to be a thread local variable.

nophead:
Seems all you have to do with the latest master to get output going to the console is to start a slow running preview in one window and then open another file. The new window prints the usual stuff and then the when the original window finishes the preview its stats go to the command line because the second window has called clearCurrentOutput().

I haven't really noticed this before because I normally let one render complete before opening another window as the GUI is painfully slow. I think it only processes events to update the progress bar.

nophead:
I think the simplest solution to getting the output to the correct destination is to add a parameter to PRINT and its cousins. It would be a pointer to a Functor object or NULL. This parameter would need to be passed all the way from the MainWindow event handlers to the functions that print. A bit cumbersome but it would be guaranteed to work instead of the current system of setCurrentOutput and clearCurrentOutput that doesn't work.

PR #2815, opened 2019-02-25
Inputdriver ui improvement

MichaelPFrey:
- bar changes color when above deadzone value
- output status information in the axis config dialog
- implementation details improved
- making better use of the status information provided by qgamepad

### DIFF
--- a/src/input/AxisConfigWidget.cc
+++ b/src/input/AxisConfigWidget.cc
@@ -49,7 +49,7 @@ void AxisConfigWidget::AxesChanged(int nr, double val) const{
 	int value = val * 100;
 	progressBar->setValue(value); //set where the bar is
 
-	//QProgressBar generates the shown string form the format string.
+	//QProgressBar generates the shown string from the format string.
 	//By setting a format string without a place holder,
 	//we can set arbitrary text, like a custom formated double.
 	//(Note: QProgressBar internally works on int, so has no formating for double values)
@@ -158,7 +158,7 @@ void AxisConfigWidget::init() {
 	initDoubleSpinBox(this->doubleSpinBoxZoomGain, Settings::Settings::inputZoomGain);
 
 	//use a custom style for the axis indicators,
-	//to prevent getting operating specific
+	//to prevent getting operating system specific
 	//(potentially animated) ProgressBars
 	int textLightness = this->progressBarAxis0->palette().text().color().lightness();
 	this->darkModeDetected = textLightness>165;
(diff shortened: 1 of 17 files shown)


MichaelPFrey:
bildschirmfoto von 2019-02-25 06-39-12

src/input/AxisConfigWidget.cc
```diff
@@ -51,11 +51,24 @@ void AxisConfigWidget::AxesChanged(int nr, double val) const{
 	if(progressBar==nullptr) return;
 	progressBar->setValue(value);
 	progressBar->setFormat(s);
+	
+	QDoubleSpinBox* deadzone = this->findChild<QDoubleSpinBox *>(QString::fromStdString("doubleSpinBoxDeadzone"+number));
```
kintel: Tiny suggestion: You can concatenate numbers directly to a QString: `QString("doubleSpinBoxDeadzone") + nr`, without going through a std::string.

src/input/AxisConfigWidget.cc
```diff
@@ -51,11 +51,24 @@ void AxisConfigWidget::AxesChanged(int nr, double val) const{
 	if(progressBar==nullptr) return;
 	progressBar->setValue(value);
 	progressBar->setFormat(s);
+
```
kintel: I didn't catch this earlier, but `setFormat()` accepts a format string, not a number. Not sure if your way triggers some undocumented compatibility mode perhaps?

src/input/DBusInputDriver.cc
```diff
@@ -150,3 +150,10 @@ const std::string & DBusInputDriver::get_name() const
 {
     return name;
 }
+
+std::string DBusInputDriver::get_info() const
+{
+	std::ostringstream stream;
+	stream << get_name() << " " << (isOpen() ? "open" : "not open");
+	return stream.str();
```
kintel: You can replace this by `return STR(get_name() << " " << (isOpen() ? "open" : "not open"))`

-> We have a STR macro for this common operation: https://github.com/openscad/openscad/blob/master/src/printutils.h#L100

src/input/AxisConfigWidget.cc
MichaelPFrey: that makes things simpler - thank you

src/input/AxisConfigWidget.cc
MichaelPFrey: Hmm... that needs documentation from my side.

QProgressbar does not have a way of setting a text directly.
All I can do is setting a format, that is used to create the text.
By setting a format that does not contain a placeholder for the value, only the format string is shown.
By using a format string that is the way I want the string to be formatted, I get the result I want.

The alternative is, to overlay the progressbar with a label. Sounds messy.

Alternatively, I could look at QSlider or QScrollbar, but they do not have any text. (where I would like to have it)

Or customize the QProgressbar.

MichaelPFrey:
axiswindowsspaceexplorer

MichaelPFrey:
This good for review for now, otherwise I will increase the scope more and more.

kintel:
Ok, will review. Limited scope is good :)

PR #2816, opened 2019-02-25
Fix missing warning about double assignment

knielsen:
Ensure that mainFilePath is an absolute path during parsing.

The logic for generating warnings about double assignment generally
seems to assume relative paths, and in fact mainFilePath is made
absolute in the recursive parse() call for a use <...> statement. So
if mainFilePath is not absolute from the start the code gets confused.


kintel:
Can anyone (e.g. @MichaelPFrey) think of a way of getting a test for relative main files added to our test suite, so we don't refactor this bug back into existence in the future?

knielsen:
Agree a test would be good. The Debian packaging has only a few smallish patches to be able to run tests out-of-tree (and triggering this issue), maybe that can be unified with upstream at some point. Though right now I'm focusing primarily of getting the packaging in a good shape for the Debian release freeze.

PR #2820, opened 2019-02-25
HIDAPI: Unify decoder

t-paul:
I left the framework for having different functions in, it's not really a big thing and might come in useful later.

### DIFF
--- a/src/ext/hidapi/hid.c
+++ b/src/ext/hidapi/hid.c
@@ -36,6 +36,10 @@ typedef LONG NTSTATUS;
 #define _wcsdup wcsdup
 #endif
 
+/* The maximum number of characters that can be passed into the
+   HidD_Get*String() functions without it failing.*/
+#define MAX_STRING_WCHARS 0xFFF
+
 /*#define HIDAPI_USE_DDK*/
 
 #ifdef __cplusplus
@@ -61,6 +65,9 @@ extern "C" {
 
 
 #include "hidapi.h"
+
+#undef MIN
+#define MIN(x,y) ((x) < (y)? (x): (y))
 
 #ifdef _MSC_VER
 	/* Thanks Microsoft, but I know how to use strncpy(). */
@@ -148,7 +155,7 @@ static hid_device *new_hid_device()
 	dev->read_pending = FALSE;
 	dev->read_buf = NULL;
 	memset(&dev->ol, 0, sizeof(dev->ol));
-	dev->ol.hEvent = CreateEvent(NULL, FALSE, FALSE /*inital state f=nonsignaled*/, NULL);
+	dev->ol.hEvent = CreateEvent(NULL, FALSE, FALSE /*initial state f=nonsignaled*/, NULL);
 
 	return dev;
 }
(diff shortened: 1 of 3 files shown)


t-paul:
@MichaelAtOz could you please give that a try if the 3D-Mouse is still working? I hope with this change we can easier support the Compact Mouse too and it works for me on Linux but it's not able to open the mouse on my Windows VM.

https://circleci.com/gh/openscad/openscad/1957#artifacts

t-paul:
Oh, another fix incoming, hopefully that makes things work on Windows 10. See https://github.com/signal11/hidapi/issues/231

MichaelAtOz:
The first time I ran it the SM didn't work, it was working yesterday. After unplugging/replugging it worked. So I don't know whether that is related, but I've not had that problem before.
I enabled the Advanced setting, restarted. 
SM worked, I did a bit of jiggling and buttoning, exited, but couldn't find a log file.

t-paul:
The log file should show up in the backup folder as shown in the Library Info window. E.g. on my test VM it's `c:\Users\tp\Documents\OpenSCAD\backups` (it will not create that folder though).

MichaelAtOz:
Log hidapi.log](https://github.com/openscad/openscad/files/2903210/hidapi.log)

Note for the first run I mentioned, there was another OpenSCAD running. Does the SM control switch with the active window?

MichaelAtOz:
p.s. Still Windows 7/64.

t-paul:
I'm not sure how multiple instance behave. I think up to now it would have prevented multiple programs to open the same device.

That actually changes with the latest HIDAPI version and that seems to fix Windows 10, at least on my test VM. Hopefully that's not breaking other Windows versions.

The log looks pretty normal. That's good, so moving to just a single decoder seems to work so far.

@FlipSchaumal maybe give that a try too, it should fix at least case when the mouse is connected via cable.

https://circleci.com/gh/openscad/openscad/1958#artifacts

MichaelAtOz:
Log via wireless USB dongle 
[hidapi.log](https://github.com/openscad/openscad/files/2903258/hidapi.log)

t-paul:
Great, that looks good. So I'll merge this and we can try if that also covers the Compact variant on MacOS which was discussed on the forum.

MichaelPFrey:
awesome. works under linux with a 3Dconnecxion Space Explorer 3D Mouse.
I will also test under Windows 10.

MichaelAtOz:
> I think up to now it would have prevented multiple programs to open the same device. That actually changes with the latest HIDAPI version 

That new version isn't in the above snap?

I just tested, run one of above, run a 2nd, the SM moves the first even with the 2nd on-top-with-focus.

MichaelAtOz:
> awesome. works under linux with a 3Dconnecxion Space Explorer 3D Mouse.

@MichaelPFrey 
All the buttons are customisable?

MichaelAtOz:
> All the buttons are customisable?

Reason I ask, I note the Shift/Alt/Control which lesser SMice lack;
a. It'd be nice to allow reuse, but,
b. Regardless, I have been thinking about Axes Constraint mechanisms.
It is often that you can't prevent a rotation when you attempt a translation etc.
In other S/W Shift/Alt/Control is commonly used to constrain movement.

So, depending on whether the S/A/C keys on your SM are fixed function or customisable may affect the best approach.
I imagine, one-handed, it would be difficult to hold the SM-Shift button while doing a non-trivial movement, so perhaps a toggle mechanism would be better.
...perhaps this should be continued elsewhere...forum?

t-paul:
Yep, I was thinking along the same line, having some way of modifier keys, both via Keyboard and maybe also via the Mouse buttons would be great. But that's food for a new feature request :-).

The snapshot ci1958 from https://github.com/openscad/openscad/pull/2820#issuecomment-467245164 has the latest code. That's the stuff merged to master now.

PR #2821, opened 2019-02-26
Fixes F4 and auto preview not working on file load

nophead:
If the root module is compiled and hasn't changed F4 would do nothing, even if an included file had been fixed.

### DIFF
--- a/src/MainWindow.h
+++ b/src/MainWindow.h
@@ -113,6 +113,7 @@ private:
 	bool checkEditorModified();
 	QString dumpCSGTree(AbstractNode *root);
 	static void consoleOutput(const std::string &msg, void *userdata);
+	static void noOutput(const std::string &, void*) {};  // /dev/null
 	void loadViewSettings();
 	void loadDesignSettings();
 	void updateWindowSettings(bool console, bool editor, bool customizer, bool toolbar);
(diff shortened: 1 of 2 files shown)


nophead:
As well as fixing the modulecache-tests: Test13 I think this also fixes the second part of #2731.

nophead:
I don't think #2731 can be fixed easily. A missing used file is a warning. I think when it is converted to an error with stop on first warnings the parser doesn't create a root_node because PRINT causes an exception, so there are no dependencies to look for. So if the file subsequently turns up it isn't seen until F5 is pressed to force the parser to run again.

Turning parser warnings into errors can't be done by an exception because it needs to carry on and return the node. After that it can stop.

nophead:
Thinking about it some more stop on first warning is incompatible with the parser because it has to run to completion to generate the list of dependencies, regardless of whether they exist or not. 

Since it doesn't take long it could be faked by having a flag that defers exceptions. It would print the first warning and suppress the rest. When the parser has finished and returned the FileModule it could check if any warnings had been generated and then throw the exception, or simply not go on the next phase.

nophead:
Any objections to me merging this? It doesn't do what I set out to do but it fixes not previewing when a file is first opened with auto reload and preview on and fixes the error highlight not updating with F4,

Issue #2814, opened 2019-02-25
Work towards a functioning MSVC Build


thehans:
I've been working on getting a Windows build to work natively within Visual Studio 2017 Community Edition.   

My changes are here if anyone wants to try for themselves:
https://github.com/thehans/openscad/tree/msvc_build

This is attempting to use [VCPKG](https://github.com/microsoft/vcpkg) for dependencies, since it seem to support nearly all OpenSCAD dependencies, aside from OpenCSG and Lib3MF.

Below are the current steps I've used to build OpenSCAD.
I have not tested it extensively, but it builds and runs for all current configurations: x86/x64 Debug/Release 

Mainly the test build, and console app `openscad.com` still need to be worked out.

I will update/edit this top post with more instructions as new things are figured out.

1) Get Visual Studio 2017 Community Edition (this has some integration with cmake)
https://visualstudio.microsoft.com/downloads/

2) Get Chocolatey for flex/bison tool packages:
https://chocolatey.org/install#installing-chocolatey
choco install winflexbison3

3) Set up VCPKG
```
git clone https://github.com/Microsoft/vcpkg.git
cd vcpkg
.\bootstrap-vcpkg.bat
# Next command integrates VCPKG with VS somehow (will prompt for elevated priviledges)
.\vcpkg integrate install
# Set an environment variable for your vcpkg path (current directory).  
# OpenSCAD's CMakeSettings.json will set CMAKE_TOOLCHAIN_FILE based on this
setx VCPKG_PATH %cd%
# Will need to restart Visual Studio (if already open) to pick up new environment variable
```
4) Configure `VCPKG_DEFAULT_TRIPLET` environment variable
vcpkg defaults to target: `x86-windows` which is 32bit, shared libraries.
If you want a 64bit build, you must run: `setx VCPKG_DEFAULT_TRIPLET x64-windows`
There are also `x86-windows-static` and `x64-windows-static` for statically linked libraries but these are not yet possible due to https://github.com/Microsoft/vcpkg/issues/4560 and https://github.com/Microsoft/vcpkg/issues/3409
Run: `.\vcpkg help triplet` to see a list of supported targets.  
Targets for other platforms such as osx, linux, arm exist but those are far out of the scope of this issue.

5) Build/Install OpenSCAD dependencies with VCPKG
```
[...]

PR #2824, opened 2019-02-27
Msvc build testing

thehans:
For #2814 
Creating PR to test if I broke anything for other platforms.


src/UserModule.cc
```diff
@@ -37,7 +37,7 @@
 
 std::deque<std::string> UserModule::module_stack;
 
-static void __attribute__ ((noinline)) print_err(std::string name, const Location &loc,const Context *ctx){
+static void /*__attribute__ ((noinline))*/ print_err(std::string name, const Location &loc,const Context *ctx){
```
MichaelPFrey: for cross-reference: https://github.com/openscad/openscad/pull/2788#discussion_r262237444

src/import_svg.cc
```diff
@@ -51,8 +54,8 @@ Polygon2d *import_svg(const std::string &filename, const double dpi, const bool
 		double width_mm = 0.0;
 		double height_mm = 0.0;
 
-		dim_t min{1.0/0.0, 1.0/0.0};
-		dim_t max{-1.0/0.0, -1.0/0.0};
+		dim_t min{MAX_DOUBLE, MAX_DOUBLE};
+		dim_t max{MIN_DOUBLE, MIN_DOUBLE};
```
MichaelPFrey: Just thinking out loud:
1.0/0.0 looks like a shorthand for infinite to me, not a shorthand for max_double.
But looking at the code, there is no special handling for +infinite or -infinite, so I guess it is assumed that min and max are overwritten anyway, so that both infinite and max_double do the job, but max_double being slightly less error prone then infinity if not overwritten for some reason.

src/linalg.h
```diff
@@ -4,12 +4,23 @@
...
+#ifdef _MSC_VER
+  #include <Eigen/StdVector> // https://eigen.tuxfamily.org/dox/group__TopicStlContainers.html
+  #if !EIGEN_HAS_CXX11_CONTAINERS
+    #warning "Eigen has detected no support for CXX11 containers and has redefined std::vector"
+  #endif
+  typedef std::vector<Vector2d, Eigen::aligned_allocator<Vector2d> > VectorOfVector2d;
+#else
+  typedef std::vector<Vector2d> VectorOfVector2d;
+#endif
+
```
thehans: I ended up fixing Eigen alignment issues the "hard way" before realizing that other builds are just setting `EIGEN_DONT_ALIGN` to workaround these issues.  This section is the main part I'm still not sure about, as I've set it to only trigger if `_MSC_VER` defined.  

This block of code is addressing the issue of "fixed-size vectorizable" eigen objects (Vector2D is multiple of 16bytes) which are used inside of STL containers: https://eigen.tuxfamily.org/dox/group__TopicStlContainers.html
Vector is apparently a special case, where Eigen documentation says to include this `<Eigen/StdVector>` header. Looking at the source, it redefines std::vector if `EIGEN_HAS_CXX11_CONTAINERS` is not defined.  All the builds are using c++11 and should support that as far as I know, so I'm not sure why but it was triggering the warning I placed here(maybe a bug in Eigen's detection).  Consequently Eigen refused to work with initializer lists after redefining `std::vector`, namely here:
https://github.com/openscad/openscad/blob/27ce666e9c9e3f80bd8f022cf94198a816b4901a/src/primitives.cc#L592

Here are some of the build logs with failures just before I wrapped it in the `#ifdef _MSC_VER`:
[Travis](https://travis-ci.org/openscad/openscad/jobs/499237309) 
[CircleCI](https://circleci.com/gh/openscad/openscad/1996?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) 
[AppVeyor](https://ci.appveyor.com/project/kintel/openscad/builds/22687164)

So *maybe* the solution, if we assume Eigen is wrong about whatever `EIGEN_HAS_CXX11_CONTAINERS` is supposed to indicate, is to unconditionally use the `Eigen::aligned allocator` but not the `<Eigen/StdVector>` include?

Anyways the qmake builds on this branch remain unchanged so they still set `EIGEN_DONT_ALIGN` as they have been, to err on the side of caution this close to a new release.  So I think its fine to merge if we want to figure this out in depth later. 

I did briefly test qmake with `EIGEN_DONT_ALIGN` removed in this commit: https://github.com/openscad/openscad/pull/2824/commits/f62d8144ecfcdac224d39ceaf4d3da3415b7c656 
and the builds all passed, (although CircleCI doesn't actually run tests for those builds).
Then reverted that change in most recent commit here: https://github.com/openscad/openscad/pull/2824/commits/35ebcceab82359f096743fc5a38f43ccd98d9342

src/linalg.h
```diff
@@ -31,12 +42,15 @@ template<typename Derived> bool is_nan(const Eigen::MatrixBase<Derived>& x) {
 
 BoundingBox operator*(const Transform3d &m, const BoundingBox &box);
 
-class Color4f : public Eigen::Vector4f
+// Vector4f is fixed-size vectorizable
+// Use Eigen::DontAlign so we can store Color4f in STL containers
+// https://eigen.tuxfamily.org/dox/group__DenseMatrixManipulation__Alignement.html
+class Color4f : public Eigen::Matrix<float, 4, 1, Eigen::DontAlign>
```
thehans: Color4f was another issue with Eigen alignment but in this case I decided it was simpler and easier to specifically set `Eigen::DontAlign` since colors are not heavily computed on.

src/import_svg.cc
thehans: Yeah I changed this due to a warning [C4723](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4723?view=vs-2017) I was getting.  And actually with the current build settings `/we4316 /EHsc` the warning doesn't even show anymore when I change back to `1.0/0.0` so I could go either way with that change.  There were so many warnings when I started, I was doing my best to whittle them down.

t-paul: I'm going to change that to Eigen::AlignedBox anyway :)

t-paul: See #2844

thehans: OK I just reverted my changes in this file to avoid merge conflicts.

src/UserModule.cc
thehans: Changed to use `compiler_specific.h` / `NOINLINE` as suggested.

src/linalg.h
thehans: @kintel, @t-paul Any thoughts on this part?

Also adding permalink so the whole section can be viewed:
https://github.com/openscad/openscad/blob/35ebcceab82359f096743fc5a38f43ccd98d9342/src/linalg.h#L14-L22

src/calc.cc
```diff
@@ -26,7 +26,7 @@
 
 #include "calc.h"
 #include "grid.h"
-#include <cmath>
+#include <math.h>
```
kintel: What does this solve? `cmath` is C++ standard

CMakeLists.txt
```diff
@@ -553,6 +641,7 @@ if(ENABLE_SPNAV)
 endif()
 
 add_executable(OpenSCAD ${Sources} ${RESOURCE_FILES})
+set_property(TARGET OpenSCAD PROPERTY CMAKE_CXX_STANDARD 11)
```
kintel: Do we need this when we set CMAKE_CXX_STANDARD to 11 globally?

thehans: I'm pretty sure it wasn't taking the global setting for some reason.  

I will try again with that commented out and see if its still an issue.  I did an update/upgrade on vcpkg today and it seemed to mess up my harfbuzz install or something, I'm trying to reinstall now (in order to reinstall I had to remove it plus dependencies, namely qt5, so its taking a while to rebuild).

src/calc.cc
thehans: Good point, not sure why I changed that, I think I was just confused.  I updated to use cmath again, and fixed a few other includes.

CMakeLists.txt
thehans: Looks like it works without it, removed now.

src/linalg.h
kintel: This still scares me a bit, but as long as we don't change behavior for other platforms, we can open a new ticket to review/clean this up later.

PR #2832, opened 2019-03-02
SVG viewbox handling

t-paul:
Add viewbox handling as explained in the very nice article
https://www.sarasoueidan.com/blog/svg-coordinate-systems/


t-paul:
Ok, Draft-PRs are useless. They don't trigger Travis (looks like it's designed that way but under consideration of change?).

src/import_svg.cc
```diff
@@ -5,72 +5,140 @@
...
 {
 	try {
 		const libsvg::shapes_list_t *shapes = libsvg::libsvg_read_file(filename.c_str());
 
-		double scale = 1.0;
+		double width_mm = 0.0;
+		double height_mm = 0.0;
+
+		dim_t min{1.0/0.0, -1.0/0.0};
+		dim_t max{1.0/0.0, -1.0/0.0};
```
kintel: The min/max values got messed up here.

tests/CMakeLists.txt
```diff
@@ -1400,6 +1402,11 @@ add_custom_target(issue2342 ALL
   WORKING_DIRECTORY ${GEN_SCRIPT_DIR}
   COMMENT "Generating issue2342.scad"
 )
+add_custom_target(svg_viewbox_tests ALL
+  COMMAND ${GEN_SCRIPT_DIR}/gen_svg_viewbox_tests.py "${CMAKE_SOURCE_DIR}/../testdata/svg/viewbox" "${CMAKE_SOURCE_DIR}/../testdata/scad/svg"
```
kintel: I think running this with `${PYTHON_EXECUTABLE}` would be safer. For some reason, the Python executable in my path is actually a shell script (due to me using `pyenv` to manage python). Using shell scripts in shebangs seems to not work as expected..

src/import_svg.cc
```diff
@@ -5,72 +5,140 @@
...
-					break;
-				case libsvg::unit_t::CM:
-					height = 10.0 * h.number;
-					break;
-				case libsvg::unit_t::MM:
-					height = h.number;
-					break;
+				const libsvg::alignment_t alignment = page->get_alignment();
+
+				std::cout << "page: " << page->dump() << std::endl;
```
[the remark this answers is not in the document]
t-paul: Still trying to debug why it fails on the Linux builds on Travis. It's fine on MacOS and Windows and locally. Something going on with the parsers :(.

kintel:
This example currently breaks due to the min/max stuff: https://gist.github.com/kintel/fd9920c3a463da99b03b1d53065f29f7

src/import_svg.cc
t-paul: Oh, oops. Thanks.

kintel:
Since my one example exported from Illustrator seems to be tricky, perhaps add it as a test case

kintel:
This example demos using OmniGraffle to create the same geometry: https://gist.github.com/kintel/fb30ba45f0adca4969b4f03ef082817f

kintel:
Btw., when I modified the min/max issue, both the Illustrator and OmniGraffle examples worked out of the box. Woot!

tests/CMakeLists.txt
t-paul: It is, the files is generated as the issue2342 script. Not sure if that's the best solution or just running it explicitely with the python executable is simpler.

kintel: Yup, we can always deal with this separate from this PR, as the other python script started breaking for me as well : /

src/import_svg.cc
```diff
@@ -5,72 +5,134 @@
 #include "clipper-utils.h"
 #include "AST.h"
 
+namespace {
+
+constexpr double INCH_TO_MM = 25.4;
+
+struct dim_t {
+	double x;
+	double y;
```
kintel: Using `Eigen::Vector2d` for this could be interesting since this file isn't part of the svg library. -> future refactoring

src/import_svg.cc
```diff
@@ -5,72 +5,134 @@
...
 {
 	try {
 		const libsvg::shapes_list_t *shapes = libsvg::libsvg_read_file(filename.c_str());
 
-		double scale = 1.0;
+		double width_mm = 0.0;
+		double height_mm = 0.0;
+
+		dim_t min{1.0/0.0, 1.0/0.0};
+		dim_t max{-1.0/0.0, -1.0/0.0};
```
kintel: If we're going for Eigen, `Eigen::AlignedBox` could work here.

PR #2843, opened 2019-03-04
Fix 2826

nophead:
Makes stop on first warning and auto reload work together.

Needs more testing but I fly overseas tomorrow and will only have limited internet access.

### DIFF
--- a/src/printutils.cc
+++ b/src/printutils.cc
@@ -32,7 +32,7 @@ void set_output_handler(OutputHandlerFunc *newhandler, void *userdata)
 }
 
 
-void no_exceptions_for_warnigs()
+void no_exceptions_for_warnings()
 {
 	no_throw = true;
 	deferred = false;
--- a/src/printutils.cc
+++ b/src/printutils.cc
@@ -20,10 +20,30 @@ bool OpenSCAD::rangeCheck = false;
 
 boost::circular_buffer<std::string> lastmessages(5);
 
+namespace {
+	bool no_throw;
+	bool deferred;
+}
+
 void set_output_handler(OutputHandlerFunc *newhandler, void *userdata)
 {
 	outputhandler = newhandler;
 	outputhandler_data = userdata;
+}
+
+
+void no_exceptions_for_warnigs()
+{
+	no_throw = true;
+	deferred = false;
+}
+
+bool would_have_thrown()
+{
+    const auto would_throw = deferred;
+    no_throw = false;
+    deferred = false;
+    return would_throw;
 }
 
 void print_messages_push()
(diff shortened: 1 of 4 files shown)


src/printutils.h
```diff
@@ -36,6 +36,8 @@ namespace OpenSCAD {
 }
 
 void set_output_handler(OutputHandlerFunc *newhandler, void *userdata);
+void no_exceptions_for_warnigs();
```
kintel: typo: warnings

src/printutils.cc
```diff
@@ -20,12 +20,32 @@ bool OpenSCAD::rangeCheck = false;
 
 boost::circular_buffer<std::string> lastmessages(5);
 
+namespace {
+	bool no_throw;
+	bool deferred;
+}
```
kintel: These are global variables, but AFAIU, the tracking of deferred warnings is per open document. Not sure how to solve that without passing a whole lot more context to the print function.

nophead: The print stuff already relies on it not switching documents mid event. If it does (which it can do) the output ends up going to the command line console because the interrupting thread changes the output and then clears it. I don't think this PR makes that any worse.

The only way I can think of making PRINT work properly is to pass an object as the first parameter that holds the destination window and these variables. All instances of PRINT will need to be changed and this output window object would need to be passed to all the functions that print. Potentially a huge change to pass the object all the way down.

I think that is the only way to ensure the output goes to the right place when one event can interrupt another when process_events is called or when a thread is used to during F6.

kintel:
@nophead Could you describe a few reproducible manual test cases for this. I can write up the details, just need to know where to start. I was thinking of either cleaning up the module cache tests (https://github.com/openscad/openscad/blob/master/testdata/modulecache-tests/README.txt), or probably just make a similar folder for these tests.

nophead:
I am currently at my new winter residence and don't get broadband installed until Wed. So I can't look at it before then. Basically what happens before my changes is stop on first warning says "execution terminated" but then auto reload makes it run through dependency checking again, so it moves on to the next warning and so on and then eventually compiles and renders. So it makes a mockery of stop on first warning. It actually does completely the opposite because interrupting the dependency checking means some of the mod times don't get updated.

After the fix it processes all the dependencies in the first pass but only displays the first warning, suppresses the rest and then stops before rendering. Auto reload then doesn't do anything on the next pass because nothing has changed and all the stored mod times have been updated.

nophead:
I should also say this is only for warnings during the parser phase. Outside of that this PR doesn't change anything. I don't think exceptions cause problems later on, except perhaps memory leaks, because the data structures get thrown away next time you compile. The parser data structures are cached in the module cache.

PR #2848, opened 2019-03-05
Fix AST dump indentation to not lose track after "inlining" a ModuleInstantiation.

thehans:
Also don't strip leading/trailing whitespace in test output so we can monitor any future changes.

I've looking at the code related to AST and testing the AST dump feature to see how it all works.  I noticed that any inlined/chained module was clearing the indentation, which bugged me so I fixed it.

The only other issue I can see with AST dump, not currently addressed, is that "use"d libraries are not placed in the dump, meaning the result is not re-compilable.

### DIFF
--- a/src/ModuleInstantiation.h
+++ b/src/ModuleInstantiation.h
@@ -44,8 +44,6 @@ public:
 	~IfElseModuleInstantiation();
 	std::vector<AbstractNode*> instantiateElseChildren(const Context *evalctx) const;
 	void print(std::ostream &stream, const std::string &indent, const bool inlined) const final;
-	void print(std::ostream &stream, const std::string &indent) const final { print(stream, indent, false); };
 
 	LocalScope else_scope;
 };
-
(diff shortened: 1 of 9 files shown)


src/ModuleInstantiation.h
```diff
@@ -42,7 +43,8 @@ class IfElseModuleInstantiation : public ModuleInstantiation {
 	IfElseModuleInstantiation(shared_ptr<class Expression> expr, const std::string &source_path, const Location &loc) : ModuleInstantiation("if", AssignmentList{Assignment("", expr)}, source_path, loc) { }
 	~IfElseModuleInstantiation();
 	std::vector<AbstractNode*> instantiateElseChildren(const Context *evalctx) const;
-	void print(std::ostream &stream, const std::string &indent) const override;
+	void print(std::ostream &stream, const std::string &indent, const bool inlined) const final;
+	void print(std::ostream &stream, const std::string &indent) const final { print(stream, indent, false); };
```
kintel: What's the purpose of this? It seems to have the same implementation as its supertype, except it's `final`?

thehans: Yeah looks like its unnecessary, removed now.

PR #2858, opened 2019-03-10
Fix some compiler/clang warnings

thehans:
clang doesn't support -frounding-math, some unused declarations, and override issues.


src/cgalutils-polyhedron.cc
```diff
@@ -270,7 +270,7 @@ namespace CGALUtils {
 	{
 		bool err = false;
 		typedef typename Polyhedron::Vertex                                 Vertex;
-		typedef typename Polyhedron::Vertex_const_iterator                  VCI;
+		//typedef typename Polyhedron::Vertex_const_iterator                  VCI;
```
kintel: If it's not used, we should just kill it

thehans: done :)

PR #2864, opened 2019-03-13
Sound notification timer

RomitKumar:
I removed the "Play sound notification on render complete" checkbox from Edit->Preference->Advanced and added a text field to take as input number of seconds. Notification sound will play if rendering time is more than or equal to the provided input


kintel:
How would we disable notifications, just set it to a really large number? ..or add a checkbox in addition?

RomitKumar:
With current modifications, the only option to disable notification is to set it really large number. However, I think that adding the checkbox in addition will be more neat. I will work on implementing it.

RomitKumar:
Added the checkbox to completely disable notification

src/Preferences.cc
```diff
@@ -163,10 +163,18 @@ void Preferences::init() {
...
+	this->defaultmap["advanced/timeThresholdOnRenderCompleteSound"] = 0;
 	this->defaultmap["advanced/enableHardwarnings"] = false;
 	this->defaultmap["advanced/enableParameterCheck"] = true;
 	this->defaultmap["advanced/enableParameterRangeCheck"] = false;
 
+	connect(this->enableSoundOnRenderCompleteCheckBox, SIGNAL(toggled(bool)),
+		        this->timeThresholdOnRenderCompleteSoundLabel, SLOT(setEnabled(bool)));
+	connect(this->enableSoundOnRenderCompleteCheckBox, SIGNAL(toggled(bool)), 
+		        this->secLabel, SLOT(setEnabled(bool)));
+	connect(this->enableSoundOnRenderCompleteCheckBox, SIGNAL(toggled(bool)),
```
kintel: Since we use automatic connections for all other controls, we should probably do the same for this.

RomitKumar: Can you please elaborate. What I understood is that you are expecting a function for SLOT in which state enabled status is toggled.

kintel: See https://doc.qt.io/qt-5/designer-using-a-ui-file.html#automatic-connections

..and the slots starting here:  https://github.com/openscad/openscad/blob/master/src/Preferences.h#L34

Issue #2360, opened 2018-03-30
Possible Improvement using CGAL "Polygon Mesh Processing" for booleans and more?


thehans:
OpenSCAD is currently using CGAL Nef Polyhedra for boolean operations, which can be quite slow for complex models.

I was looking at other options within CGAL and I found this section about mesh "corefinement" and booleans:
https://doc.cgal.org/latest/Polygon_mesh_processing/index.html#Coref_section

It sounds like this could be much faster than Nef processing, though maybe not as robust in some ways?
Although even if not intrinsically as robust, this also looks like it may be capable of better detecting and even possibly fixing mesh errors with the various functions provided (the [Hole Filling](https://doc.cgal.org/latest/Polygon_mesh_processing/index.html#PMPHoleFilling) section looks pretty impressive).

According to the package [overview page](https://doc.cgal.org/latest/Manual/packages.html#PkgPolygonMeshProcessingSummary), this was introduced in CGAL 4.7.  
OpenSCAD currently only requires CGAL 3.6, so this would be a large bump to our required version if this change were made.

kintel:
I did some early experimentation with this back in 2013 (see https://github.com/openscad/openscad/tree/epec-fast). The corefinement stuff appears to be slightly less robust than Nef3 polyhedrons, but it feels like something we may be able to work around.
The CGAL requirement was a bit of a showstopper back then, but this has been available for a while now, so perhaps time to give it another shot.

MichaelAtOz:
How about 4.8? [Multithreading???](https://doc.cgal.org/4.8/Manual/devman_multithreading.html)

kintel:
I haven't looked into whether corefinement is thread-safe, but since they use the same datatypes as Nef3 polyhedrons, my guess is that the same limitations apply.

thehans:
@kintel Was the epec-fast branch in a state that built and ran at any point or was it still WIP last time you left it?

I checked it out for a quick test but the build fails looking for `Dock.h`.  

Do you have some recommendations for next steps to move this forward?  I could give it a shot rebasing the branch to get it up to date with master?

kintel:
I cannot remember how I left it. It was really just a quick and dirty test, without any performance optimizations as I converted everything back and forth to Nef3 polyhedrons. My goal was to determine the CSG operation stability implications of using corefinement, but I had too many other things on my list..
[...]

PR #2867, opened 2019-03-14
Builds without CGAL

thehans:
This is my attempt to separate CGAL functionality from the core of OpenSCAD so we can be build without it, with hopes that it will ease any future attempts to use alternate rendering libraries.

I started this after some discussion on IRC about possibly refactoring the `GeometryCache` code(still not part of this PR), where @kintel mentioned to keep in mind that any changes to cache should not necessarily assume CGAL is the renderer. That got me started down this rabbit hole of determining what it would take to actually build OpenSCAD without CGAL at this point.

During refactoring, I felt it was best to split `GeometryEvaluator` into separate classes, so it is now `BaseGeometryEvaluator` and `CGALNefEvaluator` which extends the base.  Also I realized that there are two independent parts of CGAL that we are currently using: The CGAL_Nef_Polyhedron based function for 3D boolean/Minkowksi/Hull, etc., but also the 2D tessellation and 2D Hull which don't use Nef Polyhedrons at all.

So I have made these into separately configurable build options as `ENABLE_CGALNEF` and `ENABLE_CGAL2D`

Summary of defines:
**ENABLE_CGAL2D**
 - Really important for 2D tessellation: `Polygon2D-CGAL.cc`
 -  Even OpenCSG preview depends on this for basically any 2D operations, including linear_extrude and rotate_extrude with angles(end faces must be tessellated).
 - 2D Hull

**ENABLE_CGALNEF**
 - Needed in most cases for F6/Render of 3D geometries
 - 3D boolean rendering
 - 3D Hull
 - 3D Minkowski
 - 3D Import/Export: AMF, 3MF, STL, OFF, NEF3, NEFDBG
    Some formats still work for trivial cases of import or export of single geometry, no implicit union.
 - projection(cut=true) uses CGAL
 - projection(cut=false) uses Clipper and doesn't directly require CGAL, but applying to anything more than a single object forces us to use CGALUtils::applyOperator()
  
**ENABLE_CGAL**
 - General check if CGAL is linked, automatically set to true if either ENABLE_CGALNEF or ENABLE_CGAL2D are on.
 - I try to use the more specific defines above rather than this wherever possible.

Generally I tried to minimize the scope of these preprocessor checks so they only omit what is absolutely necessary.

This was also done with #2360 in mind, which I am still interested in trying.  Summary: This would still use CGAL, but with different mesh-based methods for boolean operations etc., which I believe may be much faster than the current Nef approach.

There are some files/classes which we may want to rename since they are still compiled and run when CGAL is disabled:
 - cgalworker: this should be able to handle other types of renderer, and the code isn't really relying on CGAL in any way.
 - cgaladv[node] (minkowski, hull, resize): Clipper does 2D minkowski, and we can resize Polygon2d, and sort of resize 3D PolySets without necessarily using CGAL.  `BaseGeometryEvaluator` still includes `cgaladvnode.h` when CGAL is completely disabled.
 - CgalRenderer: ok, not terribly useful for displaying anything other than a single PolySet when CGAL is completely disabled, but still it passes some tests and allows for limited export functionality.  Tempted to call this `RenderRenderer`, but that sounds kinda crazy.
  
I set this up to be compatible with both qmake and cmake build systems.  I implemented cmake first, and the qmake settings use inverted logic compared to cmake because I didn't know of a way to have default-on options that can be disabled from command line.

When a function is not supported, I print errors but still continue rendering as much as possible(avoid crashes or triggering assertions).  For example when CGAL Nef is disabled, a `render()` node for OpenCSG preview prints an error, but still passes through as if it were a OpenCSG union.  I tried my best to pass as many tests as possible with unimplemented features, below are summaries of my results.  
```
# Default build
cmake -DENABLE_CGALNEF=ON -DENABLE_CGAL2D=ON -DEXPERIMENTAL=ON -DCMAKE_BUILD_TYPE=Debug
1239 / 1239 tests passed (100%)
qmake CONFIG+=experimental CONFIG+=debug
1239 / 1239 tests passed (100%)

# Disable CGAL Nef
cmake -DENABLE_CGALNEF=OFF -DENABLE_CGAL2D=ON -DEXPERIMENTAL=ON -DCMAKE_BUILD_TYPE=Debug
989 / 1239 tests passed (80%)
qmake CONFIG+=experimental CONFIG+=no_cgalnef CONFIG+=debug
989 / 1239 tests passed (80%)

# Disable CGAL 2D
# Kind of a nonsense option to use cgal nef and not cgal 2d in my opinion, but technically supported now
cmake -DENABLE_CGALNEF=ON -DENABLE_CGAL2D=OFF -DEXPERIMENTAL=ON -DCMAKE_BUILD_TYPE=Debug
816 / 1239 tests passed (66%)
qmake CONFIG+=experimental CONFIG+=no_cgal2d CONFIG+=debug
816 / 1239 tests passed (66%)

# Fully Disabled CGAL
cmake -DENABLE_CGALNEF=OFF -DENABLE_CGAL2D=OFF -DEXPERIMENTAL=ON -DCMAKE_BUILD_TYPE=Debug
607 / 1239 tests passed (49%)
qmake CONFIG+=experimental CONFIG+=no_cgal2d CONFIG+=no_cgalnef CONFIG+=debug
607 / 1239 tests passed (49%)
```


nophead:
I thought tessellation was done with the Tess library. Are there two different tessellations?

thehans:
Yes the tessellation for 3D PolySet is different from the tessellation for Polygon2d (which returns a 2D PolySet)

kintel:
This is hopelessly out of date now. @thehans if there are any ideas/snippets worth saving, could you surface that somewhere?

Issue #2278, opened 2018-02-01
Surface errors should reflect the file type


stuartpb:
From a file with this line:

```
  scale(raster_scale, raster_scale, h / 100) surface(file = "tubman-portrait.png", invert=true);
```

I got this error:

```
WARNING: Can't open DAT file 'C:/Users/Stuart/Documents/GitHub/tubman20-stamp\tubman-portrait.png'.
```

The actual cause of the error was that I'd made a typo in the name of the surface file (so it was trying to open a file that didn't exist), but the error message led me to believe it was because the PNG was being misinterpreted as a DAT.

thehans:
I think the solution would be to add a check before this highlighted section of code, for whether the file exists.
https://github.com/openscad/openscad/blob/master/src/surface.cc#L142-L147

As it is, it does not distinguish between "a file that is not a png" and "a file that doesn't exist".

I would write a quick fix but I'm not sure whats the preferred method for cross platform file check.

css911:
@kintel Hi... can you please assign this issue to me..

t-paul:
Ok, lets give that a go. As the comment above already mentions, it may need a little bit of staring at the current logic to come up with a nice solution. So I'd propose you share a prototype solution via PR. It may spark some discussion but please consider that a good thing.

css911:
Sure... excited to learn from the discussion as well. Thanks @t-paul.

kintel:
@css911 GitHub assignment only works for organization members, it's enough to leave a comment here that you're working on it.

PR #2876, opened 2019-03-17
Fixes#2278 Surface errors should reflect the file type

css911:
This commit Fixes issue no #2278. To solve this issue, we are checking the return value returned by the lodepng_filesize() function (from /ext/lodepng/lodepng.cpp file) which is been called from the load_file() function(from the surface.cc file), the load_file() function (is from ext/lodepng/lodepng.cpp file) .The function lodepng_filesize return a  -1   when the file to be open is not found or when the file to open is too big.

This solves the issue that the a file which was not exits was identified as a dat file, now instead a warning message is been shown which informs that the file doesn't exits or its to big to open.

### DIFF
--- a/src/surface.cc
+++ b/src/surface.cc
@@ -146,7 +146,7 @@ img_data_t SurfaceNode::read_png_or_dat(std::string filename) const
 	}
 
 	if(ret_val == 78){
-		PRINTB("WARNING: The file '%s'. does not exists or is too large to load", filename);
+		PRINTB("WARNING: The file '%s' couldn't be opened.", filename);
 		return data;	
 	}
 	
--- a/src/surface.cc
+++ b/src/surface.cc
@@ -136,9 +136,15 @@ img_data_t SurfaceNode::read_png_or_dat(std::string filename) const
 {
 	img_data_t data;
 	std::vector<uint8_t> png;
-	
-	int ret_val = lodepng::load_file(png, filename);
-	
+	int ret_val = 0;	
+	try{
+		 ret_val = lodepng::load_file(png, filename);
+	}catch(std::bad_alloc &ba){
+		
+		PRINTB("WARNING: bad_alloc caught for '%s'.", ba.what());
+		return data;	
+	}
+
 	if(ret_val == 78){
 		PRINTB("WARNING: The file '%s'. does not exists or is too large to load", filename);
 		return data;	
--- a/src/surface.cc
+++ b/src/surface.cc
@@ -137,7 +137,12 @@ img_data_t SurfaceNode::read_png_or_dat(std::string filename) const
 	img_data_t data;
 	std::vector<uint8_t> png;
 	
-	lodepng::load_file(png, filename);
+	int ret_val = lodepng::load_file(png, filename);
+	
+	if(ret_val == 78){
+		PRINTB("WARNING: The file '%s'. does not exits or is too large to load", filename);
+		return data;	
+	}
 	
 	if (!is_png(png)) {
 		png.clear();
--- a/src/surface.cc
+++ b/src/surface.cc
@@ -140,7 +140,7 @@ img_data_t SurfaceNode::read_png_or_dat(std::string filename) const
 	int ret_val = lodepng::load_file(png, filename);
 	
 	if(ret_val == 78){
-		PRINTB("WARNING: The file '%s'. does not exits or is too large to load", filename);
+		PRINTB("WARNING: The file '%s'. does not exists or is too large to load", filename);
 		return data;	
 	}
 	


css911:
Screenshot from 2019-03-17 14-40-11

The screenshot of the change.

css911:
Screen Shots of changes carried out.

 Line No 140 from the above image is calling function on line no 6640 from the below image.



line no 6642 calling function written on line no 347. and this function is returning -1 if the file is not found and hence the function call is been returned back.

src/surface.cc
```diff
@@ -137,7 +137,12 @@ img_data_t SurfaceNode::read_png_or_dat(std::string filename) const
 	img_data_t data;
 	std::vector<uint8_t> png;
 	
-	lodepng::load_file(png, filename);
+	int ret_val = lodepng::load_file(png, filename);
+	
+	if(ret_val == 78){
```
kintel: 78 is just one of the possible return values. Should this be `!= 0` instead?

src/surface.cc
```diff
@@ -137,7 +137,12 @@ img_data_t SurfaceNode::read_png_or_dat(std::string filename) const
...
+		PRINTB("WARNING: The file '%s'. does not exits or is too large to load", filename);
```
kintel: Typo: exits -> exists

css911: Sure will make the correction. :+1:

src/surface.cc
css911: Quoting **Image2** and **Image3** from the previous comment. When the file is not getting successfully open. the **lodepng_filesize()** (Image3 line no 347) returns -1 to load_file() (Image2 line no 6640) and then this function is checking whether the value(size) is less than 0 and returning 78 to the variable ret_val (Image1). 

The reason i kept it 78 as it was already in the code and changing it might affect the code base. as the function might have been called somewhere else as well.

src/surface.cc
css911: Done :)

src/surface.cc
thehans: Please don't post images of source code as references.  You can create links directly to the source on github as a better alternative.
For example :
1) go to: https://github.com/openscad/openscad
2) Press 't' hotkey, to search by filename
3) type 'lodepng' and click on the cpp file to view
4) a) For a single line you can just click the line number (374) for example
    b) To select multiple lines click the first line number in the left column (368), then shift-click the end line (381) 
5) Click the "..." that should now show on the left of your selection
6) "Copy Permalink", then paste that into a comment.  It will show up as follows:

https://github.com/openscad/openscad/blob/2a697e4090cfdb0e376dd5565eaad104dde0c133/src/ext/lodepng/lodepng.cpp#L368-L381

kintel: ok, I see what's going on now; hard to read from the screenshots:
* If the file cannot be opened or read, lodepng::load_file() will return 78
* If the file buffer cannot be allocated lodepng::load_file() will throw std::bad_alloc

You're only handling the first case, which is fine for now, but you error message should reflect that.

css911: will writing a try catch block do to handle the second case as well?

kintel: Yes, I believe so.

css911: I have written a try a try catch block for handling the bad_alloc error.

css911: > ok, I see what's going on now; hard to read from the screenshots:
> 
>     * If the file cannot be opened or read, lodepng::load_file() will return 78
> 
>     * If the file buffer cannot be allocated lodepng::load_file() will throw std::bad_alloc
> 
> 
> You're only handling the first case, which is fine for now, but your error message should reflect that.

Should i write a try and catch block?

css911: > Yes, I believe so.

I accidently deleted my previous comment is there a way to get it back? I was me asking should i write a try catch block?

src/surface.cc
```diff
@@ -136,8 +136,19 @@ img_data_t SurfaceNode::read_png_or_dat(std::string filename) const
...
+	try{
+		 ret_val = lodepng::load_file(png, filename);
+	}catch(std::bad_alloc &ba){
+		
+		PRINTB("WARNING: bad_alloc caught for '%s'.", ba.what());
+		return data;	
+	}
+
+	if(ret_val == 78){
+		PRINTB("WARNING: The file '%s'. does not exists or is too large to load", filename);
```
kintel: Now that you're catching the "too large to load" issue, that text shouldn't be part of the error message any longer.
Also, the error should probably say "couldn't be opened" instead of "does not exist", as there are multiple reasons why a file cannot be opened (e.g. wrong permissions).

css911:
Thanks @kintel @thehans and @t-paul for helping me with my first approve :+1:

PR #2881, opened 2019-03-20
Value rework

thehans:
I mentioned this some on IRC, but I have been looking into the way ValuePtr is used around the code and felt it was unnecessary to use a shared_ptr on Values with fixed size: `UNDEFINED, BOOL, NUMBER, RANGE`.   Mainly just `VECTOR` and `STRING` ValueTypes should be on managed pointers.  So I tore out ValuePtr and created a VectorPtr class which encapsulates vector types(so it is now `shared_ptr<std::vector<Value>>` instead of `std::vector<ValuePtr>`, where ValuePtr was basically `shared_ptr<Value>`), and added `shared_ptr` management to the `str_utf8_wrapper` class.

Since `ValuePtr` was used in many places, the diff is unfortunately quite huge.  I think the majority of changes are uninteresting `ValuePtr`to`Value` or `->`to`.` etc.
Most of the actually changed functionality resides in `value.h` and `value.cc`.

Another goal of mine while I was looking at this was to be more explicit about all the places we create copies.  So all constructors for Value and its variant classes have had their default copy constructor and copy assignment deleted.  This means we have to use move assignment and move constructors as much as possible.  I would have liked to try to use `unique_ptr` instead of `shared_ptr` for even less overhead, but unfortunately we still do need to make copies which rules out `unique_ptr`.  For cases where copies are necessary, I added a `Value::clone()` method that copies the value without needing an implicit copy construction.  This way it is very clear and explicit where all the places in code that copies are made, as they have to call clone().

const qualifiers have also been added to many value related methods wherever I could determine it was possible.  This really helped me squash some of the bugs I introduced as I was refactoring.  Also I have given constructors for Value and its variant classes `explicit` constructors, so we aren't accidentally converting values in any unwanted ways.  This means that a function signature with `Value` return type if its returning a double, has to do `return Value(d);` whereas before it could `return d;`

Another change was to make `Context::lookup_variable()` method return by const reference.  This should avoid some unnecessary copies.

One other note is that `Value::undefined` can't be returned by value since that just uses copy construction anyways, so many instances of that are now just `return Value();`

I ended up doing a lot of refactoring, partly to help myself debug my changes, but I think many of the changes make our Value interface a lot more robust.  

Some notes on where/why we still clone:
 - One of the sources of requiring copies is the `bracket_visitor`.  Ideally this should return a const reference, but for string types it has to create a Value representing the char.  So this local Value can't be returned by reference.  If we stored strings internally a fixed width `utf32`, then we would be actually be able to return a reference to a char within a string.
 - Another place where we have to copy is any `Expression::evaluate` function that looks up a literal or vector lookup, because that Expressions return `Value` by value, not reference (again many of them have to create local variable which get moved out, so we can't easily change this to return by reference).  
 - There are some .clone() calls being used now in `dxfdim.cc`, since it registers Builtin functions that cache values and have to return by value based on the Builtin function interface.
- In some Node types and customizer parameter widgets store their own `Value` properties, so const reference values coming from `lookup_variable` must be cloned to be assignable into them.

I hope this is received well, it was a lot of effort and hard thinking.


src/settings.cc
```diff
@@ -128,19 +135,19 @@ SettingsEntry* Settings::getSettingEntryByName(const std::string &name)
...
+Value Settings::get(const SettingsEntry& entry) const
 {
-	return entry._value;
+	return entry._value.clone();
 }
 
 void Settings::set(SettingsEntry& entry, const Value &val)
 {
-	entry._value = val;
+	entry._value = val.clone();
```
kintel: In this case, you're _always_ taking a copy of the object. We can automate this by passing the `val` argument by value and `std::move()` it here. This has an added benefit of allowing any object literal to be move-constructed when calling this function, and will reduce the need for having to manually `move()` it at the call site. See other comment for call site

src/Preferences.cc
```diff
@@ -92,9 +92,9 @@ class SettingsReader : public Settings::SettingsVisitor
 
 	std::string key = entry.category() + "/" + entry.name();
 	std::string value = settings.value(QString::fromStdString(key)).toString().toStdString();
-	const Value v = getValue(entry, value);
+	Value v = getValue(entry, value);
 	PRINTDB("SettingsReader R: %s = '%s' => '%s'", key.c_str() % value.c_str() % v.toString());
-	s->set(entry, v);
+	s->set(entry, std::move(v));
```
kintel: As mentioned elsewhere, we can push this `move()` into the `set()` function and let the object be copied when passing by value. If we, e.g. somewhere else, want to pass a Value literal, that will be move constructed and no copy is made.

thehans: I can change `set` signature to pass by value, but it wouldn't make much difference from this context.  Since one of the main things I changed was to delete the copy constructor on `Value`, and `v` is already an lvalue, I would still need to either move or clone it in this call.  If the `PRINTDB` in the line above weren't using v, I could I could do `s->set(entry, getValue(entry, value));`.  (OK I guess regardless of `PRINTDB` I could do that but it would be a redundant call to `getValue` instead of calling `std::move` *shrug*).

thehans: Deleting the copy constructor was really eye-opening and taught me a lot.  I like the idea of having to explicitly clone when needed rather than having the compiler automatically do it, it makes you more aware of where things happen. 
 Then again Values should still be fairly cheap to copy I think, so if you think it would be better to just un-delete copy ctor, we could do that too.  I was originally trying to work towards having no copies at all somehow(and being able to use `unique_ptr` rather than `shared_ptr`), so seeing every clone operation was helpful, but it seems that eradicating copies entirely is not feasible with the way OpenSCAD works.

kintel: I didn't really think about the deleted copy constructor. Perhaps let's revisit this once I've reviewed the rest..

thehans:
Well I've had a hell of a time trying to get Trusty to work with these changes, and ended up using wandbox.org C++ compiler website to test out a bunch of various changes for quicker turnaround. 
Here is the example code that fails to compile given gcc 4.84 and boost 1.54 to match Trusty environment:
 https://wandbox.org/permlink/3r7Vw0dxhrTi4dAZ
It is mostly copy/pasted from `src/value.cc`, but with various things commented out like `glib` and `double_conversion` functions since I don't think it has those headers available.

I started trying incrementally newer versions of boost and found that Boost 1.56 worked ok. 
And sure enough looking at the [changelog for Boost 1.56](https://www.boost.org/users/history/version_1_56_0.html) shows this:
> Variant
> ...
>   - Move-only classes now can be returned by visitors.

So yeah, not sure what to do about that.  

Trusty reaches EOL next month.  But I don't think Travis is really working towards having another newer LTS version(Bionic) supported any time soon.  I'm amazed they still offer environments for Precise too, so I doubt the EOL will change anything with regards to Travis support.

kintel:
I wonder if we could just perform the Travis CI builds in our own Docker image, and base that off a newer distro. I don't see any other reasons to still support Trusty.

kintel:
Otoh, it looks like Xenial is the main environment on Travis now: https://blog.travis-ci.com/2018-11-08-xenial-release

..so perhaps we can just silently turn off Trusty support?

src/value.cc
```diff
@@ -511,27 +487,48 @@ std::string Value::chrString() const
   return boost::apply_visitor(chr_visitor(), this->value);
 }
 
-const Value::VectorType &Value::toVector() const
+void Value::VectorPtr::flatten() {
+  int n = 0;
+  for (unsigned int i = 0; i < (*this)->size(); i++) {
```
kintel: Could you just do `for (const auto v : **this)` here?
Same below.

src/value.cc
```diff
@@ -511,27 +487,48 @@ std::string Value::chrString() const
...
+      n += (*this)[i].toVectorPtr()->size();
+    } else {
+      n++;
+    }
+  }
+  Value::VectorPtr ret; ret->reserve(n);
+  for (unsigned int i = 0; i < (*this)->size(); i++) {
+    if ((*this)[i].type() == Value::ValueType::VECTOR) {
+      Value::VectorPtr &vec_ptr = (*this)[i].toVectorPtrRef();
+      for(unsigned int j = 0; j < vec_ptr->size(); ++j) {
```
kintel: This for loop looks equivalent to `ret->insert(end(*ret), begin(vec_ptr), end(vec_ptr))`

src/value.cc
```diff
@@ -315,18 +274,35 @@ bool Value::getFiniteDouble(double &v) const
   return valid;
 }
 
+Value::VectorPtr::VectorPtr(double x, double y, double z) : ptr(make_shared<Value::VectorType>()) {
+  (*this)->emplace_back(x);
```
kintel: Uncertain about this, but since the `vector` constructor takes an initializer list, could you just to `: ptr(make_shared<Value::VectorType>(x, y, z)) {`

src/primitives.cc
```diff
@@ -121,20 +121,20 @@ class PrimitiveNode : public LeafNode
...
 			PRINTB("WARNING: Ignoring radius variable '%s' as diameter '%s' is defined too, %s", radius_var % diameter_var % locStr);
 		}
-		return {d->toDouble() / 2.0};
+		return Value(d.toDouble() / 2.0);
 	} else if (r_defined) {
-		return *r;
+		return r.clone();
 	} else {
-		return Value::undefined;
+		return Value();
```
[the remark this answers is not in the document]
thehans: OK, so this is another issue with the move vs copy change.  Places that return a const reference still use `Value::undefined`  Otherwise if its return by value, since `Value::undefined` is const, it requires a copy or a new construction.  So in these places I could explicitly call `Value::undefined.clone()` but that seemed weird and messy looking.  I actually toyed with the idea of using a static function `Value::undefined()` to return a newly constructed undefined value, but then the const ref instances would need to change as I don't think we can have redefined member function and static member data with the same name.   What do you think about implementing both with slightly different naming, maybe the call the static member function `Value::undefined()`, and call the const ref `Value::undef`, to match with scad nomenclature?

src/value.cc
```diff
@@ -567,29 +564,24 @@ bool Value::getVec3(double &x, double &y, double &z, double defaultval) const
...
-  if (this != &v) {
-    this->value = v.value;
-  }
-  return *this;
-}
 
 class equals_visitor : public boost::static_visitor<bool>
 {
 public:
+	typedef bool result_type;
```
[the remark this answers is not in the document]
thehans: This was something I noticed that boost technically specifies as required for static visitors
[reference](https://www.boost.org/doc/libs/1_64_0/doc/html/variant/reference.html#variant.concepts.static-visitor): see second point
> The requirements on a static visitor of a type T are as follows:
> 
> - Must allow invocation as a function by overloading operator(), unambiguously accepting any value of type T.
> - Must expose inner type result_type. C++14 compatible compilers could detect result_type automatically, but will stick to result_type if it is defined. (See boost::visitor_ptr for a solution to using functions as visitors.)
> - If result_type is not void, then each operation of the function object must return a value implicitly convertible to result_type.

I added this to see if it would help with the Trusty compile issues, but anyways as I found out, its due to the boost version not supporting move-only types as return values for visitors, so adding these definitions wasn't a solution to the problem.  Its been working fine without it so far, so I'm fine either removing it again or, or maybe just documenting in comments that boost technically requires it (but doesn't really need it i guess).

jschobben:
This looks like a great change! It would help my "project" to reduce vector copies.
Is it still being worked on? I'd be happy to help out.

src/value.h
```diff
@@ -85,159 +93,185 @@ class RangeType {
...
+		VectorPtr() : ptr(make_shared<VectorType>()) {}
+		VectorPtr(double x, double y, double z);
+		VectorPtr(const VectorPtr &) = delete; // never copy, move instead
+		VectorPtr& operator=(const VectorPtr &) = delete; // never copy, move instead
+		VectorPtr(VectorPtr&&) = default; // default move constructor
+		VectorPtr& operator=(VectorPtr&&) = default; // default move assignment
+
+		// Copy explicitly only when necessary
+		// We can't use unique_ptr because of this :(
+		VectorPtr clone() const { VectorPtr c; c.ptr = this->ptr; return c; }
```
jschobben: This clones the shared_ptr, of course. To avoid silly mistakes, maybe this (and other pointer-clone methods) could be called ```clonePtr()``` or so? Then keep the real ```clone()``` for Value.
In practice, sometimes you may have code like this that does not immediately show that a pointer is being cloned:
```
vec = result.clone();
```

src/value.h
```diff
@@ -85,159 +93,185 @@ class RangeType {
...
+
+		const Value::VectorType &operator*() const noexcept { return *ptr; }
+		Value::VectorType *operator->() const noexcept { return ptr.get(); }
+		// const accesses to VectorType require .clone to be move-able
+		const Value &operator[](size_t idx) const noexcept { return idx < ptr->size() ? (*ptr)[idx] : Value::undefined; }
+		// non-const return operator[] is only accessible from protected toVectorPtrRef
```
jschobben: ... or from ```instance.clone()```

t-paul:
As this started to bit-rot badly and the conflicts against `function-literals` are even more severe I've merged those. See #3096 for this changes here on top of `function-literals` but still as separate pull request.

I'm not closing this one yet as I suspect it might be easier to finish the review here and just put the updates into the other branch. But I'm open for other suggestions on how to get the value changes in.

src/cgaladv.cc
```diff
@@ -59,37 +59,36 @@ AbstractNode *CgaladvModule::instantiate(const Context *ctx, const ModuleInstant
...
-		else if ( autosize->type() == Value::ValueType::BOOL ) {
-			node->autosize << autosize->toBool(),autosize->toBool(),autosize->toBool();
+		else if ( autosize.type() == Value::ValueType::BOOL ) {
+			node->autosize << autosize.toBool(),autosize.toBool(),autosize.toBool();
 		}
 	}
 
-	node->convexity = static_cast<int>(convexity->toDouble());
-	node->path = path;
+	node->path = Value();
```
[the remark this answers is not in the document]
t-paul: Removed in https://github.com/openscad/openscad/pull/3096/commits/2903eba1af16cc29808574d48110bb38032470ed

src/value.cc
t-paul: Changed in https://github.com/openscad/openscad/pull/3096/commits/d42714a6cad756645f1ab635316fedf66f05af0d - using `std::accumulate()`

src/value.cc
t-paul: The `for` loop needed an additional non-const `operator *()`.

Changed to use `insert()` and move iterators via https://github.com/openscad/openscad/pull/3096/commits/ca6d46c3a9b77739395150f39ef8cd325acf2343.

src/value.cc
t-paul: I think it's possible but much too complicated to be worth it. The problem is that the value type of `VectorType` is `Value` which is move-only. That does not easily work with initializer lists. There are some example that make it work with `make_moveable_iterator` but that seems  too messy for this case.

src/primitives.cc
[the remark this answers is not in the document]
thehans: I've been continuing work on this PR lately, in t-paul's value-rework-2 branch.  I have addressed this by adding a static `Value::undef()`  member function for cases where undefined is needed by-value (I guess I ended up doing the opposite naming of what I proposed above).

And for extra completeness, I made `Value()` default constructor as private, to enforce being explicit about undefined values. Values should be immutable anyways so I figured it didn't make much sense to allow default constructing one then change it later.
This had a side effect of pointing out another place where Values were being inadvertently default constructed:  
when `unordered_map<value>::operator[]` is used to assign a value; If it requires an insert, then a value is default constructed, and this was happening for `Context`'s `ValueMap` typedef in a couple places.  
In C++17 there is `unordered_map::insert_or_assign` function which seems to be a way to workaround this default construction, so I've made a function which I think is more-or-less equivalent in operation, and put ValueMap into its own wrapper class with its own header `valuemap.h` instead of a typedef under `context.h`.  I also was able to simplify a few other operations and duplicated `unordered_map::find` calls while looking over the calling points for this class.
Lastly, `dxfdim.cc` used a similar value map cache which now is able to re-use class `ValueMap` too.

src/value.cc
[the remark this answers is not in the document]
thehans: I ended up removing these in https://github.com/openscad/openscad/pull/3102  since we're officially on C++14, and boost shouldn't need them.

src/value.h
thehans: Some sizeable changes were made to the interface of vectors in #3102.  VectorType iterators and `operator[]` *only* provide const reference access to elements now. 
This means even a non-const cloned vector can't move elements out of the original, which fits better with supposed immutability.

I guess there's still one remaining danger that a clone could have additional elements inserted into it, which would alter the original.
I'm not sure how to solve that without making vectors truly immutable upon construction, but this would require very complicated range-based constructors which can somehow forward arguments to `std::vector`, through `boost::variant`, etc.  I event spent a day thinking if this could be done, but I was unable to figure out.  Also related to @t-paul's [comment here](https://github.com/openscad/openscad/pull/2881#discussion_r336788449) about such difficulties.

src/value.h
thehans: See my reply below about the interface changes which should make this less of an issue.
This comment did cause me to think harder about when it's safe to move elements out of vectors and ultimately resulted in the new `EmbeddedVectorType` which has shown great benefits, so I do appreciate the input.

src/context.cc
```diff
@@ -90,44 +90,36 @@ void Context::setVariables(const EvalContext *evalctx, const AssignmentList &arg
...
-	set_constant(name, ValuePtr(value));
-}
-
-void Context::apply_variables(const Context &other)
+void Context::take_variables(Context &other)
 {
-	for (const auto &var : other.variables) {
-		set_variable(var.first, var.second);
+	for (auto &var : other.variables) {
+		set_variable(var.first, std::move(var.second));
```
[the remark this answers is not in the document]
thehans: This `take_variables` function was an optimization idea I had to avoid copies when and move variables from a temporary context into some other context, but I think t-paul replaced calls to this with the more traditional `apply_variables`, when he merged function-literals in value-rework-2.  
I'm guessing it may have possibly caused issues with closures keeping a context alive even if it was "temporary", and probably best to just not move things out of containers which requires too much hard thinking about cases like that.
Since it was no longer being called I just went ahead and removed the definition in #3102.

src/value.cc
thehans: flatten was rewritten again in https://github.com/openscad/openscad/pull/3102/commits/905884b916be5865cc1f7f8ca01262bfefbf7a03 for reason related to "embedded vectors".  
flatten is now basically one simple range-based for loop, utilizing a special iterator.

For reference, the discussion about embedded vectors starts around [my last paragraph here](https://github.com/openscad/openscad/pull/3102#issuecomment-652579183) while discussing how or where it would be safe to avoid copies of elements.

thehans:
Merged as #3102, closing.

Issue #2883, opened 2019-03-20
Crash with for loops and difference: Assertion `op->left() && op->right()' failed.


ali1234:
Code:

```
tri = sqrt(3)/2;

module holes() {
    union() for (x = [-10:11]) for (y = [-10:11])
        translate([x+(0.5*(y%2)), y*tri, 0]) cylinder(r=0.35, h=100, center=true);
}

module other_holes() {
    union() for (x = [-30:31]) for (y = [-30:31])
        translate([0, 0, y*tri]) rotate([90, 0, 6*x + (3*(y%2))]) cylinder(r=0.35, h=100, center=true);
}

module basket() {
    difference() {
        cylinder(r=15, h=15);
        translate([0, 0, 1]) cylinder(r=14.5, h=20);
        intersection() {
            scale(3) holes();
            translate([0, 0, -1]) cylinder(r=14, h=20);
        }
        intersection() {
            scale(3) other_holes();
            translate([0, 0, -1]) cylinder(r=14, h=20);
        }
    }
}

basket();
```

Above code crashes git master (2a697e4090cf) and 2019.01-RC3.1:

```
OpenSCAD-2019.01-RC3.1-x86_64.AppImage: src/csgnode.cc:203: void CSGProducts::import(std::shared_ptr<CSGNode>, OpenSCADOperator, CSGNode::Flag): Assertion `op->left() && op->right()' failed.
```

rcolyer:
Notably this does not crash 15.03, making it a regression from there.

kintel:
My first hunch would be to check if #2343 may have introduced this.
@thehans Does this sound familiar?

rcolyer:
Here's a more minimal piece of code that reproduces the same bug.

```
difference() {
    cylinder(r=5, h=1);
    intersection() {
        for (x = [-4:3]) for (y = [3:5])
            translate([x, 0.2*y, 0]) cylinder(r=1, h=1);
        cylinder(r=1, h=1);
    }
    for (x = [-30:31]) for (y = [0:6])
        translate([0, 0, 0.2*y]) cylinder(r=1, h=1, center=true);
}
```

thehans:
Just from a quick test it doesn't crash if CSG elements limit is set high enough. (bumping default from 100,000 to 150,000 works for both examples).
So it seems that its passing the limit momentarily during normalization, but not cleaning up properly on abort. Also possibly not counting the nodes correctly during this process.

thehans:
[...]

PR #2886, opened 2019-03-21
Rewrite iterative CSGTreeNormalizer::normalizePass to be comprehendible and correct.

thehans:
I ended up ditching my original iterative solution since it was really too complex to try to understand and maintain.  This new one is still iterative and manages its own stack but I think its much clearer and deviates minimally from the original recursive implementation.


thehans:
For issue #2883

Issue #2877, opened 2019-03-17
Add command line option to specify the export format


t-paul:
Currently the export format is determined by looking at the extension of the file given to the `-o` option (e.g. `-o export.stl` will set the format to STL).

It would be useful to have an additional parameter that can force a specific export format in cases where the file name can't have the needed extension (e.g. when using fifos or pipes, see #616 for one use case).

Suggestion for the option: `--export-format`

So if the option is given, the export format is forced ignoring the file name completely. Ideally the code should be changed to be a bit more driven by a declared list of acceptable format, removing the currently used long list of if/else constructs.

RomitKumar:
What should be done when both -o and --export-format option is used? Maybe report a error.

t-paul:
As mentioned in the last paragraph, I think that giving the options should always force what format is selected regardless of how the file is named. I'm not sure if it makes sense to add a warning in case of obvious collisions, but then the user is asking for this specifically, so I tend to say we don't want a warning.

RomitKumar:
So, when issuing commands like:
1. `openscad --export-format stl -o foo.png bar.scad`
A foo.stl file should be made
2. `openscad --export-format stl bar.scad`
A untitled.stl file should be made. 
3. `openscad --export-format stl -o foo.png bar.scad | MySTLFileProcessor`
A foo.stl file should be made and passed to MySTLFileProcessor
4. `openscad --export-format stl bar.scad | MySTLFileProcessor`
A untitled.stl file should be made and passed to MySTLFileProcessor. (And maybe untitled.stl should be removed or maybe no untitled.stl file is generated in the first place and the bit stream is directly passed to MySTLFileProcessor?)

t-paul:
1. If the user gives strange instructions we probably should still follow those, so that would actually create a STL file named `foo.png`
2. Good question, normally that would start the GUI as there is no `-o` given. So we can either continue doing that ignoring the `--export-format` or give an error for incompatible actions. I'm leaning towards the starting the GUI option.

I would not cover the pipe stuff just yet, there's a separate issue as followup. The common way to do this seems to be a special file name like `-o -`.

RomitKumar:
Made a fix for this issue at PR #2887

donbright:
IN my opinion this is bad UI. There are some basic guidelines that can be followed to determine answers to questions like this, some good old examples are here: 

https://modelessdesign.com/backdrop/401

[...]

PR #2887, opened 2019-03-22
Added --export-format option to command line argument

RomitKumar:
Fix for issue #2877 

The --export-format option will overwrite the file extension provided in -o option. If no -o option is given with --export-format option, --export-format is ignored

### DIFF
--- a/src/export.h
+++ b/src/export.h
@@ -21,6 +21,22 @@ enum class FileFormat {
 	NEF3
 };
 
+enum class ExportFileFormat {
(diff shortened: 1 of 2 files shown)


t-paul:
One of the errors on AppVeyor is:
```
Can't open file " \u0494" for export
```
Which sounds a bit like there's some variable not initialized and that is causing wrong turns in some conditions.

t-paul:
Imported as branch in #2890.

PR #2890, opened 2019-03-23
 Added --export-format option to command line argument

t-paul:
Imported #2887 as branch.


src/openscad.cc
```diff
@@ -857,11 +858,13 @@ int main(int argc, char **argv)
...
 	const char *output_file = nullptr;
 	const char *deps_output_file = nullptr;
+	const char *export_format = nullptr;
 	
 	ViewOptions viewOptions{};
 	po::options_description desc("Allowed options");
 	desc.add_options()
-		("o,o", po::value<string>(), "output specified file instead of running the GUI, the file extension specifies the type: stl, off, amf, csg, dxf, svg, png, echo, ast, term, nef3, nefdbg\n")
+		("o,o", po::value<string>(), "output specified file instead of running the GUI, the file extension specifies the type: stl, off, amf, csg, dxf, 3mf, svg, png, echo, ast, term, nef3, nefdbg\n")
+		("export-format", po::value<string>(), "format of exported scad file, arg can be any of file extension in -o option. It overwrites the file extension in -o option\n")
```
kintel: s/overwrites/overrides/

src/openscad.cc
```diff
@@ -1044,6 +1047,21 @@ int main(int argc, char **argv)
 		arg_colorscheme = vm["colorscheme"].as<string>();
 	}
 
+	ExportFileFormatOptions exportFileFormatOptions;
+    if(vm.count("export-format")) {
+        const char *tmp_format = vm["export-format"].as<string>().c_str();
+        for(auto ff: exportFileFormatOptions.exportFileFormats) {
```
kintel: Since exportFileFormats is a map, you can just look this up directly; no need for a loop.

src/openscad.cc
```diff
@@ -269,62 +269,64 @@ void set_render_color_scheme(const std::string color_scheme, const bool exit_if_
...
-		PRINTB("Unknown suffix for output file %s\n", output_file);
+	if(export_format) {
+		PRINT("Using extension of --export-format option");
+		extsn = export_format;
+	}
+	else { 
+		auto suffix = fs::path(output_file_str).extension().generic_string();
+		suffix = suffix.substr(1);
+		boost::algorithm::to_lower(suffix);
+		for(auto ff: exportFileFormatOptions.exportFileFormats) {
```
kintel: Same comment as below; we shouldn't need a loop here

src/openscad.cc
```diff
@@ -448,46 +450,46 @@ int cmdline(const char *deps_output_file, const std::string &filename, const cha
 
 		fs::current_path(original_path);
 
-		if (stl_output_file) {
-			if (!checkAndExport(root_geom, 3, FileFormat::STL, stl_output_file)) {
+		if (curFormat == ExportFileFormat::STL) {
```
kintel: Now that we have a mapping, we don't really need this huge if statement; we should be able to collect the similar calls and pass the format and filename arguments directly and thus simplify this logic (and duplication) a bit.

kintel:
@RomitKumar See above for comments

RomitKumar:
Sorry for the late reply. I was expecting a mail for any conversation on this thread but for some reasons I didn't receive any. I will implement your suggestions soon.

RomitKumar:
Made PR #2901 
Changes done in PR:
1. Removed similar if statements 
2. Used find function to check presence of key in map
3. Removed redundant enum ExporFileFormat and added the extra formats to enum FileFormat (suggest if I should revert back to previous implementation)
4. Removed unnecessary PRINTB statements

PR #2896, opened 2019-03-25
Updated the test procedure to match reality.

nophead:


nophead:
Yes they all pass due to #2821 and #2843.

PR #2906, opened 2019-03-30
Fixes #2784.

### LINKED ISSUE #2784: Crash from misc/escape-test.scad
I noticed this from manually trying out some of the tests as I was refactoring for #2783.

It only happens if you load the file `testdata/scad/misc/escape-test.scad`, hit F5, then F6.  Unfortunately there isn't any test script that is made to duplicate this exact case.

OpenSCAD exits and I see this output on the command line:
```
openscad: src/CGALRenderer.cc:48: CGALRenderer::CGALRenderer(std::shared_ptr<const Geometry>): Assertion `ps->getDimension() == 3' failed.
Aborted
```

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/69957392-crash-from-misc-escape-test-scad?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

nophead:
The bug was caused by passing 0 to the PolySet constructor when the file type is unknown. That gave a zero dimension object. The rendering code assumes it will be 2 or 3.

The simplest test is just
```import();```

F5 and then F6.

### DIFF
--- a/src/import.cc
+++ b/src/import.cc
@@ -202,7 +202,7 @@ const Geometry *ImportNode::createGeometry() const
 #endif
 	default:
 		PRINTB("ERROR: Unsupported file format while trying to import file '%s', import() at Line %d", this->filename % loc.firstLine());
-		g = new PolySet(0);
+		g = new PolySet(3);
 	}
 
 	if (g) g->setConvexity(this->convexity);
--- a/testdata/scad/templates/import_dxf-tests-template.scad
+++ b/testdata/scad/templates/import_dxf-tests-template.scad
@@ -1,5 +1,5 @@
-import();
 import("notfound.dxf");
+import();
 translate([-210,0,0]) import(file="../../../dxf/polygons.dxf");
 translate([-210,0,0]) import(file="../../../dxf/polygons.dxf", origin=[0,110]);
 translate([-210,0,0]) import(file="../../../dxf/polygons.dxf", origin=[110,110], scale=0.5);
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by nophead** [comment]
There are some test failures because import("") now produces a 3D object and as it is first in the DXF tests it makes the following 2D object be ignored. A simple fix is to swap the order so it is not first but I found that the test file is in the .gitignore file. Why is that?

```
testdata/scad/2D/features/import_dxf-tests.scad
testdata/scad/3D/features/import_stl-tests.scad
```

t-paul:
There are a couple of test cases that are generated from template files I think, maybe those are too?

kintel:
Wait, how did this work earlier? Did we ignore zero-dimension polysets and we no longer do?

nophead:
I am not sure exactly because it only failed with F5 before F6, so the
tests never saw the problem.

kintel:
ok, I was just wondering if this was some corner case we needed to keep in mind for backwards compatibility, but I guess not. Changing the templated tests would work

nophead:
I don't think we have to worry about backwards compatibility. I presume the
test is just to ensure import() doesn't crash. I can't see it mattes that
it now returns an empty 3D PolySet instead of a malformed PolySet.

PR #2907, opened 2019-03-31
Kill trusty CI support

kintel:


kintel:
@thehans This may make your life easier

PR #2926, opened 2019-04-07
Replace new/delete with std::vector

vazgriz:
This PR removes a few instances of new/delete pairs with equivalent usages of std::vector. This makes the code base a little more modern, although it has no effect on performance or correctness.

### DIFF
--- a/src/primitives.cc
+++ b/src/primitives.cc
@@ -424,7 +424,7 @@ const Geometry *PrimitiveNode::createGeometry() const
 		g = p;
 		if (this->r1 > 0 && !std::isinf(this->r1)) {
 			struct ring_s {
-				point2d *points;
+				std::vector<point2d> points;
 				double z;
 			};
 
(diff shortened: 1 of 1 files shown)


src/primitives.cc
```diff
@@ -507,11 +502,11 @@ const Geometry *PrimitiveNode::createGeometry() const
...
 
-			auto circle1 = new point2d[fragments];
-			auto circle2 = new point2d[fragments];
+			auto circle1 = std::vector<point2d>(fragments);
+			auto circle2 = std::vector<point2d>(fragments);
 
-			generate_circle(circle1, r1, fragments);
-			generate_circle(circle2, r2, fragments);
+			generate_circle(circle1.data(), r1, fragments);
+			generate_circle(circle2.data(), r2, fragments);
```
kintel: To further improve this, do you think it would be better and/or safer to pass an iterator (e.g. from the std::vector back_inserter()) instead of passing a raw pointer?

vazgriz: That depends on if you wanted to use `generate_circle` with other containers. I would personally keep it as-is, and would only change it later if necessary.

kintel: an iterator would be container agnostic, but we'd have to choose an iterator. I'll keep it like this

Issue #2882, opened 2019-03-20
Allow trailing comma in let() or lists


adrianVmariano:
Other languages like C++ and Python allow a trailing comma in contexts where comma separated lists occur.  This makes it easier to move code around, or add and remove things.  So for example it would be nice to be able to write

    let(
        a=<expression>,
        b=<expression>,
        c=<expression>,
    )

and then if I deleted the third expression, or exchange the order the code still works.  Similarly, it would be nice if I could write

    points =  [
               [1,1],
               [2,2],
               [3,3],
              ]

for the same reason.

kintel:
The second example is already possible, no?

I think it makes sense to extend to other list-like constructs, unless we find conflicting hidden bugs/features that people are abusing : /

adrianVmariano:
Hmm.  Yes, the second example does work.  I thought I had checked it.  The one that's really been driving me crazy is the last comma in let().  

The only other context I can think of is argument lists to functions and modules, and it would probably be nice to allow it there as well.   Are there any other list-like constructs?

adrianVmariano:
I must say that the comma situation in let() statements remains the cause of about 75% of my syntax errors.   It looks like somebody fixed this an issues a PR for it.

PR #2928, opened 2019-04-09
allow trailing commas for function call such as let()

leonplust:
Fix issue https://github.com/openscad/openscad/issues/2882. 

For this PR I got very important help from t-paul, I was working toward a wrong direction before. 

Besides allowing trailing comma, considering his advice, I changed arguments to argument_list making the difference is more noticeable. I personally found it hard to notice when I first started working on it. 

Also, because definition_arguments is being used in other files, I am not sure if I should change all of them to keep consistency so I had it remain the same for now. 

I am not sure if I should open another PR for this or not. Please let me know if any change is needed.


leonplust:
Since it's originally requested for let(), I created a test case based on let-module-tests, adding trailing comma for all let(). Is this good for the test case?

leonplust:
I have made some fixes and created a new test case for function calls. The previous test case has been renamed to trailing-comma-module-test.

leonplust:
I have removed those duplicate test and added the test case to the individual test

t-paul:
Superseded by #3814.

Issue #2950, opened 2019-05-19
echo() formating error


RonaldoCMP:
I have found this:

echo(1.00006e-20); //ECHO: 1.00006e-2
echo(1.00000e-20);//ECHO: 1e-20

using:

Windows 7/64
OpenSCAD 2019.01-RC4

t-paul:
Confirmed on Debian/Testing. Same output.

thehans:
Ah you found a bug in my "trim trailing zeroes" code.  I just made a quick patch for it.  The code is a little bit weird because for efficiency, I originally tried to loop through the output string no more than once.  Now with the added check for exponent, it has to scan the string in sort of redundant way now.  It could probably be written better, but I don't have time at the moment to work out another method.

I still wish the main double-conversion library supported this but the maintainer didn't think it was important.
https://github.com/google/double-conversion/issues/87

kintel:
Fixed in #2957

PR #2951, opened 2019-05-19
Fix trim of trailing zero on exponent

thehans:
Issue #2950


kintel:
@thehans Would it be easy to add a simple test for this, just so nobody refactors the bug back into existence in the future?

thehans:
Test added, and I commented in #2957

kintel:
Solved by #2957

PR #2955, opened 2019-05-26
Implement multi-tab feature in TabManager class

RomitKumar:
The commit currently has basic gui implemented. Editing features like commenting, etc needs to be handled for each separate tab.
This commit will fail the checker since support for legacy editor is removed.


t-paul:
Ok, going though the module tests now...
- [ ] Test 1: Basic cache
- [ ] Test 2: MCAD
- [ ] Test 3: Overload USEd module
- [ ] Test 4: Recursive USE
- [ ] Test 5: Circular USE
- [ ] Test 6: Dependency tracking of common file USEd by multiple modules
- [ ] Test 7: Dependency tracking of file included from module
- [ ] Test 8: Circular include
- [ ] Test 9: Missing include file appears
- [ ] Test 10: Missing include file in subpath appears
- [ ] Test 11: Missing library file appears
- [ ] Test 12: Automatic reload of cascading changes
- [ ] Test 13: Correct handling of compile errors in auto-reloaded modules
- [ ] Test 14: Dependency tracking of underlying dependencies
- [ ] Test 15: Dependency tracking with two open files
- [ ] Test 16: Correct auto-reload of errors in includes from main
- [ ] Test 17: Correct auto-reload of errors in includes from sub modules

t-paul:
Looks like there's some regressions with auto-reload (or maybe just debug output going nuts?), but all happen in the nightly too, so they are not introduced here.

PR #2957, opened 2019-05-30
A possible way to solve "echo() formating error #2950" (trailing zeroes in output from echo)

henwist:
Hi! 
I have a suggestion on how to solve the formatting error(#2950) when trailing zeroes are output by echo.
**Tested by inserting these lines in OpenSCAD editor:**

`echo(1.00006000e-20); //ECHO: 1.00006e-2
echo(1.009000e-20);//ECHO: 1e-20
echo(1.10e+34);
echo(000.93400e-23);
echo(4.003216e+10);

echo(1.00006e-20); //ECHO: 1.00006e-2
echo(1.00000e-20);//ECHO: 1e-20

echo(10);
echo(-10);

echo(-0.34e-22);`

**output from OpenSCAD:**
`ECHO: 1.00006e-20
ECHO: 1.009e-20
ECHO: 1.1e+34
ECHO: 9.34e-24
ECHO: 4.00322e+10
ECHO: 1.00006e-20
ECHO: 1e-20
ECHO: 10
ECHO: -10
ECHO: -3.4e-23`

//Best Regards, Henrik Wistbacka


t-paul:
Thanks for the PR. I do like the nicely structured and commented implementation. How does your solution compare to the other proposed fix in #2951? The comments in #2950 mention the original implementation was focussing also quite a bit on performance (due to the fact the whole double-conversion library was added to improve performance).

kintel:
@thehans Comments on this?

henwist:
I tested two variants of data - see downwards. Both the proposed fix(#2951) as well as this PR was close to each other regarding executing time for ten million lines of echo(value_with_trailing_zeroes_and_optional_exponent) in OpenSCAD script.

-----
-----
example data that the performance test was applied on:

exponent and  many trailing zeroes
echo(0.017378000000000000e+250);
echo(5.007265000000000000e+285);
echo(-6.010508000000000000e+121);
echo(-2.030840000000000000e+72);
echo(4.004671000000000000e+203);
echo(-3.009760000000000000e+248);
echo(-5.008051000000000000e+202);
echo(3.021301000000000000e+50);



no expoent and many zeroes:
echo(6.467000000000000000);
echo(2.454000000000000000);
echo(2.722000000000000000);
echo(6.073000000000000000);
echo(-7.233000000000000000);
echo(-7.006000000000000000);
echo(-1.120000000000000000);
echo(-7.799000000000000000);
echo(2.444000000000000000);

-----





----
----
current version:

exponent and  many trailing zeroes
real    2m58.521s
user    2m54.484s
sys     0m2.785s

no expoent and many zeroes:
real    2m43.887s
user    2m33.916s
sys     0m2.664s
----



----
----
this PR:

exponent and  many trailing zeroes
real    2m52.728s
user    2m47.390s
sys     0m2.725s


no expoent and many zeroes:
real    2m34.612s
user    2m29.485s
sys     0m3.012s
---

So all in all - both implementations are almost equal regarding execution time, but this PR is slightly faster.
Note: Sorry about the formatting in bold - have no idea why it is bold...

MichaelAtOz:
> Note: Sorry about the formatting in bold - have no idea why it is bold...

[Markdown.](https://daringfireball.net/projects/markdown/syntax)

 I removed the `----` under those paragraphs.

```
Headers

Markdown supports two styles of headers, Setext and atx.

Setext-style headers are “underlined” using equal signs (for first-level headers) and dashes (for second-level headers). For example:

This is an H1
=============

This is an H2
-------------

Any number of underlining =’s or -’s will work.
```

src/value.cc
```diff
@@ -61,45 +61,72 @@ const ValuePtr ValuePtr::undefined;
...
-            return;
-          }
-        }
-      } else {
-        // no zeroes after decimal
-        return;
-      }
-    }
-  }
+  char *exppos = strchr(buffer, DC_EXP);
```
thehans: I would move this inside the if block below, and start from decimal, instead of buffer to save a few cycles.

thehans:
> @thehans Comments on this?

Looks ok to me.  I made one small code review comment.

I also just updated my PR #2951 with a test case, feel free to copy that change into your PR @henwist .

For what its worth I tried performance testing on some "normal" data (not particularly heavy with zeroes, exponents etc.) and could not find any difference between my patch and this one (negligible +/- 2s over ~4 minutes "benchmark").
No problem closing my PR if this one is preferred, it was just easier for me to add the test case there than to add a PR on top of this PR.

src/value.cc
henwist: Thanks, that is true, I moved the code to the suggested position.

henwist:
@thehans Good  - I applied your patches regarding the tests for echo as you suggested to my repository.

kintel:
Let's try this one. It's less magical :)

PR #2961, opened 2019-06-03
Fix misc. typos in documentation and source comments

luzpaz:
Found via `codespell`

### DIFF
--- a/tests/test_pretty_print.py
+++ b/tests/test_pretty_print.py
@@ -475,7 +475,7 @@ def main():
     builddir = ezsearch('--builddir=(.*?) ', ' '.join(sys.argv) + ' ')
     if not builddir or not os.path.exists(builddir):
         builddir = os.getcwd()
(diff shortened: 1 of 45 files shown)


tests/test_pretty_print.py
```diff
@@ -475,7 +475,7 @@ def main():
     builddir = ezsearch('--builddir=(.*?) ', ' '.join(sys.argv) + ' ')
     if not builddir or not os.path.exists(builddir):
         builddir = os.getcwd()
-        print('warning: couldnt find --builddir, trying to use current dir:', builddir)
+        print('warning: couldn't find --builddir, trying to use current dir:', builddir)
```
t-paul: This causes the tests to fail as the outer quotes are single `'` too.

t-paul:
Very nice, thanks. There's one change in one of the python files which causes a syntax error. Could you please update that?

tests/test_pretty_print.py
luzpaz: a6a5f62

Issue #2971, opened 2019-06-16
stroke-linejoin not attributed in SVG files when importing it - a Feature Request


henwist:
I noticed that paths imported from Inkscape into OpenSCAD as SVG, containing stroke-linejoin, (Round, Bevel and Miter),  

> stroke-linejoin:round;
stroke-linejoin:bevel;
stroke-linejoin:miter;

are not showing up in latest development version. 
From the picture below (an SVG original), drawn in Inkscape, you can see how the expected outcome should look like:

--------
![line_join_expected](https://user-images.githubusercontent.com/12303430/59567779-6b288580-9072-11e9-8d4a-21df54e9882a.png)
--------

OpenSCAD renders this shown below,  when importing the SVG file (attached as zip below):

------
![line_join_actual](https://user-images.githubusercontent.com/12303430/59567863-83e56b00-9073-11e9-86d8-6f709aa786e1.png)
------

So the stroke-linejoin between segments of the path is missing in all the paths rendered by OpenSCAD latest development version to date. 
This issue is a Feature Request to include the stroke-linejoin attribute, when importing the SVG into OpenSCAD by the following lines:

----

> linear_extrude(height=20)
import(file="/home/henrik/programming/openscad_2019_05/issue/#2971/line_join.svg");

-----

Attached file to test the issue:
[line_join.zip](https://github.com/openscad/openscad/files/3294451/line_join.zip)

//Best Regards, Henrik

PR #2972, opened 2019-06-16
Commit that attribute stroke-linejoin in SVG files in issue #2971 - a…

henwist:
… Feature Request.


t-paul:
Thanks, looks good. I've added a test case for both line-cap and the new line-join properties.

Issue #2962, opened 2019-06-04
Bug in Find feature of OpenSCAD


RomitKumar:
The count of matched words in Find option displays wrong value. Steps to reproduce the error:
1. Launch Openscad and clear the contents of editor(if any) 
2. Press Ctrl+f to view find panel and clear the contents of text field in find panel 
3. Press 'a' 9 times in the editor
4. In text field of search panel, write 'a'(count of matched words is 9)
5. In editor write on more 'a' (now there are 10 'a' in editor)
6. In text field of search panel, append another 'a' such that 'aa' is being searched for(count of matched words is 5 now)
7. In text field of search panel press a backspace, so 'a' is again being searched for
However the count of matched words shown is 1

henwist:
I can confirm the bug - but the commit 92550dece3b0d15edae2982f80a0b344dc7d0640 (just a few days ago) in master don't count to ten when another 'a' is added in the editor.

RomitKumar:
The find in OpenSCAD does not do real time finding. So the count of matched word does not update automatically. What I meant by step 5 was that there are 10 'a' in editor. Updated the issue to make it clear.

kintel:
Fixed by #2974

Issue #2829, opened 2019-03-01
GUI testbed


MichaelAtOz:
There is the likes of [Autohotkey](https://www.autohotkey.com/), for Windows. Unfortunately not multi-platform.
It is not specifically designed as a testbed, but is full screen-scraper GUI driver capable (and much more).
I used to be fluent, it's... 'non-standard' may be the best description of the language.

_Originally posted by @MichaelAtOz in https://github.com/openscad/openscad/issues/2825#issuecomment-468472862_

MichaelAtOz:
@thehans 
> Even some basic tests

List a few, I'll give it a go.

MichaelAtOz:
...all though, I'll have to double check, I think QT way not list the Window components well. But maybe a MSVC build will be friendlier.

thehans:
By basic tests, I meant those such as the test from the original discussion(It looks like the we have these under `recursion-test-module.scad` and `recursion-test-function2.scad`), where we just want to know if it crashed or not, and aren't necessarily testing clicking through GUI menus/interaction etc.

So this sort of test would ideally not require any GUI scripting, if we can have some command line parameter that forces GUI creation (call it `-force-gui` or some such) when you are running command line that would normally just export without creating a window.  Also this implies that implementation of such a flag would still automatically exit the program after export was complete.

Interface scripting may still be useful for actually verifying that particular options/menus etc work as expected but this was more an idea to get something up and running faster without integrating a whole new third party tool, maybe as a stepping stone towards that tool though.

MichaelAtOz:
It would be easy to load a .scad file, do F5 &&/|| F6 (or any menu or keystrokes), monitor the window to see if the below happens
![capture recursion crash](https://user-images.githubusercontent.com/1972961/53612105-20027800-3c25-11e9-8e37-d577febabe0b.JPG)
If so capture the crash details and click 'close the program' button.
If not grab the console output, either store it or capture some specific bits and save them too.

I'm going to do a simple prototype, because one of the things I'm concerned about is performance changes, ie worse CPU or memory creeping into changes. We have nothing for that ATM.

thehans:
Yeah I forgot about crash dialogs, maybe it is a necessity to have such a scripting engine just to close that dialog.

t-paul:
[...]

PR #2974, opened 2019-06-18
Fixes issue #2962, by setting the minimum size of fieldLabel to accom…

henwist:
…odate

all digits from the search, that was hidden (truncated) in the field.

### DIFF
--- a/src/QWordSearchField.cc
+++ b/src/QWordSearchField.cc
@@ -16,6 +16,7 @@ QWordSearchField::QWordSearchField(QFrame *parent) : QLineEdit(parent)
 	auto minsize = minimumSizeHint();
 	setMinimumSize(qMax(minsize.width(), fieldLabel->sizeHint().height() + frameWidth * 2 + 2),
 								 qMax(minsize.height(), fieldLabel->sizeHint().height() + frameWidth * 2 + 2));
+	fieldLabel->setAlignment(Qt::AlignRight);
 }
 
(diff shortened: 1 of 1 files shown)


henwist:
I tested up to 11746800 'a'  the form

>  aaaaaaaaaa
aaaaaaaaaa
...
 in the editor and searched for down to one 'a' from ten 'a' in the search field and the count of 'a' seemed correct for all searches.

Is there a way to remote control the editor from the console (e.g. ./openscad --search='a' aaa.scad) and save a picture of how the correct searchfield should look like and write a test case for that?

t-paul:
There's no GUI test framwork at this point. Some ideas are discussed in #2829. I'm still hoping to use the accessibility features at some point which seem to have improved in Qt5/QScintilla. I did try a bit with [accerciser](https://help.gnome.org/users/accerciser/stable/introduction.html.en) which shows promising results, but I did not have enough time to dig deeper yet.

MichaelAtOz:
The #2829 method could technically drive the search then take a screen capture, but the process of comparing the results is hairy and probably would only be valid on specific (ie the same PC/setup) system configurations.
Capturing the field with the count *may* be possible. This sort of testing, visual GUI behaviour, will be tedious, compared to testing .scad code via the GUI, checking console results etc.

> accessibility features at some point which seem to have improved in Qt5/QScintilla

Not on windows.

PR #2889, opened 2019-03-23
 Autocomplete / CallTips for QScintilla

t-paul:
Rebase of #905 / Initial demo of AutoComplete / CallTips feature that is provided by Scintilla.

Things to decide:
- How to format the CallTips display? What information to present?
- What options are needed
  - enable/disable AutoComplete
  - AutoComplete threshold
  - enable/disable CallTips
  - ...

ToDo:
- [ ] Complete list of keywords
- [ ] Move data to separate text file, currently it's just in code

AutoComplete / CallTips
screencast

UserList providing a set of templates to insert at the cursor position (currently triggered by CTRL-I)
screencast

PR #2975, opened 2019-06-23
Keyword autocompletion

RomitKumar:
The base code is derived from pr #2889 .


openscad.pro
```diff
@@ -194,10 +194,7 @@ CONFIG += hidapi
 CONFIG += spnav
 CONFIG += double-conversion
 
-#Uncomment the following line to enable the QScintilla editor
-!nogui {
```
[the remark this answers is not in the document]
RomitKumar: I am not sure if this is the perfect solution, however, I don't see much harm in it. This change instructs to always enable scintilla. I think that failing the travis check without this change is more of a travis configuration issue since the application builds nicely on my local ubuntu system.

src/control.cc
```diff
@@ -339,13 +339,55 @@ AbstractNode *ControlModule::instantiate(const Context* ctx, const ModuleInstant
...
-	Builtins::init("children", new ControlModule(ControlModule::Type::CHILDREN));
-	Builtins::init("echo", new ControlModule(ControlModule::Type::ECHO));
-	Builtins::init("assert", new ControlModule(ControlModule::Type::ASSERT));
 	Builtins::init("assign", new ControlModule(ControlModule::Type::ASSIGN));
-	Builtins::init("for", new ControlModule(ControlModule::Type::FOR));
-	Builtins::init("let", new ControlModule(ControlModule::Type::LET));
-	Builtins::init("intersection_for", new ControlModule(ControlModule::Type::INT_FOR));
-	Builtins::init("if", new ControlModule(ControlModule::Type::IF));
+
+	Builtins::init("child", new ControlModule(ControlModule::Type::CHILD),
```
t-paul: As `child()` is deprecated, we probably don't want to show calltips same as with the `import_*` modules.

src/func.cc
```diff
@@ -1020,41 +1020,194 @@ ValuePtr builtin_is_string(const Context *ctx, const EvalContext *evalctx)
...
+				});
+
+	Builtins::init("ln", new BuiltinFunction(&builtin_ln),
+				{
+					"ln(number) -> number",
+				});
+
+	Builtins::init("str", new BuiltinFunction(&builtin_str),
+				{
+					"str(number|string, ...) -> string",
```
t-paul: I think we should have only one way of describing things, either in multiple lines or with that `number|string` way. As most other entries have separate entries, lets start with that.

src/func.cc
```diff
@@ -1020,41 +1020,194 @@ ValuePtr builtin_is_string(const Context *ctx, const EvalContext *evalctx)
...
+				});
+
+	Builtins::init("sign", new BuiltinFunction(&builtin_sign),
+				{
+					"sign(number) -> -1, 0 or 1",
+				});
+
+	Builtins::init("rands", new BuiltinFunction(&builtin_rands),
+				{
+					"rands(min, max, num_results) -> array",
```
t-paul: Array and Vector are used a bit mixed in various places. For the calltips I think we should use vector as it seems more commonly used.

t-paul:
Don't worry about the merge conflicts, we can resolve those when actually merging the PR.

PR #3004, opened 2019-07-22
added example of condensed matter rendering.

clarkebowers:
Added an example that shows a complex atomic model using the preview.

### DIFF
--- a/examples/Advanced/Atoms.scad
+++ b/examples/Advanced/Atoms.scad
@@ -1,0 +1,203 @@
+//Condensed matter
+//All measurements are in Ångstrom
+
+//OpenSCAD
+$fn = 40;  //sections per circle
+echo(version=version());
+
+//Liquid nitrogen (N2)
+AtomicRadiusN2 = 0.65;
+BondPairN2 = 1.197; 
+ColorN2 = "Silver";
+echo(pow(BondPairN2,1/3));
+
+//Silicon (Si)
+AtomicRadiusSi = 1.1;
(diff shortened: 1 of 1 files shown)


kintel:
While more examples are nice, I feel that this particular example misses on two topics:
1. It's very slow to render
2. It's not clear what OpenSCAD features or design pattern it's trying to exercise

clarkebowers:
> While more examples are nice, I feel that this particular example misses on two topics:
> 
> It's very slow to render
> It's not clear what OpenSCAD features or design pattern it's trying to exercise

I wrote this model to print on a 3D PLA printer. I wanted students to be able to hold a lattice network in their hand to better understand condensed matter. The model renders beautifully in preview mode, and quickly (less than 1 second). See the output below. Unfortunately, I cannot generate the STL to print it. It takes more than 24-hours. I added this example as a benchmark as efforts are made to improve the GeometryEvaluator. We could test how much it improves the rendering. 

Normalized CSG tree has 5019 elements
Compile and preview finished.
Total rendering time: 0 hours, 0 minutes, 0 seconds

kintel:
As a performance test that makes sense. We don't really have a framework for that yet, except we have some tests tagged as "heavy" that can be enabled on demand: https://github.com/openscad/openscad/blob/master/tests/CMakeLists.txt#L623

We could create a new folder for future tests, and write a small README snipped for each file, but that might just bury the problem. Another idea could be to open a performance issue and link to a gist of this file.

PR #3006, opened 2019-07-23
DXF import

t-paul:
This tracks the work in https://github.com/leonplust/openscad.git branch `dxf-import`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -486,12 +486,14 @@ set(CORE_SOURCES
 
 set(NOCGAL_SOURCES
   src/builtin.cc 
+  src/dxf.cc
   src/import.cc
   src/import_3mf.cc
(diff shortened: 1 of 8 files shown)


olemis:
@t-paul what needs be done so as to finally merge this PR ? should BRL-CAD code clean up be started from scratch instead?

t-paul:
Feature wise it should be relatively ok as all the test were passing, but it would be nice to convert the remaining C style code to C++. In addition it may need some more manual testing with existing files.

olemis:
> Feature wise it should be relatively ok as all the test were passing, but it would be nice to convert the remaining C style code to C++. 

Unfortunately I do not have (yet) a development environment setup for this .

> In addition it may need some more manual testing with existing files.

Should you need any help with this please let me know.

PR #3019, opened 2019-07-29
Use shared_ptr for Context::document_path

jschobben:
While working on another PR to improve tail recursion, I noticed that creating a Context can be faster; the document_path member was always copied.

This wraps it in a shared_ptr. The speedup is about 10% to 15% when doing lots of Context creation.

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -59,6 +59,7 @@ Context::Context(const Context *parent)
 	}
 	else {
 		this->ctx_stack = new Stack;
+		this->document_path = std::make_shared<std::string>();
 	}
 
 	this->ctx_stack->push_back(this);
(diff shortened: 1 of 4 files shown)


src/context.cc
```diff
@@ -59,6 +59,7 @@ Context::Context(const Context *parent)
 	}
 	else {
 		this->ctx_stack = new Stack;
+		this->document_path = std::make_shared<std::string>();
```
t-paul: I'm hesitating a bit due to this creating an empty `shared_ptr`. It does not seem to be dereferenced anywhere as the tests are green. Do you see a way to make it more obvious that the value is always set, e.g. by passing it along in the constructor? It still needs a setter for overwriting for the `import` modules though.

jschobben: It makes a shared_ptr to an empty string, if I'm not mistaken. Maybe I can make that more obvious like this: ```std::make_shared<std::string>("")```? Let's see if it can be set in the initializer list.
W.r.t. the setter, how do you mean? ```Context::setDocumentPath``` is there, and in ```FileContext::initializeModule``` the shared_ptr reassignment works its magic.
I'll do some more testing.

t-paul: Ahh, right, looks like it does create an empty string in that case (https://godbolt.org/z/6jUVWf). I guess that makes it really identical to the old version minus the copy overhead.

PR #3022, opened 2019-07-31
Update working of No option in reload

RomitKumar:
The reload confirmation dialog box counter-intuitively turns off automatic reload on choosing No option. The pr removes this behavior and No option simply does not reload the file.

### DIFF
--- a/src/mainwin.cc
+++ b/src/mainwin.cc
@@ -1743,21 +1743,11 @@ void MainWindow::autoReloadSet(bool on)
 bool MainWindow::checkEditorModified()
 {
 	if (activeEditor->isContentModified()) {
-		QMessageBox msgBox(this);
-		msgBox.setIcon(QMessageBox::Warning);
-		msgBox.setWindowTitle(_("Application"));
-		msgBox.setText(_("The document has been modified.\n"
-						"Do you really want to reload the file?"));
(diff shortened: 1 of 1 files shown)


RomitKumar:
Are you suggesting to add different preference setting(checkbox) for auto-reload and manual reload? The auto-reload or manual reload works only if the corresponding checkbox is ticked.
My intended functionality is that "Never" option simply does the work of "No"  while "No" doesn't do anything.

kintel:
I'm not sure what the best solution is. Right now it's a bit confusing since "Never" won't work for manual reloads. Getting GUI like this right is tricky.

RomitKumar:
"Never" should only stop auto-reload. It can be strange when user triggers manual reload, and the file is not reloaded because of some preference setting set sometime in the past. I guess any manual reload should ideally reload the file irrespective of preference setting.
Yeah, the name "Never" can be confusing. I guess we can replace that with "No and stop auto-reload" or something more concise.

MichaelAtOz:
OP "counter-intuitively turns off automatic reload on choosing No option."
So how about just not doing that. Leave it to the user to sort out preferences.
Mostly auto-reload is used with external editors, it would be rare to want to change that too often.
or
Maybe "Disable auto-reload" instead of "Never". But I prefer the do nothing option.

PR #3024, opened 2019-08-02
Work-around for test failure at high -jN

knielsen:
So this isn't pretty, but it does provide a possible solution to an annoying problem. With this patch I am able to consistently run the openscad testsuite with -j8 without failures. Whereas without this I'll see random failures every 2-3 test run or something like that.

----

When running the testsuite against an Xvfb $DISPLAY with -jN, there
are occasional failures with "Unable to open a connection to the X
server". This appears to be a bug in Xvfb with rapid parallel connects.

This patch implements a work-around: The testsuite sets the
environment variable OPENSCAD_X_CONN_RETRY; this causes openscad to
retry the X connection in case of failure, which avoids the test
failure.

### DIFF
--- a/src/OffscreenContextGLX.cc
+++ b/src/OffscreenContextGLX.cc
@@ -256,11 +256,25 @@ bool create_glx_dummy_context(OffscreenContext &ctx)
 	int major;
 	int minor;
 	auto result = false;
+	int tries = 0;
+	auto dpyenv = getenv("DISPLAY");
+	auto retryenv = getenv("OPENSCAD_X_CONN_RETRY");
+	int connect_retry_count = retryenv ? atoi(retryenv) : 0;
 
-	ctx.xdisplay = XOpenDisplay(nullptr);
-	if (ctx.xdisplay == nullptr) {
+	for (;;) {
+		ctx.xdisplay = XOpenDisplay(nullptr);
+		tries++;
+		if (ctx.xdisplay != nullptr) {
+			break;
+		}
+		if (tries <= connect_retry_count) {
+			std::cerr << "Warning: XOpenDisplay(DISPLAY=\""
+				  << (dpyenv?dpyenv:"")
+				  << "\") failed, retrying...\n";
+			sleep(1);
+			continue;
+		}
 		std::cerr << "Unable to open a connection to the X server.\n";
-		auto dpyenv = getenv("DISPLAY");
 		std::cerr << "DISPLAY=" << (dpyenv?dpyenv:"") << "\n";
 		return false;
 	}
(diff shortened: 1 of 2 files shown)


src/OffscreenContextGLX.cc
```diff
@@ -256,11 +256,25 @@ bool create_glx_dummy_context(OffscreenContext &ctx)
...
+	auto retryenv = getenv("OPENSCAD_X_CONN_RETRY");
+	int connect_retry_count = retryenv ? atoi(retryenv) : 0;
+
+	for (;;) {
+		ctx.xdisplay = XOpenDisplay(nullptr);
+		tries++;
+		if (ctx.xdisplay != nullptr) {
+			break;
+		}
+		if (tries < connect_retry_count) {
```
t-paul: Shouldn't that be <= to actually have 3 retries with an env setting of 3? I wonder if we should clamp the value to a maximum of 10 or so?

knielsen:
Torsten Paul <notifications@github.com> writes:

> -	ctx.xdisplay = XOpenDisplay(nullptr);
> -	if (ctx.xdisplay == nullptr) {
> +	int tries = 0;
> +	auto dpyenv = getenv("DISPLAY");
> +	auto retryenv = getenv("OPENSCAD_X_CONN_RETRY");
> +	int connect_retry_count = retryenv ? atoi(retryenv) : 0;
> +
> +	for (;;) {
> +		ctx.xdisplay = XOpenDisplay(nullptr);
> +		tries++;
> +		if (ctx.xdisplay != nullptr) {
> +			break;
> +		}
> +		if (tries < connect_retry_count) {
>
> Shouldn't that be <= to actually have 3 retries with an env setting of
> 3? I wonder if we should clamp the value to a maximum of 10 or so?

Setting OPENSCAD_X_CONN_RETRY=3 with this code will try connecting a total
of 3 times before giving up. That's why it uses tries < connect_retry_count
- when tries is 2 it will try again, when tries is 3 it will give up and
flag the error. I guess it's a bit ambiguous if 3 connection attempts is 3
retries, or one try and 2 retries (the code implements the former).

I consider this very much an internal testsuite feature, so I don't think it
is terribly important how the semantics is - I've never before seen the need
to retry an X connection in practice. I'm happy to change this in whatever
way people prefer, such as try-or-retry semantics or clamping the value.

 - Kristian.

t-paul:
I just find it a bit odd if we set the variable called `*_RETRY` to 1 and then no retry happens, so the name does not match behavior.

knielsen:
> I just find it a bit odd if we set the variable called `*_RETRY` to 1 and then no retry happens, so the name does not match behavior.

Ack; I've now updated the pull request so that OPENSCAD_X_CONN_RETRY=N will do a total of (N+1) connection attempts before giving up with an error.

PR #3029, opened 2019-08-04
Corrected file name shown in export dialog box

RomitKumar:
The filename is of previously exported file in export dialog box. This pr corrects that.

### DIFF
--- a/src/mainwin.cc
+++ b/src/mainwin.cc
@@ -2968,19 +2968,19 @@ QString MainWindow::exportPath(const char *suffix) {
 	auto path_it = this->export_paths.find(suffix);
 	if(path_it != export_paths.end())
 	{
-		path = QFileInfo(path_it->second).absolutePath();
+		path = QFileInfo(path_it->second).absolutePath() + QString("/");
 		if(activeEditor->filepath.isEmpty())
-			path += QString(_("/Untitled")) + suffix;
+			path += QString(_("Untitled")) + suffix;
 		else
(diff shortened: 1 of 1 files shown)


src/mainwin.cc
```diff
@@ -2967,13 +2967,21 @@ QString MainWindow::exportPath(const char *suffix) {
 	QString path;
 	auto path_it = this->export_paths.find(suffix);
 	if(path_it != export_paths.end())
-		path = path_it->second;
+	{
+		path = QFileInfo(path_it->second).absolutePath();
+		if(activeEditor->filepath.isEmpty())
+			path += QString(_("/Untitled")) + suffix;
+		else
+			path += QString(_("/")) + QFileInfo(activeEditor->filepath).completeBaseName() + suffix;
```
t-paul: Not newly introduced with the change, but maybe we can drop the translation marker from `"/"` (here and in the other branch too) as this is not language dependent.

src/mainwin.cc
```diff
@@ -2968,19 +2968,19 @@ QString MainWindow::exportPath(const char *suffix) {
 	auto path_it = this->export_paths.find(suffix);
 	if(path_it != export_paths.end())
 	{
-		path = QFileInfo(path_it->second).absolutePath();
+		path = QFileInfo(path_it->second).absolutePath() + QString("/");
```
t-paul: Nice, that is even better.

PR #3032, opened 2019-08-08
Fix dangling pointer in "--export-format"

jschobben:
The --export-format option did not actually work for me; with this fix it does.
Also made some localized whitespace consistency fixes (in separate commit).

### DIFF
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -283,12 +283,12 @@ int cmdline(const char *deps_output_file, const std::string &filename, const cha
 	std::string extsn;
 	std::string output_file_str = output_file;
 	const char *new_output_file = nullptr;
-	
+
 	if(export_format) {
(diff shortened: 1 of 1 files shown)


src/openscad.cc
```diff
@@ -1009,15 +1009,15 @@ int main(int argc, char **argv)
 	}
 
 	ExportFileFormatOptions exportFileFormatOptions;
-    if(vm.count("export-format")) {
-        auto tmp_format = vm["export-format"].as<string>();
-        if(exportFileFormatOptions.exportFileFormats.find(tmp_format) != exportFileFormatOptions.exportFileFormats.end()) {
-        	export_format = tmp_format.c_str();
```
jschobben: Here a pointer to ```tmp_format``` was assigned, which then went out of scope.

Issue #616, opened 2014-01-30
Allow command line to write stl to stdout. [$5]


BenFrantzDale:
Instead of 

```
$ openscad -o foo.stl /tmp/foo.scad
$ mySTLProcessor < foo.stl
```

it would be nice to be able to do

```
$ openscad -o stl:- /tmp/foo.scad | mySTLProcessor
```

FWIW: `tar` uses `-t` for "write to stdout"; `gzip` uses `-c`, `--stdout`, or `--to-stdout` interchangeably. ImageMagic's `convert` uses "extension colon dash": e.g., `jpeg:-`.

See also: #615.
##

t-paul:
Thanks @aconz2, change is merged, bounty can be claimed.

aconz2:
thanks @t-paul for taking this across the finish line

aconz2:
@t-paul I believe the bounty can only be claimed after this issue is closed

t-paul:
Oops, sorry, I just check the merge notice here assuming it would auto close.

PR #3033, opened 2019-08-09
Don't replace extension of output file

jschobben:
Not sure what the expected behavior should be, but to me the current behavior seems strange.
Running:
```openscad input.scad --export-format stl -o /dev/null```
tries to write a STL file to ```/dev/null.stl```.

With this change, the filename given by "-o" is respected; probably the user knows what they are doing.

Side effect is that extension is no longer lower-cased now:
```openscad input.scad -o output.STL```
now produces ```output.STL``` instead of ```output.stl```.

### DIFF
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -303,7 +303,7 @@ int cmdline(const char *deps_output_file, const std::string &filename, const cha
 	}
 	
 	curFormat = exportFileFormatOptions.exportFileFormats.at(extsn);
-	std::string filename_str = fs::path(output_file_str).replace_extension(extsn).generic_string();
+	std::string filename_str = fs::path(output_file_str).generic_string();
 	new_output_file = filename_str.c_str();
 
 	set_render_color_scheme(arg_colorscheme, true);


t-paul:
Disucssion is in https://github.com/openscad/openscad/issues/2877 - The concern was that giving a file name extension that does not match the format, the result would be a bit strange. I'm not sure that is really required, if the user requests a STL file with file name `out.png` I'd be ok with just a warning. The current code however tries to correct the extension.
We should try to fix the case listed above though.

jschobben:
Ah, right!

The first post in #2877 describes it like I'd expect: force the format when the extension can't match the format's for some reason.

The way it's now, is that ```--export-format``` can only be used together with ```-o``` (otherwise it's ignored and the GUI starts). That's fine, but consider this:
- if ```--export-format``` changes the file extension, why not just use ```-o``` directly?
I.e. isn't ```-o filename.bla --export-format stl``` the same as just ```-o filename.stl``` then? It kind of defeats ```--export-format```'s purpose, I think
- If I run ```openscad input.scad -o output.bin --export-format png```, and there already exists a file ```output.png```, it's silently overwritten

I guess it's a manner of preference, but personally I prefer if a tool does what I tell it, instead of trying to be smart and "fixing" my mistake 😃 

Just having a warning might be OK, but it can also become annoying for legitimate use cases (such as the above, and that of #616) .

kintel:
I kind of agree that the magic replacement is bit unexpected. As is forcing lower-case.
A warning if conflicting with another known suffix could be a nice touch, but I think this PR is an improvement.

t-paul:
I guess we should remove the now confusing warning...

kintel:
@t-paul I didn't think there was a warning there yet? ..or did I miss smth.?

t-paul:
Ok, warning is maybe wrong word, I think that message would show up

https://github.com/openscad/openscad/blob/fe3c5cb85c839e4dae6a8d39b4449d906139e7d3/src/openscad.cc#L287-L290

for `openscad input.scad --export-format stl -o /dev/null`?

t-paul:
...which is updated to make more sense with the new behavior in #3046.

PR #3046, opened 2019-08-19
Export format logs

jschobben:
Follow-up of PR #3033. Reword help and logs for --export-format a bit.

### DIFF
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -285,7 +285,6 @@ int cmdline(const char *deps_output_file, const std::string &filename, const cha
 	const char *new_output_file = nullptr;
 
 	if(!export_format.empty()) {
-		PRINTB("Using format '%s' of --export-format option", export_format.c_str());
 		formatName = export_format;
 	}
 	else {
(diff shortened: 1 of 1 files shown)


src/openscad.cc
```diff
@@ -280,29 +280,29 @@ int cmdline(const char *deps_output_file, const std::string &filename, const cha
...
 	FileFormat curFormat;
-	std::string extsn;
+	std::string formatName;
 	std::string output_file_str = output_file;
 	const char *new_output_file = nullptr;
 
 	if(!export_format.empty()) {
-		PRINT("Using extension of --export-format option");
-		extsn = export_format;
+		PRINTB("Using format '%s' of --export-format option", export_format.c_str());
```
kintel: As this is simply a confirmation that OpenSCAD understood the `--export-format` option, I wonder if we could simply remove it

t-paul: I guess that's fine too, as unknown values are warned in https://github.com/openscad/openscad/blob/d21e65b4ce7aa8cc33731feabc60e8a38d819eab/src/openscad.cc#L1018 already.

jschobben: Indeed, it's perhaps best to just remove this message.

PR #3045, opened 2019-08-19
Initial commit to implement hyperlink feature

RomitKumar:
Todo list:
- [x] Correct Hyperlink marker on switching tabs
- [x] Implement Ctrl+Click for hyperlink activation
- [ ] Proper setting of Location object

PR #3049, opened 2019-08-24
Hyperlinks for use/include

t-paul:
Based on #3045, implementing location tracking for both `use<>` and `include<>`.


kintel:
FYI: I think location tracking of `include` and `use` should really happen on the AST level. I started on that by introducing AST nodes for these, but got stuck.
I won't let that block this PR though :)

kintel:
@t-paul Needs conflict resolution

PR #3055, opened 2019-09-01
GUI: make CTRL-W close current tab

### LINKED ISSUE #3053: Ctrl-W should close single file
With multi-file editing implemented now, I would expect the Ctrl-W shortcut to close just the current file/tab. Similar to how it functions in most browsers etc.

Currently it closes the entire OpenSCAD window regardless of how many editor tabs are open.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/79651563-ctrl-w-should-close-single-file?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

agordon:
Fixes #3053.

### DIFF
--- a/src/MainWindow.ui
+++ b/src/MainWindow.ui
@@ -1252,7 +1252,7 @@
   </action>
   <action name="fileActionClose">
    <property name="text">
-    <string>&amp;Close Window</string>
+    <string>&amp;Close</string>
    </property>
    <property name="shortcut">
     <string>Ctrl+W</string>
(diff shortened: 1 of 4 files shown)


t-paul:
Nice. I think we should retain the "Close Window" though. So the File menu section would get the new action and also the `CTRL-W` shortcut and `&C` mnemonic. The existing action stays feature wise as plain action. What do you think?

agordon:
Putting aside the menu labels for  second, and focusing on common funtionality:

Almost every unix GUI app has two common actions: "close CTRL+W" and "quit CTRL+Q".
"close" closes the current 'document' (which could be in a tab or in a separate gui window),
and "quit" closes the entire application (and all tabs / windows).
It's the same under windows except the shortcuts are "CTRL-F4" and "ALT-F4"
(at least the last time I programmed for windows, which was long long ago).

In that respect, and especially after introducing the "tabs" feature to openScad,
I think it would be better to go with this common usage (that is: "close CTRL-W" closes the current tab, if it's the last tab in the current window - it closes the window, and if it's the last window - closes the app).

I believe this is consistant with typical usage (e.g. of all modern web browsers, which have both tabs and multiple windows).

I'll send an updated patch to close the current window instead of quitting the app,
and rename "close window" to simply "close".

PR #3060, opened 2019-09-02
GUI: Jump to errors

agordon:
This makes debugging easier: if the source code has an error, a quick keypress jumps to it.
The weird keyboard shortcut (CTRL-ALT-E) is just a suggestions, happy to change it.

(also: this PR conflicts with my other one(#3059), if one is accepted, I'll amend the other accordingly.

### DIFF
--- a/src/scintillaeditor.cpp
+++ b/src/scintillaeditor.cpp
@@ -999,4 +999,11 @@ void ScintillaEditor::prevBookmark()
 
 void ScintillaEditor::jumpToNextError()
 {
+	int line, index;
+	qsci->getCursorPosition(&line, &index);
+	line = qsci->markerFindNext(line+1, 1<<errMarkerNumber);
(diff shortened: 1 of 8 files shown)


agordon:
(adjusted version that merges cleanly)

src/scintillaeditor.cpp
```diff
@@ -996,3 +996,14 @@ void ScintillaEditor::prevBookmark()
 	if (line != -1)
 		qsci->setCursorPosition(line, index);
 }
+
+void ScintillaEditor::jumpToNextError()
```
t-paul: With now 3 mostly identical functions, it would be nice to have just one implementation. Maybe something along the lines of https://godbolt.org/z/wpniJJ ?

t-paul: Merging the other pull requests did cause a conflict again. Fixing via #3063...

PR #3063, opened 2019-09-05
Jump to error

t-paul:
Fix merge conflict on #3060 and ensure we end up on correct line even if the source column is larger than the target line length.


agordon:
nice!

PR #3073, opened 2019-09-12
macOS: Rewrote version checks to explicitly write the built version to file

kintel:
@t-paul This should perhaps make the CI build for macOS more robust. Still doesn't manage dependencies though.
Update: As the  rabbit hole always goes a bit deeper than we think, this took some more time, but it looks good now.

TODO:
* [x] Rebuild all dependencies on circleci and upload manually to server
* [x] Test by creating `circleci-macos` branch
* [x] If possible, test by upgrading a single package and verify that the CI rebuilds and re-uploads the archive


kintel:
@t-paul This should be good to go now, in case you want to have a look

t-paul:
Automatically scheduled build [looks good](https://app.circleci.com/jobs/github/openscad/openscad/3818). Unfortunately I can't see a way to configure something like "daily build but only when SCM changed". But maybe we can just do 5 days (like skip Tu/Th) a week which should amount to about 200 of the 500 allowed minutes.

kintel:
Yeah, sounds good!
At some point, we could auto-upload the nightlies, including the SHA, in which case we could check that first.

kintel:
Could you write a ticket (if it doesn't exist). I can try to poke at the server.

t-paul:
Sure, I'll collect the info and try to create a test case with some dummy html file.

PR #3075, opened 2019-09-14
Add Find/Find Next/Find Previous to editor context menu (closes #3069).

### LINKED ISSUE #3069: Feature request: Add 'Find' to editor context menu
When reading code, particularly someone else's where you have to go hunt for modules etc - like when debugging some gears today, I mostly do it with my trackball with scroll-wheel, with hands AFK. Due to a neck problem I'm laid-back unless I need to be over the keyboard typing.

It would be really handy to be able to select text, then `Right-click/Find` to do the equiv. of `Ctl-F`, rather than having to trackball all the way up to the menubar. Modern screens have a lot more real-estate to cover.

Bonus would be, if Find is active, context menu items of Find-Prev & Find-Next.  

I'm not sure whether any of the 'replace' button functions would be good on the context menu. (?) I don't use it much. But `Right-Click/Replace` would keep the pointer in the vicinity of the code???

p.s. What does the Comp: prefix represent in the Git issue Labels? I'm presuming Comp:GUI is correct here.

t-paul:


MichaelAtOz:
Works well. Thanks. p.s. had you noticed the builds are not making it to the snapshot web page?

t-paul:
Yes I have prepared some changes that should make it possible to automatically update but those still need some investigation why they trigger some AdBlock. I suspect it's a config issue on our web server. Once that's resolved we should be able to pull in the dev snapshots without manual steps.

PR #3077, opened 2019-09-16
Function literals

t-paul:
This is the first step in implementing [function literals](https://github.com/doug-moen/openscad2/blob/master/rfc/Functions.md).

ToDo:
- [x] Fix/Validate (lexical) scoping
- [x] Validate operator precedence and associativity
- [x] Disable function calls on `$` variables
- [x] Check parser conflict, seems to be another dangling else case
- [x] Check warning/error cases
- [x] Recursive calls
- [x] Enable tail recursion elimination (by function name)
- [x] Add `is_function`
- [x] Add test cases

ToDo later...
- Support for defining function in shortcut style
- Support directly defining curried functions
- Decide about behavior in customizer

grafik

```

hypot = function(x, y) sqrt(x * x + y * y);
mean  = function(x, y) (x + y) / 2;

func  = function(x) x == 0
    ? function(x, y) x + y
    : function(x, y) x - y;

module m(h, x, y) {
    echo(h = h(x, y));
    translate([h(x, y), 0, 0]) cube(1);
}

echo(func0 = func(0)(8, 3));
echo(func1 = func(1)(8, 3));

echo(l = let(f = function(x) x) sqrt(f(25)));

echo(f = (function(x) sqrt(x))(49));

color("red") m(hypot, 3, 4);
color("green") m(mean, 5, 8);
color("blue") m(function(x, y) sqrt(x + y), 5, 6);
```


nophead:
Cool.

What does echo(hypot) do and what does m(1,2,3) do?

t-paul:
Currently those give:
```
echo(hypot);
ECHO: [function]
```
```
m(1,2,3);
WARNING: Ignoring unknown function 'h', in file debug.scad, line 10.
ECHO: h = undef
WARNING: Ignoring unknown function 'h', in file debug.scad, line 11.
WARNING: Unable to convert translate([undef, 0, 0]) parameter to a vec3 or vec2 of numbers, in file debug.scad, line 11
```

nophead:
Makes sense.

So variables that have function literals assigned to them are normal variables with a new type.

What happens if you have a normal function with the same name?

Can they be recursive?

t-paul:
Hmm, good question. I have not checked yet what the current implementation does. But I guess we need to make sure it works.

I think normal functions will be shadowed if a variable with the same name exists.

t-paul:
Lexical scoping of variable captured in function definition
```
module scope(f) {
    a = 5;
    echo(f = f(a));
}

a = 3;
x = 9;
scope(function(x) echo(x = x, a = a) x + a);
```
```
// ECHO: x = 5, a = 3
// ECHO: f = 8
```

t-paul:
Recursive function
```
module mod(modf) {
    echo(modf = modf(5));
}

func = function(x) x <= 0 ? 0 : x + func(x - 1);
echo(func = func(8));
mod(func);
```
```
ECHO: func = 36
ECHO: modf = 15
```

t-paul:
Functions returning another function, capturing `a` statically and resolving `$a` dynamically.
```
module m2(f1, f2) {
    a = 3;
    $a = 30;
    echo(f1 = f1(0.03), f2 = f2(0.07));
}

module m1(f) {
    a = 2;
    $a = 20;
    add2 = function(x) function(y) x + y + a + $a;
    m2(f, add2(0.2));
}

a = 1;
$a = 10;
add1 = function(x) function(y) x + y + a + $a;
m1(add1(0.1));
```
```
ECHO: f1 = 31.13, f2 = 32.27
```

MichaelAtOz:
I couldn't read that until my morning caffeine kicked in...

t-paul:
What about this one, slightly extending the Wikipedia examples for [higher order functions](https://en.wikipedia.org/wiki/Higher-order_function)? This covers both the _takes one or more functions as arguments_ (`twice`) and _returns a function as its result_ (`add` and `mul`) cases in one evaluation.
```
add3 = function(x) x + 3;
mul2 = function(x) x * 2;
twice = function(f, x) f(f(x));

add = function(x) function() twice(add3, x);
mul = function(x) function() twice(mul2, x);

echo(add = add(7)(), mul = mul(7)());
```
```
ECHO: add = 13, mul = 28
```

t-paul:
Hmm, looks like we want array slicing too to get a nice way to express `fold`:
```
a = [1, 2, 3, 4, 5, 6, 7, 8, 9];

fold = function(i, v, f) len(v) > 0 ? fold(f(i, v[0]), v[1:], f) : i;

sum = function(v) fold(0, v, function(x, y) x + y);
prd = function(v) fold(1, v, function(x, y) x * y);
cnt = function(v) fold(0, v, function(x, y) x + 1);
rev = function(v) fold([], v, function(x, y) concat(y, x));

echo(sum = sum(a), prd = prd(a), cnt = cnt(a), rev = rev(a));
```
```
// ECHO: sum = 45, prd = 362880, cnt = 9, rev = [9, 8, 7, 6, 5, 4, 3, 2, 1]
```

kintel:
Some of the demo in https://github.com/openscad/list-comprehension-demos could probably be rewritten more elegantly using this feature

kintel:
@nophead It looks like a literal function (in the variable namespace) shadows a classic function (in the function namespace), but _only_ if the variable actually resolves to a function.
That's somewhat backwards compatible, but perhaps a bit surprising:

```
function some_func(t) = t*3;

module MyModule(some_func, x) {
    r = some_func(x);
    echo(r);
    sphere(r);
}

myfunc= function(t) t*2;
MyModule(0, 5);
```

Perhaps worth a warning?

t-paul:
The question is also if we want to make the builtins available as "new style" functions (in the variable namespace). A similar effect would be  a very limited support for namespaces (e.g. something like just the builtin namespace like `::sin()`).

nophead:
@kintel , is there a line missing from your example? I.e. you don't use myfunc.

I don't know what the warning would be.  Parameter name is the same as a function name and is also used to pass a function literal.  It might even be an error, rather than a warning as I don't think the result would intended.

t-paul:
Yes, the example is `MyModule(0, 5);` vs. `MyModule(myfunc, 5);`

nophead:
You can't put the builtins in the variable namespace as that breaks
compatibility. Being able to pass and old style with :: would work but is
getting a bit messy.

Can you just pass function(x) sin(x) as a lamda style literal instead of
::sin(x)?

t-paul:
It would only interact with existing scripts in case of something like `is_undef(sin)`. A variable with name `sin` would just shadow the builtin. I think.

Yes, it would be possible to pass builtins as anonymous function literal, but not calling it `sin`.

kintel:
We can still populate the built-in top-level variable namespace with new-style functions, too give people the choice. That wouldn't break anything out of the box as classic functions are still available.
The namespace stuff would be cool, but perhaps better dealt with once we implement proper namespaces?

src/value.h
```diff
@@ -3,20 +3,32 @@
...
+
+#include "AST.h"
 #include "memory.h"
 
 class tostring_visitor;
 class tostream_visitor;
+class ValuePtr;
+class Context;
+
+class Expression : public ASTNode
```
kintel: There is no need to move the `Expression` class from its own header here - you can use a forward declaration, as all references to it are using a `shared_ptr`

src/value.cc
```diff
@@ -383,6 +413,9 @@ class tostring_visitor : public boost::static_visitor<std::string>
     return (boost::format("[%1% : %2% : %3%]") % v.begin_val % v.step_val % v.end_val).str();
   }
 
+  std::string operator()(const std::shared_ptr<Expression> &v) const {
+	  return "[function]";
```
kintel: Would it have any merit to emit some unique identifier in addition?

t-paul: Not pushed yet, but I've changed that to print the actual function definition. I guess that's needed once we continue with object-literals :-) (which I'm playing with to see how that interacts with the function-literals).

src/value.h
```diff
@@ -230,7 +247,7 @@ class Value
     return stream;
   }
 
-  typedef boost::variant< boost::blank, bool, double, str_utf8_wrapper, VectorType, RangeType > Variant;
+  typedef boost::variant< boost::blank, bool, double, str_utf8_wrapper, VectorType, RangeType, std::shared_ptr<Expression>> Variant;
```
kintel: I have to think about this, but at first read, it feels odd: We're storing an Expression, which is an AST construct, in a value, which is a run-time construct. For all other types, it's the other way around: The expression embeds a literal value.

My immediate hunch would be to define a `FunctionLiteral` stored in a value and aggregate that into the `FunctionDefinition` expression.

We don't currently need a good separation between the AST and the run-time, but this makes the coupling a bit tight.

-> could be refactored later though

kintel: Oh, and as a side-question: Why store an expression and not the `FunctionDefinition` directly?

src/context.h
```diff
@@ -3,32 +3,70 @@
...
-	typedef std::vector<const Context*> Stack;
+	typedef std::vector<std::shared_ptr<Context>> Stack;
+
+	std::shared_ptr<Context> get_shared_ptr() { return shared_from_this(); }
+	void push(std::shared_ptr<Context> ctx);
+	void pop();
+
+    template<typename C, typename ... T>
+    static ContextHandle<C> create(T&& ... t) {
+        return ContextHandle<C>{std::shared_ptr<C>(new C(std::forward<T>(t)...))};
```
kintel: Usually, you do `make_shared<C>()` instead of calling the constructor in these cases, which would allow for move semantics if everything else works out.

src/control.cc
```diff
@@ -166,7 +166,7 @@ AbstractNode* ControlModule::getChild(const ValuePtr &value, const EvalContext*
 	return modulectx->getChild(n)->evaluate(modulectx);
 }
 
-AbstractNode *ControlModule::instantiate(const Context* ctx, const ModuleInstantiation *inst, EvalContext *evalctx) const
+AbstractNode *ControlModule::instantiate(const std::shared_ptr<Context> ctx, const ModuleInstantiation *inst, std::shared_ptr<EvalContext> evalctx) const
```
kintel: I _think_ there is a small optimization potential in passing a reference to the `shared_ptr` instances, as it may not have to increase the reference count for every call.

src/value.h
t-paul: Agreed, the evaluate to a copy of itself seems a bit odd too.

src/builtincontext.cc
```diff
@@ -6,27 +6,31 @@
...
 #include "evalcontext.h"
 
-BuiltinContext::BuiltinContext()
+BuiltinContext::BuiltinContext() : Context()
 {
-	for(const auto &ass : Builtins::instance()->getAssignments()) {
-		this->set_variable(ass.name, ass.expr->evaluate(this));
+}
+
+void BuiltinContext::init()
```
kintel: Not sure I understand why initialization moved from the constructor to `init()`. Feels a bit counter to how I'm used to think about C++ RAII.

kintel:
Btw., one potential issue to think about:
`use <file.scad>` only exports classic functions and modules, not function literals bound to variables.
-> how do we build libraries of such functions?

nophead:
Presumably by using include.

In my library I include all the top level files, which have a few constants
in them (usually lists defining objects) and those use a second file that
has all the functions and modules. E.g. screws.scad uses screw.scad. I
would simply put new style functions in the first file.

t-paul:
I think that's where there objects come in. Those are supposed to provide a way to customize values in a much more obvious way than the include/overwrite hack. Thinking of it they also automatically provide some kind of namespacing.

src/value.h
t-paul: Ahh, good point. I moved it before the `shared_ptr` change. Moved back.

src/context.h
t-paul: I wanted to prevent construction by other means to ensure the scope stack is maintained via the `ContextHandle` object. With protected constructor or destructor, the `make_shared<C>()` gives an error. Any tips how to get both?

src/value.cc
t-paul: Pushed the change to print the actual function definition. Will revisit when converting to a specific `FunctionLiteral` object instead of generic `Expression`.

src/builtincontext.cc
t-paul: Passing `this` to evaluate as `shared_ptr` does not work in the constructor. I think to fix the issue we will need to decouple the evaluation functions from the context objects. Same issue also in Module and Function context which want to pass `this` along.

kintel: Ah, right,  the `shared_from_this()` hack.  That's  a bit scary, but  since we always construct context  objects  held by shared_ptrs, we should be safe.

src/context.h
kintel: Oh, didn't think about the protected constructor. I wouldn't dig too  deep into that - this is where C++  gets really obscure..

src/control.cc
t-paul: Done. Micro benchmark with recursive fibonacci shows small improvement but in total still quite a bit slower than the nightly build.
```
function fib(x) = x < 2 ? x : fib(x - 1) + fib(x - 2);

28f9df23 - Only use forward declaration for Expression.
36: 1m31,538s
40: 10m35,664s

0aac8bed - Pass context via reference.
36: 1m23,405s
40: 9m36,937s

Nightly
36: 0m36,848s
40: 4m22,674s
```

t-paul:
Ok, to expand on that and provide something to play with, I pushed the Object literals branch to #3087 as draft.

I'm not sure the [proposed syntax](https://github.com/doug-moen/openscad2/blob/master/rfc/Objects.md) is necessarily what we want, but the idea there is that external scripts are imported via
```
lollipop = script("lollipop.scad"); // e.g. default of variable length is set to 12
```
Using the customizing syntax, it's then possible to overwrite specific (existing) variables without any tricks.
```
lollipop = script("lollipop.scad")(length = 10); // customize length to be 10
```
Both `script()` and geometry in objects are not yet implemented in #3087 but it has some very simple JSON import to play with the data part.

nophead:
I can't say I understand this at all. How is a script a geometric object? Most of my scripts do not create geometry, they define modules and functions that can be used to create geometry.

Yes the example is a script that creates geometry but that isn't the general case, so I don't see why it should become a first class value.

t-paul:
I think what @doug-moen tries to do there is to unify how to look at things. To me it seems that the new objects do have a lot of overlap with modules, so maybe that needs some more discussion. The introduction of "script is a geometric object" is slightly changed in the more detailed "Objects have multiple roles in OpenSCAD2." section which also explains that they can be just data.

nophead:
Well to me an object is what you get when you instantiate a module, not a script. Does "script" denote something different from a normal OpenSCAD file here?

My OpenSCAD library files are more akin to classes than objects. None of them fit the pattern of the lollipop example. Most objects in my library have multiple geometry defining modules, not just the one that draws them. For example a fan also has modules to position screws, make a panel cutout, make an assembly with fasteners in place, etc.

The only files of mine that resemble the lollipop example are the top levels of projects that define a lot of constants and then draw the entire project, which is usually and assembly with dozens of parts.

MichaelAtOz:
Terminology! 
We need an agreed term for 'a geometric thingi', so until a better term comes along, to avoid confusion I'll use Thingi, (note I did not use 'object'). 'script' may also need a better term, but will do for now - a section of OpenSCAD code.

> wikipedia: In computer science, an object can be a variable, a data structure, a function, or a method, and as such, is a value in memory referenced by an identifier.
>
>In the class-based object-oriented programming paradigm, object refers to a particular instance of a class, where the object can be a combination of variables, functions, and data structures. 

nophead's 'what you get when you instantiate a module' is a Thingi, I have previously used 'object', 'model', 'geometry' but never found the right term.
'instantiate' is the key distinction, until a module (or geometry built-in) is 'called' there is no Thingi.

But in an object-oriented sense you can, with these changes, also have an instance of [a packaged section of OpenSCAD code - modules functions variables - that is not Thingi's but can produce Thingi's] in a variable. 

I had been thinking of the subject of these change as being able to create the equivalent of 'class' (without constructors etc), if so then by convention the instantiation of that is an 'object', which is not a Thingi, but may be further 'instantiated' (aka called) to produce Thingi's.

So Script is a collection of code packaged so it can be instantiated into an Object, which can then be manipulated, including 'calling' to produce Thingi's. _Really need something better than script..._

nophead:
But the package of code that is being called an object and is a generated by a particular  format script is not something I recognise. You can't instantiate geometry in a scad file that is used. So lollipop.scad would currently need to be included each time you wanted a lollipop but then they would all have to be the same because if you include scad the variables would all be the same.

The way to make customisable lollipops is to define a module, possibly in a used file, and pass it parameters. Not make a file with geometry in it with predefined geometry and variables that gets included and overridden.

So it seem to represent a paradigm that nobody uses at the moment. Why is this weird mechanism being added to the language?

MichaelAtOz:
It's unfortunate that the wiki uses object, as in 3D Objects (cube etc), 2D Objects (square etc).

Even Doug's [description](https://github.com/doug-moen/openscad2/blob/master/rfc/Objects.md), is having difficulty with Thingi's, however the concise line is
> A geometric model is represented by an object, which encapsulates both its geometry and its parameters.

Where 'its geometry' seems to be 'code which produces Thingi', 'geometric model' is Thingi when 'its geometry' is called (with its parameters).

> A "geometric object", or "object" for short, is a first class value that represents an OpenSCAD script.

So 'geobject'?

@nophead 
> an object and is a generated by a particular format script is not something I recognise. You can't instantiate geometry in a scad file that is used. So lollipop.scad would currently need to be included each time you wanted a lollipop but then they would all have to be the same because if you include scad the variables would all be the same.

This is another paradigm,

`lollipopA = script("lollipop.scad"); // e.g. default of variable length is set to 12`

has lollipopA.length=12, whereas

`lollipopB = script("lollipop.scad")(length = 10); // customize length to be 10`

has lollipopB.length=10, and

`lollipopC = lollipopB(radius=15); // customize radius to be 15

has lollipopC.length=10 & lollipopC.radius=15

There are no Lollipop Thingi's until you do

```
translate([-50,0,0]) lollipopB;     // tho I think lollipopB(); may be better???
translate([50,0,0]) lollipopC;      // could you then do lollipopC(radius=12) too???
```

nophead:
But why? What is wrong with modules?

And how to I have a script to make a fan, that can also cut it holes and make fan assembly?

MichaelAtOz:
I raised #3088 to continue discussion & leave this issue for technical matters.

doug-moen:
I think there is some confusion about OpenSCAD2.

A "script" is an element of syntax. It's a mixture of statements (that create geometry) and definitions (that define variables, functions and modules). An entire OpenSCAD program is a script. You can also enclose a script (a sequence of statements and definitions) in curly braces.

An "object" is what is returned by a module. So it is a run-time concept, not a grammatical concept. For example, in
```
module lollipop() {
  radius   = 10; // candy
  diameter = 3;  // stick
  height   = 50; // stick
  translate([0,0,height]) sphere(r=radius);
  cylinder(d=diameter,h=height);
}
```
then the thingy returned by `lollipop()` is an object, and the text between the curly braces is a script.

Consider that a script contains both statements (that generate geometry) and definitions. Program scripts are sometimes used to generate geometry and specify a 3D model. Other times, program scripts are used as libraries: it is the definitions that are important, and libraries do not usually contain any geometry.

In OpenSCAD2, objects are generalized. An object is a first class value that is constructed from a script. An object contains geometry (if any was specified), and it contains named fields that are initialized from the definitions, if any definitions were present. You can select fields from an object using dot notation: `obj.field`.

Just as program scripts are used either for 3D models or libraries, so OpenSCAD2 objects can be used either as shapes or as library namespaces. You can import a program script into memory using script("filename.scad"), and that will return an object. If the imported script is a model, then you can use the object as a shape. If the imported script is a library script, then you can use the object as a library namespace. So now we have the concept of namespaces.

In OpenSCAD2, the object returned by `lollipop()` has named fields like `radius` and `diameter`. So now models can carry around the parameters used to generate them, and those parameters can be programmatically accessed when the object is passed as an argument to another module.

There's more about library objects here: https://github.com/doug-moen/openscad2/blob/master/rfc/Library_Scripts.md

When @kintel reviewed the OpenSCAD2 design in 2016, he suggested that the `script()` module should be called `import()` instead. Perhaps that would eliminate some confusion.

MichaelAtOz:
Please reply to the emails for #3088. I copied Doug's last post there.

kintel:
@t-paul How hard would it be to make this experimental, to give it some time to mature?

Also, should we open a ticket to address the performance regression against nightly?

t-paul:
I don't see a way to make the parser changes experimental. We could flag the actual function literal to be always be an `undefined` literal. Yes, a ticket to check how to get the performance back makes sense.

kintel:
I think just making the usage of the function literal experimental is good enough, just to raise awareness that using this feature isn't 100% set in stone yet - makes it a lot easier to iterate.

src/value.h
kintel: Did you have a chance to think about this? I think this is the last remaining concern I have about this PR..

t-paul: I've now added a `FunctionType` for use as value. It still uses `Expression` and `AssignmentList` though.

src/control.cc
t-paul: With the latest changes, this is now at `36: 1m2,891s`.

t-paul: With the `Value` rework changes, results are `36: 0m51,910s`.

src/context.h
t-paul: I found some example of how it would be possible and those look a bit crazy :). After some minimal performance testing, I don't see any change with `make_shared()` (which is supposed to potentially optimize for one allocation instead of two?). So for now I'd prefer the protected constructor to make sure we don't accidentally break the dynamic scoping hierarchy.

t-paul:
Awesome, lets go then. I'll rebase the `Value rework` stuff from @thehans (#3096) against master then as I guess the current one will auto-close with this branch gone.

MichaelAtOz:
Ronaldo has an issue on the [Forum](http://forum.openscad.org/Functions-literals-higher-order-functions-tp27287p28039.html),

> I can't see why the following generates a stack overflow:
```
    function move(a, g) =
      function(x) g(x+a) ;

    g1 = function(x) x;
    g2 = move(0,g1);
    g3 = move(0,g2);

    echo(g3_x=g3);
    echo(g3_2=g3(2));
```
> If g3 is defined as g2 no error is reported.

With 2019.10.06 I get

```
ECHO: g3_x = function(x) g((x + a))
ERROR: Recursion detected calling function 'g' in file fn literals.scad, line 4
TRACE: called by 'g', in file fn literals parkinbot.scad, line 4.
TRACE: called by 'g3', in file fn literals parkinbot.scad, line 13.
TRACE: called by 'echo', in file fn literals parkinbot.scad, line 13.
```
If I add an `echo()` 
```
function move(a, g) =
  function(x)
    echo(a,g)
      g(x+a) ; 
```
I get infinite loop of echo in console, quiting GUI crashes.

I then did
```
function move(a, g) = echo(move=a,g)
  function(x) 
      echo(g_in_move=a,x)
        (x+a) ;


g1 = function(x)    
        
          x;
g2 = move(1,g1);
g3 = move(3,g2);

echo(g1_=g1);
echo(g1_1=g1(1));
echo(g2_=g2);
echo(g2_7=g2(7));
echo(g3_x=g3);
echo(g3_5=g3(5));
```

Which gets:
```
ECHO: move = 1, function(x) x
ECHO: move = 3, function(x) echo(g_in_move = a, x) (x + a)
ECHO: g1_ = function(x) x
ECHO: g1_1 = 1
ECHO: g2_ = function(x) echo(g_in_move = a, x) (x + a)
ECHO: g_in_moveg = 1, 7
ECHO: g2_7 = 8
ECHO: g3_x = function(x) echo(g_in_move = a, x) (x + a)
ECHO: g_in_moveg = 3, 5
ECHO: g3_5 = 8
Geometries in cache: 0
```

The extra echo stops the infinite recursion, not sure why, I haven't had my morning coffee yet...

MichaelAtOz:
Oops, at line 4 of the second code, I dropped the g() in editing.
That reinstates the infinite recursion.

`assert` helps.
```
function move(a, g) = echo(move=a,g)
  function(m) 
      echo(m_in_move=a,m,g)
      assert(m<14)
        g(m+a) ;

g1 = function(x)
       echo(fx=x)
          x;
g2 = move(2,g1);
g3 = move(3,g2);

echo(g2_=g2);
echo(g3_x=g3);
echo(g3_5=g3(5));
```
Gives
```
ECHO: move = 2, function(x) echo(fx = x) x
ECHO: move = 3, function(m) echo(m_in_move = a, m, g) assert((m 
ECHO: g2_ = function(m) echo(m_in_move = a, m, g) assert((m 
ECHO: g3_x = function(m) echo(m_in_move = a, m, g) assert((m 
ECHO: m_in_move = 3, 5, function(m) echo(m_in_move = a, m, g) assert((m 
**ECHO: m_in_move = 2, 8, function(x) echo(fx = x) x
ECHO: m_in_move = 2, 10, function(x) echo(fx = x) x
ECHO: m_in_move = 2, 12, function(x) echo(fx = x) x
ECHO: m_in_move = 2, 14, function(x) echo(fx = x) x
ERROR: Assertion '(m < 14)' failed in file fn literals ronaldo.scad, line 4
TRACE: called by 'g', in file fn literals ronaldo.scad, line 5.
TRACE: called by 'g3', in file fn literals ronaldo.scad, line 15.
TRACE: called by 'echo', in file fn literals ronaldo.scad, line 15.
```
My head hurts. Feels like a scope issue...
Also see the Forum, Ronaldo has other observations.

MichaelAtOz:
As  `echo(fx=x)` is never called, g1 is never called, even tho g=g1[` function(x) echo(fx = x) x`] as echo'd by 'm_in_move', the g that gets called is g2 again.

PR #3083, opened 2019-10-02
Update for CGAL 5.0-beta1

lrineau:
- Use CGAL header-only (possible since CGAL version 4.8 and mandatory since 5.0)
- Use C++14 if CGAL is enabled, because CGAL-5.0 requires C++14.

I have not modified the CMake configuration of OpenSCAD, because `CGALConfig.cmake` will handle all the details (by setting properties on the target `CGAL::CGAL`).

### DIFF
--- a/features/cgal.prf
+++ b/features/cgal.prf
@@ -1,4 +1,6 @@
-DEFINES += ENABLE_CGAL
+DEFINES += ENABLE_CGAL CGAL_HEADER_ONLY
+
+QMAKE_CXXFLAGS += -std=c++14
 
 # Optionally specify location of CGAL using the 
 # CGALDIR env. variable
@@ -9,7 +11,7 @@ CGAL_DIR = $$(CGALDIR)
   message("CGAL location: $$CGAL_DIR")
 }
 
-LIBS += -lCGAL -lmpfr -lgmp
+LIBS += -lmpfr -lgmp
 
 *g++* {
   QMAKE_CXXFLAGS += -frounding-math
--- a/features/cgal.prf
+++ b/features/cgal.prf
@@ -1,4 +1,4 @@
-DEFINES += ENABLE_CGAL CGAL_HEADER_ONLY
+DEFINES += ENABLE_CGAL
 
 QMAKE_CXXFLAGS += -std=c++14
 
@@ -11,7 +11,14 @@ CGAL_DIR = $$(CGALDIR)
   message("CGAL location: $$CGAL_DIR")
 }
 
-LIBS += -lmpfr -lgmp
+exists($$CGAL_DIR/include/CGAL/compiler_config.h)|exists(/usr/include/CGAL/compiler_config.h) {
+  LIBS += -lCGAL -lmpfr -lgmp
+} else {
+  # CGAL header-only
+  DEFINES += CGAL_HEADER_ONLY
+  LIBS += -lmpfr -lgmp
+  message("Using CGAL header-only")
+}
 
 *g++* {
   QMAKE_CXXFLAGS += -frounding-math
--- a/src/builtincontext.cc
+++ b/src/builtincontext.cc
@@ -1,3 +1,6 @@
+#define _USE_MATH_DEFINES
+#include <cmath>
+
 #include "builtincontext.h"
 #include "builtin.h"
 #include "expression.h"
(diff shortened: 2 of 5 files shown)


kintel:
Ooh,  header-only CGAL!

To merge this, we need to make the build system detect CGAL version, as we still need to support older versions of CGAL for Linux distro targets. I believe we need to  support down to CGAL-4.6, but I'd have to look into the details.

lrineau:
CGAL header-only is available since 4.8.

t-paul:
Minimum 4.8 seems to be a problem with Ubuntu 16.04LTS. Otherwise it's not too bad.

- MXE has 4.11, so that's good
- Debian8 jessie has 4.5, Debian9 stretch has 4.9 (we probably can drop oldoldstable jessie)
- Ubuntu 16.04 has 4.7, 18.04 has 4.11.

lrineau:
Indeed you need to support older than 4.8. I do not know how to detect a library version using qmake. Have you already done something similar before?

t-paul:
The other test failures seem to be missing includes like `cmath` for M_PI probably just caused by more strict checks due to c++14.

t-paul:
I'd say the cgal-4.8/c++14 level of dependency is open for discussion. Not for a patch release which hopefully comes soon, but realistically the next full release will not happen this year. I suspect that could mean increasing the dependency version is just about ok. @kintel what do you think?

kintel:
As usual, the various CI services tends to be the slowest movers here, so as long as we don't break that we're technically fine.
Not sure what the Ubuntu landscape looks like; are people upgrading faster now after 12.04 finally died off?

t-paul:
Ok, lets see if the unrelated build errors are fixed with that change.

t-paul:
And another one in `src/builtincontext.cc`...

t-paul:
All builds except Ubuntu 16.04 are ok now. It should be possible to update the AppImage build environment to use a new CGAL version. For Travis we could drop 16.04 or find a way to get an updated package for the 16.04 build.

lrineau:
I have pushed an update, that tries to auto-detect if `<CGAL/compiler_config.h>` is present, either in `/usr/include/` or in `$$CGAL_DIR/include/`. That will fix the case for CGAL<4.8, if `<CGAL/compiler_config.h>` is correctly detected.

features/cgal.prf
```diff
@@ -9,7 +11,14 @@ CGAL_DIR = $$(CGALDIR)
...
 
-LIBS += -lCGAL -lmpfr -lgmp
+exists($$CGAL_DIR/include/CGAL/compiler_config.h)|exists(/usr/include/CGAL/compiler_config.h) {
+  LIBS += -lCGAL -lmpfr -lgmp
+} else {
+  # CGAL header-only
+  DEFINES += CGAL_HEADER_ONLY
+  LIBS += -lmpfr -lgmp
+  message("Using CGAL header-only")
+}
```
lrineau: @t-paul Do you see the error in that logic? It seems the message "Using CGAL header-only" is still displayed on Ubuntu 16.04.

t-paul: Looks like on Ubuntu 16.04 the header is installed in `/usr/include/x86_64-linux-gnu/CGAL/compiler_config.h`. I'm not sure how to check that in a portable way.

lrineau: We could try to parse the content of `<CGAL/version.h>` using qmake, with `$$cat` and `$$find`, but that will probably be tedious.

Could the non-header-only mode of CGAL be triggered by a qmake config variable, set by the user?

t-paul:
Ok, I'm changing the AppImage build to use an updated CGAL package. That will be `4.11` from Ubuntu 18.04 at this point. So that leaves only Snap which builds on Ubuntu Core 16.

With that done we should be good to drop support (scope = next full OpenSCAD release, probably next year?) for all systems which have CGAL < 4.8 as default while still building AppImages and Snap based on 16.04. That should be ok with `C++14` too, as 16.04 has `gcc-5.3.1`.

The `C++14` change should come in via a global setup (like currently in `c++11.pri`).

t-paul:
Snap is now building against CGAL-4.14.1.

lrineau:
Meanwhile, CGAL-5.0-beta2 has been published, if you want to test it out.

t-paul:
Superceeded by #3163. Master is now at C++14, both Fedora and Debian/testing are already shipping CGAL5.

PR #3084, opened 2019-10-03
Uomui

clarkebowers:
Specify the unit of measure in a DXF export file. The user sets the UOM in the preferences' advanced tab. The export to DXF now add the UOM and the AutoCAD file version to the output. This allows machining software to correctly perform routing or cutting.


t-paul:
It feels like the preferences might be a bit unusual place to store that, at least when working with different values.

The wishlist item for this so far is #649 which would put the selection into an export dialog.

PR #3087, opened 2019-10-05
Object literals

t-paul:
This is currently mostly playground for checking interaction with function literals.

Base is the [Objects](https://github.com/doug-moen/openscad2/blob/master/rfc/Objects.md) proposal.

- Only object literals
- Simple customization is implemented
- No scoping implemented yet

Example:
 ```
o = {
    a = 6;
    b = {
        f = function(x) x * x;
    };
    v = [1, 2, 3, { a = 12; }];
};

module m(x) {
    echo(mod = x.f(5));
}

echo(value1 = o);
echo(value2 = o(a = 8));
echo(vector = o.v);
echo(func = o.b.f(7));
echo(nested = o.v[3].a);

m(o.b);
```
```
ECHO: value1 = { a = 6; v = [1, 2, 3, { a = 12; }]; b = { f = function(x) (x * x); }; }
ECHO: value2 = { a = 8; b = { f = function(x) (x * x); }; v = [1, 2, 3, { a = 12; }]; }
ECHO: vector = [1, 2, 3, { a = 12; }]
ECHO: func = 49
ECHO: nested = 12
ECHO: mod = 25
```
Customization:
```
o1 = { a = 6; b = 4; d = 9; };
o2 = o1(a = 10, c = 2);
echo(o1 = o1);
echo(o2 = o2);
echo(o3 = o1(a = 1)(d = 0));
```
```
ECHO: o1 = { d = 9; a = 6; b = 4; }
ECHO: o2 = { b = 4; d = 9; a = 10; }
ECHO: o3 = { b = 4; a = 1; d = 0; }
```

### CODE REVIEW

**Comment by MichaelAtOz** [comment]
ECHO: value1 = { a = 6; v = [1, 2, 3, { a = 12; }]; b = { f = function(x) (x * x); }; }
ECHO: value2 = { a = 8; b = { f = function(x) (x * x); }; v = [1, 2, 3, { a = 12; }]; }

Shouldn't order be consistent?

**Comment by nophead** [comment]
I would think so if it is an object, rather than a dictionary.

**Comment by MichaelAtOz** [comment]
@t-paul Does this also include the Function Literals changes?

**Comment by t-paul** [comment]
Yes, also just now updated to track latest changes in the `function-literals` branch and adding lexical scoping for object literals.

**Comment by t-paul** [comment]
```
o = {
    a = 6;
    c = a + 2;
    b = {
        f = function(x) a + x;
    };
    v = [1, 2, 3, { a = 22; }];
};

module m(x) {
    echo(mod = x.f(4));
}

function ff(x) = let(result = x.f(6)) result;

echo(value1 = o);
echo(value2 = o(a = 18, c = a + 2));
echo(vector = o.v);
echo(func = o.b.f(7));
echo(nested = o.v[3].a);
echo(customized = o(a=0).b.f(5));

m(o.b);
echo(ff = ff(o.b));
```
Result shows the lexical scoping, e.g. the function `f` captures the value of `a = 6` which even the customization can't overwrite. As it's all just "normal" values, they can be passed to modules and functions.
```
ECHO: value1 = { a = 6; c = 8; b = { f = function(x) (a + x); }; v = [1, 2, 3, { a = 22; }]; }
ECHO: value2 = { a = 18; c = 20; b = { f = function(x) (a + x); }; v = [1, 2, 3, { a = 22; }]; }
ECHO: vector = [1, 2, 3, { a = 22; }]
ECHO: func = 13
ECHO: nested = 22
ECHO: customized = 11
ECHO: mod = 10
ECHO: ff = 12
```

**Comment by nophead** [comment]
Is this correct? I would expect a reference to a to be captured by lexical scope, not its value.

**Comment by t-paul** [comment]
I have no problem with throwing this prototype branch away if we find a different solution. I do like the automatic namespacing though.

**Comment by nophead** [comment]
What other problem does solve, that can't be done with module literals?

It just seems to have weird semantics, unlike any other language I know.

**Comment by rcolyer** [comment]
I think it's possible to preserve all the advantages of object literals and merge them with modules by tweaking the syntax of the object literals, and expanding access to modules, at least when defined as literals.  This retains a nice syntactic parallel to the function literals.  I wrote up some examples of this usage:

```
// Module and object merger.

o1 = import("file.scad");
o1();

o2 = import("file.scad", radius, height=20);
o2(5);
o2(radius = 7);
o2(7, 2);
o2(6, height=4);

o3 = module(radius, height=20) {
  diam = radius*2;
  resize([diam, diam, height]) sphere(radius);
  x = 3;
  cube(x);
}
o3(8);
o4 = o3;
o4.x = 15;
o4(8);

o5 = cube(5);  // Equivalent to o5 = module() { cube(5); }
o5();
o5(9); // Error, no parameters

o6 = cube;
o6(8);
o6.anything; // Would be an error, because no members defined on built-ins.

////////////////////////////////////////////
// Example from PR 3087 first post, adapted:
o7 = module() {
   a = 6;
   b = {
       f = function(x) x * x;
   };
   v = [1, 2, 3, { a = 12; }];
};

module m(x) {
   echo(mod = x.f(5));
}

m(o7.b);
////////////////////////////////////////////

module diff(obj1, obj2) {
  difference() {
    obj1();
    obj2();
  }
}
baseobj = cube(5);
diff(baseobj, sphere(3));

extobj = import("file.stl");
extobj();
```

**Comment by nophead** [comment]
I don't think ```o4.x = 15;``` should be allowed. A module has a formal parameter list that defines what variables are parameters you can change and their default values. Object literal don't have parameter lists so anything goes and it breaks normal lexical scope as @t-paul's example and encapsulation.

If you have a function definition inside a module that uses a local variable it only ever has one value for the whole scope of the module. So if it is a parameter with a default than the function uses the value passed. It doesn't capture the default and ignore what is passed.

The problem with object literals is they don't have formal parameter lists. They seem to formalise bad practice instead of provided a structured method to interface with scripts.

Instead why not allow module literals to have a body that is a script? That provides a namespace for importing libraries but also defines what parameters can be supplied. If other internal members can be accessed with dot notation then it should be read only access.

**Comment by rcolyer** [comment]
On a language for larger codebases, I would value the protections of enforced encapsulation.  Merging the concepts of module literals and objects as in my example is modeled after the Python philosophy of reflection where all but certain built-ins can be modified.  This has proven easy to use for non-expert programmers and useful for shorter programs where the flexibility is more empowering than the risks of the low safeguards.  It is then up to the programmer to follow expected usage of a component, or take on the responsibility for violating it.  In the case of OpenSCAD, the worst that can typically happen is a design doesn't render right if something is unwisely reassigned.

My problem with the separate object literal proposal is mostly that objects DON'T have an ability to specify a clean interface with parameter passing.  Thus we end up with two almost identical things where one lets you inspect it or access its pieces, and the other lets you control it through a clean interface.  I find it better to have one thing that can be used through a clean interface, and/or inspected or accessed as needed.  This would also be fewer concepts to learn for beginners to the language, as there is only one type of such thing.  They would not have to sort out which to use when.

**Comment by rcolyer** [comment]
Upon reflection (hah), I think most of the benefits of a module literal merging the object features are obtained even if the members are all read-only.  That would disallow the o4.x = 15; line that nophead points out, but still permit the highly beneficial features like modules being able to report their own calculated dimensions as member variables, or direct access to member functions or internal modules as a sort of duck-typing collection which can be passed around.

File: src/value.cc [hunk 1]
```diff
@@ -657,14 +657,6 @@ RangeType Value::toRange() const
   else return RangeType(0,0,0);
 }
 
-Value &Value::operator=(const Value &v)
```
t-paul: @kintel do you have any concerns removing this one? Newer compilers complain a lot with just the assignment operator but no custom copy constructor.

rcolyer: The rule of zero says to remove both that and the empty destructor in value.h.  It looks like the defaults should work fine.

t-paul: Oh, good point, I missed the destructor. I've pushed that to the `function-literals` branch to get it *soon-ish*...

kwikius:
From original example of how to use object_literals
> my_object = {
>    a = 6;
>    b = {
>        f = function(x) x * x;
>    };
>    v = [1, 2, 3, { a = 12; }];
> };

in current openscad (lateish 2022 master) we can do
```openscad
module my_object(){
   a = 6;
   module b() {
       f = function(x) x * x;
   };

   v = [1, 2, 3/* n/a , { a = 12; } */];
}
```
The problem is that since a module is not first class we can't easily access those members. We need a way to make module first class.

In my WIP first class module fork https://github.com/kwikius/openscad/tree/WIP-first-class-module you can do :

```openscad
my_object = module() {
   a = 6;
   b = module() {
       f = function (x) x * x;
   };
   v = [1,2,3, module() {a = 12;}];
};
```

NOTE1: "You can do" currently means it passes syntax checks, IOW the modified executable accepts the above, but member lookup etc not implemented yet. ( but object _literal code may have the necessary functionality
NOTE2: you should be able to  get rid of the module() syntax and just use module, not tested yet, but note that you can
optionall pass arguments to this structure using following syntax 

o1 = module my_object(args);

assume you can replace = module {..}; with  ={..} and I think you are all the way there? And the module can also optionally  take arguments. Also the question of how a module is instantiated is long solved so we don't need to reinvent the wheel there.

kwikius:
built from https://github.com/openscad/openscad/tree/object-literals

Get a syntax error on the first line. Should I be on a different branch?
```
o = {
    a = 6;
    c = a + 2;
    b = {
        f = function(x) a + x;
    };
    v = [1, 2, 3, { a = 22; }];
};
```

kwikius:
OK, I see there is an openscad object-literals branch. Apologies

kwikius:
so.. Is an object the same as a module or not?
Is an object  like a module but you cant put module_instantiations in it? ( If so why not?)
Do you want me to put stuff related to the issue #4336 formerly known as object_alias here or start a new issue or what?

nophead:
I think the terminology is confusing. An instantiation of a module makes a geometric object. It doesn't make something like a C++ object that has data members bound with functions that operate on them.

kwikius:
> I think the terminology is confusing. An instantiation of a module makes a geometric object. It doesn't make something like a C++ object that has data members bound with functions that operate on them.

The grammar is my reference for using the term  module_instantiation
https://github.com/openscad/openscad/blob/master/src/core/parser.y#L323

The source code provides the semantics
https://github.com/openscad/openscad/blob/master/src/core/ModuleInstantiation.h

In general instantiations may produce side effects. echo and assert are also module_instantiations
https://github.com/openscad/openscad/blob/master/src/core/parser.y#L314

kwikius:
Assuming I have the correct object_literals branch echo causes a syntax error

```openscad
o ={
    a = 1;
    echo( a = a);  // syntax error
};
```

```openscad
as does assert
o ={
    a = 1;
    assert( a ==1); // syntax error
};
```

jordanbrown0:
I agree with Nophead that the terminology is confusing.  We have two distinct kinds of objects - geometric objects, and this relatively new-to-OpenSCAD name-value-pair mechanism that is akin to what JavaScript calls objects and Python calls dictionaries.

The geometry-as-data effort (one proposal for which is #3956; I have a different one that doesn't have an issue or PR, but is described [here](https://docs.google.com/document/d/14siqbr9CyGuA2tY6dvTqiF3Hf9zFMyefk7YdaUJSn3Q)) further confuses the issue by allowing you to generate geometric objects but then *not* add them to the model, instead returning them as data to the program.  Their exact representation when returned to the program is still TBD, but they will probably be the second kind of objects, sets of name-value-pairs.

None of those are modules.  Modules are collections of executable content, akin to what other languages call functions, procedures, or subroutines.  Modules, when executed, will often cause geometric objects to be added to the model, but may also transform other geometric objects or have side effects like logging to the console.

The "module literals" effort is a parallel to the "function literals" feature, a way to have a reference to a module as a data item.  This is akin to what C would call a function pointer, or to what you get in JavaScript or Python when you utter a function's name without calling it using parentheses.

jordanbrown0:
The syntax part of my proposal is the easy part.

It adds a new entry to "primary":

```
| '{' inner_input '}'
```

Doing little more than that made it parse.  The tricky parts are pulling the assignments out, and consuming the resulting geometry.

kwikius:
> The syntax part of my proposal is the easy part.
> 
> It adds a new entry to "primary":
> 
> ```
> | '{' inner_input '}'
> ```
> 
> Doing little more than that made it parse. The tricky parts are pulling the assignments out, and consuming the resulting geometry.

Module Literals is what I have been working on, though I call it ModuleAlias.
My work on the parser you can see here
https://github.com/kwikius/openscad/blob/WIP-first-class-module/src/core/parser.y
anyway I am already finding the feture useful.  You can read the docs at https://github.com/kwikius/openscad/blob/WIP-first-class-module/doc/experimental/ModuleLiteral/moduleLiteral.md

You can also build the code and run it and test out the feature yourself in the openscad IDE ( tested on Linux only)

$ git clone -b WIP-first-class-module --single-branch https://github.com/kwikius/openscad.git
$ cd openscad
$ mkdir build
$ cd build
$ cmake .. -DEXPERIMENTAL=1
$ make -j2
#invoke local version of the openscad executable
$ ./openscad

jordanbrown0:
It seems that this discussion is plagued by terminological problems.

Let me suggest a glossary and some basic definitions. [n] marks footnotes.  [X] marks features present in today's development snapshots.  [x] marks features partially present.  [+] marks features to be added.

* module - an executable subprogram that can add geometric objects to a CSG tree.[1][X]
* module definition - a syntactic structure that creates a module.[X]
* module instantiation - an invocation of a module, or the process of evaluating a module.  Instantiating a module yields a (possibly empty) geometric object.  Also referred to as "module call" or "module invocation".[X]
* module reference - a data item that refers to a module.  A module reference can be called like a module.[4][+]
* module literal - a syntactic structure that embeds a module definition in expression context, allowing the creation of a module reference.[+]
* function - an executable subprogram that takes data values as input, and returns a data value.  Functions do not modify a CSG tree.[X]
* function definition - a syntactic structure that creates a function.[X]
* function reference - a data item that refers to a function.  A function reference can be called like a function.[4][X]
* function literal - a syntactic structure that embeds a function definition in expression context, allowing the creation of a function reference.[X]
* object - an data structure associating names with values.  Like all OpenSCAD data values, objects are immutable.[2][x]
* object literal - a syntactic structure allowing the creation of an object.[+]
* geometric object - a data description of a 3D or 2D shape.[3][X]
* statement context - the top-level syntax in OpenSCAD, composed of assignments, module instantiations, module definitions, and function definitions.[X]
* expression context - a lower-level syntax, including arithmetic, array references, function calls, and so on.[X]

[1] In today's OpenSCAD, the geometric objects are always added to the one-and-only CSG tree that forms the final model.  The geometry-as-data proposals envision ways to add geometric objects to other CSG trees.
[2] Sometimes called an associative array or map.  Akin to JavaScript objects, and to Python dictionaries and objects.  (I used to put C structures in this list, but C structures define a schema that applies to zero or more data values.  Objects here do not have a schema; each object is independent and self-describing.)  Today's OpenSCAD has a limited implementation - objects can only be created by internal OpenSCAD mechanisms; there are no OpenSCAD-language mechanisms for creating them.
[3] Geometric objects are not executable; once created they are static and immutable.  (Transforms and boolean operations create new geometric objects; they do not modify existing ones.)  In today's OpenSCAD, geometric objects are wholly separate from data values.  They implicitly form, and are formed from, the CSG tree.
[4] Module and function references are akin to C function pointers, and are akin to the data type resulting from uttering a function name in JavaScript or Python.

There are several related proposals:
* Geometry-as-data, #3956.  The goal is to have a way to represent a geometric object as a data value so that the program can examine it.  This brings a data-value representation of geometric objects into expression context.
* Objects, #3087, #3088.  The goal is to introduce objects as a data type.
* Module references, module literals, #4349 (and probably others).  The goal is to introduce module references, data values that refer to modules and can be called like traditional static modules.
* `object()` function, #4337.  Adds a function (no new syntax) to create objects from previous objects and other data values.  Roughly akin to `concat()` for arrays.
* [Unfiled proposal](https://docs.google.com/document/d/14siqbr9CyGuA2tY6dvTqiF3Hf9zFMyefk7YdaUJSn3Q).  Proposes merging the geometry-as-data and object efforts into a single syntax.

Are there any comments on these proposed definitions?  For the moment let's not talk about syntax details - let's just make sure we can use the same words for the same concepts.

jordanbrown0:
Updated previous glossary comment with a new location for my proposal: https://github.com/openscad/openscad/wiki/OEP7%3A-Unifying-Objects-with-Geometry-as-data

jordanbrown0:
Added definitions for CSG tree, mesh, and render.

kwikius:
> Are there any comments on these proposed definitions? For the moment let's not talk about syntax details - let's just make sure we can use the same words for the same concepts.

Sounds good to me. I have updated the code in https://github.com/kwikius/openscad/tree/WIP-first-class-module to (I hope) reflect these definitions

jordanbrown0:
Added:

* geometry reference - a data item that refers to a geometric object.[X][5]
* geometry literal - a syntactic structure that embeds a geometric object in an expression context, creating a geometry reference.[X]

 [5] For purposes of this glossary, it's unspecified whether a geometry reference refers to a CSG tree or a mesh. It's a shape that you can hand around like data; it is as yet unspecified exactly what operations you can do on it.

kwikius:
I got this to build with [some brute force](https://github.com/kwikius/openscad-object-literals)

(I don't think it would be too difficult to update its master brange and merge this in. [At the heart of it there are only around 5 or 6 files mainly in src/core](https://github.com/openscad/openscad/compare/object-literals) )

The example in [OP ](https://github.com/openscad/openscad/pull/3087#issue-502969809) works OK. 

The below also works. see https://github.com/kwikius/openscad-object-literals/blob/object-literals-wkg/examples/ObjectLiterals/object_literals1.scad versus https://github.com/openscad/openscad/issues/4351#issuecomment-1259080124

```openscad
extrusionInfo = [
   {position = [0,0,0]; rotation = [0,0,0]; extrusion = 1;},
   {position = [10,10,0]; rotation = [22.5,5,0]; extrusion = 2;},
   {position = [20,20,5]; rotation = [45,10,0]; extrusion = 3;},
   {position = [30,30,10]; rotation = [67.5,15,0]; extrusion =4;},
   {position = [40,40,15]; rotation = [90,20,0]; extrusion = 5;}
];
   
for(elem = extrusionInfo){
   translate(elem.position)
      rotate(elem.rotation){
         height = 40;
         echo(elem = elem.extrusion);
      }
}
```

However this isnt working ( but maybe the list comprehensions are more advanced now , since this is an old version)

```openscad
extrusionInfo = [
   {position = [0,0,0]; rotation = [0,0,0]; extrusion = 1;},
   {position = [10,10,0]; rotation = [22.5,5,0]; extrusion = 2;},
   {position = [20,20,5]; rotation = [45,10,0]; extrusion = 3;},
   {position = [30,30,10]; rotation = [67.5,15,0]; extrusion =4;},
   {position = [40,40,15]; rotation = [90,20,0]; extrusion = 5;}
];

//function obj_by_index(ob, i) = [ for (k=ob) ob[k] ][i];
function obj_by_index(ob, i) = ob[ [ for (k=ob) k ][i] ];

for(i = [0: len(extrusionInfo)-1]){
   elem = obj_by_index(extrusionInfo,i);
   translate(elem.position)
      rotate(elem.rotation){
         height = 40;
         echo(elem = elem.extrusion);
      }
}

PR #3102, opened 2019-10-23
Value rework 2

t-paul:
This is #2881 with conflicts resolved against master.


thehans:
Conflicts with master have been resolved again.
I made some significant changes to `VectorPtr` interface to hopefully simplify usage, and reduce the need to call `.clone()` in some cases.
I've also started  the re-visiting some of the review comments from the original PR.

One thing @jschobben pointed out in that PR is that `VectorPtr::clone()` is only copying the shared_ptr, which might cause problems.
I agree that this could be a source of errors, such as where some function returns a cloned value instance to its caller, which now has a non-const temporary value that it might `std::move` elements out of.  
So I'm thinking now, that as a *const* function, `Value::clone()` should probably necessitate a deep copy for vectors.
This would prevent callers from moving elements out of any *originating* `Value` that they shouldn't be modifiying, which likely came from a Context or other store.  Until now it has been very difficult for me to reason about exactly which cases it is safe to move elements like this, and doing this should make it *always* safe.
Also, I'm pretty sure that doing a deep copy would mean we can change `VectorPtr` to use `unique_ptr` instead of `shared_ptr`, which would be a nice bonus, or at least a small consolation for the potential performance hit of needing deep copies when `.clone()` is used.

For the other Value types that used `shared_ptr`: that is `RangePtr`, `FunctionPtr`, and `str_utf8_wrapper`, since these are not manipulated / built up via moving elements around in the way that vectors often are, I think it should be fine to leave these as ptr-only copies.

I also feel like many of these difficulties could be better solved by a more intelligent, "global"(per preview / render action probably) Context scheme, which @t-paul has mentioned possibly moving towards.
If results of each expression evaluation were cached in such a Context, then maybe all `Expression::evaluate()` functions signatures could be changed to return `const Value&` and references could just be passed up the call stack without any extraneous copies.  But that is more of a discussion for a whole other PR.

So right now I'm looking into this proposed deep copy / unique_ptr change.  
@t-paul I'd like to hear your thoughts on this, does it sound good to you?

jschobben:
Looking back at this, I'm not sure if we should always do a deep copy of the underlying vector, since that could have a pretty big performance penalty in some cases (maybe deep function recursion?).

That being said, it only matters when clone() is being used of course, but I guess it's used in many places. I think last year I tried to improve on this, but did not get very far so left it as-is.

What I'm trying to say is: deep copying is a trade-off of potential for bugs, vs performance. So, maybe my comment should be taken with a grain of salt, and performance is more important here :slightly_smiling_face: . This PR already makes the use of Value much more clear.

thehans:
So I already tried deep copy and one issue is of course a regression in this test: https://github.com/openscad/openscad/blob/dd7b1984ed8fb3b249e7d1db8820a93d2247d8ee/testdata/scad/misc/recursion-test-vector.scad#L1-L9
That blows up the stack before any of our current stack checking code catches it.
To continue with that idea would require some additional stack checking inside clone I guess.

So, one alternative sort of compromise would be to try only copying 1 level deep.  
I think the code which flattens or appends vectors by moving elements should only ever operate on top level elements, and not elements of elements etc, so this would hopefully be enough to keep things safe, but again it seems tricky to know if all the possible cases would be covered by that.

There is one last idea, which I've thought about before, but can't remember if I ever discussed.  The idea is to implement a way to "embed" vectors within others, without copying or moving their elements.
An embedded vector would be stored much the same as a normally nested vector, just with a small amount of added metadata. A custom iterator could be defined for `VectorPtr` which would treat an embedded vector as though its elements were a part of its parent, while the embedded vector itself would remain unmodified.  
The bracket operator would also need some extra code to handle offsetting of indices and for any embedded vectors.
In this scenario, `.clone()` would just copy the shared_ptr as its doing now, and the "embedded" vectors feature would basically be used to replace the current `VectorPtr::flatten` and `VectorPtr::append_vector` functions (the latter one I just added recently).
It would require a bit of extra code to handle this, but I'm thinking maybe it could be ideal in terms of memory usage and compute efficiency by avoiding copies.

nophead:
Doesn't the fact that vectors are immutable mean they never need to be copied, conceptually at least?

I vaguely remember it being an issue in TCP/IP stacks that each level of the protocol wraps the previous and that means prepending headers and adding trailers, so they went to a scatter-gather system to avoid lots of copying. Forgive me if my ramblings are not relevant.

jschobben:
@thehans your embedded vector idea sounds like a "persistent data structure", in which multiple versions of the data can exist while sharing memory where possible. That's a great idea, although it will probably have a noticeable impact on the constant factor in vector accesses.

@nophead in some cases you may want to have a slightly modified version of an immutable vector (similar to tcp/ip I guess but more generic), i.e. in some recursion cases where you build up a vector in the recursion. In simple cases a for-loop can work, but passing on a slightly-modified list in recursion requires a full copy on each iteration (last time I checked at least, which is several months ago :zipper_mouth_face: ).

Sometime ago I was working on another idea to avoid list copies, that builds on this PR (well, the pre-function literal version at least). Basically, the approach was to add a special "extract()" method to VectorPtr that would normally return a fresh copy of the vector and invalidate the inner pointer; but if the inner shared_ptr has a reference count of 1, then it would instead do a std::move (i.e. in constant time). Normally the ref count never becomes 1 though, because contexts also refer the VectorPtr; so I made it do some static code analysis to figure out when a Value can be safely removed from the local context, and do so. The end result being that you could essentially modify a list in openscad code in-place in tail-recursive functions, going from ```O(n^2)``` to the expected ```O(n)``` complexity.
Worked great for some simple list-append recursions, but especially the context-cleaning part required some dirty hacks. Also the "extract()" call has to be made carefully, only implemented it in a few places so far.
I should dust off those changes and put it in a draft PR sometime...

And of course the ```O(n)``` above only applies if the "modifying" works in constant time on the underlying std::vector, i.e. changing one element, or appending one to the end. I played with changing VectorType to std::deque, which had already a noticeable constant-factor impact; but it would make such "constant-time" edits more easy. Same would apply for using a persistent data structure, I bet.

Anyway it's probably best to keep these kind of improvements for a separate PR :slightly_smiling_face: .

thehans:
@nophead Consider this simple example:
```
a = [1,2,3];
b = [each a, 4,5,6];
echo(a);
echo(b);
```
In order to create `b`, the elements of `a` must be copied into it.  And if the code were to `std::move` them, we'd end up with undefined data in `a`

nophead:
I think the scatter gather scheme adds an indirection through a set of pointers to avoid copying, or moving. So ```b``` would have a reference to ```a``` as well as its own elements. 

But I think it might get too complicated for general list manipulation.

t-paul:
I was looking at persistent data structures (e.g. https://github.com/arximboldi/immer) which are supposed to handle that case. That would help even more for things like vector slicing. Maybe there's a simpler more specialized version for just vectors?

thehans:
Small update on the recent vector discussion:
@rcolyer and I had a brainstorming session about this, and we think that an "embeddable" vector type which flattens/copies *on first read access* should probably work very well.  It would have benefits such as constant time `concat`, and only a one-time penalty of `O(n)` for the first read access, after which it would be basically be a regular vector.
So even though there's potentially a big copy at some point, I think this should still save a decent bit of memory and compute time, since "Intermediate" levels of embedded vectors would not need their own flat copies as long as they are not read from.  Regular nested vectors can stay using `shared_ptrs` so those won't need deep copies either.
I'm working on an implementation of this idea now, and I'll just have to see how it performs.

thehans:
I'm happy to announce that the "embbedded vector" concept is now **a reality**, implemented as `class EmbeddedVectorType` in `value.h` / `value.cc`.  

I consider this a great success, expanding the power of the language by allowing much more efficient recursive building of vectors via concatenation, which was previously quadratic in performance!

Here is one test case/benchmark I created to measure performance against nightly snapshot builds for Ubuntu 18.04.
```SCAD
function up1(x) = x > 0 ? concat(up1(x-1), x) : [];
function dn1(x) = x > 0 ? concat(x, dn1(x-1)) : [];
function up2(x) = x > 0 ? [each up2(x-1), x] : [];
function dn2(x) = x > 0 ? [x, each dn2(x-1)] : [];

// proper tail recursion
function up3(x, v=[]) = x > 0 ? up3(x-1, concat(x,v)) : v;
function dn3(x, v=[]) = x > 0 ? dn3(x-1, concat(v,x)) : v;
function up4(x, v=[]) = x > 0 ? up4(x-1, [x, each v]) : v;
function dn4(x, v=[]) = x > 0 ? dn4(x-1, [each v, x]) : v;

/*
// Rework limits: (for release build, debug is a bit lower)
// First 4 function calls have NO TCO, so their input values correspond
// to the max I was able to run without "Recursion detected" errors.
a = up1(2865); // clang++11: 4334, g++8.4: 2865
b = dn1(2865);
c = up2(5317); // clang++11: 7698, g++8.4: 5317
d = dn2(5317); 
// 1e6 is the current arbitrary tail-recursion limit set inside OpenSCAD.
e = up3(1e6); 
f = dn3(1e6); 
g = up4(1e6);
h = dn4(1e6);
//*/
// Rework: Total rendering time: varies ~(5.5 to 6.2)seconds

//*
// Nightly snapshot limits:
//   2020.07.06.nightly (git ad5f91e)
a = up1(2713); 
b = dn1(2713);
c = up2(3847);
d = dn2(3847);
// Note: set to only 1/20th of rework's test 
// due to time constraints of quadratic performance!
e = up3(50000); 
f = dn3(50000); 
g = up4(50000);
h = dn4(50000);
// Nightly: Total rendering time: 0:05:25.321
// Rework:  Total rendering time: 0:00:00.584
//*/

for (x=[  a,b, c,d, e,f, g,h ]) {
  echo(len(x));
  echo(x[0], x[1000], x[len(x)-1], len(x));
  // use of [] indexing forces flattening, 
  // which is "more work" (but still VERY fast) for embedded vector implementation.
}
```
So in this case its roughly 500x faster than nightly.  Of course gains will be even more for larger vectors.  The commented out test goes up to 1million vector elements for the tail-call-optimized functions, and still completes in a few seconds.  I estimate that nightly would take days to do that!

Some other changes introduced in my most recent commits include:
 - `class VectorPtr` renamed (back) to `VectorType`.  
   `VectorType` is the name currently used in master, it was only changed to `VectorPtr` in previous instances of this PR, to correspond with the naming of the old `ValuePtr` class.  Now that the class provides an interface which behaves nearly identically to a true vector rather than a pointer, it made more sense to name it based on its usage, not the implementation detail that there's a `shared_ptr` in there.
 - More usage of range-based for loops over values via iterators.    
    1) `VectorType` now has an iterator which is aware of embedded vectors, and will only iterate over *true* vector elements (where I consider embedded vectors to be pseudo-elements).  This simplifies usage elsewhere in code, as a range-based for loop can be used.  Also a specialized `VectorType::emplace_back` function is used to handle the embedding of vectors.
    2) `RangeType` already had an iterator class, but wasn't yet using range-based for loops in code.
    3) `str_utf8_wrapper` now has its own iterator, (when iterating over a string as a sequence of individual characters) which replaces a previously used `utf8_split` function.
 - Added some class documentation in `value.h` comments.
 - A new `ValueMap` class (in `valuemap.h`), which provides a convenient interface for the `Context` class and the operations it requires (also now shared with "dxfdim" cache).
 - Other (mostly cosmetic) re-organizing of value.h

rcolyer:
I stress-tested the embedded vector improvements, and I concur that in typical usage (including in [a recursive design I made years ago and hit performance limits on](https://www.thingiverse.com/thing:1693052)) this improves recursive concatenation performance from approximately quadratic to linear.  The results are quite dramatic, making scales accessible that weren't before.  I also wrote some intentionally bad code to stress-test the worst case performance of the embedded vector by forcing it to flatten during each recursion, and while this pushed the embedded vector code back to quadratic, the constant multiple was still 3-4 times faster for me than the old performance.  So in the best case it scales dramatically better, and in the worst case the new code is still a lot faster.  The worst-case stress test code:
```
// Intentionally written poorly to stress value rework PR by forcing flattening.
function _integral_rec(v, i=0, arr=[]) = (i == len(v)) ? arr :
  let(
    nsum = len(arr)==0 ? v[i] : v[i] + arr[len(arr)-1]
  )
  _integral_rec(v, i+1, concat(arr, nsum));
function integral(v) = _integral_rec(v);

seed = 0;
_ = rands(0,1,0,seed);

data = rands(-1, 1, 40000);
int = integral(data);
```
I'm pretty confident the majority of naturally arising usages will fall into the best case scenario.  For example, the above worst-case code was actually a pre-existing routine which fell into the best case linear scenario in the way it was previously written.

t-paul:
The PDF export issue is a bit strange, including the output files in the artifact download showed the PDF is fine when opened via `evince`. It's not clear why `gs` chokes on it.

Issue #2113, opened 2017-09-17
New Feature: 'power of' syntax x^y, e.g x^2 etc. [$25]


MichaelAtOz:
I'm starting to need non trivial maths in some geometry calculations.
Have to do pow(x,y) is not as intuitive as x^y when transcribing equations such as
![image](https://user-images.githubusercontent.com/1972961/30518421-46711df8-9bc1-11e7-979f-19de4e4d3acd.png)

##

runsun:
I support this idea.

Corona688:
I wonder if it'd be easier to support the perl-style ** syntax.

doug-moen:
I guess that's a personal preference, depending on what languages you are
familiar with.

I'd prefer x^y, since all of the popular math and linear algebra languages
use this syntax. It's also what you use in Latex, or if you are editing
math equations on Wikipedia or in a math.stackexchange.com forum post. And
I do consider OpenSCAD to be a math language.

On the other hand, perl, python and ruby use x**y.

The R programming language supports both syntaxes: that's a potential
compromise in case this turns out to be a contentious issue.

kintel:
Challenge is that if you come from C/C++, `^` means XOR. Introducing this operator may actually make code _less_ readable at first glance..

nophead:
Can it be ** like Python instead.

thehans:
I implemented the operator as '^' for a few reasons.  First, since that was the original request, and second because I agree with @doug-moen that ^ is typical for general math syntax.  I also view OpenSCAD as more of math language than a C-family language.

Lastly this was my first time messing with yacc so I am still not sure if there's any special handling that would need to be done to differentiate between multiplication '*' and exponent '**' tokens.

MichaelAtOz:
I usually do my maths research in wikipedia initially. When you come across a relevant equasion you can edit that section and copy the maths bit, which as others have said uses `^`.
e.g. [Torus](https://en.wikipedia.org/wiki/Torus) 
![wiki torus](https://user-images.githubusercontent.com/1972961/31590642-7e924d6a-b25f-11e7-9735-2a7f412df335.PNG)

`<math> f(x,y,z) = \left(R - \sqrt{x^2 + y^2}\right)^2 + z^2 - r^2.</math>`
So `^` would be really handy.

MichaelAtOz:
Added $25 bounty, for '^'.

MichaelAtOz:
> I implemented the operator as '^' for a few reasons. 

[...]

PR #3104, opened 2019-10-25
Exponent test case for #2125

seam345:
Added test case to @thehans pull request #2125
For Issue #2113

I had to update theHans parser.y edits as there has been a lot of reformatting since his initial change.

this is my first edit to a compiler before I would recommend someone take a quick look over my merge, however all the tests ran the same as before the merge and it runs fine when I use openscad.


tests/regression/echotest/exponent-operator-test-expected.echo
```diff
@@ -0,0 +1,12 @@
+ECHO: 25
+ECHO: 25
```
t-paul: It looks like pretty much all languages assign a higher priority to exponentiation than unary, so it behaves as -(5 ^ 2). I guess that becomes more obvious when looking at the actual [math representation](https://www.wolframalpha.com/input/?i=-5**2).

[Python](https://docs.python.org/3/reference/expressions.html#operator-precedence) gives it a higher priority than unary
```
$ python3
>>> -5**2
-25
>>> (-5)**2
25
```
[JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence#Table) (v10.15.2) assigns a lower priority but nodejs can't even parse it without parenthesis (hu?)
```
$ nodejs 
> 5**2
25
> (-5)**2
25
> -(5**2)
-25
> -5**2
-5**2
  ^^
SyntaxError: Unexpected token **
```

MichaelAtOz:
The original request #2113 was very specific re '^' in regards to transcribing equations particularly maths from e.g. [wikipedia](https://github.com/openscad/openscad/issues/2113#issuecomment-336750990), LaTeX and other sites using it for markup.

> With the concerns about using ^

What concerns?

src/parser.y
```diff
@@ -419,20 +420,27 @@ addition
...
 multiplication
-        : unary
-        | multiplication '*' unary
+        : exponent
+        | multiplication '*' exponent
             {
               $$ = new BinaryOp($1, BinaryOp::Op::Multiply, $3, LOCD("multiply", @$));
             }
-        | multiplication '/' unary
+        | multiplication '/' exponent
```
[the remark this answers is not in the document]
seam345: it is now right-associative
I have updated the test to make sure both brackets and non brackets return 262144

t-paul:
See disucssion in issue #2113, specifically https://github.com/openscad/openscad/issues/2113#issuecomment-336727814.

tests/regression/echotest/exponent-operator-test-expected.echo
seam345: it now executing unary after exponent, however I had to make an awkward exception for 5**-2 as the parser wasn't accepting it as valid syntax.
With my very limited experiance with bison I am unsure of a nicer way to resolve this, so have settled on this line:
https://github.com/seam345/openscad/blob/24b9d404b4309c18aaba4006d2f885d477606cdc/src/parser.y#L465

seam345:
I have addressed all the execution problems in 1d62878
and have updated it to the ** operatorin 24b9d40

tests/regression/echotest/exponent-operator-test-expected.echo
t-paul: Right, that does look like it might have other side effect. I'll have a look if there's a way without special cases.

MichaelAtOz:
`^ == XOR` in C?
See https://rosettacode.org/wiki/Operator_precedence `^` seems more common to anyone coming from other than C/python. How frequently will exponent be used v's XOR in OpenSCAD, `^` is a much better fit.

src/parser.y
```diff
@@ -450,6 +454,18 @@ unary
...
+       : call
+       | call TOK_EXPONENT exponent
+           {
+              $$ = new BinaryOp($1, BinaryOp::Op::Exponent, $3, LOCD("exponent", @$));
+           }
+       | call TOK_EXPONENT '-' exponent
+          {
+             $$ = new BinaryOp($1, BinaryOp::Op::Exponent, new UnaryOp(UnaryOp::Op::Negate, $4, LOCD("negate", @$)), LOCD("exponent", @$));
+          }
+       ;
```
t-paul: I need to do some more testing, but the following seems to work without any special cases and agrees with the results producted by `python3` for everything I've tried so far.
```
exponent
       : call
       | call TOK_EXPONENT unary
```
Looking at this [python grammar](https://docs.python.org/3/reference/grammar.html) it also matches the `power` declaration.
```
factor: ('+'|'-'|'~') factor | power
power: atom_expr ['**' factor]
atom_expr: [AWAIT] atom trailer*
```
This way we can have any `unary` in the expression.
```
$ python3
>>> 4 ** -+- 3 ** 2
262144
>>> 4 ** +-+ 3 ** 2
3.814697265625e-06
```
Lower priority expressions are resolved without problems too.
```
$ python3
>>> 4**3*3**2
576
>>> 4**3+3**2
73
```
Technically this allows also the unary `!`, I'm not sure if it would make sense trying to prevent that.

thehans:
Thanks for carrying this forwards @seam345 

> `^ == XOR` in C?

FWIW, I agree with Michael's sentiment on this.
A couple other points I would make in favor of `^`:

1. No other bitwise operators exist in OpenSCAD, why should a user assume `^` to be a bitwise operator(and the only one at that)?
2. As far as I know, no one in the history of OpenSCAD has requested or had a direct need for bitwise operators in the language.  (Searching github issues and forum.openscad.org for "bitwise" I came up with nothing)
3. I feel that `^` is more universally recognized/interpreted as exponentiation than anything else.  C/Python are more like the exceptions, not the rule.
4. I think its perfectly OK to expect a C user to recognize and adapt that OpenSCAD is not an 1:1 replica of C syntax.

t-paul:
> See https://rosettacode.org/wiki/Operator_precedence `^` seems more common to anyone coming from other than C/python.

Bad example. Not counting the duplicates, the numbers are pretty much identical when just looking at what is directly listed on the page (13 for `**`, skipping one of the Algols and 12 for `^`, counting the various Basic variants as 1).

IRC discussion with Julia pointed out that ImplicitCAD already accepts `^` so that would be a plus for that one.

kintel:
I'm ok with moving forward with `^`

t-paul:
@seam345 do you have the time to update this PR? As far as I can tell it's only the 2 open things, a) changing back to `^` (sorry about the back and forth discussion, core language changes are always critical as there's no way back once they are out in the world; but there's finally concensus on this) and b) the parser change from https://github.com/openscad/openscad/pull/3104#pullrequestreview-307589962 pointing back from `exponent` to `unary` to get rid of the need of special cases.

seam345:
@t-paul Yes I should have some time over the weekend to sort out the #3104 review, and changing back to ^ is just a simple revert, I'll just quickly do that now.

src/parser.y
seam345: I have implemented the solution you suggested above in the latest commit, haven't had the chance to see what ! would do but all the test pass.

t-paul: Thanks, for all tests I can currently think of, we get the same result as with `python` which is probably a good reference.

PR #3110, opened 2019-10-30
Install Start shortcut for all users on Windows

### LINKED ISSUE #2763: OpenSCAD not included on Start Menu - Windows 10.
I'm not sure that this is an OpenSCAD issue. I've tried treating it as a Windows issue and haven't had any luck coming up with a solution.

This is with Windows 10, upgraded from Windows 7.

Whether performing an install as a user with Administrator permission allowed or as an administrator directly, the OpenSCAD is not available from the start menu. The following is observer:

The OpenSCAD link is present in the start menu of the administrative user used to install.
The OpenSCAD link is present in the "Default" user start menu directory
The OpenSCAD link is not present for any other user.

I'm not sure if it's relevant but openscad is not present in C:\ProgramData\Microsoft\Windows\Start Menu or subdirectories.

This isn't a show-stopper as I can create my own links. I have had no problem installing on Windows 7.

KeenRivals:
The default shell context in NSIS is for the current user. This causes $SMPROGRAMS to put the link in the current user's profile. Other users on the system get no start menu link.

Docs for SetShellVarContext at https://nsis.sourceforge.io/Docs/Chapter4.html 
Fixes #2763

### DIFF
--- a/scripts/installer.nsi
+++ b/scripts/installer.nsi
@@ -17,6 +17,7 @@ File /r /x mingw-cross-env locale
 File /r /x mingw-cross-env color-schemes
 File /r /x mingw-cross-env templates
 ${registerExtension} "$INSTDIR\openscad.exe" ".scad" "OpenSCAD_File"
+SetShellVarContext all
 CreateShortCut $SMPROGRAMS\OpenSCAD.lnk $INSTDIR\openscad.exe
 WriteUninstaller $INSTDIR\Uninstall.exe
 # see https://msdn.microsoft.com/en-us/library/aa372105(v=vs.85).aspx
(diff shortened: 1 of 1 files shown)


t-paul:
How complicated would it be to make that an option with default to just the current user? I'm not sure forcing for all users works for everyone.

KeenRivals:
I don't know for sure. I'm not super good at NSIS. I made this PR because I'm trying to deploy OpenSCAD in a school with the silent install option. I saw the user in the referenced issue was having the same problem so I thought this might be a quick fix for others.

Some installers provide a checkbox to choose an all user install or a current user install, based on that they can do a few things:
- If it's an alluser install they'll elevate to admin, install to the C:\program files directory and create an alluser start shortcut. 
- If it's a currentuser install they'll install to %appdata%\local and create a current user start menu. Apps that install this way can avoid requesting admin access.

OpenSCAD's installer is a bit inconsistent currently because it acts like an alluser install except for the shortcut. It requests admin, installs to programfiles, and creates a system-wide uninstaller, but then it puts the shortcut in the currentuser profile. Making the shortcut system-wide by default would make OpenSCAD's entire install process system-wide by default.

I think to support both flows the installer would have to:

1. Show a checkbox for all/current user.
1. Provide default install locations based on all/current user.
1. Provide a silent switch to select the install type.
1. Store the type of installation somewhere so the uninstaller can handle it.

RichyT:
Nice. Thanks guys.

PR #3115, opened 2019-11-02
Array/String slicing and half open ranges.

t-paul:
This adds slicing notation for arrays and strings. Syntax is based on OpenSCAD ranges, so it looks like `[begin:end]` or `[begin:step:end]` (however step is not yet implemented for slicing). Both begin and end can be omitted and default to begin and end of the base object respectively.

Usage is very similar to `python`, but not identical. Mainly because OpenSCAD ranges are by default closed intervals where python slicing definitions are half-open. Important to note is that the behavior for negative values is different for slicing as compared to ranges.

To allow more flexibility the end defintion for OpenSCAD ranges now allow to specify a half-open interval by prefixing the end value with `<`. This is similar to Swift ranges that use `...` for closed and `..<` for half-open intervals. For OpenSCAD that looks like `[begin:<end]` and `[begin:step:<end]`.

ToDo:
- [ ] Decide on negative values for simple indexing `a = v[-1]`
- [ ] Decide on dropping the deprecated auto swap on ranges (slices will never do that)
- [ ] Implement step for slicing
- [ ] Handle UTF8 strings correctly
- [ ] Add warning for not supported data types
- [ ] Add error when trying to use non integer values for slicing
- [ ] Add test cases
- [ ] Mark as experimental

For later:
- Find vector implementation which supports at least the basic slicing operations ([immer](https://github.com/arximboldi/immer), [immutable-cpp](https://github.com/rsms/immutable-cpp))

**Examples**
```
a = [0,1,2,3,4,5,6,7,8,9];

echo(all = a[:]); // ECHO: all = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
echo(first = a[0:0]); // ECHO: first = [0]
echo(last = a[-1:-1]); // ECHO: last = [9]
echo(skip_first = a[1:]); // ECHO: skip_first = [1, 2, 3, 4, 5, 6, 7, 8, 9]
echo(skip_last_inclusive = a[:-2]); // ECHO: skip_last_inclusive = [0, 1, 2, 3, 4, 5, 6, 7, 8]
echo(skip_last_exclusive = a[:<-1]); // ECHO: skip_last_exclusive = [0, 1, 2, 3, 4, 5, 6, 7, 8]
echo(range1 = a[2:8]); // ECHO: range1 = [2, 3, 4, 5, 6, 7, 8]
echo(range2 = a[-6:6]); // ECHO: range2 = [4, 5, 6]
```
**Implementation of fold/reduce**
```
a = [1, 2, 3, 4, 5, 6, 7, 8, 9];

reduce = function(v, i, f) len(v) > 0 ? reduce(v[1:], f(i, v[0]), f) : i;

sum = function(v) reduce(v, 0, function(x, y) x + y);
prd = function(v) reduce(v, 1, function(x, y) x * y);
cnt = function(v) reduce(v, 0, function(x, y) x + 1);
rev = function(v) reduce(v, [], function(x, y) concat(y, x));

echo(sum = sum(a), prd = prd(a), cnt = cnt(a), rev = rev(a));

// ECHO: sum = 45, prd = 362880, cnt = 9, rev = [9, 8, 7, 6, 5, 4, 3, 2, 1]
```

### DIFF
--- a/src/comment_parser.y
+++ b/src/comment_parser.y
@@ -63,11 +63,11 @@ expr:
     }
     | '[' num ':' num ']'
     {
-        $$ = new Range($2, $4, Location::NONE);
+        $$ = new Range($2, new Literal(1.0), $4, false, Location::NONE);
     }
     | '[' num ':' num ':' num ']'
     {
-        $$ = new Range($2, $4, $6, Location::NONE);
+        $$ = new Range($2, $4, $6, false, Location::NONE);
     }
     ;
 
--- a/src/expr.cc
+++ b/src/expr.cc
@@ -294,33 +294,43 @@ void Literal::print(std::ostream &stream, const std::string &) const
 }
 
 Range::Range(Expression *begin, Expression *end, const Location &loc)
-	: Expression(loc), begin(begin), end(end)
+	: Expression(loc), exclusive(false), begin(begin), end(end)
 {
 }
 
-Range::Range(Expression *begin, Expression *step, Expression *end, const Location &loc)
-	: Expression(loc), begin(begin), step(step), end(end)
+Range::Range(Expression *begin, Expression *step, Expression *end, bool exclusive, const Location &loc)
+	: Expression(loc), exclusive(exclusive), begin(begin), step(step), end(end)
 {
 }
 
 ValuePtr Range::evaluate(const std::shared_ptr<Context>& context) const
 {
-	ValuePtr beginValue = this->begin->evaluate(context);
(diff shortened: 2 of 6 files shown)


MichaelAtOz:
I didn't recall 'half-open interval', for those in the same boat, half-open == exclusive of end point.
See https://en.wikipedia.org/wiki/Interval_%28mathematics%29

> Decide on negative values for simple indexing a = v[-1]

Sounds logical & useful.

Issue #3112, opened 2019-10-31
CRASH - Display CSG Products with large number of objects


MichaelAtOz:
I was using Ronaldo's cache demo code from the [forum](http://forum.openscad.org/When-modules-are-cached-td27677.html), and a got a crash after F5, Design/Display-CSG-Products.
2019.05 & 2019.10.25 64bit, Window7/64.
![Capture crash - display CSG Products - stack overflow](https://user-images.githubusercontent.com/1972961/67914858-73c6c600-fbe5-11e9-9e26-d65da87672fc.JPG)
Google says C...FD is Stack Overflow.
It does not crash with `cube();` or one of my complex models, I presume 50,000 is the problem.
Code:
```
//C();
B();
A(50000) 
C();
//B();

module C() square(10,center=true);
  
module B() { 
   projection() 
    linear_extrude()
      intersection(){
        projection() rotate([10,20,30]) cube(1000,center=true); 
        projection() rotate([20,30,10]) cube(1000,center=true); 
        projection() rotate([30,10,20]) cube(1000,center=true); 
        square(10,center=true);
      }
  }
  
module A(n=0)
  for(i=[for(i=0;i<n;i=i+1) i]) rotate(360*i/n) children();

```

thehans:
Can confirm that this crash also occurs on Linux (Mint 19/ Ubuntu 18.04)  
And yes, its a stack overflow issue due to recursion in `std::string CSGOperation::dump()` function from `src/csgnode.cc`.

Got this from a stack trace on a debug build:
```
[Thousands of frames at src/csgnode.cc:166]
...
#16360 0x00005555556f1722 in CSGOperation::dump[abi:cxx11]() (this=0x555561365e70) at src/csgnode.cc:166
#16361 0x00005555556f1722 in CSGOperation::dump[abi:cxx11]() (this=0x555561365f20) at src/csgnode.cc:166
#16362 0x00005555556f1776 in CSGOperation::dump[abi:cxx11]() (this=0x555561365fd0) at src/csgnode.cc:166
#16363 0x00005555559618a5 in MainWindow::actionDisplayCSGProducts() (this=0x5555566bff10) at src/mainwin.cc:2210
[...]

PR #3122, opened 2019-11-08
Use iteration for CSGOperation::dump and custom CSGOperation deleter

thehans:
For issue #3112

### DIFF
--- a/src/csgnode.cc
+++ b/src/csgnode.cc
@@ -63,6 +63,21 @@ primitives, each having a CSG type associated with it.
 	A CSGProduct is a vector of intersections and a vector of subtractions, used for CSG rendering.
 */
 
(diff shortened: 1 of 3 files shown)


src/csgnode.cc
```diff
@@ -159,19 +150,62 @@ std::string CSGLeaf::dump()
 	return this->label;
 }
 
-std::string CSGOperation::dump()
+// Recursive traversal can cause stack overflow with very large loops of child nodes,
+// so tree is traverse iteratively, managing our own stack.
+std::string CSGOperation::dump()
```
kintel: Optional: It feels like this method should be const

thehans: Agreed, I tried making it const early on, before I had finalized what exact types I was using on stack etc, but ran into some snags.  Just gave it another try and got it working this time.

PR #3138, opened 2019-11-19
Issue350 Lazy union (aka. no implicit union)

thehans:
I've updated the original code from kintel, and implemented a few fixes and additional features.

This PR implements lazy union on `for`, `children`, `echo`, `assert`, `let`, and `if[/else]` builtin modules.

Other modules are still planned to be converted to lazy-union, but can be done as separate PRs so we can make some incremental progress.

For more info see recent discussion in issue #350


src/openscad.cc
```diff
@@ -429,10 +429,11 @@ int cmdline(const char *deps_output_file, const std::string &filename, const cha
...
 			if (!root_geom) root_geom.reset(new CGAL_Nef_polyhedron());
 			if (viewOptions.renderer == RenderType::CGAL && root_geom->getDimension() == 3) {
-				auto N = dynamic_cast<const CGAL_Nef_polyhedron*>(root_geom.get());
-				if (!N) {
-					N = CGALUtils::createNefPolyhedronFromGeometry(*root_geom);
-					root_geom.reset(N);
+				if (!dynamic_pointer_cast<const CGAL_Nef_polyhedron>(root_geom) &&
+						!dynamic_pointer_cast<const GeometryList>(root_geom)) {
+					// FIXME: We should probably recursively convert all leaf nodes in the list
+					// to Nef polyhedron
```
kintel: We may want to deal with this FIXME

kintel:
In terms of documenting changes in behavior, we should highlight the following constructs that will behave differently:
* difference() for(...)
* interection() for(...)
* minkowski() for(...)

src/openscad.cc
thehans: OK, I changed this section.  It was a little tricky since nested geometries are const, but I ended up calling `GeometryList::flatten` first (you had implemented it in the original code, but it wasn't yet used anywhere), then looping through that to convert and reset shared_ptrs where necessary.  I don't 100% know the reason to force all CGAL Nef, but if we are doing this, should the conversion maybe be moved into `GeometryEvaluator::evaluateGeometry` instead so that export from GUI works the same?  Maybe even cache the converted result?

kintel: This was an  old comment, but I think the  rationale was that since the cmd-line arguments specifically asks for a CGAL evaluation, we should probably do that. `viewOptions.renderer == RenderType::CGAL`

kintel: ..and that option doesn't exist in the GUI

kintel:
@thehans Feel free to merge if you don't have any more planned work on this

src/openscad.cc
thehans: ok, makes sense

PR #3144, opened 2019-11-25
Allow initial seeds to stick between rands calls

rcolyer:
The current code uses lessdeterministic_rng with an initial time-based seed if no seed value is given to a particular rands call, and uses deterministic_rng with the specified seed applied each time a seed is given.  This changes to a single deterministic_rng, initially set with the same time-based seed, but a specific seed is given each time one is provided by a rands call.

With this change, seeded rands calls will behave identically.  Programs which do not seed will also behave identically.  However, this change results in unseeded rands calls inheriting the seeded sequence from a previous rands call.  This permits the valuable option of setting an initial seed (e.g., _ = rands(0,1,1,seed); ), and allowing consistent random values to be generated throughout the program without having to continually come up with new seeds, or pre-generate and distribute very large amounts of random numbers across a complicated program.


MichaelAtOz:
So this is for repeatability? Do `mySeed=1234; initialSeed=rands(0,1,1,mySeed);`, then all subsequent un-seeded `rands()` will generate identical results as the last Preview/Render??

t-paul:
Nice change. Extra usability with less code.

I wonder if we would want to switch `boost::` to `std::` while we are at it?

rcolyer:
Correct Michael, that's the result with this.

kintel:
Yeah, porting from boost to C++11 should be trivial and would be nice.

rcolyer:
I switched it over to the equivalent stdlib rng.  I then updated the tests that changed because the stdlib has a different uniform_real_distribution implementation, which changed some of the test result values.  I think this should be good for a merge now.

PR #3149, opened 2019-11-30
Range Warnings begin<end

MichaelPFrey:
based on http://forum.openscad.org/Infamous-warnings-td28054.html

- [x] remove the now redundant code in value.cc
- [x] consider generalizing warning
   - (just focusing on literals for now)
   - Warn for
      - [x] `for(i=[10:1])` -> swap, deprecated
      - [x] `for(i=[1:-1:10])` -> warning
      - [x] `for(i=[1:0:10])` -> bad range warning (not very specific, but "fair enough" with the line number)
   - correct (do not warn)
      - [x] `for(i=[0:10])` -> fine
      - [x] `for(i=[10:-1:1])` -> fine
   - not addressed with this PR
      -  expression containing variables like `for(i=[0:a])` `for(i=[a:0])` `for(i=[0:a:10])`  `for(i=[a:1:10])`  `for(i=[0:1:a])` 
      - non-numbers like
         - for(r=[1:true:5]) echo(r);
         - for(r=[1:"test":5]) echo(r);
         - for(r=["a":"z"]) echo(r);
- [x] test cases
   - [x] (check list from above)
   - [x] manually review changed test cases
- [x] move the warning to a NOINLINE function to minimize memory footprint?
- [x] rephrase warnings for consistency

### DIFF
--- a/src/expr.cc
+++ b/src/expr.cc
@@ -316,10 +316,10 @@ ValuePtr Range::evaluate(const std::shared_ptr<Context>& context) const
 	if (beginValue->type() == Value::ValueType::NUMBER) {
 		ValuePtr endValue = this->end->evaluate(context);
 		if (endValue->type() == Value::ValueType::NUMBER) {
+			double begin_val = beginValue->toDouble();
+			double end_val   = endValue->toDouble();
+			
 			if (!this->step) {
-				double begin_val = beginValue->toDouble();
-				double end_val   = endValue->toDouble();
-				
 				if(end_val < begin_val){
 					std::swap(begin_val,end_val);
 					PRINT_DEPRECATION("Using ranges of the form [begin:end] with begin value greater than the end value is deprecated. %s", loc.toRelativeString(context->documentPath()));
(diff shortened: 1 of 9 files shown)


nophead:
Isn't for(i=[10:-1:1]) the legal way to iterate backwards? I thought what was deprecated was automatically changing the implicit increment to -1 when the range was backwards. Not banning backwards ranges altogether.

MichaelPFrey:
concate-tests has something interesting:
Bildschirmfoto von 2019-11-30 13-59-57-1

`echo(concat([3, 4, true, 6], [4:1:3]));`

for-tests has something interesting:
```
// Illegal step value
for(r=[1:true:5]) translate([r*10-60,50,0]) cylinder(r=r);
```

some more examples:
Bildschirmfoto von 2019-11-30 14-15-41-1

MichaelPFrey:
> Isn't `for(i=[10:-1:1])` the legal way to iterate backwards?
Yes it is - I should separate out "should not create warnings" from "should create warnings" and "not my concern right now" in my to do list to be more clear.

Issue #3023, opened 2019-08-01
ERROR: CGAL error in CGAL_Nef_polyhedron3():


mogra09:
Hello!
I have a problem.
Please help me to solve the problem.

platform: Windows10-64bit, OpenSCAD version 2019.05

ERROR: CGAL error in CGAL_Nef_polyhedron3(): CGAL ERROR: assertion violation! Expr: e->incident_sface() != SFace_const_handle() File: /mxe/usr/x86_64-w64-mingw32.static.posix/include/CGAL/Nef_S2/SM_const_decorator.h Line: 329

the code:
```
module cover()
{
 CubePoints = [
  [ 9.5, 0  ,1  ], //0
  [-9.5, 0  ,1  ], //1
  [-9.5,21.3,1  ], //2
  [-6.0,26  ,1  ], //3
  [ 6.0,26  ,1  ], //4
  [ 9.5,21.3,1  ], //5
  [ 1.8, 6  ,2.5], //6
  [-1.8, 6  ,2.5], //7
  [-1.8,15  ,2.5], //8
  [ 1.8,15  ,2.5], //9
  [ 9.5, 0  ,0  ], //10
  [-9.5, 0  ,0  ], //11
  [-9.5,21.3,0  ], //12
  [-6.0,26  ,0  ], //13
  [ 6.0,26  ,0  ], //14
  [ 9.5,21.3,0  ]  //15
 ];
 CubeFaces = [
  [0,1,7,6],[1,2,8,7],[2,3,8],[3,4,9,8],[4,5,9],[0,5,9,6],[6,7,8,9],
  [0,1,11,10],[1,2,12,11],[2,3,13,12],[3,4,14,13],[4,5,15,14],[5,0,10,15],
  [10,11,12,13,14,15]
 ];
 difference(){
  polyhedron( CubePoints, CubeFaces );
  hull(){
   translate([0, 6,-1]) cylinder(d=4, h=4, $fn=16 );
   translate([0,15,-1]) cylinder(d=4, h=4, $fn=16 );
  }
 }
} 
cover();
```

t-paul:
Have you checked that all faces have the correct winding order? -> [FAQ Entry](https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/FAQ#Why_is_my_model_showing_up_with_F5_but_not_F6?)

mogra09:
Hello t-paul
Thank you very much for your kind advice.
I don't understand the writing order and now solved with your advice.

```
[...]

Issue #2873, opened 2019-03-16
Polyhedron does not render when rendering with another object


xesxen:
Given this SCAD file, the prism is showing on the preview render, but it disappears when rendering.

Sometimes, the console shows `UI-WARNING: Object may not be a valid 2-manifold and may need repair!` (only two triangles of the polyhedron will be visible in that case when rendering), but most of the times it will say `ERROR: CGAL error in CGAL_Nef_polyhedron3(): CGAL ERROR: assertion violation! Expr: e->incident_sface() != SFace_const_handle() File: /usr/include/CGAL/Nef_S2/SM_const_decorator.h Line: 330` (with the polyhedron not rendering at all)

```OpenSCAD
union(){
    cube([5,2,4]);
    translate([0,0,4])
    triangular_prism(5,2,3);
}

module triangular_prism(x, y, z)
polyhedron(
    points=[
        [0,0,0], [0, y, 0], [0, y, z],
        [x, 0, 0], [x, y, 0], [x, y, z],
    ],
    faces=[
        [0,1,2],
        [3,4,5],
        [0,2,5,3],
        [0,1,4,3],
        [1,2,5,4]
    ]
);
```

t-paul:
The polyhedron has some faces pointing in the wrong direction. See related [FAQ](https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/FAQ#Why_is_my_model_showing_up_with_F5_but_not_F6?) entry.

Issue #2847, opened 2019-03-05
Segmentation fault on rendering


plut:
openscad 2019.02.07, libcgal13 4.13-1+b2 (using the Debian repositories mentioned in http://www.openscad.org/downloads.html).

Running on the following file (which is a stupid model, granted) reliably produces a segfault when rendering, either in the GUI or in command-line via `openscad bug.txt -o bug.stl`.

[bug.txt](https://github.com/openscad/openscad/files/2932430/bug.txt)

Since the model is correctly previewed, might this be a CGAL bug instead?

t-paul:
The polyhedron is invalid (some faces are pointing in the wrong direction - [see this FAQ entry](https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/FAQ#Why_is_my_model_showing_up_with_F5_but_not_F6?)). It should never crash OpenSCAD though.

kintel:
I can reproduce on macOS using CGAL 4.11. It works (reports error instead of crashing) with CGAL 4.8.1.

This was tested using old dev snapshots of OpenSCAD, so there's theoretically a possibility that we have code changes influencing this, but the stack trace points quite strongly to an uncaught (uncatchable?) issue inside CGAL.

kintel:
I suspect this to be the culprit: Someone commented out an assertion check in CGAL:

https://github.com/CGAL/cgal/commit/f1f82785acb767ed3f0c7918ab19b45771ec4a5d#diff-2d25c6c5eefcde7a443ef7269d33ee5eL210

kintel:
If someone can organize an upstream bug report that would be awesome

adrianVmariano:
I found that this code

    for (a=[0:120:240])
       rotate(a)
          polyhedron(points = [[0,.0911,0],[1,0,0],[0,-.0911,0],[0,0,.1]],
              faces = [[0,1,2], [1,2,3], [0,1,3], [0,3,2]]);

crashes  2019.01-RC3 under Windows when I attempt to render.  The first face of the polyhedron is backwards.  If I fix that it doesn't crash any more.  Is this the same bug?

MichaelAtOz:
Did anyone 'organize an upstream bug report'?

thehans:
I have now ^^^

> Did anyone 'organize an upstream bug report'?

thehans:
Well, that didn't quite go as planned.  Already closed and completely ignored the issue of commented-out assertion which I described as carefully as I could.

[...]

Issue #2616, opened 2018-12-01
Incorrect polyhedron causes segmentation fault


sschroe:
With the current master i get a segmentation fault when trying to render the following file with an incorrect polyhedron:

```
cube(1);

polyhedron(
       points=[[3, 0, 0],
               [5, 0, 0],
               [8, 6, 0],
               [0, 6, 0],

               [3, 0, 2],
               [5, 0, 2],
               [8, 6, 3],
               [0, 6, 3]],

       faces=[ [0, 1, 2, 3],
               [0, 4, 5, 1],
               [1, 2, 6, 5],
               [3, 0, 4, 7],
               [2, 3, 7, 6],
               [4, 5, 6, 7]]
);
```
This polyhedron contains several incorrect faces that are backwards. With the 2015 release (and also older library versions) openscad does not crash but the polyhedron simply does not show up in the render (which is expected). I also checked an commit from around August and the crash seemed to exist there as well so it does not seem to be an very recent issue.

#2464 seems to report a very similar crash.

A debug build on commit 1270e5a6b5b with gdb yields the following:
```
#0  0x00005555557b057e in CGAL::HalfedgeDS_halfedge_base_base<CGAL::HalfedgeDS_list_types<CGAL::Cartesian<CGAL::Gmpq>, CGAL::I_Polyhedron_derived_items_3<CGAL::Polyhedron_items_3>, std::allocator<int> >, CGAL::Boolean_tag<true>, CGAL::Boolean_tag<true>, CGAL::Boolean_tag<true> >::opposite() (this=0x0) at /usr/include/CGAL/HalfedgeDS_halfedge_base.h:56
#1  0x00005555559665c4 in CGAL::opposite<CGAL::Cartesian<CGAL::Gmpq>, CGAL::Polyhedron_items_3, CGAL::HalfedgeDS_default, std::allocator<int> >(boost::graph_traits<CGAL::Polyhedron_3<CGAL::Cartesian<CGAL::Gmpq>, CGAL::Polyhedron_items_3, CGAL::HalfedgeDS_default, std::allocator<int> > >::halfedge_descriptor, CGAL::Polyhedron_3<CGAL::Cartesian<CGAL::Gmpq>, CGAL::Polyhedron_items_3, CGAL::HalfedgeDS_default, std::allocator<int> > const&) (h=...) at /usr/include/CGAL/boost/graph/graph_traits_Polyhedron_3.h:367
[...]

PR #3152, opened 2019-12-06
Perform some consistency checks in polyhedron primitive

tturpin:
This is a proposal to check for several geometric issues in the polyhedron primitive and output warnings which refer to the arguments. It follows my comments about this on #2351 and in [the forum](http://forum.openscad.org/User-Poll-What-do-you-want-to-see-from-OpenSCAD-development-td27802i60.html).

The implementation may not be of the best quality, but it demonstrates the idea. Here are the checks that are performed and cause an error or warning:
- the points must be vectors of 3 finite doubles (check unchanged)
- indices in the points vector must be in range (this was previously included in the "Unable to convert point ..." check)
- faces must have at least 3 vertices
- faces must not have consecutive identical vertices
- faces must not self-intersect
- polyhedron must be 2-manifold (this check is only an approximation, it uses the edge dictionary and will miss cases where 2n faces meet on the same edge, half of them with each half-edge).
- duplicate vertices are checked, I'm not sure if there should be a warning (this is explicitly supported in the manual)

Some other issues are not checked, most notably intersecting faces (other than self-intersection) and planarity of faces. But the implemented checks (unconnected edges in particular) give a sensible explanation for: #2351, #2398,  #2616, #2847, #2873, #3023. Unconnected edges are also reported for #2395, #2412 but in those two cases this is not the only issue.

I added a few cases to polyhedron-tests.scad, and looked at the messages ; other tests already trigger some warnings too (issue802.scad, line 109 for duplicate vertices, issue1246.scad, primitive-inf-tests.scad...)

I tried to keep the behavior unchanged, other than the messages. One exception is that invalid vertices (not a vector of 3 finite doubles) are now checked before starting to build the polyhedron, so the empty polyhedron is return in this case (previously the function returned whatever had already been inserted in the polyhedron). The returned result in case of errors warnings could deserve more reflection to either be more strict or try to "fix" some issues better (I'm not qualified for that as I don't know how robust CGAL is to various issues).

Let me know if the idea sounds reasonable !


thehans:
You updated the test case but it doesn't look like the expected result images are actually different, so I would just revert those pictures to previous version if that's true.

Also if the changes are just intended to add warnings, then you should probably edit tests/CMakeLists.txt to have that test included with the other echotests

tturpin:
Thank you for the hint about echotests, I hadn't found the way to test output. This made me realize that the new warnings / errors did not appear in the echo output, and were not properly handled in the GUI, as I implemented them in PrimitiveNode::createGeometry and not in PrimitiveModule::instantiate, where most errors and warnings are. I have submitted a new version with all checks in instantiate, and with echotests updated (and polyhedron-tests added). I also reverted the image tests.
Please note that I don't really know C++, so there are probably many mistakes in my code, especially about memory which I don't understand at all ;-)

tturpin:
Oups, this second commit has a serious bug, I'm currently running all tests before submitting again.

PR #3160, opened 2019-12-18
rotate(from=VECTOR,to=VECTOR)

revarbat:
I've added `from=` and `to=` arguments to the `rotate()` module, which can take vectors.  The `rotate()` function still works the same way with `a=` and `v=` arguments, but it now also allows you to do something like `rotate(from=[0,0,1], to=[1,1,0]) cube();` or `rotate(a=30, from=[0,1,0], to=[1,1,1]) cube();`.  The `a=` parameter is only allowed to be a number when used with `from=`/`to=`, and is the number of degrees to rotate around the `from=` axis before rotating around the axis perpendicular to both `from=` and `to=` by enough to go from `from=` to `to=`.


nophead:
This is commonly done in user code in a simpler way: https://github.com/nophead/NopSCADlib/blob/master/utils/sweep.scad#L53 avoiding the need for acos. Not sure if the result is the same.

revarbat:
In C, the two methods look like they should be pretty close in speed. one has 2 more cross products, the other has near equivalent length calculations that add on 2 total `sqrt()`s.  The `acos()` is likely offset by the extra matrix multiplcation in the other.

revarbat:
The point is more, though, that a newbie to OpenSCAD will find it far easier to point a thin cylinder at the corner of a cube with `rotate(from=[0,0,1], to=[1,1,1]) cylinder(d=1,h=10);` than it would be for them to look up the right math, or find the right library code and figure out how to install it.  Note: yes, my own BOSL2 library has a `rot(from=, to=)` as well.

revarbat:
Actually, the length calculations are somewhat slower than a dot product, due to the `sqrt()`s.  They are somewhat less than twice as fast as the cross products, though.

nophead:
Yes but the issue is there are an infinite number of ways to rotate from
one vector to another because rotation around the Z axis can be chosen
arbitrarily. The common user land method does the minimum rotation
possible. Although that is not always what is wanted, perhaps the built in
version should do the same for consistency.

revarbat:
The one I created does the minimal rotation from one vector to the other.  It also allows spinning around the starting vector first, to allow compensation.
```openscad
rotate(a=30,from=[0,0,1], to=[1,1,1]) cylinder(d=1,h=10);
```
should be nearly exactly equivalent to:
```openscad
rotate(90-atan2(1,sqrt(2)), v=[-1,1,0]) rotate(30) cylinder(d=1,h=10);
```
(code just off the top of my head.)

```openscad
rotate(from=[0,0,1], to=[1,1,0]) cube([10,10,0.1], center=true);
```
will make leave a corner of the flat square plate pointing upwards, while
```openscad
rotate(45, from=[0,0,1], to=[1,1,0]) cube([10,10,0.1], center=true);
```
will leave an edge parallel to the XY plane on top.

revarbat:
Actually, there may be an error in my code for exactly opposing vectors.I'll check.

t-paul:
What's the benefit of adding this to the built-in `rotate()` over having a user level function defined?

revarbat:
The benefit is in not requiring newbie users to write non-intuitive code, for a behavior that seems like it *should* be simple, and intuitive. A simple matter of pointing a shape at the corner of a cube shouldn't take trig to do.

Since MCAD doesn't seem to have something for this, the end user's only library option is to dive into the deep end, figuring out how to install a third-party library.  Useful, but you're adding to the learning curve.

This isn't an obscure use case, either.  I use `rot(from=,to=)` from my own library *all the time*.  I use it so ubiquitously, that my library of shapes can all take `spin=` and `orient=` arguments, and I have standard `UP`, `DOWN`, `LEFT`, `RIGHT`, `FWD`, and `BACK` vectors defined.

I'm a library author, and even *I* think this should have been built-in.

t-paul:
That sounds more like an argument for better documentation and easier library access.

revarbat:
# *Why, yes. Yes it is.*

(Markdown lacks enough emphasis options to properly display my frustration.)

To be quite honest, this entire PR has been a experiment in attempting to end-run around the extremely vocal everyone-should-write-their-own-library crowd that severely discourages a proper standard library.

One of three things has to happen if OpenSCAD is to progress in any meaningful way:

1. Add a UI for downloading, installing, and managing libraries like the Arduino app has. This is the extremely preferred solution.
2. Include a more comprehensive standard library with proper documentation. I’ll gladly donate BOSL2 to the cause if you want, once I finish revising it. Or get nophead’s library if you can. As long as it is a comprehensive and well documented library. 
3. Add useful but non-obvious features to OpenSCAD itself, so the libraries aren’t needed. This is what this PR is a test case for. 

You need to choose.

kintel:
@revarbat Making libraries easier to manage, as well as bundling a good standard library are indeed the preferred ways forward. Both are quite resource intensive, so right now it's more of a wish than a plan. Adding C++ features really just increases the existing maintenance burden.

t-paul:
Ok, so I'm closing this now as I think the consensus is that the only sustainable way forward is to have better library support and some kind of package management.

As for a standard lib, I don't think it makes sense to use BOSL or NopSCADlib. Both are awesome and I'm promoting them best I can, but they are way too specialized and big for a standard lib that needs to have synchronized releases with the application and no way to change API after it's out. I believe there's a number of things that would fit well into this kind of library like `torus()` or `hexagon()` but certainly not `screw()` or `nema_stepper()`. Imposing those kind of limitations onto BOSL or any other library would massively impact their possibilities to progress.

Issue #3118, opened 2019-11-03
Tail recursion limit not working


t-paul:
```
sin = function(x) sin(); 
echo(sin(30)); 
```

thehans:
I did some debugging and profiling and found that the limit counter is counting just fine, but its just slowing to a crawl before reaching the limit, because it spends ever increasing time in prepareTailCallArguments every time it recurs.

Problem is `this->defaultArguments` gets another duplicate `x` variable added to it for each recursion(Line 556)
Then all those get inserted into `variables` (Line 564).
Then the vast majority of runtime is spent updating the hashmap of `tailCallContext`  setting `x=undefined` repeatedly (Lines 570-572).
https://github.com/openscad/openscad/blob/2a943a57f9d23511c0021b228dee6b6eb9ae440d/src/expr.cc#L554-L572

So the problem is clear, but I'm a little confused about the overall handling of contexts and variables here to know the proper fix.

thehans:
@jschobben It looks like this regression is a result from some of your tail recursion changes in #3020.  Any ideas how to best resolve this?

jschobben:
Hmm, this looks like a silly bug. Line 549-559 is supposed to only execute once (sort of a lazy init or cache). Maybe with the function literals that now happens each time for this example. The "cache" is tied to a node in the AST, which probably works differently in this new use case.

A fix should probably be in restoring the "running once" of the if-block. Might be a bit tricky with the dynamic-ness of function literals.

Right now I'm not around my pc for a while, building openscad on my phone works but takes up to 2 hours :D
I'll look into this when back home, in about 2 weeks, if it's still open then.

jschobben:
A temporary workaround might be to change `defaultArguments` from vector to map, which should fix the quadratic behavior but not the proper usage of the cache. Not sure about importance of argument order though, has to be checked.

jschobben:
Or maybe the call to `EvalContext->resolveArguments` on line 552 just returns an empty map? That would also cause the if-block and line 556 to always execute. If so then maybe tracking the "not-filled-yet" status of `resolvedArguments` just needs to be improved (better than `.empty()`), or that for-loop in 554-558 needs an extra guard.

I need my debugger :D

thehans:
Yes `EvalContext::resolveArguments` returns a AssignmentMap based on the number of parameters from the *call site* which in @t-paul's example is empty: `sin()`.  

[...]

PR #3161, opened 2019-12-20
Issue3118: Tail recursion limit not working

thehans:
These changes address issue #3118 and some other things that came up while looking into it.

`EvalContext::resolveArguments` now automatically include default expressions from function/module definition so they don't need to be looped over separately when setting variables.
The return type has changed to a vector rather than a map, which I think should be slightly faster and also keeps same argument order from function/module definition.
`resolveArugments` also emites more comprehensive and specific argument override warnings, with new test case.

`expr.cc:evaluate_function` has also had `c_local` "stack" handling simplified.  Object lifetime should already be handled by "context.parent", so no need to manage a separate stack.


src/expr.cc
```diff
@@ -546,30 +546,39 @@ FunctionCall::FunctionCall(Expression *expr, const AssignmentList &args, const L
...
+	variables.reserve(this->resolvedArguments.size());
 	// Set default values for unspecified parameters
-	variables.insert(variables.begin(), this->defaultArguments.begin(), this->defaultArguments.end());
 	// Set the given parameters
-	for (const auto &ass : this->resolvedArguments) {
-		variables.emplace_back(ass.first, ass.second->evaluate(context));
+	for (const auto &arg : this->resolvedArguments) {
+		variables.emplace_back(arg.second.name,
+				arg.second.expr ?
+				(arg.first ? arg.second.expr->evaluate(context->getParent()) : arg.second.expr->evaluate(context)) :
```
thehans: This is one of the parts I'm still unsure if its 100% correct.

`arg.first` is a bool where true indicates that the assignment is defaulted, so it should be evaluated in the context of the function *definition* rather than the function call.  So the code evaluates from `context->getParent()`, which is the same way that `Context::setVariables` handles default arguments(both before and after my changes). 

I'm just not sure if the parent context is really correct or sufficient in the case of some recursive function where the function itself could be re-defined within its own `let()` expressions.

thehans:
The original bug is now addressed by #3181 with much less code changes.

I will keep this one open since I still think the additional warnings by resolveArguments are useful, and the context stack is unnecessary.  But I will need to resolve some conflicts now, and probably change the behavior of defaultArguments to be evaluated by UserFunction/FunctionDefinition instead of by the FunctionCall.

src/expr.cc
thehans: Test case from IRC:
`foo = function() let(foo = function() 5) foo(); //infinite recursion`

redlizard:
#3745 includes a version of the additional warnings and more systematic parameter parsing that was still unmerged, so I think that should close off this PR.

thehans:
As discussed on IRC, I would mainly just like to see the `arg-permutations.scad` test from this PR make it into master before closing this out.

PR #3164, opened 2019-12-22
Some comparison operator fixes

thehans:
See updated expected test result for effective changes.


kintel:
The primary breaking change here is the behavior of `true == 0` or `1` and `false == 0` or `1`.
Do you have a good idea if/how often this is encountered in the field, to gage what breakage we can expect to see?

thehans:
The discussion of fixing the comparison operators came up while I was thinking about implementing sorting algorithms for a library, and wondering if comparisons could be robust in the case where someone happens to mix different types in a single vector (and decided to sort their weird vector).  

However, I'd guess that actual mixing of bool and num in that way happens incredibly rarely to never in practice.

More likely would be a case where `nan` or `undef` get accidentally mixed in with some numbers and failed to sort, but handling that is outside the scope of this PR.


The old behavior of evaluating :
 - `false == 0` as `false `
 - `true == 1` as `false`
-  `false != 0` as `true`
 - `true != 1` as `true` 

...was particularly messed up because it also evaluated:
 -  `false <= 0` to `true` 
 - `false >= 0` to `true`
 - `true >= 1` to `true`
 - `true <= 1` to `true`

Having something be "less than or equal to" a value, **and** "greater than or equal to" the same value, **but not** just plain "equal to" didn't make any sense.  
The new expected results(first 4 bullet points are now inverted, second 4 remain same) are now consistent with most other common languages as far as I know, so I think its a safe bet for users to expect that behavior if the case ever arises.

I would predict the change to be pretty low impact overall, since apparently no-one has noticed/complained about these inconsistencies before.

Also if someone still needs to absolutely differentiate between false vs 0 or true vs 1, the `is_bool` and `is_num` functions can be used; their behavior hasn't changed.

Finally, just to clarify (maybe just a typo on your part), but the results of `true == 0` and `false == 1` haven't changed.  Those  remain `false`.

kintel:
Thanks for the thorough answer. @t-paul do you feel there's any significant risk in merging this?

t-paul:
I'm not sure, but with the current slightly random behavior, I hope there's not that much dependency on it. I think moving to a saner result for comparison which is documented and testable is nice.

PR #3169, opened 2019-12-25
Patches

t-paul:
More patches from master.

This includes the C++14 update as otherwise it would not build on an recent Linux distro that started shipping CGAL5 (e.g. Debian testing, Fedora Rawhide, ...).


t-paul:
Whoops, wrong target branch.

PR #3178, opened 2020-01-02
AssignmentList now holds shared_ptr to Assignment

thehans:
Carried from AST-WIP branch, and mostly separated from other AST changes.

### DIFF
--- a/src/localscope.cc
+++ b/src/localscope.cc
@@ -16,7 +16,7 @@ LocalScope::~LocalScope()
 {
 }
 
-void LocalScope::addChild(shared_ptr<ASTNode> &&node)
+void LocalScope::addChild(shared_ptr<ASTNode> node)
 {
 	// FIXME: Move this out of the ASTNode subtype
 	if (auto modinst = dynamic_pointer_cast<ModuleInstantiation>(node)) {
--- a/src/localscope.cc
+++ b/src/localscope.cc
@@ -20,16 +20,16 @@ void LocalScope::addChild(shared_ptr<ASTNode> node)
 {
 	// FIXME: Move this out of the ASTNode subtype
 	if (auto modinst = dynamic_pointer_cast<ModuleInstantiation>(node)) {
-		this->addModuleInst(modinst);
+		this->addModuleInst(std::move(modinst));
 	}
 	if (auto module = dynamic_pointer_cast<UserModule>(node)) {
-		this->addModule(module->name, module);
+		this->addModule(std::move(module));
 	}
 	if (auto function = dynamic_pointer_cast<UserFunction>(node)) {
-		this->addFunction(function);
+		this->addFunction(std::move(function));
 	}
 	if (auto assignment = dynamic_pointer_cast<Assignment>(node)) {
-		this->addAssignment(assignment);
+		this->addAssignment(std::move(assignment));
 	}
 // 	FIXME: Add remaining
 	this->children.emplace_back(std::move(node));
@@ -41,11 +41,11 @@ void LocalScope::addModuleInst(shared_ptr<ModuleInstantiation> modinst)
 	this->children_inst.emplace_back(std::move(modinst));
 }
 
-void LocalScope::addModule(const std::string &name, shared_ptr<class UserModule> module)
+void LocalScope::addModule(shared_ptr<class UserModule> module)
 {
 	assert(module);
-	this->modules.emplace(name, module);
-	this->astModules.emplace_back(name, std::move(module));
+	this->modules.emplace(module->name, module);
+	this->astModules.emplace_back(module->name, std::move(module));
 }
 
 void LocalScope::addFunction(shared_ptr<class UserFunction> func)
--- a/src/localscope.cc
+++ b/src/localscope.cc
@@ -16,7 +16,7 @@ LocalScope::~LocalScope()
 {
 }
 
-void LocalScope::addChild(shared_ptr<ASTNode> node)
+void LocalScope::addChild(shared_ptr<ASTNode> &&node)
 {
 	// FIXME: Move this out of the ASTNode subtype
 	if (auto modinst = dynamic_pointer_cast<ModuleInstantiation>(node)) {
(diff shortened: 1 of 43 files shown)


src/localscope.cc
```diff
@@ -14,34 +14,50 @@ LocalScope::LocalScope()
...
+	if (auto module = dynamic_pointer_cast<UserModule>(node)) {
+		this->addModule(std::move(module));
+	}
+	if (auto function = dynamic_pointer_cast<UserFunction>(node)) {
+		this->addFunction(std::move(function));
+	}
+	if (auto assignment = dynamic_pointer_cast<Assignment>(node)) {
+		this->addAssignment(std::move(assignment));
+	}
+// 	FIXME: Add remaining
```
kintel: FIXME can be killed I think

src/localscope.cc
```diff
@@ -78,9 +94,10 @@ std::vector<AbstractNode*> LocalScope::instantiateChildren(const std::shared_ptr
 	NB! for loops are special as the local block may depend on variables evaluated by the
 	for loop parameters. The for loop code will handle this specially.
 */
-void LocalScope::apply(const std::shared_ptr<Context> ctx) const
+void LocalScope::apply(const std::shared_ptr<Context> &ctx) const
 {
+	// FIXME: Filter assignments before doing this
```
kintel: Kill FIXME

src/localscope.h
```diff
@@ -11,23 +12,30 @@ class LocalScope
...
+	void addChild(shared_ptr<ASTNode> astnode);
+private:
+	void addModuleInst(shared_ptr<class ModuleInstantiation> astnode);
+	void addModule(shared_ptr<class UserModule> module);
+	void addFunction(shared_ptr<class UserFunction> function);
+	void addAssignment(shared_ptr<class Assignment> assignment);
+public:
+	void apply(const std::shared_ptr<Context> &ctx) const;
 	bool hasChildren() const {return !(children.empty());};
+	void resolveAssignments();
```
kintel: Leftover from WIP branch

src/parser.y
```diff
@@ -184,15 +184,16 @@ statement
         | '{' inner_input '}'
         | module_instantiation
             {
-              if ($1) scope_stack.top()->addChild($1);
+              if ($1) scope_stack.top()->addChild(std::move(shared_ptr<ModuleInstantiation>($1)));
```
kintel: I don't think std::move() does anything here. The called function does perform the actual move.

src/parser.y
```diff
@@ -184,15 +184,16 @@ statement
...
             }
         | assignment
         | TOK_MODULE TOK_ID '(' arguments_decl optional_commas ')'
             {
-              UserModule *newmodule = new UserModule($2, LOCD("module", @$));
+              shared_ptr<UserModule> newmodule = make_shared<UserModule>($2, LOCD("module", @$));
```
kintel: `auto newmodule =`

src/parser.y
```diff
@@ -184,15 +184,16 @@ statement
...
               newmodule->definition_arguments = *$4;
-              scope_stack.top()->addModule($2, newmodule);
+              auto top = scope_stack.top();
               scope_stack.push(&newmodule->scope);
+              top->addChild(std::move(newmodule));
```
kintel: Redundant std::move()?

src/parser.y
```diff
@@ -202,8 +203,8 @@ statement
             }
         | TOK_FUNCTION TOK_ID '(' arguments_decl optional_commas ')' '=' expr ';'
             {
-              UserFunction *func = new UserFunction($2, *$4, shared_ptr<Expression>($8), LOCD("function", @$));
-              scope_stack.top()->addFunction(func);
+              shared_ptr<UserFunction> func = make_shared<UserFunction>($2, *$4, shared_ptr<Expression>($8), LOCD("function", @$));
```
kintel: `auto func =`

src/parser.y
```diff
@@ -202,8 +203,8 @@ statement
...
+              scope_stack.top()->addChild(std::move(func));
```
kintel: redundant move?

src/parser.y
```diff
@@ -303,7 +304,7 @@ child_statement
         | '{' child_statements '}'
         | module_instantiation
             {
-                if ($1) scope_stack.top()->addChild($1);
+                if ($1) scope_stack.top()->addChild(std::move(shared_ptr<ModuleInstantiation>($1)));
```
kintel: redundant move?

src/parser.y
```diff
@@ -719,52 +716,52 @@ void handle_assignment(const std::string token, Expression *expr, const Location
...
-			assignment.setLocation(loc);
+			assignment->expr = shared_ptr<Expression>(expr);
+			assignment->setLocation(loc);
 			found = true;
 			break;
 		}
 	}
 	if (!found) {
-		scope_stack.top()->addAssignment(Assignment(token, shared_ptr<Expression>(expr), loc));
+		scope_stack.top()->addChild(std::move(assignment(token, shared_ptr<Expression>(expr), loc)));
```
kintel: Redundant move?

src/parser.y
thehans: The called function takes shared_ptr by value, so to move it the value out of parser.y without copying I thought this was necessary?  Are you sure it would still do the move automatically?  

As far as I understand, each call to a function taking a shared_ptr by value would need to use std::move to avoid copy by value.

I could also change the `LocalScope::add*` function signatures to take `shared_ptr<...>&&` arguments which I think should be functionally equivalent with maybe a bit clearer interface, but I think the move would still be required at each call site to relinquish its value to the callee?

thehans: Nevermind, resolved in IRC conversation

src/localscope.h
```diff
@@ -13,12 +13,12 @@ class LocalScope
...
+	void addChild(shared_ptr<ASTNode> &&astnode);
 private:
-	void addModuleInst(shared_ptr<class ModuleInstantiation> astnode);
-	void addModule(shared_ptr<class UserModule> module);
-	void addFunction(shared_ptr<class UserFunction> function);
-	void addAssignment(shared_ptr<class Assignment> assignment);
+	void addModuleInst(shared_ptr<class ModuleInstantiation> &&astnode);
+	void addModule(shared_ptr<class UserModule> &&module);
+	void addFunction(shared_ptr<class UserFunction> &&function);
+	void addAssignment(shared_ptr<class Assignment> &&assignment);
```
kintel: With your latest changes, I think these should go back to pass-by-value.

thehans: The latest changes were only on calls from `parser.y` to `LocalScope::addChild`,  I made these all private so they should only be called directly by addChild, which still does `std::move` on the specific `dynamic_pointer_cast` variables.

So, I think it would make sense to change addChild back to by-value now, but not these?

thehans: Or could the calls in addChild also be changed in some way to avoid std::move?

kintel: ah, I overlooked that. You're right - these are private methods and it all makes sense now :)

..so just changing addChild() to by-value should be good!

PR #3181, opened 2020-01-04
Minimal fix for #3118  Tail recursion limit not working

thehans:
This small change addresses the issue defaultArguments growing without bounds when a call has no arguments, which was causing recursion limit to not be reached in any reasonable time frame.


kintel:
One comment: To make sure this doesn't happen again, should we add a testcase?

thehans:
OK, testcase added.

Issue #2842, opened 2019-03-04
Inconsistent F6 behavior: polyhedra with holed faces


RonaldoCMP:
A good way to represent polygons with holes is by what we call a "fake simple polygon" (FSP) (I guess nophead call it a "keyhole representation"). It is a polygon whose edges do not cross but some edges are used twice when the polygon edges are circulated. Her is an example:

p = [ [100,0], [0,100], [-100,0], [0,-100], [100,0], [50,0], [0,-50], [-50,0], [0,50], [50,0] ];

Although a FSP is not a simple polygon, it is acceptable by the primitive polygon() and can be extrude by linear_extrude and rotate_extrude to model a solid with a internal hole.

However, the render of polyhedrons where some faces are FSP are subject to a variety of results under F6. A striking example follows:

r = 0.0001;
verts =  [[0, 0, -50], [0, 50, 0], [0, 0, 50], [0, -50, 0], 
          [100, 0, -50], [100, 50, 0], [100, 0, 50], [100, -50, 0], 
          [0, 0, -12], [0, -12, 0], [0, 0, 12], [0, 12, 0], 
          [200, 50, -50], [177.639, 94.7214, 0], [200, 50, 50], [222.361, 5.27864, 0],
          [200, 50, -12], [205.367, 39.2669, 0], [200, 50, 12], [194.633, 60.7331, 0], 
          [100, -12, 0], [100, 0, 12+r], [100, 12, 0], [100, 0, -12+r]] ;
         
 faces = [ [0, 4, 5, 1],    [1, 5, 6, 2],         // 4 external faces
           [2, 6, 7, 3],    [3, 7, 4, 0], 
           [9, 20, 21, 10], [10, 21, 22, 11],     // 4 internal faces
           [11, 22, 23, 8], [8, 23, 20, 9],      
           [0, 1, 2, 3, 0, 8, 9, 10, 11, 8],      // back face with hole
           [7, 6, 5, 4, 7, 20, 23, 22, 21, 20]];  // front face with hole
 
polyhedron(verts, faces);
cube(50);

As is, this code generates the expected result on F6: a cube added to a extrusion of a polygon with a hole. But if the value of the variable r is changed to zero, the render will show the polyhedron missing the holed faces.

thehans:
Your 2d example seems like an odd hack, as you can already represent holes if you supply multiple paths.
```openscad
[...]

PR #3191, opened 2020-01-18
Initial support for multi-export.

thehans:
Allows multiple `-o` options from same command line.  

I did a quick manual test, but probably should have a test or two added.  
I think it will require some more extensive changes to the test framework or a new test "type" since all previous tests assume single output.  I'm just not really sure the best way to go about that yet.

### DIFF
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -269,7 +269,7 @@ void set_render_color_scheme(const std::string color_scheme, const bool exit_if_
 	}
 }
 
-int cmdline(const char *deps_output_file, const std::string &filename, const char *output_file, const fs::path &original_path, const std::string &parameterFile, const std::string &setName, const ViewOptions& viewOptions, Camera camera, const std::string &export_format)
+int cmdline(const char *deps_output_file, const std::string &filename, const std::string &output_file, const fs::path &original_path, const std::string &parameterFile, const std::string &setName, const ViewOptions& viewOptions, Camera camera, const std::string &export_format)
 {
 	Tree tree;
 	boost::filesystem::path doc(filename);
@@ -827,7 +827,7 @@ int main(int argc, char **argv)
 
 	auto original_path = fs::current_path();
 
-	const char *output_file = nullptr;
+	vector<string> output_files;
 	const char *deps_output_file = nullptr;
 	std::string export_format;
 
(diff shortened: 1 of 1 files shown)


src/openscad.cc
```diff
@@ -1048,7 +1044,9 @@ int main(int argc, char **argv)
 				rc = info();
 			}
 			else {
-				rc = cmdline(deps_output_file, inputFiles[0], output_file, original_path, parameterFile, parameterSet, viewOptions, camera, export_format);
+				for(auto output_file : output_files) {
+					rc = cmdline(deps_output_file, inputFiles[0], output_file, original_path, parameterFile, parameterSet, viewOptions, camera, export_format);
```
kintel: While this is safe (since it recompiles for every output file), it can also be quite inefficient.
Also keep in mind that this can yield unexpected results for designs that echo'es information based on the `$preview` variable.

I don't like the implications of this. My initial hunch would be to start only allowing .echo as the second output format (and perhaps other safe outputs).

thehans: It might be a little bit inefficient, but AFAIK any GeometryCache will persist across those calls, as long as its the same process.

One of my motivations for this PR overall was that it could potentially be used to help test cache-related issues such as the one discussed here: https://github.com/openscad/openscad/issues/2842#issuecomment-469530635
By doing something like `-o foo.png -o foo.stl` forces preview before F6 for example, which would give different results (in that case) from only `-o foo.stl`.

I guess I'm not familiar with how echo from command line and `$preview` work together normally, so I don't know if this has potential to change it.

kintel:
Example to think about:
```
$ cat test.scad
echo("Start");
if ($preview) echo("Preview");
else echo("Render");
echo("End");
```


```
$ openscad test.scad -o out.png --preview
ECHO: "Start"
ECHO: "Preview"
ECHO: "End"
```

```
$ openscad test.scad -o out.png --render
ECHO: "Start"
ECHO: "Render"
ECHO: "End"
```

..but `openscad test.scad -o out.echo` currently thinks it's always in "render" mode. I guess we forgot to think about this when adding $preview. Maybe now is a good time to clean that up :)

nophead:
When outputting echo files I set $preview with a -D because there is no
obvious automatic choice for OpenSCAD to set it automatically.

Issue #2915, opened 2019-04-04
ERROR: CGAL error in CGALUtils::applyBinaryOperator union: CGAL ERROR: assertion violation! Expr: itl != it->second.end() File: /usr/include/CGAL/Nef_3/SNC_external_structure.h Line: 1106


gratefulfrog:
I have some simple code that won't compile and produces the above error.

Any help would be greatly appreciated.

If I execute ```all(20);```using the preview, it works perfectly, if I use Render, either it renders nothing, or gives the above error.

I am using the nightly builds on Ubuntu 16.04 version 2019.04.04.nightly (git 6bfcb5e) 

I would be happy to help if I can.
Thanks,
Bob

[donutTests.scad.zip](https://github.com/openscad/openscad/files/3045335/donutTests.scad.zip)

MichaelAtOz:
I don't have an answer yet, but interestingly, the first OpenSCAD version I tried it with happened to be 2018.11.04.ci339, which worked, both F5 & F6. Then tried 2015.03, errored, then 2019.01-RC4 errored.

MichaelAtOz:
Here is a workaround `$fn=32`.
I think this is probably another floating point misalignment, that changes the angle of incidence of the cylindrical shape with the squared edge.

MichaelAtOz:
But if you use View/Thrown-together, there is purple showing, so I suspect a STL will be bad...
Yep:
![Capture donuttests](https://user-images.githubusercontent.com/1972961/55591766-3058c600-5782-11e9-9a53-9e05f717450a.JPG)

MichaelAtOz:
2018.11.04 doesn't get the error, but produces a bad STL.

The sphere, cylinder & donut is where the problem is.

MichaelAtOz:
I've pared it down to: (note I increased the size * 10)

```
$fn=32;
hullRad = 15;
angle   = 20;
cylH     = 400;
e=0.2;

rotate([0,angle/2.,0])
  translate([hullRad,0,0]){
    // donut
    #rotate([-90,angle/2.,0])
      rotate_extrude (angle = 180-angle)
        translate([1*hullRad,0,0])
          circle(r=hullRad);
    // angleCyl
    a=-angle; h=cylH;
    ratio = a == 0 ? 1 : abs(a)/a;
      translate([0,0,0])
        rotate([0,a/2.,0])
          translate([hullRad*ratio,0,-e]){
            cylinder(h=h,r=hullRad);
            *translate([0,0,h])
[...]

PR #2399, opened 2018-05-22
Another attempt at multi-threaded geometry evaluation

justbuchanan:
This builds on @devilman3d's work in https://github.com/openscad/openscad/pull/1980 and re-uses most of the general changes to openscad. This PR rewrites the `ThreadedNodeVisitor` class to be a bit simpler and more efficient (and possibly resolve a couple bugs).

Note that I haven't addressed the gmpq thread-safety issues pointed out in the other PR. Does anyone have a good example scad model that triggers this? I haven't run into it in the many examples I've tried, but I also haven't done anything to specifically fix the issue.

The new implementation starts a fixed number of worker threads at the start of traversal and schedules work amongst them, which is a bit more efficient than starting a new thread for each item of work.

As before, the geometry tree is traversed top-to-bottom recursively, running prefix traversal for each of the nodes on the way down. This happens serially, ensuring that each parent's prefix traversal happens before any of it's children.

The postfix traversals are usually (much) more expensive to run than the prefix traversals and these are what we run in parallel. The important constraint on running these in parallel is that a given node's postfix traversal can not be run until the postfix traversals for each of the child nodes is run.

As the tree is recursively traversed top-to-bottom, a tree of pending postfix traversals is built. In order to keep track of which nodes are ready to run, each node keeps a counter of how many child nodes have yet to complete their postfix traversals. Each time a postfix traversal is run, the pending child count of it's parent node is decremented. When the counter gets to zero, that node is pushed onto the work queue and is executed as soon as a thread is available.

I've run some tests of the new implementation on several scad models and can see a fairly significant speedup, especially for complex models. Simple models often run a little faster with single-threaded evaluation because there's not much to parallelize and the multi-threading code adds some overhead. Below are some timing comparisons run on my computer of several example models. Note the last one is a large scad model that I downloaded from https://github.com/CarlosGS/Cyclone-PCB-Factory.

These numbers are from my computer which has a fairly fast cpu (Intel Core i7-6700K). Please try it out on your machines and report how well it works for you. I add a quick script at `scripts/thread-comparison-report.md` you can use to run several examples.

Example|Serial|Parallel
---|---|---
examples/Old/example001.scad | 1.4s | 1.6s
examples/Old/example002.scad | 0.2s | 0.2s
examples/Old/example003.scad | 0.2s | 0.1s
examples/Old/example004.scad | 0.4s | 0.4s
examples/Old/example005.scad | 2.6s | 3.0s
examples/Old/example006.scad | 104.6s | 74.5s
[...]

PR #3193, opened 2020-01-19
Parallelize render using multiple processes

slaviber:
Hello,
I tried to implement multiprocess support for OpenSCAD according to #391. 
(With all the previous failed attempts, maybe we should start working in **parallel** to succeed...)

### This is my approach:
First of all, it is obvious for me that if there can be any huge optimization/parallelism, that is in the CGAL library itself - by computing the operations in small independent chunks. But unfortunately, GCAL is not even thread-safe.  
There is not much room for optimization of OpenSCAD. Of course, that is because one real CAD model has a closely linked structure - every next geometric element depends on one or rarely a few previous. So  there isn't much to be computed in parallel on the 'operation' level.  
Of course, there are exceptions. And I'm making use of them to achieve a nice (solid) decrease in the rendering time in some corner cases. At almost negligible performance cost for the rest of the time.  
  
I first did some performance tests and I found that many constructions parsed by the OpenSCAD language are reduced to Unions before other operations are being applied to them. For example all **for** loops, all the geometry in the sourcefile's root and so on. (And maybe even more in the future).   
And I also found that some optimizations can be done before even making calls to CGAL (ex. combining the PolySets in the Unions into multivolume objects, thus reducing the Boolean OR overhead.).  
  
### The actual implementation:
So my implementation resides entirely in the **cgalutils-applyobjects** sources mainly because that's the only place where the 3D objects can be parallelized to some extent.  

First I'm taking all order-independent operations of a given branch (Unions and Intersections).  
Then I'm merging the objects wherever possible, which compensates for the added overhead of the consequent operations.  
Then, if there are enough 3D objects left, i'm spawning child processess, which compute the operations in parallel wherever possible, then combining child processess and so on until the number of objects left is insufficient for spawning more childs.  
I'm combining the left 3D objects serially and returning the root.  
  
#### That's one of those corner cases:  
[box.scad](https://gist.githubusercontent.com/slaviber/5cc68978950f3a857c692aa24b4f0215/raw/132e03b855256a78bfee2bb3fb05aedcc03eb865/box.scad)
 That box has one *patterning* operation - union of 30 holes and one Intersection of 6 spheres.  
- First the Intersection gets evaluated: 3 worker threads in parallel, then serial combination. 
- Then some non-optimizable Differences are evaluated
- Then comes the pattern - depending on the complexity of the patterning object, there may be one big merged union, or as in this case, again: 15 worker threads, then 7 worker threads, then 4 and so on until only < 4 objects are left which are then combined serially.  
- ...
  
When I did the test on my PC the results were - 15'47'' for the master branch and  8'15'' for my PR. That's almost double increase in rendering speed (at the expense of all that RAM for all those workers in parallel).  
  

The actual results vary from roughly equal rendering time for non-branched low-polygon count objects, to slightly slower for moderately-branched low polygon count objects to massive reduce in time for high polygon count branched models.


kintel:
Thanks for the PR!
We don't yet use clang-format on this codebase, so this would be easier to review if you didn't pull in the reformatting commit.

slaviber:
Thanks for the response!
I'm sorry, didn't know about that. It's never too late to make another PR with just my copy-pasted changes without any (re)formatting. 
By the way, almost all automated builds are failing because Boost.Process (for spawning the child processess) requires boost version 1.64 or later. Which is almost 3 years old now but seems that OpenSCAD is compiled against an older version.

t-paul:
We can't easily use boost 1.64. We can update the one listed in the readme which has currently 1.35, but the base line is probably 1.58 (as of Ubuntu 16.04LTS).

slaviber:
I understand. If Ubuntu 16.04LTS has to be supported, then i'd need to use either QProcess (which seems hard) or another 3rd-party multiplatform library that supports multiprocessing.

t-paul:
Ubuntu 16.04LTS is mostly required because it's currently still the base for AppImage and Snap. It's possible to update dependencies for those (and we already need to do that), but this really needs to be justified as the effort spend on upgrading the build environments is not exactly small.

A separate process seems to have some advantages over the threading solution in #2399. Do you think the benefits are bigger than the drawbacks (like the need to serialize the data structures to the other processes).

slaviber:
Too much code is not thread safe. I think that separate processes are the only possible solution for now.   
The serialization overhead is negligible compared to the CGAL operations. Even if a multi-threaded solution was possible, the serialization is almost pure RAM > RAM copy anyway, which is fast. 
  
Below are some *perf* log images of my PR, stdlib, gmplib and CGAL are the biggest factors:

kintel:
@slaviber Multiple processes would indeed make me sleep better at night :)

A few hints to make PRs like this easier to manage:
* If you perform any refactoring as part of your PR, consider opening separate PRs for green refactoring which is generally useful or harmless, as those can be reviewed and merged quickly, making your final PR less daunting
* If possible, mark your feature Experimental (see e.g. https://github.com/openscad/openscad/blob/807fe20482813c4d4bccaeef71b82970215d824c/src/feature.h#L17), to make it possible to turn on/off. This could allow us to merge before it's 100% ready and better allow others to test it.

slaviber:
I understand.  
The experimental Feature looks pretty neat.

mhier:
I can't compile this on Ubuntu 19.10 (gcc 9.2.1), I get the following linker error:
```
[100%] Linking CXX executable openscad
/usr/bin/ld: CMakeFiles/OpenSCAD.dir/src/cgalutils-applyops.cc.o: undefined reference to symbol 'shm_unlink@@GLIBC_2.2.5'
/usr/bin/ld: /lib/x86_64-linux-gnu/librt.so.1: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/OpenSCAD.dir/build.make:3193: openscad] Error 1
make[1]: *** [CMakeFiles/Makefile2:73: CMakeFiles/OpenSCAD.dir/all] Error 2
make: *** [Makefile:95: all] Error 2
```

If I add rt to the COMMON_LIBRARIES, I get a lot of undefined references for boost::archive symbols.

slaviber:
@mhier Yes, the build scripts aren't fixed properly.  

So I made some huge changes to my code.  
Fixed bugs like:  
- shared memory not properly destructed, remaining in RAM after program exit
- making the qmake (mostly) and cmake builds actually compile and link.
- fixed failing tests  

Added / Reverted functionality:
- removed meaningless (comment) blocks
- optimized the memory usage and allocation count
- switched to Qprocess (finally) so Boost.Process is not required any more.
- split the implementation in a different .cc file that contains only the multithreaded part
- reverted the cgalutils-applyops file to its original form, including only the multiprocessing calls and checks
- the multiprocessing can now be enabled/disabled as a Feature without changing any functionality in the **master**  

And I found out that the lazy-union feature makes rendering slower. Without multiprocessing it's just slower, with multiprocessing there are never enough objects in the Union to spawn a child.  

@kintel Now I can't decide which will be better - to push my changes to this branch or to create another cleaner PR.

kintel:
@slaviber Keep in mind that if you use QProcess, this would be a GUI-only feature, as the cmd-line version of OpenSCAD is built without Qt.

In terms of managing this PR, you can always pull out refactoring as a separate PR, and force-push a rebased version to this PR. If that's too much git magic, opening new PRs from scratch works too.

slaviber:
OK, I'm staying with QProcess mainly because of the lack of an alternative. And it's 5 lines of code to swap anyway.  
I'm done with the changes, the PR is now ready for a review/testing.

MichaelAtOz:
Is there a problem with ci? There is no executable in the `details` `#artifacts`

t-paul:
Works for me now - https://circleci.com/gh/openscad/openscad/4399#artifacts - maybe there was some issue or delay yesterday? It may also depend on old/new UI in CircleCI as the new one seems to require a login :-(.

MichaelAtOz:
The exe is there now....shrug.

On Windows7 F6 I get:
#3193 multi-assert

MichaelAtOz:
I tried examples/basics/logo, no error, but examples/basics/text_on_cubes, got the above assertion.
On most examples that I tried it is almost immediate, example024/023/022, but example021 the progress bar got to ~60%, example020 worked*.

* I exported to STL, STL had some inverted faces, but 2019.05 had the same, so not multithreads problem.

slaviber:
Thanks for the feedback @MichaelAtOz.  
This assertion indicates that my implementation is not behaving as I expect in some cases.  
I'll check what's happening and if it is platform dependent.

slaviber:
It was a silly mistake that has nothing to do with the multiprocessing, but with the merging of non-intersecting PolySets in a Union.  
Checked text_on_cube, example021, example022, example023, example024 - they now all render correctly.

Efroymson:
Has there been any behind the scenes update on this? I am working on some monster files that could surely benefit from faster renders. If there is any way I can help, let me know

slaviber:
Hello @Efroymson,  
You can compile my branch and test if it speeds up your work.  
Aside from this, I have no idea why my PR lost the devs' attention.

t-paul:
If it helps, I could try retriggering the CI-builds (binaries are deleted after 30 days). That would produce ready to try apps for Windows and Linux (no MacOS at this point).

Efroymson:
@t-paul That is probably a good idea, but it doesn't help me much, as my new machine is OS X only, at the moment.

@slaviber I would like to try building your branch, but the last time I tried to build OpenScad for OS X I ran into a slew of problems, and was unable to solve all of them.

Efroymson:
Just FYI, I can't seem to build this version on my system because the boost build fails.  It can't find unistd.h, which is strange.

slaviber:
You are trying the OS X building procedure from the [README.md](https://github.com/openscad/openscad/blob/master/README.md#building-for-mac-os-x) file, right?  
Which step fails? Can you show the error log?

Efroymson:
This is the tail end of what it prints out:
./boost/config/detail/posix_features.hpp:18:15: fatal error: 'unistd.h' file not found
#     include <unistd.h>
              ^~~~~~~~~~
1 error generated.
...skipped <pbin.v2/libs/regex/build/clang-darwin-9.0.1/release/threading-multi>libboost_regex.dylib for lack of <pbin.v2/libs/regex/build/clang-darwin-9.0.1/release/threading-multi>c_regex_traits.o...
...skipped <p/Users/robert/Documents/gitHub/libraries/install/lib>libboost_regex.dylib for lack of <pbin.v2/libs/regex/build/clang-darwin-9.0.1/release/threading-multi>libboost_regex.dylib...
...failed updating 94 targets...
******

There are many more such errors, complaining about unistd.h being missing.  I have a copy of it, and tried copying it into some likely place, but it doesn't work.

thehans:
Hi @slaviber, I did a bit of testing from your branch and found a few things.

First, a quick trick you can use to enable your feature throughout the entire test suite:
https://github.com/openscad/openscad/blob/40b93c0f337615e2cf0b864a3f4a75bad9e6cd82/tests/CMakeLists.txt#L267
Uncomment this line and put `multi-processing` insead of `lazy-union`.
Then re-run `cmake .` and `make` from the tests directory, and you can run ctest with feature enabled everywhere.

I ran `ctest -C All` which includes some of the longer duration tests, and some related to issues that are not yet solved and are known to fail.  Right now the master branch has **41** failing tests for me when running `All`, and I got **46** with your code.  The extra failing ones are:
```
        367 - cgalpngtest_linear_extrude-scale-zero-tests (Failed)
        501 - cgalpngtest_candleStand (Failed)
        904 - csgpngtest_linear_extrude-scale-zero-tests (Failed)
        1036 - csgpngtest_candleStand (Failed)
        1321 - cgalstlsanitytest_normal-nan (Failed)
```
Not sure how familiar you are with ctest, but you can also do `ctest -R blah` to only run tests that contain "blah" in their name if you just want to quickly try out a specific one.

For `linear_extrude-scale-zero` the test report shows an error such as:
```
WORKER 0: CGAL error: assertion violation!
Expression : itl != it->second.end()
File       : /usr/include/CGAL/Nef_3/SNC_external_structure.h
Line       : 1152
Explanation: 
Refer to the bug-reporting instructions at http://www.cgal.org/bug_report.html
```

The `candleStand` test is more like some strange logic error where the holder outer cylinders also get copied to the stand base/feet where they shouldn't be.

The `cgalsanitytest_normal-nan` looks like an error in the test python script, where its interpreting `--enable=multi-processing` as a filename for some reason (probably nothing to do with your changes, but no idea why only that test has the issue).

So that's all the errors I found with `ctest -C All`  ...  but I normally run multiple parallel jobs using ctest's `-j[NUM]` option and I noticed that a significant number of more seemingly random tests fail when I use any more than one job in parallel.

Without multiprocessing enabled I can do up to `-j32` or more on a Ryzen 8C/16T CPU and it causes no extra errors.  (Something weird about the test suite which I still don't understand: it doesn't fully load all my cores for most of the test duration at these settings of double all the available threads).

So I'm curious if the code was written with the possibility of multiple separately started processes in mind?  I haven't yet reviewed the source code changes, and don't really know the ins and outs of QProcess, but is there any way that worker processes can coordinate only with the process which spawned them?  It seems like for now maybe they are jumbling their results together when multiple concurrent "masters" are run.

@kintel wrote
> @slaviber Keep in mind that if you use QProcess, this would be a GUI-only feature, as the cmd-line version of OpenSCAD is built without Qt.

Just wanted to point out that regular builds can of course still use Qt classes even when not *displaying* a GUI.  Its only the case that builds with `CONFIG+=nogui` option in qmake or `-DHEADLESS=ON` in cmake won't even be linked with Qt.  As far as I know for our CI servers, only Travis builds actually do this currently.

MichaelAtOz:
I'm concerned with the Qt/command-line limitation. As one of the bounty source contributors, I wouldn't support non-multi-[thread/process] solution for command-line use.
Whether adding a Qt dependency to command-line use, which I understand Qt has been avoided, is not up to me. So that is a possibility perhaps?

t-paul:
The core should stay independent of the GUI part. As said above we can't simply force boost 1.64 but it should be possible to enable the feature only if boost >= 1.64 is available. That would mean people compiling themselves may need some extra steps to get the feature and also it's more work for our Snap and AppImage releases, but that would be a way forward which drops the ugly extra steps at some point.

slaviber:
Hi @thehans ,  
Thanks for testing.  

> So I'm curious if the code was written with the possibility of multiple separately started processes in mind? I haven't yet reviewed the source code changes, and don't really know the ins and outs of QProcess, but is there any way that worker processes can coordinate only with the process which spawned them? It seems like for now maybe they are jumbling their results together when multiple concurrent "masters" are run.  

That's a bug for sure. Currently, the child processes are not aware if multiple root processes are running or not. And what's worse, the main app cannot know which children is which. And based on the OS specifics, there is no implicit way of doing this. I can source the children with the PID of the root, but if a program crash occurs, shared memory won't be freed for a very long time.   
That makes me think why programs such as Chrome always have a single root, haha.  
  
I'll check if there's some reasonable way of fixing this so multiple real instances could be run together, and not just multiple "tabs", and at the same time somebody wouldn't need to restart the OS after every OpenSCAD crash (because the lack of RAM).

nophead:
I run separate instances of OpenSCAD so that they are separate. I don't want them to run in the same instance. If I did I would use tabs, or open files from the file menu of an existing instance. They share the same model but only update sequentially. Separate instances update in parallel and can use other cores.

slaviber:
Still a bit WIP and some tests are failing in a strange way but i've addressed most of the problems.  Overall performance is up approx. 13% on Debian for the whole test suite with -j1. And approx. 2.1+ times for heavy tasks like the example in the first comment. Multiple instances are working, memory is not leaking and deinitialization/crashes are handled in a proper way.

slaviber:
Okay, that's about it.  
Lastly there are 2 failing tests more than **master** with enabled multi-processing on all tests:
- cgalpngtest_linear_extrude-scale-zero-tests - which triggers #2915 because of the order in which parallelized objects are getting combined and I can't do much about it. Messing with the .scad file fixes it in some way.
- pdfexporttest_* - seems to me like another python script parsing error caused by the multi-processing argument because of the "FileNotFoundError" in the log and because of the zero execution time, but it doesn't fail on Linux and I can't debug it any further.   
  
What do you think? @t-paul @thehans

t-paul:
Nice! Could you please merge `master` again? That should hopefully bring in the updated CircleCI config and build the Windows installers. This would allow easier testing on Windows.

slaviber:
Done!

mholiv:
A quick chime in here. I had the opportunity to build and test this PR on my own projects. 

The difference was pretty stark. :D

On my open [source project](https://github.com/mholiv/Open-Nutrient-Lattice/blob/master/hardware_designs/vegetation_tray.scad) it went from:
Total rendering time: 0 hours, 1 minutes, 3 seconds
Total rendering time: 0:00:44.294

For two non-open source projects I got these:
Total rendering time: 0 hours, 4 minutes, 0 seconds
Total rendering time: 0:02:50.135

(This one was mostly looped objects)
Total rendering time: 0 hours, 4 minutes, 3 seconds
Total rendering time: 0:00:42.223

Great work here.

mholiv:
@Efroymson Here is a mac build btw. :)

[mac.zip](https://github.com/openscad/openscad/files/5389769/mac.zip)

MichaelAtOz:
> The difference was pretty stark. :D

@mholiv How many CPUs?

mholiv:
@MichaelAtOz 

This was tested on a Macbook Pro with a Intel(R) Core(TM) i5-8259U CPU @ 2.30GHz.

aka 4 cores / 8 theads. 

I could test it again on my linux 6c/12t Ryzen 1600x system later if people are interested.

MichaelAtOz:
@slaviber I did a test with n=5000 on my 8 core 32GB Debian system.
A few times it brought it to its knees, even the mouse stopped moving, and more importantly my VLC media stopped playing! The first time, in one F6, it was frozen for maybe a minute, and came back, later in same F6, maybe 15s a few times.

This is after I changed their priority to low:
Screenshot from 2020-10-21 13-55-05-multiprocess

I suspect the memory operations are in a critical section and/or it was swapping, 10*2.xGB+10.7GB.
I'm not a Linux guru, but on Debian it seems to use much more memory than Windows, is
that perhaps an Appimage side-effect?

The previous attempt kept the thread parallelism at or below the # of CPUs.
Used https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency.

Another thing would be to add a [yield()](https://www.boost.org/doc/libs/1_38_0/doc/html/thread/thread_management.html#thread.thread_management.this_thread.yield) in an appropriate loop.

For testing, and if you want to leave some CPU for other things on your system (like VLC),
an option to override the CPU count was added. See [here](https://github.com/openscad/openscad/pull/2399/commits/22eb72a02842b4500f09e014192afa445e945f6c).

On reflection, I suggest a slightly different option would be nice, `--parallelism=num` set a fixed num, or `--parallelism=-num` reduce the auto-count by num, ie -2 reserves 2 CPUs for other stuff. Alternatively an edit/preference??

slaviber:
There are no threads and no critical sections, but swapping is a possible explanation. This or the CPU can't handle the load.   
  
I tried different builds but didn't notice such a huge memory footprint difference. A couple of 100s of MB at most. The numbers from the screenshot aren't an accurate representation of the memory in use, the **free** command is more accurate.  
  
The number of CPUs currently is irrelevant because the number of workers depends on the number of children objects in an OpenSCAD operation. Limiting the number of cores would be complicated and inefficient.   

I can programmatically lower the priority of the workers, though, so foreground tasks can remain unaffected by the computations. A decent amount of RAM would still be needed for a fast operation.

Efroymson:
> @Efroymson Here is a mac build btw. :)
> 
> [mac.zip](https://github.com/openscad/openscad/files/5389769/mac.zip)

@mholiv Thank you!  I would probably be better if I were able to build a version myself, which is a hurdle I have yet to pass.  Interestingly I had a familiar problem with the binary you provided, it failed to run with the following error:
Dyld Error Message:
  Library not loaded: /usr/local/opt/harfbuzz/lib/libharfbuzz.0.dylib
  Referenced from: /private/var/folders/*/OpenSCAD.app/Contents/MacOS/OpenSCAD
  Reason: Incompatible library version: OpenSCAD requires version 20702.0.0 or later, but libharfbuzz.0.dylib provides version 20401.0.0

I am not sure at the remove of several months, but I think that is the same library that gave me pain the last time I tried to build locally.

I've been involved in some other projects, so OpenScad hasn't been on my priorty list, but when I get back to my 3d work, a faster version will be useful!

Thanks All,

Robert

mholiv:
Any updates on this? I've been using this build in "production" and was wondering if there way any update.

qcasey:
I've also been using this build successfully for a while. My intel 8700k goes from

`Total rendering time: 5 hours, 1 minutes, 23 seconds`

to

`Total rendering time: 0:22:01.743`

when multithreaded. My model is naive and also has a ton of looped objects but I think a 1300% performance uplift is respectable. It's now a coffee break instead of the day off :smile:

Captayne:
Hi, I am a user of OpenScad and have heavy calculations to do for geometry design. Some designs take hours to render.
I am using the Windows version, which only loads one CPU Core, OK. But yesterday we tried it on an apple mac book air, and there all cores were 100% busy on my scad-file. The Mac Air was pretty much as fast as my 4GHz i5 PC. 
Now I wonder why the Apple loads all cores while the PC only one. Can the Multithreading be activated somewhere? I did not find a snapshot version. Thank you very much for your answer.
Andreas

MichaelAtOz:
@Captayne What version of OpenSCAD? Only one or two development (not Snapshot) versions try paralleling.
Is it the two or four core (not thread) Macbook?

Are your designs memory hungry too? Perhaps the other core is swapping?

Efroymson:
Like the others who chimed in above, I am interested in the progress of this project.  I have started up again with some designs that definitely give the standard OpenScad build a hard time.  It would be great to be able to use more than 1/16th of my processor ...

amb-iota:
@MichaelAtOz I was reading your comment and my experience with any multitasking systems is that a default value of maximum Threads counted down by one is for most systems a good value and if the systems are using processors that do two threads in one core best speed performance is using only as much threads as cores are existing. Anything that is away from this is depending other usages of the system. 
My interest was to know why openScad did not even use more than one thread on my ubuntu 20.04 because I've made one sample that does take very long for rendering and I run on my system some script to run a miling job as long as there are not to much threads been used. I was very surprised, that while rendering my job did not stop mining until i did check with top that openScad did only use one thread and even with watching a movie while rendering the possible threads were open so that my script did even not recognise that there ist much to calculate.

After reading this all in here it looks like there is a version to compile where you get some parallelism and in my case I think it would do a good job as I've got lots of separate objects that could be build quick and separate. Else I had to wait more than 30 Min. Which version do I have to take to get the multitasking feature on my system working.

MichaelAtOz:
@amb-iota

> After reading this all in here it looks like there is a version to compile where you get some parallelism and in my case I think it would do a good job as I've got lots of separate objects that could be build quick and separate. Else I had to wait more than 30 Min. Which version do I have to take to get the multitasking feature on my system working.

This is still in development and has had limited testing, and as I noted can flood the system with jobs ATM.
Looks like the compiled executables have been deleted, development ones are not retained for long.
I'm not a GitHub developer, so someone else will need to point you to the correct source if you wish to compile it. @t-paul?

t-paul:
Updated against `master`, so we have fresh artifacts for 30 days...

MichaelAtOz:
@amb-iota if you want to run the current state of this, you can download the relevant executable from the links below, under 'all checks have passed' scroll down, click 'details' of the appimage, or for Windows the openscad-mxe-xxbit. 
You will need to register with Circleci - you can authenticate with GitHub, and because it is dumb and forgets what you wanted to do this first time, come back here to click 'details' again. 
Then select the Artifacts tab, download the relevant file.

> it looks like there is a version to compile

If you want to compile, what environment are you using?

amb-iota:
> @amb-iota if you want to run the current state of this, you can download the relevant executable from the links below, under 'all checks have passed' scroll down, click 'details' of the appimage, or for Windows the openscad-mxe-xxbit.
> You will need to register with Circleci - you can authenticate with GitHub, and because it is dumb and forgets what you wanted to do this first time, come back here to click 'details' again.
> Then select the Artifacts tab, download the relevant file.
> 
> > it looks like there is a version to compile
> 
> If you want to compile, what environment are you using?

I am using Ubuntu 20.04 with 64bit.

mholiv:
Thinking it might be worth hosting these binaries as an unofficial build somewhere. 🤔

Enough people seem to care.

MichaelAtOz:
> 
> 
> > @amb-iota ... 
> > If you want to compile, what environment are you using?
> 
> I am using Ubuntu 20.04 with 64bit.

I'm not an expert but as I understand it, I believe you want the code from [here](https://github.com/slaviber/openscad/tree/multithread).
Instruction are [here](https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/Building_on_Linux/UNIX). I'm not sure if they are up to date re the changes from qmake to cmake @t-paul?

Also @t-paul re mholiv's comment above, is it worth dropping the artifacts to the ftp site?

t-paul:
I stashed the files at http://files.openscad.org/pr/PR3193/ but we don't have the storage space to do that all the time nor does it scale if I have to do this manually. Maybe at some point we can get that automated based on a PR tag.

lf94:
@t-paul when will this be merged? :smiley:

MichaelAtOz:
@if94 This has had limited testing and can have bad effects on system stability with complex models.
It likely wont get merged without more fine tuning, & it appears nobody is actively working on it.
We can always use more developers and/or testers.

lf94:
> can have bad effects on system stability with complex models  

All I've seen as bad so far is it will eat all CPUs, but it was suggested to use n - 1 CPUs by default.

What else is there, so we can start a list of things to do / look out for?

MichaelAtOz:
> All I've seen as bad so far

It crippled my system, a Debian Skull-canyon with high-end NVMe SSDs, I'm surprised it didn't go fatal, but it was a worrying for a few minutes. Unix is meant to be resilient but something really clobbered it.

> @slaviber I did a test with n=5000 on my 8 core 32GB Debian system.
> A few times it brought it to its knees, even the mouse stopped moving, and more importantly my VLC media stopped playing! The first time, in one F6, it was frozen for maybe a minute, and came back, later in same F6, maybe 15s a few times.

`n=5000` is not a big test in my stress test, the multi-threaded version was civil even with `n=100000`, I had tests running for hours. This one spat out in very quick succession huge numbers of processes, and compared to the overhead of a thread, the overhead of a full OpenSCAD process is large, particularly memory. [this may be related to the use of Appimage perhaps?? I'm no expert]

Some people seem to see it as usable for less challenging designs.

Not being a Linux dev, I imagine setting up a worker controller to doll-out & manage jobs is not trivial (unless there is some library?). Unfortunately with a crippled system it is difficult for me to see what was actually happening, it needs a Linux guy who get into weeds.

> What else is there, so we can start a list of things to do

- [x] Start a list
- [ ] Find a developer
- [ ] Test/debug what caused 'freeze' with complex design
- [ ] Design approach to avoid it (presuming CPU count limit ATM, but it could be memory or something else)
- [ ] Preference and command-line option for CPU# v's process spawning limits
- [ ] Develop Worker management mechanism or process spawning limiting code
- [ ] Develop test cases
- [ ] Test
- [ ] Stress test
- [ ] Etc

Start with the second point. Sorry to be snarky but I thought I spelled it out with my first reply.

lf94:
I am a developer, so that's why I've asked. :laughing: 

Thank you very much @MichaelAtOz for taking the time to describe what needs to be done. It doesn't seem like much more works is needed - just some investigation, adding some limit options, and job batching.

MichaelAtOz:
> Edit: Do you know where the complex design is which is causing the issue?

Lucky I came here to see the edit.

[The latest here is this V6](https://github.com/openscad/openscad/issues/3465#issuecomment-711492149), I have V8 on the Debian system, (car engine term is purely coincidental), I'll need to compare them to se if the new one is relevant, that was months ago in a pandemic...

You should also read #3465 on what reality is in relation to 3D output, particularly the merits of identical output.
I must admit I lost interest with unexplained different STLs for the same .scad.
If a different tessellator is the cause, is it inherently part of multi-processing or just incidental.

These theological discussion should probably occur in #3465.

lf94:
@MichaelAtOz I've compiled and run the example which is freezing. I want to be explicit to be sure we are seeing the same problem: When rendering with F6, it becomes stuck at 999/1000, correct?

MichaelAtOz:
> I want to be explicit to be sure we are seeing the same problem: When rendering with F6, it becomes stuck at 999/1000, correct?

Not my 'freeze' no. 
Excuse the mansplain, did you enable multi-processing in Preferences/Features? [but see Hmm below]

What you see at 999/1000 is the final global union, unavoidable* and not readily parallelable, tho here-somewhere is/was(?) discussion of strategies to optimise it, sorting by facet count etc, I can't recall if that was in this, or another activity. Perhaps @someone-else can recall. 

*you may want to see what preferences/features/lazy-union does to that.

My 'freeze', 

> it brought it to its knees, even the mouse stopped moving, and more importantly my VLC media stopped playing!
> The first time, in one F6, it was frozen for maybe a minute, and came back, later in same F6, maybe 15s a few times.

Frozen, it just started to get bad, progressively worse responsiveness then quickly locking up, looking dead, then 'shit I don't really want to hold down the power button'. In the one F6, freeze, coming back to a slow jittery GUI, locking up again for 15s etc repeat, then coming slowly out of it, where I was able to run system monitor and lower the priority (see the image [above](https://github.com/openscad/openscad/pull/3193#issuecomment-713290338) with 26GB of Appimage OpenSCAD processes) . I expect if I was able to see it earlier there may have been a lot more spawning processes choking the kernal. In the middle I never had enough control to cancel the render.

I was about to re-run that a couple of days ago, with System monitor already running, the initial GUI OpenSCAD at lower priority, so the spawned process will inherit it. But some IRL things hit. I'll try do that today.

Hmm, I just tried a smaller test, but I'm not seeing and extra processes, something funny going on. 
Even n=8000, I only see one --spawnchild process running at 12% (100% of one core). Looking into it.

> I also read the thread about the topology changing. I have no real opinion on this right now. My immediate reaction is: I can't imagine this is a big deal.

I was also about to add my 2c to that discussion, will do so.

MichaelAtOz:
Well, of course as soon as you post, it changes behaviour.

I got the freeze again, but I had left it pre-freeze to get caffeine.
Locked solid, then partial recover etc, and yes 999/1000. 
Other funny things happening, including 32GB RAM and 32GB swap consumed, oooodels of 2.x GB processes.
It is not CPU as the spawnchild's were LOW, so it has to be kernal and/or swap* or memory copy in a critical section.
*while I was typing that, i t   w e  n  t   i nto an oth er cris is, but memory was not 32+32, it was 14+21.

Easy to tell when it's about to happen if you have media playing, when it hits, it goes like a bad CD, then freezes.

It's still 999/1000, process count is slowly reducing. 
I wonder if this bit is the final union optimisation?? 
If it is spawning a process for every pair of n=8000 (yes that was unwise), that could explain the final hit.

I'll need to do it again an be present when it first starts.

Can anyone recommend a better process monitor than the default one?

MichaelAtOz:
I'm continuing testing related stuff in #3465.

amb-iota:
Hi Michael, your freezing remembers me on a tiny thing I did program in the past. I was trying to calculate a game. 
But what I found out is, parallel working is only interesting as long as you do not share a problem in more pieces as processors are available reduced about one thread as else you get stuck because the machine itself isn't freeing there own processes anymore and that means you cause a problem that multiplies itself and does not only depend on the job itself. 

Don't know if your parallel processing is limiting itself to that kind of value, because if not you allways will run out of normal range not only with processors because your Memory will than get filled compleetly over time. 

maybe that's causing your freezing because it sounds a bit like it.

MichaelAtOz:
@Efroymson I don't believe this branch had a Mac build.
The automatic builds for branches only have Windows 64 & 32 bit & Linux Appimage builds.
This branch is not in master hence is not in the snapshot developer nightly builds.
It has deficiencies and needs some love from a Developer.

src/cgalutils-applyops.cc
```diff
@@ -85,6 +87,15 @@ namespace CGALUtils {
 		bool foundFirst = false;
 
 		try {
+			// DIFFERENCE is a strictly ordered operation and cannot be multithreaded ?! :(
```
dgatwood, line 90: Unless I'm missing something, most of the computation for difference() need not be strictly ordered.  It is *almost* embarrassingly parallel, with the exception of the first item.

```
difference() {
  a;
  b;
  c;
  d;
  e;
  f;
}
```

 is equivalent to

```
difference() {
  a;
  union() {
    b;
    c;
    d;
    e;
    f;
  };
};
```

In fact, in an ideal world,  the work operations would treat it as though it looked like this:

```
difference() {
  a;
  union() {
    union() {
      union() {a; b;}
      union() {c; d;}
    }
    f;
  }
}
```

So you do the unions in a massively parallel way, and when both child nodes are completed, you do the parent union operation, until you get up to the final node, and then you complete the difference operation.  And as long as you design the work unit system correctly, you can throw the leaf nodes (a, union{bc}, union{de}, and f) onto separate worker threads, and everything should "just work".  When each work unit completes, it checks to see if the parent node's children are all complete, and if so, you dispatch that node to a worker thread.  And so on.

Am I missing something?  (I haven't dug into this code much yet; I was just looking for a way to shave several hours off a render that takes a fraction of a second for the GPU-quality rendering, and found this pull request, and started looking at it, and saw this comment, and it struck me as confusing.)

slaviber: Hey, thank you for your interest in my PR.

It has been quite a long ago, 5 years actually, and I lost track of what has been done to this project since then.

If I remember correctly, I made numerous tests and found out that the Diff operation, as it was coded back then, resulted in wildly different output based on the order of combining the solids. The GCAL library was buggy in this regard.

It may or may not have been the case, but since the dispute on what output was considered valid, I focused on more productive activities and got my bachelor in Engineering instead. 

If your observation is true and the Diff operation can be subdivided to a Union and a Difference, and this does not cause a further operation bottleneck, and, most of all, does not alter the final render, then great. 

Maybe you can bring this code back from the dead and finally fix the paralellism problem in OpenSCAD? 
I gave up and patched another open source sw for my final thesis.

I have to mention that the biggest bottleneck was the CGAL library itself not being thread-safe, which required very computationally-heavy serialization. And lots of RAM usage. Things may have changed for better.

jordanbrown0:
OpenSCAD has almost completely moved from CGAL to Manifold, which is parallel and far faster.

src/cgalutils-applyops.cc
dgatwood: From what I see, CGAL is currently thread-safe as long as you aren't sharing objects across threads, hence why I suggested work units that combine results from previous stages; that way, you aren't ever sharing data across threads, presumably, and you'll also get deterministically consistent results every run without the need for strictly ordering anything beyond ensuring that [work item a] doesn't start until after [work item a's nested dependencies] are finished.

It's interesting that you saw different results depending on order.  I could see there being subtle differences, e.g. different triangulation of a face or something, but I'd expect the results to be congruent to the originals, just assembled differently.  If that's not the case, that's... interesting.  😁

dgatwood:
> OpenSCAD has almost completely moved from CGAL to Manifold, which is parallel and far faster.

Holy crap.  That reduced my render time from... well, giving up after half a day to being done in just three *seconds*.  Why isn't this on by default?  😂

t-paul:
Closing, it's long stalled and superseded by adding of the Manifold back-end.

jordanbrown0:
As for the "why not by default", see #5833, which currently says to see #5485 and #5866.

Issue #3186, opened 2020-01-11
Editor: Cut to end of line (ctrl-k) doesn't copy to clipboard


kintel:
Usually, operations like cut should preserve the cut text in the clipboard.
Cut to end of line (Ctrl-K) doesn't do this.

PR #3205, opened 2020-02-02
Cut to end of the line

tbharathchandra:
#3186 is solved in this branch. Cltr+K will now cut the text after the cursor to the end of the line and place it in the clipboard.

### DIFF
--- a/src/scintillaeditor.cpp
+++ b/src/scintillaeditor.cpp
@@ -1208,3 +1208,12 @@ void ScintillaEditor::jumpToNextError()
 {
 	findMarker(1, 0, [this](int line){ return qsci->markerFindNext(line, 1 << errMarkerNumber); });
 }
+
+QString ScintillaEditor::stringToEndOfTheLine(){
+    int line,index,line_len,position_start;
+    qsci->getCursorPosition(&line,&index);
+    position_start = qsci->positionFromLineIndex(line,0);
+    line_len = qsci->lineLength(line);
+    QString textToEnd = qsci->text((position_start+index),(position_start+line_len));
+    return textToEnd;
+}
--- a/src/scintillaeditor.cpp
+++ b/src/scintillaeditor.cpp
@@ -1209,11 +1209,12 @@ void ScintillaEditor::jumpToNextError()
 	findMarker(1, 0, [this](int line){ return qsci->markerFindNext(line, 1 << errMarkerNumber); });
 }
 
-QString ScintillaEditor::stringToEndOfTheLine(){
-    int line,index,line_len,position_start;
+void ScintillaEditor::stringToEndOfTheLine(){
+    int line,index,line_len;
     qsci->getCursorPosition(&line,&index);
-    position_start = qsci->positionFromLineIndex(line,0);
+    //position_start = qsci->positionFromLineIndex(line,0);
     line_len = qsci->lineLength(line);
-    QString textToEnd = qsci->text((position_start+index),(position_start+line_len));
-    return textToEnd;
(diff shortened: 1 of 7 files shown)


kintel:
There is already support for Ctrl-K in QScintilla. Did you look into adjusting that instead of writing a another global keybinding in MainWindow?

tbharathchandra:
I am unaware about that and thanks for mentioning it. I will rewrite the code using QScintilla:QSciCommand class.

tbharathchandra:
Is there any problem with the appveyor check, It is being in pending state from the last two days. It also shows "Build success" in log message.

t-paul:
Not sure, sometimes the notification of the finished build seems to get lost. I triggered the build again, but the first one was successful.

src/scintillaeditor.cpp
```diff
@@ -1208,3 +1208,13 @@ void ScintillaEditor::jumpToNextError()
 {
 	findMarker(1, 0, [this](int line){ return qsci->markerFindNext(line, 1 << errMarkerNumber); });
 }
+
+void ScintillaEditor::stringToEndOfTheLine(){
+    if(qsci->hasFocus()){
+        int line,index,line_len;
+        qsci->getCursorPosition(&line,&index);
+        line_len = qsci->lineLength(line);
```
t-paul: This calculation seems to be wrong in case of characters that are more than one byte, e.g. with cursor after the `>` and then `CTRL+K` gives:

grafik

grafik

tbharathchandra: Yeah, that lineLength method has got some problem while dealing with characters having size of more than one byte. But the problem is there is no other method available that give us the exact character count. I have also tried using positionFromLineIndex method which return the byte index in the whole text. But this will not help us because the setSelection method needs two pairs of line and relative position as input.

t-paul: Right, there's a hint in the docu of [`lineLength()`](https://www.riverbankcomputing.com/static/Docs/QScintilla/classQsciScintilla.html#ab5ecfa6bee9e78e5306986367e1194af): _In order to get the length in characters use text(line).length()._ 

I guess it should work factoring [this calculation](https://github.com/openscad/openscad/blob/5696d0703583d3d53b34683bbc59d3b1cb284a78/src/scintillaeditor.cpp#L1210) out into a separate function and use that. This makes sure to ignore the trailing newline if it exists.

t-paul:
Hmm, I found another example which is cutting too much: `te🙃st`. I'm not yet sure why that's not covered.

t-paul:
Seems just using `text()` does not work in all cases as the documentation implies. The upside-down smiley is `0x1f643` and scintilla stores that internally as 2 bytes breaking the calculations again.

Best idea I have so far is to force conversion to a 32bit string which should collapse those sequences into a single character.
```
qsci->text(line).remove(QRegExp("[\n\r]$")).toStdU32String().length();
```
That really gets convoluted, so I'm open for better ideas :upside_down_face: .

tbharathchandra:
Yeah converting string into 32-bit format will definitely gonna us with this and mostly every character that the user can think off will be in 32-bit unicode.

src/scintillaeditor.cpp
```diff
@@ -1220,5 +1220,5 @@ void ScintillaEditor::stringToEndOfTheLine(){
 }
 
 int ScintillaEditor::getCharCount(int line){
-    return qsci->text(line).remove(QRegExp("[\n\r]$")).length();
+    return qsci->text(line).remove(QRegExp("[\n\r]$")).toStdU32String().length();
```
tbharathchandra: converted to 32-bit format

Issue #3155, opened 2019-12-10
Less intrusive autocomplete


thehans:
The autocomplete editor feature is nice, but I have been getting some false-positive matches recently, which I'm finding tedious to avoid.

For example I needed a function that returns mathematical modulo.  (like `%` operator but always return non-negative values).  So I wrote a function called `mod(x,y)`, but then whenever I type to call the function: `mod(` it automcompletes to `module(`
Also, trying to call a function named `f()` autocompletes to `false(`, or a function named `rot(...`, autocompletes to `rotate(`, etc.

Tab or Enter keys (and maybe others?) already are ways to accept the autocomplete, so I'd rather not have `(` trigger it too.

`module(` isn't even correct syntax anyways (without a module name).

Secondly, I think autocomplete should be disabled when typing inside comments.

MichaelAtOz:
Also, I think the pop-up dialogue should be moved, say, ~2 lines away from the cursor and/or to the right. Frequently you want to see the adjacent lines. Particularly a problem at the bottom of the page:
![Capture autocomplete](https://user-images.githubusercontent.com/1972961/70582205-cac0b180-1c0d-11ea-8a0b-cdb3f1687617.JPG)

t-paul:
I agree that it would be a useful change, but the implementation of those features is inside QScintilla, so the control on our side is limited.
I found that setting the trigger to 2 characters helps a lot with the misfires (but would not help for `mod`). Maybe there's at least some control over the keyboard handling as I think the most problematic part is the auto-accept of the selected value when just typing.

Sharma-Hrishabh:
Hey @t-paul, I want to work on this issue. Can you guide me, from where should I begin?

rcolyer:
Hi @Sharma-Hrishabh, if memory serves these features came in with the merge of PR #2975.  That should point you to the relevant subset of the source code.

t-paul:
Fixed via  #3207.

diabl0w:
how do I enable auto-complete? I cannot find documentation on this anywhere and it is not in preferences from what I can see.

Sharma-Hrishabh:
hey @diabl0w , the option to enable/disable autocomplete is present in Preferences->Editor. [ at the bottom ]

diabl0w:
> hey @diabl0w , the option to enable/disable autocomplete is present in Preferences->Editor. [ at the bottom ]

[...]

PR #3207, opened 2020-02-03
fix intrusive autocomplete

Sharma-Hrishabh:
Fixed issue #3155 

When you don't select an autocomplete suggestion and press '(', then the first suggestion is selected for auto-completion.
This PR fixes this intrusive behaviour.

### DIFF
--- a/src/scintillaeditor.cpp
+++ b/src/scintillaeditor.cpp
@@ -273,7 +273,8 @@ void ScintillaEditor::applySettings()
 	if(Preferences::inst()->getValue("editor/enableAutocomplete").toBool())
 	{
 		qsci->setAutoCompletionSource(QsciScintilla::AcsAPIs);
-		qsci->setAutoCompletionFillupsEnabled(true);
+		qsci->setAutoCompletionFillupsEnabled(false);
+ 		qsci->setAutoCompletionFillups("(");		
 		qsci->setCallTipsVisible(10);
 		qsci->setCallTipsStyle(QsciScintilla::CallTipsContext);
 	}


src/scintillaeditor.cpp
```diff
@@ -273,9 +273,11 @@ void ScintillaEditor::applySettings()
 	if(Preferences::inst()->getValue("editor/enableAutocomplete").toBool())
 	{
 		qsci->setAutoCompletionSource(QsciScintilla::AcsAPIs);
-		qsci->setAutoCompletionFillupsEnabled(true);
+		qsci->setAutoCompletionFillupsEnabled(false);
+ 		qsci->setAutoCompletionFillups("(");
```
thehans: Does this line matter if fillups are not enabled?

src/scintillaeditor.cpp
```diff
@@ -286,6 +288,7 @@ void ScintillaEditor::applySettings()
 
 	int val = Preferences::inst()->getValue("editor/characterThreshold").toInt();
 	qsci->setAutoCompletionThreshold(val <= 0 ? 1 : val);
+	// qsci->setAutoCompletionReplaceWord(false);
```
thehans: Maybe you were testing something options here, but I would recommend not committing new unused code like this in your Pull request.

thehans:
I tested this branch and it does seem to perform as expected now.  I think that's a great improvement over previous behavior!

I did add some small comments to your PR code, and I have just one other question: why are the lexer and parser files touched in your commit but showing no changes?

src/scintillaeditor.cpp
Sharma-Hrishabh: Sorry for that

Sharma-Hrishabh:
I didn't make any changes in those files. Maybe, the changes were made by a Linux system like mod changes. That's why no change is being shown here.

src/scintillaeditor.cpp
Sharma-Hrishabh: I don't think, it will matter when fillups are enabled.

src/scintillaeditor.cpp
Sharma-Hrishabh: I have removed that comment in a new commit.

kintel:
@Sharma-Hrishabh Could you remove all the zero changes and the .vscode folder?
PS: It might be easier to just reset this branch and re-commit your 2 lines of code. You can force-push to the same branch in order to keep this PR open.

Sharma-Hrishabh:
Hey @kintel , I have cleaned the changes and removed the .vscode folder. Kindly take a look.

kintel:
Thx, looks great

PR #3208, opened 2020-02-04
Cut to end of the line 

tbharathchandra:
This PR is created after the review of @kintel on #3205 PR. #3186 is solved in this branch. This branch uses QScintilla for implementing Ctrl+K feature to the editor.


t-paul:
@kintel we may need to go back to #3205. I'm not sure there's an existing command doing cut-end-of-line to clipboard.

kintel:
Well, we tried. @tbharathchandra if you resurreect #3205, we should probably add the shortcut so it's only active if the editor has focus.

tbharathchandra:
As @kintel suggested I reopened the PR #3205 and added a condition which make sure that this shortcut work only when the editor is in focus.

t-paul:
Yep, so lets close this one and we go back to #3205.

Issue #2670, opened 2019-01-01
Editor - Line Wrap Preference - Indent?


MichaelAtOz:
The pref;
View/Preferences/Editor/Line-Wrap/Line-wrap-indentation/Style:Indented is either wrong or non-intuitive.
It uses the Indentation-width from the Indentation box above, not the Indent value to the right, ie;
![linewrap indent](https://user-images.githubusercontent.com/1972961/50576028-64f47300-0e5d-11e9-8241-85a19683d039.JPG)

This may be intentional, forcing the same indent, and thus the adjacent Indent value is only meaningful for Fixed indent, if so the adjacent field should be greyed or blanked for other than Fixed. 
However, I suspect it should use the adjacent Indent value.

Sharma-Hrishabh:
Hey, this issue looks interesting to me, and I would get an overview of the GUI portion code of openscad. Should I go ahead? 
It would be great if I can get some hints.

t-paul:
Sure, that would be great. I guess the first step would be to find out what's going on by checking the current GUI against the features documented for [QScintilla](https://www.riverbankcomputing.com/static/Docs/QScintilla/classQsciScintilla.html). There's a huge amount of settings so it could be a simple bug in the code (referencing the wrong input box to read from), a misunderstanding of the features provided (maybe there's not even a separate setting), or something else completely.

Sharma-Hrishabh:
Hey @t-paul , what I observed after exploring the editor and preferences file is that all things are correct there. I read the documentation and observed 'WrapIndentMode' enum, which is being used in our editor to control that LineWrap Indent functionality.
It is written that-
-'Same' will not take any value from Indent spin box in LineWrapIndent, and will indent as per the first sub-line.
-'Fixed' takes value from Indent spin box in LineWrapIndent and put indentation accordingly.
-'Indent' mode is something different, it indents as per first sub-line plus an extra indentation.

[...]

PR #3212, opened 2020-02-09
disable indent-spinbox in IndentLineWrap style

Sharma-Hrishabh:
This commit disables the 'Indent' spin-box when 'Indented' option is chosen from 'Line Wrap Indentation Style'.
The purpose of doing this is explained in issue #2670  (for future contributors)

### DIFF
--- a/src/Preferences.cc
+++ b/src/Preferences.cc
@@ -555,8 +555,9 @@ void Preferences::on_comboBoxLineWrap_activated(int val)
 
 void Preferences::on_comboBoxLineWrapIndentationStyle_activated(int val)
 {
-
+	//Next Line disables the Indent Spin-Box when 'Same' or 'Indented' is choosen from LineWrapIndentationStyle Combo-Box.
 	spinBoxLineWrapIndentationIndent->setDisabled(comboBoxLineWrapIndentationStyle->currentData() == "Same" || comboBoxLineWrapIndentationStyle->currentData() == "Indented");
+	
 	applyComboBox(comboBoxLineWrapIndentationStyle, val, Settings::Settings::lineWrapIndentationStyle);
 }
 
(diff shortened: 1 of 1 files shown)


Sharma-Hrishabh:
Yaa, you were right @t-paul , that line is not required at both the places. I tested, it out and concluding that it is required only inside the function. It's not working properly outside that function. 

I'm looking over that comparing things.

Sharma-Hrishabh:
Hey @t-paul , I think we can also use the current index of combo-box for comparison, and that would have no relation with the language.

Sharma-Hrishabh:
Hey @t-paul , after checking the documentation and surfing the StackOverflow and forums, I get to know that ->currentText() gives you the second value(i.e displayed on UI) and ->currentData() gives you the first value(i.e untranslated value) on a combo-box
S.o, I changed that to ->currentData() instead.
I have made proper checks by changing the language on my Ubuntu 19.04 System. All is working fine.

t-paul:
Sounds good, Qt documentation says _This property was introduced in Qt 5.2._ and we have minimum version 5.4. now.

Sharma-Hrishabh:
Cool, this is a helpful addition. So, I think this change is serving our purpose.

t-paul:
This works now nicely even when setting non-english language. But I think I found the reason for the other place that was checking the condition. When selecting a value that has the spinbox disabled and restarting the application, the spinbox is not disabled anymore. Changing the combobox fixes it again.

Sharma-Hrishabh:
Ooh, yes you are right @t-paul, I was just restarting the application each time after making, so the default value was fixed. So I didn't observe this. Thanks, you observed it nicely.
It would be nice if we leave comments to explain the existence of this line. Is it allowed @t-paul ?

t-paul:
Yes, adding this comment now is a very good idea so we don't fall into that trap again in case we need to modify this code sometime later.

Sharma-Hrishabh:
Ok @t-paul , I have added proper comments. This will help future contributors.

PR #3222, opened 2020-02-16
Shortcuts for autocomplete

t-paul:
Add shortcuts for:

- Autocomplete: CTRL + Space (pretty much the default for IDEs)
- Calltips: SHIFT + CTRL + Space (function parameters)

### DIFF
--- a/src/scintillaeditor.cpp
+++ b/src/scintillaeditor.cpp
@@ -17,7 +17,7 @@
 
 namespace fs=boost::filesystem;
 
-QString ScintillaEditor::cursorPlaceHolder = "^~^";
+const QString ScintillaEditor::cursorPlaceHolder = "^~^";
 
 class SettingsConverter {
 public:
--- a/src/scintillaeditor.cpp
+++ b/src/scintillaeditor.cpp
@@ -7,6 +7,7 @@
 
 #include <QString>
 #include <QChar>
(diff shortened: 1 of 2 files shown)


src/scintillaeditor.cpp
```diff
@@ -152,6 +156,14 @@ ScintillaEditor::ScintillaEditor(QWidget *parent) : EditorInterface(parent)
 	c = qsci->standardCommands()->boundTo(Qt::Key_Insert | Qt::CTRL);
 	c->setAlternateKey(0);
 
+	QShortcut *shortcutCalltip;
+	shortcutCalltip = new QShortcut(Qt::CTRL | Qt::SHIFT | Qt::Key_Space, this);
+	connect(shortcutCalltip, &QShortcut::activated, [=]() { qsci->callTip(); });
+
+	QShortcut *shortcutAutocomplete;
+	shortcutAutocomplete = new QShortcut(Qt::CTRL | Qt::Key_Space, this);
```
kintel: These two should probably be `Qt::META` on macOS, to correctly bind to Ctrl, as Qt:CTRL binds to the Cmd key, which is not what we want to use (since Cmd-space is Spotlight/Launchbar etc. search)

t-paul: Updated.

Issue #2977, opened 2019-06-26
Vimlike Mappings or User Defined Mappings


iasj:
Hi there. Great software.

It would be wonderful to be able to control the 3D view with some user defined keys. 

I know it is possible to control zoom with `ctrl+[` and `ctrl+]`, but I'd to have more control in order to not use the mouse at all.

As a Vim user for years, I'd like to control for example the draging of the view with the Vim keys as in

```text
       k

       ^
       |
h  <--- --->  l
       |
       v
   
       j
```

This would improve my workflow a lot once I'm always resting my right hand on top of those keys.

I'd also like for this feature to be user defined if possible. I'd like to put zoom action to keys like `i` and `o`.

In fact, I think it is a waste of fingers to use `ctrl+[number]` to change views. Why is it not as it is in Blender, where you can change view directly with number keys. It would make more sense.

Is this already possible in the 2019 version? If so, how?

MichaelAtOz:
Is this with the editor window hidden, as it has focus and gets plain keycodes? Or are you proposing a command mode toggle?

There is currently no keyboard remapping built-in.

I agree, some form of keyboard control of the preview window is needed, both translation and rotation on top of the existing zoom. Particularly helpful for [accessibility](https://en.wikipedia.org/wiki/Computer_accessibility).
The Ctrl+n is pretty baked-in, but if there was some toggle, maybe 'n' just could be used.

For info [this list](https://github.com/openscad/openscad/wiki/Keyboard-Shortcuts), is probably a little out of date, but it looks like k/h/l/j are almost unused, i/o not so much. See also the links, in particular to Qt, which has 4 OS flavours which need to be corralled. Choosing key combos is like herding cats on ice [the frozen water variety, I imagine the other would be difficult too..].

iasj:
I think a command mode toggle, such as Vim's, would be wonderful, but once I'm not aware about the guts of Openscad's code development, I'm afraid it would be a pain to implement such feature.

Instead, I'm proposing the idea of leaving all it's current default keyboard mappings untouched - so it doesn't change everyone's lives - and implement some sort of API of it's keyboard mappings for advanced users to tweak.

For example, we could have a hidden file in Home folder (.openscadrc) where we put the new mappings and possibly other settings, so every time openscad opens, it reads the file and sets itself based on it. 

And yes, without the command mode toggle, this would only make sense with the editor being hidden.

DLFW:
[...]

PR #3226, opened 2020-02-17
add user-defined key shortcuts

Sharma-Hrishabh:
Through this enhancement, the user can add custom keyboard shortcuts by editing the shortcuts.json file directly


Sharma-Hrishabh:
Hey @t-paul , I have added the support to customize the shortcuts through a JSON file for all the actions. Still, some shortcuts defined though other ways, are not in my knowledge, so that would be a challenge as of now.

src/mainwin.cc
```diff
@@ -567,6 +576,140 @@ MainWindow::MainWindow(const QStringList &filenames)
...
+	//Define every shortcut
+
+	std::string absolutePath = PlatformUtils::resourceBasePath()+"/shortcuts/shortcuts.json";
+	QString finalPath = QString::fromLocal8Bit(absolutePath.c_str());
+
+	//reading json file to get shortcuts
+	QFile jsonFile(finalPath);
+	if (!jsonFile.open(QIODevice::ReadOnly | QIODevice::Text))
+	{
+	qDebug() << "Unable to open file, exiting...";
```
t-paul: I think we did not use `qDebug()` so far. There's a separate debug logic that can be enabled via `--debug=<c++-filename>`. See `PRINTD("text");` and `PRINTDB("string %s, %s", param1 % param2);` macros.

Sharma-Hrishabh:
Cool, @t-paul . I'm confused like, you are suggesting to make a class and use it for modifying default menu shortcuts. Is it so?
But, if we don't use that JSON file, it won't be that easy for users to change shortcuts again.

t-paul:
Yes, separating the code into a different small class. The `mainwin.cc` is already huge so having code a bit split up helps with maintenance.
We will still use the JSON file, but only for changes the default, but not for the default. So basically the user can add that file to overwrite/define/remove shortcuts. Removing the file would revert back to the default application setup.

src/mainwin.cc
Sharma-Hrishabh: I removed this, as it was only needed for debugging.

Sharma-Hrishabh:
I have added, the new changes. Now, shorcuts.json file is completely optional and the user can define new shortcuts in the JSON file, and other actions will continue having their default shortcuts.

DLFW:
Hi! I built and tested this feature (commit 75de544). Three remarks:

a) Will the location of the config file changed? `PlatformUtils::resourceBasePath()+"/shortcuts/shortcuts.json"` doesn’t make sense to me. I would expect `PlatformUtils::userConfigPath()+"/shortcuts.json"`.

b) It seems that single characters don’t work as shortcut.

c) From what I can see, the available actions don’t cover turning the 3D view angle. “Bottom”, “Left”, etc. just jump to a base view but don’t allow freely rotating the view.

Regarding #2977, b) is “not nice” and not what people think about when talking about “vim-like”. But c) is a severe problem. If I don’t have missed an action in the list, this change does simply not solve what #2977 is asking for.

Sharma-Hrishabh:
I agree with you @DLFW. The whole discussion on the issue involves many views and enhancements.
Implementing the custom shortcut file is one of them, but I think, it is a really important feature as most of the IDEs/Text Editors have that feature nowadays.
Still, a single button shortcut feature is also a challenging thing, like I'm not finding any clue from the documentation, because when we add a single key shortcut, qscintilla takes over the control.
I will continue on looking at that.

DLFW:
Hi @Sharma-Hrishabh!
I just realized two things:
 • Even combinations like ‹CTRL›+c cannot be used since they are consumed by qscintilla.
 • Such combinations, and even single characters work fine if qscintilla does not have the focus.

Sadly, the editor seems to get the focus always on startup and – pretty funny – even when the editor is hidden. If this could be changed somehow, it would solve at least my point b). Maybe OpenSCAD could just put the focus on the console widget when the editor is hidden.

Any idea if freely rotating the 3d view will be possible? (My point 3) from above.) I see that this implementation uses "Actions" from some "tool bars", so it might be limited to stuff for which a button exists. Are mouse-drag operations in the 3d-view also realized with such actions? (I have no clue about Qt, as you may recognize. :) )

Thanks for your efforts!
Daniel

Sharma-Hrishabh:
Yes, this implementation is using the existing actions list and can define new shortcuts for them only. So, at this stage even I don't know if it is possible to implement shortcuts for mouse drags, as even I'm a new contributor. But, as I'm here for learning, so will definitely try to implement this feature :)
Hey @t-paul , do you agree that a separate issue would be better for this- keyboards shortcuts for mouse-drags?, as its implementation will be in some other part of the codebase, I assume so.

DLFW:
I think I figured out how rotating the view can be done via actions which can then be remapped by your shortkey solution. If I find the time in the next days, maybe I can provide a solution.

iasj:
Oh! Good to see my question bred a lot of progress on the matter. Thank you.

About that focus problem when both 3D view and editor are coexisting. Well, I insist on leaving all the current keyboard shortcut untouched for the normal user, but create fun options for the advanced user.

For example, a command line argument such as `--3d-view-only` would probably avoid the need for checking mouse hover positions. Again, not using the mouse at all would be the unicorn land for me.

Another question I have is: it would be terrible to implement it with YAML instead of JSON? But I'm guessing now it's too late to ask that right?

t-paul:
@Sharma-Hrishabh Yes, I agree that this PR should focus on a single topic - keyboard shortcuts based on reading the user defined config file. We can handle other ideas and configs via separate issue/PR later.

@iasj at this point we will stay with JSON as that is used for other files too. Having a mix of different formats is not useful and we'd like to have a GUI to maintain shortcuts at some point anyway.

src/shortcuts.cc
```diff
@@ -0,0 +1,53 @@
...
+	QByteArray jsonData = jsonFile.readAll();
+	QJsonDocument doc = QJsonDocument::fromJson(jsonData);
+	QJsonObject jObject = doc.object();
+	QVariantMap json_map = jObject.toVariantMap();
+
+    for (auto &action : actions) 
+    {
+        const QString shortCut(action->shortcut().toString(QKeySequence::NativeText));
+        //handle the case, if shortcut is not defined for an action
+        if (shortCut.isEmpty())
```
t-paul: That's checking the default shortcut of the action, correct? Why prevent setting a shortcut for actions that have none by default? With the configuration in place we may end up removing a couple of more obscure shortcuts and leave it to the user to define something that fits their needs.

src/shortcuts.cc
```diff
@@ -0,0 +1,53 @@
...
+        {
+        continue;
+        }
+    
+        QString actionName = action->objectName();
+        QString shortcut =  json_map[actionName].toString();
```
t-paul: Maybe add a `trimmed()` here, that would make sure to catch also `"  "` with `.isEmpty()` later.

src/shortcuts.cc
```diff
@@ -0,0 +1,53 @@
...
+        if(!shortcut.isEmpty()) {
```
t-paul: In case the value given in the file is empty, we could remove the default assigned shortcut. That way the shortcut can be reused without the need to assign a different one to the old action.

src/shortcuts.cc
```diff
@@ -0,0 +1,53 @@
...
+        const char *customShortcut=shortcut.toStdString().c_str();
+        action->setShortcut(q_(customShortcut, nullptr));
```
t-paul: I think we should skip the translation here to make sure the users always gets what they setup regardless of translation data. I believe just `action->setShortcut(QKeySequence(shortcut))` should be enough.

Sharma-Hrishabh: Ok, @t-paul . I have now skipped the translation part in the last commit.

Sharma-Hrishabh:
So, till now, this PR provides a user-defined keyboard shortcut customization through a JSON file feature, which a user has to make at UserConfigPath.
Another small change I made, is to setFocus at ConsoleWidget whenever the user hides the editor.

src/shortcuts.cc
Sharma-Hrishabh: I thought this would create a bad experience for the user. If he tries to add more shortcuts, as there will be conflicts of shortcuts and the number of shortcuts keys available are limited.
So, if you want we can change this. What would you suggest?

src/shortcuts.cc
Sharma-Hrishabh: Yups, I missed that, fixed now.

src/shortcuts.cc
Sharma-Hrishabh: Ok, so I think we need to check if the key exists in the JSON or not, rather than checking if the value is empty or not.
I followed this approach.

src/shortcuts.cc
Sharma-Hrishabh: Now I think you are correct @t-paul . So, now I have removed this check and the user can introduce new shortcuts. That makes sense.

DLFW:
I have created #3231 as a follow-up issue for the rotation via keyboard.

@Sharma-Hrishabh: :+1:  and thanks for the focus-thing!! :)

DLFW:
Hi again,
I have raised PM #3232 for the rotation. It can be integrated with this just like this:
https://github.com/DLFW/openscad/commit/a3c17e5e424b324e2b5287af70332901dfb88432

thehans:
Disclaimer: I haven't thoroughly examined this branch to know exactly how its currently handling these cases yet...

But I think we should carefully consider and be explicit about the expected behavior for all possible cases of "conflict" when a user assigns Action X to Shortcut Y:
1. Neither X or Y are used by default 
    No problem, add new mapping
2. X has default shortcut Z (Z != Y) defined, Y is not used by any defaults
   Would this replace Z as the shortcut, or should it be possible to have multiple shortcuts for a single action?  (I would argue that yes multiple shortcuts per action should be allowed)
3. Action X has no default shortcut, Shortcut Y defaults to some other Action W.
   User is explicitly defining their own use for Shortcut Y, so default Action W should be overridden.
4. X has default Shortcut Z **and** Y defaults to some other Action W
  This is just a combination of cases 2. and 3. and I would expect it to handle each component in the same manner as those.

I think it would be useful to allow multiple shortcuts for a single action, (and on the other hand, I think all would agree that multiple actions for a single shortcut does not make any sense).
This means the relationship of Action to Shortcut is potentially "One to Many", hence it might make more sense for the json key:value pair to be swapped.  I think this would model the relationship more clearly since the json "keys" would always be unique, while allowing for a single action to appear as "values" for multiple Shortcuts.

Alternatively, I suppose another way to implement multiple shortcuts per action would be to leave Action as the json key, but allow for the value to be a json array of Shortcuts on each.

MichaelAtOz:
Should there be some action/shortcut mappings which are fixed, such as F5/F6 (as they are characteristic of OpenSCAD), maybe cut/copy/paste & quit; allowing the action to be duplicated on another shortcut, but leaving the original mapping?

Sharma-Hrishabh:
Hey @thehans , I thought about it, and your every point makes sense. This relationship of one to many should be there between action to shortcuts. So, I would try to implement it asap.

Hello @MichaelAtOz , I also agree with your point in one view, like some shortcuts are really standard for some actions, and should not be changed.

thehans:
I was just thinking there's another case I didn't mention:

5. By default, Action X is  already assigned to Shortcut Y
   This shouldn't have any effect, but would be good test case to verify that the code can handle it without issues(for example it would be a problem if it performs the same action twice per shortcut press).

Also if we are implementing non-overridable shortcuts that Michael suggests, I would do a quick test for both cases of overridable and non-overridable versions of the above.

Sharma-Hrishabh:
I have not pushed my new changes yet, as I wanted to discuss it more.
@thehans , I have followed your second approach to allow multiple shortcuts, when the user provides the array of shortcuts in the config file.
I had a discussion on the IRC Channel, about whether giving a custom shortcut, should replace the default shortcut or should just be added as a new alternative shortcut.
So, I think now, that if a user wants to keep the default shortcut, he/she can put it in the array in the config file.
The reason for supporting this view is actually behind the case if a user wants to remove a default shortcut to free the shortcuts, as suggested by @t-paul .
And for that case, just replacing is the better option.
So, the method of removing shortcut will work by giving an empty array or just leaving the value of action in JSON file as blank.
Now, another point @MichaelAtOz raised, about keeping some shortcuts fixed like Render/Preview(F6,F5) etc. For that, automatically an error is being raised 'Ambiguous shortcut overload: F5', if we open opened though terminal and try to use F5 shortcut for some other action
'. I'm still trying to figure out, why these actions - Render(F6),Preview(F5) are having so much high priority. Although we want them not to be replaced, this is already happening.

t-paul:
I don't think we should forbid users to override some of the actions. Yes, things like `F5` have become synonymous with the action Preview, but if people _really_ want to change that, why not? I guess if we support multiple shortcuts, it's less of an issue.
We could add some special shortcut name (`DEFAULT`?) to be able to refer to the default shortcut even with support for multiple shortcuts per action.

So lets collect the cases
```
{
  // remove all shortcuts
  "action1" : "",

  // overwrite default shortcut
  "action2" : "CTRL-X", 

  // overwrite default shortcut with multikey shortcut (handled internally by Qt)
  "action3" : "CTRL-X, CTRL-Y", 

  // add shortcut leaving the default one active as default
  "action4" : ["DEFAULT", "CTRL-X"],

  // add shortcut making default shortcut a secondary one
  "action5" : ["CTRL-X", "DEFAULT"],

  // set multiple shortcuts (first one will be default shortcut)
  "action6" : ["CTRL-X", "CTRL-Y"],
}
```

DLFW:
Agree to @t-paul: to prevent the user from overriding certain keys is an unnecessary restriction for the user and an comes with unnecessary complexity for the code.

At the same time, I actually don’t see a benefit for the “DEFAULT” value. If a user really wants to have multiple shortcuts for one action and one of them shall be its default, the user will certainly know that default value and is able to type it in his config file.

I wouldn’t make it unnecessary complicated. From my humble user-opinion, this feature is already pretty much done, once the F5/F6 issue has been sorted out. And it adds a lot of value. :)

t-paul:
No, the default shortcut runs through language translation, that information is not recoverable in a configuration file easily.

Sharma-Hrishabh:
Hey @DLFW. It looks complicated at once, but actually it will help users if they want to just keep the old-default shortcut, along with new shortcut. They just have to put "DEFAULT" in the config file with another shortcut of his/her choice, and the DEFAULT shortcut will not be replaced.

I think you can assign F5/F6 to other shortcuts now. You just need to define new-shortcuts for those actions which are already using F5/56 which are (Render & Preview). 
If you try to assign a shortcut key, which is already bound with other actions, then you get "Ambiguous shortcut overload" error.
So as such I think, this idea to provide user-defined shortcuts is good, but still, the user should have knowledge if he is not assigning a bad shortcut(being already used).
So, that's why I now believe, that this feature is okay, to some extent. But we will be requiring a GUI, where all the shortcuts are visible and you can change from there.

In this JSON file approach, the user has to handle the ambiguous shortcut error on his own

I will surely like to go for implementing the GUI thing if you agree @t-paul , after my exams  :sweat_smile:

Sharma-Hrishabh:
I also have another idea, we can write all the current shortcuts to the config file. In this way, the user will have a clear view of what shortcut is assigned to which action and can change any shortcut, if he wants.

Sharma-Hrishabh:
> I don't think we should forbid users to override some of the actions. Yes, things like `F5` have become synonymous with the action Preview, but if people _really_ want to change that, why not? I guess if we support multiple shortcuts, it's less of an issue.
> We could add some special shortcut name (`DEFAULT`?) to be able to refer to the default shortcut even with support for multiple shortcuts per action.
> 
> So lets collect the cases
> 
> ```
> {
>   // remove all shortcuts
>   "action1" : "",
> 
>   // overwrite default shortcut
>   "action2" : "CTRL-X", 
> 
>   // overwrite default shortcut with multikey shortcut (handled internally by Qt)
>   "action3" : "CTRL-X, CTRL-Y", 
> 
>   // add shortcut leaving the default one active as default
>   "action4" : ["DEFAULT", "CTRL-X"],
> 
>   // add shortcut making default shortcut a secondary one
>   "action5" : ["CTRL-X", "DEFAULT"],
> 
>   // set multiple shortcuts (first one will be default shortcut)
>   "action6" : ["CTRL-X", "CTRL-Y"],
> }
> ```

One more thing, we can't actually get the case mentioned with action-5, as the first KeySequence in the list is assigned as the default shortcut in Qt.

t-paul:
The `action5` case would mean:
Assign `CTRL-X` as new default shortcut but add the original (application defined one before applying the file) as secondary shortcut.

Sharma-Hrishabh:
Okay @t-paul , got it. Just to mention, action-5 is already being handled by Qt.

DLFW:
Hi @Sharma-Hrishabh! Do you know if there any further effort taken on this feature, maybe in the context of anther issue? Or is this feature on-hold/withdrawn?

Just asking because I’m heading to my next 3d-printer project and was thinking of updating my branch for #3213.

Sharma-Hrishabh:
@DLFW Progress of this feature has already been made in #3338

PR #3228, opened 2020-02-18
Only enable 30ms joystick timer when needed

### LINKED ISSUE #2921: Constant 4% cpu load from rapid poll syscalls with 1-25 ms timeout
Running `strace` on an idle openscad (gui) process shows it constantly calling poll() with extremely short timeouts. This yields a constant 4% load on my laptop and likewise fast battery drain. Interspersed with the polls are stat syscalls on the open file, so I tried turning off automatic reload, but the polls remain. I haven't traced where they're coming from but it looks like somewhere in Qt or another library, possibly as a result of incorrect use of QTimer or internal bugs in QTimer.

Ideally automatic reload should use `inotify` with no timeout in the `poll`, rather than short timeouts, on systems where `inotify` is available. Short of that, the `poll` interval should be the whole refresh rate (currently hard coded as 200 ms), not 25 ms, and it would be nice if it were configurable.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/72507457-constant-4-cpu-load-from-rapid-poll-syscalls-with-1-25-ms-timeout?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ijackson:
Fixes #2921 

The effect is that the timer is now only enabled when (i) there is a joystick and (ii) either (a) it has been moved or buttons pressed recently and we haven't processed that yet or (b) the joystick is sufficiently off-centre and out of the dead zone that we must generate input events.  When this ceases to be true, the timer is disabled again.  So this should solve the cpu usage (and battery wastage) problem for all users, even those with joysticks.

I have tested this by rebasing these patches onto Debian's 2019.05-3 package, as found in Debian testing.  I observe in strace that the program now only polls every 200ms again.  I don't have a joystick-like device so it is difficult for me to test that my change is correct, but I think it should be.

(As far as I can tell the InputMapper class is only used for joysticks, or similar input devices - ie ones which generate a stream of input events based on a position.  It would have been nice if this had been mentioned somewhere..)

### DIFF
--- a/src/input/InputEventMapper.cc
+++ b/src/input/InputEventMapper.cc
@@ -160,7 +160,7 @@ bool InputEventMapper::generateDeferredEvents()
 void InputEventMapper::considerGeneratingDeferredEvents()
 {
     if (!timer->isActive()) {
-        timer->start(30);
+        QMetaObject::invokeMethod(timer, "start", Qt::QueuedConnection, Q_ARG(int, 30));
     }
 }
 
--- a/src/input/InputEventMapper.cc
+++ b/src/input/InputEventMapper.cc
@@ -157,9 +157,16 @@ bool InputEventMapper::generateDeferredEvents()
     return any;
 }
 
+void InputEventMapper::considerGeneratingDeferredEvents()
+{
+    if (!timer->isActive()) {
+        timer->start(30);
+    }
+}
+
 void InputEventMapper::onTimer()
 {
-    generateDeferredEvents();
+    bool generated_any_events = generateDeferredEvents();
 
     //update the UI on time, NOT on event as a joystick can fire a high rate of events
     for (int i = 0; i < max_buttons; i++ ){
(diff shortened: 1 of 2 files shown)


t-paul:
I'll give it a test with joystick once I'm through fighting with the keyboard shortcut PR :smile:.

src/input/InputEventMapper.cc
```diff
@@ -140,13 +144,29 @@ void InputEventMapper::onTimer()
...
+        any = true;
+    }
+
+    return any;
+}
+
+void InputEventMapper::considerGeneratingDeferredEvents()
+{
+    if (!timer->isActive()) {
+        timer->start(30);
```
t-paul, line 163: It does not work for me (Debian/testing, Qt5.12.5), just getting
```
QObject::startTimer: Timers cannot be started from another thread
```
This still reports at least the buttons to the 3d view, but the preferences dialog does not get any events in the Axis and Buttons tabs.

Using this to forward the start call to the thread the timer is running on makes it work for me:
```suggestion
        QMetaObject::invokeMethod(timer, "start", Qt::QueuedConnection, Q_ARG(int, 30));
```

Issue #3231, opened 2020-02-20
Free rotation of the 3D view via keyboard


DLFW:
The 3D view can be zoomed and brought to some basic positions via keyboard shortcuts. But it cannot be freely rotated.

**Request:** Add keyboard shortcuts for rotating the 3D view around all three axes and for each in both directions.

Optional: Provide shortcuts for small and big rotation steps, maybe 1° and 8°.

This request is a follow-up for #2977, which has been decided to be used as tracker for “configurable” shortcuts, only. (By @t-paul in https://github.com/openscad/openscad/pull/3226#issuecomment-589074106_.)

t-paul:
The comment is on the PR and separate review of smaller things helps with limited time, but I guess it's even better to have a specific feature request too.

thehans:
Something else that might be worth thinking about is the possibility of combining the shortcut key assignment functionality with input device configuration which already exists as "Buttons" tab in preferences.  
So that some of these new actions could be be assigned to joystick buttons, and shortcut keys could be set alongside buttons through some common interface.

t-paul:
Indeed. If we go with the additional actions, it should be just a matter of registering those with the input driver manager.

jordanbrown0:
Let's generalize this to rotate, pan, and zoom; they all seem to go together and it allows closing #4448 as a duplicate.

jordanbrown0:
The "Button" mechanism doesn't currently support rotate/pan/zoom; it is largely tied to the actions available from the menus.  It does pick up a few actions from other places, notably the animation control panel and the "toggle perspective" action.  We could add actions for "rotate X by 1 degree", "rotate X by 15 degrees", "rotate Y by 1 degree", et cetera.  (Or maybe they could rotate by one of two or three preference-controlled amounts.)  Similarly for pan and zoom.  I don't think they would need to appear in any menu, other than the pull-down in the Button preference UI.

Note also that buttons are one-shot operations; they fire on press.  Continuing to hold the button has no effect.  For keyboards, autorepeat might help, depending on how it's done.

Or there could be a variation where buttons are treated as extremely-low-resolution axis controls that can report only 0.0 and +1.0; the normal axis-mapping stuff could then map that to pan, rotate, and zoom, with a configured gain.

PR #3232, opened 2020-02-20
add rotation of 3D view by shortkeys

DLFW:
Solves #3231 .

I'm not a Qt or C++ programmer but I hope it’s not too bad.

Used default shortcuts with their action names:

```
viewActionRotateSmallForwardX
Rotate X Forward Small Step
Ctrl+J

--
viewActionRotateSmallBackwardX
Rotate X Backward Small Step
Ctrl+M

--
viewActionRotateBigForwardX
Rotate X Forward Big Step
Ctrl+Shift+J

--
viewActionRotateBigBackwardX
Rotate X Backward Big Step
Ctrl+Shift+M

--
viewActionRotateSmallForwardY
Rotate Y Forward Small Step
Ctrl+K

--
viewActionRotateSmallBackwardY
Rotate Y Backward Small Step
Ctrl+,

--
viewActionRotateBigForwardY
Rotate Y Forward Big Step
Ctrl+Shift+K

--
viewActionRotateBigBackwardY
Rotate Y Backward Big Step
Ctrl+Shift+,

--
viewActionRotateSmallForwardZ
Rotate Z Forward Small Step
Ctrl+L

--
viewActionRotateSmallBackwardZ
Rotate Z Backward Small Step
Ctrl+.

--
viewActionRotateBigForwardZ
Rotate Z Forward Big Step
Ctrl+Shift+L

--
viewActionRotateBigBackwardZ
Rotate Z Backward Big Step
Ctrl+Shift+.
```

They are not exactly VIM keys but close to them: Rotation can be done with the left hand’s home row ( j k l ) and the buttons beneath ( m , . ), each combined with CTRL. When also using Shift, 8° are used instead of 1° steps.

One observation I don’t understand: The X and Y axis seem to relate to  the camera while the Z axis seems to relate to the model. Does somebody had a clue why?

### DIFF
--- a/src/MainWindow.h
+++ b/src/MainWindow.h
@@ -40,6 +40,7 @@ public:
 	QTimer *waitAfterReloadTimer;
 	QTime renderingTime;
 	EditorInterface *customizerEditor;
+    QList<QAction *> rotationActions;
 
 	ContextHandle<BuiltinContext> top_ctx;
 	FileModule *root_module;      // Result of parsing
@@ -122,6 +123,7 @@ private:
 	void setDockWidgetTitle(QDockWidget *dockWidget, QString prefix, bool topLevel);
 	void addKeyboardShortCut(const QList<QAction *> &actions);
 	void updateStatusBar(class ProgressWidget *progressWidget);
+    void rotateViewAngle(int axis, int step);
 
   class LibraryInfoDialog* library_info_dialog;
   class FontListDialog *font_list_dialog;
@@ -255,7 +257,18 @@ public slots:
 	void viewModeShowCrosshairs();
 	void viewModeShowScaleProportional();
 	void viewModeAnimate();
-    void rotateViewAngleSmallX();
+    void viewRotateSmallForwardX();
+    void viewRotateSmallBackwardX();
+    void viewRotateBigForwardX();
+    void viewRotateBigBackwardX();
+    void viewRotateSmallForwardY();
+    void viewRotateSmallBackwardY();
+    void viewRotateBigForwardY();
+    void viewRotateBigBackwardY();
+    void viewRotateSmallForwardZ();
+    void viewRotateSmallBackwardZ();
+    void viewRotateBigForwardZ();
+    void viewRotateBigBackwardZ();
 	void viewAngleTop();
 	void viewAngleBottom();
 	void viewAngleLeft();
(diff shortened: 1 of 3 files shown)


t-paul:
We may need to leave the default shortcuts empty as I think this hits the annoying MacOS behavior again where `Qt-CTRL` translates to `MacOS-Command` and that clashes with a couple of defaults in the PR.

DLFW:
Hey @t-paul , thanks for the input.
So, you mean that we really remove all of those defaults? Then no user can use that keyboard navigation out of the box, right? I don’t know anything about this MacOS issue, but aren’t there any feasible keys for this feature that can be used?

I personally won’t care since I would anyway override the defaults with my own shortcuts. I just thought that not providing defaults could kind of “hide” this feature. 

Is this PR otherwise fine for you for getting merged? I didn’t get your comment https://github.com/openscad/openscad/issues/3231#issuecomment-589636389. Is that something I have to consider within this PR?

MichaelAtOz:
hmmm...[Apparently](https://doc.qt.io/qt-5/qkeysequence.html), Ctrl+k is cut-to-end-of-line on Linux. #3186 & #3205 implies on Mac too (@kintel is an Apple man). 

And on [Mac](https://support.apple.com/en-us/HT201236) "Command-Comma (,): Open preferences for the front app." (where `Command` == `Control`); I'm not a Mac user, so don't know if OpenSCAD makes this happen, or if it is important.
Also j,k,l are listed under 'Finder and system shortcuts' on the above link - @kintel 'system' does that apply when OpenSCAD has focus?

What about using `Alt`/`Option`(mac)? It doesn't seem to get used much, tho the Alt+arrows are used already.

Also note this [wiki page](https://github.com/openscad/openscad/wiki/Keyboard-Shortcuts).

DLFW:
Hm. Since this feature would need six well placed keys, it seems there is no real option left on the alpha-numeric part of the keyboard. I thought that the numpad might be a last resort. But after some web-research and playing with the QTCreator, it seems that Qt cannot distinguish between numbers from the alpha-numeric part of the board and numbers from the numeric keypad. Pretty surprising for me.

Does somebody know if there is a way to specifically assign the numpad-keys to shortcuts?

The very last resort before giving up on standard shortcuts on this feature might then be an combination of ALT and the numbers. They seem to be free. E.g. ALT+4/6 for the z-axis, ALT+7/9 for the y-axis and ALT+2/8 for the x-axis or something similar. That would somehow make sense when looking at the numpad.

Any thoughts on this?

PR #3241, opened 2020-03-05
Print statistics in cmdline mode as in GUI

### LINKED ISSUE #2639: Output same info from GUI via command line [$50]
As per [forum](http://forum.openscad.org/Get-console-information-in-order-to-store-it-tp25022.html), the output in the console from render:
```
ECHO: version = [2018, 11, 4]
Rendering Polygon Mesh using CGAL...
Geometries in cache: 5
Geometry cache size in bytes: 754208
CGAL Polyhedrons in cache: 3
CGAL cache size in bytes: 26485440
Total rendering time: 0 hours, 0 minutes, 21 seconds
   Simple:        yes
   Vertices:     4276
   Halfedges:   20384
   Edges:       10192
   Halffacets:  11816
   Facets:       5908
   Volumes:         2
Rendering finished.
```
should be able to be collected somehow via command line.

Similarly the preview info would also be useful
```
ECHO: version = [2018, 11, 4]
Geometries in cache: 5
Geometry cache size in bytes: 754208
CGAL Polyhedrons in cache: 3
CGAL cache size in bytes: 26485440
Compiling highlights (1 CSG Trees)...
Normalized CSG tree has 4 elements
Compile and preview finished.
Total rendering time: 0 hours, 0 minutes, 0 seconds
```

Note @nophead comment 'The statistics about edges and times, etc, are generated by the GUI code in mainwin.cc, so simply don't exist in command line mode.'

Having the output would also be useful to expand the test suite and for performance comparisons between releases.

Discussion on format;  plain text output or would an encoded format be more useful, presumably JSON?

<bountysource-plugin>

---
There is a **[$50 open bounty](https://www.bountysource.com/issues/67865416-output-same-info-from-gui-via-command-line?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Fat-Zer:
- Added a helper visitor class to uniformly process Geometry
 - Move statistic printing to a dedicated class
 - Resolves #2639

### DIFF
--- a/src/RenderStatistic.cc
+++ b/src/RenderStatistic.cc
@@ -1,0 +1,99 @@
+/*
+ *  OpenSCAD (www.openscad.org)
+ *  Copyright (C) 2020 Golubev Alexander <fatzer2@gmail.com>
+ *
+ *  This program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  As a special exception, you have permission to link this program
+ *  with the CGAL library and distribute executables, as long as you
+ *  follow the requirements of the GNU GPL in regard to all of the
+ *  software in the executable aside from CGAL.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program; if not, write to the Free Software
+ *  Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ *
+ */
+
+#include "printutils.h"
+#include "GeometryCache.h"
+#include "CGALCache.h"
+#include "polyset.h"
+#include "Polygon2d.h"
+
+#ifdef ENABLE_CGAL
+#include "CGAL_Nef_polyhedron.h"
+#endif // ENABLE_CGAL
+
+#include "RenderStatistic.h"
+
+
+void RenderStatistic::printCacheStatistic()
+{
+  GeometryCache::instance()->print();
+#ifdef ENABLE_CGAL
+  CGALCache::instance()->print();
+#endif
+}
+
+void RenderStatistic::printRenderingTime(std::chrono::milliseconds ms)
+{
+  PRINTB("Total rendering time: %d:%02d:%02d.%03d",
+    (ms.count() /1000/60/60     ) %
+    (ms.count() /1000/60 % 60   ) %
+    (ms.count() /1000    % 60   ) %
+    (ms.count()          % 1000 ) 
+  );
+}
+
+void RenderStatistic::print(const Geometry *geom)
+{
+  geom->accept(*this);
+}
+
+void RenderStatistic::visit(const GeometryList& geomlist)
+{
+  PRINT("   Top level object is a list of objects:");
+  PRINTB("   Objects:     %d", geomlist.getChildren().size());
+}
+
+void RenderStatistic::visit(const PolySet& ps)
+{
+  assert(ps.getDimension() == 3);
+  PRINT("   Top level object is a 3D object:");
+  PRINTB("   Facets:     %6d", ps.numFacets());
+}
+
+void RenderStatistic::visit(const Polygon2d& poly)
+{
+  PRINT("   Top level object is a 2D object:");
+  PRINTB("   Contours:     %6d", poly.outlines().size());
+}
+
+#ifdef ENABLE_CGAL
+void RenderStatistic::visit(const CGAL_Nef_polyhedron& Nef)
+{
+  if (Nef.getDimension() == 3) {
+    bool simple = Nef.p3->is_simple();
+    PRINT("   Top level object is a 3D object:");
+    PRINTB("   Simple:     %6s", (simple ? "yes" : "no"));
+    PRINTB("   Vertices:   %6d", Nef.p3->number_of_vertices());
+    PRINTB("   Halfedges:  %6d", Nef.p3->number_of_halfedges());
+    PRINTB("   Edges:      %6d", Nef.p3->number_of_edges());
+    PRINTB("   Halffacets: %6d", Nef.p3->number_of_halffacets());
+    PRINTB("   Facets:     %6d", Nef.p3->number_of_facets());
+    PRINTB("   Volumes:    %6d", Nef.p3->number_of_volumes());
+    if (!simple) {
+      PRINT("UI-WARNING: Object may not be a valid 2-manifold and may need repair!");
+    }
+  }
+}
+#endif // ENABLE_CGAL
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -456,6 +456,7 @@ set(CORE_SOURCES
   src/FontCache.cc
   src/DrawingCallback.cc
   src/FreetypeRenderer.cc
+  src/RenderStatistic.cc
   src/ext/lodepng/lodepng.cpp
   src/PlatformUtils.cc 
   src/libsvg/circle.cc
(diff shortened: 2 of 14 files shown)


t-paul:
The AppVeyor failure seems to be just a timeout, kicking off a new build...

Fat-Zer:
By the word, I've forgot to mention on the commit that I've changed the format of time elapsed by rendering from

    Total rendering time: h hours m minutes s seconds

to

    Total rendering time: mm:ss.zzz

I hope it's not a big deal... Even for complicated models the rendering rarely takes up a whole hour, so IMHO the user might be more interested in microseconds rather than hours.

MichaelAtOz:
a. I've had models take over 24 hours
b. the formal should be consistent between GUI & CMD line, so it can be parsed regardless.

If you want milliseconds do it on both.

t-paul:
The code is nicely factored out so same format is not a problem. Maybe there's a resonable way to achieve a more dynamic format, like put in hours only when actually needed?

Fat-Zer:
@MichaelAtOz, are such renders anyway common? Or is it just a simple worst-case example?<br> 
I maybe was wrong in my assumptions, if you wish, I won't hesitate reverting the change...

And sure, it has the same format for GUI&CLI ;)

Fat-Zer:
>The code is nicely factored out so same format is not a problem. Maybe there's a resonable way to achieve a more dynamic format, like put in hours only when actually needed?

It will be harder and error-prone to parse...

t-paul:
@MichaelAtOz what's that parsing about anyway? Easy parsing and user friendly output are sort-of opposites.

Fat-Zer:
@t-paul, the issue suggests that this output might be used in a test suite and performance comparisons... It will require some basic text parsing of the log. It won't be hard to do as long as the output is pretty consistent, but by introducing of some disappearing value (e.g. flickering hour field) the date parsing would become quite painful, because you won't be able to just match it against a simple date format string e.g. `date` utility. Also if a user won't consider that the hour may appear on the output, he won't even bother to prepare the script for it.

The log-like output is pretty dumb. So it's a good fit as both: a user info messages and a parsable text stream for a specific information.

PS: sorry, for being over verbose, I haven't actually intended it ;)

t-paul:
If we are talking test suite, I don't see much of an issue. Python would be able to parse this relatively easily and it's pretty unlikely there will be automatic testcases > 1h anyway. But in general it's a valid concern.

Another option would be to use the original format but allow fractional seconds.

MichaelAtOz:
@t-paul yes as @Fat-Zer says, text parsing the log.

```
Total rendering time: h:mm:ss.zzz
```
Where `h` is just an int, so it would go to `hh` or `hhh` etc in those unlikely situations would be OK.

> are such renders anyway common? Or is it just a simple worst-case example?

24+ is not frequent*, but can happen, certainly >1h is not every day but happens, particularly with high $fn. If I was running it in the background I'd be unimpressed if it didn't actually show the time because it overflowed.

\* more often in the past when I didn't have 32GB and RAID SSDs

src/RenderStatistic.cc
```diff
@@ -0,0 +1,98 @@
...
+{
+  GeometryCache::instance()->print();
+#ifdef ENABLE_CGAL
+  CGALCache::instance()->print();
+#endif
+}
+
+void RenderStatistic::printRenderingTime(std::chrono::milliseconds ms)
+{
+  PRINTB("Total rendering time: %02d:%02d.%02d",
```
t-paul: Shouldn't the last one be `%03d`?

As discussed we also should add back the hours. I don't mind either way, just as leading value like `1:20:01.238` or going back to the original format (with added milliseconds).

MichaelAtOz:
Almost...
The preview format should be the same format as render time. (below is GUI console)

>Parsing design (AST generation)...
Saved backup file: C:/Users/MeB/Documents/OpenSCAD/backups/unsaved-backup-llSPQEhJ.scad
Geometries in cache: 146
Geometry cache size in bytes: 2559904
CGAL Polyhedrons in cache: 46
CGAL cache size in bytes: 67006192
Normalized CSG tree has 1 elements
Compile and preview finished.
**Total rendering time: 0 hours, 0 minutes, 0 seconds**
>
>Parsing design (AST generation)...
Saved backup file: C:/Users/MeB/Documents/OpenSCAD/backups/unsaved-backup-llSPQEhJ.scad
Rendering Polygon Mesh using CGAL...
Geometries in cache: 146
Geometry cache size in bytes: 2559904
CGAL Polyhedrons in cache: 46
CGAL cache size in bytes: 67006192
**Total rendering time: 00:00.21**
   Facets:          6
Rendering finished.

As mentioned above, I'm happy with h:mm:ss.zzz, where h is flexible length.

and the preview command line output is not complete ie: 
e.g. `openscad -o cube.png cube.scad`

>Compiling design (CSG Products normalization)...
Normalized CSG tree has 1 elements

it should include the same cache info as the GUI preview console above.
@t-paul unless that code is not called from command line preview??

kintel:
Minor request regarding the copyright notice: We request that authors allow us to adjust the license of this codebase in the future, either through copyright assignment, a more liberal copyright (like 2-clause BSD) on submitted code, or through an explicit grant like in https://github.com/openscad/openscad/blob/master/doc/contributor_copyright.txt.

t-paul:
@MichaelAtOz right, good point. There's no statistics for preview (as there's no actual geometry in the general case). But the time output should match.

src/RenderStatistic.cc
Fat-Zer: thanks, good catch on the microseconds format.

Fat-Zer:
Pushed suggested changes in a separate patch for review. Which includes:

 * change time format to `h:mm:ss.zzz` as suggested.
 * fix a time display for a preview in GUI.
 * print rendering statistic for rendering to PNG in preview mode
 * added missing virtual destructor declaration to `GLView` (mostly to suppress a warning)

If everything ok, I'll squash the patches...

@t-paul, fixed the time.

@MichaelAtOz, thanks, I've actually overlooked that preview prints anything... fixed both

PS: does it actually make any sense to print CGAL cache statistic for preview in both CLI and GUI? It supposed to be either zero or (for GUI) whatever left from the last run.

@kintel, I'm absolutely OK with granting you right to sublicense. But I'm not sure if I'm allowed to put a different license (or add a closure) to the file, because I've actually reused some lines from `mainwin.cc` which contains neither...

t-paul:
I'd say we leave the cache part in as there's a feature request for persistent cache. If that gets implemented, the command line call would start showing cache content too. In addition there's a couple of modules that run though CGAL even in preview mode, so I would expect something to show up in case there's calls to `minkowski()`, `resize()` or one of the other modules forcing mesh calculation.

MichaelAtOz:
- fix a time display for a preview in GUI.

Isn't working:
(OpenSCAD version 2020.03.09.ci4620)
>Compiling design (CSG Products generation)...
Geometries in cache: 124
Geometry cache size in bytes: 1907424
CGAL Polyhedrons in cache: 14
CGAL cache size in bytes: 10240096
Normalized CSG tree has 72 elements
Compile and preview finished.
**Total rendering time: 0 hours, 0 minutes, 9 seconds**

Command line preview is good.

Fat-Zer:
@MichaelAtOz, shouldn't be so... are you sure you rebuilt everything correctly/running correct version of the binary? if `openscad --version` displays a git hash as minor (`ci4620`, I'm not actually sure what it is) it should be `50cf23c3`...

By clicking on the preview in GUI I see:

    ...
    Total rendering time: 0:00:00.257

Are there any additional actions needed to reproduce that?

MichaelAtOz:
> @MichaelAtOz, shouldn't be so... are you sure you running correct version of the binary?

@Fat-Zer oops sorry, I need to tidy up my Start menu...I didn't pin the new one as I usually do...
Capture-#3241

t-paul:
@MichaelAtOz thanks for checking this out. I'll have another look, but I think we are good to go.

t-paul:
@kintel anything open regarding your licensing question?

Fat-Zer:
I'm squashed the fixup commit & rebased upon master

kintel:
Added some minor cleanup and resolved a comment, plus the conflict here: #3330

kintel:
I'm ok with the explicit right to sublicense given here.

Issue #3243, opened 2020-03-08
Customizer vectors round floats with initial integer


rcolyer:
In customizer:
v = [1.1, 2.2, 3.7]; // yields float spinboxes

v = [1.0, 2.2, 3.7]; // and
v = [1, 2.2, 3.7]; // yield int spinboxes, rounding the last two.

This is probably a simple fix, but could also be considered together with #3012.

Sharma-Hrishabh:
Hey @rcolyer , I want to look over this bug. Should I go ahead?

Issue #3098, opened 2019-10-21
Allow to set the number of decimals in customizer


stenci:
I'm not able to set the number of allowed decimals in the customizer.

For example, in my model the `dia_hole` parameter represents the diameter of a hole. The value is usually an integer between 5 and 11 mm, but sometimes can be 6.5 mm or 11.1125 mm which represents a fairly common drilling bit of 7/16 inches.

If I set the initial value in the code to the most common value `dia_hole = 6`, then I can only type integer values.

Today I need to set the default value to any (usually) meaningless value with 4 digits and ask the user to (almost) always change that value.

It would be nice if I could do something like this:

    dia_hole = 6 // Min:5 Max:20 Decimals:4 SpinnerRounding:0.5

This would allow me to:
- constrain the value between 6 and 20 (included)
- allow the user to type any number with 4 decimals
- while keeping the spinner useful by rounding the next up/down value to the closest multiple of 0.5

t-paul:
The current customizer syntax is meant to be compatible with what [Thingiverse supports](https://customizer.makerbot.com/docs). We should have a native way of specifying additional information at some point. In addition the GUI could be extended to be more flexible.

stenci:
Here I am again, two days in a row: a cylinder with diameter = 10 mm is usually printed as 10 mm, but with some resins it may be 10.15 mm. So I added a parameter called `th_comp` used to compensate for thickness errors. I would like for the default to be 0, but that would prevent me from typing any decimal value. So I am forced to set the default to 0.01.

Setting the default values to the wrong values only to make the interface happy doesn't feel right.

Are there other workarounds?

t-paul:
Maybe
```
th_comp = 0; // [0:0.01:0.5]
cylinder(r = 10, h = 10 + th_comp);
```
For further discussion, please use the [forum/mailing-list](http://forum.openscad.org/).

buteomont:
This is a big problem for me as well.  Seems that you should be able to enter any number directly into the customizer's input fields, regardless of the original value.

pedjas:
I am supporting this request.

I even tried to set 

`thickness = 9.0;`

But GUI still shows only integer and does not allow user to enter decimal value. User can type in decimal point but no figures after that.

ronsmits:
I was just playing with this and setting the thickness to 9.01 makes it possible

1950sRobot:
[...]

Issue #3092, opened 2019-10-13
Transition double-conversion to "Ryu"


thehans:
I recently learned about a new float conversion algorithm called "Ryu", which sounds quite nice.

It quickly and automatically converts to the shortest representation which is round trip compatible, so converting to string then from string will result in the exact same input.

This could potentially be even more performant than the `double-conversion` library which was recently implemented for performance reasons.

The main repo is here:
https://github.com/ulfjack/ryu

Link to the technical paper: 
https://dl.acm.org/citation.cfm?id=3192369

A presentation video from the algorithm author: 
https://www.youtube.com/watch?v=kw-U6smcLzk

And another video from Stephan T. Lavavej about how Microsoft is putting this algorithm into their C++17 `charconv` STL implementation:
https://www.youtube.com/watch?v=4P_kbF0EbZM

So `charconv` could be another way to go if/when OpenSCAD transitions to C++17, but since the base Ryu code is in C only, there would be no need to raise the stdlib requirements if OpenSCAD were to implement only that.

PR #3246, opened 2020-03-09
fix customizer vectors highest precision bug

Sharma-Hrishabh:
I have tried to do a quick fix for #3243 
Two approaches came into my mind-
1) To look over the highest precision by dividing the value by 10.0 and keeping the count until the value becomes 0.0
2) Converting the value into the string and using string manipulation (I found this approach on the internet), while I was searching for some inbuilt C++ function to do the same.

As, toDouble() already truncates(roundoff) the doubles to 5 decimal places, so both the approaches will take nearly constant time.
I chose the second approach, as working with doubles/floats is always tricky.

### DIFF
--- a/src/parameter/parametervector.cpp
+++ b/src/parameter/parametervector.cpp
@@ -52,9 +52,12 @@ void ParameterVector::setValue()
 	double minV = object->values->toRange().begin_value();
 	double step = object->values->toRange().step_value();
 	double maxV = object->values->toRange().end_value();
+
+	double highestPrecisionValue = getHighestPrecision(vec);
+	
 	if(step==0){
 		step=1;
-		setPrecision(vec.at(0)->toDouble());
+		setPrecision(highestPrecisionValue);
 	}else{
 		setPrecision(step);
 	}
--- a/src/parameter/parametervector.cpp
+++ b/src/parameter/parametervector.cpp
@@ -52,12 +52,10 @@ void ParameterVector::setValue()
 	double minV = object->values->toRange().begin_value();
 	double step = object->values->toRange().step_value();
 	double maxV = object->values->toRange().end_value();
-
-	double highestPrecisionValue = getHighestPrecision(vec);
 	
 	if(step==0){
 		step=1;
-		setPrecision(highestPrecisionValue);
+		setPrecision(vec);
 	}else{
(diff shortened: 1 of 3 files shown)


t-paul:
That seems to duplicate the behavior of `ParameterVirtualWidget::setPrecision()` which does that calculation in a different way.

I wonder if there's a way to use the double-conversion library. I've seen the library mentioned in #3092 has a way of getting the highest precision that can restore the same number, not sure if double-conversion allows that. That is supposed to get rid of those 3.0000000001 issues.

Sharma-Hrishabh:
I tested the Ryu library, it is working fine, but still, I'm confused, why in converting back the obtained character array to double, it always pads with extra zeroes up to 6 decimal places, if required, that is not desirable. Actually, the paper and video mentioned that it would cut off as many digits as possible, but while converting back to the double again, it is rounding the result to 6th decimal place. I think I need to give some more time in understanding the paper.

Also, I got your point of repetition, of the same code. So, for time being, is it a good idea to modify setPrecision method to get the highest precision from the vector and then set the precision according to it, keeping the same approach?

t-paul:
Ah, interesting. I need to watch that video again.

To keep this PR simple, lets just unify those 2 functions so we don't calculate twice with different strategies. It sounds like we do want to have some rounding to prevent that crazy high number of decimals case.

Sharma-Hrishabh:
Yes, I agree. The current implementation supports the highest precision up to 7 decimal places and rounding off according to that. So, 3.00000004000001 truncates to 3.0000000 & 3.00000006000001 truncates to 3.0000001.
It doesn't cut automatically to the possible as many as digits. It is being hardcoded to look up to 7 decimal places and rounding off accordingly if the number of decimal digits are more than that.

Sharma-Hrishabh:
I just read #3098 , extending the GUI for this would be a good idea.

Sharma-Hrishabh:
Hey @t-paul , I think, rather than checking the highest Precision, it would be great to add a GUI box in the customizer itself to set the precision.

t-paul:
Yes, I think we should have more options in the GUI, to give the user actual control. But this PR is still useful as it could apply for the hopefully soon to be released bugfix version.

Issue #3229, opened 2020-02-18
projection() crashes after Boolean op


Parkinbot:
While this code works on OpenSCAD 2019.12.21.ci4161 on Windows 8.1 (64bit):
```
projection()
  cube([1,1,0]);  // malformed
```
the following code crashes.
```
projection()
intersection()    // crashes also on difference()
{
  cube([1,1,0]);  // malformed
  cube();    // any object
}
```
Looks to me like projection() has a problem with a "empty" object returned from CGAL

tbharathchandra:
yeah, the error is a segmentation fault, which i think is caused by the CGALUtils::createPolySetFromNefPolyhedron3 in the process of accessing Nef polyhedron of the childnode.

tbharathchandra:
I think the condition `if (this->x > 0 && this->y > 0 && this->z > 0 && !std::isinf(this->x) && !std::isinf(this->y) && !std::isinf(this->z))`  in PrimitiveNode::createGeometry() method an empty nef_polyhedron is created. and the application crashes when CGAL_forall_halffacets method in createPolySetFromNefPolyhedron3 tries to access that empty nef_polyhedron.

PR #3248, opened 2020-03-11
fix to issue #3229 

tbharathchandra:
I think the condition `if (this->x > 0 && this->y > 0 && this->z > 0 && !std::isinf(this->x) && !std::isinf(this->y) && !std::isinf(this->z))`  in PrimitiveNode::createGeometry() method creates an empty nef_polyhedron. and the application crashes when CGAL_forall_halffacets method in createPolySetFromNefPolyhedron3 tries to access that empty nef_polyhedron. issue #3229


t-paul:
The crashes in the test suite might be caused by `chN` actually being `nullptr`. So maybe we need both the `nullptr` check and the new `isEmpty()` check?

PR #3253, opened 2020-03-23
Range precision fixes

thehans:
I wanted to change the range calculation to return `begin_val + i * step_val`, instead of performing an addition on each iteration which accumulates errors unnecessarily.  

This solves most of of the floating point precision type of issues that can arise with fractional steps in ranges.

The calculation of `RangeType::numValues` has also been tweaked to account for floating point rounding errors, plus some other edge cases noticed while testing.

Here is a test I used to check if ranges generat the correct number of values, all the way up to maximum 9999 elements:
`echo("Total Mismatches (of 9999): ", len([for (d=[1:9999]) let(l=len([for(x=[0:1/d:1]) 0])) if (l != d+1) 0]));`

On the nightly release, this makes many reports 4877 mismatches.
With these changes, it is gets 0 mismatches.

A cut down version of the test has been added (only divisors up to 1000), since the full test takes 12s or so on my dev machine which seemed a little too long.


rcolyer:
This appears much more robust than the previous method.  I tried making some pathological inputs and checking for integer equality on various values.  (The old approach infinite looped on some of my tests, so this new one is definitely better.)  It passed except for some anomalous behavior near 2^32 where the uint32_t logic breaks down and confuses it before the 9999 limit is caught, and it actually does strange things like generate an iteration over just the first element, or fails to trigger the limit warning.

These pathological cases are all resolved if you change value.cc line 1029 to read:
```
  return std::nextafter((end_val - begin_val) / step_val,
      std::nextafter(
        static_cast<decltype(step_val)>(std::numeric_limits<uint32_t>::max()),
        0));
```

The test code:

```
module PathologicalInputs() {
  // Expected:  too many elements
  // Actual:  0
  echo("[0:1:4294967296] end capped");
  for (i=[0:1:4294967296]) { if ((i<2) || (i>=4294967295)) {echo(i);} }

  // Expected:  too many elements
  // Actual:  0
  echo("[0:1:8589934592] end capped");
  for (i=[0:1:8589934592]) { if ((i<2) || (i>=8589934592)) {echo(i);} }

  // Expected:  0, 1, 4294967294, 4294967295  or too many inputs
  // Actual:  no output
  echo("[0:1:4294967295] end capped");
  for (i=[0:1:4294967295]) { if ((i<2) || (i>=4294967294)) {echo(i);} }

  // Expected:  0, 1, 4294967293, 4294967294  or too many inputs
  // Actual:  too many elements (4294967295)
  // Note:  Oddly inconsistent with the 0:42949679295 case right above.
  echo("[0:1:4294967294] end capped");
  for (i=[0:1:4294967294]) { if ((i<2) || (i>=4294967293)) {echo(i);} }

  // Expected:  0, 1, 4999, 5000
  // Correct
  echo("[0:1:5000] end capped");
  for (i=[0:1:5000]) { if ((i<2) || (i>=4999)) {echo(i);} }

  // Expected:  "diff=0, i==1", "diff=0, i=5000"
  // Correct
  echo("[0:1:5000] difference from 1, 5000");
  for (i=[0:1:5000]) {
    if ((i>0.5)&&(i<1.5)) {
      echo("diff=", i-1, (i==1)?", i==1":", i!=1");
    }
    if (i>4999) {
      echo("diff=", i-5000, (i==5000)?", i==500":", i!=5000");
    }
  }

  // Expected:  "diff=0, i==1", "diff=0, i=5000"
  // Correct
  echo("[0:1:5000] difference from 1, 5000");
  for (i=[0:1:5000]) {
    if ((i>0.5)&&(i<1.5)) {
      echo("diff=", i-1, (i==1)?", i==1":", i!=1");
    }
    if (i>4999) {
      echo("diff=", i-5000, (i==5000)?", i==500":", i!=5000");
    }
  }

  // Expected:  "diff=0, i==1"
  // Correct
  echo("[0:1] difference from 1");
  for (i=[0:1])
    if (i>0) {
      echo("diff=", i-1, (i==1)?", i==1":", i!=1");
    }
}


PathologicalInputs();

**Comment by rcolyer** [comment]
This appears much more robust than the previous method.  I tried making some pathological inputs and checking for integer equality on various values.  (The old approach infinite looped on some of my tests, so this new one is definitely better.)  It passed except for some anomalous behavior near 2^32 where the uint32_t logic breaks down and confuses it before the 9999 limit is caught, and it actually does strange things like generate an iteration over just the first element, or fails to trigger the limit warning.

These pathological cases are all resolved if you change value.cc line 1029 to read:
```
  const uint32_t max = std::numeric_limits<uint32_t>::max();
  uint32_t num_steps = std::nextafter((end_val - begin_val) / step_val, max);
  return (num_steps == max) ? max : num_steps + 1;
```

The test code:

```
module PathologicalInputs() {
  // Expected:  too many elements
  // Actual:  0
  echo("[0:1:4294967296] end capped");
  for (i=[0:1:4294967296]) { if ((i<2) || (i>=4294967295)) {echo(i);} }

  // Expected:  too many elements
  // Actual:  0
  echo("[0:1:8589934592] end capped");
  for (i=[0:1:8589934592]) { if ((i<2) || (i>=8589934592)) {echo(i);} }

  // Expected:  0, 1, 4294967294, 4294967295  or too many inputs
  // Actual:  no output
  echo("[0:1:4294967295] end capped");
  for (i=[0:1:4294967295]) { if ((i<2) || (i>=4294967294)) {echo(i);} }

  // Expected:  0, 1, 4294967293, 4294967294  or too many inputs
  // Actual:  too many elements (4294967295)
  // Note:  Oddly inconsistent with the 0:42949679295 case right above.
  echo("[0:1:4294967294] end capped");
  for (i=[0:1:4294967294]) { if ((i<2) || (i>=4294967293)) {echo(i);} }

  // Expected:  0, 1, 4999, 5000
  // Correct
  echo("[0:1:5000] end capped");
  for (i=[0:1:5000]) { if ((i<2) || (i>=4999)) {echo(i);} }

  // Expected:  "diff=0, i==1", "diff=0, i=5000"
  // Correct
  echo("[0:1:5000] difference from 1, 5000");
  for (i=[0:1:5000]) {
    if ((i>0.5)&&(i<1.5)) {
      echo("diff=", i-1, (i==1)?", i==1":", i!=1");
    }
    if (i>4999) {
      echo("diff=", i-5000, (i==5000)?", i==500":", i!=5000");
    }
  }

  // Expected:  "diff=0, i==1", "diff=0, i=5000"
  // Correct
  echo("[0:1:5000] difference from 1, 5000");
  for (i=[0:1:5000]) {
    if ((i>0.5)&&(i<1.5)) {
      echo("diff=", i-1, (i==1)?", i==1":", i!=1");
    }
    if (i>4999) {
      echo("diff=", i-5000, (i==5000)?", i==500":", i!=5000");
    }
  }

  // Expected:  "diff=0, i==1"
  // Correct
  echo("[0:1] difference from 1");
  for (i=[0:1])
    if (i>0) {
      echo("diff=", i-1, (i==1)?", i==1":", i!=1");
    }
}


PathologicalInputs();
```

thehans:
@rcolyer OK, I've added your fix and additional test cases, thanks.

Issue #3262, opened 2020-03-27
Large text fails


RonaldoCMP:
```
s=2000;
text("Text", size=s);
translate([0,-1.5*s])
text("Text", size=1.5*s);
translate([0,-2*s])
text("Text", size=2*s);
```
![Text_failure](https://user-images.githubusercontent.com/19966253/77803515-34f24a80-7075-11ea-9ec0-56daf023db5a.PNG)

thehans:
Hrm, I think the error is related to the call here causing some sort of overflow issue:
https://github.com/openscad/openscad/blob/cc2dfa14eb3922d93c15f30734e8211c77ceada1/src/FreetypeRenderer.cc#L212
And the scale constant is set to `1000` here:
https://github.com/openscad/openscad/blob/cc2dfa14eb3922d93c15f30734e8211c77ceada1/src/FreetypeRenderer.cc#L49

Looks like Freetype uses a weird type (signed fixed-point 26.6) for its size value:
https://www.freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_set_char_size
https://www.freetype.org/freetype2/docs/reference/ft2-basic_types.html#ft_f26dot6

Although I'm still not quite sure I understand what's happening, because 26bit signed should go up to maximum of `33554431`, or `33554.431` before applying "scale".  And your example gives errors at just `3000`.

Maybe we would need some extra bounds checking and scaling "in post" for very large sizes.

Also just a note that there are console messages the first time this is previewed:
```
Could not get glyph 55 for char at index 0 in text 'Text'
Could not get glyph 72 for char at index 1 in text 'Text'
```

PR #3263, opened 2020-03-28
Fix for Issue #3262

thehans:
Size is now applied outside of freetype, from DrawingCallback


thehans:
creating link to issue #3262

testdata/scad/2D/issues/issue3262.scad
```diff
@@ -0,0 +1,7 @@
+s=2000;
+text("Text", size=s);
+translate([0,-1.5*s])
+text("Text", size=1.5*s);
+translate([0,-2*s])
```
t-paul, line 5: Should that be
```suggestion
translate([0,-2*s-1.5*s])
```
so there's no overlap of the 3 texts?

thehans: Yes, i was just somewhat-lazily copying the test case from the original issue without changes, but that makes it looks a bit nicer.  I have pushed the update.

PR #3265, opened 2020-04-01
Add Export STL option to Viewer-Toolbar when Editor Pane is hidden 

### LINKED ISSUE #1900: Feature request - When Editor Pane hidden add Export STL Icon to preview icon list [$10]
Or just add it regardless.
When working with an external editor I miss the convenience of the export icon, on the now hidden editor pane.

[$10 Bounty](https://www.bountysource.com/issues/40462106-feature-request-when-editor-pane-hidden-add-export-stl-icon-to-preview-icon-list)

<bountysource-plugin>

---
There is a **[$10 open bounty](https://www.bountysource.com/issues/40462106-feature-request-when-editor-pane-hidden-add-export-stl-icon-to-preview-icon-list?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Sharma-Hrishabh:
Fixes #1900

### DIFF
--- a/src/mainwin.cc
+++ b/src/mainwin.cc
@@ -352,6 +352,7 @@ MainWindow::MainWindow(const QStringList &filenames)
 	connect(this->designActionDisplayCSGTree, SIGNAL(triggered()), this, SLOT(actionDisplayCSGTree()));
 	connect(this->designActionDisplayCSGProducts, SIGNAL(triggered()), this, SLOT(actionDisplayCSGProducts()));
 	connect(this->fileActionExportSTL, SIGNAL(triggered()), this, SLOT(actionExportSTL()));
+	connect(this->fileActionExportSTL2, SIGNAL(triggered()), this, SLOT(actionExportSTL()));
 	connect(this->fileActionExport3MF, SIGNAL(triggered()), this, SLOT(actionExport3MF()));
 	connect(this->fileActionExportOFF, SIGNAL(triggered()), this, SLOT(actionExportOFF()));
 	connect(this->fileActionExportAMF, SIGNAL(triggered()), this, SLOT(actionExportAMF()));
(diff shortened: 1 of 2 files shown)


t-paul:
It should be possible to use the same action in multiple places. That would also prevent 2 showing up when editing keyboard shortcuts.

MichaelAtOz:
@Sharma-Hrishabh  Looks good, thanks.
Once @t-paul is happy, you can claim the bounty.
IIRC you need to 'Solve Issue', [here](https://www.bountysource.com/issues/40462106-feature-request-when-editor-pane-hidden-add-export-stl-icon-to-preview-icon-list), then I close the issue and you can claim it.

Sharma-Hrishabh:
Surely it is possible @t-paul . I tried this at first, but keeping the same action in both the toolbar, will not allow to show the option in one condition and hide it in another condition. Consider-

`viewerToolBar->actions().at(2)->setVisible(true);`
`editortoolbar->actions().at(12)->setVisible(false);`

If we have used the same action, then both the above lines would have returned reference to the same action, and the line which will be executed recently will have the final effect, so by the above snippet, export STL will be hidden at both the places- editor as well as viewer toolbar.

To avoid that condition, I made a different instance of the action.

t-paul:
If we want to hide the action, we'd need to remove if I suspect. Is the hiding used currently?

Sharma-Hrishabh:
Yes, I used hiding currently. But your point, that it won't be a good idea for keyboard shortcuts lead me to these two methods in the documentation. 
[https://doc.qt.io/qt-5.12/qwidget.html#insertAction](https://doc.qt.io/qt-5.12/qwidget.html#insertAction)
[https://doc.qt.io/qt-5.12/qwidget.html#removeAction](https://doc.qt.io/qt-5.12/qwidget.html#removeAction)
I think they will help,

src/mainwin.cc
```diff
@@ -2757,8 +2757,13 @@ void MainWindow::hideEditor()
 {
 	if (viewActionHideEditor->isChecked()) {
 		editorDock->close();
+		QAction *exportSTL = editortoolbar->actions().at(12);
```
t-paul, line 2760: I think that should be needed. The action is globally available in the MainWindow class via `this->fileActionExportSTL` (see the connect call using that).

t-paul:
I think it's trying to catch the event of hiding the editor in the wrong place. It seems to work when using the "Hide Editor" menu entry, but not when just closing the editor pane via the GUI close button.

Maybe `MainWindow::on_editorDock_visibilityChanged()`?

Sharma-Hrishabh:
@t-paul , I tried the same but it has a problem. if we leave the editor dock hidden and then relaunch the software, the option is not there until you tweak with the visibility of editor. So, I think we can also add one line in the constructor of MainWindow, to handle that.

src/mainwin.cc
Sharma-Hrishabh: Noted :)

t-paul:
Yes, good point. That's the same issue which makes the preferences dialog code so annoyingly verbose. I don't think there's much way around having both the init and the update case handled.

Issue #3188, opened 2020-01-13
Feature Request: PDF Export


sbridger:
There is not (for me at least) a workable way to get a correct size paper 2D drawing from Openscad without pain. 
.
DXF export -> Online-DXF->PDF_Convert results in vanishingly fine lines that do not print on a laser printer.
SVG is not directly printable, nor is PNG or anything else. PDF is the only format that can be universally printed from any computer and give a correct size print.

There are many cases where a 2D design is needed and an accurate template for drilling or cutting is required, or for laser cutting/engraving.
It would also help reduce wastage of 3d printer plastic by allowing much cheaper easier size checking. (today I am 3D printing a 14g ABS jig, that could have been made easier with wood scrap and a drill press - if I could print the paper template)

Export to PDF should have options (in decreasing order of importance to me) to:
 
 1. Print the DXF export outline with specified (configurable), visible line width - OR-
 2. Print the PNG/SVG export view
 3. have an X,Y scale ruler along the edge for calibration, and/or a calibration grid
 4. print all the "echo" output on the page
 5. filename, datetime in footer, Title etc
 6. Multiple Pages: do multiple renders (each adding one PDF page) with an iterator parameter ($t). This is really much the same as the existing Animation functionality. (of course this is also useful for stl export as well, but PDF doesn't need multiple files). This would be very useful for laser cutting multiple parts that make an object up.
7. able to obey color() : https://github.com/openscad/openscad/pull/2035

 6b. This https://github.com/openscad/openscad/issues/98#issuecomment-7512305 was someone elses take on 6

t-paul:
Sorry, this is way to broad a feature request including features that are out of scope for OpenSCAD language (`export()`).
I'd suggest to provide a clearer example/description of the core feature for discussion.

sbridger:
Not sure what is unclear: Print the DXF outline to PDF so it can be printed on paper.

Core feature: is 1 or 2 (either works for a drill template. 1 is probably more important for laser cutter)
Others are in decreasing order of importance and optional options.

3 is fairly important as it's quite easy to print pages with "scale to fit paper" and get a slightly wrong sized print

thehans:
Line thickness is not desirable for laser cutters.  A toolpath is ideally a zero-width line, and most laser cutter software can already load DXF directly.  Adding width to DXF lines would mean that laser would do 2 passes on everything, one for each side of the line.

sbridger:
[...]

Issue #3088, opened 2019-10-08
Discuss Function literals & Objects etc.


MichaelAtOz:
I'm raising this issue so the technical issues don't get bloated with discussion.
Function literals, @doug-moen's [description](https://github.com/doug-moen/openscad2/blob/master/rfc/Functions.md), and initial discussion in @t-paul's implementation issue #3077, and Geometry Objects, Doug's [description](https://github.com/doug-moen/openscad2/blob/master/rfc/Objects.md) and tp's initial work #3087, are proposed as the next evolution of OpenSCAD.

Note:
>Terminology!
We need an agreed term for 'a geometric thingi', so until a better term comes along, to avoid confusion I'll use Thingi, (note I did not use 'object'). 'script' may also need a better term, but will do for now - a section of OpenSCAD code.

[posting now so this OP remains as the overview, next post continues discussion]

MichaelAtOz:
@nophead said:

>But why? What is wrong with modules?
>And how to I have a script to make a fan, that can also cut it holes and make fan assembly?

Nothing is wrong with modules, but as it stands you can't do object-oriented programming.

Having quickly reread Doug's doco, I need to go over it again, as I was interpreting some things with my own presumptions, so I need to think again about some of the fine print (like when the object is instantiated into a Thingi). I'm also not natively imbued with object-orientation.
But as a generalisation I see the adoption of these two changes as a means of better supporting libraries and class like containerisation. The namespace like 'fields' helps.

I'll re-read it today and comment further.

nophead:
So instead of
include <screws.scad>
screw(M3_cap_screw, 10);

I could do?
screws = script(screws.scad);
screws.screw(screws.M3_cap_screw, 10);


Can I also do?
screws = script(screws.scad);
screw = screws.screw;
the_screw = screws.M3_cap_screw;
screw(the_screw, 10);

t-paul:
I think the answer with the design so far is: it depends... on how things are defined inside the `screws.scad`. Right now I don't yet see how to handle lets call them "old style module definitions" imported from a script.

**ext.scad**
```
// data
a = 42;
b = a + 8;

// function literal
func = function(x) x * x;

// object literal
o = {
    r = 2;
    h = 10;
    data = [ 1, 2, 3, 4];
    nested_func = function(x) x + r;
    cylinder(r = r, h = h);
};

// old style function
function func_old(x) = x + x;

module mod_old(x) {
    cube([x, x, a]);
}
```

```
[...]

PR #3266, opened 2020-04-02
PDF export of 2D Outlines

tbharathchandra:
This adds Export of 2D model to PDF as requested in #3188 using [Cairo](https://www.cairographics.org/) 

- [x] Add Cairo to the project
- [x] Add 'Export as PDF' to the export menu. 
- [x] Add Export 2D model to PDF single page portrait, no scaling, origin bottom left
- [x] Add Header: Title, Filename.
- [x] Add X/Y scale
- [x] Follow suggestion regarding border and scales from https://github.com/openscad/openscad/pull/3266#issuecomment-612693457
- Set meta data (may need to be conditional as it's only availabe with cairo 1.16)
  - [x] "Title" = just source file name (no path)
  - [x] "Creator" = "OpenSCAD (https://www.openscad.org/)" (no version number for now)
  - ~~"Producer" = "cairo (https://www.cairographics.org)" (removing default version number, adding https instead of http)~~ not possible with cairo 1.16
  - [x] Remove create & modification date (or set to empty string?)
- [x] Add test case


tbharathchandra:
Travis build passed https://travis-ci.org/github/openscad/openscad/builds/670874847

t-paul:
Trying the current state, that looks nice already. One (hopefully simple) additional idea would be to fill out some of the meta data (screenshot of the file properties in evince pdf viewer):

grafik

Maybe we could at least set `Creator` to OpenSCAD (`openscad_shortversionnumber` from `version.h`). I'm not sure if it's easy enough to pass the source file name through to the export function for use as `Title`.

t-paul:
The coordinate transformation needs some changes though, e.g. script
```
n = 20;

p = [
    for (i = [0 : n - 1])
        let(a = 360 / n * i)
            (50 * (i%2) + 30) * [ -sin(a), cos(a) ]
    ];
    
translate([100, 150]) difference() { polygon(p); circle(20); }
translate([1, 1]) text("Hello World!", 26);
```
produces:
grafik grafik

src/export_pdf.cc
```diff
@@ -0,0 +1,101 @@
...
+    int maxx = (int)ceil(bbox.max().x());
+    int miny = (int)ceil(bbox.min().y());
+    //int width = maxx - min
+    //int height = maxy - miny;
+
+
+    if(minx>=0 && miny>=0 && maxx>=0 && maxy>=0){
+
+        cairo_translate(cr, MARGIN, HPOINTS-MARGIN);
+        cairo_rotate(cr,270.*(M_PI/180.));
```
t-paul, line 82: The build failure on CircleCI is caused by missing definition of `M_PI`. I'm not sure why there are different setups, on my local machine (Debian/Testing) the code works fine as it is.

Error is:
```
../src/export_pdf.cc:82:31: error: 'M_PI' was not declared in this scope

         cairo_rotate(cr,270.*(M_PI/180.));
```

To make sure it compiles on those other systems too, you can use that include (example from `calc.cc`):
https://github.com/openscad/openscad/blob/2e22499abade7f5540071bcd236dd162feeb4d07/src/calc.cc#L27-L28

tbharathchandra: Yeah I replaced math.h with cmath.

tbharathchandra:
Travis build passed https://travis-ci.org/github/openscad/openscad/builds/671328177

t-paul:
I've added cairo to the MXE docker image, so the windows builds on CircleCI are fine too now. I will leave the AppVeyor with cairo disabled to verify this will build too. Right now it's probably still ignoring the ENABLE_CAIRO flag, so it fails with
```
src/export_pdf.cc:1:10: fatal error: cairo.h: No such file or directory
```
The `export_3mf.cc` file could serve as example for handling this case I think.

t-paul:
Nice! All green. I'll have a look to get it enabled for the AppImage too.

tbharathchandra:
```
n = 20;

p = [
    for (i = [0 : n - 1])
        let(a = 360 / n * i)
            (50 * (i%2) + 30) * [ -sin(a), cos(a) ]
    ];
    
translate([100, 150]) difference() { polygon(p); circle(20); }
translate([1, 1]) text("Hello World!", 26);
```
Now this script will generate pdf like 
[testbottom.pdf](https://github.com/openscad/openscad/files/4439839/testbottom.pdf)

Translating the entire geometry above to [-100,-40] will generate pdf like
[testcenter.pdf](https://github.com/openscad/openscad/files/4439845/testcenter.pdf)

t-paul:
This looks great already. Some things I'd list for discussion (lets not just change the code yet, maybe we decide it's fine as it is :smile:)

- Not sure about the red color. Maybe we want to stick to gray scales until we can give the user a choice?
- Maybe add minor ticks to the scale? Only for the first couple of cm?
- Align the scales text either centered or on one side?
- For the centered view, maybe still put the scales at top&left, but don't let them overlap, so they stand separately. Maybe some extra thick tick mark for the center points?

tbharathchandra:
- I added red color in an only intention to make geometry differ from the axis lines. Well, we can do that by using the alpha parameter. 
- Smaller ticks would make it look clean but why for only a couple of cm?
- Text aligning on one side would be better.
- Yeah, that would be great, I was also thinking to remove those lines over which scales are drawn and draw them at the end of the paper. Top X and Left Y.

t-paul:
I'm not sure if it looks too "full" when having the minor ticks everywhere. I guess we can just try and see if it's ok to just add them for the whole ruler. Due to the alpha settings, it's probably fine.

Hi @sbridger, maybe you want to have a look too :-).

sbridger:
> Hi @sbridger, maybe you want to have a look too :-).

Thanks I will, great (and unexpected) to see this happen. 
How do I get an exe?

My default choice would be edge scales as they are out of the object area.
Don't get obsessive about the scales. They are there so that print accuracy (x and Y) can be checked, they don't have another actual function. (Printers tend be be accurate in X, but a bit random in the friction feed direction Y)

Filename in header is the name of the openscad file, not the pdf file.

My intial use case for this was to make a cutting and drilling template on the laser printer. The default line colour should be solid black for that.

t-paul:
For now only the windows builds are enabled, AppImage for Linux is not ready yet (but I can build one if needed).

So for Windows find the last change like that:
grafik
and click the green check mark.

grafik

`openscad-mxe-64bit` would be the Windows 64bit version, follow the `Details` link.

For some strange reason, this link does not directly show the download tab, so you have to modify from
`https://circleci.com/gh/openscad/openscad/4758?...some...sufff...`
to
`https://circleci.com/gh/openscad/openscad/4758#artifacts`

sbridger:
Tested, works OK, prints OK on my laser. Meets my need with these changes:

- Should be black
- default to scales at edge.
- filename should be source (openscad) file, not pdf filename

I suggest that 2D only export should be a bit more explicit so people don't expect to get a picture
eg
export->"as PDF 2D Outline"
or put some labelled dividers in the export menu
export  --3D--
   as XXX etc
   --Outline 2D--
   as DXF
   as PDF
  --Image--
  as PNG
  as SVG

------------------
How much further do y'all want to go with PDF export?

tbharathchandra:
Hi @sbridger, Thanks for testing it. 

There is no problem with making the color black and taking the scales to edges. I think the problem may arise with the third modification you suggested. Though it is a useful feature in the case of PDF export, source filename is not used in any other export feature. The code is also designed accordingly. Even then, we can add this feature but this may bring some inconsistency in the code. 

I think it's a tradeoff here, @t-paul should make a call for it.

tbharathchandra:
About the testing, converting the PDF into an image and then running the check on that would be a better option. We can use Ghostscripts for converting PDFs into PNG.

t-paul:
I think it sounds sensible to have the source file in the PDF. Currently there's not much info passed into the export functions, but I think it may make sense to add an ExportInfo or ExportContext object that could later be extended to pass even more data like the options gathered by a future export options GUI.

For the menu entries, I would like to see first how things look in various combinations, we can leave that for later maybe. I think at this point it's still strictly 2D or 3D for each of the export types, but there already was some code to export a 3D view as SVG.

t-paul:
Oh, about converter, I guess we have to check what's available best on all platforms without too much installation going on.

So far I've seen:
- Ghostscript
- pdftocairo (from poppler utilities)
- imagemagick/graphicsmagick

MichaelAtOz:
> Currently there's not much info passed into the export functions, but I think it may make sense to add an ExportInfo or ExportContext object that could later be extended to pass even more data like the options gathered by a future export options GUI.

There is space in STL for more info, I think we should include filename, date/time & 'OpenSCAD=\<version\>'. And similar where feasible in other formats.

tbharathchandra:
samples:
[Untitled.pdf](https://github.com/openscad/openscad/files/4461842/Untitled.pdf)
[starcenter.pdf](https://github.com/openscad/openscad/files/4461844/starcenter.pdf)

t-paul:
The examples look great. May need some real world printing, if the scales show up, but I guess todays printers should be able to handle that fine.
Is the centered one a bit too high? It somehow looks if it's offset maybe twice as much as it should?

nophead:
No don't put date/time inside STL files. They have a date and time on the file system. Making files change when their contents haven't is not good for automatic build systems and source control. Similarly with OpenSCAD version. I would just have to strip it all out.

t-paul:
Yeah, I'm seeing the "Reproducible Build" topic too. I guess lets have that as default and maybe give an option later to add more info if people want to. I do see some cases where at least having the version number of the creating application could be useful.

src/mainwin.cc
```diff
@@ -1947,7 +1948,15 @@ void MainWindow::sendToOctoPrint()
...
 	}
 
-	exportFileByName(this->root_geom, exportFileFormat, exportFileName.toLocal8Bit().constData(), exportFileName.toUtf8());
+    ExportInfo exportInfo;
+    exportInfo.format = exportFileFormat;
+    exportInfo.name2open = exportFileName.toLocal8Bit().constData();
+    exportInfo.name2display = exportFileName.toUtf8();
+    auto fnameba = activeEditor->filepath.toLocal8Bit();
+    const char* fname = activeEditor->filepath.isEmpty() ? "" : fnameba;
+    exportInfo.sourceFileName = fname;
```
t-paul, line 1957: Looks like this block is identical in 3 places, so it should go into a separate function. This could be one in a anonymous namespace like in https://github.com/openscad/openscad/blob/a51718b59b5b7033753a02942c1913c5a75764d8/src/expr.cc#L48-L51

sbridger:
[nb. My actual requested requirement is 2D outlines and is essentially met. The below are things that might be needed long term if you kept developing pdf export...]

For ExportContext object the following should be considered:

It should also have the echo output available to it.
For example the part I used this to make, also requires some wire, and I calculate how much in the openscad, and show that with echo statements. In the perfect world, this information would be able to print in the pdf. The PDF export might simply print everything from echo, or just strings that begin with a keyword (e.g "PDF:" or "PRINT:")

I also see that if you export the rendered 3D view to PDF, then multipage PDF is very desirable. I normally have a min of 2 or different views (normal, cross-section/cutaway, assembled) in the openscad file, as well as many designs having multiple separate components. For documentation it would be desireable to print these to one file. (the current animation mechanism seems a fairly workable way to do this)

It is desirable to be able to pass Title string/s to the PDF, and this should ideally be done in a way that they can be set from the commandline for batch builds.

sbridger:
@tbharathchandra tbharathchandra
> [Untitled.pdf](https://github.com/openscad/openscad/files/4461842/Untitled.pdf)

Black is good. Linewidth is good .
Printers can't print at the very edge of the paper - only the  "0" of left hand scale is on the printable area of my printer. 
If your tickmarks reach 10mm in from the edge, I think you are pretty safe that they will print on all lasers I have used. My printer has a 5mm side gutter.
As I said, the purpose is to check calibration of physical printers, so as long as the tick marks are legible that can be done. (i.e. it doesn't matter so much if the numbers don't print on some printers, so numbers should probably be above tick marks)

MichaelAtOz:
> 
> No don't put date/time inside STL files. They have a date and time on the file system. 

Once the file is moved the FS datetime no longer reflects the actual creation, for example posting the file on Thingiverse/Shapeways. Knowing when that export was created, plus the filename & OpenSCAD version used, spells the environment that generated the geometry , thus allowing it to be replicated, to tie a flaw to a specific version (of source or OpenSCAD version) etc. (yes it doesn't include libraries etc)

I have SpiderOak which saves versions of files as they change, I can recover a source back to any version since it was created, but I'd need to know which one, hence filename & datetime. 

> Making files change when their contents haven't is not good for automatic build systems and source control. Similarly with OpenSCAD version. 

How am I proposing to modify it, I'm talking exported files not source, these are added at export time and not meant to change.

> I would just have to strip it all out.

Why? It doesn't affect geometry. It's like Exif metadata in photo files, the equivalent 'properties' is stored in Office documents and plenty of other applications.

nophead:
Why, because I want to know if my STL files have actually changed or not to know if they need to re-rendered for the build instructions and re-sliced and reprinted.

With a fully parametric design it is often difficult to know the knock on effects of changing something.

nophead:
Also when OpenSCAD changes I want to know if it changes my STLs.

sbridger:
@nophead @MichaelAtOz   I suggest a make new issue to discuss this. It's a great topic for an argument and I'd love to weigh in too. In a new thread.

t-paul:
I don't see much need for discussion actttually. Both use cases are totally valid. Base line target for this PR will be to add no special changeable attributes. I'll add the suggestions to #649 which would allow the user to opt-in for extra information.

I wonder if we should try making the PDF export actually reproducable - that way we could just add a reference PDF to the test framework instead of converting things to image for comparison.

tbharathchandra:
@sbridger 

> If your tickmarks reach 10mm in from the edge, I think you are pretty safe that they will print on all lasers I have used.

So are you suggesting to increase the size of the tick marks? 
I have also gone through your suggestions to further improve PDF export, I am not sure about the first one. The thing export means about geometry export. If that is really required, I think should be in other PR.

> It should also have the echo output available to it.

And the second one sounds really interesting but I think the views should be selected by the user before rotating and projecting them. If we set some default views, that may not fulfill the user requirement because we may loose some hidden details of the part which are represented by the dashed lines in the drafting sheets.

tbharathchandra:
@t-paul 

> I wonder if we should try making the PDF export actually reproducable - that way we could just add a reference PDF to the test framework instead of converting things to image for comparison.

So by this, do you mean that we can compare the resultant and reference PDFs in their binary form?

sbridger:
@tbharathchandra 
> So are you suggesting to increase the size of the tick marks?
Yes they should reach min of 10mm from paper edge. I would suggest 12mm.
You would probably be safe starting the text 8mm from edge. As I say, unprintable is 5mm on my printer and I think <=8mm on most printers I have had.
BTW scale should to be from the 0,0 PDF corner of the page, so that even if paper size is a4, it will still be on the edge of letter or ?? paper. [0,0 is bottom left](https://www.pdftron.com/documentation/web/guides/coordinates/) , so scales on bottom and left edges

> I have also gone through your suggestions to further improve PDF export, I am not sure about the first one. The thing export means about geometry export. If that is really required, I think should be in other PR.
If you mean 3D rendered view, then yes I agree it should be a separate job. 2D PDF is working.

Printing echo would likely be best with an export dialog (#649) so you can have checkboxes e.g. "print ECHO output", "linewidth" "scale" etc

t-paul:
Yes, I think regarding features, we don't want to add anything in this PR. It's a nice useful feature set already. I'll summarize the open topics into the list of the issue description...

t-paul:
@tbharathchandra, yes, if we can make sure the file is generated identically, then we can just do a checksum compare.

It looks like you might be working with an older cairo (the file lists 1.14.12). If that's the case, just let me know, I'll try adding the meta data stuff on my system. I have not checked yet, but I hope cairo provides a way to access the version at compile time.

tbharathchandra:
Yeah @t-paul , I am using cairo (1.14.12). This method can change the metadata.

`cairo_pdf_surface_set_metadata (surface, CAIRO_PDF_METADATA_TITLE, name2display);       cairo_pdf_surface_set_metadata (surface, CAIRO_PDF_METADATA_CREATOR, "OpenSCAD");`

I am able to jump to this method declaration in cairo-pdf header but I am facing undefined reference error. May be the defination is not provided in verion 1.14.12. This method is mentioned in version 1.16.0 documentation.

t-paul:
That is strange, the docs say it's only available since 1.16 so it should not be in the header files either. No problem, I'll have a look at that part.

t-paul:
Hmm, so even with cairo 1.16 it's not possible to overwrite the producer meta data, so we can't ensure producing the same PDF across different OpenSCAD installations.

t-paul:
Using `evince` it shows dummy (epoch) timestamp, I guess we have to check if other tools accept that without problems.

grafik

tbharathchandra:
Adjusted the tick marks as suggested by @sbridger and also decreased the tick marks count as they are just used for calibration. 
samples
[buttom.pdf](https://github.com/openscad/openscad/files/4481045/buttom.pdf)
[center.pdf](https://github.com/openscad/openscad/files/4481046/center.pdf)

sbridger:
> Adjusted the tick marks as suggested by @sbridger and also decreased the tick marks count as they are just used for calibration.
The tick marks do need to run the full length of the page. 
The reason is that you are measuring them by eye against a ruler . So if you can eyeball to (say) 0.4mm this is 1% of 40mm usable length but 0.14% of 280mm .
(Of course you can only measure between scales, not off the edge of the paper so 50mm scale length is only 4 intervals - 40mm)
Scales should be as long as possible.

tbharathchandra:
@sbridger 
I will increase the scale count as suggested because you want to compare it against a ruler. 

> Make usage of the scales self documenting by adding this text into the bottom scale: "Scale is to calibrate actual printed dimension. Check both X and Y. Measure between tick 0 and last tick"

I think the above will make the document bit untidy.

> The first tick should be labelled 0 (not the edge of paper), so the user is led to naturally measure ticks (and not from the paper edge)

Yeah I will do the above change.

> Label every tick, no reason not to when they are 1cm apart.

Having more numbers will increase the chances of overlapping of original geometry with these numbers. That is the only reason why I placed them in that fashion.

t-paul:
Please import the latest changes from `master`. That should fix the build issue on Travis/MacOS.

sbridger:
I would suggest trying it, rather than asserting that it will make it untidy without trying it. I am printing these out and checking them on a printer, so I will agree if I find it untidy. I rather doubt it as it is tiny grey text along the edge and is being viewed on an a4 paper (not zoomed in on a monitor).

Checking physical dimension is the _only_ real reason to have the scales. This 2D output exists for functional purposes (not documentation purposes) so appearance comes behind function anyway.
I have been using laser tooling (since 1988 actually), and have been bitten by this, but many users will be unaware of the issues, what to do, and how. They are unlikely to read the openscad manual about 2D pdf export (once it gets written) 
In this case there is a common and likely fault case: "fit to page" which can be a printer default, as well as the not exactly rare feed slip error.

It is generally a benefit to make things self-documenting, and should be a project guidline.
I just created #3285 
That particular issue is something where I have wasted >3hours over the years mucking about and asking questions. (and Torsten has wasted time answering my question)

tbharathchandra:
@sbridger 
samples 
[buttom.pdf](https://github.com/openscad/openscad/files/4496173/buttom.pdf)
[center.pdf](https://github.com/openscad/openscad/files/4496174/center.pdf)

sbridger:
> [buttom.pdf](https://github.com/openscad/openscad/files/4496173/buttom.pdf)

This is what I wanted, the text is nearly invisible, but someone attentive should see it.

> [center.pdf](https://github.com/openscad/openscad/files/4496174/center.pdf)

Not sure what this is supposed to do, since the star is offset. Was the star still supposed to be centred in the page? 
Anyway, Buttom.pdf is the way I wanted it.

sbridger:
@t-paul The [manual](https://en.wikibooks.org/w/index.php?title=OpenSCAD_User_Manual/The_OpenSCAD_Language&oldid=3674322&diff=cur&diffonly=0) seems to have no sections about export (or the commandline)

Am I missing something?
Should I just edit there to create them, or is the manual source somewhere else?

t-paul:
That's all in the [User Manual](https://en.wikibooks.org/wiki/OpenSCAD_User_Manual) part, but it's a bit unorganized for different formats.

I'm not sure what the best way is, probably one Page per Format + Import/Export (as applicable) like [SVG_Import](https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/SVG_Import) and [STL_Export](https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/STL_Export).

sbridger:
I think having export and import chapters makes sense, even though the format is the same, import and export are really totally different ideas and purposes in openscad. (compared to say file I/O in C)
I will make a page for PDF 2D export later, and try and move the DXF extrude page from export to imports

Two quick questions: 

Are $special variables what I would call full globals i.e. if you change them anyscope, then the next time they are used in any other scope they have the new value? (I want to clarify the manual)

There are now function vars/reference, but don't seem to be module vars/references (which are 10x more useful)? Is there long term reason for this? The [PEP](https://github.com/openscad/openscad/wiki/OEP3%3A-Function-Values-and-Variable-Scoping-improvements) doesn't even mention modules.
(a link to discussion would answer this. strangely google isn't giving me answers)

<whinge> Its a **huge** failing of wikibooks that they lack linear navigation i.e.  you can't browse the book from start to finish. It results in missing and hard to discover features.
 I hoped that export to pdf would be able to, but only seems to output each page.</whinge>

MichaelAtOz:
> Its a **huge** failing of wikibooks that they lack linear navigation i.e. you can't browse the book from start to finish. It results in missing and hard to discover features.
> I hoped that export to pdf would be able to, but only seems to output each page.

There is the 'printable' version, tho I think some parts may not be as up to date, as some people don't check it when they reorganise the 'paged' wiki.

wiki-printable-Capture

nophead:
No special variables are not how you describe. They have dynamic scope,
i.e.thy are accessible down the call stack, so one at the top level is
accessible anywhere, but if you change the value not at the top level you
create a new one that shadows the original. So you can't change a variable
in a module and have it affect a global one.

$fn is a good example. Setting it in a module only affects that module and
functions, module and children it calls. It won't change the top level $fn.

t-paul:
>  I will make a page for PDF 2D export later, and try and move the DXF extrude page from export to imports

That would be nice. Thanks.

> There are now function vars/reference, but don't seem to be module vars/references (which are 10x more useful)? Is there long term reason for this? The PEP doesn't even mention modules.

Discussion is in #3087 and #3088, based on the [OpenSCAD2 proposal](https://github.com/doug-moen/openscad2/).

tbharathchandra:
@t-paul 
>  Using evince it shows dummy (epoch) timestamp, I guess we have to check if other tools accept that without problems.

okular
Okular is not displaying the time stamp for the PDF generated by us.

t-paul:
Looks like ghostscript might be the best common way for pdf -> png. Seems to work fine on MacOS too using:
```
gs -dNOPAUSE -dBATCH -sDEVICE=png16m -r300 -sOutputFile="debug-%d.png" debug.pdf
```
This even generated binary identical files on Linux and MacOS, that was with identical version though (GPL Ghostscript 9.52).

t-paul:
Imported to branch [pdf-export](https://github.com/openscad/openscad/tree/pdf-export).

PR #3276, opened 2020-04-10
initial support for a mouse selecting renderer

Tomatower:
**Click on the Model Jump to the file.**

Or: Make the internal editor more powerful than any external editor.

it is started from two sides: `CSGNode`s do now carry their Index, whereby using the CSGTreeEvaluator we can get back to their parser information and therefore to Line/Column of the definition.

secondly the mouseselector defines its own shader encoding the newly created index into the color of the object.

Currently Working:

* Upon a Click in the renderer, the scene is redrawn into a FBO, and the Object ID is selected
* The object is selected based on its ID.
* The stored location is used to put the cursor in the editor to the beginning of the object defined.
* use tabmanager::open to open the referenced file and jump to the location
* display a context menu with the path through the csg tree to select the "right" node

Major Changes:

* Added support for different shaders for the rendering pass
* Added search-by-id into the root object tree
* Added "jump-to-row/col" for the Editor interface

ah - and sorry for a whole bunch of witespace fixes :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,7 @@ endif()
 cmake_minimum_required(VERSION 3.3)
 
 # CMP0071 introduced in 3.10.2
-if (CMAKE_VERSION VERSION_GREATER "3.10.1") 
+if (CMAKE_VERSION VERSION_GREATER "3.10.1")
   cmake_policy(SET CMP0071 OLD)
 endif()
 
@@ -81,7 +81,7 @@ macro(find_graphics)
     find_package(OpenGL REQUIRED QUIET)
     list(APPEND COMMON_LIBRARIES ${OPENGL_LIBRARIES})
 
-    if(MSVC) 
+    if(MSVC)
       find_package(GLEW CONFIG REQUIRED QUIET)
       message(STATUS "GLEW: Using target GLEW::GLEW")
       list(APPEND COMMON_LIBRARIES GLEW::GLEW)
@@ -156,7 +156,7 @@ if (MSVC)
 
   find_package(unofficial-fontconfig CONFIG REQUIRED)
   list(APPEND COMMON_LIBRARIES unofficial::fontconfig::fontconfig)
-  
+
   find_package(unofficial-glib CONFIG REQUIRED)
   include_directories(${GLIB2_INCLUDE_DIRS})
   list(APPEND COMMON_LIBRARIES unofficial::glib::gio unofficial::glib::glib unofficial::glib::gmodule unofficial::glib::gobject)
@@ -310,7 +310,7 @@ if(NOT HEADLESS)
   else()
     message(STATUS "DBus input driver disabled as the QtDBus module could not be found.")
   endif()
-  
+
   find_package(Qt5Gamepad QUIET)
   if (Qt5Gamepad_FOUND)
     message(STATUS "Gamepad input driver enabled")
@@ -403,62 +403,62 @@ set(CORE_SOURCES
   src/linalg.cc
   src/colormap.cc
   src/Camera.cc
-  src/handle_dep.cc 
-  src/value.cc 
-  src/calc.cc 
-  src/hash.cc 
+  src/handle_dep.cc
+  src/value.cc
+  src/calc.cc
+  src/hash.cc
   src/expr.cc
   src/degree_trig.cc
-  src/func.cc 
-  src/function.cc 
+  src/func.cc
+  src/function.cc
   src/stackcheck.h
-  src/localscope.cc 
-  src/module.cc 
-  src/FileModule.cc 
-  src/UserModule.cc 
-  src/GroupModule.cc 
-  src/AST.cc 
-  src/ModuleInstantiation.cc 
-  src/ModuleCache.cc 
+  src/localscope.cc
+  src/module.cc
+  src/FileModule.cc
+  src/UserModule.cc
+  src/GroupModule.cc
+  src/AST.cc
+  src/ModuleInstantiation.cc
+  src/ModuleCache.cc
   src/StatCache.cc
-  src/node.cc 
-  src/NodeVisitor.cc 
-  src/context.cc 
+  src/node.cc
+  src/NodeVisitor.cc
+  src/context.cc
   src/builtincontext.cc
-  src/modcontext.cc 
-  src/evalcontext.cc 
+  src/modcontext.cc
+  src/evalcontext.cc
   src/version.cc
   src/feature.cc
-  src/csgnode.cc 
-  src/CSGTreeNormalizer.cc 
-  src/Geometry.cc 
-  src/Polygon2d.cc 
-  src/csgops.cc 
-  src/transform.cc 
-  src/color.cc 
-  src/primitives.cc 
-  src/projection.cc 
-  src/cgaladv.cc 
-  src/surface.cc 
-  src/control.cc 
-  src/render.cc 
-  src/rendersettings.cc 
-  src/dxfdata.cc 
-  src/dxfdim.cc 
-  src/offset.cc 
-  src/linearextrude.cc 
-  src/rotateextrude.cc 
-  src/text.cc 
-  src/printutils.cc 
-  src/fileutils.cc 
-  src/progress.cc 
-  src/boost-utils.cc 
+  src/csgnode.cc
+  src/CSGTreeNormalizer.cc
+  src/Geometry.cc
+  src/Polygon2d.cc
+  src/csgops.cc
+  src/transform.cc
+  src/color.cc
+  src/primitives.cc
+  src/projection.cc
+  src/cgaladv.cc
+  src/surface.cc
+  src/control.cc
+  src/render.cc
+  src/rendersettings.cc
+  src/dxfdata.cc
+  src/dxfdim.cc
+  src/offset.cc
+  src/linearextrude.cc
+  src/rotateextrude.cc
+  src/text.cc
+  src/printutils.cc
+  src/fileutils.cc
+  src/progress.cc
+  src/boost-utils.cc
   src/FontCache.cc
   src/DrawingCallback.cc
   src/FreetypeRenderer.cc
   src/RenderStatistic.cc
   src/ext/lodepng/lodepng.cpp
-  src/PlatformUtils.cc 
+  src/PlatformUtils.cc
   src/libsvg/circle.cc
   src/libsvg/ellipse.cc
   src/libsvg/group.cc
@@ -477,7 +477,7 @@ set(CORE_SOURCES
   src/libsvg/util.cc
   src/clipper-utils.cc
   src/Assignment.cc
-  src/annotation.cc 
+  src/annotation.cc
   src/ext/polyclipping/clipper.cpp
   ${PLATFORM_SOURCES}
   ${FLEX_openscad_lexer_OUTPUTS}
@@ -486,7 +486,7 @@ set(CORE_SOURCES
   ${BISON_comment_parser_OUTPUTS})
 
 set(NOCGAL_SOURCES
-  src/builtin.cc 
+  src/builtin.cc
   src/import.cc
   src/import_3mf.cc
   src/import_stl.cc
@@ -507,15 +507,15 @@ set(NOCGAL_SOURCES
 
 set(CGAL_SOURCES
   ${NOCGAL_SOURCES}
-  src/CSGTreeEvaluator.cc 
-  src/CGAL_Nef_polyhedron.cc 
+  src/CSGTreeEvaluator.cc
+  src/CGAL_Nef_polyhedron.cc
   src/export_nef.cc
   src/import_nef.cc
-  src/cgalutils.cc 
-  src/cgalutils-applyops.cc 
-  src/cgalutils-project.cc 
-  src/cgalutils-tess.cc 
-  src/cgalutils-polyhedron.cc 
+  src/cgalutils.cc
+  src/cgalutils-applyops.cc
+  src/cgalutils-project.cc
+  src/cgalutils-tess.cc
+  src/cgalutils-polyhedron.cc
   src/CGALCache.cc
   src/Polygon2d-CGAL.cc
   src/svg.cc
@@ -523,9 +523,9 @@ set(CGAL_SOURCES
 
 include_directories("src/ext/libtess2/Include")
 set(COMMON_SOURCES
-  src/nodedumper.cc 
-  src/GeometryCache.cc 
-  src/clipper-utils.cc 
+  src/nodedumper.cc
+  src/GeometryCache.cc
+  src/clipper-utils.cc
   src/Tree.cc
   src/comment.cpp
   src/parameter/parameterset.cpp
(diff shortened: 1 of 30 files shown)


Tomatower:
so far most of it is working, I would appreciate feedback and change suggestions for code as well as behaviour - please test it with your more complex models that range over multiple files and libraries.

MichaelAtOz:
As I mentioned on the Forum:

I downloaded the zip file.
I noticed it was significantly SMALLER than others I downloaded.
The difference is in openscad.exe 37,138 KB v's 40,746 KB.
AVG blocked it (as it usually does for new versions) sends it off for full
checking.
Returned as clean.

So I presume the build may have missed something.

I ran it, F5, clicked on the model, crash Exception Code:	c0000005.
With W7/64.

t-paul:
I'm not seeing the file size difference, I checked only the 64-bit zips though:

Master (OpenSCAD-2020.04.16.ci4890-x86-64.zip)
41724416 Apr 16 19:13 ./openscad-2020.04.16.ci4890/openscad.exe

This PR (OpenSCAD-2020.04.16.ci4872-x86-64_PR3276.zip)
41730560 Apr 16 12:54 ./openscad-2020.04.16.ci4872/openscad.exe

The PDF export PR (OpenSCAD-2020.04.17.ci4900-x86-64_PR3266.zip)
43334144 Apr 17 17:27 ./openscad-2020.04.17.ci4900/openscad.exe

Last one is slightly bigger due to a new library added for the PDF export.

nophead:
This doesn't draw anything in the viewport on my Win7 / 64 machine.

image

Copy viewport gives an image.

image

It also gives this dialogue box on launch, which can be dismissed.

image

Tomatower:
Yep, this seems to be a side effect to the different handling between QOpenGLWidget v. QGLWidget (Windows still uses QGLWidget https://github.com/openscad/openscad/pull/1463).

The select technique uses another Framebuffer for rendering with object colors. 
It seems that this breaks the QGLWidget (which is according to Qt documentation rendering directly to the screen renderbuffer, whereby QOpenGLWidget is rendering to a dedicated Framebufferobject.)

I will experiment with defaulting to QOpenGLWidget for Windows, following the argument in the linked PR (https://github.com/openscad/openscad/pull/1463#issuecomment-160518801).

Will inform you when I got a working windows version.

MichaelAtOz:
> I noticed it was significantly SMALLER than others I downloaded.

Ahh, I followed the link in the Forum, which was **32bit**.      \:|
Via Github 64bit looks the right size...

When I run 64bits & F5, I get the same non-preview garbled window as Nophead. 
When I click in the garbled preview the editor does change to, presumably, the code involved.

A further note in case it matters, this is in a VM, and using S/W rendering, ie opengl32.dll in the openscad.exe directory.

When I ran the 32bit version I got:
> Warning: Missing OpenGL capabilities for OpenCSG - OpenCSG has been disabled.
It is highly recommended to use OpenSCAD on a system with OpenGL 2.0 or later.
Your renderer information is as follows:
GLEW version 1.12.0
GDI Generic (Microsoft Corporation)
OpenGL version 1.1.0

I seem to remember the opengl32.dll is 64bit, so that is likely the cause of the crash.

Tomatower:
The rendering issue for windows should be fixed now.
Selecting in newly created unnamed documents should be fixed as well.

nophead:
It now works to some extent on Win7. Rendering is fixed and it no longer causes a problem with my virus checker.

Some objects I click on take me to relevant code and some do nothing. I can't spot a pattern. For example in this scene:

image

Clicking on the tower, any of the printed parts in the hinge mechanism, the brown spacers or the light guide take me to relevant code. Clicking on either electronics boxes or the counter weight printed parts does nothing. 

Clicking on the M6 studding holding the counter weight and the M8 studding forming the vertical bearing works but clicking on the horizontal M6 studding in the hinge does nothing.

Clicking on the motors works but clicking on the microswitch or the pulleys doesn't. Clicking on the belt does. Clicking on the front microswitch screw and washers works but not the back ones. This is regardless of zoom level.

Often the file selected is lower level than one would want. For example clicking on the motor body brings up rounded cylinder. Clicking on the Dibond sheets brings up sheet but I would rather see where sheet was instantiated to make the base and platform. Similarly if I click on a fastener such as a screw nut or washer. Not sure how that could be deduced.

Things I don't like about the latest snapshots, not this particular PR:

The editor has random underlines in it and ctrl left click takes me to random files.

image

Here ```//! A solar track``` takes me to pulley.scad. and even ```include <NopSCADlib/core.scad``` takes me to inserts.scad. ```translate([``` takes me to geared_stepper.scad and ```if($preview)``` to ```belt.scad```. The other include and uses link to the correct files.

I don't like the multi-tabbed interface because it only seems to have one preview image. So if I click on a link and it opens another tab and draws that, when I switch back to main I have to hit F5 to render it again. Either it should have multiple frame buffers or do an auto F5 when switching tabs.

Tomatower:
thanks for the testing!

that behaviour when something is selected or not is strange.
Could you try launching openscad from a command console `cmd`?
I included some debug output there to identify what kind of object was selected and where (or none at all)

Do you have any large negative volumes/differences that are covering the elements you cannot click?
Any % or # highlighted areas?
Any `render()`?
although it should not matter, any `$preview`-conditionality renderings?

Currently the click is opening the lowest element in the tree, the `CSGLeaf` responsible for that particular mesh. @t-paul suggested to keep the PR as concise as possible, partly because small PRs are easier to merge partly because touching rendering code always requires a lot of testing.
So for this PR, this behaviour has to be considered 'expected'.

Regarding multitabbed render - does this affect the clicking in the model too?

Tomatower:
one idea regarding "cannot click" is - you must not move the mouse _at all_ between down and up to trigger the click.

nophead:
The click / cannot click behaviour is consistent, so it isn't due to moving the mouse. The simplest example is the studding where two instances work and one doesn't. This is the code:
```
module studding(d , l, center = true) { //! Draw a threaded rod with specified diameter and length
    vitamin(str("studding(", d, ", ", l,"): Threaded rod M", d, " x ", l, "mm"));

    chamfer = d / 20;
    pitch = metric_coarse_pitch(d);

    translate_z(center ? 0 : l / 2)
        if(show_threads && pitch)
            male_metric_thread(d, pitch, l, colour = rod_colour);
        else
            color(studding_colour)
                hull() {
                    cylinder(d = d, h = l - 2 * chamfer, center = true);

                    cylinder(d = d - 2 * chamfer, h = l, center = true);
                }
}
```

> Do you have any large negative volumes/differences that are covering the elements you cannot click?

No

>Any % or # highlighted areas?

I don't think so. Certainly not in the objects I have mentioned.

>Any render()?

There are rendered objects in the scene but that doesn't seem to be a factor. if they work or not. Sometimes is puts the cursor on the render().

>although it should not matter, any $preview-conditionality renderings?

Yes the whole scene is conditional on $preview.
```

if($preview)
    main_assembly();
else
    platform_dxf();
```

The debug just shows ```←[33mDid not click on object←[0m``` on the clicks that don't work.

When it opens another tab some draw the sub assembly and clicking in those is the same. For example clicking on the red hinge assembly show it on its own and the same parts respond or not to the clicks. I.e. the studding and the microswitch don't work, one of the motors, one of the microswitch screws, etc.

One thing that might be a pattern is when screws are placed with a for loop, often only one is clickable. I.e. one of one pair of motor screws, one in each triplet of bearing screws. However both of the other motors screws are clickable.

```
    translate([motor_x, motor_y, -gs_lug_t(motor)]) {
        geared_stepper(motor);

        geared_stepper_screw_positions(motor)
            vflip()
                screw_and_washer(screw, 8, true);
    }
```

This motor is clickable and so are both of its screws.

```
    translate([motor2_x, motor2_y, motor2_z])
        rotate([180, -90, 0]) {
            geared_stepper(motor);

            geared_stepper_screw_positions(motor) {
                translate_z(gs_lug_t(motor))
                    screw_and_washer(screw, 8, true);

                insert(insert);
            }
        }
```
This motor is not clickable and only one if is screws.

I spoke too soon about the virus checker problem. It didn't happen on first launch but it does on subsequent ones.

Tomatower:
- I have dug through my pile of models - and found one that has the same issues as your observe. testing can begin (first glimpse is, that somehow the id reverse search seems to be not quite right - either i rendered the wrong id or the search is broken). Thanks for spotting it!

- `for` does not seem to be the problem, `for(i=[0:10]) translate([i, 0, 0]) cube();` works

- the auto-preview seems to be behave on windows and linux - for me (linux) the preview stays the same until F5 or saving. - supposed a model click will just spawn a context menu with the call hierarchy, what would you call the "correct" behaviour?

nophead:
A call hierarchy menu would be nice. Another idea I had would be an option to pop off calls within a library (found via OPENSCADPATH). In my code clicking on a nut and opening nut.scad is not very useful and it just draws a blank screen because nut.scad is generic and doesn't instantiate an arbitrary nut. Seeing the code that placed the nut is more useful and that will be outside the library.

The annoying thing about the preview is each time it opens a new tab it it usually draws a blank screen. To go back to the original model I have to select the first tab and hit F5 to draw it again. That would be a pain in large projects that might take 20s to do an F5 with no changes. In my view it should always show an image relevant to the current tab, so a frame buffer for each one.

Tomatower:
I am thinking to use both - display the call hierarchy and add a preference to add libraries (default: off).
I agree, that the current way the selection works is not how I like it to work, but it is a necessary first step.

But this is all part of a future PR, so far i want to get this one out and working for as many people as possible.

I think you should describe your problem with the re-rendering in a separate issue, because the behaviour became apparent with the Ctrl+Click on `include/use` feature.

nophead:
I agree the multi-tab behaviour has never been correct in the way it shares the view and also the bug in what it highlights as a file. I haven't been using anything past the release because of that.

Tomatower:
Found the issue. - you could not click on elements with  `128 < index < 256` due to a floating point error I introduced by converting the index to [0:1] by dividing by 256 not 255.

nophead:
How many elements do you support? There must be thousands in that scene.

Tomatower:
The current implementation uses the R/G/B (float) channels to encode the identifier into the color of an object while rendering into a background buffer and reading the clicked pixel, recreating the generated id. 
A known good implementation is to split it into three 8-bit color components resulting in 2^24 distinguishable objects, which will work on any machine since the early/mid 2000's.
(see https://github.com/openscad/openscad/pull/3276/files#diff-50848d1954855a8fdcab791c6a7f55f1R51 line 51 and https://github.com/openscad/openscad/pull/3276/files#diff-50848d1954855a8fdcab791c6a7f55f1R192 line 192 for implementation details)

Theoretical maximum per channel is the full precise range for `float` which is 2^24. (resulting in `2^(24*3) = 2^72`) But graphics cards sometimes have not 100% precise float units, which may lead to strange problems, so i would not do more than 2^16 per channel (resulting in `2^48`), if we ever have to increase number of supported objects.

Every syntactic element within the tree has a unique `index`, so each `translate()`, `rotate()`, `if()`, `for()`, ... receives its unique id.
In real objects (in my tests) a roughly max 1/3rd of declared elements are rendered, which still leaves enough keyspace for any reasonable (and renderable!) model - in my opinion.

nophead:
Seems to be working fine now. I have a test file that instantiates one of each object my library and that works as well. 

It does tend to flood the recently used file list. Not sure what the solution to that is.

I will raise the other bugs as separate issues because I don't think they are specific to this PR.

MichaelAtOz:
> I spoke too soon about the virus checker problem. 

While not related to this PR, if 'problem' is that it says that this is a new unique file, clobbers it, sends it off for checking, then eventually says it is OK; I've had that with AVG for every new exe for some time. So that it 'expected' behaviour for a brand new program in the wild.

nophead:
No it isn't a message from the virus checker, it comes from OpenScad.exe and is complaining about a dll that is part of my virus checker. No problems with any other version of OpenSCAD but I haven't run any versions since the release before. 

I think the virus checker must somehow inject a DLL into OpenSCAD so that it can monitor it but OpenSCAD has suddenly decided it doesn't like it.

The only other time I have seen this problem is with Cubbit's desktop app, which is a brand new C++ program. I am wondering if the C++ runtime has been updated and no longer works properly on Win7.

Tomatower:
Has anyone tried it on macOS if there are any problems?
If so, I would remove the last bit of unnecessary debug output and it would be ready for merge.

ioquatix:
I can test it on Mac OS if you give me the instructions to follow.

t-paul:
We'll need to find a way to build. Unfortunately MacOS builds are very restricted, so we can't just build for every PR like for Windows and Linux.

ioquatix:
I have a developer account I can build it if you point me at the instructions.

t-paul:
@ioquatix that would be great. There's 2 ways for building:

### Using homebrew
With `brew` already installed, it should be essentially
  - ./scripts/macosx-build-homebrew.sh
  - qmake CONFIG+=debug CONFIG+=experimental
  - make

This is used on Travis and is probably the easier way for getting OpenSCAD compiled for local use.

### Using the build script
This is downloading all dependencies and building those, that can take quite some time initially
  - install some base dependencies, e.g. via MacPorts
  - ./scripts/macosx-build-dependencies.sh
  - ./scripts/release-common.sh -snapshot

A full example (with some additional stuff for incremental builds and caching so it stays in the time limits on the hosted builds) can be found in:
https://github.com/openscad/openscad/blob/f09962ec893b5178334cca71dc880ea0caa8f3f1/.circleci/config.yml#L109-L195

kintel:
Did a quick test on macOS:

OpenGL vertex shader Error:
ERROR: 0:1: '' :  version '130' is not supported


OpenGL fragment shader Error:
ERROR: 0:1: '' :  version '130' is not supported


OpenGL Program Linker Error:
ERROR: One or more attached shaders not successfully compiled


GL symbol retrieval went wrong, id is negative

OpenGL error switch FBO:
 invalid operation

OpenGL error load shader:
 invalid operation

Segmentation fault: 11

Tomatower:
Thanks for testing, i further simplified the code of the shader itself by moving the bit banging outside of the shader code.

This now rips the encoding and decoding of the id apart, but the shader should be compatible now.

t-paul:
Works fine on my ancient MacBook (2009)
```
OpenSCAD Version: 2020.05.30 (git 19147872)
System information: Mac OS X Version 10.11.6 (Build 15G22010) MacBookPro5,5 2 CPUs 4.00 GB RAM
OpenGL Version: 2.1 NVIDIA-10.0.52 310.90.10.05b46
GL Renderer: NVIDIA GeForce 9400M OpenGL Engine
```

kintel:
This might be an issue to tackle after merging: How do we test this feature?

I did a quick test, and the second example I loaded seems to return the wrong pick result.
In the following small example, any picking done in the negative green surface will indicate that we actually picked the red ring:

Screen Shot 2020-06-06 at 17 58 29



```
difference(){
  union() {
    color("Green")
      translate([25, 0, 0]) 
        cylinder(h=7, r=4);
        
    color("Red")
      translate([8,0,0])
        linear_extrude(4)
          difference() {
            circle(15);
            circle(13);
          }
  }
  translate([25, 0, 2])
    cylinder(h=8, r=3);
}
```

t-paul:
My suggestion would be to merge. it's flagged as experimental and we probably need more systems to run on. It worked on every system I tried,

t-paul:
Merge conflict was just some white spaces, I'll move the example with the wrong picking to a separate issue.

ioquatix:
Great work everyone!

PR #3280, opened 2020-04-12
Add --animate command line option.

### LINKED ISSUE #1632: Animate with dump pictures from command line
Adding option to use animation would be useful for specifying output image resolution, which would allow downsampling for better animation, as well as animation of more complicated models.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/33297105-animate-with-dump-pictures-from-command-line?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Eelis:
Fixes #1632.

Notes:
- I split up the huge `cmdline` function because for animation the second half of this function (which does the actual exporting) needs to be run once for each frame.
- If you don't specify an output filename and only use `--animate N`, the frame filenames will be `frame00000.png`, `frame00001.png`, etc, same as the filenames picked by the GUI's "Dump pictures" feature.
- If you _do_ specify an output filename, e.g. `-o bla.png --animate 20`, the frames are output as `bla00000.png`, `bla00001.png`, etc.

### DIFF
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -269,26 +269,37 @@ void set_render_color_scheme(const std::string color_scheme, const bool exit_if_
 	}
 }
 
-int cmdline(const char *deps_output_file, const std::string &filename, const char *output_file, const fs::path &original_path, const std::string &parameterFile, const std::string &setName, const ViewOptions& viewOptions, Camera camera, const std::string &export_format)
+struct CommandLine
+{
+    const char *deps_output_file;
+    const std::string &filename;
+    const char *output_file;
+    const fs::path &original_path;
+    const std::string &parameterFile;
+    const std::string &setName;
+    const ViewOptions& viewOptions;
+    Camera camera;
+    const std::string &export_format;
+    unsigned animate_frames;
+};
+
+int do_export(const CommandLine &cmd, Tree &tree, ContextHandle<BuiltinContext> &, FileFormat, FileModule *root_module);
+
+int cmdline(const CommandLine &cmd)
 {
 	Tree tree;
-	boost::filesystem::path doc(filename);
+	boost::filesystem::path doc(cmd.filename);
 	tree.setDocumentPath(doc.remove_filename().string());
-#ifdef ENABLE_CGAL
-	GeometryEvaluator geomevaluator(tree);
-#endif
 
 	ExportFileFormatOptions exportFileFormatOptions;
 	FileFormat curFormat;
 	std::string formatName;
-	std::string output_file_str = output_file;
-	const char *new_output_file = nullptr;
 
-	if(!export_format.empty()) {
-		formatName = export_format;
+	if(!cmd.export_format.empty()) {
+		formatName = cmd.export_format;
 	}
 	else {
-		auto suffix = fs::path(output_file_str).extension().generic_string();
+		auto suffix = fs::path(cmd.output_file).extension().generic_string();
 		suffix = suffix.substr(1);
 		boost::algorithm::to_lower(suffix);
 		if(exportFileFormatOptions.exportFileFormats.find(suffix) != exportFileFormatOptions.exportFileFormats.end()) {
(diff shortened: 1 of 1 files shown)


kintel:
One initial comment: The animate feature in the GUI doesn't move the camera between each frame. This feature uses the default camera settings, causing the camera to slightly jitter between each frame, probably due to `viewAll`. Perhaps we should try to avoid resetting the camera unless the user specifies camera parameters calculated from `$t`?

nophead:
I don't think setting the camera with the special variables works from the command line does it? And I don't think you can read them either.

kintel:
@nophead that's right, which may make bringing the animation feature to the cmd-line a bit more work than anticipated..

nophead:
It would be useful to me to be able to get the value of $vpd after a viewAll to get a measure of how big the object is but I think echo($vpd) would be executed too early.

t-paul (approved):
Imported to #3461, will need some updates due to merge of #3454 and #3191.

muesli:
@t-paul Exciting PR, thank you!

Could you explain how one would use the `--animate` flag in combination with the `--camera` flag, for example to achieve a 360 degree tracking shot around an object?

rocktronica:
> @t-paul Exciting PR, thank you!
> 
> Could you explain how one would use the `--animate` flag in combination with the `--camera` flag, for example to achieve a 360 degree tracking shot around an object?

It's been awhile and you may've already figured it out, but looks like `--animate` will render all frames w/ the arguments supplied and unchanged, so `--camera` will be fixed, if provided. If it's not provided, you'll get wiggling -- I'm guessing `--viewall` is implicit w/o `--camera`.

But there's another way! Leave out those flags and instead specify camera, distance, and rotation in OpenSCAD:

```
$vpt = [0,0,0]; // assumes object is centered at axes
$vpd = 281; // arbitrary but based on object size
$vpr = [0,$t * 360,0]; // rotate on Y axis
```

apc-12-48-600-600-256

(Rotating isn't technically the same as camera tracking as you suggested but is similar enough and _I think_ will look identical w/ OpenSCAD's lighting. It's also much simpler than the geometry you'd otherwise need.)

gingerbeardman:
Amazing! Excited to get this in and replace my old workflow (below). Should see a huge increase calling openscad 1x rather than 32x!

```
  for i in {0..31};
  do
    ANGLE=$(( 180-11.25*(i) ))
    
    NUM=$(( $TOTAL-(($p-1)*32+$i+1) ))
    printf -v NUMBER "%03d" $NUM

    openscad \
    -q \
    --autocenter \
    --camera 0,2,28,45,0,$ANGLE,500 \
    --imgsize 320,320 \
    -o $FOLDER/$NUMBER-$CAR-$FRAME.png \
    _$CAR.scad 2> /dev/null &
  done
```

gingerbeardman:
Actually, my old workflow is much faster. I run multiple instances of openscad in multiple threads.

**shell script "parallel" loop**
```
$ time ./render.sh forklift
./render.sh forklift  22.13s user 32.47s system 459% cpu 11.891 total
```

**--animate**
```
$ time ./render202101.sh forklift202101
./render202101.sh forklift202101  8.01s user 3.84s system 91% cpu 12.875 total
```

So it seems that in this instance the single thread app is the performance bottleneck.

gingerbeardman:
Am I correct that --animate ignores --projection?

It seems to default to Orthogonal projection in my tests.

And I can't get output respecting perspective projection.

t-paul:
Looks normal for just a cube:
`echo 'rotate($t * 360) cube(10, center = true);' | openscad --camera 0,0,0,55,0,25,100 --animate 10 --projection ortho -o ortho.png -`
ortho00000

`echo 'rotate($t * 360) cube(10, center = true);' | openscad --camera 0,0,0,55,0,25,100 --animate 10 --projection perspective -o perspective.png -`
perspective00000

t-paul:
Oi! Did I just reply to a deleted comment?

gingerbeardman:
Haha! yes, sorry I realised I was talking crap :)

Cheers for the reply though

src/openscad.cc
```diff
@@ -320,46 +331,89 @@ int cmdline(const char *deps_output_file, const std::string &filename, const cha
...
 
 	AbstractNode::resetIndexCounter();
-	absolute_root_node = root_module->instantiate(top_ctx.ctx, &root_inst, nullptr);
+
+	if (cmd.animate_frames == 0) {
+		return do_export(cmd, tree, top_ctx, curFormat, root_module);
+	}
+	else {
+		// export the requested number of animated frames
+		for (unsigned frame = 0; frame < cmd.animate_frames; ++frame) {
```
tncbbthositg, line 375: This never gets to $t = 1.0.  This works great for 360 degree animations but not so well if you want to rotate from -45 to 45 degrees.

jordanbrown0:
The precise issue is that for looping animations you want 0.0 and 1.0 to generate the same frame, and you want to omit one of those frames, but for non-looping animations (where, if you loop, you'll jump back to the start), you want 0.0 and 1.0 to yield different frames and you want to include both.

PR #3289, opened 2020-04-20
Create hy.po file

Avag-Sayan:
This is a full Armenian translation for OpenSCAD.

### DIFF
--- a/locale/hy.po
+++ b/locale/hy.po
@@ -1,0 +1,1454 @@
+msgid ""
+msgstr ""
+"Project-Id-Version: OpenSCAD 2020.02.01\n"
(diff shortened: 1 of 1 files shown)


t-paul:
Thanks!

t-paul:
While testing the file, I've noticed we did not refresh the template file for a while, sorry. I'll push that via #3325. It would be great if you could have a look once the builds are finished after merging. Please also let me know if you'd like the attribution differently in the about dialog (I've copied from the hy.po file).

https://github.com/openscad/openscad/pull/3325/commits/e1931225f4deb0eade380dbafbd2627a5eea8fd7#diff-72d61d5e17095fbd292650316b3e1334L226-L228

PR #3316, opened 2020-05-17
Persistent cache using Redis

tbharathchandra:
This PR adds a Persistent cache feature for OpenSCAD. We are going to serialize the Geometry (Polygon2d and Nef_polyhedron_3) using boost::serialization and store it as a key-value pair into a Redis database.


src/pcache.cpp
```diff
@@ -0,0 +1,260 @@
...
+    shared_ptr<const CGAL_Nef_polyhedron> N(new CGAL_Nef_polyhedron());
+    return N;
+}
+
+shared_ptr<const Geometry> PCache::getGeometry(const std::string &key){
+    PRINTDB("Get: %s", key.c_str());
+    std::string data;
+    shared_ptr<const Geometry> geom;
+    Geom_cache_entry ce;
+    if(get("GEOM-"+key, data)){
```
iliakonnov, line 160: It looks like this condition missing `&& !data.empty()`. I'm getting following error when trying to use persistent cache:
```
terminate called after throwing an instance of 'boost::archive::archive_exception'
  what():  input stream error
```
`PCache::getCGAL` [contains](https://github.com/openscad/openscad/blob/b0ceaad03a13ce089112be72acda3eac5214adec/src/pcache.cpp#L139) such check. Same applies to LCache: `LCache::getCGAL` contains check but `LCache::getGeometry` does not.

t-paul: Thanks, added to the imported branch.

t-paul:
Imported to branch https://github.com/openscad/openscad/tree/persistent-cache

orogvany:
Hm, while testing this, it appears to work well for rendering STLs, however, when running png exports, I see a lot of segfaults like

```
rashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Segmentation fault: 11
Termination Reason:    Namespace SIGNAL, Code 0xb
Terminating Process:   exc handler [51123]

VM Regions Near 0:
--> 
    __TEXT                 000000010ef42000-000000010f602000 [ 6912K] r-x/r-x SM=COW  /Applications/OpenSCAD-redis.app/Contents/MacOS/OpenSCAD

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   openscad-cache                	0x000000010f145101 GeometryCache::get(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) const + 433
1   openscad-cache                	0x000000010f13744c GeometryEvaluator::visit(State&, TextNode const&) + 204
2   openscad-cache                	0x000000010f1376cd non-virtual thunk to GeometryEvaluator::visit(State&, TextNode const&) + 13
3   openscad-cache                	0x000000010f131939 NodeVisitor::traverse(AbstractNode const&, State const&) + 217
4   openscad-cache                	0x000000010f13219d GeometryEvaluator::evaluateGeometry(AbstractNode const&, bool) + 941
...
```

I have tried a few scad files, even simple ones

something like 
`openscad-cache --cache=redis,127.0.0.1,6379 --colorscheme=Cornfield -o img.png --camera=5,7,2,0,0,-0.75 --projection=p --imgsize=1920,1080  testfile.scad`

first time succeeds, second time, segfault `Segmentation fault: 11`

Is there perhaps some exception handling or bounds checking required?

tbharathchandra:
Is this problem coming up with all second renders with PNG export or this is happening for some scripts? and seems like this came up from GeometryCache::get method, so once try the same scripts without pcache enabled.

orogvany:
I removed caching for TextNodes, and I have not seen it since, so it may just be specific to text

It works great on release version without cache, just something with cache and text nodes in "GeometryEvaluator::visit(State&, TextNode const&)" method

Though, now I see a merge conflict above, so maybe it was fixed in mainline and just needs merged in?

t-paul:
@orogvany could you check with the https://github.com/openscad/openscad/tree/persistent-cache branch mentioned above?

If it's easier, we could create an updated pull request...

orogvany:
@t-paul Yes, it exists in the persistent-cache branch.  I can check if merging master into `persistent-cache` fixes it, as it might have same conflict/changes in GeometryEvaluator.cc that may fix issue.

t-paul:
Ok, merge conflicts resolved in #3483.

orogvany:
Thanks! i rebuilt the persistent_cache branch after pull, and still get segfault with repro case

workaround is still
```
Response GeometryEvaluator::visit(State &state, const TextNode &node)
{
	if (state.isPrefix()) {
		shared_ptr<const Geometry> geom;
//		if (!isSmartCached(node)) {
			std::vector<const Geometry *> geometrylist = node.createGeometryList();
			std::vector<const Polygon2d *> polygonlist;
			for(const auto &geometry : geometrylist) {
				const Polygon2d *polygon = dynamic_cast<const Polygon2d*>(geometry);
				assert(polygon);
				polygonlist.push_back(polygon);
			}
			geom.reset(ClipperUtils::apply(polygonlist, ClipperLib::ctUnion));
//		}
//		else {
//		    //geom = GeometryCache::instance()->get(this->tree.getIdString(node));
//		}
		addToParent(state, node, geom);
		node.progress_report();
	}
	return Response::PruneTraversal;
}
```
which, admittedly doesn't actually fix root cause, just ignores whatever is passing in segfault causing value into cache.get()

I'm not sure why tree.getIdString(textNode) is not returning a valid ID for text nodes?  But, if nobody else is having the issue, certainly could be user error or something.

t-paul:
Can you please give a full step by step description to reproduce? I can't seem to get the issue it with that linear_extrude example + generating a PNG twice.

t-paul:
Oh, also lets continue in #3483 as that's the code being tested :smile: .

PR #1935, opened 2017-02-09
Opengl vbo renderer

obliviand:
A new experimental vbo based renderer for preview. ("#1518 reloaded")

TODO:
* [ ] Code review (@kintel)
* [x] Extract refactoring into separate PR (see #3323)
* [ ] Look into OS X performance issues
* [x] Go through the TODO list from #1518 and move over any applicable issues
* [x] Port to ThrowntogetherRenderer
* [x] Port to CGALRenderer
* [x] Use Feature::Experimental for new rendering code
* [x] Use unique_ptr for vertex sets
* [x] Clean up error handling for shader initialization
* [x] Implement VBO indexing
* [ ] Make render initialization time close to fixed pipeline renderer
* [ ] Add OpenGL error handling

Issue #1769, opened 2016-08-11
Provide antialias / multisampling settings for the 3d view


dronus:
While it can be enforced by some video card settings, for some cards it can't, so it would be cool if OpenSCAD's settings provide a mean to enable polygon antialiasing or multisampling in the 3d view. 

I do not have a complete picture, but it may be true that the screen space operations needed for preview display of CGAL operations do not allow for native multisampling, in this case rendering to a larger off-screen buffer could be used. This needs to be provided by OpenSCAD as it can't be enforced by video card settings.

t-paul:
There is indeed a comment that might be related at https://github.com/openscad/openscad/blob/master/src/openscad.cc#L700
```
// turn on anti-aliasing
disabled by clifford wolf: adds rendering artefacts with OpenCSG
```
[Updating the OpenGL render code](https://github.com/openscad/openscad/wiki/Project:-Improve-OpenGL-rendering) is a big topic on the wishlist, this could be reviewed in scope of that once someone is working on that.

dronus:
I thought that the "preview" mode uses depth buffer mechanisms for CSG bool operations. Most 3rd party scenegraph frameworks won't support those tricks properly I guess, so the OpenGL code could maybe not rely on this, or the preview feature would be dropped I fear.
On the other hand, I experience full CGAL computation even for preview quite often on newer versions. I will look into this.

t-paul:
CGAL render while doing preview would be caused by:
- `render()` (explicitely forces CGAL)
- `minkowski()` (not supported by OpenCSG)
- `hull()` (not supported by OpenCSG)

(That what I currently know about, we probably need to check if there are more triggers)

MichaelAtOz:
- resize()

dronus:
Yeah, that's it. My stupid. Preview works well on designs not using these features.

bebr42:
Maybe I don't understand the solution. 
But on my (Windows 7) machine the preview is quite coarse and the edges use no antialiasing at all. 
Would be nice to fix this.

nophead:
On my last computer with an Nvidia card I could turn on antialiasing for
OpenSCAD. Now I have an Intel graphics chip the only option seems to be to
disable it or leave it to the application to decide.

bebr42:
Exactly, same situation here, Lenovo Thinkpad, which delivers otherwise nice performance, even on larger external monitors resolutions.
So we want to be able to control the 3D settings from within openScad.

[...]

PR #3331, opened 2020-05-25
Simplified OpenCSG edge shader, with added smoothing.

thehans:
This change greatly simplifies the code and reduces the size of vertex attributes for the Preview edge shader, as well as incorporating antialiasing to edge lines.

Related to #1769 
and also spurred on by some recent discussion in #1935 (unfortunately it doesn't fix the spotted artifacts from that discussion though)


thehans:
Sample image before (click for unscaled fullsize):
example001_old
After:
example001

src/GLView.cc
```diff
@@ -304,6 +286,17 @@ void GLView::enable_opencsg_shaders()
 }
 #endif
 
+
+#if #DEBUG
```
kintel, line 290: -> `#ifdef DEBUG`?

src/GLView.cc
```diff
@@ -312,6 +305,11 @@ void GLView::initializeGL()
   glEnable(GL_BLEND);
   glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
 
+#if #DEBUG
```
kintel, line 308: -> `#ifdef DEBUG`?

src/GLView.cc
```diff
@@ -312,6 +305,11 @@ void GLView::initializeGL()
...
+  glEnable              ( GL_DEBUG_OUTPUT );
+  glDebugMessageCallback( MessageCallback, 0 );
```
kintel, line 310: This is an OpenGL 4 function, and will crash under OpenGL <4

thehans: Oops, I just copied an example from an OpenGL wiki without thinking much.  It was kind of helpful for me during debugging, but I didn't realize it was tied to such a high-ish version. 
And yeah I hastily put those busted if directives in there just before commiting.

Should I 1) just remove it, 2) comment it out with a note in case any dev wants to temporarily use it, or 3) try to wrap it in a proper version availability check?

kintel: Commenting is fine for now. I'm not actually sure why Qt for Mac doesn't ever select a newer OpenGL version..

thehans:
closing this PR and reopening under openscad repo branch for test result upload

PR #3332, opened 2020-05-25
Simplified OpenCSG edge shader, with added smoothing

thehans:
Copy of original PR #3331, now in openscad repo for test result upload

This change greatly simplifies the code and reduces the size of vertex attributes for the Preview edge shader, as well as incorporating antialiasing to edge lines.

Related to #1769 
and also spurred on by some recent discussion in #1935 (unfortunately it doesn't fix the spotted artifacts from that discussion though)


nophead:
I wish the normal preview faces had anti-aliased edges. I think on an older version they did if you enabled it on your graphics card but I came a across a comment in the OpenSCAD source that said it was disabled. Is it possible to bring it back without a special graphics card as I no longer use that machine?

kintel:
I experimented with AA a loong time ago, but it didn't play well with OpenCSG internals, and I didn't want to dive too deep into that codebase..

kintel:
@thehans To catch shader compatibility issues, would adding a `#version` tag to the source help making the compiler complain?

thehans:
I added `#version` in my latest code, but it doesn't seem to do much.  For example testing with `#version 110`, and then calling builtin `smoothstep` I would expect it to error out, but couldn't produce any error on my Linux machine or CI server.  Honestly not sure what the purpose of `#version` is if it doesn't check for version compatibility.  

I stripped out various parts of the shader and it appears that [fwidth](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/fwidth.xhtml) is the function causing the glitches on mac.  This function should be in since GLSL 1.1 according to that doc, so I don't know if its just Apple Software Renderer that Travis uses is bugged.  
I googled a bit, but didn't find any reference to such a known issue.
Anyways, I'm now testing the equivalent function definition of `abs(dFdx(p)) + abs(dFdy(p))` to see if that is any better supported.

thehans:
OK, I've narrowed the problem down to the dFdy function which I sort of suspected, since the missing bits are perfectly horizontal lines per triangle.

I searched specifically for dFdy issues on apple and stumbled on this note:
https://github.com/martijnberger/blender/blob/master/source/blender/gpu/intern/gpu_extensions.c#L84

> /* workaround for different calculation of dfdy factors on GPUs. Some GPUs/drivers
>    calculate dfdy in shader differently when drawing to an offscreen buffer. First
>    number is factor on screen and second is off-screen */

Not sure if that's exactly the issue I'm running into, but it may be specific to offscreen buffer, and not visible in GUI.

Currently testing if [`glHint(GL_FRAGMENT_SHADER_DERIVATIVE_HINT, GL_NICEST);`](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glHint.xhtml) improves anything.

thehans:
> Currently testing if [`glHint(GL_FRAGMENT_SHADER_DERIVATIVE_HINT, GL_NICEST);`](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glHint.xhtml) improves anything.

And of course it does absolutely nothing.

I thought the shader code that I replaced was superfluous in sending the other triangle vertices as attributes and doing a bunch of extra calculations, but its looking like that might have actually been the only way to calculate screen based line thickness if we have to support awful mac software renderer drivers that don't work.  :(

I'm tired and out of ideas at this point.

thehans:
I decided to look into this again, went back and checked the IRC logs, and remembered that t-paul could not reproduce the error on his macbook.  So this seems to confirm that its an issue specifically with the "Apple Software Renderer", and probably not an issue when GPU is available.  

```
no such artifacts on my old macbook
OpenGL Version: 2.1 NVIDIA-10.0.52 310.90.10.05b46
```
_Originally posted by @t-paul (in IRC) on 25 May 2020_

So I'm thinking of a couple options:
1) Just disable edge shader tests on mac CI (simplest solution, but slightly weakens our test suite for mac)
2) Attempt some sort of detection if GPU drivers are present, and if not then try using mesa instead of current CGL / Apple Software Renderer ?

Option 1 would only involve excluding these 2 failing tests which use the edge shader:
`openscad-viewoptions-axes-scales-edges_view-options-tests`
`openscad-viewoptions-edges_view-options-tests`

As for Option 2, I wouldn't really know where to begin. I dread developing for mac, and have 0 understanding of what is going on with their graphics stack.  So I would need some help and/or example code for 
a) detecting software vs hardware rendering and 
b) if there's any platform specific mesa initialization needed, Or could `OffscreenContextGLX` possibly be used directly?  I assume that mesa would be able to handle the shader without these artifacts as long as we can initialize it properly on mac.   

I'd really like to get this change merged so it can be integrated into the obliviand's VBO PR, where it should help reduce the GPU memory usage significantly.

@t-paul, @kintel, @obliviand any thoughts or other ideas?  Would you be okay with option 1, at least as a stopgap?

t-paul (approved):

PR #3346, opened 2020-06-02
Test cmake build on 18.04 and 20.04.

t-paul:


MichaelAtOz:
I have a vague recollection boost:thread is used in one of the mutil-thread attempts, possibly #2399, that seems dead in the water tho.

PR #2416, opened 2018-06-03
Flipped face triangulation on linear_extrude with twist (possibly fixes #901)

jameswilddev:
#784 fixed specific twist angles triangulating differently during a linear_extrude but appears to now consistently triangulate the quadrilaterals with the split along the long axis, creating quite unsightly geometry (see #901 for examples).

This pull request attempts to flip the edge, similar to "Rotate Selected Edge" in Blender.

I am unable to build this to test it however, as I am currently on metered internet and do not have the remaining usage to download a Linux image or Visual Studio.

Issue #2259, opened 2018-01-08
linear_extrude with scale vector renders differently to (correct) preview


speendo:
The preview of the following code looks like expected. The rendered model seems to be broken:

    difference() {
        translate([-30,0,0]) {
            linear_extrude(height = 10,center = false, twist = 0, scale = [1,0], slices = 40) {
                polygon(
                    points=[[30, 0],[0, 10],[0, -10]],
                    paths=[[0,1,2]]
                );
            }
        }
        translate([0,0,-1]) {
            cylinder(h=12,r=1);
        }
    }

I'm quite confident that this already worked before, because I developed this code with some substantial help from the mailing list in 2014 (http://forum.openscad.org/curved-surface-td8757.html) and published a model on thingiverse that uses a similar code (https://www.thingiverse.com/thing:370976).

By the way: if you comment out the `cylinder` it seems to render correctly...

MichaelAtOz:
Confirmed with OpenSCAD 2015.03-2 & 2018.01.06 (git 3473eb1) on Windows 7/64.
(you want a convexity to have F5 look good, but that is not an Issue)

kintel:
The problem is that we make certain assumptions on the convexity of objects to optimize processing.
In this case, the question is whether this object is convex or not:
`linear_extrude(height = 10, scale = [1,0], slices = 40) polygon([[30, 0],[0, 10],[0, -10]]);`

..or more generally, is there a way we can tell if a linear_extrude object is convex without evaluating it?

kintel:
Changing the line at https://github.com/openscad/openscad/blob/master/src/GeometryEvaluator.cc#L689
to `PolySet *ps = new PolySet(3, unknown);` fixes the problem, but it would be nice to catch more known-to-be-convex cases.

kintel:
I think this is accurate:
* Twist always yields concave objects
* Scale with non-square aspect ratio yields concave objects if any vertex angle is not 90 degrees

Issue #3336, opened 2020-05-28
Reduce number of faces


espr14:
There is only one operator that reduces number of faces which is hull(). But it works on convex objects only. It would be nice to have something like:

`module reduce(min_face_size mfs, max_deviation md);
`
This would remove (merge with neighbor) all faces smaller than `mfs` but only in case they don't cause deviation of surface more than `md` at any place.

This would be handy after use of more operators (union, difference etc.) which can lead to too many small faces that are not necessary. It's also great to do it before export to reduce file size.
For example I do export, load it in Meshmixer, reduce faces to half and the model looks the same.

thehans:
Personally I don't see much purpose for such a feature.  If you need simpler model, it would make more sense to adjust $fn, or $fs/$fa to not create so many faces in the first place.

I mean, maybe its useful to simplify an imported STL that wasn't created in OpenSCAD, but I feel like its too rare of a case to bother implementing when other free tools can also do the job.
Do you have any specific example case besides importing, where adjusting $fn etc is not a feasible solution?

espr14:
STL are not a problem - I use external tools.
I use $fn, $fs, $fa but they do not apply to union, difference, intersection and scale. These operators will create small faces no matter what. One reason is floating point operations that are not absolutely precise. Unaligned objects can multiply number of faces.
Small faces propagate through the hierarchy generating more and more faces. Since these operators have high time complexity (like O(n^3) or more) every face counts a lot.

nophead:
Can you give an example? I only get faces in my design where I specify them. Scale should not create any new ones. Union, difference and intersection only create the expected ones for me.

espr14:
```
$fn=32;
scale([10,1,1]) hull() sphere();
```
Scale does not create new ones but due to scale much higher $fn is needed. Scale preserves even the smallest faces so it is a mix of big and small ones. Here the reduce() would be helpful.

espr14:
```
render()
for(b=[0,1])
translate([b*.99,b*0.01,b*0.01])
cube();
```
This should be simple cube but due to (simulated) float arithmetic it will generate much more triangles.

nophead:
I don't understand why you think the result should be a cube. It is exactly
as you coded it. Two overlapping cubes with the second one slightly offset.
Nothing to do with floating point errors.

dpellegr:
There would be a pretty strong use case for this feature now that, thanks to the power of the manifold engine, we can easily perform boolean operations with thousands of objects.
[...]

Issue #3210, opened 2020-02-06
Example020 spring produces inverted faces


MichaelAtOz:
![Capture-example020-faces](https://user-images.githubusercontent.com/1972961/73907467-80506e00-48fa-11ea-8a2c-8bcc907f72c4.JPG)
Also showed the same in exported STL with Netfabb.
OpenSCAD 2015.03, 2020.01.04.nightly on Debian; 2019.05, 2020.01.30 on W7.

t-paul:
Looks like #901.

PR #3351, opened 2020-06-07
Fixes for linear_extrude with scale and/or twist.

thehans:
Triangulates across the shortest diagonal, except in cases of any 0-scaled axis.
For those it uses the longer diagonal on the uppermost slice which helps avoid
creation of zero thickness faces.

First I should mention that I am aware of PR #2416, and I originally started this journey with the intention of getting that merged.  But I quickly found that it doesn't solve the main issue, only shifts it around a bit.  To illustrate, I'm using `example020.scad`, with a modification where the large extruded `spring()` module is called, in order to get a good look at its cross section:
```
intersection() {
  spring();
  cube([120,1,120]);
}
```
Previewed with F5, using Front view (Ctrl-8) and zoomed into a single section for the screenshots:

Old Behavior (current master, nightly build):
example020_nightly

PR #2416 (chooses opposite diagonal)
example020_PR2416

This PR:
example020_new

So it can be seen that simply reversing which diagonal is chosen, while it might appear better for some or even the majority of models, its not the proper solution.

The main change of this PR is to calculate the length of diagonals for each quad, and use the shorter of the two.  That smooths out a lot of linear_extrudes with twist as shown above, but I found that with that change alone, it would fail in cases where one component of scale was 0, by generating some zero thickness faces or "ears".  (e.g. issue #2259 )

To solve this I noticed that choosing the **longer** diagonal actually seems to avoid this type of error.  So for this corner case there is a check which reverses its diagonal decision:
 - only if either scale component is 0
 - only on the final, topmost slice

Here is an example test I was using (I'll probably add it to test cases after writing up this PR description, maybe just alter the `issue2259.scad` test to include more cases)
```
module tri() { polygon([[20, 0],[0, 10],[0, -10]]); }

translate([-22,-15]) linear_extrude(height = 10, scale = [0,1], slices = 2) tri();
translate([  0,-15]) linear_extrude(height = 10, scale = [0,0], slices = 2) tri();
translate([ 22,-15]) linear_extrude(height = 10, scale = [1,0], slices = 2) tri();

translate([-22, 15]) linear_extrude(height = 10, scale = [0,1], slices = 1) tri();
translate([  0, 15]) linear_extrude(height = 10, scale = [0,0], slices = 1) tri();
translate([ 22, 15]) linear_extrude(height = 10, scale = [1,0], slices = 1) tri();
```

Thrown Together Preview from Nightly (note that it triangulates asymmetrically for these type of tests)
linear_extrude_scale

Thrown Together from this PR (consistently symmetrical, no zero thickness ears)
linear_extrude_scale_new


rcolyer:
The following code testing generation of a screw thread does not appear to go across the shortest diagonal.  Did I misunderstand the goal, or does it miss somehow in this case?

```
linear_extrude(height=10, twist=5*360)
  translate([1, 0, 0])
   circle(3);
```

screwtest1

thehans:
@rcolyer It chooses the shortest diagonal *per quad*.  Each quad being composed of two adjacent vertices from an outline, and their corresponding pair one slice up.  If the twist is large enough relative to the number of slices, then it will still look a bit wonky, but at least seems to avoid self intersecting polyhedron for all but cases where twist is at or very near 180 degrees per slice.

I guess you were expecting a possible further optimization as mentioned in #901, of finding the closest diagonal across slices by shifting though the list of vertices, but I did not attempt to go that far.  Not sure but I feel like trying that would cause even more hard to fix weird corner cases.

rcolyer:
@thehans Agreed.  That sounds like a separate issue.  Perhaps something more suitable for a reduce operation like #3336, so that calling for it would be explicit.

thehans:
I also encountered an issue with some of additional tests which I planned to add for #2259.

These two objects result in meshes that appear ok at first glance, but it ends up with edges meeting at a T-junction due to the vertex lying on the X axis scaling to [0,0], and the choice of diagonals for neighboring quads:
```
module tri() { polygon([[20, 0],[0, 10],[0, -10]]); }
translate([-25,-20]) linear_extrude(height = 10, scale = [0,1], slices = 2) tri();
translate([-25, 20]) linear_extrude(height = 10, scale = [0,1], slices = 1) tri();
```
 - These objects give non-manifold warning from CGAL, which I believe are due to those T-junctions
 - As a result, they fail `monotonepngtest` and `stlcgalpngtest`, which I think has to do with export/import failing due to the non-manifoldness. Its somewhat strange that the "expected" image from `TEST_GENERATE=1 ctest` is blank, while the "actual" images look correct, but maybe that's just how the test flow goes.
 - I mentioned in my original PR comment that I was trying to deal with single component 0-scale by choosing the longer diagonal for the topmost slice.  This was just something I noticed from testing that seemed to produce ok results, but now I'm thinking that's not sufficient.

So I'm looking for any other ideas to solve these sort of problems.  So far I think that copying an idea from this [comment on #901](https://github.com/openscad/openscad/issues/901#issuecomment-291585016) might be a workable solution:
> Split on both diagonals, generating 4 triangles from the nonplanar quad, rather than 2

And as with my previous attempt, this would be an alternate construction that's only used on the top slice when a single component is scaled to 0.
Does that sound like a good solution?

thehans:
I also wanted to mention that I've done some searches for issues containing "linear_extrude" and tried testing them on this PR.
I found that at least these four are fixed in this PR:
 - #2510 Pineapple looking thing, plus 2nd example of thin helix intersected with sphere
 - #3117 A vertical wind turbine with twisted linear_extruded blades
 - #3209 another thin helix, intersected with linear_extruded cylinder
 - #3210 example020 has back faces showing in Thrown Together view

thehans:
@rcolyer wrote:
> The following code testing generation of a screw thread does not appear to go across the shortest diagonal. Did I misunderstand the goal, or does it miss somehow in this case?
> 
> ```
> linear_extrude(height=10, twist=5*360)
>   translate([1, 0, 0])
>   circle(3);
> ```

I tried your code again after implementing the helical arc length calculation I was talking about, and I think its working. :smile:
inphase_test_helix_fs

Much nicer looking slices are chosen on default settings now, and the largest edges look to be around 2 units long as would be expected. (`$fs=2` by default)

MichaelAtOz:
I haven't played with this yet.
Many of the `linear_extrude()` issues are due to, or exacerbated by, the default `layers=` being small. Do your fixes resolve this, or should the default be increased?

thehans:
@MichaelAtOz Yes, absolutely it fixes what was previously considered using too few slices. 
Now that the shorter diagonal is chosen, it turns out that you can twist most shapes by nearly a full 180 degrees *per slice*, without any self intersections.

Take issue #2510, where the 2nd comment gives concise a simplified example:
```
intersection() {
linear_extrude(150, twist = 360)
square([0.5, 120], center = true);
sphere(50);
}
```

Plug that into the nightly build, preview Thrown Together mode, and you'll see this:
2510_2_old

The extrusion Its almost 100% backwards faces!

Compare that to a build on this PR branch:
2510_2_new

Not a single self intersection(yes this screenshot is still in Thrown Together mode!), and F6 render also works with no warning or error

MichaelAtOz:
@thehans would you expect this to affect `rotate_extrude()` or `minkowski()` or maybe `difference()`, a slight change in triangles
From [forum](http://forum.openscad.org/rendering-problem-td29209.html), one of my recent-files I just happened to try.

2020.05.06 v's 2020.06.10
Capture2020 05 06 v 06 10 B

In thrown-together that location, 05.06 shows z-fighting, 06.10 shows solid reverse face.

thehans:
The main fix here is choosing the shorter diagonal for a non-planar quad face, and the changes are specific to `linear_extrude`'s  code so no it won't affecf `rotate_extrude`, and will only possibly affect `minkowski` or `difference` if they are operating on a `linear_extude`d object that uses `twist` and/or `scale` params.

That said, the concept of triangulating a quad across its shortest diagonal might be applicable in some other places in code, but I haven't touched or looked at those as part of this PR.

nophead:
Not sure if this is relevant but when triangulating quads for sweep I think we concluded the diagonal should be chosen to make the kink in the quad match the local curvature type, i.e. convex or concave. If you put a concave quad on a convex surface or vice versa the shading looks wrong.

thehans:
@nophead I'm not sure either.  
The shortest diagonal approach appears to be the best looking option as far as I can tell.
How would you define curvature in the case of linear_extrusion?  Along what axis? its possible for a single point on a surface to have positive curvature along one axis and negative on another, a saddle point .  And then how do you correlate that with choosing a diagonal exactly?

For all I know it may even be true that the shortest diagonal and your description of matching convexity/curvature are geometrically equivalent anyways.

nophead:
I think it would be the curvature along the direction that is most pronounced. For example a helical spring would be considered convex even around the inside where it is saddle like in that inside of the coil is concave but the wire it is made of is more convex.

Having said that I think I just picked the shortest side and looks right for anything I sweep at the resolution I use. The quad patches look like rectangles when the edges are hidden because the two triangles are at such similar angles they get shaded the same.

When it is wrong the triangles alternate in brightness instead of monotonically getting brighter or darker but it is only noticeable when there is significant curvature across the quad.

Probably not worth worrying about unless it looks wrong like the old sweep does when twisted on one direction but not in the other.

thehans:
Going back to this:
> I could maybe use some help with math in the case of scale + twist though.
> If scale is uniform, then I _think_ I can just use `effective_r = (max_r + max_r * scale)/2` ?
> But non-uniform scaling+twist seems like a whole other can of worms and I'm not sure how to approach that. I think the `max_r` at bottom could be a different vertex altogether from the `max_r` at top, then what?

I'm hoping for some more discussion, ideas, opinions of how to handle this.
Its about calculating an appropriate number of slices based on `$fn`,`$fs`,`$fa`, when `slices` parameter is not specified by the user.

Let's break it down into the possible cases:
 - No scale, No twist: 
   **OK** basic `linear_extrude`, `slices=1`

 - Uniform scaling `scale.x == scale.y`, No Twist:
   **OK** Faces remain planar, so `slices=1`

 - Non-uniform scaling `scale.x != scale.y`, No twist: 
   **???** Currently this situation still defaults to 1 slice.  And while the vertices follow linear paths, the edges between them can form curved surfaces if given enough slices.  Eg:
   `linear_extrude(60, scale=[2,0], /*slices=20*/) circle(20, $fn=4);`
   ind_scale_1slice
   ind_scale_20slice
   a) Pretty sure we can ignore `$fa` here, although there's the possibility we could calculate the edge direction before/after scaling, and the use the max angle swept any edge (all edges "rotate" by 45 degrees in example above, despite 0 twist). Don't know if its worth bothering, but wouldn't be too difficult to implement I don't think.
   b) `$fn` could either be ignored or maybe just assign `slices = $fn` directly, not sure.
   c) `$fs`i think would make most sense to divide the length of the longest diagonal formed by any vertex from bottom to top.  Would need to loop over all vertices in 2D shape and determine the max delta before/after scaling.
   d) Or just ignore all of them and leave it at `slices=1`

 - No scale, With Twist:
    **Improved!** Vertices follow a helical arc (straight cylindrical).  I've implemented calculating this arc length for the most distant 2D point from origin.  So `slices = arc_length / $fs`, when `$fs` is the limiting factor.
   `$fn` and `$fa` still function same as before.  `$fn` is slices per turn, and `$fa` is degrees per slice.
   This greatly helped with the "default" look of objects when slices is not given.

 - Scaling (uniform or not), With Twist
   ***???*** There's no special handling for this yet, it just uses the same calculations as the above twist without scaling.  For uniform scaling, the vertices follow a helical arc on the surface of a cone (truncated, unless [0,0] scale) .  Otherwise its a funky smushed cone that's a circle on bottom and an ellipse on top. (see example code below)
   I'm not sure if its worth trying to do any further calculation for either of these, and I don't have the math chops to solve such arc length equations unless someone can provide them.

I did play around with a defining a parametric equation and visualizing the path of a scaled+twisted vertex, for anyone interested:
```
r = 5;
h = 10;
twist = 720;
s = [0.4,2];
T = twist*PI/180;
steps = 48;
zstep = h/steps;
// plot points from our function definition
color([0,1,0]) for(i=[0:1:steps]) translate(f(i/(steps))) cube(zstep,center=true);
// Visualize the contour that a single vertex would follow.
color("blue") for(i=[0:1:steps])
  translate([0,0,i*zstep])
    scale(lerp([1,1],s,i/steps)) cylinder(h=zstep,r=r,$fn=steps);
// verify function against linear extrude of a small "point" square
color("red") linear_extrude(height=h, twist=twist, scale=s, slices=steps) 
  translate([r,-0.05]) square([2*zstep,0.1]);
function lerp(a,b,t)=(1-t)*a+t*b;
function f(t) = let(st=lerp([1,1],s,t)) [st[0]*r*cos(-twist*t), st[1]*r*sin(-twist*t), t*h];
```

I also had an idea of making the slice height vary along Z when scaling is involved.  I feel like this could help produce better looking models, but not yet sure how that could be implemented.

nophead:
When I extrude a shape like that I add extra vertices along along the edges so the quads are roughly square. That gets rid of the ugly stripes that give a rippled appearance. It would be nice if linear extrude did that when necessary.

thehans:
@nophead I've been thinking about that issue too.  
Particularly when long, thin rectangle is extruded with twist, it ends up looking like a staircase no matter how many slices, because the long edge just sweeps too much for its length.

A small example:
```
translate([-20,0]) linear_extrude(50, twist=360,slices=20) translate([0,-.1]) square([10,0.2]);
translate([  0,0]) linear_extrude(50, twist=360, slices=20) polygon([for(x=[0:10]) [x,-0.1], for(x=[10:-1:0]) [x,0.1] ]);
translate([ 20,0]) for(x=[0:10]) linear_extrude(50, twist=360, slices=20) translate([x,0]) square([1,0.2]);
```
First line illustrates the problem.
Second line was an attempted workaround, but polygon seems to automatically merge colinear vertices, so that is just the same model as the first.
Third line works, but then you gotta waste time in CGAL to union all the separate extrusions.

So yeah it would be best if linear_extrude could automatically split long edges into shorter segments.  There's a few different ways this could be done... 

First I would think `$fn` could count as a minimum # of vertices *per outline*.  So any outline with less than `$fn` vertices would have extras inserted, but then there's the question of how to divide up  new vertices among edges.  Thinking as I write this... you could basically track a list of `<edge_length, segment_count>` pairs.  Each edge would start at `segment_count=1`, and whichever `edge_length/segment_count` is largest gets its count incremented until `$fn is reached.  
Then go through and update the polygon once you know where and how many splits to make.
That should make things about as even as possible I think.

For `$fs`, the code could split each edge into segments as `segments = round(edge_length / $fs)` (or maybe ceil instead of round?). Definitely the most straightforward.

I don't think `$fa` would be applicable for segmenting a 2D polygon though.  I suppose you could cast rays out from the "center of mass" at specific angles and spit the first edge each ray hits, but that would only work for convex shapes, and probably would result in an ugly mess most of the time.

The other question is should `linear_extrude` always do this?  If user has `$fn` or `$fs` set should it just split according to those (with `$fn` overriding `$fs` as usual), or should it require a new parameter flag, like `split=true` to enable that feature?
When `slices` is given, then linear_extrude uses that without the need for the special variables in splitting up the height.  In that case should it also use `slices` as a number of segments per outline?
Or should it have additional, completely independent parameter(s) like `max_edge_length=[num]` or `min_outline_segments=[num]`?

nophead:
To get around the merging of colinear vertices I perturbed the slightly when modelling a fan blade with this:

```
    module squarish(s, n) {
        polygon([
            for(i = [0 : n]) [i * s.x / n,  s.y + (i % 2) * eps],
            for(i = [0 : n]) [s.x - i * s.x / n, (i % 2) * eps],
        ]);
    }
```

image

I think the aim is to make the quads square, so dividing the helix length made from the height and the twist by the number of slices gives the quad height and then the perimeter is divided into segments roughly the same length.

It is probably only needed when there is a twist. I think it should be optional for backwards compatibility. What would happen to my fan blades that are already divided up roughly to the same rule? I wouldn't want extra vertices to appear. 

Making it configurable seems more logical, perhaps ```segments = [n]``` to match ```slices = [n]```' but it requires a little more work from the user assuming roughly square quads is always optimum. 

I personally don't think it should use $fs because I have it  and $fa globally set for 3D printing curves. I think it would in general be too fine for linear_extrude unless $fa was also used in the calculation. I.e. calculate the perimeter length and treat it as a circle circumference to apply $fn, $fa and $fs but that seems a bit hacky.

thehans:
@nophead thanks for the input.  I decided its probably best to limit scope on this PR and save "segmentation" options for another day.  The discussion would be good to reference in a new issue though (if there isn't already a similar feature request yet).

I've been trying to get to a good stopping point, and I think I've finally made it.

 - I cleaned up the logic around choosing diagonals when they are within error of being equal.
 - I asked around a little bit regarding finding the helical curve length with scaling and twist.  For the uniformly scaled case, it was pointed out to me that the XY view is just an archimedes spiral, which has known equations for length.  And that can be combined with the height using Pythagorean theorem to get the total length.  So I've created a function to do just that.  
 - For the non-uniform case, I'm told its unlikely to even have a closed form solution, so I've left it using the same calculation as if there were twist with no scaling, as a "close enough" approximation.
 - I also implemented the `$fs` slices calculation for non-uniform scaling, without twist.  Basically same as I proposed in my previous comment.
 
Lastly, I had noted before that equal length diagonals occur when edges touch the origin (when considering twisting and uniform or no scale).  I realized later that the edge actually doesn't need to touch origin, just be co-linear with it.  So in polar coordinates, any two points that have the same `theta`, regardless of `r` define such an edge.
And I had been experimenting with splitting quads into 4 triangles, with an added midpoint.  I found that this approach looks really nice when twisting those types of edges, but looks way worse in other cases.  I mean, I guess it makes sense to "split the difference" when diagonals are equal.

So there's some code that that will use 4-way splitting when it detects equal length diagonals, but I've left it disabled with a compile-time conditional [here](https://github.com/thehans/openscad/blob/29fd948d5d4df6a9ee60c9efd4be6d8d30f0ce54/src/GeometryEvaluator.cc#L818), because I feel like it might be viewed as inconsistent.  For example when twist extruding a not-centered square, two edges would get the nice surface finish and the others won't, etc.  Also it changes up a couple of the test results significantly, and may have higher chance of self-intersection with nearby non-4-way split faces, for cases with extreme twist per slice.
I just figured I'd mention it in case anyone else felt like experimenting with it, but you'll need to build it from source with that line changed.

Here's a small script that shows it off compared to regular diagonal split:
```
function arc_seg_f(r1=0,r2,a) = let(ca=cos(a/2),sa=sin(a/2)) r1==0 ?
  [[0,0], r2*[ca,-sa], r2*[ca,sa]] : [r1*[ca,-sa], r2*[ca,-sa], r2*[ca,sa], r1*[ca,sa]];
module arc_seg(r1=0,r2,a) polygon(arc_seg_f(r1,r2,a));
linear_extrude(20, twist=360, slices=20) arc_seg(3,10,a=10);
```
2way
4way

Anyways I'm trying to wrap up this PR and I'm done making all the changes I had in mind.
So unless anyone has issues with any of the changes, I'm hoping we can get it merged soon.

RonaldoCMP:
`The main change of this PR is to calculate the length of diagonals for each quad, and use the shorter of the two. That smooths out a lot of linear_extrudes with twist as shown above, but I found that with that change alone, it would fail in cases where one component of scale was 0, by generating some zero thickness faces or "ears". (e.g. issue #2259 )`

I don't know why but linear_extrude is virtually the last place where you may use a scale 0 without a warning. Is there any good reason for that? If there is none, why bother?  Forbid it.

thehans:
Trying to think of a reason for it, I was about to say that its for making pyramids, but `cylinder()` with `r2=0` also works fine for that).  I guess the main reason to keep it is so you can make twisted pyramids then :P

Another issue that zero scale can cause is when extruding anything with a hole in.  This will guarantee a non-manifold result, because the hole meets the top surface at the same point, giving a 0 thickness spot.  
Example (F6 render gives warning):
```
size = 10;
render() {
linear_extrude(size, scale=0) difference() { square(size,center=true); square(0.9*size, center=true); } 
cube();
}
```
So, I guess I wouldn't be opposed to enforcing non-zero scale.  It would simplify the code a bit to not have to worry about it.
I think it would be best to give a warning, automatically override the value, and carry on, much like setting `$fs < 0.01` does.

That raises the question of what the lower limit should be.  If the point is to avoid non-manifold results, then just `scale > 0` is not enough.  Because of OpenSCAD's Grid snapping, vertices that scale to something very close will get merged anyways.

In the example above, keeping `size=10`, I found that `scale=1e-7` reports non-manifold, but `scale=1e-6` doesn't.  Using other values for `size` would proportionally change this cutoff point between valid/invalid.
Based on that, I'd say `scale >= 1e-4` should probably cover most of our bases while leaving some wiggle room for other types of shapes that differ from that one example.  And for typical scale of 3D-printed objects, that should still be plenty smaller than minimum printable feature size.

MichaelAtOz:
> This will guarantee a non-manifold result, because the hole meets the top surface at the same point, giving a 0 thickness spot.
> Example (F6 render gives warning):

Interestingly, with this PR (2020.06.10) it doesn't. Exported STL is OK to Netfabb.

We should be careful with artificial limits that are significantly greater than the grid size. Some people work with very small numbers.

thehans:
@MichaelAtOz I may have caused some confusion by editing my last comment above, shortly after I wrote it.  

I had been testing without any call to the `render()` module, which the edited comment now reflects, and it does produce error for me in that case.

I hastily added `render()` in there when I first wrote it, thinking that would guarantee the error in either F5 or F6 mode, but what actually happens is even weirder...  
For some reason it passes manifold testing and exports "without error", but if you view with F12, you will see that the entire object becomes inside-out (and the resulting STL export too)!

So its still a problem, but maybe better left for a separate Pull Request to fix it, if changing that would be controversial or necessitating more discussion?

PR #3366, opened 2020-07-03
Issue432

robertlugg:
Added a check to confirm that the directory of the output file exists.  If not that output is skipped.  I'm happy to adjust behavior but this seems close to what the requestor asked for.

### DIFF
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -309,13 +309,15 @@ int cmdline(const char *deps_output_file, const std::string &filename, const std
 	std::string filename_str = fs::path(output_file_str).generic_string();
 	new_output_file = filename_str.c_str();
 
-	// Do some minimal checking of output before rendering (issue #432)
+	// Do some minimal checking of output directory before rendering (issue #432)
 	fs::path output_directory = fs::path(output_file_str).parent_path();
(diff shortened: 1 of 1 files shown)


robertlugg:
I’d be happy to implement others. but I struggled to see how I might do them. I could “touch” a file in the directory to check for write access. Maybe these checks belong after it fails so that they don’t waste time in the normal case??

t-paul:
Boost filesystem should work with the `permissions()` call looking at this example: https://stackoverflow.com/questions/9776050/how-to-get-file-permissions-with-c-boost-library - At least on Linux that seems to work nicely.

robertlugg:
Ok. I’ll experiment a bit more.  I don’t know how to see what group applies to me. But there must be a way. So am I in the owner, group. Or world?  The touch method will Tell through experimentation.

t-paul:
Right, I don't see boost handling that. The low-level way on Linux is `access()` (https://www.gnu.org/software/libc/manual/html_node/Testing-File-Access.html) which is not directly available on Windows, but there seems to be `_access()`,

robertlugg:
Would you take a look at the pull request I just put in?

PR #3369, opened 2020-07-09
Number scroll by mouse

Sharma-Hrishabh:
This feature-PR will allow the user to scroll the numbers in the text-editor by scrolling the mouse wheel.

Done:
- [x] Change the value by mouse wheel scroll.
- [x] Can change the step size from Editor Preferences.
- [x] Can change the modifier from Editor Preferences.
- [x] Can disable the number scroll via mouse wheel from Editor Preferences.
- [x] Handle the cases, so that the numbers present in variable names or some other identifiers are not changed.
<hr>

Todo:
- [ ] Disabling Mouse Movements while changing numbers  -> Abrupt Jump of caret on a single-digit number, when a new digit is added at the front.[Right now, not finding any suitable method/API in Scintilla to override the default behaviour]
- [ ] Delay Preview Until Done

### DIFF
--- a/src/Preferences.cc
+++ b/src/Preferences.cc
@@ -205,7 +205,7 @@ void Preferences::init() {
 	initComboBox(this->comboBoxLineWrapVisualizationEnd, Settings::Settings::lineWrapVisualizationEnd);
 	initComboBox(this->comboBoxLineWrapVisualizationStart, Settings::Settings::lineWrapVisualizationBegin);
 	initComboBox(this->comboBoxShowWhitespace, Settings::Settings::showWhitespace);
-	initComboBox(this->comboBoxTabKeyFunction, Settings::Settings::tabKeyFunction);
+	initComboBox(this->comboBoxModifierNumberScrollWheel, Settings::Settings::modifierNumberScrollWheel);
 	initSpinBoxRange(this->spinBoxIndentationWidth, Settings::Settings::indentationWidth);
 	initSpinBoxRange(this->spinBoxLineWrapIndentationIndent, Settings::Settings::lineWrapIndentation);
 	initSpinBoxRange(this->spinBoxShowWhitespaceSize, Settings::Settings::showWhitespaceSize);
@@ -651,6 +651,11 @@ void Preferences::on_lineEditStepSize_textChanged(const QString &text)
 	emit stepSizeChanged(text.toInt());
 }
 
+void Preferences::on_comboBoxModifierNumberScrollWheel_activated(int val)
+{
+	applyComboBox(comboBoxModifierNumberScrollWheel, val, Settings::Settings::modifierNumberScrollWheel);
+}
+
 void Preferences::on_enableHardwarningsCheckBox_toggled(bool state)
 {
 	QSettingsCached settings;
@@ -917,6 +922,7 @@ void Preferences::updateGUI()
 	updateComboBox(this->comboBoxShowWhitespace, Settings::Settings::showWhitespace);
 	updateComboBox(this->comboBoxIndentUsing, Settings::Settings::indentStyle);
 	updateComboBox(this->comboBoxTabKeyFunction, Settings::Settings::tabKeyFunction);
+	updateComboBox(this->comboBoxModifierNumberScrollWheel, Settings::Settings::modifierNumberScrollWheel);
 	initSpinBoxDouble(this->spinBoxIndentationWidth, Settings::Settings::indentationWidth);
 	initSpinBoxDouble(this->spinBoxTabWidth, Settings::Settings::tabWidth);
 	initSpinBoxDouble(this->spinBoxLineWrapIndentationIndent, Settings::Settings::lineWrapIndentation);
--- a/src/Preferences.cc
+++ b/src/Preferences.cc
@@ -613,6 +613,13 @@ void Preferences::on_checkBoxEnableLineNumbers_toggled(bool checked)
 	writeSettings();
 }
 
+void Preferences::on_checkBoxEnableNumberScrollWheel_toggled(bool val)
+{
+	Settings::Settings::inst()->set(Settings::Settings::enableNumberScrollWheel, Value(val));
+	comboBoxModifierNumberScrollWheel->setDisabled(!checkBoxEnableNumberScrollWheel->isChecked());
+	writeSettings();
+}
+
 void Preferences::on_enableSoundOnRenderCompleteCheckBox_toggled(bool state)
 {
 	QSettingsCached settings;
@@ -931,6 +938,7 @@ void Preferences::updateGUI()
 	initCheckBox(this->checkBoxBackspaceUnindents, Settings::Settings::backspaceUnindents);
 	initCheckBox(this->checkBoxHighlightCurrentLine, Settings::Settings::highlightCurrentLine);
 	initCheckBox(this->checkBoxEnableBraceMatching, Settings::Settings::enableBraceMatching);
+	initCheckBox(this->checkBoxEnableNumberScrollWheel, Settings::Settings::enableNumberScrollWheel);
 	initCheckBox(this->checkBoxShowWarningsIn3dView, Settings::Settings::showWarningsIn3dView);
 	initCheckBox(this->checkBoxMouseCentricZoom, Settings::Settings::mouseCentricZoom);
 	initCheckBox(this->checkBoxEnableLineNumbers, Settings::Settings::enableLineNumbers);
@@ -942,8 +950,7 @@ void Preferences::updateGUI()
 	For normal cases, a similar line, inside the function 'on_comboBoxLineWrapIndentationStyle_activated()' handles the disabling functionality.
 	*/
 	this->spinBoxLineWrapIndentationIndent->setDisabled(comboBoxLineWrapIndentationStyle->currentData() == "Same" || comboBoxLineWrapIndentationStyle->currentData() == "Indented");
-
-
+	this->comboBoxModifierNumberScrollWheel->setDisabled(!checkBoxEnableNumberScrollWheel->isChecked());
 	BlockSignals<QLineEdit *>(this->lineEditOctoPrintURL)->setText(QString::fromStdString(s->get(Settings::Settings::octoPrintUrl).toString()));
 	BlockSignals<QLineEdit *>(this->lineEditOctoPrintApiKey)->setText(QString::fromStdString(s->get(Settings::Settings::octoPrintApiKey).toString()));
 	updateComboBox(this->comboBoxOctoPrintAction, Settings::Settings::octoPrintAction);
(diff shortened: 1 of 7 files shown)


Sharma-Hrishabh:
Heyy @thehans @rcolyer , I'm confused about finding a valid approach to normalize the value of numbers changed by scrolling.
I found some clue from https://doc.qt.io/qt-5/qwheelevent.html#angleDelta 
In the current state of PR, when I scroll the wheel one time then the wheel event is triggered twice.[Above link mentions that general mouse works in the step of 15 degrees]. So, the value gets changed by two units.
I can use this angleDelta() for normalizing and write an overloaded method. But, I'm not able to understand this twice capture of wheel event in one scroll.

thehans:
@Sharma-Hrishabh 
I found it even more strange that our eventFilter would not receive wheel events when the scad file is possible to be scrolled up or down.  I spent some time trying to debug this today, and then searching for answers I finally found this blog post:  
http://www.setnode.com/blog/mouse-wheel-events-event-filters-and-qscrollarea/
So it seems that `QAbstractScrollArea` (which `QsciScintilla` inherits from) has some weird way of handling wheel events which requires us to install the eventFilter on its viewport instead its main object.

Our `ScintillaEditor` class's currently calls `qsci->installEventFilter(this);` in its constructor, 
which can be changed to `qsci->viewport()->installEventFilter(this)` to receive wheel events properly.  And this also seems to avoid the problem of duplicated wheel events for reasons I do no know.

However it also seems that putting the eventFilter on the viewport interferes with the keyEvents, and I can no longer use those to edit numbers up and down.  So we might need to `installEventFilter` separately for each?  I honestly am not sure if its ok to install our same filter `this` on both qsci and its viewport, or if we need implement them as separate filters in some way.

Sharma-Hrishabh:
I just tried to install the same event filter on both the viewport and qsci object. It's working fine, but still not sure whether it's good or not.
I also read some articles, which mention the last event filter installed will receive the event first and if it rejects the event then it's propagated to the parent object.

thehans:
`Left click + Wheel` worked ok for me but `Alt + Wheel` is not usable, since it is currently set to magnify my desktop (zoom in/out for the window manager, Cinnamon) on Linux Mint 20.
I think it may be difficult to find a single setting that works for everyone across different platforms.  It would be very nice if the keyModifier / button for actions like this was also configurable through preferences, but that may be significantly more work to implement.

One other improvement I would really like to see is if multiple changes could delay preview until "done".  
I was testing this on a scad file that took a couple seconds to compile for preview, which made it difficult to adjust numbers this way.
So for example, if a number change is initiated with (whatever modifier/button + wheel), then it would show the changed number in the editor, but not immediately trigger preview.  Instead, begin to watch for the *release* of modifier/button before invoking preview.  I think this watching for release could be also handled by the eventFilter, by tracking the state in an extra boolean data member of `ScintillaEditor` such as `wheelEditInProgress`.

Sharma-Hrishabh:
Hey @thehans, this navigation on number uses just modifiers. As they are not actions, I think they are not meant to be present in that Shortcut GUI. So, what I'm thinking is to provide the option to choose the free modifier from checkboxes in that Number Scroll Group Box, which I have added in the Preferences. What's your view on this?

t-paul:
It would be nice if there's a way to integrate the configuration via the ShortCut settings at some point even if they are not actions. I'm not sure how complicated that is, but from user perspective it would be good to have it in one place.

Sharma-Hrishabh:
Acceptable @t-paul , I will look at it. Maybe making dummy actions just for extracting the modifiers here could work.

Sharma-Hrishabh:
hey @thehans @t-paul , I think I may need the changes of Shortcut-GUI branch here, what should I do? Should I go for merging it locally on this branch?

thehans:
We may need to discuss some of the interface changes more in-depth, and how these will all relate with other input configuration such as "Axes" and "Buttons" tabs before diving into changing things.  

I think that parts of the shortcuts-gui will be very useful in achieving a good overall solution.  
But it may not be wise to combine dummy / "Modifier Actions" into the same table as actual actions.
For example: the shortcutCatcher currently ignores a modifier-only keypress such as `Ctrl`, and this makes sense for a true action, but we are now talking about settings which are only concerning modifier keys.

I think the most flexible option would be an ability to choose an "input axis" separately from the modifier key (currently the "axis" could be considered as the mousewheel delta in this case).
But say I wanted to use [mouse button #] click-and-drag vertically (y-axis) to edit a number, or use a joystick input axis (maybe optional modifier even makes sense in that case).  Also in the click-drag case, i think the mouse button could be considered part of the modifier.
I don't know if that would be best handled by an additional tab, or if we should use a separate table view on the shortcuts tab, or modifying the "Buttons" interface, or adding to the "Axes" tab in some way, etc.

Sharma-Hrishabh:
Yes, I think some discussion is needed first.
I directly proceeded for the dummy action work, but it doesn't look logical, as we are dealing with the modifiers only.
We have the following possibilities which I'm seeing for the number scroll:
1. through Alt + Scroll
2. through Left_Mouse Button + Scroll
3. through Drag + Mouse Button.
Another thing that is coming to mind is - these options mentioned above are constrained like we can't use the Right Mouse Button because of the context menu. Similarly, the Ctrl + Scroll is occupied for Zooming QAction. We can't directly give users the options in this case, like in the context menu via the mouse button.
So, I thought at first, these are just Editor Preferences, so putting the direct option to check/uncheck the above-mentioned ways near the 'Step-Size' option.

One more thing I'm doubtful about is, will the joystick controls for this feature will be really useful. As we have to keep the cursor in the right position to scroll. Can joystick do that?. I'm asking as I don't know much about them. Also, I think the axes tab manage translation, rotation, zoom of the model only, right?

thehans:
>    Another thing that is coming to mind is - these options mentioned above are constrained like we can't use the Right Mouse Button because of the context menu. 

Yes, I think that Right mouse button with no modifier would be another special case that should probably not be overridden, but it may still be possible/useful to filter right mouse for some uses *depending on a modifier key*.

> Similarly, the Ctrl + Scroll is occupied for Zooming QAction. We can't directly give users the options in this case, like in the context menu via the mouse button.

I was also thinking about this, and was picturing that zooming (text size) could be considered as another user-configurable setting alongside "number scrolling" if we wanted to provide that option.   On the other hand, I don't want to suggest too much feature creep or over-complication.  So trying to support every possible combination of settings like that is maybe going too far, especially if users are not likely to ever stray from the default (I think Ctrl+wheel is considered a fairly standard method of text document zoom in GUI applications, works in web browser for example).

> So, I thought at first, these are just Editor Preferences, so putting the direct option to check/uncheck the above-mentioned ways near the 'Step-Size' option.

A drop down to choose between a few preset options could be a decent solution without complicating things too much.

> One more thing I'm doubtful about is, will the joystick controls for this feature will be really useful. As we have to keep the cursor in the right position to scroll. Can joystick do that?. I'm asking as I don't know much about them. Also, I think the axes tab manage translation, rotation, zoom of the model only, right?

You're right, joystick is probably not as useful for text-based editing like this, since it would require placing the text cursor via keyboard or mouse first anyways.
I had been thinking about configuration in a very general sense where keyboard, mouse, and joystick would ideally be treated as being on the same level of configurability.   The overall idea relates back to #2977, where I believe the whole configurable shortcut project originated even!  
Then similar issue #3231 was created (and PR #3232 implements something for this but seems like it stalled due conflicting shortcut usage and not having enough configurability builtin to OpenSCAD yet).  
So hopefully that explains my thought process a little better, basically I would like to see all these input methods being more-or-less interchangeable... eventually.

Lastly, I noted in IRC, but just for the record here: I realized my Alt+Mousewheel desktop zoom feature is configurable under "Accessibility", and mine is maybe not the standard/default configuration.  I think I configured it some time ago and forgot about it.  (Even though Mint 20 is only recently released, I did upgrade from Mint 19.3 so some of my older configurations carried forward into this setup.)
So in hindsight Alt+Wheel is not so bad for me, since I can choose some other setting for desktop zoom.

rcolyer:
```
foo = 20;
foo2 = 6;
x = foo2;
sphere(x);
```
Click to the right of the 2 in x = foo2; then scroll down 3 times.  The conversion to foo-1 is a bit alarming, and I think these are not numbers we want to be scrolling.

Similarly, include&lt;foo2.scad&gt; promptly becomes include&lt;foo1scad&gt;, with even the period disappearing.

Clicking on the 1 in 2e1 results in scaling the float by powers of 10.  Is this a feature with an argument in favor of it, or is that a byproduct of the context-free text parsing?

echo(1);  Clicking to the right of the 1 and scrolling works as expected.  Clicking to the left of the 1 and scrolling up converts to 11.  Clicking to the left of the 1 and scrolling down converts to -9.  While I can see how this arises as a byproduct of trying to scroll larger values through zero from a higher step size, it comes across to me as very unintuitive.  I'm raising it for discussion/consideration.

The left click and scroll malfunctions in step size as step sizes change.  For example, echo(8);  Click right of the 8, hold left click and scroll up, holding the mouse steady.  Rather than preserve the digit place originally clicked, as it transitions to 10 it readjusts the click position to between the 10 and 0, causing the scroll to become 8, 9, 10, 20, 30.  Given that the alt-scroll conflicts with scroll+zoom features, the behavior of the left-click zoom option is pretty important as the only other choice.

Modifier option "Both" should say "Either", as it does not require activating both modifiers to enable the scrolling.

nophead:
The mouse wheel normally scrolls in the editor. How does this differentiate between modifying numbers and scrolling?

Sharma-Hrishabh:
@nophead Currently there are two modifiers(Alt or Left Mouse Button) for making it different from normal editor scroll action. You can choose from the Preferences.

Sharma-Hrishabh:
Collecting the cases, in which this feature should ignore changing the numbers-
- A number followed by a period only.
- When the cursor is at the left of the most significant digit.
- The text immediately followed by the digits ( e.g foo2 ). Right now I added a RegExp, which checks that the number is always followed by a '=' or '('. Am I missing more possibilities?

And that abrupt jump of the cursor, is due to the default behavior of caret. One possible way I'm thinking is to shift the caret by one place if it is being changed by mouse button modifier and a new digit is added on right.

Sharma-Hrishabh:
Even that shifting of caret will not work, I guess. The mouse click will select the new place when the number shifts to the right.

rcolyer:
A test suite for you:

1. The only numbers in this code that should be scrollable should be the 9 and the 0.  After updates, validate that none of the other digits can.
2.  Click-scroll up on 9 should continue to increment by 1 as long as the left click is held.
3.  Click-scroll down on 0 should continue to decrement by 1 as long as the left click is held.

```
x9 = 9;
e5 = 0;
e17 = e5;
function _1e2(a,e) = a+e;
module e2(_1, _2_2) {
  echo(_1e2(_1, _2_2));
}
e2(x9, e17);
```

t-paul (approved):

Issue #432, opened 2013-07-09
cmd line file output error when no file present


tenaciousRas:
Using the -o switch from the command line, for example:
/path/to/openscad -o path/to/foo.stl foo.scad

If path/to/foo.stl doesn't exist then OpenSCAD complains:
Can't open file "/path/to/foo.stl" for export

OpenSCAD apparently overwrites /path/to/foo.stl if it exists, otherwise it fails.  However it waits until processing foo.scad to report the error thus wasting time and resources.

It should either:  create the file if one doesn't exist; or check for existence of the export file first, before processing the SCAD.
- this is on OpenSCAD 20120822 on Mac OSX Lion
##

t-paul:
Closed via #3387.

PR #3372, opened 2020-07-12
#432 Added write check

robertlugg:
Fixed empty directory (ie current) case.
If the path is not specified (at least in the Linux case), fs::is_directory() doesn't operate as I expected.  If parent_path() is empty then set to current directory.

In addition to checking the directory, I also "touch" a new file in that directory to confirm that it is at least writable.

### DIFF
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -325,15 +325,6 @@ int cmdline(const char *deps_output_file, const std::string &filename, const std
 		);
 		return 1;
 	}
-	// Create a file in the directory to confirm that its at least writable (issue #432)
-	auto temp_file = output_path / fs::path("_tmp_open_scad.tmp");
-	std::ofstream temp_fstream(temp_file.generic_string());
-	if (!temp_fstream.is_open()) {
-		PRINTB("Can't write to directory \"%s\" for export - Skipping\n", output_path.generic_string());
-		return 1;
-	}
-	temp_fstream.close();
-	fs::remove(temp_file);
 
 	set_render_color_scheme(arg_colorscheme, true);
 
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -311,6 +311,9 @@ int cmdline(const char *deps_output_file, const std::string &filename, const std
 
 	// Do some minimal checking of output directory before rendering (issue #432)
 	fs::path output_directory = fs::path(output_file_str).parent_path();
+	if (output_directory.generic_string() == "") {
+		output_directory = fs::current_path();
+	}
 	if (!fs::is_directory(output_directory)) {
 		PRINTB(
 			"%s is not a directory (output file: %s). - Skipping\n",
@@ -318,6 +321,15 @@ int cmdline(const char *deps_output_file, const std::string &filename, const std
 		);
 		return 1;
 	}
+	// Create a file in the directory to confirm that its at least writable (issue #432)
+	fs::path temp_file = output_directory / fs::path("_tmp_open_scad.tmp");
+	std::ofstream temp_fstream(temp_file.c_str());
+	if (!temp_fstream.is_open()) {
+		PRINTB("Can't write to directory \"%s\" for export - Skipping\n", output_directory.c_str());
+		return 1;
+	}
+	temp_fstream.close();
+	fs::remove(temp_file);
 
 	set_render_color_scheme(arg_colorscheme, true);
 
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -293,6 +293,11 @@ int cmdline(const char *deps_output_file, const std::string &filename, const std
 	else {
 		// else extract format from file extension
 		auto suffix = fs::path(output_file_str).extension().generic_string();
+		if (suffix.empty()) {
+			PRINTB("No suffix for output file %s\n", output_file_str);
+			PRINT("Either add a valid suffix or specify one using --export-format\n");
+			return 1;
+		}
 		suffix = suffix.substr(1);
 		boost::algorithm::to_lower(suffix);
 		if(exportFileFormatOptions.exportFileFormats.find(suffix) != exportFileFormatOptions.exportFileFormats.end()) {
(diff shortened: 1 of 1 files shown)


src/openscad.cc
```diff
@@ -311,13 +311,25 @@ int cmdline(const char *deps_output_file, const std::string &filename, const std
...
 	if (!fs::is_directory(output_directory)) {
 		PRINTB(
 			"%s is not a directory (output file: %s). - Skipping\n",
 			output_directory.c_str() % output_file_str.c_str()
 		);
 		return 1;
 	}
+	// Create a file in the directory to confirm that its at least writable (issue #432)
+	fs::path temp_file = output_directory / fs::path("_tmp_open_scad.tmp");
+	std::ofstream temp_fstream(temp_file.c_str());
```
thehans, line 326: It looks like the windows mxe builds don't like this.  I think you need to call `.generic_string()` first.  That will return a `std::string` which should also be fine to send to ofstream constructor without `.c_str()`

Also, couldn't we just test opening `new_output_file` directly instead of writing and deleting a temp file?

thehans: And if ~~`new_output_file`~~ `filename_str` is opened here in append mode, then it should even help keep previous file contents, if any, just in case the render process fails somehow before it starts writing the new results.

robertlugg: I will switch it over to .generic_string().  Perhaps what I did was only valid in c++ 2017.  I didn't want to touch the actual filename because as you suggest, it might already exist.  I *think* that opening it in append would be fine but I didn't want to touch/contaminate the existing file.  Are you ok leaving it as is?

nophead:
My file systems are monitored by backup and sync programs, so I don't want unnecessary temporary files created as they will get transferred to the cloud and then deleted again as it creates network traffic and wears out SSDs, fills up logs, etc.

robertlugg:
This pull request cleans up my string use so should build on Windows.  If the user does not specify an extension for the -o output file, I was seeing a segfault, so added additional logic to give a better error in that case.

thehans:
In addition to nophead's use case, it looks like the appveyor CI caught another unintended consequence of creating a temporary file like this.  The test report is here:
https://ci.appveyor.com/api/buildjobs/62f3oukneaa0p6yq/artifacts/win_8664_llvmpipe-_llvm-8.0%2C-256_sdvi_report.html

```
stderr output: 
This application has requested the Runtime to terminate it in an unusual way.
Please contact the application's support team for more information.
terminate called after throwing an instance of 'boost::filesystem::filesystem_error'
  what():  boost::filesystem::remove: The process cannot access the file because it is being used by another process: "C:/projects/openscad/tests/opencsgtest-output\_tmp_open_scad.tmp"

Error: openscad.exe failed with return code 3
```
I'm guessing this happened from running multiple parallel tests (`-j <jobs>` command line argument for `ctest`), and one of them got caught removing a file while the other had it open.

So I still think the solution should be to attempt opening the exact filename of the export (in append mode, and skip the file removal step).

Also I would change the message text in that case, to be more about the specific path+filename, than the directory. 
So something like `Export filepath \"%s\" cannot be opened for writing - Skipping\n`

nophead:
I think even opening the file for append will cause unnecessary cloud uploads if it writes a zero length file and sometimes you can get errors with GoogleDrive locking a file if you change it twice in succession.

What problem is this change trying to solve? Can't it just be fixed with script if you are in the habit of rendering STLs to non writable paths. I would rather OpenSCAD just writes the file in one go when it has finished rendering. I can't think of any other applications that do trial writes followed later by the actual write.

t-paul:
I believe using `access()` (as mentioned in https://github.com/openscad/openscad/pull/3366#issuecomment-654427561) should not cause those side effect. I had no time to test that yet though. The different name of the function on Windows could be handled via the PlatformUtils functions.

robertlugg:
This work is based on issue #432 where a user asks to fail earlier...instead of waiting for rendering to be complete and then failing on write, fail before that time-consuming step.  So, it seemed a reasonable desire.
 
To me it would be pretty hard to handle all the cases; disk being full, existing file can't be overwritten.  So, I'm assume the goal is "get it working better than it currently does".  It is straight-forward to check if the directory exists.  But beyond that, I'm not as confidant.  Do I need to check the directory for write permissions?  If so, I have to figure out if I'm user, group, or world and have that work cross platform.  Its certainly possible but seems to be too much effort.

I will change the text as mentioned by @thehans.

I need to fix the concurrent access case no matter what.  I can remove the temp file check completely if you wish.  I really don't want to do anything to the user's existing files...like if I open them for append, will it change the datestamp?  If we can't agree I'll just remove the temp file checking...The check if directory exists is part of the original request and does provide some improvement.

t-paul:
Yes, I think it's a reasonable request. Failing fast in case the output folder/file is not accessible makes sense. It's not going to be possible to check every possible failure mode, so the actual export still needs to handle errors as well.

nophead:
Yes it makes sense to check the directory is accessible but I don't think it should write extra files. In my use case I generated dozens of STLs, they never take more than a few minutes and the write never fails because my scripts write them to somewhere sensible.

Everything gets written to my cloud server, which is RPI4 with 1 TB SSD. SSDs wear out after a few years (I am on the third SSD and second hard disk in this PC), so I never want extra writes to them.

t-paul:
Yep, I guess we can agree that we don't want extra write operations. The extra check would be for people who don't have everything scripted in a safe way or just general ad-hoc usage.

thehans:
OK.  I didn't mean to suggest that we need to cover every potential case, just that the option of opening in append mode would be able to handle an extra (but probably unlikely) case of writing to a read-only file "for free".
But if that's going to still cause problems for cloud syncing then I would agree to just take out any pre-emptive file writing and leave it with the previous implemented check for directory existence.

robertlugg:
Unless I hear otherwise, I'll just remove the writing of a temp file.  I was thinking about an alternative.  It seems there are two types of users.  One is like @nophead who has built and can own the infrastructure around openscad to do exactly what he needs.  And there is the casual user, like me, who doesn't care about things like writing temp files...we just want openscad to work "as easy as possible for us".  These user types are at odds with one another.

What about introducing a new command-line option such as --run-with-scissors.  Most casual users wouldn't use it.  If not specified we do all sorts of checks like the temp file writing of this pull request.  More advanced, professional users would need to specify that option and it would reduce the checks assuming those users handle the infrastructure themselves.

This does require more though for developers.  And it requires that advanced users typically need to specify an additional command line switch.  So, it isn't without its downside.  Like I say I'll just revert the writing of temp file.  That is by far the easiest path but I wanted to suggest this to see if it resonates with any of you.  Regards.

nophead:
How often do people try to write STL files not non-writable locations? I can see misspelling a directory could be a common mistake and checking the directory exists would catch that. After that I would think it is too rare an error to be worth adding an extra command line option. After all nothing bad happens, it just wastes some time. I doubt many people use the command line to render long running STLs. Most people would do it in the GUI and I think most command line use is via scripts.

And I think it should be possible to test that the directory is writable without actually writing to it. That only leaves having a read only file already there. I think that could be tested for as well. So if you really want that level of checking I think it can be done, again without and extra command line option.

MichaelAtOz:
> OpenSCAD apparently overwrites /path/to/foo.stl if it exists, otherwise it fails. However it waits until processing foo.scad to report the error thus wasting time and resources.

Why not just open the output file for write before processing the .SCAD instead of afterwards?

robertlugg:
@MichaelAtOz The behavior is to overwrite any existing file of the same name.  Opening the file sooner risks us deleting the old file and then failing our write.  I've backed off my changes to just check if the directory exists.  I had also added a check if the user entered an incorrect suffix (extension) from the --export-format CL switch

robertlugg:
Hi @t-paul , can you accept this pull request or would you like to have me do something more?  I'd rather close it out with my current changes and move on to another issue.

robertlugg:
All functionality that we want went into a different pull request.  This one, which created output files to better ensure that things would save is too much.  So closing.

PR #3376, opened 2020-07-18
Include tests/CMakeLists.txt as part of ./CMakeLists.txt

thehans:
`tests/CMakeLists.txt` now should work for either in-source or out-of-source builds.
Changed output  path from `[test-name]-output` to `output/[test-name]` for cleaner `tests/` directory (or `build/tests/...`).


t-paul:
Whops, that merge with master went a bit strange.

Issue #1555, opened 2016-01-14
export to binary STL


EastEriq:
For a smaller file size than ASCII. Or is it just me not finding the right place to control it?
My current workflow involves exporting as ASCII stl, and reopening the stl file in MeshLab to resave it as binary.
##

t-paul:
No, binary STL is not supported yet. This also depends on #649 providing a way to specify export options.
In case there's scripting in the workflow, there's also `admesh` which allows converting on command line.

EastEriq:
I see. Thanks for the admesh tip, looking into.

TLC123:
why not bundle admesh with openscad and have some piping trough it in the export dialog that'll solve stl repair issue

t-paul:
I'm not sure general repair on export is reliable enough and it would cover only a single output format. We probably can add the library version of admesh for optional repair at STL import.

doug-moen:
If the requirement is "a smaller file size than ASCII STL"  then we can
satisfy this by supporting compressed AMF, which is on average half the
size of binary STL.

>

EastEriq:
> If the requirement is "a smaller file size than ASCII STL" 

For me it is. Our workshop accepts only STL and clouding 200MB chunks is not always practical. But see #1557...

kintel:
@TLC123 Admesh doesn't perform the type of repair we need. Also, the admesh code for binary STL export is not ideally suited for integration; it's probably less code to implement our own exporter than to write the glue to integrate admesh. See #1042 for more info.

TLC123:
Somewhat off topic. If some one may clarify. if no CSG is used, say we used
just plain polyhedron, couldn't export just bypass CGAL?

kintel:
@TLC123 yes, if no CSG or minkowski operations are used, we typically won't run it through CGAL. That doesn't influence this discussion though (repair issues should be discussed in an appropriate issue).

jtoy:
+1

donbright:
the maintenance headaches of this far outweigh the potential benefits. STL ascii vs binary was designed decades ago when the difference mattered for speed. nowdays that speed question is irrelevant because computers and storage have gotten so much cheaper and faster. modern computers have gigabytes of storage inside gadgets that are smaller than a fingernail and cost 5 dollars. there is no good reason to increase the complexity of a program in order to save a little bit of space on output.



[...]

PR #3381, opened 2020-07-26
Enable default binary stl output, and ascii option

rcolyer:
This provides binary stl output, resolving issue #1555.  Since the binary stl files are about 27% the size of the ascii stl files, I've set this as the new default.  Ascii stl file output is still supported, now selectable in the gui under preferences, advanced, export, use ascii stl.  From command-line, ascii stl file output can still be selected by using --export-format asciistl

The binary stl files seem to show up as a correct match to the ascii ones in my tests, load in other programs, and can be successfully reimported into OpenSCAD with the import command.

### DIFF
--- a/src/Preferences.cc
+++ b/src/Preferences.cc
@@ -660,6 +660,12 @@ void Preferences::on_enableRangeCheckBox_toggled(bool state)
 	settings.setValue("advanced/enableParameterRangeCheck", state);
 }
 
+void Preferences::on_useAsciiSTLCheckBox_toggled(bool checked)
+{
+	Settings::Settings::inst()->set(Settings::Settings::exportUseAsciiSTL, Value(checked));
+	writeSettings();
+}
+
 void Preferences::on_enableHidapiTraceCheckBox_toggled(bool checked)
 {
 	Settings::Settings::inst()->set(Settings::Settings::inputEnableDriverHIDAPILog, Value(checked));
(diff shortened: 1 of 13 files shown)


nophead:
This will break my build scripts. Why not keep the default backwards compatible and make binary an option?

rcolyer:
The default for command line is now set back to ascii stl, with a flag to manually specify asciistl or binstl at the command line, and a suggestion in the --help to phase in updating scripts that require it to include the explicit --export_format for asciistl.  The default for gui outputs is left as binary since this is a much more compact format for uploading and transferring around designs with a very large number of triangles, but it is easily clickable over to a persistent setting of ascii in the gui preferences for anyone who prefers this.  The gui preferences are not used at the command line, hence the need for two separate treatments for user preferences.

The testing scripts were updated to validate binary stl files equivalently to ascii stl files, checking that things are manifold and valid numbers, and should now be validating the equivalence between these two outputs by the standard method used elsewhere in the tests of reimporting, exporting an image, and doing a reference image comparison.  In this case, the binary stl image outputs are being compared to previously generated images from ascii stl image outputs.

donbright:
https://stackify.com/premature-optimization-evil/

donbright:
Binary and ASCII decimal are not equivalent. 

Binary floating point does not have the same numbers as ascii decimal floating point. They are two different sets of points in space. 

Decimal floating point for example can be evenly divided by 10 without loss of information. 

Binary floating point cannot be evenly divided by 5.

rcolyer:
@donbright It's correct they're not exactly equivalent at the very low values, even though this should be negligible for almost any usage.  But the internal values for OpenSCAD are actually binary, and the GRID_FINE grid that points are snapped to is set to the binary value 1/(2 to the 20), which is representable exactly with binary floats, but not ascii.  It's the ascii output that was giving the approximation.  That the differences are negligible is why we got away with the ascii output being an inexact representation of the generated data.

As for optimization, this PR was not about making things faster, even though it does that an extremely tiny bit.  It was about making the output stl files 4 times smaller, which matters a significant amount when they become hundreds of megabytes and you're trying to transfer them around, post them in stl form in places with file size limits, and/or deal with them on space-constrained embedded systems.  I was motivated to work on it after encountering such limits.

Issue #3165, opened 2019-12-22
Comparisons across types and for function literals


thehans:
I am working on a library that provides some general algorithms and utilities to take advantage of function literals, and while testing some features I realized some issues with the builtin comparison operators.  These are mostly a problem for example if trying to sort/binary_search a vector (which might become "poisoned" by an `undef`, or by other mixing of types).

### 1. Comparison between differing types always returns false. ###

I am wondering if OpenSCAD should support comparisons between any two differing types.
One possible point of reference is to look at how Python handles things since it has many similar qualities such as lists allowing mixed types, `None` analogous to `undef`, function literals, etc. 

I ported our `operators-tests.scad` to Python and posted a [comparison of results here](https://gist.github.com/thehans/0c5727488206c1f365af2fd6db465bea/revisions) (edit: updated OpenSCAD expected output with latest changes from #3164).  The first "revision" is the Python results (in red), while the second is OpenSCAD (in green).
The Python output was search & replaced in a text editor to match with OpenSCAD output format more closely (see comments in the `operators-tests.py` within the same gist link).

I tested this on Python 2.7, but one thing to note is that while Python 2 allowed comparison between any type of object, Python 3 seems to have dropped this feature.  It mostly just throws `TypeError`s when attempted.  
I've been told that the python 2 ordering between different types is (mostly) based on lexicographical ordering of the type name.
So the results seem to show an ordering like: `None < number(aka int or float) < list < string`.

`None` (or `undef` for us) compares less than any other value, which I think makes sense.

If we did similar to Python 2 then we could basically sort any list regardless of contents.
Another option would be to follow Python 3's lead and throw an error when incompatible types are compared.
If we leave things as they are, a library could still theoretically implement its own ordering like that with a complicated comparison function, but without that it fails *silently* when a sort is attempted on mixed-type elements, leading to undefined behavior.

### 2. Comparison among function literals always return false ###

Then there is the question of if we can compare function literals in any meaningful way. 
Currently a function literal is not even equivalent to itself, which I think is problematic.

One idea is that the `toString()` of functions could be compared.  This could at least give a reasonable way to order things, however functions whose definition strings match I think could still return different results depending on their closure context.  So maybe equality specifically could be defined based on the actual pointer address or source location of the function declaration.

kintel:
@thehans After merging #3164, could the above text be cleaned up to capture the remaining issues?

thehans:
OK, edited

thehans:
[...]

PR #3383, opened 2020-07-27
Make comparison operators on mixed types return undef.

thehans:
edit: updated description to reflect current behavior 

 - The Inequality operators `<`, `>`, `<=`, and `>=` now return `undef` with a warning on mixed type operands.
 - `==` and `!=` allow checking equality across mixed types, returning `false` instead of `undef`, and do not generate a warning.
 - `==` and `!=`  can now also be checked between two function literals, but ordering of function literals is undefined
 - other binary operators, (and unary minus) which already returned `undef` for invalid type inputs will now generate warnings


thehans:
After recent discussion on mailing lists and IRC, it seems that the changes to comparison between numbers and bools, from #3164, should be reverted.

Additionally I have been convinced that, for consistency, comparison of mixed types should always be `undef`, as we do for other operators such as addition, subtraction, etc.

This also basically resolves issue #3165 which I created, questioning how we should handle comparison of mixed types in some consistent way.  It was originally posed in context of being able to sort any vector regardless of contents, but in hindsight this ability seems not particularly useful.

The other question from #3165 was how to compare between function literals.  I have opted to return `undef` for any comparison between two functions for now.   
Maybe some other approach can be developed if we ever decide there is a good reason to compare them (*and* some reasonable method of comparison).

I would have liked to add a useful warning when mixed comparisons occur, but encountered significant difficulty trying to figure out a way to get source file and line number into the warning.  
The difficulty is that no `Context` is available within `value.cc`.  This could possibly be handled by throwing exceptions, and catching within `expr.cc: BinaryOp::evaluate()` but would also require extra care to ensure that any comparison usage in the rest of the source tree would handle the exceptions properly.
Without source location info, I feel it would be viewed as worthless message spam, so some generic warnings were added, but commented out for future improvement.

nophead:
Shouldn't function literals simple be equal when they refer to the same function and not otherwise. E.g. you could have some hacky module that was passed a function and have some special case test if the function passed was a particular function. Horrible but one might expect it to work.

thehans:
I've tagged this as "Work in Progress" since I am still considering better methods of dealing with warnings.  
I'm not happy with the idea of silently proliferating more `undef` values into scripts without better messages to accompany them.

I am also thinking it probably makes more sense to change `(undef == x)` and `(x == undef)` back to only returning booleans.

@nophead Yeah, defining `==` and `!=` may be a good way to go for function literals.  
I think one particular issue with how to define function equality is that even if two function literals come from the same definition line of code, they may have formed closures over different values, so may not really evaluate in the same way.  Then again maybe just saying if they come from the same definition that is close enough for general purposes.
I think that other inequality comparisons between functions should probably remain undefined though.

thehans:
Warning messages are now implemented for invalid type combinations on comparisons and other binary operators.

The only binary operators which are not capable of throwing warnings are `&&` and `||` since any operand type is allowed to be implicitly cast to bool.

I also decided to make `==` and `!=`always return bool.  So checking equality between different types will return false, but not throw a warning/undef.
Other comparisons between mixed types are undefined, with warnings.
This is how Python3 and Ruby behave for example, which are both considered strongly typed languages.

One *maybe questionable* operation that remains is inequalities between two bools (eg `true > false' is `true`, not `undef`), but I chose to keep that since it has been allowed in the past, seems somewhat logical, also fairly harmless in my opinion.

I've also implemented `==` and `!=` for function literals, with a new test to verify that feature.

The new `UndefType` lays down some groundwork which should ease implementing additional warnings related to undefined behavior in the future.  Eventually I think most instances of undef should have some form of reason message attached to them (though not necessarily immediately triggering a warning, depending on the situation). 

I've removed the "Work In Progress" tag and consider this PR to be complete and ready for review.

PR #3387, opened 2020-07-29
Additional checks for command line writing

robertlugg:
This work was instigated by issue #3372 / #432.
This is my second merge request.  I believe I've fixed all the concerns raised.

The following were used to test the behavior:
# Command line tests
Necessary for regressions to run:
`unset DISPLAY`

# Will error as output file type is unknown (none)
`./openscad non-manifold.scad -o foo`

# Will error as output file type is unknown
`./openscad non-manifold.scad -o foo.abc`
"""
Unknown suffix 'abc' for output file foo.abc
Either add a valid suffix or specify one using --export-format
"""

# Will error as 'nonexist' does not exist
`./openscad non-manifold.scad -o ./nonexist/foo.png`
"""
`'./nonexist' is not a directory for output file: ./nonexist/foo.png - Skipping`
"""
# Will error as a file (openscad) was specified as a directory
`./openscad non-manifold.scad -o ./openscad/foo.png`

# Will error as export-format is not valid
`./openscad non-manifold.scad -o ./openscad/foo.png --export-format xxx`
"""
Unknown --export-format option 'xxx' ignored. Use -h to list available options.
"""


src/openscad.cc
```diff
@@ -288,33 +288,45 @@ int cmdline(const char *deps_output_file, const std::string &filename, const std
 
 	if(!export_format.empty()) {
 		// If command line option --export-format is specified, use that format.
-		formatName = export_format;
+		// Confirm that it is a valid export format
```
robertlugg, line 291: This confirms that the extension (suffix) entered in `--export-format` is valid.  There wasn't an early check like this before.
It is not part of the original issue but seems appropriate.

src/openscad.cc
```diff
@@ -288,33 +288,45 @@ int cmdline(const char *deps_output_file, const std::string &filename, const std
...
-
 	curFormat = exportFileFormatOptions.exportFileFormats.at(formatName);
 	std::string filename_str = fs::path(output_file_str).generic_string();
 	new_output_file = filename_str.c_str();
 
 	// Do some minimal checking of output directory before rendering (issue #432)
-	fs::path output_directory = fs::path(output_file_str).parent_path();
-	if (!fs::is_directory(output_directory)) {
+	auto output_path = fs::path(output_file_str).parent_path();
+	if (output_path.empty()) {
```
robertlugg, line 322: If the output file is left unqualified (ie `my_file.png`) so it doesn't have any path...not even `./`, it would be invalid before this change.

rcolyer:
This fixes the important current errors in the master branch of failing to write to the current directory if unspecified, and also the error in the master branch of throwing std::out_of_range when it cannot determine the file type.

The one additional thing it should probably do though is not continue processing if the --export-format is invalid.  That should exit out.  Current behavior in PR:
```
Unknown --export-format option 'asdfasdf' ignored. Use -h to list available options.
```
and then attempting to process the file anyway.

Better behavior, in keeping with the "don't waste time processing if something is wrong" philosophy of this PR:
```
Unknown --export-format option 'asdfasdf'. Use -h to list available options.
```
and then exiting.  A possible solution (needing testing) might be to not do this check during argument parsing, but to rely on the check which is present (but I think not reachable) in cmdline which will then exit out, but adding in the -h suggestion there.

robertlugg:
@rcolyer:
The logic is a bit convoluted.  --export-format is allowed even in GUI mode (it issues 2 warnings if invalid).  I'm happy to return early but this would cause the GUI case to exit instead of just warn.  Is that OK?  I think it is.

src/openscad.cc
```diff
@@ -566,8 +578,8 @@ static QString assemblePath(const fs::path& absoluteBaseDir,
   auto qsDir = QString::fromLocal8Bit(absoluteBaseDir.generic_string().c_str());
   auto qsFile = QString::fromLocal8Bit(fileName.c_str());
   // if qsfile is absolute, dir is ignored. (see documentation of QFileInfo)
-  QFileInfo info(qsDir, qsFile);
-  return info.absoluteFilePath();
+  QFileInfo fileInfo(qsDir, qsFile);
+  return fileInfo.absoluteFilePath();
```
robertlugg, line 582: resolve lint warning where info here shadowed the function info in this file.

src/openscad.cc
```diff
@@ -1041,6 +1054,7 @@ int main(int argc, char **argv)
 		}
 		else {
 			PRINTB("Unknown --export-format option '%s' ignored. Use -h to list available options.", tmp_format.c_str());
+			return 1;
```
robertlugg, line 1057: Per recent discussion.  Now, in command or GUI mode if --export-format is invalid, exit program.

src/openscad.cc
```diff
@@ -1077,7 +1091,6 @@ int main(int argc, char **argv)
 	}
 	else {
 		PRINT("Requested GUI mode but can't open display!\n");
-		help(argv[0], desc, true);
```
robertlugg, line 1080: Removed help texts which just diminishes the error.

rcolyer:
@robertlugg That does not need to happen.  I was suggesting that the exit after detecting the invalid format should come from inside cmdline, which isn't reached in gui mode.  There's no need to do a redundant check in main for validity only to ignore the error, when it's handled more cleanly in cmdline by existing code for the command line case.  It also feels strange to validate a parameter that is explicitly declared to the user as ignored.

robertlugg:
@rcolyer, I noticed that the command line supported multiple -o arguments and processes them in a loop 'cmdline' is called for every output file.  So, I didn't want to see that same error pop up again and again.  I also wanted that once I discovered a problem that it gets handled right away.  I modified the text and removed the check from within cmdline (it was unnecessary and I believe in the wrong place).

Ideally, the whole command line processing could use some work.  I'd prefer to validate all the arguments up front by, for instance, looping through all the output files and checking all the extensions first.  But I also wanted to make as few changes as I needed to.  Can you take a look at the code as it stands now?  I want to get past this change :), so I'm happy to modify it any way you would like.  Thanks for taking the time to review.

robertlugg:
@rcolyer.  Assuming the regressions run, we might be set.  Since there can be multiple -o options, I didn't want anything global like the suffix override in there.  I ran the following tests and they all seem reasonable:
```
# These work
./openscad non-manifold.scad -o dummy.stl -o dummy2.stl
./openscad non-manifold.scad -o ./dummy.xxx --export-format stl
# These error nicely
./openscad non-manifold.scad -o ./nonexistent/dummy.stl
./openscad non-manifold.scad -o dummy
./openscad non-manifold.scad -o dummy.abc -o dummy2.stl -o dummy3.def
./openscad non-manifold.scad -o ./openscad/dummy.stl
./openscad non-manifold.scad -o ./dummy.stl -o ./dummy2.stl --export-format xxx
```
I appreciate your time and comments.

PR #3390, opened 2020-08-02
Changing apply() to ClipperUtils::apply() to prevert name clash in C++17 (default for GCC 11)

anonymous-14523:
GCC 11 defaults C++ standard to C++17 (http://www.gnu.org/software//gcc/projects/cxx-status.html#cxx17)

openscad.pro selects the default standard ('CONFIG += c++std')

C++17 introduces a std::apply() function (https://en.cppreference.com/w/cpp/utility/apply)

this leads to a name clash in a call to apply() in src/clipper-utils.cc

Thus adding a namespace qualifier.

### DIFF
--- a/src/clipper-utils.cc
+++ b/src/clipper-utils.cc
@@ -144,7 +144,7 @@ namespace ClipperUtils {
 			if (!polygon->isSanitized()) ClipperLib::PolyTreeToPaths(sanitize(polypaths), polypaths);
 			pathsvector.push_back(polypaths);
 		}
-		auto res = apply(pathsvector, clipType);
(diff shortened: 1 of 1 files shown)


t-paul:
The `c++std` is actually referring to file `c++std.pri` which should set the current default to C++14.

https://github.com/openscad/openscad/blob/df175ad3d04959f133f72e7e0edce1a30cafebf5/c%2B%2Bstd.pri#L29

For some reason qmake translates that to ` -std=c++1y` but that's just an alias for C++14. I suspect that's to catch some older compiler version that did not have the actual `-std=c++14` flag.

But it's still useful to prepare for C++17 even though we don't want to require it just now. Thanks for the patch :+1: .

Issue #3388, opened 2020-07-29
GSoC 2020 - Feature-3 : Better Error Reporting &  a separate Error Log


Sharma-Hrishabh:
This issue is meant for discussion on enhancing the error-reporting of OpenSCAD.
The work during GSoC on this feature will comprise of following phases:
1) Adding New Infrastructure 
2) Converting Places calling warning/error 
3) Remove the old method.

Step-1:
- [x] Make a new function like PRINT, and pass different arguments individually, or we can use JSON objects and pass it after stringifying.
- [x] A separate class for ERRORS and warnings will receive all the new messages.

Step-2: 
- [x] Make the console output compatible with the new print function.

Step-3:
- [x] Code Refactoring and remove the old version of PRINT from the codebase

Step-4:
- [x] Handle failing tests

Step-5:
- [x] Make the messages in log clickable to jump to the location.

Step-6:
- [x] Add a filter on the basis message groups in GUI

Step-7:
- [x] Hyperlink, console messages with their location.

Feedback:
- [x] Combine Output Handlers

Todo:
 - [x] Handle Hyperlinks in Unsaved Files.
 - [x] Set the Default location of ErrorLog Widget as tabbed.

Sharma-Hrishabh:
Existing WorkFlow: 
set_output_handler() -> coupled with console widget
PRINT takes the message and set the output_handler_data.

Sharma-Hrishabh:
<a href="https://ibb.co/svx4hzz"><img src="https://i.ibb.co/jv0Ydss/Screenshot-from-2020-08-06-21-15-02.png" alt="Screenshot-from-2020-08-06-21-15-02" border="0"></a>

t-paul:
Closed via  #3392 / #3435.

PR #3392, opened 2020-08-03
Better Error Reporting & a separate Error Log

Sharma-Hrishabh:
PR addressing #3388

ToDo:
- [x] Handle Hyperlinks in Unsaved Files.
- [x] Set the Default location of ErrorLog Widget as tabbed.

Work Done:

Step-1:
- [x] Make a new function like PRINT, and pass different arguments individually, or we can use JSON objects and pass it after stringifying.
- [x] A separate class for ERRORS and warnings will receive all the new messages.

Step-2: 
- [x] Make the console output compatible with the new print function.

Step-3:
- [x] Code Refactoring and remove the old version of PRINT from the codebase

Step-4:
- [x] Handle failing tests

Step-5:
- [x] Make the messages in log clickable to jump to the location.

Step-6:
- [x] Add a filter on the basis message groups in GUI

Step-7:
- [x] Hyperlink, console messages with their location.


src/ErrorLog.h
```diff
@@ -11,10 +11,16 @@ class ErrorLog : public QWidget, public Ui::errorLogWidget
 
 public:
 	ErrorLog(QWidget *parent = nullptr);
+	ErrorLog(const ErrorLog& source) = delete;
+    ErrorLog(ErrorLog&& source) = delete;
```
tombsar, line 15: Indentation type is wrong here.

Sharma-Hrishabh: Thanks for pointing. I will fix it :)

src/mainwin.cc
```diff
@@ -3047,9 +3097,21 @@ void MainWindow::consoleOutput(const QString &msg)
...
+	this->errorLogWidget->toErrorLog(log_msg);
+}
+
 void MainWindow::setCurrentOutput()
 {
 	set_output_handler(&MainWindow::consoleOutput, this);
+	set_output_handler2(&MainWindow::errorLogOutput,this);
 }
 
 void MainWindow::hideCurrentOutput()
```
t-paul, line 3117: Both `hideCurrentOutput()` and clearCurrentOutput()` need to do something with the 2nd handler too.

src/printutils.cc
```diff
@@ -31,6 +37,11 @@ void set_output_handler(OutputHandlerFunc *newhandler, void *userdata)
 	outputhandler_data = userdata;
 }
 
+void set_output_handler2(OutputHandlerFunc2 *newhandler, void *userdata)
```
t-paul, line 40: I wonder if it's better to combine the 2 setters, we don't expect different `userdata` (different MainWindow) in any case I think. So one pointer should be enough.

Sharma-Hrishabh:
The message format for overwriting:
<a was assigned on line 5 but was overwritten> (in file test.scad, line 6)

<from args>(from loc)
 
So, it looks like a workaround, with what are you trying to say @t-paul 
Right?

src/printutils.cc
Sharma-Hrishabh: Done :)

src/mainwin.cc
Sharma-Hrishabh: Done

rcolyer:
- The current hyperlink system for the error log is failing (and ungracefully) if used on an unsaved document.  It opens a new tab to the current directory from which OpenSCAD was launched, and then produces a "Failed to open" "file to open is a directory" error.
- The error log is showing up as split-window for the console, rather than as a tab, which is resulting in not enough horizontal space for either when the editor is open on the left of them.  Most IDEs use a tabbed system for this type of thing to provide enough display space.  In our case, upon preview or render the tab should then auto-switch to the error-log if an error or warning has occurred to draw attention to the issue that has likely scrolled past.

t-paul (approved):

t-paul:
Imported to branch `error-info-window` in  #3435.

PR #3399, opened 2020-08-10
Replace post-increment with pre-increment in counter based loops.

thehans:
This change was prompted by some strange false warnings from LGTM, about comparison always being true in some counter based loops.  I suspected that maybe it was not handling post-increment correctly, and only written to handle pre-increment. So I went about trying to replace most instances of that in simple counter based loops where the result is not used. 

I am mostly curious if this change will cause LGTM to mark these false warnings as fixed.
Its not likely to make any difference in terms of efficiency on a release build with optimizations enabled, since most of these are simple int types.
But maybe some very small improvement on debug builds without optimization.  Also I just consider this a "Best Practice", to be in the habit of using pre-increment (or better yet range-based for), particularly when it comes to complex types like iterators.
(for some reference on the question of performance: https://isocpp.org/blog/2016/08/quick-q-efficiency-of-postincrement-v.s.-preincrement-in-cpp )

Changes were mostly made by a regex replacement across the project, excluding a few paths such as external lib code in `src/ext`.  The regex was also made to add a space after semicolons within for loops, if not present.

A few other changes made while looking over such loops:
 - removed boosty.h, a compatibility header for obsolete version of boost, which our minimum required version now exceeds.
 - a few plain loops replaced by range-based for loops and auto specifier.
 - use const on various RangeType functions and used range-based for loops whenever iterating over them.
     - fixed up `RangeType::iterator`'s iterator traits to make sense as far as I understand their intended meaning


src/FontCache.cc
```diff
@@ -287,7 +287,7 @@ void FontCache::check_cleanup()
 	}
 
 	cache_t::iterator pos = this->cache.begin()++;
-	for (cache_t::iterator it = this->cache.begin(); it != this->cache.end(); it++) {
+	for (cache_t::iterator it = this->cache.begin(); it != this->cache.end(); ++it) {
```
rcolyer, line 290: Inconsistent with range for over the same object in line 277 above.

PR #3407, opened 2020-08-15
Feature/aconz2 stdio

aconz2:
These change allow reading and writing from/to stdin and stdout.

I couldn't figure out a nice way to rewrite `EchoStream` so it uses a conditional on every handle of `output`. Unless I'm missing something, it doesn't need to be a subclass of `ofstream`.

Things I'm unsure about:

1. Windows; does stdout on windows care about binary?
2. The parser takes in the filename and I'm guessing uses it to report errors. When reading from stdin, should the name displayed for reporting get changed to something else than `-`? I think other tools use something like `<stdin>`.
3. Updating the help text. Is there more than one place to document this?

Relevant issues:
#615 #616

### DIFF
--- a/src/export.cc
+++ b/src/export.cc
@@ -80,11 +80,12 @@ void exportFile(const shared_ptr<const Geometry> &root_geom, std::ostream &outpu
 void exportFileByName(const shared_ptr<const Geometry> &root_geom, FileFormat format,
 	const char *name2open, const char *name2display)
 {
+#ifndef WIN32
 	if (strcmp(name2open, "-") == 0) {
-		// TODO: on windows, is cout okay for binary?
 		exportFile(root_geom, std::cout, format);
 		return;
 	}
+#endif
 	std::ios::openmode mode = std::ios::out | std::ios::trunc;
 	if (format == FileFormat::_3MF || format == FileFormat::STL) {
 		mode |= std::ios::binary;
--- a/src/export.cc
+++ b/src/export.cc
@@ -80,6 +80,11 @@ void exportFile(const shared_ptr<const Geometry> &root_geom, std::ostream &outpu
 void exportFileByName(const shared_ptr<const Geometry> &root_geom, FileFormat format,
 	const char *name2open, const char *name2display)
 {
+	if (strcmp(name2open, "-") == 0) {
+		// TODO: on windows, is cout okay for binary?
+		exportFile(root_geom, std::cout, format);
+		return;
+	}
 	std::ios::openmode mode = std::ios::out | std::ios::trunc;
 	if (format == FileFormat::_3MF || format == FileFormat::STL) {
 		mode |= std::ios::binary;
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -883,7 +883,7 @@ int main(int argc, char **argv)
 	po::options_description desc("Allowed options");
 	desc.add_options()
 		("export-format", po::value<string>(), "overrides format of exported scad file when using option '-o', arg can be any of its supported file extensions.  For ascii stl export, specify 'asciistl', and for binary stl export, specify 'binstl'.  Ascii export is the current stl default, but binary stl is planned as the future default so asciistl should be explicitly specified in scripts when needed.\n")
-		("o,o", po::value<vector<string>>(), "output specified file instead of running the GUI, the file extension specifies the type: stl, off, amf, 3mf, csg, dxf, svg, png, echo, ast, term, nef3, nefdbg. (May be used multiple time for different exports)\n")
+		("o,o", po::value<vector<string>>(), "output specified file instead of running the GUI, the file extension specifies the type: stl, off, amf, 3mf, csg, dxf, svg, png, echo, ast, term, nef3, nefdbg. (May be used multiple time for different exports) Use '-' for stdout\n")
 		("D,D", po::value<vector<string>>(), "var=val -pre-define variables")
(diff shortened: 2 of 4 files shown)


thehans:
> These change allow reading and writing from/to stdin and stdout.
> 
> I couldn't figure out a nice way to rewrite `EchoStream` so it uses a conditional on every handle of `output`. Unless I'm missing something, it doesn't need to be a subclass of `ofstream`.

I think if it `EchoStream` were changed to hold a reference to a `std::ostream` base class, then that should allow it to work for both `std::cout` and ofstreams, without requiring a conditional.

> 
> Things I'm unsure about:
> 
> 1. Windows; does stdout on windows care about binary?

Yes.  If I recall correctly linux defaults to binary output, but windows requires to explicitly set the mode for binary output.  
There's also some weird things that openscad has to do to be compatible with windows console and as a GUI windowed application at the same time.  
I don't remember much about the specifics for how it works, but a separate `openscad.com` is built in addition to `OpenSCAD.exe`, which acts as a some sort of wrapper for IO.  Not sure if something will need changing for that.  See source here: `src/winconsole.c`

> 2. The parser takes in the filename and I'm guessing uses it to report errors. When reading from stdin, should the name displayed for reporting get changed to something else than `-`? I think other tools use something like `<stdin>`.

Sounds reasonable to me.

> 3. Updating the help text. Is there more than one place to document this?

In addition to the command line `--help`, probably just the wiki would need updating but I think that can wait til this is merged.

aconz2:
I think the failed tests are Windows specific? I am passing the failed tests locally on Linux. Will try to cross-compile and test on a Windows machine here

t-paul:
The error message is: `decoder error 28: incorrect PNG signature, it's no PNG or corrupted` so there's maybe some additional stuff written to the file at the beginning as the signature is in the first bytes of the files.

Test report: https://ci.appveyor.com/project/kintel/openscad/builds/34678272/artifacts

t-paul:
Oh, it might be possible to just download the MXE build and run that via `wine`. Maybe that already gives some idea what's going on.

aconz2:
Yes ostream reference works! I don't write much C++ as you can tell!

Wasn't passing along the file mode so I think that was killing the PNGs. If that isn't the culprit, I will proceed with doing a real investigation on Windows.

Example of the `<stdin>` substitution for the parser errors.
```
→ echo 'this is not an openscad file' | ./openscad --export-format echo -o - -
ERROR: Parser error in file "/home/andrew/Repos/openscad/<stdin>", line 1: syntax error

Can't parse file '<stdin>'!
```

And after thinking about it, is this even useful? On Linux at least this suffices now that `--export-format` exists (which is something I didn't know about until recently because it wasn't yet on my system's version)
```
→ echo 'echo("Hi");' | ./openscad --export-format echo -o /dev/stdout /dev/stdin
ECHO: "Hi"
```
The one thing that the above doesn't handle as expected is includes, since it looks for the filename in `/dev`. So maybe it is useful.

aconz2:
I updated the help section for `-o` but I'm not sure where exactly the right place for the input file would be.

And something I discovered yesterday which might give you a good laugh is that I had a similar PR open from almost 5 years ago #1440 (just closed it). I have zero recollection of working on that!

thehans:
Thanks for your work on this @aconz2 

Do you think you could add an additional test or two, to verify the new functionality?
This will likely require a whole new test type in `tests/CMakeLists.txt` so openscad could be invoked using IO redirection, something like:  `openscad --export-format=echo - < input.scad > output.echo`

That shell syntax is supported cross-platform as far as I know, but there may be some complication when going through the python test running scripts that we use, not sure.

aconz2:
Made a first pass at adding some tests. I made a few changes to the python test runner to support stdin and stdout redirections and a slight tweak to the `add_cmdline_test` cmake bit to pass along the necessary options, but it works now and reuses the existing test files and expected outputs.

Currently added a few tests for formats ast, echo, and png. Not all echo tests are suitable to reuse the expected output because the test fails with the following (the filename used is `<stdin>` which causes the diff failure). I don't think its necessary to cover all existing tests with stdio tests, so I think these are fine to leave out.
```
-ERROR: Assertion '((a < 20) && (b < 20))' failed: "Test! <html>&</html>" in file assert-expression-fail2-test.scad, line 3
+ERROR: Assertion '((a < 20) && (b < 20))' failed: "Test! <html>&</html>" in file <stdin>, line 3
```

Only other thing maybe worth adding is another binary format. PNG is already tested but the actual export of STL for example happens in a different codepath than it does for PNG

aconz2:
Okay everything passes now, but it doesn't support binary output over stdout on Windows. Currently, on Windows:

```
openscad foo.scad --export-format ast -o -  # writes to stdout
openscad foo.scad --export-format png -o - # writes to the file '-'
```

Writing to the file '-' is consistent with current behavior but could be surprising to the user. We could warn, error, or disable the feature entirely on windows to avoid confusion.

If we can decide on a direction for the above I can finish up by adding a test for binary STL output over stdout. I just didn't want to start on that in case the above is a deal breaker.

aconz2:
@thehans have you had a chance to look/think/discuss this?

t-paul:
Imported to branch [stdio-import-export](https://github.com/openscad/openscad/tree/stdio-import-export) -> #3454.

t-paul:
Trying `_setmode(_fileno(stdout), _O_BINARY);` in #3454 which supposedly allows binary stdout on Windows...

t-paul (approved):

Issue #3420, opened 2020-08-25
Editor still receives key presses even when hidden


ali1234:
Version: openscad-nightly snap, version 2020.07.30 

Reproduce:

1. Open a file.
2. Tick view->hide editor.
3. Type some nonsense.
4. Untick view->hide editor.

Result: the nonsense you typed is in the file.

This is a problem because you can mess up your code with stray key presses, without realising it has happened.

rcolyer:
This bug happens as far back as 15.03.  In the master branch however it has become worse as tooltips even pop-up while the editor is hidden.  All text entry to the editor needs to be actively disabled upon hiding and re-enabled upon unhiding.

t-paul:
I can't reproduce that with locally built OpenSCAD, so I'm wondering if that's a Qt5 bug that was fixed between Qt5.5 (used in Snap package) and Qt5.14 (currently on my local system).

MichaelAtOz:
I can reproduce on W7/64 with 2020.07.05 & 2020.08.21, both show Qt 5.13.

t-paul:
I found https://bugreports.qt.io/browse/QTBUG-82939 which sounds similar but is still open.

ali1234:
I can no longer reproduce this on nightly snap (Qt version: 5.15.13).

PR #3423, opened 2020-08-27
[#3420] disable editor when hidden

eltonlaw:
Addresses https://github.com/openscad/openscad/issues/3420

- Setting the active editor to read only when hidden 94cc6b4 
- If an autocomplete is being shown, cancel it 2b470a0

### DIFF
--- a/src/mainwin.cc
+++ b/src/mainwin.cc
@@ -2857,6 +2857,9 @@ void MainWindow::hideEditor()
 	auto e = (ScintillaEditor *) this->activeEditor;
 	if (viewActionHideEditor->isChecked()) {
 		e->qsci->setReadOnly(true);
+		if (e->qsci->isListActive()) {
+			e->qsci->cancelList();
+		}
 		editorDock->close();
 	}else {
 		e->qsci->setReadOnly(false);
--- a/src/mainwin.cc
+++ b/src/mainwin.cc
@@ -48,6 +48,7 @@
 #include "FontListDialog.h"
 #include "LibraryInfoDialog.h"
 #include "RenderStatistic.h"
+#include "scintillaeditor.h"
 #ifdef ENABLE_OPENCSG
 #include "CSGTreeEvaluator.h"
 #include "OpenCSGRenderer.h"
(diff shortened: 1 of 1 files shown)


t-paul (changes requested):
I think that's actually a Qt bug. It might even be fixed as I can't reproduce the issue on Debian/Testing with Qt5.14.2. There's a similar sounding issue https://bugreports.qt.io/browse/QTBUG-82939 which is still open.

As a number of builds use older Qt versions and are affected by the issue, it seems sensible to have this fix, it would just be nice to have a code comment on why this code is here and that it is likely a workaround.

eltonlaw:
Huh you can't reproduce it? That's weird, on that ticket it says it affects version 5.14.1 and I was able to reproduce on 5.15.0 . Not too sure what's going on.

For the comments, is this alright? 8624880

t-paul:
I'm still puzzled why it does not happen with my local build. It does happen with the Snap package. I do run Wayland but OpenSCAD is not running as native Wayland app.

Anyway, the patch looks good. Thanks.

t-paul (approved):

t-paul:
Hmm, looks like the updated snap does not fix the issue.

MichaelAtOz:
Should there be Windows builds there?

t-paul:
Those should show up automatically on the download page at about 8 EDT.

MichaelAtOz:
Has the build changed or is it sill running?
It used to show e.g.
>  ci/circleci: openscad-appimage-64bit Your tests passed on CircleCI!
Details
ci/circleci: openscad-mxe-32bit Your tests passed on CircleCI!
Details
ci/circleci: openscad-mxe-64bit Your tests passed on CircleCI!
Details

and the files accessed via artifacts.

t-paul:
Oh, on the PR. Not sure why, but it looks like CircleCI did not see that PR at all. It's been a bit problematic for a while now. Other PRs just start showing "Unauthorized" and fail to run too, but I can trigger manually.

t-paul:
Looking again at the Snap it does make the editor read-only. It still triggers the autocomplete though, e.g. by typing `cu` / hide editor / typing `b` (or even any other key).

t-paul:
And it's possible to trigger call-tips via `cube(` + pressing `(`.

Issue #3404, opened 2020-08-14
Feature request: Set the default filename from the program. 


rewolff:
I have a bunch of programs that can render to either different objects from the same assembly or different variants. 

Sometimes it's useful to save the STLs under the same name, so in the slicer you can just do "reload STL from disk" to get the other variant. But often it would be nice to save the objects under say myprogram_objectA.stl and myprogram_objectB.stl . Or myprogram_3x16.stl. This then results in the gcode files also to carry that basename and I can simply hit print again on the printer if I still have a sliced version from a while back. 

I could do all this by hand, but in practice I give up after once or twice and start saving to the default filename again after one or two iterations. 

I'm suggesting something like: 

$default_basename = $mybasename + "objectA";

Where $mybasename is an automatic variable holding the basename of the main file. 
$default_basename defaults to $mybasename, and holds the filename that is suggested when I do "save as STL".

t-paul:
Setting via variable has too many issues and does not scale well for multi file export. The current idea is to be able to annotate top level modules and that could include a default file name too.

See https://github.com/openscad/openscad/wiki/Meta-Data-Use-Cases

kelvie:
@rewolff I have a quick and dirty patch that does this since I also find myself doing this over and over when trying to export different sub-parts of a single scad file.

t-paul:
As already mentioned above, doing it via a variable is not seen as viable solution.

kelvie:
Right, I didn't intend to mainline this, just to share how to do it if someone wanted something (right now) for this use case!

DaleSchultz:
> Setting via variable has too many issues and does not scale well for multi file export. The current idea is to be able to annotate top level modules and that could include a default file name too.
> 
> See https://github.com/openscad/openscad/wiki/Meta-Data-Use-Cases

My use case for this is that I have a _single_ module, and I use parameters to alter the output that the module produces. When I export to .stl file I need that file name to reflect the parameters that were in effect. I can build that into a string myself, and we need to be able to inject that/append it to the default filename. So having one file name per module would not solve the problem.

For example say my module simply creates a cube based on L, W & H variables - which are in the customizer.
When I generate a 23 x 45 x 16 cube I want to be able to generate a file name like this:
$OutputFile = str(L,"x",W,"x",H);

When I then save the stl file, the file defaults to 23x45x16.stl

I don't see the need for additional customization or meta data, we already have the data we need, we just have no programmatic way to get it into the filename. And one filename per module won't work.

[...]

PR #3425, opened 2020-08-30
Add $export_suffix

kelvie:
This will allow you to add a special suffix that will be the default export
name.

For example, for mypart.scad, if you set $export_suffix to "top-part", the
default STL export name will be "mypart-top-part.scad".

See https://github.com/openscad/openscad/issues/3404


kelvie:
Forgot to mention this is quick and dirty, I didn't do much error checking nor worry about unicode.

t-paul:
See related issue for discussion.

PR #3426, opened 2020-09-01
InputManager: Explicitly set current window on startup.

### LINKED ISSUE #3367: DBus not accepting actions on startup
I would like to control OpenSCAD from emacs via DBus. I have enabled DBus in Preferences-> Features -> input-driver-dbus and Axes -> DBus.  It appears that DBus is not being initialized correctly.  (To me, it seems as though it is only able to execute commands after the user has manually invoked some actions from the menu.)

To start with, I have built the (helpful) openscad-dbus-reference app by @MichaelPFrey

To reproduce:
* Start openscad (with a scad file, doesn't seem to matter which)
* Start the dbus reference app (it indicates connected status, good)
* Press "get actions" to populate the list of all available actions (works, good)
* Select an action, such as "viewActionViewAll" or "viewActionHideConsole" and press "send action" to send it

Actual result:
* Nothing happens

Expected result:
* The action should be executed

Now, manually choose some items from the menu (in my case it was View -> Zoom in, Preferences, etc). After enough tinkering, you can try sending the action from the reference app and the action succeeds. From then on, the DBus interface seems to be working (both from the reference app, and from my test functions within emacs).

I am on Ubuntu 20.04 using openscad-nightly 2020.06.14.nightly (git be16bd8) but get the same behaviour with release 2019.05

Lenbok:
The cause of #3367 seems to be independent of DBus, and is a general InputManager layer initialization problem.

I noticed that if I start openscad while "Hide Editor" is unchecked, the problem does not manifest (However, I have this checked by default since I'm using an external editor). Without the editor showing on startup, it seems that InputDriverManager `currentWindow` wasn't being set until after some kind of actions ended up calling `onFocusChanged()`. When the editor is showing, I guess a suitable `onFocusChanged` call happens when the editor is created.

This PR explicitly allows setting the `currentWindow` when the main window is initialized. However, at this point it's mostly a quick hack, as I don't know anything about QT. Please review and suggest a prettier solution and I can whip it into shape :-).

Fixes: #3367


t-paul:
I think the question is how that interacts with the global input (e.g. via 3D-Mouse/Gamepad). The focus filter is there to ensure only the focused window processes those events, even with multiple windows and/or multiple OpenSCAD instances running.

Lenbok:
Right, so I put a debug print in `onFocusChanged`, got openscad to open two windows, both of which have the editor enabled. 

When I mouse from one window to the other (I use focus follows mouse), I see it getting called, and whichever one most recently had focus is the one that acts on the DBus events, even though when I actually send the DBus events, I've put OS-level window focus back to my regular text editor. That's essentially what I would want to happen (except without the requirement to have the built-in editor showing).

If I hide the editor in one window, I never see any `onFocusChanged` events coming from the one without the editor as I mouse back and forth between them (even though it is being focused by the OS). Is that the real bug then?

Lenbok:
I saw that the console output widget definition has a click to focus property set, and indeed it seems that clicking in that area (if it is showing) is sufficient to cause an `onFocusChanged` event to set the focus to that window if you have multiple open. I added that same property to the main preview widget, so you can click on that if you have neither the editor nor the console is showing.

Do you have any other input devices to check whether this PR has any bad interaction with them? (Note that this PR doesn't remove the focus filter, the events still only go to the window that QT last gave focus to).

Lenbok:
I updated this PR to get the new release goodness. It's still working well for me.

Lenbok:
Updated again to current master.
@t-paul Do you or anyone have a 3-D mouse or gamepad to test whether this PR has any undesired side effects? It's still working well for me as a fix for #3367

kintel:
@t-paul This looks pretty innocuous to me. Are you still thinking testing is needed with a 3D mouse or gamepad?

Lenbok:
Still of interest.

PR #3429, opened 2020-09-06
[#3420] disable autocomplete on hide - pt.2

eltonlaw:
I hadn't thought to test additional text input after setting the editor to read only, thought that would already block everything, my bad. Thanks for the comments and sorry for the delay!

Second PR addressing: https://github.com/openscad/openscad/issues/3420
First PR: https://github.com/openscad/openscad/pull/3423

- 7ab7624 fixes https://github.com/openscad/openscad/pull/3423#issuecomment-684084541
- 8d5e046  fixes https://github.com/openscad/openscad/pull/3423#issuecomment-684124142
- f92e0e1 fixes another issue I was encountering where the active call tip wasn't disappearing after hiding the editor

### DIFF
--- a/src/mainwin.cc
+++ b/src/mainwin.cc
@@ -2867,6 +2867,9 @@ void MainWindow::hideEditor()
 		}
 		e->qsci->setAutoCompletionSource(QsciScintilla::AcsNone);
 		e->qsci->setCallTipsStyle(QsciScintilla::CallTipsNone);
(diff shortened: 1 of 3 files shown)


t-paul:
Thanks for taking care of those cases too. To me it's really surprising all of that is needed, showing the popups on a read-only editor seems like a bug in the editor code.

t-paul (approved):

PR #3459, opened 2020-10-11
Opengl vbo renderer

thehans:
This is an update to PR #1935 by @obliviand.

I've made up to date with master, fixed some compiler warnings, implemented suggested changes from my own code review :stuck_out_tongue:, and optimized some of the hotspots I found from profiling.

From my testing obliviand's changes were already a massive improvement in performance.  I previewed an insanely detailed model (I don't dare render it with CGAL, so not sure how many triangles, but Geometry Cache size was 162MB to give some idea) which I clocked nightly snapshot at 1 frame per 13s, and this code gives closer to a guesstimated 4fps, so that's up to 50x improvement!

The optimizations I added mainly just affect the initial creation of VBOs, which add some delay getting just the first frame displayed, but its only noticeable on a debug build with no compiler optimizations anyways.


thehans:
After looking some more at the code and trying some other test cases, I'm less satisfied than I originally was with the changes.

I did see a huge speedup for an object which had smoothed surfaces (my "star knob" https://www.thingiverse.com/thing:1562501 ), essentially low number of polysets, with very high triangles each (essentially composed of spheres and torus which I increased detail until I found some limits where it started to noticeably slow down).  But for other cases such as a large number of cubes, there is no speedup in FPS, and the delay time to first frame starts becoming more noticeable over nightly snapshots.

I think the biggest issue at the moment is the high GPU memory usage due to redundant data.  
I know that VBOs will necessarily use more memory than the previous code which basically streams individual triangle data during each frame, but there's an overabundance of duplicated data with this implementation.
Taking a simple example of rendering a cube:
Cube has 6 faces, 2 triangles per face, 3 vertices per triangle, (6*2*3 = 36) and ALL 36 vertices get stored on GPU, each with its own separate vertex attributes.
And, due to the way the edge shader currently works, each vertex's attributes even includes the 2 other triangle vertex positions. 
So for a single cube it adds up to 108 vec3 positions, essentially 13.5x duplication of the original 8 vertices.  That's not even accounting for the other attribute data such as color, and edge metadata. 
Color should be uniform across each polyset, so no reason to store that data per vertex either.

The changes in PR #3332 would remove the redundancy of the other 2 triangle vertices, getting it back down to 36.

A further improvement would be to refactor Polyset to represent data internally as indexed faces, more in line with how the `polyhedron` OpenSCAD module works.  This should allow us to use `glDrawElements`, which also accepts indexed vertices, and avoids duplication.
But I'm pretty sure #3332 is a pre-requisite to even being able to attempt that.

Reducing vertex duplication would not only reduce GPU memory usage, but also help with the delay to first frame that the VBO changes introduce.  Because most of this time is spent pre-calculating matrix transformations on the triangle vertices, we should be able to get it down to only applying the matrix once per vertex.

obliviand:
I've pulled the changes here back into PR #1935 and have additional changes.

PR #3477, opened 2020-11-09
Refactor methods into PolySet class to provide polygon indexing

obliviand:
This PR adds indexed polygon features to the PolySet class as suggested in PR #3459 comments. This supports creating Indexed VBOs, and using glDrawElements in PR #3459.

The change remove polyset-utils.* and Reindexer.h files since they are now incorporated into PolySet functionality.

Test results are the same (82 fails for Release, non Experimental build, no 3mf). There may be unintended regressions that are not seen with the current tests.

Further performance enhancements are probably needed, but unordered_map is used to generate unique vertices. Places in the code where unique vertices should be used have been updated as well.

Using the Reindexer class as is would also work to create indices and unique vertices during create_ render calls, so having this set of changes will allow for comparison between the two methods.


obliviand:
So, 3mf export is apparently broken. Cannot currently troubleshoot on my build machine since 3mf v1 library not available. I'll have to build a package to install.

t-paul:
Maybe that helps?
https://build.opensuse.org/package/show/home:t-paul:lib3mf/lib3mf-1.8.1

t-paul:
Looking at the test failure, it seems it's just the ordering of vertices and faces which is different.

`ExportMesh` exists to also ensure a canonical ordering which is useful for both our test cases but also users who try to limit the amount of differences between generated meshes.

obliviand:
@t-paul thanks for the lib3mf info. I was able to get lib3mf 1.8.1 installed, and tested that it worked on master branch first with no changes. I reverted export.cc and export.h.

Also forgot to move some of the numeric validity checks over to PolySet. I had to add an empty faces check to PolySet tesselate, but empty faces should be removed while building the PolySet.

kintel:
Superseded by #4842

Issue #3116, opened 2019-11-02
Support for lib3mf 2


diabonas:
Version 2 of lib3mf has been [released on 2019-09-26](https://github.com/3MFConsortium/lib3mf/releases/tag/v2.0.0). This is a major release that breaks ABI compatibility, see the changelog:

> - **autogenerated** language bindings for C++, C, Python, Pascal, C#, Golang and NodeJS have been added
(the defunct COM interface has thus been removed)

Since OpenSCAD relies on the mentioned COM interface to allow import and export of 3MF files, it doesn't build with the most recent version of lib3mf. It would be nice to make OpenSCAD compatible with the new version 2.0.0 in order to allow distributions to drop support for the legacy version 1.8.1.

t-paul:
Which distribution supports lib3mf 2.0? So far it's enough struggle to have 1.8 supported.

diabonas:
Well, I just added [lib3mf-1](https://www.archlinux.org/packages/community/x86_64/lib3mf-1/) to Arch Linux to get 3MF support for OpenSCAD. I would love to use the most recent version instead (see the new [lib3mf](https://aur.archlinux.org/packages/lib3mf/) Arch User Repository (AUR) package), but failed to compile OpenSCAD with it due to the different interfaces.

t-paul:
We can't update until 2.0 is available at least for: Debian/Ubuntu, MacOS, Windows (MXE). So without help, that's not going to be very soon.

diabonas:
Sure, I can have a look into that if I have some time. The build process itself is unchanged from the old version, the only potential issue I found is that the unit tests require the most recent version 1.10.0 of gtest.

As an interim solution, it might be possible to support both versions simultaneously with a version switch, although I haven't looked into that yet.

t-paul:
Sorry, I did mean help in general. I ended up as maintainer of lib3mf in Debian and for 2.0 I assume the interface generator needs to be packaged too. I hope to get some help for that. Also I have to look if the options to use system LIBZIP and ZLIB still exist as Debian will not accept packages with a local copy of those.

If the API difference is not too big it might be an option to support both. We don't use much of the API, really just the core mesh store. Not the perfect solution, but might help to get things earlier. There's even more work going on for the AppImage and Snap packages and doing that as just one person is not scaling too well.

diabonas:
> Also I have to look if the options to use system LIBZIP and ZLIB still exist as Debian will not accept packages with a local copy of those.

[...]

PR #3481, opened 2020-11-10
Add support for lib3MF v2.0 API

obliviand:
In working #3477, needed to build lib3MF v1.8 package, and already had distribution lib3MF v2.0.

These changes keep the existing lib3MF v1.8 API export/import unchanged. Adds new code in ifdef to support lib3MF v2.0 API depending on which version is found during compile. Disable lib3MF tests if library not found.

Implements enhancement #3116


t-paul (approved):
Nice, thanks. I guess the full upgrade to v2.x will take a while as we need to get the lib into all platforms, but this is a great first step.

hroncok:
Awesome! I'll see how far can I get backporting this to the latest release.

PR #2505, opened 2018-10-07
Accept assert() as stable

t-paul:
- Add example for assert() statement and function.
- Remove experimental flag of assert().

PR #3486, opened 2020-11-19
Straight skeleton / voronoi diagram based roof

khumarahn:
This PR implements a "roof" function, which is a 45 degree roof on top of a 2d sketch, with height of vertices following timing of the straight skeleton or Voronoi diagram.

screenshot-20201119@133024
screenshot-20201119@133043
screenshot-20201119@134510


khumarahn:
Here is what I don't know how to do. The input to the algorithm is `Polygon2d poly`. Then `poly.outlines()` contains outlines of all polygons in the shape, but does not seem to contain information of which outlines are holes and which are not.

I need to break this into an array of disjoint polygons with holes.

t-paul:
Very nice!

As for the polygons, I believe the definition of what is cut-out and what is solid is contained in the winding order of the polygon and the first one being the outline (which I think can be any winding order).

So this example:
```
difference() {
    circle(20);
    difference() {
        circle(10);
        circle(8);
    }
}
```
should have 3 polygons with alternating winding order.

To make features available to people for testing and discussion, there is the option to make it an "experimental" feature which would automatically disable it in release builds (see #2505 for a PR showing the final removal of the experimental flag from `assert()`).

The build errors flagged are due to OpenSCAD currently still having 2 build systems. The `qmake` build is still the official one, but the plan is to move fully to `cmake` after the next release. It should be enough to add the new files to the top level `CMakeLists.txt`.

TLC123:
> which outlines are holes and which are not.
Check the winding of each polygon to see outlines and holes. One is usually clock wise and the other is counter clock wise.

khumarahn:
@t-paul @TLC123 thanks!!

What if there are many outlines?

Or if there is ring containing another ring? I would need to separate the two rings into distinct polygons with holes.

khumarahn:
Looks like clipper's `PolyTree` is what I am looking for :-)
http://www.angusj.com/delphi/clipper/documentation/Docs/Units/ClipperLib/Classes/PolyTree/_Body.htm

khumarahn:
YAY!!! clipper's PolyTree does the job. I am very happy.
screenshot-20201119@215332

khumarahn:
At this point all I thought about is working. The code needs polishing and there may be bugs... Feel free to point me to where my code is suboptimal or is not in openscad style.

In the future I would like to implement slices and pass a roof height function for slices, this would be very nice for fillets. (And I have the CGAL version working.) But let's first do the simple roof - it is already very cool on its own.

TLC123:
A very simple way to trunkate the roof and save one difference operation, would be to take MIN of all vertices z components and a threshold height .

khumarahn:
> A very simple way to trunkate the roof and save one difference operation, would be to take MIN of all vertices z components and a threshold height .

This would not give the same result as a difference operation, though.

khumarahn:
I removed my debug messages.... And there is not much left of this PR. It is amazing to which extent everything was already in openscad. Besides the interface part, my contribution reduces to half a page... but even that is an interface too: all calculations are readily served by cgal and clipper.

khumarahn:
I added my files to `CMakeLists.txt` but I'm still getting lots of build check failures. On my machine (gentoo linux) everything works fine (with qmake, I did not try cmake).

khumarahn:
I updated the first messagein this PR: the `roof()` code is now complete and ready for a review.

Though I still need to understand how the tests work, and maybe to add more examples.

khumarahn:
I have been looking for bugs. I found something, and I am not sure where the source of the problem is.

I take the sketch with a hole with sharp corners, and do a small offset with large `$fn`:

```
offset(0.001, $fn=64)
difference() {
    square([12,3], center = true);
    polygon(points=[[-5,-1],[5,-1],[5,-0.05],[4,0],[5,0.05],[5,1],
        [0.1,1], [0,0], [-0.1,1], [-5,1]]);
}
```
test

I can build its roof, and its geometry looks all right in all the ways I can detect (although it has some very small edges):
test

But I cannot subtract it from a cube:
```
!mirror([0,0,1])
difference() {
    translate([0,0,1])
        cube([20,20,2], center = true);
    roof()
        offset(0.001, $fn=64)
        difference() {
            square([12,3], center = true);
            polygon(points=[[-5,-1],[5,-1],[5,-0.05],[4,0],[5,0.05],[5,1],
                [0.1,1], [0,0], [-0.1,1], [-5,1]]);
        }
}
```
fails with
```
ERROR: CGAL error in CGALUtils::applyBinaryOperator difference: CGAL ERROR: assertion violation! Expr: G.mark(v1,0)==G.mark(v2,0)&& G.mark(v1,1)==G.mark(v2,1) File: /usr/include/CGAL/Nef_S2/SM_overlayer.h Line: 294, location unknown 
```
Not sure where the problem might be.

khumarahn:
I found a potential source: my roof generates a mesh where faces may be not convex polygons, and I am storing all that in a PolySet which is designed to only have convex faces... Now I am looking for a way to divide a non-convex polygon into convex polygons.
screenshot-20201120@150247

TLC123:
>Now I am looking for a way to divide a non-convex polygon into convex polygons.

If the concave polygon is co planar and mostly then triangulation is trivial
simple ear clipping algorithm might do it.
Maybe you can find it in cgal
https://doc.cgal.org/latest/Partition_2/index.html

khumarahn:
@TLC123 yeap... Cgal' page looks good... What is the openscad data type for floating point numbers, like that used by Vector2d?

khumarahn:
I.e. am I right that it is double? Can I use double in my calculations?

TLC123:
> I.e. am I right that it is double? Can I use double in my calculations?

As far as i know.

>OpenSCAD's math functions use the C++ 'double' type, inside Value.h/Value.cc,
https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/Mathematical_Functions

khumarahn:
thanks!

TLC123:
> @TLC123 yeap... Cgal' page looks good... What is the openscad data type for floating point numbers, like that used by Vector2d?

The benefit of a custom homebrew partition/triangulation function  over a cgal  is you can avoid the hassle of 3d to 2d by stripping out all z components before decomposition , and then rematch every point in 2d solution with the correct 3d z value after.

khumarahn:
> The benefit of a custom homebrew partition/triangulation function over a cgal is you can avoid the hassle of 3d to 2d by stripping out all z components before decomposition , and then rematch every point in 2d solution with the correct 3d z value after.

Yes. But it is more complicated too, and a badly thought through algorithm may give lots of narrow faces (which is not so good e.g. on export to blender).

t-paul:
See https://github.com/openscad/openscad/blob/master/src/GeometryUtils.cc#L191 which uses `libtess2`.

khumarahn:
I am hitting numerical errors.  If the base polygon is complicated, I may get, as a result of straight skeleton, faces with very narrow spikes, in the spirit of
```
polygon([[0,0], [4,0], [1,-1e-16], [0,1])
```
screenshot-20201120@183029
Here the small number actually should be positive, but because of whatever rounding errors it is not. Such a polygon is not simple (has intersections) and cannot be partitioned into convex parts by cgal.

khumarahn:
There are good and complex news. The good part is that I did not find any non-manifold problems except those caused by antennas. I did a primitive clean-up of antennas for a test, and now some complex test models are working and I cannot detect any non-manifoldness:
test

The complex part is that straight skeletons are tricky:
* we need a reliable way to remove antennas, my method sometimes fails on straight skeletons with very short segments
* there could be lots of very short segments, as on this picture from Huber and Held:
screenshot-20201123@115132
* it would be nice to have our own implementation of straight skeleton - for top-and-bottom offsets, removal of antennas and for proper rounding of corners - but this does not look easy

I have a couple of questions:
* is there a natural way in openscad for me to construct a CGAL polyhedron instead of a polyset?
* can many  short segments lead to boolean errors?

t-paul:
This looks great! Maybe that cleanup would help in other cases too.

I don't have a quick answer to the CGAL questions, I'll have to look into that a bit more myself as I'm not that familiar with that part of the code.

Maybe we can get some of the smaller topics covered in the meantime.

The CI failures seem to be pretty much one problem only and that's a missing reference file for the example `dumptest-examples_roof`. The test suite automatically picks up all example files and uses those as input for the test run. See [testing.txt](https://github.com/openscad/openscad/blob/master/doc/testing.txt).

It should be enough to just generate the missing file(s) via:
```
cd tests
cmake .
make
TEST_GENERATE=1 ctest -R examples_roof
```
This should add the required test reference files in the `tests/regression/` folder.

I'm not sure about the CodeQL failure. That seems to be a linker error that does not happen with the normal builds.

thehans:
While I'm very happy to see some rapid progress being made here and with great results, I don't like the idea of doing this as a separate `roof` module, when there is a PR for `offset_extrude` already.

Having also recently learned about straight skeletons, I *do* believe that it will be the ideal way to implement `offset_extrude`, at least in one particular mode of operation.  That is, when given a `delta` parameter and no `chamfer`; which is precisely the operation that straight skeleton conveys, while containing full information about the exact "height" where vertices merge etc.  That missing info was the main issue of the clipper-based implementation, and so extraneous slices had to be used as a band-aid.

For a quick reference of the three different modes of `offset` (and hence the same modes inherited by `offset_extrude`), see the [pics in this wiki section](https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/Transformations#offset), and hopefully we all agree the straight skeleton corresponds exactly with the center examples in those pictures (delta with no chamfer).

So in terms of the OpenSCAD language/API, I see this as a strict subset of the `offset_extrude` functionality.  I prefer the API of that module since it follows from the existing `offset` feature, and I hope that we can get this straight skeleton implementation integrated into the work of that PR.

And a small note about the other two modes of operation `offset` / `offset_extrude`:
1) When given `r` (for radius, instead of `delta`):  I'm not 100% sure but I believe that the voronoi representation you commented on in the other PR could be ideal for this case.  So that could be another future improvement to the implementation.
2) Lastly `delta` with `chamfer=true`: I'm not sure yet if there's a perfect way to do this using CGAL / straight skeleton, but that's something we can save for future improvement if someone figures it out, and just keep the clipper version for now.  It may be slightly buggy in some cases, but better than having no feature at all.

khumarahn:
@thehans rapid progress is deceiving. I get some exciting things to work right, but I am not sure I am close to a reliable implementation.

I'd be very happy to integrate this with offset_extrude, or with any other command in openscad as long as I can use it. I have been spamming in the offset_extrude PR hoping that the cgal' straight skeleton stuff can be picked up and integrated, and preferably without me :-), it just did not look realistic.

Right now I am looking at using Voronoi diagrams / Delaunay graphs instead of straight skeleton: it is much more complicated to set up, but there may be less numerical problems (say, antennas should not be happening by design). I have it mostly ready, hope to post some pictures soon.

khumarahn:
Voronoi with CGAL works...
test
test 1
I like this offsetting more than that with straight skeleton (though both have their own uses).

CGAL is rather slow on large datasets. It may be better to switch to boost.

khumarahn:
> The CI failures seem to be pretty much one problem only and that's a missing reference file for the example `dumptest-examples_roof`. The test suite automatically picks up all example files and uses those as input for the test run. See [testing.txt](https://github.com/openscad/openscad/blob/master/doc/testing.txt).
> 
> It should be enough to just generate the missing file(s) via:
> 
> ```
> cd tests
> cmake .
> make
> TEST_GENERATE=1 ctest -R examples_roof
> ```

I did not forget about this, will do.

khumarahn:
Ah, I forgot that I have a question. The edges of Voronoi diagrams are straight lines and parabolic arcs. How to discretize the parabolic arcs in a way that is compatible with openscad? (I.e. maybe taking into account things like `$fn`.) Right now I just make the discretization fine enough so that maximum angular deviation is no more than pi/64 or similar, but that is temporary.

MichaelAtOz:
I presume if it is CGAL this does not preview, so will be slow like render() & minkowski()?

khumarahn:
> I presume if it is CGAL this does not preview, so will be slow like render() & minkowski()?

It does preview. I use cgal to do the math, but then generate a usual PolySet. Which then gets converged to cgal in a way I don't understand :-)

Yes, cgal feels rather slow. Still, this thing is way faster than minkowski.

khumarahn:
This is using Voronoi diagram from boost. A picture like this generates in less than 6 seconds (and probably would be faster if I compile it without debug and with optimization).
test

khumarahn:
Today only good news. I translated Voronoi from cgal to boost, this gave a very strong speedup (by a factor of 10 or 100).

The issue with antennas https://github.com/CGAL/cgal/issues/5177 for straight skeleton has been addressed and should make its way into future cgal releases. It is likely that we won't need the ugly workaround, and then straight skeleton offsetting in openscad will be again very short and simple.

khumarahn:
I did a bit of cosmetic work, making the code look more like the real thing.

Besides code cosmetics, I added a parameter `method` which selects whether roof is build using `voronoi diagram` or `straight skeleton`.
screenshot-20201128@121346

I removed my cleanup for the straight skeleton non-manifold mesh. Looks like it is being fixed in cgal, and it is better that we keep our code simple and just fail where we fail.

Two questions:
* What do I do if the algorithm fails? Right now I use `assert(2*2 = 4)` which in case of a problem kills everything.
* How do I write to openscad log? I'd like the user to see messages like 
```
Generating straight skeleton...
Straight skeleton generated.
```

There are optimizations and improvements to do, and more code cleanup. But that is maybe for later. At this point we already have something useful and usable.

khumarahn:
ok, I found how to write messages
```
LOG(message_group::None,Location::NONE,"","Straight skeleton computed.");
```
Now to figure out how to manage errors.

MichaelAtOz:
> 
> 
> ok, I found how to write messages
> 
> ```
> LOG(message_group::None,Location::NONE,"","Straight skeleton computed.");
> ```

I don't think it should do that. If used in a loop it will flood the console.

khumarahn:
> I don't think it should do that. If used in a loop it will flood the console.

You are probably right... I wanted to see these messages because computation of Voronoi diagram and straight skeleton are costly operations, taking a lot of time on complex shapes (though boost' Voronoi is actually pretty fast).

What is a proper way to terminate if a skeleton cannot be computed, or is computed with an error?

t-paul:
In most cases it's currently just giving an error message (hopefully including source code line) and returning an empty object.

You can also use `PRINTDB` for debug info which is not shown normally but with `--debug <c++-filename>`.

For `M_PI` you can use this:
https://github.com/openscad/openscad/blob/df9719630005a2eec4d0e4d08c78077e40dc172d/src/calc.cc#L27-L28

khumarahn:
@t-paul Thank you. I removed my debug messages, will redo them with PRINTDB next time I debug something. I added error messages, and return an empty `PolySet` in case of error.

khumarahn:
Little things.

Voronoi now respects `$fn`, `$fa` and `$fs`:
```
roof(method = "voronoi diagram", $fn=4) sketch();
```
screenshot-20201130@123413
```
roof(method = "voronoi diagram", $fs=0.25) sketch();
```
screenshot-20201130@123523

Error messages got better:
```
roof(method = "straight skeleton")
    offset(0.001, $fn=64)
    difference() {
        square([12,3], center = true);
        polygon(points=[[-5,-1],[5,-1],[5,-0.05],[4,0],[5,0.05],[5,1],
            [0.1,1], [0,0], [-0.1,1], [-5,1]]);
    }
```
causes
screenshot-20201130@123621
The interpretation of `$fn`, `$fa` and `$fs` may not be perfect (e.g. may not respect symmetry of parts). But it is discretization of parabolas, not circles, so not very straightforward.

On the negative side, I see some annoying multi-page gcc warnings of the type
```
/usr/include/eigen3/Eigen/src/Core/CwiseBinaryOp.h:77:7: warning: implicitly-declared ‘constexpr Eigen::internal::no_assignment_operator::no_assignment_operator(const Eigen::internal::no_assignment_operator&)’ is deprecated [-Wdeprecated-copy]
```
not sure what to do with them. Also not sure what I am doing wrong so that ubuntu-20 and mac compile but windows and ubuntu-18 don't.

khumarahn:
I did everything I could think about (except things like layers at varying heights which aren't complicated but could be done separately and later). All build checks are now green... I added examples:
screenshot-20201130@214235

khumarahn:
More good news: the issue with the straight skeleton https://github.com/CGAL/cgal/issues/5177, fixed upstream, will be a part of cgal 5.2 release which should come out this month. This means we really don't have to do any ugly workarounds, and since roof construction is a very direct application of the straight skeleton structure, the corresponding openscad code will stay very short and clear.

I am essentially done, probably I should remove the [WIP] tag.

@t-paul @thehans @mrkmg and everyone else: What should we do here? I am happy if `roof` became a part of the `offset_extrude` functionality https://github.com/openscad/openscad/pull/2079. But I am not sure how to achieve this. For offsets I don't do as much: I have less options, my offsets are primitive and are not the same as in `offset` function. At the same time, as far as the straight skeleton and Voronoi diagram `roofs` go, I get very good and consistent results without the problems of `offset_extrude`.

I mean to say that currently this PR and https://github.com/openscad/openscad/pull/2079 do related things but are not compatible, and it is not clear to me if and how they can be made compatible.

thehans:
As I understand, your roof goes all the way to the peak regardless of offset required to do so.

What is needed is to stop after reaching a specified offset, and cap the top with a flat face. (preferably not requiring a difference operation to do remove the top)  Without looking at the code, filling the top face should be similar to however you fill the bottom face.

It would also need support for `height` parameter, which should be a trivial scaling of Z coordinates by `height/offset`.

If those features can be done then the straight skeleton should work exactly as `offset_extrude(delta=x, chamfer=false, ...)`
and voronoi/delaunay would be equivalent fo `offset_extrude(r=x, ...)` I think.

If those two can be done then, the only remaining feature would be implementation of `delta` with `chamfer=true`, but I wouldn't worry much about that, as it could wait for a future improvement.

khumarahn:
> If those features can be done then the straight skeleton should work exactly as `offset_extrude(delta=x, chamfer=false, ...)`
> and voronoi/delaunay would be equivalent fo `offset_extrude(r=x, ...)` I think.

It isn't hard to stop after reaching certain height, making a truncation without a boolean operation. What bothers me is that the offset would not be quite the same as the one from the `offset` function. The reason is that the algorithms are quite different, and there is little chance they produce identical results in realistic examples even when they should. So if a user tried to stack offsets on top of each other, this may be a problem.

thehans:
Can you show an example of a case where it would be different? 

Roof basically only applies negative offsets from what i've seen, is that the issue?  The CGAL documentation shows some examples of how to apply positive (external) offsets here:  https://doc.cgal.org/latest/Straight_skeleton_2/index.html#title14

khumarahn:
I mean, the mitred offsets would be similar but different numerically: it is hardly possible that such different algorithms produce identical results.

The "rounded" offsets would be quite different... I don't know how to imitate offsets from `offset` here, expect it to be quite delicate.
screenshot-20201213@095818

thehans:
I think I see what you mean about stacking them.  One workaround which could be done in userspace, even if they are not numerically identical, would be to retrieve the top face by positioning it a z=0 and doing projection(cut=true).  Then offset_extrude from that, etc.

If we really need offset and offset_extrude to match exactly, then maybe the implementation of 2D offset should be altered to use straight-skeleton etc in its underlying implementation also.

khumarahn:
yeap, and it gets quite tricky. Offsetting is one of the main purposes of the whole clipper library.

So we get offsets and "roofs" relatively cheaply (by using external libraries) but it is hard to make them consistent between themselves.

thehans:
There is another possible option which should eliminate any need to stack multiple offset/offset_extrude profiles.  

I think the main reason to stack them would be to produce rounded fillets or similar edge profiles, passing different amounts of offset and height per slice.  Instead of having this logic done outside of offset_extrude over multiple iterations, it could accept the offset (`r` or `delta`) and height as parametric functions, by passing function-literal arguments.  That way computation of slices could all be made consistent as it would be handled in a single call.

So a call for rounded fillet with radius 10 could look something like
```
offset_extrude(delta= function(t)-10*(1-cos(90*t)), height=function(t) 10*(1-sin(90*t))) )
  ...
```

This would be a first for a builtin module to accept a function literal argument, but has the potential to be a very powerful feature.

khumarahn:
I too was thinking of passing a function, something in the spirit of
```
offset_extrude(depth=10, slices=4, height=function(t) 10*sin(90*t))
```
Also there might need be a two sided offset like
```
offset_extrude(depth_inside=5, depth_outside=5, slices=4, height=function(t) 10*sin(90*t))
```
(I already did something similar in cgal for straight skeleton in the offset_extrude pr.) These type of things can actually be done as a separate command which divides a shape into slices and warps them along the z axes. Like
```
warp(slices=10, height=function(t) 10*sin(90*t))
    cylinder();
```

I planed to work on this once it is clear what is happening, when this PR is either merged or becomes a part of `offset_extrude`, or when things clear up in any other way.

khumarahn:
Happy new year!

t-paul:
Happy new year and sorry for the quiet days. Release preparation takes most of the time currently, so I hope you are ok to pick things up once the release is out.

I'm with @thehans that this would be a perfect thing to integrate into the `offset_extrude()` feature, but to get some earlier feedback, here's what I'm thinking:

- Push out as `roof()` as implemented in the dev snapshots
- Add extra and explicit warning message that it's currently an experimental feature and may change

khumarahn:
> ```
> offset_extrude(delta= function(t) -10*(1-cos(90*t)), height=function(t) 10*sin(90*t) )
>   ...
@thehans @t-paul Hi!! I am trying to implement custom heights for slices, but frankly don't know how to start with passing a function to the c++ core. I wonder if there an example or guidance on this?

t-paul:
There's no example yet, the function literals are just new in the language and there's no built-in yet accepting those. So this is a new journey we beginning, but also something which would bring quite some extra power to users. It would be great if we could get this working.

khumarahn:
Yeap, I also think that this is a very powerful thing. How do I start?

t-paul:
What I'm thinking is that we need to remember the `Value` given to the parameter instead of resolving it to a low level data type. This would bring in the function definition itself into the `Node`. We probably also need to retain the `Context` as this could have captured information needed for evaluation. With those 2 things in the `Node` we then need to find a way to evaluate the function at visit time with a new `EvalContext` getting the input for the slice being processed.

But as I said, it's not tried before, to this is my current understanding. This certainly will be interesting also from memory handling perspective but I guess getting the evaluation working would be a nice first step.

khumarahn:
This is what I am trying to do:
screenshot-20210213@200054

I implemented refining the mesh into sufficiently small slices (in all x,y,z direction), so that a smooth transformation can be applied. Now I need a way to pass the transforming function. (On the screenshot, the function is coded in C++.)

Frankly, I understand very little here. I'll try to understand what `Context` and `EvalContext` are...

khumarahn:
or rather this is what I want: well 3d printable text on a curved surface
screenshot-20210214@001712

I have it working well for my specific goal, but it remains to pass a general function for a transformation.

richfelker:
Is something blocking merge of this?

richfelker:
Ping. Is anything blocking this, and is there anything I can do to help?

khumarahn:
Hi! I see the situation as follows. The `roof` function works well, does something small and useful (the unix way) with minimal code. Somehow it seemed not enough to be merged, or I had such impression. Also, there is a possible conflict with https://github.com/openscad/openscad/pull/2079 which implements related functionality in `offset_extrude`, a more standard interface but failing in many cases.

It looked to me like it would take a few years before a decision is made, so I started working, in the same PR, on a function `warp`, which would take a mesh, subdivide it and transform, as on the later pictures. Then `roof` and `warp` would work very nice together! This part is not yet ready - I am still to learn how to pass a function from openscad to the C++ backend (but all other things are figured out). I did not have much time to look at it.

khumarahn:
> Ping. Is anything blocking this, and is there anything I can do to help?

I can clean up and only leave what relates to `roof`. Then it can be merged, and I would be very very happy if it did. I can work on `warp` separately (I really want to do it and most is done, the problem it time to concentrate and finish it).

richfelker:
I think `roof` is plenty useful to have already, and sufficiently general to do anything you could do with `offset_extrude` if you apply some transformations and CSG. I'd really like to see this move forward rather than getting shelved again..

khumarahn:
I agree, but merging is not up to me.

richfelker:
> I agree, but merging is not up to me.

Indeed, but do you have any feel for whether maintainers are open to merging, and if so, whether there are any open concerns (aside from resolving the conflicts) that need to be addressed before merging?

khumarahn:
> What I'm thinking is that we need to remember the `Value` given to the parameter instead of resolving it to a low level data type. This would bring in the function definition itself into the `Node`. We probably also need to retain the `Context` as this could have captured information needed for evaluation. With those 2 things in the `Node` we then need to find a way to evaluate the function at visit time with a new `EvalContext` getting the input for the slice being processed.
> 
> But as I said, it's not tried before, to this is my current understanding. This certainly will be interesting also from memory handling perspective but I guess getting the evaluation working would be a nice first step.

@t-paul I think of coming back to this after a rather long break. Is there any progress now on passing function to the c++? I looked for `EvalContext` in the source code, but I can't find it.

khumarahn:
Is it right to expect a function literal to be passed as this?
https://github.com/openscad/openscad/blob/4a02387d997b0b4d20b24e9a17a68270692219ad/src/expression.h#L191

t-paul:
I'm not aware of any active work for passing functions to built-ins. I believe what's passed to the module implementation would be a `FunctionType` or to be exact the `FunctionPtr` which is the variant of Value that can be assigned to variables. This is sort-of the instance of a `FunctionDefinition` bringing also the possibly captured context, e.g. when referencing an outside/global variable.

```
a = 3; // will be captured, not a function parameter
f = function(b) a + b; // f is a value of type FunctionPtr
r = f(5); // call function, resulting in a value of type double
echo(r); // ECHO: 8
```

For moving forward, my suggestion would be to limit this PR to really just the initial `roof()` implementation, so we could get that merged as experimental feature, essentially what was said in https://github.com/openscad/openscad/pull/3486#issuecomment-754800502 is still valid. I'll try to find some time from my side too, so we can get progress. Sorry for the delays, the feature is still awesome, it's really just the problem finding the time while juggling other stuff and real life ;-).

khumarahn:
@t-paul thanks!

I dropped `warp` and rebased on master. I'll need to look at it again, it is not final.

t-paul:
As note regarding latest `master`, we've just dropped the `qmake` support, removing the `openscad.pro` and associated `*.pri` and `*.prf` files.

t-paul:
Testing some examples, I think we need to add a `convexity` parameter, just like `linear_extrude()` has. It's just passed into the nodes, for use by the preview renderer.

khumarahn:
I added the `convexity` parameter. I am not sure I did it right. Is it enough to set `node->convexity` before calling `children.instantiate(node)`?

I also added back `sanitize` for voronoi diagram. I removed it when I was rebasing on the clipperutils changes, but without it some examples break. (Input data for VD is polygon with integer coordinates, and the discretization may break polygons, say by introducing zero length segments.)

rcolyer:
@khumarahn My outstanding concern is about the unaddressed semantic redundancies between this and offset_extrude plans in #2079.  However, offset_extrude is a complicated endeavor that has stumbled across issues for many years.  Still, we want to strongly avoid introducing any language elements that we would then feel compelled to deprecate later after a bit more development, as this risks breaking scad code.  I have a proposed solution for this.  How about if you took roof() and renamed it as offset_extrude() with no or the few parameters specified here?  Then, as the implementation of offset_extrude fills in, the behavior of this PR can correspond to the default behavior of offset_extrude without extra parameters.  I think this is free of conflict, because this happens to be the closest possible thing to a useful result for a no-parameter offset_extrude, and I think this offers us flexibility in future development to make modified versions of the behavior here (for example, the function features discussed above).  Optionally in the future the backing for this and the backing for the other offset_extrude behaviors worked on in #2079 can share code to bring about this flexibility, but how we handle the internal implementation is something the end-user is largely unconcerned with.  My primary concern at this point is setting up a flexible and stable syntax for future development.  I think that would facilitate the balance of merging this base implementation soon, while buying time to figure out the more involved features of all the more flexible offsetting options.  Bite-sized PRs are easier to work with if viable long-term stable features can come in with each one.

What do you think?

Also, in terms of syntax proposed I'd also advocate for truncating method="voronoi diagram" and method="straight skeleton" to the more compact method="voronoi" and method="straight".  Documentation can explain these with more wordiness, but user code should be more compact.

khumarahn:
@rcolyer Hi. I would be happy to leave the naming decisions to the maintainers of the project.

My feeling to consider is that `roof` implements a useful and simple primitive which fits well within the openscad language and is worth keeping as a command. If we now rename `roof` to `offset_extrude`, then it would be quite misleading: `roof` is not an `offset_extrude` on its own. And at the current progress rate, a quality implementation of `offset_extrude` will take many years, I'd expect. Besides, when there is a proper implementation of `offset_extrude`, it may be possible to replace `roof` by a specific call to `offset_extrude` without loss of anything.

P.S.: Calling it `roof` was not even my idea, blame @TLC123 ;-) https://github.com/openscad/openscad/pull/2079#issuecomment-729833993

khumarahn:
<table border="0">
 <tr>
    <td>I attempted to apply this PR to create something useful. These should be bacteriophages on a bicycle handlebar grip. The model came out nicely, with all good chamfers, but it turned out that I don't have a soft enough filament to print them right.</td>
    <td></img></td>
 </tr>
</table>

khumarahn:
> To make features available to people for testing and discussion, there is the option to make it an "experimental" feature which would automatically disable it in release builds (see #2505 for a PR showing the final removal of the experimental flag from `assert()`).

I have trouble understanding how experimental features should be organized. I deleted the example, tests, added `ExperimentalRoof` feature and made `roof` return an empty `PolySet` if it is not enabled. This does not seem to exclude `roof` from the compilation though.

t-paul:
I've tried to collect the info about experimental features in https://github.com/openscad/openscad/wiki/Experimental-Features

As `roof()` is a new module, it should be enough to declare it experimental in `register_builtin_roof()`.

I think the suggestion @rcolyer has brought up is good, but as mentioned earlier we can sort that while `roof()` is in experimental state so people can try it out in the dev builds. Whether we end up keeping `roof()` as separate feature or use it as base for `offset_extrude()` does not need to stop merging.

khumarahn:
> I've tried to collect the info about experimental features in https://github.com/openscad/openscad/wiki/Experimental-Features

Thanks! This makes the last commit easy. But now I cannot make `roof` work, even with `--enable=roof`...

t-paul:
The `--enable` option *only* works on command line. In GUI mode, it needs to be checked in Preferences.

khumarahn:
ah, thank you!!! It works as expected then!

khumarahn:
I added a fix of `tests/regression/dumptest-examples/roof-expected.csg` in the last commit. For some reason, test output changed when I made roof experimental. It is not clear to me is the test is actually checking anything, but it passes (at least on my machine).

src/roof_ss.cc
```diff
@@ -0,0 +1,156 @@
+// This file is a part of openscad. Everything implied is implied.
+// Author: Alexey Korepanov <kaikaikai@yandex.ru>
```
t-paul, line 2: Just to clarify, most OpenSCAD code is GPL2+. You've seen that and your code contributions will be covered by that license too?

Also for completeness, the application, once built with CGAL is actually GPL3+ due to CGAL using that license and the combination of GPL2+ and GPL3+ results in the final deliverable be GPL3+.

tests/CMakeLists.txt
```diff
@@ -686,6 +686,7 @@ list(APPEND ALL_2D_FILES ${FILES_2D} ${SCAD_DXF_FILES} ${SCAD_SVG_FILES})
 #
 # Add experimental features tests
 #
+experimental_tests(examples_roof)
```
t-paul, line 689: The test output changed because the experimental flag was not applied. It needs the full name of the test case which is `<test-type>-<test-filename>`.
```suggestion
experimental_tests(dumptest-examples_roof)
```

t-paul (changes requested):
I found a couple of minor issues, with those solved we can merge. Thanks for hanging in there, I know how those finishing touches can get a bit boring but with the tiny dev team it helps to keep things a bit more managable.

khumarahn:
Thanks! I will take care of this soon. I'm fine with any opensource license.

Can you share an example where you see that convexity does not work? (I don't yet understand what it is and how it should work.)

t-paul:
Rumor has it that it's not always the same picture depending on the GPU/Driver, but for me, the simplest example is
```
difference() {
    roof() offset(10) text("H", 100);
    translate([0, 0, 6]) cube([1000, 1000, 500]);
}
```
Rotating this such that the different concave parts of the shape overlap each other shows the effect. This is caused by parts of the image not drawn enough times so it "paints" the cut part of the `roof()` as white (or background) instead of the object color from "behind" that cut part.

grafik

A convexity value of >= 2 should fix that. See also the chapter in the [manual](https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/2D_to_3D_Extrusion#Description_of_extrude_parameters).

tests/CMakeLists.txt
khumarahn: I made the change you suggest, but this did not change the test output...

khumarahn:
@t-paul I think did all the changes you asked for. Again, I am fine with any license. Convexity now works on your example using `roof(convexity = 2)` or similar.

Strangely, the output of the test did not change.

tests/CMakeLists.txt
t-paul: Did you rerun the top-level make to process the cmake file update and `TEST_GENERATE=1 ctest -R roof`?

khumarahn: I was builing in a separate folder (`build`). I did re-run cmake,make stuff in the test folder... Now I did a fresh cmake,make routine, building in the main folder, and it looks like it helped.

t-paul (approved):

khumarahn:
Amazing, very many thanks! :-)

MichaelAtOz:
This seems to be the `roof` genesis issue, so I'm posting here, if someone wants it elsewhere or in a new issue let me know.
I'm still behind in my GitHub reading & the Mailing-list posts, so sorry it's coming in at this late point.

But, I don't think `roof` is a good name, I do note the change on discussion:

> @rcolyer Hi. I would be happy to leave the naming decisions to the maintainers of the project.
> 
> My feeling to consider is that `roof` implements a useful and simple primitive which fits well within the openscad language and is worth keeping as a command. If we now rename `roof` to `offset_extrude`, then it would be quite misleading: `roof` is not an `offset_extrude` on its own. And at the current progress rate, a quality implementation of `offset_extrude` will take many years, I'd expect. Besides, when there is a proper implementation of `offset_extrude`, it may be possible to replace `roof` by a specific call to `offset_extrude` without loss of anything.
> 
> P.S.: Calling it `roof` was not even my idea, blame @TLC123 ;-) [#2079 (comment)](https://github.com/openscad/openscad/pull/2079#issuecomment-729833993)
> 
> EDIT: And I support shortening "voronoi diagram" to "voronoi" and "straight skeleton" to "straight".

Where it was mentioned `Alternative? roof ( ) / hip_roof ( ) / roof_extrude ()` to `offset_extrude()`

Basically it resembles a "roof" only on basic cubism like shapes.
When used, for example, on text or bacteriophages it is not roof-like.
Most (all?) build-ins have a geometry like name.

I don't have a good suggestion right now, but I will review the various usages in the issues and do some thinking. (taper?)
I presume we don't want to use `offset_extrude`?

Issue #3452, opened 2020-10-02
Conditional geometry should be completely removed from render tree


t-paul:
Reported via [forum post](http://forum.openscad.org/The-behaviour-if-intersection-has-changed-td30152.html)
```
intersection() {
    cube();
   
    if(false)
        sphere();
}
```
The `if()` should take out the whole subtree, so the result is the cube, not the empty intersection between a cube and an empty object.

jordanbrown0:
Note that this will in theory affect children-based patterns:

```
    module foo() {
        children(1);
    }

    foo() {
        if (false) sphere();
        cube();
    }
```

will generate a cube today, and with this would be an error.

nophead:
True, I wonder how many people do that as it looks ambiguous, compared to how many do it in an intersection?

You can either reference the last child instead of the second one or add an else empty() to force an empty child.

Or perhaps if(false) should still add an empty group node but that should not get converted to an empty() node because it is not the same as empty geometry.

jordanbrown0:
And since children aren't evaluated until they are referred to with `children()`, you can't know how many objects they will create until you refer to them, and you can't refer to them because you don't know their indexes because you don't know how many objects their earlier siblings will have returned.  And you can't know the value of $children until you've evaluated all of them.

bneuen:
I agree with @nophead here.  In the example @jordanbrown0 provided, if you want children(1) to give you an empty(), then you need to add an else that adds the empty.  Another way to look at this, is what if you want the if to actually change/determine that first child?  So I think the if conditional should completely remove its subtree.  I'm running in to this with a module I wrote that worked great on previous releases of OpenSCAD, but is now broken.

bneuen:
So given the current structure of the code, I'm not sure having the if remove its subtree is going to be easily done, so might have to look at what @nophead mentioned.

It seems reasonable to interpret a null geometry differently from an empty geometry group, but I'm not familiar what existing interpretations/assumptions come with these.  From what little bit of the code that I have seen, it seems like nullptr (null geometry) is return on most error conditions.  Are there cases where an empty geometry group is returned to indicate something?

I think evaluation of a false 'if' expression is currently returning nullptr; so would it be safe to make that an empty group like @nophead said, and then ignore empty groups in operations (certainly the boolean ones, but probably minkowski too).

PR #3505, opened 2020-11-30
Allow 2D objects in 3D hull and minkowski operations

bneuen:
Allow 2D objects in 3D operations by promoting the 2D object to a
3D object of zero height positioned on the XY plane.


t-paul:
What's your experience with those zero height replacements? Where would that be used? CGAL is usually quite picky regarding zero area triangles (one of the main issues with STL import). Also I wonder if it makes sense to just generally enable this. It seems like there's a number of cases where mixing 2D and 3D does not make much sense. The most useful case is probably hull() but with 2D objects having not just Z = 0 but no Z at all it's not so simple to handle.

PR  #2796 introduces the idea of 2.5D object, so 2D object that can move in 3D space. That might be a possible path for this change too.

bneuen:
The motivation for this change was wanting to Minkowski sum a 2D shape with a 3D object, so the 2D shape would basically define a path/region to sum/union the 3D object over.  This could be mimicked by linear extruding the 2D shape first by a infinitesimal height, but that will introduce unnecessary regions in the calculations (we only need the bottom of the extrusion for that, the sides and top can be omitted).

I was initially only going to make the change if the operation was Minkowski, but then I realized that it could have applications in some of the other 3D operations (such as hull), so I made it so 3D operations would just promote a 2D object in general.  However, this probably makes less sense with union/intersection/difference operations.  I could see intersection/difference for a 2D object and a 3D object, in that the 3D object could be used to cut the 2D object, but the output of this should probably be a 2D object, which is currently not the case.  I'm not sure union really makes sense at all unless the 3D object is sliced at the xy-plane, and that resulting 2D object is then unioned.  A similar slice approach could be used for intersection/difference as well.

Since only the 2D polygon is promoted to a 3D PolySet, and the sides/bottom are not generated, there should be no zero-area triangles created as long as the original polygon did not have such triangles.  I'm not sure what operations could cause zero-area triangles though, so that would probably be something to reason through.

I agree that PR [#2796 ](https://github.com/openscad/openscad/pull/2796) does seems similar in some respects, but I think its goals are indeed different.

t-paul:
Right, `minkowski` would be another interesting candidate, even with 2D objects only in x/y plane. Maybe it's better starting out with the special case and see if other cases could be covered later too. From backward compatibility view it's better to start more restrictive and then look how to generalize. That's not ideal from coding perspective, but as description language, it's better to minimize impact on user scripts at the expense of maybe a bit more work on code side.

MichaelAtOz:
There is another use; simple labelling with `text()` or otherwise documenting with "lines"/pointers/outlines (2D).
If such zero height objects do not produce geometry, unless involved with applicable operations (`minkowski` etc) then they are ignored. 

Obviously not good for documenting 2D...

As currently F5 is happy with mixed 2D/3D, F6 gets `WARNING: Mixing 2D and 3D objects is not supported, in file , line 2`,
an alternative is such documentation with 2D could be handled with $preview used to exclude it in render.

May want to change that to ~`WARNING: 2D objects with 3D are zero height`

bneuen:
@t-paul I plan to come back to this soon, but I'm currently trying to resolve issue [#3452](https://github.com/openscad/openscad/issues/3452).  When you get a chance, can you give additional insight on that issue?

@MichaelAtOz I'm not completely following.  Are you saying that this promoting 2D objects would break that documentation method or perhaps help it.  It seems strange to document something by exploiting the lack of a feature (which does produce a warning, which should be an indication to not do it).  If we restrict promoting 2D objects to hull and minkowski, I think that should not affect such 2D documentation?

MichaelAtOz:
> @MichaelAtOz I'm not completely following. Are you saying that this promoting 2D objects would break that documentation method or perhaps help it. 

@bneuen Help it. I'm not doing anything like that ATM, but have done - [gist](https://gist.github.com/MichaelAtOz/7e6b243bbc03dff381e41ee857f0b77c):
text definitions

Think instead some complex 3D design where you want to build-in such documentation.

Or something like:
Capture-netfabb-measure

Presuming zero-height-promoted-2D do NOT create any geometry, unless affected by e.g. minkowski().
That they could then co-exist in preview and be used to document/label things, would be ignored in render.
That could be done with `if ($preview) do_docs()`, but lazily dropping 2D's around is a perfect misuse...
```
cube(10);
label("cube(10);",[-10,12,15],[45,0,0]);

module label(label,t,r,c="black",s=5) 
translate(t) rotate(r) color(c) text(label, size=s);
```
Capture-labelled-cube




What happens to a promoted 2D rotated by x or y?

nalimcos:
Hello! This looks like this would be a lot more efficient than using the hull operation on infinitesimally thin linear extrusions, and would probably make a big difference for some of my models (woodwind instrument bores with several folds and variable non-round profiles...); any chance of seeing this in nightly builds soon? :smiley:

nophead:
I don't see how hull and minkowski would need much work to accept 2D and 3D objects as both are well defined when mixing 2D and 3D. I think most hull algorithms use points anyway.

t-paul:
> I don't see how hull and minkowski would need much work to accept 2D and 3D objects as both are well defined when mixing 2D and 3D. I think most hull algorithms use points anyway.

Yep, hence the dedicated mention of minkowski and I agree hull is a similar candidate that could be a starting point without bigger internal changes or new data structures.

bneuen:
Sounds like I should pick this back up.  Last time I looked at it, I ran into issues with running the verification suite on the change (some operations in the tests that fail before will now succeed, so the expected result is different and therefore they will need updated).

kintel:
@bneuen Yeah, the way to go about it would be to write up some test designs that thoroughly exercise the new features, then generate expected output for those, and potentially remove/clean up elements from existing tests.

t-paul:
@bneuen do you have 2 or 3 minimal test scripts per module (hull, minkowski) that exercise the new allowed cases? I can help getting those added in case https://github.com/openscad/openscad/blob/master/doc/testing.txt is not clear enough.
It would be great to get this as first step with the next one maybe introducing 2.5D (a.k.a. 2D objects in 3D space). But small steps are good, so lets get this one ready first :grinning:.

bneuen:
@t-paul I think I have that done.  Let me know if there are tweaks we should make to those tests or additional tests you want added.

kintel:
What do people feel about corner cases? Should we add a bunch of tests here, or open new issues for that?

bneuen:
Test failure:
> The following tests FAILED:
>	1937 - amfrendermanifoldtest_mirror-tests (Failed)

I actually saw this the first time I ran the test suite.  I looked at the PNG, and it was blank.  I never did figure out what might have gone wrong (the intermediate output looked fine as I remember).  Upon simply re-running the test suite again, everything passed, and I haven't seen this failure again locally.  Is this a known issue?

bneuen:
> What do people feel about corner cases? Should we add a bunch of tests here, or open new issues for that?

What corner cases are you thinking?

kintel:
Corner case ideas:
* Complex 2D polygons; multiple outlines, holes and islands
* Self-touching 2D polygons, self-intersection polygons
* Complex 3D polyhedrons; multiple volumes
* 3D manifold connectivity corner cases; vertices toucing, edges touching etc.

Most of these are likely implicitly tested, but could be good to at least give them a manual spin, e.g. check that the files in this folder work as expected with hull/manifold: https://github.com/openscad/openscad/tree/master/tests/data/scad/3D/misc

kintel:
There are some flaky tests, especially when running on the CI - if it goes away, don't worry. We haven't had the resource to figure that out..

PR #3508, opened 2020-11-30
Fix error messages (fixes #3507).

### LINKED ISSUE #3507: Broken error messages
```
echo(version());
polygon([[1]]);
```

Gives:
```
ECHO:[2020, 10, 11]
ERROR: Unable to convert points[%d] = %s to a vec2 of numbers, in file , line 2 
```

I.e. the %d and %s are not expanded here: https://github.com/openscad/openscad/blob/fabef1f871231f987fac2bbc451bcdc592da3cb5/src/primitives.cc#L633

t-paul:


nophead:
Thanks for a very quick fix but the build system seems to have changed, it looks complete alien to me. Is it still possible to download build artefacts to get a Windows executable?

t-paul:
Yes, the process is unchanged but unfortunately CircleCI allows the download only after login (e.g. via Github) with their new web UI. As the fix is merged, it should show up automatically over night on the download page too.

PR #3516, opened 2020-12-06
Update animation.scad

jlpoolen:
Instructions for newbies so that they can quickly view the animation without having to discern the subtleties of the interface.

Also, the current 2019.05 version has "Animate" rather than "Animation", so the instruction is updated with the correct term.


t-paul (approved):
Thanks!

PR #3546, opened 2020-12-20
RFC: mouse selector with ctrl+click

Tomatower:
This will change the selection-Click to **CTRL + Click**. (The modification is 1 line of code - this here should be a discussion about the "correct" UX!)

**Benefits**

* No possible confusion with double click to center object (talking to you @nophead :smiley: )
* This allows for a "Selection-Mode" while Ctrl is pressed in future work, for example to select vertices, edges and faces for measurements.
* has roughly the same meaning in source code, when ctrl + clicking on `include`d/`use`d files

**Downsides**

* I think Apple handles Ctrl + Click not always as ctrl + click, did not find good ressources on that (Maybe @kintel can help out here or suggest a "apple-feel" way to do this.)
* The feature is not as prominent as before, and users might miss it

@t-paul This has to be resolved for the next RC, as it will be nearly impossible to change later.

### DIFF
--- a/src/QGLView.cc
+++ b/src/QGLView.cc
@@ -298,7 +298,8 @@ void QGLView::mouseReleaseEvent(QMouseEvent *event)
   mouse_drag_active = false;
   releaseMouse();
 
-  if (!mouse_drag_moved) {
+  if (!mouse_drag_moved
+      && (event->button() == Qt::RightButton)) {
     QPoint point = event->pos();
     //point.setY(this->height() - point.y());
     emit doSelectObject(point);
@@ -351,7 +352,7 @@ void QGLView::zoomCursor(int x, int y, int zoom)
   const auto old_dist = cam.zoomValue();
   this->cam.zoom(zoom, true);
   const auto dist = cam.zoomValue();
-  const auto ratio = old_dist / dist - 1.0; 
+  const auto ratio = old_dist / dist - 1.0;
   // screen coordinates from -1 to 1
   const auto screen_x = 2.0 * (x + 0.5) / this->cam.pixel_width - 1.0;
   const auto screen_y = 1.0 - 2.0 * (y + 0.5) / this->cam.pixel_height;


t-paul:
Pushed to `mouseselection-click-macos` branch to trigger MacOS build...

t-paul:
I believe MacOS actually does Right-Mouse-Click via CTRL+Click for both mouse and touchpad?

So in the end it might make sense to use right mouse click as it's essentially a context menu we show, even though it's currently very specialized. There could be additional actions later. This might also make it slightly easier to discover.

Would that interfere with the double click to center and/or future options for things like selecting objects?

nophead:
Right click is already used for pan. This is what gives me the false
triggers as well as left double click.

Logically it would be right click to bring up a context menu on Windows but
unfortunately it has been used for pan for years.

t-paul:
That's right-drag, not right-click. I have no issues with that on Linux, maybe the event handling is different per platform to cause the trouble? I see how it can cause problems with left-click for things like activating the window when it's in background. But I would not expect that to be an issue with right-click.

I'm not against using the control modifier but it makes it quite a bit more annoying to use when it needs both keyboard and mouse to activate.

nophead:
Yes but right drag starts with a click, so I keep getting the menu when I am panning. Not only that but the menu gets activated and I lose my view. I thought this was supposed to be disabled when the editor was hidden but it isn't on RC3.

t-paul:
RC3 prevents opening files.

Maybe we can find out how to fix that click vs. drag issue on Windows. I can't even reproduce that via Wine. It even works fine separating both actions via touchpad which always is a bit annoying with right clicks.

nophead:
Does it work when the view is laggy due to it being complex?

nophead:
I often get it when I release the right button after a drag.

Tomatower:
The context menu is created upon the `MouseRelease` Event, If the mouse has not moved between mouse down and mouse up.

I (maybe naivly?) thought that Qt triggers `MouseRelease` only for left mouse button - if it were triggering on a right click would be a bug - ~~I will check that~~ (Confirmed) Fix incoming!

But this would again be a one-line fix, the more difficult question is: What is the "correct" way to trigger that?

Ctrl + Left or Right click with no move?


@nophead: I cannot reproduce in daily use. Do you often do very minimal movements, only moving the mouse very few (and therefore also maybe 0) pixels?

nophead:
No it can happen on long moves. It is possible my mouse has switch bounce on the right button.

MichaelAtOz:
> 
> No it can happen on long moves. It is possible my mouse has switch bounce on the right button.
 
If it doesn't happen during left-drag rotations (same Qt handling due to bug mentioned above), then I suspect you are right. I have no issues with right-drag on Windows.

> I'm not against using the control modifier but it makes it quite a bit more annoying to use when it needs both keyboard and mouse to activate.

+1. 
Unless I'm editing code, I drive much of my use with customised settings on a four-button trackball with scroll wheel, with most programs a combination of menus/icons/other-gui-stuff/hover/right-click can access most required functions. 
With modifiers I need to reach for the keyboard. I get neck/arm/hand nerve compression so need my neck in a neutral position, am not in the conventional desk/chair typing setting. My stuff is Autohotkey driven, so I could configure it to do control-lock-for-next-click, but for me the look-up UI-action suits right-click in a conventional UI.

No offence @nophead but I think a new mouse may be a better solution. I know it's a personal preference, I can recommend the Kensington Expert Mouse Wireless Trackball - not-a-mouse but that its stupid name, you don't need desk real-estate and all that lift-relocate stuff...

nophead:
This mouse hasn't been used much, so it hasn't worn out. It is a Microsoft optical mouse and they don't seem to know how to debounce switches. It works fine for everything else and it used to be fine for OpenSCAD. Not easy to get a new one where I am at the moment.

The reason I suspect switch bounce is I have another Microsoft mouse at mt other residence that is old and it developed switch bounce on the left button, so I physically swapped it with the right one, as it normally doesn't matter on right click.

t-paul:
Quick test on MacOS looks ok. It's using CMD+Click which is expected Qt behavior.

Is there a way to determine if we are chasing mouse button bounce problems? @MichaelAtOz highlights the concerns I had, while functionally ok with additional keyboard modifier, it requires the extra effort I would like to not enforce on users if we can help it. But if there's an actual issue, e.g. with designs that are slow to preview we may need that extra modifier for now and hope we can make things configurable at some point.

nophead:
My PC is actually a laptop with a track pad and that does not seem to have the problem, so I think it is switch bounce on my right button.

t-paul:
Ok, so lets go with simple right mouse click only then (no keyboard modifiers), bascically Windows context menu style (show on mouse button release). That should work ok on MacOS too.

Tomatower:
Alrighty, now with right click only, no modifier keys anymore (and no left click).

MichaelAtOz:
@nophead if you want to confirm it is bounce, swap the buttons, see if the physical-right-logical-left has the problem while drag-rotating.

nophead:
Good  idea. Yes that confirms it only happens with the right button. I.e.
rotating instead of panning.

t-paul:
Looks good on MacOS, triggers on right-click and ctrl+left-click on touchpad.

Sounds like we can go ahead with this setup.

t-paul (approved):

Issue #3143, opened 2019-11-23
$children is not computable?


thehans:
This is a problem related to work being done for lazy union, where we would like to be able to loop over individual children from a ListNode, when that ListNode is instantiated via `children()` from a user-defined module.

Here is my original example:
> Giving a `for` loop as the child in a module call, results in `$children==1` rather than one for each iteration.
> ```scad
> // Note: Enable "View -> Animate"
> 
> // "particle" module with short name
> module p() { square(); }
> 
> module explode() {
>   echo($children);
>   for(j=[0:1:$children-1]) {
>     // angle is multiple of golden ratio to get "sunflower" pattern
>     a = j*1.61803398875*360; 
>     translate(j*$t*[sin(a),cos(a)]) children(j);
>   }
> }
> 
> // single child, does not move
> explode() for(i=[0:1:200]) p();
> 
> // vs. specifying each child (works as intended, with or without lazy union)
> /*
> explode() {
>   p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();
>   p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();
>   p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();
>   p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();
>   p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();
>   p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();
>   p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();
>   p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();
>   p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();
>   p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();p();
> }
> //*/
> ```
_Originally posted by @thehans in https://github.com/openscad/openscad/issues/350#issuecomment-554828820_

So, the problem is OpenSCAD currently bases `$children` off of the number of `ModuleInstantiation` under the call to a user defined module.  In this case the `explode()` call has 1 child which is `for(i=[0:1:200])`.   
What we'd like to be able to do is actually loop over the list of nodes created by this `for` loop.  

[...]

PR #3555, opened 2020-12-26
Don't return empty group for "if" statement with failing condition and no "else".

### LINKED ISSUE #3452: Conditional geometry should be completely removed from render tree

thehans:
fixes #3452


jordanbrown0:
Does a failing "if" count as a child?

```
module count() {
    echo($children);
}

count() {
    if (false) {
        cube();
    }
}
```

thehans:
Yes it does. 

`$children` is currently based on AST nodes, not CSG tree, because it needs to be determined before/during the expression evaluation stage (ie it needs to be known at the time that `if` conditions will be evaluated, so it can't really depend on results of those conditions).  Also see https://github.com/openscad/openscad/issues/3143

t-paul (approved):

nophead:
What about the true if case when the child is not geometry, i.e. echo?

thehans:
An `if` with true conditional will group any and all children as usual.  If the children don't happen to contain geometry then the result will be the empty set due to implicit union.

If you think groups without geometry shouldn't count as empty geometry, then you're basically asking for the `lazy-union` feature.

nophead:
Does echo return a special "no geometry" node? If so, can it be propagated upwards so that `if(true) echo()` behaves the same as `echo()` and a module that wraps `echo()` behaves like `echo()` in an intersection?

I don't think it makes sense that `echo()` and `if(false) echo()` don't create an empty group but `if(true) echo()` does.

jordanbrown0:
I'm very nervous about deviating from the "one module invocation, one geometry node" rule, but I agree that once you **do** deviate from it for `echo()` and `if(false) ...`, you should also deviate from it for `if(true)echo();` and similar.  I suspect this also extends to `for(...)echo` and modules containing no "real" geometry.

Note that constructs that *do* contain real geometry, but that result in emptiness - intersections of non-overlapping objects, in particular - should continue to be represented as "real" geometry nodes.

thehans:
So far, group nodes have basically functioned identically to union.  
But it sounds like what people are asking for is that any group with no geometry nodes inside should be treated as a special case and removed from the render tree instead.  
This might be feasible to implement without full lazy union support, but I also worry that this could open another can of worms of unintended consequences and regressions.  Especially this late in a release cycle.

Issue #3547, opened 2020-12-24
Hull bug in RC3


revarbat:
I’ve found a bug that appears in OpenSCAD 2020 RC3 and 2020.12.21.

The code below incorrectly previews AND renders a hull.  The code looks silly with all sorts of unecessary `translate([0,0,0])`s and `mirror()`s, overridden `$fn=4` and `convexity=` args, but if you remove or change any one of those, then the hull will be suddenly preview and render correctly.

```openscad
hull() {
  translate([-4,-1,1]) cube(2, center=true);
  translate([3,0,0]) {
    intersection() {
      rotate([90,0,0]) {
        linear_extrude(height=4, center=true) {
          circle(r=2, $fn=4);
        }
      }
      translate([0,0,0]) {
        translate([0,0,0]) {
          mirror([0,0,1]) {
            translate([0,0,0]) {
              linear_extrude(height=4, center=true, convexity=2, $fn=4) {
                circle(r=2, $fn=4);
              }
            }
          }
        }
      }
    }
  }
}
```

The correct hulling, as shown by OpenSCAD 2019.05 is:
<img width="450" alt="Screen Shot 2020-12-24 at 1 41 03 AM" src="https://user-images.githubusercontent.com/350001/103079797-70c92a80-4589-11eb-8910-30406bc8fd10.png">

The INCORRECT hulling that RC3 exhibits is:
<img width="450" alt="Screen Shot 2020-12-24 at 1 40 47 AM" src="https://user-images.githubusercontent.com/350001/103079835-82123700-4589-11eb-945b-73d3e838ca9f.png">

revarbat:
This code posted was pared down from the CSG tree of a shape in the BOSL2 library that ran afoul of this bug.

nophead:
It looks correct for me with RC3 on Windows.

[image: image.png]

revarbat:
If it helps any, I'm using OSX OpenSCAD 2020.12.21.  The bug was found by someone running Windows OpenSCAD 2020.12-RC3.  The original code that exhibited the behavior was using the BOSL2 library (github.com/revarbat/BOSL2):

```openscad

revarbat:
The full untrimmed CSG tree is:
```openscad
group() {
  group() {
    group() {
      group();
      group();
      group();
    }
    group();
    group() {
      multmatrix([[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]) {
        group() {
          group() {
[...]

PR #3568, opened 2021-01-01
Add constness to CGAL_Nef_polyhedron for cache safety.

thehans:
Fix for issue #3547


src/CGAL_Nef_polyhedron.cc
```diff
@@ -23,25 +25,28 @@ CGAL_Nef_polyhedron CGAL_Nef_polyhedron::operator+(const CGAL_Nef_polyhedron &ot
...
 	return *this;
 }
 
 CGAL_Nef_polyhedron &CGAL_Nef_polyhedron::minkowski(const CGAL_Nef_polyhedron &other)
 {
-	(*this->p3) = CGAL::minkowski_sum_3(*this->p3, *other.p3);
+	this->p3.reset(new CGAL_Nef_polyhedron3(CGAL::minkowski_sum_3(
+		const_cast<CGAL_Nef_polyhedron3&>(*this->p3),
+		const_cast<CGAL_Nef_polyhedron3&>(*other.p3)
+	)));
```
thehans, line 49: > Postcondition
>     If either of the input polyhedra is non-convex, it is modified during the computation, i.e., it is decomposed into convex pieces.
> 
from https://doc.cgal.org/latest/Minkowski_sum_3/group__PkgMinkowskiSum3Ref.html

There are some other places where I had to `const_cast` which seemed like mistakes in CGAL's api (e.g. see `CGAL_Nef_polyhedron3::bytes()` below), but in this case I'm not sure if a full copy should be made of each opearand (might make 3D minkowski even slower).

I have no idea if this potential decomposition could have an adverse effect on subsequent operations, in cases where a modified input object has been already cached, then re-used elsewhere in the script.

t-paul (approved):

src/import_nef.cc
```diff
@@ -25,8 +25,9 @@ CGAL_Nef_polyhedron *import_nef3(const std::string &filename, const Location &lo
 	std::string msg="";
 	CGAL::Failure_behaviour old_behaviour = CGAL::set_error_behaviour(CGAL::THROW_EXCEPTION);
 	try{
-		N->p3.reset(new CGAL_Nef_polyhedron3);
-		f >> *(N->p3);
+		auto nef = new CGAL_Nef_polyhedron3;
+		f >> *nef;
```
rcolyer, line 29: Mem leak of nef here on exception throw.

src/CGAL_Nef_polyhedron.cc
rcolyer: Yes, I'm suspicious of this specific const_cast, although I think the others are safe.  I think this one is formal UB because of the modifications, and thus ambiguously hazardous beyond the cache issues.  The copy should not be a significant performance difference on a minkowski.

PR #3578, opened 2021-01-03
don't show autocomplete in strings (fixes #3563)

### LINKED ISSUE #3563: 2020.12-RC3 Autocomplete activated by decimal point in string
I presume this is not by design.
Capture-RC3-autocomplete-decimal

When I typed the `.` the pop-up poped-up.
Also type `a="x.` pops-up, or `a="a.` not `a=".`

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/95587754-2020-12-rc3-autocomplete-activated-by-decimal-point-in-string?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

donc42:
Autocomplete will be disabled inside string literals. When the editor wants to show an auto complete this code will scan the current line up to the cursor position using '"' characters to determine if the user is in a string literal. 

This will fail on the second and subsequent lines of a multi-line string literal. To fix that would require a lot more scanning of the current text.

### DIFF
--- a/src/scadapi.cc
+++ b/src/scadapi.cc
@@ -19,6 +19,26 @@ ScadApi::~ScadApi()
 
 void ScadApi::updateAutoCompletionList(const QStringList &context, QStringList &list)
 {
+    //first see if we are in a string literal. if so, don't allow auto comlete
+    bool lastWasEscape=false;
+    bool inString=false;
+    int line, col;
+    qsci->getCursorPosition(&line, &col);
+    auto sLine = qsci->text(line);  //get the current line of text
+    QByteArray ba = sLine.toLocal8Bit();
+    const char *c_str = ba.data();
+    const char *p = c_str;
+    while (col-- > 0) {
+        if (*p == '\\')
+            lastWasEscape = true;   //next character will be literal handle \"
+        else if (lastWasEscape)
+            lastWasEscape = false;
+        else if (*p == '"')        //string toggle
+            inString = !inString;
+        p++;
+    }
+    if (inString) return;  //we are in string literal, don't return autocomplete list
+    //not in string literal, proceed as normal
 	const QString c = context.last();
 	for (int a = 0; a < funcs.size(); ++a) {
 		const ApiFunc &func = funcs.at(a);
--- a/src/scadapi.cc
+++ b/src/scadapi.cc
@@ -17,8 +17,6 @@ ScadApi::~ScadApi()
 {
 }
 
-int inString=false;
-
 void ScadApi::updateAutoCompletionList(const QStringList &context, QStringList &list)
 {
     //first see if we are in a string literal. if so, don't allow auto comlete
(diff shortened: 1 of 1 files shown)


src/scadapi.cc
```diff
@@ -19,6 +19,26 @@ ScadApi::~ScadApi()
 
 void ScadApi::updateAutoCompletionList(const QStringList &context, QStringList &list)
 {
+    //first see if we are in a string literal. if so, don't allow auto comlete
+    bool lastWasEscape=false;
+    bool inString=false;
+    int line, col;
+    qsci->getCursorPosition(&line, &col);
+    auto sLine = qsci->text(line);  //get the current line of text
+    QByteArray ba = sLine.toLocal8Bit();
```
t-paul, line 28: Using local 8-bit data breaks the column calculation if the string contains multi-byte UTF-8 characters, e.g. typing:
```
echo("😀😀😀"); c.c.c.
```
will popup the autocomplete only at the 3rd `c.` as each of the [smiley emoji](https://emojipedia.org/grinning-face/) is encoded in 4 bytes (U+1F600 = f0 9f 98 80).

I think for the current logic of iterating though the beginning of the line, converting via `sLine.toStdU32String();` and using `char32_t` should work as 32-bit per character are wide enough to always map 1:1.

t-paul (changes requested):

src/scadapi.cc
donc42: ok, I thought it would just see the '"' as an 8 bit.

I will make the change.

t-paul: Yes, the quote is handled just fine and there's no possible collision with the utf-8 encoding as multi-byte cases will have the top-most bit set. But the line length is 20 (including the internal newline) but with `toLocal8Bit()` the `QByteArray is actually 29 long. So the closing quote is only seen at position 19.

donc42: ok, that makes sense. I didn't realize that the local 8 bit routine would return multiple 8 bit characters for multi-byte. 

First time using Qt. Live and learn, die and forget it all. ;-)

t-paul: Yeah, no worries, it's just a thing I'm trying to watch out for as we had a couple of those issues in the past like this dance from #3071 for the search highlight:
https://github.com/openscad/openscad/blob/8d63488af35a46011899ee42a4ee0bb7de7f291c/src/scintillaeditor.cpp#L662-L665

There's still a couple of places where those multi-bytes encodings cause trouble, like the error marker is also off a bit in some cases.

t-paul: It should and it does output smileys on my Linux machine. I just did a fix on the error output so if it still does not work for you with the latest master, that's a bug. I'm not sure, but I think there's a bug report for Windows having issues with UTF-8 in the console.

donc42: I have a master from a couple of days ago. I can try on the latest later.

t-paul (approved):

PR #3580, opened 2021-01-04
Fix uncaught exception from boost::fs

### LINKED ISSUE #3579: uncaught boost::filesystem exception when using relative path for -o .ast/.csg
This short snippet causes openscad to crash with uncaught exception:

```
  (mkdir -p z/a z/b/c && echo 'sphere(r=10);' > z/b/c/1.scad && cd z/a && ../../openscad ../b/c/1.scad -o 1.ast)
terminate called after throwing an instance of 'boost::filesystem::filesystem_error'
  what():  boost::filesystem::current_path: No such file or directory: "/kvm/src/openscad/openscad.git/z/b/c/../b/c"
```

Same with -o 1.csg.

I think the problem is just leftover code in do_export(), extra fs::current_path(fparent) calls when this is already done in cmdline(). I will submit a patch.

 - Kristian.

knielsen:
The do_export() function wants to switch to the parent directory of
the .scad source file for .ast and .csg output. But the path to that
directory is computed at the wrong place, when a different directory
is the current one. This causes failure when the source file is
specified by relative path.

Fixes #3579

### DIFF
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -463,9 +463,6 @@ int do_export(const CommandLine &cmd, Tree &tree, Camera& camera, ContextHandle<
 {
 	const std::string filename_str = fs::path(cmd.output_file).generic_string();
 
-	auto fpath = fs::absolute(fs::path(cmd.filename));
-	auto fparent = fpath.parent_path();
-
   unique_ptr<OffscreenView> glview;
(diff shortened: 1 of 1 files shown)


t-paul (approved):
Good catch, thanks.

PR #3591, opened 2021-01-06
Fix export_import_pngtest.py for relative source path

knielsen:
This patch fixes a bunch of testsuite failures in the Debian packaging. The packaging uses the testsuite in a different way from normal upstream build (eg. running it outside of the build directory for autopkgtests), which triggered this issue.

I do not think this patch is critical to get into the release, I'll just carry the one-line patch in the packaging along with a few other patches. So no need to do an RC6 just for this, for sure.

----

Use an absolute destination path for .csg output. This is necessary to
put the destination file in the correct place when the source path is
a relative path.

When outputting .csg, openscad interprets the destination path
relative to the parent directory of the source file, not relative to
the current directory when the program was started. This means that
something like `openscad a/b/c.scad -o a/b/c.csg` will try to write
a/b/a/b/c.csg, not a/b/c.csg as was expected by
export_import_pngtest.py.


knielsen:
Closing this pull request to redo it against the `master' branch

PR #3595, opened 2021-01-08
Update zh_TW translation

JustinSDK:
fix a typo and add more translation.


t-paul (approved):
Thanks!

t-paul:
Also thanks for the reminder the template file needs updating, I missed that yet again. This should make some more texts translatable.

PR #3599, opened 2021-01-09
handle missing file on command line

donc42:
This patch handles the case of a file specified on the command line not able to be opened. It will leave the current tab empty with the title untitled.scad.

### DIFF
--- a/src/tabmanager.cc
+++ b/src/tabmanager.cc
@@ -481,14 +481,18 @@ void TabManager::openTabFile(const QString &filename)
     const auto knownFileType = par->knownFileExtensions.contains(suffix);
     const auto cmd = par->knownFileExtensions[suffix];
     if (knownFileType && cmd.isEmpty()) {
-        setTabName(filename);
-        par->updateRecentFiles(editor);
+         setTabName(filename); //refeshDocument gets filename from editor
+         if (refreshDocument()) {     //doc opened
+            par->fileChangedOnDisk(); // force cached autoReloadId to update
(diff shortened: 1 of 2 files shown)


src/tabmanager.cc
```diff
@@ -481,14 +481,18 @@ void TabManager::openTabFile(const QString &filename)
...
     if (knownFileType && cmd.isEmpty()) {
-        setTabName(filename);
-        par->updateRecentFiles(editor);
+         setTabName(filename); //refeshDocument gets filename from editor
+         if (refreshDocument()) {     //doc opened
+            par->fileChangedOnDisk(); // force cached autoReloadId to update
+            par->updateRecentFiles(editor);
+        }
+        else {
+            setTabName(""); //can't open file
```
t-paul, line 490: I'm wondering if this case should still set the file name. This allows for `openscad new-file.scad` to give it a name already which will then be the default for save.

donc42: I guess my thought was openscad new-file.scad doesn't actually open a file so it is misleading for the filename to be on the tab title bar. For example, openscad new-file.scad would act the same as openascad empty-file.scad even though the latter would open a real physically present file. 

Also, suppose you did this in a directory without write access. When you went to save it you would get an error (the next issue I was going to look at).

It is up to you. Let me know if you want me to make the change.

Thanks

t-paul: Hmm, good point, I'm not sure what the SaveFileDialog does when given a non-existent name. I'll need to have a look at that.

PR #3600, opened 2021-01-09
Update zh_TW translation based on the updated template

JustinSDK:
polish and add more translation based on the updated template.


t-paul (approved):
Thanks!

Issue #3570, opened 2021-01-01
Timestamps for console messages


t-paul:
In some cases it might be useful to have timestamps for console messages. That should be added in a way that does not that screen space for every message line or so. For initial discussion, see #3540.

Also needs some further discussion to cover both command line and GUI usage.

thehans:
Another idea idea for timestamping things, as opposed to the absolute date and time, would be to log seconds since program start.  A bit like how `dmesg` under linux writes seconds since boot for every line in the log (at least on my distro).

jordanbrown0:
Maybe only system-produced messages, not echo().  That would address some compatibility concerns.  (Though really we should ensure that the CLI lets you cleanly separate system output from user output, without needing pattern matching.)

jordanbrown0:
Two ideas:
* Have the timestamp be in a collapsible column, so that when collapsed it doesn't take screen space.  Maybe show the timestamp as a tooltip when you hover over a message.
* Have there be a preferences item for how the timestamp is shown:  yyyy-mmm-dd hh:mm:ss, just hh:mm:ss, local or UTC, or relative to OpenSCAD startup, or relative to clearing the log, or relative to program start, in hh:mm:ss, or just seconds, with or without milliseconds.  Implementing any number of variations is of course easy; the hard part is which variations to implement.

jordanbrown0:
If format is a preferences item, then it would presumably be controlled by a CLI option, and the default would be "none", preserving output compatibility.

PR #3602, opened 2021-01-09
Buffer messages to GUI Console for better performance.

### LINKED ISSUE #3421: Using "include" causes script output to be delayed until the preview completes
... which may never happen if the script contains an infinite loop.

Version: openscad-nightly snap version 2021.01.03 (0+git.a980a3a) on Ubuntu 20.04

See https://github.com/openscad/openscad/issues/3421#issuecomment-753754721 for simpler test-case.

Original description: 

```
//include <empty.scad>

//data = [1, 2, 3, 4];
//unit = 10;

function f(arr, _i=0, _acc=[]) =
	_i==len(arr) ? _acc : f(
		arr, _i+1,
		concat(
			[_i==0 ? 0 : _acc[0]+(arr[_i]*unit)], _acc
		)
	);


//x = f(data);
echo(x);
```
`empty.scad` should be an empty file.

This code intentionally has two errors: `data` and `unit` are not defined. The former causes an infinite loop and a warning about `arr` being undefined. The latter causes a second warning about `unit` to be printed in every recursion. The call to `f` is commented for convenience - uncomment it to see the results. The include is commented because its presence affects the result.

**Note:** The infinite loop here also seems to be undetected by the recursion checker, so it will also force you to kill openscad.

When the include is commented, the code will spam two warnings - the limit of 5 warnings fails because there are two different errors. That is a separate problem, and is expected.

However, when the include is uncommented, no warnings at all will appear. Even if `unit` is defined so that the warning limit still works, then the 5 warnings about `arr` being undefined will not appear.

Lastly, if `data` is defined so that there is no infinite loop, then the warnings about `unit` get printed with the include, but there is a slight delay before the compiler produces any output. So it seems that including the empty file causes output to be delayed until processing is complete.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/92481284-using-include-causes-script-output-to-be-delayed-until-the-preview-completes?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

thehans:
Also should fix #3421.


MichaelAtOz:
Fixed on Windows-7.

thehans:
I realized I accidentally change the behavior of the autoscrolling, once you click cursor somewhere you lose it forever.  I will return that back to previous behavior, and also still testing some other possible optimizations.

Due to the two `appendHtml` (colored Warnings/Errors etc.) vs `appendPlainText` (echos), it is much faster than before to display many echos at once, or many warnings at once, but not as much advantage if there is a mix of them.

thehans:
I think this PR is at a good stopping point now.

Here is the basic echo speed test I used, which goes until recursion limit of 1 million is reached (console display is still limited to 5000 by `maximumBlockCount` setting though):
```
function f(x=0) = echo(x) f(x+1);
a=f();
```

With this PR:
Console Hidden: `Total rendering time: 0:00:12.569`
Console Visible: `Total rendering time: 0:00:12.433`
No significant difference.

Compared to nightly:
Console Hidden: `Total rendering time: 0:01:13.822`
Console Visible: DNF, gets to maybe 1000 in same time the PR would have finished.

I also rewrote how color-formatting and links are inserted, so it calls QTextCursor functions directly instead of parsing HTML.  
This should help speed up Warnings and Errors in addition to echos but at this point Error Log becomes the most noticeable bottleneck.
This change should also help facilitate the possibility of user-specified coloring for echo without requiring full html support, if we decide to add that.

Test with warnings (unknown variable y):
```
function f(x=0) = echo(x, y) f(x+1);
a=f();
```
Duplicate warnings after the first 5 are suppressed in Console, as before.  But logging becomes increasingly slow and memory hungry, since Error Log continues to add all 1 million of them.  Hiding Error Log helps speeds things up, but it still took me over an hour to finish one run on this branch, and used up 1.7GB memory.

As @rcolyer and I discussed on IRC, a simple fix for this would be to implement a max limit of say 1000 warnings (maybe user configurable) before script is aborted.

Another option: Have Error Log keep a `std::set` of messages whose text and location match, and just increment a new `count` property for duplicates.  This count could be displayed in its own column, instead of adding a row for every duplicate.
I guess this could potentially lose some information about the overall ordering if there are multiple different warnings recurring, but I'm not sure how much that matters to anyone.  Anyways the Console would still have the exact order if needed (at least for the last 5000 lines).

There are many other further improvements I've been considering, but I think I'd rather leave those for separate PRs
 - Timestamps could be added as tooltips per line (#3570)
 - Include line **+ column** in links to source, and move cursor to that column on click.
 - Change mouse cursor when hovering a Console link to indicate when clickable.
 - Getting our script evaluation off of the main GUI thread, which I left a `FIXME` comment for.  Then we should be able to remove the `MainWindow::processEvents` call and probably even speed things up further.
 - A console message after suppressed duplicate warnings to indicate how many were suppressed.
 - Idea for debugging utility: associate a Context to each Error Log message, which could be inspected (displayed in tooltip maybe) to show local variable values at time of warning.
 - Buffer and throttle Error Log updates in same manner as Console does in this PR.  Probably not as useful especially if a max limit is implemented.

MichaelAtOz:
Would it be difficult to make `maximumBlockCount` a command-line option?
`--Debug` is very limited with 5000.

thehans:
It shouldn't be much work to make it configurable.  Although I think it would be better to put in the Preferences menu somewhere, since its GUI specific, and wouldn't make sense for ex. on any `-o` command line which runs without gui.

Also, its *even faster* when you set `maximumBlockCount` to 0 (unlimited), since Qt takes some time to remove the old lines on each update.
~7.5s vs ~12.5s for the first example of echos without warnings.  The only cost is some memory usage but its not too crazy: ~228MB, vs ~52MB with default 5000.

t-paul (approved):
Nice, that's a massive speed improvement.

MichaelAtOz:
~1,000,000 echo's
Visible
```
2019.05 1492s
RC2     1603s
RC5      440s (no live echo's, gets Not Responding) 
PR        85.4s
```
~5 - 18.75 x faster

Hidden
```
2019.05  454s
RC2      404s
RC5      340s (no live echo's, gets Not Responding) 
PR        81s
```
~4 - 5 x faster

PR #3608, opened 2021-01-12
Fix for issue 3421

thehans:
This is a minimal fix for #3421, which I decided to separate from #3602 since those are larger changes which there isn't much time to test before release.
Also fixed a segfault which was noticed during testing of the console logging (close app during long/infinite recursion echos).


MichaelAtOz:
Does that segfault fix cover part of #1251?

t-paul:
No, I don't think this one is covered by any of the points mentioned in #1251.

t-paul (approved):

PR #3623, opened 2021-01-19
Add GUI Console font to Preferences.

thehans:
Also changed appearance of some other preferences on advanced tab, to not stretch fill the entire width.

BEFORE:
Screenshot from 2021-01-19 07-40-03

AFTER:
Screenshot from 2021-01-19 07-39-10

### DIFF
--- a/src/Console.cc
+++ b/src/Console.cc
@@ -83,6 +83,10 @@ void Console::addHtml(const QString& html)
 	this->setTextCursor(this->appendCursor);
 }
 
+void Console::setFont(const QString &fontFamily, uint ptSize) {
+	this->document()->setDefaultFont(QFont(fontFamily, ptSize));
+}
+
 void Console::update()
 {
 	// Faster to ignore block count until group of messages are done inserting.
--- a/src/Console.h
+++ b/src/Console.h
@@ -81,5 +81,6 @@ public slots:
 	void actionClearConsole_triggered();
 	void actionSaveAs_triggered();
 	void hyperlinkClicked(const QString& loc);
+	void setFont(const QString &fontFamily, uint ptSize);
 	void update();
 };
--- a/src/Preferences.cc
+++ b/src/Preferences.cc
@@ -111,11 +111,16 @@ void Preferences::init() {
 	// Editor pane
 	// Setup default font (Try to use a nice monospace font)
 	const QFont font = QFontDatabase::systemFont(QFontDatabase::FixedFont);
-
 	const QString found_family{QFontInfo{font}.family()};
 	this->defaultmap["editor/fontfamily"] = found_family;
  	this->defaultmap["editor/fontsize"] = 12;
 	this->defaultmap["editor/syntaxhighlight"] = "For Light Background";
+
+	// Leave Console font with default if user has not chosen another.
+	const QFont font2 = QFontDatabase::systemFont(QFontDatabase::GeneralFont);
+	const QString found_family2{QFontInfo{font2}.family()};
+	this->defaultmap["advanced/consoleFontFamily"] = found_family2;
+	this->defaultmap["advanced/consoleFontSize"] = 10;
 
 #if defined (Q_OS_MAC)
 	this->defaultmap["editor/ctrlmousewheelzoom"] = false;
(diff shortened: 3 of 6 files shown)


t-paul (approved):
I'm pretty sure the formatting was intentional, but the new version looks good to me too, so lets just go with that :smile:.

thehans:
> I'm pretty sure the formatting was intentional, but the new version looks good to me too, so lets just go with that .

OK, thanks.  The stretch just felt a little excessive to me.  And I was comparing it with Editor tab which used added spacers in a similar way (although I didn't go to full grid layout like most of the groups on that tab).

MichaelAtOz:
Note in the above image, the middle is Sans-serif, left is Serif and doesn't look like a scalable font (TT),
as the preference 'stored' family comes from QFontInfo, ie
```
	const QFont font2 = QFontDatabase::systemFont(QFontDatabase::GeneralFont);
	const QString found_family2{QFontInfo{font2}.family()};
```
It isn't doing a good job of matching - in that case, as I said, I don't know how it got set.

After deleting the registry entries for console fonts, rerun a dev version, it looks like size 10, preferences shows it selected `MS Shell Dlg 2`, a system font, not in the normal font list.

Using that font and visually matching to 2019.05:
Capture-2019-font
Looks like it is that internal font @ 8pts.

That use of internal fonts was OK in 2019, but externalising them to the drop-down in prefs is probably not the way to go?
I wonder what the family was prior to QFontInfo selecting the 'instantiated' internal font?

MichaelAtOz:
Having gone thru the list it looks like Tahoma 8pt, matches 2019.05 & 2021.01 console (on W7 & a W10-VM).

MichaelAtOz:
Right...
```
QFont QFontDatabase::systemFont(QFontDatabase::SystemFont type)
Returns the most adequate font for a given type case for proper integration with the system's look and feel.
```

Capture-windows-display-settings

I changed that to 10pt, expecting the 2021.01 console to match it, no.
But when I went to prefs, the drop-down says 'Tahoma'. So I suspect that Window Display setting is the source of `::SystemFont`.

thehans:
> I changed that to 10pt, expecting the 2021.01 console to match it, no.

If you changed the font as your screenshot shows, then you only affected the font of an "Active Title Bar" item.  The item settings you should change for comparison is the "Window Text".
Also, all that code dealing with SystemFont etc only applies to choosing a default for the *first time* you run an OpenSCAD version that has user configurable console font.
After you close and re-open OpenSCAD, it will have saved that setting/preference to a config file (I think path is something like `%APPDATA%\Local\OpenSCAD\...` on Windows?) 
If you find this file you can open it in notepad and look for the lines which set `consoleFontFamily` and `consoleFontSize`.  Delete those before running OpenSCAD again if you want to check how your theme settings apply.

MichaelAtOz:
Just for the record.

> The item settings you should change for comparison is the "Window Text".

I went through all the entries in the 'Item:' drop-down which have a font setting.
They are Active Title Bar, Icon, Inactive Title Bar, Menu, Message Box, Palette Title, Selected Items, ToolTip.
There is no 'Window Text' item. None of them affected the console text in 2019.05.
BTW all such font setting were Tahoma.

Most Windows setting are in the Registry, in this case at `HKEY_CURRENT_USER\Software\OpenSCAD`.
Under `HKEY_CURRENT_USER\Software\OpenSCAD\OpenSCAD\advanced` is where those prefs are located.

----

So when I delete those two prefs, run a dev version (2021.02.19 - fast CGAL), 
edit/preferences (without scroll-ring), the font displayed is 'MS Shell Dlg 2' size 10.

That font is not in the normal font list. It looks visually the same as Tahoma.

This is what I meant by

> That use of internal fonts was OK in 2019, but externalising them to the drop-down in prefs is probably not the way to go?

```
	// Leave Console font with default if user has not chosen another.
	const QFont font2 = QFontDatabase::systemFont(QFontDatabase::GeneralFont);
	const QString found_family2{QFontInfo{font2}.family()};
	this->defaultmap["advanced/consoleFontFamily"] = found_family2;
	this->defaultmap["advanced/consoleFontSize"] = 10;
```
`QFontInfo`
> The QFontInfo class provides the same access functions as QFont, e.g. family(), pointSize(), italic(), weight(), fixedPitch(), styleHint() etc. But whilst the QFont access functions return the values that were set, a QFontInfo object returns the values that apply to the font that will actually be used to draw the text.
>
> For example, when the program asks for a 25pt Courier font on a machine that has a non-scalable 24pt Courier font, QFont will (normally) use the 24pt Courier for rendering. In this case, QFont::pointSize() returns 25 and QFontInfo::pointSize() returns 24.

Hence `found_family2`; my conjecture is `::GeneralFont` is probably returning 'Tahoma', but `QFontInfo` will be whatever best match is used internally, ie 'MS Shell Dlg 2'.

Perhaps
```
void Console::setFont(const QString &fontFamily, uint ptSize) {
	this->document()->setDefaultFont(QFont(fontFamily, ptSize));
}
```

Should use `QFont_Info` (or maybe `setDefaultFont` does that internally) while `Preferences::init()` should use `QFont`, so it doesn't externalise the matched font.
I suspect this was not noticed in the editor prefs as `QFontDatabase::FixedFont` has limited or better matches.

thehans:
> @thehans I think you missed my first line, to be more specific,
> line 147 `if (static_cast<uint>(size) == savedsize) {` should use `consavedsize`.
> It made my console 14pt from editor settings.

Sorry @MichaelAtOz, I did see it and was planning to fix it, just hadn't gotten around to it.
But now I have, plus actually correct default font detection in #3696

In case you want to verify that it picks up the font in the same way as 2019.05, see my comment there about using Qt5 Configuration Tool to set alternate "General Font".  Not sure if Qt5 under Windows uses the same tool for such settings though.

PR #3636, opened 2021-01-29
Optimize disjointed unions & differences (+ introduce caching of costly nef<->polyset conversions)

ochafik:
Hi all!

This PR introduces a `fast-union` feature that avoids using Nef polyhedra for unions of "clearly disjoint" solids (as per their respective bboxes). It simply concatenates PolySets in that case (converting from Nef polyhedra if needed), which is much quicker (Nef joins themselves are very slow, but creating Nef polyhedra can be very slow too at the first place). It plays well with lazy-union (and some upgrades to it I'm proposing in https://github.com/openscad/openscad/pull/3637).

*Edit*: As suggested below, now also testing for fast-unionability of subtracted terms in differences (`fast-difference` feature).

It also makes full use of the caches for Nef polyhedra <-> polysets conversions in the process.

As it turns out, spheres love caching: the following example now renders in 5 sec (instead of 4min30sec; will add more examples [here](https://gist.github.com/ochafik/2db96400e3c1f73558fcede990b8a355)):

```js
// openscad --enable=fast-union test.scad -o test.stl

N = 5;

// When this is true, we end up with lots of overlaps.
// Rendering is still faster than before thanks to the caching of the spheres,
// but the true benefits will come from the reordering of unioned bodies, see details below.
overlap = false;

// A simple grid of smooth spheres
union() {
  for (i=[0:N-1], j=[0:N-1])
    translate([i, j, 0])
      sphere(d=(overlap ? 1.1 : 0.9), $fn=50);
}
```

(less dramatic, but this [brick wall](https://gist.github.com/deckar01/ef11def51de7e71d9f288c6e5819fdb7) benchmark's rendering time is almost cut in half)

Some implementation comments: (suggested entry point for review: CGALUtils::applyUnion3DFast):

- Added a `BoundingBoxes` helper, which allows testing a new solid's bbox against all the previously unioned solids' bounding boxes in the union (allows optimizing non-overlapping grid patterns). It currently does a crude linear search on those bounding bboxes (could speed up with some smart CGAL structure but this isn't the new bottleneck anyway)

- Added a `LazyGeometry` helper, which does nef ↔ polyset conversions on demand, provides their bboxes and does disjointed unions or normal ones (using / filling the cache in the process). Because of some pathological cases, even disjointed unions may fall back to Nef polyhedra operations (as a result, need to keep the original polysets, which prevents in-place concatenation: copying and checking increasingly large PolySets over can become the bottleneck in large grids of objects, for instance the example above with N=10). May want to revisit the costly detection of such cases + fallback strategy (edit: added an optional `-DOPTIMISTIC_FAST_UNION` define to disable these checks: the spheres then render in 700ms).

- The order of unions matters now even more than ever. Currently I'm only using the order of the children, but I plan to find "easy" groups of disjoint operands that can be unioned as polysets (say, looking at morton invervals overlaps of the operands' bounding boxes), to minimize the number of Nef polyhedra built and operated on. Much potential for parallelism here too.

- I had to switch some APIs to return shared_ptr and propagate the changes, hence a bit of diff noise, sorry.

- There are other cache-unaware code paths (e.g. intersection / difference / minkowski) that I'd be happy to tackle in a follow up.

Managed to get all tests green locally (macosx), except for differences-tests for which I had to do some compromise.

Welcoming comments & suggestions / happy to refactor.


thehans:
Seems like a very good ideal overall.  I get substantial speedup for the sample script with `overlap=false;`
3.5s  vs  82s with fast-union off, about 23x speedup.

And just for fun, I wrote a variation of your script which is more cache-friendly (it makes the inner loop / entire columns of spheres cache-able)
```OpenSCAD
union() {
  for (i=[0:N-1]) translate([i,0,0])
    for(j=[0:N-1]) translate([0,j,0])
      sphere(d=(overlap ? 1.1 : 0.9), $fn=50);
}
```
That gets my render time down to 0.732s, which is outstanding!

...However, going back to your original script and changing to `overlap=true;`, my testing actually shows that fast-union is significantly slower than before:
fast-union OFF: Total rendering time: 0:01:30.999
fast-union ON:   Total rendering time: 0:03:08.003
So just over 2x *slower* with fast-union enabled.  
And I similar speeds of ~1m30s on latest nightly snapshot, and even 2019.05 release as well.

> `// Rendering is still faster than before thanks to the caching of the spheres,`

Since I can't reproduce any speedup, I'm curious what you mean about caching spheres?  I haven't had a chance to look over the whole PR yet, but spheres should have already gone into Geometry Cache, no different from other Polysets.


And on the topic of prioritizing order of operations, I don't know if you've seen it, but the discussion in #1234 and my subsequent PR #3121 are maybe worth looking at.

As a quick example of why combining least-complex-geometries first, is a reasonable heuristic:
Say we have 8 objects A,B,C...H, which are all disjoint and have the same complexity.
The main cost here is basically memory accesses to read input operands, and write the output.
Since these all have same complexity, lets call it 1 Standardized Memory Access Cost Unit (SMACU)

If we just union them in order: ((((((A+B)+C)+D)+E)+F)+G)+H
Then each subsequent operation costs increasingly more:
1+1 (read) = 2 (write)    (4 total SMACUs for this op)
2+1 = 3   (6 SMACUs)
3+1 = 4   (8 SMACUs)
...
7+1 = 8   (16 SMACUs)
In the end we get (4+6+8+10+12+14+16) = **Grand Total 70 SMACUs**

If we instead pick the lowest complexity pair between inputs *and* intermediate results at each step, then we end up looking more like this:  ((A+B)+(C+D))+((E+F)+(G+H))
1+1=2 (4)
1+1=2 (4)
1+1=2 (4)
1+1=2 (4)
2+2 = 4 (8)
2+2 = 4 (8)
4+4 = 8 (16)
(4+4+4+4+8+8+16) = **Grand Total 48 SMACUs**

Differences would be even more dramatic if for example the first object A had a cost of 100, while the remaining objects stayed at cost = 1.

Anyways, using the "Morton" [Z-Order curve ](https://en.wikipedia.org/wiki/Z-order_curve) you mentioned (i had to look it up)  seems promising as well.  I had also thought about mapping bounds or centroid to a position on a 3D [Hilbert Curve](https://en.wikipedia.org/wiki/Hilbert_curve) to facilitate sorting by proximity.  More or less the same idea, just using a different space filling curve.

ochafik:
> Seems like a very good ideal overall. I get substantial speedup [...]

Thanks for trying things out!

> variation of your script which is more cache-friendly

Hah, neat trick!

> what you mean about caching spheres? [...] spheres should have already gone into Geometry Cache, no different from other Polysets.

Sorry I was a bit cryptic: the spheres' PolySets are indeed in the geometry cache, but when unioning N of them we convert each one to a Nef polyhedron independently (`createNefPolyhedronFromGeometry`  in `applyUnion3D`) and don't cache that Nef. When I ran with $fn=100 (*that* is where I saw the speedup), each conversion alone was taking 7 seconds on my laptop.

The questionably-named LazyGeometry helper provides a transparent on-demand conversion & cache behaviour and I think could potentially be used beyond the union case.

> I don't know if you've seen it, but the discussion in #1234 and my subsequent PR #3121 are maybe worth looking at.

I hadn't, thanks for the great pointers! I hadn't understood the extent of the problem that queue was solving, thanks for elaborating. I wonder about the actual complexity of Nef unions, I feel like there might be some log complexity lookups (do they use some bounding box hierarchy of faces to compute intersections?) that would make the costs even less linear?

I'm working on refactoring the code to preserve the queue logic ("almost" done), will ping this thread when it's ready (moving the BoundingBoxes object inside the LazyGeometry, lots of the code should stay the same).

> pick the lowest complexity pair between inputs and intermediate results at each step

I'd assume the most efficient way to multithread this would be work-stealing from the same queue with a proper lock? 

> mapping bounds or centroid to a position on a 3D Hilbert Curve to facilitate sorting by proximity. 

I haven't played with Hilbert's curve yet, could be a better pick (thanks for adding the links ;-)). In any case, overlaps of the 1D intervals corresponding to a bounding box in the curve space may not result in actual overlaps of the 3D bounding boxes, so will still need some tests on the bboxes but it should make it more likely for the heuristic to find good non-overlaps in some bounded time. Can't wait to give it a try :-)

ochafik:
> more cache-friendly (it makes the inner loop / entire columns of spheres cache-able)

Oh, and this opens an entirely new can of worms. A rendering tree optimizer might be able to detect repeating subpatterns and reproduce this manual optimization... Something I'd love to explore (see my ramblings in https://github.com/openscad/openscad/pull/3637 about tree rewrites).

ochafik:
I've retrofitted the original queue logic and did see sensible improvements, although haven't done thorough benchmarks yet.

thehans:
> > what you mean about caching spheres? [...] spheres should have already gone into Geometry Cache, no different from other Polysets.
> 
> Sorry I was a bit cryptic: the spheres' PolySets are indeed in the geometry cache, but when unioning N of them we convert each one to a Nef polyhedron independently (`createNefPolyhedronFromGeometry` in `applyUnion3D`) and don't cache that Nef. When I ran with $fn=100 (_that_ is where I saw the speedup), each conversion alone was taking 7 seconds on my laptop.

~~I tried again with $fn=100 and i still get 2x slowdown:~~
```
N=5, overlap=true, $fn=100
nightly:    Total rendering time: 0:06:31.726
fast-union: Total rendering time: 0:12:57.206
```
**Edit:** Took me a while from starting this reply to finishing it, and in the meantime I pulled your latest changes and retested.
I am seeing some speedup now even with `overlap = true;`, so that's good.
```
N=5, overlap=true, $fn=50
nightly:    Total rendering time: 0:01:32.899
fast-union  Total rendering time: 0:01:17.869
```

> The questionably-named LazyGeometry helper provides a transparent on-demand conversion & cache behaviour and I think could potentially be used beyond the union case.

OK, this sounds a bit similar to the `smartCache*` functions in GeometryEvaluator.  It is supposed to handle getting from and inserting to the proper cache: GeometryCache vs CGALCache, but I think in many ways it is suboptimal, and not particularly "smart" in how it operates.  Maybe the LazyGeometry functions can be used to replace this instead?  
I definitely would prefer more general solutions that can be applied in multiple places over having tons of niche classes for every little thing.

> I wonder about the actual complexity of Nef unions, I feel like there might be some log complexity lookups (do they use some bounding box hierarchy of faces to compute intersections?) that would make the costs even less linear?

I don't know precisely, but I'm pretty sure its very close to linear time by total number of faces.  IIRC @rcolyer helped show this to me a while ago by writing some scripts and benching them at various complexities, maybe around when I was writing PR 3121.

> > pick the lowest complexity pair between inputs and intermediate results at each step
> 
> I'd assume the most efficient way to multithread this would be work-stealing from the same queue with a proper lock?

I don't know much about work stealing, but i guess that implies each thread would have maintain its own separate queue of work to be stolen from?  I think it would be better to have a centralized queue so that ordering can be applied across threads.  So I would probably start with a basic design where all threads share a single priority queue and just pull from and push to that (with locking), something like:
1) lock priority queue
2) if worker has a result (from previous loop iteration) then insert onto the queue
3) if (queue.size < 2) { release lock; break; }
4) pop next two objects
5) release lock
6) do work
7) loop to 1)

But yes proper locking will be crucial for *any* shared resources (and there's a lot of static / global data around the codebase, for various geometry caches, stat cache, module cache, stackchecker, progress bar, console output handlers etc.  which will require careful consideration).  There has been some talk, about working towards minimizing such global state wherever possible so that its more compatible with multi-threading.

> I haven't played with Hilbert's curve yet, could be a better pick (thanks for adding the links ;-)). In any case, overlaps of the 1D intervals corresponding to a bounding box in the curve space may not result in actual overlaps of the 3D bounding boxes, so will still need some tests on the bboxes but it should make it more likely for the heuristic to find good non-overlaps in some bounded time. Can't wait to give it a try :-)

Another idea is maybe to just calculate volume of bbox intersections.
If you assume uniformly distributed vertices/faces, then you could estimate that the number of faces removed by a union would be roughly:
`NumFaces(A) * Volume( bbox(A) ⋂ bbox(B) ) / Volume( bbox(A) ) + NumFaces(B) * Volume( bbox(A) ⋂ bbox(B) ) / Volume( bbox(B) )`

I don't think this would be compatible in any way with the priority queue as-is, since that only calculates cost of individual geometries and not pairs of them.
Evaluating all possible pairs would also be a O(n^2) operation for n geometries, which might be too much overhead to do in-between every operation, maybe there could be some threshold of face count, below which the code doesn't bother to find optimal bounding box pairs etc, and falls back to priority queue.  For example if a script is only doing union on a huge number axis aligned cubes, I would think checking bounding boxes etc could be as much overhead or more compared to just doing the operation itself.

Its hard to predict the impact (positive or negative) of various optimization attempts without just implementing and measuring them directly.

thehans:
> Oh, and this opens an entirely new can of worms. A rendering tree optimizer might be able to detect repeating subpatterns and reproduce this manual optimization... 
> (Also, I tried caching intermediate results to see if it could somehow approximate your manual optimization: it resulted in lots of cache evictions, so not a free pass just yet; left it out of the last batch of commits)

I don't think this type of optimization will be possible, due to the way caching works.  Cache keys are basically just the substring / subtree of the CSG Tree which you see from "Design -> Display CSG Tree..."

The manual optimization works by splitting the translate into two separate operations.  So the inner loop gets repeated verbatim, as a child of the outer translate.  Without splitting up the translates, then each one will always be unique, so it would be pretty tough I think to match patterns like that, and even if something was devised it would probably have to be specific to grid-like translations, which is a niche solution.

Also, I will probably expand some more in your other PR eventually, but keep in mind that `CSGTreeNormalizer` already basically converts to a sum of products form.  You can see `CSGTreeNormalizer::match_and_replace` for the replacement rules etc.  
But this is only used for preview and not tied to CGAL evaluation.  The OpenCSG rendering library *requires* that the CSG terms be in this form, but I have doubts that it would be helpful for CGAL due to the way that the number of terms tend to multiply to quite large values for certain combinations of ops.

ochafik:
> I am seeing some speedup now

Yeah for me it only seems significant with large $fn values (i.e. when nefs are costly to create)

>> The questionably-named LazyGeometry helper provides a transparent on-demand conversion & cache behaviour and I think could potentially be used beyond the union case.

> OK, this sounds a bit similar to the smartCache* functions in GeometryEvaluator. It is supposed to handle getting from and inserting to the proper cache: GeometryCache vs CGALCache, but I think in many ways it is suboptimal, and not particularly "smart" in how it operates. Maybe the LazyGeometry functions can be used to replace this instead? I definitely would prefer more general solutions that can be applied in multiple places over having tons of niche classes for every little thing.

I've seen it, and agree! I've split this logic to a mixed_cache.h file as a starting point (with a meta-templated `getOrSet` that picks the right cache(s) for the asked type). Would it be fine if I attempted to unify cache access in a subsequent PR? (this one is already getting large)

> So I would probably start with a basic design where all threads share a single priority queue and just pull from and push to that (with locking), something like [...]

Glad we're on the same page on this.

I've actually taken a stab at this a few weeks ago: [parallel_reduce.h](https://github.com/ochafik/openscad/blob/faster-integ-1/src/parallel_reduce.h). I wasn't using the priority queue mechanism yet, and the rest of that branch would need rethink or refresh. It does naive parallelism using futures: I essentially swapped `shared_ptr<const Geometry>` for `lazy_ptr<const Geometry>`, which is a drop-in replacement that blocks when the actual geom is dereferenced ([lazy_ptr.h](https://github.com/ochafik/openscad/blob/faster-integ-1/src/lazy_ptr.h), pardon my clunky C++). I unlocked a few blocking calls (i.e. in caching code or everywhere that wants to know about the actual geometry's class or dimension) and make a few calls async (operations, transforms), and despite being globally thread-unsafe, I still seemed very stable (prolly didn't try to crash it hard enough).

My take away from that experiment was that unions were still too slow even with multithreading (and the evaluation tree is too deep, which serializes lots of operations), hence the lazy union extravaganza from https://github.com/openscad/openscad/pull/3637 (to flatten the tree) and now this fast union route.

FYI [here's the real-world model](https://gist.github.com/ochafik/6d19e3b131153305b44ea275d5d930e3) that motivated this month of hacking: it renders 20x faster with all my (monothreaded) features combined.

BTW: would you be open to having some benchmark model collection + scripts? (I'm sure there's cool tools to use?) I'd be happy to contribute.

> [...] calculate volume of bbox intersections. [...]

Interesting idea! (even if we don't apply it here, could help with the cost function of some global optimizer... thinking out loud, but imagine if we could build a gross, octree-based approximation of the model with just the leaf #facet information? we might have wild estimates of the number of vertices in each quadrant after CSG operations, which could help rule out some global tree rewrites)

> Evaluating all possible pairs would also be a O(n^2) [...] some threshold [...] Its hard to predict the impact (positive or negative) of various optimization attempts without just implementing and measuring them directly.

So... I couldn't resist, I did a crude heuristic that tries to find pair-wise disjoint geometries in the following neigbhours, examining up to K geometries unsuccessfully before calling it a day. But that K allowance is reset if it gathered something in its snowball. Since K is a constant, the search isn't quadratic, but I'm too lazy to find out if how much more than linear it actually is.

I ran a few tests and it cuts the N=5 overlap=true example by a sizeable amount ([full test results + code](https://gist.github.com/ochafik/8aa845c8b9ee3a241e771dda99a2d5ef)): the single loop case is 2x faster than the base line, your nested loops trick is 35% faster than baseline.

The code is a bit crude so I'll work on it it to ensure we fast-union the solids in the right order before I commit anything, but I'm pleased :-D

Btw, added some copyright headers to the new files as per my employer's wishes (I'm doing all this just for fun, though), please let me know if that's an issue.

>> caching intermediate results to see if it could somehow approximate your manual optimization: it resulted in lots of cache evictions, so not a free pass just yet; left it out of the last batch of commits)

>I don't think this type of optimization will be possible, due to the way caching works. Cache keys are basically just the substring / subtree of the CSG Tree which you see from "Design -> Display CSG Tree..."

What I did is to retain a sorted set of the cache keys of the geometries that have been unioned in the LazyGeometry, so that I can create a semantic key (`"union(){" + join(componentKeys) + "}`) that's order invariant (order of children and/or operations, e.g. same key for `union{ a; union { b; c; } }` as for `union{ a; b; c; }` or for  `union{ c; b; a; }`, especially if we flatten the unions with the other PR). An issue though is the priority queue would need to dynamically prioritize already cached subsets of the operands (should be possible?)

I haven't looked enough at the cache size issue but maybe with persistent caching it could become okay to store all that extra data?

ochafik:
>> Maybe the LazyGeometry functions can be used to replace this instead? I definitely would prefer more general solutions that can be applied in multiple places over having tons of niche classes for every little thing.

> I've seen it, and agree! I've split this logic to a mixed_cache.h file as a starting point

(oh, I think I now see what you mean, did you suggest also moving the on-demand conversion logic to that centralized cache facade? lots more code could get speed ups with that, e.g. intersections / differences)

src/bounding_boxes.h
```diff
@@ -1,3 +1,17 @@
+/*
+ * OpenSCAD (www.openscad.org)
+ * Copyright 2021 Google LLC
+ *
+ * This program is free software; you can redistribute it and/or
```
t-paul, line 5: GPL2-only does not work for OpenSCAD, while the OpenSCAD code itself is GPL2+, CGAL which is a very core part of the application is [GPL3+](https://github.com/CGAL/cgal/blob/master/Installation/LICENSE), so we need the GPL2+ `(or at your option) any later version` definition to build an application that can be distributed.

rcolyer: Hi Olivier,

```
 * OpenSCAD (www.openscad.org)
 * Copyright 2021 Google LLC
```
I would also note that this part is a little confusing, since under the standard structure for copyright notices, it looks like Google wrote OpenSCAD (which of course was not your intention).  Of course, Google via you is a contributor now, and it's perfectly reasonable to note that to satisfy your employer's conditions on your flex time.  To prepare ahead for the new files being modified by other contributors after being integrated, and to include the GPL2+ status without being wordy, perhaps a good notice is something like:

```
// Portions of this file are Copyright 2021 Google LLC, and licensed under GPL2+.  See COPYING.
```
Do you think this would satisfy it adequately?

ochafik: Oops good to know, I should have checked, now fixed.

ochafik: @rcolyer sorry, replied before seeing your comment. I'll reword to something less ambiguous indeed!

ochafik: Used your suggestion verbatim :-)

thehans:
> I think could potentially be used beyond the union case 
> ...
> (oh, I think I now see what you mean, did you suggest also moving the on-demand conversion logic to that centralized cache facade? lots more code could get speed ups with that, e.g. intersections / differences)

Yeah, I guess I was just meaning that it would be a shame if on demand conversion code ended up *not* being used beyond the union case, and that it sounds like there's now two places where we have some code choosing between two caches in one way or another.  Definitely needs some unified solution, I had at one point looked into making a single GeometryCache which took take a tuple of shared_ptrs for the different types, not sure if that's ideal either, but it was one idea.  

Just advocating the DRY principle basically, after all I'm quite proud that my sum contributions to OpenSCAD is in the negative LoC  :D

> What I did is to retain a sorted set of the cache keys of the geometries that have been unioned in the LazyGeometry, so that I can create a semantic key (`"union(){" + join(componentKeys) + "}`) that's order invariant (order of children and/or operations, e.g. same key for `union{ a; union { b; c; } }` as for `union{ a; b; c; }` or for `union{ c; b; a; }`, especially if we flatten the unions with the other PR). An issue though is the priority queue would need to dynamically prioritize already cached subsets of the operands (should be possible?)

I see, that's clever and maybe useful in some cases, but I'm not sure.  I still don't see it helping with the original script's `translate([i,j,0]) ...;`  because that single loop's `translate`**\*** never has any repeating values for the **combined** `[i,j,0]` ... every single translate is unique.  Does that make sense?

**\*** Well its technically a `multmatrix` node, after being formatted as CSG Tree text.

> I haven't looked enough at the cache size issue but maybe with persistent caching it could become okay to store all that extra data?

If there is still some use case for caching intermediate results like that, then maybe instead of appending disjoint geometries to `Polysets`, we could just use`GeometryList` (or maybe a new derived `DisjointGeometryList`?) so that only the `shared_ptr`s are copied.

As an aside, if you are interested in a more immediate / informal discussion, I'd invite you to join the IRC channel: #openscad on irc.freenode.net
The three of us: t-paul, rcolyer, and I are pretty much on there 24/7 along with some other regular users, and happy to discuss new ideas for improving the project, help with scad scripts, etc.

ochafik:
Ok, so I'm currently reworking the code to be faster, and less complex.

My current approach is to do an analysis phase where I partition union operands into clusters of mutually non-overlapping geometries (no LazyGeometry class anymore). Some of these clusters will be of size 1. Then I reduce each of these non-overlapping clusters by appending in-place to a PolySet (avoids many copies, especially in pathological cases with many small geometries like. At the end of each cluster reduction I do a single check of integrity and abort that cluster if it's not closed or valid ("explode" it to clusters of size 1 for subsequent normal union), to amortize the costs of integrity checks (which should rarely fail anyway?). The resulting clusters of size 1 are slow-unioned as usual.

I've got a working proof of concept (using [CGAL::spatial_sort](https://doc.cgal.org/latest/Spatial_sorting/index.html) and [CGAL::Union_find](https://doc.cgal.org/latest/Miscellany/classCGAL_1_1Union__find.html)) and I'm working on switching to [CGAL::box_intersection_d](https://doc.cgal.org/latest/Box_intersection_d/index.html#chapterBoxIntersection) for faster bbox tests.

All this to say, please hold off before reviewing too much of this as I hope the code will become smaller (will also try and merge some of the LazyGeometry methods with existing cache helpers as suggested by @thehans ).

ochafik:
So, that algorithm wasn't such a good idea after all, it did find large patches of fast-unionable polysets and that bit was fast, but if there were any overlaps left it ended up unioning much larger nef polyhedra at the end, making things slower for some large models.

I've now pursued a different route, which is to use the

Issue #481, opened 2013-09-17
CGAL excess memory usage (EPEC kernel)


MichaelAtOz:
With complex models (a lithophane) 2013.08.13 uses huge amounts of memory compared to 2013.06. (also just checked, simple models also uses more memory but less noticeable side effects, e.g. 187KB v's 660KB)

Both 64bit versions.

2013.06 Commit size for F6 went to max ~850KB
2013.08.13 >4GB (with system paging and slowness left & right)

The F6 out of 1000 progress indicator & memory:

```
        2013.06     2013.08.13
F6/1000       KB        KB
   0     365,000     692,000 
  47     507,000     1,700,000 
  95     615,000     3,350,000 
 238     752,000     4,080,000 
 904     845,000     4,150,000 
 952     611,000     4,215,000 
1000     622,000     4,248,000 
```

F6 Stats are the same (except time)

```
2013.06
Module cache size: 0 modules
Compiling design (CSG Tree generation)...
Rendering Polygon Mesh using CGAL...
PolySets in cache: 9
PolySet cache size in bytes: 2602936
CGAL Polyhedrons in cache: 8
CGAL cache size in bytes: 89615496
   Top level object is a 3D object:
   Simple:        yes
   Valid:         yes
   Vertices:    12876
   Halfedges:   72232
   Edges:       36116
   Halffacets:  46488
   Facets:      23244
   Volumes:         2
Total rendering time: 0 hours, 2 minutes, 18 seconds
Rendering finished.

2013.08.13
Compiling design (CSG Tree generation)...
Rendering Polygon Mesh using CGAL...
PolySets in cache: 9
PolySet cache size in bytes: 2602936
CGAL Polyhedrons in cache: 8
CGAL cache size in bytes: 89615496
   Top level object is a 3D object:
   Simple:        yes
   Valid:         yes
   Vertices:    12876
   Halfedges:   72232
   Edges:       36116
   Halffacets:  46488
   Facets:      23244
   Volumes:         2
Total rendering time: 0 hours, 20 minutes, 37 seconds
Rendering finished.
```

PR #3641, opened 2021-02-02
Blazingly fast rendering using Polygon Mesh Processing when possible

ochafik:
Experimental `fast-csg` feature for 10x faster rendering (or more... or less :-D) of lots of challenging models (see [some benchmarks here](https://gist.github.com/ochafik/2db96400e3c1f73558fcede990b8a355))!

This PR introduces the class `CGALHybridPolyhedron`, a lazy hybridization of [CGAL::Polyhedron_3](https://doc.cgal.org/latest/Polyhedron/index.html) / [CGAL::Surface_mesh](https://doc.cgal.org/latest/Surface_mesh/index.html) (using [corefinement functions](https://doc.cgal.org/latest/Polygon_mesh_processing/index.html#Coref_section) for CSG ops) and [CGAL::Nef_polyhedron_3](https://doc.cgal.org/latest/Nef_3/index.html) (as a fallback for cases not supported by corefinement functions).

As mentioned by @thehans, CGAL's [Polygon Mesh Processing](https://doc.cgal.org/latest/Polygon_mesh_processing/index.html) libraries (PMP) provide much faster CSG operations (union, intersection, difference) than the Nef polyhedron. Unfortunately, they also have strict input requirements: any shared ~~vertices~~ edges between two unioned polyhedra or mis-oriented faces and they just bail out.

This PR has being chunked and sent as separate PRs, now available in nightly builds. This [blog post](https://ochafik.com/jekyll/update/2022/02/09/openscad-fast-csg-contibution.html) summarizes the process and some next steps:

- [x] https://github.com/openscad/mxe/pull/4 will make CGAL 5.2 available in the OpenSCAD's mxe environment: I'll then clean up the code that switches mxe build setups to using my branch.
- [x] Send off some separate PRs for some of the non-feature changes (shared pointers, slight utils refactorings) -> [PR#3961](https://github.com/openscad/openscad/pull/3961), [PR#4082](https://github.com/openscad/openscad/pull/4082), [PR#4083](https://github.com/openscad/openscad/pull/4083)
- [x] A good squashing of changes -> [PR#4087](https://github.com/openscad/openscad/pull/4087) and future follow ups
- [ ] FAST_CSG_SAME_KERNEL, still needs work to make it actually faster (pick a good kernel to rule them all, ideally not a lazy one)
- [x] `fast-csg-exact-callbacks` feature to fix dramatically pathological cases like [this one](https://github.com/openscad/openscad/pull/3641#issuecomment-782550842) -> [PR#4099](https://github.com/openscad/openscad/pull/4099) (includes `fast-csg-debug-corefinement` feature to dump operands of failing operations to `.off` files to help create self-contained bugs for CGAL)
- [x] `fast-csg-trust-corefinement` feature to disable costly checks before calling corefinement methods (let them fail politely or not) --> [PR#4101](https://github.com/openscad/openscad/pull/4101)

Some notes:
- The grid is gone (it's still there for minkowski operations, though, which only indirectly use the new code paths).
- ~~Currently requires CGAL 5.1+ (because Polyhedron_3 + corefinement doesn't work well together before) but compiles fine with older versions (the `fast-csg` feature is then compiled away). A new github CI workflow builds and tests with various versions of CGAL (`fast-csg`-specific tests are disabled when its' not available). Now that there's a Surface_mesh codepath, might be able to lower the CGAL version requirement: stay tuned!~~
- Weird performance of Lazy numbers from Epeck kernel is mitigated by calling `CGAL::exact` as suggested by @sloriot in this thread. In general, calling that after transforms, corefinement operations and upon conversion from nef to polyhedron/surface mesh, but in the case of surface meshes I'm calling exact during corefinement after a face is done being split up (which creates new numbers) rather than after. This makes some rare models faster.
- The hybrid polyhedron uses a different kernel ([CGAL::Epeck](https://doc.cgal.org/latest/Kernel_d/structCGAL_1_1Epeck__d.html)) than the current nef. After CGAL 5.2, it's possible to use the same kernel, to saves many conversions  (and memory for the nef fallback cases). The switch is done automatically based on the version of CGAL. The 5.2.x behaviour (w/ x > 0) isn't tested in CI yet as the 5.2.x branch isn't building out of the box on Ubuntu 18.04 (boost version issue?) and CGAL 5.2.1 hasn't been released yet.
- Altered some signatures to return shared pointers
- Generalized a few existing utils so they can operate on the `CGAL::Epeck` kernel (which seems, with its inaccurate `Epick` sibling, to be the only kernels that work well with PMP's corefinement functions until after 5.2.0)
- Made sure to split the object files into many `cgalutils-*` so this builds fine in all environments with older GCC or limitations of RAM
- No more green display on faces that result from differences. Tests which expectations have rightly changed because of that are run with and without the feature so they can run with binaries that lack the feature.
- Using  [PMP::orient_to_bound_a_volume](https://doc.cgal.org/latest/Polygon_mesh_processing/group__PMP__orientation__grp.html#gafdbc256d06536f6aca2558bc1a0156e2) as PMP::orient can leave polyhedra corrupted (will attempt to file a reproductible bug to CGAL)
- The initial code adapted from https://github.com/openscad/openscad/pull/3636 was trying also detects fast-union cases by keeping track of the bounding boxes of unioned bodies (and then just concatenate disjoint polyhedra). It doesn't seem this beneficial with corefinement anymore, it had some annoying quadratic time search and needed reimplementing for surface meshes (Iso_cuboid3 sharing a single vertex don't seem to count as intersecting, but dumb concatenation of surface meshes doesn't work in that case: vertices need to be unique)

Future work:
- Simplify facets after corefinements once https://github.com/CGAL/cgal/pull/5461 lands (thanks @rcolyer for spotting increase, e.g. `cube(1); translate([0.5, 0, 0]) cube(1);` goes from 6 (12 triangles) to 44 ([before](https://imgur.com/f1M2dxy), [after](https://imgur.com/FJ3KFKI))
- Fix the few remaining tests that I couldn't enable fast-csg on (they're still run without the feature, see annotated list in `tests/CMakeLists.txt`), and generally address bug reports from users
- **Edit**: tried it, there's no point / that's not where the time is spent. ~~Have a version of minkowski with the new kernel (the "native" nef minkowski is already wired, but not the custom alternative from CGALUtils, which currently incurs unnecessary conversions)~~
- [x] Abandon `CGAL::Polyhedron_3` to only use `CGAL::Surface_mesh`. It's definitely faster!
- Wireframe rendering? Currently CGALHybridPolyhedron is converted to a PolySet at GeometryEvaluator's root.
- Retire `CGAL_Nef_polyhedron` and use a single kernel to rule them all
- CGALHybridPolyhedron::dump


thehans:
Linking to original issue #2360 for reference.

sloriot:
Hi, I'm the author of the corefinement code. If you have any question please do not hesitate to get in touch with me. In particular I'm a bit surprised by "one of them could just disappear: cube(1); translate([1, 0, 0]) cube(1)": this should not be an issue. You will have an issue if the cubes share only an edge or a vertex. In such a case the function will return false. Also the code should not leave the mesh structure invalid if the input are valid.

MichaelAtOz:
@ochafik thanks for this, great improvement. I've taken it for a drive, she's pretty sporty...

> Note that this seems to come at the cost of higher memory usage (maybe up to 2x).
> ...Generalized a few existing utils so they can operate on the CGAL::Epeck kernel 

There was a major memory growth problem with Epeck a long while ago (#481), got reverted.
But I'm not seeing higher memory, more the opposite, significantly. I'll post figures separately.

I've had a couple of crashes. 

The first was after setting the feature in the GUI, after I had done some renders, the GUI was doing strange things, I'm putting that down to feature disabled/enabled conflicts in the one invocation.

The second is repeatable. This .csg does it all the time.
[For Dyson Cinetic - v7-03-dodgy-2021.zip](https://github.com/openscad/openscad/files/5992824/For.Dyson.Cinetic.-.v7-03-dodgy-2021.zip)
(.csg as I don't want to share the .scad, but the .scad crashes the same way)

There is also a problem with console fonts in this build, at first it just looked like it was bigger, but it is variable:
Capture-2021 02 1-ci7176-fast-console-font

Most of my tests have shown significant improvement, but one is much worse.
Use Examples/Old/example024.scad, set `n=4;`, was ~35-40 minutes on my system, it went to 9.5 hours, (but ~3Gb v's ~6Gb commit size).
This may give a hint:
Capture-console-2021 01-vs-2021 fast

I've noticed the exported STL files are from 2-4 times the (binary) size, I'm yet to look at this in detail, but note the facet counts above.

MichaelAtOz:
My initial look at the STL output, feature off v's on, both clean .exe runs.
Doesn't look good...
Capture-netfabb-stls-2021 fast-on-off

MichaelAtOz:
Note that design is [condensed matter rendering - even smaller.zip](https://github.com/openscad/openscad/files/5993067/condensed.matter.rendering.-.even.smaller.zip).

It has self-intersections. Possibly related to #3644. It also has degenerate faces.

Below, the left-top is 2021.02.14 with the feature off, left-bottom, feature on, right 2021.01.
White is degenerate faces, red self-intersections, that magenta at the top of the sphere is duplicate face.
CapturedenSed-2021 01-v-2021 fast off on

I imagine the sheer number of triangles makes degenerate ones inevitable.

I've only looked at one other STL, example024 (n=3), not nearly as bad, but does introduce degens.
More details later.

rcolyer:
> Hi, I'm the author of the corefinement code.

While there are still some outstanding issues to figure out on this with the facet count and some edge cases having issues yet (to be expected), I'm thoroughly impressed with the performance boost this approach has given here.  Solid work, @sloriot, on the corefinement code, and to @ochafik on making that all work here so effectively.  I tested on some authentic geometrically heavy designs I had laying around, and it's no stretch to say this is expanding by approximately a factor of 5-10 the scale of what can be reasonably done with OpenSCAD.

MichaelAtOz:
Note the 'condensed matter rendering' code is not mine. I can't find the source to credit the author, I think it was submitted as a possible example for the wiki, but Googlebing (tm) fails me...

MichaelAtOz:
Another example of bad geometry.
scalemail.scad use_sweep=false.
Capture-scalemail-false-2021 01-v-2021 fast

MichaelAtOz:
logo.scad
Capture-logo-2019-2021 01-2021 fast

NetFabb doesn't show these as SI's. (I hate it when they don't agree...)

ochafik:
Thanks a lot for the feedback!

- @sloriot: infinite kudos and thanks for the corefinement goodness! I clearly mixed up the conditions for the unsupported cases and my current check for shared vertices is just plain bogus / only catches a (frequent) subset of problematic cases. As you say corefinement works perfectly fine with `cube(1); translate([1, 1, 1]) cube(1);` ([self-contained example](https://pastebin.com/Hg8i2iXC)), while `cube(1); translate([1, 1, 0.5]) cube(1);` explodes (partially shared edge) and that case isn't detected by my code, so... I'm a bit sad now 😰. Do you think there would be a way to detect those shared edge cases in reasonable time? (or maybe, throw an exception from within the corefinement code when this pops up?) I would hate to tell people to add some epsilons to their models to render faster :-/.
  - Also, looks like some shared edges are okay (if they come with colinear faces?), which is probably what fooled me into a sense of security:
    ```
    case = 0;
    
    // OK
    color("green") {
    if (case==0) { cube(1); translate([1, 1, 1]) cube(0.5); } // shared vertex: caught by check but would have worked
    if (case==1) { cube(1); translate([1, 0, 0.25]) cube(0.5); } // shared edge & coplanar faces
    if (case==2) { cube(1); translate([1, 0, 0.25]) cube(1); } // same
    if (case==3) { cube(1); translate([1, 0, 0]) cube(0.5); } // shared vertex & edge: caught by check
    }
    
    // BAD
    color("red") {
    if (case==4) { cube(1); translate([1, 1, 0.5]) cube(1); } // shared edge
    if (case==5) { cube(1); translate([1, 1, 0.25]) cube(0.5); } // shared edge
    }
    ```
  -  Re/ corruption of the polyhedron, the only cases I got were with PMP::orient, not with corefinement (will try and file a bug... eventually :-D)

- @MichaelAtOz: thanks a lot for the test cases, will look into them!
  - re/ the slowdown, I'd bet it's due to the "optimization" of `difference() { A; B; C...}` -> `difference() { A; union() { B; C... }}` I kept from https://github.com/openscad/openscad/pull/3636 and which makes no sense now that differences are also using corefinement): will check!
  - re/ font change, I would expect this to come from other recent changes. Note that I've (hopefully) fenced all my changes behind the `fast-csg` feature so you can test before/after with the same build (just make sure to flush the cache from the Design menu between renderings if you change the option)

- Facet count increase (as reported by @rcolyer) seems an expected difference of corefinement vs. nef unions: `cube(1); translate([0.5, 0, 0]) cube(1);` looks [like this w/ nef](https://imgur.com/f1M2dxy), [and like that w/ corefinement](https://imgur.com/FJ3KFKI). I'm hoping to prototype some constrained edge removal as a follow up, but probably less important than the rest

- Memory usage: I saw a x2 increase on some models and indeed it seems to be an epeck / #481 issue. I finally took a serious look at switching kernels and found out that https://github.com/CGAL/cgal/issues/5322 was the blocker (corefinement was assuming epick or epeck), it's fixed in their 5.2.x branch (great timing that @dpellegr reported it last month!). I've managed to align the kernel of the hybrid w/ the existing CGAL_Kernel3, which also helps avoid a few copies and definitely seems to result in *less* memory usage (yay!) than the original (maybe thanks to inplace updates), so being on the bleeding edge of CGAL should help here.

sloriot:
If I understand correctly what you are saying, you are saying that you are having memory issues when using the Epec kernel but not  `Simple_cartesian<Gmpq>`. Epeck kernel internally features a DAG where each operation done to construct an object (only a new Point_3 in the case of corefinement) is stored in the tree with a relationship to the parents objects used to compute that point (3 points of a triangle face + 2 points of a segment). At each node of the tree an interval version of the object is stored so that all predicate operations are first done using those intervals to speed up the computation. Only if the precision of the intervals is not enough to conclude, the exact computation is done and the DAG locally collapsed to feature a single node. If you are doing a single Boolean operation, then you will have a DAG of height 2. But each time you are applying a Boolean operation on top of a model that is the result of a previous operation, you will have another level in the DAG.
If you are having such cases and memory available is shrinking, what can be done is to force the evaluation of the DAG so that it only features leaves. If you call `CGAL::exact(p)`, it will force the exact evaluation for `p` and will make it a leaf with no parent (at the cost of doing an exact computation, but at least you would still benefit from the interval filtering).

We have been working on an alternative providing interval filtering + exact representation without the DAG in https://github.com/CGAL/cgal/pull/4495 but did not conclude. @afabri @MaelRL maybe we should consider this PR again and do some test with the corefinement code.

I can also see another possible improvement to limit the DAG depth by keeping the original supporting plane of the triangle faces while a future Boolean operation might be needed. That way all new point creations will occur only on input points.

As a bonus, note that https://github.com/CGAL/cgal/pull/5461 might be interesting to retriangulate a mesh once all Boolean operations are done (this code will turn a cube meshed with many triangles to a mesh made of 2 triangles per face of the cube).

MichaelAtOz:
Figures, note memory use is significantly less in the new code.

Capture-2021 fast-figures
(It was easier as a picture, let me know if anyone wants numbers)

MichaelAtOz:
BTW I wonder whether the new CGAL use is perhaps more impacted by the OpenSCAD Grid-ification of vertices? Perhaps contributing to higher rate of self-intersections? This is not my area of expertise.

sloriot:
@MichaelAtOz going from the exact rational world to the double grid already creates self-intersection. If Openscad uses a coarser grid then it will increase the risk of having self-intersections. If you have such cases, I'm very interested as I'm working on a solution to fix such issues.

MichaelAtOz:
@sloriot yes my recent tests have shown lots of self-intersections (Sis), e.g. see the screen-grab [above](https://github.com/openscad/openscad/pull/3641#issuecomment-780317476), of Meshlab showing the increase in a pretty simple object, particularly with the version here (the one labeled ...fast). The STL with this version is rejected by NetFabb, my usual tool, it won't open it.

And [this one](https://github.com/openscad/openscad/pull/3641#issuecomment-780282019), note that NetFabb shows 'holes', I believe it can class a bad SIs as a hole, due to the shifted vertex no-longer aligned with the next face. The 'duplicate face' is interesting, I'm not sure where it would come from.

NetFabb also flags 'degenerate faces', as a configurable minimum face length, I use 0.001mm, these tend to be slivers, candidates for 'problematic' geometry susceptible to floating point issues, such as one I investigated recently #3644, causing a slicer problems.

I can share the .STLs if you are interested, or other export formats.

If the doubles format keeps producing such SI's, it will be practically unusable.

sloriot:
@MichaelAtOz, yes I'm interested but I need also the exact representation to be able to properly handle the rounding. We can call the generic CGAL self-intersection repair function but we should get better result if we have the self-intersection free exact input in addition. Then maybe I need a openscad script and run it with this branch, dumping the exact representation + the rounded one? Alternatively, if you already know the place where the rounding is happening, I can tell you what code to add to get that. Actually, it might even be interesting to keep that code in the source for debugging purpose.

MichaelAtOz:
@sloriot I'm not a developer, you would need such support. 

By exact representation, do you mean OpenSCAD-pre-grid EPEC, or the Gmpq? 
Prior OpenSCAD versions were all rational, with no SI's, but more recently some aspects transition to floats, which is where the grid came in apparently as floats causes some faces to be non-planer.

> Alternatively, if you already know the place where the rounding is happening, I can tell you

And by 'rounding' you mean the grid? The grid is done [here](https://github.com/openscad/openscad/blob/9a08c222f2486c8af0a839692e644df07ca0c657/src/cgalutils-polyhedron.cc#L31).

Anyway you will need a developers interest to advance this aspect.

ochafik:
> what can be done is to force the evaluation of the DAG so that it only features leaves. If you call CGAL::exact(p)

@sloriot Hah! An easier option could have been to use Epeck with `-DCGAL_DONT_USE_LAZY_KERNEL` but this doesn't seem to build (even after the fix for https://github.com/CGAL/cgal/issues/5322). Re/ calling exact explicitly, I'm not sure we have a guarantee to be able to find all the numbers lurking around inside a Nef polyhedron, is enumerating all of its vertices enough?

Anyway, another consideration for now is compatibility with the existing nef (including maybe binary compatibility of the undocumented .nef3 serialized nef output format), so when I detect 5.2.x (x > 0) I'm using the same kernel for all nefs and save lots of conversions in the code that still operate on the old nef (custom minkowski and more). Later on we could just switch the kernel for all of OpenSCAD to whatever is faster :-)

> wonder whether the new CGAL use is perhaps more impacted by the OpenSCAD Grid-ification of vertices

@MichaelAtOz I've now disabled the grid logic for fast-csg, and it seems okay! (except for a weird failure only manifesting on CI, which I need to investigate). In fact, it now renders [issue1165.scad](https://github.com/openscad/openscad/tree/master/testdata/scad/3D/issues/issue1165.scad) properly, so... yay! I couldn't check the difference in terms of degenerate faces though (I don't quite master MeshLab yet, and there's no Netfabb on MacOS X it seems).

> I've had a couple of crashes. ... [For Dyson Cinetic - v7-03-dodgy-2021.zip](https://github.com/openscad/openscad/files/5992824/For.Dyson.Cinetic.-.v7-03-dodgy-2021.zip)

There shouldn't be crashes anymore (catching exceptions FTW!), and also, that file should now render fine (in a bit less than 5min on my laptop): thanks for sharing it!

> the code should not leave the mesh structure invalid if the input are valid.

@sloriot This could definitely be invalid inputs, but I ran into cases where corefinement union leaves the polyhedron in a state that prevents the later creation of a Nef (cases where the function returns false, but also where it throws an exception in some triangulation code).

I've struggled a bit to build a reproductible error case outside OpenSCAD but [collected some build instructions + example failure here](https://gist.github.com/ochafik/8d0db8fca505858cfe7a6fc38cff645d).

> note that CGAL/cgal#5461 might be interesting to retriangulate a mesh once all Boolean operations are done

Oh great, will follow your progress on that one!

ochafik:
@MichaelAtOz so as it turns out, https://github.com/openscad/openscad/blob/master/examples/Old/example024.scad was hitting some of what [CGAL::Epeck](https://doc.cgal.org/latest/Kernel_d/structCGAL_1_1Epeck__d.html)'s documentation euphemistically refers to as "surprising running time". I did what @sloriot suggested (calling CGAL::exact on all the points I could get to) and the file now renders 3x faster than w/o fast-csg (instead of astronomically slower). Also, what a beautiful model!

@sloriot Not entirely sure yet, but found 2 potential issues with corefinement difference:
- Isolated a self-sufficient test case where it seems slower than nef difference:
[corefinement-difference-slower-than-nef.zip](https://github.com/openscad/openscad/files/6014048/corefinement-difference-slower-than-nef.zip) (obtained the .off files using a new -`DFAST_CSG_DEBUG_SERIALIZE_COREFINEMENT_OPERANDS` option)
- Depending on when I call exact (on inputs vs. outputs), in-place update seems much slower than outputting in a fresh polyhedron (which is what I'm now doing for all corefinement operations, out of precaution). Couldn't reproduce this with my little test case, which lacks complex lazy numbers.

Lazy issues won't be a problem when building against >=5.2.1 (running the new `forceLazyToExact` helper only when we use Epeck in the hybrid), but before then this leaves some potential numeric issues when we fallback to Nefs, as I couldn't get my hands on Nefs' mutable vertices (it only has const iterators). It feels wrong to const-cast them, although it could potentially work :-S

MichaelAtOz:
Initial look at logo:
Capture-logo-2021 02 19-fast
No SI's.

Capture-logo-2021 02 19-fast-v-02 14
Better triangulation & fewer.

I'll try the bigger ones tomorrow.

ochafik:
@MichaelAtOz  That's promising, thanks for testing!

I've pushed more changes, I *think* example024.scad with `n=4` is now about the same as w/o fast-csg.

I've also added a mode that uses [CGAL::Surface_mesh](https://doc.cgal.org/latest/Surface_mesh/classCGAL_1_1Surface__mesh.html) instead of [CGAL::Polyhedron_3](https://doc.cgal.org/latest/Polyhedron/classCGAL_1_1Polyhedron__3.html). It can be enabled from the settings (or in command line w/ `--enable=fast-csg-mesh`). It's still WIP, and I'm only seeing marginal speed benefits so far (haven't compared memory usage yet: that's where it's meant to shine). To complicate the measurements, the surface mesh code currently does the "force lazy numbers to exact" as soon as the faces are split during the corefinement (using a visitor / callback), while the polyhedron code does it at the end of the operation (I'll unify the two paths soon). Also, I had to disable the fast "disjoint union" route for Surface_mesh, so it's a bit apples vs. oranges right now.

sloriot:
About in-place/out-of-place https://github.com/CGAL/cgal/pull/5429 might improve performances. I'm looking at the slow case vs nef and let you know.

sloriot:
WIth the aforementioned PR, I have:
```
Running Nef difference...
real	0m2.419s

Running corefine_and_compute_difference (distinct output)...
real	0m1.722s

Running corefine_and_compute_difference (in-place output)...
real	0m1.640s
```

With 5.2 I could reproduce the slow down mentioned. So as soon as the PR is merged it should be fine.

ochafik:
Some quick updates:

- `--enable=fast-csg-mesh` ([Surface_mesh](https://doc.cgal.org/latest/Surface_mesh/index.html)) does have an edge over `--enable=fast-csg` ([Polyhedron_3](https://doc.cgal.org/latest/Polyhedron/index.html)). This is especially noticeable on Minkowski models ([some quick benchmark](https://gist.github.com/ochafik/3c165eeaa74aa33103597595cea01a58)). I'll probably make it the default.

- Speaking of minkowski: I've benchmarked some alternative kernels (mostly `CGAL::Filtered_kernel<CGAL::Simple_cartesian<CGAL::Gmpq>>` to get speed from filtering, and multithreading-friendliness from refcount-free simple cartesian; also tried with normal cartesian) and that big minkowski sum is 2x slower than with `Epeck`. Multithreading that union on my 12-cores laptop recoups that loss but doesn't do better than the original.

- The lazy kernel thing is a bit annoying. I've compared doing the `CGAL::exact` calls in batch *after* corefinements vs. *during* them (in the facet created callbacks: currently enabled with `--enable=fast-csg-exact-callbacks` only for `--enable=fast-csg-mesh`): the corefinement callbacks seem to add a bit of overhead for almost all the models I've tested (something like 10% slower), but perform surprisingly well on [maze.scad](https://www.thingiverse.com/groups/openscad/forums/general/topic:34699) (2x faster;  see full [benchmark results](https://docs.google.com/spreadsheets/d/1Am8PERfepM0uz2wZElpPGWr4MxlRiPKlf3D6IdSI4i0/edit#gid=50765664))

thehans:
Running `OpenSCAD-2021.02.24.ai7374-_PR36412021.02.24.ai7374-x86_64.AppImage` from CircleCI
I get a CGAL assertion error from this script:
```
union() for(i=[1:10],j=[1:10],k=[1:10]) translate([i,j,k]) cube();
```

```
...
ERROR: CGAL error in CGALUtils::applyUnion3DPolyhedron: CGAL ERROR: assertion violation!
Expr: !CGAL::assign( v, o)
File: /usr/include/CGAL/Nef_3/Binary_operation.h
Line: 432
...
```

Renders fine with fast-csg off.

rcolyer:
@ochafik I was testing this PR on the scad file in the zip on the first post of #3644, as I thought this might resolve the associated issue (as it's a triangulation problem, and not an output issue as originally thought).  Unfortunately the render collapsed into a big fail.  So I'd like instead to highlight it as a useful test case for this PR.

pr3641_issue3644

ochafik:
This is now mostly all in from other PRs, time to close this one! (see high-level summary in [this blog post](https://ochafik.com/jekyll/update/2022/02/09/openscad-fast-csg-contibution.html))

PR #3673, opened 2021-02-16
Remove "constants" from context.  Only used for PI, which can really …

thehans:
…just considered the same as a "builtin variable".

This change should have no observable difference in behavior.


t-paul (approved):
I was a bit concered about the removed warning, but as discussed there does not seem to be a way to trigger that anyway. So lets go with the simplification.

PR #3682, opened 2021-02-21
Cleanup

jordanbrown0:
Tiny cleanups (take 2).

- Add .gitignore entries for some build artifacts,
- Build warnings in src/func.cc.


jordanbrown0:
The std::move fixes have problems in some builds that exceed my understanding.  Hopefully I'll come back to them later.

Issue #3081, opened 2019-09-25
text(halign="center", spacing!=1.0) is not centered


davidhbrown:
Centered horizontal alignment of text is not correct except when spacing is at the default 1.0:

    text("XXXXX", halign="center"); // correctly centered
    translate([0,10]) text("XXXXX", halign="center",spacing=0.8);  // shifted to right
    translate([0,-10]) text("XXXXX", halign="center", spacing=1.2); // shifted to left

Version 2019.05, Windows.

MichaelAtOz:
I know it's annoying, but it is he nature of glyphs.
In LTR text spacing is to the right of a ghyph, and there is also usually a small amount of spacing to the left. See [here](https://www.freetype.org/freetype2/docs/glyphs/glyphs-3.html#section-1) for 'advance'.

This shows what happens:
```
scale(0.75) {
color("blue") translate([0,-12,-2.5]) 
linear_extrude(1)
  text("▄▀▄", halign="center");  
color("red") translate([0,5,-1.5]) 
linear_extrude(1)
  text("▀▄▀", halign="center",spacing=0.8);  // shifted to right
color("green") translate([0,-14,-1.5]) 
linear_extrude(1)
  text("▄▀▄", halign="center", spacing=1.2); // shifted to left
}

```
![Capture text spacing annotated](https://user-images.githubusercontent.com/1972961/65671643-f14d6280-e08a-11e9-9006-b2c5fc76ef48.jpg)

I added the yellow boxes and lines manually, so they a a little off.
The yellow shows the bounding box for a glyph, the short orange lines show the origin of the next glyph, the last of which is the end position, the black lines are centered.
The yellow of the green boxes shows the expanded glyph where the extra space is on the right. The long orange lines are centered.

I beleive when the glyph is 'painted' OpenSCAD does not know what actual content is painted inside the bounding box. That would require extra processing to map the actual geometry, not just the typography.

MichaelAtOz:
Also note in the above link 'width' is notional, the actual 'ink' of a glyph can be outside of the min-max box, intended to overlap with either the previous or next glyph.

t-paul:
As @MichaelAtOz visualized, the `text()` parameters are based on glyph properties, not generated geometry. So it works as intended.

PR #3684, opened 2021-02-21
textmetrics

jordanbrown0:
This is a draft that I do not expect to be accepted.  In particular, the feature needs to be made experimental.  However, that's completely orthogonal to what it does and how it works, so I thought I would give people an opportunity to review.

It also needs to be run through the test suites, and a test for the feature should be developed.  It's expected that this change will change text rendering slightly; see below.

What it does:  It gives you information about how a particular piece of text will be formatted, and general information about a font and how it's formatted.

`textmetrics()` takes a list of arguments like text(), and returns a nested array describing how that text would be formatted:

```
echo(textmetrics("hello", size=20, font="serif", halign="center", valign="top"));
```
yields
```
ECHO: [[-27.5024, -19.5584], [54.223, 19.5584], [19.2768, -0.2816], [-27.7712, -19.2768], [55.5424, 0]]
```
Breaking that down:

- The first two pairs are the bounding box of the text.  This represents the actual ink laid down; it does not include any bearing left of the text or any advance right of the text.
- [-27.5024, -19.5584] - lower left corner
- [54.223, 19.5584] - width and height
-  [19.2768, -0.2816] - the ascent and descent (how much the text goes above and below the baseline)
- [-27.7712, -19.2768] - the offset (what halign/valign are doing to the text)
- [55.5424, 0] - the advance, where you should put the next piece of text

`fontmetrics(size, font)` returns a similar array with font-global information:

```
echo(fontmetrics(size=20, font="serif"));
```
yields
```
ECHO: [[24.7532, -6.0086], [27.2624, -8.4228], 31.9418]
```
Again, breaking that down:
- [24.7532, -6.0086] - the design ascent and descent of the font.  Exactly what that means seems to depend on the font.
- [27.2624, -8.4228] - the actual max ascent and descent of the font - the vertical bounds of the entire font.
- 31.9418 - the designed spacing from the baseline of one line of text to the baseline of the next line.

The nested-array structure is ugly and hard to work with (though adding constants for the indexes helps).  The tentative plan is to integrate with the ObjectType support from the object literals project so that one might refer to, e.g., `tm.bbox,origin` and `tm.bbox.dims`.

Could this be done in a library?  No.  A library doesn't have access to the font data or to the layout algorithms that OpenSCAD uses to render the text.

Notes:
- This change eliminates the 2.4% metrics scaling legacy.  This means that text will spread out slightly.
- It tries to clean up the behavior of halign/valign with vertical text.  One subtle implication is that the defaults are different for horizontal and vertical text, so I changed the default value for those two variables to "default" so that the calculation function can then do the right thing for each.
- Several text() messages did not print the associated location in the source; I made them do so.
- The Context::lookup_variable_with_default changes to support boolean variables might be sensible, but they were only needed for a variable that I ended up removing yesterday.  They should be backed out.  (Sorry, only just now noticed that they weren't needed any more.)

Although I'm very strong in C, I have almost no C++ experience; my OO experience is all Java and JavaScript.  I am very interested in comments on proper idiom and style.

Brutal comments solicited :-)


jordanbrown0:
textmetrics-demo

Here's a graphical demonstration of the metrics returned.
The yellow box under the Hello World is the bounding box.
The vertical black lines are the bearing on the left and the advance on the right.
The horizontal black lines are the baseline and the text ascent and descent.
The red lines are the font ascent and descent.
The green lines are the font max ascent and descent.
The blue lines are the positions of the next and previous lines.
The ABCDEF is built in two strings, using the advance of the "ABC" to position the "DEF".
(Note:  This doesn't work in general, because text layout depends on the sequences of characters involved.  For instance, "ffi" will, depending on the font, be rendered as a single ligature, but if you split it into two parts that can't happen.)

jordanbrown0:
With respect to making the functions be experimental... that's straightforward enough, but eliminating the 2.4% fudge factor in text metrics, and in particular in spacing, is not so easily hidden behind the experimental flag.  I'm sure it can be done if people think it best.

jordanbrown0:
Skimming the test results - to nobody's surprise, most of the text tests failed.  I need to look at them, but that's almost certainly because of the 2.4% issue.

jordanbrown0:
Could somebody take a look at this, and give me an idea where I should head next?

jordanbrown0:
Talked to @rcolyer about this.  His big concern is the form of the return value, since that's hard to change once it's introduced.  The plan is for me to look at the [Objects](https://github.com/doug-moen/openscad2/blob/master/rfc/Objects.md) project (#3087, #3088) to see if the relatively-easy parts of that project that provide for accessing internally-generated objects could be extracted and integrated as a first phase.

jordanbrown0:
The object support basically works:
```
echo(textmetrics("hello", size=20, font="serif", halign="center", valign="top"));
```
yields (wrapped for readability)
```
ECHO: {
    pos = [-27.5024, -19.5584];
    dims = [54.223, 19.5584];
    ascent = 19.2768;
    descent = -0.2816;
    offset = [-27.7712, -19.2768];
    advance = [55.5424, 0];
}
```

```
echo(fontmetrics(size=20, font="serif"));
```
yields
```
ECHO: {
    nominal = { ascent = 24.7532; descent = -6.0086;};
    max = { ascent = 27.2624; descent = -8.4228;};
    interline = 31.9418;
}
```

What works:
- The two functions yield the right values.
- You can create objects from C++ using straightforward calls similar to VectorType.
- You can assign them, pass them as arguments, et cetera.
- You can access the values using obj.name and obj["name"].
- echo(obj) and str(obj) yield the output shown above.

What doesn't work:
- There is no way to traverse an object from OpenSCAD.  If you don't know the names, you can't get them.
- There is no way to create an object from OpenSCAD.
- There are no comparison operators.  Greater-than and less-than wouldn't be meaningful, but equal and not-equal would be.

I don't propose to do those as part of this project, since they aren't needed for it and this project doesn't make them any harder.

jordanbrown0:
Oh, and I need to pull in the latest changes, notably including the cmake changes.

src/FreetypeRenderer.cc
```diff
@@ -256,49 +319,186 @@ std::vector<const Geometry *> FreetypeRenderer::render(const FreetypeRenderer::P
...
+        }
+
+        advance_x += glyph->get_x_advance() * params.spacing;
+        advance_y += glyph->get_y_advance() * params.spacing;
+    }
+
+    if (has_ink) {
+        if (HB_DIRECTION_IS_HORIZONTAL(
+            hb_buffer_get_direction(hb_buf))) {
+            calc_offsets_horiz(params);
```
t-paul, line 390: Maybe we could use one of the Eigen library types to return x/y instead of having `calc_offsets_*` set the values directly? That way those could stay `const` making it slightly easier to reason about the data flow.

t-paul (changes requested):
I'm not seeing any showstopper at this point, with the features locked behind the experimental flag, it should be ok to merge after getting the test suite back to green and cleaning up the dummy `obj()` function.

src/text.cc
```diff
@@ -81,8 +84,8 @@ AbstractNode *TextModule::instantiate(const std::shared_ptr<Context>& ctx, const
 	node->params.set_direction(c->lookup_variable_with_default("direction", ""));
 	node->params.set_language(c->lookup_variable_with_default("language", "en"));
 	node->params.set_script(c->lookup_variable_with_default("script", ""));
-	node->params.set_halign(c->lookup_variable_with_default("halign", "left"));
```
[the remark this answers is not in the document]
jordanbrown0: Before this change, the default alignment values are halign=left / valign=baseline.

With this change:
The default for horizontal-layout text (in calc_offsets_horiz) is halign=left / valign=baseline.
The default for vertical-layout text (in calc_offsets_vert) is halign=center / valign = top.

I don't remember all of the details, but if you experiment with direction=ttb and valign=top and valign=bottom you'll see some ... interesting... behaviors.  Also check out all three values of halign.  I believe I corrected all of the ttb alignment behaviors, but making the defaults yield the same output meant making the defaults for ttb text be different from the defaults for ltr text.

src/FreetypeRenderer.cc
```diff
@@ -196,31 +244,46 @@ void FreetypeRenderer::detect_properties(FreetypeRenderer::Params &params) const
...
+bool FreetypeRenderer::ShapeResults::shape(
+    const FreetypeRenderer::Params &params)
+{
+    FT_Error error;
+    FT_Face face = get_font_face(params);
+    if (face == nullptr) {
+        return false;
+    }
+
+	hb_ft_font = hb_ft_font_create(face, nullptr);
```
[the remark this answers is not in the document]
jordanbrown0: `shape()` is only ever called once.

The buffer cleanup can't be in shape(), because one of shape()'s results is `glyph_array`, and `glyph_array`'s `GlyphData` members include a pointer to the position data that comes back from `hb_buffer_get_glyph_positions()`.  The HB data has to stick around until `text()` is done using `glyph_array`.

I was unhappy about that asymmetry too.  I considered doing the real work in the constructor, but there's no good way to return an error from a constructor.

In the old implementation, all of the shape calculation is in `FreetypeRenderer::render` using local variables.  Here, they're extracted out into `ShapeResults::shape`, and are consumed by both `FreetypeRenderer::render()` and `FreetypeRenderer::TextMetrics::calculate()`.  Moving the HB pointers into ShapeResults meant that ShapeResults needed to clean them up.

src/FreetypeRenderer.cc
```diff
@@ -256,49 +319,186 @@ std::vector<const Geometry *> FreetypeRenderer::render(const FreetypeRenderer::P
...
+        // Note that glyphs can extend left of their origin
+        // and right of their advance-width, into the next
+        // glyph's box.  In theory they could extend
+        // arbitrarily through other glyphs' boxes.  Hence the
+        // need for the min/max here.
+
+        // Glyphs with null bounding boxes do not contribute
+        // ink and so do not contribute to the bounding box or
+        // ascent and descent.
+        if (bbox.xMax > bbox.xMin && bbox.yMax > bbox.yMin) {
```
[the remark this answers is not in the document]
jordanbrown0: Not sure.  Maybe, but not very much.  I think it would turn the following min/max sequences into constructing boxes with the bounding box for the current character, and then using `Eigen::AlignedBox::extend()` to merge them into the current box.  I think it would be very close to the same amount of text.  Whether it would be clearer would depend on whether you think of it as calculating the X and Y positions of the bounding boxes, or as unioning a set of bounding boxes.

src/FreetypeRenderer.cc
jordanbrown0: I don't think so, or at least not without making it so some of the shaping results are returned as the return value from `shape()`, while others are returned as members of the `ShapeResults` object.  (Or maybe I'm just not understanding.)

Note that `calc_offsets_*()` are private functions called only once.  Their reason for existence is purely to factor out the sequences that calculate those offsets as readily identifiable components.

jordanbrown0 (commented):
> I'm not seeing any showstopper at this point, with the features locked behind the experimental flag,

Note that the new features are experimental, but there are implementation changes to text(), and a couple of subtle behavioral changes:
- 2.4% spacing scaling removed; text spreads out slightly.
- vertical text alignment behavior was just wrong; fixed it (but kept the default behavior the same).

Between the implementation changes and the behavioral changes, even without enabling the new features the risk is non-zero.

jordanbrown0:
Next steps:  obj(), test cases.

src/func.cc
```diff
@@ -914,6 +916,150 @@ Value builtin_cross(const std::shared_ptr<Context> ctx, const std::shared_ptr<Ev
...
+
+    // The bounding box, ascent/descent, and offset values will be zero
+    // if the text consists of nothing but whitespace.
+    VectorType bbox_pos;
+    bbox_pos.emplace_back(metrics.bbox_x);
+    bbox_pos.emplace_back(metrics.bbox_y);
+
+    VectorType bbox_dims;
+    bbox_dims.emplace_back(metrics.bbox_w);
+    bbox_dims.emplace_back(metrics.bbox_h);
```
jordanbrown0, line 1002: Q:  should bounding boxes be represented as:
- [ left, right, bottom, top ]
- [ [minx, miny], [dimx, dimy] ]
- [ [minx, miny], [maxx, maxy] ]
?

Probably not the first one, because it's not a point or a dimension.  But:  it's equally inconvenient to get the coordinates of any of the four corners.

The second works well for things like a translate and a square.
The third works well for positioning to the lower left or upper right corner.

Translating between the second and third is just a matter of adding or subtracting.

MichaelAtOz: I vote the second, I can see resize() being used with this, having the width/height at hand is easier.
I presume that these are axis-aligned, and don't change with `direction=`?

jordanbrown0: They change with `direction` in that the bounding boxes for horizontal text tend to be wide and not very tall, and for vertical text tend to be tall and not very wide.  But the position is always the lower-left corner, the minimum X/Y, and the dimension is always positive.  (Or, if the third variant, the second point always has X and Y greater or equal to the first point.)

It's a bounding box.  It's the minimum and maximum X and Y for the actual ink laid down.

Put another way, it's the translation and size of the smallest square() that will cover the ink laid down.

jordanbrown0:
Current generation of results from the functions:

textmetrics():
```
{
    position = [0.96, -0.1408],
    size = [27.8024, 10.208],
    ascent = 10.0672,
    descent = -0.1408,
    offset = [0, 0],
    advance = [29.3443, 0]
}
```

fontmetrics():
```
{
    nominal = { ascent = 12.5733, descent = -2.9433 },
    max = { ascent = 13.6109, descent = -4.2114 },
    interline = 15.9709,
    font = { family = "Liberation Sans", style = "Regular" }
}
```

I thought I'd switched it to match JavaScript syntax, but it doesn't look like it.  I'd prefer to match JS syntax.

jordanbrown0:
I'd forgotten that a design intent of pull request #3087 was that an object literal looked like an ordinary block with assignments.  I've gone back to that syntax for the string/echo form of an object:
```
ECHO: { nominal = { ascent = 12.5733; descent = -2.9433; }; max = { ascent = 13.6109; descent = -4.2114; }; interline = 15.9709; font = { family = "Liberation Sans"; style = "Regular"; }; }
```

jordanbrown0:
[ This is not anywhere near done, but I need to save it and work on something else for a while. ]

To-do list from self-review:

- [ ] Add test cases for alignment of vertical text.

Comments:
- `halign="left"` is based on the text's origin, and `halign="right"` is based on the text's advance (as before).  These are not the same as the bounding box, so neither of them slams the text right up against the axis.  I'm not sure which is more right, but for compatibility I'm going to leave it alone.  If you want a bounding-box solution, you can use the bounding box returned by `textmetrics()` to adjust.  Similarly, but more subtly, for `halign="center"`.
- This has subtle effects on horizontal centering of some characters.  In Liberation Sans, for instance, a lower-case "j" is centered along the vertical stroke, even though that's well to the right of the bounding-box center of the glyph.  Again, if you want the bounding-box behavior you can use the `textmetrics()` bounding box.
image

jordanbrown0:
I'm mostly done with my personal review.  I want to take another pass, but I don't know if there's much more for me to do.  That's greatly limited by my C++ knowledge; I would be very interested in commentary.

I realized that it wouldn't be hard to implement `for (i=obj) ...`, so I did that.  There's a question of what values `i` should get - I have it getting keys.  Apparently the same functions support both the `for()` module and the `for()` list comprehension, because they both seem to work.

I was a bit more aggressive in cleaning up what I thought were minor organizational issues in the classes around `FreetypeRenderer`; I'm interested in comments there.  There were a couple of `FreetypeRenderer` methods that existed only to manipulate a `FreetypeRenderer::Params` object, so I moved them to being methods on `FreetypeRenderer::Params`.

jordanbrown0:
Oh, and I implemented is_object(v), because that's trivial.  I left that non-experimental, but could put it under the same experimental umbrella.

src/value.h
```diff
@@ -501,6 +502,32 @@ class Value
...
+    const Value& get(const std::string& key) const;
+    void set(const std::string& key, Value&& value);
+    Value operator==(const ObjectType &v) const;
+    Value operator< (const ObjectType &v) const;
+    Value operator> (const ObjectType &v) const;
+    Value operator!=(const ObjectType &v) const;
+    Value operator<=(const ObjectType &v) const;
+    Value operator>=(const ObjectType &v) const;
+    const Value& operator[](const str_utf8_wrapper &v) const;
+    const std::vector<std::string> keys() const;
```
t-paul, line 528: The suggested from LGTM seems not ideal. I think it's probably better to return a `const &` instead of removing the `const`.

jordanbrown0: Done, thanks.

jordanbrown0:
Except for rebasing away the churn, I've run out of things that I feel need to be done.

Note that one change today was to notice that the import of LibSans was getting an error both for my tests and for other tests, and so I fixed them.

jordanbrown0:
Let me know when you're happy and I'll rebase away the churn (and resync as required).

t-paul (changes requested):

t-paul:
I wonder if we should add the `font` info to `textmetrics()` too.

src/FreetypeRenderer.cc
```diff
@@ -97,49 +95,95 @@ int FreetypeRenderer::outline_cubic_to_func(const FT_Vector *c1, const FT_Vector
...
-	hb_direction_t param_direction = hb_direction_from_string(params.direction.c_str(), -1);
+	hb_direction_t param_direction = hb_direction_from_string(direction.c_str(), -1);
 	if (param_direction != HB_DIRECTION_INVALID) {
 		return param_direction;
 	}
 
 	hb_direction_t direction = hb_script_get_horizontal_direction(script);
-	PRINTDB("Detected direction '%s' for %s", hb_direction_to_string(direction) % params.text.c_str());
+	PRINTDB("Detected direction '%s' for %s", hb_direction_to_string(direction) % text.c_str());
 	return direction;
```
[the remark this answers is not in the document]
jordanbrown0: OK.  Might be a couple of days.

src/parameters.h
```diff
@@ -16,7 +16,7 @@
 class Parameters
 {
 private:
-	Parameters(ContextFrame&& frame);
+	Parameters(ContextFrame&& frame, const Location& loc);
```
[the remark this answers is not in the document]
jordanbrown0: Parameters gained the `valid` and `valid_required` type-and-presence checking functions, so that you can say
```
if (!parameters.valid("somenumber", Value::Type::NUMBER)) {
    handle error somehow
}
```
and it will spit out a warning.  To spit out a warning with a file name and line number, it needs a documentPath and a Location.  Note that the creation of the Parameters already got the location as an argument; this just retains it for use in subsequent checks.

tests/regression/echotest/text-metrics-test-expected.echo
```diff
@@ -0,0 +1,20 @@
...
+WARNING: textmetrics() parameter could not be converted: text: expected string, found number in file text-metrics-test.scad, line 20
+WARNING: textmetrics() parameter could not be converted: spacing: expected number, found string in file text-metrics-test.scad, line 20
+WARNING: textmetrics() parameter could not be converted: font: expected string, found bool in file text-metrics-test.scad, line 20
+WARNING: textmetrics() parameter could not be converted: direction: expected string, found number in file text-metrics-test.scad, line 20
+WARNING: textmetrics() parameter could not be converted: language: expected string, found range in file text-metrics-test.scad, line 20
+WARNING: textmetrics() parameter could not be converted: script: expected string, found number in file text-metrics-test.scad, line 20
+WARNING: textmetrics() parameter could not be converted: halign: expected string, found number in file text-metrics-test.scad, line 20
+WARNING: textmetrics() parameter could not be converted: valign: expected string, found number in file text-metrics-test.scad, line 20
+ECHO: { position = [0, 0]; size = [0, 0]; ascent = 0; descent = 0; offset = [0, 0]; advance = [0, 0]; }
```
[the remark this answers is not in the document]
jordanbrown0: Well, maybe, but I think it's precisely reporting what `text()` would do with the data provided.

`text()` is very forgiving.  In the current implementation, you can say `text()` or `text(true)` or `text(5)` and they will all emit nothing, silently.

Right now, I've got emitting nothing defined to be all zeroes.  (Except if you've got whitespace; then the advance might be non-zero.)  Note the comment at func.cc line 804.  But maybe some of that stuff should be `undef` for cases where no ink was emitted. 

I'm not sure which is more likely to cause trouble.  Advance is meaningful; the rest are all pretty ill-defined.  But maybe zeroes will work better as the degenerate case.

jordanbrown0:
> I wonder if we should add the `font` info to `textmetrics()` too.

I assume that you mean the font family and style?  Or did you mean all of the `fontmetrics()` data?

My inclination for either of those is no, that they're parameters of the font (and size, for metrics) and not of the text being rendered.

But I guess it doesn't really matter, and you could just call `textmetrics()` with no text or with dummy text if you wanted the font data, and we could drop `fontmetrics()` entirely.

MichaelAtOz:
Sorry, I haven't had time to look deeper into this.

a. I assume the font returned is the matched font?

b. re fontmetrics() is there other font info that may want to be extracted in the future which would make moving it into textmetrics() awkward? I'm inclined to keep them separate too.

c. Re the earlier point re zeros, I'm a fan of undef, I think your comment re ink is good, if there is nothing there it would be good to be able to test that. With advance non-zero for whitespace.

My 2c worth.

jordanbrown0:
> a. I assume the font returned is the matched font?

Yes.

> b. re fontmetrics() is there other font info that may want to be extracted in the future which would make moving it into textmetrics() awkward? I'm inclined to keep them separate too.

It currently returns everything that seemed relevant.  Could there be something else?  Shrug.  Would it awkward to put it into textmetrics()?  Probably not, because the object style means that you can add arbitrary new data without breaking old consumers.

> c. Re the earlier point re zeros, I'm a fan of undef, I think your comment re ink is good, if there is nothing there it would be good to be able to test that. With advance non-zero for whitespace.

I can add tests for those cases.

tests/data/scad/misc/text-metrics-test.scad
```diff
@@ -0,0 +1,25 @@
...
+
+echo("Normal...");
+
+// Force Liberation Sans since it's present on all platforms.
+// Exercise normal operation with default and explicit settings.
+
+// textmetrics()
+echo(textmetrics("hello", font="Liberation Sans"));
+echo(textmetrics("hello", font="Liberation Sans", size=20, direction="rtl",
+    language="en", script="latin", halign="right", valign="center", spacing=2));
```
jordanbrown0, line 11: Add tests for empty string and for whitespace only.

Consider adding tests for no text specified (but maybe not, because maybe `text()` should complain if you don't specify text).

jordanbrown0:
> > c. Re the earlier point re zeros, I'm a fan of undef, I think your comment re ink is good, if there is nothing there it would be good to be able to test that. With advance non-zero for whitespace.
> 
> I can add tests for those cases.

I may have misunderstood your request.  I thought you were asking for a test case for this situation.  Now, rereading it, I think you might be saying that you want it to be easy for a program to look at `textmetrics()` results and say "there's no ink there", perhaps by having some of the elements come back undef.

Note that a program can easily detect a no-ink situation by checking whether the "size" result is [0,0].

I do think that zeroes are the right answer for most of the values, so that the degenerate cases still "work".  In a "text on a plate" case that I was just doing the other day, I accidentally gave it an empty string, and it correctly gave me a block that was all margin, around a zero-sized piece of text.  I think that's much more right than a "you tried to use undef" error would have been.

t-paul:
I think for cases like empty string, or maybe just invisible codepoints, that sounds ok. I'm was wondering about the case where text is undefined, e.g. `text = 123` which generates a warning for invalid type, but then proceeds with an empty text.

jordanbrown0:
Here's a first cut at the documentation for `textmetrics()`.  This should probably be accompanied by an illustration.

--

`textmetrics()` returns information about the size and positioning that would be applied to text if laid out as specified by its arguments.

The arguments to `textmetrics()` are exactly the same as the arguments to `text()`.

The return value from `textmetrics()` is an object with these properties:
- position:  an [x,y] pair giving the lower left corner of the smallest box that would completely surround the text.
- size:  an [x,y] pair giving the size of that box.
- ascent:  the vertical distance (normally positive) from the baseline of the text to the highest point in the text.
- descent:  the vertical distance (normally zero or negative) from the baseline of the text to the lowest point in the text.
- offset:  an [x,y] pair giving the distance from the origin to the starting point of the baseline of the text.  This value is normally [0,0] but can be non-zero when non-default alignments.
- advance:  an [x,y] pair giving the distance from the starting point for this text to the starting point for a subsequent piece of text.  Informally, it says how far the pen should be moved before starting the next piece of text.  This value is directly helpful only with default alignments - horizontal text aligned left/baseline and vertical text aligned center/top.

--

Questions:

- What are these name-value collections?  Other languages call them "objects", "structures", "classes", or "dictionaries".  I've used the word "object" so far, but given that OpenSCAD already sort of uses the word "object" for geometric constructs (like what `cube()` produces) maybe a different word should be used here.  Or maybe the two concepts can be melded together, but I'm not sure they can.
- what are the named things that make up these constructs?  Are they "members" or "elements" or "properties" or what?  In JavaScript they are properties.  In C and C++ they are "members".  In Java they are "fields".  Python calls them name-value pairs when discussing dictionaries and attributes when discussing objects.

For the moment I'm going to use "object" and "property", but I'm not wedded to those terms.

- `advance` isn't directly useful with non-default alignments.  For some alignments this seems sensible.  In horizontal text, it seems inevitable that you will get strange results with non-default vertical alignments on a sequence of pieces of text, because the positioning is evaluated independently for each of them.  If you use valign=bottom and draw "Xg" and then "Xb", the two Xs will not be aligned with one another; the first will be bumped up by the descender on the g.  On the other hand, perhaps `advance` should be useful for halign=right. Perhaps it should allow you to concatenate a sequence of right-adjusted strings leftward along the axis.  As currently defined it won't:  unmodified, the "advance" will move the pen to the right where it would need to be moved to the left.  The caller could compensate by negating it.  Perhaps the `advance` for halign=right text should be negative.  Or perhaps the `advance` should be negated for RTL text, with the assumption that RTL text will normally use halign=right.  (In fact, probably the default for RTL text should be halign=right, but I don't think I want to go there.)

jordanbrown0:
First cut at documentation for `fontmetrics()`:
---
`fontmetrics()` returns information about the specified font and size.

The arguments to `fontmetrics()` are the size and font name.

The return value from `fontmetrics()` is an object with these properties:

- nominal: an object with `ascent` and `descent` properties describing the "usual" ascent and descent of the font.  For most fonts, this will be the ascent of upper-case letters and the descent of lower-case letters with descenders.
- max:  an object with `ascent` and `descent` properties describing the maximum ascent and descent of the font.
- interline:  gives, as a positive number, the designed inter-line spacing for the font.
- font:  an object with members `family` and `style` that gives the name and style of the selected font.  These may not be the same as the name supplied, if the specified font is not available, or if a generic name like "sans" or "serif" is specified, or if only a style is specified.
---
Note:  I'm not sure that "max" is really the absolute maximum for the font.  As a very narrow example, Liberation Sans Regular lists 13.6109 as its maximum ascent, but U+5C0 HEBREW PUNCTUATION PASEQ has an ascent that's slightly larger at 13.6128.  I thought I'd seen larger examples, but I'm not finding one right now.

jordanbrown0:
> I'm was wondering about the case where text is undefined, e.g. `text = 123` which generates a warning for invalid type, but then proceeds with an empty text.

Well, do note that `text()` emits nothing for those cases.  (Before my work, it doesn't even give a warning.  With my work, it does.)

I'm not clear whether cascading errors would be helpful here.  `text()` soldiers on; perhaps `textmetrics()` should too.

t-paul:
Yeah, good point, regarding geometry results the definitions are a bit fuzzy. We had those discussions of empty vs. no-geometry, e.g. #3452. So that might be a point that needs consideration too. I'm not sure if that necessarily needs to drive the decision for `textmetrics()`.

jordanbrown0:
In IRC, Torsten says he tends to think that textmetrics() called on a non-string should yield undef.  (I think he means that the function should return undef, rather than that any particular property should be undef.)

I don't care a lot, because you're in error territory already.  The question is whether you should get back a value that lets you (sort of) soldier on, or get back a value that will cause cascading errors.

What to do with errors in other parameters is murkier still.  Passing a garbage string as halign, for instance, yields a warning but is then ignored; the text is still rendered.  Passing a number as halign or as the font name is silently ignored in current OpenSCAD; it yields a warning and is then ignored in this version.

jordanbrown0:
There's also a tie-in to RTL behavior in #3839 in exactly what the "offset" and "advance" values should be for RTL text.

src/FreetypeRenderer.cc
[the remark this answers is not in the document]
jordanbrown0: Did this back then, but forgot to push it from the desktop.  Pushed in now.  Suggest you take a look at FreetypeRenderer::Params::get_direction() in src/FreetypeRenderer.cc.

Fingers crossed, this should be a clean test run, and I'll double-check the fix for #3701, and this should be good to go.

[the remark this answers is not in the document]
jordanbrown0: Never mind, it's stupidly wrong.  I'm going to do a full re-review pass.

jordanbrown0:
Note that `textmetrics()` will help with issue #3081.  If you want alignment based on the bounding box of the ink laid down (rather than based on the advance and character box), you can use the bounding box that `textmetrics()` returns to do your own alignment.

Perhaps halign/valign should be based on the bounding box, but I'm not going there with this project.

src/FreetypeRenderer.cc
```diff
@@ -97,49 +95,105 @@ int FreetypeRenderer::outline_cubic_to_func(const FT_Vector *c1, const FT_Vector
...
+    } else {
+        LOG(message_group::Warning, params.loc, params.documentPath,
+            "Unknown value for the valign parameter"
+            " (use \"baseline\", \"bottom\", \"top\" or \"center\"): '%1$s'",
+            params.valign);
+    }
 }
 
-hb_direction_t FreetypeRenderer::get_direction(const FreetypeRenderer::Params &params, const hb_script_t script) const
+hb_direction_t FreetypeRenderer::Params::detect_direction(const hb_script_t script) const
```
jordanbrown0, line 174: I was bothered that `get_direction` and `get_script` were not simple getters; they were not even symmetric with the type accepted by `set_direction` and `set_script`.

I renamed them to `detect_xxx` to make the distinction a little clearer, but I'm still not sure that the structure is what it should be.  It's tempting to have the string value be *only* the value passed in, and have a separate `hb_direction_t` / `hb_script_t` value that's the result of heuristics, defaulting, et cetera.  In that model, `detect_xxx` would probably move over into `ShapeResults` and would just set `hb_xxx` values there.  That model would avoid the mildly silly behavior of taking a string, converting (and defaulting, applying heuristics, et cetera) it into a HB enum, converting that enum back into a string, and then finally converting the string into a HB enum again.  The downside would be that the final value wouldn't be directly printable, but the HB enum could be converted if desired.  (But:  Maybe it would also affect what shows up in CSG dumps, making them show the input value rather than the final value.)

src/value.cc
```diff
@@ -1284,3 +1322,61 @@ std::ostream& operator<<(std::ostream& stream, const FunctionType& f)
...
+  ptr(shared_ptr<ObjectObject>(new ObjectObject()))
+{
+  ptr->evaluation_session = session;
+}
+
+const Value& ObjectType::get(const std::string& key) const
+{
+    auto result = ptr->map.find(key);
+    // NEEDSWORK it would be nice to have a "cause" for the undef, but Value::undef(...)
+    // does not appear compatible with Value&.
```
jordanbrown0, line 1342: Is there anything that can be done about this?

jordanbrown0 (commented):
In any function where I made non-trivial changes, I converted it over to what I've been told is the preferred indentation form:  4-column indentation using spaces only, no tabs.  Let me know if you'd like me to be less aggressive about this.

jordanbrown0:
Pending any final review and any comments on my review comments, and final test runs, I think this is ready to go.

t-paul (approved):
Nice, lets get this into the hands of more people.

PR #3693, opened 2021-02-26
OpenSCAD with 3D glasses

koendv:
Adds support for seeing your models in 3D with red/cyan colored 3D glasses.
There's a more complete write-up at https://github.com/koendv/openscad-raspberrypi


openscad.qrc
```diff
@@ -106,6 +106,7 @@
     <file>images/SVG-white.png</file>
     <file>images/PDF.png</file>
     <file>images/PDF-white.png</file>
+    <file>images/Anaglyph-32.png</file>
```
t-paul, line 109: We will need 2 icons, one for light, one for dark color desktop scheme. I'll have a look, we did generate most of the others based on an OpenSCAD script. I have to dig that out, I think it might be in some branch.

src/Camera.cc
```diff
@@ -1,22 +1,29 @@
 #include "Camera.h"
+#include "QSettingsCached.h"
```
t-paul, line 2: Unfortunately that's not possible, `Camera.cc` is part of the core which may not depend on Qt, so we have to find a way to pass that setting along from outside.

There's currently only some hints in `CMakeLists.txt` so it's not very obvious, hopefully this can be made more clear at some point by getting the source files better organized. 
https://github.com/openscad/openscad/blob/edd69c2792b191bbd577fcc0acf2bbd836aa0f76/CMakeLists.txt#L451-L453

src/builtin.cc
```diff
@@ -113,6 +113,7 @@ Builtins::Builtins()
 	this->assignments.emplace_back(new Assignment("$vpr", make_shared<Literal>(VectorType(0.0, 0.0, 0.0))) );
 	this->assignments.emplace_back(new Assignment("$vpd", make_shared<Literal>(500)) );
 	this->assignments.emplace_back(new Assignment("$vpf", make_shared<Literal>(22.5)) );
+	this->assignments.emplace_back(new Assignment("$anaglyph", make_shared<Literal>(Value::undefined.clone())) ); // true if rendering anaglyph 3d
```
t-paul, line 116: What's the use case for exposing this to the script?

src/dubois.cc
```diff
@@ -0,0 +1,37 @@
+#include "dubois.h"
+
+bool change_color = false;
```
t-paul, line 3: I think that might cause issues with multiple windows as this would switch globally for the whole application. Maybe the code could be converted to a class which is instantiated for each main window?

t-paul (changes requested):
Thanks for reviving this topic, it would be nice to get this feature in. I'll have to find some of those glasses now... :smile:.

src/Camera.cc
koendv: OK. Will leave Camera.cc alone.

openscad.qrc
koendv: OK. I'll add a "white" icon. Does this also mean there have to be two color schemes for 3D, one with black and one with white background?

src/Camera.cc
koendv: How about the following:
- remove dubois.h and dubois.cc files
- add function duboisColor(), float eyeSeparation and bool anaglyph to class Renderer
Opinion?

src/builtin.cc
koendv: We can drop this variable. I have used it like this:
```
color($anaglyph?"Yellow":"Red")
cube(10);
```
but now the dubois algorithm modifies colors automatically, so this variable is no longer needed.

src/dubois.cc
koendv: Or this code could be added to Renderer as a function?

koendv:
I think this will be a fun and useful feature. Let's do it.

t-paul:
I've got me some of those cheap paper cyan/red glasses, so I can literally have a good look at the PR :joy:.

koendv:
:) 
[colorscheme.zip](https://github.com/openscad/openscad/files/6136733/colorscheme.zip)

koendv:
- added command line option --projection=(a)naglyph
- added build tests
- scaled "eye separation" setting so default is middle of scale

PR #3696, opened 2021-02-28
Fix initial default console font to pick up correct font family and size.

thehans:
Initial default font (before preferences are saved) should now behave same as 2019.05.

If previous version of OpenSCAD with console font selection has already been run, then will require clearing `consoleFontFamily` and `consoleFontSize` from `~./.config/OpenSCAD/OpenSCAD.conf` (linux) 
or corresponding registry keys (windows).

For @MichaelAtOz :gift:


thehans:
To verify correctness, I first tried changing my various Linux Mint system fonts, and checking if 2019.05 release displayed differently in console.

Mint settings didn't change anything, so that wasn't where the font came from.  Then I realized Qt5 has its own global settings which can be set in an app "Qt5 Settings" (labled in my ) aka "Qt5 Configuration Tool": `qt5ct`

Here is me setting the default to an unmistakably different font family and size:
Screenshot from 2021-02-27 21-54-24

Then viewing from 2019.05 OpenSCAD (installed as a .deb from mint/ubuntu repos):
Screenshot from 2021-02-27 21-49-29

And comparing with my local build (with OpenSCAD.conf settings cleared to remove previously set default):
Screenshot from 2021-02-27 21-53-19

Note that my `qt5ct` changes didn't seem to have any effect on my *AppImage* version of OpenSCAD 2019.05, I guess those bundle their own qt5 config settings.

t-paul (approved):

Issue #3411, opened 2020-08-19
Feature Request: "segmentation" options for linear_extrude


thehans:
I figured I'd open an issue for this since there's currently a related  [discussion topic on the mailing list](http://forum.openscad.org/Problem-using-difference-with-linear-extrude-tc29998.html).

The problem is when using `linear_extrude` with `twist` (and to a lesser extent a non-uniform `scale` parameter) on objects with long thin edges, such as this:
```
linear_extrude(50, twist=360,slices=20) square([10,2], center=true);
```
Results in skinny triangle faces with large differences in their normal vectors.  Attempting to increase `slices` only makes the triangles skinnier, but the zigzag ripples are still there.
If the long edges were divided into smaller colinear line segments, this annoying modeling artifact would go away.

So I think that a builtin option for `linear_extrude` to split these long edges would be pretty helpful, and fairly simple to implement.  The main question is how we want to design the interface: what new parameter name(s) to use and how they work?

Based on previous discussion (see quotes below), I'm thinking that a new parameter should be added to `linear_extrude`, call it `max_edge_length` or something like that, and it would function analogously to how `$fs` does on circles: 
any edge longer than `max_edge_length` would be divided into `n` equal segments, where `n` is roughly `current_edge_length / max_edge_length` (I say roughly since we could use ceil or round on the result, but I'm not sure which would be preferred)

Alternatively we could have a boolean `split_edges` parameter, where setting to `true` would let this "segmentation" feature use `$fs` (and maybe `$fa`?) **or** `$fn` to calculate edge segment lengths, depending on which are defined in the current scope.
If this approach were used, then `split_edges` should probably default to false for backwards compatibility.

There was also some comments about this in my PR #3351, but I thought it would be better to separate this into its own issue / PR for further discussion.
For reference, I've quoted the relevant comments from that PR below:

_Originally posted by @nophead in https://github.com/openscad/openscad/pull/3351#issuecomment-643145170_
> When I extrude a shape like that I add extra vertices along along the edges so the quads are roughly square. That gets rid of the ugly stripes that give a rippled appearance. It would be nice if linear extrude did that when necessary.

_Originally posted by @thehans in https://github.com/openscad/openscad/pull/3351#issuecomment-643523179_
> @nophead I've been thinking about that issue too.  
> Particularly when long, thin rectangle is extruded with twist, it ends up looking like a staircase no matter how many slices, because the long edge just sweeps too much for its length.
> 
> A small example:
> ```
[...]

PR #3698, opened 2021-02-28
Preserve colinear edges in 2D.

thehans:
Related to #3411 (doesn't directly add the feature, but makes user workaround a little easier).

Also brought up again in recent forum discussion:
http://forum.openscad.org/Better-way-of-creating-a-curved-slope-tp31988p32021.html

Sample script:
```OpenSCAD
linear_extrude(50, twist=360) polygon([for(x=[0:10]) [x,-0.1], for(x=[10:-1:0]) [x,0.1] ]);
```

Before:
colinear_before

After:
colinear_after


thehans:
I don't think there is any reason to keep this around after #3770

PR #3699, opened 2021-03-01
WIP: Upload files to DO Space instead of droplet

jfhbrook:
Hey hey,

Many moons ago, DigitalOcean introduced an S3-compatible product. Currently, this project uploads build artifacts (.dmg files, etc., ) to a DigitalOcean droplet (that I'm hosting) running nginx and taking uploads via scp, but I think Spaces is a better fit for this problem.

I set up a space for this and pointed a domain I own to point to it as a CDN. Unfortunately, I think I need to own the DNS in order to make a subdomain for openscad.org work, but I think openscad.thisistheremix.dev will be OK. Happy to use a different domain, as long as I can make my DigitalOcean account control the TLD.

In order to use this new process, you'll need an AWS style access key ID and secret key. I can hand those out to whoever. I don't have keys to files.openscad.org anymore, but am happy to use that machine to transfer working creds for that bucket given that someone can upload my current machine's pubkey. Let me know.

I haven't considered data migrations at all, because I don't know the scope of what would need to be moved and because I don't have keys to files.openscad.org to actually do the migration with. If I can get access, I can write the script to do the migration and make it happen.

Aside from all of that, what this needs now is testing. I'm trying to test the MacOS release process locally, since that's the machine I happen to be on, but the others will need their levers pulled as well.

### DIFF
--- a/Info.plist
+++ b/Info.plist
@@ -113,6 +113,6 @@
 	<key>SUPublicDSAKeyFile</key>
 	<string>dsa_pub.pem</string>
 	<key>SUFeedURL</key>
-	<string>https://files.openscad.org/appcast.xml</string>
+	<string>https://openscad.thisistheremix.dev/appcast.xml</string>
 </dict>
 </plist>
--- a/Info.plist.in
+++ b/Info.plist.in
@@ -103,6 +103,6 @@
 	<key>SUPublicDSAKeyFile</key>
 	<string>dsa_pub.pem</string>
 	<key>SUFeedURL</key>
-	<string>https://files.openscad.org/appcast.xml</string>
+	<string>https://openscad.thisistheremix.dev/appcast.xml</string>
 </dict>
 </plist>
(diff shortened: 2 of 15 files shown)


jfhbrook:
I've been told that the current architecture of artifact hosting involves a cron task on the DO droplet pulling artifacts from CircleCI, so this won't work as-is.

I'm willing to put work into overhauling this more thoroughly. Like, perhaps I set up a real S3 bucket and a [scheduled lambda](https://docs.aws.amazon.com/eventbridge/latest/userguide/run-lambda-schedule.html) + IAM roles so everyone can access them in the UI? That's more work than I was counting for and it's pretty clear that I've lost track of the thread in between when I set things up aeons ago and decided to finally revisit it today. Let's get a conversation going and figure out what will actually work for everyone.

jfhbrook:
An alternate approach that sounds viable after talking to teepee is to stick to the current system but build out a fresh box and devops it up a little. The current pricing of the droplet is a non-problem and I think that once scheduled tasks come into the equation s3 + lambdas struggle to justify themselves at the scale y'all need.

In my head what this would look like is pretty similar to what exists today - with no script updates needed - but with some ansible/terraform files and so on to maintain the instance, maybe improved monitoring/alerting and a circleci hook to automatically update the instance on pushes or something. My weekend is over but I'll try to build out something next weekend!

kintel:
The current box is (or rather was) maintained with an ansible playbook, but since it was the only thing I was using ansible for, it kind of fell out of maintance. I have a private repo somewhere that I'll add you to. It probably has very little value, but there could be some nice pieces there that you want to keep; like public ssh keys.

jfhbrook:
Oh word! Yeah OK I'll take a look! and yeah good to know they fell out of maintenance - maybe some shell scripts would be better given how things are used. I'll think about it and come up with something!

Issue #3616, opened 2021-01-15
--hardwarnings causes warnings to terminate the command line mode but the exit code is 0


nophead:
I expected the exit code to be 1, the same as errors when --hardwarnings is active as it promotes a warning to an error in my mind.

Tomatower:
Is this related to #3660 / #3646?

julianstirling:
Some more detail

If creating an STL `--hardwarnings` does create a non-zero exit code as there is no top level geometry. If creating say a PNG it exits with 0, an any script running it is unaware of the problem.

We have just his this problem on our automatic renders. We expected the script to say it had failed like it does on the automatic builds.

PR #3743, opened 2021-04-12
support --hardwarnings for PNG and ECHO output (#3616)

beerriot:
As reported in #3616, --hardwarnings didn't always cause openscad to
exit with a non-zero error code when a warning was encountered. It did
for STL (and likely other geometry outputs) because those exports exited
with an error code if the root node was empty, but PNG and ECHO don't
check that.

I thought that relying on the geometry check seemed a little accidental,
so the fix I've made is to prevent the file context initialization from
swallowing the HardWarningsException.

I also fixed the hardwarnings test. The test passed only because it was
running the openscad executable with invalid commandline parameters.
OpenSCAD printed usage and exited with an error. The SUFFIX parameter is
not supported by the Cmake add_failing_test function, and shouldfail.py
does not automatically add a - for stdout output.


beerriot:
FWIW, I think the other tests added via `add_failing_test` may also be accidentally passing, but I wanted to keep this PR focused on the one issue.

t-paul (approved):
Thanks!

PR #3744, opened 2021-04-15
Improved editor call-tips (fixes #3655).

### LINKED ISSUE #3655: 2021.01 - fix/improve editor call-tips
I made this a bit generic, so if other's notice anything similar they can be mentioned here. (?)
For next bug-fix release.

Call tip | Issue
------- | -------
 [  ] text | missing `font=` duplicate `string`, consider `[text=]"string"`

---------------------------
text() call-tip, missing `font=`, duplicate `string`, 
e.g:
Capture-text-parameters

debatable whether it should say `text([text=]"string", ...`

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/96480359-2021-01-fix-improve-editor-call-tips?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

abshk-jr:
In text() call-tip, removed duplicate `string`, added `text=""` and `font=""`
changed the position of `spacing = 1` to match similar code in other files.

### DIFF
--- a/src/text.cc
+++ b/src/text.cc
@@ -110,6 +110,6 @@ void register_builtin_text()
 {
 	Builtins::init("text", new TextModule(),
 				{
-					"text(text = \"\", size = 10, spacing = 1, font = \"\", halign = \"left\", valign = \"baseline\", direction = \"ltr\", language = \"en\", script = \"latin\"[, $fn])",
+					"text(text = \"\", size = 10, font = \"\", halign = \"left\", valign = \"baseline\", spacing = 1, direction = \"ltr\", language = \"en\", script = \"latin\"[, $fn])",
 				});
 }
--- a/src/text.cc
+++ b/src/text.cc
@@ -110,6 +110,6 @@ void register_builtin_text()
 {
 	Builtins::init("text", new TextModule(),
 				{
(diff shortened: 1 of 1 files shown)


t-paul (changes requested):
I'd propose a small change as the following code:

https://github.com/abshk-jr/openscad/blob/1bc96a09c62f5ccb85b9237d1d6f44e51701dd62/src/text.cc#L51-L55

means that it's possible to give `text`, `size` and `font` as positional args in this order. There's no good way yet to indicate that in the call-tips, but giving those 3 first is probably helpful.

I think that's what @MichaelAtOz meant with
> debatable whether it should say text([text=]"string",

but if we do that, we probably want to update that for all other calls too. So I think that's a topic for a different PR.

Thanks for the attention to detail, adding the reference to the original issue.

abshk-jr:
Thank you for the information, I've updated the file and the new order is `text`, `size`, `font` and `(optional args)`.

I've left the order for optional arguments as they were beforehand.
Please let me know if any other changes are required.

t-paul (approved):
Thanks for the update, looks good.

Issue #569, opened 2013-12-11
relative import paths in included files are relative to including file


kborer:
bash> find
.
./scad
./scad/included_file.scad
./including_file.scad
./stl
./stl/cube.stl
bash> cat including_file.scad
include < scad/included_file.scad>
bash> cat scad/included_file.scad
import("../stl/cube.stl");

Given the preceding file structure, the expected result is that both including_file.scad and included_file.scad will compile correctly. Actual behavior is that included_file.scad compiles correctly, but including_file.scad fails to find the cube.stl file. 

This is because the import path in included_file.scad is relative to included_path.scad. When included_file.scad is compiled by itself, the path correctly  leads to cube.stl. When including_file.scad is compiled, the import path is relative to including_file.scad, so cube.stl is not found.
##

t-paul:
Fixed via #3745.

Issue #3174, opened 2019-12-28
Capture Memory Leak


rcolyer:
Capture from function literals leaks memory.  Repeatedly previewing with the leaker line uncommented leaks 2GB more each preview.  Having just the nonleaker line uncommented does not do this.

```
function nonleaker(v) =
  0;

function leaker(v) = let(
    capture = function() 0
  )
  0;

//x = [ for (j=[1:100000]) nonleaker(rands(0,1000,200))[0] ];
x = [ for (j=[1:100000]) leaker(rands(0,1000,200))[0] ];
```

thehans:
Here's a single line simplified case which leaks about 50MB per F5 preview.
```
x = [for (i=[1:100000]) let(capture = function() 1 ) 0 ];
```

I've also built OpenSCAD using LLVM's LeakSanitizer and made this log of the results (scroll down to the first "Indirect leak" to see the more significant leaks):
https://gist.github.com/thehans/9a2b58c9f400730c5223b8f67b1a8c19

The build was basically created using clang and enabling `-fsanitize=address` during compile and link (and linking with clang, not ld ).  More info in the [AddressSanitizer](https://clang.llvm.org/docs/AddressSanitizer.html) documentation.  LeakSanitizer is automatically included as part of AddressSanitizer.

thehans:
From some discussion on IRC, its suspected that the issue relates to `shared_ptr` objects forming cycles somewhere, which are then unable to be automatically destroyed in the normal way.

But I still haven't pinpointed the exact cause yet.

rcolyer:
The cause of this appears to be that in value.h, class FunctionType for function literals stores a shared_ptr<Context>.  Context, defined in context.h, contains a std::unordered_map<std::string, ValuePtr> (where ValuePtr in value.h is derived from shared_ptr) of every variable, including the function literal itself.  This produces a direct shared_ptr loop which preserves the variables in the context of the function literal until program termination.

I was not able to spot a simple resolution.  It appears to be a complicated side-effect of the closure nature of function literals requiring their captured contexts to stick around as long as the function literal does, combined with the extremely varied ways that function literals can be returned past the lifetime of the function.  Because multiple function literals in the same context can call each other, and multiple function literals can be returned (as the function literals can reside inside of vectors), it seems to be a complicated issue of redefining ownership for this data in some more sensible manner.

If anyone has ideas of how to do this, please speak up or help.

kintel:
[...]

Issue #3672, opened 2021-02-15
positional parameter silently overrides previous named parameter


jordanbrown0:
```
module foo(a) {
    echo(a=a);
}

foo(a=1, 2);
```

results in

```
ECHO: a=2
```
It sure seems like this should be a warning, just like `foo(a=1, a=2)` is.

thehans:
I worked on fixing this sort of thing as part of #3161 but it was out of scope for the original issue and never ended up getting merged.  The `resolveArguments` overhaul was pretty comprehensive in terms of avoiding these sort of odd overrides or warning when they occur.

thehans:
Fixed by #3745

PR #3745, opened 2021-04-17
Context cleanup

redlizard:
I reworked and cleaned up much of the evaluation subsystem, in preparation for several bugfixes and language features I plan to work on.

This rework has two main goals. One is to tighten up the way data flows through the evaluation subsystem, and to make this more explicit and transparent. I have tried to make sure functions and data structures store/receive exactly those pieces of data that is relevant to them, improving interface orthogonality and separation of concerns. To that aim I introduced data structures such as Arguments, Parameters, Children.

The other goal, specifically in service of https://github.com/openscad/openscad/issues/3174 , is to make the role of Contexts as simple as possible. In this rework, a Context is a nested interpretation of terms that can be captured by a function literal, and Contexts are only created when both these requirements apply.

This branch fixes a number of minor language interpretation bugs that we discussed recently, as a side effect of regularizing and refactoring things. I also intend to look over the bugtracker for old issues that have now become low hanging fruit, of which I think there are a fair few.


redlizard:
This rework fixes issues #569, #1968, #2768, and #3672, and obsoletes #3161.

tests/regression/echotest/children-warnings-tests-expected.echo
```diff
@@ -1,6 +1,5 @@
 DEPRECATED: child() will be removed in future releases. Use children() instead.
-WARNING: Negative child index (-1) not allowed in file children-warnings-tests.scad, line 1
-WARNING: Child index (5) out of bounds (1 children) in file children-warnings-tests.scad, line 4
-WARNING: Child index (5) out of bounds (2 children) in file children-warnings-tests.scad, line 4
+WARNING: Children index (5) out of bounds (1 children)
```
t-paul, line 2: Looks like this lost the file/line-number info. Can we get that back?

tests/regression/echotest/rands-expected.echo
```diff
@@ -1,8 +1,13 @@
+WARNING: rands() parameter could not be converted in file rands.scad, line 12
```
t-paul, line 1: The extra warnings are nice. Is there more info available?
* name or position of the parameter
* target type of the conversation

t-paul (approved):

tests/regression/echotest/children-warnings-tests-expected.echo
redlizard: Yes. It was previously present on child(), but not on children(); those code paths got merged, and were based on the children() path. But it could easily be added back.

tests/regression/echotest/rands-expected.echo
redlizard: Yes, that information is available. The existing code doesn't have that and I didn't add it, but it could be added easily. No parameter names, because builtin functions mostly do not use named parameters, but parameter position and expected type could be added.

A minor complication is that some functions have multiple possible signatures, which would make the expected parameter type a bit ambiguous. We might need to write some custom warning messages for those cases; for most other cases, the general func.cc check_arguments() function could just add the expected type to the warning message.

t-paul: Ah, good point. As those here are new warnings, we probably should leave it for a separate PR then as it would likely also affect a couple of other places.

I'm thinking if we would have more structured info in code about the parameters, we could generate the documentation instead of the inline text in `Builtins::init()`. That would even bring things one step closer to calltips for user functions.

redlizard: Well too late, I added them :)

jordanbrown0: > No parameter names, because builtin functions mostly do not use named parameters,

Perhaps that should change; perhaps they should always have names for their parameters.

t-paul: While it's possible, it's not really important for things like the math functions. I guess we can have `sin(angle = 30);` but that's not adding much.

redlizard: > Perhaps that should change; perhaps they should always have names for their parameters.

I agree that this would be better. There are a few builtin functions for which this would be tricky, though: min() max() str() chr() are variadic, and log() has a signature of log([base, ] number), and neither of those really fits in the notion of named parameters currently in use -- we'd have to switch to a substantially more elaborate model of parameters to support those.

MichaelAtOz: > 
> 
> > No parameter names, because builtin functions mostly do not use named parameters,
> 
> Perhaps that should change; perhaps they should always have names for their parameters.

If you mean mandatory names then -1; I've already had two carpal tunnel operations & I'd prefer to avoid joint replacements.

jordanbrown0: Not mandatory for the user.  Mandatory for the definition of the function, so that the infrastructure can always know the name of a parameter - in the same sense that `cube()` has `size` and `center` parameters even if you invoke it as `cube([1,1,1], true)`.

PR #3749, opened 2021-04-19
Add functionality for Indentation of Collapsed blocks (fixes #2371).

### LINKED ISSUE #2371: Feature Request: Indent Collapsed Blocks
When the 'indent' function on the editor is used on a collapsed block, it only indents the first and last line.  It would be nice if it indented the entire collapsed section instead.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/56649038-feature-request-indent-collapsed-blocks?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

abshk-jr:
This adds functionality for Indenting Collapsed Block/lines with Parent as requested in #2371.

Earlier, applying the Indent feature only Indented the first line of the collapsed/folded block.
oldindent

Now, with the addition of this, indentation gets applied to the collapsed/folded lines as well.
newindent


t-paul (commented):
I did some tests and it works well. Slightly scary is the end of the loop relying on at least returning one visible line again. It does seem to work even when collapsing a range going to the last line of the text. Do you have found any documentation this is guarantied? I only found the note that the first line is always visible. Is it maybe returning some kind of error (-1?) if the give line does not exist?

It would be nice to integrate with the Tab key at some point, but I think that should be handled separately.

However it would be nice to see 2 improvements that are related but not only related to the new code:
- Undo currently goes line by line, as user I would expect to undo the whole indent in one step. This should be possible with `qsci->beginUndoAction()`/`qsci->endUndoAction()` and would improve the current behavior too.
- It would be nice to not indent empty lines. The current Tab handler does that already, our existing Indent implementation does not. So empty lines would stay empty instead of collecting spaces/tabs according to the indenting level.

abshk-jr:
SCI_GETLINEVISIBLE returns 1 for a non-existing line.
But now the code depends on the length of the line to check if it doesn't exist or is empty.
This is made possible by using SCI_LINELENGTH which returns 0 for a non-existing line and 1 for an empty line.

The code now doesn't apply indentation on Empty lines
os_empty1
os_empty2

Also, the Undo behavior now works on the whole indent instead of going line by line.
os_undo

t-paul (commented):
Nice, this way it's using a clearly documented value. A shortcut could be `qsci->lineLength()` (which has different return values!) but as we are using `SendScintilla()` here already, that's fine.

I found one corner case, and that's with the text just ending with a '}' on the last line. I'm not sure but I suspect this is caused by all other lines having a NewLine character at the end (hence the test for <= 1) but that last line maybe just has the `}', so it's skipped due to line length of 1?

The documentation says "If you want the length of the line not including any end of line characters, use SCI_GETLINEENDPOSITION(line) - SCI_POSITIONFROMLINE(line)." - maybe that's the solution, or maybe even using the simpler `lineLength()` already covers that.

abshk-jr:
`qsci->lineLength()` returns 1 for an empty line as well as for a line ending with  '}' 
this result is the same as using `SCI_LINELENGTH`

So I instead used `SCI_GETLINEENDPOSITION(line) - SCI_POSITIONFROMLINE(line)` as recommended 
This allows for indentation of '}' if present  at the last of the document.

t-paul (approved):
Thanks! This turned out to be a bit more tricky than originally anticipated, but that seems to happen at times with GUI topics.

PR #3755, opened 2021-05-01
Customizer rework

redlizard:
This branch redesigns both the UI and the internal structure of the customizer.

It integrates the customizer with the tabbed editor framework, so that there is one customizer widget associated with each editor tab. As a consequence, it is not necessary for the customizer widget to store and reload its settings to disk when switching tabs. Instead, the customizer persistently stores the configured parameter values for everything configured using its editor, and only saves this data to disk when the editor tab itself is saved. As a consequence, the customizer now behaves much more predictably when switching tabs, and when source file edits require a change in customizable variables.

This branch also changes the code structure of the customizer to entirely separate the interpretation of parameter constraints from the assorted editor widgets. It adds a ParameterObject subclass metaobject for each supported type of parameter, containing the pre-parsed constraints applicable to that particular parameter type.

It removes support for editing vector parameters other than vectors of 1-4 numbers. This functionality was very buggy in a hard-to-fix way, not supported by MakerBot, and not very valuable, because it only presented a text editor widget where you could edit the raw openscad code for the variable (which is not hugely more useful than just editing the source code directly). This can be added back later, if someone can think of an editor component for arbitrary vectors that is both reliable and useful.


MichaelAtOz:
> Instead, the customizer persistently stores the configured parameter values for everything configured using its editor, and only saves this data to disk when the editor tab itself is saved.

Currently with unsaved source edits, the editor saves to the backup location on F5 & F6. I presume that still happens, and does/should the corresponding parameters be backed-up?

redlizard:
> Currently with unsaved source edits, the editor saves to the backup location on F5 & F6. I presume that still happens, and does/should the corresponding parameters be backed-up?

Yes, yes, and yes.

redlizard:
Good call. Both issues fixed.

Issue #3567, opened 2020-12-31
linear_extrude-scale-zero-tests fail for CGAL render


thehans:
I just ran ctest -C All to see if there were any newly fixed or failing tests since last release, and I saw that these tests are now failing:
 - `cgalpngtest_linear_extrude-scale-zero-tests`
 - `csgpngtest_linear_extrude-scale-zero-tests`

These aren't reported as errors from regular CI tests, since they are omitted from the default test suite due to time constraints.

The problem is that many of the cases in this test generate non-manifold geometry when scaled to 0 with linear_extrude.

And due to #3526, this bad geometry is now detected earlier on and returns empty geometry instead of attempting the implicit top level union on them.
As intended by that PR, the test now emits many messages of:
`ERROR: The given mesh is not closed! Unable to convert to CGAL_Nef_Polyhedron.` for those objects.

I guess CGAL managed to not crash on them before, possibly since none of the extrudes actually intersect with neighbors, but the question remains what to do with these type of zero scaled extrusions.

There was some previous discussion in PR 3351, regarding what to do with these types of issues caused by linear_extrude with scale=0, starting around here: https://github.com/openscad/openscad/pull/3351#issuecomment-644874848 but a conclusion was not reached.

Here's a screenshot clipped from the test report:
![Screenshot from 2020-12-31 17-48-09](https://user-images.githubusercontent.com/252233/103431091-6f52c000-4b90-11eb-82a6-1faaef078539.png)

kintel:
@pca006132 An interesting observation here is that this test works with fast-csg, but fails with Manifold and CGAL Nef polyhedrons. I don't really expect Manifold to manage non-manifold objects, but it could be an interesting corner case.

pca006132:
Not sure about the linear extrude implementation in OpenSCAD, the constructor in manifold will produce a self-intersecting geometry but the result is still a manifold (topologically speaking...). How does Manifold fail with this? Is it throwing an exception, return some weird stuff or segfault? I would expect one of the first two but not segfault.

kintel:
I haven't investigated in detail, but the OpenSCAD symptom is this log output, and no resulting geometry:
```
ERROR: [manifold] Input mesh is not closed!
ERROR: [manifold] Surface_mesh -> Manifold conversion failed: NotManifold
```

pca006132:
ok, this looks fine

kintel:
Status as of 2024-03-24: Still broken, also with Manifold.

Manifold issue is #5046

[...]

PR #3770, opened 2021-05-13
Add "segments" parameter for linear_extrude.

thehans:
This PR addresses issue #3411 for linear_extrude by splitting long edges into multiple segments when called with twist and/or non-uniform scaling.

#### How segments are calculated
By default it will use values from $fn or ($fs and $fa).  
If shape being extruded has multiple outlines / holes, each outline is processed independently.

All methods of calculating segments incorporate a concept of each edge having a "max edge length", across all slices.
  If scale is uniform and < 1 then the max edge length is the same as the original edge length.
  If scale is uniform and > 1 then the max edge length is after applying the full `scale` value.
  If twist *and* non-uniform scale is used, then the determination is non-trivial, requiring calculation of edge length over each slice.

- `$fn`:  Set up some metadata about edges, each starting with a `segment_count=1`.  While (total segments < fn), increment segment_count for edge with largest value of (max_edgelen / segment_count).
- As usual, if `$fn = 0` then whichever between `$fs` and `$fa` results in less segments is used for the result.
- `$fs`:  Each edge is split into `ceil(max_edgelen / $fs)` number of segments.
- `$fa`: Works the same as setting `$fn = ceil(360 / $fa)`
- An additional `segments` parameter can also be used to override `$fn` | `$fs` | `$fa`, analogous to how `slices` works.  This also uses the same calculation method as `$fn`.
Edges may be split up into additional segments, but no edges are ever removed, so the segmenting feature can be disabled by simply setting `segments=0`.

#### Other changes
 - CSG output for linear_extrude was made a little less verbose by omitting parameters with default values.
 - Some examples/tests were updated with more appropriate values for nice smooth results.

#### Further possible improvements
 - **RFC** regarding the chosen parameter interface here: defaulting to using special vars, and overriding with `segments`.  I think that in most cases non-segmented extrudes will be vastly improved by this using good `$fs` and `$fa` values with this change, and previous results were undesirable enough that it makes sense to default ON.  Ability to disable (via `segments=0`) is *available* in case user is going for a very specific "style".
 - IMPLEMENTED: ~~It was suggested in IRC to have `$fn` / `$fa` / `segments` group same-length edges and only split one if all of the same-length can be split.  So for example a an equilateral triangle (or other regular polygon) will stay 100% symmetrical even when `$fn` is not a multiple of the number of sides.  (In the case where `$fs` is the limiting variable, it automatically keeps things symmetrical)~~
 - There is an existing issue with scaling a single axis to 0 (non-uniformly).  This forms a ridge line at the top, and edges to each side of the ridge do not necessarily match up (each edge should belong to exactly 2 faces).  This can lead to a polyhedron which is not considered closed, and cannot be converted to CGAL Nef.
 See `testdata/scad/3D/issues/issue2259.scad` (there are already 2 commented out tests which fail, and using segmentation can make this effect worse (two more would fail so I've effectively disabled segments in this test).
I have some vague ideas of how this could be dealt with by special crafting of the faces between top ridge and final slice, but it was tricky enough that I put it off for now.
 - Similarly but probably requiring another independent solution, scaling *both* axes to 0 (uniformly), when the shape is not simple (has a hole, or multiple disjoint outlines, etc), will also result in non 2-manifold shapes (related: #3567)


thehans:
Update: when splitting by a fixed number of segments (`$fn`, `$fa`, or `segments`), similar length segments are now grouped to be split at the same time.  Splitting stops when the limit of total segments would be exceeded by the next group of longest edges.

Edges are grouped when segment length of smallest / largest is >= 0.9

src/linearextrude.cc
```diff
@@ -187,6 +203,6 @@ void register_builtin_dxf_linear_extrude()
 
 	Builtins::init("linear_extrude", new BuiltinModule(builtin_linear_extrude),
 				{
-					"linear_extrude(number, center = true, convexity = 10, twist, slices = 20, scale = 1.0 [, $fn])",
+					"linear_extrude(height = 100, center = true, convexity = 10, twist = 0, slices = 20, scale = 1.0 [, $fn])",
```
rcolyer, line 206: Mismatching defaults

rcolyer (approved):
The code looks good, and it tests very well with very aesthetically pleasing results.  All concerns and issues were sorted out on the IRC channel, and I approve of the current state (pending a minor tooltip fix).

Issue #3617, opened 2021-01-15
Customiser sliders seem to have broken


nophead:
![image](https://user-images.githubusercontent.com/566149/104777267-4a0aa800-5773-11eb-8a52-901b49af9720.png)

```
X = 0; // [-1 : 0.1 : 1]
Y = 0; // [-1 : 0.1 : 1]
extruder = -1; //[-1 : 3]

/* [Hidden] */
```

The Y slider position in the picture corresponds to the minimum, which is -1, but the spinbox is -0.9 and that is value being used. Dragging it to 0.8 causes it to jump back to the 0.9 position with the spinbox showing 0.8.

Sometimes moving slider does not update the preview and sometimes it does. Sometime it matches the spinbox and sometimes it doesn't. 

Clicking on the track of the slider makes it jump and update the spinbox but doesn't update the preview. 

I am pretty sure it used to work until recently. Adjusting via the spinbox seems reliable.

OpenSCAD 2021.01-RC6 Win7 / 64bit

thehans:
@nophead I just created PR #3778 which should address the issues you list here.
I also changed the automatic preview behavior a bit (details in PR comment), and would appreciate any testing/feedback.  
I've only tested on Linux Mint 20, so it would help to verify on other platforms.

nophead:
Seems a bit broken to me, but so is a recent snapshot: OpenSCAD 2021.05.11.ci7927 I was using, I must just have not used the customiser with it until now. 

It now shows all the variables included from other scad files in the customiser whereas previous versions only showed variables defined in the main file. It also doesn't stop at my `/* [Hidden] */` line and shows too many variables in my main file. Even if I move my customiser definitions to the top of the file it still shows the includes after them.

It also seemed to create a new set of values called "New set 1" of its own accord and changed some array values to 0 breaking my code as they are constants that should never be 0. 

It also marked the file as modified but I didn't have the editor visible, so how can it be modified? Changing a customiser setting now seems to make it look the file is modified.

nophead:
Also $variables seem to be broken now. Setting them on the command line no longer seems to work.

redlizard:
As you say, these issues are part of an unrelated recent customizer development at #3755.

Showing variables that should not be parameters as part of the customizer is a bug, proposed fix above.

The "New set 1" system is a deliberate change introduced in #3755. The "<design default>" set is now immutable, allowing you to always reset things to what the source file says; attempting to modify the design default set will instead create a New Set.

[...]

PR #3780, opened 2021-05-19
Make sure include<>d variables and variables outside the initial section are not interpreted as customizer parameters

redlizard:
Fix customizer variable identification issue identified in https://github.com/openscad/openscad/issues/3617#issuecomment-843587035 .


thehans:
Included as part of #3778

PR #3783, opened 2021-05-23
update qt api usage to conform with 5.15

sielicki:
various deprecation fixes, should eventually make a port to qt6.2 much easier (when it's eventually released -- waiting on multimedia).

### DIFF
--- a/src/FontListTableView.cc
+++ b/src/FontListTableView.cc
@@ -48,7 +48,7 @@ void FontListTableView::startDrag(Qt::DropActions supportedActions)
 	mimeData->setText(text);
 
 	QFontMetrics fm(font());
-	QRect rect(0, 0, fm.width(text) + 8, fm.height() + 8);
+	auto rect = fm.boundingRect(text);
 	QPixmap pixmap(rect.width(), rect.height());
 	pixmap.fill({240, 240, 240, 160});
 
--- a/src/editor.cc
+++ b/src/editor.cc
@@ -7,8 +7,8 @@ void EditorInterface::wheelEvent(QWheelEvent *event)
 	QSettingsCached settings;
 	bool wheelzoom_enabled = Preferences::inst()->getValue("editor/ctrlmousewheelzoom").toBool();
 	if ((event->modifiers() == Qt::ControlModifier) && wheelzoom_enabled) {
-		if (event->delta() > 0) zoomIn();
-		else if (event->delta() < 0) zoomOut();
+		if (event->angleDelta().y() > 0) zoomIn();
+		else if (event->angleDelta().y() < 0) zoomOut();
 	} else {
 		QWidget::wheelEvent(event);
 	}
--- a/src/MainWindow.h
+++ b/src/MainWindow.h
@@ -42,7 +42,7 @@ public:
 
 	QTimer *autoReloadTimer;
 	QTimer *waitAfterReloadTimer;
-	QTime renderingTime;
+	QElapsedTimer renderingTime;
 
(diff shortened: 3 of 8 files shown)


t-paul:
Thanks for the update, we certainly want to have those deprecations updated.

Looks like there's a surprising difference in Travis vs. Github builds. Looking at the defines collected by `cmake`, I'm seeing a number of differences we need to clarify:

Travis builds with extra flags: `OPENSCAD_NOGUI`, `OPENSCAD_SNAPSHOT`
Github with extra flags: `USE_QOPENGLWIDGET`, `USE_SCINTILLA_EDITOR`, `ENABLE_MDI`, `ENABLE_DBUS`, `ENABLE_JOYSTICK`

So I guess it's mostly due to `NOGUI` that Travis succeeds. I suppose this means we need to add conditionals to the cases where the build fails for older distros. With Ubuntu 16.04 now EOL, we don't need to consider that anymore though.

src/FontListTableView.cc
```diff
@@ -48,7 +48,7 @@ void FontListTableView::startDrag(Qt::DropActions supportedActions)
 	mimeData->setText(text);
 
 	QFontMetrics fm(font());
-	QRect rect(0, 0, fm.width(text) + 8, fm.height() + 8);
+	auto rect = fm.boundingRect(text);
```
sielicki, line 51: is this actually a (roughly) equivalent statement?

src/editor.cc
```diff
@@ -7,8 +7,8 @@ void EditorInterface::wheelEvent(QWheelEvent *event)
 	QSettingsCached settings;
 	bool wheelzoom_enabled = Preferences::inst()->getValue("editor/ctrlmousewheelzoom").toBool();
 	if ((event->modifiers() == Qt::ControlModifier) && wheelzoom_enabled) {
-		if (event->delta() > 0) zoomIn();
-		else if (event->delta() < 0) zoomOut();
+		if (event->angleDelta().y() > 0) zoomIn();
+		else if (event->angleDelta().y() < 0) zoomOut();
```
sielicki, line 11: `angleDelta` was preferred here as the [documentation claims there's an issue](https://doc.qt.io/qt-5/qwheelevent.html#pixelDelta) with `pixelDelta` on X11.

sielicki:
[18.04 is stuck on 5.9](https://packages.ubuntu.com/search?keywords=qtbase5-dev) and unfortunately a few of these changes were [introduced in 5.10](https://doc.qt.io/qt-5/qplaintextedit.html#tabStopDistance-prop).

I'm not sure what the right path forward is, outside of polluting with `ifdev`s. Maybe we just close this?

t-paul:
Yeah, having lots of defines is a bit annoying, but I guess that's the price for both moving forward while still allowing people to build on older platforms.

For things that are luckily very specific and unlikely to clash with anything else, e.g. `setTabStopDistance` vs. `setTabStopWidth`, it seems ok to just leave the new name in there and have a `#define` at the top redirecting it for older Qt versions. That way the code effective on new platforms is pretty much clean and at some point we can just drop the rename.

If you don't like that, the probably neater option would be introducing a list of helper functions in a separate file which handles the switch in a central place, we we'd call `qtcompat::setTabStopDistance(e, tabStopWidth);`. We did not do that so far but who knows what Qt6 will bring eventually. Note that we also need to consider the C++17 switch for that, as OpenSCAD is currently at C++14. But with Ubuntu 16.04 gone, C++17 seems doable.

t-paul:
Superseded by #3896. Dropped 16.04 now.

PR #3784, opened 2021-05-23
Remove loop causing EXC_BAD_ACCESS on QApplication::quit()

TheJKM:
Looks like I found the reason for the crash reported in #3775. It happens as soon as something calls `QApplication::quit()`, which is done by the `Cmd+Q` command and also by the menu entry "Quit". 
It happens in [`src/openscad.cc` in line 857](https://github.com/openscad/openscad/blob/176ea07abba6f5f8fb5a130143f23b397416c789/src/openscad.cc#L857). The loop iterates through all main windows and frees their memory. On macOS, I get an EXC_BAD_ACCESS on requesting the vector of all main windows.
openscad_mac_crash
I think this loop is rather useless, because `MainWindow` has the attribute `Qt::WA_DeleteOnClose` set [(file mainwin.cc, line 272)](https://github.com/openscad/openscad/blob/176ea07abba6f5f8fb5a130143f23b397416c789/src/mainwin.cc#L272), which makes their close routine call `delete` automatically.
I did some tests on Manjaro Linux with gdb and Qt 5.15.2, and it is throwing a segmentation fault at the same call if I'm closing OpenSCAD via the menu -> quit or `Ctrl+Q` (see screenshot). So, my suggestion is simple: remove it.
openscad_linux_crash

If anyone thinks this is not the right solution, I'd be happy to hear their suggestion!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,7 @@ if(POLICY CMP0072)
   cmake_policy(SET CMP0072 NEW)
 endif()
 
-set(CMAKE_OSX_DEPLOYMENT_TARGET "10.9" CACHE STRING "Minimum OS X deployment version")
+set(CMAKE_OSX_DEPLOYMENT_TARGET "10.13" CACHE STRING "Minimum OS X deployment version")
 
 project(openscad)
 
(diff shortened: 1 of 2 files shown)


gingerbeardman:
Looks good to me.

TheJKM:
Closing this one as there is a better solution mentioned in #3837.
Also, because this is not related to macOS, the new merge request is for master rather than for the qt5-upgrade branch.

PR #3789, opened 2021-05-26
Fix styling of tab bar on macOS 11

TheJKM:
This fixes the ugly-looking tab bar on macOS 11 (Big Sur), mentioned in #3775.

The fix is an improved modified backport of the official fix for [QTBUG-86513](https://bugreports.qt.io/browse/QTBUG-86513). It is only required for the time OpenSCAD stays on Qt 5, as the bug is fixed in Qt 6.

Before:
tabbar_before

After:
tabbar_after


t-paul (approved):
Nice!

PR #3790, opened 2021-05-28
Clean up comma parsing

### LINKED ISSUE #2882: Allow trailing comma in let() or lists

thehans:
This change is based on PR #2928, with the parsing of commas changed to be consistent in all cases, and updated to be current with master.

A single optional trailing comma is now allowed in all cases where commas are applicable: parameters, arguments, and vector contents.
Leading commas or extra "middle" commas are considered invalid syntax.  In order to be valid, a comma must be preceded by a valid element.  e.g. `v=[,]` and `v=[1,,3]` are not valid under this PR.

Resolves #2882


thehans:
Obsoleted by #3814

PR #3794, opened 2021-05-30
Implement variable clipper scale for wider range of effective precision.

thehans:
The Clipper library used for 2D operations represents vertex coordinates internally with `signed long long` integers (should be 64bit on most if not all supported architectures).

To convert double precision values into integers we have been scaling those doubles by a constant `CLIPPER_SCALE = 1 << 16;` before coercing into integral types.  This can leave a lot of imprecision on the lower end of the scale.

This PR replaces the static with an adaptive scale, which tries to maximize the usable bits of integer precision, depending on the scale of the geometry involved.

Here is a short example I used to show the differences between this nightly and this PR:
```
scalepow2 = -10;

$fn = 100;
d = 2^scalepow2;
linear_extrude(d/10) offset(r=d/10) difference() {
  circle(d=d);
  translate([d/2,0,0]) circle(d=d);
}
```

nightly_test

PR_test

With these changes, `scalepow2` can be set down to `-16` before some noticeable geometry rounding starts to happen.  It should be capable of much more precision than that, but I think the PolySet Grid snapping starts to become the limiting factor rather than clipper's precision.

### DIFF
--- a/src/GeometryEvaluator.cc
+++ b/src/GeometryEvaluator.cc
@@ -1173,6 +1173,89 @@ Response GeometryEvaluator::visit(State & /*state*/, const AbstractPolyNode & /*
 	return Response::AbortTraversal;
 }
 
+shared_ptr<const Geometry> GeometryEvaluator::projectionCut(const ProjectionNode &node)
+{
+	shared_ptr<const class Geometry> geom;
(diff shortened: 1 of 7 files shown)


t-paul (approved):
Nice!

PR #3806, opened 2021-06-10
Port 3MF vertex ordering improvements to STL

tjhowse:
I don't have the build pipeline set up locally, so I'm unable to test this for now. I'm raising this draft PR for comment.

### DIFF
--- a/src/export_stl.cc
+++ b/src/export_stl.cc
@@ -67,76 +67,88 @@ void write_vector(std::ostream &output, const Vector3f& v) {
     }
 }
 
-	
 size_t append_stl(const PolySet &ps, std::ostream &output, bool binary)
 {
     size_t triangle_count = 0;
     PolySet triangulated(3);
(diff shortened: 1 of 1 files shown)


t-paul:
Thanks for taking on this topic. Starting yesterday there seems to be some server issue for some of the CI runs, those are not able to install some of the custom packages.

The CircleCI builds should not be affected by that. That build gives some type conversion error:
```
/root/workspace/src/export_stl.cc:83:27: error: conversion from 'float' to non-scalar type 'Eigen::Vector3f' {aka 'Eigen::Matrix<float, 3, 1>'} requested
   83 |             Vector3f p0 = (float)p[0];
```

tjhowse:
One build did get to the point of throwing compiler errors, so I should be able to look at that log when I get back to a computer. I'm pretty sure that one is a case of casting an int[3] to Vector3d or something. Shouldn't be too hard to fix.

PR #3812, opened 2021-06-12
Move testdata

thehans:
Moved `testdata` path into `tests/data`.


t-paul (approved):
Looks good! Nothing unusual in both Travis and github test run logs.

PR #3813, opened 2021-06-13
Call CGAL::set_error_behaviour(CGAL::THROW_EXCEPTION) only once from main.

thehans:
edit: I had originally tried removing `#pragma push_macro("NDEBUG")` etc, in addition to the error behavior change, but had to revert those changes.
It appeared to work fine on my local machine at the time that I created the PR, but I guess I just hadn't properly cleaned the build.


t-paul (approved):
Looking at the comment in `openscad.cc` all that dance seems to be originally setup for debugging. There are probably different ways than optimizing for that use case at the expense of all that extra code.

PR #3819, opened 2021-06-19
Update pl.po

PLSomebody:
Original translation was incomplete and consists loan translation. I was trying to translate all of missing strings and fix bugs and wrong translation.


t-paul (approved):
Thanks!

PR #3845, opened 2021-07-25
Update macOS deployment target for QT 5.15.2 upgrade

TheJKM:
This updates the previously forgotten macOS deployment target value.

Re-issuing this separately because I closed #3784. For details, see #3775.


TheJKM:
Looks live the CI fails because CGAL 5.3 fixes are not in `qt5-upgrade-macos`

t-paul:
I'll pull in the fix CGAL 5.3 fix from master...

PR #3848, opened 2021-07-30
Add support for defs in svg files.

### LINKED ISSUE #3847: Import svg file with <defs> usage
I am using "Silhouette studio business edition" to create svg files which I would like to import in openscad.
This software puts all the paths in the `<defs>` tag in the svg file and then uses `<use xlink:href="....">` to reference to the paths in the `<defs>` tag.

Openscad has code to ignore the `<defs>` tag and has no support for `<use>` tags in the svg.

I would like to request support for `<defs>` and `<use xlink:href="....">` in svg files.

Some extra information:
In an svg file the `<defs>` tag can be used to store elements for later use. Without a `<use>` tag they should not show [(documentation)](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/defs) 
A use tag can reference to a thing in the `<defs>` section by it's id. There are 2 ways, the deprecated way by using `xlink:href="...."` and the non-deprecated way by using `href="...."`. Silhouette studio uses the deprecated way, which I would like to use. [(documentation)](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/use)

Silhouette studio is free, but in order to save SVG files you need to buy "business edition" which costs $50-$80 depending on the store. This is an example SVG file that silhouette studio business edition creates:

```
<?xml version="1.0" encoding="utf-8"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<!-- exported from Silhouette Studio® Business Edition V4.4.285ssbe -->
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="304.8mm" height="304.8mm" viewBox="0 0 304.8 304.8">
<defs>
<path id="path_9243898cc3d88891b5c5be22c468ec50" d="M 36.8663 33.67314 L 36.8663 72.57143 L 83.31201 72.57143 L 83.31201 33.67314 L 36.8663 33.67314 Z" fill-rule="evenodd" fill="none" fill-opacity="1" stroke="#FF0000" stroke-opacity="1" stroke-width="0.1" stroke-linecap="round" stroke-linejoin="round"/>
<path id="path_be886460d7bffa85c0d1cb7d783b634f" d="M 122.2103 38.89828 L 122.2103 76.63543 L 172.72 76.63543 L 172.72 38.89828 L 122.2103 38.89828 Z" fill-rule="evenodd" fill="none" fill-opacity="1" stroke="#FF0000" stroke-opacity="1" stroke-width="0.1" stroke-linecap="round" stroke-linejoin="round"/>
</defs>
<g>
<use xlink:href="#path_9243898cc3d88891b5c5be22c468ec50"/>
</g>
<g>
<use xlink:href="#path_be886460d7bffa85c0d1cb7d783b634f"/>
</g>
</svg>
```

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/100149788-import-svg-file-with-defs-usage?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

vedelaar:
Fixes #3847

The idea is that we create a `map<string, shared_ptr<shape>>` (defs_lookup_list) and fill it with the items found in the `<defs>` tag. The ID of the thing being the string in the map. Then when we encounter a `<use>` tag we check the map for the ID it references to, we clone the thing in the map (including the children of the thing) and make it the child of the use tag. 

X and Y information can be added to the `<use>` tag and that is stored as a "translate" in the transform string in the `<use>` shape. That way when transformations are applied the children of the use tag have the proper location.


t-paul:
Nice! I'm wondering about the test case image update. Looking at the SVG source, I think it should place the filled ellipsis not only in the initial position but also overlay in all the 4 arc examples.

I did some testing with
```
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<svg xmlns:svg="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="20mm" height="2mm" viewBox="0 -1 20 2">
  <defs>
    <circle id="circle" cx="5" cy="0" r="1" />
    <rect id="rect" x="5" y="1" width="1" height="1" rx="0.1" />
  </defs>

  <circle cx="0" cy="0" r="1" />
  <rect id="rect" x="0" y="1" width="1" height="1" rx="0.1" />

  <use xlink:href="#circle" />
  <use xlink:href="#rect" />

  <use x="10" xlink:href="#circle" />
  <use x="10" xlink:href="#rect" />

  <g transform="translate(15 0)">
    <use xlink:href="#circle" />
    <use xlink:href="#rect" />
  </g>
</svg>                                                                                                                                                                                                             
```
And that looks good.

vedelaar:
I looked into the test case image. There was a problem in my code that prevented children 2 deep or more in a `<defs>` to have the correct parent and because of that were missing on some translate calls. I fixed that.

After fixing that the test case has 5 times the same image and the original test gets lost. I changed the svg file so that the original test is kept, and also added an extra test case to test my change better.

I rebased and force-pushed the commit to keep history clean, if you are interested in what I changed after your comment: https://github.com/openscad/openscad/compare/40e8fec..b844d7c

t-paul (approved):
Thanks!

PR #3853, opened 2021-08-11
Add Offline User Manual.

abshk-jr:
This Pull Request add the functionality to download the OpenSCAD User Manual from [OpenSCAD Documentation](http://files.openscad.org/documentation/).

A new CMakeLists.txt has been created inside the resources folder which checks for an active internet connection, and runs the python script to download the documentation.

The downloaded documentation can be accessed from the help menu, which opens the Manual in the default browser present on the system.

offline documentation

By default, the option for downloading the User Manual has been turned to `OFF` in CMakeLists.txt.
If the user doesn't download the documentation, the option in the help menu is hidden away.

offline docs off

https://github.com/openscad/openscad/pull/3853/commits/949c456586658a07794db61b73f5762d5977552d adds the functionality to access an Offline version of the Cheat Sheet.

offline cs


src/MainWindow.cc
```diff
@@ -459,6 +459,18 @@ MainWindow::MainWindow(const QStringList &filenames)
 	connect(this->helpActionCheatSheet, SIGNAL(triggered()), this, SLOT(helpCheatSheet()));
 	connect(this->helpActionLibraryInfo, SIGNAL(triggered()), this, SLOT(helpLibrary()));
 	connect(this->helpActionFontInfo, SIGNAL(triggered()), this, SLOT(helpFontInfo()));
+		
+	// Checks if the Documentation has been downloaded and hides the Action otherwise
+	fs::path resPath = PlatformUtils::resourcePath("resources");
+    fs::path fullPath = resPath/"docs"/"OpenSCADUserDocs"/"openscad_docs"/"OpenSCAD_User_Manual.html";
+	if (fs::exists(fullPath) && fs::is_regular_file(fullPath))
```
t-paul, line 466: This looks fine on it's own but as we have an almost identical part in UIUtils, there's the opportunity to move everything to this class. So it could have an internal function generating the path and provide 2 methods to the outside.
- hasOfflineUserManual() - for use here in the `if`
- openOfflineUserManualURL() - this will remain as it is now

t-paul (changes requested):
Looks good, I'll give it a go on my notebook. Just one small suggestion as there's the possibility to centralize the generation of the path to the documentation.

abshk-jr:
I have changed the function `openOfflineUserManualURL()` and replaced it with `hasOfflineUserManual()`.
The new function now returns a boolean for if the User Manual exists or not, and also accepts a boolean argument `openManual` which opens the User Manual if set to true.

src/UIUtils.cc
```diff
@@ -176,6 +176,22 @@ void UIUtils::openUserManualURL()
     openVersionedURL("https://www.openscad.org/documentation.html?version=%1");
 }
 
+bool UIUtils::hasOfflineUserManual(bool openManual = false)
```
rcolyer, line 179: Can you provide a rationale for these bool returning functions labeled "has" being reused as the routine which opens the manual and cheatsheet?  Generally a function should do only what it says on the label, unless there is a compelling reason otherwise.

rcolyer (commented):
Solid work, and looks good.  Just the one in-line comment to address or fix.

t-paul (approved):

Issue #3852, opened 2021-08-05
Save depth map to png


TLC123:
Feature request:
I propose a function to export or display render buffers from the current viewport display, depth maps, normal map or other.
Why: liberate as much user created data as possible. Why not enable the display/export of all available image buffers.
How: On png export dialog select other buffer. Alternatively select viewport mode in "view" menu and export what ever is on.

jordanbrown0:
Duplicate of #2043.

PR #675, opened 2014-02-26
Speed up OpenGL 3d mesh drawing significantly by batching whole polyheda

OskarLinde:

PR #3855, opened 2021-08-12
Load shaders from files

thehans:
Allows for modifying / testing shaders without recompiling OpenSCAD itself.

This should help make future features such as #3852 easier to implement, and such modifications more accessible to end-users.


t-paul (approved):
Nice, later we could even extend that to allow user specified shaders.

Just two observations:
- There's no `loadShaderSource()` for the MouseSelector shaders?
- Maybe we could move the files to `resources`. It would be nice to reduce the number of subfolders that need to be given in all those scripts

thehans:
> Nice, later we could even extend that to allow user specified shaders.
> 
> Just two observations:
> 
> * There's no `loadShaderSource()` for the MouseSelector shaders?

Yeah I had edited that file but accidentally left it out of the commit.  Had some other changes mixed in locally, so i was selectively adding files.

> * Maybe we could move the files to `resources`. It would be nice to reduce the number of subfolders that need to be given in all those scripts

I was thinking about that, but as I understand, Qt resources get compiled into the executable?  I like the idea of having shaders potentially editable by end user, much like color-schemes directory.  So I more-or-less copied how color-schemes was handled for packaging etc.

TLC123:
Looking forward to play around with shaders again. Will we see this in next snapshot?

t-paul:
Yes, it's merged, but there seem to be a packaging issue. Note that this is not yet setup for easy user change of the files, you'll need to poke into the installation directories and find the `Preview.{vert,frag}` files.

thehans:
@TLC123 Also, the Preview shader is currently only active when previewing with "Show Edges" enabled.

I would like to eventually add a separate shader for preview *without* edges as well.  This PR was just sort of a minimal "laying the groundwork" for future shader expansion.

TLC123:
Guys! I literally editing the shaders in OpenSCAD, **LIVE** as i use it. The preview _with_ edges situation is perfekt in case of invalid shaders, everything default gracefully to _without_ 
Thank you. Grazie mille.
shaders

t-paul:
If we can collect a couple of interesting shaders, that would make it even more useful to move this to a better user accessible feature. This https://github.com/openscad/openscad/pull/675#issuecomment-46676728 shows some nice ones but I don't think @OskarLinde posted the shader sources.

TLC123:
I will struckure my experiments like this. 

In the shaders folder I create subfolders named by each experiment plus one called original that I keep writeprotected . In each folder I have the set of preview.frag/.vert . 

The switching of shader is a simple copy of one subfolders content to the shader folder.

PR #3860, opened 2021-08-17
Small collection of basic shaders for tests

TLC123:
Unpolished shaders for tests purpose.


t-paul:
Neat! 
```
#!/bin/bash

TMP=$(mktemp -t XXXXXXXXXX.png)

trap EXIT "rm -fv '$TMP'"

for F in $(find shaders -mindepth 2 -name Preview.frag)
do
	N=${F%.frag}
	V="${N}".vert
	T=$(basename $(dirname "$N"))
	P="/tmp/shader-${T}.png"
	cp -vf "$F" "$V" shaders/
	./b/openscad -o "$TMP" --imgsize 1024,1024 --view edges examples/Basics/CSG.scad
	convert "$TMP" -background gray80 -font Open-Sans -gravity South -pointsize 22 -splice 0x64 -resize 500x500 -annotate +0+2 "$T" -append "$P"
done
```
shader-checkerboard
shader-graphpaper
shader-RGBcheckerboard
shader-screennormal
shader-worldnormal

TLC123:
Yeah of course. How interesting. The  subtractions obviously, must , have  inverted normals.
 + depth-map and its cousin screenPos in today's commit.
 
 One can go crazy with procedural patterns but o find it way out of scope here.
 I think these class of shaders is about what you can do now with current uniforms,
   without major work on multiple render pass buffers and .
  
 I have few more concepts to explore.

t-paul:
Looks a bit scary :smiley: 
shader-metallic

I think the path will probably be to start slowly with adding more uniforms as parameters to the shader(s). But if you have specific suggestions, please list those. What's the benefit of multi pass shaders here? I know them from RetroArch where they can produce amazing effects, I'm not yet sure how that applies to the type of things we want to display.

TLC123:
> Looks a bit scary 😃
> shader-metallic
> 
> I think the path will probably be to start slowly with adding more uniforms as parameters to the shader(s). But if you have specific suggestions, please list those. What's the benefit of multi pass shaders here? I know them from RetroArch where they can produce amazing effects, I'm not yet sure how that applies to the type of things we want to display.

I see your point. 
Purpose and opportunity.
In line with the main purpose to increase OpenSACDs usefulness there are a few visual modes  of interest.
Analysis like normals and so on. Visual improvements that better conveys the shape of the model.
Secondary there is an opportunity to  give OpenSCAD a "sexier" look for new comers.

For the  Visual improvements i have previous mentioned _ambient occlusion_ as a mean to convey depth and proximity of nearby objects. As little i know i am sure that even the simplest form of screen space AO  can't be done on a single render passes
  
Then there is an unintended residual end user opportunity to hack around with shaders.
Shaders are such a low risk joy to be creative with.

The list. I'll have a think.

PR #3866, opened 2021-08-23
Replace deprecated .delta() with angleDelta().y()

themactep:


t-paul:
Something strange is going on here. Looking at the [documentation](https://doc.qt.io/qt-5/qwheelevent.html#angleDelta), I think the replacement would be `.y()`. So I suspect this whole code is actually not needed and the default handler already takes care of the zooming.

themactep:
Paul, I think you're right, it should be `.y()` since mouse wheel typically provides vertical scrolling. I amended the patch.
`delta()` is deprecated and throws warnings during compilation.

t-paul (approved):
Thanks, lets get rid of the deprecation. We can come back later to check if the whole event handler could be dropped.

PR #3869, opened 2021-08-31
dxf export as polylines

IanSC:
dxf export converted to polyline, so each island comes out as single entity when imported.
Most users would probably need it to be as such, after all what's the use of 100's of separate line 
segments for a circular path.
Removed extra dxf sections for inkscape compatibility as it is not needed anymore.

Tested in inkscape and https://sharecad.org/# only.
If this is acceptable will do more test on other apps.

### DIFF
--- a/src/export_dxf.cc
+++ b/src/export_dxf.cc
@@ -30,347 +30,224 @@
 #include "dxfdata.h"
 
 /*!
-	Saves the current Polygon2d as DXF to the given absolute filename.
+    Saves the current Polygon2d as DXF to the given absolute filename.
  */
 
 void export_dxf_header(std::ostream &output,double xMin,double yMin,double xMax,double yMax) {
-	
-	// https://dxfwrite.readthedocs.io/en/latest/headervars.html
-	// http://paulbourke.net/dataformats/dxf/min3d.html
+    
+    // https://dxfwrite.readthedocs.io/en/latest/headervars.html
+    // http://paulbourke.net/dataformats/dxf/min3d.html
 
-	// based on: https://github.com/mozman/ezdxf/tree/master/examples_dxf
-	// Minimal_DXF_AC1009.dxf - not working in Adobe Illustrator
-	// Minimal_DXF_AC1006.dxf - not working with LibreCAD (due to 3DFACE?)
+    // based on: https://github.com/mozman/ezdxf/tree/master/examples_dxf
+    // Minimal_DXF_AC1009.dxf - not working in Adobe Illustrator
+    // Minimal_DXF_AC1006.dxf - not working with LibreCAD (due to 3DFACE?)
 
-	// tested to work on:
-	// - InkScape 1.0.0
-	// - LibreCAD
-	// - Adobe Illustrator
-	// - https://sharecad.org
-	// - generic cutters
+    // tested to work on:
+    // - InkScape 1.0.0
+    // - LibreCAD
+    // - Adobe Illustrator
+    // - https://sharecad.org
+    // - generic cutters
 
-	output
-		<< "999\n" << "DXF from OpenSCAD\n";
+    output
+        << "999\n" << "DXF from OpenSCAD\n";
 
-	//
-	// SECTION 1
-	//
+    //
+    // SECTION 1
+    //
 
-	/* --- START ---
-	0
-	SECTION
-	2
-	HEADER
-	9
-	$ACADVER
-	1
-	AC1006
-	9
-	$INSBASE
-	10
-	0.0
-	20
-	0.0
-	30
-	0.0
-	*/
+    /* --- START --- */
 
-	output
-		<< "  0\n" << "SECTION\n"
-		<< "  2\n" << "HEADER\n"
-		<< "  9\n" << "$ACADVER\n"
-		<< "  1\n" << "AC1006\n"
-		<< "  9\n" << "$INSBASE\n" // for 3D only?
-		<< " 10\n" << "0.0\n"
-		<< " 20\n" << "0.0\n"
-		<< " 30\n" << "0.0\n"
-		;
+    output
+        << "  0\n" << "SECTION\n"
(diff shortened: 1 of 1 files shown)


src/export_dxf.cc
```diff
@@ -35,61 +35,84 @@
...
-				 << "ENTITIES\n";
+	output  << "  0\n" << "SECTION\n"
+			<< "  2\n" << "BLOCKS\n"
+			<< "  0\n" << "ENDSEC\n";
+
+	output  << "  0\n" << "SECTION\n"
+			<< "  2\n" << "ENTITIES\n";
+
+	// Some importers (e.g. Inkscape v1.1) needs a layer to be specified, but just once is enough
+	output  << "  8\n" << "0\n"; // layer 0
```
t-paul, line 48: This change seems to break LibreCAD import.

t-paul (changes requested):
Adding support for PolyLine would be a welcome change.

I'm a bit concerned regarding compatibility though, especially with mentions of Inkscape 1.1 which is *really* new and we want to support at least Ubuntu 18.04 going forward. Unless adding an option that allows using to the old logic, we probably want to minimize the changes.

src/export_dxf.cc
IanSC: Updated the code based on LibreCAD's output.
Can you send the link to the test files breaking LibreCAD in case it still breaks?

I understand the need to add extra stuff to be compatible with other softwares even though they are not necessary to be conformant. But I just wanted to keep it minimal to avoid incompatibilities with other softwares. The old version does not work with some versions of Illustrator.

Since you are open to this change, I will go test with other softwares.

IanSC: It seems DXF format need more info that what is currently generated.
I checked various dxf files/libraries and applied those for 2D.

This one is working on app I have access to.
It outputs complete dxf header for 2D.

btw, so of the comments in the code is from the original, such as what Inkscape needs, etc.

t-paul:
Looks great!

I've tested with https://github.com/openscad/openscad/pull/3869/commits/b3b875ecd7c9823656ae3f6e097b6975ef55b5a5 via the AppImage using this file: [christmas-tree.dxf.txt](https://github.com/openscad/openscad/files/7110119/christmas-tree.dxf.txt)

Ubuntu 18.04 (pretty much the oldest platform supported going forward)
- LibreCAD 2.1.2 ✔
- Inkscape 0.92.3 ✔
- KiCAD 4.0.7 ✔
- FreeCAD 0.16.6712 (DXF Import 1.39) ✔
- Blender 2.79b (Import DXF v.0.8.6) ✔

Debian Testing
- LibreCAD 2.1.3 ✔
- Inkscape 1.0.2 ✔
- KiCAD 5.1.10 ✔
- FreeCAD 0.19.1 ✔
- Blender 2.83.5 ✔

t-paul (approved):
Do you have anything open from your side? If not, I'm good with merging.

IanSC:
My last commit is a lot cleaner. The view in github messes all the tabs. I specifically aligned everything so it's easier to maintain later. I converted all tabs to space. So it's red/green all over.
This is my very first contribution. New OS. New IDE. New language (after a long long long time).
Is there some list what tools you guys use? I had a hard time starting because I don't know what to editor/IDE to use coming from Windows Visual Studio. Finally settled with VSCode. It's a good thing OpenSCAD has a build instruction right there in the readme to get the ball rolling.

t-paul:
Github diff has a flag to ignore whitespace changes. I'm not too worried about that, as there's preparation to use `uncrustify` for formatting. It's not applied yet as there's somehow never the "right" time, but it will happen eventually.
I'm using NetBeans on Linux for development. This is maybe not perfect, but works ok. Still hoping for the C++ support to move forward as it's currently in holding pattern, but it does handle at least C++14 pretty well.
I suppose VSCode is fine, from what I've seen their C++ support got quite some love, but I have not used it myself for that. Other options (I've not used) would be QtCreator and CLion.

PRouzeau:
Unfortunately, this break DXF import in NanoCAD 5, which is the free version of NanoCAD (2D CAD). This is critical for me, so I've gone back to a former snapshot.

t-paul:
Any error messages that would indicate technical details for the problem?

PRouzeau:
Unfortunately, not. I had to inspect the DXF file and to compare to a working one to discover the format output was modified.
NanoCAD 5 is old (2013) and no longer evolve but as it is Free, it is still well used on Windows.

t-paul:
I would assume it does support and export polylines, those are at least in the 2012 spec and probably even older. Can you create a simple drawing with just a multi-line polygon? Maybe there's some simple change in the file that would help.

PRouzeau:
Nanocad in export commands description tells that it export AutoCAD DXF 2013 format, no mention of DXF version in imports. I checked that it does effectively export polylines and that it can Import the polyline file it created. In the Nanocad created DXF file, the polyline command does look like the one created by new OpenSCAD snapshots, though the file created had a lot more commands in it (file size is 34kb for a single three lines polyline). 

Tested import with a simple square created on OpenSCAD (only one polyline) (with Openscad October Windows snapshot),  no success (invisible)
Removed all DXF headers, left only the polyline, still invisible
Added simple lines to the existing polyline (still with all headers removed), these are visible and import correctly, so the polyline don't stop the import, it is only not visible.
Removing the LWPOLYLINE command does stop the import (while its empty in OpenSCAD output)
Added the Polyline command as created by Nanocad (only AcDbPolyline with NO other command) is not visible.

So the problem seems to be in other commands associated with the polyline command, which are difficult for me to test as I don't know the DXF syntax.

t-paul:
The spec is quite big and a bit confusing, but here's the link for the 2012 version: https://images.autodesk.com/adsk/files/autocad_2012_pdf_dxf-reference_enu.pdf - maybe create a ticket and attach the 2 sample files, e.g. just a square with both Nanocad and OpenSCAD. Someone might find a bit of time to look at the issue.

PR #3897, opened 2021-09-23
Eye candy for About dialog.

themactep:
My attempt to make About dialog more appealing.

I used donation button from https://opencollective.com/openscad/donate/button.png?color=blue with slight modifications to reduce file size, so you might want to check thier clause 13 at https://opencollective.com/tos regarding copyright and licensing, if you decide to merge this.

Other changes include simplifying html structure, removing dependency on particular fonts, removing duplicate contributor's name entry, updating website URLs if redirects or switched to HTTPS, proper GitHub name spelling, etc.

If you have any questions, I'll gladly explain every change made.

### DIFF
--- a/src/AboutDialog.html
+++ b/src/AboutDialog.html
@@ -251,6 +251,8 @@ lABPVqVoVsd6gA2YoAVC+QMLSLCBWoV1pWONaxTeygm52rWqnAgCADs=" alt="Donate"></a></p>
     License: CC0 1.0 Universal (CC0 1.0) Public Domain Dedication.<br>
     Author: Gabriel Araujo.</p>
 
+<p>Donation button design is copyright of <a href="https://opencollective.com/">Open Collective, Inc.</a></p>
+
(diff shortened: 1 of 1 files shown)


t-paul:
Very nice!

As for the button image, the TOS does not sound to promising unfortunately and the [Button & Banners](https://docs.opencollective.com/help/collectives/widgets) page does not even mention any licence conditions. But we'll see, I've sent an email asking if there's any more specific note about usage.

t-paul:
Great news :tada:

Initial response said
> I checked and our buttons don't have a clear license today. They should and we're happy to add one.

and mentioned a license that would be good to go. So we just need to hang in there for a bit and maybe need a tiny update if attribution will be required but it's unlikely the button would have to be removed.

t-paul:
Ok, update is live now at https://docs.opencollective.com/help/collectives/widgets - `CC-BY-SA 4.0`, so I think we need to add the attribution and then we are good to go.

themactep:
Added attribution referencing the copyright holder.

t-paul (approved):
Thanks!

PR #3900, opened 2021-09-25
Fix alt-uparrow tab crash

### LINKED ISSUE #3899: Editor crash after alt-uparrow and new tab
This was reported as a bug on the IRC channel.  In the 2021.01 release or the master branch, launch openscad with a blank file, type "x = 5;", click between the "5" and the ";", press alt-uparrow to change 5 to 6, press the "New file" button to open a new tab, and then type a character such as "a".  The program crashes.

If openscad was launched from console, the following additional output is available (reporting line 427 in the master branch, line 263 in the 2021.01 release):
```
Assertion [undoSequenceDepth > 0] failed at ../src/CellBuffer.cpp 427

Aborted
```

src/CellBuffer.cpp is part of QScintilla, in which I did find said assert, in a function "EndUndoAction":  https://github.com/opencor/qscintilla/blob/master/scintilla/src/CellBuffer.cpp#L427

Note that this still crashes if the tab is already open while the alt-uparrow press is performed, and one clicks over to the other tab.  But it does NOT crash if any other character is entered between alt-uparrow and the tab switch.

Still unresolved is why this is happening.

rcolyer:
This fixes #3899.  (Also some tab/space issues were fixed in the area of work.)

### DIFF
--- a/src/ScintillaEditor.cc
+++ b/src/ScintillaEditor.cc
@@ -986,7 +986,6 @@ bool ScintillaEditor::handleKeyEventBlockCopy(QKeyEvent *keyEvent)
 
 bool ScintillaEditor::handleKeyEventNavigateNumber(QKeyEvent *keyEvent)
 {
-    static bool wasChanged = false;
     static bool previewAfterUndo = false;
 
 #ifdef Q_OS_MAC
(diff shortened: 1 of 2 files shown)


t-paul (approved):
Thanks, lets get the crash fixed, we can revisit the undo handling later.

Issue #3628, opened 2021-01-22
2021.01-RC6 - OpenCSG S/W rendering significantly slower


MichaelAtOz:
I use S/W rendering, accessing a W7 VM (on real W7 box via VMWare), from Debian via FreeRDP.
As without S/W rendering it only supports OpenGL V1 and preview is not good.

For simple models it works well, for complex models and higher resolution/preview-panel-size it is slow.

In 2019.05 in a 'normal' preview size a complex model took many seconds to render, as in paint, the preview, ie change front view to top view, many seconds.
In 2021.01-RC6 it is unusable, it is as if it is continuing to repaint but not finishing, then repainting again, the GUI is very unresponsive, it gets the hourglass & 'Not Responding' repeatedly, typing a short-cut, eg Alt-F may sometimes work after many seconds, but often not.

To get a measure, I had to arrange a small preview size.
![Capture-tiny-review-2019-blank](https://user-images.githubusercontent.com/1972961/105436071-a4938f00-5cb2-11eb-804d-602160b22279.JPG)

Note that this is not F5-preview time, that is 
```Compile and preview finished.
Total rendering time: 0 hours, 0 minutes, 0 seconds
```
The design as .csg is 
[For Dyson Cinetic - v7-03-dodgy-2021.zip](https://github.com/openscad/openscad/files/5854259/For.Dyson.Cinetic.-.v7-03-dodgy-2021.zip)

With 2019.05 after F5, it paints, clicking View-all or changing to other views takes ~4s to repaint.

With 2021.01-RC6 (`Total rendering time: 0:00:00.939`) after F5, it paints, View-all is slow

Here is a recording.
[clip0041.zip](https://github.com/openscad/openscad/files/5854275/clip0041.zip)
At 0:29 (18:29:09 on the clock at the top) you can see 2019.05 View-all takes ~4s.
With 2021.01-RC6, clicking View-all is at 1:12 (18:29:52), ... 4:59 (18:33:37) repaint finishes.
That is 10 frames/s video time & clock time seem out a bit.
So based on clock time, 4s v's 255s, ~64 times longer.
Time is proportional to preview window size, at a reasonable size it is ~14s, v's I had dinner and it didn't finish.

MichaelAtOz:
Could someone please have a look at this prior to release cut-off?


I think it is also repainting to often, such as associated with how it repaints when moving/resizing windows, it could benefit with deferred/skipped repaints if repaint time is a significant portion of a UI event-loop.

[...]

PR #3904, opened 2021-09-26
Default convexity to 10.

jordanbrown0:
Also, in linear extrude, allow "h" as an alias for "height".

This is pretty straightforward, but I'd like to discuss a little.  In particular, I'm wondering about adding a $convexity so that those who are nostalgic for rendering artifacts can easily globally set $convexity=1.

Questions:

Is there a better place for the definitions of DEFAULT_CONVEXITY and QUOTE/QUOTED?  (Note that QUOTE/QUOTED are defined in several other places; it would be nice to commonize them.)

If adding $convexity, and even if not, it'd be nice to commonize the "get convexity parameter and default it to DEFAULT_CONVEXITY" pattern.  Is there a good place to put such a helper function?

The sample I happened to pick to standardize on uses `static_cast<int>`, but as far as I can tell a simple `(int)` would work too.  Is there a reason to prefer one over the other?

### DIFF
--- a/src/cgaladv.cc
+++ b/src/cgaladv.cc
@@ -42,6 +42,7 @@ static AbstractNode* builtin_minkowski(const ModuleInstantiation *inst, Argument
 	
 	Parameters parameters = Parameters::parse(std::move(arguments), inst->location(), {"convexity"});
 	node->convexity = static_cast<int>(parameters["convexity"].toDouble());
+    if (node->convexity <= 0) node->convexity = DEFAULT_CONVEXITY;
 
 	return children.instantiate(node);
 }
@@ -62,6 +63,7 @@ static AbstractNode* builtin_resize(const ModuleInstantiation *inst, Arguments a
 
 	Parameters parameters = Parameters::parse(std::move(arguments), inst->location(), {"newsize", "auto", "convexity"});
 	node->convexity = static_cast<int>(parameters["convexity"].toDouble());
+    if (node->convexity <= 0) node->convexity = DEFAULT_CONVEXITY;
 	node->newsize << 0,0,0;
 	if ( parameters["newsize"].type() == Value::Type::VECTOR ) {
 		const auto &vs = parameters["newsize"].toVector();
--- a/src/import.cc
+++ b/src/import.cc
@@ -112,9 +112,9 @@ static AbstractNode* do_import(const ModuleInstantiation *inst, Arguments argume
 			node->layername = "";
 		}
 	}
-	node->convexity = (int)parameters["convexity"].toDouble();
 
-	if (node->convexity <= 0) node->convexity = 1;
+	node->convexity = static_cast<int>(parameters["convexity"].toDouble());
+    if (node->convexity <= 0) node->convexity = DEFAULT_CONVEXITY;
 
 	const auto &origin = parameters["origin"];
 	node->origin_x = node->origin_y = 0;
--- a/src/linearextrude.cc
+++ b/src/linearextrude.cc
@@ -55,7 +55,7 @@ Parameters parse_parameters(Arguments arguments, const Location& location)
 	{
 		Parameters normal_parse = Parameters::parse(arguments.clone(), location,
 			{"file", "layer", "height", "origin", "scale", "center", "twist", "slices", "segments"},
-			{"convexity"}
+			{"convexity", "h"}
 		);
 		if (normal_parse["height"].isDefined()) {
 			return normal_parse;
@@ -69,7 +69,7 @@ Parameters parse_parameters(Arguments arguments, const Location& location)
 	// then assume it should be the height.
 	return Parameters::parse(std::move(arguments), location,
 		{"height", "file", "layer", "origin", "scale", "center", "twist", "slices", "segments"},
-		{"convexity"}
+		{"convexity", "h"}
 	);
 }
 
@@ -92,13 +92,14 @@ static AbstractNode* builtin_linear_extrude(const ModuleInstantiation *inst, Arg
 
 	if (parameters["height"].isDefined()) {
 		parameters["height"].getFiniteDouble(node->height);
-	}
+	} else if (parameters["h"].isDefined()) {
+		parameters["h"].getFiniteDouble(node->height);
+    }
 
 	node->layername = parameters["layer"].isUndefined() ? "" : parameters["layer"].toString();
 
-	double tmp_convexity = 0.0;
-	parameters["convexity"].getFiniteDouble(tmp_convexity);
-	node->convexity = static_cast<int>(tmp_convexity);
+	node->convexity = static_cast<int>(parameters["convexity"].toDouble());
+    if (node->convexity <= 0) node->convexity = DEFAULT_CONVEXITY;
 
 	bool originOk = parameters["origin"].getVec2(node->origin_x, node->origin_y);
 	originOk &= std::isfinite(node->origin_x) && std::isfinite(node->origin_y);
@@ -117,9 +118,6 @@ static AbstractNode* builtin_linear_extrude(const ModuleInstantiation *inst, Arg
 		node->center = parameters["center"].toBool();
 
 	if (node->height <= 0) node->height = 0;
-
-	if (node->convexity <= 0)
-		node->convexity = 1;
 
 	if (node->scale_x < 0) node->scale_x = 0;
 	if (node->scale_y < 0) node->scale_y = 0;
@@ -190,9 +188,7 @@ std::string LinearExtrudeNode::toString() const
 	if (!(this->has_slices && this->has_segments)) {
 		stream << ", $fn = " << this->fn << ", $fa = " << this->fa << ", $fs = " << this->fs;
 	}
-	if (this->convexity > 1) {
-		stream << ", convexity = " << this->convexity;
-	}
+    stream << ", convexity = " << this->convexity;
 	stream << ")";
 	return stream.str();
 }
@@ -202,7 +198,9 @@ void register_builtin_dxf_linear_extrude()
 	Builtins::init("dxf_linear_extrude", new BuiltinModule(builtin_linear_extrude));
 
 	Builtins::init("linear_extrude", new BuiltinModule(builtin_linear_extrude),
-				{
-					"linear_extrude(height = 100, center = false, convexity = 1, twist = 0, scale = 1.0, [slices, segments, $fn, $fs, $fa])",
-				});
+        {
+            "linear_extrude(height|h = 100, center = false, convexity = "
+            QUOTED(DEFAULT_CONVEXITY)
+            ", twist = 0, scale = 1.0, [slices, segments, $fn, $fs, $fa])",
+        });
 }
--- a/src/node.h
+++ b/src/node.h
@@ -138,3 +138,7 @@ public:
 
 std::ostream &operator<<(std::ostream &stream, const AbstractNode &node);
 AbstractNode *find_root_tag(AbstractNode *node, const Location **nextLocation = nullptr);
+
+#define DEFAULT_CONVEXITY 10
+#define QUOTE(x__) # x__
+#define QUOTED(x__) QUOTE(x__)
--- a/src/primitives.cc
+++ b/src/primitives.cc
@@ -516,7 +516,7 @@ public:
 	
 	std::vector<point3d> points;
 	std::vector<std::vector<size_t>> faces;
-	int convexity = 1;
+	int convexity;
 };
 
 std::string PolyhedronNode::toString() const
@@ -635,8 +635,8 @@ static AbstractNode* builtin_polyhedron(const ModuleInstantiation *inst, Argumen
 		faceIndex++;
 	}
 
-	node->convexity = (int)parameters["convexity"].toDouble();
-	if (node->convexity < 1) node->convexity = 1;
+	node->convexity = static_cast<int>(parameters["convexity"].toDouble());
+    if (node->convexity <= 0) node->convexity = DEFAULT_CONVEXITY;
 
 	return node;
 }
@@ -802,7 +802,7 @@ public:
 	
 	std::vector<point2d> points;
 	std::vector<std::vector<size_t>> paths;
-	int convexity = 1;
+	int convexity;
 };
 
 std::string PolygonNode::toString() const
@@ -931,8 +931,8 @@ static AbstractNode* builtin_polygon(const ModuleInstantiation *inst, Arguments 
 		return node;
 	}
 
-	node->convexity = (int)parameters["convexity"].toDouble();
-	if (node->convexity < 1) node->convexity = 1;
+	node->convexity = static_cast<int>(parameters["convexity"].toDouble());
+    if (node->convexity <= 0) node->convexity = DEFAULT_CONVEXITY;
 
 	return node;
 }
--- a/src/projection.cc
+++ b/src/projection.cc
@@ -42,6 +42,7 @@ static AbstractNode* builtin_projection(const ModuleInstantiation *inst, Argumen
 
 	Parameters parameters = Parameters::parse(std::move(arguments), inst->location(), {"cut"}, {"convexity"});
 	node->convexity = static_cast<int>(parameters["convexity"].toDouble());
+    if (node->convexity <= 0) node->convexity = DEFAULT_CONVEXITY;
 	if (parameters["cut"].type() == Value::Type::BOOL) {
 		node->cut_mode = parameters["cut"].toBool();
 	}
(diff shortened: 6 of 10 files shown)


jordanbrown0:
And I expect that the CSG-dumping tests will all fail and their reference output will need to be updated.

Also:  in their example calls, some modules say "convexity", others say "convexity=number", and others say "convexity=<the actual default>".  Divergence seems wrong.  Any preference between the three?

t-paul:
Unifying the various places for convexity is a nice step, but combining language (parameter) changes and the convexity change in one PR is not going to be mergable, that needs to be separated.

I dont see any reasoning for convexity = 10, it's likely not going to make sense for convex polygons (with single outline = no cutouts), so a fixed value is maybe not the ideal solution.

`$convexity` seems strange, dynamic variables travel down the call chain, convexity increases up the call chain, I believe. How would that work?

nophead:
Convexity seems something best worked out by a computer, not specified by a
human. What is the convexity of a gear with 39 teeth? I would have to draw
it and put a ruler across it and start counting.

t-paul:
Agreed, but I don't know any algorithm that will give the exact value in reasonable time. If there's one, that would be great. Otherwise we may need to work with some simpler solution that may not always be right, but hopefully much better than the current defaults.

nophead:
The algorithm must exist in OpenSCAD already because if I make a 3D shape
with arbitrary convexity by combining 3D primitives it will display
correctly but If I make it with 2D primitives I have to specify the
convexity of the 2D bit.

jordanbrown0:
> The algorithm must exist in OpenSCAD already ...

Well, maybe.  It seems to know how to do it, at least sort of, *if* it knows the convexity of the input objects.  If you operate on a cube and a cylinder, it knows they are both convexity 1.  But if you give it some arbitrary shape, whether from linear_extrude, import, or polyhedron, it needs to be told what that shape's convexity is.  I get the impression that the CSG stuff doesn't *really* do geometry calculations, that it does something more akin to hidden-line removal.  Its algorithms may not be clever enough to derive the convexity of an arbitrary shape.  (And certainly I don't know enough about the geometry processing to do anything useful there.)

> combining language (parameter) changes and the convexity change in one PR is not going to be mergable

Are you talking about the height/h change, or about the hypothetical $convexity change?  I certainly don't mind splitting out the height/h change; that was just an "as long as I'm here" thing for something that annoys me every time I use linear_extrude.  It does seem like the $convexity thing, if implemented, goes with the convexity default change.

> I dont see any reasoning for convexity = 10, it's likely not going to make sense for convex polygons (with single outline = no cutouts), so a fixed value is maybe not the ideal solution.

There is almost precisely no reasoning for the particular value 10, other than the fact it's enough for a lot of cases, without being large enough to cause performance problems.  It's not that there aren't objects with convexity greater than 10, it's that there are a lot fewer of them than there are with convexity 2.  Performance tests (nophead's, I think) indicate that large convexity values don't start hurting until they are *very* large, like hundreds or thousands, and that the performance impact of small numbers like 10 is negligible.  The convexity-related e-mail complaints would basically disappear; very few people would ever need to set it.

> $convexity seems strange, ...

$convexity would be the default convexity for any module that has a convexity parameter, replacing my current internal DEFAULT_CONVEXITY.  Note that union(), for instance, does not have a convexity parameter even though it can generate concave objects; it would not be affected by $convexity.  It would be like $fa/$fs/$fn, which exist largely (albeit not completely) so that you can express global preferences for performance vs some other metric.  I would expect it to be used **very** rarely; it would largely exist only so that we could say that there was an easy way to get back to the old behavior.  Somebody who routinely works with high-convexity objects might set it higher than 10 to avoid needing to set it on each linear_extrude/import/polyhedron.

Any answers for the internal design/layout questions I asked?

nophead:
Top level linear_extrudes, etc don't need convexity setting, so if this forces them all to be 10 that might be a performance hit. I rarely set convexity in linear_extrude because not many are what I call exposed differences / intersections. I render void those and render the ones I can't avoid.

Perhaps the 3D OpenCSG just adds the convexities together. Is it true that if you union or difference two objects with convexity 1 the result has a maximum convexity of 2. If that is all it does then 2D union, difference and intersection could do the same. That would be you only need to set it for polygon, polyhedron and import.

jordanbrown0:
I thought I remembered you doing a comprehensive performance test, but I don't find it.

I do find [this message](https://forum.openscad.org/2D-preview-convexity-issue-tp29548p29557.html):
> [...] The odd thing is setting it to 100 is no problem. People warn it will slow down but I have never noticed that. Even setting it to 10000 doesn't cause a noticeable slow down in an animation! [...]

`render()`, on the other hand, is just about guaranteed to be a performance hit.

I had wondered whether it had some algorithm like adding together convexities, but hadn't worked through enough cases to be sure what they really are.  (It seems, for instance, like intersection might yield the maximum of the two, rather than the sum.)

nophead:
Render is a one off performance hit and then it cached. On the other hand OpenCSG has to do its tricks to do differences and intersections on every redraw of the screen. When you have thousand of objects in a scene, like I do, I can't afford that time, so render is a win. Maybe less so VBO rendering. 

But there are other artifacts that render() avoids. For example negative objects from different parts can interfere with each other and cause Z-fighting and you also have the negative object clipping problem when you zoom in.

Mostly my objects are unions of extruded 2D parts and union is free as it just draws them in the correct order, I think.

Whatever the 3D rules used to combine convexities should be applied to 2D is my point. I presume they are buried somewhere in OpenCSG.

nophead:
Does a convexity of 10 on thousands of objects that don't need it have a noticeable effect? I.e. is it similar to 1000 on one object.

nophead:
For example CSG products shows me there are 1046 extrudes in this scene. All but a handful are +, which means they are unions that are simple drawn, so don't need convexity setting at all. Would they slow down if they were all 10?

image

There are 11 PCB / perfboards and four fans crammed in this box.

thehans:
> The algorithm must exist in OpenSCAD already because if I make a 3D shape with arbitrary convexity by combining 3D primitives it will display correctly but If I make it with 2D primitives I have to specify the convexity of the 2D bit.

It doesn't exist in OpenSCAD, and has nothing to do specifically with extrudes or 2D vs 3D..  All of our 3D primitives just happen to be convex.
If you "make a 3D shape with arbitrary convexity by combining 3D primitives" then you are just sending a bunch of convex CSG product terms to OpenCSG.  

To demonstrate, try opening `Examples -> Old -> example024.scad`
Preview that as is, and everything looks fine (because its composed entirely of cubes, with convexity 1)

Now insert a `render()` without setting convexity on line 13 and try previewing:
```
    render() for (v=[[0,0,0], [0,0,90], [0,90,0]])
```

This sends an object with high convexity to OpenCSG, and it absolutely doesn't know what to do with it.  No 2D or extrudes necessary.

nophead:
I see, so no matter how complex the 3D object is it is always made from convex primitives when it is rendered.

I wonder if there is a simple rule for combining 2D shapes all the same.  It might be pessimistic but if you union convex shapes I think it can be no more than the sum of their convexities. So perhaps the default value could be that and you only need to override it. And perhaps the maximum for a polygon is related to the number of vertices. 1 until you get to 5 which is 2, six is 3 I think.

Or perhaps just analyse the perimeter of 2D polygons by looking at the angle between segments and counting the number of times it changes from curving outwards to inwards and do that for each intermediate 2D result.

jordanbrown0:
Chris: I don't know enough about the algorithms to torture test them, but this test puts 2000 overlapping cubes into the model, each a linear extrusion with an unnecessary convexity of 200, and performance appears unaffected.

```
nx = 10;
ny = 10;
nz = 10;
size = 2;
space = 3;
convexity = 100;

for (t = [[0,0,0], [1,1,1]*space/2])
    translate(t)
    for (x = [0:nx-1], y = [0:ny-1], z=[0:nz-1])
        translate([x*space, y*space, z*space])
        linear_extrude(convexity=convexity, height=size, center=true)
            square(size, center=true);
```

Hans:  Amusingly, I tried that Menger Sponge example, added the render(), and didn't see any problems... until I realized that I was running my own build where the default convexity is 10.

jordanbrown0:
The Menger Sponge example, with a render added on line 13, appears to be convexity 15.  At least, I can't see any artifacts at 15.  It's pretty good at 10, but when you look at the bottom and move it around, there are little sparkles.  It is noticeably slower at 15 than at 1, but then again at 1 it's a complete disaster visually.  I wouldn't call it sluggish at 15, but it's not snappy.  Just eyeballing it (and with basically no experience with this sort of thing), maybe 5-10fps?  Completely usable, but not fluid.

nophead:
It obviously depends on n but not obvious how. Also I think it depends on viewing direction.  So it isn't the true convexity of the object but the convexity in the viewing direction. I.e. you don't have to pass a ray through all possible directions, just perhaps all the rays from the camera.

Perhaps OpenCSG should iterate until no pixels change.

jordanbrown0:
On 9/28/2021 12:56 AM, Chris wrote:
>
> Perhaps OpenCSG should iterate until no pixels change.
>

Perhaps.  But anything that involves changing the display algorithms is
much harder and much riskier than bumping up the convexity default. 
Bumping up the convexity default is easy and, especially if we provide a
way to set the default, very low risk... and it would make a lot of
cases just work.

jordanbrown0:
So, questions:
* Provide a way for the user to set the default?
  * no
  * $convexity
  * preference

and the internal structure questions above.

nophead:
The trouble with a preference is it has to be passed on the command line when generating PNGs and old versions won't accept it.

The problem with $variables is they are now broken, so adding another doesn't seem a good idea.

t-paul:
My vote would be: No user setting, but a simple heuristic based on polygon convexity and number of outlines, e.g. something like:

- polygon convex, only 1 outline (= no cut-outs) = 1
- polygon convex, multiple outlines = `min(ceil(sqrt(outlines)), 10)` (= assuming we normally don't have all cuts lined up in view direction, maybe limit to something less to 10 but at least have some upper bound)
- if polygon is not convex, add 1 or 2?

jordanbrown0:
@t-paul That sounds a heck of a lot more complex than changing a bunch of constant 1s to constant 10s, or even than adding a preference or $convexity.  And what about render(), polyhedron(), import(), and surface()?

Here's a simple demonstration of the problem using only 3D primitives:
```
difference() {
    render(convexity=1) {
        cube([10,5,10]);
        cube([5,10,10]);
    }
    cube([1,1,1], center=true);
}
```

@nophead Although the CLI interaction is interesting, for this particular case it doesn't seem relevant:  for a value of 10, the observed impact even for a complex object is measured in small fractions of a second per frame - negligible when generating a single frame.

nophead:
Yes but if anybody changed the default to something higher than 10 then they might get incorrect PNGs if they don't pass it on the command line.

Does it slow down for 1000 linear extrudes with convexity 10 the same as one linear extrude with convexity 10,000?

Does it have any effect at all on speed when there isn't a following CSG operation?

t-paul:
How is that complicated? It's literally a single line function to be called with the polygon as argument. It's far from perfect, but will not punish convex polygons or trivial cases with a random high `convexity` value that has no good justification and unknown performance impact.

We could start with covering 2D cases first and have a go at 3D later.

jordanbrown0:
@t-paul  I don't know much about geometry processing, but can you really determine whether a polygon is convex in one line?

While it's formally true that setting the convexity to 10 has unknown performance impact, there is a great deal of anecdotal and experimental evidence that the performance impact is generally negligible.  We can never know for sure that there isn't a pathological case that would get awful performance.  That possibility is one of the reasons for considering providing a way for the user to set the default.  Plus of course it is only a default; if there's a case that it really hurts you can always explicitly say "convexity=1" on the relevant calls.  We have a *lot* of evidence that people routinely run into rendering problems with the default of 1; it's almost certainly in the top five newcomer complaints.  (Alongside variables that don't vary, Z-fighting, and non-manifold-ness.)

@nophead I've always assumed that preferences provide the default for command-line operations just as they do for GUI operations.  Is that not correct?  You wouldn't be able to **set** the convexity default from the CLI, but that doesn't mean that it wouldn't use the value set in the preferences.

We're poking in the dark a bit with respect to what operations will be impacted, but it seems that there's no impact until you do something like a difference.  My 2000-cube test above does get a bit sluggish when you subtract away a cube and have convexity=10 - at a guess, it's around 2-3fps.  But a difference *elsewhere* in the model doesn't seem to have any impact; I created a separate cube and subtracted from it and performance was still just fine.

jordanbrown0:
Indeed preferences do not control the CLI.

jordanbrown0:
The fact that preferences do not control the CLI suggests that if control over this default is via a preference, there would need to be a CLI option to control it.

An older OpenSCAD would not support that option.

If the reason the user wants to set it is to *increase* the default convexity, they were already going to get crummy results on that older OpenSCAD.  So they have a choice:  (a) their script runs only on newer OpenSCAD, or (b) they don't set it, and they get crummy results on older OpenSCAD and less crummy (but still not ideal) results on newer OpenSCAD.  (It is tempting to have an option that says "set this configuration parameter to this value", that ignores unknown configuration parameters for compatibility, but such a mechanism can't detect typos that *should* be reported as errors.)

If the reason that the user wants to set it is to *decrease* the default complexity, that's theoretically a problem but (with no testing so far) it seems very unlikely to be an appreciable performance hit since the worst performance hit we've observed in the GUI is still well less than a second per frame.

thehans:
To contribute a performance test case, I've modified the example024 to the point where I get ~2.5 FPS with convexity=1, and about 0.5 FPS with convexity 10, on an NVidia GTX 1660.

```
n=3;
D=3^(n-1)*2+1;
convexity = 1;
echo(D);

module menger() {
  difference() {
    cube(D, center=true);
    render(convexity=convexity) for (v=[[0,0,0], [0,0,90], [0,90,0]])
      rotate(v) menger_negative(side=D, maxside=D, level=n);
  }
}

module menger_negative(side=1, maxside=1, level=1) {
  l=side/3;
  cube([maxside*1.1, l, l], center=true);
  if (level > 1) {
    for (i=[-1:1], j=[-1:1])
      if (i || j)
        translate([0, i*l, j*l])
          menger_negative(side=l, maxside=maxside, level=level-1);
  }
}

for(x=[0:1:9],y=[0:1:9],z=[0:1:0]) 
  translate((D+1)*[x,y,z]) menger();
```

Enabling the experimental `vertex-object-*` features make it fast enough to increase the z range to `[0:1:9]` for similar performance.

Also to confirm what @nophead was asking about, if you comment out the `difference()` and `cube(...)` lines of module `menger()`, then changing convexity to arbitrarily large values doesn't seem to make any difference in performance.

t-paul:
> ...but can you really determine whether a polygon is convex in one line?

Yes, [`poly.is_convex()`](https://github.com/openscad/openscad/blob/master/src/Polygon2d.h#L47)

nophead:
I see that function looks for outward turns in the perimeter. So a similar function could count changes in direction from inwards to outwards and that would be a worst case convexity value, not too bad an overestimate. Then the only use of the convexity parameter would be to lower it if anybody found a case where it was too bad a performance hit, which is unlikely. It would always be 1 in convex cases and rarely as high as 10.

jordanbrown0:
A straightforward gear would have one such transition per tooth.

nophead:
Yes when I make gears I set it equal to the teeth but that is an overestimation because it depends on the tooth depth. You would have to work out the most teeth a cord could pass through.

jordanbrown0:
For Hans's 100x Menger-3 model, it takes about 2m on my system to generate a PNG, and the difference between convexity=1 and convexity=10 is about 10s.  It took about three and a half hours for a full render, with a total of ~700K facets.  I think it's fair to say that that's an extremely complex model.

thehans:
There may actually be some OpenCSG configuration settings that would be worth experimenting with.
The documentation is a bit vague, but I think the "depth complexity algorithm" if set, may no longer require setting convexity on individual CSG terms.

See the Algorithm and DepthComplexityAlgorithm [descriptions here](http://opencsg.org/#algorithm) (if I'm not mistaken, I think our current code uses Automatic, unless "Force Goldfeather" preference is on, and does not explicitly choose a DepthComplexityAlgorithm):

MichaelAtOz:
Just a note on Software rendering (ie OpenGL hence preview).
I don't know if it is used much, but I do it in a remote desktop setting.

#3628 has some detail, but since 2021.01 it performs very badly, measured in SPF not FPS.
I would expect that noticeable change in H/W rendering will be much worse in S/W.

I'm still catching up on my reading, I'll try some of the above with S/W rendering.

t-paul:
See #4081.

PR #3906, opened 2021-09-29
Strip execute bit where isn't due.

themactep:
Linux guy's pet peeve :)


t-paul (approved):
Sounds good, we really don't want to execute those :grinning:.

PR #3941, opened 2021-10-14
Define CGAL_DEBUG to simplify header includes.

thehans:
Just noticed the usage of this in CGAL's assertions.h, so we don't need to push and pop NDEBUG around all CGAL includes anymore.
https://github.com/CGAL/cgal/blame/master/STL_Extension/include/CGAL/assertions.h

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Run Test Suite
         run: |
           cd build
-          cp -v "${RUNNER_TEMP}/msys/msys64/mingw64/bin/opengl32.dll" .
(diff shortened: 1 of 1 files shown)


thehans:
Just for curiosity's sake, I tested how much overhead this causes, by comparing an unsafe `CGAL_NDEBUG` build with the "standard" `CGAL_DEBUG`.  It ends up being about 6.6% overhead averaged over all the tests ran, so nothing *too* crazy.

I ran the full suite `-C All`, filtered by anything with "png" in the name to get mostly cgal tests
`time ctest -C All -R png`

standard/safe build
```
real    9m43.132s
user    9m21.790s
sys     0m54.193s
```

unsafe build
```
real    9m6.351s
user    8m47.067s
sys     0m53.620s
```

PR #3942, opened 2021-10-15
Update expected test results for valid tests.

thehans:
These tests aren't run for normal "ctest", but are included in "ctest -C All"
Changes presumably caused by textmetrics PR (2.4% size difference?).

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Run Test Suite
         run: |
           cd build
-          cp -v "${RUNNER_TEMP}/msys/msys64/mingw64/bin/opengl32.dll" .
+          cp -v "${RUNNER_TEMP}/msys64/mingw64/bin/opengl32.dll" .
           cd tests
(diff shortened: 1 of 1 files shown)


jordanbrown0:
Yes, all text spacing changed 2.4%.  I'm a little puzzled by the change for the single-letter cube case, but probably centering changed just a hair, because I think centering is over the "advance" and it is the advance that increased 2.4%.

jordanbrown0:
Thanks, sorry for not fixing that as part of the initial change - I didn't know about the "extra" tests.

thehans:
> Thanks, sorry for not fixing that as part of the initial change - I didn't know about the "extra" tests.

No problem, they aren't run by any of the automated systems due to time constraints, and get missed somewhat often when subtle changes are made.  
I just tend to run the full suite every now and then, particularly when profiling performance changes, so I noticed it recently.

For future reference, the different test configs are documented here:
https://github.com/openscad/openscad/blob/master/doc/testing.txt#L52-L58

PR #3944, opened 2021-10-16
MacOS: Silence warning about OpenGL deprecation.

t-paul:


kintel:
I briefly looked into the zlib issue a few days ago, and it looks like the problem is that the CMake zlib finder as used in lib3mf is broken; zlib comes with macOS, but the MacPorts version is somehow preferred by CMake. Could be a pkg-config issue as well.

That said, bundling our own zlib is a perfectly good workaround.

kintel:
PS! When manually inspecting a macOS filesystem, it may look as if zlib is missing. That's because Apple invented a clever overlay filesystem to be able to pull the run-time libraries from the appropriate SDK at runtime. I don't quite understand why and how that works though.

t-paul:
Ah, I was wondering if the better option would be to exclude it from beeing copied to the App bundle. But I'm not sure yet how to do that and without a local dev machine I can't spend a lot of time on that.

Issue #3907, opened 2021-09-29
Icon for macOS Big Sur


MohanMaker:
macOS Big Sur introduced a new icon style (square icons instead of round icons). It would be great if OpenSCAD had a new icon to agree with this design scheme. I know some individual users have already created their own versions of a square icon here: https://macosicons.com

TheJKM:
The only file that has to be updated is `icons/OpenSCAD.icns`, this container format is only used on macOS. This will change the icon for all versions of macOS, as Apple writes:

> IMPORTANT
> When you update your app for macOS 11, use your new app icon design to replace the icon you designed for earlier versions. You can’t include two different app icons for one app, and the macOS 11 app icon style looks fine on a Mac running Catalina or earlier.
(https://developer.apple.com/design/human-interface-guidelines/macos/icons-and-images/app-icon/)

I'm not a fan of the linked icon, I'd prefer the current icon on a white square to differ the icon from the background. In the linked version, they become blurred. An Example:

<img width="469" alt="image" src="https://user-images.githubusercontent.com/21987613/136023391-9c227fe0-dae4-4026-9710-0803dd46ebb7.png">

TheJKM:
Now that the icon is merged, I think we can close this.

t-paul:
Closed via #3970.

PR #3947, opened 2021-10-18
Bugfix: Correctly associate nightly icon with the app bundle

kintel:
After porting the build system to CMake, the nightly icon wasn't associated with the app bundle, causing the snapshot build to be iconless.


t-paul:
Ah, I did not mean to imply you did not. I was just wondering if you know about that special case :grin:. Related other issue... icon suggestion for BigSur #3907? Sounds like Apple changed things a bit.

PR #3952, opened 2021-10-23
MacOS: Homebrew on Apple M1 and Bundle resource fixes + add Unix Mode

ljanyst:
It seems that the MacOSX build with CMake is a bit rusty, so I fixed a bunch of things and added a Unix mode build. 

This PR:
 * Adds a QT search path compatible with Apple M1 installation of Homebrew
 * Fixes bundle versioning
 * Copies the necessary resources inside the bundle
 * Adds the Unix mode where you can `make install` stuff the Unix way (OFF by default)


ChrisCoxArt:
Kintel updated the build for M1 and both homebrew and dependency script builds should both be working now.

Issue #2076, opened 2017-07-29
Render() exposed as a function


TLC123:
I thought i posted this before but i cant find it.

If render() were exposed as a function, points and faces returned as lists could be subjected to some arbitrary operation before drawn as polyhedron or polygon in 2D case.
Related:  #1324  #301 

Benefit of exposing geometry this way is that a hole bunch of issues can safely be resolved in user space. with little commitment to future support.

```
My_polymesh = render_to_polyhedron(){ sphere(1) ; }
returns [points = [ [x, y, z], ... ], faces = [ [p1, p2, p3..], ... ]] for polyhedron.

My_polymesh = render_to_polyhedron(){ import("example012.stl", convexity=3); }
returns [points = [ [x, y, z], ... ], faces = [ [p1, p2, p3..], ... ]] for polyhedron.

My_polygon = render_to_polygon (){ circle(1) ; }
returns [points = [ [x, y, z], ... ], paths= [ [p1, p2, p3..], ... ]] polygon .

My_polygon = render_to_polygon (){import("example013.dxf" }
returns [points = [ [x, y, z], ... ], paths= [ [p1, p2, p3..], ... ]] polygon .
```

nophead:
Yes it would be nice but the problem is OpenSCAD evaluates expressions before it creates the geometry, so it would be a major redesign.  You might have geometry that depends on expressions that depend on geometry, etc.

TLC123:
Sure, pre-sub-evaluation of anything  under render treated as a separate model and successiv replacement with returned geometry as an expression. Finally evaluate parent model. It could be several recursive evaluations. Cyclic references wouldn't be an issue though. I specifically proposing naming the feature render() since a time penalty is already expected of render().

I reintroduced this issue because i wanted to add a specific bounty, a bounty i'm willing to raise considerably considering its a major challenge if there is serious support.

I think the feature carries way more value than it seams at first glance. 
virtually every mesh algorithm in the book becomes low hanging fruit to solve in user space.
usage example: mesh post processing like 
subdivision, edge rounding, decimation, import repairs, object distribution over surfaces, measurements, collision tests.
In a singel blow one feature expands OpenScads usefulness considerably without breaking backwards compatibility. 

cheers

nophead:
I think there was a working PR submitted a while ago that did nearly the same thing called probe but it was a module rather than a function.

TLC123:
Disclaimer: Thinking out loud readings ahead.

[...]

PR #3956, opened 2021-10-26
data = render() feature.

sublim9:
This feature allows render() to be used as a function.
This function returns information about the targeted module on the right.
Information such as bounding box, vertex, face and path (for 2d objects) data.
The syntax, which is a bit quirky and will be subject to change, at this moment is:
~~myvar = render() @ some_module(){};~~

```
data = render() module();
```

or

```
data = render() { module(); ... };
```

But at this moment not like this:

```
data = render() module() { ... };
```

This is still a work in progress.


jordanbrown0:
It might be wise to add in a layer of indirection, so that `render()` returns a list of objects, each of which has a list of faces/paths, so that it would be clear how a multi-color / multi-material model would be represented.

Even without multi-color / multi-material support, it looks like CGAL keeps track of "volumes"; if it does then returning them separately might allow for, e.g., automatic or semi-automatic exploded diagrams.

If that capability isn't designed in now, it's at least a bit unclear how it would be compatibly added later.

sublim9:
Interesting, I'll look into it. Thank you for your input.

t-paul:
Yes, good point. I'd suggest to discuss the design/ideas first before directly implementing things.

jordanbrown0:
Without looking at the implementation, my guess is that adding the indirection that I suggest is just a matter of how the results are marshaled into the result object, and whether you interpose another vector or object or two.

More than that if you return multiple volumes separately.

Of course, there will be more work when we eventually do multi-color/multi-material, but the idea is that it won't be necessary to change the interface here.

jordanbrown0:
There might also be an interaction with lazy union.

jordanbrown0:
This looks like it's strongly related to (or perhaps identical to) issue #2076.

jordanbrown0:
In other areas, I've been thinking about the syntax for creating objects.  (Ref #3684 , #3088 , et cetera).

In JavaScript and Python, the equivalent syntax looks something like so:
```
obj = {
    mem1: val1,
    mem2: val2,
    ...
};
```

We could emulate that, but it would be a wholly new syntax for OpenSCAD.

But we already have a syntax that assigns values to names in a particular scope:  ordinary assignments.  Perhaps we might consider something that looks like that, something like:

```
obj = {
    mem1 = val1;
    mem2 = val2;
    ...
};
```

That in turn might make one wonder:  what happens if you include some geometry in there?

```
obj = {
    mem1 = val1;
    mem2 = val2;
    cube();
    ...
};
```

Maybe the result is an object that includes the user-specified members, plus the geometry in some representation.

The biggest question in my mind would be what name the geometry would get.

---

But:  should there be both a mechanism to return the CSG tree and a mechanism to return the rendered triangle soup?

Perhaps the structure above would return the CSG tree, and render(obj) would return the triangle soup.

jordanbrown0:
I've been thinking a bit more about my comment of May 11, having { ... } be an expression that returns an object that includes both ordinary values and introspectable geometry.

I'm pretty happy with the high-level concept, that the geometry that it returns would be introspectable CSG, and that render(csg) as a function would return introspectable faces and vertexes.

But then what?  How could you incorporate that data into the model?  As you've noted, faces-and-vertexes could be fed to polyhedron().  But what could you do with a CSG tree?

Maybe render(csg) as a module would add it to the model, but that implies a perhaps-unwanted CGAL render.

Alternatively, you could use that data purely for introspection, and for anything to be included in the model re-evaluate it in an ordinary way.  But that seems redundant.

Maybe the render(csg) function would return a CSG tree where the geometry has been rendered down into polyhedra, perhaps with color() nodes.

jordanbrown0:
Maybe import(csg)... not exactly a match, but not exactly not a match either.

jordanbrown0:
I implemented something like this in PR #4478 .

PR #3957, opened 2021-10-27
Building OpenSCAD with Visual Studio

dannypike:
Instructions and a few modifications to support building and debugging OpenSCAD from within Microsoft Visual Studio on a windows PC. See the file `VisualStudio.md` for details.


t-paul:
Thanks for the update and sorry for the current churn on the build stuff. I just want to mention this is a very much appreciated change. It might take a bit more time for someone to review but it will be great to have.

dannypike:
You're welcome. I'll try to keep it in synch with the master branch, and vcpkg updates, as they appear.

CMakeLists.txt
```diff
@@ -123,6 +123,44 @@ set(BOOST_DIRECTLY_REQUIRED_LIBRARIES filesystem system regex program_options)
...
+  add_definitions(/wd4244)  # conversion, possible loss of data
+  add_definitions(/wd4267)  # '': conversion from '' to '', possible loss of data
+  add_definitions(/wd4305)  # '': truncation from '' to ''
+  add_definitions(/wd4996)  # '': This function or variable may be unsafe. Consider using [...]
+
+  # pkg-config is required to build OpenSCAD, but is not installed by default as a part of Windows.
+  # Installation instructions are available online e.g. as a comment posted by "E_net4 the curator":
+  #   https://stackoverflow.com/questions/1710922/how-to-install-pkg-config-in-windows
+  #
+  find_path(PKGCONFIG_EXE pkg-config.exe REQUIRED)
```
thehans, line 139: This should NOT be REQUIRED, if you want the subsequent conditional to ever be reached.

thehans:
I started to review this but didn't get through reproducing the build process before I had to prepare for a trip.  Will be back in about a week.

CMakeLists.txt
dannypike: The reason why this is REQUIRED is in the build instructions.

t-paul: I think what that comment meant to say is that the following `if (NOT PKGCONFIG_EXE)` will not be reached with the `REQUIRED` in `find_path` as `cmake' will stop processing there already. So it will not give the custom message.

dannypike: Ouch, sorry; I am blind. :(

Fixed and pushed in 49191e57c7f76405dbadd19776196cc578d71fec. I can't test i,t because qt5 won't build for me at the moment, but hopefully it's a simple-enough change.

t-paul: No worries, we build test all the current platforms. As long as those go fine, adding this is an improvement even if there's some small glitches we may find later. It's great to get VS suport.

thehans:
I've tested this and got the build to succeed... up to the post-build `translation-make.sh` script where it fails with:
```
/bin/bash: C:/Users/hans/Documents/GitHub/openscad_vs/scripts/translation-make.sh: No such file or directory
```

It looks like cmake finds a shell at this location: `C:/Windows/System32/bash.exe`, from this line in CMakeLists.txt:
```
find_program(SHELL_EXE NAMES sh bash $ENV{SHELL})
```
As far as I can tell, this bash expects the path to the script to look like  `/mnt/c/...` instead of `C:\...`.
I read somewhere that this is some conflict with having WSL installed, and the commenter said that uninstalling WSL fixed it, but I'd rather not require that if there's any other solution.

@dannypike Does your `translation-make.sh` script succeed?  What `SHELL_EXE` path does your cmake find?

thehans:
Some other notes:

- VisualStudio.md should be moved under `/doc/` path.
- I'd like to simplify the overall steps as much as possible, with less reliance on links to 3rd party instructions etc.  There are a few things I did differently for my own build, including a couple changes to CMakeLists.txt which I can push to your branch.  But I figured I'd run them by you beforehand:
  -  `pkg-config` can be provided by vcpkg, [with a bit of a workaround](https://github.com/microsoft/vcpkg/issues/20689#issuecomment-996483391) (its actually `pkgconf`, technically different from `pkg-config`, but compatible).
  - What part of your build requires "Microsoft MPI" being installed?  I don't even know what this does (is it a Message Passing Interface?), but I don't believe it is actually necessary.  I uninstalled it and still was able to build (up until the translation-make step mentioned above).
  - For flex/bison I just used the [latest release from here](https://github.com/lexxmark/winflexbison/releases).  Unzip to your preferred path and make sure it is added to system PATH.  I feel that is a little simpler than going through mingw.  I would have liked if vcpkg could provide these tools itself, as it already uses them in builds of a few of our dependencies, but it seems that its only available to [maintainers of vcpkg portfiles](https://vcpkg.readthedocs.io/en/latest/maintainers/vcpkg_find_acquire_program/), and not for "end-users" of vcpkg.

I think our doc should also mention that changes to PATH require a restart of Windows to take effect, or more precisely just a restart of Windows Explorer.  This can be done with Ctrl-Alt-Delete to open Task Manager,  scroll to "Windows Explorer" Process, right click and Restart.  This avoids having to logout of your session or doing a full reboot.

thehans:
I just remembered one other annoyance I had to deal with, where Microsoft (on Windows 10) tries to force you through their microsoft store for python even when you have a suitable version installed.  They [put fake exes on the path which just link to the store, and put their path before others so the fake exes are found first](https://stackoverflow.com/questions/58754860/cmd-opens-window-store-when-i-type-python), really gross.

I ended up removing any `*\WindowsApps` from path so that doesn't happen.

dannypike:
>     * VisualStudio.md should be moved under `/doc/` path.

w.r.t. eb7f2943a629a04673c1924dca9274cfa03a0609

The file is renamed to `doc/Windows-VisualStudio.md` and a separate file called `doc/Windows-MinGW.md` contains the instructions for cross-compiling. I felt that having two separate files emphasizes that the instructions are independent of each other; you only need to do the one that applies to you, but it ought to be OK to do both, if you want to do that, too (though I haven't tested the MinGW version).

dannypike:
> I think our doc should also mention that changes to PATH require a restart of Windows to take effect, or more precisely just a restart of Windows Explorer. This can be done with Ctrl-Alt-Delete to open Task Manager, scroll to "Windows Explorer" Process, right click and Restart. This avoids having to logout of your session or doing a full reboot.

You do not need to restart Windows Explorer. As long as you update the environment variables in the registry, using `setx` or the Control Panel applet, it is only necessary to restart `Visual Studio`.

As a long-term Windows user, I think that it is overkill to tell a Windows user that they need to do it. I'm willing to be overridden, but I agree that we're already at risk of being too verbose with the amount of detail in the instructions, as you mentioned elsewhere. There's always a line "not to cross", and my line excludes teaching people about how the environment works. :)

dannypike:
> I just remembered one other annoyance I had to deal with, where Microsoft (on Windows 10) tries to force you through their microsoft store for python even when you have a suitable version installed. They [put fake exes on the path which just link to the store, and put their path before others so the fake exes are found first](https://stackoverflow.com/questions/58754860/cmd-opens-window-store-when-i-type-python), really gross.
> 
> I ended up removing any `*\WindowsApps` from path so that doesn't happen.

My earlier comment about not including too much detail also applies to this. I haven't ever had problems with installing Python and maybe that sort of thing doesn't belong in our instructions for building OpenSCAD?

thehans:
@dannypike thanks for the replies.  Can you comment on the shell script issue I was having?  I was not able to resolve that.

Regarding environment variables, I'm not sure what my exact situation was, where it seemed that Windows Explorer needed a restart.  I only edit them through the control panel applet, and IIRC simply restarting Visual Studio was not seeing the updated values, but maybe I was mistaken.

dannypike:
> @dannypike thanks for the replies. Can you comment on the shell script issue I was having? I was not able to resolve that.

@thehans thank you for all of your work. Yes, I will look at and comment on them all. I only have an hour or so spare each day, at the moment. I did forget to mention that I did test the instructions on my development machine and on virtual machines that only contain the latest Windows, Git and Visual Studio.

> Regarding environment variables, I'm not sure what my exact situation was, where it seemed that Windows Explorer needed a restart. I only edit them through the control panel applet, and IIRC simply restarting Visual Studio was not seeing the updated values, but maybe I was mistaken.

Would you try it again, for me? How are you verifying the environment variables in Visual Studio? When I checked it yesterday, I used the SysInternals [Process Explorer](https://docs.microsoft.com/en-us/sysinternals/downloads/process-explorer) utility to look at the environment, and I also used a powershell+cmd debug window. I didn't have time to try the full build.

dannypike:
> @dannypike Does your `translation-make.sh` script succeed? What `SHELL_EXE` path does your cmake find?

@thehans I have both ```sh``` and ```bash``` available on my PATH:
```
D:\work\github.forks\openscad>where bash
C:\Program Files\Git\usr\bin\bash.exe
C:\MinGW\msys\1.0\bin\bash.exe

D:\work\github.forks\openscad>where sh
C:\Program Files\Git\usr\bin\sh.exe
C:\MinGW\msys\1.0\bin\sh.exe
```

So CMake chooses ```sh``` for me:
```
find_program(SHELL_EXE NAMES sh bash $ENV{SHELL})
message(STATUS "will run translation script using the shell at: ${SHELL_EXE}")
```
produces this output on compiling ```CMakeLists.txt```:
```
1> [CMake] -- will run translation script using the shell at: C:/Program Files/Git/usr/bin/sh.exe
```
and the script runs without reporting any errors when I build with ```F6```:
```
[245/245] cmd.exe /C "cd . && "C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe" -E vs_link_exe --intdir=CMakeFiles\OpenSCAD.dir --rc=C:\PROGRA~2\WI3CF2~1\10\bin\100203~1.0\x64\rc.exe --mt=C:\PROGRA~2\WI3CF2~1\10\bin\100203~1.0\x64\mt.exe --manifests  -- C:\PROGRA~1\MICROS~4\2022\COMMUN~1\VC\Tools\MSVC\1430~1.307\bin\Hostx64\x64\link.exe /nologo @CMakeFiles\OpenSCAD.rsp  /out:openscad.exe /implib:openscad.lib /pdb:openscad.pdb /version:0.0 /machine:x64 /debug /INCREMENTAL /subsystem:windows -subsystem:windows -ENTRY:mainCRTStartup -stack:8388608  && cmd.exe /C "cd /D D:\Users\dan\CMakeBuilds\7224ddf8-773a-4354-a378-45639c171864\build\x64-Debug && C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe -noprofile -executionpolicy Bypass -file D:/work/github.forks/vcpkg/scripts/buildsystems/msbuild/applocal.ps1 -targetBinary D:/Users/dan/CMakeBuilds/7224ddf8-773a-4354-a378-45639c171864/build/x64-Debug/openscad.exe -installedDir D:/work/github.forks/vcpkg/installed/x64-windows/debug/bin -OutVariable out && cd /D D:\Users\dan\CMakeBuilds\7224ddf8-773a-4354-a378-45639c171864\build\x64-Debug && "C:\Program Files\Git\usr\bin\sh.exe" D:/work/github.forks/openscad/scripts/translation-make.sh ""
  msgfmt -c -v -o ./locale/cs/LC_MESSAGES/openscad.mo ./locale/cs.po
  244 translated messages, 74 fuzzy translations, 160 untranslated messages.
  msgfmt -c -v -o ./locale/de/LC_MESSAGES/openscad.mo ./locale/de.po
  478 translated messages.
  msgfmt -c -v -o ./locale/es/LC_MESSAGES/openscad.mo ./locale/es.po
  218 translated messages, 113 fuzzy translations, 147 untranslated messages.
  msgfmt -c -v -o ./locale/fr/LC_MESSAGES/openscad.mo ./locale/fr.po
  382 translated messages, 57 fuzzy translations, 39 untranslated messages.
  msgfmt -c -v -o ./locale/hy/LC_MESSAGES/openscad.mo ./locale/hy.po
  384 translated messages, 55 fuzzy translations, 39 untranslated messages.
  msgfmt -c -v -o ./locale/pl/LC_MESSAGES/openscad.mo ./locale/pl.po
  410 translated messages, 64 fuzzy translations, 4 untranslated messages.
  468 translated messages, 10 untranslated messages.
  msgfmt -c -v -o ./locale/tr/LC_MESSAGES/openscad.mo ./locale/tr.po
  ./locale/tr.po:8: warning: header field 'Project-Id-Version' still has the initial default value
  ./locale/tr.po:8: warning: header field 'PO-Revision-Date' still has the initial default value
  ./locale/tr.po:8: warning: header field 'Language-Team' still has the initial default value
  471 translated messages, 3 fuzzy translations, 4 untranslated messages.
  msgfmt -c -v -o ./locale/uk/LC_MESSAGES/openscad.mo ./locale/uk.po
  283 translated messages, 68 fuzzy translations, 127 untranslated messages.
  msgfmt -c -v -o ./locale/zh_CN/LC_MESSAGES/openscad.mo ./locale/zh_CN.po
  448 translated messages, 8 fuzzy translations, 22 untranslated messages.
  msgfmt -c -v -o ./locale/zh_TW/LC_MESSAGES/openscad.mo ./locale/zh_TW.po
  408 translated messages, 53 fuzzy translations, 17 untranslated messages.
```
This is on my development machine, not a clean VM. What happens for you?

dannypike:
@thehans. I have completed all of the changes that I think were raised in your comments. I'm sorry for the tangled Git history over the last few days, but I have been testing this across a couple of machines (my development one and a clean test VM). Git is the way that I like to ensure that I keep them both in sync. Here is a summary of what I have done:

1. I have removed the requirement for Microsoft MPI from the build instructions. I do not know why I was getting those errors before, but I agree that it is not needed to compile any of the OpenSCAD dependencies.
2. I have changed `tests\CMakeLists.txt` script to look for a suitable version of python in the `vcpkg` tools folder. It might pick a different version depending on what packages you have installed, but it should ensure that it is at least version 3.4. Please take a look at what I have done (the logic inside the `if(MSVC)` statement on line 226) and see if you agree that this is a valid way to do it. It's the first time I have tried to use this technique in `vcpkg`. I've removed from the build instructions any need to install Python explicitly.
3. I've removed ```pkg-config``` from the build instructions and added ```pkgconf``` to the list of packages that will be installed by ```vcpkg```, as per your [workaround](https://github.com/microsoft/vcpkg/issues/20689#issuecomment-996483391). Thank you for that suggestion; it simplifies things a lot!
4. I've changed the suggested way to install ```flex``` and ```bison```. It now suggests that you use Chocolatey, or install them from github, as per your suggestion. Chocolatey is good at this sort of thing; there's no need to set the PATH or do other ancilliary tasks. It also adds one of the most useful commands that is missing from standard Windows: ```refreshenv```.
5. I've moved and split the Windows build instructions into two files: ```doc/Windows-VisualStudio.md``` and  ```doc/Windows-MinGW.md```.
6. I've updated to a newer version of the master branch (33d9cea27dc51f8f7eca455adf832b8d79d30b36) and verified that it still builds, installs and runs for me.

Please try this new version and let me know what you find.

CMakeLists.txt
```diff
@@ -132,8 +132,41 @@ endif()
 
 set(BOOST_DIRECTLY_REQUIRED_LIBRARIES filesystem system regex program_options)
 
+find_program(PKG_CONFIG_EXECUTABLE NAMES pkgconf.exe pkg-config.exe)
 find_package(PkgConfig)
 
+# Note: Saving CMAKE_MODULE_PATH as CGAL will overwrite it.
+# Reconsider this after CGAL 5.4: https://github.com/CGAL/cgal/pull/6029
+if ("${CMAKE_MODULE_PATH}" STREQUAL "")
```
thehans, line 140: This whole section is doing some odd things.  `CMAKE_MODULE_PATH` Is actually a list of paths to search over, not a single path, so I don't think there should be any check for empty string.  
The previous behavior was actually appending a path to the existing value, instead of completely replacing it.  

If you look at the automated builds (click the red X next to your commits), you can see that MXE builds have been failing to find libzip since your changes.  This is likely because the MXE toolchain sets its own CMAKE_MODULE_PATH, and so our additional path is not appended here.

If a platform/toolchain, such as VCPKG or MXE provides `Find[package].cmake` files etc, its best to use those before our own.  (Ideally we wouldn't need to maintain our own *at all*, if every platform had all the needed find modules).

CMakeLists.txt
```diff
@@ -132,8 +132,41 @@ endif()
...
+  set(ORIGINAL_CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/Modules")
+  set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/Modules")
+else()
+  message(STATUS "Saving CMAKE_MODULE_PATH '${ORIGINAL_CMAKE_MODULE_PATH}'")
```
thehans, line 144: This message is printing the value of a variable before it is set.

CMakeLists.txt
```diff
@@ -235,18 +278,25 @@ if (MSVC)
   target_include_directories(OpenSCAD SYSTEM PRIVATE ${HARFBUZZ_INCLUDE_DIRS}/harfbuzz)
   list(APPEND COMMON_LIBRARIES harfbuzz::harfbuzz)
 
-  find_package(unofficial-fontconfig CONFIG REQUIRED)
-  list(APPEND COMMON_LIBRARIES unofficial::fontconfig::fontconfig)
+  find_package(fontconfig REQUIRED)
+  list(APPEND COMMON_LIBRARIES ${VCPKG_LIBRARY_DIR}/fontconfig.lib)
```
thehans, line 282: The whole point of `find_package` is that it sets [various variables](https://cmake.org/cmake/help/latest/manual/cmake-developer.7.html#standard-variable-names) eg `[packageName]_LIBRARIES`, so that you don't have to construct their paths yourself.  

In this case vcpkg was previously providing a package with unofficial in the name, but it seems they must have dropped that recently.  Still there should be either standard variable name set, or a namespaced target (with double colons `::` like in the removed code)

CMakeLists.txt
```diff
@@ -235,18 +278,25 @@ if (MSVC)
...
 
-  find_package(unofficial-glib CONFIG REQUIRED)
+  find_package(glib2 REQUIRED)
   target_include_directories(OpenSCAD SYSTEM PRIVATE ${GLIB2_INCLUDE_DIRS})
-  list(APPEND COMMON_LIBRARIES unofficial::glib::gio unofficial::glib::glib unofficial::glib::gmodule unofficial::glib::gobject)
+  list(APPEND COMMON_LIBRARIES
```
thehans, line 286: Use namespaced library targets

CMakeLists.txt
```diff
@@ -235,18 +278,25 @@ if (MSVC)
...
+    ${VCPKG_LIBRARY_DIR}/gobject-2.0.lib
+    )
 
   find_package(double-conversion CONFIG REQUIRED)
   list(APPEND COMMON_LIBRARIES double-conversion::double-conversion)
 
-  find_library(GETTEXT_LIBRARY libintl)
-  list(APPEND COMMON_LIBRARIES ${GETTEXT_LIBRARY})
+  find_library(GETTEXT_LIBRARY intl)  
+  list(APPEND COMMON_LIBRARIES ${VCPKG_LIBRARY_DIR}/intl.lib) # In MSVC, 'libintl' is called 'intl'
```
thehans, line 297: `find_library` sets `GETTEXT_LIBRARY` to the path, so use it.

thehans:
@dannypike your PR is showing 487 files changed now.  There was a major reformatting of the codebase recently (#4095) that touched pretty much every source file (mostly whitespace, settling on a standard indentation model), so it seems that something went wrong during one of your latest merges.

I also added some code review comments regarding your changes in `CMakeLists.txt`.  Up to a certain point there were no changes to the various `cmake/modules/Find[package].cmake` files, and then suddenly you modified a number of them.  I feel these are unnecessary changes (since it was working for you before) as long as you don't completely overwrite vcpkg's `CMAKE_MODULE_PATH`.

t-paul:
Sorry, I did not expect the code formatting to cause such issues. As this has progressed quite a bit, how is that suggestion:

I'll try to rebase this onto latest master branch and we import the changes into a dedicated branch in the openscad repo.
That way it might also be a bit easier trying to setup a github action running this (at least I believe that's possible). I would not want to run that later on each commit, but a schedule like once a week or daily if we get enough caching could be quite usable too.

dannypike:
@thehans I'm sorry about the number of commits; it's a side-effect of the way that I want to check that my changes to the build instructions match want actually happens when a user downloads and tries to build OpenSCAD on a new machine. As @t-paul says, I think that these are caused by me keeping this branch in sync with the master by back-merging them.

You can rest assured that I will never deliberately reformat any of the OpenSCAD files (I hate it when people do that, too!) and I will not change any of the logic without either: raising a separate ticket to talk about it, or by adding or removing `#if _MSVC_VER` gates around my modification. Every other change to this port will have come from a fetch-and-remerge of the from the `master` branch.

As for he current state of this `vs2019-port` branch:

As I have learned more about the inner workings of OpenSCAD, I have been simplifying and adjusting the build instructions as much as I can. This is partly because I don't want to scare off new users and partly because I don't want to generate work for myself later, in the form of support tickets. By automating as much of the process as I can using a manifest file, `'vs`' configurations etc, I think I have got it quite close to saying `please clone this repo and hit `F5'`. If I can get the TL;DR section down to three or four steps, then I will be happy.

To avoid breaking things, as I only have a short amount of time each day, I use Git to do this in very small steps and test each one by cloning on a clean VM. That means there are lots of small commits that you can safely ignore for the moment. When I have something ready for you to check, I will write a commit that mentions you directly, and detail the things that you need to compare against the last image that you , so you can do a `git diff` between the new "published" HEAD and the last repo that you commented on.

Incidentally, this is the reason why I have not yet reacted to your comments of "12 days ago". I haven't forgotten about them, I only want to keep them "safe and separate" from this storm caused by my many small adjustments to build instructions.

I hope that makes sense!

Dan

dannypike:
> Sorry, I did not expect the code formatting to cause such issues. As this has progressed quite a bit, how is that suggestion:
> 
> I'll try to rebase this onto latest master branch and we import the changes into a dedicated branch in the openscad repo. That way it might also be a bit easier trying to setup a github action running this (at least I believe that's possible). I would not want to run that later on each commit, but a schedule like once a week or daily if we get enough caching could be quite usable too.

@t-paul Please give me a few days to finish what I am doing to the build instructions. Then I will work through the comments that @thehans suggested and it will be fine for you to merge the branch with `master`.

The Visual Studio build for OpenSCAD have been working for a quite a few days, now. Hans' comments relate more to doing things in the correct style; they are not so much about build failures with Visual Studio on his machine. When I have finished this current burst of changes to the build, it should be safe for you to merge and I can then fix the "style" problems separately.

I will announce when I have finished adjusting the build instructions for you to choose whether you want to merge the pull request.

dannypike:
The git tree for this seems to have become corrupted somewhere around 67ccb8b4dc68a4fe409972a568616d60da2c1499. The graph says that should have been a merge commit, but it isn't one. so a whole bunch of files are now missing and it won't build.

So, I'm closing this and will do a fresh PR, by manually copying over my changes to a new fork of the master branch. That will have two benefits: it'll get me out of this hole that I seem to have fallen into and it will also make the PR a lot cleaner for you guys to analyse. There are very few files to copy over, despite the long history: almost all of the changes have been confined to `README.md`, `doc/*` and `CMakeLists.txt`, so this seems like the most sensible thing to do.

dannypike:
This PR has now been superceded by [PR 4126](https://github.com/openscad/openscad/pull/4126). I shall copy-quote comments from here to that PR, as appropriate.

PR #3959, opened 2021-10-28
Default uni-build-dependencies.sh to CGAL 5.3

ochafik:
Note: gutted this PR of the multiple CGAL versions testing github workflow. Might reintroduce it for https://github.com/openscad/openscad/pull/3641 to test w/ and w/o FAST_CSG_USE_SAME_KERNEL, TBC.


t-paul:
@ochafik what's your view? Would it be ok to just require 5.2 now and consider 5.3 once 5.3.1 is out (with the quite annoying build issue fixed that should make it easier to upgrade the MSYS2 package too).

t-paul:
MSYS2 is now also on [CGAL-5.3](https://packages.msys2.org/package/mingw-w64-x86_64-cgal?repo=mingw64). I'm not sure why the github action did not grab that yet, but I guess that's a go for version 5.3.

ochafik:
Sorry for the radio silence! Covid hit me in the early days of November, then it's been busy months :-S

@t-paul 5.3 is only useful to use the same kernel in the hybrid polyhedron as in the nef3 (see FAST_CSG_USE_SAME_KERNEL in https://github.com/openscad/openscad/pull/3641). I could remove support for that (although I'm not sure how much bandwidth I have these days), but if we keep it I'd rather it be tested and not dead code.

Note however that I haven't enabled FAST_CSG_USE_SAME_KERNEL by default on 5.3+ as it needs more benchmarking.

So I'd suggest we just test for 5.2 and 5.3? (with a view to only test on 5.3 once all testing / build environments support it; we may still test two different 5.3-based builds, w/ and w/o FAST_CSG_USE_SAME_KERNEL, until this graduates properly to on or off).

t-paul:
No worries, I know the busy state quite well too :smile:.

I think we are on 5.3 now everywhere with the official builds, and 5.4 was just released (breaking homebrew builds already... fixed yesterday). I guess we can double check, but I'm pretty sure we could declare 5.3 as baseline.

ochafik:
@t-paul Cool, dropped the multibuild substance out of this PR then :-)

t-paul (approved):

Issue #3920, opened 2021-10-04
MacOS build broken because "poppler 21.01.0" does not compile


ChrisCoxArt:
To be clear for those not used to the MacOS builds, this library is downloaded and built by /scripts/macosx-build-dependencies.sh

"poppler 21.01.0" cannot compile due to missing JPEG library and header references
This appears to be a compile options error, it is not fixed with later library versions.
It appears to be fixed if the options "-DENABLE_LIBOPENJPEG=none -DENABLE_DCTDECODER=none" are removed from macosx-build-dependencies.sh
But since the full MacOS build still does not work, I haven not been able to actually test the compiled poppler library in OpenSCAD.

```
bash-3.2$  ./scripts/macosx-build-dependencies.sh
Detected BigSur (11.x) or later
Building for 10.9 or later
Setting number of CPUs to 24
Using basedir: /Volumes/Development/openscad/../libraries
double_conversion eigen gmp mpfr glew gettext libffi freetype ragel harfbuzz libzip libxml2 fontconfig hidapi libuuid lib3mf glib2 boost poppler pixman cairo cgal qt5 opencsg qscintilla
Building packages: double_conversion eigen gmp mpfr glew gettext libffi freetype ragel harfbuzz libzip libxml2 fontconfig hidapi libuuid lib3mf glib2 boost poppler pixman cairo cgal qt5 opencsg qscintilla 

double_conversion 3.1.5 already installed - not building
eigen 3.3.7 already installed - not building
gmp 6.1.2 already installed - not building
mpfr 4.0.2 already installed - not building
glew 2.1.0 already installed - not building
gettext 0.21 already installed - not building
libffi 3.2.1 already installed - not building
freetype 2.9.1 already installed - not building
ragel 6.10 already installed - not building
harfbuzz 2.3.1 already installed - not building
libzip 1.5.1 already installed - not building
libxml2 2.9.9 already installed - not building
fontconfig 2.13.1 already installed - not building
hidapi 0.11.0 already installed - not building
libuuid 1.6.2 already installed - not building
lib3mf 1.8.1 already installed - not building
glib2 2.56.3 already installed - not building
boost 1.74.0 already installed - not building
Building poppler 21.01.0 ...
-- The C compiler identification is AppleClang 13.0.0.13000029
-- The CXX compiler identification is AppleClang 13.0.0.13000029
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
[...]

PR #3962, opened 2021-10-29
macOS package upgrade: 

kintel:
gmp 6.2.1, libffi 3.4.2


kintel:
Took out poppler for now, see comment in https://github.com/openscad/openscad/issues/3920

t-paul:
Maybe we can integrate the updates/fixes from https://github.com/openscad/openscad/commits/qt5-upgrade-macos here too.

kintel:
Nice! I might separate out the qt stuff and continue on that branch instead. Trying to limit the  scope of my PRs in case I get too busy elsewhere:)

kintel:
Pulled out qt and qscintilla as well; will create separate PRs for those.

kintel:
Slimmed this down it its minimal state. Also uploaded built libraries from CircleCI and update the sha. Not sure if this is the ideal way of doing that these days. Let me know if there is a better process.

t-paul:
No, nothing better at this point. It's not happening too often, so I've not spend much time thinking about a better solution.

t-paul (approved):

PR #3964, opened 2021-10-31
Put back accidentally removed -DCGAL_DEBUG

kintel:
This was originally added in https://github.com/openscad/openscad/pull/3941, but must have gotten lost during refactoring in https://github.com/openscad/openscad/pull/3948


t-paul (approved):
Oops :smiley:.

PR #3966, opened 2021-11-01
Upgrade Qt and QScintilla for macOS

kintel:
Qt: 5.15.2
QScintilla: 2.11.6

Notes:
* Qt 5.15.2 requires macOS >= 10.13 "High Sierra". They actually use system calls introduced on 10.13 with no way of #ifdef'ing out of it, so there is no trivial way of supporting older versions of macOS.
* x86_64 only, but this upgrade does offer a theoretical pathway to arm64 (and universal binary support)


kintel:
Giving it some soaking time to let people in #3775 test this.

t-paul:
Good idea. Unfortunately I'm out at this point as the MacOS version of my old MacBook is too old. But I guess we have to move eventually.

kintel:
If your MacBook Pro is from 2010 or later, you _should_ be able to put 10.13 on it

t-paul:
Hmm, it already feels slower than a Raspi1 as it is, but I guess there's not a huge amount to lose trying to upgrade. If they would sell Mac Minis for at least half reasonable prices, I might be a bit interested. We'll see :grinning:.

kintel:
heh, yeah, slow is expected. I use an 8 year old laptop running 10.13 in a VM. Now, _that_ is slow :)

t-paul:
Looks like the github build (MacPorts) asks for a rebuild too:
```
Error: Current platform "darwin 20" does not match expected platform "darwin 19"
```
I'll trigger that...

kintel:
Which build is that?

t-paul:
`macos-release`
https://github.com/openscad/openscad/runs/4063856881

kintel:
Looks like the GitHub build has other issues; MacPorts' cmake pick up libbz2 and libz from MacPorts instead of the system:

```
% otool -L /Volumes/T7/code/OpenSCAD/ci-cache/GitHub/libraries-gh/lib/libzip.dylib 
/Volumes/T7/code/OpenSCAD/ci-cache/GitHub/libraries-gh/lib/libzip.dylib:
	@rpath/libzip.dylib (compatibility version 5.0.0, current version 5.0.0)
	/opt/local/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
	/opt/local/lib/libbz2.1.0.dylib (compatibility version 1.0.0, current version 1.0.8)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1292.100.5)
```

May be better solved against master.

Ironically, this sounds a bit like the same problem which caused us to manually build libzip in the first place. See similar problem I had with Poppler: https://github.com/openscad/openscad/issues/3920#issuecomment-954904503

kintel:
-> I'll try to pull together a GitHub Actions build without using MacPorts tools and see where that leads..

t-paul:
Ok, so lets merge that state for getting the snapshot builds updated and check the release build separately?

kintel:
Sure, I'm fine merging this branch for now (after disabling the release build job for this branch).
I just wanted to give you some time in case you decided to upgrade your Mac and give this a spin, but happy to just merge for now  :)

t-paul:
I'd say go :rocket:...

t-paul (approved):
/me closes eyes and clicks merge :scream:.

t-paul:
FWIW, no High Sierra for this MacBookPro, it's Mid-2009, but required is "2010 or later" :disappointed:.

PR #3968, opened 2021-11-05
Rewrite macos-release to use Homebrew and Github Actions caching

kintel:
TODO
* [x] Use Homebrew tools to build dependencies, to avoid using MacPorts' cmake which hardcodes the MacPorts library path
* [x] Use Github Actions built-in caching support
* [x] Enable OpenSCAD build
* [x] Validate OpenSCAD build
* [x] (optional) Common workflow for release and snapshot -> postponing to later (not as easy as it seems)


kintel:
@t-paul Btw., is there a reason for running snapshots on CircleCI, but tests and releases on GitHub? ..apart from spreading the load?

PR #3970, opened 2021-11-06
Update macOS icons to Big Sur style

TheJKM:
Now that OpenSCAD looks good again on modern macOS versions, the icon is left making it feel wrongly placed in Big Sur and above. After discussion in #3907, I created new icons for macOS for release and nightly versions.
I added images (plain and in the macOS dock). If you want, I can also contribute the GIMP sources.

Release icon:
Icon1024

Nightly icon:
Icon1024Nightly

Release icon in dock:
Screen Shot 2021-11-06 at 21 49 34

Nightly icon in dock:
Screen Shot 2021-11-06 at 22 20 31


kintel:
Awesome! Gimp sources wouldn't hurt, just in case we need to tweak the icons later.

TheJKM:
Done

kintel:
One thing I noticed: This only updates the app bundle icon, not the application icon; the icon in the dock, and when using Cmd-Tab didn't change.
Do you have an easy way of making your new icon appear in those places as well?

TheJKM:
That's intresting. I'll have a look at it, but it may take some time.

kintel:
There is also the document icon, not sure if that needs a update as well: https://github.com/openscad/openscad/blob/master/icons/SCAD.icns

TheJKM:
Good catch! I didn't notice the difference until now. Will update it as well.
Screen Shot 2021-11-08 at 19 58 42

kintel:
@TheJKM Did you plan to update the remaining icons? I'm doing some work on the macOS build these days, so it would be good to get this committed as part of that effort :)

TheJKM:
It's still on my to do list, however I couldn't find time for it. I hope the holidays will change that.

TheJKM:
The new file type icon for macOS is here. Now I'll look into the issue that the new normal icon is not shown every time, and after that, we should be ready for merging.
SCAD

TheJKM:
Now I also fixed the bug causing the old icon to show while the application was running. MacOS shows the icon embedded in the binary when it is running, and the bundle icon when it's closed. So, when we build for macOS, now we integrate the new icon into the binary.

Because I was really unhappy with my gimp creation of the document icon, I did a rework. I reverted to the old one, which is actually the correct one up until macOS 10.15. For macOS 11 and newer, I set the Info.plist property causing macOS to automatically create the document icon according to the new style.

From my point of view, this merge request is now ready to merge. However, as the snapshot is not building automatically, I could only test my local build.

kintel:
I'll do a sanity check using the snapshot & release builds and get this merged soon. Thanks a lot!

kintel:
I couldn't find any docs for the `CFBundleTypeIconSystemGenerated` setting, but I found a news article from a year ago saying that macOS auto-generates document icons by default? Are you sure this setting actually does anything?

TheJKM:
Yes, macOS generates the document icons automatically by default if the application is running on macOS 11 or newer. This already works for the current release with file formats like 3MF. The problem is, that we target macOS 10.13 and newer, so we have 10.13, 10.14 and 10.15 not supporting automatic document icons. They use the `.icns` file defined in Info.plist. If we would leave it like that, macOS 11 and 12 would also use the `.icns` file. So, we have to tell it that we want the auto-generated icon instead of the `.icns` file, which is what `CFBundleTypeIconSystemGenerated ` does say.

I found this key by creating a clean Mac app in Xcode, configuring automatic document icons with a legacy `.icns` for older systems, exporting the app and reading it's Info.plist.

On my main system, I had problems with caching, so the automatic icon didn't show up. I tested the fresh build on a system which never saw OpenSCAD before running Monterey 12.1, and it worked as it should.

TheJKM:
Another small update, for some weird reason I added a white drop shadow to the icon. Now it's black as it should be.

kintel:
After testing this on macOS 12 on M1, I noticed that `icons/openscad-nightly-macos.png` and `icons/openscad-macos.png` are not actually being used since these files are not included in `openscad.qrc`.

On this version of macOS (I don't know about others), it falls back to using the icon specified in `Info.plist` if the file isn't found in the qrc archive.

TheJKM:
I include them in src/openscad.cc, they should be included into the binary - or not? My Qt knowledge mainly comes from PyQt, sorry.

kintel (approved):

kintel:
Thanks, I think this looks great now!

Issue #3142, opened 2019-11-20
Backwards compatibility with experimental or optional features


thehans:
This is a continuation of some discussion from issue #350 regarding how to handle backwards compatibility as new/experimental features are added.

Today we have 2 experimental features in the nightlies "lazy-union" and "function-literals".  We manage these through "Experimental" option checkboxes on nightly releases which are toggled by the user in GUI preferences(Qt settings), and default to disabled. They can also be enabled per command line call with 
```--enable=feature-name``` (how our tests work).  

This is useful so that users can opt-in to testing experimental code, but there are some aspects to compatibility that this scheme doesn't address.

1) User A shares a script which they developed while making use of experimental features.  User B downloads script and but is using an older release or has not specifically enabled the same feature, potentially causing errors in parsing or geometry.
2) User A wants to develop a script using a new feature, but also depends on an old library from User B which is not aware of this feature.  On the one hand, if the feature is a new builtin module, this wouldn't be a problem since the if the old library just doesn't use it.  But in the example of "lazy union", its not really an addition to the parser, but a change in overall behavior which could break this older library in insidious ways.

So the main concern is that, as more features get added, it becomes more difficult to share and re-use older code, and to even be able to tell what features or OpenSCAD version any given script relies on.

Here is one suggestion which I think would help in both of these cases:
_Originally posted by @nrdvana in https://github.com/openscad/openscad/issues/350#issuecomment-554807479_
> Best way to solve back-compat is with a opt-in directive that affects *only* the current source file.  To borrow design from Perl,
> ```
> use feature 'lazy_union';
> ```
> 
> or since 'feature' already means something in this codebase that can't be toggled on the fly (or can it?)
> ```
> use options implicit_union=false;
> ```
> 
> Then if people decide it should be the default:
> ```
> use 2020.xx;  // opt-in to default options as of version 2020.xx
> ```
> 
> 

So, instead of having the user globally enable "lazy union" in their preferences, they would need to specify per scad file.  This would make scripts self-documenting about the feature set they require.  
So when user B tries to run this on an older release, it would at least "fail fast" by giving a parser error, which would be more obvious to troubleshoot than some weird geometry difference.

Looking further ahead, lets say we do implement such a feature-enabling scheme.   Then later another release adds an experimental feature called "magic".  The releases before "magic" could at least recognize `use feature` (or whatever final syntax we decide on), and print an even more useful error message than a parser syntax error: 
[...]

Issue #462, opened 2013-08-20
Sphere coordinates should be fixed


MarkJeronimus:
A common [CG sphere](https://www.google.com/search?tbm=isch&q=wireframe+sphere) has coordinates at integral positions along the longitude and latitude axes, but OpenSCAD spheres have vertices on the latitude axis on half positions instead.

To elaborate, if $fn=12, the sphere has longitude vertices at 0°, 30°, 60°, etc. but on the latitude, the vertices are at 15°, 45°, 75°, etc. creating an unwanted <i>flat</i> top and bottom.

This creates scenarios where geometry doesn't match with other geometry, or geometry ends up too thin. The full discussion with examples is on http://forum.openscad.org/Fix-the-sphere-with-vertex-on-top-td5332.html

Temporary workaround (rotation of a half-circle):

<pre><code>module my_sphere() {
    rotate_extrude()
    intersection() {
        circle();
        translate([0, -1]) square(2);
    }
}</code></pre>

(note that a rotation of a full circle causes a void object when rendering F6).
##

thehans:
I thought I would post my version of this sphere implementation.  

By generating list of points for a semicircle and rotate_extrude'ing that, you save a boolean operation.  And you also don't have to throw away half of the circle points you just computed.

This uses the arc, and fragments functions that I orignally developed for my "[Functional OpenSCAD](https://github.com/thehans/FunctionalOpenSCAD)" library which operates directly on vertex data.  

```
module my_sphere(r=1, d) { 
  let(R = d == undef ? r : d/2)
  rotate_extrude()
    polygon(arc(r=R, angle=180, offsetAngle=-90));
}

// Draw a circular arc with center c, radius r, etc.
// "center" parameter centers the sweep of the arc about the offsetAngle (half to each side of it)
// "internal" parameter enables polyhole radius correction
function arc(r=1, angle=360, offsetAngle=0, c=[0,0], center=false, internal=false) = 
  let (
    fragments = ceil((abs(angle) / 360) * fragments(r,$fn)),
    step = angle / fragments,
    a = offsetAngle-(center ? angle/2 : 0),
    R = internal ? r / cos (180 / fragments) : r,
    last = (abs(angle) == 360 ? 1 : 0)
  )
  [ for (i = [0:fragments-last] ) let(a2=i*step+a) c+R*[cos(a2), sin(a2)] ];

function fragments(r=1) = ($fn > 0) ? 
  ($fn >= 3 ? $fn : 3) : 
[...]

PR #3973, opened 2021-11-10
change for  $fa <=120

UBaer21:


thehans:
This change breaks a large number of tests (52 of 1423 on Ubuntu 20.04).  
The test results are uploaded as artifacts, which are linked at the bottom of the build "Summary" page, e.g. here for ubuntu: https://github.com/openscad/openscad/actions/runs/1445300625
If you download and unzip one of those artifacts you can view the html report page which shows the expected vs actual results images side by side for all failed tests.

All tests need to be passing before this can be considered for merging.  So a number of tests would need their expected images updated and/or the test scripts updated to change the 

The problem with this change as I see it, is the way fragments are calculated has `$fa` tied to `$fs`.   It calculates the limit imposed by each and then takes whichever would result in less fragments.  This means in probably 99% of the cases where this limit comes up, its going to be from `r` or `d` being too small relative to `$fs` and not because the user is trying to set a very large `$fa` (until now I've never heard anyone wanting something *larger* than the default `$fa=15`).

I suspect the original reasoning for this limit of 5 instead of 3 when using `$fs` and `$fa` is that it results in something that is "reasonably" similar to a circle, at the worst limits.
In the case where a specific number of sides is needed, the user should specify `$fn` instead.

Below is a short analysis of two metrics comparing regular polygons of n sides to an idealized circle.
1) `r_pct` represents the ratio of the effective inradius, to the provided radius (circumradius).
   Inradius is especially relevant for circular holes, which determines the largest circle that can fit *inside* the 
2) `A_pct` represents the ratio of the regular polygon area to the ideal circle area.
```
n = 3,  r_pct = 50%,      A_pct = 41.3497%
n = 4,  r_pct = 70.7107%, A_pct = 63.662%
n = 5,  r_pct = 80.9017%, A_pct = 75.6827%
n = 6,  r_pct = 86.6025%, A_pct = 82.6993%
n = 7,  r_pct = 90.0969%, A_pct = 87.1026%
n = 8,  r_pct = 92.388%,  A_pct = 90.0316%
n = 9,  r_pct = 93.9693%, A_pct = 92.0725%
n = 10, r_pct = 95.1057%, A_pct = 93.5489%
n = 11, r_pct = 95.9493%, A_pct = 94.6502%
n = 12, r_pct = 96.5926%, A_pct = 95.493%
```
At `n = 3`, inradius is only 50% of the circumradius, and the area is even less at 41%.
At `n = 5`, inradius is 80.9% of circumradius, and area is 75.7% which seems at least somewhat more reasonable.

Additionally spheres can end up as triangular or rectangular prisms in many cases of the test results, which is even less recognizable as a sphere than the current five sided(longitude) polyhedron.

UBaer21:
can we set `$fs=1;` ? (or .5)  as we already limit fn to 30 with `$fa=12`  - there is no need for $fs=2 which also too coarse for 3D-printing and make oSCAD look unprofessional.

nophead:
I can't see the point of this change. To get polygons with defined numbers of sides I set $fn. I set $fa to 6 and $fs to half my extrusion width to get 3D printable smooth curves and OpenSCAD chooses the number of sides.

UBaer21:
i am using $fa to get a fix number of fragments in rotate_extrude with angle<360 (fa=deg/fn), and sure there are workarounds but the question is why should we have a limitation that exist only to fix something wrong on a different location. Mathematically $fa can have 120° so this should be the proper value. The problem with "let OpenSCAD choose" is that e.g 
```
$fs=2;
$fa=12;
difference(){
  circle(r=2.1);  
  circle(r=1.9);
}
```
a mismatch of fragments can cause walls with changing width so slicing thin walls causes breaks. 
 But i totally agree that the default $fs should be <1;
And yes $fa>72 is rare but i can't see why this should not be possible.

nophead:
Why don't you just use $fn = 3 instead of $fa = 120?

UBaer21:
i could use $fn=ceil((360/grad)*fn)  -  when e.g.  rotate_extrude(angle=180)  using $fn=3 will not give you 3 fragments in that case not even a 3D object.  And lets not talk about that everything jumps 180°  when angle 359 gets 360

nophead:
The way I understand it is if you don't specify $fn, OpenSCAd computes it from $fa and $fs. If you don't like that you set $fn and then $fa and $fs are ignored.  I have this function my library that does the same thing: 
```
function r2sides(r) = $fn ? $fn : ceil(max(min(360/ $fa, r * 2 * PI / $fs), 5));    //! Replicates the OpenSCAD logic to calculate the number of sides from the radius
```

Why do you need to specify fn indirectly with $fa?

UBaer21:
i made a module that way a long time ago and recently wondered why i got 4 segments when hitting 288° as there is no mathematical sound reason for it.  This is not "i have a problem and need a solution"  this is " i love open scad because it is so pure and mathematical and here is a flaw that just should not exist.  Sure if you tell people how they have to use it - you also can use CAD Software with fix workflows. I think this just show that $fs should not be 2 so we don't need to cover for it by limiting $fa to 72°. I already thought this PR could fail and assume that $fs=.5 as PR would also fail because everything would be more round.  It is terrible that the first thing you have to do in a new file is to define fragments anyhow.

t-paul:
I don't see why we would need to throw out everything out of principle. The PR came out of the IRC discussion observing that slightly unexpected `$fa` limitiation assuming this is a simple, trivial fix. Turns out that's not the case, so we can't just use this solution. That does not mean we need to stomp out the whole discussion on that issue.

The defaults are a different thing, they are probably that low as OpenSCAD is quite old by now and when it started higher values could have been an issue. That's a totally separate discussion, e.g. people have asked to introduce a configuration option so it's possible to match the environment.

jordanbrown0:
This change seems fine to me.  Mathematically, if you set $fa to anything greater than 72 you should get a square or a triangle, and if you set it to anything greater than 90 you should get a triangle.  If you set it to anything greater than 120 the result should be undefined and probably limited to 3 sides.

Similarly, if your $fs is set to more than 2πr/4 you should get a triangle, and if it's set to more than 2πr/5 you should get a square or a triangle.  If it's set to more than 2πr/3 then again the result should be undefined and probably limited to 3 sides.

$fa and $fs are not the right way to control the number of fragments in a circle, but there's no reason for them to have an arbitrary limit.  They should follow the math.  Explaining them is hard enough without introducing this limit.

And yes, very small objects will be silly.  But at least they will be silly in a way that follows the definitions.

---

It does seem like this discussion touches on a different issue:  the interaction between the `angle` parameter to `rotate_extrude` and the `$f*` variables is unclear and does not allow for variations that you might want.  If for whatever reason you want a 75° extrusion in one fragment... you can't have it, because that's a 4.8-sided figure.  Nor can you have it in 2, 3, or 4 segments (9.6, 14.4, or 19.2-sided).

Best might be to introduce another variable, call it `$fna`, that lets you explicitly set the angle for each fragment.

thehans:
In general I'm all for making OpenSCAD more mathematically consistent, but in this case the proposed fix would potentially alter behavior for a **large** portion of existing scripts, to satisfy a very specific niche case which I don't think has ever come up before and may not actually come up again for anyone besides UBaer21.

Having the most fundamentally simple example of:
```
sphere();
```
Change from this:
sphere_default
to this:
sphere_120
is not something to be taken lightly.

Granted the first pic is still a poor approximation of a sphere, but at least it's more recognizable as such compared to a elongated cube rotated by 45 degrees.

Somewhat related, I would also like to see sphere eventually changed to have a vertex at the poles (see #462), but this has been avoided so far due to the same backwards compatibility issues.
The backwards compatibility discussion has been going on for a long time, and Its a very contentious topic.   There is issue #3142 for discussion on how to handle such issues, but no consensus seems to be reachable.
 - Newer users want to know why can't we just change behavior across the board so that everything makes perfect logical/mathematical sense.
 - And older users (or even just newer users which stumble over someone else's old scad script) want to know why openscad keeps breaking their existing designs which they've already perfected to work with previous versions.  This absolutely happens even for much less-wide-sweeping changes to behavior, which are often assumed at the time to affect hardly anyone.

UBaer21:
so if i understood this right then these old scrips using the default definitions and is controlled by $fs- if we set `$fs=1.04720  `everything should look the same - so we could allow $fa to be up to 120. I think $fa shouldn't be used to limit something caused by $fs in the first place.

nophead:
The normal use of $fa and $fs is to set set $fa small to get smooth curves
and $fs to something reasonable, so that tiny radius curves don't have
silly numbers of sides that break 3D printing. I have never seen $fa have a
value more than 6 and $fs more than 0.5. $fn is when a specific number of
sides is needed.

UBaer21:
exactly but at the moment oscad is using $fa to limit objects to 5 sides
with $fa max. 72.
while $fs should be used for that.

nophead:
Setting $fs gives you a number of sides that depends on the radius until it gets big enough for $fa to limit it.

jordanbrown0:
If you want a particular number of sides, do not use $fa or $fs.  That's not what they are for.  They are for controlling the accuracy of the **approximation** used for creating circles, and although they have mathematical definitions they should be viewed only as hints.

jordanbrown0:
@thehans yeah, I suppose.  If we were starting from zero I'd push for the mathematically correct answer, but the compatibility answer is pretty compelling.

BTW, I don't fall well into your two camps.  I'm not new (~4y), but at the same time I would generally rather take a modest compatibility hit to get long-term improvements in consistency and simplicity.  I hate it when a product is so mired in compatibility concerns that it cannot improve cleanly.

thehans:
Yeah, it was perhaps an over-generalization that not all users fit into, but I was just meaning to summarize the two sides of the issue.  Replace with "some users", and "other users" to your taste ;-)

PR #3976, opened 2021-11-13
Build macOS snapshot on GitHub Actions

kintel:
This is currently mostly for completeness of build configs and to test that CircleCI and Github build are equivalent.

### DIFF
--- a/.github/workflows/macos-snapshot.yml
+++ b/.github/workflows/macos-snapshot.yml
@@ -1,0 +1,69 @@
+name: macos-snapshot
+
+on:
+  push:
+    branches:
+      - "*-macos"
+  schedule:
+   - cron: "30 6 * * 0,1,3,5,6"
+
+jobs:
+  build:
+    runs-on: macos-latest
+    env:
+      OPENSCAD_LIBRARIES: /Users/runner/work/openscad/libraries/install/
+      LIBRARIES_CACHE: libraries.tar.gz
+    steps:
+    - name: Checkout OpenSCAD
+      uses: actions/checkout@v2
+      with:
+        submodules: 'recursive'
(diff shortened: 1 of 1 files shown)


t-paul:
Cool, good to have multiple options. Right now CircleCI has the benefit of the simple automatic download for the website, I have not managed to get that working for github without the need to grant extensive access rights to some access token.

kintel:
Thanks, good to know.
Btw., how are snapshots published these days? I see the macOS snapshot hasn't updated for a couple of weeks..

t-paul:
This is supposed to run via cron:
```
#8 8 * * * /home/openscad/bin/fetch-snapshots.sh
```
I have that disabled currently due to the build issue with the Windows snapshots after the MXE update as I don't want to publish those. There's an issue with the lib3mf, using 1.8.1 and dropping the patches broke the static build.

t-paul:
Ok, I changed the `./update-snapshots.sh` generating the files used by the download webpage to skip the Windows builds for now. So macOS and AppImage should be back to updating automatically.

kintel:
Thx, found the scripts - it all makes sense now :)

kintel:
Looks like artifact access is a GitHub limitation that a lot of people struggle with: https://github.com/actions/upload-artifact/issues/51

I'll merge this and disable the build for now, and if this materially changes in the future we can revisit it.

t-paul:
The workaround would probably be adding a release action similar to what https://github.com/openscad/openscad-documentation is doing. But maybe it's not a bad idea to have more than just one options. Now with Travis requiring credit card validation after the `.com` move, it's really just down to 2 services (unless you have some way to get the validation through to reactivate Travis).

PR #3977, opened 2021-11-13
Rewrite CircleCI macOS snapshot to use Homebrew and CircleCI caching

kintel:
Similar to #3973 but for the CircleCI snapshot build:
* Use Homebrew tools to build dependencies, to avoid using MacPorts' cmake which hardcodes the MacPorts library path
* Use CircleCI built-in caching support
* Validate OpenSCAD binary

I've also validated that the 3 different macOS builds (CircleCI snapshots, GitHub snapshots, CircleCI release) have exactly  the same dependency chain of all direct and implicit library dependencies (i.e. all libraries also have exactly the same dependency chain).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,7 +103,7 @@ jobs:
           destination: 64-bit
   openscad-macos:
     macos:
-      xcode: 12.3.0
+      xcode: 12.5.1
     steps:
       - checkout
       - run: git submodule update --init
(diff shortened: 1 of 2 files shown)


kintel:
@t-paul I believe this means that there should no longer be any need to update shasums for library and macports caches, as the built-in caches will be used. I'm still packaging the libraries as artifacts, for easy download and inspection if needed.

t-paul:
Very nice! Yeah, if the caches are internal and not coming from our server anymore, then there should be no need for the checks. Does it manage to rebuild the whole cache in one run? One of the issues I had before was that a full rebuild was hitting CI timeouts.

kintel:
Yeah, looks like a full rebuild take 1h40m: https://app.circleci.com/pipelines/github/openscad/openscad/2040/workflows/1b578ef7-4770-4843-b872-7f7fc2f2f0f9/jobs/9555

t-paul:
Cool. I'm not sure about the actual timeout, but it seems to be at least > 2h. Github-CI is so much more relaxed with the 6h I've read somewhere.

PR #3979, opened 2021-11-18
Fix deprecated use of <mimetypes> in appdata

knielsen:
Use of &lt;mimetypes&gt; is deprecated in appdata [1]. Replace it by the now
recommended &lt;provides&gt; => &lt;mediatype&gt;.

Motivated by a warning shown for package openscad  in the Debian infrastructure [2].

[1] https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-mimetypes
[2] https://appstream.debian.org/sid/main/issues/openscad.html


t-paul (approved):
Thanks!

PR #3991, opened 2021-12-04
Add timeout option for macos dependency build.

t-paul:
With this change we can incrementally build the cache by just running the build multiple times. The always stored cache should also ensure we are not completely losing the cache after 15 days.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -124,7 +124,7 @@ jobs:
       - restore_cache:
           keys:
             # CircleCI library caching. If we ever need invalidate the cache (e.g. to remove any old files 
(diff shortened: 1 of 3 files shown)


kintel:
Nice! I initially thought the cache TTL was based on _usage_, not since creation. This should work quite nicely.

kintel (approved):

t-paul:
Yeah, I hope so. Build time is cutting it a bit close with the Qt5 build alone taking about 56 minutes. I did not find anything else to disable, but it's still building some stuff like designers that are not needed. But lets see if there's some response from CircleCI next week...

PR #4000, opened 2021-12-09
Use built-in libz on macOS

kintel:
Now that the build use Homebrew instead of MacPorts, we can remove the workaround for the bad MacPorts CMake build

### DIFF
--- a/scripts/macosx-build-dependencies.sh
+++ b/scripts/macosx-build-dependencies.sh
@@ -11,8 +11,7 @@
 #  -f   Force build even if package is installed
 #  -v   Verbose
 #
-# Prerequisites:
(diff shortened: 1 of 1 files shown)


kintel:
@t-paul Any idea why the Windows build for this, and other PRs, fail?

t-paul:
No, I'm a bit puzzled by that. It seems to crash after successful build when trying to upload the test result data. The error does not indicate what the problem is. Maybe it's hitting some sort of size limitation? One option could be to restrict to the actual log file and see if that helps for now.

kintel:
Yeah, looks like it uploads all output files, even if tests were successful. I'll try a few things.

PR #4005, opened 2021-12-10
Support M1 build of dependencies

### LINKED ISSUE #3928: MacOS build fails on M1 (ARM64) based machines because of GMP compilation errors
The version downloaded is GMP 6.1.2, which works fine on Intel processor Macintoshes.
But GMP 6.2.1 is the first version to support ARM64 on MacOS.

Also, the dependency script forces the host to --host=core2-apple-darwin13.0.0
And that will only work on Intel x86 processors, not the ARM64 based M1.

But I suspect many changes will be needed to make an M1 native build of OpenSCAD.
I'll try some fixes to at least get it building.

kintel:
Fixes #3928

This makes `macosx-build-dependencies.sh` support building for the local architecture on either x86_64 or arm64.
Cross compilation not yet supported.


ChrisCoxArt:
With a clean copy of the source and download of dependencies, mpfr does not build.
```
Building mpfr 4.0.2 ...
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for arm64-apple-darwin17.0.0-strip... no
checking for strip... strip
configure: WARNING: using cross tools not prefixed with host triplet
checking for a thread-safe mkdir -p... ./install-sh -c -d
checking for gawk... no
checking for mawk... no
checking for nawk... no
checking for awk... awk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking whether to enable maintainer-specific portions of Makefiles... yes
checking build system type... Invalid configuration `arm64-apple-darwin': machine `arm64-apple' not recognized
configure: error: /bin/sh ./config.sub arm64-apple-darwin failed
```

ChrisCoxArt:
Building libzip fails because it uses wget, which is not installed by default.
Also, why is this not using the libzip that ships with MacOS?
```
Building libzip 1.8.0...
./scripts/macosx-build-dependencies.sh: line 444: wget: command not found
```

ChrisCoxArt:
build error still exists with mpfr 4.1.0 - so most likely it is something in the configure line that it doesn't like.

kintel:
As for the libzip question: I don't think macOS ships with libzip. Let me know if you've found otherwise.
_libz_ otoh does come with macOS, which was recently removed from the build script now that we use a better cmake build.

kintel:
Thanks for giving this a spin btw., really helpful to get my assumptions and local system config second guessed!

kintel:
You may have an old autoconf installed. Here's mine:
```
$ autoconf --version
autoconf (GNU Autoconf) 2.71
```

ChrisCoxArt:
> You may have an old autoconf installed. Here's mine:
> 
> ```
> $ autoconf --version
> autoconf (GNU Autoconf) 2.71
> ```
same version here
```
autoconf --version
autoconf (GNU Autoconf) 2.71
```

kintel:
I can reproduce here. Will fix in a moment

ChrisCoxArt:
> As for the libzip question: I don't think macOS ships with libzip. Let me know if you've found otherwise.

Hmmm, I seem to have the library on all my machines. But that may also mean that something I use commonly installs it.
But it's not worth installing a clean OS to debug.

kintel:
Do you actually have libzip in `/usr/lib/` ?

ChrisCoxArt:
> Do you actually have libzip in `/usr/lib/` ?

Not in that directory.  I'll solve that mystery later.  Probably just means I have too many apps that use open source frameworks...

ChrisCoxArt:
mpfr configure doesn't seem to like the last 2 arguments "--build=$ARCH-apple-darwin --host=$ARCH-apple-darwin17.0.0 "
It builds without those arguments.

kintel:
My patch should fix those flags.
If they aren't actually needed, we can remove them. Let me know how your build goes.

ChrisCoxArt:
Dependencies build worked, now building the app.

kintel:
Yah, documented wget as a dependency in the script, together with automake, libtool, cmake and pkg-config

Issue #912, opened 2014-08-19
Vector swizzling


davidson16807:
[Swizzling](http://en.wikipedia.org/wiki/Swizzling_%28computer_graphics%29) is a useful concept you can find built in to computer graphics languages like [GLSL](http://en.wikipedia.org/wiki/OpenGL_Shading_Language). There's been more than one occasion where I would have liked to use it. It can be very useful for converting 3d vectors to 2d and rearranging elements of a vector.

To pose an example, the following statements would produce equivalent behavior:

```
foo3d = [1,2,3];

foo2d = [foo3d.x, foo3d.y];  // without swizzling
foo2d = foo3d.xy;            // with swizzling

bar = [foo3d.z, foo3d.y, foo3d.x];
bar = foo3d.zyx;

baz = [foo3d.x, foo3d.x, foo3d.x, foo3d.x];
baz = foo3d.xxxx;
```

This easily handles conversion to lower dimensionality, though it wouldn't support all scenarios when converting to higher dimensionality. The latter is particularly important because it allows you to convert vectors to a form that is compatible with 4d matrices. As an optional extension to this proposal, I suggest a method be found to convert to higher dimensionality, perhaps borrowing GLSL's vec*() functions:

```
baz = [foo3d.x, foo3d.y, foo3d.z, 1];
baz = vec4(foo3d, 1);
```

or Python's asterisk notation:

```
baz = [foo3d.x, foo3d.y, foo3d.z, 1];
baz = [*foo3d, 1];
```
##

TLC123:
I keep a bunch of swizzling  functions at hand. Like so:


function xyy(v)=[v.x,.v.y,v.y];
function zyx(v)=[v.z,.v.y,v.xy];
function xyzw(v)=[v.x,.v.y,v.y,v[3]];
function w(v)=v[3];
function rgg(v)=[v.x,.v.y,v.y];

t-paul:
Closed via #4010.

PR #4009, opened 2021-12-13
Vector swizzling feature

sublim9:
Gives the ability to swizzle vectors like in GLSL. Also, names the fourth, unnamed vector element as 'w'.
Usage:
`myvec = [10,20,30,40];`
`echo(myvec.w); // outputs 40`
`echo(myvec.xy); // outputs [10,20]`
`echo(myvec.zyx); // outputs [30,20,10]`
`echo(myvec.xxyy); // outputs [10,10,20,20]`
`echo(myvec.wxwxwxwx); // outputs [40,10,40,10,40,10,40,10]`


t-paul:
I already have 2 suggestions :-).

* Maybe we could partially follow https://www.khronos.org/opengl/wiki/Data_Type_(GLSL)#Swizzling and and allow `rgba` too. I guess the texture version does not make much sense at this point.
* It would be great to have a test case too, just testing the examples you gave above.

jordanbrown0:
I wouldn't do this.  We're moving toward "foo.bar" being an object-element-access mechanism.  The existing foo.x, foo.y, foo.z are not that, because they operate on arrays rather than objects, but they look a lot like it.

`[myvec.y, myvec.x]` does not seem like an onerous way to do what this proposal does as `myvec.yx`.  It seems more obvious, and doesn't introduce magic interpretation of the name after the dot.  More complex variants would of course be longer, but more complex variants are also much less likely in real life.

Note that
```
function yx(a) [ a.y, a.x ]
```
is a simple way to get a similar level of terseness, for any cases that are common enough to warrant doing something.

t-paul:
I don't see an issue with that. Swizzling is just an extension of the current .x/y/z lookups which only work for vectors. There's no collision with objects due to the type the lookup is executed on. It's a long standing feature request via #912 which indicates people already used this via functions in a not so concise way. And it's following known behavior of GLSL.

jordanbrown0:
It's not a direct conflict, no, but I think that from a language design perspective we'd be better off regarding the existing .x/.y/.z mechanism as a legacy mechanism than to introduce more magic in that area.

Other languages may have had this construct, but that can be said for almost any construct - there have been all sorts of weirdness in various languages.  Tellingly, I don't think you'll find this construct in any major language.

Look at it this way:  take somebody experienced with any major language, and show them `a.yx`.  How many will figure out what it means without having to look at the documentation?

("Major language" is of course a gray distinction, but I'd draw the line at top-10 or top-20.  Put differently, what's the most popular language with this construct?  I'm probably familiar with a broader spectrum of languages than most, and I had to look up what GLSL is.)

sublim9:
I accidentally based this PR on my data-render code instead of the current master so I closed #4009 and redid the PR correctly as #4010. We should continue talking there. 

jordanbrown0, I don't know, you may be right. I'm just looking at it from a productivity point of view. I'm trying to find ways to reduce the amount of typing I do in a scad file and increase it's readability at the same time. When someone learns and uses the .yx mechanism, I find the scad code produced is much more readable.

I'm not very familiar with the vision the openscad community may have about the future of openscad's type system. Thus, I am a bit confused. From my perspective I fail to see any issues going forward. Are future vectors going to be ObjectType objects instead of VectorType? In that case there is seemingly a clash.  But even then, couldn't we have reserved member names for exposed intrinsic members like 'x', 'y', 'z' for certain (or all) object types? Even if we don't integrate the proposed change, will there not still be a clash with the existing mechanism for accessing vector 'x', 'y' and 'z' elements? Will element accessing syntax for vectors, in the future, differentiate itself from the object member access syntax? If not I see no harm in supporting the proposed changes. It's just an extension on what already is. 

Pesonally, I find the syntax intuitive as I discovered it playing around with GLSL code and did not have to read about it. I expected it to work and to my delight, it did.

If there are no technical reason and if it's a taste vs productivity thing, I'm for productivity. I don't see this change being "ugly" enough to disregard. 

But that's just me. It's now up to you.

jordanbrown0:
There's no technical conflict.  I don't know if anybody truly has a long-term roadmap, but I don't see any conflicting future meaning for <vector>.<name>.

My concern is in usability, which isn't quite the same thing as productivity.  Sure, it's possible to learn this syntax, but pretty much nobody will understand it immediately.  It's a bit of unnecessary syntactic sugar that slightly reduces typing in somewhat obscure cases - and approximately that same savings could be achieved using trivial functions, without any new syntax.  (a.yx is four characters; yx(a) is five.)

Generally:  new syntax is Bad, and should be added only when there's a compelling reason for it.

PR #4010, opened 2021-12-13
Vector swizzle feature.

sublim9:
Enables vector swizzling like in GLSL. Names the unnamed fourth vector member as 'w' and adds the 'r', 'g', 'b', 'a' nomenclature. Swizzling results are limited to 4 elements as in GLSL.
Usage:

```
myvec=[10,20,30,40];

echo(myvec.xy); // output: [10,20]
echo(myvec.yx); // output: [20,10]
echo(myvec.wxwx); // output: [40,10,40,10]
```


src/expr.cc
```diff
@@ -371,12 +372,31 @@ MemberLookup::MemberLookup(Expression *expr, const std::string &member, const Lo
...
+          VectorType ret(context->session());
+          for(const char& ch : this->member)
+            switch(ch)
+            {
+              case 'r': case 'x': ret.emplace_back(v[0]); break;
+              case 'g': case 'y': ret.emplace_back(v[1]); break;
+              case 'b': case 'z': ret.emplace_back(v[2]); break;
+              case 'a': case 'w': ret.emplace_back(v[3]); break;
+            }
+          return ret;
```
t-paul, line 390: This triggers a problem with the Ubuntu 18.04. I assume that's because the older boost version does not support the automatic type conversion. This should fix the build issue:
```suggestion
          return Value(std::move(ret));
```

t-paul (changes requested):

thehans:
Continuing the conversation from the previous, closed PR...

> It's not a direct conflict, no, but I think that from a language design perspective we'd be better off regarding the existing .x/.y/.z mechanism as a legacy mechanism than to introduce more magic in that area.
> 
> Other languages may have had this construct, but that can be said for almost any construct - there have been all sorts of weirdness in various languages.  Tellingly, I don't think you'll find this construct in any major language.
> 
> Look at it this way:  take somebody experienced with any major language, and show them `a.yx`.  How many will figure out what it means without having to look at the documentation?
> 
> ("Major language" is of course a gray distinction, but I'd draw the line at top-10 or top-20.  Put differently, what's the most popular language with this construct?  I'm probably familiar with a broader spectrum of languages than most, and I had to look up what GLSL is.)
> 
_Originally posted by @jordanbrown0 in https://github.com/openscad/openscad/issues/4009#issuecomment-992979724_

Personally I think swizzling is a welcome addition to the language.  It's apples and oranges to compare OpenSCAD with "major" (inherently general purpose) languages, because it is in fact the opposite of that; its a Domain Specific Language(DSL).  

GLSL is also a DSL, and since it is designed to run GPU shader programs, the "domain" in question is graphics/geometry, much like OpenSCAD (i.e. more "apples to apples").
I have personally used GLSL; OpenSCAD itself relies on it, as does every other modern OpenGL-supported application or game.  Even in cases of using DirectX rather than OpenGL, their shader language (HLSL) also does swizzling.
Swizzling is a nice convenience in that context.

Also, from a performance perspective, builtin swizzling should be capable of much better performance than any OpenSCAD (userspace) function-based approach.  Because OpenSCAD expression evaluation actually spends a significant portion of its time doing function/variable name lookups, and every additional operator involved in an expression adds overhead.  Swizzling would reduce to a single "operation" with one vector name lookup, staying in a tight c++ loop, as opposed to doing a function lookup plus 3 or 4 individual (although redundant since they would all be the same) vector name lookups, plus separate array indexing operation for each.

thehans:
>...
> It's a bit of unnecessary syntactic sugar that slightly reduces typing in somewhat obscure cases - and approximately that same savings could be achieved using trivial functions, without any new syntax.  (a.yx is four characters; yx(a) is five.)
>...
_Originally posted by @jordanbrown0 in https://github.com/openscad/openscad/issues/4009#issuecomment-993852333_

This ignores the definition of the function, and if you want to support all combinations of `x,y,z,w` for 2 to 4 elements, then you'll need to define `4^2 + 4^3 + 4^4 = 336` individual functions.  Double that number to add support for `r,g,b,a`.  That's assuming you don't allow mixing between `r,g,b,a` and `x,y,z,w` component names, which would require `8^2 + 8^3 + 8^4 = 4672` function definitions.

t-paul:
@thehans agreed. Yes it's syntactic sugar but in my view only extending the one we already have and in a way that follows a reasonable and well-known example. At this point we follow pretty closely the GLSL restrictions (maximum of 4 combinations, no mixing between sets, although no stpq as we don't really have much of texturing yet).

I think we are pretty much set, I'm just considering 2 minor questions
* should we make it an experimental feature? (I tend to go with NO here)
* should we give a warning for `[1, 2, 4].a`, so accessing out of bounds?

jordanbrown0:
Yes, OpenSCAD is domain-specific.  That doesn't mean that commonality with common languages isn't important.  Consider:  how often do people ask "how do I do this in Python", versus "how do I do this in GLSL"?  The fact that OpenSCAD and other OpenGL programs depend on it is immaterial; it's an internal implementation detail invisible to the users.

Yes, of course there's a performance advantage to a built-in mechanism.  How many OpenSCAD models spend a significant amount of their time on the expression evaluation?  The vast majority of models have their processing time dominated by the rendering phases.

Yes, I didn't count the cost of typing the function.  If you're doing this enough to make it interesting, the cost of the function disappears.

I didn't mean that there should be built-in functions, but that anybody who cared could easily define one.  That also means that nobody has to define the hundreds of combinations, only the ones they actually need.  (And especially not the nearly-useless color variants and the completely useless color+coordinate variants.)

I put this request in the same category as the request to have `translate()` be `t()` to save keystrokes, only far less generally applicable.  Saving keystrokes should almost never be a syntactic goal.  The primary syntactic goal should be reducing *complexity*.

But it seems that I'm outvoted, so I'll concede.

nophead:
As a matter of interest, why do people swizzle vectors? I sometimes drop the last one to go from 3D to 2D or from homogeneous to 3D but just as often add a third or a fourth element to go the other way, which isn't a swizzle. I don't think I ever need to reorder them.

t-paul:
It's syntactic sugar, so by definition there's never going to be a *need* for it other than maybe (hopefully) making expression more concise or readable. That also may not be the case for everyone. Java has no custom operator overloading, so you never can write `o1 + o2` for any of your own classes. Lots of languages decided, that's useful enough to have, even though it would be bad if someone implements the `*` operator for subtracting values.

Just looking at some examples from [OpenFlexture](https://gitlab.com/openflexure/openflexure-microscope/), there seems to be a number of cases where it could make expressions quite a bit more readable.
```
cutout_size = [motor_connector_size().y+2, motor_connector_size().x+2, 2*h];

translate([oms.x, objective_mount_y() - 1.2 - 2.5, oms.z])

translate([housing.x/2-3, housing.y/2-3, 0])
```
The "add zeros" is a good point though, that seems to be needed in a large number of cases. Covering that might make it much more useful, even when staying from the GLSL template where they use constructors to do that.

Just for sake of discussion, lets imagine we have also the option to give a `0` (there's maybe better options) to force that place to 0.
```
cutout_size = motor_connector_size().yx0 + [2, 2, 2*h]; // I like that

translate(oms.x0z + [0, objective_mount_y() - 1.2 - 2.5, 0]) // meh?

translate(housing.xy0 / 2 - [3, 3, 0]) // ok-ish
```

nophead:
If you allow zero then I think you need 1 for promoting to homogeneous coordinates.

nophead:
Also the third example is simply: `translate(housing -[3, 3])`. There is no need to supply a Z value of 0.

t-paul:
I'm not sure what you mean. The maximum number of components is currently set to 4, so this would be covered too:
```
p = [1, 2, 3];
echo(p.xyz0); // ECHO: [1, 2, 3, 0]
```
Is that what you meant by *promoting to homogeneous coordinates* (for the 3d case)?

jordanbrown0:
I'm not convinced:
* I'm not sure that the replacements are better than the originals.
* I don't see a significant advantage over `yx0(motor_connector_size())`.
* If given those lines, I would have had to look up what they meant... without any clear idea of where to look.  I guess it'd have to be in the definition of vectors.  Remember to document in both directions:  "I want to do X, how do I do it?" and "I see a program that does Y, what does it mean?".

But I don't need to be convinced; I've conceded.

t-paul:
Oh, wait, it would want a 1 as 3rd (2D) or 4th (3D) value?

nophead:
Yes,  I meant:
`
p = [1, 2, 3];
echo(p.xyz1); // ECHO: [1, 2, 3, 1]
`

src/expr.cc
```diff
@@ -371,12 +372,31 @@ MemberLookup::MemberLookup(Expression *expr, const std::string &member, const Lo
 Value MemberLookup::evaluate(const std::shared_ptr<const Context>& context) const
 {
     const Value &v = this->expr->evaluate(context);
+    static const boost::regex re_swizzle_validation("^([xyzw]{1,4}|[rgba]{1,4})$");
```
thehans, line 375: I would use `{2,4}` here just to be precise, even though I realize it is inconsequential since `member.length() > 1` is also checked below.

t-paul (approved):

t-paul:
I'll merge the current stuff as it seems to be a pretty simple and safe extension of the existing `.x` logic and create a new ticket for handling of `0` and `1` values and also warnings on invalid combinations.

Issue #4003, opened 2021-12-09
WIndows 64-bit 2021.12.07 fails to start due to missing DLL entry point


jcwren:
OpenSCAD 2021.12.07 fails to start on Windows 7 64-bit with an error message of "The procedure entry point CreateEventA could not be located in the dynamic link library api-ms-win-core-synch-l1-2-0.dll".  

![image](https://user-images.githubusercontent.com/529578/145444454-dfb05887-217f-4d0b-923e-67caba260e30.png)

t-paul:
Hmpf, not sure if https://github.com/mxe/mxe/pull/2660 means the default is Win10 now. We have to find out if there's a way around that.

jcwren:
Hopefully. Not supporting Windows 7 would be double-plus ungood. It still represents about 18% of all machines running any version of Windows.

nophead:
Which is the last snapshot that works on Win7 please?

t-paul:
Probably https://files.openscad.org/snapshots/Win7/

thehans:
> Hopefully. Not supporting Windows 7 would be double-plus ungood. It still represents about 18% of all machines running any version of Windows.

Win7 was released over 11 years ago and is coming up on 2yrs past End of Life.  How many years **past** EOL is reasonable for an open source project with very limited development resources to support?  I would argue the answer is 0.

t-paul:
While I do agree in general, it's a bit annoying that this randomly broke with the MXE upgrade and maybe without strict need for that from OpenSCAD side.
As the linked change was introduced due to a build failure of `qtbase` there might be no way to get things back even with the hinted build option. At this point it's not clear to me how and where to set that though.

nophead:
I agree with @thehans and 0. The problem is Windows peaked at XP and has gone downhill since then. I reluctantly moved to Win7 and had to replace a perfectly working scanner with a new model. It has less capable optics, so can now only scan flat sheets, whereas I could scan PCB's with the old one. The software interface that came with is better though.

I looked at upgrading to Win10, which is already replaced, but it said I needed a new computer and presumably I would need new peripherals as well. It also updates itself without permission, so is no use to me. I don't need a new PC as they don't get much faster for single threaded programs. And I have 3, so I would need to replace all of them. Bonkers as they still work flawlessly with Win7.

I don't need a new OS because all it needs to do is launch applications and no reason that needs to change.

[...]

PR #4012, opened 2021-12-17
Declare Windows7 as minimum version.

t-paul:


t-paul:
Did not help with #4003, still giving the missing DLL error.

PR #4013, opened 2021-12-17
Fix deprecations and warnings found on MacOS

ChrisCoxArt:
Fix a bunch of deprecations, warnings, UBSan, and analyzer issues.
Several remain: some bogus, some about GL deprecation, and some in externally supplied code.  But I went through all of them and fixed what I could.

### DIFF
--- a/src/OffscreenContextCGL.mm
+++ b/src/OffscreenContextCGL.mm
@@ -70,6 +70,8 @@ OffscreenContext *create_offscreen_context(int w, int h)
   ctx->openGLContext = [[NSOpenGLContext alloc] initWithFormat:pixFormat shareContext:nil];
   if (!ctx->openGLContext) {
     std::cerr << "Unable to create NSOpenGLContext\n";
+    [ctx->pool release];
+    delete ctx;
     return nullptr;
   }
 
@@ -79,12 +81,18 @@ OffscreenContext *create_offscreen_context(int w, int h)
   GLenum err = glewInit();
   if (GLEW_OK != err) {
     std::cerr << "Unable to init GLEW: " << glewGetErrorString(err) << std::endl;
+    [ctx->openGLContext release];
+    [ctx->pool release];
+    delete ctx;
     return nullptr;
   }
   glew_dump();
 
   ctx->fbo = fbo_new();
   if (!fbo_init(ctx->fbo, w, h)) {
+    [ctx->openGLContext release];
+    [ctx->pool release];
+    delete ctx;
     return nullptr;
   }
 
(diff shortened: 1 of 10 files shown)


ChrisCoxArt:
Sigh. All tests passed on my system.  Best guess is that path separator is breaking some very old code that somehow doesn't fire on my system.

ChrisCoxArt:
Ok, the path separator seems to have fixed it. Now I'll have to track down why the old MacOS 9 path separator is being used in the code somewhere (and so far I haven't found it, so most likely in a library).

t-paul:
The Path Separator as `:` seems correct. It's meant to separate multiple full/relative path not directories. So something like `/Dir1/Dir2`**:**`/Dir3/Dir4`.

src/OffscreenContextCGL.mm
```diff
@@ -39,7 +37,7 @@
   std::ostringstream out;
   out << "GL context creator: Cocoa / CGL\n"
       << "PNG generator: Core Foundation\n"
-      << "OS info: Mac OS X " << majorVersion << "." << minorVersion << "." << bugFixVersion << " (" << name.machine << " kernel)\n"
+      << "OS info: Mac OS X " << osVersion << " (" << name.machine << " kernel)\n"
```
t-paul, line 40: Sorry for nitpicking, but should that maybe show `macOS` instead of `Mac OS X`?

ChrisCoxArt:
Ah, normally the term "path separator" applies to separating directories. And ":" was the separator used in MacOS 9.
That change is reverted anyway, and this should be ready to integrate.

t-paul (approved):

src/OffscreenContextCGL.mm
t-paul: Grepping shows also one in PlatformUtils-mac.mm.

ChrisCoxArt: Yeah.  I tried to preserve the style and strings that already existed.
The OS label could go either way (Mac OS X 12.1  or  MacOS 12.1) and be acceptable. It's just informative text, which doesn't get used often (if at all).

t-paul: Right, we can always change it later if someone complains :grinning:.

jordanbrown0:
The comment added to InputEventManager::scale() appears to be simply incorrect.  The mapping is symmetric; -1 maps to -10.72, just as +1 maps to +10.72.

| val | -1 | +1 |
|--------|--------|--------|
| `x =4*val` | -4 | +4 |
| `xx = x < 0 ? -exp(-x) + 1 : exp(x) - 1` | -53.60 | 53.60 |
| `xx / 5.0` | -10.72 | 10.72 |

I'll fix it.

Issue #1172, opened 2015-01-22
Syntax highlighting: # affects the whole line


kintel:
This is likely due to QScintilla's C++ parser tagging the whole line as a preprocessor directive.
The solution is probably to write our own parser component for QScintilla.

![screen shot 2015-01-21 at 19 13 38 pm](https://cloud.githubusercontent.com/assets/186094/5848010/a506ffe4-a1a1-11e4-8169-f927fec8f049.png)
##

ChrisCoxArt:
The problem is in the scintilla CPP lexer used by Qscintilla.

QScintilla/scintilla/lexers/LexCPP.cpp LexerCPP::Lex()
```
			} else if (sc.ch == '#' && visibleChars == 0) {
				// Preprocessor commands are alone on their line
				sc.SetState(SCE_C_PREPROCESSOR|activitySet);
```
The logic after setting the preprocssor state is very C/C++ specific and not very customizable.
The rest of the state machine under 	
```
		case SCE_C_PREPROCESSOR:
```
is also veryC/C++ specific and non-customizable.

We can do a little better and highlight everything after the first word by setting 
        newLexer->setStylePreprocessor(true);
in the OpenSCAD code (ScintillaEditor.cc / void ScintillaEditor::setColormap() ). That will highlight everything after the first word, but the first word will remain unhighlighted until there is a change to the scintilla code.

t-paul:
For reference, the separate lexer code is in branch: https://github.com/openscad/openscad/tree/lexertl

ChrisCoxArt:
Relevant bits of the branch (which hasn't been updated since 2015):
qscintillaeditor.cpp
scadlexer.cpp
lex.cc
lex.h
lexertl/*

ChrisCoxArt:
The lexertl library could be added to the dependency downloads, but homebrew doesn't list it, which would break the homebrew builds.
I guess it could go into source/ext or somewhere in the openscad tree.

t-paul:
It's a not too big header only library, so integrating into src/ext should be ok. I guess we would need to switch to the C++14 version though, which seems to be maintained https://github.com/BenHanson/lexertl14.

ChrisCoxArt:
Do we want a partial fix for the time being (allow syntax coloring for everything after the first word after the '#'), or wait for a complete fix ?
The partial fix is one line
```
 newLexer->setStylePreprocessor(true);
```

t-paul:
[...]

PR #4018, opened 2021-12-19
partial fix for issue #1172, allow partial syntax coloring on lines starting with #

ChrisCoxArt:
This is a partial fix for issue #1172
While we can't allow full syntax coloring on lines starting with # (without major changes), we can color everything after the first word.


ChrisCoxArt:
103 minutes to finish all checks. ouch.

t-paul (approved):

PR #4031, opened 2021-12-30
Minor build system fixes

kintel:
Mostly for macOS:
* Removing libxml2, as it comes with macOS
* QScintilla wasn't properly cleaned before build, potentially leaving caches from old build. This caused the build to break after a compiler update.
* Removed qmake-specific comments
* `scripts/setenv-macos.sh` can now be used interactively, for manual builds.
* Fixed `sed` invocation causing a couple of stray `-e` backup files.
* GLIB2 may use built-in libs, don't limit lib searches to our install dirs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 *~
 *.tar*
 *.new
+*.bak
 out.*
 Makefile
 Makefile.Debug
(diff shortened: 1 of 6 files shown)


t-paul (approved):
Looks good.

PR #4040, opened 2022-01-07
SVG import loses transform on <use> element

cwalther:
In an imported SVG, a `<use>` element that has a `transform` attribute is mistakenly imported without that transform. This is because the translation inserted to place the element at the position given by the `x` and `y` attributes overwrites the original transform, while the [SVG specification](https://www.w3.org/TR/SVG11/struct.html#UseElement) says that it should be appended.

These commits add a regression test and a fix.

Actual:
use-transform-fail

Expected:
use-transform-expected


t-paul (approved):
:heart:

PR #4041, opened 2022-01-07
SVG import: Respect "display: none"

cwalther:
Currently, SVG import renders all elements, even those that should not be due to their `display="none"` attribute or `display: none` style. The attached commits add a regression test and a fix.

In particular, this properly hides hidden layers in Inkscape documents.

Includes some minor related fixes, as well as #4040 which the test relies on.

Fixes #2736.

Actual:
display-actual-fail

Expected:
display-expected

PR #4042, opened 2022-01-07
SVG import: Allow selecting parts by id

### LINKED ISSUE #2383: Feature Request: Add layer= for import() of SVG files.
Not urgent, can be worked around handraulicly , but would be useful. 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/57342913-feature-request-add-layer-for-import-of-svg-files?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cwalther:
This adds an `id` parameter to the `import` module that allows selecting parts of an SVG to import, either single elements or whole groups (including layers).

In particular, Inkscape layers can be selected using `import("file.svg", id="layer2")`.

I was unsure whether to reuse the `layer` parameter for this, which already performs a similar function for DXF import (although layers in DXF are a different thing than layers in vector graphics applications), but since this is more general than selecting layers, I decided to introduce a new one. They are aliased internally though, so you can just as well use `layer` instead of `id`.

The internal mechanism could easily be extended to select by other criteria than the `id` attribute, e.g. by stroke color, if a suitable query language were to be implemented (CSS selectors?). I have no plans to do that at this point.

Includes #4041 which it builds upon, only the last commit is new.

Fixes #2383.


cwalther:
Shall I rebase this to get the extra commits out of the way, now that they are merged?

t-paul:
Hmm, yeah, lets try that. Feel free to force-push your branch and keep this PR.

I'll have to think a bit about the new `id` attribute, so maybe there's some small changes before merging and having it more focused would be nice. In general it looks great, and big thanks for the separate PRs with dedicated test cases.

The concern is mostly what's discussed in #2383, as SVG has no official layers. But as also mentioned there, it seems too useful of a feature to not have it because of that. Supporting mostly the input of other open source tools, specifically Inkscape sounds like a sensible approach.

cwalther:
OK, rebased on master, with no changes to the code.

Sure, I am open to discussion. Other than the aliasing with the `layer` parameter, which can easily be removed if preferred, there is nothing in this that makes any assumptions about SVG having layers or not. It just happens to be useful with the way Inkscape and Illustrator and [apparently](https://github.com/w3c/svgwg/issues/68#issuecomment-196253175) many other vector graphics applications encode their own concept of layers into SVG (not CAD applications, which tend to have a different concept of layers).

t-paul:
Don't worry about the ubuntu-18.04 fail, looks like there's some certificate issue on the build server.
```
Certificate verification failed: The certificate is NOT trusted.
The received OCSP status response is invalid. 
Could not handshake: Error in the certificate verification. [IP: 195.135.221.134 443]
```
I'll just re-trigger the build sometime later.

ChrisCoxArt:
Looks like ubuntu had a problem downloading a library.

t-paul:
Sorry for the delay, too much going on. So here's what I'm thinking after some testing with a sample Inkscape file:
- I like the feature, it's been asked for a couple of times, we certainly want this
- Looking at how Inkscape handles layers, we probably don't want the `id` / `layer` alias
- Keeping the `id` filter seems like a useful thing and it matches what SVG provides. Also `id` might be generic enough to be eventually useful for other imports too. It would mostly be useful for picking objects or groups where it's possible to set the id in the GUI
- The `layer` may need to actually check `inkscape:groupmode="layer"` and `inkscape:label="layer-2"`. I can't find an option to actually control the `id` of a (logical Inkscape) layer, so the value is basically unknown to the user. This makes it specific for Inkscape, but that's ok for a start.

The "branch has conflicts" is caused by the application of a defined code style. No need to care much about this, the branch currently applies fine with the "ignore whitepace" option.

```
color("red") import("layers.svg", id="layer1");
color("green") translate([0, 60]) import("layers.svg", id="layer2");
color("blue") translate([0, 60]) import("layers.svg", id="layer3");
color("cyan") translate([-10, -60]) import("layers.svg", id="id-group-1");
color("yellow") translate([10, -60]) import("layers.svg", id="id-star-2");
```
layers

grafik

celer:
I've updated this PR so it compiles against master again, I can create a new PR if desired:

[
](https://github.com/openscad/openscad/compare/master...celer:openscad:selectid)

I'd really like to see this merged as it's a feature that would dramatically improve my openscad workflow.

t-paul:
If it's still just the whitespaces, that's fine, I can handle that.

What's your feedback on my comments? It seems to me the current way of handling this misses a lot of potential use, at least in combination with user level Inkscape input (like not manually or programmatically modified SVG files).

celer:
> If it's still just the whitespaces, that's fine, I can handle that.

Some of the files were moved around and it was the white space changes. I had to resolve a few other merge conflicts.

> 
> What's your feedback on my comments? It seems to me the current way of handling this misses a lot of potential use, at least in combination with user level Inkscape input (like not manually or programmatically modified SVG files).

I'm not the original author of this PR I'm just somebody who really wants openscad to have a bit more functionality when it comes to importing SVG files.

I'm entirely use case-driven, I have a specific use case that I need to solve using this functionality, unfortunately I will be in the hospital for the next week so I won't get to test this out. But I believe this PR resolves my use case.

I understand the complexity of the issue with layers, but at this point any additional functionality when it comes to importing SVGs to select specific objects or layers is useful.

cwalther:
@t-paul, I’m sorry for the silence, I must have missed your comment from 8 Feb or ignored it because I had too much going on myself, and then forgot about it.

@celer, thanks for the update. I’ll take a look and can push it here.

I’d like to remove the `id`/`layer` aliasing before it’s merged, since that seems to be preferred, but I can’t commit to a time frame for that. Implementing `layer` in the Inkscape-specific way can then be done later.

t-paul:
@celer, Oh, oops, sorry, I totally missed you are not the PR owner.

@cwalther integrating the merge fix would be nice. Maybe I'll find some time to look at some minimal changes, right now I'm also a bit short on time to spare, but I'd really like to see this feature integrated.

cwalther:
@celer, your version still doesn’t totally follow the new code style, it still has some tabs that I originally wrote that should now be spaces and other minor style differences. I have cleaned those up (by hand, so I might have missed some – is there an automated style check?) and pushed the result.

I have not tested whether this compiles or works, just that it matches my original diff, and I have not separated `id` and `layer` yet.

t-paul:
There's `scripts/beautify.sh` running the `uncrustify` tool using the configuration file on top level of the repo. If it runs in a branch it tries to touch only the files modified by the branch.

cwalther:
OK, will try running that when I get set up for it.

Meanwhile, I am guessing (untested) that separating `id` and `layer` should be this additional commit. (I’ll squash them all together at the end.) Let’s see whether CI builds that. The test should fail now since I haven’t adjusted it yet. Will fix that next.

cwalther:
So far, so good. So here is a preliminary fix for the test. A proper fix would be to remove the redundant lines 5+6 in _id.scad_ and generate a new _id-expected.png_ with the top right square removed. Maybe someone else who is set up for building and running tests will get around to that sooner than I will.

cwalther:
Well, looks like that didn’t fix the test. Figuring out why will need to wait until I can run it locally. Probably some stupid mistake.

celer:
Thank you guys for looking at this. I got motivated to try to get this working right before I went and got heart surgery. So I'm hanging out in the hospital recovering from heart surgery.

cwalther:
Alright, I have fixed it. Evidently adding an attribute isn’t as simple as adding a member, it must also be added to the `toString()` output, because that is apparently used as a cache key – otherwise, all `import`s of the same file would be replaced by the first one even though they have different `id`s. Adding it always however makes some `dumptest` tests fail that have expectations on the output, so only add it when it’s set. I have not checked whether it would be okay to adjust the expected output of these tests instead or whether backward compatibility is important there.

So, here is the result, simplified into one commit, with an improved test, and some code style amendments by _scripts/beautify.sh_. This is ready for review and merging as far as I am concerned.

JelleHissink:
Any idea on how I can get my hands on the windows build output? I really need/want this feature, currently I spend my time running a script next to openscad to preprocess my svg into layer seperated svg files (easy xml transformation). But I'd like to test this and throw my script in the trash!

t-paul:
Win64: https://app.circleci.com/pipelines/github/openscad/openscad/2890/workflows/25033df3-d2e3-4417-a053-e7fce1ec9a80/jobs/12116/artifacts

Grab it, while it's still hot, I'm not sure how long they store the builds :grin:.

JelleHissink:
Thanks! I will give this a go.

jordanbrown0:
I built this fork for a project where I wanted to import a 2.5D drawing from Inkscape, and it worked perfectly.

Note that you can set the ID of individual paths using the Object Properties panel.

cwalther:
Thanks for merging, and thanks for the extensions in #4378!

PR #4052, opened 2022-01-16
catch exceptions in CGAL worker threads, fixing issue #3756

ChrisCoxArt:
catch std::exception derivatives inside CGAL code, and report error.
catch ALL exceptions thrown to root level of worker thread, and report errors.


t-paul (approved):

t-paul:
Merged via separate PRs #4089 and  #4090. Thanks!

Issue #4061, opened 2022-01-21
OpenSCAD crashes on MacOS if a file is opened but the application is denied access to the folder


ChrisCoxArt:
Branching this from #4036
Unhandled exception is the root cause.
Failure to correctly handle files that cannot be opened due to permissions is the larger problem.

I reproduced the crash if you denied access to the directory that the file is in.
```
bool parse(SourceFile *&file, const std::string& text, const std::string &filename, const std::string &mainFile, int debug)
{
  fs::path filepath = fs::absolute(fs::path(filename));
  mainFilePath = fs::absolute(fs::path(mainFile));
```
Throws a std::exception that the code path does not catch, resulting in a call to terminate().
Good news - this is not specific to Kintel's changes or M1 port.

ChrisCoxArt:
This was fixed with the pull request, and can be closed now.

kintel:
Fixed by #4062

PR #4062, opened 2022-01-21
Fix crash issue #4061

ChrisCoxArt:
Multiple layered fixes to handle exceptions thrown when permissions deny access to files opened, fixing issue #4061

1) catch the boost::fs exception near the cause in parse(), and return false from the parser when the file could not be opened.
2) if the file cannot be opened for some reason, don't try to parse the file.
3) add catch(...) around parsing and report an unknown exception error instead of crashing with uncaught exceptions.
4) if parsing fails for any reason, don't leave some variables with stale contents that could cause access violations later.

### DIFF
--- a/src/MainWindow.cc
+++ b/src/MainWindow.cc
@@ -1059,9 +1059,8 @@ void MainWindow::compile(bool reload, bool forcedone)
 		if (reload) {
 			// Refresh files if it has changed on disk
 			if (fileChangedOnDisk() && checkEditorModified()) {
-				shouldcompiletoplevel = true;
-				tabManager->refreshDocument();
-				if (Preferences::inst()->getValue("advanced/autoReloadRaise").toBool()) {
+				shouldcompiletoplevel = tabManager->refreshDocument();  // don't compile if we couldn't open the file
+				if (shouldcompiletoplevel && Preferences::inst()->getValue("advanced/autoReloadRaise").toBool()) {
 					// reloading the 'same' document brings the 'old' one to front.
 					this->raise();
 				}
@@ -1127,9 +1126,11 @@ void MainWindow::compile(bool reload, bool forcedone)
 		}
 
 		compileDone(didcompile | forcedone);
-	}catch(const HardWarningException&){
+	} catch(const HardWarningException&) {
 		exceptionCleanup();
-	}
+	} catch (...) {
+        UnknownExceptionCleanup();
+    }
 }
 
 void MainWindow::waitAfterReload()
@@ -1565,7 +1566,7 @@ void MainWindow::actionReload()
 {
 	if (checkEditorModified()) {
 		fileChangedOnDisk(); // force cached autoReloadId to update
-		tabManager->refreshDocument();
+		(void)tabManager->refreshDocument();    // ignore errors opening the file
 	}
 }
 
@@ -1824,7 +1825,9 @@ void MainWindow::parseTopLevelDocument()
 	auto fnameba = activeEditor->filepath.toLocal8Bit();
 	const char* fname = activeEditor->filepath.isEmpty() ? "" : fnameba;
 	delete this->parsed_file;
-	this->root_file = parse(this->parsed_file, fulltext, fname, fname, false) ? this->parsed_file : nullptr;
+    this->parsed_file = nullptr;    // because the parse() call can throw and we don't want a stale pointer!
+    this->root_file = nullptr;      // ditto
+    this->root_file = parse(this->parsed_file, fulltext, fname, fname, false) ? this->parsed_file : nullptr;
 
 	if (this->root_file!=nullptr) {
 		//add parameters as annotation in AST
@@ -2367,6 +2370,14 @@ void MainWindow::exceptionCleanup(){
 	LOG(message_group::None,Location::NONE,""," ");
 	GuiLocker::unlock();
 	if (designActionAutoReload->isChecked()) autoReloadTimer->start();
+}
+
+void MainWindow::UnknownExceptionCleanup(){
+    setCurrentOutput();     // we need to show this error
+    LOG(message_group::Error,Location::NONE,"","Parsing aborted by unknown exception");
+    LOG(message_group::None,Location::NONE,""," ");
+    GuiLocker::unlock();
+    if (designActionAutoReload->isChecked()) autoReloadTimer->start();      // should we stop timers on unknown errors?
 }
 
 void MainWindow::actionDisplayAST()
--- a/src/MainWindow.cc
+++ b/src/MainWindow.cc
@@ -2377,7 +2377,7 @@ void MainWindow::UnknownExceptionCleanup(){
     LOG(message_group::Error,Location::NONE,"","Parsing aborted by unknown exception");
     LOG(message_group::None,Location::NONE,""," ");
     GuiLocker::unlock();
-    if (designActionAutoReload->isChecked()) autoReloadTimer->start();      // should we stop timers on unknown errors?
+    if (designActionAutoReload->isChecked()) autoReloadTimer->start();
 }
 
 void MainWindow::actionDisplayAST()
--- a/src/parser.y
+++ b/src/parser.y
@@ -759,10 +759,17 @@ void handle_assignment(const std::string token, Expression *expr, const Location
 
 bool parse(SourceFile *&file, const std::string& text, const std::string &filename, const std::string &mainFile, int debug)
 {
-  fs::path filepath = fs::absolute(fs::path(filename));
-  mainFilePath = fs::absolute(fs::path(mainFile));
-  parsingMainFile = mainFilePath == filepath;
+  fs::path filepath;
+  try {
+    filepath = fs::absolute(fs::path(filename));
+    mainFilePath = fs::absolute(fs::path(mainFile));
+  } catch (...) {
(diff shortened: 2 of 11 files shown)


src/MainWindow.cc
```diff
@@ -2369,6 +2372,14 @@ void MainWindow::exceptionCleanup(){
 	if (designActionAutoReload->isChecked()) autoReloadTimer->start();
 }
 
+void MainWindow::UnknownExceptionCleanup(){
+    setCurrentOutput();     // we need to show this error
+    LOG(message_group::Error,Location::NONE,"","Parsing aborted by unknown exception");
+    LOG(message_group::None,Location::NONE,""," ");
+    GuiLocker::unlock();
+    if (designActionAutoReload->isChecked()) autoReloadTimer->start();      // should we stop timers on unknown errors?
```
kintel, line 2380: I think this it's pretty innocent to keep it like this, as long as it doesn't keep spewing errors (which it shouldn't). I think we can remove the comment.

src/parser.y
```diff
@@ -759,10 +759,17 @@ void handle_assignment(const std::string token, Expression *expr, const Location
...
 bool parse(SourceFile *&file, const std::string& text, const std::string &filename, const std::string &mainFile, int debug)
 {
-  fs::path filepath = fs::absolute(fs::path(filename));
-  mainFilePath = fs::absolute(fs::path(mainFile));
-  parsingMainFile = mainFilePath == filepath;
+  fs::path filepath;
+  try {
+    filepath = fs::absolute(fs::path(filename));
+    mainFilePath = fs::absolute(fs::path(mainFile));
+  } catch (...) {
```
kintel, line 766: Is there a more specific exception we can listen for?
Not a big deal, as I don't expect there to be many other failure cases, and oom errors are probably not expected to occur here.

kintel (approved):
This all looks reasonable to me.
Caveat: Since I don't experience the problem, I haven't verified this other than reading code.

src/parser.y
ChrisCoxArt: std::exception should work, but I wanted to be safe and catch everything.

src/MainWindow.cc
ChrisCoxArt: I looked at the timer code, but couldn't understand how it was used very quickly.
If it is safe, then we can remove the comment.

ChrisCoxArt:
Verifying also means peeling the onion and testing the behavior at each layer - which takes a while.
And, of course, Apple makes it behave a bit differently in XCode than in a signed application (so I had to keep switching between debugging in XCode, and running the command line built app).

ChrisCoxArt:
This is why I miss working on a team in an office - over the shoulder code reviews go sooooo much faster.

src/MainWindow.cc
kintel: It's a bit of a hack used to discover file changes for the auto-reload UI feature. Ideally this should be rewritten into using a filesystem monitor.

src/parser.y
kintel: std::exception doesn't really make a difference IMO, I was more wondering if there's an actual subtype being thrown for the permission error. But I'm fine keeping it like this, we probably have bigger fish to fry :)

kintel:
I also just discovered https://mac.getutm.app/, which apparently supports "M1 guest on M1 host" virtualization. I'll try to give it a spin and see if I can reproduce this behavior on a vanilla install

kintel:
In a clean VM with macOS 12.1, I get the "repeated dialog" issue, but even if I answer "don't allow", I cannot reproduce the crash (it just keeps asking over and over again).

ChrisCoxArt:
Is that with, or without the fix?
I've never been able to reproduce the "asking many times" prompts -- it only prompts when opening (or saving) a document.

kintel:
Without the fix - it's the exact same binary everyone else is using. I really just want to reproduce the crash the be able to verify this, and later understand it to debug any related issues.

Btw., what version of macOS are you using to reproduce this?

ChrisCoxArt:
MacOS 12.1, Intel Xeon W processor, 160 GB RAM.

kintel:
oh, x86. The other crash report was macOS 12.0.1 on M1, so that doesn't sound isolated to neither arch or macOS version..

kintel:
Btw., can your reproduce this crash with a regular development snapshot, or only using the universal binary?

ChrisCoxArt:
That's what I said previously - it is not related to your changes or the M1 processor.

ChrisCoxArt:
And I can reproduce it with a downloaded dev build, or a local build.
No M1 build or universal binary needed.
Again, that's why I broke this out into a different bug report, because it was not related to your M1 work - just an app bug on MacOS.

ChrisCoxArt:
Report from a dev build I just downloaded, matches user report and my debugger findings:
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Process:               OpenSCAD [59472]
Path:                  /Volumes/VOLUME/*/OpenSCAD.app/Contents/MacOS/OpenSCAD
Identifier:            org.openscad.OpenSCAD
Version:               2022.1 (2022.01.22)
Code Type:             X86-64 (Native)
Parent Process:        launchd [1]
User ID:               501

Date/Time:             2022-01-22 20:24:58.8210 -0800
OS Version:            macOS 12.1 (21C52)
Report Version:        12
Bridge OS Version:     6.1 (19P647)
Anonymous UUID:        096D5B68-B94E-63F0-17DD-11136D1F6B88


Time Awake Since Boot: 260000 seconds

System Integrity Protection: enabled

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Application Specific Information:
terminating with uncaught exception of type boost::filesystem::filesystem_error: boost::filesystem::current_path: Operation not permitted
abort() called


Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	    0x7ff8014f9112 __pthread_kill + 10
1   libsystem_pthread.dylib       	    0x7ff80152f214 pthread_kill + 263
2   libsystem_c.dylib             	    0x7ff80147bd10 abort + 123
3   libc++abi.dylib               	    0x7ff8014ec0b2 abort_message + 241
4   libc++abi.dylib               	    0x7ff8014dd1e5 demangling_terminate_handler() + 242
5   libobjc.A.dylib               	    0x7ff8013da511 _objc_terminate() + 104
6   libc++abi.dylib               	    0x7ff8014eb4d7 std::__terminate(void (*)()) + 8
7   libc++abi.dylib               	    0x7ff8014edd55 __cxxabiv1::failed_throw(__cxxabiv1::__cxa_exception*) + 27
8   libc++abi.dylib               	    0x7ff8014edd1c __cxa_throw + 116
9   libboost_filesystem.dylib     	       0x10581387f boost::filesystem::emit_error(int, boost::system::error_code*, char const*) + 431
10  libboost_filesystem.dylib     	       0x105819728 boost::filesystem::detail::current_path(boost::system::error_code*) + 472
11  OpenSCAD                      	       0x104c55942 parse(SourceFile*&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, int) + 50
12  OpenSCAD                      	       0x104e8e7b7 MainWindow::parseTopLevelDocument() + 871
13  OpenSCAD                      	       0x104f195a1 TabManager::openTabFile(QString const&) + 641
14  OpenSCAD                      	       0x104f186e1 TabManager::createTab(QString const&) + 1873
15  OpenSCAD                      	       0x104f17bc1 TabManager::TabManager(MainWindow*, QString const&) + 401
16  OpenSCAD                      	       0x104e7e62c MainWindow::MainWindow(QStringList const&) + 3532
17  OpenSCAD                      	       0x1048b77d8 gui(std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > > >&, boost::filesystem::path const&, int, char**) + 2088
18  OpenSCAD                      	       0x1048bd2cf main + 17823
19  dyld                          	       0x11194d4fe start + 462

ChrisCoxArt:
Hmm, I'm getting inconsistent results on M1, with an intel dev snapshot.  (but I only have MacOS 11.6 on here)

kintel:
I'm installing 12.0.1 in a VM to test. If I don't learn anything, I'd say it's probably safest to just merge this.

It's a little weird though, that nobody else have reported this crash..

kintel:
I get the exact same behavior on 12.0.1 as I did for 12.1. 

I'll merge this PR if you're content with it.

ChrisCoxArt:
Yes, I'm pretty sure this fixes this issue and should help catch any similar issues (and prevent them from crashing).
Go ahead and merge if you're happy with the code.

The crash requires a new install (or build) of the app plus denying access to the location of the file you selected -- the combination of which are unlikely... until some starts testing a new build.

PR #4067, opened 2022-01-24
Use cmake for install and packaging with Windows

thehans:


cmake/Modules/info.cmake
```diff
@@ -44,5 +44,3 @@ else()
   message(STATUS "- DBus Remote Driver (disabled)")
 endif()
 message(STATUS " ")
-message(STATUS "DEFINES = ${DEFINES}")
-message(STATUS " ")
```
t-paul, line 48: Do you know why this is not showing anything anymore? It was quite useful before, when it was showing all the defines.

Is that maybe: https://cmake.org/cmake/help/latest/prop_dir/COMPILE_DEFINITIONS.html ?

thehans: Oops, I removed that because I didn't remember it working before, figured it was just always messed up or been that way a long time.  I just realized that its actually because I moved the properties from being directory-based to target-based when I was cleaning things up for mimalloc.  The `DEFINES` variable gets set on the first line of `info.cmake`, and is used in the various input driver conditionals as well.

Fixed it just now and force pushed.

cmake/Modules/info.cmake
```diff
@@ -1,4 +1,4 @@
-get_directory_property(DEFINES COMPILE_DEFINITIONS)
+get_target_property(DEFINES OpenSCAD COMPILE_DEFINITIONS)
```
t-paul, line 1: Ah, nice. I missed that. Good to have that back, it has been quite useful when looking at CI build issues. Using target properties seems much better.

thehans:
I decided to revive this PR.  Not necessarily for getting shared library mimalloc working on windows, but mostly for the work done towards cmake-based install and packaging.

I think that consolidating our various packaging methods with cmake would make things simpler and more manageable overall.
It appears to be capable of producing packages for pretty much all of our supported platforms, via different "cpack generators", including:
- [CPack NSIS Generator](https://cmake.org/cmake/help/latest/cpack_gen/nsis.html) - which I was testing in this PR already.
- [CPack Bundle Generator](https://cmake.org/cmake/help/latest/cpack_gen/bundle.html) - for macos
- [CPack DEB Generator](https://cmake.org/cmake/help/latest/cpack_gen/deb.html)
- [CPack RPM Generator](https://cmake.org/cmake/help/latest/cpack_gen/rpm.html)
... and more, with the [full list here](https://cmake.org/cmake/help/latest/manual/cpack-generators.7.html#manual:cpack-generators(7))

So I plan to use this PR as a space to explore testing and implementing more of those generators.

thehans:
I rebased this branch after merging #4416

thehans:
MXE builds were failing to package up a test file with utf8 characters in the filename.  I found this related issue: https://gitlab.kitware.com/cmake/cmake/-/issues/16505  which points out that it succeeds after setting the shell locale to `en_US.UTF-8`

thehans:
I think this PR is in a good place for merging now.
I would still like to see if macos bundles from cpack would work well, but I think it is better to do that in a separate PR.

cmake/Modules/FindLib3MF.cmake
```diff
@@ -27,9 +27,10 @@ set(LIB3MF_FOUND ${PC_LIB3MF_FOUND})
 find_path(LIB3MF_INCLUDE_DIRS
     NAMES Model/COM/NMR_DLLInterfaces.h lib3mf_implicit.hpp
     HINTS $ENV{LIB3MF_INCLUDEDIR}
+          ${PC_LIB3MF_INCLUDEDIR}
```
thehans, line 30: TODO: remove this accidental duplicate line

PR #4068, opened 2022-01-25
Update windows installer links

Rotzbua:
* use secure https instead of http
* default openscad homepage moved to non www.


t-paul (approved):
Awesome, thanks, I did not even remember the installer has those links.

PR #4074, opened 2022-01-29
Increase input device button count to 24 (fixes #4073).

### LINKED ISSUE #4073: Increasing max_buttons
I tested gamepad navigation with a PS3 controller, which has 17 buttons. The right button on the left arrow pad would be `Button16`, but max_buttons is only 16, so it is not possible to assign it.

My OpenSCAD version is `2022.01.25.ai10450 (git 6f3767b99)`

t-paul:


MichaelPFrey:
`getButtonCount()` and `getAxisCount() `are intended to return the number of the Buttons/Axis the controller has (so that for e.g. the UI can disable irrelevant fields).

A proper implementation (as in asking the operating system which gets the info from the device) is in 
https://github.com/openscad/openscad/blob/master/src/input/JoystickInputDriver.h / https://github.com/openscad/openscad/blob/master/src/input/JoystickInputDriver.cc

As we do not necessarily get access to the values from the device, but we can make an educated guess:
   * QGamePad just assumes an XBox360 Controller - so we should to the same
      * 16 Buttons
      * 6 Axis
      * more are not supported by QGamePad. The way QGamepad is written makes adding buttons tricky - so I do not expect that to happen. Even if: We would have to update our implementation to support more buttons.
   *  space navigator
      * all space navigators currently have six axis - even if in the future one would be created with more, it would require a driver change to access/map the new axis
   * hid api input driver
      * all space mice currently have six axis - even if in the future one would be created with more, it would require a driver change to access/map the new axis
      * hidapi_decode_button has a hard coded limit for 16 buttons

Room for improvement is and was that for a space mouse with 0 buttons (e.g. "3Dconnexion Space Mouse Compact"), we could/should report back 0 buttons. (The hip api driver knows which mice is connected)

Reducing the axis number from 9 to 8:

Note that 8 axis is not a lot: An Xbox 360 controller natively under linux reports 8 axis.
(QGamepad remaps the two axis for the dpad to 4 buttons)
I can imagine more exotic devices ()

t-paul:
> Reducing the axis number from 9 to 8:

I don't think I did that. It was using 0 - 9 and I fixed that to actually be 9 with index 0 to 8.

MichaelPFrey:
> I don't think I did that. It was using 0 - 9 and I fixed that to actually be 9 with index 0 to 8.

My bad - that sounds correct. Thank you for the hint.

t-paul:
I'd like to check the spacenav driver again anyway. It's currently not enabled on some of the builds and people have asked for that. Maybe a chance to see if there's some more changes needed.

t-paul:
@MichaelPFrey the QGamePad still confuses me. The test application from the documentation shows setup for 18 buttons.

PR #4078, opened 2022-02-01
Fix type error on Setting `entries`

unDeadHerbs:


unDeadHerbs:
Someone should check how #4074 passed the checks; because, to my understanding, that can't have built correctly.  There is no [`std::array::operator*`](https://en.cppreference.com/w/cpp/container/array).

t-paul:
My first idea would be "release build" but that also would mean we don't have a single debug build, which is a bit surprising. If that's the case we probably want to have at least one of the platforms span a build matrix for both release and debug builds.

Also, ignore the MacOS fail. Looks like Homebrew just updated `boost` causing the failure.

t-paul (approved):

PR #4100, opened 2022-02-07
Only trade AbstractNodes in shared pointers

ochafik:
This is a mostly mechanical refactoring meant to make it easier to transform the AST (stuff like https://github.com/openscad/openscad/pull/3637)

- No more naked AbstractNode pointers anywhere (nor explicit deletions).
- AbstractNode extends std::enable_shared_from_this for the occasion (so it can find itself and so visitors who only have a reference to it can also play the shared_ptr game).
- Threw a few autos in the mix.


ochafik:
Oh, please wait before merging, need to confirm this doesn't regress performance

t-paul (approved):

PR #4103, opened 2022-02-09
Convert remaining instance of applyUnion3D in import_3mf.cc

michaelshiel:
I wasn't able to build against lib3mf 2.2.0 because of this error:

```src/import_3mf.cc: In function ‘Geometry* import_3mf(const string&, const Location&)’:
src/import_3mf.cc:379:53: error: cannot convert >‘std::shared_ptr<const Geometry>’ to ‘CGAL_Nef_polyhedron*’ in initialization
379 |     CGAL_Nef_polyhedron *N = CGALUtils::applyUnion3D(children.begin(), children.end());
    |                              ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    |                                                     |
    |                                                     std::shared_ptr<const Geometry>
```

I was able to fix it with this patch, and I think potentially this instance of the `applyUnion3D` call was missed in this commit: https://github.com/openscad/openscad/commit/3b822d59a3355415aabe72a64a77520392304db4

### DIFF
--- a/src/import_3mf.cc
+++ b/src/import_3mf.cc
@@ -369,18 +369,21 @@ Geometry *import_3mf(const std::string& filename, const Location& loc)
   } else if (meshes.empty()) {
     return first_mesh;
   } else {
-    PolySet *p = new PolySet(3);
+    PolySet *p = nullptr;
 #ifdef ENABLE_CGAL
     Geometry::Geometries children;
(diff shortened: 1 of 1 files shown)


t-paul (approved):
Thanks, there's currently no build environment for lib3mf v2.x so this path gets no coverage unfortunately.

PR #4111, opened 2022-02-11
Fast csg repair self intersections

ochafik:


ochafik:
Folding this line of experimentation into https://github.com/openscad/openscad/pull/4121

Issue #4114, opened 2022-02-12
fast-csg performance regressions when combined with trust-corefinement and/or lazy-union


wolfwood:
#3641 is exciting work that is giving me real improvements, however I am seeing `fast-csg-trust-corefinement` take significantly longer in several cases (though the problem models, _thumb_, _fingers_, and _assembly_, are related). I'm also seeing that enabling `lazy-union` with `fast-csg` can give slower results than `fast-csg` alone (for _plate_ and _assembly_).

I'm currently using the Linux nightly `OpenSCAD-2022.02.09.ai10824-2022.02.09.ai10824-x86_64.AppImage`

I gathered commandline render times on my branch https://github.com/wolfwood/tryadactyl/tree/fast-csg-test using the script there called [`time.sh`](https://github.com/wolfwood/tryadactyl/blob/fast-csg-test/time.sh)

|Model | stock render | `lazy-union` | `fast-csg` | `fast-csg` `lazy-union` |   `fast-csg` `trust-corefinement` |   `fast-csg` `trust-corefinement` `lazy-union` |
|:---:|---:|---:|---:|---:|---:|---:|
| _rest_          | 4.238 | 4.229 | **3.806** | 3.869 | 3.849 | 3.855 |
| _thumb_      | 63.618 | 62.283 | 21.795 | **19.468** | 115.209 | 35.996 |
| _plate_        | 396.910 | 349.068 | 72.449 | 108.148 | **24.474** | 63.494 |
| _fingers_     | 168.728 | 166.485 | 84.211 | **76.821** | 864.384 | 840.219 |
| _assembly_ | 705.781 | 605.382 | **190.565** | 216.741 | 1005.702 | 935.563 |

ochafik:
Amazing, thanks for the test cases and for taking the time to crunch the numbers!

I couldn't run *fingers*, *thumb* (missing from the repo) or *assembly* (depends on missing `../KeyV2/includes.scad`), but I reckon these very slow runs would probably be improved (or fixed) by the remeshing (can try it out https://gist.github.com/ochafik/28bde2491d1f8cc0992a51c8e6f25716 if you're willing to build OpenSCAD :-D)

wolfwood:
oh, whoops. I fixed the missing files, and dropped the unnecessary dependency.

successfully followed that gist, I'll give that a shot.

wolfwood:
the numbers for `thumb` are not significantly different with the new build.

ochafik:
[...]

PR #4117, opened 2022-02-12
[fast-csg] Custom remeshing after corefinements

ochafik:
Remeshing logic that "fixes" the results of corefinement operations to avoid an explosion of number of faces that hangs many models.

`fast-csg-remesh` operates as follows (needs `fast-csg`):

- A corefinement visitor keeps track of faces being split and copied (as suggested by @sloriot in https://github.com/CGAL/cgal/pull/5461#issuecomment-1035033500), giving each face a unique patch id that's inherited over copies and splits.
- If no face was split we skip the remeshing (note: we could force remeshing of all meshes, maybe when importing them).
- Maximal patches of neighbouring coplanar faces are then found with a flood-fill algorithm starting from descendants of split faces + exact coplanarity tests. 
- Patches that contain holes are skipped for now
- Patch borders are found and we mark their collapsible vertices (those between collinear edges). If exactly two neighbour patches agree, those vertices will be dropped
- Each patch is replaced with a polygon that espouses its borders, and we let CGAL [triangulate](https://doc.cgal.org/latest/Polygon_mesh_processing/group__PMP__meshing__grp.html#ga70d65044f8c7309c24ade88fa280124a) it.
- A new mesh is created with the edits, as in-place edits seem tricky.

Note: supersedes https://github.com/openscad/openscad/pull/4107 as by design it should lead to less coplanarity tests (for many faces we know they're coplanar because we know they have a common ancestor thanks to the corefinement visitor). Also this means we're not tied to the upcoming release of CGAL.

## Some examples:

| Model | fast-csg | fast-csg + remesh | baseline (nef) |
|-|-|-|-|
| [condensed matter](https://github.com/openscad/openscad/pull/3641#issuecomment-780282019) (@MichaelAtOz) | 383k faces, 3min28sec | 177k faces, 44sec | 172k 5min31sec |
| [PushPullFeeder](https://github.com/markmaker/PushPullFeeder) (@markmaker) | hangs | 89k faces, 44sec | 47k faces, 1min30sec |
| minkowski example from https://github.com/openscad/openscad/pull/4113 | 19k faces, 38sec | 9k faces, 25sec | 4min26sec |

And some simpler ones:

*   This goes from 644 faces to 12:

  ```js
  for (i=[0:2],j=[0:2],k=[0:2])
    translate([i*0.8, j*0.8, k*0.8])
      cube(1);
  ```
  
  image

*   This goes from 40 faces to 28

  ```js
  cube(1); translate([0.5, 0.5, 0]) cube(1);
  ```

Screen Shot 2022-02-21 at 02 10 20

* This goes from 44 to 12 (instead of 12 w/ nef, which is able to figure out it's now just a "cube"):

  ```js
  cube(1); translate([0.5, 0, 0]) cube(1);
  ```

image

## Known limitations

- No support for remeshing of faces with holes (as there are in the models in https://github.com/openscad/openscad/issues/4114). Will probably need an exact version of GeometryUtils::tessellatePolygonWithHoles
- Currently only remeshing *exactly* coplanar faces (the [example below](https://github.com/openscad/openscad/pull/4117#issuecomment-1053731672) with separately-rotated cubes has infinitesimal rounding errors due to transforms not using exact numbers).
- Still some bugs, e.g. trying to remove too many vertices in [For Dyson Cinetic](https://github.com/openscad/openscad/pull/3641#issuecomment-780265132) (@MichaelAtOz)
- The corefinement visitor slows down all operations a bit and uses a bit more memory (although some care was taken to allocate data structures sparingly) but that's quickly recouped in complex models
- This PR depends on https://github.com/openscad/openscad/pull/4146 as it adds STL export tests, this is (the main reason) why the tests aren't green yet.
- The resulting triangulation is sensitive to the order of the corefinement output, which is currently altered by optimization flags (https://github.com/CGAL/cgal/issues/6363). Because of this, a `fast-csg-remesh-predictibly` feature was introduced that takes quite a bit longer to run (+30% total run time on the minkowski example mentioned above) but always starts each face from the "smallest" vertex. It is used by tests in conjunction with the `sort-stl` feature I've recently introduced (https://github.com/openscad/openscad/pull/4146) so we can test stable remeshing outputs against golden STL files.


thehans:
Here is a case which appears to generate a number of T-junctions (or maybe just degenerate faces?   It claims there are 46 facets, but I only count 32 by eye)

```
r = [1,1,1];
render() union() {
	rotate(r) cube();
	rotate(r) translate([0.5,0,0]) cube();
}
```

Tjunctions

MichaelAtOz:
2021.01 also has bad triangulation for that. Slightly different, but T's & degenerate faces.
I notice different triangulation edges display with Preview(F5) with render() <- thin lines no points, v's Render(F6) with render() <- no lines on edges, no points, Render(F6) without render() <- lines on edges & points.
The F5 before F6 v's just F6 problem, different triangulation in cache.

t-paul (approved):

PR #4126, opened 2022-02-17
Support building under Visual Studio on Windows

dannypike:
This is a condensed version of all of the changes that have been as a part of [PR 3957](https://github.com/openscad/openscad/pull/3957), rebased to master beeff28b14a0de55c93ace0255f50e711e9bd841. It includes the recent fix to the "Black Window" bug, but excludes everything after that and also the comments by @thehans on 02 Feb2022. I shall address his suggestions next.

I apologise for the delay in looking at those comments, but I have been focussed on getting the core build working, and making sure that the instructions are readable and accurate.

OpenSCAD now compiles under VS2022 using these files and works for the simple models that I have tried. The build instructions have been completely rewritten to support versioning by using a vcpkg manifest file. The initial vcpkg.json manifest file only selects the CGAL version 5.2.2.

The other changes in this PR are historical and were in the original PR.


dannypike:
@thehans. re your comment [here](https://github.com/openscad/openscad/pull/3957#pullrequestreview-8710254630):
> This whole section is doing some odd things. `CMAKE_MODULE_PATH` Is actually a list of paths to search over, not a single path, so I don't think there should be any check for empty string. The previous behavior was actually appending a path to the existing value, instead of completely replacing it.
> 
> If you look at the automated builds (click the red X next to your commits), you can see that MXE builds have been failing to find libzip since your changes. This is likely because the MXE toolchain sets its own CMAKE_MODULE_PATH, and so our additional path is not appended here.
> 
> If a platform/toolchain, such as VCPKG or MXE provides `Find[package].cmake` files etc, its best to use those before our own. (Ideally we wouldn't need to maintain our own _at all_, if every platform had all the needed find modules).

I agree. I don't understand it either, but it was an attempt to integrate `vcpkg` with what was in the original `CMakeLists.txt` around line 294:
```
# Note: Saving CMAKE_MODULE_PATH as CGAL will overwrite it.
# Reconsider this after CGAL 5.4: https://github.com/CGAL/cgal/pull/6029
set(ORIGINAL_CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH})
set(CGAL_DO_NOT_WARN_ABOUT_CMAKE_BUILD_TYPE TRUE)
find_package(CGAL 5.0 REQUIRED COMPONENTS Core)
message(STATUS "CGAL: ${CGAL_MAJOR_VERSION}.${CGAL_MINOR_VERSION}")
target_compile_definitions(OpenSCAD PRIVATE ENABLE_CGAL)
```

There was some code that also changes it for an MSVC compiler around line 303:
```
if (MSVC)
.
.
.
  # needed for Qt5QScintilla, maybe others
  set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/Modules/")
else()
  set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/Modules/")
. 
.
.
```

after setting up the graphics libraries (the non-MSVC version sets it beforehand).

It's all rather confused and is very likely to be wrong now, but are you familiar with what the block of code in the `master` version of `CMakeLists.txt` is trying to do here?

It restores the value at line 396, but there is a commented-out version that is supposedly ready for CGAL >= 5.4:
```
# revert any changes to module path from CGAL_Macros.cmake; see note above
# Commented out code should work in CGAL>= 5.4
#if (CGAL_MODULE_PATH_IS_SET)
#  set(CMAKE_MODULE_PATH ${ORIGINAL_CMAKE_MODULE_PATH})
#endif()
set(CMAKE_MODULE_PATH ${ORIGINAL_CMAKE_MODULE_PATH})
```

I imagine, at the time, that I was just trying to get something working, to see if I would understand "sometime later", as a result of what else I discovered in the porting process.

So, what do you advise we should with `CMAKE_MODULE_PATH`? I think we need to document a consistent way that we set it / edit it, with comments explaining what happens and why.

dannypike:
@thehans, re: your comment [here](https://github.com/openscad/openscad/pull/3957#pullrequestreview-871026942):

> This message is printing the value of a variable before it is set.

Yes, you are correct. I think that it was probably an copy-paste error from my attempts to understand what is happening with `CMAKE_MODULE_PATH` and `ORIGINAL_CMAKE_MODULE_PATH` as per my comment above. I expect that it should have been printing the value that is being saved, but it is redundant, now.

I've deleted it in the current pull request until I can better understand what we should with `CMAKE_MODULE_PATH`; thank you for noticing it.

dannypike:
@thehans re: your comment about the `glib` libraries [here](https://github.com/openscad/openscad/pull/3957#pullrequestreview-871050455)

> Use namespaced library targets

Do you know what the line should look like? This does not work in Visual Studio:
```
list(APPEND COMMON_LIBRARIES unofficial::glib::gio unofficial::glib::glib unofficial::glib::gmodule unofficial::glib::gobject)
```

neither does this:
```
list(APPEND COMMON_LIBRARIES glib::gio glib::glib glib::gmodule glib::gobject)
```
nor this:
```
list(APPEND COMMON_LIBRARIES glib2::gio glib2::glib glib2::gmodule glib2::gobject)
```
the `find_package()` command sets several variables for `Glib2`, but only seems to think we need `glib-2.0` and `prce`, e.g. this one:

```
1> [CMake] -- GLIB2_LIBRARIES=[...]/x64-Release/vcpkg_installed/x64-windows/lib/glib-2.0.lib;[...]/x64-Release/vcpkg_installed/x64-windows/lib/pcre.lib
```

 That doesn't match what the original `CMakeLists.txt` implies I should use, so I put the filenames in explicitly in the `CMakeLists.txt` that I pushed.

Would you please give me a `CMake` command that you think will work with Visual Studio?

dannypike:
@thehans re: your comment [here](https://github.com/openscad/openscad/pull/3957#pullrequestreview-871051225):

> `find_library` sets `GETTEXT_LIBRARY` to the path, so use it.

You are right. Now fixed, thank you.

thehans:
> I agree. I don't understand it either, but it was an attempt to integrate `vcpkg` with what was in the original `CMakeLists.txt` around line 294:
> 
> ```
> # Note: Saving CMAKE_MODULE_PATH as CGAL will overwrite it.
> # Reconsider this after CGAL 5.4: https://github.com/CGAL/cgal/pull/6029
> set(ORIGINAL_CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH})
> set(CGAL_DO_NOT_WARN_ABOUT_CMAKE_BUILD_TYPE TRUE)
> find_package(CGAL 5.0 REQUIRED COMPONENTS Core)
> message(STATUS "CGAL: ${CGAL_MAJOR_VERSION}.${CGAL_MINOR_VERSION}")
> target_compile_definitions(OpenSCAD PRIVATE ENABLE_CGAL)
> ```
> 

What that section is doing with [`CMAKE_MODULE_PATH`](https://cmake.org/cmake/help/latest/variable/CMAKE_MODULE_PATH.html), is not relevant to integrating vcpkg. 
It backs up the variable, calls some commands that might alter the value, and then restores it to the backed up value.  It is self contained in that block of code, so there's absolutely no reason to try emulating that behavior at the top of the file too.

`CMAKE_MODULE_PATH` is used by `find_package` to know where to look for cmake modules.  The thing that matters is what the value is **at the time of `find_package` call**.    So when you override it at completely different spot in the file compared to previous commits, then its going to change the behavior of all the `find_package` calls in between.

> There was some code that also changes it for an MSVC compiler around line 303:
> 
> ```
> if (MSVC)
> .
> .
> .
>   # needed for Qt5QScintilla, maybe others
>   set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/Modules/")
> else()
>   set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/Modules/")
> . 
> .
> .
> ```
> 
> after setting up the graphics libraries (the non-MSVC version sets it beforehand).
> 
> It's all rather confused and is very likely to be wrong now, but are you familiar with what the block of code in the `master` version of `CMakeLists.txt` is trying to do here?

This is **appending** OpenSCAD's own [cmake/Modules](https://github.com/openscad/openscad/tree/master/cmake/Modules) directory to the list of paths.  For MSVC / VCPKG this is done as late as possible so that the cmake modules provided by VCPKG are guaranteed to be used in all the preceding `find_package` calls.

> So, what do you advise we should with `CMAKE_MODULE_PATH`? I think we need to document a consistent way that we set it / edit it, with comments explaining what happens and why.

I would try leaving it alone and not change the behavior at all, until there is some specific need to do so.

thehans:
> @thehans re: your comment about the `glib` libraries [here](https://github.com/openscad/openscad/pull/3957#pullrequestreview-871050455)
> 
> > Use namespaced library targets
> 
> Do you know what the line should look like? 

I had to dig around vcpkg site a bit to see, but it looks like they've moved to using pkgconfig instead of a find module (here is an [example of commit source](https://github.com/microsoft/vcpkg/pull/13100/files#diff-d307033c577938e6a57e03f03d984cf3fb2082b2fba7a31454339e178cf84066) where they transitioned away from the `unofficial-glib` package name)
So apparently the preferred way should look like this now:
```
pkg_check_modules(GLIB2 glib-2.0 gobject-2.0 gmodule-2.0 gio-2.0 IMPORTED_TARGET)
target_include_directories(OpenSCAD SYSTEM PRIVATE ${GLIB2_INCLUDE_DIRS})
list(APPEND COMMON_LIBRARIES PkgConfig::GLIB2)
```
As long as that comes after finding pkgconfig itself.

dannypike:
@thehans: re: using `pkg_check_modules`:

What would you like me to do to get it to load the libraries when building in Debug mode? Windows authors sometimes have the annoying habit of putting a 'd' on the end of the filename (as well as putting it in a different subfolder). So this means that debug builds will be broken by your request to replace my hack that builds the filenames "explicitly".

For example, we need to get CMAKE to find and load a file called `[...]vcpkg\installed\x64-windows\debug\lib\qscintilla2_qt5d.lib`, instead of `[...]vcpkg\installed\x64-windows\lib\qscintilla2_qt5.lib`, which is what the current `find_package(Qt5QScintilla 2.8.0 REQUIRED QUIET)` finds.

If we don't use the debug version of the `.lib` file, the "build"phase will appear to succeed, but the "install" will not be able to find the correct DLL and it will refuse to start up without manually copying files around.

I did spend a few hours with Google, but I couldn't find any documentation on `pkg_check_modules` that made any sense to me.

kintel:
@dannypike Hey, I finally bit the bullet, purchased a PC and installed Windows, for the sole purpose of faster OpenSCAD dev cycles. I'm currently using msys2, but I could be convinced to try our Visual Studio (especially if it works in a VS Code remote ssh session). Are you still playing with this, and do you think it would be mergeable if I review and test it?

dannypike:
@kintel I haven't updated it in a while because the person who reviewed my version lost interest in helping me do it the way that they need.

So this is an "unofficial" fork.

With that, I have been quietly merging updates right up until recently, when I changed to using F360 for a project that needs a faster filleting system than OpenSCAD+minkowski can handle. However, I notice that there are conflicts in the latest commit, but I might be able to fix them up if you're keen to try it out.

My OpenSCAD needs are simple. If you need something complex (and are an expert at CMAKE/Linux), perhaps you can do a better job than me at producing a build system that TPTB like?

kintel:
PS! With the recent support for the Manifold library, some things are crazy fast with OpenSCAD now: https://gist.github.com/ochafik/70a6b15e982b7ccd5a79ff9afd99dbcf#results

But yeah, authoring build system configs is one of the hardest parts of SW development, especially for Windows, and it's very important to keep everything working, so every hack will come back and bite us really hard later.

I don't know much about Windows unfortunately, so I won't be able to drive this, but I can help test (now that I actually possess a Windows box) and suggest when a piece of infrastructure could be cleaned up and submitted separately etc., but it may indeed be a long road ahead.

roel-v:
Is anyone still interested in / working on this issue? I have a working vcpkg setup on my local machine, but that needed some changes to CMakeLists.txt, as well as some code changes (which have more to do with switching to C++20, which one of the dependencies required).

However, it's quite a gnarly vcpkg setup, since vcpkg switched to Qt6. I can manually specify to use Qt5 for Openscad, but qscintilla can be linked against both Qt5 and 6, and due to the way the vcpkg ports are arranged/configured, it's impossible to change qscintilla's dependency on Qt6. So what I had to do was configure a baseline from before vcpkg's QScintilla switched to Qt6, and then manually override the version of (most) other packages.

Anyway, the current state is that the Release build seems to work fine; the Debug build doesn't, because the FindQt5QScintilla.cmake module in Openscad doesn't make that distinction. I do have a version that does, and with which I can correctly get a debug version running, which I can contribute as a step towards getting a full Openscad VS debug build working.

My goal would be to have an easy way to build an environment on Windows to work on Openscad. In my current setup, all it takes is running a single batch file to get a working build (assuming one has Visual Studio and vcpkg installed). Keeping this working would require some sort of continued effort to keep the build in a functional state. Is there someplace where CI builds are being done? I could make one here locally if this is a useful thing, and if there is someone who is interested in this or at least in guiding me on how things work, and test on other platforms.

roel-v:
OK, I will then make my own branch to work in, since I worked off master instead of this branch (I didn't know of this branch existed when I started...), and then I can request feedback on various individual issues there. As long as I know someone is interested in providing that feedback so that I'm not wasting my time :) Thanks.

kintel:
Sounds good. I can help test stuff on a Windows box, but I'm pretty clueless about Windows. We also have a couple of other devs which potentially could give it a spin.

Also note: There is a bunch of old Visual Studio-related stuff in the codebase from older attempts at making this work. Feel free to delete/remove/rewrite as you see fit. You can probably expect all of that to be unused, and don't trust any of that to be working :)

roel-v:
OK so I've created a fork @ https://github.com/roel-v/openscad/tree/vcpkg-build . Shall I open a new issue there or here to discuss? And what would a suitable place to add some notes on how to use the fork? Sorry for noob questions but the last time I was involved in open source was back in the sourceforge days, I'm still getting the hang of git/github.

kintel:
Best to open a ticket here. Notes should eventually make it into the main README or some convenient location, but feel free to just keep it updated in the ticket description as a start.

As mentioned earlier, as your fork grows, it's generally advisable to split out separate, smaller part into individual branches and open multiple PRs. That way, we can get some progress along the way and avoid that a large PR gets stuck in review at the end.

Example of a similar ticket for tracking a body of work: #4782

Issue #4109, opened 2022-02-11
fast-csg eats fabric


rcolyer:
I tried many old designs, and fast-csg mostly performed admirably with good performance boosts.  One that it completely failed on (with just the fast-csg feature enabled and no related optionals enabled) is fabric_v2.scad, the second file on this download page:  https://www.thingiverse.com/thing:1693052/files

With the master branch and fast-csg off, it renders correctly for me in 4min 41s.  With it on, it completes in 6s without error, but the output is empty.

ochafik:
Thanks for reporting this (and cool models!!).

I can't seem to reproduce this on macosx (tried a fresh build and yesterday's nightly snapshot), it builds fine in 13s here.
Which platform did you get this on? (and did you build OpenSCAD yourself / with which CGAL version?)

ochafik:
Ok, unclear if it's the same bug but it looks like `fast-csg` eats [lizards](https://www.thingiverse.com/thing:1034488/files) (not all of them if not using `fast-csg-trust-corefinement`, to make things more interesting)

rcolyer:
I did a fresh pull and clean build, so I'm up-to-date, and it still displays nothing on render.  But I remembered you saying you were doing most of your tests from the command line, so I did a command line stl output of the same, and it succeeded!  So I went back and did the gui render, nothing displays, but the stl output from the gui still succeeds even with nothing shown on render!  So fast-csg has technically done a successful render of this, but for some reason enabling fast-csg results in the display of the render result going empty.

t-paul:
```
module base() {
    linear_extrude(5) offset(3) square([20, 8]);
}

module lid() {
    difference() {
		base();
		cube(); // <-- 
    }
}

base();
translate([0, 0, 18]) mirror([0, 0, 1]) lid();
```
Another one, disabling the `cube()` changes the behavior.

UBaer21:
```
// with lazy union activated the objects (min 2) in union below top level will vanish in render view (but are still there)
cube();  // remove this
translate([10,0])cube(2); // and this and the union will apear in render again

translate([10,10])union(){  // these vanish
  sphere(); // remove one from the union and the other re apear
  translate([10,0])cylinder();
}
```

UBaer21:
```

//translate([0,0,10])cube();  // even without this if more "groups" are in scene

translate([20,20]) {
		cube(5);
		cube(); // <-- 
}

translate([10,0])intersection() {
		cube(5);
		cube(); // <-- 
}

[...]

PR #4132, opened 2022-02-20
[fast-csg] Add missing hybrid-awareness / casts to fix top-level disappearing geom (esp. w/ lazy-union)

ochafik:
Fixes https://github.com/openscad/openscad/issues/4109 (Except for the [lizards](https://www.thingiverse.com/thing:1034488/files) 😢 )

Actually found an embarrassingly large amount of places that assumed to only find legacy nefs.
- I've renamed `createNefPolyhedronFromGeometry` to `getNefPolyhedronFromGeometry`: it now avoids a useless copy if the geom was already a nef
- Using the new helper everywhere applicable to avoid explicit type casts (kept some explicit tests where the hybrid can be used instead of converting it to a nef)
- Incidentally had to add `CGALHybridPolyhedron::isValid`
- Added tests for `lazy-union` + `fast-csg` + `fast-csg-trust-corefinement` combo, including AMF & 3MF export tests


rcolyer:
I confirm this fixes every known issue in #4109 except the lizard eating.  It is still missing the internals of the lizard with regular fast-csg, and missing most of the lizard with trust-corefinement included.  (Maybe that's properly a separate issue?)

t-paul (approved):

PR #4146, opened 2022-02-25
Experimental sort-stl feature for stable outputs (fixes #420)

### LINKED ISSUE #420: STL and AMF output should have a canonicalization option

ochafik:


src/export.cc
```diff
@@ -138,18 +138,22 @@ namespace Export {
...
+  vertices.reserve(vertexMap.size());
+
+  size_t index = 0;
   for (const auto& e : vertexMap) {
-    indexTranslationMap.emplace(e.second, index++);
+    vertices.push_back(e.first);
+    indexTranslationMap[e.second] = index++;
   }
 
   for (const auto& i : triangleIndices) {
```
ochafik, line 159: Technically, we might want to rotate the indices of the triangle *after* the remapping to ensure the first index comes first, in case some mesh algorithm rotates our faces arbitrarily. Can do that later if that comes to be an issue.

PR #4153, opened 2022-03-03
Fix various typos

luzpaz:
Found via `codespell -q 3 -S *.po,./src/ext,./libraries/MCAD,./submodules/mimalloc -L childs,clen,nd,synopsys,tesselate,tesselation,tesselator,uncomplete,vie`

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -439,7 +439,7 @@ endfunction()
 function(add_failing_test TESTCMD_BASENAME)
   cmake_parse_arguments(TESTCMD "" "RETVAL;EXE;SCRIPT;SUFFIX" "FILES;ARGS" ${ARGN})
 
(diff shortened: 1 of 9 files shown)


tests/CMakeLists.txt
```diff
@@ -439,7 +439,7 @@ endfunction()
 function(add_failing_test TESTCMD_BASENAME)
   cmake_parse_arguments(TESTCMD "" "RETVAL;EXE;SCRIPT;SUFFIX" "FILES;ARGS" ${ARGN})
 
-  if ("${TESTCMD_SUFFIX}" STREQUAL "") # Suffix "off" counts as a falsy value, so check diectly for empty string.
```
t-paul, line 442: I think the "[falsy](https://en.wiktionary.org/wiki/falsy)" is not actually a typo, but I guess we can keep "false" as it's clear in this context.

t-paul (approved):
Nice! Thank you!

PR #4162, opened 2022-03-07
Remove -pthread for WASM build

DSchroer:
Threading support has been dropped from the WASM build. This makes it much easier to consume and with that this flag can be removed. 

The flag does not break the new build but its not needed anymore.


t-paul:
Oh, nice, so that means the scary "everything will be slow" warning will be gone too?

I'll trigger a rebuild of the docker base image...

t-paul:
Is it possible to move the `npm` processing into the build container? It would be nice to keep the external dependencies to a minimum. I'll push the current state anyway, that should still fix the CI-build.

DSchroer:
You can avoid using `npm` by using the specific make rules for `build/openscad.js` rather than `build`. If you need a work around for now.

t-paul:
I've pushed the new base docker container, could you update the `.circleci/config.yml` file with the modified `em++` flags too? That should hopefully fix the build.

t-paul (approved):

PR #4163, opened 2022-03-07
Support performance timings of heavy operations

ochafik:
Now that there are tests being done comparing our CGAL (corefinement) rendering with alternatives (e.g. [against exciting new Manifold library](https://elalish.blogspot.com/2022/03/manifold-performance.html)), it's time to look at the details of the timings :-)

(reused code from https://github.com/openscad/openscad/pull/3641)

To build for performance comparisons
```bash
git clone https://github.com/ochafik/openscad.git --single-branch -b perf-timings openscad-perf-timings
cd openscad-perf-timings
git submodule init && git submodule update

mkdir build && cd build && cmake .. \
  -DCMAKE_BUILD_TYPE=Release -DEXPERIMENTAL=ON -DPERFORMANCE_TIMINGS=ON
```

I tried to match the testing conditions mentioned by the author of the [Manifold library](https://github.com/elalish/manifold) in [his post](https://elalish.blogspot.com/2022/03/manifold-performance.html):

```js
$fn=320; // Gives ~101k faces per sphere
// $fn=1000 // Gives ~1M faces per sphere
// $fn=3200 // Gives ~10M faces per sphere

difference() {
  sphere();
  translate([0.5, 0, 0]) sphere();
}
```

Tested with:
```bash
# MacOS:
# ln -s ./build/OpenSCAD.app/Contents/MacOS/OpenSCAD openscad
 
time ./openscad bigspheres.scad --enable=fast-csg --enable=fast-csg-trust-corefinement \
    -o diff.off '-D$fn=1000'

# STL is just so much slower
```

Results:

| $fn | # faces in each sphere | Total time (STL) | Total time (OFF) | PolySet -> hybrid | Corefinement Difference time |
| -- | -- | -- | -- | -- | -- |
| 100 | ~10k | 0.3sec | 0.26sec | 0.01sec | 0.08sec |
| 320 | ~101k | 1.5sec | 0.9sec | 0.06sec | 0.35sec |
| 1000 | ~1M | 13.9sec | 7.5sec | 0.63sec | 3.3sec |
| 3200 | ~10M | 173sec | 87sec | 7.7sec | 38.7sec |

I need to instrument imports & exports to understand where the most time is spent, but it seems the underlying union time in corefinement is more comparable to the (*mono-threaded CPU*) figures published for Manifold at the 100k and 1M faces.

@elalish could you share your raw figures and methodology?

### DIFF
--- a/src/scoped_timer.h
+++ b/src/scoped_timer.h
@@ -1,6 +1,7 @@
 #pragma once
 
-class ScopedTimer {
+class ScopedTimer
+{
 public:
   ScopedTimer(const std::string& name) : name(name) {
     t.start();
--- a/src/scoped_timer.h
+++ b/src/scoped_timer.h
@@ -1,0 +1,25 @@
+#pragma once
+
+class ScopedTimer {
+public:
+  ScopedTimer(const std::string& name) : name(name) {
+    t.start();
+  }
+  ~ScopedTimer() {
+    t.stop();
+    printf("[%s] time %f s\n", name.c_str(), t.time());
+  }
+private:
+  std::string name;
+  CGAL::Timer t;
+};
+
+#ifndef PERFORMANCE_TIMINGS
+#define PERFORMANCE_TIMINGS 1
+#endif
+
+#if PERFORMANCE_TIMINGS
+#define SCOPED_PERFORMANCE_TIMER(name) ScopedTimer timer(name)
+#else
+#define SCOPED_PERFORMANCE_TIMER(name)
+#endif
--- a/src/scoped_timer.h
+++ b/src/scoped_timer.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <CGAL/Real_timer.h>
+
 class ScopedTimer
 {
 public:
@@ -12,7 +14,7 @@ public:
   }
 private:
   std::string name;
-  CGAL::Timer t;
+  CGAL::Real_timer t;
 };
 
 #ifndef PERFORMANCE_TIMINGS
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,6 +60,7 @@ option(PROFILE "Enable compiling with profiling / test coverage instrumentation"
 option(MXECROSS "Enable setup for MXE cross platform build" OFF)
 option(OFFLINE_DOCS "Download Documentation for offline usage" OFF)
 option(USE_MIMALLOC "Use mimalloc as malloc replacement." ON)
+option(PERFORMANCE_TIMINGS "Print time spent in various costly CSG operations." OFF)
 
 if(APPLE)
   option(APPLE_UNIX "Build OpenSCAD in Unix mode in MaxOS X instead of an Apple Bundle" OFF)
@@ -83,6 +84,10 @@ if(WASM)
   set(ENV{PKG_CONFIG_PATH} "/emsdk/upstream/emscripten/cache/sysroot/lib/pkgconfig")
   target_compile_options(OpenSCAD PRIVATE "-pthread")
   target_compile_definitions(OpenSCAD PRIVATE CGAL_DISABLE_ROUNDING_MATH_CHECK)
+endif()
+
+if(PERFORMANCE_TIMINGS)
+  target_compile_definitions(OpenSCAD PRIVATE PERFORMANCE_TIMINGS)
 endif()
 
 if(NULLGL)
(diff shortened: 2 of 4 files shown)


ochafik:
@sloriot FYI first benchmarks of your corefinement routines against fresh new [Manifold](https://github.com/elalish/manifold) library (see [post](https://elalish.blogspot.com/2022/03/manifold-performance.html)).

ochafik:
Also, ported one of Manifold's microbenchmarks to CGAL in https://github.com/elalish/manifold/pull/84

elalish:
@ochafik Looks like you've made a much better comparison than I did. I was running OpenSCAD's GUI, did an F5 render to warm things up, then F6 and recorded the "total rendering time" it conveniently displays. Of course I don't know exactly what this means, but considering the large difference between F5 and F6, I knew it didn't contain export and import was probably pretty trivial. Anyway, here's my [spreadsheet](https://docs.google.com/spreadsheets/d/11Vj3zJnoQ7W5-6SNOv7lIL36bJebtqDldJjIvTb7GxU/edit?usp=sharing).

If you want to benchmark the sponge too, I used this SCAD code:
```
difference(){
    cube(center=true);
    holes();
    rotate([90,0,0])holes();
    rotate([0,90,0])holes();
}

module holes()
import("mengerHoles3.stl");
```
And I exported "mengerHolesX.stl" from this line, to avoid a bunch of lazy unions that last I checked were still not lazy in OpenSCAD: https://github.com/elalish/manifold/blob/master/samples/src/menger_sponge.cpp#L50

How do these libraries compare performance-wise on your machine?

sloriot:
Hey, thanks for the notification. Very interesting. You are using EPECK as input kernel while Manifold uses double, right? I wonder what difference it makes if we are also using double as input with corefinement. Also do you know if Manifold is taking advantage of several threads?

ochafik:
@sloriot I’ve tested Epick and Epeck against manifold in a very limited microbenchmark here, Manifold monothreaded cpu mode seems >2x faster than coref on epick for Union: https://github.com/elalish/manifold/pull/84

(Not sure corefinement would accept a simpler ‘Cartesian<double>’?)

Manifold can leverage CUDA and OpenMP but I haven’t tested either, stuck to mono-threaded CPU comparisons for now (unfair to Manifold haha) but @elalish seems to say we can expect at least 2x speedups, even before he proceeds to optimizing more steps in his library.

Could corefinement use TBB to parallelize some operations btw? Looks like Epeck is rather thread safe from reading https://github.com/CGAL/cgal/issues/6141.

sloriot:
We could indeed use TBB to process intersections and retriangulation but I never spent time on it as no one asked for it. I saw the bench. It is strange that difference is faster but not union. One possibility could be that union need to import more triangles than difference. If that the case, then we might be impacted by the generic import rather than using something dedicated to `Surface_mesh`.

ochafik:
> running OpenSCAD's GUI, did an F5 render to warm things up, then F6 and recorded the "total rendering time" it conveniently displays. Of course I don't know exactly what this means, but considering the large difference between F5 and F6, I knew it didn't contain export and import was probably pretty trivial

@elalish This should indeed skip import/export, but may still have some conversions I've left to avoid reimplementing rendering logic for fast-csg's hybrid polyhedron's data structures (which can be either Surface_mesh or a differently-kernelled Nef than the one already supported in the renderer). Given how big the meshes are, even just copies to the cache might take time here.

src/scoped_timer.h
```diff
@@ -0,0 +1,26 @@
...
+  ScopedTimer(const std::string& name) : name(name) {
+    t.start();
+  }
+  ~ScopedTimer() {
+    t.stop();
+    printf("[%s] time %f s\n", name.c_str(), t.time());
+  }
+private:
+  std::string name;
+  CGAL::Timer t;
```
thehans, line 15: Is this really the timer you want?
https://doc.cgal.org/latest/Miscellany/classCGAL_1_1Timer.html

> When used to profile a multithreaded algorithm, this timer will add up the process time of each thread, which may be unexpected. To compute the wall-clock time of an algorithm, please consider using [CGAL::Real_timer](https://doc.cgal.org/latest/Miscellany/classCGAL_1_1Real__timer.html).

ochafik: Ooops indeed, fixed! I hope the different matters soon, maybe if/when corefinement becomes super parallel (@sloriot ;-))

PR #4167, opened 2022-03-12
adding support for gradient backgrounds

MichaelPFrey:
I mainly use openscad for animations. This merge request adds the ability to define a color gradient background.

_Background Color_ in the color scheme is used as the starting color of the gradient.
_Background Stop Color_ in the color scheme is used as the stop color of the gradient.

The code only applies if a _background stop color_ is applied and that color is different from the _background color_.

The _background stop color_ in the color scheme is optional. Old Color schemes still work without change.

Provided is a color scheme I named "clearsky", which is essentially cornfield, just with a light blue to light blue gradient:

LetterBlock1

The effect is subtle, so here is a blue to magenta gradient for demonstration:

LetterBlock3

An other example inspired by an other cad software:

LetterBlock2

I would not want to change legacy color schemes - having solid colors does have benefits.

Developing new color schemes is something I want to leave to people that have a better grasp at design and colors.

What could be discussed if a two tone gradient is enough. It would be easy to add a third color, to allow for finer control of the gradient.

It would also be possible to add control to four corner colors:
```
  if(bgcol != bgstopcol){
    glDisable(GL_DEPTH_TEST);
    glMatrixMode(GL_PROJECTION);
    glLoadIdentity();

    //draw screen aligned quad with color gradient 
    glBegin(GL_QUADS);  
    glColor3f(1, 0, 0);  
    glVertex2f(-1.0f, +1.0f);

    glColor3f(0, 1, 0);  
    glVertex2f(+1.0f, +1.0f);

    glColor3f(0, 0, 1);  
    glVertex2f(+1.0f, -1.0f);
    
    glColor3f(1, 1, 1);  
    glVertex2f(-1.0f, -1.0f);
    glEnd();  
    glEnable(GL_DEPTH_TEST);
  }
```
LetterBlockRGB
This is an extreme example, more practically this could be used to have left to right gradients and/or subtle diagonal gradients.

I personally think that a top to bottom gradient is enough, but when there is a good use case, it would be easy to make it more flexible.


t-paul (approved):
Looks great. I agree it's probably ok to go with just the horizontal gradient. I don't remember seeing a diagonal one.

I suppose if we ever go with a more complex background, it may need some more complex setup too, not just the 3 colors.

MichaelAtOz:
In the future, `$vpcolor...` could allow some fancy animation. The model zooms into the distance as the background changes to sunset then night, and you finish off with some stars twinkling in the distance...

PR #4168, opened 2022-03-12
only process syntactically relevant brackets for folding (fixes #4144)

### LINKED ISSUE #4144: brackets in comments are parsed for collapsible block
version 2022.02.18

- Issue

following code will collapse the union only till the comment

```
union(){
// parsed }
// inside [ will collapse the curly brace from the union
}
```

- Feature Request
collapse All/expand All  (show All / hide All) for bracket sections

MichaelPFrey:
this fixes #4144
```
union(){
// parsed }
// inside [ will collapse the curly brace from the union
}
```
```
str1 = "Hallo { ";

cube();

str2 = "} Welt";
```

this has one side effect worth noting:
currently brackets within commented out section can be folded:
```
/*
if(true){
    while(true){
        cube();
    }
}
*/
```
This fix sees this whole comment block as one fold-able comment, ignoring the brackets within.

I could expand the code so that the code makes a "best effort" of matching brackets within the comment, but given that comments are by no means guaranteed to be well formatted, commented out code, this seams more trouble then it is worth.

**Todo:** I will to some more testing and at screenshots for documentation.

### DIFF
--- a/src/gui/ScadLexer.cc
+++ b/src/gui/ScadLexer.cc
@@ -261,9 +261,9 @@ void ScadLexer2::fold(int start, int end)
     bool style = (editor()->SendScintilla(QsciScintilla::SCI_GETSTYLEAT, i - 1) == Comment);
     bool startstyle = (editor()->SendScintilla(QsciScintilla::SCI_GETSTYLEAT, i) == Comment);
 
-    if ((ch == '{') || (ch == '[') || ((!style) && (startstyle))) {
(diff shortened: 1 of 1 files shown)


MichaelPFrey:
fold1
fold2
fold3
fold4

MichaelPFrey:
ignoring brackets in the comments is not uncommon.

Geany for e.g. also does not offer folding for commented out code:

geany

t-paul (approved):
Nice! Thank you.

Issue #4166, opened 2022-03-12
Fast CSG render without Preview missing face from quad


UBaer21:
// F5 F6  differs from  (flush cache)  F6 without F5 (no preview)
// v. 2022 02 18

```
fn=8;
r=[5,3];

p=[
for(i=[0:fn-1])[sin(360/fn*i)*r[i%2],cos(360/fn*i)*r[i%2],0],
for(i=[0:fn-1])[sin(360/fn*i)*r[i%2],cos(360/fn*i)*r[i%2],10],
];


f=[ // non planar quad
[for(i=[fn-1:-1:0])i],
[for(i=[fn:1:fn*2-1])i],
for(i=[0:fn -1],i=i%fn)[i,i+1,i+fn,i+fn -1]
];


f2=[ // planar quad
[for(i=[fn-1:-1:0])i],
[for(i=[fn:1:fn*2-1])i],
for(i=[0:fn -1])[i,(i +1)%fn,(i+1)%fn+fn,i+fn]
];



union(){
polyhedron(p,f2); // planar quad no issue
translate([20,0,0])polyhedron(p,f); // non planar 
}
```

MichaelAtOz:
The second face is wrong. `ECHO: "f[2]=[0, 1, 8, 7]"`
I have not checked the others.
```
fn=8;
r=[5,3];

p=[
for(i=[0:fn-1])[sin(360/fn*i)*r[i%2],cos(360/fn*i)*r[i%2],0],
for(i=[0:fn-1])[sin(360/fn*i)*r[i%2],cos(360/fn*i)*r[i%2],10],
];


f=[
[for(i=[fn-1:-1:0])i],
[for(i=[fn:1:fn*2-1])i],
for(i=[0:fn -1],i=i%fn) [i,i+1,i+fn,i+fn -1]
];

for (i=[0:len(f)-1]) echo(str("f[",i,"]=",f[i]));
echo();
for (i=[0:len(p)-1]) echo(str("p[",i,"]=",p[i]));

union(){
polyhedron(p,f);
cube();
}

polydebug(p, echoPoints=false);

module polydebug(p, echoPoints=true) {
  
  for( i=[0:len(p)-1]) {
    if (echoPoints) echo(i,p[i]);
    translate(p[i]) {
      color("black") sphere(0.2, $fn=12);
      translate([-0.5,0.5,( (p[i].z>0) ? 0.2 : -0.4) ]) 
        color("red") 
            linear_extrude(height=0.2) 
[...]

Issue #4148, opened 2022-02-27
fast-CSG fails without preview data


UBaer21:
```
Test(); // render has no preview data - due to random generation - ends in error "mesh not closed"
   
module Test(i=0,t=5,a=90){
    a2=rands(10,120,1)[0];
    t2=rands(-15,15,1)[0];
    t22=abs(t2)<3?5:t2;
    dir=t22>0?0:180;
    t3=abs(t22);

    if(i<6)rotate(a)translate([t-t3,0])rotate([0,dir])Test(i=i+1,t=t22,a=a2*sign(t22));
    rotate_extrude(angle=a)translate([t,0])circle(3);
}
```

kintel:
Closing as fast-csg has been deprecated

PR #4170, opened 2022-03-12
Better triangulation for fast-csg inputs + big hull improvements for all

ochafik:
Reuse most of the logic from `createNefPolyhedronFromPolySet` in `createHybridPolyhedronFromPolySet` (grid quantization, nearly-coplanar-accepting triangulation, hull for convex polys) for a much better backwards-compatibility of `fast-csg`.

This fixes https://github.com/openscad/openscad/issues/4166 and https://github.com/openscad/openscad/issues/4148, along with many `fast-csg` tests I was skipping (the ones I thought were about bad face orientations or something).

And while we're at it, speed up hull (both in these PolySet -> Nef/Hybrid conversion functions and in CGALUtils::applyHull) by giving it a list of unique vertices (in the case of the conversion functions, added an optional points sink parameter to `PolySet::quantizeVertices` since it can already give us the deduped points for free).

This silly example goes from 15sec to 2.5 sec, so a 6x speedup (no `fast-csg` involved here):

```js
$fn=320;
// This hull now has 102k points to process instead of 661k
hull() {
  sphere();
  translate([0.5, 0, 0]) sphere();
}
```

(a more complex model doing [lots of hulls here](https://gist.github.com/ochafik/6d19e3b131153305b44ea275d5d930e3) shows a more modest 20% overall speedup for `-DN=1`)

As a result I've taken the step to enable `fast-csg`, `fast-csg-remesh` and `fast-csg-trust-corefinement` on all tests except the known failures / different expectations.

**Note**: as a downside, conversion from PolySet to Hybrid mesh is a bit slower (the above example with `union` instead of `hull` with `fast-csg` is 1.6x slower), but robustness of imports seems more important here.


t-paul (approved):

MichaelAtOz:
Is it normal that the artifacts are dated 2022.03.13? It being 15th here, but minimum 14th 12:58 elsewhere e.g. [near date line](https://time.is/U.S._Minor_Outlying_Islands).

t-paul:
Yes, last commit to the branch was 13th: https://github.com/openscad/openscad/pull/4170/commits/a9dd531c9152cd9270bb424bfd49b8ba42f1fd60

The builds from the merge (happening right now) will not show up here, but only in the master branch commit list.

Issue #4172, opened 2022-03-13
using echo on a vector can crash openscad


MichaelPFrey:
```
rec();

module rec(a=1,b=6000)
{
 if(b==1) echo(a);
 if(b>0)
  rec([a,10,10],b-1);
}
```
b=6000 has to be adjusted. Alternatively, ` if(b==1) ` can simply be removed.

This issue is for documentation and as justification of a side effect of https://github.com/openscad/openscad/pull/4173

MichaelPFrey:
I am working on a test case for the fix, so I am working in the command line with the quite option.

The quite option does two things:
   * it makes the run time much faster
   * the output is more stable

interestingly, in the command line, this code does trigger the recursion exception:
```
rec();

module rec(a=1,b=6000)
{
 if(b==1) echo(a);
 if(b>0)
  rec([a,10,10],b-1);
}
```
```
openscad 'issue4172-echo-vector-stack-exhaust-1.scad' -o 1.png -q
ERROR: Recursion detected calling module 'rec' in file issue4172-echo-vector-stack-exhaust-1.scad, line 12
echo $?
0
```

while this this triggers a segmentation fault:
```
rec();

module rec(a=1)
{
  echo(a);
  rec(
   [[[[[[[[[[[[
   [[[[[[[[[[[[
   [[[[[[[[[[[[
   [[[[[[[[[[[[
   [[[[[[[[[[[[
   [[[[[[[[[[[[
   [[[[[[[[[[[[
   [[[[[[[[[[[[
   [[[[[[[[[[[[
   [[[[[[[[[[[[
   [[[[[[[[[[[[
   [[[[[[[[[[[[
    a
   ]]]]]]]]]]]]
   ]]]]]]]]]]]]
   ]]]]]]]]]]]]
   ]]]]]]]]]]]]
   ]]]]]]]]]]]]
   ]]]]]]]]]]]]
   ]]]]]]]]]]]]
   ]]]]]]]]]]]]
   ]]]]]]]]]]]]
   ]]]]]]]]]]]]
   ]]]]]]]]]]]]
   ]]]]]]]]]]]]
   );
}

```
```
openscad 'issue4172-echo-vector-stack-exhaust-2.scad' -o 1.png -q
Segmentation fault (core dumped)
echo $?
139
```

in UI mode, both code crashes OpenSCAD.

PR #4173, opened 2022-03-13
WIP: show usermodule parameters in trace, add stack check for vectors to toEchoString()

MichaelPFrey:
The idea _show arguments in trace_ is inspired by a mailing list [discussion](https://lists.openscad.org/empathy/thread/G2O3X6JB5SUY34HTERKMSM32BHA2AKS6).

The change in toEchoString() coincidentally fixes a crash: https://github.com/openscad/openscad/issues/4172

`Value::toEchoString(const tostring_visitor *visitor) ` seams tobe unused.

I am open for input.

Todo:
- [ ] add option to disable trace?
   - could be useful for some automated tests to stabilize the output
- [ ] add option to disable show argument?
   - could be useful for some automated tests to stabilize the output
- [ ] manual testing
- [ ] some screenshots where it is useful for demonstration
- [ ] improve comments
- [ ] regenerate tests as necessary
- [ ] minimize diff

PR #4182, opened 2022-03-19
Stack usage optimization for tostream_visitor

thehans:
This change was tested against the sample code in #4172, and while it does not fix the crash, it delays the crash until higher values of `b`, by going through fewer function calls, and re-using `tostream_visitor` instead of constructing new ones for each nested vector. 

My testing showed the lowest value to trigger a crash went up by about 1000 after changes.
Lowest b value required to trigger a crash, built with `g++-10`
```
RelWithDebInfo build:
  unedited master: b=5390
  with changes:    b=6378
Release build:
  unedited master: b=4936
  with changes:    b=6081
Debug build:
  unedited master: b=3465
  with changes:    b=3905
```

Also removed some `Value::toString|toStream` functions which were not being used.

### DIFF
--- a/src/core/Value.cc
+++ b/src/core/Value.cc
@@ -291,6 +291,68 @@ const str_utf8_wrapper& Value::toStrUtf8Wrapper() const {
   return boost::get<str_utf8_wrapper>(this->value);
 }
 
+// Optimization to avoid multiple stream instantiations and copies to str for long vectors.
+// Functions identically to "class tostring_visitor", except outputting to stream and not returning strings
+class tostream_visitor : public boost::static_visitor<>
+{
+public:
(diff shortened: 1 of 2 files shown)


MichaelPFrey:
#4173 contains a similar approach partially implemented. This implementation is a lot cleaner.

As I want to discard #4173 anyway to get a cleaner implementation, merging this here is welcome by me.

I think fixing #4172 (_add stack check for vectors to toEchoString()_) will also be its own merge request (before _show usermodule parameters in trace_).

Splitting it up makes things a lot easier.

thehans:
Thanks for looking it over, I was hoping you would see this since it affects some of the related code you were working on, and alters the required b value for the test case by a significant amount.

PR #4183, opened 2022-03-19
Remove unused qscintilla-devel build dependency for fedora

Lukas-Heiligenbrunner:
Remove unused (and non existing in package archive) fedora build dependency: qscintilla-devel
Screenshot_20220319_144752

Openscad builds just fine without this dependency and fails while trying to install this non-existing package. 
qscintilla-qt5-devel should include all the neccessary files.


t-paul (approved):
Thanks!

PR #4185, opened 2022-03-20
Fix error handling when creating an offscreen OpenGL context on Windows.

### LINKED ISSUE #4180: Animation from command line fails on the second frame (Windows)
On Windows, when I give this command line:

```
>openscad --animate=30 test.scad
```

the first frame is generated, but the second fails with:

```
MS GDI - RegisterClass failed
last-error code: 1410
Can't create OpenGL OffscreenView. Code: -1.
```

Error code 1410 is "Class already exists."  (The details of test.scad don't seem relevant.)

I believe the relevant code is here:  https://github.com/openscad/openscad/blob/3b6312b819d29df4329ec187b3f936ea3af21277/src/glview/OffscreenContextWGL.cc#L108

When the second frame begins, it tries to register the window class again, which fails because there's already a class registered with that name.  (I would have expected the second call to RegisterClassW to return the original atom.  Alas, that doesn't seem to be the case.)

The code doesn't use the atom returned by RegisterClassW for anything other than the error check--the window class is referenced by name in the CreateWindowW call.  So the simplest fix would be to ignore the return value of RegisterClassW.  If it fails for a legitimate reason, the subsequent CreateWindowW call will fail.  If it fails because the class already exists, the CreateWindowW call will succeed and all will be well.

Other possible fixes would be to adjust the control flow to create the class just once or even to create the class and the window once and then re-use the window for each frame.  That might require persisting some state in a way that doesn't fit well with the current cross-platform abstractions.

Finally, the code could unregister the class after the window is deleted, so the next time it's registered it won't be a duplicate.

aidtopia:
Fixes #4180

When animating from the command line interface, the code registers the
window class on each frame. This worked for the first frame, but the
registration for the second frame seemed to fail. Actually,
RegisterClassW was merely indicating that the second registration was
redundant.

This change ignores ERROR_CLASS_ALREADY_EXISTS, which seems to be the
least intrusive way to solve the problem.

I tested the solution in a proof-of-concept program, but I haven't tested
this actual patch because building OpenSCAD for Windows is onerous.

(There will probably still be bugs here.  It looks like the code never
destroys the windows it creates, and it creates a new one for each frame.
And each of those windows has its own DC because of the CS_OWNDC in the
window class.  So a many-frame animation is going to burn up GDI
resources.  Those lingering issues will be easier to detect and debug
once we can get past the initialization issues.)

### DIFF
--- a/src/glview/OffscreenContextWGL.cc
+++ b/src/glview/OffscreenContextWGL.cc
@@ -97,6 +97,7 @@ bool create_wgl_dummy_context(OffscreenContext& ctx)
   //  and ctx->dev_context if successful
 
   // create window
+  LPCWSTR lpClassName = L"OpenSCAD";
 
   HINSTANCE inst = GetModuleHandleW(0);
   WNDCLASSW wc;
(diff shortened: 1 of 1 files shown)


t-paul (approved):
Thanks!

PR #4186, opened 2022-03-21
add stack check to vector to echostring

### LINKED ISSUE #4172: using echo on a vector can crash openscad

MichaelPFrey:
fixes #4172 and is based on #4182


src/core/Value.cc
```diff
@@ -328,6 +329,10 @@ class tostream_visitor : public boost::static_visitor<>
   }
 
   void operator()(const VectorType& v) const {
+    if (StackCheck::inst().check()) { 
+      LOG(message_group::Error, Location::NONE, "", "Stack exhausted while trying to convert a vector to EchoString");
```
thehans, line 333: The LOG should not be done immediately since, particularly in Qt GUI, this process can add a lot to the stack.  It should be deferred to somewhere up the stack after catching the exception somewhere.

I'm also wondering where the following exception ends up getting caught, and wouldn't that cause a duplicated message?

I would suggest to just throw the exception here, without the LOG call.  
Then try/catch for it inside the **`tostring_visitor`**'s  `std::string operator()(const VectorType& v) const`, with the LOG call after catching it.

Basically `tostream_visitor` only is called through `tostring_visitor`'s vector handling function, so catching this will ensure that the exception will pop off the stack a good amount (to the top level) before attempting to print the log message.

MichaelPFrey: I will respond in detail.
The response in short is, that I used `RecursionException` as a design pattern.

MichaelPFrey:
> I'm also wondering where the following exception ends up getting caught, and wouldn't that cause a duplicated message?

It is a `EvaluationException`
https://github.com/openscad/openscad/blob/e34e756d264800b701c07414ceffd9dd768cdf84/src/core/SourceFile.cc#L182-L185

The history is a bit tricky to track down due to multiple renames.

https://github.com/openscad/openscad/commits/06711a4f74db7ac70c04460892570cd3b06134f6/src/SourceFile.cc

https://github.com/openscad/openscad/commits/82730fc5752ea1f9313c4803ae3b1daf85397d22/src/FileModule.cc

From my memory:
Originally - the artifact is still there - the error message was generated there.

The intention was then to add location to all errors and warnings.
"The obvious" implementation was to move the message generation to where we have the location available.
(The alternative would have been to add the location to the exception)

Next, I implemented Trace. Trace catches and re-throws `EvaluationException`
https://github.com/openscad/openscad/blob/e34e756d264800b701c07414ceffd9dd768cdf84/src/core/ModuleInstantiation.cc#L82-L88
https://github.com/openscad/openscad/blob/e34e756d264800b701c07414ceffd9dd768cdf84/src/core/Expression.cc#L600-L606

There, it also made sense that all error and warnings (see hardwarning/stop on first warning) Log/Print their message. That way, the error/warning is comes before of the unrolling call stack/trace.

> I would suggest to just throw the exception here, without the LOG call. Then try/catch for it inside the **`tostring_visitor`**'s `std::string operator()(const VectorType& v) const`, with the LOG call after catching it.
> 
> Basically `tostream_visitor` only is called through `tostring_visitor`'s vector handling function, so catching this will ensure that the exception will pop off the stack a good amount (to the top level) before attempting to print the log message.

This is an interesting idea. Only thing I have to add is, that the exception has to be re-thrown.

I will look into implementing this suggestion.

src/core/Value.cc
```diff
@@ -389,7 +393,12 @@ class tostring_visitor : public boost::static_visitor<std::string>
   std::string operator()(const VectorType& v) const {
     // Create a single stream and pass reference to it for list elements for optimization.
     std::ostringstream stream;
-    (tostream_visitor(stream))(v);
+    try {
+      (tostream_visitor(stream))(v);
+    } catch (EvaluationException& e) {
+      LOG(message_group::Error, Location::NONE, "", "Stack exhausted while trying to convert a vector to EchoString");
```
thehans, line 399: I would just use `e.what()` here for the string, so that the message from the exception constructor is not needlessly duplicated / unused.

src/utils/exceptions.h
```diff
@@ -54,6 +54,18 @@ class LoopCntException : public EvaluationException
   LoopCntException(const std::string& what_arg, const Location& loc) : EvaluationException(what_arg), loc(loc) {}
 };
 
+class VectorEchoStringException : public EvaluationException
+{
+public:
+  static VectorEchoStringException create() {
+    return VectorEchoStringException{STR("ERROR: Stack exhausted while trying to convert a vector to EchoString")};
```
thehans, line 61: Small nitpick, but the `STR` macro is a helper that enables inline usage of stream operators `<<` to compose a string from variables etc.  Its not needed if there's no `<<` operators and just a single const string.

thehans:
@MichaelPFrey Looks good to me, I just added 2 minor notes.  Thanks for working on this.

MichaelPFrey:
Thank you for the feedback.
I will implement the feedback.

PR #4187, opened 2022-03-22
Missing ImageMagick dependency on Arch

tleb:
Nothing much to add. Great script by the way, a bit surprising that it automatically installs packages though. I did not expect half a GB to download without confirmation.


t-paul (approved):
Thanks!

PR #4191, opened 2022-03-27
Remove obsolete QGLWidget preprocessor checks etc.

thehans:
Only QOpenGLWidget should be used now.


MichaelPFrey:
For reference:

[qopenglwidget](https://doc.qt.io/qt-5/qopenglwidget.html):
> This class was introduced in Qt 5.4.

[Relation to QGLWidget](https://doc.qt.io/qt-5/qopenglwidget.html#relation-to-qglwidget)

t-paul (approved):
Yeah, I think we accidentally updated the Windows builds to `QOpenGLWidget` when switching to `cmake` and it just worked now with the newer Qt versions. I'd say lets go with that.

PR #4193, opened 2022-03-27
Fix python 'testing equality to None' issue

luzpaz:
"Testing whether an object is 'None' using the == operator is inefficient and potentially incorrect."


t-paul (approved):
:+1:

PR #4196, opened 2022-03-30
Move html,sound,icon and .qrc files under /resources path.

thehans:
A few notes about these changes:
- Since our CI tests aren't really using the interactive GUI, I hope we can get a few people to test and look over these changes before merging.
    - Particularly on Mac, where I separated out two image resources (originally from #3970) into mac-specific qrc file.  Those two images were particularly large compared to the rest, accounting for 1.3 MB or so.
 - I'm not sure if this change requires any additional changes to packaging scripts, configs, etc, including @t-paul's OBS builds.
- I also added some RCC options in CMakeLists.txt, which should compress resources more agressively than default.  This change in conjunction with the 2 removed mac images made my Linux release build nearly 3 MB smaller.
- Since `openscad.qrc` file has been moved to `/resources/common.qrc`, I had to change a number of bash scripts which were using that file to detect if running from the top directory.  I chose `openscad.appdata.xml.in`, which seemed as good as any of the other options.
- Paths to resources within .qrc files are relative, so no longer require "resources/" prefix within the qrc, nor when referencing from Qt function calls in C++ source.


thehans:
I've created a separate branch in order to build a mac snapshot, as suggested by @t-paul in IRC.
https://github.com/openscad/openscad/tree/resources-macos

MichaelPFrey:
I build and went trough the dialogs under Ubuntu 21.10.
It builds like normal and looks normal.

t-paul (approved):

PR #4197, opened 2022-03-31
Add wavefront obj export

emogenet:
Hi everyone,

Long time OpenSCAD user, and I have lots of command line tools that consume wavefront OBJ file format as input.

Right now, I have to export to STL / OFF / ... and use another tool (eg meshlab) to get an OBJ, which complexifies my workflow. 

So I bit the bullet and added an OBJ exporter. to OpenSCAD.

Please consider adding the feature.


src/gui/MainWindow.cc
```diff
@@ -2017,6 +2021,8 @@ void MainWindow::sendToOctoPrint()
   const QString fileFormat = QString::fromStdString(Settings::Settings::octoPrintFileFormat.value());
   FileFormat exportFileFormat{FileFormat::STL};
   if (fileFormat == "OFF") {
+    exportFileFormat = FileFormat::OBJ;
+  } else if (fileFormat == "OFF") {
     exportFileFormat = FileFormat::OFF;
```
t-paul, line 2026: Typo, one shoudl be `OBJ`.

t-paul (changes requested):
Looks good.

Would you mind adding a simple test case along the lines of this one:
e72b6c412724153a86ca28bd9956fb3636cc0edc

See https://github.com/openscad/openscad/blob/master/doc/testing.txt for some details about the tests. Section "Adding a new test" explains how to generate the reference file automatically.

src/gui/MainWindow.cc
emogenet: typo fixed
tests added
see 4 commits below

t-paul (approved):
Thanks!

PR #4198, opened 2022-04-02
add trace to assignment, store location of original assigment and  assigment overwrite

MichaelPFrey:
**Background:**
Based on a mailing list discussion.

We have the feature, that assignments from includes can be overwritten in the main file without a warning.

The counter intuitive thing is, that they are evaluated  in the location of the original assignment,
meaning variables declared after the original assignment, but before the overwrite are unknown during evaluation.
This can cause warnings that are hard to understand.

**Example:**

inc.scad
```
a = "test"; //original assignment
```
main.scad
```
include <inc.scad>
b = true; //variable declared after the original assignment, but before the overwrite 
a = assert(b); //overwrite
```

Current error message:
```
WARNING: Ignoring unknown variable 'b' in file main.scad, line 3
ERROR: Assertion 'b' failed in file main.scad, line 3
```

Improved error message:
```
WARNING: Ignoring unknown variable 'b' in file main.scad, line 3
ERROR: Assertion 'b' failed in file main.scad, line 3
TRACE: overwritten assignment to 'a' (this is where the assignment is evaluated) in file inc.scad, line 1
TRACE: overwriting assignment to 'a' in file main.scad, line 3
```

Note that the solution to the problem - moving the include after the deceleration of b - still is not obvious.

main-fixed.scad
```
b = true; //variable declared before the original assignment
include <inc.scad>
a = assert(b); //overwrite
```

This has to be written into the documentation,
but having a oddly specific message helps finding the issue and documentation.

**Interaction with other features**
Trace in general and this specifically is most useful,
when "stop on first warning" is active.

**Implementation:**
Location of the assignment
* The location of the assignment will not be updated to the location of the overwrite.
   * this location is relevant for the evaluation and thus should not be changed
* The location of the overwrite will instead be stored separately.

Trace _(following the pattern used for modules and functions)_
* trace catches EvaluationException, outputs a message and rethrows the Exception


MichaelPFrey:
Location interacts with Customizer. I need to look into it.

MichaelPFrey:
This is some food for though regarding includest:

**main.scad**
```
//main description
included_variable = "I am in main";

include <include.scad>

echo(included_variable);
```
**include.scad**
```
//include description
included_variable = "I am from an include";
```

Having both locations could allow for included_variable to be influenced by the customizer.
But as included_variable is overwritten by an include, it creates a warning, so not having the variable in the customizer is somewhat explained.

Thinking about compatibility: https://customizer.makerbot.com/docs
>  Although you cannot include external libraries, the following are automatically available to be included in your script: 

So really: there is not much to worry about includes in therms of compatibility.

MichaelPFrey:
thinking about what location matters more:
```
//description 1
a = 1;

//description 2
a = 2;
```
This creates a warning in openscad. https://customizer.makerbot.com/docs creates two inputs for a.

Using the overwrite location seams like a good enough default and it is compatible with openscad before this change.

t-paul (approved):

PR #4200, opened 2022-04-03
Fix lib3MF version 2.0 build

obliviand:
src/io/export_3mf.cc: Missing wrapper parameter in new append_polyset
  call.


obliviand:
Commit bbd4da62cd7f0183f239c2fb600643b73b36f8d2 was missing the wrapper parameter for the 2.0 API.

t-paul (approved):
:tada:

PR #4205, opened 2022-04-07
syntax highlight: include and use have a unique syntax

MichaelPFrey:
`use` and `include` are not normal keywords. The usage of angle brackets `<>` is unique and has some interesting effects, that should be reflected by the syntax highlighting.

I will post some examples in future comments.

The main reference for openscad syntax for the purpose of this discussion is the lexer:

https://github.com/openscad/openscad/blob/667abf11c60a514ab28456e410309c621557f03a/src/core/lexer.l#L142-L147
https://github.com/openscad/openscad/blob/667abf11c60a514ab28456e410309c621557f03a/src/core/lexer.l#L151-L166

This merge request should bring syntax highlight closer inline with the current syntax interpretation.
If any of the example look like bugs (not features) this is out of scope for this merge request.

### DIFF
--- a/src/gui/ScadLexer.cc
+++ b/src/gui/ScadLexer.cc
@@ -111,11 +111,19 @@ Lex::Lex()
 
 void Lex::default_rules()
 {
+  rules_.push_state("PATH");
   rules_.push_state("COMMENT");
 
-  std::string keywords("module function use echo include import projection render "
+  std::string keywords("module function echo import projection render "
                        "return if else let for each assert");
   defineRules(keywords, ekeyword);
+
+  //include and use have a unique syntax
+  rules_.push("INITIAL", "use", ekeyword, "PATH");
+  rules_.push("INITIAL", "include", ekeyword, "PATH");  
+  rules_.push("PATH", ".|\n", etext, "INITIAL"); // catches things like (); but also acts as a "fail safe" so that the path state is allways leaved
+  rules_.push("PATH", "[\\+\\-\\*\\/%\\^!=]", eoperator, "INITIAL"); //include and use can be used as variable names
+  rules_.push("PATH", "[ \t\r\n]*<[^>]*>", eQuotedString, "INITIAL");
 
   std::string transformations("translate rotate scale linear_extrude "
                               "rotate_extrude resize mirror multmatrix color "
@@ -194,7 +202,7 @@ void Lex::lex_results(const std::string& input, int start, LexInterface *const o
 
   int isstyle = obj->getStyleAt(start - 1);
   if (isstyle == ecomment)    // WARNING - hardcoded number, not positive about this!
-    results.state = 1;
+    results.state = rules_.state("COMMENT");
   lexertl::lookup(sm, results);
 
   while (results.id != eEOF) {
--- a/src/gui/ScadLexer.cc
+++ b/src/gui/ScadLexer.cc
@@ -201,7 +201,7 @@ void Lex::lex_results(const std::string& input, int start, LexInterface *const o
   lexertl::smatch results(input.begin(), input.end());
 
   int isstyle = obj->getStyleAt(start - 1);
-  if (isstyle == ecomment)    // WARNING - hardcoded number, not positive about this!
+  if (isstyle == ecomment)
     results.state = rules_.state("COMMENT");
   lexertl::lookup(sm, results);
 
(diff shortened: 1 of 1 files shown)


MichaelPFrey:
Left: Nightly Build, Right: This Merge Request

I want to begin with an simple example of what is currently wrong:
```
//Test for failed assertion in an "use"

cnt =-2;

use<../assert-test-module1.scad>

row(-2);
```
assert is not a keyword in the context of being used as a filename.
Screenshot from 2022-04-07 19-07-36

similar pattern here:

```
translate(){
    //test
    cube();
}
include <test.scad>
include <test.scad>
include <include.scad>
include <include-test.scad>
use<use.scad>;
use  <use-421.scad>
use  <use-fasfa.scad>

use  <use-fasf.scad>
use <echo-search.scad>

echo1 = "test";
search = "test";
echo(echo1, search);

/*
include <test.scad>
include <include.scad>
include <include-test.scad>
*/
```

Screenshot from 2022-04-07 19-10-34

MichaelPFrey:
Now lets build up some context: Keywords are not reserved as in: they can be used as variable names.
Meaning: include and use are perfectly valid variable names:

```
use = 1;
include = 2;

echo(include);
echo(use);
echo(use * include);
echo(use + include);
echo(use - include);
echo(use - include); //comment
echo(use*include);
echo(use+include);
echo(use/include);
echo(use-include); //comment
search = test;
echo(search);
```

Screenshot from 2022-04-07 19-19-37

This shows a minor bug of this implementation: If `include`/`use` is followed by a space, the operator is shown as such.

if there is no white space between operator and `include`/`use`, the operator is incorrectly shown as text.

Note that the next post will show, why `include`/`use` are not good variable names and thus, this bug is somewhat acceptable.

MichaelPFrey:
Now I begin with the interesting quirks of include/use syntax.

Assume we have created the following two files:

` a && b `
```
echo("this is included") +
```
`funny.scad`
```
include = 1;
a = 5;
b =7;

if(include < a && b > 5){
    echo("that would make sense");
}
```
The current syntax highlighting makes it looks like a comparison of the variable include, when in fact, it is a perfectly working include:
Screenshot from 2022-04-07 19-30-02
_(This example can be hard to recreate. I had to rename the file in the command line to get the leading and trailing space in the file name. Regardless: The syntax of include allows for leading and trailing spaces.)_

an easier to recreate example
`st`
```
echo("Hello World!");
```
`8.scad`
```
include <te

st>;
```
Screenshot from 2022-04-07 19-38-58

Note: I do not want to argue the include/use syntax as it stands now, but syntax highlighting matching the actual syntax makes it easier to understand what happens.

MichaelPFrey:
This example is also interesting:
```
include <//test.scad>
include </*/my.scad>
echo("this is a test");
```
Screenshot from 2022-04-07 20-13-41

The nightly build not knowing that include<> has special meaning, tries to highlight comments within the angled brackets of the include. This makes is it look like the echo is commented out - but as the warning and output log show: there is no comment in this example.

MichaelPFrey:
One might think that the underline feature indicates the include path. Maybe in most cases, but it can be thrown off track:

`st`
```
echo("Hello World!");
```
`10.scad`
```
include <te

st>

use <te
st>
 
use
    
<no-55
non-existent>
```
Screenshot from 2022-04-07 20-31-02

jordanbrown0:
I keep expecting the underlined parts to act as links that would open the appropriate file.  I'd rather *not* have them highlighted, than have them be underlined and not act as links.

t-paul:
Looks good to me. Not sure with the discussion if you want to add some more changes?

The underlining is currently handled separately (by the main lexer/parser) so maybe that should not be mixed with the lexer changes unless we move it to actually use this lexer.

nophead:
@jordanbrown0 The underlined lines are links to open the file. You just need to use the non-obvious ctrl-click to open them.

jordanbrown0:
Thanks.  I thought I'd tried that, but it seems not.  Maybe it needs a tooltip.

MichaelPFrey:
> Looks good to me. Not sure with the discussion if you want to add some more changes?

**I would like to see this merged and then work on the finer details/impact in separate merge requests.**

To name one: Currently the file seams tobe only lexed/analyzed based on changes.
That improves performance, but the draw back is, that both comments and include/use use lexter states, where looking at a change (without the full context) might not yield the same results as lexing the whole file. I guess the solution is lexing the whole file on F5/F6 and/or save. But that is a separate task. _(That has to be noted here, as to the user, the issues explained here are not yet fixed while editing code. The fix only fully works when loading a file from disc and thus full file is lexed/analyzed.)_

> The underlining is currently handled separately (by the main lexer/parser) so maybe that should not be mixed with the lexer changes unless we move it to actually use this lexer.

Good to know, but it looks like the interface between lexer and also has edge cases: 

```
use<test>
use <hello>
use <1
2>
use <3
4>
```
Screenshot from 2022-04-08 06-37-28

That will be a separate merge request.

t-paul (approved):

PR #4208, opened 2022-04-09
only show hyperlinkIndicator as active when a modifier key is pressed and add tooltip

MichaelPFrey:
this goes back to #4205
> I keep expecting the underlined parts to act as links that would open the appropriate file.
> I'd rather _not_ have them highlighted, than have them be underlined and not act as links.

and approaches as follow:
   * hyperlinkIndicator are now hidden by default (`QsciScintilla::INDIC_HIDDEN`)
      * this also keeps the cursor "I-Beam", which is how the mouse courser should be by default in a text editor
   * when a modifier key (CTRL or ALT) is pressed, the hover indicator style is set to `PlainIndicator`
      * when the mouse is over an hyperlinkIndicator, this causes QT/QScintilla/Scintilla to
         * change the courser to "Hand"
         * to underline the text
      * _this is similar to QT creator_ 
   * in case of any event without modifier key, the hover indicator style is set to `HiddenIndicator`
      * this prevents mouser courser changes

This somewhat hides the "CTRL+Click to open" feature, so:
   * a tool tip is shown on click-release over an indicator, when no modifier key was pressed
Screenshot from 2022-04-09 13-18-20

The implementation is a bit of a hack.
QsciScintilla, Scintilla and QT use different concepts and do not allow full access to each others internal structures.
QT it self does not provide some events.

Namely:
   * there is a modifier pressed event, but no modifier relased event
      * note that relaying on a relase event could get messy, when the press event is in the application, but the release is outside
      * so having not having that event means not having such cases, which is a benefit
   * the indicator style remains always plain, only hoverstyle is used
      * this hides the fact that we do not get a modifier released event
      * If the indicator style would be set when the modifier is pressed, the user would see that we only clear it after an other event like mouse movement. By using hover-style, that quirk is hidden, as leaving the hoverstyle active after releasing the modifier key is not wrong and by moving the mouse at all, it is quickly cleared.
   * the link open on clicked, the tool tip shows on release
      * opening on click is what is to be expected
      * showing the tooltip on release makes sense as
         * showing it after clicked, makes it only shown as long as the click is held (qt quirk?)
         * the click might have a different intention (like selecting text) - showing the tooltip then could make the selection harder
   *  bool indicatorsActive
       * reduces code duplication
       * checking a bool should be cheaper then calling QsciScintilla


src/gui/ScintillaEditor.cc
```diff
@@ -1317,10 +1317,17 @@ void ScintillaEditor::setIndicator(const std::vector<IndicatorData>& indicatorDa
...
+
 void ScintillaEditor::onIndicatorClicked(int line, int col, Qt::KeyboardModifiers state)
 {
-  if (!(state == Qt::ControlModifier || state == (Qt::ControlModifier | Qt::AltModifier))) return;
-
+  if (!(state == Qt::ControlModifier || state == (Qt::ControlModifier | Qt::AltModifier))){
+    setToolTip("to open links, use <b>CTRL + Click</b>");
+    QTimer::singleShot(5000,this,SLOT(resetToolTip()));
+    return;
+  }
```
t-paul, line 1330: The behavior is indeed a bit strange regarding when the tooltip actually shows. After toying a bit with that, how about moving to the `indicatorRelease` instead of `indicatorClicked` event and go:
```
void ScintillaEditor::onIndicatorClicked(int line, int col, Qt::KeyboardModifiers state)
{
  qsci->SendScintilla(QsciScintilla::SCI_SETINDICATORCURRENT, hyperlinkIndicatorNumber);

  int pos = qsci->positionFromLineIndex(line, col);
  int val = qsci->SendScintilla(QsciScintilla::SCI_INDICATORVALUEAT, ScintillaEditor::hyperlinkIndicatorNumber, pos);

  // checking if indicator clicked is hyperlinkIndicator
  if (val >= hyperlinkIndicatorOffset && val <= hyperlinkIndicatorOffset + indicatorData.size()) {
    if (QGuiApplication::keyboardModifiers().testFlag(Qt::ControlModifier) || QGuiApplication::keyboardModifiers().testFlag(Qt::AltModifier)) {
      emit hyperlinkIndicatorClicked(val - hyperlinkIndicatorOffset);
    } else if (QGuiApplication::keyboardModifiers() == Qt::NoModifier) {
      QTimer::singleShot(0, this, [this] {
        QToolTip::showText(QCursor::pos(), "Use <b>CTRL + Click</b> to open the file", this, rect(), toolTipDuration());
      });
    }
  }
}
```
This almost retains the CTRL/ALT limitation of the original (just allows either combination of the two to open the file) and when clicking without any modifier it shows the tooltip directly.
The default time seems a bit long but as it hides when clicking somewhere else that might be ok instead of hard coding a time to show.
This also would not need the `resetToolTip()` method anymore.

t-paul (changes requested):

src/gui/ScintillaEditor.cc
t-paul: Oh, also, I'm not sure if that's currently relevant, but it seems better to have the tooltip logic inside the "ok, we checked it's actually a hyperlink indicator" as the callback triggers for any indicators set in the text.

t-paul (changes requested):

src/gui/ScintillaEditor.cc
MichaelPFrey: I will look into it. The suggestions sound good.

Currently, there are no other uses for indicators.

MichaelPFrey: ah, clever. The `Qt::NoModifier`check  is to ignore Shift+Click, which is used by the editor.

Regarding [`toolTipDuration()` ](https://doc.qt.io/qt-5/qwidget.html#toolTipDuration-prop)

> Specifies how long time the tooltip will be displayed, in milliseconds.
> If the value is -1 (default) the duration is calculated depending on the length of the tooltip.

Sound fine.

> moving to the `indicatorRelease` instead of `indicatorClicked` event

so that the tooltip is not shown on click+drag?

> The default time seems a bit long but as it hides when clicking somewhere else that might be ok instead of hard coding a time to show.

Strange. I am under `Ubuntu 21.10` which is using `Qt version 5.15.2 `.
There, using your code, the tool-tip disappears immediately when releasing the mouse button.

Or your describe what happens after _moving to the `indicatorRelease` instead of `indicatorClicked`_?

MichaelPFrey:
A few more thoughts independent of tooltip and implementation and looking at the user experience:

When the mouse pointer is over an indicator, it changes to `hand`.
By changing to `hand`, the editor changes what the user expect to happens when he clicks.
With `hand`, I expect the include to be opened on click.
As we break this exception, we now add a tool-tip (and try a balance act of not being annoying when editing text and still prominently showing the text).
Also side note: Precisely selecting text with the hand does not feel right.

Looking at it from this angle:
By leaving the courser `Ibeam` until CTRL or ALT is pressed, we would not create the exception mismatch in the first place. The down side of this approach is, that it makes the `CTRL+Click to open` feature harder to discover.

Starting up QT-Creator: Yupp - QT-Creator does not show the hand until I press CTRL.

Looking at geany: It also has CTRL Click features, but no hand.

-------

Taking yet an other angle:
We could add "open link" to the right click menu and indicate the short cut by writing out the menu entry with "open link (CTRL+Click)".
(Note: Implementing that sounds messy, as that would need graying out the option and supporting https or even generic URLs in strings and comments to not break user expectation of what a grayed out "open link (CTRL+Click)" _could_ mean)

-------

I think I will dwell on that a bit more.

src/gui/ScintillaEditor.cc
t-paul: Yes, the *disappears immediately* happens with the `indicatorClicked` event. I have not found a clear documentation describing that behavior, but I suppose it's the release event doing the cancel action. And not even the `QTimer` trick works. But I think the release event does make sense anyway.
For some reason I can't get the keyboard modifiers in the `indicatorRelease` event, it's always `0`. But the global keyboard state should be fine.

MichaelPFrey:
note: first comment and title changed to better reflect the current pull request and changes

t-paul (approved):

PR #4211, opened 2022-04-10
debug GUI indicator, lexter line numbers, lexter column numbers

MichaelPFrey:
fixes #4210, adds regression test for #4210

note: this could interact with #4209, so after merging either, the other merge requests tests should be rerun.

PR #4213, opened 2022-04-10
regenerate expected echo to fix interaction of two pull requests

MichaelPFrey:
#4211 added a test case

#4209 added details to an error message

#4211 triggers the changed message in #4209


t-paul (approved):

t-paul:
Thanks for the quick update :+1:.

PR #4215, opened 2022-04-12
Fold/Unfold all

osfesa:
Pull request for issue #4145

PR #4220, opened 2022-04-14
fix and regenerate ord-tests

MichaelPFrey:
That was a puzzle.

The ord-tests did not match what I got, when I regenerated the expected files.
No matter how I changed the expected file, the test would always passed.

The problem is here:

https://github.com/openscad/openscad/blob/d777dc24af688b2becff8ac01581ebf1612b42bf/tests/test_cmdline_tool.py#L126-L133

Python throws an exception, as it is not a valid UTF8 file.

Which ord-tests claims to be:
https://github.com/openscad/openscad/blob/d777dc24af688b2becff8ac01581ebf1612b42bf/tests/data/scad/misc/ord-tests.scad#L2-L4

As exception handling, the python script sets the text to empty string.
As both files are not valid UTF-8 Files, both are represented as empty string, which off-course matches.

see http://python-notes.curiousefficiency.org/en/latest/python3/text_file_processing.html#files-in-an-ascii-compatible-encoding-best-effort-is-acceptable for reference why I picked latin-1 as "de-fault" encoding.

### DIFF
--- a/tests/regression/echotest/ord-tests-expected.echo
+++ b/tests/regression/echotest/ord-tests-expected.echo
@@ -1,26 +1,27 @@
 ECHO: 97
 ECHO: 97
+WARNING: ord() number of parameters does not match: expected 1, found 0 in file ord-tests.scad, line 9
 ECHO: undef
 ECHO: undef
-WARNING: ord() argument undef is not of type string, in file ord-tests.scad, line 11
+WARNING: ord() parameter could not be converted: argument 0: expected string, found undefined in file ord-tests.scad, line 11
 ECHO: undef
-WARNING: ord() argument inf is not of type string, in file ord-tests.scad, line 12
+WARNING: ord() parameter could not be converted: argument 0: expected string, found number in file ord-tests.scad, line 12
 ECHO: undef
-WARNING: ord() argument -inf is not of type string, in file ord-tests.scad, line 13
+WARNING: ord() parameter could not be converted: argument 0: expected string, found number in file ord-tests.scad, line 13
 ECHO: undef
-WARNING: ord() argument nan is not of type string, in file ord-tests.scad, line 14
+WARNING: ord() parameter could not be converted: argument 0: expected string, found number in file ord-tests.scad, line 14
 ECHO: undef
-WARNING: ord() argument 3.1416 is not of type string, in file ord-tests.scad, line 15
+WARNING: ord() parameter could not be converted: argument 0: expected string, found number in file ord-tests.scad, line 15
 ECHO: undef
-WARNING: ord() argument [] is not of type string, in file ord-tests.scad, line 16
+WARNING: ord() parameter could not be converted: argument 0: expected string, found vector in file ord-tests.scad, line 16
 ECHO: undef
-WARNING: ord() argument [1, 2, 3] is not of type string, in file ord-tests.scad, line 17
+WARNING: ord() parameter could not be converted: argument 0: expected string, found vector in file ord-tests.scad, line 17
 ECHO: undef
-WARNING: ord() argument ["a", "b"] is not of type string, in file ord-tests.scad, line 18
+WARNING: ord() parameter could not be converted: argument 0: expected string, found vector in file ord-tests.scad, line 18
 ECHO: undef
-WARNING: ord() argument [1 : 1 : 5] is not of type string, in file ord-tests.scad, line 19
+WARNING: ord() parameter could not be converted: argument 0: expected string, found range in file ord-tests.scad, line 19
 ECHO: undef
-WARNING: ord() called with 2 arguments, only 1 argument expected, in file ord-tests.scad, line 20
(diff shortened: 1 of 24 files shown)


tests/regression/echotest/ord-tests-expected.echo
```diff
@@ -1,26 +1,27 @@
...
+WARNING: ord() parameter could not be converted: argument 0: expected string, found number in file ord-tests.scad, line 12
 ECHO: undef
-WARNING: ord() argument -inf is not of type string, in file ord-tests.scad, line 13
+WARNING: ord() parameter could not be converted: argument 0: expected string, found number in file ord-tests.scad, line 13
 ECHO: undef
-WARNING: ord() argument nan is not of type string, in file ord-tests.scad, line 14
+WARNING: ord() parameter could not be converted: argument 0: expected string, found number in file ord-tests.scad, line 14
 ECHO: undef
-WARNING: ord() argument 3.1416 is not of type string, in file ord-tests.scad, line 15
+WARNING: ord() parameter could not be converted: argument 0: expected string, found number in file ord-tests.scad, line 15
```
t-paul, line 14: Looks like this is not caused by your changes, but would you mind having a look if it's easily possible to get the actual value back into this message, that seems useful? Otherwise we can create a ticket so the topic is not lost.

t-paul, line 14: Looks like this is not caused by your changes, but would you mind having a look if it's easily possible to get the actual value back into this message, that seems useful? Otherwise we can create a ticket so the topic is not lost.

MichaelPFrey: sure, I can look into it.

The merge request was partly to see if the CI-Test uncovers unintended side effects (it did not for now), to confirm that my expected files works on Windows and Mac OS X (it does) and in part to get feedback what actually is desired output of ord() and if my approach to workaround the issue is acceptable.

t-paul: Cool, thanks. I'm still puzzled why your pull requests don't trigger the CircleCI builds. I guess the test suite runs are more important, but still strange.

MichaelPFrey: > I'm still puzzled why your pull requests don't trigger the CircleCI builds. 
> I guess the test suite runs are more important, but still strange.
Did you notice that I still use my own fork on github and thus use "across forks"?

I think that github treads "across fork" merge request differently from merge request within a project. (Regardless of the user being a member or not)

I could push to OpenSCAD instead of my fork, but I do not know the precise implications.
(e.g. I remember it used to be that we had restriction on Mac OS X Builds)

MichaelPFrey: just as note for myself:

https://github.com/openscad/openscad/blob/4db149f831e7b1c4382b1110f64b8ce340ed8492/src/core/builtin_functions.cc#L396-L400

https://github.com/openscad/openscad/blob/4db149f831e7b1c4382b1110f64b8ce340ed8492/src/core/Parameters.cc#L241-L263

MichaelPFrey: a train of thought: toEchostring can generate an error and thus throw an exception, for e.g. when the to vector to string conversion exhausts the stack.

This effectively can turn a warning into an error.

But... people running into that quirk, like have long running script, that should use "stop at first warning".

More problematic is, that the failing toEchoStrings prevents the warning from being generated.
That can be a real problem, as that can easily happen, when two arguments in a call are swapped. Knowing that there is a type error is more helpful, than knowing that converting the vector would exhaust the stack (which might never happen during normal runtime).

MichaelPFrey:
since  #4186, toEchostring can throw an exception (before it could crash which was not helpful either).
That would not be useful while trying to generate a warning.

Catching the exception that can be thrown by toEchostring while trying to generate a warning, means, that we get an Stack Exhaust Error and the Warning (with the value shown as `...`).

Note that toEchostring throws on stack exhaust. Stack is used up by - among other things:
   * Vector Depth while trying to create a vector string
   * recursive user module calls
   * recursive user function calls
Note that a stack exhaust is typically caused by a combination of the above.
Meaning: It might be impossible to generate an echostring for the same data while deeply nested in user modul calls - but it might be possible in the root document. Also, some Vectors are never mend to be outputed into the console, as they are used to generate 3D geometry.

It is not likely that someone would want to give a lot of vertices to for e.g. ord(), but error handling is part helping users to find typos. Having a Stack Exhaust Error (alone) would not be helpful, while the warning that the wrong data type was passed to a function is helpful.

tests/regression/echotest/ord-tests-expected.echo
t-paul: Yes, cross fork is different, but it's configured to still build just without the secret keys. It did work for example in #4215. Very strange.
Maybe it's confused by the combination of you having access to the openscad main repo and creating a pull request from the separate account. I'll try that with my account with the next change.

t-paul (approved):

PR #4221, opened 2022-04-15
warning when a \n is found in use<> or include<>

MichaelPFrey:
this is supported syntax:
```
use <test>
use
<test>
use
   <test>
```

This is a quirk of the lexter, not a feature:
```
use
<
test
>
use <
ignored line
used line
>
assert(false);
```
**In release**, it even throws of the line counting.
`[\n\r]` was unhandled, only the last path/filename was used as such.

**In nightly**, I recently fixed the line counting.
Doing so, I specifically added handling of `[\n\r]`.
That feels wrong, as it really is not well defined. But I do not want to make it an error for compatibility.
Making it a warning is a fair compromise.

We could argue, that this is a feature:
```
include<folder/
file.scad>

include
<folder/
file.scad>
```
but it is inconsistent
```
include<folder/
subfoler/
file.scad>

include<folder\
file.scad>

include
<folder\
file.scad>

use
<folder/
file.scad>

use
<folder\
file.scad>
```
that I want to support it for compatibility, but not as a feature.

### DIFF
--- a/src/core/lexer.l
+++ b/src/core/lexer.l
@@ -151,7 +151,8 @@ include[ \t\r\n]*"<"	{ BEGIN(cond_include); filepath = filename = ""; LOCATION_C
 <cond_include>{
 [\n\r]					{
 							LOCATION_ADD_LINES(parserlloc, yyleng);
-							LOG(message_group::Warning,LOCATION(parserlloc),"","new lines in 'include'-statement is not definded - behavior may change in the future");
+							// see merge request #4221
+							LOG(message_group::Warning,LOCATION(parserlloc),"","new lines in 'include<>'-statement is not defined - behavior may change in the future");
 }
 [^\t\r\n>]*"/"			{ filepath = yytext; }
 [^\t\r\n>/]+			{ filename = yytext; }
@@ -164,7 +165,8 @@ use[ \t\r\n]*"<"		{ BEGIN(cond_use); LOCATION_COUNT_LINES(parserlloc, yytext); }
 <cond_use>{
 [\n\r]					{
 							LOCATION_ADD_LINES(parserlloc, yyleng);
-							LOG(message_group::Warning,LOCATION(parserlloc),"","new lines 'use'-statement is not definded - behavior may change in the future");
+							// see merge request #4221
+							LOG(message_group::Warning,LOCATION(parserlloc),"","new lines 'use<>'-statement is not defined - behavior may change in the future");
 }
 [^\t\r\n>]+				{ filename = yytext; }
  ">"					{
--- a/src/core/lexer.l
+++ b/src/core/lexer.l
@@ -149,7 +149,10 @@ LOCATION_NEXT(parserlloc);
 
 include[ \t\r\n]*"<"	{ BEGIN(cond_include); filepath = filename = ""; LOCATION_COUNT_LINES(parserlloc, yytext); }
 <cond_include>{
-[\n\r]					{ LOCATION_ADD_LINES(parserlloc, yyleng); }
+[\n\r]					{
+							LOCATION_ADD_LINES(parserlloc, yyleng);
+							LOG(message_group::Warning,LOCATION(parserlloc),"","new lines in 'include'-statement is not definded - behavior may change in the future");
+}
 [^\t\r\n>]*"/"			{ filepath = yytext; }
 [^\t\r\n>/]+			{ filename = yytext; }
 ">"						{ BEGIN(INITIAL); includefile(LOCATION(parserlloc));  }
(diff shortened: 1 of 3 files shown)


src/core/lexer.l
```diff
@@ -149,7 +149,10 @@ LOCATION_NEXT(parserlloc);
 
 include[ \t\r\n]*"<"	{ BEGIN(cond_include); filepath = filename = ""; LOCATION_COUNT_LINES(parserlloc, yytext); }
 <cond_include>{
-[\n\r]					{ LOCATION_ADD_LINES(parserlloc, yyleng); }
+[\n\r]					{
+							LOCATION_ADD_LINES(parserlloc, yyleng);
+							LOG(message_group::Warning,LOCATION(parserlloc),"","new lines in 'include'-statement is not definded - behavior may change in the future");
```
t-paul, line 154: Typo `definded`. Maybe we could call it `include<>` to make it even clear what's meant?

src/core/lexer.l
```diff
@@ -159,7 +162,10 @@ include[ \t\r\n]*"<"	{ BEGIN(cond_include); filepath = filename = ""; LOCATION_C
 
 use[ \t\r\n]*"<"		{ BEGIN(cond_use); LOCATION_COUNT_LINES(parserlloc, yytext); }
 <cond_use>{
-[\n\r]					{ LOCATION_ADD_LINES(parserlloc, yyleng); }
+[\n\r]					{
+							LOCATION_ADD_LINES(parserlloc, yyleng);
+							LOG(message_group::Warning,LOCATION(parserlloc),"","new lines 'use'-statement is not definded - behavior may change in the future");
```
t-paul, line 167: Typo `definded`. Maybe we could call it `use<>` to make it even clear what's meant?

t-paul (changes requested):
Sounds good to me. Just noticed a typo in the message.

jordanbrown0:
I think it's far far more likely that this is a user mistake than intentional, and would be in favor of declaring it an error, especially if that makes anything (e.g. highlighting in the editor) simpler.

MichaelAtOz:
A path/filename can be very long, [260 chars, or 32K in W10](https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation).

 Formatting with a line break may make sense.

MichaelPFrey:
> A path/filename can be very long, [260 chars, or 32K in W10](https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation).
> 
> Formatting with a line break may make sense.

I agree in principle, but my issue is, that currently (and looking back to older versions especially releases), the lexter is messy when in comes to line break in use and include. If someone writes a scad file, using a new defintion of line breaks, it breaks compatibility to older version. I would not mind that if the old version would give out an meaning full error. But older version do not give any useful message and even loose track of line count, when they find a line break in between <>.

This is also where my long therm strategy is going: Making it a warning now, so that people that exploit this lexter quirk get warnings, when opening their files in a newer version of openscad (while the scad file still works as before).
Then later, when a release with the warning is established, someone can change the behavior (knowing that users have been warned that the behavior was undefined in the first place).

Side note: I would be careful to give use and include to much of a "non c-style" syntax.

> I think it's far far more likely that this is a user mistake than intentional,

seeing that it caused the line numbering to be off, it is likely a mistake and never (really) noticed before.
The lexter behaves as if it does not know what do to with newlines.

> and would be in favor of declaring it an error,

I think a warning (continue with "best effort") is good enough.
That way, all scad files with that quirk still work fine, but users are encouraged to change their files.
An error (do not continue - fully stopping the lexting) is not necessary.

> especially if that makes anything (e.g. highlighting in the editor) simpler.

The highlighting is fine (for the most part, you can get the syntax highlighting confused while editing, but that is fixed after pressing F5).
Debugging the syntax highlight lexer brought me to this issue in the first place.

The hyperlink indicators are confused, but this is a direct result of the lexter being confused.

MichaelPFrey:
I add in a `// see merge request #4221` for good measure.
This discussion here is a good reference.

For readers from the future, also see #4211 for context.

That way, it is clear that both the code and the test are a mix of backwards compatibility and provisional improvement in the context of backwards compatibility.

If there is ever a big language- or lexer-overhaul: Feel free to make up a new definition and implement it consistently. Feel free to write a new line number test - all the current line number test wants to be is a regression test based on the current lexer. For a new lexer, it can make sense to disregard the current test.

t-paul:
Yeah, going directly to error seems excessive here as there's no technical reason to go with warning first. This looks good to me, I would hope multi-line include<> or use<> is not a common thing. If there's unexpected impact we can always reconsider.

t-paul (approved):

PR #4222, opened 2022-04-15
trace: show parameters of user modul calls, make trace depth configurable

MichaelPFrey:
reworked based #4173. intended as a debugging features helping for cases such as [this](https://lists.openscad.org/empathy/thread/G2O3X6JB5SUY34HTERKMSM32BHA2AKS6).

Reworked the preference UI to group warnings and trace.

Setting the trace depth to zero effectively disables the trace functionality.

Having the option to disable _show parameters of user modul calls_ has a few indention:
   * when calling a user modul with a generated vector, the output can be come to long to practically handle
   * for the test framework, the number/recursion counter is not always deterministic (e.g. for a recursive call, the number depends on stack size and stack usage)

The output of this new trace message is ideally such, that it can be copy pasted and run.

When the original error was due to stack exhaust, changes are that we can not convert the parameters or vector within the parameters to a string. In that case, parameters or the vector respectively is represented by `...`.


MichaelPFrey:
I want to document one or two design decisions:

traceUsermoduleParameters also lowers the traceDepth parameter, despite not actually showing a new level, as the information is 'call of' and 'called by' is partially redundant (as in the user modul name).

https://github.com/openscad/openscad/blob/8e7f72b823e9748027d1639f304af1b4a6c4dde4/tests/regression/echotest/assert-fail3-test-expected.echo#L3-L4

https://github.com/openscad/openscad/blob/4db149f831e7b1c4382b1110f64b8ce340ed8492/tests/data/scad/misc/assert-fail3-test.scad#L1-L7

But note that one message contains the location being called and the other contains the location where the call happened.

In a small, single file this is of little use, but in a larger project having both pieces of information helps.

traceUsermoduleParameters affecting traceDepth is also the reason, why I combined traceUsermoduleParameters and traceDepth into one merge request: To get the same number of user module calls in the trace, the user needs to double the trace-depth.

Maybe the name  traceDepth is not accurate anymore, as it is now closer to traceMessagesLeft?

t-paul (approved):

PR #4225, opened 2022-04-18
error log improvement

MichaelPFrey:
* add Message Nr.
   * static constexpr -> enum
   * map various resizes to resizeRowsToContents()
      * the effect is that row height auto adjusts, which leads to line break
   * add tool tip 
   * disable drag and drop
      * I did not find functionality linked to drag and drop

### DIFF
--- a/src/gui/ErrorLog.cc
+++ b/src/gui/ErrorLog.cc
@@ -21,10 +21,10 @@ void ErrorLog::initGUI()
   logTable->verticalHeader()->hide();
   logTable->setModel(errorLogModel);
   logTable->setSelectionMode(QAbstractItemView::SelectionMode::SingleSelection);
-  logTable->setColumnWidth(COLUMN_NR, 20);
-  logTable->setColumnWidth(COLUMN_GROUP, 80);
-  logTable->setColumnWidth(COLUMN_FILE, 200);
-  logTable->setColumnWidth(COLUMN_LINENO, 80);
+  logTable->setColumnWidth(errorLog_column::nr, 20);
+  logTable->setColumnWidth(errorLog_column::group, 80);
+  logTable->setColumnWidth(errorLog_column::file, 200);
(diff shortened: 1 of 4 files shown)


MichaelPFrey:
Screenshot from 2022-04-18 07-43-07

t-paul:
Hmm, I'm not sure this is a good UI for this. The forced number column and essentially making it impossible to sort by other columns seems to indicate it may want to be handled differently.
I don't know what the best way is yet, some detail field or window, a more tree style display with trace at 2nd level or something else.

MichaelPFrey:
The column numbering and sorting are not important to me.
The sorting was just flipping on a setting.
If you want, I can remove those or make them optional.
Or we could move that to a separate discussion.
I don't mind.

What is important to me is calling `resizeRowsToContents()` as this leads to usable line breaks (instead of Ellipsis/`...`) and adding trace to the error log.

I kind of mixed various issue into one merge request to get a better feel for the whole interacting with each other.
As such, I would not mind splitting it or cutting out things.

Or are there plans for a totally different error log UI?

t-paul:
I'd prefer to keep the Trace out for separate discussion. All the other changes look good.

Plan for different UI: No. But there's some vague ideas on possible improvements. E.g. while it does not mention the error log specifically, the GSoC project idea https://github.com/opencax/GSoC/issues/53 might be integrated with the error log too. For example if we see the "unknown parameter to function" we could add the call info from the cheat sheet and link to the documentation and so on.

MichaelPFrey:
I think I get the rough idea.

The error-log aims at being "beginner an high level tool" (only relevant information and to the point - but with jumping of points to the documentation),

where as the console in general and trace specifically aim at being power user tools (at the same time noisy and highly compacted information)?

MichaelPFrey:
I close this merge request and open a new one.
That way, I do not loose code & ideas and the new pull request will be clean.

Issue #3937, opened 2021-10-12
Axes Mapping preferences not loaded from Windows registry


verglor:
If I change Axes Mapping it is saved in `HKEY_CURRENT_USER\SOFTWARE\OpenSCAD\OpenSCAD\input` registry. However when I close and open OpenSCAD (2021.01) these preferences are not loaded.
Also `rotationVPRelGain` is not saved at all.

PR #4226, opened 2022-04-18
fix preference setting for RotateVPRelGain

MichaelPFrey:
this adresses on aspect of #3937

> Also `rotateVPRelGain` is not saved at all. (Seems to be missing at [AxisConfigWidget.cc](https://github.com/openscad/openscad/blob/8d59f235d40f0b34a87a24f1121e54913311be48/src/input/AxisConfigWidget.cc#L157))


t-paul (approved):
:+1:

PR #4227, opened 2022-04-18
ignore mouse wheel in preferences

MichaelPFrey:
The preferences are a scroll area. When I scroll in the preferences, I tend to accidentally change setting, while trying to scroll to the setting I want to change.

The code duplication was a historic artifact.
Input driver was a feature branch and Customizer experimental.


MichaelPFrey:
I need to delete obsolete files

CMakeLists.txt
```diff
@@ -838,14 +838,13 @@ set(GUI_SOURCES
   src/gui/TabWidget.cc
   src/gui/UIUtils.cc
   src/gui/WindowManager.cc
+  src/gui/IgnoreWheelWhenNotFocused.cc
   src/gui/input/AxisConfigWidget.cc
   src/gui/input/ButtonConfigWidget.cc
   src/gui/input/InputDriver.cc
   src/gui/input/InputDriverManager.cc
   src/gui/input/InputEventMapper.cc
-  src/gui/input/WheelIgnorer.cc
```
t-paul, line 846: Unifying those is nice.

src/gui/input/AxisConfigWidget.cc
```diff
@@ -31,10 +31,9 @@
 #include "QSettingsCached.h"
 #include "input/InputDriverManager.h"
 #include "SettingsWriter.h"
-#include "WheelIgnorer.h"
+#include "../IgnoreWheelWhenNotFocused.h"
```
t-paul, line 34: Is the `../` needed? There should be a `-I/openscad/src/gui` to bring in the file without explicit path.

t-paul (approved):

PR #4232, opened 2022-04-20
fix depsfile when using multiple outputs

MaienM:
The depsfile was written for each output file, resulting in only the very last output filename actually being in the dependency file. This means that all outputs but the last won't have their dependencies known to make.

This PR changes this so that the depsfile is only written once, after all outputs have been created.


t-paul (approved):
Thanks! The change ignores the stdout case `-`, but we did that before too. So I suppose we can look into that separately.

PR #4234, opened 2022-04-21
gradient background needs GL_MODELVIEW

MichaelPFrey:
the gradient back only worked properly with show axis enabled.

### DIFF
--- a/src/glview/GLView.cc
+++ b/src/glview/GLView.cc
@@ -131,7 +131,11 @@ void GLView::paintGL()
 
   if(bgcol != bgstopcol){
     glDisable(GL_DEPTH_TEST);
+
     glMatrixMode(GL_PROJECTION);
+    glLoadIdentity();
+
+    glMatrixMode(GL_MODELVIEW);
     glLoadIdentity();
 
     //draw screen aligned quad with color gradient 


src/glview/GLView.cc
```diff
@@ -131,9 +131,13 @@ void GLView::paintGL()
 
   if(bgcol != bgstopcol){
     glDisable(GL_DEPTH_TEST);
+
     glMatrixMode(GL_PROJECTION);
     glLoadIdentity();
 
+    glMatrixMode(GL_MODELVIEW);
+    glLoadIdentity();
```
t-paul, line 139: Is the `GL_PROJECTION` part still needed? It looks to me like that would do nothing?

t-paul (changes requested):

src/glview/GLView.cc
MichaelPFrey: when I comment out `glMatrixMode(GL_PROJECTION);`, there is a square centered on the screen, which size depends on zoom level.

I followed https://community.khronos.org/t/gradient-background/54348

The rough idea seams to be to setup openGL that the corners of the view are a 1 unit away from center (independent of aspect ratio of the view), then look straight on, then look onto a square with gradient.

thehans: Both ModelView and Projection matrices are combined for the overall transformation (sometimes referred to as MVP).  Loading the identity resets whichever matrix is currently specified by `glMatrixMode`, but both are used simultaneously on all vertices going through the fixed function pipeline.

t-paul: Ah, ok. Thanks.

t-paul (approved):

PR #4235, opened 2022-04-22
Show manual overwrite confirmation only in the correct context

### LINKED ISSUE #3706: "Save as..." to existing filename prompts twice to overwrite.
Running nightly dev snapshot on Mint 20

1) Create 2 files: `foo.scad`, `foo2.scad`
2) Start OpenSCAD and open `foo.scad`, click "File -> Save As..."
3) Select foo2.scad, a confirmation appears:
> **Save File -- OpenSCAD**
> foo2.scad already exists.
> Do you want to replace it?
> No / Yes

4) Click yes, file dialog closes, and 2nd confirmation appears (only the dialog title bar differs slightly)
> **foo.scad -- OpenSCAD**
> foo2.scad already exists.
> Do you want to replace it?
> No / Yes

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/96972087-save-as-to-existing-filename-prompts-twice-to-overwrite?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jbinvnt:
Closes #3706

### DIFF
--- a/src/gui/TabManager.cc
+++ b/src/gui/TabManager.cc
@@ -699,6 +699,16 @@ bool TabManager::saveAs(EditorInterface *edt)
     filename.append(".scad");
   }
 
+  // Manual overwrite check since Qt doesn't do it, when using the
+  // defaultSuffix property
+  const QFileInfo info(filename);
+  if (info.exists()) {
+    const auto text = QString(_("%1 already exists.\nDo you want to replace it?")).arg(info.fileName());
+    if (QMessageBox::warning(par, par->windowTitle(), text, QMessageBox::Yes | QMessageBox::No, QMessageBox::No) != QMessageBox::Yes) {
+      return false;
+    }
+  }
+
   bool saveOk = save(edt, filename);
   if (saveOk) {
     setTabName(filename, edt);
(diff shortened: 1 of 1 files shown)


src/gui/TabManager.cc
```diff
@@ -697,15 +697,15 @@ bool TabManager::saveAs(EditorInterface *edt)
 
   if (QFileInfo(filename).suffix().isEmpty()) {
     filename.append(".scad");
```
jbinvnt, line 699: An alternate fix for this issue would be to remove this line, and the manual overwrite confirmation box, entirely. However, that course of action would change the behavior of the "Save As" function. The current behavior is to add `.scad` regardless of whether the user checks the "Automatically select filename extension" box in the Qt window. Even if the user elects not to select an extension and types their filename as `foo2`, this line causes it to be saved as `foo2.scad` anyway. From my investigation, this behavior is why the secondary, manual overwrite confirmation was added in the first place: because without that second box, if a file called `foo2.scad` already existed, it would be overwritten silently because that `.scad` extension is added after Qt has done all its internal checks.

rcolyer (approved):
All the confirmation pathways seem to have the correct behavior now.

PR #4237, opened 2022-04-24
moved animate do dockable widget

MichaelPFrey:
directly related to dockable widget:
   * some layout switching depending on width

side changes:
   * _pause button_ added
      * various related icons added 
   * infrastructure for _dockable window_ <-> _input driver_ connection
      * _animation pause_ in button mapping preference menu
      * _animation pause/resume_ in button mapping preference menu
   * debugged icons in button mapping preference menu
   * input boxes are set to red background, when an invalid value is set
      * (this can be helpful, when the user mixes up `.` and `,` as decimal separator) 
      * I do not want to set an input qt validator or similar, as some users currently use invalid values to stop the animation


MichaelPFrey:
three screenshots, showing the three layout states:

emulating the current look:

Screenshot from 2022-04-24 13-25-24

narrow look:

Screenshot from 2022-04-24 13-25-07

Screenshot from 2022-04-24 13-25-14

MichaelPFrey:
icons in button mapping preference menu:

Screenshot from 2022-04-24 15-12-12

the bug was simply that the actions where processed before the icons where added to the actions.
The icons are added in code, as we have a light/dark mode switch.

MichaelPFrey:
`Exec=env QT_STYLE_OVERRIDE=Adwaita-Dark ./openscad`

Screenshot from 2022-04-24 19-15-36

Screenshot from 2022-04-24 19-17-30

MichaelPFrey:
undock to seperate window works:

Screenshot from 2022-04-24 19-36-31

_(maybe we should change those messages describing the options? It works on much more then edition and console)_

Screenshot from 2022-04-24 19-40-34

t-paul:
I'm only getting the single line layout which looks a bit awkward for most window sizes. Maybe I need a clean rebuild.

The button seems to have the images flipped? I think the video player convention is to show what happens when you press the button, so in stopped state the "play" triangle is shown.

Something strange is going on with window selection. The idea is that ALT+W + N/P switches between editor / console / error log and makes them directly usable with keyboard. Animate does not get focus into the edit fields and the switching order does not work after going there with ALT+W + N.

MichaelPFrey:
I assume the feedback means a general approval?
Feel free to suggest renames or clean ups in the source code and interfaces.
That is what will stay for years to come and/or is tricky to change once merged.

Mean while the animate UI will likely grow, so the layout is somewhat contemporary.
(but some optimization is also important)

> I'm only getting the single line layout which looks a bit awkward for most window sizes. Maybe I need a clean rebuild.

https://github.com/openscad/openscad/blob/18096a60193f42b8f930fc665fd4fc923b94d47d/src/gui/Animate.cc#L215-L227

Note that this switches purely on width.
Maybe I should at a height > ...?
Yeah, I think that is it.
I will change that.

> The button seems to have the images flipped? I think the video player convention is to show what happens when you press the button, so in stopped state the "play" triangle is shown.

In my mind, both work, but when thinking about it: your suggestions makes the most sense.
 
> Something strange is going on with window selection. The idea is that ALT+W + N/P switches between editor / console / error log and makes them directly usable with keyboard. Animate does not get focus into the edit fields and the switching order does not work after going there with ALT+W + N.

I will look into that.

t-paul:
Yes, I think this is certainly a good step forward compared to the current strange "hide extra toolbar to stop animation" behavior. And as already asked for on the mailing list, it's much more open for further extension/improvements.

I'm just watching out for the window handling as that is likely critical for some people who need to minimize mouse usage. So including the animation stuff into that workflow is also a good thing.

The switch on width might not trigger for me on my 4k notebook screen :smile:.

MichaelPFrey:
> The switch on width might not trigger for me on my 4k notebook screen smile.

My initial concern where/are our users on miniature screens that are concerned with every pixel and do not like change.
I am on Full HD, with secondary screen.

Triggering on height looks and feels a lot nicer either way - the code is now also a bit more robust.

> Something strange is going on with window selection. The idea is that ALT+W + N/P switches between editor / console / error log and makes them directly usable with keyboard. Animate does not get focus into the edit fields and the switching order does not work after going there with ALT+W + N.

Nothing strange, just a missing "set focus".
That was also a good hint - ALT + W + A is a good keyboard shortcut.

t-paul:
Looks good.

We should add the icons to `resources/icons/svg-default/icons.scad`.

SVGs can be generated via `./scripts/generate-svg-icons.sh`.

Here's a suggestion:
```
module animate_disabled() {
    translate([width / 2, height / 2]) {
        outline(thin) circle(d = 0.8 * width);
		circle(d = 0.5 * width, $fn = 3);
		rotate(50) square([thin, 0.8 * width], center = true);
    }
}

module animate_pause() {
    translate([width / 2, height / 2]) {
        outline(thin) circle(d = 0.8 * width);
		for (x = [-thick, thick]) {
			translate([x, 0]) square([thick, 0.4 * width], center = true);
		}
	}
}
```

src/gui/MainWindow.cc
```diff
@@ -2788,16 +2725,7 @@ void MainWindow::animateUpdateDocChanged()
...
-    bool fps_ok;
-    double fps = this->e_fps->text().toDouble(&fps_ok);
-    if (fps_ok && fps <= 0 && !animate_timer->isActive()) {
-      animate_timer->stop();
-      animate_timer->setSingleShot(true);
-      animate_timer->setInterval(50);
-      animate_timer->start();
-    }
-  }
+    //animateWidget->animateUpdate();
```
t-paul, line 2728: `MainWindow::animateUpdateDocChanged()` and `MainWindow::animateUpdate()` look to be still connected via the `TabManager`.

MichaelPFrey: I look into it.

t-paul: Ah, I see, thanks for the update.

t-paul (approved):

PR #4238, opened 2022-04-26
Include <fstream> to fix the build with boost >= 1.79.0

heirecka:
boost::filesystem deprecated the string_file.hpp header with 1.79.0 [1],
which implicitly pulled in <fstream>.
In this case only std::fstream is used anyway, so just include <fstream>
directly.

[1] https://www.boost.org/users/history/version_1_79_0.html


t-paul (approved):
Thanks!

PR #4241, opened 2022-04-27
Add Option to Invert Mouse Buttons on 3DView

### LINKED ISSUE #3981: Add preference option to swap left and right mouse buttons.
Many other applications that I use, usually user the left-mouse-button for translation and the right-mouse-button for rotation. 
Can a swap option (checkbox) be added to the Preferences?
Thank you.

osfesa:
Please, let me know if you are not happy with the change.

Fixes  #3981.

### DIFF
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -352,6 +352,7 @@ MainWindow::MainWindow(const QStringList& filenames)
 
   QSettingsCached settings;
   this->qglview->setMouseCentricZoom(Settings::Settings::mouseCentricZoom.value());
+  this->qglview->setMouseInvertButtons(Settings::Settings::mouseInvertButtons.value());
 
   animate_timer = new QTimer(this);
   connect(animate_timer, SIGNAL(timeout()), this, SLOT(updateTVal()));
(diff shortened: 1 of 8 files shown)


MichaelPFrey:
I am unsure about the name. I personally would call this "Swap Mouse Buttons".
When I read "Invert Mouse Buttons" I think inverting of pressed/released.
But am not a native English speaker, so feel free show me a reference.

Microsoft uses _swap_ and _switch_. [1](https://support.microsoft.com/en-us/windows/change-mouse-settings-e81356a4-0e74-fe38-7d01-9d79fbf8712bl)
Others call it right and left handed mouse. (but this is not the point here)
"remap" is also used sometimes for such a feature, but that would be to broad.
Reverse Mouse button is also sometimes used.

Other thoughts on the name?

t-paul:
I'm bringing more votes for "Swap Mouse Buttons" :smile:.

Otherwise the change looks good.

osfesa:
I like "Swap mouse buttons" too. I'll submit the new name soon.

t-paul (approved):
:tada:

PR #4242, opened 2022-04-29
Fix using a relative path to `import` JSON in a `use`d file

mmalecki:
Previously we took the filename as-is, which caused us to attempt to
open skewed paths.

Add a test.

### DIFF
--- a/src/core/builtin_functions.cc
+++ b/src/core/builtin_functions.cc
@@ -37,6 +37,7 @@
 #include "FreetypeRenderer.h"
 #include "Parameters.h"
 #include "import.h"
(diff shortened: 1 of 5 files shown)


mmalecki:
Hm, I just realized there's a deprecation warning there (duh), but it seems like the most intuitive approach, especially for JSON (with modules wanting to ship their data).

t-paul (approved):

t-paul:
No, I think it's ok that way. It could be argued to not enable the fallback for the new stuff, but as it gives a deprecation warning I don't see a big problem just keeping it in sync with other imports.

PR #4243, opened 2022-05-01
viewport controll widget

MichaelPFrey:
Setting the view size is useful, when creating animation.
Some codecs require the height and resolution to be divisible by 4. 
Some  people might want to create a bounce of animations with consistent resolutions (e.g 640 x 480).

The technical reason why it can only crop not expand is, that it is tricky for a widget to trigger its parent elements to resize.
Depending on the request, it might be enough to move the docks, but it might require resizing the whole window - in some cases even above the systems resolution. By only allowing to crop, we can avoid having to deal with resizes.

The camera can be controlled by variables in the OpenSCAD script, but when the preview takes some time (and the render even more), that is not always practicable. 

Maybe "view control widget" would be more accurate.

To some extend, this is also just infrastructure work.
Expanding the mainwindow is always tricky, while working in a specialized dock is much easier.
Some example what additional controls could be added later:
* adding buttons like "isometric view" or "rotate by x units"
* adding generic sliders
* adding a custom 3D control widget
* save/restore current camera

----

Other professional CAD packages have a similar dialog with numeric entry and sliders.
I had to use one of those when working remotely, as for some reason somewhere my modifier keys where not applied to my mouse clicks. By providing the doublespinboxes, we now also have a default, when neither modifier keys nor input driver are available.


MichaelPFrey:
Screenshot in action:

Screenshot from 2022-05-01 13-37-00

t-paul:
Nice, I think that's useful especially for preparing and exporting images. It could maybe even provide a way to copy the command line parameters for producing the save view.

I'm not sure about the sizing. I certainly agree, this would be a very useful feature, but we tried that before and hit similar issues like you describe. At that point we reverted to just show the current size leaving it to the user as minimal compromise.

I think the correct solution would be to decouple the image export from the on-screen buffer and render the image to a dedicated off-sceen buffer that can have any size within the limits of what the graphics driver supports (usually at least 4096x4096).

What might make sense from viewport perspective is allowing for a fixed viewport aspect ratio with a couple of standard values like 4:3, 16:9 and so on.

t-paul:
Oh, maybe "Vewport control" would fit better as title, matching the naming in the status bar.

MichaelPFrey:
> Oh, maybe "Viewport control" would fit better as title, matching the naming in the status bar.

OK - I will rename it.

MichaelPFrey:
A general design idea: I think that "mid term" the groups will become collapsible and/or the ViewportControlWidget will become a scroll area (like the customizer/ParameterWidget)

> I think the correct solution would be to decouple the image export from the on-screen buffer and render the image to a dedicated off-sceen buffer that can have any size within the limits of what the graphics driver supports (usually at least 4096x4096).

That would be an option in the animation widget, not here.
That would also imply moving the export code to the animation widget - which sounds like a good plan when spoken out loud.

> What might make sense from viewport perspective is allowing for a fixed viewport aspect ratio with a couple of standard values like 4:3, 16:9 and so on.

I like that idea.
We could then connect that "somehow" with the animation resolution.

> It could maybe even provide a way to copy the command line parameters for producing the save view.

sure - that would be a button that puts the text into the clipboard. "no big deal".
(currently, editing the MainWindow Source Code is complicated - by creating specialized widgets, such "auxiliary functions" become easier to justify, write and maintain)

thehans:
With regards to resizing the viewport to *larger* sizes, I would expect that to probably create scrollbars in the GLView widget if its not able to fit in its pane, and not to affect the MainWindow size.

thehans:
To clarify my last comment some more, the scrollbars should be automatically handled by Qt, via the *immediate container* of the `QGLView`.  (I don't think [`QOpenGLWidget`](https://doc.qt.io/qt-5/qopenglwidget.html) itself ,which our `QGLView` class inherits, supports scrollbars directly.)

So I think we would just need to wrap our QGLView inside a new [QScrollArea](https://doc.qt.io/qt-5/qscrollarea.html) container, in the ui file here:
https://github.com/openscad/openscad/blob/5a0eb0a10299fe7d6ab59081080bb81340e3d39c/src/gui/MainWindow.ui#L153-L166

t-paul:
Scrolling might help with bigger viewports, but that also means the window scaling will not work anymore. Especially with the docked windows, the size calculation of Qt is behaving quite strange.

thehans:
So, behavior up to this point has been to simply expand viewport to fill the whole pane, when the user manually resizes the main window or drags the borders of viewport-surrounding widgets.

This brings up the question of how to prioritize user window resizing versus values entered in this new control widget.  Should ui resizing always immediately override values set in these new spinboxes?  

Or can we make them default to some "non-value" (for example it could display "Auto" in the spinbox textarea), so that only after the user enters something, then resizing window no longer affects the viewport size?

Maybe a separate checkbox for Auto size vs these controls?

t-paul:
I would drop the sizing till we can render to an off-screen buffer. And even then it would probably be better located in the animation window anyway, plus the (not yet existing) image export dialog.

All other things should be possible to handle without too much messing with the UI (well, none, except when we try to get the aspect ratio option in).

MichaelPFrey:
setting the aspect ratio here is good compromise and infrastructure, so lets go with that

MichaelPFrey:
> I think the correct solution would be to decouple the image export from the on-screen buffer and render the image to a dedicated off-sceen buffer that can have any size within the limits of what the graphics driver supports (usually at least 4096x4096).

Good description :+1:  - I have implemented working "prove of concept" code based on this description and it was straightforward. That will be a separate pull request.
I want to do some testing, UI polish etc. and I would like to bind it to the aspect ratio here.

t-paul:
Oh, nice! I've had some issue with the window handling again, but the code looks good. I'll give it another try and merge if there's nothing coming up.

As future idea, I wonder if maybe some additional movement buttons would be useful, similar to what Octoprint uses for the printhead https://octoprint.org/slides/fabcon14/img/screenie_application.png - selection of amount and buttons for directions.

MichaelPFrey:
> As future idea, I wonder if maybe some additional movement buttons would be useful, similar to what Octoprint uses for the printhead https://octoprint.org/slides/fabcon14/img/screenie_application.png - selection of amount and buttons for directions.

totally - creo has sliders/sideway spinners in a non-modal dialog to manipulate the camera and that is what inspired this widget in the first place (I had to use those as the modifier keys did not work in my remote work setup). I also like altiums rotation controls, but just by locking at it, I know tricky that would be to implement.

But currently, I do not have a good idea for the right controls. I mean: Likely it is just picking a style, adding some sort of timer for repeat while holding the button/slider position, and polishing it till it feels right... at which point I have my gamepad in hand, because that's "good enough". So, currently I think leaving it "as is" is better then doing something aimless.

Working in the widget feels a lot nicer then editing mainwindow.cc, so someone with the right idea has now a lower barrier of entry.

MichaelPFrey:
there was a typo that made the hide viewport control option not work.

t-paul:
Aha, thanks, I thought I checked that but still missed that. The tab order is a bit wonky, it starts at the absolute Y and finishes with the lock checkbox. If possible it should loop inside the window. However if that needs fighting with Qt, it's ok to leave it going back to the editor.

MichaelPFrey:
https://doc.qt.io/archives/qt-4.8/designer-tab-order.html makes it easy to improve the tab order - thank you for the hint.

Keeping the tab within the widget is also surprisingly easy.

t-paul:
Looks like you pushed the latest changes to the openscad repo (https://github.com/openscad/openscad/compare/2022-04-25-camControl). The PR comes from your user account so it did not pick those up. So maybe we switch to that branch and use a new PR?

MichaelPFrey:
> This PR comes from your user account so it did not pick those up. So maybe we switch to that branch and use a new PR?

I simply pushed it again to my user account. I prefer to keep the discussion together.

Side question: Would it make sense to have such tab order code in/for the customizer? I think with just a minor performance penalty I could generalize the code.

t-paul:
Yep, that works too.

I'm not sure about the customizer. If there's a way improving keyboard usage, that would be great. I think right now it's not really covered in that regard.

t-paul (approved):

PR #4247, opened 2022-05-02
Various fixes for source code

ntfshard:
This patch contains various changes spotted by cppcheck, compiler with enabled warnings and random files wandering
* fixes for format output
* passing arguments by const reference
* `StatCache::stat` took `std::string` as a 1st argument
* removed unused variables
* we should use `throw;` to rethrow an exception
* we should check index before accessing an element (src/core/customizer/CommentParser.cc)
* memory leak in error handler in src/glview/OffscreenContextWGL.cc

Found suspicious but not fixed: 
src/geometry/GeometryUtils.cc:296 Non-boolean value (-1)  returned from function returning bool. I need some clarification from codeowners is it false or is it true should be.

### DIFF
--- a/scripts/chrpath_linux.c
+++ b/scripts/chrpath_linux.c
@@ -380,7 +380,7 @@ chrpath(const char *filename, const char *newpath, int convert)
 
   if (strlen(newpath) > rpathlen)
   {
-    fprintf(stderr, "new rpath '%s' too large; maximum length %i\n",
+    fprintf(stderr, "new rpath '%s' too large; maximum length %u\n",
             newpath, rpathlen);
     free(dyns);
     free(strtab);
--- a/src/core/Expression.cc
+++ b/src/core/Expression.cc
@@ -330,12 +330,12 @@ Value Vector::evaluate(const std::shared_ptr<const Context>& context) const
     } else {
       VectorType vec(context->session());
       vec.emplace_back(std::move(val));
-      return std::move(vec);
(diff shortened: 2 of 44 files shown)


t-paul (approved):

t-paul:
The changes look good in general. I think there's some issue with older compilers in regard to that "using `std::move()` on return" for the boost variant type. Maybe there's a better way to solve this?

ntfshard:
Ok, I see. NRVO (named return value optimization) not working well for such configuration. I'll rollback this changes until C++17 (it will be mandatory in this case)

> Maybe there's a better way to solve this?

Get rid of old configurations or smart pointers =)

ntfshard:
What about my question, about src/geometry/GeometryUtils.cc:296 ?

t-paul:
> Get rid of old configurations or smart pointers =)

If only that was so easy, people like to cling to ancient installations :smiling_face_with_tear:. That said, C++17 is likely happening soon-ish now. The rough timeline is supporting old-stable on Ubuntu/Debian. But we need to update at least the AppImage build to 20.04 now that 22.04 is out.

> What about my question, about src/geometry/GeometryUtils.cc:296 ?

I don't know much about this code. Looking at how it's used, I would expect this to return false (= error, tesselation failed) at this point. Which means we would change the current logic. So I'd say, we fix it but via a separate mini-PR that can be easily rolled back in case strange things start to happen.

PR #4250, opened 2022-05-03
tessellatePolygonWithHoles() return value fix

ntfshard:
Changes for src/geometry/GeometryUtils.cc:296: Non-boolean value (-1) returned from function returning bool.
According to discussion in #4247 `-1` changed to false


t-paul (approved):
:tada:

Issue #4254, opened 2022-05-06
Editor: comment unused lines


Gregor-Gregor:
Add a menu command to comment when executed all the lines containing unused code (eg. not used variable definitions, unreferenced modules etc.).

Gregor-Gregor:
Purpose: to help cleaning up the complex code from all the dead ends, temporary solutions etc.

MichaelPFrey:
I need a minimum example, so...
```
if(0){
    echo("you should not see this");
}else{
    echo("run always");
}
```
More interesting is
```
a = true;

if(a){
    test1();
}else{
    test2();
}

//never run because overwritten
module test1(){
    cylinder();
}


module test1(){
    cube();
}

module test2(){
    sphere();
}
```
Detecting unused lines is a hard problem.
   * `a` is picked up by the customizer.
   * `a` can be overwritten, when the file is included.
 
 Much easier, but similar would be just to indicate lines that have actually been interpreted/run/visited.

I have a rough idea for one possible implementation: We could add a flag "visited"/"run" to every ASTNode. Every ASTNode already has a Location. Just run the code as normal, just adding a flag when the Node was visited. We could then walk trough the AST-Tree and add a green background to every line (or even character - we have the column) that ran. By not resetting the color until the code is changed, multiple runs (triggered by customizer or animation) could accumulate.

I have following main concerns:
   * much runtime impact would such a feature have?
      * adding and setting a bool should be cheap
      * having the feature "default off" should avoid the biggest isssue
      * by doing the UI changes after render/preview, it should feel like an unnecessary slow down
   * is there any way to setup an automated test?
      * this describes a GUI feature, but requires core work
      * would a "run lines of code" count help?
   * how robust is our location source code?
      * hidden bugs in location could be uncovered during the work - which is a good thing - 

I take a look into it

t-paul:
I'm not sure the "comment lines" is a good strategy. That could make things even harder if there's for example multiple assignments on one line.

It might be better to go with the logic other editors use by highlighting the issue in some way (marker on the sidebar, underlining the assignment, something like that).

I think using the AST is probably the way to go, meaning it needs at least the "compile" run. Traversing that should not be a huge concern, but having a feature toggle is probably still a good idea.

At this point there's no GUI tests, it might be possible for this scenario that we extend the AST export to include the extra information so it can be referenced via expected files. There's actually a number of features that could do with a better AST handling, like some simple optimizations and collecting timing information.

[...]

PR #4262, opened 2022-05-19
option to highlight evaluated AST-Nodes in the editor

MichaelPFrey:
implementing #4254 "mark unused code" is not trivial, but highlighting evaluated nodes is very close in effect.

This works very well for a single file. The statistic is very useful for finding death code.
Note that it works very specifically - if a default value of a user modul is unused, that is shown and counted as such.

Implementation details: gatherChilderen creates a linear way to access the ast-nodes. (The live time is limited to the live time of the corresponding ast-tree). gatherChilderen is based on and can be checked against print.
setEvaluated is a bit of a hack (mutable/const), but it makes sense, as it is a "set only flag".

The infrastructure changes and the feature it self have a lot of untapped potential.

This feature can later be expanded to cover multiple files, which would allow for a "code coverage" analyzes for library maintainers.

The infrastructure change (gatherChilderen) allows to transport meta information from the AST-Node into the editor (or other components).

qt handels invalid inputs to the color field gracefully.

### DIFF
--- a/src/core/AST.h
+++ b/src/core/AST.h
@@ -59,7 +59,7 @@ protected:
     this->evaluated=true;
   }
 private:
-  volatile mutable bool evaluated=false;
+  mutable bool evaluated=false;
 };
 
 std::ostream& operator<<(std::ostream& stream, const ASTNode& ast);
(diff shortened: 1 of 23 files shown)


MichaelPFrey:
**demonstration screenshoot:**
Screenshot from 2022-05-21 07-58-03
**demonstration video:**
   * [animation](https://www.youtube.com/watch?v=pVXRDVkpy1w)
   * [customizer](https://www.youtube.com/shorts/ucyPG74d64o)
   * [change background color](https://www.youtube.com/watch?v=jUvqtCxs99o)

Issue #4244, opened 2022-05-01
(minor) core dumped when openscad is started in a formerly available but now deleted directory.


st2000:
Tried starting openscad while in a directory that was deleted and saw the core dump.  Simple workaround ... just change to any not deleted directory.  

```
openscad
terminate called after throwing an instance of 'boost::filesystem::filesystem_error'
  what():  boost::filesystem::current_path: No such file or directory
Abort (core dumped)
[1]  + Abort                         openscad (core dumped)
```
What's running:
```
> openscad --version
OpenSCAD version 2019.05
> lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.4 LTS
Release:	20.04
Codename:	focal
```

PR #4267, opened 2022-05-23
core dumped when openscad is started in a formerly available but now deleted directory

osfesa:
Issue [4244](https://github.com/openscad/openscad/issues/4244)

Please let me know if you don't find it acceptable.

Thanks.


jordanbrown0:
While you're there, please fix "does not exists" to be "does not exist".

(And I would remove the "Try ..."; I don't think it adds value.)

osfesa:
Sorry guys. Holidays and merge/rebase problems, I've been able to submit the changes.

I guess, the circleci test failed as WASM has no QT.  I'll investigate how to compile it.

PR #4261, opened 2022-05-16
Animate Widget: Offscreen Rendering and Improved UI

MichaelPFrey:
* Offscreen rendering allows for high resolution [(example)](https://www.youtube.com/watch?v=OLwsw_gTG54)
   * the UI improvement increases interactivity, the slider also acts as a visual progress indicator [(example)](https://www.youtube.com/watch?v=SCJjOjilNwo)
   * the view all option helps with animations that automatically grow [(example)](https://www.youtube.com/watch?v=iOBT1hFiUVE)

PR #4273, opened 2022-05-29
adjustable edge in preview

MichaelPFrey:
[demonstration](https://www.youtube.com/watch?v=tUHdCCiM-sU)

This feature can be used for high dpi screen, artistic effect [(example)](https://www.youtube.com/watch?v=mSmYnRA4sY4) and/or to find issues in the model by offering better contrast.

todo
   * [ ] wait for #4261
   * [ ] update the UI File
   * [ ] is there a way to have the constant 1.414 in fewer places?

Currently, this feature is not supported in command line and thus the automated testframework.

### DIFF
--- a/shaders/Preview.frag
+++ b/shaders/Preview.frag
@@ -1,6 +1,8 @@
 #version 110
 
 uniform vec4 color1, color2;
+uniform float totalHalfEdgeThickness; // total thickness of half-edge (per triangle) including fade, (must be >= fade)
+uniform float edgeFadeThickness; // thickness of fade (antialiasing) in screen pixels
 varying vec3 vBC;
 varying float shading;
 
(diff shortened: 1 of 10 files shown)


thehans:
> * [ ]  is there a way to have the constant 1.414 in fewer places?

The `edgeFadeThickness` variable could be re-formulated as something like `edgeFadePercentage` (0 to 100% of the halfEdgeThickness), to cut those constant occurrences in half.

Then, line 17 in the shader could be changed from:
```
vec3 a3 = smoothstep((th-fade)*d, th*d, vBC);
```
to (assuming I got the math right)
```
vec3 a3 = smoothstep(th*(1 - fade * 0.01) * d, th * d, vBC);
```

thehans:
Also in case its not clear, (I don't think the code comments mention it) the reason behind the chosen constant value was just a rough approximation for the distance of one pixel diagonal `sqrt(2) ~= 1.414`. 
As far as I remember, that just *felt* like a reasonable distance to anti-alias across, but I don't think there was any "more scientific" reasoning beyond that somewhat arbitrary choice.

PR #4285, opened 2022-06-21
fixed overwrite of feature enable from commandline

parvit:
Hi, I think i've found an issue in the Feature functionality.
If you run openscad with "--enable <a functionality disabled in preferences>" the `setupFeaturesPage ` method in Preferences.cc execution actually resets the value to false and so it remains disabled.

This change syncs the QT value to the command line value requested as one would expect.

### DIFF
--- a/src/gui/Preferences.cc
+++ b/src/gui/Preferences.cc
@@ -327,7 +327,12 @@ void Preferences::setupFeaturesPage()
     cb->setFont(bold_font);
     // synchronize Qt settings with the feature settings
     bool value = getValue(featurekey).toBool();
-    feature->enable(value);
+    if( !feature->is_enabled() ) {
+      feature->enable(value); // only if not already set from command-line
+    } else {
+      value = feature->is_enabled(); // otherwise sync QT with command-line request
+    }
+
     cb->setChecked(value);
     cb->setProperty(featurePropertyName, QVariant::fromValue<Feature *>(feature));
     connect(cb, SIGNAL(toggled(bool)), this, SLOT(featuresCheckBoxToggled(bool)));


MichaelAtOz:
I haven't done any specific OpenSCAD testing for a while, but I used command-line in testing via GUI, I had shortcuts with things enabled so I could just click to get the specified behaviour. I have used `--debug` with GUI, and also load specific `.scad` files for subsequent GUI-ing. I can see `--hardwarnings` `--check-parameters` & `--check-parameter-ranges` being on/off for different testing situations. 
I agree command-line shouldn't make permanent changes, but I don't think there should be a hard separation with GUI use.

thehans:
`--enable` basically assumes that openscad is being run in non-interactive mode (i.e. if an input file and at least one `-o` output is specified, then the program just processes and exits without invoking the Qt based GUI).

I have never even considered using the command line options as a temporary override of the GUI preferences (specifically experimental Features), but I'm not sure what this would look like in terms of a sensible interface.

Say that `--enable` did act this way, what would you expect the user to see if they open Preferences -> Features ?
Should it not reflect the features enabled via command line?  That seems confusing to not show the settings that are currently in effect.
But then *if it did* show overridden values, then the changes would be automatically written to disk (making this no longer temporary).  I think this would happen as soon as the Preferences window opens and the widgets are initialized (or maybe only upon closing the Preferences window, I don't recall exactly).

Another issue is that `--enable` is currently "additive" only.  Command line usage starts with no features enabled, and you can specify any number to enable.   It can "toggle on" something which was not set in GUI, but cannot toggle off any of those settings. 
So would we need to implement a new `--disable` option to allow for this?
*OR* should it be such that, if any `--enable` is option is provided, then all other features not specified are disabled?

thehans:
Sorry, I misread this as a feature request and not a PR.

I have not yet run a build with this change, to observe which exact behavior it exhibits, but I think the points above are still valid.

parvit:
@thehans @MichaelAtOz  After your comments i think it is appropriate then to only let the switch be a temporary value, not override the persistent preferences.

PR #4290, opened 2022-06-25
Add error check to userConfigPath on mac

hexane360:
This PR adds an error check to `userConfigPath` on MacOS. Without this, I was getting segfaults in some edge cases (like running openscad as a system user).

In the event of error, it returns an empty string, which is the same behavior as on Linux (in `PlatformUtils-posix.cc`).


t-paul:
Thanks, that Linux solution is probably not perfect, but if following that prevents crashes on MacOS, it's worth having.

PR #4293, opened 2022-07-05
Feature multicore geometry

parvit:
This change responds to #391.

It creates a new feature called "multicore-ops" and allows the multithread execution of the geometry collecting operations:
* Union
* Subtraction
* Hull 
* Minkowsky

It supports both standard (Nef) and hybrid polygon operations (when the corresponding feature is also enabled).

The subdivision of work is simple divide-and-conquer:
1) if the subdivision reaches the maximum number of executors or the list cannot be subdivided further then the operation is applied and returned
2) if not then two balanced lists of geometries are generated
3) the two threads are launched recursively and awaited for their result
4) the operation is then applied on the two partial results and returned to the parent thread

From my tests i've found that the Nef objects do not significantly benefit from the change (because the last operation still dominates the running time and cannot be simplified via threading), while the hybrid operations do seem to have a certain boost.

I cannot give precise measurements because i've develop this in a virtual machine so my timings are not really indicative.

A word of caution: this is not a silver bullet to improve performance, the scaling is _by geometry_ not _by facets_ so there could be anomalous scenarios where the added operations can slow down significantly the entire operation

If there is anything else to document or explain, please let me know.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -739,6 +739,7 @@ set(CGAL_SOURCES
   src/geometry/GeometryEvaluator.cc
   src/geometry/cgal/cgalutils.cc
   src/geometry/cgal/cgalutils-applyops.cc
+  src/geometry/cgal/cgalutils-applyops-multicore.cc
   src/geometry/cgal/cgalutils-applyops-hybrid.cc
   src/geometry/cgal/cgalutils-applyops-hybrid-minkowski.cc
   src/geometry/cgal/cgalutils-closed.cc
--- a/src/Feature.cc
+++ b/src/Feature.cc
@@ -50,6 +50,8 @@ const Feature Feature::ExperimentalTextMetricsFunctions("textmetrics", "Enable t
 const Feature Feature::ExperimentalImportFunction("import-function", "Enable import function returning data instead of geometry.");
 const Feature Feature::ExperimentalSortStl("sort-stl", "Sort the STL output for predictable, diffable results.");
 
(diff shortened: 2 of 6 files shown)


nseinlet:
Is there a way to make openscad able to use more than 2 cpu at a time? I'm basically using `difference` and `union` . logs shows launch of threads, more than one cpu core is used, but, as far as I can see, not more than 2.

jordanbrown0:
@nseinlet That's exactly the point of this project and issue #391 .

parvit:
@nseinlet this patch _does_ allow to use all the cpus you have, but only if the operations have enough object to parallelize (it splits the list of objects into two threads until it can't anymore with the max number of cores).

The limiting factor is that the total runtime is limited by the two final biggest operands so not much is actually achieved in terms of running time (for real models).

I'll rework this to do the parallelization on a by-vertex way, but it's going to be some time before i can do it.

nseinlet:
@jordanbrown0 Hi, sorry, if I was unclear. I know it's the point to use more than one core, but I wasn't able to use more than 2. My question is ` how can I refactor my code to use more than 2 cores`  or ` What' s the ideal way to architecture the design to maximise the number of used cores` 

@parvit I will try to refactor my code to have more objects that can be rendered in //. Thanks for the job, BTW.

jordanbrown0:
And it sounds like I may have misunderstood the context of your original comment - it sounds like you are asking about multicore operations *with* this not-yet-integrated change.  I had originally thought that you were asking about released OpenSCAD.

PR #4294, opened 2022-07-05
add dict() builtin function with example

### LINKED ISSUE #4203: Feature request: Associative arays
Hello friends,

I am thinking why this very good software has not the **associative arrays**... (sorry vectors..)
I found a lot of workarounds using some kind of search functions, e.g.: [here](https://www.thingiverse.com/groups/openscad/forums/general/topic:5610) or [here](https://stackoverflow.com/questions/55448011/associative-arrays-in-openscad).

What is currently working is that first three items from the array can be accessed by x,y,z indexes like:
```
array.x
array.y
array.z
```
So why we cannot access the any item by similar way?

I can imagine that the array can be specified by the following example (or something similar):
```
array1 = [
   index0: "value0",
   index1: [ "value1-0", 1.1 ]
   index2: [ 
      index2-0: "value2-0",
      index2-1: "value2-1",
   ]
]
```
Then items might be accessible by normal numerical indexes or by "named indexes"
So these variants will be equivalent:
```
array1[0]
array1.index0
array1["index0"]    // probably better variant?...
```
And some nested arrays samples:
```
array1[1][1]
array1.index1[1]   // nested array has only numeric indexes
array1.index2.index2-1
array1["index2"]["index2-1"] 
array1.index2["index2-1"]    // combined way
array1[2]["index2-1"]    // combined way
array1["index2"][1]    // combined way
```

The array can be stored in memory by the same way as now with additional info which will allow to convert the "named indexes" to the numeric indexes - then Internally the item will be retrieved by numeric index as usually.

This can be very useful feature. What do you think?...

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/112040485-feature-request-associative-arays?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

adam-trhon:
dict() is a variable parameter function (all arguments must be named)
that constucts ObjectType with attributes corresponding to given
arguments. See the attached example comments for dict() purpose.

The pull request closes #4203 and partially addresses #3088 (see [this comment](https://github.com/openscad/openscad/issues/3088#issuecomment-539626365)).
I added only a function in order to not introduce any new syntax.

### DIFF
--- a/examples/Advanced/dict_split_definition_instantiation.scad
+++ b/examples/Advanced/dict_split_definition_instantiation.scad
@@ -1,0 +1,102 @@
+// dict_split_definition_instantiation.scad - usage of dict()
+
+// dict() construct an associative array, where keys are strings. One of
+// possible usages is to separate definition of the model from its
+// instantiation. The model defintion then can be used elsewhere, for example
+// to define mounting holes or attachment points, as shown in this example.
+
+// The example could be split into three source files, as the modules are
+// supposed to not know about each other. However, for simplicity, the files
+// were fused into one and the content split by markers.
+
+
+$fn = 32;
+
+/* ---------------------- BOX MODULE DEFINITION ---------------------------- */
+
+function box_definition(length, width, height)
+    = dict(
+        length = length,
+        width = width,
+        height = height,
+    );
+
+module box_instance(definition) {
+    translate([-definition.length/2, -definition.width/2, 0])
+        difference() {
+            cube([definition.length, definition.width, definition.height]);
+            translate([4, definition.width/2, -0.01])
+                cylinder(r=2, h=definition.height+0.02);
+            translate([definition.length-4, definition.width/2, -0.01])
+                cylinder(r=2, h=definition.height+0.02);
+        }
+}
+
+module at_box_mounting_holes(definition) {
+    translate([-definition.length/2, -definition.width/2, 0]) {
+        translate([4, definition.width/2, 0])
+            children();
+        translate([definition.length-4, definition.width/2, 0])
+            children();
+    }
+}
+
+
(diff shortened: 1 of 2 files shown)


kwikius:
Rather tha dict(...)[You could use an array with a minor mod to the vector_element syntax ](https://github.com/kwikius/openscad/blob/WIP-first-class-module/src/core/parser.y#L742)
see https://github.com/openscad/openscad/pull/4337#issuecomment-1242141901

adam-trhon:
@kwikius I could, but I saw the discussion about a new syntax in other threads and wanted to avoid it.

kwikius:
> @kwikius I could, but I saw the discussion about a new syntax in other threads and wanted to avoid it.

OK, the modification to syntax is a minor and pure extension to the language and doesn't break existing code. Arguably adding a builtin function is more of a breaking change!

t-paul:
With the big discussion going on in #4337 we probably end up with that one as both PRs cover the same topic. I'm not finally sure if there's maybe a dedicated dictionary data structure, but so far it seems PR #4337 might cover all the important use cases, notably the request to also allow for calculated keys.

@adam-trhon it would be nice if you could have a look at that implementation, if it covers all the use case you would like to see.

adam-trhon:
@t-paul Yes, it covers the usecase.

BTW, does it make sense to update the example in this PR after #4337 is merged, or is the example out of scope?

jordanbrown0:
Implemented (with a different name) in #4337 .

PR #4300, opened 2022-07-16
Improve axis scale markers when view is off center.

### LINKED ISSUE #1206: Improve axis scale markers
If the design is not centered in the origin and you zoom into it, the scale markers eventually disappear since you assume that the center of focus is the origin.
-> needs more testing.

SoapSeller:
This is my attempt to fix  #1206.
There is some more work to be done, see bellow.

Before this fix, centering the following cube:
```
translate([2000,-10,-10])
cube(20);
```
Looks like:
openscad-bad
Afterward it will look like:
openscad-good

However, this fix is not complete.
I've made an effort to make minimal changes to current code structure to get this reviewed for basic concept before moving on to make bigger changes.
The main issue is that currently  the code treat all axis as equal, rendering the markers in one pass. It creates an issue when different lengths of axis are shown on screen(as my change try to find the global min&max over all axis) due to rendering of too much details and slowing the view to a crawl(at least on my oldish NVIDIA GTX 960), for example, in this view:
openscad-slow

Implementation details:

Introduce new functionality in GLView:
* Extraction of frustum planes from opengl.
* Calculating angle between vectors.
* Intersection of axis with frustum planes.

And use the new functionality to define&use min/max bounderies in
GLView::showScalemarkers to draw markers - instead of previous zoom
level usage.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .uuid
 .vs/
 .vscode/
+.devcontainer/
 /*.scad
 **/out*.*
 *.dmg
--- a/src/glview/GLView.cc
+++ b/src/glview/GLView.cc
@@ -6,6 +6,7 @@
 #include "printutils.h"
 #include "Renderer.h"
 #include "degree_trig.h"
+#define _USE_MATH_DEFINES
 #include <cmath>
 #include "boost-utils.h"
 #ifdef _WIN32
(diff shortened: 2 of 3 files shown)


t-paul:
Nice, yes that should fix the build, I was just about to comment `_USE_MATH_DEFINES` is need for `cmath` to provide `M_PI`.

t-paul:
I'd suggest moving all the `<>` includes to the top. It still failes due to `degree_trig.h` importing `cmath` already without the `#define`.

SoapSeller:
@t-paul thanks for the hint - I've found that degree_trig.h define M_RAD2DEG which is actually what I've used M_PI for, so I've replaced it. Much nicer like that.

PR #4305, opened 2022-07-22
Select shader location

jbinvnt:


jbinvnt:
The shaders included in this PR are imported from #3860. For the fastest progress on the new feature, at least initially I've structured the shader switching mechanism to read `Preview.frag` and `Preview.vert` from a parent folder which names the shader to be used (i.e. the structure that was already used to present the examples). In order to extract just the relevant files I ran the following command in the `shaders` directory after checking out the branch in the PR:

```
find . -name 'Preview.*' -exec cp --parents {} keptshaders/ \;
```

jbinvnt:
The GSoC 2022 work is now consolidated in #4330

PR #4306, opened 2022-07-24
Add "em" parameter to text functions to set size of an em-square.

### LINKED ISSUE #4304: text scaled wrong (resolution = 100 should be 72)
That seems like a pretty blunt statement, but after some analysis I come to the conclusion that it's true:  OpenSCAD text processing scales the text wrong by about 30%.

If you tell OpenSCAD to draw an X at 10 units in the default Liberation Sans, you get an object that's 9.552 units tall.  What happened to the other 0.45 units?

If you tell it to draw an X in Arial at 72 units, you get an object that's 71.6 units tall.  But if you tell any *other* application (Word, PowerPoint, CorelDraw) to draw an X in Arial at 72pt, noting that 72pt is one inch, you get an X that's 0.716 inches or about 51.6pt  tall.  In OpenSCAD, you get an X that's close to the specified size, but in all of the other programs you get an X that's about 70% of the specified size.  Why the mismatch?

The answer is buried in the text processing...

https://github.com/openscad/openscad/blob/ea8950bd3cd7f06da333c15463492e484098e591/src/core/FreetypeRenderer.cc#L279

```
  FT_Error error = FT_Set_Char_Size(face, 0, scale, 100, 100);
```

The two constant 100s are the answer.  They tell FreeType to return results scaled so that one unit is 1/100 inches.  (Yes, OpenSCAD is unit-less, but FreeType is not - with this call, the input parameter "scale" is in points (1/72 inch) and the output coordinates are in 1/100 inches.)

The "size" parameter controls the *body size* of the text, which is roughly and nominally the font ascent plus the font descent.  The cap-height of the font is typically 70% or so of that, to allow for descenders.

Thus with size=72 we ask for a 72pt body size, or a body size of an inch, but because of the two 100s we get glyphs with a body size of 100.

It's a total coincidence that the typical ratio for cap height to body size is 72:100 and so the scaling mismatch roughly yields the size equaling the cap height.

This inaccuracy affects both `text()` and  the text metrics functions `textmetrics()` and `fontmetrics()`.  (But at least they match.)

---

The picture in font-land is really more complex than this, because body height is not necessarily related to the actual size of the glyphs.  Some fonts have glyphs that are taller than their nominal body height, while other fonts have glyphs that are smaller than their nominal body height.

---

The picture in OpenSCAD internals is really more complex than this, because OpenSCAD has FreeType render glyphs at 100,000 units tall, where units are 1/64pt, and then scales them to the specified size.

jordanbrown0:
Adds a new parameter `em` to `text()`, `textmetrics()`, and `fontmetrics()`.  Where `size` has a 100/72 error and only accidentally more or less matches the height of a capital letter, `em` directly specifies the em size, also known as body size, of the text.  This is the standard way to specify the size of text.  `size` continues to work bug-compatibly.

Fixes #4304.

Note that this includes temporary support for a "raw" return from fontmetrics() that returns raw values reported by FreeType.  Do people think that this should be made a permanent feature, perhaps triggered by fontmetrics(raw=true)?


MichaelAtOz:
I'm all for raw and more (underline_position & underline_thickness - or is that a [FT version](https://freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_facerec) thing?). 
I'm wondering if, maybe later, some of the esoteric fields (\_flag & numb\_ etc) could help with things like doing a table style layout of glyphs, so maybe `raw=false` would be good. ?

Why default 13.9 for `Em`? I realise it is close  to the magic [72](https://en.wikipedia.org/wiki/72_(number))**\*** ratio (13.88¯), I would think **10** would allow easier default line layouts. **\***_number of planets between hell and heaven_

t-paul:
I'm concerned about exposing too much internals of FreeType into the OpenSCAD interface. I'd be fine to add some of the information for debug purposes but making data structures based on libraries can cause headaches if they ever change anything. It's difficult enough with the datastructures we define ourselves.
Using `raw = true` does not help at all as even if it's documented as "may change", it's bound to be used an cause issues when actually changed.

jordanbrown0:
Why 13.9?  Because there can only be one default size, and that default has to be the current definition of size=10, which corresponds to em=13.88888.

Exposing more information?  Yes, but probably not via "raw".  If there's something that we want to expose, it should be OpenSCAD-ized - for instance, if there's a bitmask, we should expose it as a set somehow, rather than exposing it as an integer that you have to do bitwise arithmetic on.

With respect to revealing internal information... yes, that's a very real concern.  Maybe instead of `raw=true`, have a "debug mode" flag in the UI, so that it's emphasized that programs can't depend on it?  Or maybe a compile-time flag?  I'm not strongly pushing for it, but the information *was* useful in checking some stuff out, and it'd be nice not to have to recreate it next time.

jordanbrown0:
But based on a very quick look at the test results, I have more to do.  My guess is that changing the resolution from 100 to 72 subtly changed the shape of all of the characters, even with the scaling difference canceled out.

MichaelAtOz:
28% reduction in DPI = more chunky glyphs.
Assuming there is no workaround, and you really don't like default of 10 (as they are two different measures, they don't need to be the same), I'd round the Em default to 14.

MichaelAtOz:
What would happen if you left the DPI at 100, but internally multiplied the Em value by 0.72?

jordanbrown0:
Yes, that's what I expect.  It's not as bad as you might expect, since FreeType thinks it's rendering type that's about 22" tall, so it's dropping from a grid of 2200 to a grid of 1562... neither exactly low-resolution.

We ask FreeType to generate text that is 100,000 times 1/64pt tall, and then we reverse that scaling on the result.

And now I'm wondering where the 1/64 factor went in the equation.  More thought and experimentation required.

Yes, there are other possibilities for where to apply the correction - perhaps by scaling the em-size to compensate, or perhaps by setting the resolution to either 100 or 72.  My initial thought was that I wanted to keep the lower levels as clean as possible, and hide the bug compatibility up as close to the top as possible, since it's the "size" parameter that needs to be bug compatible.

jordanbrown0:
Some thought and research later... I think the 1/64 factor is cancelled on the output side by the coordinates being passed back from FreeType *also* being in 1/64 units.  Unfortunately, the [FreeType documentation](https://freetype.org/freetype2/docs/reference/ft2-outline_processing.html#ft_outline_funcs) doesn't explicitly say what the coordinate system is for the values passed to the callback functions.  It gives the C types (FT_Vector, FT_Pos), but [FT_Vector](https://freetype.org/freetype2/docs/reference/ft2-basic_types.html#ft_vector) doesn't say and [FT_Pos](https://freetype.org/freetype2/docs/reference/ft2-basic_types.html#ft_pos) says "it depends on context".

I still need to look at the details of the test failures.

jordanbrown0:
Found the big problem, I think.  The csgpngtest tests operate by generating a .csg file and then processing that .csg file.

My change didn't fix the CSG dump, so a CSG dump would show size= with the corrected size.  That is, for an input size=10, it would show size=13.888889.  When reprocessed, that size=13.88889 would be scaled up *again*.

The fix is to have it emit "em=" instead of "size=".  Or it could use the uncorrected value and emit "size=", but my overall hope is to limit the use of the "bad" size to as few cases as possible, preferably isolated to the argument processing for text() and the metrics functions.

jordanbrown0:
At the moment, all of the "dump" tests fail because they dump the "corrected" value, which does not match the golden reference.  Emitting "em" as described above won't fix that, of course.  The fix is to update the golden reference after inspecting the new results.  (Or in some fashion go back to CSG reporting size= and the uncorrected size.)

jordanbrown0:
The push above changed strategies:  instead of correcting the "size" parameter at the top, the top records whether it was set using size= or em= and the bottom uses that to select either 100 or 72 as the resolution.  This should yield precisely compatible results when using size=, instead of only 99.9% compatible.

jordanbrown0:
And I #if'ed out the raw stuff.  Still need to decide what the final answer is there, but #if'ing it out should let the fontmetrics() test pass.

jordanbrown0:
Any opinions on what I should do with the "raw" stuff?  Rip it out, leave it under #if, or do something more elaborate?

Will it be helpful for me to squash out non-final commits in preparation for merging?

jordanbrown0:
That push ripped out the "raw" stuff.  It was easy enough to add; if there's a need for it I can add it again.

MichaelAtOz:
> Any opinions on what I should do with the "raw" stuff?

@t-paul said "to add some of the information for debug purposes", ie change it to PRINTDEBUG, see https://github.com/openscad/openscad/blob/master/src/utils/printutils.h

MichaelAtOz:
...but I think the underline info (underline_position & underline_thickness) is not debug info.

jordanbrown0:
I agree that underline position and thickness are not debug information.  (Or, rather, they would be in their original units, but not if converted to OpenSCAD units.)

I can add them if you like.  I would add a new entry to the `fontmetrics()` return, an object `underline` with members `position` and `thickness`.

jordanbrown0:
Here's how those new elements might be used:
```
f = "Liberation Sans";
size = 20;
text = "Hello, world";

fm = fontmetrics(font=f, size=size);
tm = textmetrics(text, font=f, size=size);

echo(fm=fm);
echo(tm=tm);

text(text, font=f, size=size);
translate([0, fm.underline.position - fm.underline.thickness/2, 0])
    square([tm.advance.x, fm.underline.thickness]);
```
Note that a truly clever underlining scheme puts gaps in the underline around descenders.  That's left as an exercise for the reader.  (Hint:  difference, offset.)

jordanbrown0:
Do we want an issue or PR for the addition of those two values?

jordanbrown0:
That last push added underline parameters to the data returned by fontmetrics():

`ECHO: { nominal = { ascent = 12.5733; descent = -2.9433; }; max = { ascent = 13.6109; descent = -4.2114; }; interline = 15.9709; font = { family = "Liberation Sans"; style = "Regular"; }; `**`underline = { position = -2.48895; thickness = 1.0173; };`**` }`

Note that while FreeType reports the position at the vertical center of the underline, this reports the position at the bottom of the underline.  This allows a simple
```
translate([0, fm.underline.position, 0])
    square([tm.advance.x, fm.underline.thickness]);
```
to draw an underline, rather than requiring the user to translate down by `fm.underline.thickness/2`.

jordanbrown0:
Assuming that the tests run OK (which I expect them to), I think I'm done with this wad, and it's ready for final review and perhaps acceptance.

jordanbrown0:
Still on my to-do list.

jordanbrown0:
Replaced by #6656.

PR #4307, opened 2022-07-25
Update README.md

wenqingl:
This README is great but I made some comments on the grammar

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 <p><a href="https://opencollective.com/openscad/donate"><img align="right" src="https://opencollective.com/openscad/donate/button@2x.png?color=white" width="200"/></a>
 
 OpenSCAD is a software for creating solid 3D CAD objects. It is free software and
-available for Linux/UNIX, MS Windows and macOS.</p>
+is available for Linux/UNIX, MS Windows, and macOS.</p>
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -99,8 +99,8 @@ development. Other versions may or may not work as well.
 
 If you're using a newer version of Ubuntu, you can install these 
 libraries from aptitude. If you're using Mac, or an older Linux/BSD, there 
-are build scripts that download and compile the libraries from source. 
-Follow the instructions for the platform you're compiling on below.
+are build scripts that download and compile the libraries from the source. 
+Follow the instructions for the platform you're compiling below.
```
froqstar, line 103: Are we compiling the platform? I think the current wording is better here, we are compiling OpenSCAD _on_ the platform of choice.

froqstar: Also, I think "compiling from source" is a common term in software dev, even though it may not be 100% correct language wise (not an expert on that :) )

jordanbrown0: "Follow the instructions below for the platform you're compiling on" might be better, avoiding the strange "on below" sequence.

t-paul: Thanks for the comments, this helps me as non-native speaker. @wenqingl would you mind updating the PR with the suggested changes (or give an argument for keeping the original version)?

PR #4333, opened 2022-08-21
Save and restore GL context on double click.  Fixes #4332.

### LINKED ISSUE #4332: Right double click sometimes messes up rendering
Start with a simple cube(10);.

F5 preview, then right double click outside the model.  It might take a few times, but usually not many.  Click and drag to pan or rotate, and observe the ... interesting ... rendering.

image
image

It may (or may not) be relevant that Thrown Together shows it in purple.

image

jordanbrown0:
Take special note of the voodoo warning, and of the NEEDSWORK note for the copyright notice on a new file.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -829,6 +829,7 @@ set(GUI_SOURCES
   src/gui/PrintService.cc
   src/gui/ProgressWidget.cc
   src/gui/QGLView.cc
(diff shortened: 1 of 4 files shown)


t-paul (approved):

t-paul:
Ok, lets do some voodoo then :crossed_fingers:.

Issue #4336, opened 2022-08-28
Feature request: alias-module


kwikius:
An id aliasing a module and its arguments
(For simplicity allow alias for builtin module {cube, cylinder etc} only to start with).
The alias-module can use dot member operator to access the arguments.

Consider a simple rectangular plate with holes near the corners.

proposed syntax:
```openscad
/*
 create a rectangular plate with holes near the corners.
*/
plate : cube([100,200,3],center = true);  // module_alias
hole : cylinder(d = 5, h = plate.z + 2);  // module_alias
border = 5; // distance from outside of hole to edge of plate

difference(){
   assert(plate.center);
   plate();  // module_instantiation
   for( x = [-1:2:1]){
      for( y = [-1:2:1]){
         pos = [
             (plate.x/2 - (border + hole.d/2)) * x,
             (plate.y/2 - (border + hole.r)) * y, //hole.r == hole.d/2
             -(plate.z/2 + 1 )
         ];
         translate(pos){
           hole();
         }
      }
   }
}
```
existing syntax ( edited based on input from @UBaer21 )
```openscad
/*
 create a rectangular plate with holes near the corners.
*/
plate_size = [100,200,3];
hole_dia = 5;
module plate() cube(plate_size,center = true); 
module hole() cylinder(d = hole_dia, h = plate_size.z + 2); 
border = 5; 

difference(){
   plate();
   for( x = [-1,1]){
      for( y = [-1,1]){
         pos = [
             (plate_size.x/2 - (border + hole_dia/2)) * x,
             (plate_size.y/2 - (border + hole_dia/2)) * y,
             -(plate_size.z/2 + 1 )
         ];
         translate(pos){
             hole();
         }
      }
   }
}
```
syntax tested  with parser.y
https://github.com/openscad/openscad/blob/master/src/core/parser.y
```
module_alias
        : TOK_ID ':' module_id '(' arguments ')'
           {
              /* ... */
           }
        ;
```  
        tested with :
```
@>:  bison -fsyntax-only parser.y 
```

UBaer21:
```
size=[100,200,3];
d=5;
module plate(size=size)cube(size,center = true); 
module hole(d=d)cylinder(d = d, h = size.z + 2); 
border = 5; 

difference(){
   plate();
   for( x = [-1,1]){
      for( y = [-1,1]){
         pos = [
             (size.x/2 - (border + d/2)) * x,
             (size.y/2 - (border + d/2)) * y,
[...]

PR #4337, opened 2022-08-29
Added object() function to make an object from arguments.

revarbat:
The `textmetrics()`, `fontmetrics()`, and `import("foo.json")` functions can return an Object type value that contains sub-values accessible via foo.bar syntax.  But creating a new object isn't available.  This PR adds an `object()` function that takes named args and makes them into object named sub-values.

```openscad
obj = object(foo=42, bar="fee", baz=[4,6,8]);
echo(obj.foo, obj.bar, obj.baz);
```

This also suggests an `newobj = obj_set(object, key=value, ...);` function that returns a copy of the given object, with new values set or overridden by the other given key=value arguments.  I have not implemented that in this PR as I am not clear enough on the function internals.

### DIFF
--- a/src/core/builtin_functions.cc
+++ b/src/core/builtin_functions.cc
@@ -427,6 +427,19 @@ Value builtin_concat(Arguments arguments, const Location& loc)
   return std::move(result);
 }
 
+Value builtin_object(Arguments arguments, const Location& loc)
+{
+  ObjectType result(arguments.session());
+  for (auto& argument : arguments) {
+    if (argument.name == boost::none) {
+      LOG(message_group::Warning, loc, arguments.documentRoot(), "object() argument is not named");
+    } else {
+      result.set(argument.name->c_str(), std::move(argument.value));
+    }
+  }
+  return std::move(result);
+}
+
 Value builtin_lookup(Arguments arguments, const Location& loc)
 {
   if (!check_arguments("lookup", arguments, loc, { Value::Type::NUMBER, Value::Type::VECTOR })) {
@@ -1084,6 +1097,11 @@ void register_builtin_functions()
   Builtins::init("concat", new BuiltinFunction(&builtin_concat),
   {
     "concat(number or string or vector, ...) -> vector",
+  });
+
+  Builtins::init("object", new BuiltinFunction(&builtin_object),
+  {
+    "object(key=value, ...) -> object",
   });
 
   Builtins::init("lookup", new BuiltinFunction(&builtin_lookup),


revarbat:
This helps address issue #4203

t-paul:
With the `object` discussion haing a diffent focus regarding the syntax, would it make sense to move this (and similar recent #4294) to focus on dictionary as discussed some time ago? Specifically supporting calculated keys and other features that will not be available via a "normal" `object` syntax as discussed in #3087. I think there's room for both features as they seem to have enough difference that prevent unification without losing functionality.

I've been meaning to dig out that discussion for a while, but spare time is extremely rare lately.

t-paul:
cc @adam-trhon

jordanbrown0:
@t-paul 's point about calculated keys is valuable.  A related topic would be where even the number of keys is dynamic.  Probably one of the forms should be
```
object([[name, value], ...])
```
That is, where the argument is a list of name-value pairs.

t-paul:
Ref: https://lists.openscad.org/empathy/thread/3DVXSTCBWJPKPUVBJBFMMQ27BVYUS352

revarbat:
I admit I hadn't even considered the calculated key issue. But @jordanbrown0 's idea for a single un-named argument with a list of key-value pairs is a clean and simple solution.

jordanbrown0:
Getting more general, I don't think it would need to be limited to a single such argument.  There are probably some devilish details, but it seems sensible to allow one to mix explicit name=value arguments with both [[name,value],...] arguments and existing objects, all unioned to form the final object.

revarbat:
Allow `object()` to take a mix of named and unnamed arguments (each un-named one containing either another Object or a List of `[KEY, VALUE]` pairs). When a duplicate key appears, we keep the value associated with the last.

revarbat:
The previous discussion I started about dictionaries sadly suffers from having to modify the parser.y syntax.  It's syntactically nice, but doing it all from inside an `object()` function is far simpler to implement, and not that much more verbose.

jordanbrown0:
I think we do want a `{ ... }` object construction syntax.  It's just nicer for building static objects.  But also I think it plays well into the #3956 data=render() mechanism.

t-paul:
Modifying the parser is fine, we just need to try to not cause issues with the existing syntax. Using bison does introduce a couple of restrictions on what is possible. The last changes for supporting the dot notation did introduce some unwanted side effects, but it's a bit cleaner now so hopefully even less risk for further changes.

Unfortunately most punctuation characters are already used otherwise, but maybe we can still find some nice and short expressions.

revarbat:
Err, tweaking the parser syntax, making it mergeable with #3956, and working with the new-fangled (to me.  Get off my lawn you darned kids!) C++11 rvalues syntax and optimizations pretty much raises this above the amount of effort I'm willing to put into it.

revarbat:
In any case, I'm not so worried about this PR not making it in, as much as I was just trying to make sure the idea of an `object()` function was considered.

jordanbrown0:
Having done only "architectural" review - that is, what does this function look like to the OpenSCAD program, and not any code review - I'm OK with letting it in as-is, with just name=value pairs.  It could then be compatibly extended to include [[name,value],...] pairs and objects, and could reasonably be parallel with the { ... } object constructor syntax that I'm suggesting elsewhere.

revarbat:
Is the redundancy really appropriate, though?  I'm perfectly willing to add the `[[name,val], ...]` support, if you do want it.

jordanbrown0:
Well, we'll see what others say, but I think it's OK.  I want objects to be a first-class citizen, which suggests dedicated syntax, and I like the idea of combining it with the geometry-as-data mechanism because they could use the same syntax.  At the same time, that syntax doesn't work at all for calculated keys, and those seem important.

revarbat:
Okay, I've added support in `object()` for `[[key,val],...]` and object arguments.

revarbat:
```openscad
obj1 = object(foo=43,bar="fee",baz=object(r=.3,g=.2,b=[3,7,4]));
obj2 = object(obj1, [["fee",22],["fie",true]], blatz=PI);
echo(obj2);
```
results in:
```
ECHO: { foo = 43; bar = "fee"; baz = { r = 0.3; g = 0.2; b = [3, 7, 4]; }; fee = 22; fie = true; blatz = 3.14159; }
```

revarbat:
Oh, and later values override earlier values for duplicate keys, letting you make new copies of an object with altered entries.

jordanbrown0:
Should probably be experimental.  Look at examples for ExperimentalImportFunction, one line each in src/core/builtin_functions.cc, src/Feature.cc, src/Feature.h.

revarbat:
Experimental now as "object-function".

MichaelAtOz:
> Oh, and later values override earlier values for duplicate keys, letting you make new copies of an object with altered entries.

There are two ways to do that, update the value for the key in the original order offset, or delete the original key from its old offset and create a new key in the position of the new (duplicate).

```
o1 = object( the=1, quick=2, brown=3, fox=4 );
o2 = object(  jumped="a", over="b", the="c", lazy="d", dog="e" );

o3 = object( o1, slash="/", o2 );
echo( o3 );

a. ECHO: { the="c", quick=2, brown=3, fox =4, slash="/", jumped="a", over="b", lazy="d", dog="e" }
OR
b. ECHO: { quick=2, brown=3, fox =4, slash="/", jumped="a", over="b", the="c", lazy="d", dog="e" }
```
I can see arguments for both. The second seems more right to me.
Perhaps it should match behaviour of an existing language?
Does it matter when this gets extended later?

revarbat:
I replaced the value in place, so the keys() ordering stays the same when making an altered copy of an object.

revarbat:
That is to say, the code currently in this PR will get result "a".

MichaelAtOz:
```
o1 = object( the=1, quick=2, brown=3, fox=4 );
o2 = object(  jumped="a", over="b", the="c", lazy="d", dog="e" );
```
I considered how I would go about arranging to get the "b" result with the current behaviour.
Then how would you remove a key?
As `undef` is valid that killed my initial thought.
So a list comprehension to build a name-value list and test each for the name you want to remove to exclude it.

Would a simpler syntactic solution be better?

Maybe `o3 = object( o1, the=, o2);`

MichaelAtOz:
BTW is `concat()` valid for these objects, so `o3=object(o1,o2)` == `o3=concat(o1,o2)`?

jordanbrown0:
concat is not yet valid for objects.

For removing entries, I was debating whether `name=undef` should set the value to undef, or should remove it.  Although undef is a theoretically valid value, it usually isn't a very *useful* value, and the only difference between deletion and a value of undef is in what you get out of iteration, so I was slightly leaning towards saying that `name=undef` would remove name.

Another possibility would be to work from the `[ [ name, value ], ... ]` variation and say that an entry like `[ [ name ], ...` with no value might delete that member.

I would *not* use a syntactic answer like `name=` if I could possibly avoid it.

jordanbrown0:
I have mixed feelings about whether `concat()` should ever be valid for objects.  If it could completely replace `object()` then I'd say yes, but it can't (at least not well) replace `object()`'s ability to create objects with calculated keys.  Because it can't do that, I would be inclined to say that `concat()` is for lists and `object()` is for objects.

revarbat:
I see value in being able to store an `undef` value in an object element.
The `[[KEY], ...]` syntax has clear merit, though the extra brackets make me want to be able to just use a list of key name strings for deletion:  `["foo", "bar", ...]`.

jordanbrown0:
Maybe, though there's also an argument for allowing standalone [name, value] pairs and that would be ambiguous.

That is, `object(n1=v1, [n2, v2], n3=v3)`.  Doesn't technically add anything, but for the simpler cases it might be nice.  The questions would be (a) whether to have a simpler syntax for either case, and (b) which one takes precedence for the simpler syntax.

revarbat:
I just enabled object key deletion with an argument like `[[KEYSTR1], [KEYSTR2], ...]` or just `[KEYSTR1, KEYSTR2, ...]`.  I also made it complain if a key name string contains a non-alphanumeric character.

I can remove the `[KEYSTR1, ...]` syntax if you prefer.

jordanbrown0:
We can let other people comment on how `[s1, s2]` should be interpreted - whether it should mean "set name s1 to value s2" or "delete s1 and s2".  I think the former, but not strongly.  More importantly, I'm not sure that it's an important enough case to provide a shortcut for *either* of them.

> I also made it complain if a key name string contains a non-alphanumeric character.

Why?  Yes, you won't be able to refer to them using `obj.elem` syntax, but you will be able to using `obj[s]` syntax.  And for constructed keys punctuation can be very useful.

MichaelAtOz:
Debating: without the wrapping `[` `]` some could think `[KEYSTR1, KEYSTR2]` was `[KEYNAME, KEYVALUE]`?

I'd say it is less confusing to say entries should always be in a list. `[...]`
An entry is also a list. `[...]`
An entry list can be a name/Value pair, or a name (in which case it is for deletion). `KEYNAME,KEYVALUE` | `KEYNAME`

@jordanbrown0 
Why? Identifiers are `'A-Z','0-9','a-z','_'` ATM & not unicode. I'd say `_` is enough. ?
Extend it here, extend it to "variables"?

jordanbrown0:
I'm coming from JavaScript, where there's a similar mechanism.  For keys that follow identifier rules, you can say `obj.key`, but for any key you can say `obj[key]`.  JavaScript is, for instance, perfectly happy to have a key consisting of the empty string.
```
> x = {}
{}
> x[''] = 5
5
> x
{ '': 5 }
```

And I think you would find that virtually all dictionary / associative-array implementations allow arbitrary-string keys.

Useful for what?  For looking up values where the natural key might not be limited to identifier characters.

```
material_colors = object([
    ["Dark Wood Walnut PLA", "brown"],
    ["Translucent Purple PLA", "purple"]
]);

module material(name) {
    color(material_colors[name]) children();
}

material("Dark Wood Walnut PLA") cube(10);
```

jordanbrown0:
Note that I edited that comment to add
```
    echo("Material", name);    // for bill of materials
```

revarbat:
@jordanbrown0 That is a very good argument for non-alphanumerics.

MichaelAtOz:
?? Unicode

jordanbrown0:
Unicode probably doesn't hurt, as long as you don't expect to be able to use a different form to retrieve the value.  (No storing it with a key in composed form and expecting to retrieve it using a key in decomposed form.)  If we were doing case-insensitive matching it would be harder, but we're not.

revarbat:
Okay, I removed the alphanumeric key restriction, and the [KEY,...] deletion syntax.

revarbat:
One last thought... setting and deleting nested objects.  From hard experience with other languages I've written, making a copy of a nested structure, while changing a deeply nested value is a pain. Having a shortcut to do that would be helpful, but possibly annoying to implement.  If I were to give it a syntax here, I'd use something like
`[[["key", "nested_key", "even_more_nested_key", ...],value], ...]`

```openscad
obj1 = object(type="foostruct", num=23, data=object(counts=object([["item1",73],["item2",29]])));
obj2 = object(obj1, [[["data","counts","item1"],30]]);
```

revarbat:
Possibly allow scalar indices for setting nested vector values.

revarbat:
Just had another thought: if we allow `undef` values, I don't see any efficient way to determine if a key exists.
```openscad
obj = object(item1=34, item2=undef, item3=82);
// obj["item2"] and obj["nonexistent"] both return undef, even though one exists.
exists = [for (x=obj) if(x=="item2") 1]!=[]; // Seems to be the only way to determine item2 exists.
```

revarbat:
```openscad
function obj_has_key(obj,key) =
    obj[key]!=undef || [for (k=obj) if(k==key) 1]!=[];
```

revarbat:
Yet another idea, though this will need minor parser changes: Enabling varargs in functions and modules by using objects.  Maybe something pythonish like:
```openscad
function foo(arg1=7, arg2, *args, **kwargs) = ...
```
where `args=` contains a list of extra un-named positional arguments, and `kwargs=` contains an object with named keyword arguments and their values.

MichaelAtOz:
Built-in `iskey(obj,key)`? Proper would be has key, but for the sake of congruency...

I'd leave varargs for a different discussion.

revarbat:
`is_*()` checks data type.  I'd keep with `has_key()` for testing is an object has a key.

revarbat:
Current state of the PR:
```openscad
obj1 = object(the=32, quick="foo", brown=[4,5,6]);  // creating an object.
obj2 = object(obj1, fox=false, jumps=21); // Adding entries to a copy of obj1.
obj3 = object(obj2, [["over",25], ["fwee",12], ["da",true], ["fie",17]]); // Alternate add syntax.
obj4 = object(obj3, fox=undef, quick="bar"); // Altering entries in a copy of obj3.
obj5 = object(obj4, [["brown",[2,8,3]],["the",33]]); // Alternate alteration syntax.
obj6 = object(obj5, [["fwee"],["da"]]);  // Remove "fwee" and "da" entries from a copy of obj5.
fox_exists = has_key(obj6,"fox"); // Verify key "fox" is in obj6.  (true, in this case.)
fox1 = obj6.fox;  // Gets the value associated with the key "fox" in obj6.  (undef, in this case.)
fox2 = obj6["fox"]; // Also gets the value associated with the key "fox" in obj6.  (undef, in this case.)
obj7 = object(obj6, [["lazy",64],["fie"]], dog="done");
echo(obj7=obj7);  //ECHO: obj7 = { the = 33; quick = "bar"; brown = [2, 8, 3]; fox = undef; jumps = 21; over = 25; lazy = 64; dog = "done"; }
```

jordanbrown0:
Hmm.  Building deeply nested structures is straightforward, albeit tedious without the real object construction syntax.  Replacing one element deep inside a nested structure... yeah, yuck.  You'd have to peel out each layer's object, and reconstruct that subtree.

One could imagine an entry looking like `[n1, n2, n3..., v]` implying creating or updating intervening objects and updating instead of replacing.  Unlike most of the rest of this proposal, I'm not immediately sure that I'm comfortable with the semantics there.  Maybe, but only maybe, if you have object(o1, o2), and o1 and o2 are both objects that have a member m that is an object, it should update o1.m with o2.m rather than replacing it.  Again, I'm not immediately comfortable.

If you want to replace a.b.c.d, I think you would currently need
```
object(a, b=object(a.b, c=object(a.b.c, d=newval)))
```
which isn't as awful as I thought it would be, but took me a long time to work out.

revarbat:
>    object(a, b=object(a.b, c=object(a.b.c, d=newval)))

That's honestly a lot better than I was expecting.

jordanbrown0:
I would leave the "does this key exist" and varargs for another project.  And then there's sort of the inverse of varargs:  given an object {d=10; h=10; $fn=20; center=true;}, construct a cylinder.  (It should work for all cylinder parameters, not just those four.)  Also for a follow-on project.

We are kind of stuck with the possibility of object members with a value of `undef`, and that being different than them just being not present.  I don't think that's a good thing; I think it would be better if those two cases were indistinguishable.  Having multiple kinds of nothingness confuses people.  People have a hard enough time with empty arrays and empty strings.

jordanbrown0:
Also, if your function/module makes a distinction between "element does not exist" and "element has a value of undef", I would say that you should change your design...

jordanbrown0:
Can you tell whether or not a particular argument to a module was supplied?

revarbat:
I am of two minds.  The main opposing arguments:
1. Pro undef-in-object: If you can store undef in a "variable", you should be able to store undef in an object, which is like a little package of "variables".
2. Con undef-in-object: If `obj["nonexistent"]` returns undef, then storing an undef into an object should delete that key.

MichaelAtOz:
```
// notHere - doesn't exist
echo(notHere=notHere, is_undef(notHere) );
undefHere=undef;
echo(undefHere=undefHere, is_undef(undefHere) );
/*
WARNING: Ignoring unknown variable 'notHere' in file , line 2
ECHO: notHere = undef, true
ECHO: undefHere = undef, true
*/
```

You can't tell the difference between a non-variable and an undef variable, apart from the warning. **

I vote for 1. store `undef` in 'identifier' in an object. Much useful.
If a function can return `undef` you need to be able to store that result, equally in a normal variable or in an object.
What does `is_object(object( [ ["nonexistent"] ] ) // an object with nothing` return?

** please be judicious with any warnings, I hate warnings where it is legit but perhaps unusual usage.
Legit like `obj["nonexistent"]` or in above `object( [ ["nonexistent"] ] )` - no 'Warning: can't remove non existing element'.
(perhaps we need an `is_identifier()` - I didn't want to use the incorrect term 'variable')

jordanbrown0:
>  What does is_object(object( [ ["nonexistent"] ] ) // an object with nothing return?

There are two questions you might be asking here.

1.  What does `object([["anything"]])`, a request to start with nothing, and remove "anything" from it, return?  My inclination is that removing a member that isn't there is a no-op.  *Maybe* it's a warning case.
2. What does `object()`, a request to create an object without supplying any members, return?  That one is easy:  an object with no members.  Definitely *not* a warning case.

---

With respect to `obj["nonexist"]` or equivalently `obj.nonexist`, my immediate reaction is that by analogy with `list[out_of_bounds]`, `list.z` when there are fewer than three entries, and failing to supply arguments to a module or function, the result is `undef` with no warning.  I suppose that if we had a `has_key(obj, key)` I could maybe believe in a warning, but I don't think I'd like it.  (Those who follow my positions closely might notice an inconsistency:  normally I'm in favor of calling things that might be bugs errors, and requiring more explicit action if you want to do something weird.    My immediate reaction is *not* to do that here, because of the analogies above and because of the parallels to JavaScript.  OTOH, Python considers list[out_of_bounds] and dict["nonexist"] both to be errors.)

kwikius:
Surely this is dependent on  resolving https://github.com/openscad/openscad/pull/3087.  You cant really return an object without defining what an object  is?

kwikius:
for most uses the syntax seems obscure.
```openscad
// creating an object.
obj2 = {the=32; quick="foo"; brown=[4,5,6];};
//obj1 = object(the=32, quick="foo", brown=[4,5,6]);
```

jordanbrown0:
@kwikius We have had internal support for objects since #3684 textmetrics.  That included the ability to create them internally and to retrieve values from them, including echo and iteration.  What we don't have is the ability to create them from OpenSCAD programs.

The plan is to add something like #3087 to allow you to create them with syntax something like
```
o = { the=32; quick="foo"; brown=[4,5,6];};
```
... but that syntax doesn't answer two key questions:
* How do you create keys that aren't valid identifiers?  We want to use this as a replacement for what some languages would call dictionaries, which means arbitrary keys.  (Or at least arbitrary string keys.)
* How do you merge two objects, or take an object and create a new one with one key added, replaced, or removed?

In conventional languages those are answered through mechanisms like `delete obj.key` or `obj[key] = value`, but because everything in OpenSCAD is immutable those are not available; we need a way to create an object that's more flexible than the syntactic mechanism.

kwikius:
> @kwikius We have had internal support for objects since #3684 textmetrics. That included the ability to create them internally and to retrieve values from them, including echo and iteration. What we don't have is the ability to create them from OpenSCAD programs.
> 
> The plan is to add something like #3087 to allow you to create them with syntax something like
> 
> ```
> o = { the=32; quick="foo"; brown=[4,5,6];};
> ```
> 
> ... but that syntax doesn't answer two key questions:
> 
>     * How do you create keys that aren't valid identifiers?  We want to use this as a replacement for what some languages would call dictionaries, which means arbitrary keys.  (Or at least arbitrary string keys.)
> 
>     * How do you merge two objects, or take an object and create a new one with one key added, replaced, or removed?
> 
> 
> In conventional languages those are answered through mechanisms like `delete obj.key` or `obj[key] = value`, but because everything in OpenSCAD is immutable those are not available; we need a way to create an object that's more flexible than the syntactic mechanism.

Have you thought of an  array style syntax:
```openscad
ar = [ a = 1, b="hello", f = f ];  
```
It is possible to change  the  vector_element in the grammar to 
vector_element
        : list_comprehension_elements_p
        | expr
        | TOK_ID '=' expr
         {
           $$ = $3; free($1);  // return somethng else, just to get it working
           }
        ;
Tested and works ok

then you need some mechanism to get an index from a named element and then range, slice , etc

kwikius:
The immutability is no great problem as can be seen from some C++ style template metaprogramming, which  can be used for all these tasks. If you want to ad done to a list, you just copy the list with the element added, etc.  The inefficiency isnt a great concern think since most time is taken in rendering

jordanbrown0:
Had not considered an array-based syntax.  Yes, it would work, but so will the { ... } syntax, and that can then be identical to what I think I would call "statement context", rather than being a new thing.

But also:  an array-based syntax wouldn't help solve the problems that this PR addresses.

jordanbrown0:
Also:  with only one of the "explicit" syntaxes - `{ a = 1; b = 2; }` or `[ a = 1, b = 2 ]` - there is no obvious way to add, replace, or remove an entry.  (Maybe if you also map them closely to arrays and use list comprehensions, however they might apply.  But it isn't instantly obvious how a list comprehension would work for the purpose.)

kwikius:
> Also: with only one of the "explicit" syntaxes - `{ a = 1; b = 2; }` or `[ a = 1, b = 2 ]` - there is no obvious way to add, replace, or remove an entry. (Maybe if you also map them closely to arrays and use list comprehensions, however they might apply. But it isn't instantly obvious how a list comprehension would work for the purpose.)

You seem to be arguing to be allowed to modify variables. That is quite a large breaking change.
```openscad
x = 1;
y = x;
assert(y == 2);
x = 2;
```

kwikius:
Are you not going to break the code of anyone who has defined object(..) or has_key(...) as a function?

kwikius:
I have to say I like the array approach
An array is a sequence and there are many standard operations on sequences. Also try to avoid breaking existing code
```openscad
obj = [a=1,b=2,c=3];
assert(obj.has_member(a));
assert(! obj.has_member(x) );
assert(obj1.index_of(b) == 1);
obj1a = obj.range(1,2);  
assert(obj1a==[b=2,c=3]);
obj1b  = ob1.push_back([d = 4, e = 5]);

sum = obj1b.fold(0,f1= function(a,b) a+b);
assert(sum == 15);

etc
```

jordanbrown0:
> You seem to be arguing to be allowed to modify variables. That is quite a large breaking change.

Sorry, no, I used imprecise terminology.  When I said "add, modify, or remove" I really meant making a copy with those modifications.

> Are you not going to break the code of anyone who has defined object(..) or has_key(...) as a function?

Strictly, no.  User-defined modules and functions override builtin ones.  However, practically, they'll eventually want to rename theirs.  That's an unfortunate corollary of adding any new built-in modules, functions, or $ variables.

> An array is a sequence and there are many standard operations on sequences.

But objects / dictionaries / associative-arrays are *not* sequences.

revarbat:
- I would argue that removing a non-existent key from an object would be a NO-OP *without* a warning.
- `object()` without any args should return an empty object.  I would also argue that an empty object should be considered `false` for boolean tests.  (ie: `if (!obj) ...`)
- For `{key=val; ...}` (or `[...]`) syntax, you could modify from a copy of another object like `{each object; key=value, ...}`
- The `{key=val; ...}` (or `[...]`) syntax is clean and generally preferable, but does not address the need for calculated keys. It also might only address key deletion if we sacrifice the ability to store undef in an object. (ie: `{each obj; key=undef}`)  I like the idea of having `{key=val; ...}` as the preferred literal syntax, and `object()` as the more flexible solution for calculated keys.  (ie: `obj2 = object(obj1, [["foo"], [str("item",itemnum), itemval]]);`)

revarbat:
Actually, thinking on it, I would argue against using array syntax for object construction.  It makes it far too simple to make mistakes which mix list and object creation, and it may not be immediately visually obvious.  Also, I think some thought may need to be put into how to make object-comprehensions.

```openscad
obj = [
    foo = 43,
    bar = [5,6,3],
    [23,17,9]
];
```

revarbat:
Or more insidiously:
```openscad
obj = [
    foo = 43,
    bar = [5,6,3], [
        fee=23,
        fie=17,
        foe=9
    ]
];
```

jordanbrown0:
My first though is that an empty object should be truthy, because that's the way that it is in JavaScript.  But OpenSCAD is not JavaScript, and in OpenSCAD (unlike JavaScript) an empty array is falsey.  Right now, an empty object is truthy.  (You can get one with JSON import.)

Yes, we should probably think about object comprehension.  That thought process might well lead to "no", because it seems like object comprehension must necessarily involve calculated keys, and it's not at all obvious how to fit calculated keys into a reasonable syntax-based construct.  "each" works, but "for" and "if" wouldn't work because there's no way to supply different keys.  Note that `object()` will indirectly give you object comprehension, because you can use list comprehension to create a list that you can then give to `object()` to convert into an object.  Also remember that `{...}` syntax, at least in my proposal, is OpenSCAD statement context... and would we want to try to leak comprehension into that context?

revarbat:
This aaaaaaaaalmost works.  We're one proper `this` keyword away from being able to write object oriented OpenSCAD code.

```openscad
Person = object(
    new = function(first_name, last_name, ssn) let(
        this = object(
            first_name = first_name,
            last_name = last_name,
            ssn = ssn
        )
    )  object(this,
        full_name = function() str(this.last_name, ",", this.first_name),
        hello = function() echo(str("Hello, ",this.full_name()),  // Falls apart here.
    )
);

p1 = Person.new("John", "Doe", 123-45-6789);
p1.hello();
```

revarbat:
I take it back.  We *can* do full object oriented with just a bit of redundancy.
```openscad
Person = object(
    new = function(first_name, last_name, ssn)
        object(
            Person,
            first_name = first_name,
            last_name = last_name,
            ssn = ssn
        ),
        full_name = function(self) str(self.last_name, ",", self.first_name),
        hello = function(self) echo(str("Hello, ",self.full_name()),
    )
);

p1 = Person.new("John", "Doe", 123-45-6789);
p1.hello(p1);
```

nophead:
Using an object to make a class that makes objects!

The repetition would get a bit tedious though. Currently I have screw_clerance_radius(M3_pan_screw) it would become M3_pan_screw.clearance_radius(M3_pan_screw) .

revarbat:
Actually, never mind, I see flaws with that.  You would have to only set `$this`  when *calling* the function literal dereferenced from the object.

adam-trhon:
The repetition could be solved by introducing the `->` operator. If `p1->hello("world")` is a syntactic sugar for `p1.hello(p1, "world")`, then the consistency of accessing attributes stays and we have a convenient way to call methods.

nophead:
Or perhaps look for a $self parameter declaration in the function and fill it in if found.

kwikius:
> ```scad
> obj = [
>     foo = 43,
>     bar = [5,6,3],
>     [23,17,9] // Danger, Will Robinson!
> ];
> ```
not really  lost, still accessible by index ...
```scad
obj = [
    foo = 43,
    bar = [5,6,3],
    [23,17,9] // Danger, Will Robinson!
];
echo(obj[2]);
```
would output
ECHO: [23, 17, 9]

kwikius:
> ```scad
> obj = [
>     foo = 43,
>     bar = [5,6,3], [
>         fee=23,
>         fie=17,
>         foe=9
>     ]
> ];
> ```

obj = [
    foo = 43,
    bar = [5,6,3], [
        fee=23,
        fie=17,
        foe=9
    ]
];

echo(obj[2]);
 output
ECHO: [23, 17, 9]

kwikius:
> > An array is a sequence and there are many standard operations on sequences.
> 
> But objects / dictionaries / associative-arrays are _not_ sequences.

 with  (optionally) named assignments  an array gets to be a sequence and an associative container. You can also add names to existing array elements without breaking old lookup that use indices

kwikius:
In fact the named array element style is useful documentation, but would also help to avoid having to have external enums
(Following using the proposed ModuleLiteral as the array r-values)
```scad
choices = [
  square = { translate([-5,-5,0]) cube([10,10,10]);},
  cylinder = module cylinder(d = 10, h = 10),
  triangle = module triangle_prism(10,5),
  octagon = module octagon_prism(10,5)
];
shape2 = choices[choice];
// shape2 = choices.square ;
color("blue"){
  translate([10,0,0]) shape2();
}
```
Can be tried out at 
https://github.com/kwikius/openscad/blob/WIP-first-class-module/examples/ModuleLiterals/shape_choice.scad#L37

revarbat:
>> ```scad
>> ```scad
>> obj = [
>>     foo = 43,
>>     bar = [5,6,3],
>>     [23,17,9] // Danger, Will Robinson!
>> ];
>> ```
> 
> not really  lost, still accessible by index ...
> ```scad
> obj = [
>     foo = 43,
>     bar = [5,6,3],
>     [23,17,9] // Danger, Will Robinson!
> ];
> echo(obj[2]);

Parser-wise, that would really clearly imply that you would have to access object elements after dereferencing it's list position:

```openscad
obj = [
    foo = 43,
    bar = [5,6,3]
];
echo(obj[0].foo);
```

revarbat:
In any case, `[key=val, ...]` vs. `{key=val; ...}` syntax discussion, while an interesting diversion, is probably better on-topic in #3087 or #3956.  Or at https://lists.openscad.org/empathy/thread/3DVXSTCBWJPKPUVBJBFMMQ27BVYUS352

kwikius:
> > > ```scad
> > > obj = [
> > >     foo = 43,
> > >     bar = [5,6,3],
> > >     [23,17,9] // Danger, Will Robinson!
> > > ];
> > > ```
> > 
> > 
> > not really  lost, still accessible by index ...
> > ```scad
> > obj = [
> >     foo = 43,
> >     bar = [5,6,3],
> >     [23,17,9] // Danger, Will Robinson!
> > ];
> > echo(obj[2]);
> > ```
> 
> Parser-wise, that would really clearly imply that you would have to access object elements after dereferencing it's list position:
> 
> ```scad
> obj = [
>     foo = 43,
>     bar = [5,6,3]
> ];
> echo(obj[0].foo);
> ```
I don't see the parsing issue? You can try something equivalent in the current language:

```openscad
obj = [
   /* x = */ 43,  
   /* bar = */ [5,6,3]
];

echo( obj[0] );
echo( obj.x );  // Cheating ;)
echo(obj[0].foo);
```

output:
ECHO: 43
ECHO: 43
ECHO: undef

thehans:
I have a number of notes and concerns on the topic of objects, some of them I brought in IRC during some back and forth with @jordanbrown0, but I figured I would recount them here for posterity and so more eyes can see.

First, I feel strongly that an object-literal syntax *on its own* should be capable of representing the full range of objects (including those with "generated" keys).  For this reason I would really rather see us going towards a Python dictionary-like syntax, which is also very similar to Javascript object literals.  Namely that:
     - The object literal is enclosed by curly braces `{}`.  (this seems to mostly be the consensus already)
     - the key and value should be separated by a colon `:`.  This very clearly distinguishes itself from an assignment, and allows for r-values as keys.  There is also some precedent of similar syntax in our customizer drop-downs with labeled entries.
     - and entries are separated by commas `,` (same as lists, let expressions, parameters, and arguments)

IMO, one important aspect of our value types *thus far* is that, when echo'd, they are output such that it **is a valid OpenSCAD literal**.  (FWIW, Python also seems to represent dictionaries in this way, as far as I can tell).  I would prefer not to break that for objects.
If an `object()` function is the only way to make string based keys, then how would you echo an object as a literal where e.g. a key contains a space?

I know this format goes directly against Jordan's proposal where every element of an object *must be* a statement.  I'm sorry I don't have a better response or solution to that, other than, maybe its a sign that geometry and dictionary-like objects should be two distinct types?

As for the question of object/dictionary comprehensions, I think it's possible to do this in a way that adds a lot of power in terms of generated values, and also correlates well with existing list-comprehensions (while remaining distinct enough, due to curly braces, to discern the difference).
Again, Python has [dictionary comprehensions](https://peps.python.org/pep-0274/), which I think would be a good example to take ideas from.  Also see the more [technical reference detailing the the exact grammar here](https://docs.python.org/3/reference/expressions.html#dictionary-displays)
Their syntax in its simplest form is more-or-less:
```
{ key_expression : val_expression for iter in iterable }
```
Here is one simple example
```Python
>>> print {i : chr(65+i) for i in range(4)}
{0 : 'A', 1 : 'B', 2 : 'C', 3 : 'D'}
```
Note that Python dictionaries allow keys of *any* type, not just strings.  But I think we would be fine to just coerce the value into a string (as Javascript does, iirc).

For our purposes, I would re-order the `for` and the `key:value` to be consistent with our list comprehensions.
```
{ for(iter = iterable) key_expression : val_expression }
```

`each` would be straightforward to just copy elements into the new object  (I think it should probably be an error if object comprehension's `each` is attempted on an array or string rather than an object)
This would make appending elements very simple:
```
{ each old_obj , new_key1: new_val }
```

I should also mention that Python has a convenient tuple unpacking syntax which facilitates dictionary comprehension (though that is not strictly a prerequisite for them).
So a key and value can be assigned at the same time from a `for` clause, as in:
```Python
{ k : v for (k, v) in dict.items() if <condition> }
```
I'm not sure the best way we could handle that situation of looping over an object.
But also note they allow for `if` as a filtering mechanism (same as our list-comprehensions), which would be equivalent to "deleting" items, as has been discussed here.

```OpenSCAD
{ for(iter = iterable) if (condition) key_expr : val_expr }
// and of course we could also do if/else  which is very nearly equivalent to the conditional ternary operator `? :`
{ for(iter = iterable) if (condition) key_true : val_true else key_false : val_false } 
```

thehans:
In summary, to me an `object()` function feels like an unattractive stopgap / band-aid in lieu of a proper object-literal syntax.
Why should object be the only type that requires a special builtin function/"constructor"?

revarbat:
While I'd love to see a pretty object literal syntax, experience tell me that the fancier you make it, the more years get added before it becomes available to users.  I just don't have too many years left in me to wait for it.

jordanbrown0:
Python syntax does have the advantage of being friendly to calculated keys.  It also has the distinct disadvantage of being unfriendly to *constant* keys.  You can't say `{a:1, b:2}`.  You must say `{'a':1, 'b':2}`.  JavaScript syntax is friendly to constant keys with any values, because it allows either an identifier or a constant string as the key.  That also allows the string form of an object to always be parsable as input.  But it doesn't help with calculated key values; there JavaScript relies on assigning values into the object, a mechanism that OpenSCAD doesn't have available.

BTW, OpenSCAD echo format is not completely parsable.  It doesn't handle `"\""`, `"\n"`, or `"\\"` correctly.

One of the reasons for my proposal was to try to avoid having two distinct types with somewhat similar syntaxes.  I suspect that trying to do that would require putting a prefix in front of one of them, much as JavaScript puts `function (args)` in front of inline functions.  It sure seems like it would be nice to be able to say `x = { sphere(10); }` without needing much more text than that.  (Noting that the intent there is to evaluate the sphere() into geometry right there, not store away a module reference to be evaluated later.)  "Module literal" syntax would throw another wrinkle into the picture.

The object comprehension syntax you propose seems plausible, give or take the question of whether the left side of the colon is an rvalue or an identifier.  I strongly agree that the key/value to be generated belongs on the right side of the for/if.  You shouldn't have to look to the right of a construct to see whether or not you're going to use it, or whether you might use it more than once.

In the short term, yes, object() would be the only way to create objects.  That's because it doesn't require any new syntax; its semantics are easy to define.  In the long term it's true that some of its uses could be replaced by object syntax... on the other hand, I believe that every single use of `concat()` can be replaced by list and list comprehension syntax and yet I suspect that many prefer concat() because it does simple cases simply.

Net, I think that object() provides useful functionality in a way that clearly works (give or take the echo-output-is-parsable question), without the need to solve all of the syntactic and semantic problems that circle around object literals, geometry-as-data, and module literals.  Absent object(), it would seem that we can't have user-constructed objects without having a pretty firm grasp of all of those other components.

kwikius:
I am pretty sure you can do much of this in the current language. 

```openscad
find = function(dict, key, pos = 0) 
    (pos < len(dict) )
       ? (dict[pos][0] == key)
          ? dict[pos][1]
          : find(dict, key , pos +1)
       : undef
;

d = [
  ["a",1],
  ["b",2],
  [0,"hello"],
];

res1 = find(d,"b");
echo(res1);

res2 = find(d,0);
echo(res2);

res3 = find(d,-1);
echo(res3);

/* output:
ECHO: 2
ECHO: "hello"
ECHO: undef
*/
```
What problem is this (and indeed the object) trying to solvethatcan't be solved in the existing language?
Could we have a real world example?

kwikius:
What is the difference between an object and a module, apart from the fact you can't currently access the members of a module?
https://github.com/openscad/openscad/pull/3087#issuecomment-1239275040

revarbat:
@kwikius While BOSL2 already has a user-space implementation of objects that we call structs, there is a vast difference between the O(n) efficiency of a user-space implementation and the O(1) efficiency of a builtin.  I could rewrite my VNF validation code to be orders or magnitude faster if I could use built-in objects.

revarbat:
Of note, in BOSL2 we implemented a [hashmap](https://github.com/revarbat/BOSL2/wiki/fnliterals.scad#function-hashmap)  using function literals, but interpreter overhead means it only becomes faster than `search()` at more than 4000 items size.

kwikius:
BOSL2 is very impressive, but[ in my useage](https://github.com/kwikius/dlg) of OpenSCAD, I havent found the need to spend a great deal of time optimising things. I mostly find a few seconds parsing time is dwarfed by the rendering time.

revarbat:
One of my biggest problems with efficiency is in VNF validation. Validating VNFs with `vnf_validate()` is extremely similar to validating STLs, and becomes very slow (multiple minutes runtime) when working with non-trivial shapes. It is a powerful debugging tool, though. I have used `vnf_validate()` to debug construction of complicated shapes like threads, spiral beveled gears, and `linear_sweep()`/`rotate_sweep()` with surface texturing.  I have to simplify my cases as much as possible to even get a reasonable runtime, though.  Think screws with only 3 sides, gears with 5 teeth.  Sadly, such simplified shapes sometimes fail to expose errors that a larger model will run afoul of.

Rendering the shape and showing where the errors are is far far faster than that analysis calculation.  By about 3 orders of magnitude.

Creating shapes with textured surfaces via `linear_sweep(texture=)` or `rotate_sweep(texture=)` is another case where construction of the shape is far slower than previewing or rendering the result.  (By texturing, think things like having knurling, or a brick wall texture.)  And you cannot really do what these functions do, using pure geometry.

So, efficiency actually DOES matter.

kwikius:
> One of my biggest problems with efficiency is in VNF validation. Validating VNFs with `vnf_validate()` is extremely similar to validating STLs, and becomes very slow (multiple minutes runtime) when working with non-trivial shapes. It is a powerful debugging tool, though. I have used `vnf_validate()` to debug construction of complicated shapes like threads, spiral beveled gears, and `linear_sweep()`/`rotate_sweep()` with surface texturing. I have to simplify my cases as much as possible to even get a reasonable runtime, though. Think screws with only 3 sides, gears with 5 teeth. Sadly, such simplified shapes sometimes fail to expose errors that a larger model will run afoul of.
> 
> Rendering the shape and showing where the errors are is far far faster than that analysis calculation. By about 3 orders of magnitude.
> 
> Creating shapes with textured surfaces via `linear_sweep(texture=)` or `rotate_sweep(texture=)` is another case where construction of the shape is far slower than previewing or rendering the result. (By texturing, think things like having knurling, or a brick wall texture.) And you cannot really do what these functions do, using pure geometry.
> 
> So, efficiency actually DOES matter.

It is again impressive, but the obvious question is why do it in Openscad? ( I would also point out that the source data is an array, so the object () function doesnt seem to be relevant here?
I mean we would be doing 
```openscad
vnf = object( vertices=[...],faces = [...] );
```
It would seem easier just to write the array direct?  Note with an extended array syntax it works nicely:
```openscad
vnf = [ vertices=[...], faces = [...] ],
```
Overall though for a big dataset I would have chosen another language ( in my case C++, since I know it well)

For example I wouldnt consider using openscad (or python or java for that matter) to generate a wing shape
https://github.com/kwikius/f3res/blob/master/outer.stl
I use a c++ application and  script
https://github.com/kwikius/f3res/blob/master/outer_panel.wing_stl

I found it quite easy to convert between C++ and openscad since the syntax is quite similar
https://github.com/kwikius/Trilateration/blob/master/trilateration.scad
https://github.com/kwikius/Trilateration/blob/master/trilateration.cpp

and of course to write a c++ app that output openscad source
https://github.com/kwikius/Trilateration/blob/master/trilateration_transform.cpp#L263

jordanbrown0:
> @thehans What is the difference between an object and a module, apart from the fact you can't currently access the members of a module?

An object is a data structure.  A module is an executable subprogram.

An object is analogous to a JavaScript object, a Python dictionary, a Python object, or a C structure.
A module is analogous to a JavaScript function, a Python function, or a C function.

jordanbrown0:
Why objects instead of arrays with name-value pairs?  Because `obj.member` is a heck of a lot more convenient than `find(obj, "member")`.

A fontmetrics() object looks like so (reformatted for readability):
```
{
  nominal = {
    ascent = 12.5733;
    descent = -2.9433;
  };
  max = {
    ascent = 13.6109;
    descent = -4.2114;
  };
  interline = 15.9709;
  font = {
    family = "Liberation Sans";
    style = "Regular";
  };
}
```
The nominal ascent is `fm.nominal.ascent`, or `find(find(fm, "nominal"), "ascent")`.  I known which one I'd rather type and read.

revarbat:
Since the shapes I'm debugging are *generated* in OpenSCAD, it's useful and a lot simpler to *debug* them in OpenSCAD.
And `object()` could be used to vastly speed up detection of non-manifold holes or extra faces sharing an edge, among other things.

kwikius:
> > @thehans What is the difference between an object and a module, apart from the fact you can't currently access the members of a module?
> 
> An object is a data structure. A module is an executable subprogram.

When I asked originally  I got a different answer:
https://github.com/openscad/openscad/issues/4336#issuecomment-1229522281

I asked again on the official objects thread #3087 since there seemed to be various answers and got no answer
https://github.com/openscad/openscad/pull/3087#issuecomment-1239275040

I'm not too bothered how they are rep'ed,  but it seems that objects have crept into the language, without knowing their semantics and  with no way to write them in source code ( hence this proposal which still doesnt answer the OP  basic questions about them)

And that is a problem!

kwikius:
> Since the shapes I'm debugging are _generated_ in OpenSCAD, it's useful and a lot simpler to _debug_ them in OpenSCAD. And `object()` could be used to vastly speed up detection of non-manifold holes or extra faces sharing an edge, among other things.
I don't understand how using maps rather than arrays will speed things up, though. I would have thought the opposite would be the case?

jordanbrown0:
Let's take the overall design question over to #3087.

One tidbit that we might be able to close out here:
> I don't understand how using maps rather than arrays will speed things up, though. I would have thought the opposite would be the case?
Presumably that would depend on whether your data is naturally searched by an index, or by some other kind of key.  There are good reasons why practically every modern language offers associative arrays either directly in their core or in a standard library.

jordanbrown0:
Note that the part starting "Presumably" shouldn't have been included in the quote.  Edited.

revarbat:
> I don't understand how using maps rather than arrays will speed things up, though. I would have thought the opposite would be the case?

Maps have O(1) lookup time.  Lists have O(n) lookup time.  That means if you are working with 1000 items, the map will take a short fairly constant lookup time, while the list will have to search the 1000 items linearly.

To detect holes in a VNF, you just need to find edges which are not shared with any other face.  To detect multiply connected edges, you look for edges which are shared by more than two faces.  This can be sped up a lot by using a map keyed on edges (which is just a pair of vertex indexes, in sorted order), with a list of face numbers that share that edge as the value.

src/core/Value.cc
```diff
@@ -1320,9 +1320,46 @@ const Value& ObjectType::get(const std::string& key) const
...
+  if (ptr->map.find(key) == ptr->map.end()) {
+    ptr->map.emplace(key, std::move(value));
+    ptr->keys.emplace_back(key);
+    ptr->values.emplace_back(std::move(value));
+  } else {
+    ptr->map.erase(key);
+    ptr->map.emplace(key, std::move(value));
+    for (int i = ptr->keys.size() - 1; i >= 0; i--) {
+      if (ptr->keys[i] == key) {
+	ptr->values[i] = std::move(value);
```
ntfshard, line 1332: access to moved object (already moved at 1329)

sandeep-datta:
@revarbat are you still working on this?

revarbat:
> @revarbat are you still working on this?

Not very actively. I can fix the noted bug if you need, but this PR kinda got pulled into a more comprehensive PR with other object creation syntaxes and module literals. I'm unclear if any fixes here will get merged into the other PR, and which PR will eventually get used anyways.

jordanbrown0:
That larger PR is #4478.

revarbat:
This has been included into mainline elsewhere, so I'm closing this.

PR #4346, opened 2022-09-03
Update GEB.scad

UxuginPython:
Change font to Liberation Mono

### DIFF
--- a/examples/Advanced/GEB.scad
+++ b/examples/Advanced/GEB.scad
@@ -1,6 +1,4 @@
-font = "Liberation Sans";
-// Nicer, but not generally installed:
-// font = "Bank Gothic";
+font = "Liberation Mono";
 
 module G() offset(0.3) text("G", size=10, halign="center", valign="center", font = font);
 module E() offset(0.3) text("E", size=10, halign="center", valign="center", font = font);


t-paul:
The tests include the example files for verification so if we change this, the test image needs to be updated too. From the branch name it sounds like you changed this via the Web interface so I'm not sure if you would be setup to update the image. I can take care of that.

Otherwise it sounds like a sensible change, but can you still give a note why you propose it?

UxuginPython:
I don't have much experience with tests and images. I did do it through the GitHub web interface. As for the reason, changing to a fixed-width font could make it easier to use letters other than G, E, and B. I chose Liberation Sans because it's another Liberation font and also because it's one of the nicer fixed-width fonts.

UxuginPython:
This is a two-year-old pull request for an extremely minor change that never really made sense in the first place. I haven't looked at this in a really long time, but if I remember right, Liberation Mono kind of cuts off the tops of the letters, which Liberation Sans does not. I don't see any good in this pull request, which would almost certainly never have activity again, staying open, so I am now closing it. If, for whatever reason, this pull request is rediscovered and the change is (however unlikely) still desired, contact me if you want, or just make your own pull request. I don't care about getting credit for something this small.

Issue #4340, opened 2022-08-30
Removing holes from 2D geometries


nickzoic:
I had a need to get an outline around a 2D geometry created partly from dynamic text.  The usual `offset(-n) offset(n)` doesn't follow closely enough around the outline and doesn't fill all the holes.  What I'm looking for is a outline which fits tightly around all external edges of the geometry but ignores the internal edges.

Searching the web, I found a few people[^1][^2] asking if this was possible but no workable solutions ...

An alternative to *removing* the holes would be an operator which *found* the holes (so I could deal with them separately).

[^1]: https://forum.openscad.org/How-to-get-an-outline-of-a-2D-object-td30564.html
[^2]: https://computergraphics.stackexchange.com/questions/5333/openscad-using-outline-of-2d-shape

nickzoic:
... I've had a couple of sketchy goes at implementing this:

* as an `outline()` operator [^1]  which was simple but I don't like to add keywords to the syntax!
* as a `remove_holes` flag to `offset()` [^2] which is a little uglier but should be more compatible.
  it's pretty likely this will be used with a small offset radius as well, so combining the operations makes sense.

[^1]: https://github.com/nickzoic/openscad/tree/nick-outline-1
[^2]: https://github.com/nickzoic/openscad/tree/nick-offset-remove-holes-1

nickzoic:
Another way of expressing the same idea just occurred to me: `offset()` could allow a different `r` for internal and external offsets.

* `offset(r=0, r_internal=100)` would not expand the outside but would remove the holes.
* `offset(r=5, r_internal=0)` would expand the outline but leave the holes the size they were.

etc, or `internal` and `external` could be boolean flags (default true) to indicate which types of outline this offset will run on:

* `offset(r=100, external=false)` would not expand the outside, but would remove the holes.
* `offset(r=5, internal=false)` would expand the outline but leave the holes the size they were.

nickzoic:
From @t-paul on the mailing list:
>
> I've seen this discussed before and I think there's no
> general solution with the currently available features.
> 
> So a PR implementing this would be great. I would suggest
> to separate this from offset() though. A dedicated fill()
> seems like a good choice.

nickzoic:
[...]

PR #4348, opened 2022-09-09
Adds a fill() 2D operator

nickzoic:
#4340 ... Adds a fill() 2D operator much like hull() which gets rid of all the holes in a 2D shape.

TODO:

- [x] Is it possible/desirable to implement a 3D fill()?
- [x] Check I haven't done anything stupid in C++ 
- [ ] Add to docs / examples.
- [x] Tests still need work to cover more cases & check for correctness.


nickzoic:
I'm ticking off "is it possible to implement a 3D fill" because yeah it might be possible but it's out of my depth for the moment so it can go be its own bug.

Along which lines, "have I done something stupid in C++?" I could use some help with.

I've added some tests for things like nested holes and fixed a bug I found as a result.

t-paul:
Nice thanks! Yes, a 3D version would be awesome, but it's better to keep it separate in any case. I'll try to review as soon as possible, very much lacking time to spare lately. The test cases already look great :+1:.

nickzoic:
On Tue, 13 Sep 2022, at 18:04, Torsten Paul wrote:
> Nice thanks! Yes, a 3D version would be awesome, but it's better to keep it separate in any case. I'll try to review as soon as possible, very much lacking time to spare lately. The test cases already look great 👍.
> 

OK thanks!  I wanted to test with all the 2D primitives but thinking about it I'm a bit worried that the default font one won't be pixel-identical on all build platforms.

t-paul:
Yes, fonts can be tricky in that regard, we even had an issue with different versions of the Liberation fonts at some point.

For that reason the test cases come with their [own dedicated font files](https://github.com/openscad/openscad/tree/822a8c457a50618b70f0c0657b40d6355972574d/tests/data/ttf) trying to minimize the risk of platform differences breaking things.
https://github.com/openscad/openscad/blob/822a8c457a50618b70f0c0657b40d6355972574d/tests/test_cmdline_tool.py#L276-L279

nickzoic:
Oh that's great, you're way ahead of my last-minute-before-bed conjectures :-)

kwikius:
@nickzoic Looks like a nice feature. Will try to review in more detail when I get time. Also used your pull request to figure out how to add tests. Thanks for that :)

src/geometry/GeometryEvaluator.cc
```diff
@@ -342,6 +355,8 @@ Polygon2d *GeometryEvaluator::applyToChildren2D(const AbstractNode& node, OpenSC
     return applyMinkowski2D(node);
   } else if (op == OpenSCADOperator::HULL) {
     return applyHull2D(node);
+  } else if (op == OpenSCADOperator::FILL) {
```
kwikius, line 371: Am wondering if this section should now be changed to a switch statement to match switch statement a little later in the function?

t-paul: Yep, might be nice, but I also don't mind keeping the if/else.

t-paul (approved):
Thanks, nice new feature!

Issue #4353, opened 2022-09-14
"ERROR: Compilation aborted by exception: Unauthorized intersections of constraints" after enabling fast-csg


thehans:
Running openscad 2022.09.16.nightly on Linux Mint 21.

When previewing the following script (which has some tweaks to resolution made by me), I get the message in Console:
<strike>"ERROR: Parsing aborted by unknown exception"</strike>
"ERROR: Compilation aborted by exception: Unauthorized intersections of constraints"

```OpenSCAD
// Fillet function
// from https://github.com/clothbot/ClothBotCreations/blob/master/utilities/fillet.scad

$fa = 1;
$fs = 0.25;

module fillet(r=1.0,steps=3,include=true) {
  if(include) for (k=[0:$children-1]) {
	children(k);
  }
  for (i=[0:$children-2] ) {
    for(j=[i+1:$children-1] ) {
	fillet_two(r=r,steps=steps) {
	  children(i);
	  children(j);
	  intersection() {
		children(i);
		children(j);
	  }
	}
    }
  }
}

module fillet_two(r=1.0,steps=3) {
  for(step=[1:steps]) {
	hull() {
	  render() intersection() {
		children(0);
		offset_3d(r=r*step/steps) children(2);
	  }
	  render() intersection() {
		children(1);
		offset_3d(r=r*(steps-step+1)/steps) children(2);
	  }
	}
  }
}

module offset_3d(r=1.0) {
  for(k=[0:$children-1]) minkowski() {
	children(k);
	sphere(r=r);
  }
}

r = 2;
steps = ceil(2*PI*r/4/$fs); // rough estimated fragments for a 90 degree fillet
fillet(r=r,steps=steps) {
	cylinder(r=5,h=10);
	cube([10,10,2]);
	rotate([30,30,30]) cylinder(r=1.0,h=50,center=true);
}
```

Here is a view of which experimental features I have enabled:
(Disabling the topmost "fast-csg" feature allows the script to preview without error)
![Screenshot from 2022-09-14 17-54-38](https://user-images.githubusercontent.com/252233/190278561-75a175a9-672c-49c0-a38d-7ca37512deaf.png)

rcolyer:
With #4354 this is revealed to be "ERROR: Compilation aborted by exception: Unauthorized intersections of constraints"

A perusal of the CGAL source finds this explanatory comment for the exception condition in Constrained_triangulation_2.h:
```
[...]

PR #4354, opened 2022-09-15
Catch and log std::exception for compilation

rcolyer:
In response to #4353, we really need to be catching std::exception and logging to the console its .what() message for exception throws during compilation, which this PR does.  The unknown exception remains as a fallback.

### DIFF
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -1103,7 +1103,7 @@ void MainWindow::compile(bool reload, bool forcedone)
     compileDone(didcompile | forcedone);
   } catch (const HardWarningException&) {
     exceptionCleanup();
-  } catch (std::exception &ex) {
+  } catch (const std::exception& ex) {
     UnknownExceptionCleanup(ex.what());
   } catch (...) {
     UnknownExceptionCleanup();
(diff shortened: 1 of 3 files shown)


src/gui/MainWindow.cc
```diff
@@ -1103,6 +1103,8 @@ void MainWindow::compile(bool reload, bool forcedone)
     compileDone(didcompile | forcedone);
   } catch (const HardWarningException&) {
     exceptionCleanup();
+  } catch (std::exception &ex) {
```
t-paul, line 1106: Should we go with `const` ref?
```suggestion
  } catch (const std::exception& ex) {
```

src/gui/TabManager.cc
```diff
@@ -501,6 +501,8 @@ void TabManager::openTabFile(const QString& filename)
       par->parseTopLevelDocument();
     } catch (const HardWarningException&) {
       par->exceptionCleanup();
+    } catch (std::exception& ex) {
```
t-paul, line 504:
```suggestion
    } catch (const std::exception& ex) {
```

src/gui/MainWindow.cc
rcolyer: Ah, yes, sure.

t-paul (approved):

PR #4371, opened 2022-10-03
linear_extrude slices calculation

UBaer21:
using max slices of twist and non uniform scale to prevent reduction of slices due to twist component in non uniform extrusion

### DIFF
--- a/src/geometry/GeometryEvaluator.cc
+++ b/src/geometry/GeometryEvaluator.cc
@@ -1046,7 +1046,7 @@ static Geometry *extrudePolygon(const LinearExtrudeNode& node, const Polygon2d& 
       size_t slicesTwist;
       slicesNonUniScale = (unsigned int)Calc::get_diagonal_slices(max_delta_sqr, node.height, node.fn, node.fs);
       slicesTwist = (unsigned int)Calc::get_helix_slices(max_r1_sqr, node.height, node.twist, node.scale_x, node.fn, node.fs, node.fa);
-      slices = max(sclicesNonUniScale,slicesTwist);
+      slices = max(slicesNonUniScale,slicesTwist);
     } else { // uniform scaling with twist, use conical helix calculation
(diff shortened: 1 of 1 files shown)


src/geometry/GeometryEvaluator.cc
```diff
@@ -1034,10 +1034,19 @@ static Geometry *extrudePolygon(const LinearExtrudeNode& node, const Polygon2d&
...
-    // **** so just use this straight helix calculation anyways.     ****
-    if ((node.scale_x == 1.0 && node.scale_y == 1.0) || node.scale_x != node.scale_y) {
+    if (node.scale_x == 1.0 && node.scale_y == 1.0) {
       slices = (unsigned int)Calc::get_helix_slices(max_r1_sqr, node.height, node.twist, node.fn, node.fs, node.fa);
+    } else if ( node.scale_x != node.scale_y) { // non uniform scaling with twist using max slices from twist and non uniform scale
+      double max_delta_sqr = 0; // delta from before/after scaling
+      Vector2d scale(node.scale_x, node.scale_y);
+      for (const auto& o : poly.outlines())
+        for (const auto& v : o.vertices)
+          max_delta_sqr = fmax(max_delta_sqr, (v - v.cwiseProduct(scale)).squaredNorm());
```
rcolyer, line 1044: Let's make these {} wrapped for loops to avoid future maintenance issues.

src/geometry/GeometryEvaluator.cc
```diff
@@ -1034,10 +1034,19 @@ static Geometry *extrudePolygon(const LinearExtrudeNode& node, const Polygon2d&
       for (const auto& v : o.vertices)
         max_r1_sqr = fmax(max_r1_sqr, v.squaredNorm());
```
rcolyer, line 1035: I see you didn't edit this line, but maybe you could {} wrap these two for loops also while you're at it, since it's right there, or any others you see nearby.  (It's a good standard practice for the code to have those, in case someone goes to insert an indented line and misses that they are absent.)

PR #4376, opened 2022-10-07
Fix recent file list bug

### LINKED ISSUE #3690: 2021.01 - Save & Save-As do not populate recent file list
Create a new file, Save or Save-As, look at File/Recent-files, they don't show up.
File/Open does update the list.
On Windows & Debian.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/96790604-2021-01-save-save-as-do-not-populate-recent-file-list?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Paulwhy2:
Fixes #3690

The source of this bug was that the 'filename' field in the editor interface object does not always hold the filename and path of the recently saved file.  

I change the call to `updaterRecentFiles()` argument from a pointer to the editor interface to a QString with the recent filename and path, which is available in both of the functions which call `updateRecentFiles()`.

### DIFF
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -984,16 +984,14 @@ bool MainWindow::network_progress_func(const double permille)
   return (progresswidget && progresswidget->wasCanceled());
 }
 
-void MainWindow::updateRecentFiles(EditorInterface *edt)
+void MainWindow::updateRecentFiles(QString FileSavedOrOpened)
 {
   // Check that the canonical file path exists - only update recent files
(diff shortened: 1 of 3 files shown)


t-paul (approved):
Looks good, thanks!

PR #4380, opened 2022-10-15
FreetypeRenderer::GlyphData destruction simplification

ntfshard:
- Avoiding redundant inheritance from `std::vector`
- Releasing resource in dtor, instead of external functor, should improve readability
- Straightforward memory allocation pattern: one chunk for all GlyphData of each `text` command. Technically I found a small difference in average runtime: from 34.5 to 33.9 seconds 734 symbols of 'Lorem ipsum" text, and J5040 CPU, but I believe it's not the main goal of this patch.


jordanbrown0 (commented):
Looks reasonable to me, as one of the last people to make substantive changes to these files.  However, I have to note that I'm not really a strong enough C++ person to be sure of the changes.

t-paul (approved):
:+1:

PR #4382, opened 2022-10-15
Update to github ci to migrate to node 16 compatible versions

Scopeuk:
inline with the blog entry below project CI is generating warnings relating to the soon to be deprecated node runtime.
project ci files checked for uses: clauses and updated where current imported actions were using node 12 runtime.
https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/


t-paul (approved):
Nice, thank you!

Issue #3908, opened 2021-09-30
echo() should not include any decoration - extra text, punctuation or formatting


dannypike:
This ticket is opened in response to [#3905](https://github.com/openscad/openscad/issues/3905) which started as a discussion about adding a setting to clear the console on each build and migrated to a discussion between @t-paul, @MichaelAtOz and @jordanbrown0, about having more control over the text that is output to the log by the ```echo()``` statement.

My current position is that the ```echo()``` command should not output any text other than that which is specified in the arguments to that call. No "ECHO:" header, no commas, no double quotes, just the arguments. It should continue to support a "suitable" subset of ANSI escape sequences, such as the colours and newlines.

So this set of statements:

```
echo("\r\n   cone_narrow_od:", cone_narrow_od
, "\r\n   cone_wide_od:", cone_wide_od
, "\r\n   cone_length:", cone_length
, "\r\n");
```

which currently produces this output:

```
ECHO: "
   cone_narrow_od:", 16.4, "
   cone_wide_od:", 32, "
   cone_length:", 88.472, "
"
```

would then produce this output:

```

   cone_narrow_od: 16.4
   cone_wide_od: 32
   cone_length: 88.472
```

(Note the blank line at the start of the block.)

rcolyer:
@dannypike Does it match your use-case thinking that this is only important for command-line execution?  The "-o output.echo" already captures only echo, plus warnings, it's just that it includes the "ECHO: ".  The rest is already obtainable with the following syntax, utilizing str:

```
cone_narrow_od = 16.4;
cone_wide_od = 32;
cone_length = 88.472;
echo(str("\r\n   cone_narrow_od: ", cone_narrow_od
, "\r\n   cone_wide_od: ", cone_wide_od
, "\r\n   cone_length: ", cone_length
, "\r\n"));
```
Output of:
```
ECHO: "
   cone_narrow_od: 16.4
   cone_wide_od: 32
   cone_length: 88.472
"
```

I could certainly see value in a clean echo command-line flag which additionally purges the "ECHO: " and the quote marks.

The operative question that should be discussed in concert, is what should happen with warnings and errors?  Ignoring them falls under "very bad idea", leaving a few options like:
- Bundle them even with clean echo, and it's the scad implementer's responsibility to clean up warnings and errors if output without them is desired.  (Which feels pretty reasonable to me.)
- Provide an alternate output for warnings and errors such as to stderr, and provide no capture to file support.
[...]

PR #4384, opened 2022-10-17
Decimal round trip fixes

### LINKED ISSUE #3932: Cache representational precision errors
Following a similar issue raised on the IRC channel, the cache key text representation precision can in some cases trigger surprising errors.  I've distilled this to a testcase (Design, Flush cache, after switching between these):

cache_issue.scad
```
linear_extrude(5) translate([-1, 0, 0]) polygon([[0.9999999, 5], [0.9999999, 10], [3, 7]]);
rotate_extrude() translate([-1, 0, 0]) polygon([[1, 5], [1, 10], [3, 7]]);
```

reliably_works.scad:
```
rotate_extrude() translate([-1, 0, 0]) polygon([[1, 5], [1, 10], [3, 7]]);
linear_extrude(5) translate([-1, 0, 0]) polygon([[0.9999999, 5], [0.9999999, 10], [3, 7]]);
```

The only difference is the order of the lines, determining which is cached first.  The source of the problem is 0.9999999 being cached as 1, but while the calculation continues to happen as the double corresponding to 0.9999999.  This results in the rotate_extrude input violating the y-axis line crossing condition in the polygon incorrectly pulled from the cache.

An obvious solution is just increasing the precision representation of doubles in the caching keys to correspond to the amount used for calculations, but this will increase many cache key lengths.  I thought it best to raise this for discussion to see if there are issues for this.  If not, this is probably low-hanging fruit to fix.

thehans:
Spreading usage of double-conversion library throughout more of the code, including CSG cache keys, for more precision.
Switched to double-conversion ToShortest (shortest string representation without losing any data)

Fixes #3932

I thought this would be a quick change, and mainly wanted to see how much it affected echo tests, dump tests, etc.  
But I quickly realized that only changing the double toString within `Value.(h|cc)` did not actually handle the code paths used for CSG tree dump/cache ids etc, which required quite a bit more refactoring.

This also comes with the risk of causing more bug reports about "ugly" number representations for some computed values.

One thing that still might be worth tweaking is the decimal low/hi exponents (now found in `src/core/scadstream.h`)
```
#define DC_DECIMAL_LOW_EXP -6
#define DC_DECIMAL_HIGH_EXP 21
```

[double conversion documentation](https://github.com/google/double-conversion/blob/master/double-conversion/double-to-string.h#L108-L118) explains the usage as follows:

>  // When converting to the shortest representation the converter will
>   // represent input numbers in decimal format if they are in the interval
>   // [10^decimal_in_shortest_low; 10^decimal_in_shortest_high[
>   //    (lower boundary included, greater boundary excluded).
>   // Example: with decimal_in_shortest_low = -6 and
>   //               decimal_in_shortest_high = 21:
>   //   ToShortest(0.000001)  -> "0.000001"
>   //   ToShortest(0.0000001) -> "1e-7"
>   //   ToShortest(111111111111111111111.0)  -> "111111111111111110000"
>   //   ToShortest(100000000000000000000.0)  -> "100000000000000000000"
>   //   ToShortest(1111111111111111111111.0) -> "1.1111111111111111e+21"


thehans:
I decided to settle on exponent values of -5 and +6, which is closest to how previous behavior(master HEAD) was.
I have also changed printing of doubles so that positive sign for exponents is omitted. `1e+6` => `1e6`

thehans:
Well, I didn't realize how much the various platforms differed at the last couple significant digits.  I suspect this is down to differences in compiler optimizations, re-ordering double operations, or maybe the accuracy of trig function implementations?

The test reports appear to differ only in text based CSG dumps.  I looked at a couple of these tests and tried to correlate the csg lines to the original script.  It seems that they differ mostly around `rotate()` modules, which are converted to `multmatrix` in .csg, and more specifically where the input angles were calculated with some reasonably complex math including trig functions.

There was some conjecture in IRC that the rounding mode was differing between platforms (see std::fesetround, std::fegetround).  But I'm pretty sure they should all default to `FE_TONEAREST`?

I did try wrapping just the double-conversion calls with various different `fesetround()` values(and restoring the previous setting afterwards), but this had no effect on test results.  So if it is anything to do with the rounding, it would have to be the calculations that occur before converting to string.

Right now, I'm considering doing some tests with using `long double`s for all trig: [`long double std::sinl(long double)`](https://en.cppreference.com/w/cpp/numeric/math/sin) etc, for increased precision, particularly within the function that converts `rotate` args to `multmatrix`.

thehans:
OK, well my attempts at better precision caused Ubuntu 18.04 and 20.04 to have successful matching results.  Previously 18.04 failed on 2 tests IIRC.

Mac is still failing 9 tests, but it is down from 14 before precision changes.  
I'm not sure what the exact failures are though since test reports aren't uploaded.

Mac failed before
```
	204 - dumptest_intersection_for-tests (Failed)
	211 - dumptest_module-recursion (Failed)
	260 - dumptest-examples_animation (Failed)
	261 - dumptest-examples_assert (Failed)
	271 - dumptest-examples_logo_and_text (Failed)
	291 - dumptest-examples_example014 (Failed)
	296 - dumptest-examples_example020 (Failed)
	300 - dumptest-examples_example024 (Failed)
	301 - dumptest-examples_candleStand (Failed)
	304 - dumptest-examples_module_recursion (Failed)
	305 - dumptest-examples_list_comprehensions (Failed)
	306 - dumptest-examples_polygon_areas (Failed)
	1122 - lazyunion-dump_lazyunion-hull-for (Failed)
	1124 - lazyunion-dump_lazyunion-intersection-for (Failed)
```
Mac failed after:
```
	204 - dumptest_intersection_for-tests (Failed)
	260 - dumptest-examples_animation (Failed)
	291 - dumptest-examples_example014 (Failed)
	296 - dumptest-examples_example020 (Failed)
	301 - dumptest-examples_candleStand (Failed)
	305 - dumptest-examples_list_comprehensions (Failed)
	306 - dumptest-examples_polygon_areas (Failed)
	1122 - lazyunion-dump_lazyunion-hull-for (Failed)
	1124 - lazyunion-dump_lazyunion-intersection-for (Failed)
```

Windows tests are still running as I write this comment, due to a random msys2 package download failure and having to re-run the build.

src/core/scadstream.h
```diff
@@ -0,0 +1,89 @@
+#pragma once
+
+#include <ostream>
+#include <sstream>
+#include <string>
+#include <iterator>
+#include <double-conversion/double-conversion.h>
+
```
kwikius, line 8: IMHO would be preferable to make these defines static members of the scad::ostringstream class 
```c++
class ostringstream
{
public:
   using dc_flags_t = double_conversion::DoubleToStringConverter::Flags;
   static constexpr dc_flags_t dc_flags = dc_flags_t::UNIQUE_ZERO;
   static constexpr std::intmax_t dc_buffer_size = 128;
   static constexpr char dc_inf[] = "inf";
   static constexpr char dc_nan[] = "nan";
   static constexpr char dc_exp = 'e';
   static constexpr int dc_decimal_low_exp = -5;
   static constexpr int dc_decimal_high_exp = 6;
   static constexpr int dc_max_leading_zeros = 5;
   static constexpr int dc_max_trailing_zeros = 0;
  ostringstream() :
      builder(buffer, dc_buffer_size),
      dc(dc_flags, dc_inf, dc_nan, dc_exp,
          dc_decimal_low_exp, dc_decimal_high_exp,
          dc_max_leading_zeros, dc_max_trailing_zeros)
  {}
...
};
```

thehans: Good point, always nice to replace `#define`s.  Those were moved over from Value.cc without much thought to refactoring.

thehans:
I'm done with changes for now, if anyone else would like to review the code.

The test failures on mac (slight differences in calculated values between platforms) were resolved by re-enabling number rounding in `test_cmdline_tool.py`.  
Long ago, before double-conversion was implemented *at all*, numbers in test dumps were rounded to 6 significant digits before performing a diff of expected vs actual test results.  This time the rounding limit has been increased to 13 significant digits, which was the most precision that still passed tests.

I also tried to reduce the number of changes by splitting out some of the tangentially related changes into their own PRs (4395, 4396, 4399; already merged).

thehans:
I'm setting this back to "WIP" status, as I think it would be best to find a way to set CSG cache key precision, **separately** from user `echo()`'s etc.  
That, and/or it should be released alongside a scad user-space option for number formatting.

I plan to change `scad:ostringstream`'s constructor to take a parameter which specifies the format for doubles.  
And maybe also see how much work to switch to [fmt](https://github.com/fmtlib/fmt) library instead of double-conversion, after reading some discussions in #3908

PR #4389, opened 2022-10-20
Warnings: annihilated

thehans:
I got tired of scrolling through pages of warnings and went on a compiler warning fixing spree until I ended up with 0 on my Linux Mint, building with both GCC 11.2 and clang++-14


thehans:
Summary of results after looking over the CI logs:

- Ubuntu 18.04: No Warnings
- Ubuntu 20.04: No Warnings
- Ubuntu 22.04 (distro which Linux Mint 21 is based on; my own computer): No Warnings
- Mac:  No Warnings except during linking `ld: warning: dylib ([VARIOUS LIBRARIES].dylib) was built for newer macOS version (11.0) than being linked (10.13)`
  No idea if that is actually a problem or not.
- Windows:  
    - I found some very concerning warnings about out of bounds array access when building mimalloc, but it turns out its a false positive from gcc on Mingw64, and a [bug exists here](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105523)
    - Looks like still a ton of boost pragma messages which I can't make any sense of.  I try to follow the list of includes but can never find this alleged file which is including a deprecated header `/include/boost/function_output_iterator.hpp`  nor the other one about bind placeholder.  Example snippet from log below.

```
[ 13%] Building CXX object CMakeFiles/OpenSCAD.dir/src/Feature.cc.obj
In file included from D:/a/_temp/msys64/mingw64/include/boost/smart_ptr/detail/sp_thread_sleep.hpp:22,
                 from D:/a/_temp/msys64/mingw64/include/boost/smart_ptr/detail/yield_k.hpp:23,
                 from D:/a/_temp/msys64/mingw64/include/boost/smart_ptr/detail/spinlock_gcc_atomic.hpp:14,
                 from D:/a/_temp/msys64/mingw64/include/boost/smart_ptr/detail/spinlock.hpp:42,
                 from D:/a/_temp/msys64/mingw64/include/boost/smart_ptr/detail/spinlock_pool.hpp:25,
                 from D:/a/_temp/msys64/mingw64/include/boost/smart_ptr/shared_ptr.hpp:29,
                 from D:/a/_temp/msys64/mingw64/include/boost/shared_ptr.hpp:17,
                 from D:/a/_temp/msys64/mingw64/include/boost/format/alt_sstream.hpp:22,
                 from D:/a/_temp/msys64/mingw64/include/boost/format/internals.hpp:24,
                 from D:/a/_temp/msys64/mingw64/include/boost/format.hpp:38,
                 from D:/a/openscad/openscad/src/utils/printutils.h:6,
                 from D:/a/openscad/openscad/src/utils/exceptions.h:6,
                 from D:/a/openscad/openscad/src/Feature.h:9,
                 from D:/a/openscad/openscad/src/core/module.h:6,
                 from D:/a/openscad/openscad/src/core/SourceFile.h:8,
                 from D:/a/openscad/openscad/src/core/customizer/CommentParser.h:3,
                 from D:/a/openscad/openscad/src/openscad.cc:28:
D:/a/_temp/msys64/mingw64/include/boost/function_output_iterator.hpp:14:1: note: '#pragma message: This header is deprecated. Use <boost/iterator/function_output_iterator.hpp> instead.'
   14 | BOOST_HEADER_DEPRECATED("<boost/iterator/function_output_iterator.hpp>")
      | ^~~~~~~~~~~~~~~~~~~~~~~
D:/a/_temp/msys64/mingw64/include/boost/bind.hpp:36:1: note: '#pragma message: The practice of declaring the Bind placeholders (_1, _2, ...) in the global namespace is deprecated. Please use <boost/bind/bind.hpp> + using namespace boost::placeholders, or define BOOST_BIND_GLOBAL_PLACEHOLDERS to retain the current behavior.'
   36 | BOOST_PRAGMA_MESSAGE(
      | ^~~~~~~~~~~~~~~~~~~~
In file included from D:/a/_temp/msys64/mingw64/include/boost/smart_ptr/detail/sp_thread_sleep.hpp:22,
                 from D:/a/_temp/msys64/mingw64/include/boost/smart_ptr/detail/yield_k.hpp:23,
                 from D:/a/_temp/msys64/mingw64/include/boost/smart_ptr/detail/spinlock_gcc_atomic.hpp:14,
                 from D:/a/_temp/msys64/mingw64/include/boost/smart_ptr/detail/spinlock.hpp:42,
                 from D:/a/_temp/msys64/mingw64/include/boost/smart_ptr/detail/spinlock_pool.hpp:25,
                 from D:/a/_temp/msys64/mingw64/include/boost/smart_ptr/shared_ptr.hpp:29,
                 from D:/a/_temp/msys64/mingw64/include/boost/shared_ptr.hpp:17,
                 from D:/a/_temp/msys64/mingw64/include/boost/format/alt_sstream.hpp:22,
                 from D:/a/_temp/msys64/mingw64/include/boost/format/internals.hpp:24,
                 from D:/a/_temp/msys64/mingw64/include/boost/format.hpp:38,
                 from D:/a/openscad/openscad/src/utils/printutils.h:6,
                 from D:/a/openscad/openscad/src/utils/exceptions.h:6,
                 from D:/a/openscad/openscad/src/Feature.h:9,
                 from D:/a/openscad/openscad/src/Feature.cc:9:
D:/a/_temp/msys64/mingw64/include/boost/function_output_iterator.hpp:14:1: note: '#pragma message: This header is deprecated. Use <boost/iterator/function_output_iterator.hpp> instead.'
   14 | BOOST_HEADER_DEPRECATED("<boost/iterator/function_output_iterator.hpp>")
      | ^~~~~~~~~~~~~~~~~~~~~~~
[ 13%] Building CXX object CMakeFiles/OpenSCAD.dir/src/FontCache.cc.obj
```

thehans:
I found that the deprecated header include messages are due to msys builds using a downgraded CGAL 5.2, which had not yet updated their boost includes.

After changing to latest CGAL (5.5.1), I found that a single test was failing on windows: 
`fastcsg-cgalpng_polyhedron-tests`

As far as I know, this test is intended to verify that no crashes occur for operations on malformed polyhedrons.
Since the polyhedrons are malformed, I don't feel that there is really a right or wrong result as long as the crash doesn't occur.

polyhedron_test

In addition to the malformed polyhedron point/faces data, the polyhedrons in this test were each touching at a single point with at least two neighbors which if I'm not mistaken makes them not 2-manifold. I think fastcsg has specifically had past issues with joining polyhedrons that contain coincident points, so I spaced them a tiny bit so they don't touch.  That would at least narrow down what type of malformed input we are testing with this. and luckily that seemed to work.  

I'm not sure if we have any other tests that involve polyhedrons touching at a single point, but I also don't think there would be much value in making another separate test for it.  Particularly because it would have added difficulty tracking differences between CGAL 5.2 and 5.5.1.

thehans:
I also submitted a couple of minor warnings as an issue to mimialloc (the only ones that weren't the gcc bug about array bounds).

src/geometry/Polygon2d.h
```diff
@@ -40,7 +40,7 @@ class Polygon2d : public Geometry
 
   typedef std::vector<Outline2d> Outlines2d;
   const Outlines2d& outlines() const { return theoutlines; }
-
+  using Geometry::transform;
```
rcolyer, line 43: This looks like an incorrect "using" include (to fix that warning), and maybe indicative of a failed is-a relationship in the inheritance if Geometry was expected to take a 3D transform (although it's not even implemented in the base).  To think about.

src/io/DxfData.cc
```diff
@@ -419,7 +419,7 @@ DxfData::DxfData(double fn, double fs, double fa,
         auto lv = grid.data(this->points[lines[idx].idx[j]][0], this->points[lines[idx].idx[j]][1]);
         for (size_t ki = 0; ki < lv.size(); ++ki) {
           int k = lv.at(ki);
-          if (k < 0 || k >= lines.size()) {
+          if (k < 0 || k >= static_cast<int>(lines.size())) {
```
rcolyer, line 422: This inequality can be cast the other undefined-behavior-safer way, k to unsigned int, since k is guarded as not less than 0.

src/io/DxfData.cc
```diff
@@ -449,7 +449,7 @@ next_open_path_j:;
       auto lv = grid.data(ref_point[0], ref_point[1]);
       for (size_t ki = 0; ki < lv.size(); ++ki) {
         int k = lv.at(ki);
-        if (k < 0 || k >= lines.size()) {
+        if (k < 0 || k >= static_cast<int>(lines.size())) {
```
rcolyer, line 452: Also here

src/io/DxfData.cc
```diff
@@ -491,7 +491,7 @@ found_next_line_in_open_path:;
...
```
rcolyer, line 494: And here

src/geometry/Polygon2d.h
thehans: I was trying to think of a nice clean solution, but haven't come up with anything yet.

But also keep in mind, that if OpenSCAD is to officially support 2d-in-3d geometry at some point, then we would probably still want some way to call a 3d transform on 2d object.  
Then again, maybe the class to do that would be something entirely separate from Polygon2d.

thehans: added note in comments

PR #4392, opened 2022-10-22
Add "Save a Copy" menu option

### LINKED ISSUE #4314: Feature Request: Save a Copy menu option [$25]
Add a new File menu option titled "Save a Copy." This option exists in other software such as LibreOffice, and its behavior is described [here](https://askubuntu.com/questions/398969/what-is-the-difference-between-save-as-and-save-a-copy). This would be helpful for when the user wants to save a benchmark copy or alternate version with a different name, but then keep making modifications to the original file without needing to reopen it.

Most users would likely expect to see the option placed below "Save As," which is where it is located in LibreOffice et al.

<bountysource-plugin>

---
There is a **[$25 open bounty](https://app.bountysource.com/issues/110341311-feature-request-save-a-copy-menu-option?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

RayyanAnsari:
Closes #4314


t-paul (approved):
Looks good :+1:. Adds the saved copy to recent file history which I think is good.

PR #4407, opened 2022-11-04
C++17 Update

thehans:
Testing if builds still work when switching from C++14 to C++17.  

I also may have gotten carried away trying out `std::variant` in lieu of `boost::variant`.
I'm still not sure if there's any benefit to doing so, and I haven't seen any measurable difference in performance.
The visitors in `Value.cc` are slightly cleaner by not having to inherit `boost::static_visitor<return_T>`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,8 +41,8 @@ endif()
 
 add_executable(OpenSCAD)
 
-# Strict c++14, e.g. no GNU extensions
-set_property(TARGET OpenSCAD PROPERTY CXX_STANDARD 14)
+# Strict c++17, e.g. no GNU extensions
+set_property(TARGET OpenSCAD PROPERTY CXX_STANDARD 17)
 set_property(TARGET OpenSCAD PROPERTY CXX_EXTENSIONS OFF)
 set_property(TARGET OpenSCAD PROPERTY CXX_STANDARD_REQUIRED ON)
 # Output compilation database (compile_commands.json), so we can e.g. run clang-tidy or other tools separately
--- a/scripts/macosx-build-homebrew.sh
+++ b/scripts/macosx-build-homebrew.sh
@@ -58,7 +58,7 @@ for formula in boost; do
   fi
 done
 
-for formula in pkg-config eigen cgal glew glib opencsg freetype libzip libxml2 fontconfig harfbuzz qt5 qscintilla2 lib3mf double-conversion imagemagick ccache ghostscript; do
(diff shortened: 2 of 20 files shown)


thehans:
I guess my motivation was that the more `std` library things we can use instead of boost, the better.
I also looked into switching to `std::optional`, but realized that it does not support reference types, which are already being used in a few places.

thehans:
It looks like our external json library: `src/ext/json/json.hpp` is causing problems, as it tries to `#include <filesystem>` header when it detects C++17.

Ubuntu 18.04 builds are using GCC 7.5.0, which supports most of C++17, but the filesystem TS is one thing that version is missing.
Also apparently the Mac server on `AppleClang 13.0.0.13000029` has incomplete filesystem support and is failing on `json.hpp` as well.

thehans:
Updating `json.hpp` to latest release resolved the std filesystem issue.
Ubuntu 18.04 is passing.

Mac is now failing to build for a different reason.
Somehow the generated `parser.cxx` file is defining a `NE` macro (for Not Equal '!=') which is clashing with a boost `NE` macro (for "no except").
The build log shows that it is using Bison 2.3, which is ancient (2006).  

My local machine uses Bison 3.7.6, and has 'NE' defined in `parser.hxx` as an *enum*, not `#define`.
So I'm hoping that installing a recent Bison 3.8 via homebrew will resolve it.

Windows builds are also failing tests, but that seems to be an issue with updated runner images #4408, and has been failing for a few days now.

thehans:
Homebrew's bison would not override the old version without further changes:
```
bison is keg-only, which means it was not symlinked into /usr/local,
because macOS already provides this software and installing another version in
parallel can cause all kinds of trouble.

If you need to have bison first in your PATH, run:
  echo 'export PATH="/usr/local/opt/bison/bin:$PATH"' >> /Users/runner/.bash_profile

For compilers to find bison you may need to set:
  export LDFLAGS="-L/usr/local/opt/bison/lib"
```
I don't know why macOS provides an ancient version or what kind of trouble a second homebrew version might cause, so...

As an alternative simpler fix suggested by @jordanbrown0, I have gone back to the default Bison, and just renamed our symbol `NE` to `NEQ`.

t-paul:
I think that Keg thing mostly means we need to include the package into that `brew link` list which already has a couple of those "keg-only" entries:
https://github.com/openscad/openscad/blob/master/scripts/macosx-build-homebrew.sh#L73-L76

But it looks like the rename works, so maybe just going with that is even easier.

PR #4415, opened 2022-11-12
Copy over the Save As's manual overwrite check to Save a Copy

### LINKED ISSUE #4397: Save a Copy silently overwrites existing file if extension is empty
I observed the behavior described below with the new Save a Copy option on the current [master](https://github.com/openscad/openscad/commit/18b0dbd0c41a97791ca2ce17c5f21d8f3d307c13). I believe the fix will be the same as it was for #4235.

        \[...\] in the very specific set of inputs as shown in the screenshot, the user will get no confirmation that they are about to overwrite an existing file.
Screenshot_20220413_180620

_Originally posted by @jbinvnt in https://github.com/openscad/openscad/issues/3706#issuecomment-1098561884_

jbinvnt:
Closes #4397 

I did some manual local tests and it appears that Save a Copy is functioning as expected now. However, it would help if someone else can do their own checks as well before this is merged.

### DIFF
--- a/src/gui/TabManager.cc
+++ b/src/gui/TabManager.cc
@@ -729,6 +729,16 @@ bool TabManager::saveACopy(EditorInterface *edt)
 
   if (QFileInfo(filename).suffix().isEmpty()) {
     filename.append(".scad");
+
+    // Manual overwrite check since Qt doesn't do it, when using the
+    // defaultSuffix property
+    const QFileInfo info(filename);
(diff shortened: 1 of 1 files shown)


bitbasher:
this PR adds a check and a popup outside of what the Qt File Save dialog is already doing .. which is the code i deleted in PR #6258 

@kintel so this PR should probably be abandoned

PR #4425, opened 2022-11-26
Major rearrangement of tests/CMakeListst.txt

thehans:
I found `tests/CMakeListst.txt` difficult to follow, and wanted to give it a more clear structure.
So I did a large refactoring, mostly re-ordering chunks of the file so that similar commands are grouped into major sections, each having its own prominent label.

#### List of sections
 - Check Dependencies
 - Platform Specific Configurations
 - Define Macros and Functions
 - Configure Templated Files and Generated Tests
 - Define test configurations
 - Define Various Test File Lists
 - Add experimental features tests
 - Add tests
 - Disable Tests with Known Issues
 - Extra Debug Info

Due to the way certain data is stored or generated, some steps should only be done before/after others. 
One of the reasons for splitting things up this way, was to make it easier to verify/enforce this ordering.  For example
 - "Define test configurations" must go before "Add Tests"
 - "Add experimental features tests" must go before "Add Tests"
 - "Disable Tests with Known Issues" must go after "Add Tests"
 - Writing out the list of test configurations in "Extra Debug Info" must go after "Add Tests"

#### Change summary from commit message
Major rearrangement of tests/CMakeListst.txt into sections for better understandability.
 - Each section has a named header outlined in # characters
 - Fixed a small scoping bug, only relevant to debug message of test configurations.
 - Simplify logic/conditionals where possible.
 - Use cmake `configure_file` command for CTestTestCustom.template
 - Shorten line lengths where possible
   - create new variables for repeated strings
   - abbreviate some of the longer existing variable names
   - rework path definitions containing `../`
 - File lists in their own section, before adding tests.
   - Don't reference partially built lists to build other lists.
   - Try not to build lists of files that have only one usage.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -35,6 +35,7 @@ jobs:
           mkdir build && cd build
           cmake .. -G"Ninja" -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=Release -DEXPERIMENTAL=ON -DSNAPSHOT=ON
           cmake --build . -j2
+          cmake --install . --prefix=.
       - name: Run Test Suite
         run: |
           cd build
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -14,53 +14,63 @@ cmake_policy(SET CMP0057 NEW)
 project(openscad_tests)
 enable_testing()
 
-# MCAD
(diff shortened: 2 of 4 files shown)


thehans:
These changes started while I was working on #4067, but then decided to split them into their own PR.
I needed to fix/debug some issues with (cross-built, and packaged) tests and was not confident that things were always processed in the correct order, so I set out to make that ordering more clear.  I have also made an effort to simplify / shorten things wherever I saw an opportunity.

Overall test behavior should not change, as no tests were intentionally added, removed, or altered.

Issue #4432, opened 2022-12-04
linear_extrude regression for `slices=0`


thehans:
This was originally brought up in IRC by @UBaer21 

In development snapshots since mid october 2022, linear_extrude can be passed `slices=0` and results in broken geometry. eg:
```
linear_extrude(twist=90, slices=0) square(10);
```

 - Previous behavior:  `slices` less than 1 was considered invalid, and treated the same *as if slices were undefined*. This means slices would be auto calculated, resulting in either `slices=1` or some value involving `$fn`/`$fs`/`$fa` (dependent upon the existence of `twist` or non-uniform `scale` arguments).
 - Current behavior: slices=0 results in invalid geometry with a top and bottom face and no sides.
 - Proposed solution: `slices<1` should get clamped to a *minimum value* of 1, and *not* treated as undefined.  I believe this is more fitting with the principle of least surprise.

I have identified the logic error in code which caused this, and will create a PR shortly.  Mainly opening an issue just to produce an issue number, for labeling a new test case.

thehans:
Fixed by #4433

PR #4433, opened 2022-12-05
Fix for issue #4432

thehans:
Plus some additional type WARNINGS, and reformatting of named argument warnings.


src/core/Parameters.cc
```diff
@@ -261,3 +278,27 @@ void print_argConvert_warning(
...
+  if (expected.size() == 1) {
+    message << Value::typeName(expected[0]);
+  } else {
+    assert(expected.size() > 0);
+    message << "one of (" << Value::typeName(expected[0]);
+    for (size_t i = 1; i < expected.size(); i++) {
+      message << ", " << Value::typeName(expected[i]);
+    }
+    message << ")";
+  }
```
jordanbrown0, line 301: Suggest "expected a, b, or c"

src/core/Parameters.cc
```diff
@@ -261,3 +278,27 @@ void print_argConvert_warning(
...
+void print_argConvert_warning(
+  const std::string& calledName,
+  const std::string& argName,
+  const Value& found,
+  std::vector<Value::Type> expected,
+  const Location& loc,
+  const std::string& documentRoot
+  ) {
+  std::stringstream message;
+  message << calledName << "(..., " << argName << "=" << found.toEchoStringNoThrow() << ") parameter could not be converted: expected ";
```
jordanbrown0, line 291: You could drop the "parameter could not be converted" part. You've given the name in context, so it is clear that it is a parameter you are complaining about, and the "expected x, found y" makes the reason for the displeasure clear.

jordanbrown0 (approved):
Optional wording changes.

src/core/Parameters.cc
thehans: I agree the wording there is odd since there isn't really any "conversion" attempted.
I am guessing this phrase originally comes from situations like `cube(size=1);` where size gets "converted" to a "vec3" `[1,1,1]`.

It feels like its missing something to remove the phrase entirely though.  I'm thinking maybe replace with "Invalid type:"

PR #4439, opened 2022-12-07
Clang tidy

thehans:
Applying many clang-tidy automated fixes for `modernize-*` checks.

There are many more that could still be applied, but I wanted to open a PR to make sure that no other platforms are seeing regressions, before applying much more.

So far the following have been applied:
```
modernize-loop-convert
modernize-use-emplace
modernize-redundant-void-arg
modernize-use-noexcept
modernize-make-unique
modernize-concat-nested-namespaces
modernize-use-equals-delete
modernize-use-bool-literals
modernize-deprecated-headers
modernize-use-transparent-functors
modernize-make-shared
modernize-avoid-bind
modernize-use-using
modernize-use-nullptr
```
With the exception of `modernize-use-using`, the above fixes were applied in order from least to most frequent.

The remaining `modernize*` checks which have not had fixes applied yet, are basically those which occur most frequently.
They are listed below with counts* of occurrences to the left of each.
```
   3986 [modernize-use-trailing-return-type]
    744 [modernize-use-nodiscard]
    177 [modernize-use-auto]
    160 [modernize-use-equals-default]
    148 [modernize-use-override]
    133 [modernize-use-default-member-init]
    121 [modernize-avoid-c-arrays]
    115 [modernize-return-braced-init-list]
    105 [modernize-raw-string-literal]
     78 [modernize-pass-by-value]
```
*counts may contain duplicates and should be considered a rough estimate.

There are also many other groups of checks have not been covered by this PR  (e.g. `cppcoreguidelines-*` `bugprone-*`, `concurrency-`, `clang-analyzer-`, `performance-`, `protability-`, etc.)


thehans:
I ended up applying fixes for almost all checks of type: `modernize-*` and `clang-analyzer-*`.

The few exceptions which remain can be seen in the top level `.clang-tidy` file (checks prefixed with minus `-` are disabled)
```
  modernize-*,
    -modernize-use-trailing-return-type,
    -modernize-avoid-c-arrays,
    -modernize-use-nodiscard,
  clang-analyzer-*,
    -clang-analyzer-cplusplus.NewDelete,
    -clang-analyzer-cplusplus.NewDeleteLeaks
```

 - `use-trailing-return-type` was not touched since it would just be an absurdly large change to the codebase, changing every function signature.
 - `avoid-c-arrays` had more occurences than I cared to deal with.
 - `use-nodiscard` actually has fixes applied to all sources, *except* those under `src/gui`.  The reason for this is that Qt MOC is not able to parse `[[nodiscard]]` on Ubuntu 18.04 and 20.04.
 - The `NewDelete` and `NewDeleteLeaks` checks were mostly pointing to some internal boost and CGAL header code.  I could not comprehend the sequence of messages that went with those, but I'm guessing they are pretty much false positives.

PR #4441, opened 2022-12-07
Rough distance measurement function

Paulwhy2:
This is a rough draft of a distance measurement function.  You can Shift-double-click at 2 points on the model and get an approximate distance in the Log. It works in the preview, and after a render.

How it works: The existing QGLView::mouseDoubleClickEvent(QMouseEvent *event) function provides the screen x and y coordinates of any doubleclick event and then finds the z cordinate with glReadPixels().  These x, y,and z are in screen coordinates which are no good for measuring, but that same event function converts these to model coordinates with gluUnProject().  You can use these model coordinates to measure, but they are only approximate, +/- a pixel width I think.  If you zoom in and click, you get higher accuracy.

The existing DoubleClickEvent function then moves the center of rotation of the model to the doubleclicked location.  I added an if/else at this point: if the shift key is held down while doubleclicking, it calls a new function: void measureDist(GLdouble px, GLdouble py, GLdouble pz), else it continues with the existing code.

The new measureDist() saves the coordinates on the first shift-doubleclick and puts a crosshairs on the location, then if you shift-doubleclick again, it subtracts the old coordinates from the new, does the pythagorean calculation and prints the result to the log.

It still needs a lot of work, but at this point I'd like some advice on if I am on a good track, or if anyone else is working on this or a better solution, or if anyone wants to help with this development.

### DIFF
--- a/src/gui/QGLView.cc
+++ b/src/gui/QGLView.cc
@@ -174,6 +174,49 @@ void QGLView::mousePressEvent(QMouseEvent *event)
   last_mouse = event->globalPos();
 }
 
+bool firstPoint = true;
+
+void measureDist(GLdouble px, GLdouble py, GLdouble pz){
+  static float firstx, firsty, firstz;
+  float deltax, deltay, deltaz, deltadistance;
+  float secondx, secondy, secondz;
+  if(firstPoint){
+    firstx = px;
+    firsty = py;
+    firstz = pz;
+    glBegin(GL_LINES);
+    glVertex3f(firstx-1, firsty, firstz);
+    glVertex3f(firstx+1, firsty, firstz);
+    glVertex3f(firstx, firsty+1, firstz);
+    glVertex3f(firstx, firsty-1, firstz);
+    glVertex3f(firstx, firsty, firstz+1);
+    glVertex3f(firstx, firsty, firstz-1);
+    glEnd();
+    LOG(message_group::None, Location::NONE, "", "first point stored, select second point");
+    firstPoint = false;
+    return;
+  }else{
+    secondx = px;
+    secondy = py;
+    secondz = pz;
+    glBegin(GL_LINES);
+    glVertex3f(secondx-1, secondy, secondz);
+    glVertex3f(secondx+1, secondy, secondz);
+    glVertex3f(secondx, secondy+1, secondz);
+    glVertex3f(secondx, secondy-1, secondz);
+    glVertex3f(secondx, secondy, secondz+1);
+    glVertex3f(secondx, secondy, secondz-1);
+    glVertex3f(firstx, firsty, firstz);
+    glVertex3f(secondx, secondy, secondz);
+    glEnd(); 
+    deltax=px-firstx;
+    deltay=py-firsty;
+    deltaz=pz-firstz;
+    deltadistance=sqrt(pow(deltax,2)+pow(deltay,2)+pow(deltaz,2));
+    LOG(message_group::None, Location::NONE, "", "distance x=: %1$.4f  y=: %2$.4f   z=: %3$.4f  Total= %4$.4f",deltax,deltay,deltaz, deltadistance);
+    firstPoint = true;
+  }
+}
 /*
  * Voodoo warning...
  *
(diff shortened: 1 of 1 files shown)


jordanbrown0:
Better would be something that selected interesting points on the model (vertexes, edges, et cetera), but that's hard.  What you're doing seems like a very reasonable approach.

(Haven't looked at the actual implementation.)

I'm not sure about shift-double-click as the gesture; that's a bit obscure.  I would think perhaps of a menu item with a control-key shortcut that would put you into "measure" mode, and then have you click twice to select points.

Similarly, I'm not sure about the result going into the log.  I would have thought in terms of a popup, or maybe temporarily taking over the Viewport status bar.  *Also* putting it into the log might be good so that you could copy/paste from there.

Paulwhy2:
Hi Jordan, I agree with all your comments.  
1)  It would be better to select vertices.  There must be a database or array of vertices, and the code could loop through these to find the closest and call that the measurement point.  I'm not sure where to find that array yet, but coding that might be straightforward.
2) yes again, shift double click was just an easy way of hijacking the existing code.  This function needs a proper UI.  What you describe would be better.
3)  Yes, again , putting the results in the log was just an easy way of getting this going.  A popup that shows you are in measure mode could have an output section.

jordanbrown0:
You would think that there would be a list of vertices, but you would be wrong.

Remember what I said about preview abusing OpenGL?  Preview doesn't *know* which vertices survive the various boolean operations, or what vertices are created by those operations; all of that is done in OpenGL in 2D.

It knows the raw triangles that go into the boolean operations, but that's it.

Paulwhy2:
No wonder I couldn't find it.  What if we built up an array of vertices as the model is rendered, and then used that array for snapping the measurement point?  

It seems that some (or maybe all?) elements are drawn in Renderer::render_surface() with:
glBegin(GL_TRIANGLES);
...some iteration and calculation code...
glVertex3d(x,y,z);  glVertex3d(x2,y2,z2);  glVertex3d(x3,y3,z3);
glEnd();

Could we 'collect' those vertices at that point in the code?

Paulwhy2:
I think I understand now.  In **preview** I'll only find the vertices of the centered cube (and optionally the hidden vertices of the subtracted cube) but not the vertices formed in that operation.  They only exist in old GL 2d at this point.    

Can you tell me where in the code it does these operation of throwing the negative and positive triangles into openGL?  and where it does the intersection magic?

OK, how about this:  Those points are an intersection of three planes.  We know the planes.  Could we make an array of fabricated vertices to search along with the specified vertices?  That may be a tough algorithm to develop.  I may be grasping at straws here.

Or maybe we can have the user select 3 planes to determine an exact measurement point?

Or maybe the specification is: preview gives you approximate measurements, for exact measurements you'll have to render?

jordanbrown0:
I think the preview rendering is in here:
https://github.com/openscad/openscad/blob/master/src/glview/preview/OpenCSGRenderer.cc

There may be two variations - the default, and the "VBO" variation.

jordanbrown0:
> Or maybe we can have the user select 3 planes to determine an exact measurement point?

You assume that the three planes are not themselves the result of a Boolean operation.

> Or maybe the specification is: preview gives you approximate measurements, for exact measurements you'll have to render?

That seems plausible, that with a preview you can do the pixel-based thing while with a render you can select vertexes.

mikejwaters:
Is this still being worked on? I want to endorse this feature. It would be especially useful for dealing with complex or obtuse scad code that I find online.

Paulwhy2:
Hello Michael,

I have not been working on this for many months.  I got it to the point
where I could do approximate measurements, but I need to learn more about
the openscad source code to continue.  Feel free to download this or
continue work on it.  I may pick it back up later this fall when I have
more time.

Paul

kintel:
I guess this is superseded by #4813

t-paul:
Yes, #4813 does match points and vertices from the mesh (although currently it works only with Manifold enabled in render mode / F6) so it goes one step further. I'd suggest that if there's ideas for measurement improvements, they should be based on the recently merged code.

Issue #4297, opened 2022-07-12
[Feature request] Allow touch-pad friendly 3D navigation


fancsali:
Using a touchpad, and navigating in the 3D view can be somewhat difficult: clicking the right button and moving around at the same time is a bit tricky, as right click is usually implemented as "keep two fingers on the pad and keep it depressed".

Most users, who use a _scripted CAD_ are probably approaching this from a programmers perspective anyway, and do not necessarily use a separate mouse, as they are more likely to be in the keyboard-first camp.
 
Hence, I am proposing using FreeCAD's already established approach for the purpose: holding down `Shift` and moving around with the mouse will **pan**, while keeping `Alt` depressed and moving around would **rotate** the view.

Still, for those amongst us, who do like to use the mosue: this is something that IMO works equally nice with a touchpad and an actual mouse and matches the already established habit with the _other big FOSS CAD package_.

Besides not reinventing the wheel, this also caters for anyone using both - which I expect is a considerable amount of fellow geeks.

fancsali:
Looking at the [code](https://github.com/openscad/openscad/blob/ea8950bd3cd7f06da333c15463492e484098e591/src/gui/QGLView.cc#L240), and the [documentation](https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/The_OpenSCAD_User_Interface#View_navigation); and discussing this in the matrix channel, it turns out there are a few more aspects of this:

* Some (and in particular the `Shift`) modifier key(s) is/are already bound to functionality around manipulating the view.
* On Mac machines `Ctrl`-`Left mouse` is actually an alias of `Right mouse`
* There are many other thoughts and ideas around this topic, e.g. https://www.reddit.com/r/openscad/comments/vnegj2/change_mouse_view_setting/

After some discussion on the [matrix](https://app.element.io/#/room/#openscad:libera.chat) channel we've established, that:
* This could be achieved to various degrees on various OS's with mouse emulation for touch-pads (which might or might not be built into the OS/default driver)
* Most modern OSs support multi-touch, so left/right click two/three finger touch and click can be achieved - but that's not the focus here, it's rather a way of achieving this with modifiers rather multi-touch, if (due to the script-CAD nature) the user's hands are on the keyboard anyway.
[...]

PR #4445, opened 2022-12-10
Allow touch-pad friendly 3D navigation

fleiner:
issue #4297, only changes the behaviour of mouse/touchpad on Liunx systems:

This patch changes <ctrl> left mouse drag  to pan the 3d view, while &lt;ctrl>-&lt;shift> left mouse drag  zooms. Without this patch the view gets rotated in both cases.

This helps in particular with touchpads as Linux does not support right-click and drag with the touchpad in most windows managers.

### DIFF
--- a/src/gui/QGLView.cc
+++ b/src/gui/QGLView.cc
@@ -278,6 +278,8 @@ void QGLView::mouseMoveEvent(QMouseEvent *event)
     if (event->buttons() & button_compare
 #ifdef Q_OS_MAC
         && !(event->modifiers() & Qt::MetaModifier)
+#elif !defined(_WIN32)
+       && !QApplication::keyboardModifiers().testFlag(Qt::ControlModifier)
 #endif
         ) {
       // Left button rotates in xz, Shift-left rotates in xy


stevensmedia:
Hey wow this is exactly what I was going to request: an alternative to right-drag for touch pads. It's also not functional for me on Windows 11.

In fact it won't surprise me if it only works for touch pads on MacOS where Control is the long-standing alias for right-click.

kintel:
How about two-finger-drag? That's usually how I pan under macOS.

codingcatgirl:
Hi, yes, this would be *great*. I'm currently enjoying my time trying to use OpenSCAD on my framework laptop with absolutely no way to translate the viewport. It's not fun.

t-paul:
Which OS is that. For me 2-finger dragging works fine without any special configuration on Debian/Gnome/Dell-XPS13.

codingcatgirl:
I managed to get right click drag to work eventually. Using a modifier would still be *much* more comfortable though.

M1cha:
> Which OS is that. For me 2-finger dragging works fine without any special configuration on Debian/Gnome/Dell-XPS13.

I'm on hyprland and moving two fingers around scrolls instead of dragging (as expected). Thus there's no way to pan the view right now.

evan-goode:
I'm working on a PR to improve the touchpad situation. So far, I've got pinch-to-zoom and pinch-to-rotate working just like Blender: https://github.com/evan-goode/openscad/tree/evan-goode/gestures. I've only tested on Wayland so far; I'm fresh out of Windows and macOS machines with touchpads.

My plan for panning is to extend the recently-added "Mouse" preference pane to allow changing what is mapped to scrolling and shift+scrolling. For example, to get Blender-style touchpad navigation, you would map "Scroll" to "Rotate in altitude/azimuth" and "Shift + Scroll" to "Pan left/right & up/down".

evan-goode:
https://github.com/openscad/openscad/pull/6681

PR #4446, opened 2022-12-10
Added linear_extrude(1) to render 3D shape instead of 2D shape

hackwin:
Updated examples to use linear_extrude so that F6 renders a 3D shape instead of a 2D shape


jordanbrown0:
None of these **need** 3D objects to accomplish their demonstration purposes, and the polygon one, since it is explicitly about polygons, should definitely stay 2D.

PR #4456, opened 2022-12-22
Adapt Version numbers in uni-build-dependancies

gsohler:
Existing version numbers did not compile or were not available for download anymore.
With  the proposed changes, the dependencies and OPENSCAD compiles without any warning/issue in my FC37
(glib 2.75 uses meson & ninja  now to compile), this could be added the the documentation as base requirement)


t-paul (approved):
Thanks! It would be nice to keep code changes separate (matching the PR title).

gsohler:
Hi Torsten,

Yes, this was my intent. But then I found out that  I cannot create the
pull request for a specific commit, but only the whole branch.
Will do more branches in the future.

PR #4465, opened 2022-12-26
New function evalstring

gsohler:
evalstring("variablename") => content of the variable.

Such a function is missing in openscad and it appears that many people have developed tricky workarounds  to overcome this "lack" 

This function looks up the  variable name in the current context and  extracts its value. it returns undef the the variable is not found.

many applications are possible with this feature.
E.g.  together with the context variable scope an associative array can be established
e.g.

refoff=evalstring(str("$",ref));

is  equivalent to an associative index in other languages like
$var=$array[$string];  ( python here)

Hope that this makes OpenSCAD even better


t-paul:
What are those other use cases? It seems better actually having associative arrays like in #4478 instead of a difficult to use feature poking into the internal data structures.

gsohler:
Hi Torsten,

Thank you for coming back to me.

I have seen that test compilations with mac and with windows went totally
wrong, but of course I did not yet have time to look into logs
(just wondering, is the windows  release created with MSYS2  ? )

In order to provide the evalstring function, I just grabbed some pieces of
existing code and put it together in a new way. I can;t  really claim, that
I changed internal data structures. At most I made them accessible to the
user. haha

And Yes, I did look into #4478.  The ideas of Jordan Brown are
extraordinary.  A lot of new ideas and concepts to implement, cleaning up,
simplifying, refactoring
 Mankind must assume he is one of the main programmers these days. I am
confident that he knows what he is doing, whereas I am really new to the
openscad source code and propose just one encapsulated additional function.

And yes, about use cases:

I wanted to redo something like relativity,scad.
Relativity needs associative arrays to access the corners/faces/edges of an
object. But it does in a very clever way: The index is not a string, but
it's a unity vector  instead.
And here comes the clever  trick: when using an unity vector as a key to
the associative array, "relativity"  does not need to dereference the key
but just needs to multiply it with the object dimensions to get the "point
of interest".

But for the sake of completeness: I have attached my own "relativity.scad"
to the mail. it depends on the new evalstring function right in the
beginning.
Each object (like box) defines arbitrary positions in the object.  When
cascading modules with the "child" concept, parameters "ref" and "dest" are
used to concatenate
the objects at certain points and orientations.

Relativity.scad is able to circumvent associative arrays by using unity
vectors as indexes. and it does succeed. I am sure, that many other use
cases will become possible
when having the option to actually use associative arrays

long story short:

evalstring(str("$",string_variable))
is equivalent to

(openscad $ global namespace) [ string_variable ]
this is what  python can do natively

In the meantime I  decided to use actual python instead(see attached
picture). It's not any of the available python-openscad wrapper but uses
an embedded python interpreter compiled in.

Thank you for your attention.

t-paul:
Don't worry about the failures yet, those are very likely caused by updates that happened in those environments lately. Hopefully the failures will be fixed soon.

MSYS2 is only used in Github for testing, the releases are built on CircleCI via [MXE](https://mxe.cc). For reference, some maybe not fully up-to-date pointers at https://github.com/openscad/openscad/wiki/CI-Setup

t-paul:
As for the python stuff, I toyed with something slightly different a while ago. This was more some sort of plugin that could be called :smiley:.

Anyway, I think this could be an interesting option for some people but it would be sad if it's just goes completely separate ways. If you want to keep this going, I'd suggest to discuss a way of making it an official option using all the benefits of an existing project. Easiest option would be in our IRC channel.

t-paul:
For IRC just go to https://openscad.org/ and click the Chat bubble icon (4th one at the top of the left side bar).

src/gui/ScadLexer.cc
```diff
@@ -125,7 +125,7 @@ void Lex::default_rules()
 
   std::string functions("abs sign rands min max sin cos asin acos tan atan atan2 round "
                         "ceil floor pow sqrt exp len log ln str chr ord concat lookup "
-                        "search version version_num norm cross parent_module dxf_dim "
+                        "search version version_num norm cross parent_module parent_param dxf_dim "
```
MichaelPFrey, line 128: this looks odd

src/core/builtin_functions.cc
```diff
@@ -721,6 +721,25 @@ Value builtin_parent_module(Arguments arguments, const Location& loc)
   return {UserModule::stack_element(s - 1 - n)};
 }
 
+Value builtin_evalstring(const std::shared_ptr<const Context>& context, const FunctionCall * func_call)
+{
+  Arguments arguments(func_call->arguments, context);
+  Location loc = Location::NONE;
```
MichaelPFrey, line 727: `Location::NONE` looks odd to me.
Other build in functions have a  `const Location& loc` parameter.

LOC is used in check_arguments, so lets say I  pass the number 3.14 to the function, then check arguments will create a warning, but not say the location?

gsohler: Hi MichaelPFrey,

Thank you for code  review.
Yes i am am aware, that evalstring  does not use a valid Location for error debugging.
Issue is that the available function  prototypes which are valid arguents to new BuiltinFunction are very sparse.
The Function prototype needs access to the Context.
The Function prototype needs to  be able to return a Value
Optionally  the Function Prototype has access to the correct Location.

Unfortunately i did not find such a prototype, this is why I traded off the Location use use Location::None instead.

Actually this is same as 'is_undef' function,which does  not use Location either.

If you show me which function prototype  has all the stuff i need, I eager to  use  a good "Location" parameter.

PR #4469, opened 2022-12-26
Adjust 3mf regression test

kintel:
Adjust regressing testing of 3mf files to account for variations in xmlns schemas to account for differences between V1.x and V2.x.


kintel:
I'll yield to the fix in #4067

PR #4473, opened 2022-12-27
Fixes for clang-tidy `performance-*` checks

thehans:


thehans:
Summary of changes:
```
    138 [performance-unnecessary-value-param]
     15 [performance-for-range-copy]
      9 [performance-unnecessary-copy-initialization]
      8 [performance-no-automatic-move]
      6 [performance-move-const-arg]
      5 [performance-no-int-to-ptr]
      5 [performance-noexcept-move-constructor]
      3 [performance-inefficient-string-concatenation]
      2 [performance-faster-string-find]
      1 [performance-inefficient-vector-operation]
      1 [performance-type-promotion-in-math-fn]
      1 [performance-move-constructor-init]
```
All fixed now (or bypassed with `NOLINT` comments in a few cases).

Issue #4479, opened 2022-12-29
Build failure with mpfr-4.1.1 and cgal-5.5


dwightkelly:
OpenSCAD fails to build against mpfr-4.1.1 and cgal-5.5.
This is most likely an upstream bug in MPFR:
* https://github.com/CGAL/cgal/issues/7064
* https://gitlab.inria.fr/mpfr/mpfr/-/issues/1
* https://www.mpfr.org/mpfr-4.1.1/#bugs

Looks like it will be fixed in mpfr-4.2

Was:

Apple clang version 14.0.0 (clang-1400.0.29.202)
Target: x86_64-apple-darwin21.6.0
OS: macOS 12.6.1 Intel
openscad @2015.03-3_9 (installed using macport)

```
:info:build In file included from src/LibraryInfo.cc:15:
:info:build In file included from src/cgal.h:24:
:info:build In file included from /opt/local/include/CGAL/Gmpq.h:26:
:info:build In file included from /opt/local/include/CGAL/Gmpz.h:34:
:info:build In file included from /opt/local/include/CGAL/Gmp_coercion_traits.h:33:
:info:build /opt/local/include/CGAL/GMP/Gmpfr_type.h:208:25: error: cannot initialize a variable of type 'mpfr_ptr' (aka '__mpfr_struct *') with an lvalue of type 'mpfr_srcptr' (aka 'const __mpfr_struct *')
:info:build                         mpfr_custom_get_kind(f),
:info:build                         ^~~~~~~~~~~~~~~~~~~~~~~~
:info:build /opt/local/include/mpfr.h:1030:14: note: expanded from macro 'mpfr_custom_get_kind'
:info:build     mpfr_ptr _x = (x);                                                  \
:info:build              ^
:info:build /opt/local/include/mpfr.h:1010:9: note: expanded from macro 'mpfr_custom_init_set'
:info:build   _k = (k);                                                    \
:info:build         ^
```

kintel:
"openscad@2015.03-3_9" - this is an ancient version of OpenSCAD. If you really want to get this working on MacPorts, I would suggest either upgrading the MacPorts build or work with the MacPorts maintainer (if they're still active), or report a bug against MacPorts: https://ports.macports.org/port/openscad/

For an easier path, we do offer pre-compiled binaries at https://openscad.org/downloads.html

dwightkelly:
I would recommend deleting the macport listing on your website.

[...]

Issue #4470, opened 2022-12-26
Test case fastcsg-cgalpng_highlight-modifier2 hangs forever on macOS


kintel:
When building locally on M1 on macOS 12.6.1, Xcode 13.4.1, `fastcsg-cgalpng_highlight-modifier2` hangs forever:
```
ctest -R fastcsg-cgalpng_highlight-modifier2
```

It hangs both when building Homebrew dependencies and when building dependencies locally from source.

The same test case passes in ~4 seconds on the CI, so it's not expected to be a heavy test.

kintel:
Update: This seems to be an issue with CGAL: Downgrading to CGAL-5.3 makes it work again, but it fails with the latest (CGAL-5.5.1).

kintel:
Update: Also fails on x86_64 on macOS 12.6.2, Xcode 14.0.1, CGAL 5.5.1:
https://github.com/openscad/openscad/actions/runs/3880238492/jobs/6618132903

kintel:
This just failed on macOS 11 as well: https://github.com/openscad/openscad/actions/runs/3909140299/jobs/6679925007

macOS 11.7.2, GitHub runner image 20221215.1, Xcode 13.2.1, CGAL-5.5.1

kintel:
Closing as fast-csg has been deprecated

PR #4480, opened 2022-12-29
Upgraded dependencies for macOS

kintel:
A few notes:
* mpfr upgraded only to 4.1.0: #4479
* CGAL upgraded only to 5.5: #4470
* glib2 kept at 2.71.0: Fix may be available at https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3177
* lib3mf kept at 1.8.1: Needs #4067 and we may want to stay conservative for a while longer, until lib3mf 2.x can be release on other platforms as well


kintel:
This looks fine now. I'll run some manual tests on old versions of macOS/Intel before merging.

t-paul:
@kintel can we merge this, the master branch build is currently failing when building Qt, I assume this should fix things?

PR #4488, opened 2023-01-03
Correct use of "-nightly" icons suffix for snapshots builds

kintel:
There was some mixup in how the CMake `-DSUFFIX` and `-DSNAPSHOT` parameters were used to determine executable name, install location and icon name.

This change makes all builds with `-DSNAPSHOT` use the "-nightly" version of icons, but otherwise doesn't interfere with how `-DSUFFIX` is managed.

Tested on macOS, could use some testing on other platforms.


thehans (approved):
LGTM.  
It see now what you meant about there being two orthogonal concepts.
The remaining usages of `SUFFIX_WITH_DASH` are for basically naming the executable and directory names, etc.
While `SNAPSHOT_SUFFIX` is specifically for selecting "nightly" alternate icons and images.

PR #4490, opened 2023-01-10
Downgrade macOS runner to macos-11, to silence #4470

kintel:
This really just buys us some time, so we can investigate the issue offline.


kintel:
Windows fail and CircleCI fails are unrelated as this PR is only meant to fix the macOS GitHub build

PR #4492, opened 2023-01-10
Depend on icu-devel package.

t-paul:


t-paul:
Not needed, fixed by https://github.com/msys2/MINGW-packages/issues/15021

PR #4498, opened 2023-01-13
OpenSCAD with a Python Engine

gsohler:
I have added python interpreter next to the yacc interpreter in openscad,  which is used
when the source file extension is ".py" and Python-Engine checkbox is ticked in the Features
(it is a security risk to execute unknown python files)
Almost *all* functions are ported to python already. apparently language constructs and math functions are available in python already.
Many new python embedded functions have OO extensions. Main advantage of python is that it resolves many shortcomings of openscad language
(variable assigning) . or you can create the bits of a qr code on the fly ...

e.g.

def thing():
   return square([3,5]).linear_extrude(height=3)

output(translate(thing(),[1,2,3])
(think its correct syntax)

Child keyword is not handled, but I believe its not needed and/or can be done with python function pointers.

I know it will be a big task to get it compiled everywhere and add the documentation and all the test scripts, but i believe its worth the effort.

lets work forth and back on python-support until its good enough for the world,


jordanbrown0:
This is a very interesting project (give or take the security concerns that we discussed).

In terms of source file layout, I would *not* add the Python support to the files that contain the OpenSCAD-language support.

For a typical module, like "color", I would have three files:
* One that processes an OpenSCAD built-in module call and yields a CSG node.
* One that processes a Python function/method call and yields a CSG node.
* One that implements the CSG node.

Ideally, the three would be in three separate directories, so that the support for each language, and the common support for CSG, are all cleanly separated.

gsohler:
Hi Jordan, i understand your Idea.
Separating the parts would almost multiply the total Files by 3 and Most
Files would only contain one function only. Maybe compilation time is Not
an issue for you But my Personal Focus would rather be to Compile IT on all
platforms and write Regression Tests.


jordanbrown0 ***@***.***> schrieb am Fr., 13. Jän. 2023,
19:01:

kwikius:
Sounds like a great project.
Is it possible to use python source files and openscad sources in one openscad project?
Might be best to wrap your Python code in a #define, so that it can be included or ignored based on a makefile option.

src/core/builtin_functions.cc
```diff
@@ -721,6 +721,25 @@ Value builtin_parent_module(Arguments arguments, const Location& loc)
   return {UserModule::stack_element(s - 1 - n)};
 }
 
+Value builtin_evalstring(const std::shared_ptr<const Context>& context, const FunctionCall * func_call)
```
t-paul, line 724: Please keep that separate, it's not clear if that's going to be merged, so we don't want to combine it with the Python changes.

src/io/import_svg.cc
```diff
@@ -27,7 +27,7 @@
 #include <Eigen/Core>
 #include <Eigen/Geometry>
 
-#include "import.h"
+#include "ioimport.h"
```
kintel, line 30: Are these renames needed, or could you revert that to make this PR cleaner?

gsohler:
Hi
Yes i Had to rename two Header files in openscad as they Had conflicts with
Header files coming from Python . Renaming Them in OpenSCAD was the easier
option

Marius Kintel ***@***.***> schrieb am Sa., 14. Jän. 2023,
21:40:

src/io/import_svg.cc
kintel: Right, python header file conflict..
I'm wondering whether we should just nudge this towards a cleaner solution, by e.g. removing "src/io" from OpenSCAD's `-I` compiler arg and instead `#include "io/import.h". Otherwise we're ending up playing some sort of cat and mouse game with external headers.

kintel: This belong in a separate PR though, if we decide to do it, so there's nothing actionable here.

t-paul: Yep, I was wondering about this too. Sounds like it's good to go for now, but something to consider simetime later for some better header file structure ideally with a single top-level `-I`.

src/core/pyfunctions.cc
```diff
@@ -0,0 +1,1510 @@
...
+#include "iofileutils.h"
+#include "handle_dep.h"
+
+//using namespace boost::assign; // bring 'operator+=()' into scope
+
+
+// Colors extracted from https://drafts.csswg.org/css-color/ on 2015-08-02
+// CSS Color Module Level 4 - Editor’s Draft, 29 May 2015
+extern  std::unordered_map<std::string, Color4f> webcolors;
+extern boost::optional<Color4f> parse_hex_color(const std::string& hex);
```
kintel, line 60: I feel that parse_hex_color() could benefit from being made a utility function declared in a header file.
Now that webcolors is used outside of a single compilation unit, writing a lookup function instead of accessing the data structure directly would also be nice.

kintel: PS! This could be done in a separate, "green refactoring" CL.
(I'm a huge fan of splitting out general cleanup into separate CL, so that reviewers can focus on actual functional changes)

jordanbrown0: Better, read the color table from a text file.

t-paul: The list is coming from a fixed spec that's rarely updated. I don't see a huge benefit from putting it into a separate file.

src/core/pyopenscad.cc
```diff
@@ -0,0 +1,261 @@
...
+    {"rotate_extrude", (PyCFunction) python_rotate_extrude_oo, METH_VARARGS | METH_KEYWORDS, "Rotate_extrude Object."},
+    {"offset", (PyCFunction) python_offset_oo, METH_VARARGS | METH_KEYWORDS, "Offset Object."},
+    {"roof", (PyCFunction) python_roof_oo, METH_VARARGS | METH_KEYWORDS, "Roof Object."},
+    {"output", (PyCFunction) python_output_oo, METH_VARARGS | METH_KEYWORDS, "Output the result."},
+    {"color", (PyCFunction) python_color_oo, METH_VARARGS | METH_KEYWORDS, "Output the result."},
+    {NULL, NULL, 0, NULL}
+};
+
+
+PyTypeObject PyOpenSCADType = {
```
kintel, line 145: You may be able to benefit from the fact that C++ zero-initializes structs, and initialize only the non-zero members, smth. like this:

```

PyTypeObject PyOpenSCADType = {
  .ob_base = PyVarObject_HEAD_INIT(NULL, 0),
  .tp_name = "PyOpenSCAD",

  ...etc for all non-zero field members
};
```

src/gui/MainWindow.h
```diff
@@ -55,6 +55,7 @@ class MainWindow : public QMainWindow, public Ui::MainWindow, public InputEventH
   SourceFile *parsed_file; // Last parse for include list
   std::shared_ptr<AbstractNode> absolute_root_node; // Result of tree evaluation
   std::shared_ptr<AbstractNode> root_node; // Root if the root modifier (!) is used
+  static int python_active;
```
kintel, line 58: Why static? You're checking this for each MainWindow instance, so it should probably be a regular instance member.

src/gui/MainWindow.cc
```diff
@@ -1790,14 +1796,30 @@ void MainWindow::parseTopLevelDocument()
...
+	  int len=strlen(fname);
+	  if(len >= 3 && ! strcmp(fname+len-3,".py")) {
+		  if(Feature::ExperimentalPythonEngine.is_enabled()) this->python_active = 1;
+		  else  LOG(message_group::Warning, Location::NONE, "","Python is not enabled");
+	  }
+  }
+
   this->parsed_file = nullptr; // because the parse() call can throw and we don't want a stale pointer!
   this->root_file = nullptr;  // ditto
+  if(this->python_active) {
```
kintel, line 1815: Eventually, doing this also for cmd-line mode would be cool

src/gui/MainWindow.h
t-paul: I assume it's there to prevent any concurrent access, across multiple MainWindows. I have not checked but it might not be needed as we already have that annoying global lock via `GuiLocker::gui_locked`.

CMakeLists.txt
```diff
@@ -93,6 +93,10 @@ if(("${CMAKE_CXX_COMPILER_FRONTEND_VARIANT}" STREQUAL "GNU") OR
   target_link_options(OpenSCAD BEFORE PRIVATE "-v")
 endif()
 
+  find_package (Python REQUIRED COMPONENTS Interpreter Development)
+  target_include_directories(OpenSCAD PRIVATE ${Python_INCLUDE_DIRS})
+  target_link_directories(OpenSCAD PRIVATE ${Python_LIBRARY_DIRS})
+  target_link_libraries(OpenSCAD PRIVATE ${Python_LIBRARIES})
```
kintel, line 99: Two suggestions:
1. Wrap find_package() in a cmake check (e.g. -DENABLE_PYTHON) to enable turning on/off Python support on the cmake cmd-line, and make Python REQUIRED in that case.
2. Set a preprocessor macro (using target_compile_definitions()) if ENABLE_PYTHON was set

src/io/import_svg.cc
kwikius: Perhaps Better would be #include <openscad/xxx/yyy.h> which gives openscad its own namespace

jordanbrown0:
I built this in my Windows MSYS2 environment.  It built cleanly, but failed on execution with
```
C:/msys64/home/Jordan/openscad/python-support/b/openscad.exe: error while loading shared libraries: libzip.dll: cannot open shared object file: No such file or directory
```
... which is at least something of a lie, since libzip.dll is in /mingw64/bin, and that's in $PATH.

I got it to work with
```
$ PATH=/mingw64/bin:/c/Users/Jordan/AppData/Local/Programs/Python/Python310 b/openscad
```

It seems wrong that CMake found the Python that was installed over in Windows-land, rather than the one installed in MSYS2-land.  I don't see anything in my environment variables that would lead it there.  (But I haven't looked at any of the CMake configuration to try to figure it out.)

In a few minutes of playing with it, I found two things that surprised me.

First, I couldn't figure out how to do a union.  It kept giving me a type error, no matter what I did.  I finally had to look at the sources to see that it expects a list of objects like `union([obj1,obj2,...])`; I kept trying to give it `union(obj1,obj2,...)`.  I also tried `obj1.union(obj2)`, but that failed with an obvious "no such method" error.

Second, I was surprised that if you call output() twice, the results from the first call are discarded.

Minor:  print() output went to the process standard out, rather than to the OpenSCAD console.

This is very promising work.

gsohler:
Jordan, Output is Used to Register one final Mode for displaying. Bit of
course you can also Use an Array in the latest Version .Last Output wins

jordanbrown0 ***@***.***> schrieb am Mo., 16. Jän. 2023,
05:08:

openscad_cpack.cmake
```diff
@@ -0,0 +1,15 @@
+# CPack-time project CPack configuration file.
+# This file is included at cpack time, once per generator
+# after CPack has set CPACK_GENERATOR to the actual generator being used.
+# It allows per-generator setting of CPACK_* variables at cpack time.
+
+if(CPACK_GENERATOR MATCHES "NSIS")
+  set(CPACK_INSTALL_CMAKE_PROJECTS "/home/gsohler/git/openscad/build;OpenSCAD;Unspecified;/")
```
t-paul, line 7: We need to either make this file generic or exclude from being checked in.

src/gui/MainWindow.cc
```diff
@@ -1796,6 +1809,28 @@ void MainWindow::parseTopLevelDocument()
...
+	  }
+  }
+
+  if(this->python_active) {
+    auto fulltext_py =
+    std::string(this->last_compiled_doc.toUtf8().constData());
+
+    char *error  = evaluatePython(fulltext_py.c_str());
+    if(error != NULL) LOG(message_group::Error, Location::NONE, "", error);
+    fulltext ="cube([10,10,10]);\n";
```
t-paul, line 1831: What's this used for?

src/gui/ScadLexer.cc
```diff
@@ -19,7 +19,7 @@ ScadLexer::ScadLexer(QObject *parent) : QsciLexerCPP(parent)
     "abs sign rands min max sin cos asin acos tan atan atan2 "
     "round ceil floor pow sqrt exp len log ln str chr ord concat "
     "is_undef is_list is_num is_bool is_string is_function "
-    "lookup search version version_num norm cross parent_module "
+    "lookup search version version_num norm cross parent_module evalstring"
```
t-paul, line 22: Not really critical listing `evalstring` here, but should be cleaned up.

src/gui/ScadLexer.cc
```diff
@@ -125,7 +125,7 @@ void Lex::default_rules()
 
   std::string functions("abs sign rands min max sin cos asin acos tan atan atan2 round "
                         "ceil floor pow sqrt exp len log ln str chr ord concat lookup "
-                        "search version version_num norm cross parent_module dxf_dim "
+                        "search version version_num norm cross parent_module parent_param dxf_dim "
```
t-paul, line 128: Same with `parent_param` here, but should be cleaned up.

t-paul (changes requested):
With some minor cleanups listed above, I think we can merge.

Where are the test case result changes coming from? If those are from the PR itself, we need to remove them and keep the originals. The PR changes should have no impact on any of the existing test results.

t-paul (approved):
Going to squash-merge to `python-support` branch for some final checks, but in general it's a nice first step on the integration. :tada:

jordanbrown0:
Did a little more research on locating the Python library.  If I uninstall Python for Windows, the MSYS2 build does successfully find the MSYS2 copy of Python, and gets past that problem.

It still needs $PYTHONHOME set, e.g.
```
$ PYTHONHOME=c:/msys64/mingw64 b/openscad --enable-python
```

Issue #4462, opened 2022-12-26
Test false negative due to lax image compare: csgpngtest_issue1165


kintel:
This test passes, but it should have failed:
`ctest -R csgpngtest_issue1165`

Actual:
![issue1165-actual](https://user-images.githubusercontent.com/186094/209509008-cd74b609-f44a-46eb-85ad-8ad373eabcef.png)

Expected:
![issue1165-expected](https://user-images.githubusercontent.com/186094/209509019-b0542b17-824a-4f8c-b7f1-7f43461840fd.png)

It looks like the ImageMagick comparison is a little too forgiving:
`convert issue1165-expected.png issue1165-actual.png -alpha On -compose difference -composite -threshold 10% -morphology Erode Square:1 -format "%[fx:w*h*mean]" info:`

Dropping the threshold down to 8% makes it fail again. Need to investigate if that's a reasonable setting and what impact it has on other tests.

kintel:
In practice, this should be added to `SCADFILES_WITH_DIFFERENT_FAST_CSG_EXPECTATIONS`, but making it fail first would be a good start.

kintel:
Note: File in question was changed in #4067, so the test case will always pass now

PR #4506, opened 2023-01-18
Switch to new image_compare.py for testing

rcolyer:
Following up on #4462 and broader concerns, I wrote a new image comparison method with Python.  Upon completing it and running it, I found a fairly large number of tests that should have failed but the previous comparison method was not detecting the issue, as we had been using a number of pixels over a threshold of difference with ImageMagick.  Because of issues with small edge effects, the threshold was set very large, but this left the opportunity for some fairly significant differences to slip by.

This new test does not compare pixels, but checks all 3x3 pixel blocks (with overlap) for like-signed difference between actual and expected, and fails the test if any 3x3 block has this property.  None of the failing tests in my run were improperly failing, and all of the new fails were tests that should have been updated with prior changes.

Fortunately, no hidden errors in OpenSCAD were detected in this update to the testing, but we simply had some substantial alterations in output (on purpose) that we did not have adequate test awareness of (which undermines the utility of testing).  Hopefully this will give us more rigor in the future.

Of importance is that this update introduces generation of a venv to the cmake run, to setup numpy and PIL for the image_compare.py file.  Provided it actually ends up eventually passing correctly on all CI platforms, I think this is likely a net benefit to the codebase.  This opens up future option of a rich set of Python libraries for more elaborate statistical tests, or flexible customized image testing if a particular future test happens to not work well with the standard image comparison.

This update also removes the ImageMagick dependency requirement for running tests.  However the previous behavior of ImageMagick or diffpng remains available as an option with:  cmake -DUSE_IMAGE_COMPARE_PY=OFF


rcolyer:
Note that the remaining failing test, that happens on only the macos automated testing, is catching a real defect in the produced test result for the issue3158 test.  (Which rather than a fail for the PR is the point of it.)

tests/image_compare.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env python3
+
+from PIL import Image
+import numpy as np
+import sys
+
+
+def Compare3x3(img1, img2):
```
kintel, line 8: It wouldn't hurt with a short explanation what this actually does, as this code is pretty hard to read.

kintel (approved):

tests/image_compare.py
rcolyer: Right.  Done.

kintel: Awesome!

PR #4508, opened 2023-01-22
OpenSCAD with a Python Engine (#4498)

t-paul:
Consolidated changes from #4498 into `python-support` branch.

Initial support for integrated python scripting.

Status:
* Most modules supported
* Build time switch
* Command line switch to enable python engine
* No cross language support yet

Bringing over some of the comments made in the original PR, verify if any of that should be handled before the initial merge to `master`:
- [ ] Check access to color list, maybe introduce a simple internal API instead of direct data structure access?
- [ ] Check static variable python_active
- [ ] Check if command line version can be enabled
- [ ] Check parse() handling with dummy code


t-paul:
Rebased onto `master` to get the unrelated test case changes out of the PR.

jordanbrown0:
In IRC it was said:
> [20:42](https://libera.irclog.whitequark.org/openscad/2023-01-22#33702515) <guso78> so openscad included a whole python interpreter inside its executable

but at least in my MSYS2 build of this project that's not true - it's using an externally-supplied python310.dll.  One of the problems that I encountered was that the build was finding a Python DLL, but it wasn't being found at runtime.

t-paul:
Superseded by #4588.

PR #4510, opened 2023-01-23
Enable autocomplete for words in the document.

maximkulkin:
Currently builtin editor does autocompletion for
builtin APIs only. This patch enables autocompletion for all words in the document, thus allowing to
autocomplete variable names, which greatly increases editing comfort and speed.


t-paul:
See #6846 which tries to be a bit more specific than just enabling all words.

PR #4515, opened 2023-01-31
Yet another Path-Extrude

gsohler:
Even though, this functionality is already available in OpenSCAD, 
I feel there should be a reworked version for several  reasons:

* As a native comamnd , path_extrude is more  available as in an external library
* Beeing there  its also faster and more flexible.It can easily do more complicated 
calculations and checks
* An important path_extrude feature to me is that the cross profile is constant along the
  path and does not get narrower near the corners(even without segments/slices), I did not
  want to break existing code by changing its behaviour. In order to achieve a constant cross-section,
  not only scaling the the profile is neccessary, also the x and y axis of the profile is
  not even normal to each other most of the time.
* code for  profile morphing is already in place, in case python gets a chance. For a path with 3 points
 the python profile function would get a real number from 0.0 to 2.0

A) Most simple application

https://imgpile.com/i/dzei0R

This shows a simple application of path_extrude. You can  concatenate any coordinates  in 3d space
into your points list. The result is arbritary angles between the segments and the profiles match
up exactly. Imagine that you have a steel bar with a square cross section and you bend it,
so the result will exaclty look like this.

B) Corner roundings

https://imgpile.com/i/dze3Wu

For each corner point you can specify a 4th coordinate, which acts as the bending radius. If the
parameter is omitted,the radius=0. The Steel bar  looks even more real now.
For the rounded corners, $fn, $fa, $fs variables are considered and the number of 
segments chosen accordingly.

C) closed and xdir property

https://imgpile.com/i/dzee9G

Of course, there is also an easy way to close the path to form a ring.
xdir propery gives a hint to the extrude, where the x axis for the first segment. This is 
not equal to rotating the extracted profile itself, because xdir guides the x axis when the path 
walks arbritary directions in space and neccessarily vertical.

D) twist and scale

https://imgpile.com/i/dzjsRw

are obligatory. The Twist is evenly spread among the segments. Once twist is used, the pathes 
are segmented with guide of $fs variable.

E) Displayed path is always manifold.

Due to their nature, path_extrude, is very prone to self intersection.
A general thumb of rule is:
if the width_of_profile * sin(bend_angle) > segment high
then it is very likely, that  an intersecion is there  and you can resolve it by making sure 
the equation is not  true anymore.
Fortunately, self intersection can very easily be detected by using the 
point-to-plane distance formula.
As soon as intersection is detected, the path_extrude result is not dispayed in order to prevent furhter
processing of the wrong result . You can always override this by using 'allow_intersect=true'  parameter to debug.
Most of the time, the result will still look fine, just be aware that there is somewhere a hidden 
intersection.

Last, but not least, here is my idea/implementation. Any criticism good/bad is welcome here


froqstar:
I absolutely love this, I think the more full-featured the OpenSCAD standard library the more accessible it gets. 
I used OpenSCAD for years and only recently discovered the library ecosystem around it. Managing libraries is still a pain though.

On topic: I don't understand the xdir parameter for closed shapes yet - doesn't the X axis always point in the same direction, and if not, why does it need to point 45° in the X/Y plane for the given example?

gsohler:
Hi  martin,  I appreciate your feedback and i also found your internet/social media presence ...

xdir paramer is the direction in 3d space where the x axis of the 2D profile shall be directed for the first path segment.
Its [1,0,0] by default, but it can be overridden.  In my example there is no need to point 45 deg into the XY plane.  I just
chose it to show the abilities. you  can choose any other vector or skip this parameter (just 0/0/1 does not make sense, but the cc code captures this error for you)

PR #4516, opened 2023-02-02
offset for 3D  objects

gsohler:
Some time back I realized, that the offset() function in OpenSCAD only works for 2D objects. 
In my opinion this is functionality is missing, because if I 3D print some files from the internet
and I realize that the parts are too tight too join, I really loved to import them to OpenSCAD, offset by -0.3
and finally print the updated STL file with my printer.

https://imgpile.com/i/dXb41c

The Picture shows the offset for 3D function in action. Using this approach instead of 2 spheres will guarantee
an always constant shell thickness of 0.5 in this case.

I don't know, why it only displays correct with F6 and not with F5, but maybe somebody can shed some light on that.

I am aware, that you can always oversize objects by minkowski it with a small sphere, but the offset works
yields alternative results. It will always create sharp corners and in contrast to the minkowski approach, you can
easily undersize(and I believe its more efficient). It literally works by offsetting the faces of the object and then
recalculating the vertices. 
When offsetting with positive values, holes in the object might disappear , if you offset with negative values,
volumes could disappear without removing its  vertices. The effect on that must definitely  by studied.
In case this  yield wrong results, please demonstrate, else i am happy to see the OpenSCAD gets a new feature.


kwikius:
Potentially a very nice feature.

rcolyer:
I already provided this to gsohler on IRC, but I'll document here for record-keeping.  I attempted to design a maximally pathological testcase to stress the edge-cases of what a 3D offset would have to do.

Here are issues to look for in it:
- Deformations in the inner and outer cones.
- Deformations in the inner and outer cubes.
- Cone orientation inconsistentencies on the cone bit that slices through the cube in the back.
- Sensible cube merging of the open slit in the front on positive offsets.
- Adjacent object merge issues on positive offset.
- Successful opening of holes to internal spaces on negative offsets.

The potential issues I'm concerned about are about the difference between a particular algorithmic definition of what offset means, and user expectations for what it means in particular situations.  I think this test highlights several cases which can form a basis for considering this.  Also several elements are designed to challenge the code's ability to produce manifold output.

The current status at the time of posting is that the PR segfaults on this design (if and only if one of the offset lines are uncommented).  Once that is resolved we can evaluate the other aspects.

```
module OffsetTest() {
  difference() {
    translate([7, -4, 0]) rotate([0, 30, 0]) translate([0, 0, -10]) cylinder(r1=8, r2=0, h=10, $fn=100);
    translate([7, -15, -10]) cube(11);
  }
  rotate([-90, 0, 30]) {
    difference() {
      cube(18);
      translate([5, 0, 12]) rotate([0, 30, 0]) translate([0, 0, -10]) cylinder(r1=8, r2=0, h=10, $fn=100);
      translate([10, 0.5, 10]) cube(3);
      translate([10, 10, 3.5]) sphere(3);
    }
  }
  translate([17.5, 12, 0]) sphere(3);
}

//OffsetTest();
//offset(1) OffsetTest();
//offset(-1) OffsetTest();
offset(3) OffsetTest();
//offset(-3) OffsetTest();
//offset(6) OffsetTest();
//offset(-6) OffsetTest();
```

src/geometry/GeometryEvaluator.cc
```diff
@@ -535,12 +536,162 @@ Response GeometryEvaluator::visit(State& state, const RootNode& node)
...
+//   printf("visited is %p\n",this->visitedchildren);
+//   if(this != NULL) {
+//   for (const auto& item : this->visitedchildren[index]) {
+//      if (!item) {
+//		int   f=item.first();
+//	      printf("t %d %d\n",item.first.getDimension(), item.second.getDimension());
+//      }
+//      }
+//    }
       const Geometry *geometry = applyToChildren2D(node, OpenSCADOperator::UNION);
```
thehans, line 695: Don't call `applyToChildren2D` without checking dimension first.  See `GeometryEvaluator::applyToChildren(const AbstractNode& node, OpenSCADOperator op)` for operations which handle both 2D and 3D.  Probably would be easiest to call that, and add `OFFSET` to `enum OpenSCADOperator` in `enums.h`

PR #4531, opened 2023-02-23
Update French translation for "encrypted"

Crocmagnon:


jordanbrown0:
Just out of curiosity, what's the difference in meaning between the two variations?  Google Translate translates them both to "encrypted".

Crocmagnon:
« Encrypté » is an anglicism. The correct way to translate “to encrypt” is with the verb « chiffrer ».

There’s a website presenting (in French) the different forms we can find and why they’re suited (or not) to talk about cryptography.

https://chiffrer.info/

jordanbrown0:
Critically, although Google Translate recognizes encryptée, L'Académie Français does not - and it does recognize chiffrer.  Sounds like an authoritative answer to me.

It is amusing to read chiffrer.info through Google Translate, because it says things like
> why we say "encrypt", but not "encrypt".

t-paul (approved):
Thanks!

PR #4535, opened 2023-03-02
Fix leaking xsessions in export_png.cc

unDeadHerbs:
Found this when generating an animation resulted in the error:
> Maximum number of clients reached
> Unable to open a connection to the X server.
> DISPLAY=:0
> Can't create OpenGL OffscreenView. Code: -1.


t-paul (approved):
Nice, thanks!

Issue #4143, opened 2022-02-25
PDF export defaults to A4 paper with no way to change it


amatulic:
It's nice to see PDF export in build 2021.01. However, it seems stuck on outputting A4 size pages.

European size A4 paper is not readily available in the United States. When exporting a 2D design to PDF, the purpose is to get a drawing with a 1:1 scale. When printed on US letter-size (8.5x11 inches) paper, the bottom and right edge of the drawing are cropped off, because A4 paper is larger than US letter size. Scaling to fit the page defeats the purpose of having a 1:1 scale drawing.

There should be a setting available for default PDF output size.

Now, I designed my drawings to fit letter-size paper, so you might think, what is the problem with cropping? The problem is that the axis scale gets cropped off the bottom, and this is critical for calibrating the laser printer's scaling when printing from Adobe Acrobat.

JonMcLin:
I have been thinking about implementing this for my own purposes.  
Is this the right place for discussing the implementation?

I'm considering two additional features, and I also have some questions about best-practices with respect to some implementation decisions.

Features:
1. Add paper selection.  Probably a pick-list, A4, A3, Letter, Tabloid.   Later could add additional sizes.
2. Add selection of landscape vs format.  (I could automate this based on the bounding box proportion, but it would seem better at least initially to make it selectable, along with paper size)
3. Generalize the Origin position behavior for finer granularity.  Rather than bottom-left or center, place the origin at the center of the bounding box, to center the geometry on the paper.

I am not at all familiar with Qt beyond what I've read.  I have been going through the source code, and I've generally mapped out what I need to do.  But... 
-   I understand I need to add the above widgets to ui/preferences.ui (which I am assuming I use Qtcreator for; I'll figure that out), and add equivalent setValue method/action to the preferences.cc (or is that code inserted by Qt in the build process?)
-  It's not clear where the best place to get these values from settings would be.  For other export actions, the get methods where settings are needed is in the mainWindow.cc code, and there are no gets in the export*.cc code.  But that's because they all can use a common export signature.    So should I put the settings get calls for this purpose in the exportPdf.cc, or change the export method signature and put the gets in mainWindow.cc.

Thoughts?

t-paul:
@JonMcLin Yes, discussion here is fine, if you like to have a chat, feel free to join the IRC channel too.

[...]

PR #4537, opened 2023-03-04
Implement ExportPdfOptions Dialog with paper selection and more (#4143, partial #649)

JonMcLin:
Provides:
- Paper Size selection
- Layout Orientation selection
- Selective annotations
- A new "grid" option ("graph paper")

See [#4143](https://github.com/openscad/openscad/issues/4143) for details.

Added exportPdfDialog.ui, exportPdfDialog.h, exportPdfDialog.cc
Mods to CMakeLists.txt, export.h, export.cc, mainwindow.h, mainwindow.cc, export_Pdf.cc


t-paul (commented):
Thanks, very nice. I've updated the test files.

Note that this now keeps the design centered instead of guessing based on location of the design. In case other requiremnts are raised we can extend the options later.

PR #4538, opened 2023-03-05
implement new scale and translate parameter to rotate_extrude

schiele:
This implements two new vector parameters scale and translate to rotate_extrude with similar behavior that is known of twist in linear_extrude.

- scale: scale the base polygon along x and y axis in a linear way from [1,1] to the specified value from the start of the rotation to the end.  Since this parameter name was also used by the deprecated file importing function it is only available when not using file importing function.

- translate: translate the base polygon along x and y axis in a linear way from [0,0] to the specified value from the start of the rotation to the end.

This is an initial implementation working for me as a suggestion of how an interface and the implementation could look like. Given that I am not yet familiar with the OpenSCAD development community, I may have missed some interface or implementation guidelines. Feel free to point out name choices or design or implementation decisions that you don't like. I am open to adapting the code to the style usually applied.

### DIFF
--- a/src/core/RotateExtrudeNode.cc
+++ b/src/core/RotateExtrudeNode.cc
@@ -47,7 +47,7 @@ static std::shared_ptr<AbstractNode> builtin_rotate_extrude(const ModuleInstanti
 
   Parameters parameters = Parameters::parse(std::move(arguments), inst->location(),
                                             {"file", "layer", "origin", "scale"},
-                                            {"convexity", "angle"}
+                                            {"convexity", "angle", "translate"}
                                             );
 
   node->fn = parameters["$fn"].toDouble();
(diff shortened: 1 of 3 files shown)


gsohler:
Big difference between linear_extrude and rotate_extrude is, that the rotate_extrude object is closed whereas linear_extrude isn't.
When scale and translate are set wrong , when angle is 360 degree, the result will not be a manifold object.
If it was dummy safe to add these paramters to rotate_extrude, i believe, openscad would have these parameters for a very long time already.
it appears to be very hard to get a functionality approved when its not *ABSOLUTELY* unique. translate  parameter can very easily be applied to the 2D object instead.

schiele:
> Big difference between linear_extrude and rotate_extrude is, that the rotate_extrude object is closed whereas linear_extrude isn't. When scale and translate are set wrong , when angle is 360 degree, the result will not be a manifold object. If it was dummy safe to add these paramters to rotate_extrude, i believe, openscad would have these parameters for a very long time already. it appears to be very hard to get a functionality approved when its not _ABSOLUTELY_ unique. translate parameter can very easily be applied to the 2D object instead.

That means if we just made sure that the translation is big enough that after one full round the polygon would not intersect with the one from one round before this would solve our problem? I thought about such a test but then thought that I might better leave the responsibility to the user but could try to add such a test. Since the test likely needed to work on min/max values it might trigger when there is no actual intersection. Therefore should such a test just produce a warning if potentially intersecting or would you expect a hard error?

I am not sure how to apply the translate functionality to the base 2D object instead. Note, that similar to the twist functionality in linear_extrude this applies gradually during extraction. If I apply it to the base 2D object instead the same translation is applied globally. For instance "rotate_extrude(angle=360*10, translate=[0, 50]) translate([10, 0]) circle(1);" produces a spring while applying the translate=[0, 50] globally to the base 2D object always ends up in a torus. Did I misunderstand your comment?

gsohler:
Hi Robert,
Sorry, I was actually wrong in the interpretation of the translate parameter.
IMHI using the translate parameter gradually during one revolution is a nice feature and i'd love to see it.
Let's see, how it goes ...

schiele:
> Hi Robert, Sorry, I was actually wrong in the interpretation of the translate parameter. IMHI using the translate parameter gradually during one revolution is a nice feature and i'd love to see it. Let's see, how it goes ...

Ok, then I will put the checking for non-overlapping rotations onto my backlog then.

Meanwhile attaching some examples to increase the chances that the intention of the change is properly understood. With the patch we could get something like this:
testhelix
The corresponding code is:
```// A screw.
// The base cylinder.
cylinder(50, r=10);
// The screw thread.
rotate_extrude(angle=360*15, translate = [0,45]) translate([10, 0]) circle(d=3-1/128, $fn=4);
// At the end of the screw the thread should sink into the cylinder.
translate([0, 0, 45])
    rotate_extrude(angle=360, translate = [-1.5,3])
        translate([10, 0]) circle(d=3-1/128, $fn=4);

// A simple spring
translate([30, 0, 0])
    rotate_extrude(angle=360*10, translate=[0, 50])
        translate([10, 0]) circle(1);
        
// A flat spring
translate([0, 0, -10])
    rotate_extrude(angle=360*10, translate=[50, 0])
        translate([10, 0]) circle(1);
        
// A spring with increasing strength
translate([-40, 0, 0])
    rotate_extrude(angle=360*10, scale=2, translate=[0, 50])
        translate([10, 0]) circle(1);
```

schiele:
Rebased to the current master and factored out a fix independent of this change as in https://github.com/openscad/openscad/pull/4538#issuecomment-1455412985

schiele:
rebased to current master

schiele:
rebased to curent master

schiele:
Rebased to current master and fixed formatting issue in toString method as detected by test cases.

schiele:
Rebased to master again to retrigger the hanging Windows test.

schiele:
Changed the code to use the simplified LOG() overload in a place I missed before.

agausmann:
I believe marking this as "stalled" is a mistake; it seems like this was waiting for response from maintainer, not from the author.

If the project maintainers are never going to be interested in this then it's fine to disclose that, but it's strange to apply an automated stalebot to unreviewed PRs.

PR #4539, opened 2023-03-06
move range check outside loop in random_extrude

schiele:
We should move the range check outside the loop calculating min/max x coordinates. There are two reasons behind this:
1. With the old implementation a violation might get detected early without going through all loops but this approach makes the loop processing in the non-error case slower due to repeated testing. It is advisable to optimize for the non-error case of optimization for error case since the non-error case is the one users will execute normally.
2. With the old implementation the printed range is incorrect since it will only print the range of all the vertices processed so far instead of the full range as printed in the new implementation. This produces quite confusing values in the error case.


schiele:
rebased to current master

t-paul (approved):
Thanks, both are good points.

Issue #4444, opened 2022-12-09
rotate_extrude(angle = 180, $fn=3)  creates non-manifold


Parkinbot:
not a big thing, just a pathological case that leads to a non manifold result

  rotate_extrude(angle = 180, $fn=3) 
  polygon([[0,0],[10,0],[0,10]]);   // anything here

PR #4540, opened 2023-03-06
round up number of fragments in rotate_extrude

schiele:
The old code rounded down the number of fragments in rotate extrude, which lead to weird results when working with a low number of $fn. By rounding up that number we always at least get the number of fragments needed to span a real body when working with degrees > 180.


schiele:
fixes #4444

t-paul (approved):
Thanks! This opens some potential changes to geometry with higher values, but lets cross that bridge if we come there.

PR #4561, opened 2023-03-18
[DRAFT] Flatten CSG tree to let more lazy unions bubble up to the top

ochafik:
This is a rehash of https://github.com/openscad/openscad/pull/3637, hopefully more likely to be productionized.

Goals:
*   Get as many lazy top-level unions as possible
*   Push down transforms to reduce the # of transforms being done, mitigate precision concerns (esp. w/ the Manifold rendering backend introduced in https://github.com/openscad/openscad/pull/4533) + allow more unions to bubble up.

Note that I've improved Manifold's tree flattening in https://github.com/elalish/manifold/pull/368. We could potentially ask Manifold for a lazy top-level union instead of its full rendering (it keeps a lazy CSG AST under the hood), which could be a better approach than this PR (it might still need to deal with the issue of transforms precision). Still, I wanted to get this draft PR out there as it works even w/o Manifold.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -616,6 +616,7 @@ set(CORE_SOURCES
   src/core/SourceFile.cc
   src/core/SourceFileCache.cc
   src/core/StatCache.cc
+  src/core/TreeFlattener.cc
   src/core/UserModule.cc
   src/core/Tree.cc
   src/core/customizer/ParameterObject.cc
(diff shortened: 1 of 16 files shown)


pca006132:
> We could potentially ask Manifold for a lazy top-level union instead of its full rendering

Yeah, we can just provide a `GetMeshes` function that returns a list of meshes that should be unioned together.

PR #4576, opened 2023-03-24
Refactor offscreen context management

kintel:
The goal is to leave the functionality unchanged, but with better/more extendible APIs, in order for later to bring in the context creation code from https://github.com/kintel/offscreen


bradley-jarvis:
This patch breaks the compile for me, in src/glview/offscreen-old/OffscreenContextEGL.cc the decleration of

std::shared_ptr<OffscreenContext> CreateOffscreenContextEGL(
    uint32_t width, uint32_t height, uint32_t majorGLVersion, 
    uint32_t minorGLVersion, bool gles, bool compatibilityProfile,
    const std::string& drmNode = "")

include a default parameter when it shouldn't, removing the = "" on parameter 7 allows the code to compile completely

kintel:
@bradley-jarvis Will fix. Out of curiosity, what system and compiler do you use, and how are you building OpenSCAD?

bradley-jarvis:
Thanks, I use Gentoo and compiling from the openscad-9999.ebuild, g++ version 12.2.1_p20230304, I use it along with the latest freecad

media-gfx/openscad-9999::gentoo  USE="cairo dbus egl gui -experimental -gamepad -hidapi -mimalloc -spacenav"

kintel:
Thanks! We don't yet have continuous builds set up for the EGL build, but I'll look into setting that up, so that such build errors are caught earlier next time.

PR #4582, opened 2023-03-30
Add new, more capable, WGL offscreen support

kintel:
This is the remaining part of the new offscreen classes: WGL support.

TODO:
* [x] Revert provider selection back to "-old", until they're better tested
* [x] WGL
  * [x] Implementation
  * [x] Error handling
   * Automated tests
     * [ ] GLEW + new
     * [ ] GLEW + old
     * [ ] GLAD+ new
     * [ ] GLAD + old
* [x] (optional) NSOpenGL -> Meh, not needed. It's nice to get rid of that ObjC code
* [x] Verify that new OpenGL versions can be requested (although OpenSCAD may fail) -> postponed to later
* [x] Consider generating GLAD sources inline. This may require us to set up a more consistent Python environment on all platforms. -> out of scope; Python is already tricky enough to deal with across platforms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -656,7 +656,9 @@ elseif(WIN32)
   if(NOT NULLGL)
     set(OFFSCREEN_METHOD "Windows WGL")
     message(STATUS "Offscreen OpenGL Context - using Microsoft WGL")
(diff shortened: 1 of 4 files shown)


kintel:
This is pretty much done, but need more Windows testing, and my Windows installation didn't survive long enough to get this tested.

kintel:
Still valid - pending testing on Windows

kintel:
Will reopen

PR #4585, opened 2023-04-01
Test using msys2 python for tests

kintel:
This PR fixes some issues running test on msys2:
* Prefer using the newer FindPython3 CMake macro
* Look for venv binaries in the appropriate folder: Scripts/ on Windows (except in=f bin/ was created in Unix-like Windows environments like msys2)
* Install python packages for msys2 using pacman to work around issues with building wheels on msys


scripts/msys2-install-dependencies.sh
thehans, line 1: I think omitting the `p` suffix here might help  (and leave numpy+pillow out for the tests' venv to handle)
```
    python-pip: \
    python-numpy: 
```

The `p` means to grab the package built for a specific runtime (we are defaulting to UCRT).  For libraries we want them all the same but for tools I don't think it should matter.

https://packages.msys2.org/package/python
https://packages.msys2.org/package/python-pip

kintel: Challenge is that both gdb and glib2 requires Python, so the runtime-specific Python is already installed

thehans: You could also try a different value for `MSYSTEM_DEFAULT` in this file (OR just set the `MSYSTEM` environment variable before calling).  
  
I changed it to `UCRT64` a few months ago.  Previous behavior didn't use `MSYSTEM` but listed the package names fully prefixed with `mingw-w64-x86_64`.  (which is equivalent to setting `MSYSTEM=MINGW64`)

My reasoning was that the ucrt is supposed to be microsoft's more modern replacement for msvcrt,  and it didn't cause any issues *at the time*.  I wonder if switching it back to `MINGW64` would behave any better regarding python though?
Here's the breakdown of the differences:  https://www.msys2.org/docs/environments/#overview

Here is the code before I started changing it.  The old style didn't use pacboy, and listed the fully prefixed package names:  
https://github.com/openscad/openscad/blame/2ad1823f16acf30ddac9f0c74941f1ae3a32faea/scripts/msys2-install-dependencies.sh#L11-L27

kintel: We were just chatting about ucrt64 vs. mingw on IRC today. I'm planning to try it out. Someone mentioned that there is also a mingw-ucrt runtime.

kintel: I tried this, but it didn't really pan out. It was a while ago and I forgot to update this, so I don't remember the details..

PR #4586, opened 2023-04-02
SDF support via libfive

lf94:
the time is near for a new era of code cad

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -618,6 +618,7 @@ set(CORE_SOURCES
   src/core/DrawingCallback.cc
   src/core/RotateExtrudeNode.cc
   src/core/SurfaceNode.cc
+  src/core/SdfNode.cc
   src/core/TextNode.cc
   src/core/TransformNode.cc
   src/core/UndefType.cc
@@ -773,6 +774,10 @@ if(ENABLE_MANIFOLD)
   target_compile_definitions(OpenSCAD PRIVATE ENABLE_MANIFOLD)
   list(APPEND Sources ${MANIFOLD_SOURCES})
 endif()
+
+
+# TODO: PROPER LIBFIVE INCLUSION
+target_link_libraries(OpenSCAD PRIVATE five)
 
 #
 # Offscreen OpenGL context source code
--- a/src/core/Builtins.cc
+++ b/src/core/Builtins.cc
@@ -48,6 +48,7 @@ extern void register_builtin_transform();
 extern void register_builtin_color();
 extern void register_builtin_primitives();
 extern void register_builtin_surface();
+extern void register_builtin_sdf();
 extern void register_builtin_control();
 extern void register_builtin_render();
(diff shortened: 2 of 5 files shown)


lf94:
So uh, it segfaults. Gonna need some help on that one

lf94:
Here's a test you can use:

```
$ cat test.scad
sdf(function(x,y,z) sqrt(x^2+ y^2 + z^2)-1, [-1,-1,-1,1,1,1], 1.0);
```

pca006132:
I wonder if you are interested in trying the sdf function of manifold, which is already included in experimental features. It would also be interesting to compare the performance of manifold and libfive (we do not use eigen so it may be slower).

gsohler:
Chun Kit,

libfive SDF is already included in my python fork
watch out my homepage for a demo

this is libfive side of SDF

https://guenther-sohler.net/openscad/pictures/collosseum.png
https://guenther-sohler.net/openscad/examples/collosseum.txt

lets also test Manifold SDF ...

pca006132:
OK thanks for the pointer. We currently don't have that many builtin functions for sdf in manifold, so porting the example to manifold will likely take a bit of time. Also, I originally thought that libfive does JIT, but it seems that they don't have this function (but should not be too hard to implement I guess, considering they are all numerical operations).

Manifold currently has this nice SDF example: https://github.com/elalish/manifold/blob/master/samples/src/gyroid_module.cpp
I can try to see if I can get this to work on your fork.

pca006132:
(the sdf example is available on manifoldcad.org as well)

pca006132:
Doable because manifold supports it under the hood. It is not yet supported by OpenSCAD for now.

The Python PR by @gsohler provides libfive integration, but it is only exposed for python.

lf94:
From what we discussed in the IRC channels months ago, it seems manifold is quite slow compared to libfive. At the same time, it's not a good idea to duplicate functionality. So I'm not sure what should be done. The only reason I've abandoned this is because the lack of (understandable) motivation from others :) Feel free to pick up everything.

pca006132:
Indeed, manifold is slower for more complicated cases (at least for now) as the SDF code is very simple. But we guarantee manifoldness and we have plans to make it faster. More users can help us improve better.

lf94:
All I can wish you is good luck :slightly_smiling_face: I did this for OpenSCAD out of the motivation to hopefully pull more people into code CAD, but extending the problem space OpenSCAD can solve. Given I've spent enough time on the proof of concept, all I can do is hope someone else (maybe you!) to continue. I'm working on my own code CAD system where SDFs are first class. OpenSCAD has been a deadend for me for many years now tool-wise. I still recommend OpenSCAD to those starting out regardless since the community is extremely nice and there's lots of resources :)

t-paul:
The treading model of libfive makes it a bit difficult to call back into OpenSCAD expression evaluation. Depending on the API Manifold uses, it could be easier to integrate.

PR #4588, opened 2023-04-03
Adding Python Support to OpenSCAD

gsohler:
First trial failed due to outdated python-support branch.

Now i have created a fresh copy of master branch and merged i all required python support.
I am curious to see the actual file  change set AND ctest result as it looks promising at least locally.


kintel:
One suggestion, if it takes time to get the Python stuff merged: Extract all non-functional non-Python-related changes into a separate cleanup PR, which could be merged quickly. That should make this PR a bit more targeted.

gsohler:
Using python, The script  can be much smaller and much more performant at the same time, than the openscad language counterpart. 
The picture shows an 6th order 3D Hilbert curves and it rendered in few seconds on my linux virtual machine only
It used less than 20 lines and half of it is used for octant decoding.

https://imgpile.com/i/903Q9a

gsohler:
@teepee which milestones are left for being able to merge the python engine as an experimental feature into the OpenSCAD main branch ?

gsohler:
@t-paul  which milestones are left for being able to merge the python engine as an experimental feature into the OpenSCAD main branch ?

gsohler:
superceeded bei #4702

PR #4590, opened 2023-04-04
Python prepare

gsohler:
As Suggested by kintel:

This is a PR with only refactoring, no single functional change

Its meant to get prepared for the pyhton.

Using this the file change set is about half ...

### DIFF
--- a/src/core/primitives.h
+++ b/src/core/primitives.h
@@ -1,0 +1,181 @@
+/*
+ *  OpenSCAD (www.openscad.org)
+ *  Copyright (C) 2009-2011 Clifford Wolf <clifford@clifford.at> and
+ *                          Marius Kintel <marius@kintel.net>
+ *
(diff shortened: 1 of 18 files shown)


gsohler:
image

I don't get the error as in ubuntu-20.04 build when running ctest locally.
the shown wireframe error is probably due to my virtualization.

ochafik, do you have an idea why your mainfold test starts to fail after just structural changes and only in a single build ?

src/core/primitives.h
```diff
@@ -0,0 +1,181 @@
...
+#include <sstream>
+
+
+struct point2d {
+  double x, y;
+};
+
+struct point3d {
+  double x, y, z;
+};
```
t-paul, line 37: The names are maybe a bit generic to have them in a public header, but as there's no issue right now lets postpone changes until there's a conflict.

t-paul (approved):
Looks good, thank you.

PR #4592, opened 2023-04-09
Allow printing by custom command

asdil12:


src/gui/MainWindow.cc
```diff
@@ -2125,6 +2130,37 @@ void MainWindow::sendToPrintService()
...
+  ExportInfo exportInfo = createExportInfo(FileFormat::STL, exportFilename, activeEditor->filepath);
+  exportFileByName(this->root_geom, exportInfo);
+
+  cmd.replace("#1", exportFilename);
+  LOG(message_group::None, Location::NONE, "", qPrintable(cmd));
+  if (fork() == 0) {
+    // Run cmd in background
+    std::system(qPrintable(cmd));
+    exit(0);
+  }
```
t-paul, line 2160: This will not work on Windows.

As this is part of the GUI anyway, using `QProcess::startDetached` should work cross platform and allows simple error checking if the command actually was started.

src/gui/MainWindow.cc
```diff
@@ -2125,6 +2130,37 @@ void MainWindow::sendToPrintService()
...
+    LOG(message_group::Error, Location::NONE, "", "Could not open temporary file.");
+    return;
+  }
+  const QString exportFilename = exportFile.fileName();
+
+  //Render the stl to a temporary file:
+  ExportInfo exportInfo = createExportInfo(FileFormat::STL, exportFilename, activeEditor->filepath);
+  exportFileByName(this->root_geom, exportInfo);
+
+  cmd.replace("#1", exportFilename);
```
t-paul, line 2154: A simple replace is probably going to fail if there's spaces in the directory and/or filename.

t-paul (changes requested):

src/gui/MainWindow.cc
asdil12: updated to use `startDetached()`

src/gui/MainWindow.cc
asdil12: This should take care:
```diff
+  QStringList args = QProcess::splitCommand(cmd);
+  int var_idx = args.indexOf("#1");
+  if (var_idx >= 0) {
+    args.replace(var_idx, exportFilename);
+  }

asdil12:
@t-paul could you have a look if I everything is fine now?

jordanbrown0:
Appears to have been addressed by #5299 et seq.

PR #4597, opened 2023-04-12
GitHub CI: Cancel jobs if a new push happens while jobs are running

kintel:
This somehow doesn't work. Any ideas are welcome.
It's not a huge deal, but we're currently wasting a lot of cycles whenever a second push happens while the CI is still running.


kintel:
This didn't work out

PR #4598, opened 2023-04-12
Prefer CMake's FindGLEW

kintel:
Now that CMake has FindGLEW, let's use it!
However, CMake's FindGLEW doesn't work on MXE, so we write a custom macro just for MXE.


thehans:
Nice to see consolidating on cmake's imported target style with `GLEW::glew`.  
And a net negative LOC is always a good sign too IMHO 😉

PR #4601, opened 2023-04-14
Python delivered in tiny bites

gsohler:
python infrastructure,  and all switches added, but no actual useful function added yet


gsohler:
I have added functionality to save the hash of trusted python files along with the QSettingsCached.
On each unknown file the user is asked whether it trusts this one or all of them(during session only)
And there is a new Menu item File->Revoke trusted python files to undo all trusted files.

gsohler:
rebased to current master

t-paul:
Will merge via #4699, moving away from using crypto++ as git submodule.

PR #4608, opened 2023-04-16
Remove need to cd into tests directory

kintel:
Need testing:

* [x] macOS local
* [x] macOS GitHub
* [x] Linux local (incl coverage)
* [x] Linux GitHub (incl. coverage)
* [x] ~~Windows local: Not run as msys2 tests are not happy atm.~~
* [x] Windows GitHub
* [x] ~~Windows MXE test package: Not run as this stopped working a while ago~~


rcolyer:
- Testing ran successfully in both build and build/tests with both make test and ctest in each.
- It picked up the CTEST_PARALLEL_LEVEL environment variable correctly, as expected, for parallel building.
- make test -j8 does not shuttle the parallelization level in.  Is this possible?  Should we be using a different parallelization syntax with this approach?
- ctest -j8 works correctly for parllelization in both locations, as expected.
- The README.md update should probably more clearly state "Once built, you can run tests with make test or ctest from the build directory." since this is dislocated from the command above it that puts it into that directory.
- Now that there are two ways to run the tests, there are two differences in the log output locations, with them appearing in build/Testing/Temporary and build/tests/Testing/Temporary depending on the starting directory.  Should/can this be made consistent?

kintel:
Thanks for the review! You're right that `make test` doesn't have a good story for parallel execution. The best I can come up with is `make test ARGS=-j10`, which isn't a great cmd-line to remember. I've removed references to `make test`, so we can continue recommending running `ctest` directly.

thehans:
https://github.com/openscad/openscad/blob/master/CMakeLists.txt#L151-L157

In terms of simplifying the overall commands for build+test, I would typically run something like this:
```
mkdir build && cd build
cmake -GNinja -DCMAKE_BUILD_TYPE=Release -DEXPERIMENTAL=ON .. && cmake --build . && ctest -j32
```
Note that currently 88% of tests fail if you forget to add `EXPERIMENTAL=ON`  
I also tend to utilize bash history quite a bit with Ctrl-R to repeat long cmake commands etc.

One of the nice things about Ninja is that it automatically detects # of threads and utilizes them all without specifying any `-j`.
I decided to set Ninja as my default via environment variable in my `~/.bashrc`, and also found a environment variable setting for ctest jobs:
```
export CMAKE_GENERATOR=Ninja
export CTEST_PARALLEL_LEVEL=32
```

With that you can omit the `-GNinja` arg from cmake and `-j` arg from ctest
```
cmake -DCMAKE_BUILD_TYPE=Release -DEXPERIMENTAL=ON .. && cmake --build . && ctest
```

A long time ago I added a little convenience snippet to our CMakeLists that would set `CMAKE_BUILD_TYPE=Release` if left unset by the user.
https://github.com/openscad/openscad/blob/master/CMakeLists.txt#L151-L157
I was kind of on the fence whether or not this was a good idea, and at some point I decided to comment it out.
But, if simplifying commands is a goal, we could always re-enable that, so a basic build+test command would shorten further (still assuming env setup mentioned above) to:
```
cmake -DEXPERIMENTAL=ON .. && cmake --build . && ctest
```

Another option would be to just abort the build if not set:
```
if(NOT CMAKE_BUILD_TYPE)
  message(FATAL_ERROR "You must set CMAKE_BUILD_TYPE")
endif()
```

I view doing a build with blank `CMAKE_BUILD_TYPE` as being analogous to undefined behavior in C++.  So I think I'd like to see at least one of those implemented, in order to avoid that possibility.

kintel:
I think that a common practice for a lot of software is that the default build should be releasable, so defaulting to a release build would make sense.

kintel:
Updating docs and scripts to utilize `cmake --build` and `cmake --install` could be good too, but let's keep each PR small :)

kintel:
The experimental test issue is worth a ticket, to collect any discussion and activities related to that: #4612

t-paul:
> I think that a common practice for a lot of software is that the default build should be releasable, so defaulting to a release build would make sense.

I believe that interferes with some distro builds the expect the `CMakeList.txt`  does not change this. This setting is most confusing but I think the "not set at all" is actually a valid and expected value for CMAKE_BUILD_TYPE.

kintel:
Humm, good point. I guess cmake supports building multiple profiles at the same time. I'll read up on it, but let's keep that out of this PR then.

PR #4609, opened 2023-04-17
Added missing packages information for build on MacOS

gordio:


README.md
```diff
@@ -142,6 +142,7 @@ Prerequisites:
 
 * Xcode
 * automake, libtool, cmake, pkg-config, wget (we recommend installing these using Homebrew)
+* meson and ninja (can be installed with `pip3 install --user meson ninja`)
```
kintel, line 145: ninja isn't used by default, and not required. We recommend installing meson using homebrew, so it can be added to the list above. See here for how we install dependencies for release builds: https://github.com/openscad/openscad/blob/master/.circleci/config.yml#L182

gordio: For some reason configuration is failed without `ninja`. Probably coz I installed `meson` with pip.

PR #4610, opened 2023-04-18
Fix: Bump required version of glib2

gordio:
glib2 2.75 is required to be build on MacOS with latest clang (14.0.3). More info https://gitlab.gnome.org/GNOME/glib/-/issues/2759

### DIFF
--- a/scripts/macosx-build-dependencies.sh
+++ b/scripts/macosx-build-dependencies.sh
@@ -52,7 +52,7 @@ PACKAGES=(
     "fontconfig 2.14.1"
     "hidapi 0.12.0"
     "lib3mf 1.8.1"
-    "glib2 2.71.0"
+    "glib2 2.75.0"
     "pixman 0.42.2"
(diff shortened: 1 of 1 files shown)


kintel:
Thanks! Verified to work in a local VM, as our CI doesn't run the latest XCode

kintel:
@gordio Had to revert this as it broke the snapshot build. Something most have changed in how glib builds libintl..

kintel:
I eventually pinned it down to this patch which made it into glib-2.75.2: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3177

Next time I have motivation I'll try bumping to 2.76.3 and manually patching the glib build system to work again.

Perhaps time to revive the discussion on replacing glib, e.g. https://github.com/openscad/openscad/pull/2523

kintel:
ok, I found that 2.75.1 actually works on macOS 13, so that will buy us some time: https://github.com/openscad/openscad/pull/4663

PR #4626, opened 2023-05-05
add missing include of <cstdint> for uint32_t

schiele:
An update in the compiler toolchain to gcc 13 detected this obvious oversight.


thehans:
Thanks!

Issue #4316, opened 2022-08-02
ASCII STL export is slow


jordanbrown0:
For `sphere(10, $fn=100)`, binary STL export using
```
openscad -o /tmp/a.stl --export-format=binstl /tmp/a.scad
```
takes about 0.3 seconds, but ASCII STL export takes about 5.5 seconds.  (That's for the entire execution, as measured with "time" on my Windows system under MSYS2.)

That seemed like a lot, so I did some experimentation and ... it *is* a lot.

The time is almost all spent in the [`fromString` function](https://github.com/openscad/openscad/blob/ea8950bd3cd7f06da333c15463492e484098e591/src/io/export_stl.cc#L49).

I don't know exactly what's involved in the fancy C++ stuff that it's doing, but when I replaced it with a function that did the same thing with primitive operations, the time dropped to about 0.8 seconds, and the resulting file was bit-for-bit identical.

```
#include <cstdlib>
#include <cstring>
...
Vector3d fromString(const std::string& vertexString)
{
  double p0, p1, p2;
  const char *s = vertexString.c_str();
  p0 = atof(s);
  s = strchr(s, ' ') + 1;
  p1 = atof(s);
  s = strchr(s, ' ') + 1;
  p2 = atof(s);
  return Vector3d(p0, p1, p2);
}
```

Note:  yes, that's a horrible implementation if you don't know exactly what the string looks like.  But here we *do* know what the string looks like, and I just wanted to see how fast it would be.  Safer implementations left as an exercise for the reader.

But:  Will anybody care?  Binary STL is faster and smaller; why would you use ASCII STL?  And yes, it's slower than it needs to be, but will a few seconds really matter after you've gone through a multi-minute render?

You can make it a bit faster still by eliminating some of the gyrations that it goes through to do processing with ASCII strings rather than with floats or doubles.  An experiment says that drops it to 0.6 seconds, and the source is a lot simpler.  But then many (all?) of the normals are off in the fourth or fifth significant figure, compared to the original.  Does that matter?  Don't know.  Which one is more correct?  Don't know.  Theoretically it might also generate degenerate triangles, when vertices differ when represented as binary but become the same when represented as ASCII.

thehans:
The original reason for "processing with ASCII strings rather than with floats or doubles" can be seen here: #1853

jordanbrown0:
[...]

Issue #2651, opened 2018-12-26
STL export should contain single precision floats


3D-ES:
Hello!

I am [designing 3D KiCAD PCB parts using OpenSCAD and Wings3D](http://happyrobotlabs.com/posts/tutorials/tutorial-3d-kicad-parts-using-openscad-and-wings3d/)

When I import an OpenSCAD STL file in Wings3D, the program always crashes,
after repairing the file using [meshconv](https://www.patrickmin.com/meshconv/), the STL file no longer crashes Wings3D.

Meshconv only changed the normal and vertex locations to floating point values:

STL output from OpenSCAD test with cube(10);

    solid OpenSCAD_Model
        facet normal -0 0 1
            outer loop
                vertex 0 10 10
                vertex 10 0 10
                vertex 10 10 10
            endloop
        endfacet
        ...

STL output after repairing cube(10) with meshconv:

    solid cube
    facet normal -0.000000 0.000000 1.000000
        outer loop
            vertex 0.000000 10.000000 10.000000
            vertex 10.000000 0.000000 10.000000
            vertex 10.000000 10.000000 10.000000
        endloop
    endfacet
    ...

The [STL file format specifies](http://www.fabbers.com/tech/STL_Format):

> The numerical data in the facet normal and vertex lines are single precision floats, for example, 1.23456E+789.

Currently the STL file contains only integers when testing with a cube(10);
After meshconv the value notation is in std::fixed style, so also incorrect?...

I made the following changes in /src/export_stl.cc:

    std::string toString(const Vector3d &v)
    -   return OSS(v[0] << " " << v[1] << " " << v[2]);
    +   return OSS(std::scientific << v[0] << " " << v[1] << " " << v[2]);

    void append_stl(const PolySet &ps, std::ostream &output)
    if (is_finite(normal) && !is_nan(normal)) {
    -       output << normal[0] << " " << normal[1] << " " << normal[2] << "\n";
    +       output << std::scientific << normal[0] << " " << normal[1] << " " << normal[2] << "\n";
    }
    else {
    -       output << "0 0 0\n";
    +       output << "0.000000e+00 0.000000e+00 0.000000e+00\n";
    }

STL output of cube(10) after making the above changes:

    solid OpenSCAD_Model
        facet normal -0.000000e+00 0.000000e+00 1.000000e+00
            outer loop
                vertex 0.000000e+00 1.000000e+01 1.000000e+01
                vertex 1.000000e+01 0.000000e+00 1.000000e+01
[...]

PR #4643, opened 2023-05-19
More precise & up to 2-5x faster STL output

ochafik:
Fixes https://github.com/openscad/openscad/issues/4316 (and the precision part of https://github.com/openscad/openscad/issues/2651 )

`sphere(10, $fn=1000);` now renders 5.3x faster in ASCII STL and 2.5x faster in binary STL (tested on M1 mac).

Notes:
- Using wicked-fast double-conversion's [ToShortestSingle](https://github.com/google/double-conversion/blob/master/double-conversion/double-to-string.h#LL245C8-L245C24) to format floats (as suggested by @thehans in the thread below), which also guarantees that parsing gives back the exact same float value.
  - As a result, ASCII STL output now has the same precision as the binary format (reading it is guaranteed to yield the exact same float coordinates)
  - Also means we no longer need to parse strings after formatting, for further speedups
- Uses an indexed mesh: 
  - Added a tesselate_faces overload that outputs the intermediate indexed mesh instead of wasting time building a PolySet.
  - For ASCII export, caching the results of each vertex's toString
- Both overloads of tesselate_faces are also faster because we now skip the indexed faces stage (made possible by making the Reindexer's points vector grow incrementally / available for lookups w/o overhead)
- Skipping expensive superfluous tests in STL output (tesselation gives us guarantee of no duplicate vertices in each poly)
- Batch vector outputs in binary STL mode for extra speedups
- Preallocation of polygon vectors in the PolySet append_poly / append_vertex flow yields up to 10% render speedups on its own and was merged independently in https://github.com/openscad/openscad/pull/4642 (this PR's branch has now been rebased)


src/io/export_stl.cc
```diff
@@ -48,12 +50,19 @@ Vector3d toVector(const std::array<double, 3>& pt) {
...
-Vector3d fromString(const std::string& vertexString)
+Vector3f fromString(const std::string& vertexString)
 {
-  Vector3d v;
-  std::istringstream stream{vertexString};
-  stream >> v[0] >> v[1] >> v[2];
-  return v;
+  auto cstr = vertexString.c_str();
+  return {
+    strtof(cstr, (char**)&cstr),
```
t-paul, line 57: Does that work in locales having something else than '.' as decimal separator? The man page says it's using locale settings.

ochafik: Ah! Luckily there's a a `setlocale(LC_NUMERIC, "C");` in the parent of all append_stl functions that's meant to force '.' as decimal separator for both ways of parsing in this context (weirdly tho, I tried to force the locale to "fr_FR.UTF-8" there and the output didn't budge - I was expecting commas 🧐)

thehans: Keep in mind that `setLocale` has been shown to be incredibly slow on windows if you enable localization in preferences. #2744

ochafik: @thehans oh, does that setlocale currently make ASCII STL export slower on Windows? (sounds like something exciting to follow up on, although this PR shouldn't make things worse in that respect)

thehans: I don't know how much it is specifically affecting STL output speed.  
IIRC the test case there was particularly slow during CSG generation (sample code had tons of polyhedron data, thousands of vertex coordinates that require double to string conversion).

That was the impetus for adding `double-conversion` library, and so any time we print a number through our `Value` variant class it now goes through that.  But I don't think any of the export code was ever touched as part of those changes.

There has been some discussion more recently about switching that out for `libfmt` aka {fmt} which is supposed to be even faster than `double-conversion`.
I spent quite some time a few months ago attempting to swap out the libraries but didn't quite get it to a fully working/presentable state.  I might revisit it soon if I find the energy.

thehans: It might not be particularly relevant in this case, I had a bit of a knee-jerk reaction to seeing the mention of `setlocale`.  But as long as its not being called on every number conversion, it shouldn't be too terrible.  
The problem was that we weren't even calling setlocale directly, it was just a consequence of the code path taken deep within `printf` or  `ostream operator<<` or whatever the previous code was calling.

thehans: Also I think the latest windows builds should be linking with `ucrt` instead of `msvcrt`, which might not be affected by those same performance issues.

ochafik: @thehans Thanks for the pointers! Also, caveat: been a bit feverish this week so don't take anything I write at face value 🤒🤪 (trying to find the [ideal body temperature for coding](https://xkcd.com/323/))

I've tested double-conversion ([source here](https://gist.github.com/ochafik/fca4d509b9e8cd314c7fe49153724f2b)) and it seems faster than STR (ostringstream), but not as fast as a simple `sprintf(output, "%.6g %.6g %.6g", v[0], v[1], v[2]);`. In the other direction, it's also slower than `strtof`.

I've now also fallen into the rabbit hole of https://github.com/openscad/openscad/issues/2651 (now hoping to switch from 6 to 9 digits in this PR - might still be faster btw), and realized this all needs better test cases, which I'll add before further experiments (I've probably already silently broken the round-trip logic for distinct vertices that are formatted to the same string, for instance). Good thing this is still officially a draft 😅

thehans: You might also try casting `double` to `float` and calling [`ToShortestSingle`](https://github.com/google/double-conversion/blob/master/double-conversion/double-to-string.h#LL245C8-L245C24)

ochafik: Re/ doubleconv it’s weird indeed, maybe I’ve misused it (tried same Params as the Value code, then more naive params), or the mac stdlib is faster than average (might try [this benchmark](https://github.com/miloyip/dtoa-benchmark) if I have a chance), or… the way I naively stripped trailing zeros negated its performance? (Hadn’t seen  ToShortestSingle, will try next)

thehans: From your gist `trimTrailingZeros`  doesn't appear to actually get called anywhere.  

Anyways, simply stripping zeroes from the end is not even remotely correct behavior, in the event that:
1) There is no decimal point:  "1000" becomes "1"
2) There is an exponent
   a) "1.000000e6" would not get trimmed at all  
   b) "1.000000e10" would trim the exponent "1.000000e1" #2950

thehans: The more I think about it, I feel that using `ToShortestSingle` would make the most sense here, and would allow for simplifying the code much further:

`ToShortestSingle` creates the shortest **exact representation** of the input value (a "round trip" guarantee).  So no need to worry about trailing zeros (and no directly specifying any precision)

Additionally, all of the code going from double to string and back, was for the purpose of normal calculations matching up after reading in the decimal ASCII converted vertex positions (#1853).
If you downcast `double` to `float` and calculate normals from that, then there's no need to perform the round trip (string back to float) ourselves, because no further precision will be lost(post-downcast) by `ToShortestSingle`.

tests/data/scad/stl/stl-import-limits.stl
```diff
@@ -0,0 +1,86 @@
+solid OpenSCAD_Model
+  facet normal -1 0 0
+    outer loop
+      vertex 0 0 0
+      vertex 0 0 1.23456788
+      vertex 0 1.23456788 1.23456788
+    endloop
+  endfacet
+  facet normal 0 0 -1.47171605e-07
```
thehans, line 9: Why is there a leading zero on all the exponents?  Is `double-conversion` actually outputting that for you, and this test is being run succeessfully?  I don't think it should.

e.g.
```
echo(1.19208998e-07); // 1.19209e-7
```

src/io/export_stl.cc
```diff
@@ -39,123 +41,144 @@
 
 namespace {
 
-std::string toString(const Vector3d& v)
+/* Define values for double-conversion library. */
+#define DC_BUFFER_SIZE (128)
+#define DC_FLAGS (double_conversion::DoubleToStringConverter::UNIQUE_ZERO | double_conversion::DoubleToStringConverter::EMIT_POSITIVE_EXPONENT_SIGN)
```
thehans, line 46: I know this is basically copied verbatim from Value.cc, but I would remove the `EMIT_POSITIVE_EXPONENT_SIGN` flag here.  I was planning to remove that one from the original code too.  Also meant to convert all these to `constexpr` instead of defines.

See this [snippet from my unmerged PR](https://github.com/openscad/openscad/pull/4384/files#diff-796468af57e236dd8c69179c2126ae82e0c22250452807322bc3dd53d2a9b336R14-R23) for the converted values.  I had also tweaked a couple of other values, but I don't know if those are relevant to `ToShortestSingle` (or `ToShortest`) without reviewing the `double-conversion` "docs" (header file comments).

I basically halted work on that PR to start implementing libfmt, but that ended up being much more work than anticipated.

tests/data/scad/stl/stl-import-limits.stl
thehans: The above comment was a small nitpick for the purpose of just minimizing ASCII filesize... but actually thinking about it some more, this appears to be an error that the normal is not normalized to a unit vector.

ochafik: Ah yes I forgot to copy the output back to the input (same file is in expectations and input folder). This was from the earlier %.9g formats

src/io/export_stl.cc
ochafik: Ah, cool PR! I've dropped EMIT_POSITIVE_EXPONENT_SIGN and will keep the defines consistent-ish for now (unless you'd prefer I constexprify both places in this PR).

Re/ libfmt, once all compilers fully support C++20 (clang 14 doesn't seem there yet), would we get enough of its features through std::format?

tests/regression/stlexport/stl-import-export-expected.stl
thehans, line 1: Is this supposed to be empty?

tests/regression/stlexport/stl-import-limits-expected.stl
```diff
@@ -0,0 +1,86 @@
+solid OpenSCAD_Model
+  facet normal -1 0 0
+    outer loop
+      vertex 0 0 0
+      vertex 0 0 1.23456788
+      vertex 0 1.23456788 1.23456788
+    endloop
+  endfacet
+  facet normal 0 0 -1.47171605e-07
```
thehans, line 9: Looks like this one still has the old values too. (leading zero on exponent, and some `-0` values further down in the file)

thehans:
I guess I'm still a bit confused about what exactly the ASCII STL tests are checking for, since various discrepancies are still passing.
Should we be doing a full textual diff of actual vs expected stl file output, to better ensure correct output?

tests/data/scad/stl/stl-import-limits.stl
```diff
@@ -0,0 +1,86 @@
+solid OpenSCAD_Model
+  facet normal -1 0 0
+    outer loop
+      vertex 0 0 0
+      vertex 0 0 1.2345679
+      vertex 0 1.2345679 1.2345679
+    endloop
+  endfacet
+  facet normal 0 0 -1.471716e-7
```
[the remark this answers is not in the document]
ochafik: Oh, have now given isZero a zero tolerance to fix this.

tests/regression/stlexport/stl-import-export-expected.stl
ochafik: Sorry it wasn't 😅 (probably still had a bit of fever)

ochafik:
@thehans I've restructured the test so that `stl-import-export-limits.scad` both imports and expects the expectations of `stl-export-limits.scad` (modified the cmake and python test to accept an explicit --expected-file arg), to minimize the changes needed and still benefit from TEST_GENERATE=1

pca006132:
some optimizations are already implemented, remaining should be done with polyset cleanup.

PR #4644, opened 2023-05-19
Speed up hull() w/ better reserve & less alloc

ochafik:
Makes the following model 10% faster to render

```js
$fn=1000;

hull() {
  sphere();
  translate([0.5, 0, 0])
    sphere();
}
```

Also, the code is now simpler (reindexer was already building a vector of points for us)

### DIFF
--- a/src/geometry/cgal/cgalutils-applyops.cc
+++ b/src/geometry/cgal/cgalutils-applyops.cc
@@ -167,55 +167,49 @@ bool applyHull(const Geometry::Geometries& children, PolySet& result)
   using K = CGAL::Epick;
   // Collect point cloud
   Reindexer<K::Point_3> reindexer;
-  std::vector<K::Point_3> points;
-  size_t pointsSaved = 0;
+
+  auto addCapacity = [&](const auto n) {
+    reindexer.reserve(reindexer.size() + n);
+  };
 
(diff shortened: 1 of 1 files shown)


src/geometry/cgal/cgalutils-applyops.cc
```diff
@@ -167,55 +167,49 @@ bool applyHull(const Geometry::Geometries& children, PolySet& result)
...
 #endif
-    } else {
-      const auto *ps = dynamic_cast<const PolySet *>(chgeom.get());
-      if (ps) {
-        points.reserve(points.size() + ps->polygons.size() * 3);
-        for (const auto& p : ps->polygons) {
-          for (const auto& v : p) {
-            addPoint(vector_convert<K::Point_3>(v));
-          }
+    } else if (auto ps = dynamic_pointer_cast<const PolySet>(chgeom)) {
```
kintel, line 202: Tiny nit: You may want to retain `const auto` here, just for clarity for future readers.

ochafik: Heh on second thought, no good reason not to dynamic-cast the raw pointers so skipping the shared_ptr overhead and have used `auto const *x` for each of these declarations. Thanks for the nit with thicker roots than I thought ;-)

Issue #4620, opened 2023-04-24
fast-csg resize no-op


rcolyer:
In the latest master branch code, fast-csg is ignoring resize under numerous conditions.  For example:

```
resize([1,1,1]) { cube(5); translate([6,0,0]) cube(5); }  // Fails.
//resize([1,1,1]) union() { cube(5); translate([6,0,0]) cube(5); }  // Fails.
//resize([1,1,1]) { cube(5); } // Works correctly.
//resize([1,1,1]) render() { cube(5); translate([6,0,0]) cube(5); }  // Works correctly.
```

Correct result without fast-csg:
![image](https://user-images.githubusercontent.com/1266642/233897968-284134fc-1571-40b1-94ac-034388bd00ad.png)
Incorrect result with fast-csg:
![image](https://user-images.githubusercontent.com/1266642/233898041-5a982037-27c5-4771-a3d6-4b795ab62253.png)

PR #4646, opened 2023-05-20
Fix fast-csg's resize

ochafik:
Fixes https://github.com/openscad/openscad/issues/4620

Still unsure what caused the regression at the first place, but this aligns CGALHybridPolyhedron::resize on PolySet's and ManifoldGeometry's

~~Will take out of Draft mode once I've added tests~~
Edit: looks like fast-csg isn't tested much anymore (yay, Manifold), so will skip adding tests for this but still worth going in hopefully.


kintel:
I'm still a bit unsure on what to do about fast-csg. Neither fast-csg or manifold has reached release quality yet, and it kind of seems like Manifold has more traction. We could make a clean cut and remove it, or keep it around as experimental, but if we don't test it well, it's going to bitrot.

It might be a bit premature to kick out right now, but any opinions or insights are welcome.

PS! The main reason I removed fast-csg from the tests is that there were test framework bugs related to which backend was actually used, and removing it made it easier to understand and clean up. With some effort we could re-enabled it, at least for select tests.

PR #4648, opened 2023-05-23
Speed up script evaluation by prehashing & interning identifier strings

ochafik:
This PR substantially speeds up variable bindings & lookups by interning variable names (allows comparing string pointers instead of string contents) and precomputing their hash, using a new `Identifier` class that's built during AST creation and replaces most `std::string` name uses.

Result: a model like [poly_attached_inlined.scad](https://gist.github.com/ochafik/5b622b1075dfb2fbf537805aa925818b) (heavy scripting w/ BOSL2) gets 10% faster preview (measured w/ `-o out.echo`).

<details>
  <summary>Benchmark command-line & results</summary>

```bash
hyperfine --warmup 1 --runs 1 -L build buildMaster,build \
    '{build}/OpenSCAD.app/Contents/MacOS/OpenSCAD poly_attached_inlined.scad -o out.echo'
```

```
Benchmark 1: buildMaster/OpenSCAD.app/Contents/MacOS/OpenSCAD poly_attached_inlined.scad -o out.echo
  Time (abs ≡):         9.463 s               [User: 9.395 s, System: 0.061 s]
 
Benchmark 2: build/OpenSCAD.app/Contents/MacOS/OpenSCAD poly_attached_inlined.scad -o out.echo
  Time (abs ≡):         8.502 s               [User: 8.440 s, System: 0.057 s]
 
Summary
  'build/OpenSCAD.app/Contents/MacOS/OpenSCAD poly_attached_inlined.scad -o out.echo' ran
    1.11 times faster than 'buildMaster/OpenSCAD.app/Contents/MacOS/OpenSCAD poly_attached_inlined.scad -o out.echo'
```
</details>

Notes:
- Removed builtin function lookup cache from this PR (speedup w/ it was ~15%) ~~Some extra speedup is had by caching the resolution of builtin functions inside the `Identifier` class (caching other kinds of values is more complex, w/ unclear payoffs so far).~~
- While interning of variable names strings will grow memory of the app over time, it's expected to be negligible.
- As hinted by @jordanbrown0 below, we could technically pre-resolve variables by keeping an index into the context stack, then a variable slot index within the context. I've tried various versions of that (e.g. [this one](https://github.com/openscad/openscad/commit/8e230ef77107d0534865f3328ab28e67a3478b7c) remembering how many context parent escalations are needed) but it's tricky (same variable not always resolved at the same level, e.g. first iteration of LcForC has a different context layout than subsequent ones) and the end benefit wasn't significant.


jordanbrown0:
[ Without reviewing the implementation. ]

Precalculating the hashes seems entirely plausible.  From what you're saying, it sounds like you're thinking of this as caching, and from a quick glance it looks like you aren't touching the parser.  You could do this precalculation at parse time.

In theory you could resolve lexical-scope (non-$) variable and function references to an offset into a stack, no hashing required.  Or, because of context capture, more likely you'd want to resolve into an index into a stack of contexts and an index into that context.

$ variables are probably require hashing.

Is this worth the complexity?  For most OpenSCAD programs, probably not, because the rule of thumb is that execution time is trivial.  For programs using complex libraries like BOSL2, yeah, probably worthwhile.

ochafik:
@jordanbrown0 Thanks for the rapid feedback! I've updated the poor wording, the code indeed precomputes the hashes when building the `ASTNode`s.

> In theory you could resolve lexical-scope (non-$) variable and function references to an offset into a stack, no hashing required. Or, because of context capture, more likely you'd want to resolve into an index into a stack of contexts and an index into that context.

Ah of course, thanks! I was a bit scared of magical semantics but as you point out it's probably limited to $ vars.

> Is this worth the complexity?

I'll take a look, not sure how much more complex it would need to be.

> For most OpenSCAD programs, probably not, because the rule of thumb is that execution time is trivial. For programs using complex libraries like BOSL2, yeah, probably worthwhile.

Agree for most programs, but we do get repeated complains from people who want faster previews, and I tend to be completely constrained by evaluation time now that I've tasted the sweet BOSL2 fruit (and that I've done all I could on the rendering front, e.g. [integrating manifold](https://github.com/openscad/openscad/pull/4533), [fast-csg](https://github.com/openscad/openscad/pull/4087)).

The model linked from this PR takes 37sec to preview with the PR's change vs. 48sec before... then only 6 extra seconds to render w/ Manifold 🤦‍♂️.

src/core/Context.cc
```diff
@@ -87,27 +90,47 @@ const Value& Context::lookup_variable(const std::string& name, const Location& l
   return *result;
 }
 
-boost::optional<CallableFunction> Context::lookup_function(const std::string& name, const Location& loc) const
+boost::optional<CallableFunction> Context::lookup_function(const Identifier& name, const Location& loc) const
 {
-  if (is_config_variable(name)) {
-    return session()->lookup_special_function(name, loc);
+  if (name.resolved_builtin_function) {
```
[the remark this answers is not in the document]
kintel: Btw, the rest of this PR looks good, modulo test failures, so the quickish way forward could be to split out the function caching into a separate PR to move this one forward.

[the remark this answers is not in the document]
ochafik: @kintel sorry it took me so long to get back to this, fixed tests + removed the function caching bit as suggested

src/core/Context.cc
```diff
@@ -87,24 +87,24 @@ const Value& Context::lookup_variable(const std::string& name, const Location& l
...
 {
-  if (is_config_variable(name)) {
+  if (name.is_config_variable()) {
     return session()->lookup_special_function(name, loc);
   }
   for (const Context *context = this; context != nullptr; context = context->getParent().get()) {
     boost::optional<CallableFunction> result = context->lookup_local_function(name, loc);
     if (result) {
-      return result;
+      return std::move(result);
```
kintel, line 98: I wouldn't worry about this, but just a note to whomever is reading this that at some point we should probably go through our usage of std::move(). In this case, moving shouldn't be needed due to NRVO copy elision, although the effect is likely negligible.

src/core/Identifier.cc
```diff
@@ -0,0 +1,25 @@
...
+    hash = empty_hash;
+    return;
+  }
+
+  is_config_variable_ = name[0] == '$' && name != "$children";
+
+  auto it = interned_names_with_hash.find(name);
+  if (it == interned_names_with_hash.end()) {
+    it = interned_names_with_hash.emplace(name, std::make_pair(name, std::hash<std::string>{}(name))).first;
+  }
```
kintel, line 21: Nit: I _think_ all our supported compilers have C++17 now, so you should be able to this, if appropriate:
`interned_names_with_hash.try_emplace(name, std::make_pair(name, std::hash<std::string>{}(name)))`

src/core/Identifier.h
```diff
@@ -0,0 +1,78 @@
...
+private:
+
+  friend class Context;
+
+  void update(const std::string &name);
+
+  bool is_config_variable_ = false;
+  std::string *interned_name_ptr = NULL;
+  size_t hash = 0;
+  Location loc;
```
kintel, line 62: Another nit: We don't really have a convention for private variables. Old code used no special syntax and prefers prefixing with `this->` on access. Newer code seems to have adopted `_` suffix as popularized by certain style guides. The `_` variant makes code less ambiguous to read, so I'm very slightly tending towards it myself. When modifying existing classes, we should, IMO, try to adhere to one of the two though.

kintel (approved):

kintel:
For some reason, one of the tests seem to crash on Windows. Needs looking into.

src/core/Identifier.h
ochafik: Updated, thanks!

ochafik:
@kintel Hopefully better late than never, refreshed this and tests look green + speed improvement still there.

jordanbrown0:
I said:

> In theory you could resolve lexical-scope (non-$) variable and function references to an offset into a stack, 

It's not that simple, because of recursion.  

```
x = 1;
module foo() {
    if (rands(0,1,1)[0] < 0.5) {
        foo();
    } else {
        echo(x);
    }
}
```
The offset from the top of the stack to x differs depending on how many recursion levels there are.

PR #4653, opened 2023-05-25
Turn simplify_function_body into a virtual method

ochafik:


src/core/Expression.cc
```diff
@@ -567,7 +564,7 @@ Value FunctionCall::evaluate(const std::shared_ptr<const Context>& context) cons
   const Expression *expression = this;
   while (true) {
     try {
-      auto result = simplify_function_body(expression, *expression_context);
+      auto result = expression->simplify_function_body(*expression_context); // || Value::undefined.clone()
```
kintel, line 567: What's the purpose of the commented out code?

src/core/Expression.cc
```diff
@@ -475,77 +475,74 @@ boost::optional<CallableFunction> FunctionCall::evaluate_function_expression(con
...
+  ContextHandle<Context> let_context{Context::create<Context>(context)};
+  let_context->apply_config_variables(*context);
+  return SimplifiedExpression{evaluateStep(let_context), std::move(let_context)};
+}
 
-static SimplificationResult simplify_function_body(const Expression *expression, const std::shared_ptr<const Context>& context)
+SimplificationResult FunctionCall::simplify_function_body(const std::shared_ptr<const Context>& context) const
 {
-  if (!expression) {
+  const auto *call = this;
```
kintel, line 502: Nit: Apart from retaining the old structure, does the call variable make this more readable, or does it makes sense to just use `this`?

kintel (commented):
Minor nits

kintel:
Didn't realize this was still a draft. Looks like a decent refactor to me.

kwikius:
Does this not increase sizeof (Expression) for all expressions

kwikius:
> Does this not increase sizeof (Expression) for all expressions

OK Apologies for the noise. . It shouldnt https://stackoverflow.com/questions/63557630/why-the-size-of-class-remains-same-on-adding-more-than-one-virtual-function

Issue #782, opened 2014-05-06
Library variables recomputed on each module call


OskarLinde:
Here is a simplified synthetic case to illustrate the problem:

**_Snippet 1**_

```
lookup_table = expensive_computation();
function f(x) = interpolate(x, using = lookup_table);
```

**_Snippet 2**_

```
for (i=[1:1000]) echo(f(x));
```

If these two snippets are placed in the same file, everything executes as expected. The expensive_computation() is performed once and the cheap `interpolate` call does the work as intended. Now to the problem: if we instead put Snippet 1 in its own library: `lib.scad`, and call it like this

```
use<lib.scad>
for (i=[1:1000]) echo(f(x));
```

the expensive_computation() call will suddenly be performed 1000 times – causing a dramatic slowdown.

While this could all be as intended (there is a fringe use-case, see below), I think it is more likely an unintended consequence of doing what was easiest – avoiding having to find a way to store the global context of each `use`d file, instead recomputing it at each module instantiation and function call.

Fixing the issue is easy, but I've not attached a patch as there is a remote chance that the behavior is intended, and because there is also a remote chance of this breaking / changing the behavior of some existing code that could be relying on this.

Here is the most likely (although very contrived) theoretical scenario I've come up with that would change behavior by fixing this:

**_lib.scad**_

```
var = $fn * 5;
module m() circle($fn=var);
```

And used:

```
use<lib.scad>
module n() {
    $fn = 7;
    m(); // generates a circle with 35 sides
}
n(); 
```

The fix would make m() generate a circle with 25 sides (default $fn*5).

That behavior is a bit inconsistent, as changing `use` -> `include` already today makes the circle get 25 sides.

Is this behavior intended, and are there any more reasonable use-cases? 
##

adrianVmariano:
I stumbled across this recently and was surprised to learn about this old bug report.  

The run time implications of this bug are quite serious:  in a recent test I improved my code run time from 10.5 minutes to 5 seconds by commenting out high level examples in a library file.

nophead:
Yes I moved some test code from a library module into separate test file that used the library and got a speed up.

rcolyer:
I prepared a testcase to help benchmark and illustrate the behavior of this recomputing issue:

slow.scad:
```
//include <slowused.scad>
use <slowused.scad>

x = [for (i=[0:100000]) let($i=i) foo(i)];
echo(x[0]);
```

slowused.scad:
```
howslow = 25;
[...]

PR #4654, opened 2023-05-25
[Do not merge] Umbrella PR for non-CSG speedup initiatives

ochafik:
Now that 3D CSG operations are super fast (with experimental features [manifold](https://github.com/openscad/openscad/pull/4533) or [fast-csg](https://github.com/openscad/openscad/pull/4087)), the rest of OpenSCAD (script evaluation, import, export) can become the bottleneck on some more ambitious models.

This PR is an umbrella to list, discuss and merge-integrate some active efforts towards non-CSG speedups:

- **fast-stl3**: https://github.com/openscad/openscad/pull/4643
  Faster & more precise STL export using indexed meshes & double-conversion
- **identifier**: https://github.com/openscad/openscad/pull/4648
  Identifier to pre-hash and intern variable lookup strings
- **vector-reserve**: https://github.com/openscad/openscad/pull/4652
- **polyset-reserve**: https://github.com/openscad/openscad/pull/4642
  (both for # of polygons and # of vertices in each polygon)
- **virtual-simplify**: https://github.com/openscad/openscad/pull/4653
- **remove-dupe-vars**: https://github.com/openscad/openscad/pull/4651
  (takes some hashmaps out of hot evaluate codepaths)
- **frame-reserve**: https://github.com/openscad/openscad/pull/4650

Upcoming (not merged yet in this draft PR as not stable yet):

- **[fast-linalg2](https://github.com/ochafik/openscad/compare/master...fast-linalg2?expand=1)** Optimistic Eigen-backed MatrixType instead of VectorType when possible.

Results (WIP) (all assuming `--enable=manifold`):
- [poly_attached_inlined.scad](https://gist.github.com/ochafik/5b622b1075dfb2fbf537805aa925818b) renders 1.37x faster to binstl / 1.43x faster to asciistl
- Test your own? Ideally pick models w/ few CSG operations and high # of facets, importing large assets or that use BOSL2 to process lots of geometry in script


pca006132:
I thought about this for some time, I think it is hard to get decent performance without doing some kind of large refactor, as AST traversal and allocations for stack frames are slow in general. We also don't have jit, which can speed up numerical code by a lot.

I wonder if it will be a good idea to look into translation into lua for example, which has good FFI and a pretty fast jit. V8 will probably be faster, but not sure if the FFI is as simple. By writing the translation rules, we can also fix some unclear semantics issues...

ochafik:
> I thought about this for some time, I think it is hard to get decent performance without doing some kind of large refactor, as AST traversal and allocations for stack frames are slow in general. We also don't have jit, which can speed up numerical code by a lot.

Other possibly tricky avenues of optimizations:
* processing evaluations in parallel (would require tweaking the stack frame code to become threadsafe, and finding the right async / TBB graph way to split computations w/ low overhead - my first forays in that direction haven't been successful, overhead has been disappointing so far, but I have a few more things to try there)
* speed up linear algebra by using SIMD-optimized Eigen matrices instead of vectors of vectors of generic values. Have spent quite a bit of time on this (e.g. [fast-linalg3](https://github.com/openscad/openscad/compare/master...ochafik:openscad:fast-linalg3)), but haven't stabilized it to the point of proving any benefit yet.
* optimise / cache frame referencing (as discussed [in this issue](https://github.com/openscad/openscad/pull/4648#issuecomment-1560089935)), which is tricky in the face of OpenSCAD's current behaviour (where the same code may be run at different frame depths between the first and subsequent iterations of a loop, for instance)

> I wonder if it will be a good idea to look into translation into lua for example, which has good FFI and a pretty fast jit. V8 will probably be faster, but not sure if the FFI is as simple. By writing the translation rules, we can also fix some unclear semantics issues...

Yess!!

Tbh I've been toying with the idea of transpiling OpenSCAD to heavily templated C++ (which in turn, would use parallel constructs, Eigen matrices, etc). Advanced static analysis w/ type propagation / static tracing could bring us quite far.

Definitely worth looking at the semantics of Lua, JS, and general options that have few dependencies (and work well w/ emscripten). I've done a fair bit of V8 native interfacing (codegen), it has a solid API.

My main concern right now is I don't fully understand OpenSCAD's semantics, will need a deep dive into its possibly undocumented fine-print that enables some of the magic leveraged by libs like BOSL2.

pca006132:
There are also issues regarding caching, e.g. #782. I think I can find some time later to hack a simple bytecode interpreter and see how far can we go with something naive.

jordanbrown0:
It seems like it should be possible to fully resolve non-$ variable, function, and module references at parse time.  (But I haven't done any analysis on how much that would save.)

pca006132:
Yes, those are just static scoping. Typically you will save time in computing the hashes and creating the hash tables. These are small gains, but can be pretty large if they are in the hot path (which they are...)

jordanbrown0:
I don't immediately understand why you would need hashes or hash tables for them.  It seems like you could just have pointers directly to the underlying `Value`s, allocated as part of context blocks.

$ variables are entirely different, of course.

pca006132:
Ah, poor wording. What I meant is you will save time *from* computing those hashes. You can use pointers to point to the underlying values, or have a stack of values stored inline (avoid allocation per value) and use indices to those values directly.

PR #4656, opened 2023-06-01
Use gettext functionality from glib instead of directly from gettext

kintel:
This eliminates gettext as a direct run-time dependency, which greatly simplifies dependency management on macOS, and is potentially a small reduction in complexity on other platforms.

Note: We still need the gettext tools whenever we want to re-compile translation files, but that's done occasionally and we can install the gettext tools separately.


kintel:
The jury is still out on this one, since glib is even harder to build correctly than gettext. Also, since replacing glib is still a possibility, perhaps adding more use-cases for it isn't the best idea..

kintel:
Closing this: I'd rather remove glib than gettext, as glib is a bit of a heavy and hard to maintain dependency.

PR #4663, opened 2023-06-10
Bump glib to 2.76.3 for macOS

kintel:
Bump to latest (2.76.3) version of glib.
Includes a patch to make the latest glib build correctly for our build variant. See https://gitlab.gnome.org/GNOME/glib/-/issues/3025 for related discussion.


kintel:
This didn't pan out since glib-2.75.2 has a problem finding our static libintl (since 2.72.0).

Will investigate whether we can continue using 2.71.0, and rather patch it to work with macOS 13.

kintel:
@gordio If you're still around, could you give this a spin?

gordio:
I'm sorry, I can't.
Check please you email inbox.

PR #4670, opened 2023-06-12
Importobj fix

gsohler:
import_obj f statement now understands any number of attributes for its vertices.
Only 1st one is useful though


kintel (approved):

kintel:
Example is a bit bigger than necessary, but it's not really an issue as long as the output is deterministic.

gsohler:
hi Marius,

Example is the exact same file which was reported not to work by J23,
Now it has found its way into the REPO!
Remarkably, this file appears to have bevelled edges, whereas it has 8
sharp corners.
The effect comes from the tweaked normal vertices instead.

PR #4675, opened 2023-06-21
GSoC 2023: Create VertexStateManager

jbinvnt:
`VertexStateManager` is a class for holding common logic used by multiple renderer classes. This reduces duplicate code and is expected to provide a single place where future changes to OpenGL operation will propagate across different types of renderers.

The intended use case for this class is to be a drop-in replacement for logic that currently makes certain rendering functions long and redundant. Therefore `VertexStateManager` objects are initialized with references to stack objects. This means that lifetime of a `vsm` should be limited to the scope of the function in which it is created.

Future refactoring might repurpose `VertexStateManager` into a larger hierarchy of classes. For now though, its goal is to accelerate ongoing investigation of OpenGL modernization by allowing the multiple rendering modes to share individual functions.

This pull request does not introduce any functional changes to the codebase. More PRs will follow for [Google Summer of Code 2023](https://summerofcode.withgoogle.com/programs/2023/projects/87gAMrBX).

### DIFF
--- a/src/glview/VertexStateManager.cc
+++ b/src/glview/VertexStateManager.cc
@@ -1,12 +1,21 @@
 #include "VertexStateManager.h"
 
-void VertexStateManager::initializeSize(size_t vertices_size, bool multiple_vbo, std::vector<GLuint>::iterator current_vbo) {
+void VertexStateManager::initializeSize(size_t vertices_size) {
+  std::vector<GLuint> placeholder = std::vector<GLuint>();
+  size_t placeholder_index = 0;
+  initializeSizeHelper(vertices_size, false, placeholder, placeholder_index);
+}
+
+void VertexStateManager::initializeSize(size_t vertices_size, std::vector<GLuint> & vbos, size_t & vbo_index) {
+  initializeSizeHelper(vertices_size, true, vbos, vbo_index);
+}
+
+void VertexStateManager::initializeSizeHelper(size_t vertices_size, bool multiple_vbo, std::vector<GLuint> & vbos, size_t & vbo_index) {
   if (Feature::ExperimentalVxORenderersDirect.is_enabled() || Feature::ExperimentalVxORenderersPrealloc.is_enabled()) {
     size_t elements_size = 0;
     if (Feature::ExperimentalVxORenderersIndexing.is_enabled()) {
       if (multiple_vbo) {
-        vertex_array->elementsVBO() = *current_vbo;
-        current_vbo = current_vbo.next();
+        vertex_array->elementsVBO() = vbos[vbo_index++];
       }
       if (vertices_size <= 0xff) {
           vertex_array->addElementsData(std::make_shared<AttributeData<GLubyte, 1, GL_UNSIGNED_BYTE>>());
@@ -33,8 +42,7 @@ void VertexStateManager::initializeSize(size_t vertices_size, bool multiple_vbo,
     }
   } else if (Feature::ExperimentalVxORenderersIndexing.is_enabled()) {
     if (multiple_vbo) {
-      vertex_array->elementsVBO() = *current_vbo;
-      current_vbo = current_vbo.next();
+      vertex_array->elementsVBO() = vbos[vbo_index++];
     }
     vertex_array->addElementsData(std::make_shared<AttributeData<GLuint, 1, GL_UNSIGNED_INT>>());
   }
--- a/src/glview/VertexStateManager.cc
+++ b/src/glview/VertexStateManager.cc
@@ -1,0 +1,41 @@
+#include "VertexStateManager.h"
+
+void VertexStateManager::initializeSize(size_t vertices_size, bool multiple_vbo, std::vector<GLuint>::iterator current_vbo) {
+  if (Feature::ExperimentalVxORenderersDirect.is_enabled() || Feature::ExperimentalVxORenderersPrealloc.is_enabled()) {
+    size_t elements_size = 0;
+    if (Feature::ExperimentalVxORenderersIndexing.is_enabled()) {
+      if (multiple_vbo) {
+        vertex_array->elementsVBO() = *current_vbo;
+        current_vbo = current_vbo.next();
+      }
+      if (vertices_size <= 0xff) {
+          vertex_array->addElementsData(std::make_shared<AttributeData<GLubyte, 1, GL_UNSIGNED_BYTE>>());
+      } else if (vertices_size <= 0xffff) {
+          vertex_array->addElementsData(std::make_shared<AttributeData<GLushort, 1, GL_UNSIGNED_SHORT>>());
+      } else {
+          vertex_array->addElementsData(std::make_shared<AttributeData<GLuint, 1, GL_UNSIGNED_INT>>());
+      }
+      elements_size = vertices_size * vertex_array->elements().stride();
+      vertex_array->elementsSize(elements_size);
+    }
+    vertices_size *= vertex_array->stride();
+    vertex_array->verticesSize(vertices_size);
+
+    GL_TRACE("glBindBuffer(GL_ARRAY_BUFFER, %d)", vertex_array->verticesVBO());
+    GL_CHECKD(glBindBuffer(GL_ARRAY_BUFFER, vertex_array->verticesVBO()));
+    GL_TRACE("glBufferData(GL_ARRAY_BUFFER, %d, %p, GL_STATIC_DRAW)", vertices_size % (void *)nullptr);
+    GL_CHECKD(glBufferData(GL_ARRAY_BUFFER, vertices_size, nullptr, GL_STATIC_DRAW));
+    if (Feature::ExperimentalVxORenderersIndexing.is_enabled()) {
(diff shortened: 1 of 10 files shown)


src/glview/VertexStateManager.h
```diff
@@ -0,0 +1,19 @@
...
+#include "VBORenderer.h"
+
+class VertexStateManager {
+public:
+  VertexStateManager(VertexStates* v_s, VertexArray* v_a) : vertex_states(v_s), vertex_array(v_a) {}
+  void initializeSize(size_t vertices_size);
+  void initializeSize(size_t vertices_size, std::vector<GLuint> & vbos, size_t & vbo_index);
+  void addColorState(Color4f color);
+private:
+  VertexStates* vertex_states;
```
kintel, line 16: Since vertex_states appear not to be in use, perhaps remove it until we start using it?

src/glview/VertexStateManager.h
```diff
@@ -0,0 +1,19 @@
...
+  VertexArray* vertex_array;
```
kintel, line 17: Just to avoid having to deal with nullptr, should we make this a reference?

src/glview/VertexStateManager.cc
```diff
@@ -0,0 +1,49 @@
+#include "VertexStateManager.h"
+
+void VertexStateManager::initializeSize(size_t vertices_size) {
+  std::vector<GLuint> placeholder = std::vector<GLuint>();
```
kintel, line 4: Explicit initialization unnecessary as it does the same as default initialization.

src/glview/preview/OpenCSGRenderer.cc
```diff
@@ -165,50 +165,25 @@ void OpenCSGRenderer::createCSGProducts(const CSGProducts& products, const Rende
...
+    for (const auto& csgobj : product.subtractions) {
+      if (csgobj.leaf->geom) {
+        vertices_size += getSurfaceBufferSize(csgobj, highlight_mode, background_mode, OpenSCADOperator::DIFFERENCE);
       }
-    } else if (Feature::ExperimentalVxORenderersIndexing.is_enabled()) {
-      vertex_array.elementsVBO() = all_vbos_[vbo_index++];
-      vertex_array.addElementsData(std::make_shared<AttributeData<GLuint, 1, GL_UNSIGNED_INT>>());
     }
 
+    vsm.initializeSize(vertices_size, all_vbos_, vbo_index);
```
kintel, line 185: To make this code easier to read, would it make sense to instead make this call `vsm.initializeSize(vertices_size, all_vbos_[vbo_index++])`. Letting the caller manage this feels a bit cleaner.

jbinvnt: `vbo_index` is incremented multiple times ([here](https://github.com/openscad/openscad/pull/4675/files#diff-c7bd8e779ecb300b4770befff604ab40bd6912b4622040d2fc4734d33190c12dR18) and [here](https://github.com/openscad/openscad/pull/4675/files#diff-c7bd8e779ecb300b4770befff604ab40bd6912b4622040d2fc4734d33190c12dR45)). In theory, I could split `initializeSize` up into multiple functions to avoid this. But what I'm trying to do now is find large swatches of shared code between renderers (or slightly different code that can be consolidated with the overload-and-helper-method approach that I've proposed here).

kintel: I see now that the increment is also conditional on enabled features. Let's keep it like this and we can revisit API design when the bigger picture starts to become more clear.

src/glview/VertexStateManager.cc
```diff
@@ -0,0 +1,61 @@
...
+      GL_CHECKD(glBufferData(GL_ELEMENT_ARRAY_BUFFER, elements_size, nullptr, GL_STATIC_DRAW));
+    }
+  } else if (Feature::ExperimentalVxORenderersIndexing.is_enabled()) {
+    if (multiple_vbo) {
+      vertex_array->elementsVBO() = vbos[vbo_index++];
+    }
+    vertex_array->addElementsData(std::make_shared<AttributeData<GLuint, 1, GL_UNSIGNED_INT>>());
+  }
+}
+void VertexStateManager::addColor(Color4f last_color) {
```
jbinvnt, line 50: This `addColor` version won't work as a replacement inside `CGALRenderer` once it's time to merge the work done in #4330. That's because it introduces a new attribute for the shader `color_marked`. However, that may change (see the [comment here](https://github.com/openscad/openscad/pull/4330/commits#discussion_r1154840778))

src/glview/VertexStateManager.h
jbinvnt: Now using references

src/glview/VertexStateManager.cc
gsohler: I am interested in the color_marked attribute. When can an user see an item with color_marked ?

jbinvnt: @gsohler The `color_marked` attribute was actually a workaround so that `PolySet` can be used to display the green-colored faces that happen when a `difference()` operation gets used. This `color_marked` attribute does not currently let the user programmatically select a custom set of faces to be rendered with a different color. But maybe you could adapt some of my code from #4330 to allow that type of highlighting behavior if it's what you want.

Note that this PR #4675 is refactoring only and doesn't introduce anything related to `color_marked`. I put that note here because the now-closed #4330 from last year's GSoC interferes with this change, so it is something I'll need to figure out myself later.

src/glview/VertexStateManager.cc
```diff
@@ -0,0 +1,61 @@
...
+      } else {
+          vertex_array.addElementsData(std::make_shared<AttributeData<GLuint, 1, GL_UNSIGNED_INT>>());
+      }
+      elements_size = vertices_size * vertex_array.elements().stride();
+      vertex_array.elementsSize(elements_size);
+    }
+    vertices_size *= vertex_array.stride();
+    vertex_array.verticesSize(vertices_size);
+
+    GL_TRACE("glBindBuffer(GL_ARRAY_BUFFER, %d)", vertex_array.verticesVBO());
```
kintel, line 33: Unless you want to keep these traces around, it might be a good time to kill them. They were really only added for early debugging of the VBO code. The GL_CHECKD below will catch any run-time errors, so it's ok to keep around for now.

Could also be deferred to a later PR - up to you.

kintel (commented):
Looks good once my comments are addressed. I might not be around to approve it, but @t-paul can do a final check.

src/glview/VertexStateManager.cc
```diff
@@ -0,0 +1,61 @@
...
+  std::vector<GLuint> placeholder;
+  size_t placeholder_index = 0;
+  initializeSizeHelper(vertices_size, false, placeholder, placeholder_index);
+}
+
+void VertexStateManager::initializeSize(size_t vertices_size, std::vector<GLuint> & vbos, size_t & vbo_index) {
+  initializeSizeHelper(vertices_size, true, vbos, vbo_index);
+}
+
+void VertexStateManager::initializeSizeHelper(size_t vertices_size, bool multiple_vbo, std::vector<GLuint> & vbos, size_t & vbo_index) {
```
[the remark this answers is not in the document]
jbinvnt: Along with [this other suggestion](https://github.com/openscad/openscad/pull/4675#discussion_r1264014388), I would like to make another (very tiny) PR. My reasoning is that this current set of changes is moving around code, and introducing additional changes even just to names can make it harder to to see in the diffs what got moved.

src/glview/VertexStateManager.cc
```diff
@@ -0,0 +1,61 @@
...
+          vertex_array.addElementsData(std::make_shared<AttributeData<GLubyte, 1, GL_UNSIGNED_BYTE>>());
+      } else if (vertices_size <= 0xffff) {
+          vertex_array.addElementsData(std::make_shared<AttributeData<GLushort, 1, GL_UNSIGNED_SHORT>>());
+      } else {
+          vertex_array.addElementsData(std::make_shared<AttributeData<GLuint, 1, GL_UNSIGNED_INT>>());
+      }
+      elements_size = vertices_size * vertex_array.elements().stride();
+      vertex_array.elementsSize(elements_size);
+    }
+    vertices_size *= vertex_array.stride();
```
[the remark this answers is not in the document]
jbinvnt: [Same](https://github.com/openscad/openscad/pull/4675#discussion_r1265914815)

src/glview/VertexStateManager.cc
jbinvnt: I'll choose to defer it.

src/glview/VertexStateManager.cc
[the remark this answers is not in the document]
kintel: Sounds good!

kintel (approved):

t-paul (approved):

src/glview/VertexStateManager.cc
jbinvnt: #4705

PR #4681, opened 2023-06-26
meson required for MacOS M2 build

thejml:
Might be needed on non-ARM Macs, but my M2 wouldn't build without meson installed.


t-paul:
Yes, correct, needed to build `glib` as dependency. Why did you close it? Adding more info?

thejml:
I only closed it because I got a message saying it was rejected. I can re-open it if need be. But as long as the info is added, we're good!

Issue #4694, opened 2023-07-03
OpenSCAD does not show installed size in the Windows apps control panel


meyersjd:
**Describe the bug**
OpenSCAD does not show its size in the apps control panel of windows

**To Reproduce**
1. Install OpenSCAD
2. Go to the App control panel and find OpenSCAD
3. No app size is listed.

**Expected behavior**
App size should be shown

**Code reproducing the issue**
If applicable, add trimmed down, self-contained code demonstrating the issue.

**Screenshots**
![image](https://github.com/openscad/openscad/assets/80523069/28397727-714f-46f5-920b-29f5864d60bc)

**Environment and Version info (please complete the following information):**
 - OS: Windows 10 and 11
 - System: Intel PC
 - 2021.01 OPENSCAD.org

**Library & Graphics card information**
If applicable, add the content of the Help->Library Info window, or the output of running the `openscad --info` command.

**Additional context**

fkkjohn:
This could help... It's not really a bug as much as an optional parameter:

https://devblogs.microsoft.com/oldnewthing/20040709-00/?p=38493

If I can find where in the code the install file is setting these parameters, I'll fix it.

kintel:
When opening a Pull Request, we will automatically build an installer on CircleCI. That might be the easiest way of testing this.

fkkjohn:
I also tried adding the registry key manually, but I don't think it made a difference on my end. Looks like it was already estimating the install size, because explorer reports 68750kb, but my Add/Remove window shows OpenSCAD with a generic icon and (remove only) as being 67.1Mb.

t-paul:
Dev snapshot shows "Humanity" as publisher

![image](https://github.com/openscad/openscad/assets/1330241/727c43b7-a235-4efa-b3d5-e38ab124cbad)

thehans:
It looks like we were previously writing "Publisher" and other info directly in the registry from installer.nsi here:
https://github.com/openscad/openscad/blob/185c1f59270b7e91764f06b1557b003b66fef38f/scripts/installer.nsi#L25-L32

However, since our windows installers are now created via [CPack NSIS generator](https://cmake.org/cmake/help/latest/cpack_gen/nsis.html) this `.nsi` file and others are no longer in use.  
CPack seems to automatically copy many of these values from various other package settings, though I don't know how many others remain unconfigured or misconfigured.

[...]

PR #4696, opened 2023-07-04
Update installer.nsi Include EstimatedSize and DisplayIcon (Issue #4694)

fkkjohn:
In response to issue #4694, I've made a couple of changes to the install script for windows. It adds registry keys for the estimated install size for the Add and Remove Programs settings page as well as the correct icon for the OpenSCAD entry in the Application list.

### DIFF
--- a/scripts/installer.nsi
+++ b/scripts/installer.nsi
@@ -29,6 +29,8 @@ WriteRegStr HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\OpenSCAD" 
 WriteRegStr HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\OpenSCAD" "URLUpdateInfo" "https://openscad.org/downloads.html"
 WriteRegStr HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\OpenSCAD" "HelpLink" "https://forum.openscad.org/"
 WriteRegStr HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\OpenSCAD" "UninstallString" "$INSTDIR\Uninstall.exe"
+WriteRegStr HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\OpenSCAD" "DisplayIcon" "$INSTDIR\openscad.exe"
+WriteRegStr HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\OpenSCAD" "EstimatedSize" "68750"
 WriteRegStr HKCR ".scad" "PerceivedType" "text"
 SectionEnd
 Section "Uninstall"


kintel:
Installer should show up under branch-build/mxe*/artifacts, e.g. https://app.circleci.com/pipelines/github/openscad/openscad/3854/workflows/aa334a65-7010-4747-948c-cbde0cc6d51e/jobs/15691/artifacts

fkkjohn:
Thanks for the your time and the link.

Well, it didn't work at all. The EstimatedSize registry key is completely missing, and the DisplayIcon is included, but only points the the directory and not the executable, which is what I pointed it to. My guess is the file I edited is no longer being used for the installer. It WAS the correct file, I believe, because it had the exact same registry keys as my installed older version, but it seems very different now. Also, I have a folder in my Start menu named OpenSCAD with the uninstall icon, but there's no main app icon. When I go to the Add/Remove Programs screen, there's OpenSCAD, but it has no icon at all, and no size either. Seems like it needs a lot of work still, and we'll have to find the location of the new installer script. Screen grabs on the way. I'll edit with those in a minute.

kintel:
I didn't work on this, but it could be that CPack is used to generate Windows installers these days: https://github.com/openscad/openscad/blob/master/CMakeLists.txt#L1125
https://cmake.org/cmake/help/latest/module/CPack.html

fkkjohn:
Seems like windows continues to clean things up a little while after install. Maybe it never existed in the folder name I thought it was in. However the Add/Remove programs section looks better now. I'll keep playing with it and see what I can figure out.

openscadinstallinfonew

fkkjohn:
> I didn't work on this, but it could be that CPack is used to generate Windows installers these days: https://github.com/openscad/openscad/blob/master/CMakeLists.txt#L1125 https://cmake.org/cmake/help/latest/module/CPack.html

There does seem to be some registry information in the CMakeLists.txt file: 
```
set(CPACK_NSIS_EXTRA_PREINSTALL_COMMANDS "\
    !include \\\"FileFunc.nsh\\\"\n\
    !include \\\"${CMAKE_SOURCE_DIR}/scripts/mingw-file-association.nsh\\\"\
  ")
  set(CPACK_NSIS_EXTRA_INSTALL_COMMANDS "\
    \\\${RegisterExtension} '$INSTDIR\\\\openscad.exe' '.scad' 'OpenSCAD_File'\n\
    \\\${RefreshShellIcons}\
  ")
  set(CPACK_NSIS_EXTRA_UNINSTALL_COMMANDS "\
    \\\${UnRegisterExtension} '.scad' 'OpenSCAD_File'\n\
    \\\${RefreshShellIcons}\
  ")
  ```

t-paul:
I think that would be the way to add the estimated size: https://nsis.sourceforge.io/Add_uninstall_information_to_Add/Remove_Programs#Computing_EstimatedSize

fkkjohn:
I did some digging... it looks like there should be a template file to help make the CMakeLists.txt file. That template is where we put the regular strings that would be human readable and functional like "WriteRegDWORD HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\${PRODUCT_NAME}" "EstimatedSize" ${ESTIMATED_SIZE}"

Should have a filename like openscad.nsi.in or something with that file extension, I THINK! =) I'll keep digging, but I'm out of time for today.  

OH! and that filename.nsi.in SHOULD be listed somewhere in CMakeLists.txt.

PR #4699, opened 2023-07-09
Python prepare

t-paul:
* Python infrastructure from  #4601 (sqashed)
* Replace git submodule of `crypto++` with cmake find for system installed version

cc @gsohler


src/gui/MainWindow.cc
```diff
@@ -97,6 +97,29 @@
 #include "QSettingsCached.h"
 #include <QSound>
 
+#ifdef ENABLE_PYTHON
+extern std::shared_ptr<AbstractNode> python_result_node;
```
kintel, line 101: One comment for future PRs: With multiple MainWindows, this feels like a challenge. Should these be MainWindow members instead?

t-paul: Right, ideally we don't even need that if the processing switch between scad and python returns the root node.

PR #4703, opened 2023-07-11
Store Points in PolySet as Indexed  Points

gsohler:
The Idea is that the Polygons in PolySet do not contain Vertices but indexes to a vertice array instead.
The adavantage is that a point shared by two polygons can be the SAME index instead of the same coorindate.
So chances  that an object stays water-tight during operations is much better.
This PR is to show  my intent, but I see many places  to optimize  with the  new data structure
(many functions use their local indexed array and still store  polygons as vertices, it much better to maintain a global point array instead)


src/geometry/PolySet.h
```diff
@@ -8,12 +8,20 @@
 
 #include <vector>
 #include <string>
+#include <unordered_map>
+#include <hash.h>
+
+int operator==(const Vector3d &a, const Vector3d &b);
```
kintel, line 14: Why is this operator declared? I'm not seeing an implementation.

src/core/LinearExtrudeNode.cc
```diff
@@ -106,6 +106,7 @@ static std::shared_ptr<AbstractNode> builtin_linear_extrude(const ModuleInstanti
   scaleOK &= parameters["scale"].getFiniteDouble(node->scale_y);
   scaleOK |= parameters["scale"].getVec2(node->scale_x, node->scale_y, true);
   if ((parameters["scale"].isDefined()) && (!scaleOK || !std::isfinite(node->scale_x) || !std::isfinite(node->scale_y))) {
+	  printf("error %f %f\n",node->scale_x, node->scale_y);
```
kintel, line 109: Stray printf?

src/geometry/PolySetUtils.cc
```diff
@@ -129,4 +141,38 @@ bool is_approximately_convex(const PolySet& ps) {
 #endif
 }
 
+
+PolySet  convert_polyset(const shared_ptr<const Geometry>& geom)
+{
+  PolySet ps(3);
+  if (const auto geomlist = dynamic_pointer_cast<const GeometryList>(geom)) {
+    for (const Geometry::GeometryItem& item : geomlist->getChildren()) {
+      return convert_polyset(item.second);
```
kintel, line 150: Not sure I understand the logic: You just return the polygons from the first geometry in the list?

src/geometry/PolySetUtils.cc
```diff
@@ -129,4 +141,38 @@ bool is_approximately_convex(const PolySet& ps) {
...
+    // TODO(ochafik): Implement append_geometry(Surface_mesh) instead of converting to PolySet
+    const PolySet *psp =hybrid->toPolySet().get();
+    return *psp;
+#ifdef ENABLE_MANIFOLD
+  } else if (const auto mani = dynamic_pointer_cast<const ManifoldGeometry>(geom)) {
+    const PolySet *psp = mani->toPolySet().get();
+    return *psp;
+#endif
+  } else if (dynamic_pointer_cast<const Polygon2d>(geom)) { // NOLINT(bugprone-branch-clone)
+    assert(false && "Unsupported file format");
```
kintel, line 171: This error message is a bit misleading; the real issue is that PolySet doesn't support 2D geometry, right?

src/geometry/GeometryUtils.h
```diff
@@ -8,6 +8,7 @@ using Polygons = std::vector<Polygon>;
 
 using IndexedFace = std::vector<int>;
 using IndexedTriangle = Vector3i;
+using PolygonsInd = std::vector<IndexedFace>;
```
kintel, line 11: IMO better to spell out non-standard abbreviations -> PolygonIndices.
Otoh, since this is used only once, perhaps drop the alias and just use `std::vector<IndexedFace>` directly in PolySet?

src/geometry/PolySet.h
```diff
@@ -8,12 +8,20 @@
...
 
 class PolySet : public Geometry
 {
 public:
   VISITABLE_GEOMETRY();
-  Polygons polygons;
+  PolygonsInd polygons_ind;
+  std::vector<Vector3d> points;
```
kintel, line 21: vertices is often a more common name for this.

src/geometry/PolySet.h
```diff
@@ -8,12 +8,20 @@
...
+#include <hash.h>
+
+int operator==(const Vector3d &a, const Vector3d &b);
 
 class PolySet : public Geometry
 {
 public:
   VISITABLE_GEOMETRY();
-  Polygons polygons;
+  PolygonsInd polygons_ind;
```
kintel, line 20: Perhaps just s/polygons_ind/indices/` ?

src/geometry/PolySet.h
```diff
@@ -8,12 +8,20 @@
...
+  std::vector<Vector3d> points;
+  std::unordered_map<Vector3d, int> pointMap;
```
kintel, line 22: This adds unnecessary weight to the PolySet class, and should be part of a `PolySetBuilder` or similar.

src/geometry/PolySet.h
kintel, line 20: When reading this, I wonder if an indexed polygon mesh should ideally differ from an indexed triangle mesh. For rendering, we're going to have to convert to triangle meshes anyway, so perhaps it would be smart and efficient to make that choice earlier.

src/geometry/PolySetUtils.cc
```diff
@@ -129,4 +141,38 @@ bool is_approximately_convex(const PolySet& ps) {
 #endif
 }
 
+
+PolySet  convert_polyset(const shared_ptr<const Geometry>& geom)
+{
+  PolySet ps(3);
```
kintel, line 147: We should ideally find a way of making a better guess at the number of vertices. ochafik@ did some work on that earlier. This kind of looks like a regression.

kintel (commented):
Did a very quick scan. My main concern is that PolySet is becoming heavier. I would recommend starting out by splitting out PolySet _creation_ from PolySet _storage_ into separate components. I think mixing those two was an early design mistake

src/geometry/PolySetUtils.cc
gsohler: yep, this is the error which made my ctest fail

src/geometry/IndexedMesh.cc
kintel, line 1: Why remove the .cc file but keep the .h file?

src/geometry/PolySetBuilder.h
```diff
@@ -0,0 +1,25 @@
...
+#include <vector>
+#include <string>
+#include <unordered_map>
+#include <hash.h>
+class PolySet;
+
+class PolySetBuilder
+{
+  PolySet *ps;
+  std::unordered_map<Vector3d, int> pointMap;
```
kintel, line 19: FYI: Reindexer already solves this for any type, including Vector3d. Might not be super important right now, but could be interesting if we were to add more vertex attributes in the future.

src/geometry/PolySetBuilder.h
```diff
@@ -0,0 +1,25 @@
...
+
+#include <vector>
+#include <string>
+#include <unordered_map>
+#include <hash.h>
+class PolySet;
+
+class PolySetBuilder
+{
+  PolySet *ps;
```
kintel, line 18: This is fine for now, but I believe I've seen places where PolySet where managed as objects, not pointers. If we need non-pointers, we may want to just return a PolySet by value in `result()` and std::move() the arrays into the returned object.

src/geometry/PolySetBuilder.h
```diff
@@ -0,0 +1,25 @@
...
+  std::unordered_map<Vector3d, int> pointMap;
+public:
+  PolySetBuilder(int dim,int vertices_count);
+  int vertexIndex(const Vector3d &coord);
+  void append_poly(const std::vector<int> &inds);
```
kintel, line 23: You may want to add an append_poly() function taking a vector of Vector3d, and manage the indices internally.

gsohler:
Right now I see about 40 ctests not passing, Some of them are segfaults due to wrong memory handling with std::shared_ptr, std::make_shared, std::move, delete etc. Do not completely understand how these functions work and what's the intentional lifecycle of allocated memory in openscad.
These errors hide very likely in PolySetBuilder.cc, PolySetUtils.cc and cgalutils.cc
any help welcome

gsohler:
still the issue is, that this branch has segfaults in windows build  only which cannot be seen when enabling sanitizers.
maybe mingw64 is an option to find the cause of the crashes

gsohler:
superceded by indexed2_sqashed

PR #4705, opened 2023-07-25
Gsoc 2023: Rename vertices_size to num_vertices where appropriate

jbinvnt:
This PR also removes a couple of `GL_TRACE` statements. All these fixes address some feedback that @kintel previously left on #4675 that I decided to make into a separate PR.

### DIFF
--- a/src/glview/VertexStateManager.cc
+++ b/src/glview/VertexStateManager.cc
@@ -1,33 +1,33 @@
 #include "VertexStateManager.h"
 
-void VertexStateManager::initializeSize(size_t vertices_size) {
+void VertexStateManager::initializeSize(size_t num_vertices) {
   std::vector<GLuint> placeholder;
   size_t placeholder_index = 0;
-  initializeSizeHelper(vertices_size, false, placeholder, placeholder_index);
+  initializeSizeHelper(num_vertices, false, placeholder, placeholder_index);
 }
 
-void VertexStateManager::initializeSize(size_t vertices_size, std::vector<GLuint> & vbos, size_t & vbo_index) {
-  initializeSizeHelper(vertices_size, true, vbos, vbo_index);
+void VertexStateManager::initializeSize(size_t num_vertices, std::vector<GLuint> & vbos, size_t & vbo_index) {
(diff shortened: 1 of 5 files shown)


src/glview/VertexStateManager.cc
```diff
@@ -52,9 +48,7 @@ void VertexStateManager::addColor(const Color4f& last_color) {
   Renderer::shaderinfo_t shader_info = renderer.getShader();
   std::shared_ptr<VertexState> color_state = std::make_shared<VBOShaderVertexState>(0, 0, vertex_array.verticesVBO(), vertex_array.elementsVBO());
   color_state->glBegin().emplace_back([shader_info, last_color]() {
```
t-paul, line 50: While we are at renaming, maybe `last_color` could be just `color` here?

src/glview/VertexStateManager.cc
```diff
@@ -1,43 +1,39 @@
...
       }
-      elements_size = vertices_size * vertex_array.elements().stride();
+      elements_size = num_vertices * vertex_array.elements().stride();
       vertex_array.elementsSize(elements_size);
     }
-    vertices_size *= vertex_array.stride();
+    vertices_size = num_vertices * vertex_array.stride();
     vertex_array.verticesSize(vertices_size);
 
-    GL_TRACE("glBindBuffer(GL_ARRAY_BUFFER, %d)", vertex_array.verticesVBO());
```
t-paul, line 33: I'm not sure we really want to remove the trace, are we sure they don't bring useful extra information. I don't like the code duplication but that might be difficult to resolve.
If we decide to keep it, it probably still should move to system-gl.h along with the other macros instead of living in `VertexArray.h`.

jbinvnt: https://github.com/openscad/openscad/pull/4675#discussion_r1264014388

t-paul: Ah, missed that. So remove it is. There's still lots of other places using the macros, so we have to decide on those eventually too.

t-paul (approved):

PR #4709, opened 2023-07-29
Issue 4244 core dump on deleted dir

henriquealrs:


src/openscad.cc
```diff
@@ -732,6 +732,20 @@ void registerDefaultIcon(QString applicationFilePath) {
 void registerDefaultIcon(const QString&) { }
 #endif
 
+static void assertValidPath()
+{
+#ifndef __WIN32
+  boost::system::error_code ec{};
+  auto current_path = fs::current_path(ec);
+
+  while(ec.value() != 0)
```
t-paul, line 741: How safe is it to assume it will always reach a state where it exits the loop?

t-paul (commented):
Thanks for taking on that issue! Just one question, see inline comment.

src/openscad.cc
henriquealrs: In a UNIX like system it should always reach at least `/`. I tested this on my Linux box, and couldn't figure out a scenario where / would not be reached.  I'd like to test on macOS, but right now I have only a  Linux machine available.

On Windows I don't think it is possible ensure reaching an existing directory. Nor am I sure that reproducing the issue is even possible, as I remember, if you have Explorer of terminal in a directory, it is not possible to remove it.

src/openscad.cc
```diff
@@ -732,6 +732,20 @@ void registerDefaultIcon(QString applicationFilePath) {
...
+  {
+      fs::current_path(fs::path(".."));
```
kintel, line 743: Why not `current_path.parent_path()`

src/openscad.cc
```diff
@@ -920,6 +934,8 @@ int main(int argc, char **argv)
   init_mimalloc();
 #endif
 
+  assertValidPath();
```
kintel, line 937: Not sure I understand, is your goal to assert (fail with an error), or silently find the closest valid path?

src/openscad.cc
henriquealrs: I tried it. It doesn't work. any boost::fs operation on a current path nonexistent will fail and throw an exception.

src/openscad.cc
henriquealrs: Yeah, assert is a bat verb, I was thinking of silently finding the closes path

PR #4719, opened 2023-08-08
Make FOV changeable via GUI

rewolff:
With shift-mouse-wheel.


rewolff:
The failed test seems to be completely unrelated to my code, correct ?

t-paul:
Yes, I'm sure it's unrelated. My current suspicion is a bug in the old TBB of Ubuntu 20.04 used by Manifold for multi-threading.

t-paul (approved):
Thanks!

PR #4720, opened 2023-08-12
Update msys2-install-dependencies.sh

gsohler:
libxml2 is needed for pacboy to work.
without libxml2 , pacboy invokes pacman with unknown option "-a"


t-paul (approved):
Thanks! What a strange failure mode.

Issue #4679, opened 2023-06-25
[manifold] Difference of Cube and Complex SVG Results in Incomplete Render


degroof:
Using Manifold, a difference() between a cube and a linear_extrude()-ed complex SVG results in an incomplete render. Preview is correct but render results in incomplete model.

**To Reproduce**
1. Enable Manifold
2. Load and render the attached wodgrain.scad code

**Expected behavior**
Render should look like preview

**Code reproducing the issue**
SCAD code and SVG file included
[manifold-issue.zip](https://github.com/openscad/openscad/files/11859755/manifold-issue.zip)

**Screenshots**
![image](https://github.com/openscad/openscad/assets/24703978/835968fc-65e4-41df-b2a7-b8808b6d7159)

**Environment and Version info:**
 - OS: Windows 10
 - System: Intel PC 64-bit
 - OpenSCAD Version: 2023.06.24 (git 92eec2d43)

**Additional context**
I realize this is a bit of a torture test for the rendering engine. Submitting this mostly in the event that it reveals an issue with the Manifold library.

kintel:
It would be awesome to have an "scad-to-manifold" converter which spits out a pure manifold test case, to make it easier to report such issues upstream.

donovansmith:
I am seeing the same behavior when differencing using the same specs/version and manifold. 

It also occurs with DXFs not just SVGs. 

**Environment and Version info:**
OS: Windows 10
System: Intel PC 64-bit
OpenSCAD Version: 2023.06.24 (git https://github.com/openscad/openscad/commit/92eec2d43843bf4164475b277f96515f9744630f)

pca006132:
There are several bug fixes for manifold, wondering if this issue will still occur with the latest master (manifold).

It is also possible that the issue is caused by the mesh extruded from svg, not sure about that. For submitting an issue to upstream, you can just export the model for the svg extruded part (in gltf for example, as it preserves manifoldness), it should be enough for us to reproduce it.

degroof:
Just tried it with OpenSCAD version 2023.07.09 (git 44fdb2c46), and am getting similar results. Not sure how to export to gltf, but here's an STL file of the extruded SVG, which gives basically the same results when subtracted from a 145x100x1 cube.
[woodgrain.zip](https://github.com/openscad/openscad/files/12113817/woodgrain.zip)

pca006132:
Reproduced now, thanks!

alidaf:
[...]

PR #4728, opened 2023-08-28
Bump manifold to latest version (v2.2.1-3-gb6eea01).

t-paul:


pca006132:
is the failure related to how manifold reorders boolean operands?

t-paul:
@pca006132 I would assume so, yes. But we already have a feature that sorts faces and vertices at export to produce a deterministic output, it's just not enabled by default for STL.

I'm planning of just removing the experimental flag from that feature so we always get the sorted STL output. I think that's nothing Manifold should take care of, at least not always as there's probably a good number of use cases where the sorting order does not matter and impacting those performance wise seems not ideal.

pca006132:
FYI: discussion about deterministic results in https://github.com/elalish/manifold/issues/545

pca006132:
In particular, there should be no guarantee in how we triangulate the result, other than the triangulation should be valid.
I guess this may make testing harder as you can no longer compare the edges of the mesh.

pca006132:
Wondering what is the blocker for this PR? Is it some issue in manifold?

pca006132:
the triangulator refactor is merged, maybe we can bump the version again and get this merged?

pca006132:
For the manifold-stlexport_stl-export it is expected due to changes in the triangulation. I have no idea about those png failures, will investigate them later.

pca006132:
I am unable to reproduce the png error when using the latest master with updated manifold submodule. Maybe you can try to update this and see if it still fails? I think there are a few patches related to OpenGL recently.

t-paul:
@pca006132 thanks a lot, and I'm sorry for not having time to really look into the issues yet. I'm assuming the test case issues are on OpenSCAD side. @kintel is also currently fixing some inconsitencies in the test suite that have crept in via all the new feature.

pca006132:
Sure, I saw the PR by kintel. It looks interesting that the png tests still failed after updating to the master. Maybe I should try if I can reproduce the issue with docker.

pca006132:
I tried merging with master (after @kintel's recent PRs), and I only get the following failures now:
```
	1273 - cgalstlcgalpngtest_bad-stl-tardis (Failed)
	1292 - cgalbinstlcgalpngtest_bad-stl-tardis (Failed)
```

both got the error `ERROR: The given mesh is not closed! Unable to convert to CGAL_Nef_Polyhedron.`. And if I understand the names correctly it seems that the test is using bad input files?

kintel:
I feel like think the tartis issue is a completely different bug. If you look at the test report:
```
/home/runner/work/openscad/openscad/b/openscad /home/runner/work/openscad/openscad/tests/data/scad/misc/bad-stl-tardis.scad -o /home/runner/work/openscad/openscad/b/tests/output/cgalstlcgalpngtest/bad-stl-tardis.scad.stl --enable=fast-csg --render=cgal --export-format asciistl
Converted to Nef polyhedron
Geometries in cache: 2
Geometry cache size in bytes: 523888
CGAL Polyhedrons in cache: 0
CGAL cache size in bytes: 0
Total rendering time: 0:00:00.273
   Simple:     yes
   Vertices:      810
   Halfedges:    2880
   Edges:        1440
   Halffacets:   1262
   Facets:        631
   Volumes:         2
```


-> Why is it using CGAL for the geometry when manifold is enabled?

pca006132:
Not sure about the internals, I think it currently just loads the stl as a CGAL object, and lazily convert it into a manifold object when needed (which is not needed in the test because there is no CSG operations).

kintel:
I did some digging: Surprisingly, the two tests pass _without_ `--enable=predictible-output`, and fail _with_ `--enable=predictible-output`. Since this PR makes the predictible output default, it fails.

@t-paul Any ideas?

pca006132:
Can we merge with master to see if the CI can pass (except the known failures that are unrelated to manifold). This manifold update should fix #4679, #4766, #4769 which I think are the major issues people get when using manifold.

t-paul:
It was already rebased, but I've dropped the predictible-output for now as it seems we may not depend on it.

t-paul:
It's green, so lets go :snail: :grin:. Thank you @pca006132!

kintel:
Since this seems to be causing crashes on Windows, I wonder if we should revert this PR until we can figure that out.

t-paul:
I'll have a look, maybe just the Manifold update instead off full revert. That way we could rule out the other changes.

kintel:
Yeah, rolling back Manifold to some safe version would be good for now, as it seems to be tricky to reproduce the current issue.

pca006132:
I am thinking if the crashes are actually caused by manifold. The examples are pretty simple, with no tricky 2D geometry which may stress our new triangulator. No one managed to reproduce the crash with msys2/macos/linux, or a clean build with MXE. The crash is also very weird as we normally won't segfault directly, we usually just throw an index out of range exception. And I tried using debug build and enabled address sanitizer for manifold on Windows, which we found no memory corruption, indicating that it is quite unlikely to have such severe bug for simple model like the `logo.scad`.

This kinds of make me wonder if this is something one-off like the binary getting corrupted by random-hardware-fault or some issue with MXE. The first hypothesis can be ruled out if we managed to reproduce the crash with a clean MXE build, because the probability with 2 consecutive corruption which causes similar error is rare. For the second one we can try to reproduce it with something other than MXE...

PR #4738, opened 2023-09-10
GSoC 2023: Macro to allow building with OpenGL Core

jbinvnt:
If compiled with `-DDISABLE_FIXEDFUNCTION_GL`, all API calls to legacy OpenGL functions are removed. Otherwise this PR is a no-op. The mxe-64 test is failing because so is the current master branch.


kintel:
One note: This PR was created by generating bindings where only OpenGL 3.2 core profile entry points are available. However, modern OpenGL is defined by not only entry point but also how these entry points are used.
Example: `glGetIntegerv(GL_RED_BITS, ...` isn't available in OpenGL 3.2 core, but wasn't removed by this PR: https://github.com/openscad/openscad/blob/c3c4a658bc6421f3ed9d866688119f263270882a/src/glview/system-gl.cc#L67

There are numerous similar examples in the codebase. Finding them all may be quite some work, and I'm not convinced that partially removing legacy calls is worth merging to master, especially without a more concrete plan for how to proceed.

I'll think and tinker with this over the next few months and see if I can come up with a plan.

jbinvnt:
@kintel can you let me know if there is a more intuitive source I can look at for listing these additional removed features? My general guide has been the [instructions from the wiki](https://www.khronos.org/opengl/wiki/Legacy_OpenGL#Removed_functionality) but a quick search there doesn't return anything for `GL_RED_BITS` (or even `.*_BITS`).

I'm also curious why the compiler even let this build with Core Profile headers if I missed some legacy calls.

kintel:
You weren't missing legacy calls, but legacy _argument_ to calls.

The OpenGL 3.2 core profile spec, appendix E.2.2 is a great start: https://registry.khronos.org/OpenGL/specs/gl/glspec32.core.pdf

PR #4768, opened 2023-10-06
add unity build support

pca006132:
Unity build can speed up compilation by grouping several source files into one. This adds unity build support by setting `SKIP_UNITY_BUILD_INCLUSION` for files that does not support unity build due to symbol naming conflict.

This speeds up compilation pretty significantly. On my machine with 20 logical cores using clang 11.1.0, mold 1.11.0 and ninja 1.11.1, the compilation time went down from `2906.50s user 78.91s system 1760% cpu 2:49.57 total` to `1232.89s user 30.07s system 936% cpu 2:14.83 total`. This may be significant for CI.

To use unity build, just add `-DCMAKE_UNITY_BUILD=ON` to the cmake arguments.


gsohler:
Would be  great if  compilation time could really be speed up. I am wondering if CI Builds will pass.
How did you select  the files/submodules to skip Unity build ?

pca006132:
I did not change the CI build settings, so it should not be using unity build and should not have any effect for now.

For skipping unity build, it is sort of like trial-and-error. I know that manifold does not support this, and I saw some error messages indicating gl.h header probably doesn't work well with unity build, etc.

t-paul:
Yes, I think all 3 (different!) build failures are unrelated. MXE 64-bit is the issue you linked. Ubuntu 20.04 is likely the old thrust/tbb library making things crash and the macOS test has trouble with the cairo 1.18 upgrade. CI is awesome...

t-paul:
Trying this locally, I'm getting various build failures like the included OpenCSG and disabling that conflicts with X11 header (#)defining a lot of generic names like `Bool` and `Status`.

pca006132:
yeah I found the opengl related code incompatible with unity build, so I tried to exclude them. maybe need to exclude more

thehans:
@teepee I was getting same errors and found that it was from `spnav.h` which includes `Xlib.h`

In real time, its actually a few seconds slower for me, but i guess a significantly less load on CPU cores going by user and sys times.

No unity build:
```
real    1m43.485s
user    30m27.255s
sys     2m44.975s
```
With unity build:
```
real    1m46.305s
user    14m54.172s
sys     0m58.558s
```

There's still a macro redefinition by libtess which seems gross, but I guess its not technically an error:
```
/home/hans/projects/openscad_unity/src/ext/libtess2/Source/priorityq.c:337: warning: "Swap" redefined                                                                                          
  337 | #define Swap(a,b)   do {PQkey *tmp = *a; *a = *b; *b = tmp;} while(0)                                                                                                                  
In file included from /home/hans/projects/openscad_unity/build/CMakeFiles/OpenSCAD.dir/Unity/unity_0_c.c:7:                                                                                    
/home/hans/projects/openscad_unity/src/ext/libtess2/Source/geom.c:199: note: this is the location of the previous definition                                                                   
  199 | #define Swap(a,b)       do { TESSvertex *t = a; a = b; b = t; } while(0)                                                                                                               
```

pca006132:
to summarize:

1. Unity build cuts down the linux and macos build time by half (linux: from 22mins to 12mins, macos: 14mins to 6mins). The timing on macos is not very stable though, perhaps due to different runners...
2. Cut down the homebrew install time by half (10mins to 6mins)

It is not enabled on circleci and windows for now. msys2 can build using unity build, but someone it causes test failures without.

kintel:
Local macOS test: Build time is slashed in half, from ~100s to under 50s, both with ninja and make -j10

kintel:
Re-running Windows CI to see if the error changes...

CMakeLists.txt
```diff
@@ -1076,6 +1087,30 @@ if (ENABLE_CGAL)
...
+  src/gui/input/SpaceNavInputDriver.cc
+  src/gui/QGLView.cc
+  src/gui/QGLView2.cc
+  # things that requires _USE_MATH_DEFINES
+  src/core/BuiltinContext.cc
+  src/geometry/roof_vd.cc
+  src/io/DxfData.cc
+  src/utils/calc.cc
+  src/utils/degree_trig.cc
+  PROPERTIES SKIP_UNITY_BUILD_INCLUSION ON)
```
kintel, line 1112: List list of exclusions is hard to read and hard to maintain. We'd probably want some reasoning per file for why it needs to be excluded, to make it possible to add to or remove from this list in the future.

CMakeLists.txt
```diff
@@ -1076,6 +1087,30 @@ if (ENABLE_CGAL)
...
+  ${MANIFOLD_SOURCES}
+  src/glview/OffscreenContextFactory.cc
+  src/glview/RenderSettings.cc
+  src/glview/Camera.cc
+  src/glview/ColorMap.cc
+  src/glview/preview/CSGTreeNormalizer.cc
+  src/gui/input/SpaceNavInputDriver.cc
+  src/gui/QGLView.cc
+  src/gui/QGLView2.cc
+  # things that requires _USE_MATH_DEFINES
```
kintel, line 1106: I do wonder if it's worthwhile porting off math defines, e.g. to boost constants (https://www.boost.org/doc/libs/1_37_0/libs/math/doc/sf_and_dist/html/math_toolkit/toolkit/internals1/constants.html), or collect in our own header while we wait for C++20.

pca006132:
@kintel this is now updated to address you comment.

CMakeLists.txt
```diff
@@ -1076,6 +1087,37 @@ if (ENABLE_CGAL)
...
+  # files using opengl does not work well with unity build, again due to the
+  # macros
+  ${OFFSCREEN_SOURCES}
+  ${PLATFORM_SOURCES}
+  src/glview/OffscreenContextFactory.cc
+  src/glview/RenderSettings.cc
+  src/glview/Camera.cc
+  src/glview/ColorMap.cc
+  src/glview/preview/CSGTreeNormalizer.cc
+  src/gui/input/SpaceNavInputDriver.cc
```
kintel, line 1109: Did this file sneak in by mistake?

kintel (approved):

CMakeLists.txt
pca006132: no, see https://github.com/openscad/openscad/pull/4768#issuecomment-1754311120

but yeah I should document this as well

pca006132:
I tested on a windows VM and found some cgal tests crashed with OOM, not sure if it is related to the CI failing with unity build enabled.

Issue #4782, opened 2023-10-14
Move VBO support out of experimental


kintel:
The purpose of this ticket is to track moving VBO support, first merged in #1935, out of experimental.

TODO:
* [x] Investigate macOS performance issues (#4504)
* [x] Fix macOS performance issues: #4504
* [x] Investigate performance issue #4883 
* [x] Investigate lack of indexing support for Nef Polyhedrons
   * Not only is there no indexing support, we render using practically one glDrawArrays() call per triangle.
   * This is probably because gluTess by default outputs various primitive types. If we force gluTess to output only triangles, we might be able to combine the data into a single array as a start, and then later manage indexes.
* [x] The VBO support consists of 4 experimental flags: `vertex-object-renderers`, `vertex-object-renderers-prealloc`, `vertex-object-renderers-direct`, `vertex-object-renderers-indexing`.
   * [x] Reduce to two flags: VBO + indexed
   * [x] Ideally, reduced to a single flag; make a decision about indexed: We should keep indexed rendering as an experimental option for now.
* [x] Go through tests and verify that various combinations of render mode and view options are working.
* [x] Investigate what it would take to make the VBO support conform to the OpenGL core profile (e.g. VAO support).
   * -> Too much work right now. Address this as part of general GL cleanup.
* [x] Investigate RAM usage: #4858
* [x] Remove `vertex-object-renderers` experimental flags, keep `vertex-object-renderers-indexing`: #4966

For further opportunities, see #4968

PR #4789, opened 2023-10-24
Move VBO rendering out of experimental mode

kintel:
This makes VBO rendering non-experimental.

Indexed rendering is still kept as an experimental feature until we perform more testing. The main question is whether indexed or non-indexed rendering will misbehave in any common scenarios. Misbehavior is expected to be limited to performance issues.

This also kills legacy code paths. We could discuss whether we should keep them behind an `#ifdef`, in case it becomes necessary to test agains the old code.

Related to #4782


kintel:
After #4914, we can rewrite this as a much smaller patch

PR #4792, opened 2023-10-25
Document asciistl and binstl in the man page.

ChickenProp:
Text copied from `openscad --help` which already documents them.


t-paul (approved):
Nice, thank you!

PR #4822, opened 2023-11-12
Rename --render=cgal to --render=force to force-convert to the current backend-specific geometry

kintel:
`--render=cgal` was really meant for converting to the internal format of the current geometry engine to utilize any of its built-in mesh validation and repair tools. When using another geometry backend, we're changing this to `--render=force` to force conversion into the current backend-specific geometry.

`--render=cgal` is still useable for backwards compatibility, but not sure for how much longer we need to support it.

Also added some opportunistic refactoring.

### DIFF
--- a/src/geometry/GeometryUtils.cc
+++ b/src/geometry/GeometryUtils.cc
@@ -514,7 +514,7 @@ std::shared_ptr<const Geometry> GeometryUtils::getBackendSpecificGeometry(const 
 #if ENABLE_MANIFOLD
   if (Feature::ExperimentalManifold.is_enabled()) {
     if (const auto ps = std::dynamic_pointer_cast<const PolySet>(geom)) {
-      return ManifoldUtils::createMutableManifoldFromPolySet(*ps);
+      return ManifoldUtils::createManifoldFromPolySet(*ps);
     } else if (auto mani = std::dynamic_pointer_cast<const ManifoldGeometry>(geom)) {
       return geom;
     } else {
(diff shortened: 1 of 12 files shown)


jordanbrown0:
Yes, you should.  "--render=cgal" should use CGAL, or be an error if CGAL is not available.  Otherwise it's confusing.

jordanbrown0:
That is, you should rename it.

kintel:
`--render=cgal` is also slightly confusing as both the regular backend and fast-csg are CGAL.

I believe the initial purpose of this flag was:
* `--render`: Force concrete geometry to be rendered. If the .scad file only contains concrete geometry, e.g. only consists of a primitive, an `import()`, a basic linear extrude etc., it will use that geometry as is, i.e. a `PolySet`, otherwise, the boolean engine will be invoked and produce whatever geometry that backend provides (Nef3 polyhedrons for the release build, but could be ManifoldGeometry).
* `--render=cgal`: If the resulting geometry from evaluating the file was a `PolySet`, convert it to CGAL Nef3 polyhedrons, then back to `PolySet`. This ensures that the geometry passes all checks for valid polyhedrons in CGAL, or allows issues to be triggered instantly.

Ideally, we should rename `--render=cgal` to something more generic which doesn't involve naming the backend.

If we do ship with multiple backends (either for a release build or as experimental), having a flag to select backend would make sense, but that could be separate from `--render=`.

kintel:
Update: Renamed `--render=cgal` to `--render=force`

Issue #4815, opened 2023-11-09
Windows 2023.11.08 crash on salt pig lid model with manifold enabled


owebeeone:
**Describe the bug**
Attempting to render the model linked below crashes or wedges Openscad v Windows 2023.11.08

https://drive.google.com/file/d/1qmgVoY0NuFzueMCYkebTCH9kjReEhAtO/view?usp=sharing

**To Reproduce**
1. Load the model linked
2. Render

**Expected behavior**
Create a model

**Code reproducing the issue**
(https://drive.google.com/file/d/1qmgVoY0NuFzueMCYkebTCH9kjReEhAtO/view?usp=sharing)

**Environment and Version info (please complete the following information):**
 - OS: Windows 11
 - System Model	ROG Strix G513RM_G513RM
 - Installed Physical Memory (RAM)	32.0 GB
 - OpenSCAD Version [2023.11.08 ] installed exe installer

**Library & Graphics card information**
OpenSCAD Version: 2023.11.08 (git e7abd10a9)
System information: Microsoft Windows 11 (10.0.22621) x86_64 16 CPUs 31.26 GB RAM
User Agent: OpenSCAD/2023.11.08 (git e7abd10a9) (Microsoft Windows 11 (10.0.22621) x86_64)
Compiler: GCC "11.3.0" 64bit
MinGW build: MingW64
Debug build: No
Boost version: 1_81
Eigen version: 3.4.0
CGAL version, kernels: 5.5, Cartesian<Gmpq>, Extended_cartesian<Gmpq>, Epeck
OpenCSG version: OpenCSG 1.5.0
Qt version: 5.15.10
QScintilla version: 2.11.2
InputDrivers: 
GLib version: 2.70.2
lodepng version: 20210627
libzip version: 1.5.2
fontconfig version: 2.14.2
freetype version: 2.13.1
harfbuzz version: 7.3.0
cairo version: 1.16.0
lib3mf version: 1.8.1
Features: fast-csg, fast-csg-safer, fast-csg-debug, manifold*, roof, input-driver-dbus, lazy-union*, vertex-object-renderers, vertex-object-renderers-indexing, vertex-object-renderers-direct, vertex-object-renderers-prealloc, textmetrics, import-function, predictible-output
Application Path: C:/Program Files/OpenscadDev
Documents Path: C:\Users\gianni\Documents
User Documents Path: C:\Users\gianni\Documents
Resource Path: C:/Program Files/OpenscadDev
[...]

PR #4830, opened 2023-11-15
use newer tbb for mxe build

### LINKED ISSUE #4814: Windows: 2023.11.08 Build Crashes When Using Manifold
**Describe the bug**
Windows 2023.11.08 nightly build crashes or hangs with manifold enabled upon render or preview (dependent on file). 

2023.11.05 build does not crash. Linux 2023.11.08 AppImage build is also OK. 

**To Reproduce**
1. Enable Manifold
2. Restart OpenSCAD GUI
3. Open Examples > Basics > logo.scad
4. May crash on initial preview
5. Click Render
6. Will crash with no notification if it hasn't already

**Expected behavior**
OpenSCAD completes the preview or render.

Use Built In Example File: Basics > logo.scad

**Screenshots**
openscad_error

**Environment and Version info:**
 - OS: Windows 10, Windows 11
 - System: Intel PC 64-bit
 - OpenSCAD Version 2023.11.08 snapshot, exe and zip.

OpenSCAD Version: 2023.11.08 (git e7abd10a9)
System information: Microsoft Windows 10 (10.0.19045) x86_64 16 CPUs 127.27 GB RAM
User Agent: OpenSCAD/2023.11.08 (git e7abd10a9) (Microsoft Windows 10 (10.0.19045) x86_64)
Compiler: GCC "11.3.0" 64bit
MinGW build: MingW64
Debug build: No
Boost version: 1_81
Eigen version: 3.4.0
CGAL version, kernels: 5.5, Cartesian<Gmpq>, Extended_cartesian<Gmpq>, Epeck
OpenCSG version: OpenCSG 1.5.0
Qt version: 5.15.10
QScintilla version: 2.11.2
InputDrivers:
GLib version: 2.70.2
lodepng version: 20210627
libzip version: 1.5.2
fontconfig version: 2.14.2
freetype version: 2.13.1
harfbuzz version: 7.3.0
cairo version: 1.16.0
lib3mf version: 1.8.1
Features: fast-csg, fast-csg-safer, fast-csg-debug, manifold, roof, input-driver-dbus, lazy-union, vertex-object-renderers, vertex-object-renderers-indexing, vertex-object-renderers-direct, vertex-object-renderers-prealloc, textmetrics, import-function, predictible-output
GLEW version: 2.1.0
OpenGL Version: 4.6.0 NVIDIA 537.42
GL Renderer: NVIDIA RTX A5000 Laptop GPU/PCIe/SSE2
GL Vendor: NVIDIA Corporation
RGBA(8888), depth(24), stencil(8)
GL_ARB_framebuffer_object: yes
GL_EXT_framebuffer_object: yes
GL_EXT_packed_depth_stencil: yes

**Additional context**
Appears to have occurred after merges on 2023.11.07 — specifically #4728?

The exception codes I've noted in Event Viewer "Application Error" and ".NET Runtime" entries have been:
1. `0xc0000374`
2. `0xc0000005`

I've tried:
1. EXE and Zip versions
2. Three different machines Win 10 (Phys), Win 10, (Phys) Win 11 (VM) (Two have never had OpenSCAD use)
3. Wiping `HKEY_CURRENT_USER\SOFTWARE\OpenSCAD` settings in registry (For machine w/ previous OpenSCAD use)
4. Enabling/Disabling `vertex-*` entries
5. `fast-csg `options were not checked

pca006132:
Closes #4814 #4815.

For some reason tbb is linked differently after manifold update, and causes segfault/infinite loop on Windows for MXE builds. The build log shows
```
/mxe/usr/x86_64-w64-mingw32.static.posix/lib/libtbb_static.a(task.cpp.obj): duplicate section `.rdata$_ZTIN3tbb4taskE[_ZTIN3tbb4taskE]' has different size
/mxe/usr/x86_64-w64-mingw32.static.posix/lib/libtbb_static.a(scheduler.cpp.obj): duplicate section `.rdata$_ZTIN3tbb4taskE[_ZTIN3tbb4taskE]' has different size
/mxe/usr/x86_64-w64-mingw32.static.posix/lib/libtbb_static.a(arena.cpp.obj): duplicate section `.rdata$_ZTIN3tbb4taskE[_ZTIN3tbb4taskE]' has different size
```

This patch will patch manifold to ignore the system tbb package, and use a newer version of tbb. As the latest version is not working on MXE due to compiler internal error, I chose v2021.8.0 which does work. Not sure if the bug is caused by tbb being too old, or just some compiler options. Anyway I think there is no harm in updating the tbb version, considering the MXE one is already 5+ years old.

The build is working on Virtual Box but I have yet tried it on real windows machine. Build artifacts: https://app.circleci.com/pipelines/github/openscad/openscad/4216/workflows/f892a61f-bae6-47e6-a428-b773a12067c6/jobs/17006/artifacts

@donovansmith @owebeeone maybe you can give this a try.

### DIFF
--- a/scripts/release-common.sh
+++ b/scripts/release-common.sh
@@ -171,6 +171,45 @@ echo "Checking pre-requisites..."
 
 git submodule update --init --recursive
 
+if [ "$OS" = "UNIX_CROSS_WIN" ]; then
+  cd submodules/manifold
+  echo "Patching manifold to use newer TBB version"
+  # The MXE provided TBB is too old and somehow causes linker problem.
+  # The latest version cannot compile on MXE due to compiler internal error,
+  # so we use v2021.8.0.
+  git apply << EOF
+diff --git a/src/utilities/CMakeLists.txt b/src/utilities/CMakeLists.txt
+index b1cf20a..235fb8c 100644
+--- a/src/utilities/CMakeLists.txt
++++ b/src/utilities/CMakeLists.txt
+@@ -52,10 +52,10 @@ if (TRACY_ENABLE)
+ endif()
+
+ if(MANIFOLD_PAR STREQUAL "TBB")
+-    find_package(PkgConfig)
+-    if (PKG_CONFIG_FOUND)
+-        pkg_check_modules(TBB tbb)
+-    endif()
++       #find_package(PkgConfig)
++       #if (PKG_CONFIG_FOUND)
++       #    pkg_check_modules(TBB tbb)
++       #endif()
+     if(NOT TBB_FOUND)
+         message(STATUS "tbb not found, downloading from source")
+         include(FetchContent)
+@@ -63,7 +63,7 @@ if(MANIFOLD_PAR STREQUAL "TBB")
+         set(TBB_STRICT OFF CACHE INTERNAL "" FORCE)
+         FetchContent_Declare(TBB
+             GIT_REPOSITORY https://github.com/oneapi-src/oneTBB.git
+-            GIT_TAG        v2021.10.0
++            GIT_TAG        v2021.8.0
+             GIT_SHALLOW    TRUE
+             GIT_PROGRESS   TRUE
+         )
+
+EOF
+fi
+
 echo "Building openscad-$VERSION ($VERSIONDATE)"
 echo "  CMake args: $CMAKE_CONFIG"
 echo "  DEPLOYDIR: " $DEPLOYDIR
@@ -275,7 +314,7 @@ if [ -n $FONTDIR ]; then
   cp -a fonts/10-liberation.conf $FONTDIR
   cp -a fonts/Liberation-2.00.1 $FONTDIR
   case $OS in
-    MACOSX) 
+    MACOSX)
       cp -a fonts/05-osx-fonts.conf $FONTDIR
       cp -a fonts-osx/* $FONTDIR
       ;;


pca006132:
I think the CI errors are not related to this PR.

owebeeone:
Salt pig lid model and all other anchorscad models work fine with Openscad vers 2023, 11, 15, 18, 23, 34

thehans:
Just a couple thoughts about moving forward with this.

The CI builds are based on a docker image, which installs mxe packages based on this source file:
https://github.com/openscad/docker-openscad/blob/main/mxe/packages.lst
`intel-tbb` can be seen in the packages listed there

I was just working on some small changes to get our build scripts to install and use `intel-tbb` outside of docker as well, for anyone attempting a local MXE build.

However after seeing that you are bypassing that package with this patch, and that apparently the entire project was renamed to `oneTBB` since MXE last updated, I guess those changes are irrelevant.

Without the MXE package installed directly, my local build was attempting to download and compile tbb itself (i didn't catch which version), but that was failling for me with some errors about converting `char*` to `wchar_t*`, so I thought the MXE package was the way to go at the time.

OpenSCAD is currently maintaining our own mxe fork, and the build for intel-tbb [can be seen here](https://github.com/openscad/mxe/blob/openscad-snapshot-build_2023-09-12/src/intel-tbb.mk) (though that package is still same as upstream mxe at the moment).
I am thinking we should probably look into adding `oneTBB` into our mxe fork, as its own package to replace `intel-tbb`, then eventually try to get that into upstream?

I don't know if this is PR is intended to be merged directly, or more of a quick test/stopgap, but it feels rather messy to include patches like this in our bash scripts, particularly coming from a contributor to the project being patched.  I would hope this could be addressed in manifold project proper, as a more robust solution.  If pkgconfig usage will remain then it should probably have some added check for valid version ranges.

pca006132:
This PR is more like a quick test, an effort to avoid reverting the manifold update and make sure it is not our fault :). The best way would be to update MXE, considering openscad maintains its own fork it should be easier than trying to get this upstream (MXE). It is hard to address this in manifold, mainly due to

1. We don't really know why the mxe build fails. It can be a version issue, or compiler arguments being incompatible. In case it is the latter issue, compiling from source is just the simplest way to make sure the compiler arguments are compatible, so I did not bother to dig deeper. If we want to check the supported version in manifold, we have to do a lot more to test on different platforms...
2. Overriding system package is not always doable. I am not an expert in CMake, but I think if a package is installed in `/usr/include`, you are basically out of luck because you can't really remove this directory from the header search without a lot of work. For some systems such as NixOS, they inject installed package path into compiler flags directly, bypassing whatever `find_package` stuff you use.
So, even if we implement version check in manifold, we can only say that your tbb version is incompatible. It is platform specific to how to fix the issue, either update the package, remove the package, or give some flags to disable the `find_package` and use a version known to be working (but this flag itself may not work due to reasons mentioned above).

pca006132:
We can also merge this first, and open an issue about the mxe problem so we can fix it later. I think some Windows users rely on the nightly manifold feature and it would be better if we can let them use the latest nightly.

pca006132:
@thehans I can look into adding `oneTBB` into OpenSCAD MXE fork in a few days.

t-paul:
Using `gcc-13` plugin, I have an initial compile working using:
```
FROM openscad/mxe-base:latest

ARG JOBS=2
ARG CLEAN=clean-junk

WORKDIR /mxe

COPY tbb.mk /mxe/src/

RUN \
	--mount=type=cache,id=pkg,target=/mxe/pkg \
	echo make "JOBS=${JOBS}" "-j${JOBS}" \
		MXE_TARGETS=x86_64-w64-mingw32.static.posix \
		MXE_PLUGIN_DIRS=plugins/gcc13 \
		tbb
```
```
# This file is part of MXE. See LICENSE.md for licensing information.

PKG             := tbb
$(PKG)_WEBSITE  := https://www.threadingbuildingblocks.org
$(PKG)_DESCR    := oneAPI Threading Building Blocks
$(PKG)_IGNORE   :=
$(PKG)_VERSION  := 2021.10.0
$(PKG)_CHECKSUM := 487023a955e5a3cc6d3a0d5f89179f9b6c0ae7222613a7185b0227ba0c83700b
$(PKG)_GH_CONF  := oneapi-src/oneTBB/tags,v
$(PKG)_TARGETS  := $(BUILD) $(MXE_TARGETS)
$(PKG)_DEPS     := cc

$(PKG)_DEPS_$(BUILD) :=
$(PKG)_OO_DEPS_$(BUILD) := $(MXE_CONF_PKGS)

define $(PKG)_BUILD
    # build and install the library
    cd '$(BUILD_DIR)' && $(TARGET)-cmake '$(SOURCE_DIR)' \
        -DTBB_STRICT=OFF \
        -DTBB_BUILD_SHARED=$(CMAKE_SHARED_BOOL) \
        -DTBB_BUILD_STATIC=$(CMAKE_STATIC_BOOL) \
        -DTBB_BUILD_TESTS=OFF
    $(MAKE) -C '$(BUILD_DIR)' -j '$(JOBS)'
    $(MAKE) -C '$(BUILD_DIR)' -j 1 install

    # compile test
    '$(TARGET)-g++' -W -Wall \
        '$(SOURCE_DIR)/examples/test_all/fibonacci/fibonacci.cpp' \
        -o '$(PREFIX)/$(TARGET)/bin/test-$(PKG).exe' \
        `'$(TARGET)-pkg-config' $(PKG) --cflags --libs`
endef
```
```
DOCKER_BUILDKIT=1 docker build -t mxe-tbb .
```
```
docker run -it --rm mxe-tbb:latest bash
```
```
make -j2 JOBS=2 MXE_TARGETS=x86_64-w64-mingw32.static.posix MXE_PLUGIN_DIRS=plugins/gcc13 tbb
```
It does show warnings of static builds not being officially supported, so we need to check if this works.

pca006132:
From https://github.com/oneapi-src/oneTBB/issues/646

> The static library is not a problem itself; however, there are some scenarios when the static linkage approach can lead to the issues. The most typical problem if static and dynamic or static and static libraries are mixed in one application. In the good case, it will be a performance problem, in the worst case, something will mysteriously fail (e.g. due to symbol conflicts).

I feel like the windows problem is exactly this: symbol conflicts. I think if we check the MXE build in the CI by running it on Windows,, it should be fine.

thehans:
@t-paul did you see the actual warning from that?  I only noticed the warning from reading the CMakeLists.txt myself.
https://github.com/oneapi-src/oneTBB/blob/d1c04c902b76f47e597878c8b76c7f716357cb5b/CMakeLists.txt#L111-L120

The old values of :
```
        -DTBB_BUILD_SHARED=$(CMAKE_SHARED_BOOL) \
        -DTBB_BUILD_STATIC=$(CMAKE_STATIC_BOOL) \
```
end up giving a cmake warning about unused settings when checking the mxe build log:

As I understand we need to explicitly set  `-DBUILD_SHARED_LIBS=OFF` to get the static build (and that warning)

thehans:
Also they changed the testing option from `TBB_BUILD_TESTS` to just `TBB_TEST`
https://github.com/oneapi-src/oneTBB/blob/master/CMakeLists.txt#L95

For reference there is also this cmake-specific doc which lists the new options:
https://github.com/oneapi-src/oneTBB/blob/master/cmake/README.md

pca006132:
btw they just released tbb 2021.11.0.

PR #4832, opened 2023-11-16
Repair meshes in manifold & fast-csg routes

ochafik:
This PR introduces some crude mesh repairs in the `manifold` & `fast-csg` modes.

Since orient_to_bound_a_volume seems happy to throw assert and precondition exceptions at us when the mesh has self-intersections or isn't closed, I'm catching these, remove any self intersections & plug any holes (*then* try orienting again w/ more lenient method).

This fixes `3D/features/linear_extrude-scale-zero-tests.scad` along with a few models I had that have invalid geometry but that I prefer seeing repaired than exploding in flight.


src/geometry/manifold/manifoldutils.cc
```diff
@@ -115,31 +115,25 @@ std::shared_ptr<ManifoldGeometry> createMutableManifoldFromPolySet(const PolySet
...
-      points[i] = vector_convert<K::Point_3>(points3d[i]);
+      points[i] = vector_convert<P>(points3d[i]);
     }
     if (points.size() <= 3) return make_shared<ManifoldGeometry>();
 
     // Apply hull
-    CGAL::Surface_mesh<CGAL::Point_3<K>> r;
-    CGAL::convex_hull_3(points.begin(), points.end(), r);
-    CGALUtils::copyMesh(r, m);
+    CGAL::convex_hull_3(points.begin(), points.end(), m);
```
pca006132, line 130: You can use the hull function in manifold. We don't do much optimization for now but we may do them later.

pca006132: Oh sorry, I probably checked the wrong thing.

ochafik: Hah, good to know it's there, not sure about here (would have to inline this logic in a manifold-specific way) but I'll probably give it a spin where we perform the hull primitive itself.

pca006132:
Looking at CGAL, it seems that they do have algorithm for repairing non-manifold triangular soup called `duplicate_non_manifold_vertices`. They are called combinatorial repair https://doc.cgal.org/latest/Polygon_mesh_processing/index.html#PMPCombinatorialRepair.

I think the existing mesh processing in this PR are for dealing with self-intersections, but not sure if this can deal with topological issues (manifoldness).

pca006132:
Ah, bad news. I tried `duplicate_non_manifold_vertices` and `repair_polygon_soup` with no luck. It doesn't work on the example in #4838.

pca006132:
Good news: We need to first `repair_polygon_soup`, and then `orient_polygon_soup`. Here is a demo:

https://github.com/openscad/openscad/commit/782eb0ce498c27165a833e71088ac2ee7cd38461

This works for the simple cases where two cubes are touching at a corner/edge. I did not try more complicated cases. Should be nice to include in this PR.

PR #4833, opened 2023-11-17
Bitwise operators.  Fixes #3345.

### LINKED ISSUE #3345: Suggestion / question: Bitwise operator support
Hi all!

I started working with OpenSCAD but some things have seemed challenging for me to do programmatically and wanted to build some features if it would be considered for master branch. If I built bitwise operators into the language, would that be accepted to master, or should users be using Processing or similar tool for those sorts of things, or am I using OpenSCAD incorrectly or haven't learned the right way to use it?

Bitwise operators were one of the things I immediately ran into that complicated a project.
To implement bitwise AND, I needed to build a recursive and somewhat excessive arithmetic function, and wasteful on large operations since the CPU/GPUs can perform bitwise operations quickly, and I'm specifically using OpenSCAD to perform many-polygon type projects rather than traditional CAD.

```
// 16-bit OpenSCAD AND function -samy kamkar
function and(a, b, c=0, s=0) = 
  s == 15 ? 
  c : 
  and(a, b, 
    (c + ((floor(a/exp(ln(2)*s) % 2) * 
    floor(b/exp(ln(2)*s)) % 2)) *
    exp(ln(2)*s)
  ),
  s+1);
```

The non-return modules and single-expression functions also makes things a bit challenging. Are these restrictions due to some underlying optimization that prevents a more natural, multi-expression subroutine with a return value?

Thanks,
samy 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/91612885-suggestion-question-bitwise-operator-support?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jordanbrown0:
Adds:
* Hexadecimal constants, in C-style 0xHHHHHHHH.
* Left and right shift, using C-style `<<` and `>>`.  Right shift is zero-filling.
* Binary "and", "or", and "not", using C-style `&`, `|`, and `~`.  (No xor; C-style `^` conflicts with exponentiation.)

Precedence is C-like but not quite the same.
C precedence is:  unary (including ~), multiply, add, shift, greater/less, equality, binary-and, binary-or, logical-and, logical-or, ternary
Precedence here moves binary-and and binary-or above comparisons: unary (including ~), multiply, add, shift, binary-and, binary-or, greater/less, equality, logical-and, logical-or, ternary
This avoids the C strangeness where `x & 1 == 0` tests whether 1 is equal to 0, and then ands the result with x.  (I think that a very early C did not have logical-and and local-or, and used binary-and and binary-or for the purpose.)

Does *not* add a new integer type.  Operates on 32-bit values stored as ordinary OpenSCAD numbers.  (OpenSCAD numbers could store ~53-bit values but that seemed harder to explain.)  The conversion is straightforward C conversion from double to unsigned 32-bit integer:  fractions are dropped, values above 2^32 have the top parts chopped off, negative values are converted to 32-bit signed and then unsigned, and so small negative values turn into values near 2^32.

Mostly, you should use these operations only on integers 0 through 2^32-1.  -(2^31) through -1 behave sensibly if you understand two's complement.  It would be best not to rely on the fact that values above 2^32 are truncated; a future version might extend to 53-bit support.

Possibly missing:  a way to turn a number into a hexadecimal string.

I'm not sure that this is a good idea, but it's not clearly a bad idea.  I think there are better solutions for many of the problems that people would try to solve with bit operations, but bit operations may be more comfortable for programmers used to languages like C.  The major "cost" is consuming the `&`, `|`, and `~` operators, and using them for any *other* purpose would confuse people.


pca006132:
maybe we should add some test cases for this?

jordanbrown0:
Yes, though if key people think that this *should not* be merged then there's no point writing tests for it.

src/core/lexer.l
```diff
@@ -262,6 +262,10 @@ use[ \t\r\n]*"<"		{ BEGIN(cond_use); LOCATION_COUNT_LINES(parserlloc, yytext); }
 
 {UNICODE}+              { parser_error_pos -= strlen(yytext); return TOK_ERROR; }
 
+0x{H}{1,8}              {
```
kintel, line 265: Note: There's an ugly bug in OpenSCAD's lexer which allows variables to start with a number. Your patch might break code which relies on that.

src/core/parser.y
```diff
@@ -404,6 +407,34 @@ comparison
             }
 		;
 
+binaryor
+        : binaryand
+        | binaryor '|' binaryand
```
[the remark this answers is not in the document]
kintel: Fair, it's a bit late to add keywords, especially if that makes them unusable as symbols.

kintel:
I'm uncertain about the 32-bit thing.
Otoh, since this is implemented in JavaScript in almost the same way, perhaps it would be cleaner (and easier to explain) if we make our integer operator implementation exactly the same as JavaScript's?

JavaScript's version of integer ops is also very much battle tested and we can stea^H^H^H^Hget inspired by existing implementation details.

jordanbrown0:
Note that there are some design comments over in #3345, including comments on the 0x thing.

src/core/lexer.l
jordanbrown0: It breaks anything that uses variables that would be legal hex constants.  It does not break variables that don't match the pattern.  Whether or not disallowing `0x0f` as a variable is acceptable is probably the biggest technical question in this change.

#3345 mentions a couple of possible alternative styles:  a function that accepts a hex string and returns a number, or a Python-style syntax like `h'0f'`.

Over in #3345 you will see that apparently @nophead relies on the identifiers-that-start-with-digits behavior.

nophead: Yes but I don't think any of them look like hex constants.

kintel:
So many comments... but I'll read #3345

src/core/lexer.l
jordanbrown0: Still... allowing 0w123 and 0y123 as identifiers, but not allowing 0x123, would be odd.  Except for the incompatibility, I'd be in favor of disallowing identifiers that start with digits.  Being able to use real part numbers as names is nice, but if the part number is 4632-342 you can't use it, so maybe the starts-with-digit-otherwise-alphanumeric case is narrow enough to disallow.

Maybe we should think about some kind of pragma mechanism to enable/disable language features.  Start out with `#pragma hex-constants` to enable the new feature, but eventually migrate to `#pragma leading-digit-identifiers` to enable the legacy behavior.

kintel:
In terms of hex values, it could be interesting to let `color()` accept numeric hex values too
..or let `hex()` accept "#rrggbb" strings.

kintel:
> I will compare against what JS does. Agree that ... adopting ... behavior is best.
> 

I believe JS does pretty much the same as you; convert to 32-bit. Just nice to have the same semantics to avoid surprises.

precedence is an orthogonal topic. I don't have any opinions on that yet :)

src/core/lexer.l
kintel: Instead of a per-feature pragma, I'd say just deprecate and use a monotonically increasing version pragma for backwards compatibility instead. That way, the language always moves forward, and it's much easier on us to manage. ..but that's a completely separate discussion :)

jordanbrown0:
I investigated JS a little.  They allow arbitrary-sized hex constants, that (like decimal) just start losing precision when they get over 13 digits.  `0x100000000000000` plus one is unchanged.  My now-backed-out hex constant support rejects constants over 8 digits.

Bitwise arithmetic is limited to the low 32 bits.  `0xfffffffffffff & 0xf00000000` (13 Fs, 8 zeroes) is zero.  That's consistent with what I did.

JS bitwise arithmetic is signed(!).  `0xffffffff` (2^32-1) is a positive number.  `0xffffffff | 1` is -1.  `1<<31` is a large negative number.  I made bitwise arithmetic unsigned because nobody (or at least nobody sane) wants to to signed arithmetic with these operators.  (In a low-level language like C where you might be using bitwise arithmetic to implement normal arithmetic, maybe... but not in a high-level language where you are using bitwise arithmetic to manipulate bit masks.)  If we decided that consistency was critical I could match, but I'd have to hold my nose.

JS precedence is C-like; `a&b==c` is `a & (b==c)`, with a numeric result.

Python has two types, `int` and `float`.  `int` appears to have unlimited precision.  `int` appears to be signed, in that `~0` is displayed as `-1`.  `int` appears to convert to `float` at appropriate times - `int` divided by `int` is `float`, for instance.  (Integer division is `//`.)  Conversion to float has expected problems at 2^52; if you set `x` to `0x100000000000000` (14 zeroes) then `x+1` is precise but `x+1.0` is not.  `x+1-x` is `1`; `x+1.0-x` is `0`.  (`pow(int,positiveInt)` is int!)

Python precedence puts bitwise operators above comparison, so `a&b==c` is `(a&b) == c`, with a boolean result.

Python's behavior seems like the better one to emulate, though implementing unlimited-precision int does not seem worthwhile and I don't know what the implications would be of *not* implementing it.

jordanbrown0:
> Consider supporting "-1" as an "all bits set": a noop with "and".

-1 does naturally equal 0xffffffff, 2^32-1.  Within the bounds of 32-bit arithmetic, -1 & x == x.

jordanbrown0:
Wrote some tests, which exposed some bugs.  Fixed those bugs.

jordanbrown0:
Note from writing tests:  doing bitwise stuff without hex constants is a nuisance.

jordanbrown0:
The test failed on some systems, with differences like:
```
 ECHO: "Expect two warnings:"
-WARNING: undefined operation (number | bool) in file bitwise-operators.scad, line 36
-WARNING: undefined operation (bool | number) in file bitwise-operators.scad, line 38
+WARNING: undefined operation (number | bool) in file bitwise-operators.scad, line 61
+WARNING: undefined operation (bool | number) in file bitwise-operators.scad, line 65
```

The "-" lines are the expected lines; the "+" lines are the actual lines.

The actual lines are simply wrong; the expected lines are where the statements in question are located.

My guess is that there's a CRLF vs LF problem that is causing the lexer to count lines incorrectly when it's presented with a file with unexpected line breaks.  The lexer should really silently handle those issues, but I bet that I ended up with CRLF lines and that the MSYS2 operations didn't clean them up during the push.

Will look at it more in a few hours.

jordanbrown0:
Indeed, looking at the lexer it sure looks like a CRLF file on a UNIX system will count both the CR and the LF as starting new lines, except that CR in comments (at the end of `//`, or anywhere inside `/*...*/` won't be counted.

But I've stripped out the CRs from the test files, so (fingers crossed) the problem should go away for this PR.

jordanbrown0:
I filed the CRLF bug as #4882 .

jordanbrown0:
I fixed an error in the previous comment where 2^31 should have been 2^32.

rewolff:
On Wed, Dec 06, 2023 at 11:56:55AM -0800, Jordan Brown wrote:
> Open questions...
> 

> These first few are easy, I think, but should get concurrence.  * Is
> the non-C-like (but Python-like) precedence OK?  (My answer is
> "yes", that the C precedence is error-prone and not useful.)

Agreed! 

> * Xor, either using punctuation, a text operator, or a function.
> (I'd prefer a function, if at all.  I see xor as very much a
> secondary feature, not useful for the key use case of bitmaps.)

If you start doing things with bitmaps/bitvectors, xor becomes
"useful" pretty quickly. e.g. "somehow the user indicated that bit 5
needs to be the other way around". So I say: Yes I see the usefulness
in an xor, a function is fine if the ^ operator is not available.

> * 32-bit, or more?  And if more, 53-bit, 64-bit, or unlimited?
> 53-bit would be straightforward; the other two more difficult.  (My
> answer: leave it at 32-bit but make it clear that that might change
> in the future.)

Fully agreed. 

> * Signed, or not?  Simplest case: is `~0` equal to 2^31-1, or is it
> equal to -1?  (My answer: unsigned; signed does not seem useful for
> the use cases here.)

Agreed. But I'd like to add that ~0 and -1 are the same. There are
people to whom -1 is more natural than ~0 (me :-) ). 


> * Right shift sign-extending, or not?  (My answer: not, especially
> if these values are unsigned.)

Indeed. -1 >> 1 == 2^31-1.

> That brings us to the not-easy questions: presenting values as hex.
> (Octal and binary are theoretically possible but probably have very
> small audiences.)  There's an argument for leaving this for another
> PR, but in writing the test cases I found it very unnatural to talk
> about bitwise arithmetic in decimal.

> There are three sub-cases:
> * Hex constants.
> * Hex-string-to-number.
> * Number-to-hex-string.
> 
> Details:
> * Hex constants
>   * 0xHHHHHHHH.  Incompatible change, but obvious.

I'd vote for this one: Such incompatible changes happen from time to
time when a language evolves. I've had a C variable name suddenly be a
keyword in a later edition of C. Thus: Although 0x89ab is now a valid
identifier, it would become a numerical constant in the future. 

The 0.001% of users who have their variables or modules named that way
are out of luck. Sorry. (it makes life easier for the 1% of users who
will use hex constants). 

> * Number-to-string

>   * `str(n, base=16)` - adding a named parameter to `str()`.
>   Currently `str()` ignores names, and theoretically giving them
>   semantics would be incompatible, but it seems unlikely that
>   anybody uses this "feature".  Would allow for additional
>   numeric-formatting options, e.g. `width=10`, `pad="0"`, et cetera
>   and so lead to solving additional problems.

I like generalizations that happen to solve additional problems!


-- 
** ***@***.*** ** https://www.BitWizard.nl/ ** +31-15-2049110 **
**    Delftechpark 11 2628 XJ  Delft, The Netherlands.  KVK: 27239233    **
f equals m times a. When your f is steady, and your m is going down
your a is going up.  -- Chris Hadfield about flying up the space shuttle.

jordanbrown0:
> But I'd like to add that ~0 and -1 are the same. 
As described, they aren't the same.  They're the same *after* you feed the -1 through at least one bitwise operator.  But before that, ~0 is 2^32-1, and -1 is... -1.

The way that I would think of it is that the bitwise operators act on integer values 0..2^32-1.  If you try to use a bitwise operator on a non-integer, or on a value outside that range (like -1), it gets coerced into being an integer inside that range - for a non-integer by truncating the fractional part; for a value outside that range by stripping off the high bits in 2's-complement representation.

So there are really three behaviors there:
* Non-integers.  I won't promise that I'll never extend the bitwise operators into the fractional space.  Can't think of why I would, but the purist in me likes it.
* Numbers greater than 2^32-1.  I specifically don't promise that they will always be stripped down to 32 bits.
* Negative numbers.  This is a tough question.  Today, `-1|0` is a positive number, as is `~0`.  Arguably, we should reserve the right for one or them to both be negative, some day.  We should probably guarantee that 0x80000000 will never be negative, but if we go above 32 bits maybe we'll go signed.  I'd say we should be very clear in documentation that the future is uncertain.

jordanbrown0:
Does not currently include XOR, notably because ^ is used for exponentiation.

Note that XOR is easily implemented in OpenSCAD-language:
```
function xor(a,b) = (a|b) & ~(a&b);
```

jordanbrown0:
I don't know that that's final - see comments in source - but it's certainly something that could be reviewed.

I think we should put hex-string-to-number and number-to-hex-string off into a future wad.

jordanbrown0:
hex-string-to-number and number-to-hex-string are covered by #5586 and #5587 respectively.

src/core/lexer.l
kintel: I think we should allow more than 32 bits for hex constants; hex constants have more value than being used as a bitwise operand.
We'll naturally lose precision when going above 53 bits, but that's also the case for decimal constants.

kintel:
I think we should reconsider negative numbers. As a comparison, both Python and JavaScript evaluate `-2 << 1` to `-4`, and, IMO, the most unsurprising behavior would be for us to do the same.

src/core/lexer.l
jordanbrown0: OK

jordanbrown0:
I'm not morally opposed to handling negative numbers.  I'll have to think through exactly what it means, though, given the various transformations going back and forth between double and int32.

src/core/lexer.l
jordanbrown0: I removed the limit on the number of hex digits, though I implicitly added a limit at 2^64 since that's what you can get from an unsigned long long.

I added a warning if you go over 2^64.
I added a warning (for both hex and decimal integer constants) if the value cannot be represented precisely.
Note that for decimal constants you can suppress that warning by appending a `.` so that it is no longer considered an integer.

If we wanted to eliminate the 2^64 limit, we could DIY the strtoull(), calculating using doubles instead of ULLs.

jordanbrown0:
With regard to negative numbers:  I haven't thought them through completely, but I realized one thing.

We can never make a hex constant turn into a negative number.  Today, with 32-bit bitwise math, maybe we would think that 0xFFFFFFFF should be equal to -1.  But suppose that tomorrow we decide to extend it to 48-bit bitwise math... that would need to be equal to 2^32-1.

Similarly, it's a little worrisome that the result of the `~` operator depends on how many bits we use.  That will need a clear caveat in the documentation.

jordanbrown0:
I was incorrect when I said that the result of the ~ operator depends on how many bits we use.  If we allow negative numbers, it doesn't.

Latest pushes use signed numbers, and uses int64_t under the covers so it gets ~53 bits.  As noted above, hex constants are always positive.  (As are decimal constants; -1 is internally the negation of 1, not a constant negative 1.)  I also simplified it; it doesn't try to warn on over-large values because the limit is now so high.  It seems to be mostly OK, though I had to fix several tests.

I think the most likely use case for bitwise operations is unsigned, but it doesn't really matter here because you can't use all 64 bits anyway; if you start getting bits up near the sign bit then you start losing bits on the right.

One case that bugs me is that `1<<63` is a -2^31 and `-3<<62` is a 2^62, because they shift the wrong bit into the sign bit.  Options:
* Live with it.  You asked for it, you got it.  (This is what JavaScript does.)
* Preserve the sign bit on left shifts, essentially truncating at 63 bits.
* Implement shift as an operation on double.  (Which might have the correct-but-surprising result that `3>>1==1.5`.)
Note:  Python appears to have infinite-precision integers, and so gets these cases right.
Right now my answer is "live with it".  If you're doing these operations, it's probably what you expect.

My OCD wants me to implement these operations as operations on doubles, so that they work for values of all magnitudes.  But that's not trivial for `&` and `|`, and it's useless.

kintel:
Thanks! I’ll look at this again soon

src/core/Value.cc
```diff
@@ -269,6 +269,19 @@ bool Value::toBool() const
   // NOLINTEND(bugprone-branch-clone)
 }
 
+// Convert the value to a double with an integer value, for use in bitwise operations.
+// Since there are several possible ways to do this (floor, ceil, round) this function
+// centralizes the choice for consistency.
+double Value::toInteger() const
+{
+  return floor(this->toDouble());
```
kintel, line 277: Now that we support negative numbers, I think this should trunc() here, not round down

kintel (commented):
Just isolating one issue at a time, as they tend to affect the whole implementation.

src/core/Value.cc
jordanbrown0: I'm not strongly opposed, but It seems like truncation is generally the wrong answer for float-to-integer conversion, because it means that zero is over-represented in the results.  If you want something that's more human-obvious for negative numbers than floor(), would you consider round()?

kintel: It's not really conversion, it's just casting the result to an integer, and truncating is usually the answer.
For reference, Javascript and Python agrees. Python needs an explicit conversion, but it will trunc upon conversion.

jordanbrown0: Yeah, I think they're wrong too.  But OK.  I can't come up with any practical cases where there's obviously-right or obviously-wrong answers, in the context of bitwise arithmetic.

(But see my just-filed doc bug for the ... odd... behavior of picking a random integer from a range that crosses zero, if you use trunc().  In that case, I'd say that trunc() is one of the most wrong answers.  However, I can't come up with any way to reasonably tie that to bitwise arithmetic.)

jordanbrown0: Another alternative is to implement the bitwise operations for floating point values.  But few people would understand, and it would be an unpleasant amount of work.

jordanbrown0: Array indexes appear to truncate, so that's one rationale.

kintel: Least surprising when comparing to other implementations is also a good rationale. We probably violate that in a bunch of places, but I'd think mostly by accident, not because we did the research and designed it to fit a particular class of applications :)

jordanbrown0:
Note:  I had some strange stuff happen with OpenCSG and Clipper2, and I'm not sure that I resolved them correctly.  I think I made them be the same as upstream/master, but I would be happier if I didn't have to touch them at all - and I couldn't figure out how to do that.  Anyhow, before you merge this make sure that you double check that I didn't do anything bad there.

Also:  I will be happy to squash to a single changeset when and if that's desirable.

kintel:
Thx! Submodules sometimes act up, but it looks good now. FYI: We recently changed the OpenCSG upstream from our fork back to the official upstream as our patches are no longer needed.

src/core/lexer.l
```diff
@@ -253,22 +256,59 @@ use[ \t\r\n]*"<"        { BEGIN(cond_use); LOCATION_COUNT_LINES(parserlloc, yyte
...
 
-{D}+{E}? |
+0x{H}+                  {
+                            errno = 0;  // strtoxxx have crummy error semantics
+                            unsigned long long ull = strtoull(yytext + 2, NULL, 16);
+                            parserlval.number = ull;
+                            if (errno != 0) {
+                                LOG(message_group::Warning, LOCATION(parserlloc), "",
+                                    "Hexadecimal constant \"%1$s\" too large", yytext);
+                            } else if ((unsigned long long)parserlval.number != ull) {
```
kintel, line 266: Interestingly, the result of the cast to unsigned long long isn't the same on all platforms, causing the test to fail by not printing the expected warning:
```
-WARNING: Integer "0xffffffffffffffff" cannot be represented precisely in file ../../tests/data/scad/functions/bitwise-operators.scad, line 112
-WARNING: Integer "18446744073709551615" cannot be represented precisely in file ../../tests/data/scad/functions/bitwise-operators.scad, line 112
```

 On arm64, it looks like 0xffffffffffffffff is actually exactly representable using double. I guess it happens to cast back to the same value:

0xffffffffffffffff == 18446744073709551615 -> double -> unsigned long long = 18446744073709551615
0xfffffffffffffffe == 18446744073709551614 -> double -> unsigned long long = 18446744073709551615

Just using the latter number in the test makes the tests pass, but haven't tested on non-arm64.

jordanbrown0: That's ... interesting.  I suspect that it means that somebody isn't IEEE 754 compliant.

I'd be interested to look at the binary representation of that double.

I don't know how well-versed you are with binary floating point, but recovering that original value is sort of like this example in decimal:
* Assume that your calculator can only represent three decimal digits, plus an exponent.
* Given the input number 99999, you would convert it to either 9.99e+4 or 1.00e+5.
* And then somehow when that gets converted back, it's recovering the lost 99.

There's about 2048 (give or take a factor of two) numbers in that range that should all convert to the same double value,, because that's a 64-bit value and there's only 53 bits or so available in a double.

I'll push through a change to get a bit further away from the apparently-magic all-ones value.  `0xfffffffffffffff0` is still well more bits than will fit into a double.

kintel: It's curious - I'll run some tests to look at what's going on, just so we don't jump to conclusions

kintel: Binary representation between x86_64 and arm64 is the same, but conversion back does indeed differ: https://godbolt.org/z/sTT897Y3v

kintel: Also, result is the same on gcc vs. clang

jordanbrown0: Nit:  your display of the sign bit will be wrong, because you don't right-shift it down to being in the low-order bit.

It looks to me (without analyzing bit-by-bit) like both of them are converting that u64 to 2^64 as a double.  That makes perfect sense, since it's 2^64-1.

2^64 won't fit back into a u64.  But then arm64 is somehow converting it back to 2^64-1 (probably to the maximum u64), while x86-64 is somehow converting it to 2^63.  I can't explain that x86-64 answer; 2^63 is the middle of the range for u64, not the maximum.

So my original test probably failed because its test value was exactly the maximum u64, and converting from an out-of-range double to u64 on arm64 resulted in the maximum u64.

The warning on x86-64 is correct, even with the inexplicable conversion back to u64:  the value can't be represented precisely.

On arm64 the picture is murkier.  Because of the clamping conversion, it thinks that 2^64-1 can be represented precisely... when it isn't.  (The double representation is 2^64, not 2^64-1.)

That's nearly harmless, just a missing warning on one obscure case.  But can it be fixed in a portable fashion?  Given that we're talking here about ULLs, which strictly aren't necessarily the same as u64s... don't know.  I'm not immediately coming up with anything that I'm sure will work.

But I just realized something that is more wrong (though still pretty extreme). 36893488147419103232 will give this warning... and it's wrong.  That's 2^65, which *will* be represented precisely.  I don't have any good ideas about what to do about that.  Maybe redefine "represented precisely" to mean that N-1, N, and N+1 are three distinct values, which will be false for anything >2^53 or so.  That can be checked without any conversion to/from an integer type.  You'll get the warning for numbers that happen to end with a bunch of zero bits, when they are in fact represented precisely, but they are in a very gray area.  Such a warning might help with the user who was complaining about the behavior of ranges with values > 2^53.

kintel: I guess it's technically undefined behavior:

https://en.wikipedia.org/wiki/Double-precision_floating-point_format

A [prvalue](https://en.cppreference.com/w/cpp/language/value_category#prvalue) of floating-point type can be converted to a prvalue of any integer type. The fractional part is truncated, that is, the fractional part is discarded.

If the truncated value cannot fit into the destination type, the behavior is undefined (even when the destination type is unsigned, modulo arithmetic does not apply).

kintel: For now, I say we just need to make the warning appear in the test on all systems we're developing on :)

jordanbrown0: I think that's done.  We can see later if anybody complains about spurious or missing warnings.

(No fair cheating by reading this conversation!)

kintel (approved):
LGTM. The only thing I can think of would be more tests for boundary conditions, now that we support > 32 bits, but we can cross that bridge later.

jordanbrown0:
Do you want/need me to squash out all of my development changesets?  It would be silly for them all to end up in the permanent record.

kintel:
Nah, I'm squashing when merging.

Issue #3781, opened 2021-05-19
File scope $variables ignored if a file is used by another file


nophead:
OpenSCAD 2021.05.11.ci7927  breaks my library. OpenSCAD 2021.03.24.ci7656  was the last snapshot I tried that worked.

In my `global_defs.scad` file that I include in every file in my library and project I have constants like these:

```
$_bom           = is_undef($bom)             ? 0      : $bom;             // 0 no bom, 1 assemblies and stls, 2 vitamins as well
$exploded       = is_undef($explode)         ? 0      : $explode;         // 1 for exploded view
```

Previously I could set `$bom` and `$explode` on the command line , or in my main file, and it would change the values of `$_bom` and `$exploded` throughout the project and used modules from the library. Now it doesn't seem to have the same effect on my library modules. Some BOMs are not produced. Also setting $explode at the end of my main file doesn't work but setting it at the top before my library is included does. Also this doesn't work anymore:

```
                let($explode = 1)
                    spring_washer(w);
```

It formally showed the spring washer exploded but now shows it compressed. I.e. in the library module being called it doesn't see `$exploded` set to 1.

nophead:
My washer are drawn correctly with version OpenSCAD 2021.05.01.ci7809 but are wrong with version OpenSCAD 2021.05.03.ci7815. The only two commits between those two are d9491713914e37dbe04ec9f748c5111e7294bd9e and 9317cd98b15d3270d753ed7e6fecda024437cb3c, which are just additional warnings, so I can't see how they broke special variables.

nophead:
I tried to set up a simple test case to show this bug but found $variables are totally broken. With a simple test like this:

dollars.scad:
```
$v = 0;

module showv() {echo($v);}
```

test_dollars.scad:
```
use <dollars.scad>

showv();
```

The expect result and the result with OpenSCAD 2021.05.01.ci7809:
```
Parsing design (AST generation)...
Compiling design (CSG Tree generation)...
ECHO: 0
```

With OpenSCAD 2021.05.03.ci7815:
```
Parsing design (AST generation)...
Compiling design (CSG Tree generation)...
WARNING: Ignoring unknown variable '$v' in file dollars.scad, line 3
TRACE: called by 'echo' in file dollars.scad, line 3
TRACE: called by 'showv' in file test_dollars.scad, line 3
Execution aborted
 ```

[...]

PR #4834, opened 2023-11-18
#3781 Proof of Concept

pca006132:
It is unfortunate that #3781 sits here for 2 years without any progress, breaking some user code and blocking release. This is an attempt to fix the behavior by adding a fallback to file scope when lookup to config variables failed. This is definitely not fast nor clean code (I don't understand much about the codebase), but I think the behavior should be fine.

```
// dollars.scad
$v = 0;
$v1 = $v;

module showv() {
    echo($v);
    echo($v1);
}

// test_dollars.scad
use <dollars.scad>
showv();
// output:
// ECHO: 0
// ECHO: 0

// test_dollars1.scad
use <dollars.scad>
$v = 1;
showv();
// output:
// ECHO: 1
// ECHO: 1

// test_dollars2.scad
use <dollars.scad>
$v = 1;
$v1 = 0;
showv();
// output:
// ECHO: 1
// ECHO: 0
```

Basically, in the above examples, if a config variable is not being overridden in the dynamic scope, it will try to use the value in the file scope (and can lookup other config variables in the stack). If the value is being overridden, it will use the overriden value.

### DIFF
--- a/src/core/Context.cc
+++ b/src/core/Context.cc
@@ -63,10 +63,10 @@ std::vector<const std::shared_ptr<const Context> *> Context::list_referenced_con
   return output;
 }
 
-boost::optional<const Value&> Context::try_lookup_variable(const std::string& name) const
+boost::optional<const Value&> Context::try_lookup_variable(const std::string& name, const Location& loc) const
 {
   if (is_config_variable(name)) {
-    return session()->try_lookup_special_variable(name);
+    return session()->try_lookup_special_variable(name, loc);
   }
   for (const Context *context = this; context != nullptr; context = context->getParent().get()) {
     boost::optional<const Value&> result = context->lookup_local_variable(name);
@@ -79,7 +79,7 @@ boost::optional<const Value&> Context::try_lookup_variable(const std::string& na
 
 const Value& Context::lookup_variable(const std::string& name, const Location& loc) const
 {
-  boost::optional<const Value&> result = try_lookup_variable(name);
+  boost::optional<const Value&> result = try_lookup_variable(name, loc);
   if (!result) {
     LOG(message_group::Warning, loc, documentRoot(), "Ignoring unknown variable '%1$s'", name);
     return Value::undefined;
--- a/src/core/Context.h
+++ b/src/core/Context.h
@@ -79,7 +79,7 @@ public:
   virtual const class Children *user_module_children() const;
   virtual std::vector<const std::shared_ptr<const Context> *> list_referenced_contexts() const;
 
-  boost::optional<const Value&> try_lookup_variable(const std::string& name) const;
+  boost::optional<const Value&> try_lookup_variable(const std::string& name, const Location& loc) const;
   const Value& lookup_variable(const std::string& name, const Location& loc) const;
   boost::optional<CallableFunction> lookup_function(const std::string& name, const Location& loc) const;
   boost::optional<InstantiableModule> lookup_module(const std::string& name, const Location& loc) const;
--- a/src/core/EvaluationSession.cc
+++ b/src/core/EvaluationSession.cc
@@ -69,6 +69,11 @@ boost::optional<const Value&> EvaluationSession::try_lookup_special_variable(con
         EvaluationSession dummy(documentRoot());
         auto dummyContext = Context::create<Context>(&dummy);
         auto filecontext = Context::create<FileContext>(dummyContext->get_shared_ptr(), file);
+        // copy config variables from the stack
+        for (const ContextFrame* frame : stack) {
+          filecontext->apply_config_variables(*frame);
+        }
+
         Value result = assignment->getExpr()->evaluate(filecontext->get_shared_ptr());
         // set it in the top level context, so it is cleared when the frame is popped
         stack.back()->set_variable(name, std::move(result));
--- a/src/core/EvaluationSession.cc
+++ b/src/core/EvaluationSession.cc
@@ -24,8 +24,13 @@
  *
  */
 
+#include "AST.h"
 #include "ContextFrame.h"
 #include "EvaluationSession.h"
+#include "Expression.h"
+#include "ScopeContext.h"
+#include "SourceFile.h"
+#include "SourceFileCache.h"
 #include "printutils.h"
 
 size_t EvaluationSession::push_frame(ContextFrame *frame)
@@ -47,7 +52,7 @@ void EvaluationSession::pop_frame(size_t index)
   assert(stack.size() == index);
 }
 
-boost::optional<const Value&> EvaluationSession::try_lookup_special_variable(const std::string& name) const
+boost::optional<const Value&> EvaluationSession::try_lookup_special_variable(const std::string& name, const Location &loc) const
 {
   for (auto it = stack.crbegin(); it != stack.crend(); ++it) {
     boost::optional<const Value&> result = (*it)->lookup_local_variable(name);
@@ -55,12 +60,28 @@ boost::optional<const Value&> EvaluationSession::try_lookup_special_variable(con
       return result;
     }
   }
+  // try file level lookup
+  SourceFile *file = SourceFileCache::instance()->lookup(loc.fileName());
+  if (file) {
+    // find assignment of the special variable
+    for (const std::shared_ptr<Assignment> &assignment : file->scope.assignments) {
+      if (assignment->getName() == name) {
+        EvaluationSession dummy(documentRoot());
+        auto dummyContext = Context::create<Context>(&dummy);
+        auto filecontext = Context::create<FileContext>(dummyContext->get_shared_ptr(), file);
+        Value result = assignment->getExpr()->evaluate(filecontext->get_shared_ptr());
+        // set it in the top level context, so it is cleared when the frame is popped
+        stack.back()->set_variable(name, std::move(result));
+        return stack.back()->lookup_local_variable(name);
+      }
+    }
+  }
   return boost::none;
 }
 
 const Value& EvaluationSession::lookup_special_variable(const std::string& name, const Location& loc) const
 {
-  boost::optional<const Value&> result = try_lookup_special_variable(name);
+  boost::optional<const Value&> result = try_lookup_special_variable(name, loc);
   if (!result) {
     LOG(message_group::Warning, loc, documentRoot(), "Ignoring unknown variable '%1$s'", name);
     return Value::undefined;
--- a/src/core/EvaluationSession.h
+++ b/src/core/EvaluationSession.h
@@ -23,7 +23,7 @@ public:
   void replace_frame(size_t index, ContextFrame *frame);
   void pop_frame(size_t index);
 
-  [[nodiscard]] boost::optional<const Value&> try_lookup_special_variable(const std::string& name) const;
+  [[nodiscard]] boost::optional<const Value&> try_lookup_special_variable(const std::string& name, const Location& loc) const;
   [[nodiscard]] const Value& lookup_special_variable(const std::string& name, const Location& loc) const;
   [[nodiscard]] boost::optional<CallableFunction> lookup_special_function(const std::string& name, const Location& loc) const;
   [[nodiscard]] boost::optional<InstantiableModule> lookup_special_module(const std::string& name, const Location& loc) const;
--- a/src/core/Parameters.cc
+++ b/src/core/Parameters.cc
@@ -45,7 +45,7 @@ Parameters::Parameters(Parameters&& other) noexcept :
 boost::optional<const Value&> Parameters::lookup(const std::string& name) const
 {
   if (ContextFrame::is_config_variable(name)) {
-    return frame.session()->try_lookup_special_variable(name);
+    return frame.session()->try_lookup_special_variable(name, loc);
   } else {
     return frame.lookup_local_variable(name);
   }
--- a/src/core/builtin_functions.cc
+++ b/src/core/builtin_functions.cc
@@ -868,7 +868,7 @@ Value builtin_is_undef(const std::shared_ptr<const Context>& context, const Func
     return Value::undefined.clone();
   }
   if (auto lookup = dynamic_pointer_cast<Lookup>(call->arguments[0]->getExpr())) {
-    auto result = context->try_lookup_variable(lookup->get_name());
+    auto result = context->try_lookup_variable(lookup->get_name(), call->location());
     return !result || result->isUndefined();
   } else {
     return call->arguments[0]->getExpr()->evaluate(context).isUndefined();


pca006132:
btw, I think imported functions will not work in this case, not sure how should I fix this.

pca006132:
@t-paul @nophead @jordanbrown0 any comments on this? Is this what you want or are there other issues related to this?

pca006132:
It is pretty easy to guard this behind a configuration, considering the code is only in 1 function. Will do that later.

The re-evaluation problem is tricky. I am not very familiar with OpenSCAD codebase, so it is not easy to understand that.
In general though,

t-paul:
> It is pretty easy to guard this behind a configuration, considering the code is only in 1 function. Will do that later.

I fear it might be more complicated. The question is if that setting needs to be on the file scope, so basically each file imported via `use<>` can have it's own setting.

nophead:
I can't test anything at the moment as my laptop died two weeks ago and it has taken that long to get a new PC on Tenerife due to incompetent delivery people at a company called Seur. It will take a few days to set it up.

Putting the search back has no effect on speed because without it you get an error if you need it. Fixing the revaluation bug would be easier without $variables but this bug doesn't help.

pca006132:
> > It is pretty easy to guard this behind a configuration, considering the code is only in 1 function. Will do that later.
> 
> I fear it might be more complicated. The question is if that setting needs to be on the file scope, so basically each file imported via `use<>` can have it's own setting.

It is still pretty easy, the code already checks the file referencing the config variable, so it is fine if the setting is associated with that file. May need to change `SourceFile` class a bit to include this compatibility flag, but should be simple as well.

pca006132:
> Putting the search back has no effect on speed because without it you get an error if you need it. Fixing the revaluation bug would be easier without $variables but this bug doesn't help.

Yeah. Ideally, if config variables are really used for configs, their evaluation should not take too much time. And we probably should not spend too much time optimizing something that we want to phase out soon.

nophead:
It is the fact that any variable can depend on a $varible, directly or indirectly, perhaps by a function call into a another module. So if that $variable has changed since the last call those expressions need to change. Currently all expressions are evaluated on every function or module call into a used module so in a large project that dominates the time to run the script. It needs to be more clever to only revalue an expression if it depends on a $value that has changed.

For example, a large number of expressions in a project might depend on $extrusion_width and $layer_height but they may never change for the entire project, but they could do.

jordanbrown0:
I haven't looked at your specific proposal.  (I can if somebody wants me to.)

My (conflicting!) goals are:
* Avoid incompatible change.
* Avoid the "process all assignments on every call" slowdown.
* Have behavior that's easy to explain.
* Keep implementation simple (because then it's less fragile and less likely to have bugs).

My tentative idea is to introduce an entirely new use/include mechanism in the form of a function that reads a .scad file and returns an object populated with values, function reference, and module references.  As a new mechanism it would not have backward compatibility constraints, and the "return entire library as a single object with a name supplied by the caller" scheme avoids a ton of namespace issues.

pca006132:
@jordanbrown0 my implementation should be simple enough. The actual logic is just 20 lines. It may be a bit slow because it requires copying every config variables along the stack, but I am not sure if there is a simpler way to do it. It should also be compatible, as this only changes the behavior when nothing else along the callstack defined the config variable. The caching behavior is harder to explain, but it is not visible to the user and only affects performance a bit.

pca006132:
How should we enable the per file setting? Or should we implement it as some feature flag for now?

pca006132:
I am not sure what the consensus, whether this behavior is considered undesirable but kept for compatibility reasons (may be removed one day), or will be kept indefinitely.

nophead:
Is there a way to get a Windows binary of this branch without building it? There used to a non obvious way to get assets from the CI build but it is so long ago that I ran a new OpenSCAD I forgot how and it has probably changed, or gone.

pca006132:
https://app.circleci.com/pipelines/github/openscad/openscad/4229/workflows/ace95605-2fa0-44c4-95c4-88dddd35dbd7/jobs/17057/artifacts

nophead:
Thanks. I will give it a Spin. I have problems with user fonts on the last release in Windows 11, so I might have to update but this bug has stopped me updating for two years.

nophead:
Yes that seems to have fixed the bug. Compared to the last version I used it seems to run my fast tests a bit slower and my slow tests faster. So it seems to take about half a second longer to start up but is then about 25% faster. 

I haven't got any of the experimental features turned on when I run it from the command line. In the GUI if I enable manifold it terminates as soon as I do a preview.

kintel:
As discussed in various fora, this re-introduces behavior which we found to be a bit questionable. Before doing this we probably need a much better idea of why this would be wanted behavior.

nophead:
There was a bug where file scope $variables in used files overrode values passed up the call stack. The fix for that completely ignored file scope $variables in used files, which was worse in my opinion and the reason why I stopped updating OpenSCAD for two years. When you used a file and $values in it could not be accessed by functions and modules but if you copied them into a regular variable you could access the value, which is weird. It made it so a file could be used as the main file in the GUI but it broke if you used in another file.

pca006132:
@nophead Not sure if this will ever be merged, but for now you can use the build in https://app.circleci.com/pipelines/github/openscad/openscad/4352/workflows/64f45f84-846a-4a14-bcb2-f9ed42dc409c/jobs/17526/artifacts (building, have to wait a bit)

pca006132:
@nophead I think the problem you got is due to me not understanding those symbol scoping issues. Will fix but it is lower priority for now.

I think the consensus is the behavior implemented here is not recommended, and you should probably fix your library...

nophead:
It isn't just my library, it is every project I have ever written in OpenSCAD over the last 13 years that I would have to rewrite.

In this case it seems to be this line that is the root cause.
```
        if(is_undef($show_front) || $show_front)
```

$show_front is defined as true at the top of the file and is a customizer variable. When main.scad is opened in the GUI it works fine. When main.scad is used by bom.scad it gives this error. If I comment out this line and subsequent lines with similar $variables expressions it works. I don't understand why it produces the error message and call stack that it does.

```
[WARNING: Ignoring unknown function 'is_undef' in file W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/global_defs.scad, line 31](https://github.com/openscad/openscad/pull/31,W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/global_defs.scad)
[TRACE: called by 'is_undef' in file W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/global_defs.scad, line 31](https://github.com/openscad/openscad/pull/31,W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/global_defs.scad)
[TRACE: overwritten assignment to 'layer_height' (this is where the assignment is evaluated) in file W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/global_defs.scad, line 31](https://github.com/openscad/openscad/pull/31,W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/global_defs.scad)
[TRACE: overwriting assignment to 'layer_height' in file W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/global_defs.scad, line 31](https://github.com/openscad/openscad/pull/31,W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/global_defs.scad)
[TRACE: called by 'is_undef' in file main.scad, line 133](https://github.com/openscad/openscad/pull/133,C:/Users/palme/Nextcloud/CHRIS/Projects/RepRaptor/HydraMaker/scad/main.scad)
[TRACE: called by 'if' in file main.scad, line 133](https://github.com/openscad/openscad/pull/133,C:/Users/palme/Nextcloud/CHRIS/Projects/RepRaptor/HydraMaker/scad/main.scad)
[TRACE: called by 'children' in file W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/utils/core/global.scad, line 50](https://github.com/openscad/openscad/pull/50,W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/utils/core/global.scad)
[TRACE: called by 'translate' in file W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/utils/core/global.scad, line 50](https://github.com/openscad/openscad/pull/50,W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/utils/core/global.scad)
[TRACE: call of 'translate_z(z = 233)' in file W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/utils/core/global.scad, line 49](https://github.com/openscad/openscad/pull/49,W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/utils/core/global.scad)
[TRACE: called by 'translate_z' in file main.scad, line 101](https://github.com/openscad/openscad/pull/101,C:/Users/palme/Nextcloud/CHRIS/Projects/RepRaptor/HydraMaker/scad/main.scad)
[TRACE: called by 'children' in file W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/utils/core/bom.scad, line 124](https://github.com/openscad/openscad/pull/124,W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/utils/core/bom.scad)
[TRACE: called by 'let' in file W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/utils/core/bom.scad, line 123](https://github.com/openscad/openscad/pull/123,W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/utils/core/bom.scad)
[TRACE: called by 'if' in file W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/utils/core/bom.scad, line 122](https://github.com/openscad/openscad/pull/122,W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/utils/core/bom.scad)
[TRACE: called by 'children' in file W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/utils/core/bom.scad, line 74](https://github.com/openscad/openscad/pull/74,W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/utils/core/bom.scad)
[TRACE: called by 'let' in file W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/utils/core/bom.scad, line 74](https://github.com/openscad/openscad/pull/74,W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/utils/core/bom.scad)
[TRACE: call of 'no_pose()' in file W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/utils/core/bom.scad, line 73](https://github.com/openscad/openscad/pull/73,W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/utils/core/bom.scad)
[TRACE: called by 'no_pose' in file W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/utils/core/bom.scad, line 121](https://github.com/openscad/openscad/pull/121,W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/utils/core/bom.scad)
[TRACE: call of 'assembly(name = "main", big = undef, ngb = false)' in file W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/utils/core/bom.scad, line 114](https://github.com/openscad/openscad/pull/114,W:/CHRIS/Projects/RepRaptor/OpenSCADLibs/NopSCADlib/utils/core/bom.scad)
[TRACE: called by 'assembly' in file main.scad, line 95](https://github.com/openscad/openscad/pull/95,C:/Users/palme/Nextcloud/CHRIS/Projects/RepRaptor/HydraMaker/scad/main.scad)
[TRACE: call of 'main_assembly()' in file main.scad, line 94](https://github.com/openscad/openscad/pull/94,C:/Users/palme/Nextcloud/CHRIS/Projects/RepRaptor/HydraMaker/scad/main.scad)
[TRACE: called by 'main_assembly' in file bom.scad, line 2](https://github.com/openscad/openscad/pull/2,C:/Users/palme/Nextcloud/CHRIS/Projects/RepRaptor/HydraMaker/scad/bom.scad)
```

In the last release `$show_path` would be defined. and it works After the bug was introduced that ignored $variables in used files `$show_path` would not be defined. With this fix it should be defined again but the odd thing is it should not matter in this line because it is guarded by a check for it being defined. And why it says `is_undef()` in `main.scad`  is calling another `is_undef()` in an included file and that `is_undef()`  is an unknown function is beyond me.  Yes that file is included all over the place but there isn't anything being called in it at that point, so why is it shown on the call stack?

jordanbrown0:
@nophead Is that "if is_undef" line at the top level of the "used" file, or is it inside a module?

nophead:
The line in main.scad is inside a main_assembly() module. It references a $variable defined at file scope in main.scad, which is a used file, by bom.scad.

The one it says is ignored is at file scope in an included file.

MichaelAtOz:
For what it's worth, I support the original behaviours, as I [extensively argued at the time](https://github.com/openscad/openscad/issues/3881).
The proposed [alternate .scad coding fix](https://github.com/openscad/openscad/issues/3781#issuecomment-920799080) is worse. Bringing [other unpleasant work-arounds](https://github.com/openscad/openscad/issues/3781#issuecomment-1827536662).
The original simplicity is how I perceive OpenSCAD as a language, easy to learn and easy to extend your coding proficiency. 
Requiring such alternative coding does not engender that.
Anyway, I'm not going to comment further. With my recent limitations, I'm sticking to my Admin role.

pca006132:
relevant irc discussion: https://libera.irclog.whitequark.org/openscad/2023-11-26#1701007784-1701043184;

I am personally in flavor of this fix, as it is not really causing any incompatibility, just a fallback behavior. We already have some difficult to explain variable behavior, and this fallback behavior is not too hard to understand.
I don't have too strong opinion about this, so I am still waiting for a real consensus as to whether or not this is desired instead of trying hard to argue about it...

nophead:
I don't think the IRC discussion understands the problem at all.

It certainly isn't the case that I can change all my $variables in the library to functions and it won't affect the client code. And is certainly isn't the case that my library and projects work without this fix. I have to use the last release for all of the command line work because I use files and invoke modules in them to make specify assembly views and STLs, etc. Those files might work in the GUI but don't from command line because of the extra level of use.

I don't know why you think I am the only person using file scope $ variables when I only had to look as far as the second official example to find code that uses them.  CSG-modules.scad has the following:
```
// Global resolution
$fs = 0.1;  // Don't generate smaller facets than 0.1 mm
$fa = 5;    // Don't generate larger angles than 5 degrees
```

Now if you use CGS-modules.scad and invoke the modules it has defined you will be surprised to find that the $fs and $fa values are ignored and the default values are used.

pca006132:
@rcolyer maybe it is easier to just list out the pros and cons of implementing the feature in this PR (with some further fix for function resolution I guess). With this we may come up with a conclusion, or let everyone vote or something. Otherwise we are in another ideological debate which will never end. What do you think?

MichaelAtOz:
> Anyway, I'm not going to comment further.

(Noting I have yet to read the IRC discussion, and am tempted not to)

Except to say, it is REALLY **not good communication** to _use_ the '**use'** or **'include'** words in general text to imply syntax in posts, when you really mean `use<...>` or `include<...>`. It is not helping in comprehension, from back them to just now. 

Please _[use](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#quoting-code)_ markup to flag when you actually mean `use <>` or `include<>`.
SO you don't have to look it up, just utilise (see what I did there) a _back-tick_ to wrap the text  'use<>' or 'include<>' and it will be clear as `use<>` & `include<>`.

Communication is bad enough in disjointed text or Email posts. Haven't we all been in the communication business long enough? (Yes I am a pedantic bastard, someone needs to be)

pca006132:
Reading those discussion threads in full, I think my implementation here is flawed because it does not handle $variable from another `use<>`. I am closing this, wait for #3881, then decide if I should fix this work-around and work on a full fix that also deals with the re-evaluation issue.

nophead:
Not sure what you mean by $variables from another use<>?

I have just realised the last release has a an OPENSCAD_FONT_PATH environment variable, so I can make it work in Win11. So happy to stay with that for command line work. Might be odd if there is ever a new release to be saying people have to use the old one with NopSCADlib, but so be it.

My framework relies on use<>ing a file and invoking modules from it gets the same result as opening the file in the GUI and invoking the modules. That was true for about the first 11 years of OpenSCAD, but no longer true if it has any $variables at file scope just like the second official example and many others.

It seems very strange to me that the result of a function or module is different when you open it in the GUI, compared to use<>ing it. If I have a multifile project I often have subassemblies in separate files that get use<>d in main.scad. When I work on them individually I open them in the GUI to just view that part of the assembly. Odd that can give a more correct result now.

Also odd that a module can say $xyz is is undefined when it can be defined at file scope on the line before. And even odder that you can get around it being ignored by taking a copy in a normal file scope variable and using that in a module or function, which seems to break any arguments about dynamic scope only.

PR #4842, opened 2023-11-21
Indexed2 squash

gsohler:
Squashed commit of my indexed-point PR


src/core/primitives.cc
```diff
@@ -333,51 +300,49 @@ const Geometry *CylinderNode::createGeometry() const
...
       }
     }
   }
 
   if (this->r1 > 0) {
-    p->append_poly(fragments);
+    builder.append_poly(fragments);
     for (int i = 0; i < fragments; ++i)
-      p->insert_vertex(circle1[i].x, circle1[i].y, z1);
+      builder.prepend_vertex(builder.vertexIndex(Vector3d(circle1[i][0], circle1[i][1], z1)));
```
pca006132, line 336: why prepend here?

src/geometry/GeometryEvaluator.cc
```diff
@@ -1316,19 +1316,20 @@ static Geometry *rotatePolygon(const RotateExtrudeNode& node, const Polygon2d& p
...
-        ps->insert_vertex(rings[j % 2][(i + 1) % o.vertices.size()]);
-        ps->append_poly(3);
-        ps->insert_vertex(rings[j % 2][i]);
-        ps->insert_vertex(rings[(j + 1) % 2][i]);
-        ps->insert_vertex(rings[(j + 1) % 2][(i + 1) % o.vertices.size()]);
+        ind1=builder.vertexIndex(rings[j % 2][i]);
+        ind2=builder.vertexIndex(rings[(j + 1) % 2][(i + 1) % o.vertices.size()]);
+        ind3=builder.vertexIndex(rings[j % 2][(i + 1) % o.vertices.size()]);
+        builder.append_poly({ind3,ind2,ind1});
+        builder.append_poly(3);
```
pca006132, line 1323: why is there an `append_poly(3)` here?

src/geometry/PolySet.cc
```diff
@@ -184,14 +131,14 @@ void PolySet::quantizeVertices(std::vector<Vector3d> *pPointsOut)
 {
   Grid3d<unsigned int> grid(GRID_FINE);
   std::vector<unsigned int> indices; // Vertex indices in one polygon
-  for (auto iter = this->polygons.begin(); iter != this->polygons.end();) {
-    Polygon& p = *iter;
+  for (auto iter = this->indices.begin(); iter != this->indices.end();) {
+    IndexedFace& p = *iter;
     indices.resize(p.size());
```
pca006132, line 136: maybe we can change the name to avoid name conflict.

src/geometry/PolySet.cc
```diff
@@ -204,10 +151,16 @@ void PolySet::quantizeVertices(std::vector<Vector3d> *pPointsOut)
     p.erase(currp, p.end());
     if (p.size() < 3) {
       PRINTD("Removing collapsed polygon due to quantizing");
-      this->polygons.erase(iter);
+      this->indices.erase(iter);
```
pca006132, line 154: > Iterators, pointers and references pointing to position (or first) and beyond are invalidated

Perhaps the iterator is invalidated after this?

pca006132 (commented):
some comments

kintel:
I'll try to find time to contribute to this, just need to wrap up some of the balls I have in the air.

src/geometry/PolySetUtils.cc
```diff
@@ -90,15 +94,15 @@ void tessellate_faces(const PolySet& inps, PolySet& outps)
...
+
 
   for (const auto& faces : polygons) {
     if (faces[0].size() == 3) {
       // trivial case - triangles cannot be concave or have holes
-      outps.append_poly(3);
-      outps.append_vertex(verts[faces[0][0]]);
-      outps.append_vertex(verts[faces[0][1]]);
-      outps.append_vertex(verts[faces[0][2]]);
+       builder.append_poly({faces[0][0],faces[0][1],faces[0][2]});
```
pca006132, line 105: This function is *very problematic*. This will not preserve manifoldness because duplicated vertices are collapsed in the hash above, and added back without knowing any topological information.

pca006132:
I think the main problem is that while the `PolySetBuilder` code is convenient, using it means that you lose manifoldness, which is kind of embarrassing...

src/core/SurfaceNode.cc
```diff
@@ -282,33 +284,33 @@ const Geometry *SurfaceNode::createGeometry() const
...
+    builder.append_poly({ind1,ind2,ind3,ind4});
   }
 
   // the bottom of the shape (one less than the real minimum value), making it a solid volume
   if (columns > 1 && lines > 1) {
-    p->append_poly(2 * (columns - 1) + 2 * (lines - 1) );
+    builder.append_poly(2 * (columns - 1) + 2 * (lines - 1) );
     for (int i = 0; i < columns - 1; ++i)
-      p->insert_vertex(ox + i, oy + 0, min_val);
+      builder.prepend_vertex(builder.vertexIndex(Vector3d(ox + i, oy + 0, min_val)));
```
pca006132, line 304: 1. Maybe use `j` for iterator.
2. Should we share the vertex id when the coordinates are the same, e.g. `i = j`?

gsohler:
thank you for the code review.
Admittedly i am not as expert in c++ as you, but I am happy to implement your proposals, if it helps stability

pca006132:
btw, another question: do we really need polygonal mesh? if we only need to work with triangular mesh (or there is no lose of information by converting the polygonal mesh into triangular mesh), it will be better performance wise to store the indices as triples instead of vectors of ints.

nophead:
CGAL produces polygonal faces, not triangles, so I expect it was designed around that.

kintel:
CGAL is very good at maintaining polygons, even yielding polygons with holes when appropriate. This is quite useful for rendering polygon outlines. We don't currently support outline rendering for Manifold, so it hasn't really come up in that context.

@pca006132 I'm kind of reading between the lines that Manifold doesn't maintain polygons, but triangulates everything. Is that so?

pca006132:
yes, we triangulate everything.

kintel:
@gsohler Thanks for working on this! This is one of the dark corners of OpenSCAD with significant technical debt.

I have one big question to help drive the approach to improving this:

The PolySet class currently contains (relatively separate) functionality for 2D polygons and 3D meshes. This makes both this class, as well as client code, unnecessary complex. What if we, prior to improving the PolySet class to support indexed meshes, refactored this into two separate classes: `PolySet2d` and `PolySet3d` ?

I haven't really given it much though it such refactoring would be easier before or after the index stuff, but I generally tend to prefer refactoring early.

gsohler:
Hi Marius,
I totally agree that it would even better to separate the 2D and 3D stuff in PolySet, so after the separarion, PolySet2D and PolySet3D would be "Geomtry" and even dimension field would probably be obsolete.
My issue here is just that the indexed PR is quite progressed and with some luck the segfaults in windows disappear with the code review output from @pca006132 .
When we do another PR first to separate PolySet into 2D and 3D,  I fear that I will get lots of conflicts when rebasing the index after due to the PolySet->PolySet3D renaming.
I would volunteer to work on the PolySet -> PolySet2D and PolySet3D separation after. what do you think ?

kintel:
@gsohler Sounds good, let's try that. I might be able to take a crack at 2D/3D separation myself. I'll start reviewing this PR now.

src/geometry/PolySetBuilder.cc
```diff
@@ -0,0 +1,148 @@
...
+void PolySetBuilder::append_poly(int nvertices){
+  ps->indices.emplace_back().reserve(nvertices);
+}
+
+void PolySetBuilder::append_vertex(int ind){
+  ps->indices.back().push_back(ind);
+  ps->dirty = true;
+}
+
+void PolySetBuilder::prepend_vertex(int ind){
```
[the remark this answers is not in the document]
gsohler: append adds to the end , prepend adds to beginning. i did (simple) code translation, but did not change the method, how polygons are built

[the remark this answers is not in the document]
pca006132: I see. After reading this I revisited the original code and see that the old `insert_vertex` is indeed prepend... Thanks for clarifying.

src/core/SurfaceNode.cc
```diff
@@ -262,17 +263,18 @@ const Geometry *SurfaceNode::createGeometry() const
...
+    ind2=builder.vertexIndex(Vector3d(ox + 0, oy + i - 1, v1));
+    ind3=builder.vertexIndex(Vector3d(ox + 0, oy + i, v2));
+    ind4=builder.vertexIndex(Vector3d(ox + 0, oy + i, min_val));
+    builder.append_poly({ind1,ind2,ind3,ind4});
+
+    ind1=builder.vertexIndex(Vector3d(ox + columns - 1, oy + i, min_val));
+    ind2=builder.vertexIndex(Vector3d(ox + columns - 1, oy + i, v4));
+    ind3=builder.vertexIndex(Vector3d(ox + columns - 1, oy + i - 1, v3));
+    ind4=builder.vertexIndex(Vector3d(ox + columns - 1, oy + i - 1, min_val));
+    builder.append_poly({ind1,ind2,ind3,ind4});
```
kintel, line 277: Some whitespace is needed in various places. Could you try running our clang-format on this?
I cannot remember exactly how it works but ask in IRC :)

src/core/SurfaceNode.cc
```diff
@@ -234,25 +235,25 @@ const Geometry *SurfaceNode::createGeometry() const
...
-      p->append_vertex(ox + j - 0.5, oy + i - 0.5, vx);
-
-      p->append_poly(3);
-      p->append_vertex(ox + j - 1, oy + i, v3);
-      p->append_vertex(ox + j - 1, oy + i - 1, v1);
-      p->append_vertex(ox + j - 0.5, oy + i - 0.5, vx);
+      ind1=builder.vertexIndex(Vector3d(ox + j - 1, oy + i - 1, v1));
+      ind2=builder.vertexIndex(Vector3d(ox + j, oy + i - 1, v2));
+      ind3=builder.vertexIndex(Vector3d(ox + j - 0.5, oy + i - 0.5, vx));
+      builder.append_poly({ind1,ind2,ind3});
```
kintel, line 241: This feels like a task for `PolySetBuilder::append_poly(const std::vector<Vector3d> &v)`

src/geometry/PolySetBuilder.cc
```diff
@@ -0,0 +1,148 @@
...
+int PolySetBuilder::vertexIndex(const Vector3d &pt)
+{
+  return allVertices.lookup(pt);
+}
+
+void PolySetBuilder::append_poly(const std::vector<int> &inds)
+{
+  ps->indices.push_back(inds);        
+  ps->dirty = true;
+}
```
kintel, line 66: I know you inherited this from PolySet, but we should try to use either snake_case or camelCase, not both

src/geometry/PolySetBuilder.cc
```diff
@@ -0,0 +1,148 @@
...
+
+
+void PolySetBuilder::append_poly(const std::vector<Vector3d> &v)
+{
+  std::vector<int> inds;
+  for(const auto &pt: v)
+    inds.push_back(vertexIndex(pt));  
+  ps->indices.push_back(inds);        
+  ps->dirty = true;
```
kintel, line 106: Isn't the convention that you generate a valid PolySet in the `result()` function, so setting the PolySet dirty should perhaps be done there?

src/geometry/PolySetBuilder.cc
```diff
@@ -0,0 +1,148 @@
...
+  ps->polygon=std::move(pol);   
+  ps->dirty=false;
+}
+
+int PolySetBuilder::vertexIndex(const Vector3d &pt)
+{
+  return allVertices.lookup(pt);
+}
+
+void PolySetBuilder::append_poly(const std::vector<int> &inds)
```
kintel, line 62: I wonder if move semantics could make this more efficient, but we can revisit that later.

src/geometry/PolySetBuilder.cc
```diff
@@ -0,0 +1,148 @@
...
+  } else { // NOLINT(bugprone-branch-clone)
+    assert(false && "Not implemented");
+  }
+
+}
+
+
+void PolySetBuilder::append_poly(const std::vector<Vector3d> &v)
+{
+  std::vector<int> inds;
```
kintel, line 102: since you know the size, you could reserve here

src/geometry/PolySetBuilder.h
```diff
@@ -0,0 +1,28 @@
...
+
+#include <Reindexer.h>
+#include "Polygon2d.h"
+#include <memory>
+#include "boost-utils.h"
+class PolySet;
+
+class PolySetBuilder
+{
+  PolySet *ps;
```
kintel, line 11: Not sure about this yet, but it could be cleaner to construct the polygon in the `result()` function (and perhaps rename it to `build()`, and retain the necessary data structures in this class.

kintel (commented):
Left a bunch of comments. Haven't finished reading all the code yet.

Warning: A bunch of my comments are rather pedantic, and if it's too much of  a hassle to address the details right away, we could save that for a post-merge cleanup job.

src/geometry/PolySet.h
```diff
@@ -24,27 +27,17 @@ class PolySet : public Geometry
...
-  void append_poly(const Polygon& poly);
-  void append_vertex(double x, double y, double z = 0.0);
-  void append_vertex(const Vector3d& v);
-  void append_vertex(const Vector3f& v);
-  void insert_vertex(double x, double y, double z = 0.0);
-  void insert_vertex(const Vector3d& v);
-  void insert_vertex(const Vector3f& v);
-  void append(const PolySet& ps);
-
+  size_t numFacets() const override { return indices.size(); }
```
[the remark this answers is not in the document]
gsohler: hmm, its still called numFacets in Geometry.h

src/geometry/PolySet.cc
```diff
@@ -184,14 +131,14 @@ void PolySet::quantizeVertices(std::vector<Vector3d> *pPointsOut)
 {
   Grid3d<unsigned int> grid(GRID_FINE);
   std::vector<unsigned int> indices; // Vertex indices in one polygon
```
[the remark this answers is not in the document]
gsohler: not really sure whats the intent of this function, rather don't touch it.

src/geometry/PolySet.h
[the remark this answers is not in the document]
kintel: Right, let's ignore this for now then.

src/geometry/PolySet.cc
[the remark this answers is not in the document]
kintel: It's messy. I'll create a new ticket tracking PolySet cleanup tasks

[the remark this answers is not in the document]
kintel: How about this: For any of my comments which you think we should postpone, add to #4851 before resolving the comment?

src/geometry/PolySet.cc
```diff
@@ -204,10 +151,16 @@ void PolySet::quantizeVertices(std::vector<Vector3d> *pPointsOut)
...
       PRINTD("Removing collapsed polygon due to quantizing");
-      this->polygons.erase(iter);
+      this->indices.erase(iter);
     } else {
       iter++;
     }
   }
 }
 
+void PolySet::reset(const PolySet *ref)
```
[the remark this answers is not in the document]
gsohler: copy constructur  does not work. only this special sequence makes all ctest pass. did yet understand he reason.

src/geometry/PolySetUtils.cc
```diff
@@ -90,15 +94,15 @@ void tessellate_faces(const PolySet& inps, PolySet& outps)
 
   // Estimate how many polygons we will need and preallocate.
   // This is usually an undercount, but still prevents a lot of reallocations.
-  outps.polygons.reserve(polygons.size() );
+  PolySetBuilder builder(verts.size(),polygons.size());
+  for(int i=0;i<verts.size();i++)
+    builder.vertexIndex({verts[i][0],verts[i][1],verts[i][2]});
```
pca006132, line 99: thinking about it, this function is using reindexer twice, once for removing degenerates and once for the builder.

src/geometry/manifold/ManifoldGeometry.cc
```diff
@@ -76,17 +78,15 @@ std::string ManifoldGeometry::dump() const {
...
-  for (const auto &tv : mesh.triVerts) {
-    for (const int j : {0, 1, 2}) {
-      poly[j] = vector_convert<Vector3d>(mesh.vertPos[tv[j]]);
-    }
-    ps->append_poly(poly);
-  }
-  return ps;
+  std::vector<int> indmap;
+  indmap.reserve(mesh.vertPos.size());
+  PolySetBuilder builder(mesh.vertPos.size(),mesh.triVerts.size());
```
pca006132, line 84: You should not use the builder with manifold. It purposely duplicate the vertices to maintain manifoldness, so deduplicating them will just cause you more trouble. Also, the mesh data structure is very much like the indexed polyset, so there is no point hashing everything again...

src/geometry/manifold/manifoldutils.cc
```diff
@@ -53,11 +48,9 @@ std::shared_ptr<manifold::Manifold> trustedPolySetToManifold(const PolySet& ps)
   const auto vertexCount = mesh.vertPos.size();
   assert(indices.size() == numfaces * 4);
```
pca006132, line 49: this does not hold now.

src/io/import_stl.cc
```diff
@@ -145,10 +145,10 @@ PolySet *import_stl(const std::string& filename, const Location& loc) {
               boost::lexical_cast<double>(results[v + 1]);
           }
           if (++i == 3) {
-            p->append_poly(3);
-            p->append_vertex(vdata[0][0], vdata[0][1], vdata[0][2]);
-            p->append_vertex(vdata[1][0], vdata[1][1], vdata[1][2]);
-            p->append_vertex(vdata[2][0], vdata[2][1], vdata[2][2]);
+	    int ind[3];
```
pca006132, line 148: maybe remove this?

src/io/import_amf.cc
```diff
@@ -134,15 +135,15 @@ void AmfImporter::set_v3(AmfImporter *importer, const xmlChar *value)
...
 }
 
 void AmfImporter::end_object(AmfImporter *importer, const xmlChar *)
 {
   PRINTDB("AMF: add object %d", importer->polySets.size());
-  importer->polySets.push_back(importer->polySet);
+importer->polySets.push_back(importer->builder->build());
   importer->vertex_list.clear();
-  importer->polySet = nullptr;
+  importer->builder = nullptr;
```
pca006132, line 146: memory leak for builder

src/geometry/cgal/cgalutils.cc
```diff
@@ -383,13 +384,15 @@ bool createPolySetFromNefPolyhedron3(const CGAL::Nef_polyhedron_3<K>& N, PolySet
...
-    ps.append_vertex(verts[t[0]]);
-    ps.append_vertex(verts[t[1]]);
-    ps.append_vertex(verts[t[2]]);
+  PolySetBuilder builder(verts.size(),allTriangles.size());
+  for(int i=0;i<verts.size();i++)
+    builder.vertexIndex({verts[i][0],verts[i][1],verts[i][2]});
+
+  for(int i=0;i<allTriangles.size();i++)  {
+	const auto &tri=allTriangles[i];
+	builder.appendPoly({tri[0],tri[1],tri[2]});
```
pca006132, line 393: Builder may remove degenerates, so the index here may not be correct?
You need to store the indices in the loop above, e.g. `indMap[i] = builder.vertIndex ...`, and do `indMap[tri[0]]`...

pca006132 (commented):
some new comments

pca006132:
Btw, as we are already doing such a large refactor, does it make sense to use some modern smart pointers? It will be much easier to reason about memory safety if we use smart pointers rather than throwing raw pointers around. (especially when you have exception)

kintel:
> Btw, as we are already doing such a large refactor, does it make sense to use some modern smart pointers? It will be much easier to reason about memory safety if we use smart pointers rather than throwing raw pointers around. (especially when you have exception)

I have a large PR in the works which changes a lot of the current raw pointers to using smart pointers: #4857

src/io/import_stl.cc
```diff
@@ -87,16 +88,15 @@ PolySet *import_stl(const std::string& filename, const Location& loc) {
...
     f.read((char *)&facenum, sizeof(uint32_t));
 #if BOOST_ENDIAN_BIG_BYTE
     uint32_byte_swap(facenum);
 #endif
     if (file_size == static_cast<std::streamoff>(80ul + 4ul + 50ul * facenum)) {
       binary = true;
-      p->reserve(facenum);
     }
   }
+  PolySetBuilder builder(0,facenum);
```
pca006132, line 99: One bug: This `facenum` is only valid when binary is true.
This causes the `std::bad_alloc` because `facenum` is nonsense and too large.

src/geometry/PolySet.cc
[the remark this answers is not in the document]
kintel: That's probably a symptom of something we should fix. Let's look at it once everything else is resolved.

src/geometry/manifold/ManifoldGeometry.cc
```diff
@@ -76,17 +78,31 @@ std::string ManifoldGeometry::dump() const {
...
-  return ps;
+  ps->indices.reserve(mesh.triVerts.size());
+  for(int i=0;i<mesh.triVerts.size();i++) {
+	auto tri=mesh.triVerts[i];
+	std::vector<int> trinew({tri[0],tri[1],tri[2]});
+  	ps->indices.push_back(trinew);
+  }
+
+  return std::shared_ptr<PolySet>(ps);
+*/
```
pca006132, line 105: what is this commented code?

src/io/import_stl.cc
pca006132: This is exactly the bug causing the CI build to fail. You can fix by changing it to
```c++
if (!binary) facenum = 0;
PolySetBuilder builder(0,facenum);
```

gsohler:
just rebased to the latest master and its green again.
Are you happy with this ?
I'd really want to see openscad to have indexed points internally because the models are much less error prone to become not-water-tight and/or manifold.
Finally it makes the measurement feature at least 3x more performant(minimal number of faces in a corner)

kintel:
I think we're close. Could you go over the open code review conversation and close the ones which are addressed, comment on the ones you'd rather postpone (or add to #4851), then see if there is anything left to resolve before merging?

gsohler:
Thank you for the feedback. Went through the history again. This time I have seen  quite some more remarks which I literally did 
not notice the time before. Especially I got rid of the ind1-4 in various places, code looks both more readable and compact in
these places. Now most of the sections are either outdated, or resolved,
For most  of the topics scheduled for #4851 , I do either not get the idea or I am lacking internal knowledge how to handle it correctly(e.g dirty flag)
Let me know if anything else is missing.

src/geometry/manifold/ManifoldGeometry.cc
kintel: Could you add exactly that as a code comment, or remove it and add it to #4851 ?

kintel:
@gsohler IMO, if you kill or explain that piece of commented out code and I think we can merge this. @pca006132 any last comments?

pca006132:
I had no computer access for the past few days, will be able to do some additional reviews this afternoon (in a few hours).

I am not too concerned about this PR though, I think most changes are relatively straightforward, as long as tests passed it should be fine. Will need some follow-up mentioned in #4851 to make this more useful for preserving manifoldness.

pca006132:
Looks good. There are some cleanup needed but I think we can leave it for #4851.

gsohler:
@kintel, this piece of code was me trying to use the Manifold output as-is without reindexing them in OpenSCAD, but the ctests failed for unkown reason yet. its already scheduled for #4851.
I have just now removed the commented code entirely  and pushed again.

kintel (approved):

Issue #4039, opened 2022-01-07
ERROR: The given mesh is not closed! Unable to convert to CGAL_Nef_Polyhedron associated with linear_extrude


djwarn:
An openscad script I have developed gives the error "ERROR: The given mesh is not closed! Unable to convert to CGAL_Nef_Polyhedron" when rendered, but not in preview. I have minimised the code to the following, which still exhibits the problem:

```
a=110;  // twist, limited to 110 to avoid th strands touching
a2=150; // to set the start point such that the strand would 
        // end on the Z axis if a = 180
h=30;   // height
R=20;   // radius
r=2;    // strand radius

// strand 1
linear_extrude(height=h,convexity=10,twist=a,scale=[0,1])
  translate([R*sin(a2),R*cos(a2),0])circle(r);

// strand 2  
rotate(90,[0,0,1])
  linear_extrude(height=h,convexity=10,twist=a,scale=[0,1])
    translate([R*sin(a2),R*cos(a2),0])circle(r);
```

The above code will exhibit the issue, commenting either strand 1 or strand 2 will render as expected. The value of a is set to avoid the strands crossing in case this was the cause of the issue, as in issue #3805.

ChrisCoxArt:
The given script does not work, because RSin and RCos are not defined functions.
If you convert those to R*sin and R*cox, then the script works just fine, without error in OpenSCAD 2021.01

Please fix your script so that it works.
And tell us what OS and what version of OpenSCAD you are using.

ChrisCoxArt:
Looks like your error is in the scale parameter to linear_extrude, resulting in infinitely thin lines at the top end of the extrusion, which results in a lot of degenerate triangles, which can't close the mesh.
Change the [0,1] to [0.01,1] so they have some width to create geometry with.

djwarn:
Thanks Chris. My code did have the R*sin etc. the '*' somehow got lost in the copy and paste. I confirm that the change to [0.01,1] fixes the issue. Thanks again for your help.

thehans:
> Edit: Ah, I see the asterisk disappears from the post. I am using Intel 64 bit 2021.01

Asterisk and other special characters are interpreted as markdown (*italics* for asterisk) in github comments.  To post multiple lines of code, you can put three backquotes on a line by themselves, immediately before and after your code.  That way nothing between those lines will be interpreted as markdown.  (That is why ChrisCoxArt's code sample displayed inside a gray block with monospaced font.)

[...]

PR #4855, opened 2023-11-24
try harder to import stl

pca006132:
This enables importing some non-manifold stl into mesh by attempting to do vertex duplication etc.
This polygon soup repair will create self-intersections because they require duplicating vertices, but this is fine for manifold because they are still ε-valid for some ε. Not sure if this will work for fastcsg as the resulting mesh has self-intersections. For Nef-polyhedron it may work, but I am not sure how to do this for Nef-polyhedron so I did not bother implementing it.

Tested with simple example such as two touching cubes (this stl is non-manifold, even if you manage to export it precisely using infinite precision, because they are indeed touching), and the `Proximity_Sensor_D18_Holder_fixed.stl` in #4039. Here is an example:

image


kintel (approved):
Looks very reasonable to me.

pca006132:
I guess we can merge this?

PR #4867, opened 2023-12-01
polyset cleanup 2

pca006132:
1. Fix `ManifoldGeometry` -> ` PolySet`.
2. Faster STL export. 4x faster for ASCII STL, slightly faster for binary, slower than #4643 because the tessellation is slow (will be addressed later).


src/geometry/PolySetBuilder.cc
```diff
@@ -60,7 +60,8 @@ int PolySetBuilder::vertexIndex(const Vector3d &pt)
 
 void PolySetBuilder::appendPoly(const std::vector<int> &inds)
 {
-  ps->indices.push_back(inds);        
+  auto& face = ps->indices.emplace_back();
+  face.insert(face.begin(), inds.begin(), inds.end());
```
kintel, line 64: This is probably food for another round of refactoring, but I've been wondering if we should start using spans for use-cases such as this. `std::span` is available since C++20, but the compatible `boost::span` has been in Boost since 1.78 (December 2021). Not sure if we can require that quite yet, but could be worth figuring out.

kintel: It's more a question of whether all target distros have boost-1.78 available..

pca006132:
btw, 44f4571 is not required for merging. I did some simple benchmark and did not see much of a difference in terms of performance, probably due to the use of mimalloc. It is more of an experiment.

src/io/export_stl.cc
```diff
@@ -39,121 +40,142 @@
...
 }
 
 
-uint64_t append_stl(const PolySet& ps, std::ostream& output, bool binary)
+uint64_t append_stl(const PolySet& polyset, std::ostream& output, bool binary)
 {
+  static_assert(sizeof(float) == 4, "Need 32 bit float");
+  // check if tessellation is needed
+  PolySet tmp(3);
+  bool needTessellation = !std::all_of(
```
kintel, line 105: This is another refactoring opportunity, let the PolySet know if it's a triangle mesh or a polygon mesh. ..either through a flag or template args or smth.

pca006132: I also found the API a bit unwieldy. There is no point in passing the result as the second argument, we can just return a `std::shared_ptr<const PolySet>` or something. This way when the input is already a triangular mesh, we can just return the input.

kintel: The second argument issue is resolved in my smart pointer PR

kintel:
> btw, [44f4571](https://github.com/openscad/openscad/commit/44f457161a19d3a85a277c4c45bed795addfd1dc) is not required for merging. 

I think it makes sense, as it makes the intend of the code clearer

src/io/export_stl.cc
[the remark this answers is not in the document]
pca006132: ah yes, forgot this

kintel (approved):

kintel:
Can merge once green

kintel:
@pca006132 Turns out that the change from `toString(Vector3d)` to `toString(Vector3f)` causes [this assert](https://github.com/openscad/openscad/blob/dbd7d4f3d93356fedc0cfac1989e2a1e85335de8/src/io/export_stl.cc#L158-L161) to fail. Not ideal, since we specifically remove coincident vertices during tessellation. We could disable the assert and allow writing coincident vertices to STL, but it might be better to go back and perform tessellation in the space in which we intend to export. ..or revert back to double.

Any ideas?

pca006132:
Revert back to double seems to be the simplest option. Though I am not sure if STL allows that much precision, apparently the binary format only supports single precision.

PR #4874, opened 2023-12-03
Polyset cleanup 3

pca006132:
1. Removed IndexedMesh.
2. Removed usage of Reindexer in tessellator. Hopefully this helps preserving manifoldness in some cases, but there is no guarantee. This also improves performance for tessellation.


src/geometry/PolySet.h
```diff
@@ -37,6 +37,7 @@ class PolySet : public Geometry
 
   bool is_convex() const;
   boost::tribool convexValue() const { return this->convex; }
+  bool isTriangular = false;
```
kintel, line 40: Should probably be a private member with a const accessor

src/geometry/PolySetUtils.cc
```diff
@@ -55,76 +55,87 @@ std::unique_ptr<Polygon2d> project(const PolySet& ps) {
...
 {
   int degeneratePolygons = 0;
+  auto result = std::make_unique<PolySet>(3, polyset.convexValue());
+  result->setConvexity(polyset.getConvexity());
+  result->isTriangular = true;
+  // ideally this should not require a copy...
+  if (polyset.isTriangular) {
+    result->vertices = polyset.vertices;
+    result->indices = polyset.indices;
+    return result;
```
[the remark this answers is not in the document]
pca006132: (this is also why I wanted to just kill fastcsg)

src/geometry/PolySet.h
kintel: I see what you mean. Let's keep this for now. This is really just another argument for making triangle meshes a separate type.

kintel (approved):

PR #4891, opened 2023-12-09
Ignore \r characters.  Fixes #4882.

### LINKED ISSUE #4882: Lexer counts lines incorrectly in CRLF files on LF/NL-only systems
**Describe the bug**
When a file has lines ending with CRLF (\r\n, 0D 0A) on a newline (LF)-only system like Linux, the lexer counts a line each for the CR and the LF.  Except that CR in a comment isn't counted, so line numbers aren't simply doubled.

**To Reproduce**
* Create a `.scad` file with lines that end with CRLF, that generates warnings or errors.
* Run it on a Linux system.
* Observe that the line numbers reported do not match the source.

**Expected behavior**
Errors and warnings should report the correct line.

jordanbrown0:
For some purposes, it was counting \r as starting a new line, so \r\n (notably on Linux) would be counted as starting two lines.  Just always ignore \r.

Note:  On Linux, lines are normally terminated by \n, with no \r.  On Windows, they are normally terminated by \r\n.  You can get \r\n files on Linux if you've moved them from Windows without converting.

There should maybe be a test that checks line numbers of error messages, but to test this \r case would require a \r\n test file on Linux, and that would be fragile - git would sometimes try to "fix" it depending on where it was changed.

lexer.l is inconsistent in its use of tabs versus spaces, and uses 4-space tabs so does not display properly by default in most UIs.  I want to clean that up, one way or another, but I'll leave that to another PR.

### DIFF
--- a/src/core/lexer.l
+++ b/src/core/lexer.l
@@ -148,7 +148,8 @@ LOCATION_NEXT(parserlloc);
 
 include[ \t\r\n]*"<"	{ BEGIN(cond_include); filepath = filename = ""; LOCATION_COUNT_LINES(parserlloc, yytext); }
 <cond_include>{
-[\n\r]					{
+\r						/* ignored */
+\n						{
 							LOCATION_ADD_LINES(parserlloc, yyleng);
 							// see merge request #4221
 							LOG(message_group::Warning,LOCATION(parserlloc),"","new lines in 'include<>'-statement is not defined - behavior may change in the future");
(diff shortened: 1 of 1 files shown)


jordanbrown0:
Please review and (assuming it's OK) merge #4892 first, and then I'll resync this one with that.

jordanbrown0:
Withdrawn in favor of #4895 .

PR #4895, opened 2023-12-10
Uniform support for \n, \r\n, \r as line breaks.  Fixes #4882.

### LINKED ISSUE #4882: Lexer counts lines incorrectly in CRLF files on LF/NL-only systems

jordanbrown0:
Fixes lexer to count lines properly in files with \n, \r\n, and \r line breaks.  (\r line breaks are used only in some old or obscure systems, but what the heck.)

Tweaks existing line number counting for (bad but legacy) include/use cases:
```
include <
line2>
```
reports a warning.  Old behavior was to report it on line 2, but really the error (the unexpected end-of-line) is on line 1 so the new behavior (and tweak to existing test) makes it report line 1.

Adds a new `RETVAL` argument to `add_cmdline_test()` to allow tests where execution is expected to fail to still get expected-output comparison, to check that the right errors (and line numbers!) are reported.  We should probably eventually move the existing add_failing_test cases to use this mechanism instead, so that they can check the error reported.

Adds a gitattributes file that suppresses text processing for the new `*.bin.scad` test files.  Those files have sections that are UNIX-style, sections that are Windows-style, and an obscure \r-only section.  In theory I think git will preserve them, but they may be too fragile for long-term use.  We'll see.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.bin.scad -text diff
--- a/src/core/lexer.l
+++ b/src/core/lexer.l
@@ -83,21 +83,29 @@ extern SourceFile *rootfile;
   Handle locations.
   Since flex doesn't handle column numbers, we deal with those manually.
   See "Advanced Use of Flex" / "Advanced Use of Bison"
+  https://www.lrde.epita.fr/~akim/ccmp/doc/gnuprog2/Advanced-Use-of-Flex.html
+  https://www.lrde.epita.fr/~akim/ccmp/doc/gnuprog2/Advanced-Use-of-Bison.html
 */
 extern YYLTYPE parserlloc;
 
 #define LOCATION(loc) Location(loc.first_line, loc.first_column, loc.last_line, loc.last_column, sourcefile())
 #define LOCATION_INIT(loc) do { (loc).first_line = (loc).first_column = (loc).last_line = (loc).last_column = yylineno = 1; } while (0)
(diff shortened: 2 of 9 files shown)


jordanbrown0:
OBTW, I cloned a copy to a new directory, and the `*.bin.scad` files seemed to survive.

jordanbrown0:
Sigh.  Looks like I fooled myself into thinking it fixed the problem, and need to fire up the Linux VM and check it out there.

jordanbrown0:
I found the (stupid) mistake, and fixed it.

One of the tests currently doesn't pass because it compares output, and the output includes a fully qualified path to the test `.scad` file, and of course those paths are different for different build environments.

t-paul:
Looking at this test fail, I wonder if we could just remove that log output. There's an error log from the parser `yyerror` before already.

jordanbrown0:
I’m fine with removing that log message too.

jordanbrown0:
Still on my to-do list.

PR #4904, opened 2023-12-27
Changes to allow building with Qt 6

cjmayo:
With Qt5Compat and without gamepad support.

and Fix slots not matching signals with Qt 6

---

These changes allow me to do a non-experimental build and run OpenSCAD on Qt 6 (and still with Qt 5 too).

N.B. I am really a Python person so do please check carefully - especially the use of QSoundEffect, and I kinda guessed what the return value of `ScadLexer2::description()` should be.

### DIFF
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -63,6 +63,9 @@
 #ifdef Q_OS_MAC
 #include "CocoaUtils.h"
 #endif
+#ifdef Q_OS_WIN
+#include <QScreen>
+#endif
 #include "PlatformUtils.h"
 #ifdef OPENSCAD_UPDATER
 #include "AutoUpdater.h"
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -679,8 +679,8 @@ MainWindow::MainWindow(const QStringList& filenames)
     // again.
     // On Windows that causes the main window to open in a not
     // easily reachable place.
-    auto desktop = QApplication::desktop();
-    auto desktopRect = desktop->frameGeometry().adjusted(250, 150, -250, -150).normalized();
+    auto primaryScreen = QApplication::primaryScreen();
+    auto desktopRect = primaryScreen->availableGeometry().adjusted(250, 150, -250, -150).normalized();
     auto windowRect = frameGeometry();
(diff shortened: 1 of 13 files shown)


src/gui/MainWindow.cc
```diff
@@ -91,13 +91,11 @@
 #include <QTemporaryFile>
 #include <QDockWidget>
 #include <QClipboard>
-#include <QDesktopWidget>
```
t-paul, line 94: This is needed for Windows builds.

t-paul (changes requested):
Nice, that looks very promising. The build error on Windows should be easy to resolve by getting the `include` back. If this can provide compatibility with both Qt5 and Qt6 without much trickery that would be great.

I'm trying to get the snapshot builds back, for some reason those did not trigger for the PR.

t-paul:
CircleCI workflow running at https://app.circleci.com/pipelines/github/openscad/openscad/4450/workflows/6ff4d24a-aae3-48fd-afb5-73ba065e353f

They will likely all show up as fail due to the manual start, but hopefully do the build successfully.

src/gui/ScintillaEditor.cc
```diff
@@ -6,6 +6,8 @@
 #include <boost/property_tree/json_parser.hpp>
 #include <QString>
 #include <QChar>
+#include <QRegExp>
```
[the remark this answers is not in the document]
t-paul: I see, no problem having it if that reduces the needed changes.

src/gui/MainWindow.cc
t-paul: Looks promising, according to the docs `QScreen` is available since 5.0, so that should work for both versions too.

t-paul (approved):
If the new QtGamepad is simpler to setup it might even become usable. I never got it to work well anyway, so lets just wait and see how it will look like if it gets into Qt6.

kintel:
@cjmayo Out of curiosity, do you have a grander plan/idea/wish for your Qt6 work? I was thinking about starting a tracking ticket for eventually moving to Qt6 as the default, but just wanted to hear about your current ideas.

kintel:
If you could share the files, that would be awesome. I'll open a ticket and tag you, feel free to dump any knowledge, ideas or experience into that ticket

PR #4908, opened 2023-12-31
Test building with Qt 6

cjmayo:
Builds master on Linux with:

-DCLANG_TIDY=OFF -DENABLE_CAIRO=yes -DENABLE_EGL=ON -DENABLE_GLX=OFF -DENABLE_SPNAV=yes -DENABLE_TESTS=no -DEXPERIMENTAL=no -DHEADLESS=no -DUSE_CCACHE=OFF -DUSE_GLAD=ON -DUSE_MIMALLOC=OFF -DENABLE_GAMEPAD=no -DENABLE_QTDBUS=yes

Draft because Qt6Gamepad doesn't exist, and I guess not everyone wants to switch right now.


t-paul:
We will need to support both Qt5 and Qt6 for a while anyway, at least for the Linux builds. Qt6-only will be a showstopper for the next couple of years or we would have to drop support for things like AppImages which seem to be quite popular.

I'll try to see if we can get the scintilla package for Ubuntu built via OBS, we have to do that for other libraries too.

I'm not sure Homebrew is a problem, as we have to custom build all the macOS libraries anyway. So the effort here is to figure out how and when to do that. On the plus side, it should be fine here to "just" switch over to Qt6 at some point.

kintel:
@cjmayo also, just FYI: The tests run here don't exercise any Qt code paths, but they are good for validating builds.

kintel:
For macOS, let's start by building our own Qt6 and test based on that. We do such builds on CircleCI, but we don't currently run tests there.

Neumann-A:
Do I read this PR correctly that only CMake changes are required to build OpenSCAD with Qt6? If so, I might try building it with vcpkg on windows.

kintel:
> With qscintilla2 on Qt 6, Qt 5 builds are broken

Argh! I'll take a look at that.

kintel:
For now, I added a request here: https://github.com/Homebrew/homebrew-core/pull/166414#issuecomment-2005226871 - let's see where that goes.

kintel:
Temporary workaround to downgrade QScintilla to the qt5 version: #5059

kintel:
We should support both Qt5 and Qt6 builds on the CI, subject to CI capacity.
See #5129 for WIP to make that work on macOS.

kintel:
If you could open a ticket for the gcc14 issue that would be cool.

The other tests look like flakes - have you tried just rerunning?

kintel:
Could be that clipper isn't updated past that fix in Manifold yet, plus we haven't pulled Manifold in a while.

t-paul:
We should be able to use the OBS packages to override that - https://build.opensuse.org/project/show/home:t-paul:polyclipping2 - right now that just builds `main`.

t-paul:
Nice! :tada:

PR #4913, opened 2024-01-01
Stop using + operator when creating QKeySequence

cjmayo:
Dropped in Qt 6:

 D:/a/openscad/openscad/src/gui/MainWindow.cc:457: error: use of deleted
function 'constexpr void Qt::operator+(QFlags<Modifier>::enum_type, QFlags<Modifier>::enum_type)'

---

Compiles on Qt 6, let's see if it does so on Qt 5. Haven't run it (MS Windows only).


t-paul (approved):
Thanks, looks good and Qt5 documentation shows the matching constructor too.

PR #4916, opened 2024-01-02
Measure, fix segfault when compile  not successful

gsohler:


t-paul (approved):

tpimh:
Just tested. As expected, https://github.com/openscad/openscad/pull/4813#issuecomment-1873851091 is no longer reproducible. Thanks for a quick fix!

Issue #4909, opened 2023-12-31
Upgrade to Qt6


kintel:
This is a tracking issue for all work needed or wanted to be able to move OpenSCAD to Qt6.

TODO:
* [ ] Create a build-time Qt selection CMake flag
* [ ] First working Qt6 build on Linux
* [ ] First working Qt6 build on macOS
* [ ] First working Qt6 build on Windows
* [ ] Linux AppImage snapshot build
* [ ] Windows snapshot build
* [ ] macOS snapshot build
* [ ] Other Linux builds, including distro packages
* [ ] Go through and manually test each Qt-based feature on each target platform. This can be a bit of work, so we might want to structure this as checklist or smth.
* [ ] Update website to offer direct download of Qt6 builds
* [ ] Game pad support
* [ ] Remove (or hide) Qt5 downloads, and make Qt6 downloads the default

kintel:
@cjmayo FYI

kintel:
Changes to `CMakeList.txt` and `FindQt6QScintilla.cmake` to make OpenSCAD build on Qt6: #4908

Scopeuk:
As a note on the qtgame pad it looks like the maintainer decided it should be re-implimented at a lower level during the 5->6 jump but other things took precedence (https://www.reddit.com/r/QtFramework/comments/13uperb/comment/jm5y9yc/ ), another developer has done a straight port at https://github.com/pumphaus/qtgamepadlegacy which could be a work arround short term

AaronVerDow:
Qt6 package for Nix: https://gist.github.com/AaronVerDow/b945a96dbcf35edfc13f543662966534

coryrc:
Is the AI to update the build system for:

* Linux AppImage snapshot build
* Linux Flatpak snapshot build

to build with Qt6? Can we just stop doing Qt5 or do you need both being built?

I'm also assuming:

*  Linux distro packages as applicable

means going through the packages per distro and adding Qt6 ones and removing Qt5 ones?

t-paul:
> to build with Qt6? Can we just stop doing Qt5 or do you need both being built?

We can drop Qt5.

The changes for Qt6 in AppImage are ready, just waiting for the bigger CI changes are merged.

Note that Flatpak would currently apply to the `beta` branch https://github.com/flathub/org.openscad.OpenSCAD/tree/beta

> Linux distro packages as applicable

[...]

PR #4929, opened 2024-01-08
Add the option to build with Qt 6

cjmayo:
I've tried to do this in the way that makes it easiest to remove Qt 5 down the line.
For now there is no gamepad support with Qt 6.


Neumann-A:
You probably want to avoid the whole code duplication and rather have a variable called `OPENSCAD_QT_VERSION` which can be either set to 5 or 6 (or auto). See e.g. https://gitlab.kitware.com/vtk/vtk/-/blob/master/CMake/vtkQt.cmake?ref_type=heads

CMakeLists.txt
```diff
@@ -513,11 +516,20 @@ if(NOT HEADLESS)
...
+    if (APPLE AND EXISTS /opt/homebrew/opt/qt@6)
+      list(APPEND CMAKE_PREFIX_PATH "/opt/homebrew/opt/qt@6")
+    endif()
+  else()
+    if (APPLE AND EXISTS /usr/local/opt/qt@5)
+      list(APPEND CMAKE_PREFIX_PATH "/usr/local/opt/qt@5")
+    endif()
+    if (APPLE AND EXISTS /opt/homebrew/opt/qt@5)
+      list(APPEND CMAKE_PREFIX_PATH "/opt/homebrew/opt/qt@5")
+    endif()
```
kintel, line 532: This is getting a bit messy - do you think there's a way of moving this out of the main CMakeLists file?

Same goes for the Qt5 vs. Qt6 logic. Perhaps a separate macro which deals with the details without making the main CMakeLists file so messy?

kintel:
FYI: I managed to build this on macOS - looks good so far

kintel:
@cjmayo If you rebase this to master, it should build on macOS too.

CMakeLists.txt
kintel: ok, let's defer this until we have both Qt5+6 working and have a better picture

kintel:
We can probably merge this soon. I'll do a more proper review, and update #4909 accordingly

kintel (approved):

PR #4930, opened 2024-01-08
Fix building with tbb when CUDA is not installed

cjmayo:
The default THRUST_DEVICE_SYSTEM is THRUST_DEVICE_SYSTEM_CUDA, causing the build to fail when looking for cuda_runtime_api.h.

Once this is set to THRUST_DEVICE_SYSTEM_TBB the build fails on linking.

----

Perhaps some people want to use CUDA for this??? In which case it looks like there is an implicit dependency here on CUDA and a missing option for those of us without NVIDIA hardware.

The device systems supported by Thrust are:

```c
#define THRUST_DEVICE_SYSTEM_CUDA    1
#define THRUST_DEVICE_SYSTEM_OMP     2
#define THRUST_DEVICE_SYSTEM_TBB     3
#define THRUST_DEVICE_SYSTEM_CPP     4
```
I've somewhat guessed here, probably don't want OpenMP as another dependency.

I'm using CUDA C++ Core Libraries (CCCL) 2.2.0. It contains Thrust's dependencies and the Thrust project is now archived.


pca006132:
The CUDA dependency is just due to thrust being some Nvidia library and has a CUDA backend (that manifold do not use). I haven't checked the thrust version with CCCL yet, I guess it is a version issue... We will automatically pull thrust from the archived repo when thrust is not found on the machine.

pca006132:
Btw, does it work with manifold alone? Not sure if it is something specific to openscad or an upstream issue.

pca006132:
The linker issue on macos is weird. With manifold we are linking against either `TBB::tbb` or `${TBB_LINK_LIBRARIES}` if the target `TBB::tbb` is not defined. Not sure if that can fix the issue.

CMakeLists.txt
```diff
@@ -857,7 +857,10 @@ if(EXPERIMENTAL AND ENABLE_TBB)
   target_compile_options(OpenSCAD PRIVATE 
     -DENABLE_TBB
     -DTHRUST_HOST_SYSTEM=THRUST_HOST_SYSTEM_TBB
+    -DTHRUST_DEVICE_SYSTEM=THRUST_DEVICE_SYSTEM_TBB
   )
+  pkg_check_modules(TBB tbb REQUIRED)
+  target_link_libraries(OpenSCAD PRIVATE ${TBB_LIBRARIES})
```
kintel, line 863: You should probably do this instead:
```
find_package(TBB REQUIRED)
target_link_libraries(OpenSCAD PRIVATE TBB::tbb)
```

It sounds like you want to patch Manifold, not OpenSCAD though.  Not sure how good it is to have both Manifold and OpenSCAD link TBB in their own way.

pca006132: I think this is not patching manifold: openscad does use thrust for parallel minkowski. As long as the tbb version is the same, I think linking tbb in both should work fine.

kintel: Oh, I hadn't realized; minkowski for Manifold does indeed use TBB.

pca006132 (approved):
LGTM

t-paul (approved):

Issue #4851, opened 2023-11-23
PolySet needs cleanup


kintel:
@gsohler is working on making PolySet support indexed polygons, and during review of their work some messy areas were highlighted. This ticket tracks mess we should clean up:

* [x] [@kintel] Separate 2D and 3D parts into separate classes. -> Opted to remove the 2D features from PolySet: #4933
   * The main purpose of the 2D code is to support rendering, so we could move those parts to rendering-specific helpers.
   * Perhaps postpone until #4782 is done, as there is a lot of duplicated rendering code.
* [x] [@kintel] Get rid of `void PolySet::reset(const PolySet *ref)` (#4857)
* [x] [@kintel] Consider push_back() instead of  emplace_back() for points, e.g. `outline.vertices.emplace_back(point[0], point[1]);` in primitives.cc
* [x] [@kintel] PolySetBuilder (#4866)
  * [x] Consider constructing the polygon in the result() function.
  * [x] Use shared_ptr/unique_ptr for the PolySet pointer to clarify ownership.
* [x] [@kintel] Look into cases where we can directly construct a PolySet without using PolySetBuilder, so we don't have to pay for the reindexer. (WIP: #4937)
* [x] import_amf: Fix memory leak for builder
* [x] [@pca006132] Faster STL export (#4643)
* [x] [@pca006132] (#4867) ManifoldGeometry: You should not use the builder with manifold. It purposely duplicate the vertices to maintain manifoldness, so deduplicating them will just cause you more trouble. Also, the mesh data structure is very much like the indexed polyset, so there is no point hashing everything again...
* [x] [@pca006132] (#4874) In PolySetUtils::tessellate_faces: void tessellate_faces(const PolySet& inps, PolySet& outps):
   * This function is using reindexer twice, once for removing degenerates and once for the builder.
* [x] [@kintel] integrate `ExportMesh` into `PolySet`. (#4949)
* [x] Output PolySet directly for `extrudePolygon` so they will not output something beyond repair (#4039): Deferred to #4970
* [x] import_3mf should preserve manifoldness: Deferred to #4971
* [x] quantizeVertices(): Deferred to #4969

pca006132:
and maybe we can remove the old `polyset-utils-old.cc`?

PR #4933, opened 2024-01-11
PolySet Refactor: Remove Polygon2d features from PolySet

kintel:
This refactors a lot of code dealing with rendering 2D polygons:
* Completely removes `Polygon2d` features from `PolySet`
* Changes rendering to convert 2D data into a "2D-in-3D" PolySet (z==0), instead of carrying the polygon2d all the way to the renderer
* Note: Changes in test golden images are mostly related to CSG bbox being slightly different since the generated 2D PolySet is 1 unit thick, causing camera position to slightly change for "View All"

Related to #4851


gsohler:
This is great!
Did you also consider to add 1D Data ?
A function which defines a radius depending on an angle could polar_extrude to a 2D shape ?

kintel:
I haven't considered any new use-cases yet - this is really just refactoring to clean up existing code, which, hopefully, should make it easier to add new geometry in the future. Still a lot more to clean up though..

PR #4939, opened 2024-01-19
fixed some compiler warnings

gsohler:
Fixed some compiler warnings and removed unused stuff.

### DIFF
--- a/src/io/export_stl.cc
+++ b/src/io/export_stl.cc
@@ -118,7 +118,7 @@ uint64_t append_stl(const PolySet& polyset, std::ostream& output, bool binary)
   if (!binary) {
     vertexStrings.resize(ps.vertices.size());
     std::transform(ps.vertices.begin(), ps.vertices.end(), vertexStrings.begin(),
-      [](const auto& p) { return toString({p.x(), p.y(), p.z()}); });
+      [](const auto& p) { return toString({(float) p.x(), (float) p.y(),(float)  p.z()}); });
   }
 
   // Used for binary mode only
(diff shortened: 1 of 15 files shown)


src/io/export_stl.cc
```diff
@@ -118,7 +118,7 @@ uint64_t append_stl(const PolySet& polyset, std::ostream& output, bool binary)
   if (!binary) {
     vertexStrings.resize(ps.vertices.size());
     std::transform(ps.vertices.begin(), ps.vertices.end(), vertexStrings.begin(),
-      [](const auto& p) { return toString({p.x(), p.y(), p.z()}); });
+      [](const auto& p) { return toString({(float) p.x(), (float) p.y(),(float)  p.z()}); });
```
kintel, line 121: Instead of using C-style casts, I'd prefer using a C++ cast, e.g. `const auto f = p.cast<float>()`, then use the resulting float vector as arguments to `toString()`
```

File: src/io/import_stl.cc [hunk 2]
```diff
@@ -53,7 +53,7 @@ static void uint32_byte_swap(uint32_t& x) {
 
 static void read_stl_facet(std::ifstream& f, stl_facet& facet) {
   f.read((char *)facet.data8, STL_FACET_NUMBYTES);
-  if (f.gcount() < STL_FACET_NUMBYTES) {
+  if ((unsigned long int) f.gcount() < STL_FACET_NUMBYTES) {
```
kintel, line 56: `static_cast<uint32_t>(f.gcount())`

src/io/import_stl.cc
```diff
@@ -53,7 +53,7 @@ static void uint32_byte_swap(uint32_t& x) {
 
 static void read_stl_facet(std::ifstream& f, stl_facet& facet) {
   f.read((char *)facet.data8, STL_FACET_NUMBYTES);
-  if ((unsigned long int) f.gcount() < STL_FACET_NUMBYTES) {
+  if (static_cast<unsigned long int >(f.gcount()) < STL_FACET_NUMBYTES) {
```
kintel, line 56: Just use `size_t` since that's what STL_FACET_NUMBYTES is defined as. Introducing a new type is just confusing.

PR #4944, opened 2024-01-20
Clean up Toolbar Export 3D/2D preferences

jordanbrown0:
Remove duplicate (and non-operative) Toolbar Export 3D/2D options from "Console" section.
Move working Toolbar Export 3D/2D options from "User Interface" to "Export Features".

Before (note that bottom one does not match toolbar buttons):
image

After:
image

Note:  There's one section of the file that's just deleted; there's another that's moved unchanged from one place to another.

### DIFF
--- a/src/gui/Preferences.ui
+++ b/src/gui/Preferences.ui
@@ -1910,60 +1910,6 @@
                  </item>
                 </layout>
                </item>
-               <item>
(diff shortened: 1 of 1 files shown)


t-paul (approved):
:+1:

PR #4946, opened 2024-01-21
Make customizer font and size selectable.

jordanbrown0:
This (probably) isn't ready to be merged.

It adds entries to Preferences/Advanced allowing control of the font and size used in the customizer data pane.

It mostly works, but:
* Some aspects are stretching my knowledge of C++ and there might be better answers.
* The text inside spinboxes resizes dynamically, but the box around the text doesn't.

I considered trying to control the stuff *above* the data pane, the preset control stuff, and decided that that should be part of a more general "UI font size" control that would affect all text in forms, menus, et cetera.

Review comments solicited.  I'll write something up tomorrow about the areas that I'm not happy with, but right now it's bedtime.

### DIFF
--- a/src/gui/Preferences.cc
+++ b/src/gui/Preferences.cc
@@ -88,33 +88,19 @@ void Preferences::init() {
   this->defaultmap["advanced/consoleFontFamily"] = font2.family();
   this->defaultmap["advanced/consoleFontSize"] = font2.pointSize();
 
+  // Leave Customizer font with default if user has not chosen another.
+  this->defaultmap["advanced/customizerFontFamily"] = font2.family();
+  this->defaultmap["advanced/customizerFontSize"] = font2.pointSize();
+
 #if defined (Q_OS_MAC)
   this->defaultmap["editor/ctrlmousewheelzoom"] = false;
 #else
   this->defaultmap["editor/ctrlmousewheelzoom"] = true;
 #endif
 
-  uint savedsize = getValue("editor/fontsize").toUInt();
-  const QFontDatabase db;
-  BlockSignals<QComboBox *> fontSize{this->fontSize};
-  for (auto size : db.standardSizes()) {
-    fontSize->addItem(QString::number(size));
-    if (static_cast<uint>(size) == savedsize) {
-      fontSize->setCurrentIndex(this->fontSize->count() - 1);
-    }
-  }
-  // reset GUI fontsize if fontSize->addItem emitted signals that changed it.
-  this->fontSize->setEditText(QString("%1").arg(savedsize) );
-
-  uint consavedsize = getValue("advanced/consoleFontSize").toUInt();
-  BlockSignals<QComboBox *> consoleFontSize{this->consoleFontSize};
-  for (auto size : db.standardSizes()) {
-    consoleFontSize->addItem(QString::number(size));
-    if (static_cast<uint>(size) == consavedsize) {
-      consoleFontSize->setCurrentIndex(this->consoleFontSize->count() - 1);
-    }
-  }
-  this->consoleFontSize->setEditText(QString("%1").arg(consavedsize) );
+  createFontSizeMenu(fontSize, "editor/fontsize");
+  createFontSizeMenu(consoleFontSize, "advanced/consoleFontSize");
+  createFontSizeMenu(customizerFontSize, "advanced/customizerFontSize");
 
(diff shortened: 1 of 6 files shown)


jordanbrown0:
No time yet to really write something, but a quick stab at the things that made me unhappy:

* Needing (wanting) a separate mySetFont() for the various widgets, versus just overloading setFont().  Qt's setFont is not marked virtual, so you can't exactly override it.  You can create a new one, but then which one you get depends on the type you're using to make the call, whether it's a specific Parameter* or ParameterVirtualWidget, or something generic like QWidget.  That was too weird for me.  (I spent a long time trying to figure out why two calls against the same object went to different places.)
* Needing a mySetFont() on ParameterVirtualWidget.  The intent was that the derived classes would all be forced to provide one, but even after they do the compiler finds a reference to the superclass's method in one of Qt's constructed files.
* One might hope that some kind of inheritance would apply, so that you could just do a setFont() on the top-level widget and it would percolate down.  But it seems not.
* The magic to update the font on a QDoubleSpinBox is ugly.

As noted above, updating the font on spin boxes doesn't update the size of the spinbox.  Need to research further.

jordanbrown0:
ComboBox also doesn't resize the container.  Here's an example of the problem, switching from 36 pt to 6 pt:
image

Note that the text box at the top *did* resize.

jordanbrown0:
`QWidget::adjustSize()` doesn't help, whether applied to the spinbox or to the underlying lineedit.  Nor does `QWidget::updateGeometry()`.

Fully rebuilding the widgets (e.g. if you change the list of customizable parameters) does reset the size - because it tears them all down and recreates them.

jordanbrown0:
OK, figured out the spinbox issue.  Or at least a workaround.  It appears that if you tell the `Q*SpinBox` to change its font, it doesn't tell the underlying `QLineEdit`.  And if you track down the `QLineEdit` and tell it directly, it doesn't tell the `Q*SpinBox` that it needs to resize.  But if you tell the `QLineEdit` and **then** tell the `Q*SpinBox`, even though it doesn't change anything it reassesses how much space it needs.

jordanbrown0:
Combo boxes are a bit more complex.  They don't seem to be misbehaving in quite the same way.  They seem to be willing to get bigger than their original size, and to shrink back to their original size, but will not shrink below that original size.

kintel:
Not sure if you saw my message on IRC:
I had a quick look at the customizer UI code and tinkered a tiny bit with style sheets. If I replace your ParameterWidget::setFontFamilySize() with this, it looks like it works:
`this->setStyleSheet(QString("font: %1px").arg(fontsize));`

If we could get this working with stylesheets, it would make the code a lot simpler I think.
Didn't look into changing font family; not sure if that's really necessary for this feature

jordanbrown0:
Using style sheets is indeed far simpler.  (Kind of annoying, because as I read the docs calling setFont() should have basically the same effect, and if I have a simple choice I'd rather use a straightforward function call than have a parsed text form as an intermediary.)  Based on previous discussions I changed the font of the parameters pane only, not the "control" stuff above it.  (But see Q below.)  I set both the font family and size, for consistency with the editor and console.

I looked a little more deeply into why the Preferences infrastructure isn't available at ParameterWidget constructor time, and that's just the ordering of a few lines in TabManager.  Moved the Preferences initialization up as far as it can go.  (It could probably go further up, if we broke color schemes away from the Editor object.  It seems like it should be one of the first things initialized.)

Q.  Should this preference really be limited to the Customizer, or should it be a global UI setting that would affect, e.g. the preferences UI, Animate UI, exports, et cetera?  The people who were asking were vision-limited; presumably all small text is a problem for them.

Give or take the scope question, this seems ready for review.  (And there's no reason that the scope question couldn't be revisited later.)

kintel:
I'd say limit to Customizer for now. Whole UI settings should ideally follow the OS settings. Not sure how well that works today..

jordanbrown0:
Ready for review and possibly for merge.  Interim commits rebased out.

CI failures are not related to these changes; they occur with a stock master too.

t-paul:
CI build of the master branch is fixed via https://github.com/openscad/openscad/commit/8b042fb06e7cafe26deae5e04f767b9b26be0fdc.

kintel (approved):

kintel:
Can be merged once green

Issue #4825, opened 2023-11-12
Move manifold out of experimental


kintel:
This is a meta issue collecting everything needed or wanted to be able to move Manifold out of experimental and make it the default/preferred/only geometry backend in OpenSCAD.
We may want to ship Manifold before all of these issues are addressed, in which case we should split out the unsolved topics into separate issues.

* [ ] Go through all export options and update code to work without CGAL if possible
* [ ] Look through the 2D backend to make sure it's well understood how it interacts with Manifold
* [ ] Consider moving all our Clipper V1 code to Clipper V2 and use the same Clipper build as Manifold
* [ ] Look into our libtess usage and see if it's possible to use the same tessellator as Manifold
* [ ] Find a way of running a larger scale test of OpenSCAD with CGAL vs. Manifold backends to attempt to exercise more cases than the test framework currently does (e.g. download some SCAD collections and run against both backends)
* [ ] Consider making the CGAL and Manifold backends co-exist and add UI + cmd-line features to choose which one to prefer
* [ ] Make sure Manifold failures and exceptions don't crash the OpenSCAD process: #4824
* [ ] Track missing Manifold features which currently requires a CGAL fallback
* [ ] Manifold end game: Be able to build _without_ CGAL without loss of functionality

pca006132:
I think one important missing feature would be minkowski sum. We are not very interested in implementing the convex decomposition algorithm required because we think there are better algorithms for 3D offsetting, but we do accept PRs if anyone is willing to do it.

kintel:
Good point about minkowski. A very common use-case for minkowski is indeed 3d offset, but there may be other use-cases for which 3d offset wouldn't suffice. We're kind of committed to supporting minkowski forever, but we could use Manifold for the resulting large union of convex objects (unless we already do that, and the CGAL API gives us easy access to those objects). We also have a special-case already which optimizes the case for which all minkowski operands are convex.

kintel:
Updated text to mention minkowski

pca006132:
I think the current minkowski already uses manifold if possible: https://github.com/openscad/openscad/blob/master/src/geometry/manifold/manifold-applyops-minkowski.cc#L196

ochafik:
> I think the current minkowski already uses manifold if possible: https://github.com/openscad/openscad/blob/master/src/geometry/manifold/manifold-applyops-minkowski.cc#L196

[...]

PR #4960, opened 2024-02-03
Rewrite OFF import

mmuman:
Rewrote the OFF import as I initially thought it didn't work, because of a badly scaled KiCad export converted through MeshLab.

This version removes the dependency on CGAL, and should be at feature parity.

Still missing is color support but we miss the API for that, and binary format support as I couldn't find any sample file.

### DIFF
--- a/src/io/import_off.cc
+++ b/src/io/import_off.cc
@@ -87,7 +87,7 @@ std::unique_ptr<PolySet> import_off(const std::string& filename, const Location&
   std::vector<std::string> words;
 
   if (has_ndim) {
-    boost::split(words, line, boost::is_any_of(" \t"));
+    boost::split(words, line, boost::is_any_of(" \t"), boost::token_compress_on);
     if (f.eof() || words.size() < 1) {
       AsciiError("bad header: missing Ndim");
       return std::make_unique<PolySet>(3);
@@ -119,7 +119,7 @@ std::unique_ptr<PolySet> import_off(const std::string& filename, const Location&
       line = line.erase(results.position(), results[0].length());
   }
 
-  boost::split(words, line, boost::is_any_of(" \t"));
+  boost::split(words, line, boost::is_any_of(" \t"), boost::token_compress_on);
   if (f.eof() || words.size() < 3) {
     AsciiError("bad header: missing data");
     return std::make_unique<PolySet>(3);
@@ -158,7 +158,7 @@ std::unique_ptr<PolySet> import_off(const std::string& filename, const Location&
       continue;
 
     boost::trim(line);
-    boost::split(words, line, boost::is_any_of(" \t"));
+    boost::split(words, line, boost::is_any_of(" \t"), boost::token_compress_on);
     if (words.size() < 3) {
       AsciiError("can't parse vertex: not enough data");
       return std::make_unique<PolySet>(3);
@@ -197,7 +197,7 @@ std::unique_ptr<PolySet> import_off(const std::string& filename, const Location&
       continue;
 
     boost::trim(line);
-    boost::split(words, line, boost::is_any_of(" \t"));
+    boost::split(words, line, boost::is_any_of(" \t"), boost::token_compress_on);
     if (words.size() < 1) {
       AsciiError("can't parse face: not enough data");
       return std::make_unique<PolySet>(3);
--- a/src/io/import_off.cc
+++ b/src/io/import_off.cc
@@ -221,9 +221,11 @@ std::unique_ptr<PolySet> import_off(const std::string& filename, const Location&
       }
       if (words.size() >= n + 4) {
         // TODO: handle optional color info
+        /*
         int r=boost::lexical_cast<int>(words[n+1]);
         int g=boost::lexical_cast<int>(words[n+2]);
         int b=boost::lexical_cast<int>(words[n+3]);
+        */
       }
     } catch (const boost::bad_lexical_cast& blc) {
       AsciiError("can't parse face: bad data");
--- a/src/io/import_off.cc
+++ b/src/io/import_off.cc
@@ -1,6 +1,5 @@
 #include "import.h"
 #include "PolySet.h"
-#include "PolySetBuilder.h"
 #include "printutils.h"
 #include "AST.h"
 #include <fstream>
@@ -15,7 +14,7 @@ std::unique_ptr<PolySet> import_off(const std::string& filename, const Location&
 
   int lineno = 1;
   std::string line;
-  //boost::smatch results;
+  double scale = 1000.0;
 
   auto AsciiError = [&](const auto& errstr){
     LOG(message_group::Error, loc, "",
@@ -30,8 +29,6 @@ std::unique_ptr<PolySet> import_off(const std::string& filename, const Location&
 
   boost::regex ex_magic(R"(^OFF$|^COFF$)"); // TODO: 4OFF?
   boost::regex ex_comment(R"(^\s*#)");
-
-
 
   std::getline(f, line);
   if (f.eof() || !boost::regex_search(line, ex_magic)) {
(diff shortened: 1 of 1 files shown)


mmuman:
Don't tell me there's actually a 3rd format called OFF!?

mmuman:
Wait, did I reimplement this because it was scaled so tiny I thought it discarded the input? :facepalm: 

At least this version doesn't depend on CGAL…

t-paul:
Also CGAL supports only 50% of the format "Note that the format defines both ASCII and binary OFF formats, but only the ASCII OFF format is supported in CGAL.", so there's even more opportunities.

kintel:
This also unblocks part of #4825

kintel:
The scale issue is likely because OFF, like OpenSCAD, does not specify unit, and you're probably importing a design which uses unexpected units.

mmuman:
Ok cool then!

kintel:
The built-in automated tests already catch some issues. If you discover something which isn't covered by the existing tests, we should add more tests.

mmuman:
This should pass all the tests now, the headers they used differed from the Wikipedia interpretation of the specs.

t-paul (approved):

kintel:
For some reason, this fails on Windows. Haven't had the chance to look at the logs. Let me know if you need any assistance with that.

mmuman:
Let's see the tests now, they pass again here.

mmuman:
Bloody Windows… That should fix it.

src/io/import_off.cc
```diff
@@ -2,24 +2,234 @@
...
+  bool got_magic = false;
+  // defaults
+  bool has_normals = false;
+  bool has_color = false;
+  bool has_textures = false;
+  bool has_ndim = false;
+  bool is_binary = false;
+  unsigned int dimension = 3;
+
+  while (line.length() == 0) {
```
kintel, line 63: -> line.empty()
Could probably use empty() in several places below to make code a tiny bit clearer.

src/io/import_off.cc
```diff
@@ -2,24 +2,234 @@
...
+    if (!getline_clean("bad header: end of file"))
```
kintel, line 64: nit: Use curly braces for multi-line if statements for readability. Same for a couple more cases below.

src/io/import_off.cc
```diff
@@ -2,24 +2,234 @@
...
+    is_binary = results[6].length() > 0;
+    if (results[4].length())
+      dimension = 4;
+    if (results[5].length())
+      has_ndim = true;
+  }
+
+  // TODO: handle binary format
+  if (is_binary) {
+    AsciiError("unhandled binary format");
```
kintel, line 83: -> "binary OFF format not supported"

src/io/import_off.cc
kintel, line 63: You copy this snippet a few times. Perhaps write another lambda next_line() or smth.?

src/io/import_off.cc
```diff
@@ -2,24 +2,234 @@
...
+    } catch (const boost::bad_lexical_cast& blc) {
+      AsciiError("bad header: bad data for Ndim");
+      return std::make_unique<PolySet>(3);
+    }
+  }
+
+  PRINTDB("Header flags: N:%d C:%d ST:%d Ndim:%d B:%d", has_normals % has_color % has_textures % dimension % is_binary);
+
+  if (dimension != 3) {
+    AsciiError("unhandled vertex dimensions");
```
kintel, line 111: Could you emit the found # of dimensions too?

src/io/import_off.cc
```diff
@@ -2,24 +2,234 @@
...
+      return std::make_unique<PolySet>(3);
+    }
+
+    try {
+      Vector3d v;
+      for (int i = 0; i < 3; i++) {
+        v[i]= boost::lexical_cast<double>(words[i]);
+      }
+      //PRINTDB("Vertex[%ld] = { %f, %f, %f }", vertex % v[0] % v[1] % v[2]);
+      int o = dimension;
```
kintel, line 171: Avoid single-character variable names unless it's very obvious what it means.
I can't really tell what the below code does

src/io/import_off.cc
```diff
@@ -2,24 +2,234 @@
...
+
+    boost::trim(line);
+    boost::split(words, line, boost::is_any_of(" \t"), boost::token_compress_on);
+    if (words.size() < 1) {
+      AsciiError("can't parse face: not enough data");
+      return std::make_unique<PolySet>(3);
+    }
+
+    try {
+      unsigned long n=boost::lexical_cast<unsigned long>(words[0]);
```
kintel, line 204: n -> face_size or smth?

src/io/import_off.cc
```diff
@@ -2,24 +2,234 @@
...
+      if (words.size() - 1 < n) {
+        AsciiError("can't parse face: missing indices");
+        return std::make_unique<PolySet>(3);
+      }
+      ps->indices.emplace_back().reserve(n);
+      //PRINTDB("Index[%d] [%d] = { ", face % n);
+      for (int i = 0; i < n; i++) {
+        int ind=boost::lexical_cast<int>(words[i+1]);
+        //PRINTDB("%d, ", ind);
+        if(ind >= 0 && ind < vertices_count)
```
kintel, line 214: Should we print an error on overflow?

src/io/import_off.cc
```diff
@@ -2,24 +2,234 @@
...
+        */
+      }
+    } catch (const boost::bad_lexical_cast& blc) {
+      AsciiError("can't parse face: bad data");
+      return std::make_unique<PolySet>(3);
+    }
+  }
+
+  //PRINTDB("PS: %ld vertices, %ld indices", ps->vertices.size() % ps->indices.size());
+  f.close();
```
kintel, line 233: close is unnecessary; the destructor will close.

kintel (commented):
Just added some minor suggestions. Feel free to disregard, but let us know when you're ready for this to me merged.

src/io/import_off.cc
mmuman: OK, haven't used st:: stuff much in Haiku :-)

mmuman: I could probably just do the loop in `getline_clean()` since we aren't interested in the comments anyway…

mmuman: > has no member named ‘empty’

Oh results[] is just the trait but not a real std::string, so I'd need `.str().empty()`… not shorter than `.length() > 0`.

kintel: Not a big deal, most of my comments are just minor nudges to make it easier for the next person reading this code :)

src/io/import_off.cc
mmuman: OK

src/io/import_off.cc
mmuman: OK

src/io/import_off.cc
mmuman: Oh, that's the offset into the split words, but I changed the code to just reuse `i` since we increment it anyway.

src/io/import_off.cc
kintel: Right, I didn't spot that detail. Interesting tidbit about boost sub_match: It has a public `bool matched` member, so you can technically do `if (results[2].matched)`. Not really helping too much though.

src/io/import_off.cc
mmuman: OK

src/io/import_off.cc
mmuman: OK

mmuman:
I think we should be good now.

src/io/import_off.cc
mmuman: OK

src/io/import_off.cc
mmuman: OK

src/io/import_off.cc
mmuman: It's a little more obvious what is done yeah.
Now, I'm not really sure the various magic flags are always in that order though, if not I'd have to rewrite that part. For now I haven't found a file that didn't read though.

mmuman: Works fine like that, it's a bit shorter.

mmuman:
Hmm, one sample I just tested failed with latest code because of missing newline on the last line…

mmuman:
Now I can load all samples from https://people.sc.fsu.edu/~jburkardt/data/off/off.html except the 4d ones. I think we're ready.

kintel:
Ideally, we should add one example of each variant to our test suite, but I'm happy to merge this as is :)

kintel (approved):

mmuman:
:partying_face:

PR #4990, opened 2024-02-15
fix append_stl infinite recursion

pca006132:
Closes #4981.

This is a simple infinite recursion.

PR #4991, opened 2024-02-15
fix manifold stuff

### LINKED ISSUE #4732: Manifold - Minkowski fails on concave polyhedron with render mode (but works if you preview first)
Minkowski fails when given certain shapes (e.g. a concave polyhedron + tall convex object) using manifold.

**To Reproduce**
1. Enable manifold
2. Use the code below
3. Use a fresh instance of OpenSCAD, or 'Flush Caches'
4. Render (F6). The following error appears:
   > Rendering Polygon Mesh using Manifold...
   > WARNING: [manifold] Minkowski failed with error, falling back to Nef operation: CGAL ERROR: precondition violation!
   > Expr: ! running
   > File: /mxe/usr/x86_64-w64-mingw32.static.posix/include/CGAL/Timer.h
   > Line: 83
5. Use a fresh instance of OpenSCAD, or 'Flush Caches'
6. Preview (F5)
7. Render (F6). No error this time.

No other experimental features were enabled apart from manifold.

**Expected behavior**
Render (F6) to pass at step 3.

```
$fn = 40;
cracked_square = [[-10,-10],[-10,10],[10,10],[10,-10],[-8,-10],[7,7]];
silo_shape = [for(a=[0:360/$fn:89.99]) 5*[cos(a),sin(a)],[0,5],[0,-10],[5,-10]];

minkowski() {
  rotate([10,30,0]) linear_extrude(20) polygon(cracked_square);
  rotate_extrude() polygon(silo_shape);
}
```

 - OS: Windows 10 22H2 (19045.3324)
 - System: AMD Ryzen 7 7700X
 - OpenSCAD Version: 2023.08.25 (git '05b962168) [latest 64-bit Development Snapshot]

P/S. Manifold is awesome, thank you all for integrating this library.

### LINKED ISSUE #4753: Manifold 1st and 2nd F6 Render return different Geometry types
**Describe the bug**
This is a follow-up to issue #4685. where a 2nd render with manifold enabled was triggering an assertion.  The assertion has been removed, but behavior remains where it seems to return different Geometry types between the 1st and 2nd render.  This is apparent by the differing render statistics displayed upon F6.

I suspect the cause is related to caching in some way, but haven't dug in to determine specifics.

**To Reproduce**
1. Enable Preferences->Features->Manifold
2. Use the following design (but pretty much any design should do):
```
union() {
  cube();
  sphere();
}
```
3. Hit F6. Then hit F6 again.
4. Review Console output (only 2nd render says `Top level object is a 3D object (manifold)` indicating a `ManifoldGeometry`, not sure what type is returned on the 1st render):
```
Saved backup file: /home/hans/.local/share/OpenSCAD/backups/unsaved-backup-YaYgPcrb.scad
Rendering Polygon Mesh using Manifold...
Geometries in cache: 3
Geometry cache size in bytes: 2560
CGAL Polyhedrons in cache: 0
CGAL cache size in bytes: 0
Total rendering time: 0:00:00.022
   Facets:         44
Rendering finished.
 
Saved backup file: /home/hans/.local/share/OpenSCAD/backups/unsaved-backup-YaYgPcrb.scad
Rendering Polygon Mesh using Manifold...
Geometries in cache: 3
Geometry cache size in bytes: 2560
CGAL Polyhedrons in cache: 0
CGAL cache size in bytes: 0
Total rendering time: 0:00:00.016
   Top level object is a 3D object (manifold):
   Status:     NoError
   Genus:      0
   Vertices:       24
   Facets:         44
   BBox.min:   -0.809016, -0.951056, -0.866025
   BBox.max:   1.000000, 1.000000, 1.000000
Rendering finished.
```
 
**Expected behavior**
Consistent Render Statistics across renders

 - OS: Linux Mint 21.2
 - System: amd64
 - OpenSCAD Version 2023.09.22 nightly

pca006132:
Closes #4732, resolves #4753, and also update manifold to get the performance improvement on windows.

For the `GeometryEvaluator`, do we really need the `getGeometryAsPolySet` for `ManifoldGeometry` regardless of whether or not `allownef` is set? I removed that part of the code and it seems fine, and I don't understand why we need that. Btw, we probably want to rename `allownef` into `asConvexPolySet` or something? The `convex` part is weird anyway (it is asking for convex polygons, not convex in the 3D sense).

Btw, this depends on #4990 or some tests will not pass because we are attempting to export an STL file with `ManifoldGeometry`. The flaky nature of #4990 is caused by the cache issue, which is fixed in this PR.

### DIFF
--- a/src/geometry/GeometryEvaluator.cc
+++ b/src/geometry/GeometryEvaluator.cc
@@ -80,9 +80,12 @@ std::shared_ptr<const Geometry> GeometryEvaluator::evaluateGeometry(const Abstra
     // We cannot render concave polygons, so tessellate any 3D PolySets
     auto ps = PolySetUtils::getGeometryAsPolySet(this->root);
     if (ps && !ps->isEmpty()) {
+      this->root = ps;
       // Since is_convex() doesn't handle non-planar faces, we need to tessellate
       // also in the indeterminate state so we cannot just use a boolean comparison. See #1061
       bool convex = bool(ps->convexValue()); // bool is true only if tribool is true, (not indeterminate and not false)
+      if (std::dynamic_pointer_cast<const ManifoldGeometry>(ps))
+        convex = true; // manifold only have triangles, which are convex
       if (!convex) {
         assert(ps->getDimension() == 3);
         this->root = PolySetUtils::tessellate_faces(*ps);
--- a/src/geometry/GeometryEvaluator.cc
+++ b/src/geometry/GeometryEvaluator.cc
@@ -72,34 +72,25 @@ std::shared_ptr<const Geometry> GeometryEvaluator::evaluateGeometry(const Abstra
     } else {
       this->traverse(node);
     }
-#ifdef ENABLE_CGAL
-    if (std::dynamic_pointer_cast<const CGALHybridPolyhedron>(this->root)) {
-      this->root = PolySetUtils::getGeometryAsPolySet(this->root);
-    }
-#endif
-#ifdef ENABLE_MANIFOLD
-    if (std::dynamic_pointer_cast<const ManifoldGeometry>(this->root)) {
-      this->root = PolySetUtils::getGeometryAsPolySet(this->root);
-    }
-#endif
+  } else {
+    this->root = GeometryCache::instance()->get(key);
+  }
 
-    if (!allownef) {
-      // We cannot render concave polygons, so tessellate any 3D PolySets
-      auto ps = PolySetUtils::getGeometryAsPolySet(this->root);
-      if (ps && !ps->isEmpty()) {
-        // Since is_convex() doesn't handle non-planar faces, we need to tessellate
-        // also in the indeterminate state so we cannot just use a boolean comparison. See #1061
-        bool convex = bool(ps->convexValue()); // bool is true only if tribool is true, (not indeterminate and not false)
-        if (!convex) {
-          assert(ps->getDimension() == 3);
-          this->root = PolySetUtils::tessellate_faces(*ps);
-        }
+  if (!allownef) {
+    // We cannot render concave polygons, so tessellate any 3D PolySets
+    auto ps = PolySetUtils::getGeometryAsPolySet(this->root);
+    if (ps && !ps->isEmpty()) {
+      // Since is_convex() doesn't handle non-planar faces, we need to tessellate
+      // also in the indeterminate state so we cannot just use a boolean comparison. See #1061
+      bool convex = bool(ps->convexValue()); // bool is true only if tribool is true, (not indeterminate and not false)
+      if (!convex) {
+        assert(ps->getDimension() == 3);
+        this->root = PolySetUtils::tessellate_faces(*ps);
       }
     }
-    smartCacheInsert(node, this->root);
-    return this->root;
   }
-  return GeometryCache::instance()->get(key);
+  smartCacheInsert(node, this->root);
+  return this->root;
 }
 
 bool GeometryEvaluator::isValidDim(const Geometry::GeometryItem& item, unsigned int& dim) const {


kintel:
`allownef` means something else: It's used by the preview renderer, e.g. for the `render()` module, to request that we only return `PolySet` or `Polygon2d` geometry, since the preview renderer doesn't deal with backend-specific geometry.
It should definitely be renamed/rewritten; we just don't have a good internal concept for "backend-specific geometry" yet.

pca006132:
btw, thinking about it, we should also perform toPolySet even when the cache is hit, because the result can be a `ManifoldGeometry`. Will fix when I actually wake up... (suddenly thought about this and woke up at 5am)

kintel:
Out of scope for this PR, but we should really rethink the cache. Having the possibility of the cache returning different geometry types is already the source of some problems.

src/geometry/GeometryEvaluator.cc
```diff
@@ -73,34 +72,25 @@ std::shared_ptr<const Geometry> GeometryEvaluator::evaluateGeometry(const Abstra
...
-#ifdef ENABLE_CGAL
-    if (std::dynamic_pointer_cast<const CGALHybridPolyhedron>(this->root)) {
-      this->root = PolySetUtils::getGeometryAsPolySet(this->root);
-    }
-#endif
-#ifdef ENABLE_MANIFOLD
-    if (std::dynamic_pointer_cast<const ManifoldGeometry>(this->root)) {
-      this->root = PolySetUtils::getGeometryAsPolySet(this->root);
-    }
-#endif
```
kintel, line 85: I believe the reason for doing this is that `GeometryCache` isn't really meant to cache any geometry. It's for caching PolySets (and Polygon2d). Otherwise, we wouldn't have a separate `CGALCache`.
..so my guess is that when extending with new geometry backends, it was easier to just convert to PolySet than to generalize the caching infrastructure.

CMakeLists.txt
```diff
@@ -55,6 +55,7 @@ option(ENABLE_TBB "Enable support for oneAPI Threading Building Blocks." ON)
 option(ENABLE_CGAL "Enable CGAL." ON)
 option(ALLOW_BUNDLED_HIDAPI "Allow usage of bundled HIDAPI library (Windows only)." OFF)
 option(ENABLE_PYTHON "Enable experimental Python Interpreter" OFF)
+option(BUILD_SHARED_LIBS "Build shared library" OFF)
```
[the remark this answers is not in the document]
pca006132: ah, I forgot why this change is added here, will remove

src/geometry/GeometryEvaluator.cc
pca006132: Ah, make sense. I think we need to rethink about this cache thing.

pca006132:
I was thinking if it would be better to just convert to PolySet when caching ManifoldGeometry, but one problem is that this triggers evaluation of the underlying Manifold object, which is expensive and forbids other useful optimizations. Avoid caching ManifoldGeometry is also problematic because we will not be able to cache things that are not top-level.

src/geometry/GeometryEvaluator.cc
kintel: I think the root cause triggering the STL issue is that `GeometryEvaluator::smartCacheInsert()` (https://github.com/openscad/openscad/blob/ded24c57b2eb91f6985697361bd435a31788941a/src/geometry/GeometryEvaluator.cc#L327) has buggy logic, causing it to insert `ManifoldGeometry` into the `GeometryCache`.

To fix that we need to rethink caching a bit.

kintel:
Yeah, the idea of caching CGAL objects separately was that conversion to PolySet was somewhat lossy, and the same logic applies to Manifold geometry. Conversion performance is less of a concern, but if it matters we should keep that in mind too.

IMO, we should probably get rid of the whole pattern of having a single GeometryEvaluator which does a bunch of 
 feature checks and `dynamic_pointer_cast` to figure out what to do. Perhaps we should implement a design where there are multiple GeometryEvaluator implementations (subclasses) and only ever a single of them is instantiated at a time (depending on the chosen backend), and where the evaluator is typed/templated and owns the cache.

src/geometry/GeometryEvaluator.cc
pca006132: Yeah, I think we should try to make it clear what different parts of the API accepts. It would be cleaner if we can keep the backend specific types within the geometry part of the code, and use PolySet for everywhere else.

pca006132:
That said, I think we will still have backend specific geometry type vs PolySet, as primitives are returned as PolySets? Agree this is the more maintainable way though.

kintel:
Yeah, that's why we might want to introduce a concept of "non-backend-specific geometries". One challenge is that "implicit unions" introduce the concept of GeometryList, which essentially yields geometry trees.

pca006132:
@kintel so what should we do with this PR? remove the cache related stuff to properly fix it? may require some refactoring if we want to do it in the way you described.

kintel:
I'll think about it for a bit. Since automated tests don't help us much, we have to be a bit careful.

pca006132:
btw, wonder if this will make more sense:

```diff
diff --git a/src/geometry/GeometryEvaluator.cc b/src/geometry/GeometryEvaluator.cc
index a721f92fb..f78ab1126 100644
--- a/src/geometry/GeometryEvaluator.cc
+++ b/src/geometry/GeometryEvaluator.cc
@@ -80,9 +80,12 @@ std::shared_ptr<const Geometry> GeometryEvaluator::evaluateGeometry(const Abstra
     // We cannot render concave polygons, so tessellate any 3D PolySets
     auto ps = PolySetUtils::getGeometryAsPolySet(this->root);
     if (ps && !ps->isEmpty()) {
+      this->root = ps;
       // Since is_convex() doesn't handle non-planar faces, we need to tessellate
       // also in the indeterminate state so we cannot just use a boolean comparison. See #1061
       bool convex = bool(ps->convexValue()); // bool is true only if tribool is true, (not indeterminate and not false)
+      if (std::dynamic_pointer_cast<const ManifoldGeometry>(ps))
+        convex = true; // manifold only have triangles, which are convex
       if (!convex) {
         assert(ps->getDimension() == 3);
         this->root = PolySetUtils::tessellate_faces(*ps);
```

so it always returns a `PolySet` when `!allownef`.

kintel:
Yeah, that looks reasonable. Something is odd about the original code though. Why isn't this->root set when the PolySet is convex?

I'll try to find a good way of running some GUI tests of this.

pca006132:
My guess is that we set convexity as unknown most of the time, except in things like simple linear extrude with convex polygon.

kintel:
Wait, manifold using only triangles doesn't imply that the PolySet is convex..

pca006132:
yes, but I think what it does it making sure the polygons are convex instead of the polyset is convex? you cannot make a non-convex polyset convex by tessellation...

kintel:
Thanks for the discussion! Closing in favour of #4996

PR #4994, opened 2024-02-17
OFF import: Handle binary format

mmuman:
Finally found a way to generate binary OFF files:
https://github.com/pmp-library/pmp-library

However it seems it saves them as little-endian, while the specs says it should be big-endian. So we just try and support both, and do best-effort at autodetecting it.

Validated with all the previous samples.


mmuman:
Seems macOS doesn't like the IEEE-754 test… I wanted to make sure we'd get compatible floats when reading…

kintel:
For tests, read https://github.com/openscad/openscad/blob/master/doc/testing.txt
..and look at e.g. the OBJ file tests as an example:
https://github.com/openscad/openscad/blob/master/tests/data/obj/cube.obj
https://github.com/openscad/openscad/blob/master/tests/data/obj/dodecahedron.obj
https://github.com/openscad/openscad/blob/55be9cf86a4059a96239e7b660481643975c8173/tests/CMakeLists.txt#L717

kintel:
Looks like you're just missing parentheses for the macOS issue

mmuman:
> Looks like you're just missing parentheses for the macOS issue

No it also gives the error

src/io/import_off.cc
```diff
@@ -11,6 +11,36 @@
 // References:
 // http://www.geomview.org/docs/html/OFF.html
 
+#if !defined(BOOST_ENDIAN_BIG_BYTE_AVAILABLE) && !defined(BOOST_ENDIAN_LITTLE_BYTE_AVAILABLE)
+#error Byte order undefined or unknown. Currently only BOOST_ENDIAN_BIG_BYTE and BOOST_ENDIAN_LITTLE_BYTE are supported.
+#endif
+
+// Specs says data is big-endian, however PMP-library used little-endian
```
kintel, line 18: I would say this is a PMP bug. Should we really follow that example and support software outputting buggy files?

mmuman: It depends if it's been used widely or not…

mmuman:
Let's see how they handle https://github.com/pmp-library/pmp-library/issues/198

kintel:
Looks like the PMP issue was resolved

mmuman:
Just rebased for now.

mmuman:
So, should we use `double-conversion`?
Do we drop little endian?

mmuman:
Oh I'll have to look into that again.

PR #5001, opened 2024-02-20
more precise manifold construction

pca006132:
Simplified the code a bit, and avoids the mesh quantization.

### DIFF
--- a/src/geometry/manifold/manifoldutils.cc
+++ b/src/geometry/manifold/manifoldutils.cc
@@ -109,38 +109,16 @@ template std::shared_ptr<ManifoldGeometry> createMutableManifoldFromSurfaceMesh(
 std::shared_ptr<ManifoldGeometry> createMutableManifoldFromPolySet(const PolySet& ps)
 {
 #ifdef ENABLE_CGAL
-  PolySet psq(ps);
-  std::vector<Vector3d> points3d;
-  psq.quantizeVertices(&points3d);
-  auto ps_tri = PolySetUtils::tessellate_faces(psq);
+  std::unique_ptr<PolySet> tmp;
+  if (!ps.isTriangular)
+    tmp = PolySetUtils::tessellate_faces(ps);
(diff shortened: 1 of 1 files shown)


pca006132:
`cgalpng_issue1165` is actually quite weird, I think both outputs are valid from the point of view of manifold. If we do exact arithmetic without doing decimation, I think the output with a membrane is fine. But manifold may also remove the membrane, as we have some heuristics to remove that (which may or may not work).

And while getting this, I think there is a bug in our cache implementation: two `multimatrix` that differs by tiny amounts may be considered the same by the cache due to its rounding behavior. And for cases like `cgalpng_issue1165`, tiny differences may make the membrane appear or disappear, and this rounding for cache can make it confusing to the user. (they have to flush the cache to get a 'correct' result). I was thinking about changing the code to use doubleconversion, with the most precise string, but it is failing a lot of tests because the csg file is now different, so it is not included in this PR.

kintel:
Issue 1165 was originally a tessellation issue; CGAL would output the top concave face as one large concave polygon, and our triangulator didn't handle the concavity properly. Removing tiny details < some epsilon is fine. Challenge is, of course, dealing with test differences across different geometry backends.

We should discuss the cache issue elsewhere. Not sure if this would be the ideal example to use for highlighting/testing that.

pca006132:
for some reason this fails https://www.thingiverse.com/thing:1958354

need further investigation

kintel:
As mentioned on IRC: A polyhedron where two triangles exactly touches also fails: https://github.com/openscad/openscad/blob/9d464b8c71c9f720cde62d468aac011cb57a134e/tests/data/scad/3D/misc/polyhedron-self-touch-face.scad

kintel:
> for some reason this fails https://www.thingiverse.com/thing:1958354
> 
> need further investigation

The objects generated by that library are terrible:
Screenshot 2024-02-26 at 00 38 39

PR #5006, opened 2024-02-22
Clean up LOG/PRINT functions.  Fixes #5004.

### LINKED ISSUE #5004: Doesn't stop execution on errors
**Describe the bug**
With "stop on the first warning" set, execution stops on the first warning.
However, it does *not* stop on errors(!).

**To Reproduce**
```
polyhedron() cube();
echo("hello");
```
correctly does not print "hello".

```
polyhedron(points=5);
echo("hello");
```
reports an error (ERROR: Unable to convert points = 5 to a vector of coordinates in file , line 1), without a stack trace, but then continues and prints "hello".

**Expected behavior**
Terminate with a stack trace.

Windows 11, OpenSCAD 2023.12.31.

**Additional context**
The problem appears to be at
https://github.com/openscad/openscad/blob/dc2a0d1c28e41016abf4f91ffbc4c6c707c4f64a/src/utils/printutils.cc#L119
where it handles two cases:
* Throws on warning if that's enabled.
* Records a deferred exception on error if suppressing throws.
but an ordinary `message_group::Error` without `no_throw` does nothing.

It seems like the second half of the `if` should just check for `Error`, and not check `no_throw`.

jordanbrown0:
Move the error-log output next to the console output.  (This means that error-log entries are suppressed after five identical ones.  Is that good?)

Move throwing on hardwarning/error out of the `PRINT()` functions and into its own function, and have LOG() call that function.  (Fixes #5004.)

Along the way, clean up a little:  nobody outside of printutils.cc should be calling `PRINT_NOCACHE()`, so scope it static.  Remove PRINTB_NOCACHE since it's unused and its one pseudo-use in a comment looks like the intent is the same as `PRINTD()`.

### DIFF
--- a/src/core/Expression.cc
+++ b/src/core/Expression.cc
@@ -424,10 +424,10 @@ void FunctionDefinition::print(std::ostream& stream, const std::string& indent) 
 }
 
 /**
- * This is separated because PRINTB uses quite a lot of stack space
- * and the method using it evaluate()
+ * This is separated because LOG uses quite a lot of stack space
+ * and the method using it, evaluate(),
  * is called often when recursive functions are evaluated.
- * noinline is required, as we here specifically optimize for stack usage
+ * NOINLINE is required, as we here specifically optimize for stack usage
  * during normal operating, not runtime during error handling.
  */
 static void NOINLINE print_err(const char *name, const Location& loc, const std::shared_ptr<const Context>& context){
@@ -435,10 +435,10 @@ static void NOINLINE print_err(const char *name, const Location& loc, const std:
 }
 
 /**
- * This is separated because PRINTB uses quite a lot of stack space
- * and the method using it evaluate()
+ * This is separated because LOG uses quite a lot of stack space
+ * and the method using it, evaluate(),
  * is called often when recursive functions are evaluated.
- * noinline is required, as we here specifically optimize for stack usage
+ * NOINLINE is required, as we here specifically optimize for stack usage
  * during normal operating, not runtime during error handling.
  */
 static void NOINLINE print_trace(const FunctionCall *val, const std::shared_ptr<const Context>& context){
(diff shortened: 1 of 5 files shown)


jordanbrown0:
Sigh.  A fair number of tests depend on Error not being fatal.  I could have Error be similar to Warning, fatal only if `hardwarnings`.  Or fix those tests.  Or maybe add a separate `harderrors`.  Thoughts?  Shelved for the moment pending comments on this and on the "suppress >5 duplicates" question above.

thehans:
Do we have a clear differentiation of what types of conditions should trigger an ERROR vs a WARNING?

Looking at the failing tests report, it appears that all the errors that terminate the tests are related to converting invalid data to points:
```
ERROR: Unable to convert points[4] = [0, 0, -inf] to a vec3 of numbers in file builtin-invalid-range-test.scad, line 16
...
ERROR: Unable to convert points[3] = [30] to a vec2 of numbers in file errors-warnings.scad, line 3
...
ERROR: Unable to convert points = undef to a vector of coordinates in file allmodules.scad, line 25
...
ERROR: Unable to convert points = undef to a vector of coordinates in file text-search-test.scad, line 15
...
ERROR: Unable to convert points = undef to a vector of coordinates in file polygon-tests.scad, line 1
```

Would it make any more sense to have those as WARNINGS?  What should ideally qualify as an ERROR?

jordanbrown0:
I don't have any real doubt that some cleanup is appropriate on what's a warning versus an error, but that seems orthogonal to the issues here around how each of them is handled.

I'd say that an error is when the result cannot be correct, and a warning is where it might or might not be right but is in some way suspicious.  Deprecations (e.g. behavior is slated for removal, behavior is known to be confusing and has been replaced) are the obvious candidates for warnings.

No matter the distinction, there will be those who will say "why did you stop my run, it worked fine before this, get off my lawn", even though there's evidence that something is *not* working the way they expect.  That's why we probably need an option to not stop on errors that aren't totally fatal.  (Parse errors might be totally fatal.)

kintel:
As discussed offline, should we just get the NOCACHE fix merged before delving into the details of errors and warnings?

jordanbrown0:
#5004 got fixed in a different way.  We probably still need some refactoring, as discussed there, but at this point I don't think this is the right approach.

PR #5007, opened 2024-02-22
Remove degenerate polygons in PolySetBuilder

### LINKED ISSUE #5000: Crash in STL export after rotate_extrude
I'm getting an assertion failure crash here:
```
src/io/export_stl.cc:131: // Tessellation already eliminated these cases.
src/io/export_stl.cc:132: uint64_t {anonymous}::append_stl(std::shared_ptr<const PolySet>, std::ostream&, bool): Assertion `p0 != p1 && p0 != p2 && p1 != p2' failed.
```
using the master branch checked out & built Feb 13 under Ubuntu 22.04.

It seems like the geometry created by rotate_extrude is in error, and/or the tessellation mentioned in the comment has not eliminated these cases.

To reproduce, I enter just the following, press F6, then press F7 to export STL. The crash happens during the STL export.
`rotate_extrude() square([10, 10]);`

It doesn't matter the value of angle or presence of the angle argument.
It doesn't matter CGAL vs Manifold.

It also crashes also on the code example from manual section 2D_to_3D_Extrusion:
`rotate_extrude($fn = 200) polygon( points=[[0,0],[2,1],[1,2],[1,3],[3,4],[0,5]] );`

These objects have an edge along the Y axis which is permitted.

kintel:
Also includes some PolySetBuilder refactoring.

Fixes #5000

Note: rotate_extrude() could be improved to better handle manifold corner cases. Deferring to #5046


kintel:
@pca006132 In case you have any comments :)

src/core/primitives.cc
```diff
@@ -404,13 +405,17 @@ std::string PolyhedronNode::toString() const
 
 std::unique_ptr<const Geometry> PolyhedronNode::createGeometry() const
 {
-  auto p = std::make_unique<PolySet>(3);
-  p->setConvexity(this->convexity);
-  p->vertices=this->points;
-  p->indices=this->faces;
-  for (auto &poly : p->indices)
+  PolySetBuilder builder(this->points.size(), this->faces.size());
```
pca006132, line 408: going through `PolySetBuilder` will lose manifoldness.

src/geometry/PolySetBuilder.h
```diff
@@ -16,20 +16,26 @@ class PolySetBuilder
...
-  void prependVertex(int n);
-  void prependVertex(const Vector3d &v);
+  void appendPolygon(const std::vector<int>& inds);
+  void appendPolygon(const std::vector<Vector3d>& v);
+
+  void beginPolygon(int nvertices);
+  void addVertex(int n);
+  void addVertex(const Vector3d &v);
+  // Calling this is optional; will be called automatically when adding a new polygon or building the PolySet
+  void endPolygon();
```
pca006132, line 29: why should we have this if it can be called automatically?

src/core/primitives.cc
kintel: Riight, this is the wrong hammer for this job; we need something that cleans up a user-specified polyhedron. I'll think about it, perhaps just tessellate to triangles early for user polyhedrons.

src/geometry/PolySetBuilder.h
kintel: Really just for API cleanliness, as most client code will use begin/end pairs to make it more readable.

kintel: We could make it required and rather assert if someone tries to do two beginPolygon() in a row.

pca006132:
I think we probably should not do this in the `PolySetBuilder`, e.g. manifold outputs can have degenerate triangles but we don't want to pass it through `PolySetBuilder` as this loses manifoldness as well.

src/geometry/PolySetBuilder.h
pca006132: > We could make it required and rather assert if someone tries to do two beginPolygon() in a row.

Sounds good.

pca006132:
Another concern (not specific to this PR): we are using `assert` for checks, but we do not build in debug mode in the CI, so these checks are not run automatically.

kintel:
I think the only issue here is `PolyhedronNode`. Apart from that, this only affects code already using PolySetBuilder, and it's a cheap feature. The goal of the assert() is to catch programming errors, not run-time errors, so it's really just a convenience for developers.

That said, there might be a better/more general place to insert a polygon cleaner which doesn't collapse manifolds.
Btw., do you know a good nomenclature to describe this type of objects (manifolds where multiple different vertices have the same position in space)?

jordanbrown0:
>we are using assert for checks, but we do not build in debug mode in the CI, so these checks are not run automatically.

Tidbit:  It's been a long time, but the last time I tried to build in debug mode for MSYS2/Windows, the build failed - it hit some kind of Windows symbol table limit.

elalish:
I'm not sure it has a specific name - it's just an artifact of various different manifoldness definitions floating around. I wrote up the major differences on our [wiki](https://github.com/elalish/manifold/wiki/Manifold-Library#manifoldness-definition). The short answer is that anything that doesn't allow disjoint vertices at the same location cannot maintain manifoldness through Boolean operations in general. It's closely related to why STL is not capable of reliably transmitting manifolds either. Hence why 3MF should be used instead, and now even glTF with my new EXT_mesh_manifold extension.

PR #5009, opened 2024-02-23
Update manifold to get TBB fix, print TBB version

kintel:
This fixes the issue that re-running cmake causes a `TBB::tbb target not found` error.

Also, added fallback to pkg-config if find_package(TBB) failed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -861,6 +861,7 @@ if(EXPERIMENTAL AND ENABLE_TBB)
     -DTHRUST_DEVICE_SYSTEM=THRUST_DEVICE_SYSTEM_TBB
   )
   find_package(TBB REQUIRED)
+  message(STATUS "TBB: ${TBB_VERSION}")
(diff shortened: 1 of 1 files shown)


pca006132:
btw should we port the pkg-config fallback?

kintel:
Yes, thanks for reminding me. I'll add it to this PR.

kintel:
@pca006132 Btw., take a look at my fallback code - could be nice for Manifold too, as a way of aligning the interfaces of the two ways of looking up TBB. I could perhaps package it as FindTBB.cmake if you want to reuse it.

pca006132:
Sounds good, but for manifold we hope to reduce users burden by doing FetchContent and build tbb from source if it is not found. I think the code here can be updated to allow that?

kintel:
Possibly, I'm not familiar with FetchContent. What would be cool about FetchContent would be to employ a standard pattern of find, then pkg-config, then FetchContent, but we can leave that to later :)

PR #5012, opened 2024-02-24
Implement "height" parameter on rotate_extrude to allow for helical threads.

### LINKED ISSUE #5011: Add a "height" parameter for rotate_extrude to help with threading
If trying to build threaded parts of any kind in OpenSCAD, there's a persistent frustration that the only in-built method to extrude a shape both vertically and rotationally is to use `linear_extrude`, which keeps the extruded polygon parallel to the XY plane. Ideally, when creating threads (or similar helical structures, like corkscrews, corkscrew marble channels, etc), it would be possible to use 2D geometry to describe the vertical cross-section of the helix and extrude that along a helical path. 

In a similar way that the `angle` parameter of `linear_extrude` extends linear (vertical) extrusion to do rotation as well, an optional `height` parameter for `rotate_extrude` could extend rotate extrusion to include vertical translation as well.

Common solutions involve implementing the rotation in OpenSCAD using `hull()` (which only works with convex cross-sections, or using `linear_extrude()` as in [this user-created library](https://www.thingiverse.com/thing:1958354), but these methods are painfully slow, and ad hoc for a function that is useful in many modeling contexts.

**Additional context**
I have implemented the `height` parameter in my copy of openscad (PR incoming), and the speedup for spiralizing a semi-complex (non-convex) thread cross-section is 100x in preview mode compared to the linked `spiral_extrude` library and 200x compared to my home-brewed `hull()`-based implementation (done by decomposing the cross-section into convex positive shapes and convex negative shapes, and differenceing the hulls). For a full render, the difference is even starker. The C++ implementation finishes in < 0.15s. Both other methods take 10s of minutes. (I'm still waiting for the faster one to finish, it's been at least 20 minutes. I can give actual benchmark numbers later, if required.)

aprotim:
Fixes openscad/openscad#5011

Details of the motivation are described in detail in the linked issue.


jordanbrown0:
I don't have the energy right now to look at the actual implementation, but two cautions:
* If your Z-increment isn't enough that the sample at 360° is above the sample at 0°, you'll generate a self-intersecting polyhedron, and that's bad.  (I'm not completely sure, but I don't think any other primitive, other than polyhedron(), can generate a self-intersecting polyhedron.)  It might be enough to do a 2D intersect of the shape and itself offset as for 360°, and confirm that the result is empty.
* While the linear extrude scheme works based on slices perpendicular to the Z axis, a rotate_extrude scheme works based on slices perpendicular to the circle that they are rotated through.  Neither is really correct for a helix.  If you want to make a helix with a circular cross-section, your circles need to be tilted so that they align with the angle of the helix.  Without that tilt, the linear extrude scheme needs a base shape sort of like a banana, and the rotate extrude scheme needs something more like an ellipse.  (Not sure if it *is* an ellipse.)

Several libraries have "sweep" functions that will do the right thing.

pca006132:
@jordanbrown0 linear extrude + twist + small number of `slices` can generate self-intersecting polygons, but it should be a known issue.

aprotim:
Existing tests should be fixed. Additional tests testing this functionality pending.

aprotim:
FWIW, here's an example of the kind of design I was trying to match with the above .scad file (the actual dimensions I needed turned out to be somewhat different, but the idea is there). As you can see, the important shapes/dimensions are perpendicular to the axis of the helix. This is true in almost every design/standard I've seen for threads, and something that cannot be accomplished "out of the box" with OpenSCAD. 

I'm offering up the change only because it's such a common design paradigm and because it turned out to be relatively simple to implement with minimal changes and significant gains to speed and geometric accuracy. (I'm also attaching screenshots of spiral_extrude's full render vs rotate_extrue w/ heights preview to demonstrate the geometric artifacts. 

[201802231348444844.pdf](https://github.com/openscad/openscad/files/14411302/201802231348444844.pdf)

spiral_extrude library
rotate_extrude w/ height parameter

pca006132:
How does this compare with `sweep` implementations such as https://github.com/BelfrySCAD/BOSL2/wiki/skin.scad#functionmodule-spiral_sweep?

aprotim:
I'm not sure, partially because I've generated a relatively complex 2D cross-section, and those modules require that to be converted into a vector of vertices, which is non-trivial, as far as I know?

This is actually another reason that I prefer this paradigm: it allows a user to take advantage of all of the available 2D methods to create their cross section, rather than having to handcraft their polygons.

jordanbrown0:
BOSL2 can do fairly complex things with lists of points.  It's a different way of doing things, but similarly powerful.  Note also that BOSL2 has a substantial thread library.

But I'm starting to come around to the idea that this sort of extension to rotate_extrude has value.  Everybody loves threads, and if threads are defined using a vertical cross-section, then this is an easy-to-understand answer.  (Not that my opinion matters much.)

aprotim:
After a cursory Google, I don't see any implementation of sweep that allows you to use the 2D subsystem to specify the input shape. I also can't think of an approach to implement that in "user space", other than the `spiral_extrude` or `hull`-based approximations.

Are you aware of a sweep module that does this, or can you point me at a way to implement one?

aprotim:
> Everybody loves threads

So say we all.

jordanbrown0:
> Are you aware of a sweep module that [is based on the 2D subsystem], or can you point me at a way to implement one?

No, and I agree that with current OpenSCAD it isn't possible.  (It *is* possible with PR#4478.)  The closest you can come is successive hulls, and that doesn't work for concave shapes.

kintel:
I think this is a reasonable approach. While there are opportunities to extend it further by allowing 2D scale and rotation, I think this in isolation should be pretty well-defined.

This definitely needs tests though. Look at existing tests for rotate_extrude and linear_extrude for ideas. We should probably create new test cases for this instead of adding to existing ones, as I guess there are some corner cases to consider.

Also, I'm not sure self-intersections are bad per se, especially when using the Manifold backend, as long as we're producing a manifold topology.

pca006132:
> Also, I'm not sure self-intersections are bad per se, especially when using the Manifold backend, as long as we're producing a manifold topology.

lt is bad, if it does not fall into the epsilon-valid requirement. manifold will probably add a self-intersection removal phase, but probably not in a short period of time.

kintel:
@pca006132 Thanks for jumping in! Just so I get it right: Self-touching volumes are fine (as long as the topology is manifold by keeping coincident vertices separate), but self-intersecting volumes are bad?

pca006132:
yup

kintel:
OK, let's keep the self-intersection test, at least until we have a better way of detecting or avoiding self-intersections (e.g. we could chop up the geometry into < 360 degree chunks and union them together).

I'd say let's merge once we have sufficient tests.

gsohler:
isn't same as https://github.com/openscad/openscad/pull/4538 ?

only difference it that the parameter name is called 'translate' and there is also a 'scale'  parameter

kintel:
Oh, hadn't paid attention to that one.
@aprotim could you take a look at that other PR and perhaps discuss with the other author how to best proceed from a technical perspective?

degroof:
I was just thinking about something similar. In my case, I thought a "pitch" parameter added to rotate_extrude would be handy. Would default to 0 and would represent the Z value added per 360 degrees. 

But, yeah, there's the issue of self-intersection if the pitch is less than the width of the 2D object being extruded.

gsohler:
you can easily fix the self intersection problem by extruding a limited
angle only, but do that any times, finally merging all the parts.
this is performed in
https://github.com/openscad/openscad/pull/5110
This PR suggests using a "v" vector which shows the direction of the spiral
axis

degroof:
I've been playing around with creating helical objects using linear_extrude and a horizontal cross-sections of the objects, and adding a twist that's 360°×height×pitch. Threaded bolts turned out to be fairly trivial, with the horizontal cross-section being essentially a polar plot of the thread profile. A coil, on the other hand, was a bit more interesting. The cross-section turns out to be a kidney bean or C shape, depending on the pitch (see below). 

This approach (as far as I can tell) avoids the whole self-intersection issue, but seems to handle a wide variety of helical constructs without having to mess around with generating polyhedrons from scratch.




```
//cross-sectional diameter of coil
d1=4;
//median diameter of coil
d2=15;
//height
h=70;
//pitch
p=6;
//granularity
gr=4;

//twist
tw=360*h/p;
//circular factor
fc=360/p;

//horizontal slice polygon for coil
coilSection=[for(a=[1:gr:360])
  let(r1=d1/2*cos(a)+d2/2,
    a1=d1/2*sin(a)*fc,
    x1=r1*cos(a1),
    y1=r1*sin(a1))
  [x1,y1]
];

linear_extrude(height=h,twist=tw,slices=tw/gr)
polygon(coilSection);
```

Cross-section of a coil:
image

Resulting coil:
image

jordanbrown0:
> with having to mess around with generating polyhedrons from scratch.

[ Assuming that you meant "without". ]

Yes.  The question is which is easier for you: defining the polyhedron, or defining the correct cross-section.

degroof:
Yep. Should've been "without". Fixed now.

And, yeah, defining the correct cross-section can been tricky. I suppose some sort of library or built-in function could simplify that. I've already made a simple thread library for my own use. The coil code I posted above only generates a circular profile. Not sure how difficult it'd be to extend that to the general case of using any polygon as the profile (e.g. a "slinky" spring with a rectangular profile).

PR #5013, opened 2024-02-25
Add manifold corner case tests

kintel:
The goal of these tests is to create some good scenarios for tricky manifold objects, in particular manifold where the same vertex position is used by different topological neighborhoods, making it hard to reconstruct manifolds from a polygon soup.

TODO:
* [x] Look at #5001 to see if there are more corner cases we could introduce -> defer; needs more investigation
* [x] Establish theoretical ability to maintain these manifolds in various file format: 3MF, OBJ, OFF
* [x] Test import and export of the capable file formats.
* [x] Take note of failures which we should, theoretically, be able to fix using Manifold and write up tickets: #5022
* [x] Optional: Perform STL export/import to see what fails
* [x] Optional: Is there a way we can verify that a file is exported as a proper manifold (i.e. doesn't collapse vertices with equal spatial positions)? This is interesting because OpenSCAD attempts to repair imported files, and with the current export-import tests, we cannot distinguish between an imported manifold vs. an imported repairable non-manifold. -> we could do this per file format, but it's quite some work to pull off. Alternatively extract our importers into a separate, small, test program, which is also a fair bit of work -> defer.


pca006132:
I think maybe we should write some code that directly converts `polyhedron` primitive into `manifold`. For some reason the round-trip to `SurfaceMesh` can sometimes make things worse...

kintel:
> I think maybe we should write some code that directly converts `polyhedron` primitive into `manifold`. For some reason the round-trip to `SurfaceMesh` can sometimes make things worse...

Yeah, the goal of theses tests is to establish some good failures to guide further development.

Issue #5015, opened 2024-02-26
OBJ import bug: Index out of range


kintel:
The OBJ importer uses PolySetBuilder to track vertices. This causes duplicate vertices to be merged, but the indices used are taken directly from the file, indexing unmerged vertices, and if any duplicate vertex is found in the file, it will result in an index out of range error.

Example file (two cubes touching in one vertex):
```
v 0 0 0
v 0 0 1
v 0 1 0
v 0 1 1
v 1 0 0
v 1 0 1
v 1 1 0
v 1 1 1
v 1 1 1
v 1 1 2
v 1 2 1
v 1 2 2
v 2 1 1
v 2 1 2
v 2 2 1
v 2 2 2
f  4 1 2
f  1 6 2
f  1 4 3
f  7 1 3
f  6 1 5
f  1 7 5
f  6 4 2
f  4 6 8
f  4 7 3
f  7 4 8
f  7 6 5
f  6 7 8
f  12 9 10
f  9 14 10
f  14 12 10
f  9 12 11
f  15 9 11
f  12 15 11
f  14 9 13
f  9 15 13
f  15 14 13
f  12 14 16
f  15 12 16
f  14 15 16
```

Error importing the file:
```
WARNING: Index /Users/kintel/code/OpenSCAD/openscad/build/tests/output/objpngtest/polyhedrons-touch-vertex.scad.obj out of range in Line 40
WARNING: Index /Users/kintel/code/OpenSCAD/openscad/build/tests/output/objpngtest/polyhedrons-touch-vertex.scad.obj out of range in Line 41
WARNING: Index /Users/kintel/code/OpenSCAD/openscad/build/tests/output/objpngtest/polyhedrons-touch-vertex.scad.obj out of range in Line 42
```

The imported is missing a few triangles because of this.

PR #5016, opened 2024-02-26
Fix for kintel's obj file cube corner testcase .

gsohler:
using information provided by builder.vertexIndex


kintel:
This implementation still doesn’t maintain more complex manifolds, but it’s an improvement from the current state, so let’s merge it!

gsohler:
This came into my mind during lunch time too.
The Solution you are looking for it the plain dumb solution without
PolySetBuilder and this without PolySetBuilder..
I just introduced the polysetbuilder everywhere when doing the indexed
mode. Apparantly that was too nice 🤣

kintel:
Fixes #5015

kintel:
I'm preparing some tests to exercise the manifold corner cases, so no action needed right now. I'll create new tickets as needed once I understand where we need to go.

PR #5031, opened 2024-03-05
Issue3649 export filename not remembered

### LINKED ISSUE #3649: 2021.01 - Export filename not remembered
In 2019.05, when exporting it remembers the filename used and presents it again on the next export.

e.g.
Open file `Some fantastic design.scad`
`File/Export`, the pre-filled filename is `Some fantastic design.stl`,
edit that to `Some fantastic design-with-2021-01-a.stl`, save.
<do some stuff>
`File/Export`, the pre-filled filename is `Some fantastic design-with-2021-01-a.stl`,
so it is easy to change 'a' to 'b', save.

In 2021.01, it doesn't remember and just offers `Some fantastic design.stl`.

The only mention in the release is 'Fix export format name handling',
a search here found #3029, that didn't get much review at the time.

I believe the original behaviour is better. 
Other's may differ?? Perhaps as an option?

otsakir:
Reviewed the proposed export path algorithm to work better per-tab (#3649)

    * Use previous export path for this suffix and tab if available.
    * Fallback to "defaultExportDir" preference.
    * If this is not available either use same path as .scad file.
    * Fallback to system wide setting if .scad has not been saved yet.

Also adds "advanced/defaultExportDir" preference option.

This partly fixes #3649. It covers only the "path" part. Suggesting also the previous _basename_ remains to be done (discussed further etc.).

### DIFF
--- a/src/gui/Preferences.ui
+++ b/src/gui/Preferences.ui
@@ -2303,6 +2303,20 @@
                  </item>
                 </layout>
                </item>
+               <item>
+                <layout class="QHBoxLayout" name="horizontalLayoutDefaultExportDir">
+                 <item>
+                  <widget class="QLabel" name="labelDefaultExportDir">
+                   <property name="text">
+                    <string>Default Export Directory</string>
+                   </property>
+                  </widget>
+                 </item>
+                 <item>
+                  <widget class="QLineEdit" name="lineEditDefaultExportDir"/>
+                 </item>
+                </layout>
+               </item>
               </layout>
              </widget>
             </item>
--- a/src/gui/Preferences.ui
+++ b/src/gui/Preferences.ui
@@ -1274,7 +1274,6 @@
            <widget class="QLabel" name="labelFeatures">
             <property name="font">
              <font>
-              <weight>75</weight>
               <bold>true</bold>
              </font>
             </property>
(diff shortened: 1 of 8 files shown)


src/gui/Preferences.ui
```diff
@@ -2303,6 +2303,20 @@
...
+                <layout class="QHBoxLayout" name="horizontalLayoutDefaultExportDir">
+                 <item>
+                  <widget class="QLabel" name="labelDefaultExportDir">
+                   <property name="text">
+                    <string>Default Export Directory</string>
+                   </property>
+                  </widget>
+                 </item>
+                 <item>
+                  <widget class="QLineEdit" name="lineEditDefaultExportDir"/>
```
kintel, line 2316: This UI element also needs to offer a way for users to select a folder using a file dialog. Manually entering a folder name isn't very user friendly

kintel:
Tested this on macOS. Looks good except for the preference UI mentioned in my review.
Someone should give this a spin on Windows too..

otsakir:
Added directory selection dialog triggered with a browse button. Currently it looks like this on Linux:

image

In terms of behavior, if the textbox already contains a directory it start from there. Otherwise, it uses `PlatformUtils::userDocumentsPath()` as a starting point.

otsakir:
Any update on that ? Any more fixes improvements needed ? 

@kintel

kintel:
I think we're only waiting on someone with a Windows machine to test it. ..but if it drags on we might just let users figure out how it works on Windows themselves :)

PR #5034, opened 2024-03-06
Add manifold check to PolySet, improve statistics output

kintel:
Also changed `isTriangular` to a private member and added accessors.


kintel:
Abandon this for now

Issue #3809, opened 2021-06-10
SVG rendering incorrect for a path with multiple simple subpaths.


DavidPhillipOster:
The python qrcode package pyqrcode generates SVG using a series of short horizontal and vertical line segments.

Example of creating the .svg attached.

```
import pyqrcode
from pyqrcode import QRCode
getqrcode = pyqrcode.create("A")
getqrcode.svg("qrA.svg", scale=1)
```

OpenSCAD version 2021.01, given the following .scad file:
```
import("qrA.svg");
```
does not parse this correctly compared to Inkscape and Google Chrome. `qrA.svg` could be worked up into a unit test.

_Originally posted by @DavidPhillipOster in https://github.com/openscad/openscad/issues/1740#issuecomment-858815675_

but reduced to a smaller file, with  expected and actual results.

## the SVG file
qrA.svg renamed so I can attach it here: [qrA_svg.txt](https://github.com/openscad/openscad/files/6634299/qrA_svg.txt)

## Expected results

qrA.svg as rendered by Chrome ![A-Chrome](https://user-images.githubusercontent.com/5148308/121594209-9b048c80-c9f1-11eb-9dd8-fb94b87fbb21.png) and Inkscape ![A-Inkscape](https://user-images.githubusercontent.com/5148308/121594219-9cce5000-c9f1-11eb-8974-1eb49254e496.png)

## Actual results

as rendered by OpenSCAD version 2021.01
![A-OpenSCAD](https://user-images.githubusercontent.com/5148308/121594225-9f30aa00-c9f1-11eb-9d90-b3740c3476f2.png)

The problem is most visible in the corners - what should be a square surrounded by a larger square is merged into a single shape.

MichaelAtOz:
Note that QR codes seem to frequently contain points which become non-manifold.
e.g. http://forum.openscad.org/Preview-OK-Render-gets-only-half-tp33306p33315.html 
That won't be the cause of the above issue, but could be your next problem.
As shown, offset() is a work-around.

DavidPhillipOster:
I don't think this is quite the same problem, since although the file renders incorrectly, it previews and renders without error (log attached: [log.txt](https://github.com/openscad/openscad/files/6635199/log.txt) )

One workaround for QRCodes generated in Python anyway: In the python, you can have it generate a `.png` file just as easily as it generates `.svg`, then bring the `.png` into OpenSCAD.

DavidPhillipOster:
Thank you. Yes: 

```
[...]

PR #5036, opened 2024-03-09
Fix svg import (stroke-linecap and stroke-linejoin)

jumostedu:
Fixes wrong default for stroke-linecap, cause of #3809

Updates (wrong) expected png for related test (svgpngtest_spec-shapes-line01). I manually compared the old and the updated png to what Gwenview renders (which for the stuff mentioned in #3809 is the same as e.g. Firefox and other applications) and the updated png matches, the old one does not, because of the stroke-linecap.

Also fixes a wrong value name I noticed in an adjacent function for stroke-linejoin. There seems no test affected.

See commit messages for links to according part in SVG spec.


t-paul (approved):
Nice! Thanks for including the spec links for easier review.

Issue #3460, opened 2020-10-13
Option to use fixed origin for SVG export instead of semi-arbitrarily decide one


dogtopus:
Currently OpenSCAD just uses the minimum xy coordinates that won't crop anything as the origin point, which really messes up use cases that require precise alignment like laser engraving over previously cut objects. It would be nice to have user set an origin point (like (0, 0)) instead of latching on shapes which could have arbitrary size and alignment.

naggie:
I had previously resolved this by exporting a DXF instead. However, there are issues with the DXF support (like curves being a collection of independent lines instead of connected).

I suppose a workaround is to somehow put a small dot at the origin and max x/y.

This would be a welcome change.

jordanbrown0:
Not to ask excessively stupid questions, but why isn't the SVG origin simply the OpenSCAD origin?

t-paul:
Because they have a different Y axis direction.

jordanbrown0:
I exported an SVG from OpenSCAD and looked at it from Inkscape.

It appears that the SVG origin is in the same place as the OpenSCAD origin, but with Y negated (as is kind of required, since SVG uses a "plus Y is down" system, while OpenSCAD uses "plus Y is up").

That's not what the primary Inkscape UI says - its rulers have the origin in the top-left of the drawing - but that appears to be a transformation layered on top of the SVG coordinate system.  I can't say that I fully understand Inkscape's coordinate system, but if I set the viewbox origin to 0,0 (File, Document Properties, Page, Scale, Viewbox, X and Y) then the rulers line up with the SVG and OpenSCAD origins.

Doing that puts the "page" in a perhaps-stupid place, down-right of the origin, in what SVG calls +X/+Y and OpenSCAD calls +X/-Y, but it does put all of the origins in the same place relative to the object..

Does that do what you need?

(Note that Inkscape's display coordinate system got flipped somewhere around 1.0.  I'm referring here to 1.0's default settings, where [0,0] is in the top left and +Y is down.  Ref https://wiki.inkscape.org/wiki/index.php/Release_notes/1.0#Y-Axis_Inversion)

chrysn:
> Because they have a different Y axis direction.

That is merely a matter of 3D orientation – but granted, doing literal coordinate translation would look surprising to users.

I propose resolving this by introducing a `literal_coordinates` argument to the importer that only applies to import format such as SVG that have right-down axis semantics. It would be documented like this:

* **literal_coordinate**
[...]

PR #5038, opened 2024-03-14
SVG import: Add normalize_coordinates option

chrysn:
This adds an option as described in https://github.com/openscad/openscad/issues/3460#issuecomment-1996843751, closing https://github.com/openscad/openscad/issues/3460:

> * **literal_coordinate**
   Boolean. If set to true, the numeric values of x and y coordinates are used without format specific transformations.
>
>   This argument only applies to SVG (or similar future formats): SVG's coordinate system applies the *y* value toward the bottom of the screen on the drawing plane, whereas in OpenSCAD's top view, *y* points toward the top of the screen. By default (i.e., when literal_coordinate=false), the import command flips an SVG file's *y* coordinates, and shifts the whole drawing into the positive area.
>
>   Using literal_coordinate=true is particularly useful when the OpenSCAD script has additional information about positions inside the SVG file's coordinate system.
>
>   This is mutually exclusive with the "center" argument.

There is open point from https://github.com/openscad/openscad/issues/3460#issuecomment-1996843751: Should literal_coordinate also disregard scaling and force the import units to be SVG units, or is it fine as it is (and the documentation just needs slight adjustment)?

### DIFF
--- a/src/core/ImportNode.cc
+++ b/src/core/ImportNode.cc
@@ -61,7 +61,7 @@ static std::shared_ptr<AbstractNode> do_import(const ModuleInstantiation *inst, 
 {
   Parameters parameters = Parameters::parse(std::move(arguments), inst->location(),
                                             {"file", "layer", "convexity", "origin", "scale"},
-                                            {"width", "height", "filename", "layername", "center", "dpi", "id"}
+                                            {"width", "height", "filename", "layername", "center", "dpi", "id", "normalize_coordinates"}
                                             );
 
   const auto& v = parameters["file"];
@@ -123,8 +123,15 @@ static std::shared_ptr<AbstractNode> do_import(const ModuleInstantiation *inst, 
     LOG(message_group::Warning, inst->location(), parameters.documentRoot(), "Unable to convert import(..., origin=%1$s) parameter to vec2", origin.toEchoStringNoThrow());
   }
 
+  const auto& normalize_coordinates = parameters["normalize_coordinates"];
+  node->normalize_coordinates = normalize_coordinates.type() == Value::Type::BOOL ? normalize_coordinates.toBool() : true;
+
   const auto& center = parameters["center"];
   node->center = center.type() == Value::Type::BOOL ? center.toBool() : false;
+
+  if (node->center && node->normalize_coordinates == false) {
+    LOG(message_group::Warning, inst->location(), parameters.documentRoot(), "Import argument center is ignored when normalize_coordinates is set.");
+  }
 
   node->scale = parameters["scale"].toDouble();
   if (node->scale <= 0) node->scale = 1;
@@ -212,7 +219,7 @@ std::unique_ptr<const Geometry> ImportNode::createGeometry() const
     break;
   }
   case ImportType::SVG: {
-    g = import_svg(this->fn, this->fs, this->fa, this->filename, this->id, this->layer, this->dpi, this->center, loc);
+    g = import_svg(this->fn, this->fs, this->fa, this->filename, this->id, this->layer, this->dpi, this->center, loc, this->normalize_coordinates);
     break;
   }
   case ImportType::DXF: {
@@ -250,7 +257,8 @@ std::string ImportNode::toString() const
   }
   stream << ", origin = [" << std::dec << this->origin_x << ", " << this->origin_y << "]";
   if (this->type == ImportType::SVG) {
-    stream << ", dpi = " << this->dpi;
+    stream << ", normalize_coordinates = " << (this->normalize_coordinates ? "true" : "false")
+           << ", dpi = " << this->dpi;
   }
   stream << ", scale = " << this->scale
          << ", center = " << (this->center ? "true" : "false")
--- a/src/core/ImportNode.h
+++ b/src/core/ImportNode.h
@@ -36,6 +36,7 @@ public:
   boost::optional<std::string> layer;
   int convexity;
   bool center;
+  bool normalize_coordinates;
   double dpi;
   double fn, fs, fa;
   double origin_x, origin_y, scale;
--- a/src/io/import.h
+++ b/src/io/import.h
@@ -16,7 +16,7 @@ std::unique_ptr<class PolySet> import_3mf(const std::string&, const Location& lo
 std::unique_ptr<class Polygon2d> import_svg(double fn, double fs, double fa,
 					  const std::string& filename,
 					  const boost::optional<std::string>& id, const boost::optional<std::string>& layer,
-					  const double dpi, const bool center, const Location& loc);
+					  const double dpi, const bool center, const Location& loc, const bool normalize_coordinates);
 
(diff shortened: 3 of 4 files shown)


t-paul:
The option name feels a bit awkward, but I suppose the feature itself is useful. I don't have a good suggestion yet though.

> This is mutually exclusive with the "center" argument.

Why? Is there a technical reason? It does not bring any additional feature as it should be identical to normal centered + y axis flipped but it might be helpful to not impose the exclusivity restriction.

chrysn:
I'm open for any other name.

As for being mutually exclusive with centering, this overrides all the coordinate processing -- scaling, flipping, viewbox adjustment and center offset. It's probably possible to apply center offsets still, but it kind of defeats the purpose of having the raw coordinates applied. All use cases where I've run into this were along the lines of "some tool generates me an SVG with a sidecar file of coordinates, and I want to put them together in 3D". (About 10 years ago it was visualizing PCBs, now it's assembling [laser cut boxes](https://festi.info/boxes.py/) in 3D). If you think it's useful to have, I think I can make it work, but I think it reduces maintainability without a practical benefit.

t-paul:
Oh, it skips transformations even on the objects? I don't think that will work, e.g. Inkscape has the tendency to modify the object transforms instead of the actual coordinates (create circle, make it a bit smaller, move slightly). Skipping those will just import random data.

MichaelAtOz:
> I'm open for any other name.

flip=<true|*false>??

> By default (i.e., when literal_coordinate=false), the import command **FLIPS** an SVG file's y coordinates, and shifts the whole drawing into the positive area.

gsohler:
default is more but only flip. maybe name it "normalize" ?

chrysn:
> Oh, it skips transformations even on the objects?

No, it skips only the transformations that OpenSCAD would apply (the handedness correction / flip, the shift, and any centering). Anything internal to the SVG is preserved.

> normalize

Sounds good, so it'd be `normalize_coordinates=true/false`. This also has the upside that (unlike the "literal" term) it'd be justifiable to stick with the current convention of OpenSCAD units being millimeters (as applying 'literal' coordinates to SVG's complex units system would be troublesome anyway).

chrysn:
With the PR updated, the documentation to be added would now be:

> * **normalize_coordinates**
   Boolean. If true (which is the default), OpenSCAD transforms coordinates from the file to be suitable in its coordinate system.
>
>   This argument only applies to SVG (or similar future formats): SVG's coordinate system applies the *y* value toward the bottom of the screen on the drawing plane, whereas in OpenSCAD's top view, *y* points toward the top of the screen. During normalization of coordinates, the import command flips an SVG file's *y* coordinates, and shifts the whole drawing into the positive area.
>
>   Using normalize_coordinates=false is particularly useful when the OpenSCAD script has additional information about positions inside the SVG file's coordinate system.
>
>   When this is set to false, the "center" argument is ignored, and a warning is shown if center=true.

chrysn:
Is there anything more I should adjust for this? From my point of view it is ready for final review.

remote

(Picture related: Assembling parts from SVG files requires precise control over the coordinate space.)

chrysn:
I've slightly extended the help text (around appearing mirrored), but that's kind of outside the PR. Is there anything I can do (add or remove) to advance this issue?

chrysn:
The continuation of work on boxes.py laser cutter 3D visualization is blocked on this. Can this PR continue, is there anything you'd rather have changed, or is that an issue you do not consider worth fixing?

chrysn:
I'm about to take the project up again where I'd have needed this.

Can this get in if the conflicts are addressed? (I'd rather not waste time if this is clearly rejected, but this PR is in the limbo of "there were review comments, I addressed them, and it's just sitting here").

kintel:
It does sound useful, but I'm not really plugged in enough to SVG to really judge it well right now.
One Q: Is this inherent to SVG or do you think there are the other supported import file formats which could also utilize this feature?

chrysn:
It'd be useful for any format where coordinates are being transformed by default -- but to my knowledge, that's only SVG right now. (To some extent this is probably also because SVG tools tend to be a bit cavalier about coordinate systems -- like, Inkscape < 1.0 used to do a very similar coordinate transformation to what OpenSCAD does; in 1.0 they went with more what's-in-the-file-is-what-you-get numbers).

kintel:
Re. Inkscape 1.0 - that's interesting and some more info about this could be useful; I haven't really used InkScape since the pre-1.0 times.

chrysn:
I've rebased this onto the latest master (squashing the discussed name change as that'd have caused more merge conflicts), run own tests, and verified that the existing tests align with my understanding of how things work.

Does this need an explicit own test in CI, or would it be fine as is?

chrysn:
Ad more useful information, the best I found was the [release notes](https://wiki.inkscape.org/wiki/index.php/Release_notes/1.0#Y-Axis_Inversion); money quote:

> This finally makes the coordinates that a user can see in the interface match the ones that are saved in the SVG data (unit conversions/transformations may be required), and makes working in Inkscape more comfortable for people who are used to this more standard behavior.

It's hard for me to tell whether OpenSCAD's existing behavior was created to emulate old Inkscape, or whether it was to fit the coordinate systems (because OpenSCAD tends to look at 2D things from positive Z, which looks like looking at the SVG in literal/unnormalized coordinates from the backside, i.e. mirrored).

[edit, added] But based on this, I'd update my documentation proposal to:

With the PR updated, the documentation to be added would now be:

> * **normalize_coordinates**
   Boolean. If true (which is the default), OpenSCAD transforms coordinates from the file to be suitable in its coordinate system.
>
>   This argument only applies to SVG (or similar future formats): SVG's coordinate system applies the *y* value toward the bottom of the screen on the drawing plane, whereas in OpenSCAD's top view, *y* points toward the top of the screen. During normalization of coordinates, the import command flips an SVG file's *y* coordinates, and shifts the whole drawing into the positive area. This also coincides with how Inkscape in versions < 1.0 presented the SVG coordinate system.
>
>   Using normalize_coordinates=false is particularly useful when the OpenSCAD script has additional information about positions inside the SVG file's coordinate system. The resulting object will appear to be mirrored upside down in the default view, but correct when viewed from the bottom.
>
>   When this is set to false, the "center" argument is ignored, and a warning is shown if center=true.

chrysn:
@kintel, how can we progress this? This PR is racking up merge conflicts again already. AIU, is just stuck in the review process.

kintel:
I think @t-paul is the best person to give this a go/no-go - I'm not plugged in enough to the SVG ecosystem to have an opinion.

chrysn:
He already tentatively confirmed usefulness in https://github.com/openscad/openscad/pull/5038#issuecomment-1998309982 -- all that was left was naming and other details, to all of which I've made changes or comments.

So from my PoV the next step is either a "no we've changed our minds" or "yes, the responses are good; please rebase and then we merge it".

Issue #5040, opened 2024-03-15
Chore: remove python 2 specific code


Rotzbua:
There are some EOL python 2 specific (compatibility) code left in the scripts which can be remove.

### Example:

```python
from __future__ import print_function
```

### Tasks

- [x] remove compatibility imports
- [x] remove outdated imports
- [x] new classes declaration
- [x] update percent format
- [x] update format specifiers
- [x] simplify to `yield from`
- [x] redundant `open` modes
- [x] fix invalid escape sequences

PR #5043, opened 2024-03-15
chore(python): migrate string format

Rotzbua:
From percent to f-string.

### Reference

Mentioned in #5040

### DIFF
--- a/tests/shouldfail.py
+++ b/tests/shouldfail.py
@@ -59,7 +59,7 @@ sys.stdout.flush()
 try:
     result = subprocess.call(export_cmd)
 except (OSError) as err:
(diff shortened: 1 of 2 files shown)


t-paul:
Would it make sense to go straight to f-strings? We probably are at >= Python 3.6 by now. I'm not sure if there's some sort of official view on string.format() vs. f-strings though.

t-paul:
Agreed, it seems the current minimum is 3.8. Added to the main readme via #5050.

* Ubuntu focal (20.04LTS): 3.8.2-0ubuntu2
* Debian 11 (bullseye): 3.9.2-3

PR #5068, opened 2024-03-25
Replace deprecated Q_OS_MAC and Q_OS_MACX

cjmayo:
Only the Q_OS_MACX ones are creating deprecation warnings at the moment, but that is something anyway. But get ahead and be consistent.

Deprecated in Qt 5:

https://doc.qt.io/qt-5/qtglobal.html#Q_OS_MAC
https://doc.qt.io/qt-5/qtglobal.html#Q_OS_OSX


kintel (approved):

kintel:
Merge when green

PR #5080, opened 2024-04-09
backport linear_extrude with vector to upstream, fixes #5079

### LINKED ISSUE #5079: Linear_extrude with 3d vector as height yields inapropriate result
**Describe the bug**
height parameter of linear_extrude incorrecly used

**To Reproduce**

Use this code:

linear_extrude(height=[0,0,10]) square(1);

It always extrudes 100 for all vector values

**Expected behavior**
Either output an error or create oblique prism 

linear_extrude(height=[0,0,10]) square(1);

Windows version 2023.12.28 (git bdb7b78ac) and probably many others

If applicable, add the content of the Help->Library Info window, or the output of running the `openscad --info` command.

**Additional context**

I am happy that anybody with scad parsing experience addresses the bug

gsohler:
Creating Oblique prisms like that:

linear_extrude(height=[1,1,1]) square([1,2]);

Turns a bug into a feature


src/core/LinearExtrudeNode.cc
```diff
@@ -89,7 +89,11 @@ static std::shared_ptr<AbstractNode> builtin_linear_extrude(const ModuleInstanti
   }
 
   if (parameters["height"].isDefined()) {
-    parameters["height"].getFiniteDouble(node->height);
+    if(!parameters["height"].getVec3(node->height[0], node->height[1], node->height[2], 0.0)) {
```
kintel, line 92: I wonder if we should name this parameter differently than "height" now that we support a vector.

gsohler:
Even though "height" is not the correct name for the new function, I
believe it's most intuitive.
When I wanted to try to create an oblique prism, I boldly tried to assign a
vector to height parametrer (and failed unexpectedly)
OpenSCAD also refers  rectangles as "square", ellipses as "circle" and
prisms as cube.
You have ownership of the repo and the right to decide. Personally I
don't know a good name. If you prefer another name, please let me know.

thehans:
The first thing that comes to my mind is to just call the vector `v`.
Much like the form for rotating about an arbitrary axis:  `rotate(a, v)`

That also raises the question of what happens if both height and v are provided.  
 - One option would be that one overrides the other (as in `circle(r=1, d=10);`, `d` just wins).
 - Another option would be to normalize the vector and then multiply it by a scalar height

t-paul:
I'm not sure using `height` is a good idea, another potential extension is passing a `function(x, y)` determining the actual height at that point. Having then a vector producing something totally different seems confusing.

If we use `v`, could that then maybe just be a direction vector instead of also defining the height? So extrusion along that vector but still using `height`?

kintel:
Generally, I think this is a good addition, and the code is pretty easy.
Before merging I suggest adding sufficient tests to test this new feature, including any corner cases.
We should also write a ticket to remember documenting this.

src/core/LinearExtrudeNode.cc
```diff
@@ -87,9 +87,27 @@ static std::shared_ptr<AbstractNode> builtin_linear_extrude(const ModuleInstanti
...
   }
-
-  if (parameters["height"].isDefined()) {
-    parameters["height"].getFiniteDouble(node->height);
+  double height = 100.0;
+  Vector3d v(0,0,1);
+
+  if (parameters["v"].isDefined()) {
+    if(!parameters["v"].getVec3(v[0], v[1], v[2], 0.0))
+      LOG(message_group::Error, "v when specified should be  a vector.");
```
kintel, line 95: If the parse failed, we should probably reset the vector to the default (0,0,1). Otherwise, the vector could be partially assigned.

src/geometry/GeometryEvaluator.cc
```diff
@@ -1130,12 +1130,12 @@ static std::unique_ptr<Geometry> extrudePolygon(const LinearExtrudeNode& node, c
 
   const Polygon2d& polyref = is_segmented ? seg_poly : poly;
 
-  double h1, h2;
+  Vector3d h1, h2;
   if (node.center) {
-    h1 = -node.height / 2.0;
-    h2 = +node.height / 2.0;
+    h1 = Vector3d(-node.height[0] * 0.5, -node.height[1] * 0.5, -node.height[2] * 0.5);
+    h2 = Vector3d(node.height[0] * 0.5, node.height[1] * 0.5, node.height[2] * 0.5);
```
kintel, line 1136: You shouldn't need to change this. Eigen Vector3d supports the `/` operator.

gsohler: Yep, only unary plus operator is not supported by Eigen

src/core/LinearExtrudeNode.cc
gsohler: An error is created anyway, so it should not matter, but vector is reset now.

tests/data/scad/3D/features/linear_extrude-tests.scad
```diff
@@ -19,6 +19,10 @@ translate([19,20,0]) linear_extrude(height=20, twist=45, slices=10) square([10,1
 
 translate([0,-15,0]) linear_extrude(5) square([10,10]);
 
+translate([15,-15,0]) linear_extrude(v=[3 ,4 ,5]) square([10, 10]);
+
+translate([30,-15,0]) linear_extrude(height=10, v=[3, 4, 5]) square([10,10]);
```
kintel, line 24: Since unspecified components of `v` in linear_extrude defaults to zero, we should test 1- and 2- dimensional v values as well.

src/core/LinearExtrudeNode.cc
```diff
@@ -87,9 +87,29 @@ static std::shared_ptr<AbstractNode> builtin_linear_extrude(const ModuleInstanti
...
   }
+  double height = 100.0;
+  Vector3d v(0,0,1);
 
-  if (parameters["height"].isDefined()) {
-    parameters["height"].getFiniteDouble(node->height);
+  if (parameters["v"].isDefined()) {
+    if(!parameters["v"].getVec3(v[0], v[1], v[2], 0.0)) {
+      v=Vector3d(0,0,1);
+      LOG(message_group::Error, "v when specified should be  a 3d vector.");
```
thehans, line 96: extra space between `be  a`

kintel (approved):

src/core/LinearExtrudeNode.cc
```diff
@@ -87,9 +87,29 @@ static std::shared_ptr<AbstractNode> builtin_linear_extrude(const ModuleInstanti
     node->filename = filename;
     handle_dep(filename);
   }
+  double height = 100.0;
+  Vector3d v(0,0,1);
 
-  if (parameters["height"].isDefined()) {
-    parameters["height"].getFiniteDouble(node->height);
+  if (parameters["v"].isDefined()) {
+    if(!parameters["v"].getVec3(v[0], v[1], v[2], 0.0)) {
```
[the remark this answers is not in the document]
t-paul: This breaks FreeCAD import. Maybe we should revisit this?

gsohler:
Of course,
Just explain a bit better what's the issue and why it breaks now ..

i am quite sure, that height is still being parsed.

PR #5095, opened 2024-04-21
Build fix: Make non-experimental build work again

### LINKED ISSUE #4875: Set up and run non-experimental tests on the CI
We should run tests for non-experimental OpenSCAD builds regularly.
Suggestion:
* Add config to Build and test OpenSCAD in non-experimental mode for one system/variant for each of Linux, Windows and macOS
* Run non-experimental tests weekly, perhaps choose a different day per platform

kintel:
Also adds a workflow to execute on each push to master, which fixes #4875

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -41,7 +41,7 @@ jobs:
     - name: Dependency fixup
       run: sudo apt-get install gettext
     - name: Build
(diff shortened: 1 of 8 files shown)


t-paul:
I've started an additional workflow that uses the build flow described in the `README.md`, specifically using `get-uni-dependencies.sh` on a bare Ubuntu. I'll activate once this is in place with the build fixes.

PR #5096, opened 2024-04-21
Don't use PolySetBuilder for linear_extrude

kintel:
This will make yield manifold meshes for corner cases of linear_extrude. This will allow us to process geometry which is currently failing.

Split out linear_extrude-related functions into a separate `linear_extrude.cc` file.

Partially fixes #5046

PR #5110, opened 2024-04-26
New parameter "v" for rotate_extrude 

gsohler:
The parameter "v" in rotate_extrude, similar to linear_extrude defines an extra linear direction of extruion which makes it easy to design helical screws and other fancy stuff.
An extra parameter method, which defaults to "center" defines the per-point elevation depending on the rotated angle. With "center" the center of gravity is calculated on the 2 objects, wherease for setting " linear" steepnees decreases 1/x from the v axis outwards
"center" looks better whereas "linear" is more natural.

image


src/geometry/GeometryEvaluator.cc
```diff
@@ -1250,67 +1249,72 @@ static void fill_ring(std::vector<Vector3d>& ring, const Outline2d& o, double a,
...
-
-  if ((max_x - min_x) > max_x && (max_x - min_x) > fabs(min_x)) {
-    LOG(message_group::Error, "all points for rotate_extrude() must have the same X coordinate sign (range is %1$.2f -> %2$.2f)", min_x, max_x);
-    return nullptr;
-  }
 
-  fragments = (unsigned int)std::ceil(fmax(Calc::get_fragments_from_r(max_x - min_x, node.fn, node.fs, node.fa) * std::abs(node.angle) / 360, 1));
 
-  bool flip_faces = (min_x >= 0 && node.angle > 0 && node.angle != 360) || (min_x < 0 && (node.angle < 0 || node.angle == 360));
+  double fact=(node.v[2]/node.angle)*(180.0/G_PI);
```
kintel, line 1260: G_PI: That's the first time I've seen that. No idea where that's coming from, but it feels wrong.

I think the cleanest way to achieve this prior to C++20 is:
```
#define _USE_MATH_DEFINES
#include <cmath>
```

tests/data/scad/3D/features/rotate_extrude-angle.scad
```diff
@@ -26,6 +26,22 @@ rotate([0,0,90]) {
...
+
+translate([80,0,0])
+rotate_extrude(angle=180*5,v=[10,-10,30],method="linear")
+    translate([5,0]) circle(4,$fn=20);
+
+translate([80,40,0])
+rotate_extrude(angle=270,v=[0,0,40])
+    translate([5,0]) circle(4,$fn=20);
+
+
```
kintel, line 44: This file is getting too crowded, let's start a new file for this new feature.

src/geometry/GeometryEvaluator.cc
```diff
@@ -1330,6 +1334,60 @@ static std::unique_ptr<Geometry> rotatePolygon(const RotateExtrudeNode& node, co
   return builder.build();
 }
 
+static std::shared_ptr<const Geometry> rotatePolygon(const RotateExtrudeNode& node, const Polygon2d& poly){
```
kintel, line 1337: See #5096 - I suggest pulling out all the rotate_extrude tools into a new file, as this file is getting a bit hard to reason about.

kintel (commented):
I didn't look into the actual parameters yet, just commented on some surrounding code.

thehans:
I understand the desire for introducing vertical offset to rotate_extrude, to allow for much simpler creation of threads and springs etc.
However I'm not sure if a fully configurable vector is the way to go.  I can't imagine a use case where I would want to offset anything other than perfectly vertical.  Has this been a requested feature or is this simply being done to match the recent `linear_extrude` work?

I am starting to think that these sort of cases (`linear_extrude` included) would better handled via a separate affine transformation of [`shear`](https://en.wikipedia.org/wiki/Shear_mapping).
I believe that the result of a vertical offset, and then `shear` would give the same result, would it not?

I think the existing `multmatrix` module can be used with a user-constructed matrix to do this, however a builtin module with more convenient parameters would certainly be more user-friendly, reducing cognitive load to construct such a matrix.

kintel:
One benefit I can think of is that if we know the shear in advance, we can tessellate slightly better.

PR #5111, opened 2024-04-26
Fix initial GUI setup.

t-paul:
Change default for "Auto-Reload and Preview" to false, so new installations have to manually turn it on.

The initial GUI state (new install / cleared configuration) is very messy:

This changes the default to dock Console/ErrorLog/Animate as tabs and hide the Customizer and ViewportControl docks:

### DIFF
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -665,7 +665,6 @@ MainWindow::MainWindow(const QStringList& filenames)
   const auto windowState = settings.value("window/state", QByteArray()).toByteArray();
   restoreGeometry(settings.value("window/geometry", QByteArray()).toByteArray());
   restoreState(windowState);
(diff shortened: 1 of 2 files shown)


src/gui/MainWindow.cc
```diff
@@ -906,7 +906,7 @@ void MainWindow::loadViewSettings(){
 void MainWindow::loadDesignSettings()
 {
   QSettingsCached settings;
-  if (settings.value("design/autoReload", true).toBool()) {
+  if (settings.value("design/autoReload", false).toBool()) {
```
t-paul, line 909: @kintel what do you think, shall we flip the default? We could go back if we ever get a working cancel button :smile:.

kintel: I feel default auto-reload is a bit tricky until we have a working way to cancel, so I'd vote to make it off by default.

PR #5113, opened 2024-04-26
Linextrude

gsohler:
getting back height parameter for linear_extrude in csg


tests/regression/dumptest/linear_extrude-tests-expected.csg
```diff
@@ -30,52 +30,52 @@ multmatrix([[1, 0, 0, 1], [0, 1, 0, 21], [0, 0, 1, 0], [0, 0, 0, 1]]) {
...
 }
 multmatrix([[1, 0, 0, 0], [0, 1, 0, -25], [0, 0, 1, 0], [0, 0, 0, 1]]) {
-	linear_extrude(v = [ 0, 0, 0], $fn = 0, $fa = 12, $fs = 2) {
+	linear_extrude(height = 0, $fn = 0, $fa = 12, $fs = 2) {
 		square(size = [10, 10], center = true);
 	}
 }
 multmatrix([[1, 0, 0, 0], [0, 1, 0, -25], [0, 0, 1, 0], [0, 0, 0, 1]]) {
-	linear_extrude(v = [ 10, 10, 0], $fn = 0, $fa = 12, $fs = 2) {
+	linear_extrude(height = 14.1421, v = [ 10, 10, 0], $fn = 0, $fa = 12, $fs = 2) {
```
kintel, line 78: How did this happen? height is negative in the test.

kintel:
I guess the purpose here is to allow old importers to read output of new files.
However, this is a bit dangerous, as old importers may now misinterpret the files, in case where we output both `height` and `v`. Perhaps a safer way would be to maintain the same output as input (i.e. drop height if it wasn't specified)?

I haven't thought through this in detail though.

gsohler:
> I guess the purpose here is to allow old importers to read output of new files. However, this is a bit dangerous, as old importers may now misinterpret the files, in case where we output both `height` and `v`. Perhaps a safer way would be to maintain the same output as input (i.e. drop height if it wasn't specified)?
> 
> I haven't thought through this in detail though.

i think we can fully maintain old functionality:
if height/v shows vertically up, its simply omited and only height is specified.
The plan is to specify the length  of the vector as "height"  and the normalized direction as "v"  ,. of course v is omited most of the times

i will look into your other "negative" bug

kintel:
My concern is this:

`linear_extrude(v=[3 ,2 ,5])`
Is written to .csg as 
`linear_extrude(height = 6.16441, v = [ 0.486664, 0.324443, 0.811107], $fn = 0, $fa = 12, $fs = 2)`

If someone, e.g. the FreeCAD importer imports this file, it will import the wrong geometry, and likely silently ignore the `v` argument.

Not sure how big of a problem this is, as we've made similar changes in the past (e.g. added `angle` to rotate_extrude), but worth being aware of.

kintel (approved):

PR #5117, opened 2024-04-29
Viewport-Control widget: Limit width to at least dynamic minimum size of widget. Add tab order back in.  Fixes #5116

### LINKED ISSUE #5116: Viewport-Control widget can be made too narrow; tab order broken.
**Describe the bug**
PR #5107 improved the Viewport-Control widget by allowing it to morph to a vertical layout when the widget became too narrow for the original horizontal layout.  

However after that change the widget could be made too narrow (issue 1) and look bad because the controls were "cut off" and only partially visible.  #5107 also lost the left-to-right top-to-bottom tab order within the widget (issue 2).  This PR fixes both of these issues.

**To Reproduce issue 1 - widget may be made too narrow**
1. Open a new document in OpenSCAD.
2. If the Viewport-Control isn't visible, select "Viewport-Control" from the Window menu
3. If the Viewport-Control isn't in the right size of the Window, move it there.
4. Resize the Viewport-Control widget to as small as you can get it.
5. Notice that its contents are no longer fully visible (screenshot below).

**Expected behavior issue 1**
The widget width should be constrained to be _at least_ the minimum width necessary to fit the contents and avoid cutting off any of the controls in the window.

**To Reproduce issue 2 - tab order is weird**
1. Open a new document in OpenSCAD.
2. If the Viewport-Control isn't visible, select "Viewport-Control" from the Window menu
3. Set the focus to the first control in the Viewport-Control widget: Aspect Ratio.
4. Hit the tab key repeatedly and watch the focus jump around in the widget in a less than useful order.

**Screenshots**
Issue 1 - Viewport-Control can be made too narrow:

image

Issue 2 - Viewport-Control tab order is weird:

Screen Recording 2024-04-28 at 11 25 07 PM 2024-04-28 23_27_18

 - OS: macOS 13.6.6
 - System: 13" M1 MacBook Pro
 - OpenSCAD Version:  Head of master as of 4/28/2024 (hash: 97184930d36d6d5cb126b5e0e0cd5ca3c606c05f ) that includes #5107

haikusw:
Addresses #5116 - Viewport-Control widget can be made too narrow & tab order was lost is now weird.

PR #5107 improved the Viewport-Control widget by allowing it to morph to a vertical layout when the widget became too narrow for the original horizontal layout.  Very nice!

However, after that change the widget could be made too narrow and look bad because the controls were "cut off" and only partially visible.  #5107 also lost the left-to-right top-to-bottom tab order within the widget. This PR fixes both of these issues.  See #5116 for screenshots and detailed steps.

### DIFF
--- a/src/gui/ViewportControl.cc
+++ b/src/gui/ViewportControl.cc
@@ -106,6 +106,11 @@ void ViewportControl::resizeEvent(QResizeEvent *event)
         gridLayout->addWidget(labelFOV, 3, 0, 1, 1);
         gridLayout->addWidget(doubleSpinBox_fov, 3, 1, 1, 1);
         scrollAreaWidgetContents->layout()->invalidate();
+      } else {
+        const auto width = scrollAreaWidgetContents->minimumSizeHint().width();
+        if (scrollArea->minimumSize().width() != width) {
+          scrollArea->setMinimumSize(QSize(width,0));
+        }
       }
     }
   }
--- a/src/gui/ViewportControl.ui
+++ b/src/gui/ViewportControl.ui
@@ -331,6 +331,19 @@
    </property>
   </action>
  </widget>
+ <tabstops>
+  <tabstop>spinBoxWidth</tabstop>
+  <tabstop>spinBoxHeight</tabstop>
+  <tabstop>checkBoxAspecRatioLock</tabstop>
+  <tabstop>doubleSpinBox_tx</tabstop>
+  <tabstop>doubleSpinBox_ty</tabstop>
+  <tabstop>doubleSpinBox_tz</tabstop>
+  <tabstop>doubleSpinBox_rx</tabstop>
+  <tabstop>doubleSpinBox_ry</tabstop>
+  <tabstop>doubleSpinBox_rz</tabstop>
+  <tabstop>doubleSpinBox_d</tabstop>
+  <tabstop>doubleSpinBox_fov</tabstop>
+ </tabstops>
  <resources/>
  <connections/>
 </ui>


t-paul:
Thanks for the tab fixes. Forcing fixed sizes on the scollarea has shown all sorts of funny effect, I'll try if that works without surprises.
Also I'm not sure it's actually better to enforce this. Sometimes it might be easier to just move the window out of the way even though it's clipping instead of closing it completely. But I'm not too focused on one solution, so if it works, we can go with it.

haikusw:
Thanks.  I tried to keep it to a very focused fix.  

I tested as much as I can on a mac. Tried moving the widget to other places in the window and combining it with other widgets in the same area; it all seems to work without any weirdness.  I don't have other platforms to test on though, so that would be my main concern.  

One possible change, now that you've pointed out there may be issues with oddities relating to fixed sizes and scrollarea, could be to rest the minimum to none when switching back to the horizontal layout.  I did not see any issues in my testing, but if you see any I'd be happy to add this change.

I think cutting off controls looks pretty bad.  Since all, or nearly all, of the other wedges prevent this from happening it seems best to be consistent.

Since  you fixed the main issue of having a minimum width that was very wide (thanks!) it won't get in the way nearly as much.

haikusw:
Thought I might add that I initially put in a hard-coded empirically determined minimum width to prove the approach of setting the minimum to fix this would work (it worked).  For the final implementation I use a dynamically calculated minimum (`scrollAreaWidgetContents->minimumSizeHint().width()`) to support potential cross platform control/font/widget sizing differences (accessibility settings the user may have chosen to increase font size, etc).

t-paul:
I'm not sure why, but it does not seem to make a difference on Linux (using the AppImage build).

image

haikusw:
Weird.  Since it's unlikely (?) that Qt behaves differently on Linux, my first thought is that the build didn't actually rebuild it, or didn't use the right branch or something.  

I'm not sure what "the AppImage build" means but I'll take a look and maybe spin up an Ubuntu VM and see how difficult it is to get it building there.

t-paul:
AppImage build in this case refers to the automatically built one which can be found via the green checkmark on the commit.

image

This *should* always build on a clean system.

I'll try to find time to check locally too, maybe adding some temporary debug output for actual values. While the general behavior of the widgets should be mostly same on platforms, there's potential for small differences in regard to specific values and/or event ordering.

haikusw:
I created a new Ubuntu 20.04.2 ARM64 VM, slogged through updating it enough to build openscad, and built the branch `haikusw:viewport-control-layout-minwidth` and I see the same thing you do - no change.

I do not have experience developing under Linux so it's going to require a lot of ramp up (what developer tools to debug? hopefully something other than gdb in terminal…) for me to look into why this isn't working on Linux, more than I can devote to this at this point.

Since the behavior on Linux is no worse than what it was without this PR, and this PR does improve macOS (and Windows?) to bring their UI behavior up to macOS platform standards (and Windows? idk), perhaps you'd consider merging this even if it doesn't fix this issue there?

t-paul (approved):
It does not seem to do any harm and at least works on one platform. So I'd say lets get this state merged and also get the tab order fix. If we find a cross platform solution, we can always use another PR based on this.

haikusw:
Thanks.  I _almost_ got VS Code to debug the project on Ubuntu last night but not quite (builds and launches it, but no debugging).  I'll try again when I get a chance and see if I can't track down what's different on linux.

PR #5127, opened 2024-05-09
Show save dialog when exiting via shortcut or menu item

timgblack:
Ctrl-Q on Linux doesn't prompt to save files like the normal window close button, and I managed to lose work several times over it.

Digging into the openscad (and Qt) source code revealed that File -> Quit has the same behaviour, and it most likely happens on OS X too (Cmd-Q).

I'm no C++ developer, but the fix appeared to be relatively simple once I found it.

### DIFF
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -3558,7 +3558,7 @@ void MainWindow::setFont(const QString& family, uint size)
 void MainWindow::quit()
 {
   QCloseEvent ev;
-  QApplication::sendEvent(QApplication::instance(), &ev);
+  MainWindow::closeEvent(&ev);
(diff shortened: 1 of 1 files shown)


kintel:
As it's possible to open multiple windows, are you sure that your fix doesn't just close the current window instead of quitting the application?

timgblack:
There's definitely some unintended interaction with multiple windows open! I'll see if I can figure out the best solution.

kintel:
Sorry, this got buried. I'm still struggling a bit figuring out how this works. In particular,  `CocoaUtils::endApplication()` now seems to be called every time a window closes, which is not quite expected.

Also, when using Qt6, this closing issue is no longer present, so I wonder what changed in Qt..

PR #5144, opened 2024-05-19
Test on Ubuntu 24.04 and Qt 6

cjmayo:
Some tidying up of github-ci-linux-get-dependencies.sh and its use with old and misnamed Ubuntus and an attempt to avoid similar problems in future by defaulting to ubuntu-latest (although that still does have to be defined in one place).

gcovr completes, but the reports show no coverage - which actually isn't any different to the existing tests on other PRs as far as I can see.

`gcovr -r ../src` does seem an improvement on `gcovr -r ..` in that it doesn't collect source files from the build directory or submodules.

### DIFF
--- a/.github/workflows/linux-tests.yml
+++ b/.github/workflows/linux-tests.yml
@@ -18,7 +18,7 @@ jobs:
             distro: focal
             qt5default: true
           - os: ubuntu-22.04
-            distro: kinetic
+            distro: jammy
             qt5default: false
 
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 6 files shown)


.github/workflows/linux-tests.yml
```diff
@@ -12,45 +12,54 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-20.04, ubuntu-22.04 ]
+        os: [ ubuntu-20.04, ubuntu-22.04, ubuntu-24.04 ]
+        qt: [ qt5, qt6 ]
+        exclude:
```
kintel, line 17: Why do we exclude these? A comment explaining why and if applicable when it can be turned back on would be great

kintel (approved):
LGTM.
@t-paul Let us know if you have any comments/wishes before merging

Issue #5153, opened 2024-05-28
3DConnexion Bluetooth Wireless SpaceMouse


rcarmo:
**Describe the bug**
The new SpaceMouse Wireless with Bluetooth is not supported in macOS

**To Reproduce**
1. Connect the SpaceMouse via Bluetooth, USB or the Universal Receiver.
2. Enable HIDAPI
3. Restart OpenSCAD
4. Try using the mouse

**Expected behavior**
The SpaceMouse should work and update the 3D view accordingly

**Code reproducing the issue**
N/A

**Screenshots**
N/A

**Environment and Version info (please complete the following information):**
 - OS: macOS 14.5 (23F79)
 - System: Mac mini M2 
 - OpenSCAD Version - both OpenSCAD-2021.01 and OpenSCAD-2024.05.27 from the main website

**Library & Graphics card information**
N/A

**Additional context**

I've gone and captured the HID logs in Bluetooth mode:

```
0: HidApiInputDriver::open()
2: Enumerating HID devices...
105: D: 256f:c63a | path = DevSrvsID:4295023432, serial = , manufacturer = 3Dconnexion, product = SpaceMouse Wireless BT
105: D: 256f:c63a | path = DevSrvsID:4295023432, serial = , manufacturer = 3Dconnexion, product = SpaceMouse Wireless BT
105: D: 05ac:0250 | path = DevSrvsID:4294973007, serial = com.3dconnexion.3DxMacCore-DriverKit-VirtualHIDKeyboard, manufacturer = com.3dconnexion, product = 3Dconnexion Virtual Keyboard
105: D: 05ac:0250 | path = DevSrvsID:4294973007, serial = com.3dconnexion.3DxMacCore-DriverKit-VirtualHIDKeyboard, manufacturer = com.3dconnexion, product = 3Dconnexion Virtual Keyboard
105: D: 05ac:0250 | path = DevSrvsID:4294973007, serial = com.3dconnexion.3DxMacCore-DriverKit-VirtualHIDKeyboard, manufacturer = com.3dconnexion, product = 3Dconnexion Virtual Keyboard
105: D: 05ac:0250 | path = DevSrvsID:4294973007, serial = com.3dconnexion.3DxMacCore-DriverKit-VirtualHIDKeyboard, manufacturer = com.3dconnexion, product = 3Dconnexion Virtual Keyboard
[...]

PR #5155, opened 2024-05-28
Add USB ID for 3Dconnexion Space Mouse Wireless BT

rcarmo:
Following up on #5153 - I cannot build this right now, but the USB ID should be of general use.


t-paul (approved):
Lets just try it out :rocket:.

PR #5156, opened 2024-05-28
Test building with Qt 6 on Windows

cjmayo:
Windows test. We do want Qt 5 and Qt 6?

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -8,8 +8,11 @@ on:
 
 jobs:
   build:
+    strategy:
+      matrix:
+        qt: [ qt5, qt6 ]
     runs-on: windows-latest
-    name: windows-latest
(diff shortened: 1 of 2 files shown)


kintel:
Technically, the tests don't exercise Qt, so it's a bit unnecessary to test both versions, but it's important that we can build and launch it so I think this is good for now.

t-paul (approved):
Yep, great step to get things moving. The challenge is with the official build based on MXE, AppImage and the distro ones. Do we have some minimum qt6.x requirement?

t-paul:
Hmm, I have not yet found a PPA for 20.04 with Qt6, but https://github.com/state-of-the-art/qt6-docker looks like a promising reference. MXE has Qt6.7, hopefully that works with QScintilla. I'm not sure yet how to handle the Distro builds, probably starting Qt6 with Debian 12 / Ubuntu 24.04.

PR #5162, opened 2024-06-01
Run apt-get update for Debian based OS's in uni-get-dependencies.sh

cjmayo:
Hopefully (= I have tested. But it won't show in the PR checks) this fixes the recent linux-build-matrix.yml failures. Also removes some code for long unsupported distros.


t-paul (approved):
Thanks!

PR #5163, opened 2024-06-03
Look up the Ubuntu code name in github-ci-linux-get-dependencies.sh

cjmayo:
Save an additional change every time ubuntu-latest is updated.

An unknown code name causes an error.

---

OK it's only every 2 years but this felt a bit half-done at the time.


t-paul (changes requested):
The [man page](https://www.linux.org/docs/man5/os-release.html) says `/etc/os-release` should be preferred. I'm not sure we need the fallback though.

t-paul (approved):
Thanks!

PR #5165, opened 2024-06-06
Reseed rng at compile for unseeded initial vals

rcolyer:
The following code should have random values at each run for the first 3 sets of values output, so that unseeded random values can be utilized easily in combination with seeded random values, without the seed sticking BETWEEN runs.  This PR resolves this by calling a new initialize_rng function when a compile is done, restoring the prior behavior of reliably having randomness before a seed is used.

```
echo(rands(0, 100, 3));
echo(rands(0, 1, 1, 42));
echo(rands(0, 100, 3));
```

### DIFF
--- a/src/core/Builtins.h
+++ b/src/core/Builtins.h
@@ -8,6 +8,8 @@
 class AbstractModule;
 class BuiltinFunction;
 
+void initialize_rng();
+
 class Builtins
 {
 public:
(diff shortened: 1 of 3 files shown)


t-paul (approved):
Right, so this is by execution order. I guess there is not much way around that without introducing API changes.
I agree the current behavior of the seed sticking across compiles is a bug.

PR #5171, opened 2024-06-12
Add daylight gem and nocturnal gem color schemes

rcolyer:
This adds two new color schemes, paired for dark and light background preferences, for an overall prettier OpenSCAD render.

The nocturnal gem color scheme:
demo_nocturnal_gem

The daylight gem color scheme:
demo_daylight_gem


t-paul (approved):
Nice!

PR #5173, opened 2024-06-13
Show STL files to import in Open File dialog #5172

### LINKED ISSUE #5172: Allow selecting STL files to import via a file dialog
It is very non-intuitive how to import an STL file into your project. If you don't know that OpenSCAD can do this, from the interface, you'd have no idea you can, or how to do so.

I would like a way to bring up a QFileDialog that has STL files in a filter so that STL files will show in the dialog. Choosing an STL and accepting the file dialog would add the import statement to the editor's cursor's current line.

This could be implemented very simply by just adding a file filter for STL files to the openFile() utility function, however that may confuse the user, since it isn't exactly opening a new file.
Alternatively, a separate menu item and toolbar button could be created for importing STLs (File->Import probably), which would call a different utility function that shows a file dialog specifically for importing.

**Additional context**
https://github.com/openscad/openscad/issues/4791 asks for any file to be openable in the editor. If for some reason the user wanted to open an STL file and have the STLs file contents shown in the editor window, this feature I am proposing would need to be implemented in some way other than with the existing openFile() utility function (and thus not through File->Open, or the open tool button).

keithel:
OpenSCAD supports opening STL files from the file open dialog - doing so will add an import line into the project for that stl file.

However, you have to know exactly what he file's name is and type it exactly to actually get the file open dialog to load it. There is no way to set a filter to show STL files. This adds some additional filters to the file open dialog - one for all supported filetypes (.scad, .csg., .stl), one for just the project files (.scad, .csg), and one for just STLs (*.stl).

Additionally, it supports saving the selected filter between calls to openFileDialog, so the next time you open the file dialog, it shows the last filter you selected.

This is to fix #5172

### DIFF
--- a/src/gui/UIUtils.cc
+++ b/src/gui/UIUtils.cc
@@ -60,8 +60,11 @@ QFileInfoList UIUtils::openFiles(QWidget *parent)
 {
   QSettingsCached settings;
   QString last_dirname = settings.value("lastOpenDirName").toString();
+  static constexpr QLatin1String defaultSelectedFilter {"OpenSCAD Designs (*.scad *.csg)"};
+  static QString filters = QString("Supported Filetypes (*.scad, *.csg, *.stl);;%1;;STL Files (*.stl)").arg(defaultSelectedFilter);
+  static QString lastSelectedFilter(defaultSelectedFilter);
   QStringList new_filenames = QFileDialog::getOpenFileNames(parent, "Open File",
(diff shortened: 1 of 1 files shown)


keithel:
There has been a request to make this be done via a separate file menu entry. I'm working on that.

jordanbrown0:
PR #6341 `"File-picking` template" would enable something like this.

PR #5175, opened 2024-06-13
Calculate Segment Distance instead of Line Distance

gsohler:
This calculates the minimal  Segment to Segment distance instead of Line to Line distance 
Matches more to what's being displayed and thus less confusing

Same goes for Point to Segment-Distance


t-paul (approved):
Maybe at some later point we can have both shown in a measurement window, but it sounds like this is more useful for now.

src/glview/cgal/CGALRenderUtils.cc
```diff
@@ -21,3 +50,23 @@ double calculateLineLineDistance(const Vector3d &l1b, const Vector3d &l1e, const
 	dist_lat=(v2.cross(n)).dot(l2b-l1b)/t;
 	return d;
 }
+
+double calculateSegSegDistance(const Vector3d &l1b, const Vector3d &l1e, const Vector3d &l2b, const Vector3d &l2e, double &dist_lat)
```
kintel, line 54: What is dist_lat?

src/glview/cgal/CGALRenderUtils.cc
```diff
@@ -1,8 +1,37 @@
 #include "CGALRenderUtils.h"
 
+
+int linsystem( Vector3d v1,Vector3d v2,Vector3d v3,Vector3d pt,Vector3d &res,double *detptr)
```
kintel, line 4: Should this rather return a bool
It wouldn't hurt with a comment describing what this function does.

src/glview/cgal/CGALRenderUtils.cc
```diff
@@ -21,3 +50,23 @@ double calculateLineLineDistance(const Vector3d &l1b, const Vector3d &l1e, const
...
+	Vector3d v2=l2e-l2b;
+	Vector3d n=v1.cross(v2);
+	Vector3d res;
+	if(n.norm() < 1e-6) {
+		return calculateLinePointDistance(l1b,l1e,l2b,d);
+	}
+	if(linsystem(v1,n,v2,l2e-l1b,res,nullptr)) return NAN;
+	double d1=res[0], d2=res[2];
+	if(d1 < 0) d1=0;
+	if(d1 > 1) d1=1;
```
kintel, line 67: I think you could just do `d1 = std::clamp(res[0], 0, 1);` to make things a tiny bit easier to read.

src/glview/cgal/CGALRenderUtils.cc
```diff
@@ -1,8 +1,37 @@
...
+}
+
 double calculateLinePointDistance(const Vector3d &l1, const Vector3d &l2, const Vector3d &pt, double & dist_lat) {
-    Vector3d d = (l2 - l1).normalized();
+    Vector3d d = (l2 - l1);
+    double l=d.norm();
+    d.normalize();
     dist_lat = (pt-l1).dot(d);
+    if(dist_lat < 0) dist_lat=0;
+    if(dist_lat > l) dist_lat=l;
```
kintel, line 34: -> `std::clamp((pt-l1).dot(d), 0, 1)`

src/glview/cgal/CGALRenderUtils.cc
gsohler: dist_lat is the length of the  lateral portion of  distance from l1 to pt parallel to  the line

kintel: Sorry, badly formulated question: This argument doesn't seem to be used; is that an oversight?

gsohler: i copied this function from my codes. Many times the caller is also interested in the lateral distance, however not in this case. sounds like you don' t want it for future extensions ;)

gsohler: using this parameter you can e.g. easily determine, if a pt is projecting onto the line.

kintel: Sounds great for a future extension, but it's, IMO, better to leave out unimplemented arguments, and rather leave a comment in the code describing possible future extensions.

PR #5176, opened 2024-06-14
cmake compatibility breaks at 3.5

Obscuretone:
The documentation implies you need 3.5. cmake 3.5 will break compatibility, and is not released yet. This can't be correct.


t-paul:
What do you mean? `cmake` 3.5 was released somewhere April 2016. The current version is 3.30.

https://cmake.org/files/

Obscuretone:
30 > 5. 

I'm an idiot

Issue #5065, opened 2024-03-24
Display colors when rendering using F6


kintel:
Currently, `color()` can be used to set the color of (both negative and positive) objects, but only works in Preview Mode (F5).

Example:
```
difference() {
    union() {
      color("darkred") cube(20);
      color("darkgreen" ) sphere(10);
    }
    color("cyan") cylinder(r=8, h=30);
}
```
<img width="484" alt="Screenshot 2024-03-24 at 13 34 52" src="https://github.com/openscad/openscad/assets/186094/aa3be768-0531-4a60-9cac-d563857b7484">

It would be great if this was extended to also support Render Mode (F6).

Using Manifold, this should be possible, as demoed by https://manifoldcad.org/#Tetrahedron%20Puzzle.
This feature is likely a generalization of #5047

Further topics related to this could be:
* Extend color support to allow importing and exporting objects with color information
   * [Export color models #4311](https://github.com/openscad/openscad/issues/4311)
   * [color(), material() and part() #1608](https://github.com/openscad/openscad/issues/1608)
   * [Allow colours in SVG export #1364](https://github.com/openscad/openscad/issues/1364)

When investigating this, it could be beneficial to look at the following related issues:
* ~~[Preserve original colors of minuend objects that are cut with difference() #4798](https://github.com/openscad/openscad/issues/4798)~~
* ~~[Color ignored by linear_extrude #3621](https://github.com/openscad/openscad/issues/3621)~~
* [Feature request: shaders #3198](https://github.com/openscad/openscad/issues/3198)
* [Add color, linecolor, and line width parameters to projection. #2145](https://github.com/openscad/openscad/issues/2145)
   * No change in capability -> back burner
* [color overrides transparency #1393](https://github.com/openscad/openscad/issues/1393)
* [Difference applies color of removed object to interface #501](https://github.com/openscad/openscad/issues/501)
* [Hierarchical color() behavior #492](https://github.com/openscad/openscad/issues/492)
   * No change in capability -> back burner
* [Hierarchically inherit colors #366](https://github.com/openscad/openscad/issues/366)
   * We could address this now.

gsohler:
My work on that is stuck currently, as
the order of the input sets described in runIndex is not predictable for me
in my place even documentation states, that
they are sorted by originalID.
sometimes their order is A, B, sometimes th eorder is  B,A and I did not
follow the approach by matching the sets by triangle count ;)

hope to get the right inspiration soon.

[...]

PR #5180, opened 2024-06-15
Color support for 3D rendering + export to glTF, Collada, STEP, PLY, X3D

ochafik:
See https://github.com/openscad/openscad/issues/5065

image

This adds color to 3D rendering, fenced by both `assimp` & `manifold` experimental features (and works well w/ `lazy-union`).

```bash
git clone https://github.com/ochafik/openscad --branch color-assimp2 --recursive openscad-colors
cd openscad-colors
# brew install assimp
# apt install libassimp-dev
cmake -B build -DEXPERIMENTAL=1 -DUSE_QT6=1 -DENABLE_ASSIMP=1 && cmake --build build -j
alias openscad=./build/OpenSCAD.app/Contents/MacOS/OpenSCAD
openscad examples/Basics/CSG-modules.scad -o out.glb --enable=assimp --enable=manifold 
```

image

Some details:

- Sets colors on Geometry, ensure it's preserved across most PolySet operations.
- Uses Manifold's OriginalID mechanism to trace original colors as described by elalish@ [here](https://github.com/openscad/openscad/issues/5065#issuecomment-2168257810)
- At the time of exporting files (or OpenGL rendering), just explodes the top-level manifold(s) into one PolySet per color.
- Uses [assimp](https://assimp-docs.readthedocs.io/en/latest/index.html) to export to the following standard formats that support colors:
  - [glTF](https://en.wikipedia.org/wiki/glTF) (`.glb`; binary variant)
  - [Collada](https://en.wikipedia.org/wiki/COLLADA) (`.dae`)
  - [STEP](https://en.wikipedia.org/wiki/ISO_10303-21) (`.stp`)
  - [PLY](https://en.wikipedia.org/wiki/PLY_(file_format)) (`.ply`; binary variant)
  - [X3D](https://en.wikipedia.org/wiki/X3D) (`.x3d`)

Now, some big caveats:
- I've left 3MF aside for now. It's actually trivial to either let `export_assimp.cc` handle it or to update `export_3mf.cc` to support colors (see below), but neither PrusaSlicer nor BambuStudio are doing anything with the result.

  <details>
  
  <summary>Show export_3mf.cc diff to support colors</summary>
  
  ```c++
  static bool append_polyset(const std::shared_ptr<const PolySet> & ps, Lib3MF::PWrapper& wrapper, Lib3MF::PModel& model)
  {
    try {
      auto mesh = model->AddMeshObject();
      if (!mesh) return false;
      mesh->SetName("OpenSCAD Model");
  
      auto & color = ps->getColor();
      if (color.isValid()) {
        LOG(message_group::Export_Warning, "Color: %1$f %2$f %3$f %4$f", color[0], color[1], color[2], color[3]);
        Lib3MF::sColor sc {
          (Lib3MF_uint8) (color[0] * 255),
          (Lib3MF_uint8) (color[1] * 255),
          (Lib3MF_uint8) (color[2] * 255),
          (Lib3MF_uint8) (color[3] * 255)
        };
        Lib3MF::PBaseMaterialGroup baseMaterialGroup = model->AddBaseMaterialGroup();
  
        auto colorResourceID = baseMaterialGroup->AddMaterial(
          "color",
          // std::format("color_{:02x}{:02x}{:02x}{:02x}", sc.m_Red, sc.m_Green, sc.m_Blue, sc.m_Alpha).c_str()
          sc);
  
        mesh->SetObjectLevelProperty(baseMaterialGroup->GetResourceID(), colorResourceID);
      }
  ...
  
  #ifdef ENABLE_MANIFOLD
    } else if (const auto mani = std::dynamic_pointer_cast<const ManifoldGeometry>(geom)) {
      return append_polyset(mani->toPolySet(), wrapper, model);
      for (auto& ps : mani->toPolySets()) {
        if (!append_polyset(std::move(ps), wrapper, model)) return false;
      }
  #endif
  ```
  
  </details>

- Generally speaking, this PR supports nice colors for rendering (should allow colorful previews in the Web Playground, I'm already on it :-)), but isn't helping with multimaterial 3D printing. What we probably need for that is distinct manifolds with each their color / material (I'm exploring something similar to what described [here](https://github.com/openscad/openscad/issues/5065#issuecomment-2167812697)).

TODO before undrafting:

- Gather consensus on approach.
- Sort out CI / libassimp-dev dependency
- Add tests


gsohler:
Thank you for sharing your code with us. i was comparing your runOriginalD magic with mine and finally got also my code working. (
Definitely your CSG output is multicolor, but can also be the input nodes/. intermediate nodes be multicolor ?
( i see a setColor per a Geometry whereas in my branch i have a color for each facet)

right now i am really excited, how well it works here aftert providing the right hints :) :) 

image

ochafik:
> Thank you for sharing your code with us. i was comparing your runOriginalD magic with mine and finally got also my code working. 

Hah cool, feel free to share your code too even if not polished yet :-)

> ( Definitely your CSG output is multicolor, but can also be the input nodes/. intermediate nodes be multicolor ? ( i see a setColor per a Geometry whereas in my branch i have a color for each facet)

Yes, I'm mostly relying on the original id tracking in each ManifoldGeometry, which gets reset when applying a color to a manifold (I then turn it into its own original, see ManifoldGeometry::setColor), and also allow colors to be set on any other Geometry, since we have some scenes (w/ `--enable=lazy-union`) which only rely on PolySet, for instance. Also making sure the color of a PolySet is carried over across ops like tesselation, etc.

And the OpenGL rendering of the rendered colors is also trivial as I added a ManifoldGeometry::toPolySets method that spits out individually-colored polysets.

> right now i am really excited, how well it works here aftert providing the right hints :) :)

Exciting times indeed!

BTW, I've pushed my updated code (w/ experimental support for solid output) as a [web demo](https://ochafik.com/openscad3/#%7B%22params%22%3A%7B%22sourcePath%22%3A%22%2Fplayground.scad%22%2C%22source%22%3A%22%2F*%5Cn%20%20Hello%20there!%5Cn%5Cn%20%20If%20you're%20new%20to%20OpenSCAD%2C%20please%20learn%20the%20basics%20here%3A%5Cn%20%20https%3A%2F%2Fopenscad.org%2Fdocumentation.html%5Cn%5Cn%20%20There%20are%20lots%20of%20amazing%20libraries%20in%20the%20OpenSCAD%20ecosystem%5Cn%20%20(see%20this%20list%3A%20https%3A%2F%2Fopenscad.org%2Flibraries.html).%5Cn%5Cn%20%20Some%20of%20these%20libraries%20are%20bundled%20with%20this%20playground%5Cn%20%20(search%20for%20%5C%22demo%5C%22%20or%20%5C%22example%5C%22%20in%20the%20file%20explorer%20above)%5Cn%20%20and%20can%20be%20included%20directly%20from%20your%20models.%5Cn%5Cn%20%20Any%20bugs%20(this%20is%20an%20Alpha!)%20or%20ideas%20of%20features%3F%5Cn%20%20https%3A%2F%2Fgithub.com%2Fopenscad%2Fopenscad-playground%2Fissues%2Fnew%5Cn*%2F%5Cn%5Cn%2F%2F%20Click%20on%20Render%20or%20hit%20F6%20to%20do%20a%20fine-grained%20rendering.%5Cn%24fn%3D%24preview%20%3F%2020%20%3A%20100%3B%5Cn%5Cncolor(%5C%22red%5C%22)%5Cntranslate(%5B-24%2C0%2C0%5D)%20%7B%5Cn%20%20union()%20%7B%5Cn%20%20%20%20cube(15%2C%20center%3Dtrue)%3B%5Cn%20%20%20%20sphere(10)%3B%5Cn%20%20%7D%5Cn%7D%5Cn%5Cncolor(%5C%22green%5C%22)%5Cnintersection()%20%7B%5Cn%20%20cube(15%2C%20center%3Dtrue)%3B%5Cn%20%20sphere(10)%3B%5Cn%7D%5Cn%5Cncolor(%5C%22blue%5C%22)%5Cntranslate(%5B24%2C0%2C0%5D)%20%7B%5Cn%20%20difference()%20%7B%5Cn%20%20%20%20cube(15%2C%20center%3Dtrue)%3B%5Cn%20%20%20%20sphere(10)%3B%5Cn%20%20%7D%5Cn%7D%5Cn%5Cncolor(%5C%22black%5C%22)%5Cntranslate(%5B0%2C%20-30%2C%20-12%5D)%5Cn%20%20linear_extrude(1)%5Cn%20%20%20%20text(%5C%22OpenSCAD%20Playground%5C%22%2C%20halign%3D%5C%22center%5C%22%2C%20valign%3D%5C%22center%5C%22)%3B%5Cn%5Cn%22%2C%22features%22%3A%5B%22manifold%22%2C%22fast-csg%22%2C%22lazy-union%22%2C%22assimp%22%5D%2C%22renderFormat%22%3A%22glb%22%7D%2C%22view%22%3A%7B%22layout%22%3A%7B%22mode%22%3A%22multi%22%2C%22editor%22%3Atrue%2C%22viewer%22%3Atrue%2C%22customizer%22%3Afalse%7D%2C%22color%22%3A%22%23f9d72c%22%2C%22showAxes%22%3Atrue%2C%22showShadows%22%3Atrue%7D%7D).

gsohler:
@ochafik no problem.

https://github.com/gsohler/openscad/blob/color/src/geometry/manifold/manifold-applyops.cc#L97-L136
you can  here find a copy of your code ;)

kintel:
Thanks a lot for picking this up. I have a draft lying around somewhere too, so I'll compare my notes with your code and give it a good review. Since you're touching a lot of stuff, I'd encourage you to split off any refactoring or general improvements which doesn't directly touch the face ID stuff into separate PRs. That would make it a lot easier to iteratively drive this work forward.

src/geometry/manifold/ManifoldGeometry.cc
```diff
@@ -162,8 +227,12 @@ std::shared_ptr<Polyhedron> ManifoldGeometry::toPolyhedron() const
 template std::shared_ptr<CGAL::Polyhedron_3<CGAL_Kernel3>> ManifoldGeometry::toPolyhedron() const;
 #endif
 
-std::shared_ptr<manifold::Manifold> binOp(const manifold::Manifold& lhs, const manifold::Manifold& rhs, manifold::OpType opType) {
-  return std::make_shared<manifold::Manifold>(lhs.Boolean(rhs, opType));
+ManifoldGeometry ManifoldGeometry::binOp(const ManifoldGeometry& lhs, const ManifoldGeometry& rhs, manifold::OpType opType) const {
+  auto mani = std::make_shared<manifold::Manifold>(lhs.manifold_->Boolean(*rhs.manifold_, opType));
```
kintel, line 231: One thing to keep in mind: We should really be using `Manifold::BatchBoolean()`. I started on that refactoring but it hasn't made it yet. Shouldn't be too hard though, just loop-insert the children into the map.

kintel: It would be great if that's the case - one less thing to worry about.

ochafik:
> Thanks a lot for picking this up. I have a draft lying around somewhere too, so I'll compare my notes with your code and give it a good review. Since you're touching a lot of stuff, I'd encourage you to split off any refactoring or general improvements which doesn't directly touch the face ID stuff into separate PRs. That would make it a lot easier to iteratively drive this work forward.

@kintel I've split out the meat of it to https://github.com/openscad/openscad/pull/5185. Realized I had to retrofit per-face colors into PolySet. Will send the Assimp stuff later, it's only OFF import / export for now. Also, the solid colors logic will need more prototyping, I'm leaning towards using CGAL's tetrahedral (re)meshing or maybe test if slicers wouldn't do it for us if we set colors properly (my first attempt w/ 3MF + BambuStudio was a bit silly).

kintel:
> Realized I had to retrofit per-face colors into PolySet. 

That's also what I did in my WIP.
Out of curiosity, what made you realize you needed that? I thought the split-into-multiple-PolySet was an alternative solution.

ochafik:
> > Realized I had to retrofit per-face colors into PolySet.
> 
> That's also what I did in my WIP.

Exciting times 🤗

> Out of curiosity, what made you realize you needed that? I thought the split-into-multiple-PolySet was an alternative solution.

I got slightly spooked by a first combinatorial explosion on a complex model. There probably are glaring bugs and lots of room for optimization, but I started suspecting the node caching for the many repeated subnodes isn't the silver bullet I'd hoped. Needs more exploration in any case.

And I realized importing formats such as OFF will require handling superficial colors anyway, maybe w/ some automagical tetrahedralization. If we have to support it somewhere, might as well do it during the export rather than during the import, this way we avoid any combinatorial shenanigans and use the "native" color support from Manifold all the way.

elalish:
I'm not sure exactly what combinatoric explosion you're talking about - we can end up creating a lot of IDs, though ideally you still map them back to a small number of colors and can concatenate all the ones that match. Still, you may also benefit from https://github.com/elalish/manifold/issues/702 which will make unique IDs opt-in to simplify portions that don't need to be individually tracked.

ochafik:
> I'm not sure exactly what combinatoric explosion you're talking about - we can end up creating a lot of IDs, though ideally you still map them back to a small number of colors and can concatenate all the ones that match. Still, you may also benefit from [elalish/manifold#702](https://github.com/elalish/manifold/issues/702) which will make unique IDs opt-in to simplify portions that don't need to be individually tracked.

Oh, I haven't run into issues w/ Manifold's id tracking yet tbh, it's worked like charm so far (didn't try to push it to its limits yet 🙂).

The combinatorial issue I was mentioning was for the other approach I prototyped, to get "solid colors", not just colored faces. My vague idea was to partition each node into distinctly colored solids, and bubble this partition up the CSG tree all the way to the top ([idea here](https://github.com/openscad/openscad/issues/5065#issuecomment-2167812697), [naive implementation here](https://github.com/openscad/openscad/pull/5180/files#diff-8de1a2254aab2771fedd7868db200f00710155682453f23eb818f3da2cc03892)). I need to get back to the drawing board for that one, think i got my algebra wrong.

But alternatively, I'm now thinking of ways to turn the final superficially-colored output into something 3d-printable (either building a tetrahedral mesh and propagating colors through its cells, or letting the slicer do its magic - just need to find out how PrusaSlicer writes its [material painting](https://help.prusa3d.com/article/multi-material-painting_262620) info, for instance).

elalish:
Thanks! And great to see 3MF living up to it's intended use cases.

jordanbrown0:
>  crude modules-as-functions feature... makes any module return geometry in a struct format

Please look at PR#4478, which has what sounds like a more comprehensive treatment of this feature.

It has:
* a way to treat geometry as an opaque data item (put in variables, arrays, return from functions, et cetera)
* a way to add a geometry value to the model
* a way to render a geometry item and return a data structure describing points, faces, et cetera

See also https://github.com/openscad/openscad/wiki/OEP8:-Objects-(dictionaries%3F),-Geometry-as-data,-and-Module-References

It also includes syntax to build "objects" (structures, associative arrays, dicts) using OpenSCAD language, and module literal/reference features.

tpimh:
Just tried to export STEP file:
```
openscad: openscad/src/io/export.cc:256: void exportFile(const std::shared_ptr<const Geometry>&, std::ostream&, const ExportInfo&): Assertion `false && "Unknown file format"' failed.
```

Issue #4903, opened 2023-12-23
libhogweed.so.6 failed to reference __gmpn_cnd_swap  


Hoek67:
Trying to compile OpenSCAD on Ubuntu (latest).

Get near the last step.

following instructions on https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/Building_on_Linux/UNIX

This step has an error.
![image](https://github.com/openscad/openscad/assets/18671600/3f8cbe8e-0b4d-4cb9-885a-ae8cd109382b)

gets error...

![image](https://github.com/openscad/openscad/assets/18671600/86ce4764-c199-41a6-9ce2-8ecad6db47cc)

![image](https://github.com/openscad/openscad/assets/18671600/cdc0516a-dab1-4aee-86fa-42ba479de2e8)

I'd like to have a look at the export code and hopefully add a zip option with .obj and .mat in it.

I'm not a Linux person so tried a few things and it didn't fix it. (ie. install nettle)

Any help appreciated.

Seems GMP got broken somewhere.

kintel:
If you're not trying to do something very specific, the install guide recommends `./scripts/uni-get-dependencies.sh`, not `./scripts/uni-build-dependencies.sh`.

t-paul:
Also duplicate of #4741.

Issue #4741, opened 2023-09-12
libhogweed.so.6 failed to reference __gmpn_cnd_swap


ElektroBOXua:
During compilation libhogweed.so.6 failed to reference  «__gmpn_cnd_swap»
You can reproduce this bug by using version of libhogweed 3.9.1-2.2

I am running OpenSUSE and this is my current libhogweed version.
I tried to install version libhogweed6 3.6-1 and that fixed it for me.

t-paul:
What is that, and how is it related to OpenSCAD?

t-paul:
It's not a direct dependency as far as I can tell and it's not linked on Debian.

That error looks like a runtime error, not compilation. So that might be some local setup problem regarding library path or mismatch of libraries. Both would more point to an OpenSUSE question.

ElektroBOXua:
That is runtime error tho, but when i compile, it also requires this library for linkage:

```
furdog@localhost:~/Documents/openscad/build> make
[  1%] Automatic MOC and UIC for target OpenSCAD
[  2%] Built target OpenSCAD_autogen
[  7%] Built target mimalloc-static
[  7%] Built target graphlite
[  8%] Built target Clipper2
[  9%] Built target utilities
[  9%] Built target collider
[ 10%] Built target cross_section
[ 10%] Built target polygon
[ 15%] Built target manifold
[ 15%] Linking CXX executable openscad
Використання вбудованих специфікацій.
COLLECT_GCC=/usr/bin/c++
COLLECT_LTO_WRAPPER=/usr/lib64/gcc/x86_64-suse-linux/13/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none:amdgcn-amdhsa
OFFLOAD_TARGET_DEFAULT=1
Ціль: x86_64-suse-linux
[...]

PR #5186, opened 2024-06-20
Adjust Debian dependencies

jvasile:
Building on a freshly installed Debian testing system, I discovered that I needed gnuib to advance while building dependencies.  Without it, I get an error about missing fseterr.c

I also hit #4903 and #4741.  So I dug into uni-get-dependencies and saw that it is stalling on a missing installation candidate for "qt5-default" among other dependency issues.  This commit installs the debian packages one at a time instead of all at once.  That way one problem doesn't sink the whole apt-get command.

After making this switch, I still get the missing qt5-default error but `check-dependencies.sh` shows a solid list of OK and the build runs just fine.

### DIFF
--- a/scripts/uni-get-dependencies.sh
+++ b/scripts/uni-get-dependencies.sh
@@ -100,14 +100,13 @@ get_mageia_deps()
 get_debian_deps()
 {
  apt-get update
- apt-get -y install \
-  build-essential curl ninja-build libffi-dev \
+ for i in build-essential curl ninja-build libffi-dev \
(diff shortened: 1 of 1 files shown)


jvasile:
```
$ uname -a
Linux silicon 6.8.12-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.8.12-1 (2024-05-31) x86_64 GNU/Linux
```

I didn't capture what depended on gnulib.  I suppose I could remove it and see what fails if you like.  But you're right that once I installed the dependencies one at a time, I went right to compiling and no longer needed to build.

PR #5195, opened 2024-06-27
Don't run parseToplevelDocument twice

gsohler:
When OpenSCAD is called with an Design File as argument, MainWindow::parseToplevelDocument is called twice Its also called from MainWindow::compile
Even customer still works with the code commented.

I dont see why this is needed.

### DIFF
--- a/src/gui/TabManager.cc
+++ b/src/gui/TabManager.cc
@@ -495,6 +495,7 @@ void TabManager::openTabFile(const QString& filename)
 
   if (opened) { // only try to parse if the file opened
     par->hideCurrentOutput(); // Initial parse for customizer, hide any errors to avoid duplication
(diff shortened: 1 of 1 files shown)


kintel:
Not sure if this is the correct fix. It looks like some interaction with automatic reload and preview, perhaps we forgot to keep track of whether the file was already compiled?

gsohler:
Quite some time back I have disabled parseTopLevelGeometry in TabManager.cc due to duplicate parseTopLEvelDocument in my
fork and I dont miss anything - not on startup, not by pressing "Save"and not by pressing F5/F6, it just works to my best expectation.
It seems, that automaticReload path does always once reliablly do the job.

kintel:
Ok, I'll go through all the possible states and trace the logic.

kintel:
There was some work done in this domain earlier this year. Is OpenSCAD still exhibiting the double parse behavior?

PR #5209, opened 2024-07-08
Measurement improvement

gsohler:
* Display Coordinates of focused lines and segments in status bar
*  Display "Ruler"  together with the measured result to see, what was measured.
* Cleanup measurement code


src/gui/MainWindow.cc
```diff
@@ -2345,7 +2345,7 @@ void MainWindow::actionMeasureAngle()
 void MainWindow::leftClick(QPoint mouse) 
 {
   QString str = meas.statemachine(mouse);
-  if(str.size() > 0) {
+  if(str != nullptr) {
```
kintel, line 2348: Do you mean `!str.isEmpty()` ?

src/gui/Measurement.cc
```diff
@@ -50,11 +50,13 @@ void Measurement::startMeasureAngle(void)
 }
 QString Measurement::statemachine(QPoint mouse) 
 {
-  if(qglview->measure_state == MEASURE_IDLE) return "";
+  if(qglview->measure_state == MEASURE_IDLE) nullptr;
```
kintel, line 53: If you want to return an empty string, return `{}` or `QString()`

src/gui/Measurement.cc
```diff
@@ -50,11 +50,13 @@ void Measurement::startMeasureAngle(void)
...
   qglview->selectPoint(mouse.x(),mouse.y());
   double ang=NAN;
   double dist=NAN;
   SelectedObject obj1, obj2, obj3;
+  Vector3d p1, p2, p3,p4, side1, side2;
+  SelectedObject ruler;
```
kintel, line 59: It might be safer to initialize it to invalid here, in case a case becomes unhandled below:
```
SelectedObject ruler = {
  .type = SelectionType::SELECTION_INVALID
};
```

src/gui/Measurement.cc
```diff
@@ -128,17 +138,36 @@ QString Measurement::statemachine(QPoint mouse)
...
         if(obj1.type == SelectionType::SELECTION_POINT && obj2.type == SelectionType::SELECTION_POINT && obj3.type == SelectionType::SELECTION_POINT)
         {
-          Vector3d side1=(obj2.p1-obj1.p1).normalized();
-          Vector3d side2=(obj2.p1-obj3.p1).normalized();
-          ang=acos(side1.dot(side2))*180.0/3.14159265359;
+	  p1=obj1.p1;
+          p2=obj2.p1;
+	  p3=obj2.p1;
+	  p4=obj3.p1;
+	  goto display_angle;
```
kintel, line 145: This makes code pretty hard to read - try to avoid goto and split out the code into a function or lambda instead. In general, large switch-case statements are hard to follow.

src/gui/Measurement.cc
```diff
@@ -50,11 +50,43 @@ void Measurement::startMeasureAngle(void)
...
+        double ang=acos(side1.dot(side2))*180.0/G_PI;
+        if(!std::isnan(ang))
+        {
+          return QString("Angle  is %1 Degrees").arg(ang);
+        }
+        qglview->selected_obj.clear();
+        qglview->shown_obj = nullptr;
+        qglview->update();
+        qglview->measure_state = MEASURE_IDLE;
+  };
```
kintel, line 88: You should probably return an empty string here.

src/gui/Measurement.cc
```diff
@@ -50,11 +50,43 @@ void Measurement::startMeasureAngle(void)
...
-  double ang=NAN;
   double dist=NAN;
   SelectedObject obj1, obj2, obj3;
+  SelectedObject ruler = {
+	  .type=SelectionType::SELECTION_INVALID
+  };
+  auto display_angle = [this] (Vector3d p1, Vector3d p2, Vector3d p3, Vector3d p4) 
+  {	
+	SelectedObject ruler;
+	Vector3d side1, side2;
```
kintel, line 63: These could be declared where they're assigned. That way they can also be const.

src/gui/Measurement.cc
```diff
@@ -50,11 +50,43 @@ void Measurement::startMeasureAngle(void)
...
+  	ruler = {
+            .type = SelectionType::SELECTION_SEGMENT,
+            .p1 = p1,
+            .p2 = p2
+        };
+	this->qglview->selected_obj.push_back(ruler);
```
kintel, line 69: Tip: As you don't need access to this variable later, it's possible to do:
```
this->qglview->selected_obj.push_back({
  .type = SelectionType::SELECTION_SEGMENT,
  .p1 = p1,
  .p2 = p2
});
```

src/gui/Measurement.cc
```diff
@@ -50,11 +50,43 @@ void Measurement::startMeasureAngle(void)
...
+        side2=(p4-p3).normalized();
+        double ang=acos(side1.dot(side2))*180.0/G_PI;
+        if(!std::isnan(ang))
+        {
+          return QString("Angle  is %1 Degrees").arg(ang);
+        }
+        qglview->selected_obj.clear();
+        qglview->shown_obj = nullptr;
+        qglview->update();
+        qglview->measure_state = MEASURE_IDLE;
```
kintel, line 87: This is just a nit, but it's sometimes more readable to handle the error inside the if statement and let the natural return path of the function be the good value.

Issue #5217, opened 2024-07-14
color() doesn't override color of render()'ed objects in Manifold mode


kintel:
The following design should render a red cube, but renders a blue cube in Manifold mode:
```
color("red") render() color("blue") cube();
```

Issue #5218, opened 2024-07-15
Color rendering: Subtracted minkowski isn't using the subtraction color in Manifold mode


kintel:
The following design subtracts a minkowski object from a simple positive object. The cutout surface should be rendered using the subtractions color (defaults to green), but is instead rendered with the default positive color (yellow):
```
difference() {
  cube(10, center=true);
  translate([5,5,5]) minkowski(convexity=2) {
    cube(8, center=true);
    cube(1, center=true);
  }
}
```

pca006132:
Probably missing an AsOriginal call after minkowski.

kintel:
This was already fixed by #5234 - forgot to close it.

@pca006132 Did you observe similar wrong behavior related to your latest commit in #5282?

pca006132:
yes but it is now fixed as well

kintel:
Reopening this: This is still broken in throwntogether mode, but we didn't include a test for that. Will create a test shortly.

kintel:
How to reproduce:
* Re-enable disabled test here: https://github.com/openscad/openscad/blob/7f35682b4bf278e3d33e3bb8293b5907ec356a29/tests/CMakeLists.txt#L1628
* `ctest -R throwntogether-manifold_issue5218`

ali1234:
Here is a much simpler reproducer for this bug:

```
difference() {
    cube(15, center=true);
    render() difference() {
        cube(20, center=true);
        cube(10, center=true);
    }
}
```

If you remove the `render` call then you get a green cube in thrown together. But with render it is yellow.

~~It happens because of this code in ThrownTogetherRenderer:~~ it isn't that...

If an object has explicit colour data (ie color_indices) then that will be used. Otherwise, ThrownTogetherRenderer chooses the appropriate colour based on other properties.

Since PolySetRenderer previously did not support colour for Nefs, they would have no colour data when they hit the preview and would be coloured automatically.

Now it does support colour, those colours override anything set in the renderer.

So effectively `render()` ends up behaving the same as setting an explicit colour, eg `color("red")`

With Manifold backend, the input to `render()` is not a Nef, so rendering it always resulted in colours being assigned.

ali1234:
Interestingly the reproducer totally fails in normal preview. Nothing is drawn at all.

ali1234:
To explain the problem in a different way, this is what ThrownTogetherRenderer does:

```
For each chunk of geometry:
    If it has explicit colour data:
        it is drawn in that colour
[...]

PR #5234, opened 2024-07-21
Fix CSG rendering of colors / modifiers when there's underlying colors

ochafik:
Fixes https://github.com/openscad/openscad/issues/5217 & https://github.com/openscad/openscad/issues/5218

cc/ @kintel


kintel:
This fixes both #5217 and #5218 right?

ochafik:
> This fixes both #5217 and #5218 right?

Oh, it does indeed!

tests/data/scad/issues/issue5218.scad
kintel, line 1: I recommend moving this file to tests/data/scad/3D/issues/ - that will automatically run this test in CGAL and Manifold mode.
..then add the file to `SCADFILES_DIFFERENT_MANIFOLD_RENDER_EXPECTATIONS` and generate expectations for everything.
This will add (rather useless) expectations for CGAL mode, but it makes the config a bit cleaner IMO.

src/glview/preview/OpenCSGRenderer.cc
```diff
@@ -261,12 +265,16 @@ void OpenCSGRenderer::createCSGVBOProducts(
...
+        bool override_color;
         if (highlight_mode) {
           colormode = ColorMode::HIGHLIGHT;
+          override_color = true;
         } else if (background_mode) {
           colormode = ColorMode::BACKGROUND;
+          override_color = true;
         } else {
           colormode = ColorMode::CUTOUT;
+          override_color = true;//c.isValid();
```
kintel, line 277: What is the purpose of leaving commented out code here?

ochafik: Removed

tests/data/scad/issues/issue5218.scad
ochafik: Done

kintel (approved):
LGTM.

FYI: Re. the issue1137 images: I think we, in general, should try to avoid checking in empty images or images representing an unwanted result. We did this in the past in an attempt to track crash bugs (empty image == at least we didn't crash).
-> let's clean that up separately.

PR #5235, opened 2024-07-22
Add backend setting / flag to enable manifold in prod builds

ochafik:
This moves Manifold (and its color support) out of experimental (https://github.com/openscad/openscad/issues/4825)

image

image

Command line usage:

```bash
echo '$fn=1000; cube(); color("red") sphere();" | \
  openscad - -o out.off --backend=manifold
```

cc/ @kintel


src/openscad.cc
```diff
@@ -1071,12 +1072,17 @@ int main(int argc, char **argv)
...
+    RenderSettings::inst()->manifoldEnabled = false;
+  }
 
   if (vm.count("preview")) {
     if (vm["preview"].as<string>() == "throwntogether") viewOptions.renderer = RenderType::THROWNTOGETHER;
   } else if (vm.count("render")) {
     // Note: "cgal" is here for backwards compatibility, can probably be removed soon
-    if (vm["render"].as<string>() == "cgal" || vm["render"].as<string>() == "force") {
+    if (vm["render"].as<string>() == "cgal" ||
+        vm["render"].as<string>() == "manifold" ||
```
kintel, line 1084: `--render=cgal` is just for backwards compatibility. `--render=force` is meant to convert rendered geometries into the format of the current geometry backend. Adding `"manifold"` as an option doesn't really make sense.

src/openscad.cc
```diff
@@ -978,6 +978,7 @@ int main(int argc, char **argv)
     ("camera", po::value<string>(), "camera parameters when exporting png: =translate_x,y,z,rot_x,y,z,dist or =eye_x,y,z,center_x,y,z")
     ("autocenter", "adjust camera to look at object's center")
     ("viewall", "adjust camera to fit object")
+    ("disable-manifold", "render everything with CGAL (disables Manifold) for legacy, slower behaviour")
```
kintel, line 981: Not sure we're ready to switch the default yet. It would be nice with an option where we could select backend, then switch the default value for that option down the road.

src/openscad.cc
ochafik: Reverted this + added `--backend=(manifold|cgal)`

src/openscad.cc
ochafik: Got it, switched it back to cgal + turned the UI to a combo (see screenshot in PR desc)

kintel:
@ochafik I'm back and ready to move this forward. What's missing from your end? Didn't look into the test failures yet..

ochafik:
> @ochafik I'm back and ready to move this forward. What's missing from your end? Didn't look into the test failures yet..

@kintel Welcome back! Giving tests a fresh run, I think it's mostly about making them green, bar any code changes you'd like.

kintel:
OK, sg, I'll give it a spin + review

kintel:
Got sidetracked by a manifold upgrade. Back on this now; did some merging as a start.

kintel:
I think all should be good now. Will merge if I don't hear anything :)

PR #5247, opened 2024-07-27
Improve readability of axis numbering; use Hershey font.

hzeller:
The axes labels were always hard to read due to their simple '7 segment' shape that easily can be mis-read from different perspectives (e.g is something a 2 or an upside 5 ?)

This change uses the Hershey Simplex font commonly used in such applications.

 * Use the Hershey Simplex font to draw the numbers on the axes, replacing the ad-hoc 7-segment font. The Simplex font just uses simple line-segments so is fast to draw and crisp to read (it is commonly used in CAD and plotting applications to emit labels for that reason).
 * Provide a text drawing function that with an interface that emits calls with drawing instructions, abstracting away the output device (the text drawing function itself does not need to know about glVertex3D()).
 * While at it, include not only numbers but all characters from the Hershey Simplex font, which allows to use this to draw other ASCII-labels as well.
 * Use that in the GLView::decodeMarkerValue() to draw the number axis annotations, _vastly_ improving readability of these labels.

While at it, also improve placement of the text

  * The numbers are now horizontally centered around the corresponding tick on the axis (including the negative axis).
  * Make all numbers readable in the default projection, looking forward on the coordinate system (Y and Z used to be weirdly 'backwards' before).

Here a before/after

before-font

after-font


hzeller:
Also numbers with additional characters such as exponent `1e+06` or with decimal point `0.16` look much better with this change

|     Before  |  After |
|---------------|-----------|
|font-1e6-old | font-1e6-new
|font-012-old |font-0 12-new

t-paul (approved):
Looks great. With the ongoing refactoring of the OpenGL code, I hope @kintel will find some time to have a look for merging.

I just have 2 questions

1) Regarding licensing, as Linux distributions usually require a clear specification of what file is covered under which license: Do you see both new files fully covered by the Hershey License?

2) The comment says "Roman Simplex font", which one is that? `rowmans.jhf`?

hzeller:
The code `hershey.{cc,h}` is covered by GPL, written by me (in fact I wrote it a couple of years ago in my [BeagleG project](https://github.com/hzeller/beagleg/blob/main/src/hershey.cc) - so here it is only with minor formatting changes)

The font is  just the data in the array and the comment above the array satisfies the license requirements to attribute Hershey.

Which of the Hershey Filename: don't know, I converted that a long time ago and I don't know if  there was a filename associated with it. Roman Simplex is sufficient to characterize it (there are also Japanese as opposed to Roman and there are Duplex and Gothic which use more strokes.) I might have gotten the original data from https://paulbourke.net/dataformats/hershey/ which I then converted into the current form.

kintel:
I'll review this.
Eventually we want to phase out immediate mode OpenGL (glBegin, glEnd, glVertex3d etc.), but this patch doesn't make it any worse, so we can probably merge it after review.

kintel:
Somewhat surprisingly, tests didn't fail, but I guess our "1-pixel-erode" to account for GPU differences kind of erases any 1-pixel fonts as well ;)

kintel (approved):

PR #5248, opened 2024-08-04
Minor fixes

mmuman:


doc/openscad.1
```diff
@@ -63,10 +63,7 @@ strings, care has to be taken that the shell does not consume quotation marks.
 More than one \fB-D\fP option can be given.
 .TP
 .B \-p
-Customizer parameter file.
-.TP
-.B \-p
```
t-paul, line 68: Oh, the typo is that this one should be an upper case `-P`. It's `-p param.json -P ParameterSetNameToUse`

```
  -p [ --p ] arg               customizer parameter file
  -P [ --P ] arg               customizer parameter set
```

t-paul (changes requested):

doc/openscad.1
mmuman: Oh, OK :-D

t-paul (approved):
:+1:

PR #5250, opened 2024-08-04
Fix OFF import of clors in float format

mmuman:
like unitcube.off

### DIFF
--- a/src/io/import_off.cc
+++ b/src/io/import_off.cc
@@ -3,7 +3,10 @@
 #include "PolySet.h"
 #include "printutils.h"
 #include "AST.h"
+#include <charconv>
 #include <fstream>
+#include <sstream>
+#include <locale>
 #include <boost/regex.hpp>
 #include <boost/lexical_cast.hpp>
 #include <boost/algorithm/string.hpp>
@@ -52,6 +55,34 @@ std::unique_ptr<PolySet> import_off(const std::string& filename, const Location&
 
     return true;
   };
+
+  auto getcolor = [&](const auto& word){
+    int c;
+    if (boost::contains(word, ".")) {
+      float f;
+#ifdef __cpp_lib_to_chars
+      auto result = std::from_chars(word.data(), word.data() + word.length(), f);
+      if (result.ec != std::errc{}) {
+        AsciiError("Parse error");
+        return 0;
+      }
+#else
+      // fall back for pre C++17
+      std::istringstream istr(word);
+      istr.imbue(std::locale("C"));
+      istr >> f;
+      if (istr.peek() != EOF) {
+        AsciiError("Parse error");
+        return 0;
+      }
+#endif
+      c = (int)(f * 255);
+    } else {
+      c = boost::lexical_cast<int>(word);
+    }
+    return c;
+  };
+
 
   if (!f.good()) {
     AsciiError("File error");
@@ -226,10 +257,10 @@ std::unique_ptr<PolySet> import_off(const std::string& filename, const Location&
       if (words.size() >= face_size + 4) {
         i = face_size + 1;
         // handle optional color info (r g b [a])
-        int r=boost::lexical_cast<int>(words[i++]);
-        int g=boost::lexical_cast<int>(words[i++]);
-        int b=boost::lexical_cast<int>(words[i++]);
-        int a=i < words.size() ? boost::lexical_cast<int>(words[i++]) : 255;
+        int r=getcolor(words[i++]);
+        int g=getcolor(words[i++]);
+        int b=getcolor(words[i++]);
+        int a=i < words.size() ? getcolor(words[i++]) : 255;
         Color4f color(r, g, b, a);
         
         auto iter_pair = color_indices.insert_or_assign(color, ps->colors.size());


src/io/import_off.cc
```diff
@@ -53,6 +53,17 @@ std::unique_ptr<PolySet> import_off(const std::string& filename, const Location&
     return true;
   };
 
+  auto getcolor = [&](const auto& word){
+    int c;
+    if (boost::contains(word, ".")) {
+      c = (int)(boost::lexical_cast<float>(word) * 255);
```
t-paul, line 59: The FAQ of [boost::lexical_cast()](https://www.boost.org/doc/libs/1_85_0/doc/html/boost_lexical_cast/frequently_asked_questions.html) mentions:
```
boost::lexical_cast has the behavior of std::stringstream, which uses num_get functions of std::locale to convert numbers.
```
which sounds like this would break with locales that do not use "." as a decimal separator?

t-paul (changes requested):

src/io/import_off.cc
t-paul: Looks like there's no way to disable that behavior and it is locale dependent:
```
#include <iostream>
#include <boost/lexical_cast.hpp>

int main(int argc, char** argv) {
    std::setlocale(LC_ALL, "");
    std::locale::global(std::locale(""));

    std::cout << "LC_ALL: " << std::setlocale(LC_ALL, NULL) << std::endl;
    std::cout << "LC_CTYPE: " << std::setlocale(LC_CTYPE, NULL) << std::endl;
    std::cout << "C++ locale: " << std::locale().name() << std::endl;

    float f = boost::lexical_cast<float>(argv[1]);
    printf("%f\n", f);
    return 0;
}
```
$ g++ -o x x.cc && LC_ALL=en_US.utf8 ./x 1.2
LC_ALL: en_US.utf8
LC_CTYPE: en_US.utf8
C++ locale: en_US.utf8
1.200000
```
```
g++ -o x x.cc && LC_ALL=de_DE.utf8 ./x 1.2
LC_ALL: de_DE.utf8
LC_CTYPE: de_DE.utf8
C++ locale: de_DE.utf8
terminate called after throwing an instance of 'boost::wrapexcept<boost::bad_lexical_cast>'
  what():  bad lexical cast: source type value could not be interpreted as target
Aborted
```
```
g++ -o x x.cc && LC_ALL=de_DE.utf8 ./x 1,2
LC_ALL: de_DE.utf8
LC_CTYPE: de_DE.utf8
C++ locale: de_DE.utf8
1,200000
```

It would be nice if we could stop messing with the global locale as some of the other imports.

Comment (hzeller) [reply src/io/import_off.cc:59]: the `std::from_chars()`, available since c++17, parses numbers locale-independently, so sounds like a good choice here.

```c++
#include <iostream>
#include <charconv>

int main(int argc, char** argv) {
    std::setlocale(LC_ALL, "");
    std::locale::global(std::locale(""));

    std::cout << "LC_ALL: " << std::setlocale(LC_ALL, NULL) << std::endl;
    std::cout << "LC_CTYPE: " << std::setlocale(LC_CTYPE, NULL) << std::endl;
    std::cout << "C++ locale: " << std::locale().name() << std::endl;

    std::string_view num(argv[1]);
    float f;
    auto result = std::from_chars(num.data(), num.data() + num.length(), f);
    if (result.ec == std::errc{}) {
      printf("%f (left over '%s')\n", f, result.ptr);
    } else {
      printf("parse error\n");
    }
    return 0;
}
```

Parsing with `en_US` locale
```
g++ --std=c++17 x.cc && LC_ALL=en_US.utf8 ./a.out 1.2
LC_ALL: en_US.utf8
LC_CTYPE: en_US.utf8
C++ locale: en_US.utf8
1.200000 (left over '')
```

Parsing with locale that usually uses comma for decimal point (`de_DE`): Parses 1.2 correctly (and then here: prints in German locale)

```
g++ --std=c++17 x.cc && LC_ALL=de_DE.utf8 ./a.out 1.2
LC_ALL: de_DE.utf8
LC_CTYPE: de_DE.utf8
C++ locale: de_DE.utf8
1,200000 (left over '')
```

Parsing 1,2 would not work: it will parse the `1` and considers the `,2` as left-over. But that is exactly how we would like to see the parsing work:

```
g++ --std=c++17 x.cc && LC_ALL=de_DE.utf8 ./a.out 1,2
LC_ALL: de_DE.utf8
LC_CTYPE: de_DE.utf8
C++ locale: de_DE.utf8
1,000000 (left over ',2')
```

... and correctly complains about things not parseable at all:

```
g++ --std=c++17 x.cc && LC_ALL=en_US.utf8 ./a.out foobar
LC_ALL: en_US.utf8
LC_CTYPE: en_US.utf8
C++ locale: en_US.utf8
parse error
```

Also advantage: no exceptions.

mmuman:
Seems to work with `std::from_chars` :-)

t-paul (approved):
Ah, `std::from_chars` sounds good and C++17 is fine.

t-paul:
Hmm, well, I *thought* we are good with C++17. Going to check the build errors...

hzeller:
mmh, I remember that earlier releases of gcc just after c++17 came out didn't have yet `from_chars()` for float yet and only worked for g++ version 11 and above (I remember I once had to implement an annoying workaround for that reason).
That it even affects ubuntu 20 is somewhat disappointing...

mmuman:
`strtof` maybe?

t-paul:
Unfortunately `strtof` is also locale aware.

I suppose the best option would be a "toFloat" in `export.cc` (so it can be used later in other imports too). For now this would either use `stringstream` or we find some `#define` that would switch to allow `from_chars()` for the platforms it available.

This is ugly, but should be both thread safe and locale independent. I'm not 100% sure about error handling. This works, but there might be a better way.

https://godbolt.org/z/d7xEbMq1M
```
#include <sstream>
#include <locale>

int main(void) {
    float f;
    std::istringstream istr("1.3e-3");
    istr.imbue(std::locale("C"));
    istr >> f;
    if (istr.peek() == EOF) {
        printf("%f\n", f);
    } else {
        printf("error");
        f = 0.0f;
    }
}
```

mmuman:
WTF is there no single way to do this simply…

t-paul:
We do have a dependency on https://github.com/google/double-conversion/ but I'm not sure I would call that a simple solution either.

t-paul (approved):
Supporting even just 1 previous (LTS) distro can be a pain. Thanks for not giving up on that.

PR #5262, opened 2024-08-18
manifold.h is a library outside openscad, use canonical include path.

hzeller:
This means that
  * Use the 'system' include angle brackets not the double-quote include used for local projects.
  * Use the canonical location under manifold/manifold.h (which is where manifold installs its headers. By accident, cmake search also allowed to leave out the prefix, which is of course not desirable).

### DIFF
--- a/src/RenderStatistic.cc
+++ b/src/RenderStatistic.cc
@@ -38,8 +38,8 @@
 #endif // ENABLE_CGAL
 
 #ifdef ENABLE_MANIFOLD
+#include <manifold/manifold.h>
 #include "ManifoldGeometry.h"
-#include "manifold.h"
 #include "manifoldutils.h"
 #endif // ENABLE_MANIFOLD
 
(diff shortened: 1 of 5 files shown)


hzeller:
mmh the CI indicates that there are some more clean-ups needed. It built locally, so will have to dig a bit further.

hzeller:
Ok, looks like this only works after CMake is modified to use the cmake locations provided by the submodule and I think that makes only sense after manifold is updated to the latest. So closing for now, to revisit later.

hzeller:
@pca006132 FYI - related to my comment about the include paths and cmake deps in the manifold PR.

 I'll re-visit this once OpenSCAD dependency is updated to a more recent manifold; then we can also try an `find_package(manifold REQUIRED COMPONENTS Core)` before having to resort to `add_subdirectory(submodules/manifold,...`

Issue #5265, opened 2024-08-22
Build script does not support Linux Mint Virginia


atossell91:
**Description of the error**
Attempting to build OpenSCAD on Linux Mint Virginia gives an error, stating that the OS is not recognized.

**To Reproduce**
Attempt to build OpenSCAD with a clean, Linux Mint Virginia install

**Expected behavior**
The build script should run using the same, or similar, settings to the previous versions of Linux Mint

**Environment and Version info (please complete the following information):**
 - OS: Linux Mint 21.3 Virginia

**Additional context**
I added this

    elif [ "`grep -i linux.mint.[2][1] /etc/issue`" ]; then
      get_ubuntu_16_deps

To the uni-get-dependencies.sh file and it seemed to work. I'm not sure if there's a better way though.

t-paul:
Closed via #5193.

PR #5193, opened 2024-06-24
Simplify Debian dependencies

cjmayo:
Linux Mint 18 & 19 are EOL

git-core is a virtual package provided by git
libboost-dev installed by libboost-all-dev
libcgal-qt5-dev related bug was resolved in libcgal-dev-4.8-1
libfontconfig1-dev is installed by a virtual package in Ubuntu 20.04, subsequently a dummy package
libgmp3-dev is a dummy package
libqt5scintilla2-dev is a virtual package provided by libqscintilla2-qt5-dev

PR #5267, opened 2024-08-22
Added Linux Virginia support for dependencies

atossell91:
Added a new case in the get-dependencies scripts to account for the new Linux Mint (Virginia).

It calls the same function as the previous Linux Mint case, but I figured a new case make sense for a new version of the OS.

#5265


t-paul (approved):
Thanks!

t-paul:
I just saw that #5193 was still pending which does some more cleanup and also updates the Mint rule. It would be great if you could check if that also works for you on recent Mint.

atossell91:
@t-paul I uninstalled all of the dependencies, and then reinstalled them with the script and built openscad. Everything went well, so #5193 should be good to go!

t-paul:
Awesome, thanks! Sorry for the overlapping effort.

PR #5276, opened 2024-08-29
Increase minimum Qt version to 5.12.0

cjmayo:
https://bugreports.qt.io/browse/QTBUG-65592
was duplicated by:
https://bugreports.qt.io/browse/QTBUG-67069
and then by:
https://bugreports.qt.io/browse/QTBUG-70571
which was fixed by:
https://code.qt.io/cgit/qt/qtbase.git/commit/?id=e2d79b496335e8d8666014e900930c66cf722eb6

Reverts:
36bc309a7 ("Add workaround for jumping dock widgets.", 2018-12-08)

---

I can't recreate the problem on Linux with Qt 5 or 6 with this patch.
On QTBUG-65592 ? and macOS users reported the bug fixed.
There was a late comment from a Windows user however, but that was 5 years ago and I can't find any subsequent bugs plus there have been 3 more Qt 5 releases since.

Ubuntu 20.04 has Qt 5.12.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -591,7 +591,7 @@ if(NOT HEADLESS)
       message(STATUS "DBus input driver disabled per user request.")
     endif()
   else()
-    find_package(Qt5 5.4 COMPONENTS Core Widgets Multimedia OpenGL Concurrent Network Svg REQUIRED QUIET)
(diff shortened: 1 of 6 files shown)


kintel:
Thx!

Issue #5281, opened 2024-09-04
Slow 2D Offset


hackwin:
**Is your feature request related to a problem? Please describe.**
The renderer is too slow when extruding SVG shapes into 3d.  The "openscad-2020.10.11.ci6009" takes about 10 seconds while the 
"OpenSCAD-2024.08.30-x86-64" version takes about 2 minutes.

**Describe the solution you'd like**
Fast SVG extrusion.  Seems like this should be 1 second or less.  Maybe some command line argument?

> Newer version of clipper2 works fine and is really fast for this one, it took less than 0.2s in total. Not entirely sure why the older version of clipper2 and clipper1 are slow...
-Chum Kit LAM (from discuss@lists.openscad.org)

**Describe alternatives you've considered**
PyMesh, Inkscape, writing a script that outputs STL triangles from simple XY SVG points

**Additional context**

![plain](https://github.com/user-attachments/assets/8c063264-b290-47f5-91cb-1ed3b351cdfa)
![render](https://github.com/user-attachments/assets/96fc792e-fa0a-4356-822c-11d52dc2c36a)

```
$fn=50;

color([45/255,252/255,132/255]){
    linear_extrude(1){
        resize([203.2+0.1,203.2+0.1,1]){
            offset(-100)offset(100){
                import("plain.svg",center=true);
            }
        }
    }
}
```

```
$fn=50;

color([245/255,203/255,15/255]){
    resize([203.2,203.2,3]){
        linear_extrude(10){
            import("plain.svg",center=true);
        }
    }
}
```

Sample files: https://mylabyrinthlab.com/data/1y4a2u/
Used here: https://mylabyrinthlab.com/

UBaer21:
the offset taking unexpected much time - use fill() instead

kintel:
Unrelated to the slowness, but the base could probably be constructed using `hull()` rather than two offsets.

kintel:
Interestingly, `openscad base.scad -o out.stl` causes an assertion failure for me:
```
Assertion failed: (s0 != s1 && s0 != s2 && s1 != s2), function append_stl, file export_stl.cc, line 161.
```

kintel:
The assert is likely caused by #4867 - we can deal with that separately.

kintel:
Some light-weight archeology:
* Last fast version I can find: 2021.11.15
* First slow version I can find: 2023.01.29

Let me know if anyone can reduce that span. I can try bisecting the source, but building old software isn't always fun..

[...]

Issue #5211, opened 2024-07-11
SIGSEGV with Minkovski operation using `manifold` (Snapshot version)


pludi:
**Describe the bug**
Rendering the preview of the attached file results in a `SIGSEGV` on macOS (2024.07.08) and Linux (2024.07.09).

**To Reproduce**
1. Enter the below code
1. Let a preview render
1. Enable `manifold` feature & restart
1. Enter or reload the code

**Expected behavior**
Preview rendered.

**Code reproducing the issue**
```openscad
difference() {
  minkowski() {
    union() {
      translate([0, .6, 0])
        cube([29.75, .5, 99.5], center = true);
      translate([-12.5, -.6, 0])
        cube([4.75, .5, 99.5], center = true);
      translate([-14.5, 0, 0])
        cube([.875, 1.2, 99.5], center = true);
    }
    sphere(.25);
  }
  translate([-15.375, 0, 0])
    cube([1, 2.2, 100], center = true);
}
```

**Environment and Version info (please complete the following information):**
- macOS:
   - OS: macOS Sonoma 14.5
   - System: MacBook Pro M3
   - OpenSCAD Version 2024.07.08, downloaded from openscad.org
- Linux:
    - OS: EndeavourOS / Arch Linux
    - System: Intel Core i7 system
    - OpenSCAD Version 2024.07.09, compiled locally from AUR

**Additional context**
Keeping the 3 `union` cubes non-overlapping by changing the last cube to
```openscad
cube([.875, .2, 99.5], center = true);
```
does not lead to this issue.

pca006132:
maybe you can compile with debug information (RelWithDebInfo) and attach the stack trace? it will be very helpful.

pludi:
Full output of the debug build:

```
> ./openscad
Could not initialize localization.
Error reading examples.json: examples.json: cannot open file
ERROR: Cannot open shader source file: 'MouseSelector.vert'
ERROR: Cannot open shader source file: 'MouseSelector.frag'
/usr/src/debug/openscad-git/openscad/src/gui/MouseSelector.cc: OpenGL vertex shader Error:
0:1(1): error: syntax error, unexpected end of file


/usr/src/debug/openscad-git/openscad/src/gui/MouseSelector.cc: OpenGL fragment shader Error:
0:1(1): error: syntax error, unexpected end of file


/usr/src/debug/openscad-git/openscad/src/gui/MouseSelector.cc: OpenGL Program Linker Error:
[...]

Issue #5069, opened 2024-03-25
Nightly Build Crashes During Render of Complicated Object


LenStruttmann:
OpenSCAD nightly build OpenSCAD-2024.03.23-x86-64 with Manifold consistently crashes while attempting to render (F6) the attached complicated model after about 15 minutes.

**To Reproduce**
Attempt to render (F6) the attached  model.

**Expected behavior**
Successful render or a graceful failure.

**Code reproducing the issue**

```
r1 = 10;
r2 = 5;
halfSide = 1;

faces = [
  [ 0, 4, 1],
  [ 3, 4, 0],
  [ 2, 4, 3],
  [ 1, 4, 2],
  [ 0, 1, 2, 3],
];


union()
{
  sphere(r1);
  for (phi=[0:1:180], theta=[0:1:360])
  {
    points = [ 
      [-halfSide, -halfSide, 0],
      [ halfSide, -halfSide, 0],
      [ halfSide,  halfSide, 0],
      [-halfSide,  halfSide, 0],
      [ 0, 0, r1 + rands(0,10,1)[0] ],
    ];

    rotate( [0, phi, theta]) 
    polyhedron(points = points, faces = faces);
  }
}
```

**Environment and Version info (please complete the following information):**
OS Name	Microsoft Windows 11 Home	
Version	10.0.22621 Build 22621	
System Manufacturer	Acer	
System Model	Predator PH315-54	
System Type	x64-based PC	
Processor	11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz, 2304 Mhz, 8 Core(s), 16 Logical Processor(s)	
BIOS Version/Date	Insyde Corp. V1.12, 6/7/2022	
Hardware Abstraction Layer	Version = "10.0.22621.2506"	
Installed Physical Memory (RAM)	16.0 GB	

**Library & Graphics card information**
OpenSCAD Version: 2024.03.23 (git 0ac066a01)
System information: Microsoft Windows 11 (10.0.22621) x86_64 16 CPUs 15.77 GB RAM
User Agent: OpenSCAD/2024.03.23 (git 0ac066a01) (Microsoft Windows 11 (10.0.22621) x86_64)
Compiler: GCC "13.2.0" 64bit
MinGW build: MingW64
Debug build: No
Boost version: 1_81
Eigen version: 3.4.0
CGAL version, kernels: 5.5, Cartesian<Gmpq>, Extended_cartesian<Gmpq>, Epeck
[...]

Issue #5032, opened 2024-03-05
Manifold surface cover when using small offsets


UBaer21:
Please first consider that general questions about using OpenSCAD or how to write code for specific models are better asked on the mailing list. See https://openscad.org/community.html for more details.

**Environment**
v2024.02.22 on  Windows

**Issue**
offset(-.0001) produces a face covering the surface of the difference when rendering

![image](https://github.com/openscad/openscad/assets/93376232/dedf3bc9-0772-4a1e-8807-614d0a9504df)

code:
```
size=60; // only on larger sizes
matrixString=[for(i=[0:size^2]) round(rands(0,1,1,i)[0]) ];
module qrCodePattern(matrixString=matrixString, dimension=size) {
    translate([1,1,-1])linear_extrude(2,convexity=5)offset(-.0001)union(){
        for (y = [0 : dimension-1]) {
            for (x = [0 : dimension-1]) {
                if (str(matrixString[y * dimension + x]) == "1") {
                    translate([x , (dimension - 1 - y) , 0]) square(1, center = true);
                }
            }
        }
    }
}

  difference(){
    cube(size+1);
    color("navy")qrCodePattern();
  }
```
preview looks fine
![image](https://github.com/openscad/openscad/assets/93376232/fbedde09-66df-4c57-90b2-3005051ef306)
render
![image](https://github.com/openscad/openscad/assets/93376232/abd82fdb-1141-427b-bfff-d164b4b6bcdd)

kintel:
Reproducible on latest master on macOS as well.

@pca006132 Any ideas? I've seen similar things when playing around with Manifold's 2D triangulator, but I haven't tried to diagnose this yet.

pca006132:
sounds like a bug in the triangulator. is it possible to get a minimal example?

kintel:
It seems tricky to trigger this on a minimal example, but I'll dig around a bit.

kintel:
Some observations:
If you replace the last construct with any of these, it works:
```
difference() {
  translate([0,0,-size-1]) cube(size+1);
  color("navy")qrCodePattern();
}

difference() {
  cube(size+1);
  translate([0,0,size+1]) color("navy")qrCodePattern();
}
```

Also, both `union` and `intersection` work where `difference` fails.

PR #5282, opened 2024-09-05
updated manifold to latest version with double precision floating point

### LINKED ISSUE #4871: CSG op issue when coordinates have resolution differences of more than 4 orders of magnitude
**Describe the bug**
The example was posted on the [mailing list](https://lists.openscad.org/empathy/thread/LDHO7C6LUQ367LH2EDCSR5WLUNNFIJ5N?hash=FKKYVQ53FBXQSUMTKWC2ANZ4JXQ5ZFON#FKKYVQ53FBXQSUMTKWC2ANZ4JXQ5ZFON). It works (slow) with CGAL but produces empty result with Manifold enabled.

**To Reproduce**
```
$fa = 6;
$fs = 0.25;
inf = 1e10;      // very big
eps = 1/128;     // small fudge factor to stop CSG barfing on coincident faces.
layer_height    = 0.25;
teardrop_angle  = 45;

function r2sides(r) = $fn ? $fn : ceil(max(min(360/ $fa, r * 2 * PI / $fs), 5));    //! Replicates the OpenSCAD logic to calculate the number of sides from the radius
function r2sides4n(r) = floor((r2sides(r) + 3) / 4) * 4;                            //! Round up the number of sides to a multiple of 4 to ensure points land on all axes

module circle4n(r, d = undef) {                     //! Circle with multiple of 4 vertices
    R = is_undef(d) ? r : d / 2;
    circle(R, $fn = r2sides4n(R));
}

module extrude_if(h, center = true)                 //! Extrudes 2D object to 3D when `h` is nonzero, otherwise leaves it 2D
    if(h)
        linear_extrude(h, center = center, convexity = 5) // 3D
            children();
    else
        children();                                 // 2D

       
module teardrop(h, r, center = true, truncate = true, chamfer = 0, chamfer_both_ends = true, plus = false) { //! For making horizontal holes that don't need support material, set `truncate = false` to make traditional RepRap teardrops that don't even need bridging
    module teardrop_2d(r, truncate) {
        er = layer_height / 2 - eps;    // Extrusion edge radius
        R = plus ? r + er : r;          // Corrected radius
        offset = plus ? -er : 0;        // Offset inwards
        hull()
            for(side = [0 : 1])
                mirror([side, 0, 0])
                    intersection() {
                        hull()
                            translate([offset, 0]) {
                                circle4n(R);

                                if (teardrop_angle > 0) {
                                    x = R * tan(teardrop_angle / 2);
                                    if(truncate)
                                        translate([0, R / 2])
                                            square([2 * x, R], center = true);
                                    else
                                        polygon([[0, 0], [eps, 0], [0, R + x * tan(min(teardrop_angle, 90 - eps))]]);
                                }
                            }
                        translate([0, -2 * R])
                            square([R, 4 * R]);
                    }
    }

    render(convexity = 5)
        extrude_if(h, center)
            teardrop_2d(r, truncate);
}

module offset_3D(r, chamfer_base = false) { //! Offset 3D shape by specified radius `r`, positive or negative.
    module ball(r)
        if(chamfer_base)
            rotate_extrude()
                intersection() {
                    rotate(180)
                        teardrop(0, r);

                    translate([0, -r])
                        square([r, 2 * r]);
                }
        else
            sphere(r);

    if(r > 0)
        minkowski() {
            children();

            ball(r);
        }
    else
        if(r < 0)
            render() difference() {
                cube(inf / 2, center = true);

                minkowski() {
                    difference() {
                        cube(inf, center = true);

                        children();
                    }
                    ball(-r);
                }
            }
        else
            children();
}

module round_3D(r, ir = undef, or = undef, chamfer_base = false) { //! Round a 3D child single radius or separate inside and outside radii
    IR = is_undef(ir) ? r : ir;
    OR = is_undef(or) ? r : or;
    offset_3D(OR, chamfer_base)
        offset_3D(-OR -IR, chamfer_base)
            offset_3D(IR, chamfer_base)
                children();
}

module shape()
    difference() {
        square(40, center = true);

        square([20, 20], center = true);
    }

module rounds() {
      translate([50, 0])
        round_3D(or = 4, ir = 2, chamfer_base = true, $fn = 16)
            linear_extrude(40, center = true)
                shape();
}

rounds();
```

**Screenshots**
Expected
grafik

With Manifold enabled:
WARNING: [manifold] Minkowski failed with error, falling back to Nef operation: CGAL ERROR: precondition violation!
Expr: ! running
/usr/include/CGAL/Timer.h
Line: 83
grafik

 - OS: Debian/Testing
 - System: Dell XPS13
 - OpenSCAD Version: 2023.11.28.nightly (git 539f1f1) / OBS Build

```
OpenSCAD Version: 2023.11.28.nightly (git 539f1f1)
System information: Linux 6.5.0-4-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.5.10-1 (2023-11-03) x86_64 Debian GNU/Linux trixie/sid 8 CPUs 15.32 GB RAM
User Agent: OpenSCAD/2023.11.28.nightly (git 539f1f1) (Linux x86_64; Debian GNU/Linux trixie/sid)
Compiler: GCC "13.2.0" 64bit
MinGW build: No
Debug build: No
Boost version: 1_74
Eigen version: 3.4.0
CGAL version, kernels: 5.5.1, Cartesian<Gmpq>, Extended_cartesian<Gmpq>, Epeck
OpenCSG version: OpenCSG 1.6.0
Qt version: 5.15.10
QScintilla version: 2.14.1
InputDrivers: JoystickInputDriver
GLib version: 2.78.1
lodepng version: 20210627
libzip version: 1.7.3
fontconfig version: 2.14.2
freetype version: 2.13.2
harfbuzz version: 8.0.1
cairo version: 1.18.0
lib3mf version: 1.8.1
Features: fast-csg, fast-csg-safer, fast-csg-debug, manifold*, roof, input-driver-dbus, lazy-union, vertex-object-renderers*, vertex-object-renderers-indexing*, vertex-object-renderers-direct*, vertex-object-renderers-prealloc*, textmetrics*, import-function*, predictible-output*
```

pca006132:
As described. Fixes #4871 #5211 (actually producing the expected output as CGAL) #5032 #5284.

New issues:
1. Test for #2841 is now failing due to CGAL not accepting the polyhedron output by manifold (consider it as self-intersection). But apparently the issue was never fixed, just behaves somewhat differently with manifold previously?
2. As manifold no longer depends on thrust, existing parallelization code (for minkowski) no longer works. Not sure what is the best way to deal with this, do we want to add thrust dependency here or use the parallelization template library provided by manifold? Those are pretty simple, and I can either expose them from manifold or copy the required function to openscad.
3. The new version of manifold allows users to disable clipper2 support, but this is currently still using the cross-section API (clipper2) to handle polygon fill rule after slicing/projecting the polyhedron. And due to #5281, I think it may be worth it to give clipper2 a try?

TODO:
* [x] Wait for https://github.com/elalish/manifold/issues/916
* [x] Update our code to use MeshGL64
* [x] Test both with and without `-DUSE_MANIFOLD_TRIANGULATOR=ON`
* [x] Go through relevant issues and update if applicable
   * [x] #5294: No change
   * [x] #5211: Fixed
   * [x] #5069: Possible fixed, still uses a lot of memory. May need more testing.
   * [x] #5032: Fixed
   * [x] #4871: Fixed
   * [x] #4617: No change


kintel:
FYI: Updating OpenSCAD to Clipper2 is described here: #4974

kintel:
If the parallelization issue only affects minkowski, we can split that work out into a separate ticket for (re-)optimizing minkowski.

kintel:
The underlying CGAL bug causing #2841 was fixed last week: https://github.com/CGAL/cgal/issues/8379

pca006132:
For #2915, it works with manifold now but the mesh is self-intersecting (epsilon-valid):

image

CGAL will reject the output though. If line 19 is changed to `translate([hullRad*ratio,0,-0.00001]){` with that -0.0001, it works fine.

pca006132:
#5069 now managed to complete in ~11 minutes using not more than ~10G of RAM. I have no idea why but it works.

image

pca006132:
For https://gist.github.com/ochafik/70a6b15e982b7ccd5a79ff9afd99dbcf#file-minkowski-of-minkowski-difference-scad, it sometimes fail with

```
Expression : sf_old->mark()
File       : /nix/store/g9spz3sbkig9l4kyyr3ks6w9nmfcsjzd-cgal-5.6.1/include/CGAL/Convex_decomposition_3/SM_walls.h
Line       : 560
Explanation: 
WARNING: [manifold] Minkowski failed with error, falling back to Nef operation: CGAL ERROR: assertion violation!
Expr: sf_old->mark()
File: /nix/store/g9spz3sbkig9l4kyyr3ks6w9nmfcsjzd-cgal-5.6.1/include/CGAL/Convex_decomposition_3/SM_walls.h
Line: 560

volume: 599809.082911
Expression : e_below != SHalfedge_handle()
File       : /nix/store/g9spz3sbkig9l4kyyr3ks6w9nmfcsjzd-cgal-5.6.1/include/CGAL/Nef_3/SNC_FM_decorator.h
Line       : 418
Explanation: 
ERROR: CGAL error in CGAL_Nef_polyhedron3(): CGAL ERROR: assertion violation!
Expr: e_below != SHalfedge_handle()
File: /nix/store/g9spz3sbkig9l4kyyr3ks6w9nmfcsjzd-cgal-5.6.1/include/CGAL/Nef_3/SNC_FM_decorator.h
Line: 418
```

Not sure if it is specific to this PR or it was broken a while ago. Need further investigation... If it is caused by this PR, we probably need to expose manifold's mesh simplification API to allow openscad to simplify the model up to certain precision to clean up degenerate triangles.

pca006132:
I did some profiling with https://gist.github.com/ochafik/70a6b15e982b7ccd5a79ff9afd99dbcf, it seems that this PR will cause some trouble with minkowski.

1. Because our precision is greatly improved, we now can have many more non-degenerate (w.r.t. our precision value) triangles that were regarded as degenerate and were removed previously. This causes the number of vertices to increase, and make minkowski slower. Currently, the slowest one is minbosl-scad and is slower by 4 times.
2. CGAL is mad at some of our outputs, and we need to do more simplification at higher precision value (worse precision). Currently, I settled on the magic value 1e-5, which seems to make CGAL happy for our test cases, but I don't think this can solve everything.

In case convex decomposition failed with CGAL, the current fallback logic may not work because it can be possible that the whole CSG tree has to rebuild from the start with CGAL operations. Or someone may have to bite the bullet and implement a convex decomposition that can work with epsilon-valid polyhedron, which can be non-trivial. Though we don't really need an optimality guarantee for this one, it can be worse than the decomposition given by CGAL, as long as the composition results are convex polyhedrons it should be fine. Again, this can be done later.

Regarding improving performance by using higher precision value, we can probably try to get the scale of everything in the CSG tree and use the minimal feature size (times some constant, e.g. 1e-5) to set the precision value? Setting precision values local to each geometry will not work because manifold uses maximum precision value when doing boolean operation, so smaller features can get loss. This can be done later, though.

@elalish pinging you as you may be interested in the discussion here.

pca006132:
related: https://github.com/elalish/manifold/issues/675

pca006132:
@UBaer21 I think the problem in that case is about the extrude function in openscad producing something non-manifold. And 3MF export does not support something non-manifold as well. One simple way to work around this is to avoid having 2D polygons that may generate non-manifold mesh when extruded, e.g. just do things in 3D. Very naive changes to your program will work:

```scad
size=20; // smaller 18 works

matrixString=[for(i=[0:size^2]) round(rands(0,1,1,i)[0]) ];

module qrCodePattern(matrixString=matrixString, dimension=size) {
    translate([1,1,-1])union(){
        for (y = [0 : dimension-1]) {
            for (x = [0 : dimension-1]) {
                if (str(matrixString[y * dimension + x]) == "1") {
                    translate([x , (dimension - 1 - y) , 0])
                    linear_extrude(2,convexity=5)
                    square(1, center = true);
                }
            }
        }
    }
}


union(){
  difference(){
    cube(size+1);
    color("navy")qrCodePattern();
  }
}
```

UBaer21:
In some cases when i got the error - i simply could add a cube(.01) outside (or inside) the object and the export error was gone - sometimes rearranging the order of objects - Which shouldn't change the geometry at all. Export as STL show no errors and later external checks and analyze show no problematic triangles. If we could get more info about this error like which triangle can't be added could help.  I had models with two parts (non touching)- Part A works, Part B works  too - but can't export both at once.  Just weird.

pca006132:
Yeah for the other cases it is weird, but I need test cases to debug. At least for the example in #5030 it should be related to extrude not producing something manifold.

elalish:
@UBaer21 if you can compile with the `MANIFOLD_DEBUG` flag enabled, you may be able to catch earlier if Manifold is hitting a problem. Of course it would also be nice to get a more specific error out of the 3MF export call.

pca006132:
Also deals with #5284 by allowing users to use locally installed manifold instead of submodule one, by specifying `USE_BUILTIN_MANIFOLD=FALSE`.

pca006132:
@kintel what do you think if we change the extrude function to manifold's extrude function? That can solve the issue above regarding extrude not giving manifold output.

pca006132:
@UBaer21 the latest artifact should have manifold debug enabled.

kintel:
In terms of using Manifold's extrude function: Last I looked, that function doesn't have the same features as OpenSCAD needs, so it's probably not trivial to align the two extrude feature sets. Also, I'm a bit concerned that aligning OpenSCAD's user-facing features with 3rd party library implementation is a bit risky in terms of allowing us to evolve such features going forward. Especially if Manifold and OpenSCAD ends up having slightly different compatibility strategies.

kintel:
@pca006132 Since this will need some careful testing and depends on some other tasks, I took the liberty of starting a TODO list in the description.

pca006132:
Thanks. Now that `MeshGL64` is added to the upstream, I can update this PR later today.

pca006132:
I think there are some test failures, will check more closely later today.

pca006132:
It seems that the test failures are caused by color differences.

pca006132:
Tested with `-DUSE_MANIFOLD_TRIANGULATOR=ON`, it works fine.

CMakeLists.txt
```diff
@@ -949,15 +950,18 @@ if(EXPERIMENTAL)
...
+      set(CMAKE_UNITY_BUILD ON)
+    else()
+      add_subdirectory(submodules/manifold EXCLUDE_FROM_ALL)
+    endif()
   else()
-    add_subdirectory(submodules/manifold EXCLUDE_FROM_ALL)
+    find_package(manifold REQUIRED)
   endif()
-  target_link_libraries(OpenSCAD PRIVATE manifold polygon)
+  target_link_libraries(OpenSCAD PRIVATE manifold polygon cross_section)
```
kintel, line 964: Are anyone already packaging Manifold and provide CMake or pkg-config configs for this? I'd expect the libraries to be namespaced to e.g. `manifold::manifold`, `manifold::polygon` etc.

hzeller: Yes, I have packaged manifold for NixOS, and as soon as we can have a` find_package()` (#5284), it will work to compile-against and link the system-provided manifold.

kintel: Cool! This PR implements the call to `find_package()`, so that part is sorted. I'm curious how you make the manifold library available in your package; e.g. do you provide a pkg-config .pc file, do you provide a CMake Config macro etc.?

hzeller: Yes :)
Manifold provides both of them, cmake and since recently also pkg-config (https://github.com/elalish/manifold/pull/901 ).

I guess `find-package()` will pick up the cmake macros.

kintel: Btw., I believe the manifold pkg-config only supports linking with the core manifold library. It won't pick up the polygon and cross-section libraries exported by manifold.

hzeller: Yeah, that might need some refinement.

kintel: Also opened https://github.com/elalish/manifold/issues/936 to discuss namespacing.

kintel: Resolving this conversation for now as it's out of scope for this PR.

kintel:
Out of curiosity, was #5211 fixed because you passed the `Polygons` result from `Slice()` and `Project()` into the `CrossSection()` constructor which cleans up duplicate vertices and self-interections?

pca006132:
> Out of curiosity, was #5211 fixed because you now pass the `Polygons` result from `Slice()` and `Project()` into the `CrossSection()` constructor which cleans up duplicate vertices and self-interections?

No, we were using `CrossSection` previously as well. I guess it is some other fixes causing the issue to disappear.

pca006132:
#5218 should now be fixed.

kintel (approved):

Issue #5216, opened 2024-07-14
Color rendering: color() without arguments renders black with Manifold


kintel:
This minimal example renders a black cube in render mode (both CGAL and Manifold), but a cube with default color in preview mode.
```
color() cube();
```

See also https://github.com/openscad/openscad/blob/master/tests/data/scad/3D/features/color-tests.scad

UBaer21:
I assume replacing "undef" with black ([0,0,0]) is also the reason why  color(alpha=.5)color("blue"); works only in render but not in preview.  As alpha can be set without defining a color but can't overwrite existing alphas without defining also a color. Would be nice if that would be possible too.

kintel:
@UBaer21 I think specifying a partial color was never intended, so if it works anywhere that's a coincidence :)
..but it's an interesting feature request. It's different from this issue though.

PR #5288, opened 2024-09-09
PovRay export

folkertvanheusden:
This patch set adds support for exporting PovRay scripts from OpenSCAD.
Note that it exports triangles, not the original CSG scripting.

By default it adds "radiosity" to the output-scripts, this makes rendering a tad slow. Fix that by removing:

     global_settings { photons { count 20000 autostop 0 jitter .4 } radiosity { Rad_Settings(Radiosity_Normal, off, off) } }

To render a .pov-script using povray, invoke:

     povray +Imy_script.pov +W3840 +H2160 +A0.01

This renders it at 4k resolution with anti aliasing enabled.

There's a crude animation added to it that can be controlled via +K. +K here lets you control the rotation of the camera around the object. This is of course optional.

To render such an animation, run e.g.:

     for i in `seq 0 360` ; do DISPLAY= povray +Irpi3.pov +W3840 +H2160 +A0.01 +k$i +O${i}.png ; done

to create an mp4 of the result:

    fmpeg -framerate 25 -i %d.png -c:v libx264 -r 25 -pix_fmt yuv420p output.mp4

Example animation render: https://vanheusden.com/permshare/openscad-povray-example.mp4
Example POV script: https://vanheusden.com/permshare/openscad-povray-example2.pov
Example PNG render output: https://vanheusden.com/permshare/openscad-povray-example2.png


t-paul (changes requested):
Nice, I'm currently rendering a simple animation :tada:.

I've just added a couple of minor notes for things to check. Also I'm getting 2 warnings from povray. Considering version 3.7 is probably old enough to be a safe target, it would be great if we could address those.

```
Possible Parse Error: assumed_gamma not specified in this POV-Ray 3.7 or later
 scene. Future versions of POV-Ray may consider this a fatal error. To avoid
 this warning, explicitly specify 'assumed_gamma 1.0' in the global_settings
 section. See the documentation for more details.
```
```
Parse Warning: This scene had other declarations preceding the first #version
 directive. Please be aware that as of POV-Ray 3.7, unless already specified via
 an INI option, a #version is expected as the first declaration in a scene file.
 If this is not done, POV-Ray may apply compatibility settings to some features
 that are intended to make pre-3.7 scenes render as designed. You are strongly
 encouraged to add a #version statement to the scene to make your intent clear.
 Future versions of POV-Ray may make the presence of a #version mandatory.
```

t-paul:
Very nice result for a really simple script: https://files.openscad.org/tmp/tetris.mp4

Just moved the camera up a bit.

```
$fa = 2; $fs = 0.2;

s1 = 60;
s2 = 80;
s3 = 2;

eps = 0.1;

module block() {
	hull()
		for (r = [[0, 0, 0], [90, 0, 0], [0, 90, 0]])
			rotate(r)
				minkowski() {
					linear_extrude(s2 - 2 * s3 + 2 * eps, center = true)
						square(s1, center = true);
					sphere(s3);
				}
}

module blocks(pos) {
	for (p = pos)
		translate(s2 * p)
			block();
}

translate(s2 *[-4, -2]) color("red") blocks([[0, 0], [1, 0], [2, 0], [3, 0]]);
translate(s2 *[2, 2]) color("khaki") blocks([[0, 0], [1, 0], [0, 1], [1, 1]]);
translate(s2 *[1, -2]) color("orange") blocks([[0, 0], [1, 0], [2, 0], [2, 1]]);
translate(s2 *[-4, 2]) color("green") blocks([[0, 0], [1, 0], [1, 1], [2, 1]]);
translate(s2 *[-1, 0]) color("purple") blocks([[0, 0], [1, 0], [2, 0], [1, 1]]);
```

jordanbrown0:
It is indeed very pretty, but I'm confused about where the light source is.  Faces that are close to us (all of the "outside" faces as it rotates) are dark, but faces that are further away but facing toward us (the same direction as the "near" faces) are lit.  Am I just not interpreting what I'm seeing correctly, or is there something off about the lighting?
{0AF0D0DA-9944-4856-A02C-5160950B3CD7}
Where would the light source be, so that the fronts of the white and yellow blocks are in shadow, but the fronts of the green, purple, and red blocks are lit?

jordanbrown0:
Also, it's backwards.  Or up-side down.  Or something.

image

folkertvanheusden:
>
>
> I've just added a couple of minor notes for things to check. Also I'm
> getting 2 warnings from povray. Considering version 3.7 is probably old
> enough to be a safe target, it would be great if we could address those.
>
> Possible Parse Error: assumed_gamma not specified in this POV-Ray 3.7 or later
>  scene. Future versions of POV-Ray may consider this a fatal error. To avoid
>  this warning, explicitly specify 'assumed_gamma 1.0' in the global_settings
>  section. See the documentation for more details.
>
> Parse Warning: This scene had other declarations preceding the first #version
>  directive. Please be aware that as of POV-Ray 3.7, unless already specified via
>  an INI option, a #version is expected as the first declaration in a scene file.
>  If this is not done, POV-Ray may apply compatibility settings to some features
>  that are intended to make pre-3.7 scenes render as designed. You are strongly
>  encouraged to add a #version statement to the scene to make your intent clear.
>  Future versions of POV-Ray may make the presence of a #version mandatory.
>
>
Ok, fixed!


> It should be possible to use std::clamp
> <https://en.cppreference.com/w/cpp/algorithm/clamp> as we declare c++17
> compatibility.
>
done


> +  output << "// hello povray!\n\n";
>
> Maybe something like "Generated by OpenSCAD from <source file>
> <https://github.com/openscad/openscad/blob/dbd7d4f3d93356fedc0cfac1989e2a1e85335de8/src/io/export_pdf.cc#L259>"
> instead?
>
I'd love to, but where do I get that info from? The name of the source-file
that is. Is it in a global parameter block or so?



> The icons are not yet automatically generated in the build process, they
> need to be updated via scripts/generate-svg-icons.sh, checked in and
> added to the Qt resource file
> https://github.com/openscad/openscad/blob/dbd7d4f3d93356fedc0cfac1989e2a1e85335de8/resources/common.qrc.in#L37
>

done


> Another place where POV should be added is in the settings for the 3d
> export toolbar
> https://github.com/openscad/openscad/blob/dbd7d4f3d93356fedc0cfac1989e2a1e85335de8/src/gui/Settings.cc#L128
>

I've added it but could not find where it is in the gui. So that's not
tested.

folkertvanheusden:
The lights are placed arbitrarily. I can add more but my idea is (open for
discussion!) that it simply emits the povray code and that the user then
determines where the camera should be located as well as the lights. Also
the radiosity can be optional as it can take quite a while to render with
it.

folkertvanheusden:
Try rendering with +K180 then it is rotated to somewhat more correctly.
Maybe the viewpoint of the camera in openscad can be made available for the
export-code?

jordanbrown0:
WRT lighting - if the results make sense for the lights used, that's fine.  But maybe the default light should match the OpenSCAD light.

WRT rotation, yes, I'm sure you can rotate to match (assuming that it's not a mirroring problem), but again the defaults should match.

folkertvanheusden:
Lighting: but what is the position of the OpenSCAD light?

nophead:
The lighting in OpenSCAD is awful. It has two diffuse light diametrically opposed for a trick with OpenCSG. One is above and behind and the other below and in front so objects facing the camera are not well lit.

jordanbrown0:
https://github.com/openscad/openscad/blob/ea7f4aab937bd61c5c35a3d4b72711c07af34052/src/glview/GLView.cc#L253
```
  GLfloat light_position0[] = {-1.0, +1.0, +1.0, 0.0};
  GLfloat light_position1[] = {+1.0, -1.0, -1.0, 0.0};
```
I don't know beans about OpenGL, but I suspect that's in a -1 to +1 coordinate space.

(It's not this topic, but I think it might be interesting to have an OpenSCAD module that adds a light with specified parameters at the origin, subject to enclosing transformations.)

folkertvanheusden:
> WRT rotation, yes, I'm sure you can rotate to match (assuming that it's not a mirroring problem), but again the defaults should match.

Will look at that.

t-paul:
> I'd love to, but where do I get that info from? The name of the source-file that is. Is it in a global parameter block or so?

PDF export is currently adding the filename to the meta data.
https://github.com/openscad/openscad/blob/master/src/io/export_pdf.cc#L260

This is contained in `ExportInfo` passed along in `export.cc:exportFile()`.

folkertvanheusden:
image

https://github.com/openscad/openscad/pull/5288

https://github.com/user-attachments/assets/b68311a2-7b8f-42e5-afab-d5f1066e2791

t-paul:
I'm not sure why the failure happens now, but it seems the root cause is the unity build colliding between Qt and boost. Can you disable this by changing `-DCMAKE_UNITY_BUILD=ON` to `OFF` in

https://github.com/openscad/openscad/blob/master/scripts/github-ci.sh#L34

src/io/export_pov.cc
```diff
@@ -0,0 +1,122 @@
...
+      auto color_index = ps->color_indices[pi];
+      if (color_index >= 0) {
+        auto color = ps->colors[color_index];
+        r = std::clamp(double(color[0]), 0., 1.);
+        g = std::clamp(double(color[1]), 0., 1.);
+        b = std::clamp(double(color[2]), 0., 1.);
+        a = std::clamp(1.0 - color[3], 0., 1.);
+      }
+    }
+    if (r == 0 && g == 0 && b == 0)  // work around for black objects in 5185
```
kintel, line 84: Is this a workaround for #5216 ? If so, I suggest leaving it unfixed and add a test case.

src/io/export_pov.cc
```diff
@@ -0,0 +1,122 @@
...
+  double avg_x = 0;
+  double avg_y = 0;
+  double avg_z = 0;
+  double tsd_x = 0;
+  double tsd_y = 0;
+  double tsd_z = 0;
+  size_t avg_n = 0;
+
+  for (size_t pi=0; pi<ps->indices.size(); pi++) {
+    const auto &t = ps->indices.at(pi);
```
kintel, line 59: What is "t"? I assume it's short for "triangle", but in reality it's a polygon.
-> in general, try to use longer variable names, unless it's painfully obvious what the name means, as code is primarily meant for human consumption.

kintel, line 59: I find it more readable to use `[pi]` than `.at(pi)`.

src/io/export_pov.cc
```diff
@@ -0,0 +1,122 @@
...
+    for (size_t i=0; i<t.size(); i++) {
+      if (i)
+        output << ", ";
+      output << "<" << ps->vertices[t.at(i)].x() << ", " << -ps->vertices[t.at(i)].y() << ", " << ps->vertices[t.at(i)].z() << ">";
+      avg_x += ps->vertices[t.at(i)].x();
+      avg_y += -ps->vertices[t.at(i)].y();
+      avg_z += ps->vertices[t.at(i)].z();
+      tsd_x += ps->vertices[t.at(i)].x() * ps->vertices[t.at(i)].x();
+      tsd_y += ps->vertices[t.at(i)].y() * ps->vertices[t.at(i)].y();
+      tsd_z += ps->vertices[t.at(i)].z() * ps->vertices[t.at(i)].z();
```
kintel, line 70: Lots of repetition here and it's easy to miss something.
Suggestion:
```
const auto& x = ps->vertices[i].x();
const auto& y = -ps->vertices[i].y();
const auto& z = ps->vertices[i].z();
```
..then use x, y an z repeated.
...or some similar construct

src/io/export_pov.cc
```diff
@@ -0,0 +1,122 @@
...
+  double tsd_z = 0;
+  size_t avg_n = 0;
+
+  for (size_t pi=0; pi<ps->indices.size(); pi++) {
+    const auto &t = ps->indices.at(pi);
+    output << "polygon { " << t.size() + 1 << ", \n";
+    for (size_t i=0; i<t.size(); i++) {
+      if (i)
+        output << ", ";
+      output << "<" << ps->vertices[t.at(i)].x() << ", " << -ps->vertices[t.at(i)].y() << ", " << ps->vertices[t.at(i)].z() << ">";
```
kintel, line 64: The negated y axis is a bit surprising here. Does PovRay use a left-handed coordinate system? This is worth a comment.

src/io/export_pov.cc
```diff
@@ -0,0 +1,122 @@
...
+      avg_x += ps->vertices[t.at(i)].x();
+      avg_y += -ps->vertices[t.at(i)].y();
+      avg_z += ps->vertices[t.at(i)].z();
+      tsd_x += ps->vertices[t.at(i)].x() * ps->vertices[t.at(i)].x();
+      tsd_y += ps->vertices[t.at(i)].y() * ps->vertices[t.at(i)].y();
+      tsd_z += ps->vertices[t.at(i)].z() * ps->vertices[t.at(i)].z();
+      avg_n++;
+    }
+    double r = 1.0, g = 1.0, b = 1.0, a = 0.0;
```
kintel, line 73: Why double? Should be fine to keep it as float as you're not trying to do any calculations needing more precision here

src/io/export_pov.cc
```diff
@@ -0,0 +1,122 @@
...
+
+  double l_x = center_x + sd_x * 5;  // '5' is chosen arbitrarily
+  double l_y = center_y + sd_y * 5;
+  double l_z = center_z + sd_z * 5;
+
+  std::vector<double> lx { -l_x, 0, +l_x };
+  std::vector<double> ly { -l_y, 0, +l_y };
+  std::vector<double> lz { -l_z, 0, +l_z };
+
+  constexpr const double brightness = 0.2;  // 1.0 is way too bright
```
kintel, line 110: const is redundant

src/io/export_pov.cc
```diff
@@ -0,0 +1,122 @@
...
+
+  for(auto cur_lx: lx) {
+	  for(auto cur_ly: ly) {
+		  for(auto cur_lz: lz)
+			  output << "light_source { <" << cur_lx << ", " << cur_ly << ", " << cur_lz << "> color rgb <" << brightness << ", " << brightness << ", " << brightness << "> }\n";
+	  }
+  }
+
+  output << "camera { look_at <0, 0, 0> location <0, " << dist * 15 << ", 0> sky <0, 0, 1> rotate <55, clock * 3, clock - 25> }\n";
```
kintel, line 119: How does your example have a real value for look_at, while it's hard-coded to 0,0,0 here?

src/io/export_pov.cc
```diff
@@ -0,0 +1,122 @@
...
+  double center_z = avg_z / avg_n;
+
+  double sd_x = sqrt(tsd_x / avg_n - pow(center_x, 2.));
+  double sd_y = sqrt(tsd_y / avg_n - pow(center_y, 2.));
+  double sd_z = sqrt(tsd_z / avg_n - pow(center_z, 2.));
+  double dist = pow(sd_x * sd_x + sd_y * sd_y + sd_z * sd_z, 1/3.);
+
+  double l_x = center_x + sd_x * 5;  // '5' is chosen arbitrarily
+  double l_y = center_y + sd_y * 5;
+  double l_z = center_z + sd_z * 5;
```
kintel, line 104: I have a hard time understanding what's going on here. Are you on purpose trying to find the center of gravity and average distance from that center? It kind of feels like you might be better off working with bounding boxes or bounding spheres.

src/io/export_pov.cc
t-paul: Looking through the povray docs, it seems the coordinate system can be set via `up`and `right` vectors. So maybe it's better to just dump the objects in OpenSCAD coordinate system?

hzeller:
Nice!
I successfully rendered some [old project](https://github.com/hzeller/hot-snipper/blob/master/band-cutter-machine.scad ) including transparency, but the pov-output of [another one](https://github.com/hzeller/digi-spherometer/blob/master/frame/frame.scad) could not be parsed by povray  (3.8.0 beta 2) with a message like:

```
File 'digi-spherometer/frame/frame.pov' line 83250:
 Possible Parse Error: Singular matrix in MInvers.
Fatal error in parser: Uncategorized error.
```

Data around there looks like this; maybe the vertices are too close together ? This is certainly a fault of my model, but maybe there are ways to repair that (not in this PR, I am happy to see this landed first).

```
...
polygon { 4, 
<-17, 17.39, 2.50025>, <-17.0002, 17.39, 2.5003>, <-17.0501, 17.39, 2.56089>, <-17, 17.39, 2.50025>
texture { pigment { color rgbf <0.501961, 0.501961, 0.501961, 0> } }
finish { MATERIAL } interior { MATERIAL_INT }
}
...
```

(also cool would be if the camera position `$vpr` was available to the export modules that that the povray export can emit the viewport chosen by the user in the UI or even programmatically by setting `$vpr`. Also something for later, as it looks like it might need to be wired all the way through to `ExportInfo`).

folkertvanheusden:
>
> Mostly looks good. Added some comments.
> Also, while it's tricky to add graphical tests, we should at the minimum
> add a sanity check that the export work, similar to how we export a simple
> object in OFF, OBJ etc. and compare the output against a golden file.
> ..plus a similar one for color support.
>
I'm not entirely sure where to locate the test-files.
Am I right that the input should be in openscad/tests/data/scad/pov and the
result in openscad/tests/data/scad/pov/ ?


> ------------------------------
>
> In src/io/export_pov.cc
> <https://github.com/openscad/openscad/pull/5288#discussion_r1776293448>:
>
> > +      tsd_y += ps->vertices[t.at(i)].y() * ps->vertices[t.at(i)].y();
> +      tsd_z += ps->vertices[t.at(i)].z() * ps->vertices[t.at(i)].z();
> +      avg_n++;
> +    }
> +    double r = 1.0, g = 1.0, b = 1.0, a = 0.0;
> +    if (has_color) {
> +      auto color_index = ps->color_indices[pi];
> +      if (color_index >= 0) {
> +        auto color = ps->colors[color_index];
> +        r = std::clamp(double(color[0]), 0., 1.);
> +        g = std::clamp(double(color[1]), 0., 1.);
> +        b = std::clamp(double(color[2]), 0., 1.);
> +        a = std::clamp(1.0 - color[3], 0., 1.);
> +      }
> +    }
> +    if (r == 0 && g == 0 && b == 0)  // work around for black objects in 5185
>
> Is this a workaround for #5216
> <https://github.com/openscad/openscad/issues/5216> ? If so, I suggest
> leaving it unfixed and add a test case.
>
I did the clamping as other code was doing it as well. I've removed it now.
Note that the black-test is there because some example .scad files I saw
use the color black for objects. That's hard to see when rendered :-)


> ------------------------------
>
> In src/io/export_pov.cc
> <https://github.com/openscad/openscad/pull/5288#discussion_r1776294513>:
>
> > +  output << "global_settings { assumed_gamma 1.0 }\n";
> +  output << "#declare MATERIAL=finish { specular 0.5 roughness 0.001 reflection{0 0.63 fresnel} ambient 0 diffuse 0.6 conserve_energy }\n";
> +  output << "#declare MATERIAL_INT=interior{ior 1.32}\n";
> +
> +  auto has_color = !ps->color_indices.empty();
> +
> +  double avg_x = 0;
> +  double avg_y = 0;
> +  double avg_z = 0;
> +  double tsd_x = 0;
> +  double tsd_y = 0;
> +  double tsd_z = 0;
> +  size_t avg_n = 0;
> +
> +  for (size_t pi=0; pi<ps->indices.size(); pi++) {
> +    const auto &t = ps->indices.at(pi);
>
> What is "t"? I assume it's short for "triangle", but in reality it's a
> polygon.
> -> in general, try to use longer variable names, unless it's painfully
> obvious what the name means, as code is primarily meant for human
> consumption.
>
Done


> ------------------------------
>
> In src/io/export_pov.cc
> <https://github.com/openscad/openscad/pull/5288#discussion_r1776296299>:
>
> > +  output << "global_settings { assumed_gamma 1.0 }\n";
> +  output << "#declare MATERIAL=finish { specular 0.5 roughness 0.001 reflection{0 0.63 fresnel} ambient 0 diffuse 0.6 conserve_energy }\n";
> +  output << "#declare MATERIAL_INT=interior{ior 1.32}\n";
> +
> +  auto has_color = !ps->color_indices.empty();
> +
> +  double avg_x = 0;
> +  double avg_y = 0;
> +  double avg_z = 0;
> +  double tsd_x = 0;
> +  double tsd_y = 0;
> +  double tsd_z = 0;
> +  size_t avg_n = 0;
> +
> +  for (size_t pi=0; pi<ps->indices.size(); pi++) {
> +    const auto &t = ps->indices.at(pi);
>
> I find it more readable to use [pi] than .at(pi).
>
Done


> ------------------------------
>
> In src/io/export_pov.cc
> <https://github.com/openscad/openscad/pull/5288#discussion_r1776298528>:
>
> > +      output << "<" << ps->vertices[t.at(i)].x() << ", " << -ps->vertices[t.at(i)].y() << ", " << ps->vertices[t.at(i)].z() << ">";
> +      avg_x += ps->vertices[t.at(i)].x();
> +      avg_y += -ps->vertices[t.at(i)].y();
> +      avg_z += ps->vertices[t.at(i)].z();
> +      tsd_x += ps->vertices[t.at(i)].x() * ps->vertices[t.at(i)].x();
> +      tsd_y += ps->vertices[t.at(i)].y() * ps->vertices[t.at(i)].y();
> +      tsd_z += ps->vertices[t.at(i)].z() * ps->vertices[t.at(i)].z();
>
> Lots of repetition here and it's easy to miss something.
> Suggestion:
>
> const auto& x = ps->vertices[i].x();
> const auto& y = -ps->vertices[i].y();
> const auto& z = ps->vertices[i].z();
>
> ..then use x, y an z repeated.
> ...or some similar construct
>
Done


> ------------------------------
>
> In src/io/export_pov.cc
> <https://github.com/openscad/openscad/pull/5288#discussion_r1776298765>:
>
> > +
> +  double avg_x = 0;
> +  double avg_y = 0;
> +  double avg_z = 0;
> +  double tsd_x = 0;
> +  double tsd_y = 0;
> +  double tsd_z = 0;
> +  size_t avg_n = 0;
> +
> +  for (size_t pi=0; pi<ps->indices.size(); pi++) {
> +    const auto &t = ps->indices.at(pi);
> +    output << "polygon { " << t.size() + 1 << ", \n";
> +    for (size_t i=0; i<t.size(); i++) {
> +      if (i)
> +        output << ", ";
> +      output << "<" << ps->vertices[t.at(i)].x() << ", " << -ps->vertices[t.at(i)].y() << ", " << ps->vertices[t.at(i)].z() << ">";
>
> The negated y axis is a bit surprising here. Does PovRay use a left-handed
> coordinate system? This is worth a comment.
>
I've changed the camera output now to use the coordinate system of
OpenSCAD. So the negating is removed now.


> ------------------------------
>
> In src/io/export_pov.cc
> <https://github.com/openscad/openscad/pull/5288#discussion_r1776301083>:
>
> > +  for (size_t pi=0; pi<ps->indices.size(); pi++) {
> +    const auto &t = ps->indices.at(pi);
> +    output << "polygon { " << t.size() + 1 << ", \n";
> +    for (size_t i=0; i<t.size(); i++) {
> +      if (i)
> +        output << ", ";
> +      output << "<" << ps->vertices[t.at(i)].x() << ", " << -ps->vertices[t.at(i)].y() << ", " << ps->vertices[t.at(i)].z() << ">";
> +      avg_x += ps->vertices[t.at(i)].x();
> +      avg_y += -ps->vertices[t.at(i)].y();
> +      avg_z += ps->vertices[t.at(i)].z();
> +      tsd_x += ps->vertices[t.at(i)].x() * ps->vertices[t.at(i)].x();
> +      tsd_y += ps->vertices[t.at(i)].y() * ps->vertices[t.at(i)].y();
> +      tsd_z += ps->vertices[t.at(i)].z() * ps->vertices[t.at(i)].z();
> +      avg_n++;
> +    }
> +    double r = 1.0, g = 1.0, b = 1.0, a = 0.0;
>
> Why double? Should be fine to keep it as float as you're not trying to do
> any calculations needing more precision here
>
Done.

(I tend to use double where possible as on some platforms double is faster
than float)


> ------------------------------
>
> In src/io/export_pov.cc
> <https://github.com/openscad/openscad/pull/5288#discussion_r1776301296>:
>
> > +  double center_z = avg_z / avg_n;
> +
> +  double sd_x = sqrt(tsd_x / avg_n - pow(center_x, 2.));
> +  double sd_y = sqrt(tsd_y / avg_n - pow(center_y, 2.));
> +  double sd_z = sqrt(tsd_z / avg_n - pow(center_z, 2.));
> +  double dist = pow(sd_x * sd_x + sd_y * sd_y + sd_z * sd_z, 1/3.);
> +
> +  double l_x = center_x + sd_x * 5;  // '5' is chosen arbitrarily
> +  double l_y = center_y + sd_y * 5;
> +  double l_z = center_z + sd_z * 5;
> +
> +  std::vector<double> lx { -l_x, 0, +l_x };
> +  std::vector<double> ly { -l_y, 0, +l_y };
> +  std::vector<double> lz { -l_z, 0, +l_z };
> +
> +  constexpr const double brightness = 0.2;  // 1.0 is way too bright
>
> const is redundant
>
Indeed! Today I learned :-)


> ------------------------------
>
> In src/io/export_pov.cc
> <https://github.com/openscad/openscad/pull/5288#discussion_r1776309292>:
>
> > +  double l_z = center_z + sd_z * 5;
> +
> +  std::vector<double> lx { -l_x, 0, +l_x };
> +  std::vector<double> ly { -l_y, 0, +l_y };
> +  std::vector<double> lz { -l_z, 0, +l_z };
> +
> +  constexpr const double brightness = 0.2;  // 1.0 is way too bright
> +
> +  for(auto cur_lx: lx) {
> +	  for(auto cur_ly: ly) {
> +		  for(auto cur_lz: lz)
> +			  output << "light_source { <" << cur_lx << ", " << cur_ly << ", " << cur_lz << "> color rgb <" << brightness << ", " << brightness << ", " << brightness << "> }\n";
> +	  }
> +  }
> +
> +  output << "camera { look_at <0, 0, 0> location <0, " << dist * 15 << ", 0> sky <0, 0, 1> rotate <55, clock * 3, clock - 25> }\n";
>
> How does your example have a real value for look_at, while it's hard-coded
> to 0,0,0 here?
>
Because initially I did not realise most (all?) drawings are centered at
0,0,0 in OpenSCAD. I did a conversion earlier for other formats and there
it worked better to calculate a camera position.
But after a bit of experimenting I noticed that for O-S, 0,0,0 works better
(together with the dist vector and the rotation)



> ------------------------------
>
> In src/io/export_pov.cc
> <https://github.com/openscad/openscad/pull/5288#discussion_r1776309909>:
>
> > +  double center_x = avg_x / avg_n;
> +  double center_y = avg_y / avg_n;
> +  double center_z = avg_z / avg_n;
> +
> +  double sd_x = sqrt(tsd_x / avg_n - pow(center_x, 2.));
> +  double sd_y = sqrt(tsd_y / avg_n - pow(center_y, 2.));
> +  double sd_z = sqrt(tsd_z / avg_n - pow(center_z, 2.));
> +  double dist = pow(sd_x * sd_x + sd_y * sd_y + sd_z * sd_z, 1/3.);
> +
> +  double l_x = center_x + sd_x * 5;  // '5' is chosen arbitrarily
> +  double l_y = center_y + sd_y * 5;
> +  double l_z = center_z + sd_z * 5;
>
> I have a hard time understanding what's going on here. Are you on purpose
> trying to find the center of gravity and average distance from that center?
> It kind of feels like you might be better off working with bounding boxes
> or bounding spheres.
>
That's exactly what I'm doing there, finding the gravity/distance.
With bounding boxes/spheres I also need to add some arbitrary distance to
each point to make the lighting "work" I think or else light_sources may
end up in the "material" (unless I misunderstood).

folkertvanheusden:
Hi,

Nice!
> I successfully rendered some old project
> <https://github.com/hzeller/hot-snipper/blob/master/band-cutter-machine.scad>
> including transparency, but the pov-output of another one
> <https://github.com/hzeller/digi-spherometer/blob/master/frame/frame.scad>
> could not be parsed by povray (3.8.0 beta 2) with a message like:
>
> File 'digi-spherometer/frame/frame.pov' line 83250:
>  Possible Parse Error: Singular matrix in MInvers.
> Fatal error in parser: Uncategorized error.
>
>
Can this be a regression in povray 3.8.0 beta 2 instead?
Because 3.7.0.10.unofficial (unofficial: ubuntu build) runs fine with it:
https://vanheusden.com/permshare/spherometer_frame_main_block-povray.tar.xz

folkertvanheusden:
That file you included runs fine with 3.7 as well.

hzeller:
That file you included runs fine with 3.7 as well.


Alright, cool, will dig into the sources of povray to see what is happening
there.


>

folkertvanheusden:
f.w.i.w.: povray 3.8 741b6f8771b8bdbd27bd4996417220128b74fce7 fails
differently:

File '../spherometer_frame_main_block.pov' line 36365: Parse Error: Viewing
 angle has to be smaller than 180 degrees.

That's with
camera { look_at <0, 0, 0> location <0, -124.945, 0> up <0, 0, 1> right <1,
0, 0> sky <0, 0, 1> rotate <125, 0, 25> }

If you like, we can take the povray discussion to somewhere else?

kintel:
In terms of tests, what I meant was adding an automated test using the smallest example possible. For an example, see:
https://github.com/openscad/openscad/blob/master/tests/data/scad/obj/obj-export.scad
https://github.com/openscad/openscad/blob/master/tests/regression/objexport/obj-export-expected.obj

Wired up like this:
https://github.com/openscad/openscad/blob/master/tests/CMakeLists.txt#L1134

The -expected file is auto-generated by setting TEST_GENERATE=1 as discussed here:
https://github.com/openscad/openscad/blob/master/doc/testing.txt

If you struggle with this, we can probably lend you a hand once the rest of the PR is good to go.

kintel:
> Because initially I did not realise most (all?) drawings are centered at
0,0,0 in OpenSCAD. I did a conversion earlier for other formats and there
it worked better to calculate a camera position.
But after a bit of experimenting I noticed that for O-S, 0,0,0 works better
(together with the dist vector and the rotation)

Centering in the origin is a design decision and not something OpenSCAD does. For generating a default view, I would suggest choosing an (arbitrary?) view angle and pointing to the center of the bounding box as a start.

kintel:
> That's exactly what I'm doing there, finding the gravity/distance.
With bounding boxes/spheres I also need to add some arbitrary distance to
each point to make the lighting "work" I think or else light_sources may
end up in the "material" (unless I misunderstood).

Not sure I understand. Today you have to introduce the (arbitrary) scale factor of 5 since you operate on _average_ vertex distance from the center of gravity. If you instead calculate a bounding box, you know exactly where the center of your object is and exactly how far you need to place the lights to not interfere with the geometry. Then you can place lights at some distance away from the object (e.g. measured relative to the object extents).

src/io/export_pov.cc
```diff
@@ -0,0 +1,124 @@
...
+  std::vector<double> lx { -light_vector_x, 0, +light_vector_x };
+  std::vector<double> ly { -light_vector_y, 0, +light_vector_y };
+  std::vector<double> lz { -light_vector_z, 0, +light_vector_z };
+
+  constexpr float brightness = 0.2;  // 1.0 is way too bright
+
+  for(auto cur_lx: lx) {
+	  for(auto cur_ly: ly) {
+		  for(auto cur_lz: lz)
+			  output << "light_source { <" << cur_lx << ", " << cur_ly << ", " << cur_lz << "> color rgb <" << brightness << ", " << brightness << ", " << brightness << "> }\n";
```
kintel, line 118: This crates 27 light sources. Not sure I understand the reasoning behind this? Are you just trying to sprinkle light everywhere so that no internal part of the object is in shadows?

folkertvanheusden:
I've now calculated the boundingbox. Then I point the camera to its center
and moved to "x1 + (x2 - x1) * distance" with distance being 1.5.
Unfortunately that gives problems with the lighting.

[image: tetris.png]

> This crates 27 light sources. Not sure I understand the reasoning behind
this? Are you just trying to sprinkle light everywhere so that no internal
part of the object is in shadows?

That's correct. Now there again are 27 light sources but their placement is
not ideal. e.g.:

[image: tetris.png]

So I would *suggest* that we *do *use the boundingbox, but only for camera
placement. And keep the other code for lighting.
Do you agree?

kintel (commented):
I didn't see any bounding box calculation code.
Tip: Instead of doing this manually, do `auto bbox = polyset.getBoundingBox()`, and `mybbox.extend(otherbbox)`

Could you update your code and pov file to show the bad lighting? This sounds like a bug in your code.

folkertvanheusden:
On Fri, Sep 27, 2024 at 4:26 PM Marius Kintel ***@***.***>
wrote:

> ***@***.**** commented on this pull request.
>
> I didn't see any bounding box calculation code.
>
I had not committed nor pushed it :-)


> Tip: Instead of doing this manually, do auto bbox =
> polyset.getBoundingBox(), and mybbox.extend(otherbbox)
>
Ok, replaced it with BoundingBox bbox = geom->getBoundingBox(); now.


> Could you update your code and pov file to show the bad lighting? This
> sounds like a bug in your code.
>
I believe I fixed it now.

kintel (commented):
It's getting better!
Let's wire up that test, so that the test files can be compared automatically.

src/io/export_pov.cc
```diff
@@ -0,0 +1,114 @@
...
+
+  double dx = max_x - min_x;
+  double dy = max_y - min_y;
+  double dz = max_z - min_z;
+
+  constexpr double move_away_factor = 2.;
+  std::vector<double> lx { min_x - dx * move_away_factor, 0, max_x + dx * move_away_factor };
+  std::vector<double> ly { min_y - dy * move_away_factor, 0, max_y + dy * move_away_factor };
+  std::vector<double> lz { min_z - dz * move_away_factor, 0, max_z + dz * move_away_factor };
+
```
kintel, line 99: the zeroes here should probably be the center of the bounding box.
FYI: bbox.center().

src/io/export_pov.cc
```diff
@@ -0,0 +1,114 @@
...
+  auto & min_y = bbox.min().y();
+  auto & min_z = bbox.min().z();
+
+  auto & max_x = bbox.max().x();
+  auto & max_y = bbox.max().y();
+  auto & max_z = bbox.max().z();
+
+  double dx = max_x - min_x;
+  double dy = max_y - min_y;
+  double dz = max_z - min_z;
```
kintel, line 93: You could probably reduce the # lines of code by doing bbox.diagonal().

folkertvanheusden:
> In src/io/export_pov.cc
> <https://github.com/openscad/openscad/pull/5288#discussion_r1779717553>:
>
> > +  std::vector<double> lx { min_x - dx * move_away_factor, 0, max_x + dx * move_away_factor };
> +  std::vector<double> ly { min_y - dy * move_away_factor, 0, max_y + dy * move_away_factor };
> +  std::vector<double> lz { min_z - dz * move_away_factor, 0, max_z + dz * move_away_factor };
> +
>
> the zeroes here should probably be the center of the bounding box.
> FYI: bbox.center().
>
Oops! yes


> ------------------------------
>
> In src/io/export_pov.cc
> <https://github.com/openscad/openscad/pull/5288#discussion_r1779718611>:
>
> > +  auto & min_x = bbox.min().x();
> +  auto & min_y = bbox.min().y();
> +  auto & min_z = bbox.min().z();
> +
> +  auto & max_x = bbox.max().x();
> +  auto & max_y = bbox.max().y();
> +  auto & max_z = bbox.max().z();
> +
> +  double dx = max_x - min_x;
> +  double dy = max_y - min_y;
> +  double dz = max_z - min_z;
>
> You could probably reduce the # lines of code by doing bbox.diagonal().
>
Tried that but I could not reduce the line count by that.
Because we probably(?) want a double dx = bbox.diagonal().dx etc as well?


Then the testing: I followed the steps from doc/testing.txt and placed an
scad under ../tests/data/pov/conversion-test.pov
then same thing under ../tests/data/scad/pov/conversion-test.scad
Then I added
list(APPEND EXPORT_POV_TEST_FILES ${TEST_SCAD_DIR}/pov/conversion-test.scad)
add_cmdline_test(povexport              EXPERIMENTAL OPENSCAD SUFFIX pov
FILES ${EXPORT_POV_TEST_FILES} ARGS --render --enable=predictible-output)
to the cmakefile under tests.

But I'm not sure what to do next. I know how to run the test
( TEST_GENERATE=1 ctest -R pov ) which runs fine:

running '/home/folkert/t/openscad/build/openscad --info' to generate
sysinfo.txt
QObject::startTimer: Timers can only be used with threads started with
QThread
Post test:/usr/bin/python3
"/home/folkert/t/openscad/tests/test_pretty_print.py"
--builddir=/home/folkert/t/openscad/build
Test project /home/folkert/t/openscad/build
    Start 1712: povexport_conversion-test
1/1 Test #1712: povexport_conversion-test ........   Passed    0.19 sec

But how/why/what? Did I just do the right thing?

kintel:
I think it's because your `tests/data/pov/conversion-test.pov` file is in the wrong location. Remove that one, and add the one generated when running the test with TEST_GENERATE=1.

kintel:
In terms of # of lines of code; it's not a big deal, but it's because you explode everything manually. I think I'd probably write it smth. like this:
```
vector<Vector3f> lightExtremes = {
  bbox.min() - 2*bbox.diagonal(),
  bbox.center(),
  box.max + 2*bbox.diagonal(),
};

for (int z=0;z<2;z++) {
  for (int y=0;y<2;y++) {
    for (int x=0;x<2;x++) {
      output << lightExtremes[x].x() << lightExtremes[y].y() << lightExtremes[z].z();
    }
  }
}
```

folkertvanheusden:
Ok but nothing is generated at all?

kintel:
Oh, the test wasn't added to tests/CMakeLists.txt - I can take a look; adding a new rule can be a bit finicky

kintel:
ok, pushed an example of how to add tests. Added a minimal cube test for now; once we're happy with that, we can add another test that exercises the remaining exporter features (camera position, light position, colors)

src/io/export_pov.cc
```diff
@@ -108,7 +108,7 @@ void export_pov(const std::shared_ptr<const Geometry>& geom, std::ostream& outpu
 
   output << "camera { look_at <" << (min_x + max_x) / 2 << ", " << (min_y + max_y) / 2 << ", " << (min_z + max_z) / 2 << "> "
 	  "location <" << min_x + dx * move_away_factor << ", " << min_y - dy * move_away_factor << ", " << min_z + dz * move_away_factor << "> "
-	  "up <0, 0, 1> right <1, 0, 0> sky <0, 0, 1> rotate <-55, clock * 3, clock + 25> }\n";
+	  "up <0, 0, 1> right <1, 0, 0> sky <0, 0, 1> rotate <-55, clock * 3, clock + 25> right x*image_width/image_height }\n";
```
kintel, line 111: You're writing your formula as an explicit string

kintel:
Resolved the conflict, but the test now fails due to your last commit

folkertvanheusden:
As intended: it tells povray to compensate for the aspect ratio.

kintel:
> As intended: it tells povray to compensate for the aspect ratio.

Wops, it looked like a typo. Could you run `TEST_GENERATE=1 ctest -R pov` and validate that the output is correct and commit the expected file changes?

tests/CMakeLists.txt
```diff
@@ -1137,6 +1139,7 @@ add_cmdline_test(objexport              EXPERIMENTAL OPENSCAD SUFFIX obj FILES $
 if (LIB3MF_FOUND)
 add_cmdline_test(3mfexport              EXPERIMENTAL OPENSCAD SUFFIX 3mf FILES ${EXPORT_3MF_TEST_FILES} ARGS --enable=predictible-output)
 endif()
+add_cmdline_test(povexport              EXPERIMENTAL OPENSCAD SUFFIX pov FILES ${EXPORT_POV_TEST_FILES} ARGS --render --enable=predictible-output)
```
kintel, line 1142: We should add --backend=manifold for color support

src/io/export_pov.cc
```diff
@@ -0,0 +1,114 @@
...
+    for (size_t i=0; i<polygon.size(); i++) {
+      if (i)
+        output << ", ";
+      const auto & x = ps->vertices[polygon[i]].x();
+      const auto & y = ps->vertices[polygon[i]].y();
+      const auto & z = ps->vertices[polygon[i]].z();
+      output << "<" << x << ", " << y << ", " << z << ">";
+    }
+    output << ", <" << ps->vertices[polygon[0]].x() << ", " << ps->vertices[polygon[0]].y() << ", " << ps->vertices[polygon[0]].z() << ">";
+    float r = 1.0, g = 1.0, b = 1.0, a = 0.0;
```
kintel, line 62: Is white a good default, or should we use the color from the current OpenSCAD color scheme?

src/io/export_pov.cc
kintel: This probably still needs addressing.

src/io/export_pov.cc
```diff
@@ -0,0 +1,114 @@
...
+    if (has_color) {
+      auto color_index = ps->color_indices[polygon_index];
+      if (color_index >= 0) {
+        auto color = ps->colors[color_index];
+        r = float(color[0]);
+        g = float(color[1]);
+        b = float(color[2]);
```
kintel, line 69: float cast is superfluous

src/io/export_pov.cc
kintel, line 62: `a` here isn't meant as alpha, probably name it `f` to avoid confusion

src/io/export_pov.cc
```diff
@@ -0,0 +1,114 @@
...
+
+  constexpr double move_away_factor = 2.;
+  std::vector<double> lx { min_x - dx * move_away_factor, bbox.center().x(), max_x + dx * move_away_factor };
+  std::vector<double> ly { min_y - dy * move_away_factor, bbox.center().y(), max_y + dy * move_away_factor };
+  std::vector<double> lz { min_z - dz * move_away_factor, bbox.center().z(), max_z + dz * move_away_factor };
+
+  constexpr float brightness = 0.2;  // 1.0 is way too bright
+
+  for(auto cur_lx: lx) {
+	  for(auto cur_ly: ly) {
```
kintel, line 103: indent to two spaces

src/io/export_pov.cc
```diff
@@ -0,0 +1,114 @@
...
+		  for(auto cur_lz: lz)
+			  output << "light_source { <" << cur_lx << ", " << cur_ly << ", " << cur_lz << "> color rgb <" << brightness << ", " << brightness << ", " << brightness << "> }\n";
+	  }
+  }
+
+  output << "camera { look_at <" << (min_x + max_x) / 2 << ", " << (min_y + max_y) / 2 << ", " << (min_z + max_z) / 2 << "> "
```
kintel, line 109: bbox.center().x() etc. would be easier on the eyes here.

tests/CMakeLists.txt
```diff
@@ -1137,6 +1139,7 @@ add_cmdline_test(objexport              EXPERIMENTAL OPENSCAD SUFFIX obj FILES $
 if (LIB3MF_FOUND)
 add_cmdline_test(3mfexport              EXPERIMENTAL OPENSCAD SUFFIX 3mf FILES ${EXPORT_3MF_TEST_FILES} ARGS --enable=predictible-output)
 endif()
+add_cmdline_test(povexport              EXPERIMENTAL OPENSCAD SUFFIX pov FILES ${EXPORT_POV_TEST_FILES} ARGS --enable=predictible-output --backend=manifold --render)
```
kintel, line 1142: I believe --render is superfluous here

kintel (commented):
Getting close. I think this is my last batch of comments.

folkertvanheusden:
>
> Out of curiosity, why are you using individual triangles, when PovRay has a mesh2 object which allows vertices + face indices, including specifying a texture list and index into it per face?
>
> I guess it's the same functionality, but the mesh2 approach is meant to
> produce more compact files.
>
>
> https://wiki.povray.org/content/Documentation:Tutorial_Section_3.1#Mesh_Object
>
It's because I was not sure if each polygon meets the requirements for a
mesh object.

Also if people like this pov-export feature, we may want to invest time in
exporting /before/ the triangle-phase.



> +    output << ", <" << ps->vertices[polygon[0]].x() << ", " << ps->vertices[polygon[0]].y() << ", " << ps->vertices[polygon[0]].z() << ">";
> +    float r = 1.0, g = 1.0, b = 1.0, a = 0.0;
>
> Is white a good default, or should we use the color from the current
> OpenSCAD color scheme?
>
The OpenSCAD yellow may be even better. I've set it to that now.


> + if (r == 0 && g == 0 && b == 0) // work around for black objects in 5185
>
> This probably still needs addressing.
>
I've let it in but removed the "in 5185" part of the comment. I left it in
because people may choose to create black objects which would not be
visible in povray.
But maybe a warning in the log-window is better?


regards

src/io/export_pov.cc
```diff
@@ -59,21 +59,21 @@ void export_pov(const std::shared_ptr<const Geometry>& geom, std::ostream& outpu
...
-        a = 1.0 - color[3];
+        r = color[0];
+        g = color[1];
+        b = color[2];
+        f = 1.0 - color[3];
       }
     }
-    if (r == 0 && g == 0 && b == 0)  // work around for black objects in 5185
+    if (r == 0 && g == 0 && b == 0)  // work around for black objects
       g = 1.0;
```
kintel, line 74: I'd say black objects are useful if you want an object that absorbs light with some light object in the background. Ideally we'd support real materials, but changing black to green feels like the wrong idea. If black is really universally unwanted in PovRay, a warning makes sense.

nophead:
Black isn't very useful in OpenSCAD ither. I use a very dark grey instead because black nerver reacts to lights. I imagine it is the same in Povray.

folkertvanheusden:
Ok, removed the check & override

kintel (approved):

tpimh:
This is what I've got on the first try, pretty impressive. Are there any tools that can help to adjust lights and camera?

https://github.com/user-attachments/assets/0dbfeb24-9bca-4bb1-9fe2-377c8a5fe1a3

folkertvanheusden:
Hi,

There are sofar no tools other than notepad :-)
If you open the .pov-file, then you see the lights defined at the bottom
(as well as the camera).
You can adjust the parameters there. Oh and at the top there's a define
which describes the "material", also parameters to tweak.

tpimh:
Thanks for your reply. Just editing the file is what I am doing now, but this is not very convenient as there is no immediate preview. I can probably write a script which would extract the "camera" object in OpenSCAD, and patch the exported PovRay file accordingly.

folkertvanheusden:
Yes for now indeed that would work.
I'm working on something that takes the viewport from openscad and uses
that in the output.

tpimh:
This sounds great! Does it mean the result would be similar to what current PNG export looks like?

jordanbrown0:
You might also consider somehow having two files:  one that represents the model, and one that represents "presentation" stuff like camera and lights.  The camera/lights file might be entirely static.  Then the user could edit the camera/lights file by hand, without making it difficult to regenerate the model file when the model changes.

We should also add some general-purpose markup features that would let the OpenSCAD model supply markup that would be interpreted by the exporter.

PR #5299, opened 2024-09-17
Add Send_to_Cura option in PrintInitDialog

gsohler:
My Daily flow is to
1) Improve Model
2) Save STL
3) Load STL in CURA
4) Slice
5) Print

This PR shortcuts 2 and 3. 
It saves the design to a temporary file and loads it in cura. No need anymore to save STL into a "safe" place and find it with cura.
works incredible well in Linux,  I am curious how it does in windows


kintel:
I guess the main challenge is that neither Windows or macOS generally use PATH to locate GUI applications, so we may need a way to determine what to launch. I'm also not 100% sure how QProcess deals with situation where the launching process dies, e.g. can you quit OpenSCAD while Cura is still running after launching with QProcess?

src/gui/MainWindow.cc
```diff
@@ -2195,6 +2199,55 @@ void MainWindow::sendToOctoPrint()
 #endif // ifdef ENABLE_3D_PRINTING
 }
 
+void MainWindow::sendToLocalSlicer(void)
+{
+#ifdef ENABLE_3D_PRINTING
+  const QString slicer = QString::fromStdString(Settings::Settings::localSlicerExecutable.value());
+
+ const QString fileFormat = QString::fromStdString(Settings::Settings::octoPrintFileFormat.value());
```
kintel, line 2207: it uses the octoprint file format for the local slicer feature. Easiest is perhaps to copy this setting into the local slicer section in Preferences. Alternatively make it a common setting, but then we have to deal with file format support filtering for each print service somewhere..

src/gui/Preferences.ui
```diff
@@ -1580,6 +1580,33 @@
...
+             <property name="text">
+              <string>Executable name</string>
+             </property>
+             <property name="alignment">
+              <set>Qt::AlignRight|Qt::AlignTrailing|Qt::AlignVCenter</set>
+             </property>
+            </widget>
+           </item>
+           <item row="0" column="2">
+            <widget class="QLineEdit" name="lineEditLocalSlicer">
```
kintel, line 1601: It's a bit inconvenient to edit the filename manually. Could you add a file dialog here?

src/gui/MainWindow.cc
```diff
@@ -2195,6 +2199,55 @@ void MainWindow::sendToOctoPrint()
...
+  } else {
+    QFileInfo fileInfo{activeEditor->filepath};
+    userFileName = fileInfo.baseName() + fileFormat.toLower();
+  }
+
+  ExportInfo exportInfo = createExportInfo(exportFileFormat, exportFileName, activeEditor->filepath);
+  exportFileByName(this->root_geom, exportInfo);
+
+  QProcess process(this);
+  if(!process.startDetached(slicer, {exportFileName})) {
```
kintel, line 2244: It's a bit annoying that, on macOS, this has to be the path to the executable itself, not to the app bundle. This makes it hard to locate for users not knowledgeable about the app bundle structure on macOS.
If anyone has a good idea or pointer to how this is solved in other Qt-based software, that would be awesome.

src/gui/MainWindow.cc
```diff
@@ -2195,6 +2199,55 @@ void MainWindow::sendToOctoPrint()
...
+    LOG(message_group::Error, "Could not start Slicer.  Is it installed?");
```
kintel, line 2245: This error message is not very informative. Would it be possible to provide more info why it failed? e.g. by using `setStandardErrorFile()` to extract error output?

t-paul: I kept that in using `readAll()` with merged channels, but I don't think it will ever report anything. This will likely happen only in the case where the call actually succeeds and then later collects the output.
Unfortunately the dedicated error message for `startDetached()` failure does not give a very useful error message for the case of a wrong path.

src/gui/Preferences.ui
t-paul: Done

t-paul (approved):

kintel (approved):

gsohler:
thx, its cewl, to have this speedup available to everyone!

PR #5304, opened 2024-09-19
Development: add a script that caches clang-tidy results.

### LINKED ISSUE #5300: Interest in having a script that runs clang-tidy and caches results ?
For various other projects I have written and refined some [script](https://github.com/hzeller/dev-tools/blob/main/run-clang-tidy-cached.cc)  (well, really a self-contained C++ program), that runs `clang-tidy` and caches all the results content-addressed out-of-tree, so that re-runs only have to look at changed files or headers. It provides a combined view of all the findings (which look like compiler-outputs, so it is easy to `cat OpenSCAD_clang-tidy.out` to step through all the findings in an editor), as well as a summary.

I know there is already a `CLANG_TIDY` config in CMake, so this might feel redundant and not interesting ... but I usually find the output of cmake harder to deal with and certainly requires more diligence in making sure the results are not lost or have to all be re-generated (and `clang-tidy` is slooow).

Dropping that script into `scripts/`, changing the configuration struct to the following could make this a useful tool:

```c++
static constexpr ConfigValues kConfig = {
  .start_dir = "src/",
  .file_exclude_re = "src/ext/",
  .revisit_brokenfiles_if_build_config_newer = false,
};
```

It uses all cores to run in parallel, but does not do Unity-builds, so it runs a bit longer as it visits each file. But since it caches the results, only changed files will be considered on re-runs. The cache is outside tree (`~/.cache/clang-tidy/` by default), it will also remember results from unchanged files between different check-outs.

Anyway, didn't want to directly send a PR, wanted to check if this is that would be interesting. If not, feel free to close.

This is how the summary output looks like for OpenSCAD:
```
491 files of interest.
Details: OpenSCAD_clang-tidy.out
Summary: OpenSCAD_clang-tidy.summary
---- Summary ----
 3985 [misc-include-cleaner]
 1153 [misc-const-correctness]
  178 [misc-use-anonymous-namespace]
   81 [clang-diagnostic-error]
   60 [modernize-use-emplace]
   42 [misc-unused-parameters]
   37 [performance-enum-size]
   31 [performance-avoid-endl]
   28 [clang-analyzer-optin.core.EnumCastOutOfRange]
   26 [bugprone-switch-missing-default-case]
   14 [modernize-use-override]
   12 [modernize-macro-to-enum]
   11 [modernize-use-auto]
   10 [performance-unnecessary-value-param]
    9 [misc-header-include-cycle]
    9 [modernize-type-traits]
    8 [bugprone-integer-division]
    7 [bugprone-assignment-in-if-condition]
    7 [clang-analyzer-deadcode.DeadStores]
    7 [modernize-redundant-void-arg]
    7 [modernize-use-equals-default]
    6 [modernize-make-shared]
    5 [bugprone-branch-clone]
    5 [bugprone-empty-catch]
    5 [bugprone-multi-level-implicit-pointer-conversion]
    5 [bugprone-reserved-identifier]
    5 [bugprone-unused-local-non-trivial-variable]
    4 [bugprone-macro-parentheses]
    4 [misc-confusable-identifiers]
    4 [modernize-use-nullptr]
    4 [performance-for-range-copy]
    3 [modernize-return-braced-init-list]
    2 [modernize-loop-convert]
    2 [modernize-use-using]
    1 [bugprone-implicit-widening-of-multiplication-result]
    1 [modernize-use-bool-literals]
    1 [modernize-use-default-member-init]
    1 [performance-type-promotion-in-math-fn]
    1 [performance-unnecessary-copy-initialization]
```

hzeller:
A self-contained c++ program that runs clang-tidy in parallel on all files that changed since last invocation. It also takes changed headers into acccount and re-processes all files that depend on these.

Fully self-contaiend, so it can be invoked as a 'script' as it compiles itself; a simple

works (but also invoking with 'sh -c' if you wish).

The results are cached in ~/.cache/clang-tidy out of tree, so different checkouts or switching between branches
will also only need to process the different files.

The output are two files

  * OpenSCAD_clang-tidy.out : contains the detailed 'compiler warning'-like output (I usually use `cat OpenSCAD_clang-tidy.out` as 'compile command' in my editor to be able to step through it.
  * OpenSCAD_clang-tidy.summary : A summary with all finding counts, grouped by type (see #5300 for an example).

A copy from run-clang-tidy-cached.cc from https://github.com/hzeller/dev-tools with configuration for openscad; so the only difference from that upstream script is the config-section:

```c++
static constexpr ConfigValues kConfig = {
  .cache_prefix = "OpenSCAD_",
  .start_dir = "src/",
  .file_exclude_re = "src/ext/",
  .revisit_brokenfiles_if_build_config_newer = false,  // not clean yet.
};
```

Fixes #5300

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -95,3 +95,4 @@ tests/data/scad/misc/use-tests.scad
 tests/data/svg/viewbox/viewbox_*.svg
 tests/regression/pdfexporttest/centered.pdf
 tests/regression/pdfexporttest/simple-pdf.pdf
+scripts/run-clang-tidy-cached
--- a/scripts/run-clang-tidy-cached.cc
+++ b/scripts/run-clang-tidy-cached.cc
@@ -1,0 +1,543 @@
+#if 0  // Invoke with /bin/sh or simply add executable bit on this file on Unix.
+B=${0%%.cc}; [ "$B" -nt "$0" ] || c++ -std=c++17 -o"$B" "$0" && exec "$B" "$@";
+#endif
+// Copyright 2023 Henner Zeller <h.zeller@acm.org>
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Location: https://github.com/hzeller/dev-tools (2024-09-19)
+
+// Script to run clang-tidy on files in a bazel project while caching the
+// results as clang-tidy can be pretty slow. The clang-tidy output messages
+// are content-addressed in a hash(cc-file-content) cache file.
+// Should run on any system with a shell that provides 2>/dev/null redirect.
+//
+// Invocation without parameters simply uses the .clang-tidy config to run on
+// all *.{cc,h} files. Additional parameters passed to this script are passed
+// to clang-tidy as-is. Typical use could be for instance
+//   run-clang-tidy-cached.cc --checks="-*,modernize-use-override" --fix
+//
+// Note: useful environment variables to configure are
+//  CLANG_TIDY = binary to run; default would just be clang-tidy.
+//  CACHE_DIR  = where to put the cached content; default ~/.cache
+
+// This file shall be c++17 self-contained; not using any re2 or absl niceties.
+#include <unistd.h>
+
+#include <algorithm>
+#include <cerrno>
+#include <cinttypes>
+#include <csignal>
+#include <cstdint>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+#include <filesystem>
+#include <fstream>
+#include <functional>
+#include <iostream>
+#include <list>
+#include <map>
+#include <mutex>
+#include <regex>
+#include <string>
+#include <string_view>
+#include <system_error>
+#include <thread>
+#include <utility>
+#include <vector>
+
+// These are the configuration defaults.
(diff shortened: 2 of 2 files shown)


hzeller:
CC @t-paul since you commented on #5300

hzeller:
Not sure what's up with the Windows msys2 build failure in the CI. But it is certainly unrelated to this PR.

kintel:
Windows CI is flaky. Rerunning it.

kintel:
One Q: As it can take significant time to get to a "clean" state, would it make sense to disable checks in .clang-tidy until we're clean, then gradually enable them as we clean things up? That way we could enable the "clean mode" of your runner earlier.
Main challenge is perhaps to write a readable/maintainable .clang-tidy file

hzeller:
Yes, I think disabling the top-most findings for now could make it easier to get to a clean state.

Top findings currently are:

```
 3985 [misc-include-cleaner]
 1153 [misc-const-correctness]
```

(And next is `178 [misc-use-anonymous-namespace]` which I am actually not necessarily agreeing with, as having a `static` prefix in front of functions is much easier to read than trying to figure out if something is in an anonymous namespace, so I'd permanently disable that rule).

It still possible for developers to manually just run one rule if they want to work on that, which reduces a lot the run-time of clang-tidy as well:

```
scripts/run-clang-tidy-cached.cc --checks="-*,misc-include-cleaner"
```

Once these numbers are low or gone it then a good time to re-add to the `.clang-tidy` file.

hzeller:
Currently, about 460 files contain some sort of warning

```
cat OpenSCAD_clang-tidy.out | egrep "^.*:.*:.*\[[a-z-]+\]" | awk -F: '{print $1}' | sort | uniq | wc -l461
461
```

If the top three are disabled, then only about 130 files are affected

```
cat OpenSCAD_clang-tidy.out | egrep "^.*:.*:.*\[[a-z-]+\]" | grep -v "misc-include-cleaner|misc-const-correct|misc-use-anonymous" | awk -F: '{print $1}' | sort | uniq | wc -l
133
```

And most of the remaining ones can be reduced quickly semi-automatically.

kintel:
sounds good. We can iterate on settings from here

PR #5306, opened 2024-09-20
Fix a bunch of  `[misc-include-cleaner]` clang-tidy findings

hzeller:
Using the clang-tidy runner ( #5304 ), found some missing includes and fixed them half-automated.

This PR adds includes needed for `size_t` (include `<cstddef>`), `std::vector` and `std::string` (including `<vector>` and `<string>`). Separated out as three different commits to not mix things too much.

This reduces `[misc-include-cleaner]` warnings by about 400, about 10%.
  * before: 3985
  * after: 3567

I found that the places how headers are ordered was not necessarily following a common style (e.g. c headers before c++ before project headers), so I put them where things fit 'best'. Longer term, adopting a style there is probably a good idea, but I didn't want to mess _too_ much with that in the first round.

There are a few places where some trailing whitespaces had been found and removed - this is because I've configured my editor to do that by default; they are so few that they hopefully don't mess with ongoing pull requests.

My plan is to chip away on the include cleaner and other clang-tidy findings (if the main contributors welcome these improvements). Rough plan is to first add all the missing includes (because we know then that everything needed is there), so that the superfluous includes can be removed afterwards (as they might be needed right now to accidentally help out other files to work).

### DIFF
--- a/src/Cache.h
+++ b/src/Cache.h
@@ -44,6 +44,7 @@
 
 #pragma once
 
+#include <cstddef>
 #include <unordered_map>
 #include "printutils.h"
 
--- a/src/FontCache.cc
+++ b/src/FontCache.cc
@@ -29,6 +29,7 @@
 
 #include <boost/filesystem.hpp>
 #include <boost/algorithm/string.hpp>
+#include <string>
 #include <utility>
 
 #include "FontCache.h"
--- a/src/FontCache.cc
+++ b/src/FontCache.cc
@@ -25,6 +25,7 @@
  */
 
 #include <iostream>
+#include <vector>
 
 #include <boost/filesystem.hpp>
 #include <boost/algorithm/string.hpp>
--- a/src/LibraryInfo.cc
+++ b/src/LibraryInfo.cc
@@ -1,5 +1,6 @@
 #include "LibraryInfo.h"
 #include <glib.h>
+#include <string>
 #include <vector>
 
 #include "version_check.h"
--- a/src/RenderStatistic.cc
+++ b/src/RenderStatistic.cc
@@ -26,6 +26,7 @@
 
 #include <fstream>
 #include <json.hpp>
+#include <vector>
 
 #include "printutils.h"
 #include "GeometryCache.h"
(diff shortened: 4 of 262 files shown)


pca006132:
@kintel I wonder if we should also adopt some code formatting requirement as well, and we can clean things up in a few PRs.

In addition, if we can encapsulate code interacting with CGAL in a few files, we can probably massively improve compile time...

hzeller:
`clang-format` improved a lot over the last decade, and it indeed works nicely with editors or just the `clangd` language server.

The output it produces is rarely annoying anymore. 

The benefits of not having to walk on egg-shells to modify some existing code to fit it in with the local style in that file **far** outweighs the occasional formatting that does not do what the programmer would decide otherwise (often when formatting tabular array initialization of sorts, but it is easy to wrap them in `// clang-format off / on` comments.

Being able to run the formatter over the whole project that produces no changes is then easy to keep and not back-slide. And it makes it far easier for contributors to confidently provide code that easily fits in.

Anyway, +1 for considering clang-format (but ... that is something for a different thread, not this PR)

kintel (approved):
LGTM. In the future, try to change a single thing per PR if you touch a lot of files. This makes it a lot easier to review.

..or if some automation was used, mention how to reproduce the results.

-> just a guard against human error.

hzeller:
Ack, next time only one commit per PR with one fix.

For reference, this is essentially what I did:

Finding all the entries that talk about a particular symbol not provided by the corresponding header, and writing a file with one line each mentioning the filename and also looking like a 'compiler output' with a line number of 3 (column 1):

```bash
grep "no header providi.*std::vector" OpenSCAD_clang-tidy.out | awk -F: '{print $1}' | sort -u | awk '{print $0 ":3:1:"}' > /tmp/tofix.txt
```

that then looke like something like
```
src/core/builtin_functions.cc:3:1:
src/core/Builtins.cc:3:1:
src/core/Builtins.h:3:1:
src/core/Children.cc:3:1:
...
```
... now, opening the editor and calling compile, but with the 'compile command' (`M-x compile-command` on emacs) being `cat /tmp/tofix.txt`. Now step through each 'error', which opens the right file and places the cursor already roughly positioned at the top of the file (because: line 3) and manually adding the header.

Once we have clang-format, it would be easy to programmatically put the header roughly at top and then clang-format will take care of putting it in the right block.

hzeller:
Can this be merged, as I plan to prepare some follow-up changes this weekend ?

kintel:
Can be merged once all checks are green. Rerunning Windows again..

PR #5307, opened 2024-09-20
Make invocation of bash scripts posix compatible.

hzeller:
The baked-in assumption that bash resides in /bin/bash is not always true, so invoke with /usr/bin/env to be more compatible.


hzeller:
This will in particular make users of NixOS happy.

kintel (approved):

PR #5309, opened 2024-09-22
Add .clang-format configuration file.

hzeller:
Configured, so that it is similar to what uncrustify would produce, but applyig it will still be a massive change, so there needs to be a large scale submit once. The output looks (to my eyes) strictly better than uncrustify, which sometimes does qeustionable

Started in a comment-thread in another PR
https://github.com/openscad/openscad/pull/5306#issuecomment-2363622267

How to apply:
```bash
clang-format -i $(find src -name "*.cc" -o -name "*.h" | grep -v src/ext/)
```

Some comments
  * I did not attempt to fully recreated the uncrustify configuration (is there an automatic tool ?), but mostly looked at the style 'manually' and applied  configuration options that mostly match it.
  * there are numerous options to tweak https://releases.llvm.org/18.1.6/tools/clang/docs/ClangFormatStyleOptions.html
  * While the typical line length is 80 in many projects, here it looked a bit more so I determined a value  determining the 98% percentile of all line length currently found
  `awk '{print length($0)}' $(find src -name "*.cc" -o -name "*.h" | grep -v src/ext/) | datamash perc:98 1`
which resulted in 106. Chosen value 105
  * I was surprised to see that pointers and references are handled differently: References are left-hugging, pointers are right hugging 'const Foo&' vs. `const Foo *`. (Personally, I'd prefer always right-hugging).
  * I don't like braces on a new line for functions, it unnecessarily eats up vertical space, but added the configuration, as it seems to be largely used.
  * Since output differs from uncrustify and `clang-format`, if this is to be added, this PR or a follow-up PR should probably also remove `.uncrustify.cfg` and update `scripts/beautify.sh` to use `clang-format` 

As Draft pull  request for now as there will probably be discussions.
Yes it will create a large change when applied (and it could be rolled-out in stages, e.g. first with often touched files), but using the more modern tool that is more commonly used will make contribution and readability much easier afterwards.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,37 @@
+# Some set of config values, then refined below
+BasedOnStyle: Google
+
+ColumnLimit: 105
+ContinuationIndentWidth: 2
+ConstructorInitializerIndentWidth: 2
+
+# public:, private: in classes: no indent
+AccessModifierOffset: -2
+
+# case on same level as switch
+IndentCaseLabels: false
+AllowShortCaseLabelsOnASingleLine: true
+AlignConsecutiveShortCaseStatements:
+  Enabled: true
+  AcrossEmptyLines: true
+  AcrossComments: true
+  AlignCaseColons: false
+
+# Always align pointer * and reference & to the right.
+DerivePointerAlignment: false
+PointerAlignment: Right
+ReferenceAlignment: Left
+
+# To be enabled later:
+#  *  once all misc-include-cleaner are addressed an no accidental inclusion
+#     order needed.
+#  * Some CGAL inclusion order has been figuured out.
+SortIncludes: false
+
+# Braces on newline in functions, classes and namespaces.
+BreakBeforeBraces: Custom
+BraceWrapping:
+  AfterClass: true
+  AfterFunction: true
+
+AllowShortIfStatementsOnASingleLine: AllIfsAndElse


pca006132:
Maybe post some examples so we can take a look at the result without actually running the formatter? Also, what is the version of clang-format used? I think we should fix a version because clang-format can change from version to version (and unfortunately cannot be configured in the config file...)

kintel:
Generally I want this, but as you say, we should deploy with care.

One idea: rerun beautify today to align the current codebase with the current style, then modify beautify to introduce a partial clang-format run as a post-process.

Then it might be less intrusive to establish a new style, or at least allow us to iterate from today's state.

One challenge is that large scale code changes may make existing open PRs tricky to merge.

I think the main difference between uncrustify and clang-format is that the latter only makes whitespace changes, while uncrustify can modify braces, commas etc.

pca006132:
I think clang-format can modify braces as well: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#insertbraces

kintel:
> I think clang-format can modify braces as well: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#insertbraces

Nice, they've added more features since last I checked :)

hzeller:
to figure out what files can be changed without affecting current pull requests, I have a [git-open-pr-files](https://github.com/hzeller/dev-tools/blob/main/git-open-pr-files) script that can be utilized for that (needs `gh`  github command line and possibly set it up to authenticate to allow hammering the API, and `jq`)

```bash
../dev-tools/git-open-pr-files openscad openscad | grep -v "export_stl copy" | sort > pr-open-files.txt
find src -name "*.h" -o -name "*.cc" | grep -v src/ext | sort > affected-files.txt
join -v1 affected-files.txt pr-open-files.txt > not-currently-in-pr-files.txt
```

Right now, 259 files are not currently in any open PR and could be modified without problem:
[not-currently-in-pr-files.txt](https://github.com/user-attachments/files/17090385/not-currently-in-pr-files.txt)

Maybe that can be done regularly on files currently not open, so we can accumulate a steadily increasing allow list of files that should be handled with clang-format vs. uncrustify (and check that in) and the `beautify.sh` script takes that into acccount until most files are converted.

kintel:
Nice! There are probably a lot of abandoned PRs which we could ignore using an allow- or denylist combined with your script.

..and sometimes just biting the bullet is necessary to avoid spending time we don't have on this :)

pca006132:
I feel like we should probably just close abandoned PRs as they are dead already... If someone still wants to push to merge it, they can reopen it.

gsohler:
Openscad has got a total of over 140 views.
It would not only yield a clear view on the remaining "mind"  of the
community,
Also there would be even more "message" to what ideas are more
"appreciated" than others ..

In every communication, a clear "No" or "Yes" is more contributing than
simple "Not acknowledge"

Also helpful if the openscad owners would sometimes remind authors on pr's
where more collaboration would result in a merge ;)


my 2 cents ....

kintel:
@gsohler Managing external contributions can be tricky and time consuming. Especially feature contributions where the design isn't anchored in the community in advance. Very often, writing the code for a new feature and opening a PR can be a small fraction of the work needed, and unless the contributor has secured the resources to finish it, it's hard for the current team of volunteers to absorb the extra work. But it's a good point that we could be more direct and discourage contribution attempts which are likely to require significant labor which won't be prioritized. Unfortunately, the end result can very quickly end up with an abandoned PR. Perhaps a cleaner approach is to review these PR briefly, then convert them to drafts if more work is needed, and then close them after N weeks of inactivity?
Reminding is best left to bots IMO, unless someone wants to play that role :)

kintel (approved):

gsohler:
@klintel, I did not mean the volunteers/ you to rush/speed up  , sorry if somebody got wrong impression. Basically
I am happy that the message about 'to more directly discourage unviable attempts' was received.

Issue #5308, opened 2024-09-21
Support externally built Manifold


kintel:
We recently added the `-DUSE_BUILTIN_MANIFOLD` CMake flag, defaulting to true, but the support for setting this flag to false is incomplete.

TODO:
* [x] macOS: Build our own Manifold for distribution (macosx-build-dependencies.sh) and build OpenSCAD against it.
   * [x] Full Manifold w/deps build (fat binaries): #5310
   * [x] Allow building with internal manifold after previously installing it (i.e. override the system version). It looks like this is fixed by https://github.com/elalish/manifold/pull/1048 - needs testing after merge.
   * [x] Bonus: Build manifold deps separately -> defer Clipper2 until we upgrade OpenSCAD
   * [x] Bonus: Allow external deps (Clipper2, glm, tbb, onedpl) but builtin Manifold -> resolved by building manifod
   * [x] Q: Should USE_BUILTIN_MANIFOLD ideally override just manifold or also manifold's deps? -> Just manifold. We trust Manifold to do the right thing.
* [x] Linux: Make sure OpenSCAD can build against a distro-packaged Manifold.
   * [x] NixOS
   * [x] Gentoo
   * [x] Q: Are there any other distros packaging a recent manifold? -> unknown, but two is a good start
   * [x] Q: How is Clipper2 managed? -> as a separately packaged library
* [x] Windows: msys2 + MXE -> out of scope for now; we can use submodule.
* [x] Manage Manifold feature requests
   * [x] Use namespaced CMake libraries and targets (e.g. manifold::manifold instead of just manifold): https://github.com/elalish/manifold/issues/936
   * [x] If manifold built and installed its own clipper, a reinstall will not rebuild Clipper, but use the existing version. This is probably not intentional, unless Manifold has a way of querying for minimal required version and rebuild if and old version is installed _and_ it was built by manifold: https://github.com/elalish/manifold/issues/942 -> resolved by linking Clipper statically
   * [x] Unified library target name: https://github.com/elalish/manifold/pull/1048

pca006132:
I think @cjmayo is packaging manifold for gentoo.

kintel:
@cjmayo Thanks! I'll ping you when/if I make changes so we don't break anything.

kintel:
@cjmayo How do you get around the manifold target name differences (`manifold` when using as submodule and `manifold::manifold` when using the installed version via `manifoldConfig.cmake`)?

hzeller:
[...]

PR #5310, opened 2024-09-22
Support building and using externally built manifold on macOS

kintel:
See #5308

* [x] Basic build script
* [x] Resolve https://github.com/elalish/manifold/issues/1039
* [x] Fat binary test
* [x] Wait for https://github.com/elalish/manifold/pull/1048 and update submodule


kintel:
Yeah, but I want OpenSCAD to use Clipper2 as well, so I'm planning to switch those together.

t-paul:
For Debian builds, there is already that: https://build.opensuse.org/project/show/home:t-paul:polyclipping2 (OBS can't do cmake-fetch as the builders have no network access at all).

PR #5312, opened 2024-09-22
Make headers include relative to project-root.

### LINKED ISSUE #5311: Header includes: would be neat if relative to project-root
Right  now, all locally used project includes are included as local files, e.g.

```c++
#include "linalg.h"
#include "CSGNode.h"
#include "GeometryEvaluator.h"
```

Which makes it harder for someone new to the code base to see where things are and 'connect the dots'.

It would be good if this was relative to the project root to provide context at one glance; also reccommended by [styleguides](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes)

```c++
#include "geometry/linalg.h"
#include "core/CsgOpNode.h"
#include "geometry/GeometryEvaluator.h"
```

If this sounds good, I have a script that can convert that.

hzeller:
That is, relative to src/ and src/ext.

Fixes #5311

This was created with a script:

```bash
# Repair some project includes sometimes being included as system header.
sed -i 's/<Reindexer.h>/"Reindexer.h"/' $(find src -name "*.h" -o -name "*.cc")
sed -i 's/<PolySet.h>/"PolySet.h"/' $(find src -name "*.h" -o -name "*.cc")
sed -i 's/<PolySetBuilder.h>/"PolySetBuilder.h"/' $(find src -name "*.h" -o -name "*.cc")

sed -i 's/<json.hpp>/"json.hpp"/' $(find src -name "*.h" -o -name "*.cc")
sed -i 's|>FontListTableView.h<|>gui/FontListTableView.h<|' src/gui/FontList.ui


# Some are almost there, but not fully
sed -i 's|#include "input/|#include "|' $(find src -name "*.h" -o -name "*.cc")

# Find all the simple files that can be represented with their path relative
# to project root
for f in $(find src -name "*.h" -o -name "*.hpp" | grep -v OffscreenContextEGL.h | grep -v OffscreenContextGLX.h); do echo "s|\(#include \"\)\($(basename $f)\)\"|\1$(echo $f | sed 's|src/\(ext/\)\?||')\"|;"  ; done > sed.script

sed -e "$(cat sed.script)" -i $(find src -name "*.h" -o -name "*.cc" -o -name "*.y" -o -name "*.l" | grep -v ext/src/)
```

Note, the two headers for `OffscreenContextEGL.h` and `OffscreenContextGLX.h` were not done with this script as there are two headers with the same name in different directories. These were fixed-up manually to avoid mistakes.

The CMakeLists file still contains all the header paths, but they are commented in a separate block and should be removed soon. This is more out of an abundance of caution, I don't think there will be too much of an issue. But might anyway be good to have a small separate commit that just removes these.


kintel (commented):
We should try to avoid editing files under src/ext and treat these as we do submodules. ..or even convert them to submodules or external libraries down the road.

hzeller:
Ups, sorry, missed that. Undid the changes in `src/ext/...`

kintel (commented):
One suggestion: use pathless includes for interface headers belonging to a .cc file.
Right now, this is more of a gut feeling; gotta think about it some more to have a stronger opinion..

This should also be the first header to be included, but that's a different type of change.

hzeller:
Yes it should be the first header to be included; clang-format can fix that, but we can't use it for that yet, as right now a lot of things are order dependent.
I do feel (somewhat strongly) that even headers needed for the corresponding *.cc file should have the full relative path. This makes it less ambiguous and confusing (less 'exceptions' to the rule).

hzeller:
Another reason of course is that it makes it easier to automatically process by tools if there is no ambiguity what header is meant.

kintel (commented):
I think we should let libsvg includes be references as `#include "libsvh/*.h"`, as that's really a separate library. Perhaps we should move that library directly under src.

@t-paul Since you wrote this library, any opinions on this?

src/io/import_json.cc
```diff
@@ -26,11 +26,11 @@
 
 #include <fstream>
 #include <string>
-#include <json.hpp>
+#include "json/json.hpp"
```
kintel, line 29: As  this is an external library, should we retain angle brackets to signal that it's not really part of the OpenSCAD codebase?

src/io/imageutils-lodepng.cc
```diff
@@ -1,4 +1,4 @@
-#include "imageutils.h"
+#include "io/imageutils.h"
 #include "ext/lodepng/lodepng.h"
```
kintel, line 2: Some places, "ext/" is included, other places not. Would be good to choose one style.
..plus the <> vs. "" discussion.

src/io/export_stl.cc
```diff
@@ -24,19 +24,19 @@
  *
  */
 
-#include "export.h"
-#include "PolySet.h"
-#include "PolySetUtils.h"
+#include "io/export.h"
+#include "geometry/PolySet.h"
+#include "geometry/PolySetUtils.h"
 #include "double-conversion/double-conversion.h"
```
kintel, line 30: Use angle brackets?

src/gui/PrintInitDialog.cc
```diff
@@ -26,9 +26,9 @@
 
 #include <QString>
 
-#include "Settings.h"
-#include "PrintService.h"
-#include "PrintInitDialog.h"
+#include "gui/Settings.h"
+#include "gui/PrintService.h"
+#include "gui/PrintInitDialog.h"
 #include "QSettingsCached.h"
```
kintel, line 32: gui/QSettingsCached.h

src/gui/Preferences.cc
```diff
@@ -37,19 +37,19 @@
 #include <QSettings>
 #include <QTextDocument>
 #include <boost/algorithm/string.hpp>
-#include "GeometryCache.h"
-#include "AutoUpdater.h"
+#include "geometry/GeometryCache.h"
+#include "gui/AutoUpdater.h"
 #include "Feature.h"
 #include "Settings.h"
```
kintel, line 43: gui/Settings.h

src/gui/Editor.h
```diff
@@ -6,7 +6,7 @@
 #include <QWheelEvent>
 #include <QScrollBar>
 #include <QTextEdit>
-#include "IndicatorData.h"
+#include "core/IndicatorData.h"
 #include "parameter/ParameterWidget.h"
```
kintel, line 10: gui/parameter/ParameterWidget.h

src/glview/OffscreenContextFactory.cc
```diff
@@ -1,26 +1,26 @@
...
-#include "printutils.h"
+#include "utils/printutils.h"
 
 #ifdef __APPLE__
 #include "offscreen-old/OffscreenContextNSOpenGL.h"
-#include "OffscreenContextCGL.h"
+#include "glview/OffscreenContextCGL.h"
 #endif
 #ifdef _WIN32
 #include "offscreen-old/OffscreenContextWGL.h"
```
kintel, line 12: glview/offscreen-old/OffscreenContextWGL.h

src/glview/OffscreenContextFactory.cc
```diff
@@ -1,26 +1,26 @@
-#include "OffscreenContextFactory.h"
+#include "glview/OffscreenContextFactory.h"
 
 #include <string>
 
-#include "printutils.h"
+#include "utils/printutils.h"
 
 #ifdef __APPLE__
 #include "offscreen-old/OffscreenContextNSOpenGL.h"
```
kintel, line 8: glview/offscreen-old/OffscreenContextNSOpenGL.h

src/geometry/manifold/ManifoldGeometry.h
```diff
@@ -1,9 +1,9 @@
 // Portions of this file are Copyright 2023 Google LLC, and licensed under GPL2+. See COPYING.
 #pragma once
 
-#include "Geometry.h"
+#include "geometry/Geometry.h"
 #include <glm/glm.hpp>
-#include "linalg.h"
+#include "geometry/linalg.h"
 #include "manifold/manifold.h"
```
kintel, line 7: We should probably use angle brackets for manifold too

src/geometry/manifold/ManifoldGeometry.cc
```diff
@@ -1,19 +1,19 @@
...
-#include "PolySetBuilder.h"
-#include "PolySetUtils.h"
-#include "manifoldutils.h"
-#include "ColorMap.h"
+#include "geometry/PolySet.h"
+#include "geometry/PolySetBuilder.h"
+#include "geometry/PolySetUtils.h"
+#include "geometry/manifold/manifoldutils.h"
+#include "glview/ColorMap.h"
 #include "src/glview/RenderSettings.h"
```
kintel, line 11: Just glview/RenderSettings.h ?

src/core/Value.cc
```diff
@@ -33,11 +33,11 @@
...
-#include "StackCheck.h"
-#include "boost-utils.h"
+#include "core/Value.h"
+#include "core/EvaluationSession.h"
+#include "utils/printutils.h"
+#include "utils/StackCheck.h"
+#include "utils/boost-utils.h"
 #include "double-conversion/double-conversion.h"
 #include "double-conversion/utils.h"
 #include "double-conversion/ieee.h"
```
kintel, line 43: angle brackets for double-conversion

src/io/import_json.cc
hzeller: Right now, we get it from `src/ext` so we should have it with double-quotes until we move it to, say, submodules and really consider it external based on cmake configuration.

Otherwise there is an ambiguity for the compiler where to pick the file.

If we put it in angle brackets, it should be `#include <nlohmann/json.hpp>` as that is the commonly used name on systems.

src/io/imageutils-lodepng.cc
hzeller: Oh, I had only modified single `"foo.h"` type headers, haven't seen that there are so many `"somtimespartial-or-even-excessive-prefix/foo.h"` headers. Fixed now.

Same here: as long as we have it in src/ext and should be included from there always, they should be quote. To get them angle-bracketified, we need to make a choice in cmake and either pull the system header or from submodules/.

src/io/export_stl.cc
hzeller: Sounds good.

src/gui/PrintInitDialog.cc
hzeller: I wonder why that was not picked up properly by the script before. Fixed.

src/gui/Preferences.cc
hzeller: Done.

src/gui/Editor.h
hzeller: done.

src/glview/OffscreenContextFactory.cc
hzeller: done.

src/glview/OffscreenContextFactory.cc
hzeller: done.

src/geometry/manifold/ManifoldGeometry.h
hzeller: Agreed. Done.

src/geometry/manifold/ManifoldGeometry.cc
hzeller: ah, that also got missed by the script that wasn't expecting that prefix.

src/core/Value.cc
hzeller: Done. Fixed all the double-conversion.

hzeller:
Updated script

```bash
# Repair some project includes sometimes being included as system header.
sed -i 's/<Reindexer.h>/"Reindexer.h"/' $(find src -name "*.h" -o -name "*.cc")
sed -i 's/<PolySet.h>/"PolySet.h"/' $(find src -name "*.h" -o -name "*.cc")
sed -i 's/<PolySetBuilder.h>/"PolySetBuilder.h"/' $(find src -name "*.h" -o -name "*.cc")

# json is imported in this project src/ext, so needs to be quote-included.
sed -i 's/<json.hpp>/"json.hpp"/' $(find src -name "*.h" -o -name "*.cc")

# Fix names that had been quote-included but should be in angle brackets.
sed -i 's|"GL/gl.h"|<GL/gl.h>|' src/glview/OffscreenContextEGL.cc
sed -i 's|"manifold/\([^"]\+\)"|<manifold/\1>|' $(find src -name "*.h" -o -name "*.cc")
sed -i 's|"double-conversion/\([^"]\+\)"|<double-conversion/\1>|' $(find src -name "*.h" -o -name "*.cc")

sed -i 's|>FontListTableView.h<|>gui/FontListTableView.h<|' src/gui/FontList.ui


# Some are almost there, but not fully
sed -i 's|#include "input/|#include "|' $(find src -name "*.h" -o -name "*.cc")

# Find all the simple files that can be represented with their path relative
# to project root
for f in $(find src -name "*.h" -o -name "*.hpp" \
	    | grep -v OffscreenContextEGL.h \
	    | grep -v OffscreenContextGLX.h); do
    echo "s|\(#include \"\)\(.*/\)\?$(basename $f)\"|\1$(echo $f | sed 's|src/\(ext/\)\?||')\"|;"
done > sed.script

sed -e "$(cat sed.script)" -i $(find src -name "*.h" -o -name "*.cc" -o -name "*.y" -o -name "*.l" | grep -v src/ext)

# It's in src/ext, but this needs to be fixed
sed 's|"system-gl.h"|"glview/system-gl.h"|' -i src/ext/CGAL/OGL_helper.h
```

src/ext/CGAL/OGL_helper.h
```diff
@@ -25,7 +25,7 @@
 #include <CGAL/Simple_cartesian.h>
 #include <CGAL/Nef_3/SNC_decorator.h>
 
-#include "system-gl.h"
+#include "glview/system-gl.h"
```
hzeller, line 28: Note, this is the only place in src/ext that needed to be updated, but this is an explicitly [manually added OpenSCAD header](https://github.com/openscad/openscad/blame/f23b1525b9cc4ea3c7c654bff898392d9ef68b45/src/ext/CGAL/OGL_helper.h#L28) anyway.

src/glview/OffscreenContextEGL.cc
```diff
@@ -11,7 +11,7 @@
 #include "utils/printutils.h"
 #define GLAD_EGL_IMPLEMENTATION
 #include "glad/egl.h"
-#include "GL/gl.h"
+#include <GL/gl.h>
```
hzeller, line 14: That was also a header that needed angle brackets.

src/io/imageutils-lodepng.cc
kintel: ok, let's follow your suggestion,

t-paul:
> I think we should let libsvg includes be referenced as `#include "libsvg/*.h"`

Yes, I think we should keep it slightly separated. Also in reference to the license, a separate directory is useful.

I don't mind moving it to some other place, I don't have any preference on that.

hzeller:
if it is something that should be considered separate from the OpenSCAD, then moving it to `src/ext/` sounds like a good choice of home (that would be out of scope for this PR though)

t-paul:
No, I think  `src/ext/` should be only code that lives originally somewhere else so it needs to be updated from time to time. `libsvg` is *only* hosted inside the OpenSCAD source tree.

kintel:
Moving libsvg to `src/` makes sense. Alternatively a new `src/libs/` and use that as it's own header path root.

hzeller:
> Moving libsvg to `src/` makes sense. Alternatively a new `src/libs/` and use that as it's own header path root.

Though that should probably be a separate PR.

kintel:
> > Moving libsvg to `src/` makes sense. Alternatively a new `src/libs/` and use that as it's own header path root.
> 
> Though that should probably be a separate PR.

Agreed that moving stuff belongs elsewhere, but in the interim, could we just revert the include path changes involving libsvg? ..as least references inside libsvg/ itself.

hzeller:
Alright, including libsvg paths now witout `io/`-prefix.

kintel (approved):

hzeller:
Looks like all tests pass.

PR #5314, opened 2024-09-23
Change (auto)reload to use render instead of preview mode

3sv:
Proposal that swaps the (auto)reload function to use rendering instead of preview mode. 

Given how fast rendering works with manifold enabled this might be a better default once manifold is out of testing?

In any case allows me already to use external editor and have the rendering updated automatically with much better quality by using auto-rendering instead of auto-preview.  (external editor as described here,  https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/Using_an_external_Editor_with_OpenSCAD)


nophead:
No thanks. I want a preview not a render. In my code preview shows
an assembly view and render shows printer parts laid out for printing.
Perhaps it could just repeat the last F5 or F6 could be repeated to make it
optional or two menu options.

nophead:
I use $preview to determine what to draw in F5 / F6. I very rarely use F6 as most of my STLs are generated from the command line.

t-paul:
Repeating the last action could get a bit confusing with multiple windows. A 3-state selection defining the action to take after reload (none, preview, render) in the specific window might be a bit more obvious and also easier to discover for new users.

nophead:
I never open multiple windows but presumably it currently works on the one in focus. It isn't confusing if it repeats the last operation on that window. What is more confusing is you do an F6 and then change the code and it does F5 and you have to repeat the F6 to get the same view before the code update.

kintel:
Based on initial feedback, I suggest shelving this PR and open a new one if we can find a solution that fit better with people's use-cases. To discuss, it might be more productive to either open a Feature Request and discuss there, or take it to the mailing list/IRC.

mjparme:
I have always wished I could just have it render instead of preview after making external changes.  There are definitely some designs where it wouldn't perform good enough even with manifold (e.g. adding knurling to a cylinder) so making it configurable via a menu option does seem wise.

3sv:
Proposal based on feedback:
Reload and view mode under Design, choose between Preview/Render

Made further changes so that this selection is also used when "auto reloading" by either the customizer or the animations. 
Changed label under Customizer accordingly.

Not sure if not too much is repeated when doing the animation render, but seems for my usage at least. Had to disable a notification, as it added multiple times in the status bar. It seems the notification is not updated in time. Didn't found out how that piece of code is supposed to work. Looks like the status indicator is lagging behind of the real work in normal cases as well, when repeating fast from the animation view this resulted in status bar getting too wide. Disabled those notifications for now as they are not really useful with new renderer anyway? 

UI:
image

jordanbrown0:
The menus are awfully cluttered.  Is this a mode that one is likely to change at all frequently?  If not, Preferences might be a better place.

t-paul:
Maybe combine the two by adding mode == off?

nophead:
I would much prefer it just remembered the last F5 or F6 on each window and repeated it. So it never switches the type of view unless you do a manual F5 or F6, so no extra menus needed.

It has never made sense that an external change in an editor changed the view mode.

t-paul:
I would not mind if it repeats the last action, it would need to be per window then. But in that case, what's the initial choice, if no F5 or F6 was done yet?

nophead:
I think it should default to F5 as it is safest and what it has always done in the past.

t-paul:
Mostly safe yes, but it will force people who want to use F6 externally to always manually run F6 first. I'm not using external editors, so I can't really judge how annoying that would be.

nophead:
Perhaps the best solution is to do what the option says (F5 or F6 ) and the option defaults to F5 but if you do a manual F5 or F6 it then repeats that and overrides the option.

nophead:
Which is another way of saying repeat the last user action on that window but if they have never done a manual F5 or F6 then use the option to decide.

nomike:
I like the idea of keeping the previous state (F5 or F6).
From a user interface perspective I envision this: 

There are the following preference settings:
render-mode-follows-last-render ∈ (t f) ;; false means: keep the current behavior
initial-render-mode ∈ ('preview 'final) ; dictates which mode is used on a new window.

There is also a command line option which can override initial-render-mode. So if someone calls OpenSCAD from a script, they can set this if they want and it will not break their workflow. 

The window itself needs to remember which mode was used the last time and maybe should indicate this somewhere in the status bar so users don't have to guess what will happen next. 

What's missing then would be a menu option and a key-binding to repeat the last render. 

I guess we're short on function keys already, but maybe something like C-F5 if that's not already used elsewhere. 

What are your thoughts on this?

jordanbrown0:
For interactive use, I don't understand the need for any cleverness.  If you like F5 Preview, hit F5.  If you like F6 Render, hit F6.

For command-line use, mostly similar.  If you ask for a PNG, you get Preview, if you ask for an STL, you get Render.  (And if you really want a rendered PNG, there's --render.)

The only gaps that I see are Design/Automatic Reload and Preview and Design/Automatic Reload and Preview.  I'd expect them to be switched between Preview and Render with a preference.

jordanbrown0:
In the last paragraph in the previous comment, it should be Design/Automatic Reload and Preview and Design/Reload and Preview (no "Automatic" in the second one).

nomike:
> The only gaps that I see are Design/Automatic Reload and Preview and Design/Reload and Preview. I'd expect them to be switched between Preview and Render with a preference.

Fair enough, you're right.

typetetris:
F6 doesn't reload the file though. Why not add a Reload and Render action?

PR #5323, opened 2024-09-25
Insert include <utility> missing in files using std::move

hzeller:
** Note: stacked on top of #5322 so merge this PR after that is merged **

## Automatically generated

```
../dev-tools/insert-header.cc '<utility>' $(grep "no header providing.*misc-include-cleaner" OpenSCAD_clang-tidy.out | awk -F: '/^src\/.*(std::move)/ {print $1}' | sort -u)
```

(w/ insert-header.cc from this [script collection](https://github.com/hzeller/dev-tools))

### DIFF
--- a/src/core/CSGNode.h
+++ b/src/core/CSGNode.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <utility>
 #include <cstddef>
 #include <string>
 #include <vector>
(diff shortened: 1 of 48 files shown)


kintel (approved):

hzeller:
mmh, should'nt github be able to figure out automatically that these come from a stacked set of branches and thus there is not actually a conflict ?

kintel:
Yeah, you'd think so, but I guess the merge commit has a different commit message and the hash is thus different, so I guess it gets confused. `git rebase HEAD~ --onto origin/master` is my goto if this happens and it's just a single commit.

PR #5327, opened 2024-09-26
In implementation compilation unit: move corresponding header to top.

hzeller:
Move header foo.h within implementation foo.cc to the top. That ensures that the header at least is self-contained.

This revealed, that the core/SurfaceNode.h header was missing an include of core/Value.h.

This PR leaves out changing src/glview, as there are too many header dependencies that break, and should be addressed separately.

### DIFF
--- a/src/Feature.cc
+++ b/src/Feature.cc
@@ -1,3 +1,5 @@
+#include "Feature.h"
+
 #include <cstdio>
 #include <iostream>
 #include <sstream>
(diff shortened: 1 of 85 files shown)


hzeller:
This looks like it revealed an issue defining the `_USE_MATH_DEFINES` directly in the source files, as now things are order-dependent. I'll make a separate PR first to move that definition to the cmake file. Marking this as draft for now.

hzeller:
Alright, rebased after #5328, making the Circle CI happy.

kintel (approved):

PR #5328, opened 2024-09-26
Don't #define _USE_MATH_DEFINES in *.cc files but as compile define.

hzeller:
Setting the #define in the compilation unit makes it dependent if the same header had been included before; if <cmath> has been included before (and we don't know that, as we don't know what system headers do between each other), the include-guard will prevent our definition to ever been seen.

So the only way to make sure <cmath> sees the _USE_MATH_DEFINES define is to set it on the compile command line, i.e. in the CMakeLists.txt

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -198,7 +198,11 @@ file(MAKE_DIRECTORY ${OPENSCAD_LIB_OUTPUT_DIR})
 
 target_compile_options(OpenSCAD PRIVATE "$<$<CONFIG:DEBUG>:-DDEBUG>")
 
-target_compile_definitions(OpenSCAD PRIVATE _REENTRANT UNICODE _UNICODE)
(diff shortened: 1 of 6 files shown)


hzeller:
The Circle CI in #5327 failed due to a new header ordering.

CMakeLists.txt
```diff
@@ -727,6 +727,9 @@ elseif(WIN32)
   endif()
 endif()
 
+# We want symbols such as M_PI visible.
+add_compile_definitions(_USE_MATH_DEFINES)
```
kintel, line 731: We use target_compile_definitions() in all other places. Perhaps do that here as well, and move this to ~line 200 where other global definitions are added.

hzeller: Done. Since I looked for `add_compile_definitions`, I didn't see the `target_compile_definitions`

kintel (approved):

PR #5331, opened 2024-09-26
Store Color information along with 3MF Files

gsohler:
We have all color information available. 
We should also store it in 3mf files when exporting

### DIFF
--- a/src/io/export_3mf_v2.cc
+++ b/src/io/export_3mf_v2.cc
@@ -121,6 +121,26 @@ static bool append_polyset(std::shared_ptr<const PolySet> ps, Lib3MF::PWrapper& 
       }
     }
 
+    Lib3MF::PBaseMaterialGroup pBaseMaterial = model->AddBaseMaterialGroup();
+
+    std::vector<Lib3MF_uint32> ids;
+    for(int i=0;i<out_ps->colors.size();i++) {
+      sLib3MFColor col;
+      std::string colname="Color" + i;
+      Lib3MF_uint32 nBaseMaterialID = pBaseMaterial->AddMaterial(colname.c_str(), wrapper->FloatRGBAToColor(
+		out_ps->colors[i][0], out_ps->colors[i][1], out_ps->colors[i][2], out_ps->colors[i][3]));
+      ids.push_back(nBaseMaterialID);
+    }
+    if(ids.size() > 0) mesh->SetObjectLevelProperty(pBaseMaterial->GetResourceID(), ids[0]);
+    sLib3MFTriangleProperties tri_prop;
+    tri_prop.m_ResourceID=pBaseMaterial->GetResourceID();
+    for(int i=0;i<out_ps->color_indices.size();i++) {
+      tri_prop.m_PropertyIDs[0]=ids[out_ps->color_indices[i]];
+      tri_prop.m_PropertyIDs[1]=ids[out_ps->color_indices[i]];
+      tri_prop.m_PropertyIDs[2]=ids[out_ps->color_indices[i]];
+      mesh->SetTriangleProperties(i,tri_prop);
+    }
+
     Lib3MF::PBuildItem builditem;
     try {
       model->AddBuildItem(mesh.get(), wrapper->GetIdentityTransform());
--- a/src/io/export_3mf_v2.cc
+++ b/src/io/export_3mf_v2.cc
@@ -125,8 +125,7 @@ static bool append_polyset(std::shared_ptr<const PolySet> ps, Lib3MF::PWrapper& 
 
     std::vector<Lib3MF_uint32> ids;
     for(int i=0;i<out_ps->colors.size();i++) {
-      sLib3MFColor col;
-      std::string colname="Color" + i;
+      std::string colname=std::string("Color") + std::to_string(i);
       Lib3MF_uint32 nBaseMaterialID = pBaseMaterial->AddMaterial(colname.c_str(), wrapper->FloatRGBAToColor(
 		out_ps->colors[i][0], out_ps->colors[i][1], out_ps->colors[i][2], out_ps->colors[i][3]));
       ids.push_back(nBaseMaterialID);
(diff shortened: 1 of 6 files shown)


src/io/export_3mf_v2.cc
```diff
@@ -121,6 +121,26 @@ static bool append_polyset(std::shared_ptr<const PolySet> ps, Lib3MF::PWrapper&
       }
     }
 
+    Lib3MF::PBaseMaterialGroup pBaseMaterial = model->AddBaseMaterialGroup();
+
+    std::vector<Lib3MF_uint32> ids;
+    for(int i=0;i<out_ps->colors.size();i++) {
+      sLib3MFColor col;
+      std::string colname="Color" + i;
```
t-paul, line 129: That does not work as it looks like as it does pointer arithmetic before assignment.

int i = 3; std::string colname="Color" + i;
Result of that is colname = "or"!

kintel (commented):
Could you add a test for this, similar to the existing 3mf test, but with a minimal file which contains color information?
i.e. similar to https://github.com/openscad/openscad/blob/master/tests/regression/3mfexport/3mf-export-expected.3mf

src/io/export_3mf_v2.cc
gsohler: good catch. It works well up to maximum of 6 colors then segfaults ...

src/io/export_3mf_v2.cc
```diff
@@ -121,6 +121,26 @@ static bool append_polyset(std::shared_ptr<const PolySet> ps, Lib3MF::PWrapper&
...
+		out_ps->colors[i][0], out_ps->colors[i][1], out_ps->colors[i][2], out_ps->colors[i][3]));
+      ids.push_back(nBaseMaterialID);
+    }
+    if(ids.size() > 0) mesh->SetObjectLevelProperty(pBaseMaterial->GetResourceID(), ids[0]);
+    sLib3MFTriangleProperties tri_prop;
+    tri_prop.m_ResourceID=pBaseMaterial->GetResourceID();
+    for(int i=0;i<out_ps->color_indices.size();i++) {
+      tri_prop.m_PropertyIDs[0]=ids[out_ps->color_indices[i]];
+      tri_prop.m_PropertyIDs[1]=ids[out_ps->color_indices[i]];
+      tri_prop.m_PropertyIDs[2]=ids[out_ps->color_indices[i]];
```
t-paul, line 140: `PolySet.h`mentions that the color index can be -1, I think in that case we need to skip adding the property.

> https://github.com/openscad/openscad/blob/master/src/geometry/PolySet.h#L24

kintel: Another reason to add a test :)

kintel (commented):
Btw. how hard would it be to implement color import support for 3mf? That way we could get color testing "for free".

gsohler:
Yeahh, i considered that, too in the past.
I was trying with v2 and i failed using 3mf function to iterate over the existing materials in the 3mf file.
maybe somebody else got more luck

kintel:
Is this still WIP? I cannot see any color info in the output files..

gsohler:
No, with commit from 21:39 I was able to see colors in  an openSCAD created 3MF file with an online 3mf viewer

kintel:
oh, you're doing v1 first - I was only testing with v2..

gsohler:
no, both of them are implemented.
I just realized that ctest is working with the coloring disabled as i need to turn manifold for that during ctest

kintel:
> Yeahh, i considered that, too in the past. I was trying with v2 and i failed using 3mf function to iterate over the existing materials in the 3mf file. maybe somebody else got more luck

It looks relatively straight-forward:
```
        Lib3MF::sTriangleProperties triangleProperties;
        object->GetTriangleProperties(idx, triangleProperties);
        PBaseMaterialGroup baseMaterialGroup = model.GetBaseMaterialGroupByID(triangleProperties.m_ResourceID);
        Lib3MF::sColor color = baseMaterialGroup.GetDisplayColor(triangleProperties.m_PropertyIDs[0]);
```

...but I have to look into it in more detail. I can try when I find some time..

t-paul:
> It looks relatively straight-forward

I think it's more complicated. The properties can be of different type and instead of material/displaycolor we probably want to first go with (single-)color or support both color and material.
I'm trying with V1, but that should work the same with V2 too, just slightly different API.

kintel:
> > It looks relatively straight-forward
> 
> I think it's more complicated. The properties can be of different type and instead of material/displaycolor we probably want to first go with (single-)color or support both color and material. I'm trying with V1, but that should work the same with V2 too, just slightly different API.

Right, I was focusing on being able to read what we write, but the superset of that can be tricky

kintel:
Is this still stuck on implementing import of colors?

gsohler:
Hi @kintel ,
I believe the code is ok. Issue is that when I update the ctest expected files my local 3mf library has some slightly different output as the 3mflib on the CI server.
Suggest o manually test it and then then update the regression files accordingly

t-paul:
I have an extended version for V1 handling both colors and materials in 3MF but that's not checked in yet. I can see if that can be checked in at least.

gsohler:
Torsten, sorry for that,
Maybe I missed some updates ? I am happy to implement when you
resend/re-push
After all, today in the morning, I failed to checkout my color-3mf branch
ven though i can clearly see it in my github page.
(probably i could accept some assistance there ...)

kintel:
I think finishing the import before merging could save us significant work testing and validating the export, and it would likely be a release blocker anyway. Happy to pick up the V2 story if work is needed there. Ideally we should deprecate V1, but that might be tricky unless we submodule tbe library..

gsohler:
I think we need both V1 and V2.

V2 is default  version for linux whereas V1 is default version for MXE.

kintel:
We control our own MXE branch, so the only blocker there is taking the time to upgrade it.

kintel:
If MXE is the only remaining V1 usage, it seems significantly cheaper to kill it and upgrade MXE

PR #5343, opened 2024-09-30
Hide fast-csg

kintel:
Easy way to disable fast-csg for now. Unless someone complains loudly, let's kill it later on.


kintel:
@ochafik

hairyhenderson:
@kintel is there context for this removal? I just noticed the feature is missing in the latest dev snapshot and have [a model](https://github.com/smkent/monoscad/blob/main/gridfinity/rugged-box/rugged-box-gridfinity.scad) that's rendering quite slowly, where it was much faster with fast-csg previously...

PR #5357, opened 2024-10-06
Add support for multiple 3D print services

kintel:
We currently only support having one 3D print service available. This adds support for offering multiple service from the downloaded JSON file, and populates the UI accordingly.

TODO:
* [x] Auto-populate settings and read/write on change
* [x] Auto-populate print dialog
* [x] Connect settings and dialog and hook up the 3D print action
* [x] (optional) Collect all 3 types of services (octoprint, local slicer, external service) under the same interface and move most of the code out of MainWindow
* [x] Add better way of selecting file format to use
* [x] For print services, supported file formats should probably be part of the downloaded JSON file
* [x] (optional) Formalize network protocol -> will defer
* Manual testing
   * [x] Test behavior of current stable release
   * [x] All prefs can be loaded from settings
   * [x] All print services dialog UI are populated from settings
   * [x] File format choices in service dialog and preferences are the same
   * [x] All print services can be remembered and persisted to settings
   * [x] All settings work immediately and after restart
   * [x] Selecting one print service from prefs or dialog doesn't delete the others from the settings file
   * [x] settings file use file format identifiers (lowercase)
   * [x] optional: Settings supports reading upper case for backwards compatibility, and "stl" as an alias for "asciistl" -> will defer
   * [x] Test no print services available -> warning printed
   * [x] Test unable to download print service JSON
   * [x] Test file format support per service


src/gui/PrintService.cc
```diff
@@ -58,7 +58,7 @@ createPrintServices() {
   // TODO: Where to call this, will we need a mutex?
   try {
     auto networkRequest = NetworkRequest<void>{
-        QUrl{"http://localhost:8000/print-service.json"}, {200}, 30};
+        QUrl{"https://files.openscad.org/print-service.json"}, {200}, 30};
```
t-paul, line 61: Just an idea, should we give it a dedicated alias name? In case we need/want to move it to a different location?

kintel: Good idea, any suggestion for URL? app.openscad.org? We need SSL for that domain though, but that should be easy to set up using the existing SSL stuff.

t-paul: Multi-domain handling for SSL is fully automatic via https://github.com/nginx-proxy/acme-companion. It's trivial to add either a dedicated nginx container or just point this to files.openscad.org and let it generate a certificate for both domains. Well, I'm sure about the first solution, 2nd according to documentation :grin:.

t-paul: `app.openscad.org` is live as alias for `files.openscad.org`.

kintel: awesome!
I guess I should switch the macOS autoupdater as well

PR #5365, opened 2024-10-09
Add missing headers in export_pov.

hzeller:
Also, replace one 'auto' variable with explicit type, so that clang-tidy does not consider PolySet an unneeded header.

### DIFF
--- a/src/io/export_pov.cc
+++ b/src/io/export_pov.cc
@@ -26,13 +26,19 @@
 
 #include "io/export.h"
 
+#include <cstddef>
+#include <memory>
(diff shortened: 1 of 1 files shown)


nophead:
Pardon my ignorance but how does any C++ compile if there are missing headers?

kintel:
The C++ preprocessor leaks symbols from indirectly included headers. For an overview over why that's suboptimal, see e.g. https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md

PR #5366, opened 2024-10-10
Remove std::-headers in files not needing them

hzeller:
There were a few files that included headers they didn't need.
These can be removed now that all the _necessary_ `std::`-headers are included (after a previous set of pull requests), so we know that there are no headers left that accidentally provide symbols in other files - all headers not needed directly in a header or *.cc file can be removed now.

One commit per include.

Auto-generated with

```bash
./scripts/run-clang-tidy-cached.cc
grep "not used directly" OpenSCAD_clang-tidy.out  | awk '{print $5 " " $1}' | egrep  "^(sstream|fstream|algorithm|memory|map|vector|array|chrono|ciso646|cstdint|forward_list|functional|streambuf|string|thread|tuple|unordered_set|utility|variant|stack|set|ostream|queue|numeric|list|limits|iterator|iostream|iomanip|locale) " | awk -F: '{print $1}' | sort > to-remove.txt
for f in $(awk '{print $1}' < to-remove.txt | uniq); do awk -vinc="$f" 'BEGIN { printf("sed -i \"/#include <%s>/d\" ", inc); } /^'$f'/ { printf("%s ", $2) } END { printf("\ngit commit -a -m \"Remove not-needed <%s>\"\n", inc); }' < to-remove.txt; done > /tmp/allremove.sh
. /tmp/allremove.sh
```

### DIFF
--- a/src/Feature.cc
+++ b/src/Feature.cc
@@ -1,7 +1,6 @@
 #include "Feature.h"
 
 #include <cstdio>
-#include <iostream>
 #include <sstream>
 #include <string>
 #include <map>
--- a/src/Feature.cc
+++ b/src/Feature.cc
@@ -1,7 +1,6 @@
 #include "Feature.h"
 
 #include <cstdio>
-#include <sstream>
 #include <string>
 #include <boost/algorithm/string/join.hpp>
 #include <boost/range/adaptor/transformed.hpp>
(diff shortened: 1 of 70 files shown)


kintel (approved):

hzeller:
Since I never compiled `PlatformUtils-win.cc` on my system, the CI revealed that that needed map. Added there in last commit.

hzeller:
After this is merged, I'll start with missing Qt headers...

PR #5368, opened 2024-10-11
Refactoring: Split openscad_gui.cc from openscad.cc, slimmed down ExportInfo

kintel:
* Split openscad_gui.cc from openscad.cc
* Removed redundant fields from ExportInfo, moved platform specific file-format handling closer to where files are opened
* Added support for writing filename metadata to PDF and POV files in cmd-line mode
* Default filename output for PDF export to false in cmd-line mode, mostly to make testing easier.

Note: We _could_ output just the filename without path in PDF exports, but I'm uncertain what the actual use-case for this is.


kintel:
@JonMcLin Could you comment on the filename output in PDF feature? Does it work as intended, or could/should we output just the filename?

JonMcLin:
That was the previous pattern, and I thought it was good.  This results in the path being saved and restored with each save.   The save dialog will come up in the last location a PDF was saved, and then if changed, that new path becomes the default.   There are certainly other ways this could be implemented, but I replicated what was existing and it certainly worked for my applications.      For command-line usage a different implementation that isolates that may be appropriate, so that persistent path is only relevant to when the dialog is used.

kintel:
@JonMcLin I was referring to the feature where the full path of the .scad filename is rendered into the resulting PDF.

JonMcLin:
Ahhh.  No particular use case for that - but probably a result simply of adding the gui.   Although I used it extensively when I was developing the code - that would point me to the specific design-project folder that had the design I'd printed (I'd have design variants, where the variant info was coded in the path. )

t-paul:
I'm not yet sure why but this breaks the lookup for the bundled files in AppImages:

Ok:
```
-rwxr-xr-x 1 tp users 52348096 Okt 23 04:45 ./OpenSCAD-2024.10.23.ai20994-x86_64.AppImage
```
Error:
```
-rwxr-xr-x 1 tp users 52348096 Okt 24 03:51 ./OpenSCAD-2024.10.24.ai21003-x86_64.AppImage
Could not initialize localization.
Error reading examples.json: examples.json: cannot open file
```

t-paul:
Hmm, I think it's caused by moving
```
#ifndef OPENSCAD_NOGUI
#include "gui/QSettingsCached.h"
#define OPENSCAD_QTGUI 1
#endif
```

So due to OPENSCAD_QTGUI not being defined, it uses the boost based PlatformUtils::registerApplicationPath() and that's not working for AppImages as they start via AppRun:

```
$  ./OpenSCAD-2024.10.24.ai21003-x86_64.AppImage --appimage-mount
/tmp/.mount_OpenSCjlunvq

$ find /tmp/.mount_OpenSCjlunvq -name AppRun -o -name openscad -type f
/tmp/.mount_OpenSCjlunvq/AppRun
/tmp/.mount_OpenSCjlunvq/usr/bin/openscad
```

Running `/tmp/.mount_OpenSCjlunvq/usr/bin/openscad` works.

t-paul:
Added AppImage specific lookup path via #5392. This fixes the AppImage startup.

I guess we can now decide if we want to fully go with the boost solution or if we find a way of still using the Qt code.

kintel:
Thanks for finding and fixing! The less Qt in common initialization the better I think. Not sure if there are other hidden corner cases though..

PR #5370, opened 2024-10-12
Fix many missing Qt includes

hzeller:
One commit per header.

Found with grepping out relevant info from the clang-tidy output.

roughly:

```bash
grep 'no header providing.*"Q[a-zA-Z]*".*misc-include-cleaner' OpenSCAD_clang-tidy.out  | awk '{print $6}' | sort | uniq
```

(Exception: not updating `src/openscad.cc `currently, as there is some boost compilation error, probably due to some other header ordering issue somewhere else. Not investigated yet).

### DIFF
--- a/src/gui/AboutDialog.h
+++ b/src/gui/AboutDialog.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "version.h"
+#include <QDialog>
 #include <QWidget>
 #include "gui/qtgettext.h"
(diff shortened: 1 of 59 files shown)


hzeller:
Not sure what is wrong with the Windows CI, but looks like a flaky result.

t-paul:
Yeah, I think I've seen that one earlier. I've re-triggered the job...

kintel (approved):

PR #5373, opened 2024-10-13
Improve quality of temporary file name  for Local Slicer

gsohler:
When using Local Slicer Feature and writing GCodes to SD cards, all of the files are called "OpenSCAD" and its
hard to distinguish them.
This fix should help

### DIFF
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -2257,7 +2257,8 @@ void MainWindow::sendToLocalSlicer()
     exportFileFormat = FileFormat::STL;
   }
 
-  const auto tmpPath = QDir::temp().filePath("OpenSCAD.XXXXXX."+fileFormat.toLower());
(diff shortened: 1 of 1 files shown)


src/gui/MainWindow.cc
```diff
@@ -2257,7 +2257,8 @@ void MainWindow::sendToLocalSlicer()
     exportFileFormat = FileFormat::STL;
   }
 
-  const auto tmpPath = QDir::temp().filePath("OpenSCAD.XXXXXX."+fileFormat.toLower());
+  QString basename = QString(std::filesystem::path(activeEditor->filePath.toStdString()).stem().c_str());
```
t-paul, line 2260: The `filePath` is empty if the file is not saved yet. Currently this is handled in various places, so we either need to do that here too, or have some dedicated method getting "name or *Untitled*" similar to "exportPath()".

kintel: I'm doing some related refactoring in #5357 - will fold this feature into that PR

kintel:
Will include this in #5357

PR #5383, opened 2024-10-17
Add null-coalescing operator '??'

PinkPandaKatie:
Adds a new binary operator, `??`. The expression `x ?? y` evaluates `x`, and if it is not undefined, returns it. If `x` is undefined, it evaluates `y` and returns that.

It is equivalent to: `!is_undef(x) ? (x) : (y)`

This solves a common issue I run into when writing library functions.  I often use special `$` variables to define debugging flags and values, and these variables may not always be defined. Referencing an undefined variable always causes a warning except when calling `is_undef` on a single variable. This means that I have to resort to the long form `if (!is_undef($some_debug_flag) && $some_debug_flag)` to check it without causing a warning.

With this new operator, the code is shortened to `if ($some_debug_flag ?? false)`.

This also makes syntax for calculated defaults much nicer. Say I have a function with 3 arguments, `a`, `b`, and `c`, but `c` is optional and if not specified, should default to `a + b`. This function could  be written as:

`function f(a, b, c=undef) let(c = c ?? a + b) [a, b, c];`

This operator has lower precedence than logical operators `||` and `??` and higher precedence than the ternary operator `a ? b : c`. This is based on its precedence in C# as documented [here](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/).

It is also right-associative, so `a ?? b ?? c` is equivalent to `a ?? (b ?? c)`. If it were left associative (i.e.) `(a ?? b) ?? c`, then the result would be the same, but would cause a warning if `b` were not defined.

### DIFF
--- a/src/core/Expression.cc
+++ b/src/core/Expression.cc
@@ -108,6 +108,15 @@ Value BinaryOp::evaluate(const std::shared_ptr<const Context>& context) const
     return this->left->evaluate(context).toBool() && this->right->evaluate(context).toBool();
   case Op::LogicalOr:
     return this->left->evaluate(context).toBool() || this->right->evaluate(context).toBool();
+  case Op::IfUndef: {
+    if (auto lookup = dynamic_pointer_cast<Lookup>(this->left)) {
+      auto result = context->try_lookup_variable(lookup->get_name());
+      return (result && result->isDefined()) ? result->clone() : checkUndef(this->right->evaluate(context), context);
+    } else {
+      Value leftValue = this->left->evaluate(context);
+      return leftValue.isDefined() ? std::move(leftValue) : checkUndef(this->right->evaluate(context), context);
+    }
+  }
   case Op::Exponent:
     return checkUndef(this->left->evaluate(context) ^ this->right->evaluate(context), context);
   case Op::Multiply:
@@ -143,6 +152,7 @@ const char *BinaryOp::opString() const
   switch (this->op) {
   case Op::LogicalAnd:   return "&&";
   case Op::LogicalOr:    return "||";
+  case Op::IfUndef:      return "??";
   case Op::Exponent:     return "^";
   case Op::Multiply:     return "*";
   case Op::Divide:       return "/";
--- a/src/core/Expression.h
+++ b/src/core/Expression.h
@@ -48,6 +48,7 @@ public:
   enum class Op {
     LogicalAnd,
     LogicalOr,
+    IfUndef,
     Exponent,
     Multiply,
     Divide,
--- a/src/core/lexer.l
+++ b/src/core/lexer.l
@@ -278,6 +278,7 @@ use[ \t\r\n]*"<"        { BEGIN(cond_use); LOCATION_COUNT_LINES(parserlloc, yyte
 "!="                    return NEQ;
 "&&"                    return AND;
 "||"                    return OR;
+"??"                    return IFUNDEF;
 
 .                       { return yytext[0]; }
 
--- a/src/core/parser.y
+++ b/src/core/parser.y
@@ -127,13 +127,14 @@ bool fileEnded=false;
 %token TOK_FALSE
 %token TOK_UNDEF
 
-%token LE GE EQ NEQ AND OR
+%token LE GE EQ NEQ AND OR IFUNDEF
 
 %nonassoc NO_ELSE
 %nonassoc TOK_ELSE
 
 %type <expr> expr
 %type <expr> call
+%type <expr> null_coalesce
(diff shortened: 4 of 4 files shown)


jordanbrown0:
A related construct (that maybe we can't implement because of backward compatibility) is a JavaScript-style `||` operator.  `a || b` is equal to `a` if `a` is truthy, else is equal to `b`.  (But we currently have `||` returning a boolean, which is subtly different.)

Note also that you can trivially implement this feature as a function.  I've had
```
function default(val, def) = is_undef(val) ? def : val;
```
for a long time, and `default(x, y)` is reasonably terse and not particularly magic-looking.

PinkPandaKatie:
@jordanbrown0 I actually already defined a function like that:

```
function coalesce(a, b, c, d, e) = (
    !is_undef(a) ? a :
    !is_undef(b) ? b :
    !is_undef(c) ? c :
    !is_undef(d) ? d :
    e
);
```

It basically works like `COALESCE` in SQL, where it takes multiple arguments and returns the first non-null value.

The problem is, it doesn't solve the problem of the possibly undefined variable, since the variable throws a warning as soon as it's referenced, before the function even runs: 

```
function default(val, def) = is_undef(val) ? def : val;

// ECHO: value = "abc" 
echo(value = !is_undef(undefined_var) ? undefined_var : 3.0)

// WARNING: Ignoring unknown variable 'undefined_var' in file test.scad, line 5
// TRACE: called by 'default' in file test.scad, line 5
// TRACE: called by 'echo' in file test.scad, line 5
echo(value = default(undefined_var, "abc"));
```

That's true for every user function and also every builtin except `is_undef`. The `builtin_is_undef` function in `builtin_functions.cc` is defined like this:

```
Value builtin_is_undef(const std::shared_ptr<const Context>& context, const FunctionCall *call)
{
  if (call->arguments.size() != 1) {
    print_argCnt_warning("is_undef", call->arguments.size(), "1", call->location(), context->documentRoot());
    return Value::undefined.clone();
  }
  if (auto lookup = dynamic_pointer_cast<Lookup>(call->arguments[0]->getExpr())) {
    auto result = context->try_lookup_variable(lookup->get_name());
    return !result || result->isUndefined();
  } else {
    return call->arguments[0]->getExpr()->evaluate(context).isUndefined();
  }
}
```

The `if (auto lookup = ...)` clause checks if the AST node of its first argument is a variable lookup, and if so, runs `try_lookup_variable`, which returns an empty `boost::optional<const Value&>` without causing a warning if the variable doesn't exist.

jordanbrown0:
Indeed, I'd forgotten the not-defined versus defined-as-undef distinction.

kintel:
Thanks for your PR. First of all, adding new language constructs takes time, as it takes time to evaluate how new constructs fit into the language and how it interacts with existing and other planned/proposed changes.

Sometimes soliciting input on the mailing list can help tease out any critical opinions. While we cannot listen to every criticism, it's important to be aware of them, especially since we currently don't have a structured proposal mechanism in place.

One note: I think one of the benefits of this operator which I didn't see you mention is that it would guarantee the each operand would be evaluated at most once.

PinkPandaKatie:
@kintel I joined the mailing list and sent a message. I don't know if it went through because it's not showing up on the archive but maybe there's just a delay.

kintel:
> @kintel I joined the mailing list and sent a message. I don't know if it went through because it's not showing up on the archive but maybe there's just a delay.

Mailing list requires approval due to high spam volumes. I just approved you.

PinkPandaKatie:
I asked about this on the mailing list and there wasn't much opposition. The thread just kinda went stale. I still think it could be useful.

kintel:
I think it could be useful too. Two challenges:
1. It's a bit risky to merge a new feature when we're sprinting towards a new release. Making it experimental makes it possible to quickly merge it.
2. No tests. If experimental, test can be added later.

PinkPandaKatie:
I'm ok with it being experimental but I'm not sure how to do that. I could also write some tests for it.

kintel:
Experimental features are managed by https://github.com/openscad/openscad/blob/master/src/Feature.h
You pretty much have to find a good place to check if the feature is enabled by the user (in Preferences -> Features).
Since you're changing the parser, we probably have to do the check some place _after_ parsing happens.

kintel:
..then to graduate it from experimental, we give is some soak time, gather feedback from users, write docs/examples and make sure test coverage is good. That may take some time, but while waiting people can test it out using the nightly builds.

PinkPandaKatie:
I'd still like to see this get merged at some point

PR #5385, opened 2024-10-18
Add geometry/Geometry.h header in files that need it.

hzeller:
Created with missing header findings of `clang-tidy`:

```
../dev-tools/insert-header "geometry/Geometry.h" $(grep "no header providing.*misc-include-cleaner" OpenSCAD_clang-tidy.out  | awk -F: '/"Geometry"/ { print $1}' | sort | uniq)
```

### DIFF
--- a/src/glview/cgal/CGALRenderer.h
+++ b/src/glview/cgal/CGALRenderer.h
@@ -6,6 +6,7 @@
 
 #include "glview/VBORenderer.h"
 #ifdef ENABLE_CGAL
+#include "geometry/Geometry.h"
 #include "glview/cgal/CGAL_OGL_Polyhedron.h"
 #include "geometry/cgal/CGAL_Nef_polyhedron.h"
 #include "core/Selection.h"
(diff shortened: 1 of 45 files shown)


src/glview/cgal/CGALRenderer.h
```diff
@@ -6,6 +6,7 @@
 
 #include "glview/VBORenderer.h"
 #ifdef ENABLE_CGAL
+#include "geometry/Geometry.h"
```
kintel, line 9: Should be outside the ifdef

kintel (commented):
The include insertion isn't aware of conditions like #ifdef ENABLE_CGAL. I didn't go through all, but marked one instance.

src/glview/cgal/CGALRenderer.h
hzeller: Oh, will double-check all files.

hzeller:
Sorry, on vacation next 1.5 weeks, will revisit when I come back. Closing this for now and start with the state then.

PR #5393, opened 2024-10-27
Povray viewport

folkertvanheusden:
This patch series makes the povray using the OpenSCAD viewport used when exporting.


kintel (commented):
Two initial comments (without reading the code in detail):
* Should camera be part of ExportInfo?
* We need more tests of various camera parameter corner cases. We have a bunch of existing camera tests; perhaps they can be reused? Search for `--camera` in tests/CMakeLists.txt

folkertvanheusden:
Ok, moved it to ExportInfo.

Now I have to figure out what is going wrong when I also change the
rotation for the y-axis.

kintel:
The script is good, but is not automated. The primary use-case for such tests is not really to verify that the code works today - I expect that to be resolved during development. The tests ensure that if someone refactors the code in the future, or change related code, that this feature doesn't break.
Ideally, we should have end-to-end tests including povray rendering, but that might be a bit big to take on right now.

folkertvanheusden:
> The script is good, but is not automated. The primary use-case for such tests is not really to verify that the code works today - I expect that to be resolved during development. The tests ensure that if someone refactors the code in the future, or change related code, that this feature doesn't break. Ideally, we should have end-to-end tests including povray rendering, but that might be a bit big to take on right now.

Added a few automated tests now

src/io/export_pov.cc
```diff
@@ -36,6 +36,14 @@
...
 #include "geometry/linalg.h"
 
+void RotMat(double& x, double& y, double a)
+{
+  double x2 = x*cos(a) - y*sin(a);
+  double y2 = x*sin(a) + y*cos(a);
+  x = x2;
+  y = y2;
+}
+
```
kintel, line 46: This looks unused

folkertvanheusden:
Oops! Fixed in
https://github.com/folkertvanheusden/openscad/commit/d921d7b0e8d1d5fb82a85367de64d4c25e6580fb

kintel (approved):

PR #5399, opened 2024-10-30
Use filesystem::relative for boostfs_uncomplete.

thehans:
Cleans up some old custom code with now standard function, since Boost 1.60.  I would have liked to directly replace the calls to boostfs_uncomplete with calls to fs::relative, but there was a difference in that empty path was treated as equivalent to "." in the old code.

I think it should be fine to update our requirements to boost 1.60 (in README and and check dependencies scripts).  However there is also the option to swap out `boost::filesystem` for `std::filesystem` here.  I think one or the other should be done before merging.


kintel:
Last I checked, compiler support (or rather stdlib support) for std::filesystem wasn't great. Do you know if this could pose a challenge for moving to std::filesystem?

kintel:
macOS is on libc++ and if we bump minimal version requirements from 10.14 to 10.15 I think we can use it.

thehans:
What's the significance of 10.15?

Also the one major difference I've encountered when swapping in `std::filesystem` is that `std::filesystem::last_write_time` returns `std::filesystem::file_time_type` as opposed to `std::time_t` in boost.
I wrote a little helper function to return the value of seconds since epoch time, but ran into a confusing issue where I was seeing very negative numbers of high magnitude.  I finally found that this is because libstdc++ is using an internal epoch time in the far future of [Jan 1 2174](https://stackoverflow.com/a/67142253/251068)  and not what I had assumed to be the typical 1970 epoch.

Question is if we care that csg output has negative value "timestamp" (for `import` and other nodes that accept filenames).
As far as I can tell the code doesn't really read in and use this timestamp info for anything?

kintel:
No real significance of requiring macOS 10.15, just that everytime we increase requirements, we lock out some users with older hardware. Not much we can do about it as other dependencies (like Qt) also impose minimum requirements.

I believe the timestamp is really only used for caching.

thehans:
Ah, I misunderstood thinking 10.15 was referring to clang or libc++ version.  I truly know nothing about the mac world, would that affect CI at all or pretty much just a documentation change?

thehans:
OK I see the errors already in CI `error: 'path' is unavailable: introduced in macOS 10.15` etc.
I will try a (manual) search and replace of 10.14 with 10.15

xcode/OpenSCAD.xcodeproj/project.pbxproj
kintel, line 1: Thus file isn't in use any longer. I'll kill this in a separate PR

kintel:
The main doc for system requirement is on the download page: https://github.com/openscad/openscad.github.com/blob/main/downloads.html#L444

kintel (approved):
Good stuff!

PR #5413, opened 2024-11-07
Fix crash if "libraries" resource path is not used

cjmayo:
```
$ openscad
terminate called after throwing an instance of 'std::filesystem::__cxx11::filesystem_error'
  what():  filesystem error: cannot make absolute path: Invalid argument []
Aborted (core dumped)
```

### DIFF
--- a/src/core/parsersettings.cc
+++ b/src/core/parsersettings.cc
@@ -154,5 +154,7 @@ void parser_init()
 
   add_librarydir(PlatformUtils::userLibraryPath());
 
-  add_librarydir(fs::absolute(PlatformUtils::resourcePath("libraries")).string());
+  if (!PlatformUtils::resourcePath("libraries").empty()) {
+    add_librarydir(fs::absolute(PlatformUtils::resourcePath("libraries")).string());
+  }
 }


kintel:
@thehans This looks related to your recent work. Could you take a look, and perhaps think about whether similar issues could be lingering elsewhere in the codebase?

thehans:
Anywhere there is a possibility to pass an empty path to `fs::absolute` it could be a problem.  

The boost::filesystem behavior was to return `fs::current_path()` when `fs::absolute` was passed an empty path.

There are currently 11 total calls to `fs::absolute`.  

- The line being addressed in this PR (I would instead use `fs::current_path()` in case of empty path, for full compatibility with previous behavior):
https://github.com/openscad/openscad/blob/fd3a9aad2bcd913ac1830e11670f0a422231e43f/src/core/parsersettings.cc#L156-L158

- Other usage in same source file. I guess someone could theoretically set an empty path between separators in their OPENSCADPATH env var, so maybe good to handle that just in case
https://github.com/openscad/openscad/blob/fd3a9aad2bcd913ac1830e11670f0a422231e43f/src/core/parsersettings.cc#L144-L153

- I doubt fontpath can be empty while fonts.conf exists as a regular file, but I don't know the details of resource path handling on every platform well enough so maybe its still possible?
https://github.com/openscad/openscad/blob/fd3a9aad2bcd913ac1830e11670f0a422231e43f/src/FontCache.cc#L131-L134

- This one explicitly is wrapped in a check that the path exists which would return false on empty:
https://github.com/openscad/openscad/blob/fd3a9aad2bcd913ac1830e11670f0a422231e43f/src/FontCache.cc#L165-L168

- Is it possible to get to this export code with an empty filename?  I think it wouldn't pass the command line parsing but not sure.
https://github.com/openscad/openscad/blob/fd3a9aad2bcd913ac1830e11670f0a422231e43f/src/openscad.cc#L329-L332

- 2 instances in parser.y fixed by my previous PR
https://github.com/openscad/openscad/blob/fd3a9aad2bcd913ac1830e11670f0a422231e43f/src/core/parser.y#L763-L764

- fileutils.cc has 2 usages already wrapped in explicit empty path check
- Platformutils-posix.cc has 2 usages wrapped in `is_directory` checks which would not succeed on empty path.

kintel:
Is this superseded by #5416 ?

thehans:
Yes the same line of code is covered there.  thanks @cjmayo for raising the issue to my attention.

PR #5414, opened 2024-11-07
Update manifold

cjmayo:
Please check carefully, I've just looked at the manifold code for how things have changed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,7 +53,6 @@ option(USE_CCACHE "Use ccache to speed up compilation." ON)
 option(ENABLE_CAIRO "Enable support for cairo vector graphics library." ON)
 option(ENABLE_SPNAV "Enable support for libspnav input driver." ON)
 option(ENABLE_HIDAPI "Enable support for HIDAPI input driver." ON)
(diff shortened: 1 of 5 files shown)


kintel:
Manifold changed a bit. I'm planning to look into this soon, hopefully next week.

t-paul:
Manifold 3.0: https://github.com/elalish/manifold/pull/1024

pca006132:
Wow, this is simpler than I thought!

kintel:
Enabling TBB only in experimental mode sounds wrong - that's probably an oversight by me when moving Manifold out of experimental.

kintel:
I would suggest: Remove ENABLE_TBB, remove the MANIFOLD_PAR check, and always enable MANIFOLD_PAR (except for Emscripten I guess). That should simplify the CMake setup a bit.

If we need to be able to build a non-parallel Manifold in the future, we can bring this back with a more concrete use-case in mind.

kintel:
@cjmayo Feel free to take on the above changes, but I'm fine to merge this and deal with the TBB/experimental stuff separately.

kintel (approved):

PR #5416, opened 2024-11-08
Add protection/workaround for every possible call to fs::absolute with empty path

thehans:
This addresses a difference in behavior between `boost::filesystem::absolute` and `std::filesystem::absolute` where passing an empty path to boost would return `fs::current_path()`, and the std library now throws a `std::filesystem_error`.

I don't know if an empty path can even necessarily reach all these calls, but better to just handle the possibility regardless, than wait for another bug report.


thehans:
relevant discussion also in #5413

kintel:
One question is whether current_path is even wanted in OpenSCAD as a fallback.
..but this is safe for now as it reverts to the previous behavior, so further tweaks can be deferred.

kintel (approved):

PR #5423, opened 2024-11-11
Italian localization, contributed via email.

t-paul:


t-paul:
Superseded by #6063 which includes most of the fixes to the issues noted above.

Export options will not be renamed generally as it's useful to have the action itself to spell out the actual function, e.g. for external call to actions.

Issue #1610, opened 2016-03-29
align(): generalized alignment


doug-moen:
The `align()` operator aligns a shape with respect to the X, Y and Z axes. For each axis, you can align the shape to be centred, above (in the positive range), below (in the negative range), or specify no change. To match the capabilities of `halign` and `valign` in `text()`, you can also request to be centred, above or below a specific coordinate value on a particular axis.

The `align()` operator is more general than the `center=` argument supported by some geometric primitives, and can also replace most uses of the `halign=` and `valign=` arguments of `text()`.

In general, the arguments are `align(all=aspec, x=aspec, y=aspec, z=aspec)child` where each *aspec* argument defaults to `undef`. An *aspec* value is one of the following:
* `"center"`, which is short for `["center",0]`
* `["center",n]` -- move the center of the bounding box to coordinate n on the specified axis
* `"above"`, which is short for `["above",0]`
* `["above",n]`-- move the bottom of the bounding box to coordinate n on the specified axis
* `"below"`, which is short for `["below",0]`
* `["below",n]`-- move the top of the bounding box to coordinate n on the specified axis
* `undef`, meaning no change

`align("center") cube(10)` is the same as `cube(10, center=true)`. It centres the cube on the origin.

`align("above") cube(10)` is the same as `cube(10, center=false)`. It places the cube in the +X,+Y,+Z octant, adjacent to the origin.

The `align` operator has the same power as the `center=[x,y,z]` argument proposed by #265. For example,
```
align(x="center", y="center") cube(10)
```
is the same as the proposed
```
cube(10, center=[true,true,false])
```

Issue #621 proposes to add a `center=` option to `import()`. This is a more general solution.

Issue #586 proposes to either add a `center=` option to `resize()`, or to add a separate `center()` primitive. This is a more general solution.

Using `align()`, it is possible to implement a `zstack()` module which stacks a list of shapes along the Z axis. This is because calls to `align()` can be nested, which is an extra degree of flexibility that you don't get from alignment options in each primitive shape module.

Initial discussion of `align()` took place in issue #1609.

t-paul:
In addition we could review if https://github.com/devlaam/openscad/blob/loop_extension/doc/Loop.pdf from the old pull request #450 has some more ideas that could be included...

doug-moen:
The `align()` operator is more general than the generalized `center=` arguments of #450. For example,
```
sphere(4, center=[0,1,-1]);
```
from the Loop.pdf document is equivalent to
```
[...]

PR #5433, opened 2024-11-16
add a per axis center argument to the square and cube primitives

mickael868:
Sometimes, I have to center a square or a cube along one axis but not another. The current solution is to translate the primitive along an axis but not the other, eg :

`
translate([0, -10]) square([10, 20]);
// or
translate([5, 0]) square([10, 20], center=true);
`

A shorthand solution could be to specify along which axis the centering should be performed:

`
square([10, 20], center=[false, true]);
`

I find this easier to read and avoid the values duplication but YMMV.

This PR implements the described feature for a square. If the idea is accepted, I can update the PR to extend it to the cube primitive. The cheatsheet should also be updated, I can also help but I did not searched where to do it yet.

This PR add a Vec2<bool> support for the center argument but still accepts a single boolean value.


jordanbrown0:
See #1610.  If you're going to generalize centering/alignment, including "put the object on the positive/negative side" is an obvious and vary useful thing to cover.

t-paul:
As very much visible in the discussion in #1610, things are nowhere near a consensus on how to specify this. This may need some stepping back and consideration if there is a different solution.

mickael868:
## Goal

Allow to center an object independently along each axis, either centered or toward the positive or negative direction.
This feature is specified in the object arguments list itself to save a translate().
It is based on the existing "center" argument, and a similar feature has been discussed in the issue [#1610](https://github.com/openscad/openscad/issues/1610).

## On which objects 

I think that, unlike the issue [#1610](https://github.com/openscad/openscad/issues/1610), this PR is limited to "centerable" primitives:

- 2D: circle, square
- 3D : sphere, cube, cylinder

I ruled out the polygon and polyhedron because one would typically use absolute points. I also think that this PR does not include operations such as addition or hull, the code and the goal being slightly different. Is this okay ?

## Parameter name

The issue [#1610](https://github.com/openscad/openscad/issues/1610) is about an "align" parameter. having both "align" and "center" may lead to confusion and incoherence, eg. `square(10, center=false, align=["above", "below"]);`. 

Removing "center" would break compatibility. 

This leads to keep using the optional "center" argument alone, right ?

## Parameter value name

Here are several possibilities:

- true/false may not be used, "-true" seems ill-formed to me
- Generic literal naming: "below", "center", and "above" is pretty good but as stated in the other issue, it is not well suited for the X axis
- Dimension specific literal naming, using the keywords cited in the other issue or the keywords used by SolidPython :
  - "left", "center", and "right" for the X axis
  - "backward", "center", and "forward" for the Y axis
  - "below", "center", and "above"  or up/down for the Z axis in 3D (and arguably for the Y axis in 2D)
- Extrema naming: "min", "center", and "max"
- Numerical naming: -1/0/+1 seems good but there is [a comment giving a good argument against this](https://github.com/openscad/openscad/issues/1610#issuecomment-779439338)
- Operator naming : "-", "=", and "+" or "<" and ">": this removes the numerical naming problem (use 0.5, 1, 2?) but the "=" is not intuitive IMHO. "c" or "center" may be used instead of "=".

Some variants:

- Allow all the naming variants, eg `square(10, center=["-1", "center", "below"]);`, this is confusing but one would typically use the same naming convention in all their files.
- Implicit  center : to avoid the "0" or "=" keywords of the operator naming, if the center argument is given as a vector,  only specify the + and -, eg. `center=[x="+", z="-"]`. Does the  dict/map variable type exists in OpenSCAD? 
- Single string: write `"abc"` instead of `["a", "b", "c"]`. This is shorter but maybe less readable.

Are there other options? Should the goal be broader?

jordanbrown0:
My personal library uses a vec3 of +1/0/-1 and I've been happy with that.  Using numbers can be helpful when you are trying to do mirror-image constructs because you can use the same +/-1 as a multiplier in an associated translate - I have a lot of cases of `for (s=[-1,1]) translate([s*10,0,0]) jcube(5, j=[s,0,0]);`. (Yes, you could also do this with mirror().) As my earlier comment says, this scheme does not extend well to fractions (e.g. center on the 2/3 mark) or larger values (e.g. put my 1-unit object from +1 to +2) but I haven't felt the need for those  - and of course literal string variants don't do those either. 

A single string would be awkward for mirror constructs where you want to have the alignment controlled by a variable. You'd have to build the string with str().

BOSL2 uses named constants that you add together - e.g. FWD+RIGHT aligns on minimum Y, maximum X, and centers on Z.  But I don't think we want to introduce a bunch of predefined constants.  

One thing to watch out for: alignment is not well defined for circles.  Or, rather, for regular polygons and polyhedra created using circular constructs like circle() and cylinder().  Should it align based on the theoretical circle, or based on the actual polygon? There are use cases for each. Note that for odd numbers of sides even centering is problematic.

t-paul:
I think it would be worth exploring if introducing attachment points is a better option that would also allow to expand this to user modules at some point.

Again pointing to the old issue which mostly showed cryptic arrays as "center" specification are not getting consensus.

mickael868:
The center taking an array of  -1, 0, or 1 feature is now implemented for sphere, cube, cylinder, square, and circle.

There are still some limitations :

- `center=[0, 0]` is equivalent to `center=true` whereas `center=[1, 1]` is equivalent to `center=false`, this seems inverted. That said, when reading a single center argument, it is still clear what it does (eg. `center=[1, 0]` is centered on y and not on x and `center=true` is centered)
- for the cylinder, there are two radii (r1 and r2), I chose to use max(r1, r2) for the offset distance when the cylinder is not centered around x or y. This is arbitrary.
- This circle offset implementation uses the theoretical circle radius, not the actual polygon radius. I think that both may be equally valid depending on the usage. As long as this is a consistent and documented behavior, it is okay for me. If someone wants to use the polygon radius, they can still use `translate()` on a centered object, probably with ` rotate()`.

The boolean center is still working and the default behavior is not changed. It fits my initial need and I find the notation pretty self explanatory. What does make it cryptic?

Does this "array center" conflicts with a broader "attachment point" feature? For me, it is 2 different things, the array center goal is to type less, eg. `square([width, 20], center=[0, 1])` instead of `translate([-width/2, 0]) square([width, 20])`. This is especially true when `width` is an expression such 2*a+b-sin(alpha), it reduces the need for duplication or temporary variables.

kwikius:
@mickael868  Reading through the thread I am afraid you are wasting your time here.

I think it is a very nice solution and it has been reinvented by at least 4 or 5 other people which reinforces the belief that it is a good solution to a real need.  

However the people above hold the power here and they have indicated above that it isnt going to happen ( I learnt this through a previous pull request to Openscad, which I wasted a lot of time on before figuring it out) They wont of course say this directly, but that is what they are saying.

The good thing is that you can implement your own version of the primitives and be on your way, which is what I did

gsohler:
Yes, I agree that for some features it's *really* hard to merge
But there is yet another viable option - fork and implement it myself.
So did I with pythonSCAD and I am happy to have it.
The characters in the center show the alignment for each axis of the cube.
I am happy to backport the functionality to scad language on demand.

https://imgur.com/a/7iQrc1E

mickael868:
@kwikius thank you for your support. I have no hard feeling about this feature not being merged though. 
I can understand that maintainers do not wish to get their project alienated. Plus, it was really pleasant to dig into the code and I did spend only few hours doing this PR thanks to the simplicity and the readability of the existing code (and cmake generating everything an LSP server needs). 

On a side note, I would prefer an `operator==()` to decide whether OpenSCAD should re-build an object instead of comparing the `toString()` output. This is an even smaller issue though, I do not expect a PR  about this to be merged given the context.

[This inner/outer radius post](https://github.com/openscad/openscad/pull/5433#issuecomment-2512408684) is interesting,  we could have "center" being an alias for "outer-center" and there would be another "inner-center" option. The inner-center would align the left side of the right pointing triangle to 0 on the x axis if the x center is 1. If the x center is 0, it would be the middle between the rightmost point and the leftmost point. Odd-numbered polygons may have an arbitrary but consistent "rotation", eg. the shape (a triangle in the given example) points to the center when the center is not [0, 0, 0].

The regular polygon alignment aside, I do not get any tangible reason preventing this PR to be merged. Without any upvote for an alternative or clear criticism of this PR, I have nothing to do to improve this PR.

@gsohler I chose SolidPython over pythonSCAD years ago because it was available in the ArchLinux repositories. You did a great job but providing an independent package (SolidPython) seems to be easier to spread than a fork (pythonSCAD). This is why I prefer to not maintain a fork for this little PR. If I stop maintaining the fork, all the code depending on it becomes obsolete because a fork is arguably more short-lived than an independent project.

jordanbrown0:
The reason that this and many other language-enhancement PRs stall is that a language change is nearly forever - compatibility means that you mostly can't ever back it out.  As a result, it needs to be the *right* change, or at least clearly a subset of the right change so that a future change can compatibly complete the picture.  When people can reasonably say "but what about...", or just can't spend the time to understand a complex change, the change stalls.

gsohler:
@miakael868,

Personally I don't bother, that you go with SolidPython, it's a fair
decision. Some time back I also tried to get SolidPython running in my
place, but I failed to get it running together with Jupyter,, there were
some strange missing libs errors, which I could not resolve. Likely this is
very similar  to people, which cannot get PythonSCAD running due to python
version conflicts ....

And, yes, I agree that an independent Software has MANY advantages over a
stacked/tandem solution. and probably the greatest reason is that you get
new features automatically when the underlying software(openscad is
improved) . Most prominent disadvantage  through is that you cannot access
the actual vertices list, modify them and get a solid again in the same
run(mesh() function)

If someone actually goes with PythonSCAD and after some time I am not
available to work on PythonSCAD anymore , the person is stuck/orphaned in
his projects ...
But is that  a realistic scenario ?

*) When pythonSCAD was not developed anymore, guys would miss latest
updates and bug fixes of OpenSCAD, but there would always be a latest
stable version + source code, which can keep their designs alive
* )Many people  use PythonSCAD and I am confident that somebody would
volunteer to follow up on PythonSCAD *just* because there is demand.
*) Maybe @kintel or @teepee merge my latest python branch, but admittedly,
that is rather unlikely
*) If all of the above fallbacks don't, appy, as PythonSCAD never removed
ANY support for SCAD language, there is still the option to export CSG.
It's valid code to run in OpenSCAD alone.

@Jordan Brown ***@***.***> Yes, it's a sad fact, that
the majority of PR's stall for almost infinite time  and all of us accept
that fact. And yes, OpenSCAD maintainer do this to stay backward compatible
and to choose the optimal path  to future whenever possible,
There is a tradeoff of staying compatible and staying attractive to  3D
model developers For openSCAD this is very  much to the left  side of the
sentence
.
However, that sadly implies that OpenSCAD has very slow development time,
it lacks killer features, which other free cad software have since ages and
IMHO OpenSCAD will not stay competive on a long term basis in the free CAD
software world.

Finally, I see daily increasing web page visits and downloads, it appears
my success proves me right 😃

PR #5456, opened 2024-11-28
Povray animation

folkertvanheusden:
Animates an scad-scene via its $t variable. It produces a number of pov-files that you can then render individually. 

I then merge them in a video using `ffmpeg -f image2 -framerate 25 -i frame%05d.png -vcodec libx264 -crf 22 video.mp4`

### DIFF
--- a/src/gui/Animate.cc
+++ b/src/gui/Animate.cc
@@ -33,6 +33,7 @@ void Animate::initGUI()
   this->anim_tval = 0.0;
   this->anim_dumping = false;
   this->anim_dump_start_step = 0;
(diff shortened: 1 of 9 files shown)


kintel (commented):
Added some minor comments.

However, I feel like the animation logic is getting a bit convoluted and hard to reason about. I'm sure this technically works, but the code is getting very hard to maintain at this point. It feels like the right thing to do is to refactor the animation code and put it into a bit better isolated component. This is, of course, not trivial, and it's all amplified by not having GUI tests, so every change needs to go through vigorous manual testing.

PR #5457, opened 2024-11-28
make File->Export code  more versatile, no per-format slot

gsohler:
By use QSignalMapper, there is no need anymore to have an
per-fileformat slot. 
The code might even become shorter now by utilizing existing 
fileformat tables.


src/gui/MainWindow.cc
```diff
@@ -2952,56 +2962,43 @@ void MainWindow::actionExport(FileFormat format, const char *type_name, const ch
...
+      break;
+    case FileFormat::AMF:
+      actionExport(FileFormat::AMF, "AMF", ".amf", 3);
+      break;
+    case FileFormat::DXF:
+      actionExport(FileFormat::DXF, "DXF", ".dxf", 2);
+      break;
+    case FileFormat::SVG:
+      actionExport(FileFormat::SVG, "SVG", ".svg", 2);
+      break;
```
kintel, line 3000: I wonder if you could use `FileFormatInfo` to express this directly: `FileFormatInfo info = fileformat::info(fmt)`

gsohler: Cannot find any FileFormatInfo in all code. Did it recently disappear or do you rather refer to  ExportFileFormatOptions ?

kintel: It was recently introduced. I wrote it a while ago, but I guess it only made it into master yesterday:
https://github.com/openscad/openscad/blob/11fb64f013b494399358dc4a5ff5de84847983b0/src/io/export.h#L41

src/gui/MainWindow.cc
```diff
@@ -509,18 +512,25 @@ MainWindow::MainWindow(const QStringList& filenames)
...
-  connect(this->fileActionExportWRL, SIGNAL(triggered()), this, SLOT(actionExportWRL()));
-  connect(this->fileActionExportPOV, SIGNAL(triggered()), this, SLOT(actionExportPOV()));
-  connect(this->fileActionExportAMF, SIGNAL(triggered()), this, SLOT(actionExportAMF()));
-  connect(this->fileActionExportDXF, SIGNAL(triggered()), this, SLOT(actionExportDXF()));
-  connect(this->fileActionExportSVG, SIGNAL(triggered()), this, SLOT(actionExportSVG()));
-  connect(this->fileActionExportPDF, SIGNAL(triggered()), this, SLOT(actionExportPDF()));
-  connect(this->fileActionExportCSG, SIGNAL(triggered()), this, SLOT(actionExportCSG()));
-  connect(this->fileActionExportImage, SIGNAL(triggered()), this, SLOT(actionExportImage()));
+  for(auto &action : { 
+		  this->fileActionExportSTL, this->fileActionExport3MF, this->fileActionExportOBJ,
```
gsohler, line 516: I am actually wondering, if this list can by a table with 2 columns here, defining the qt signal + the file format

kintel: Yeah, or probably just a map:
```
std::unordered_map<QObject*, FileFormat> = {
  {this->fileActionExport3MF, FileFormat::_3MF},
  ...
};
```

src/io/export.h
```diff
@@ -35,7 +35,8 @@ enum class FileFormat {
   PNG,
   PDF,
   POV,
-  PARAM
+  PARAM,
+  Image
```
kintel, line 39: Image doesn't really fit, but since we only support PNG at the moment, perhaps make it PNG, and either add it as a 2D file format in `initialize()` in export.cc, or add a new isImage() function.

src/gui/MainWindow.cc
```diff
@@ -2819,29 +2821,17 @@ void MainWindow::actionExportFileFormat(int fmt_)
 {
   FileFormat fmt = (FileFormat) fmt_;	
   FileFormatInfo info = fileformat::info(fmt);
-  char suffix[10]; // will need 5 bytes at most
+  const std::string suffix = "."+info.suffix;
+  printf("suffix is %s\n",suffix.c_str());
+  printf("identifier is %s\n",info.identifier.c_str());
```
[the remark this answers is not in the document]
gsohler: I removed the printf again, 
the reason why i put it:

when trying to export STL from my model, the GUI form asked me to save "binstl" files.
(its lowercase and its not stl only)
This was quite unusual to me, but info.identifier clearly returns that, so suppose thats ok
I tested all other formats and at least the save dialog did appear, but also all formats were lowercase

gsohler:
> One small comment, otherwise looks good.
> 
> Since we don't have automated GUI tests: Did you manually check that each format still works?

yes, i did :)

src/gui/MainWindow.cc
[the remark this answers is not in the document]
kintel: That sounds like a bug: we shouldn't expose the identifier to users, that's what the display name is for

kintel (approved):

Issue #5459, opened 2024-11-29
POVray export 


UBaer21:
version 2024.11.29 
`sphere(50);`  results in a belt missing most of the sphere.
![image](https://github.com/user-attachments/assets/b604b000-5bb9-49bd-ad7b-e4bfac9d9835)

PR #5460, opened 2024-11-29
Fix for #5459.

folkertvanheusden:
Issue #5459 is caused by a missing PolySetUtils::tessellate_faces.


src/io/export_pov.cc
```diff
@@ -40,6 +40,9 @@
 void export_pov(const std::shared_ptr<const Geometry>& geom, std::ostream& output, const ExportInfo& exportInfo)
 {
   std::shared_ptr<const PolySet> ps = PolySetUtils::getGeometryAsPolySet(geom);
+  if (!ps->isTriangular()) {
+    ps = PolySetUtils::tessellate_faces(*ps);
+  }
```
kintel, line 45: I believe we discussed this during the initial POV export work: POV export is able to handle non-triangular polygons, and there was some argument why that was intentional. This line only make all polygons into triangles. It's a fine workaround, but it really indicates that there is a bug downstream from here which doesn't handle such polygons.

src/io/export_pov.cc
```diff
@@ -55,18 +56,47 @@ void export_pov(const std::shared_ptr<const Geometry>& geom, std::ostream& outpu
...
-      const auto & z = ps->vertices[polygon[i]].z();
-      output << "<" << x << ", " << y << ", " << z << ">";
+    // create a PolySetBuilder with only this polygon
+    PolySetBuilder psb;
+    psb.beginPolygon(polygon.size());
+    for (size_t i=0; i<polygon.size(); i++)
+      psb.addVertex(ps->vertices[polygon[i]]);
+    auto temp_polyset = psb.build();
+    // triangulate it
+    temp_polyset = PolySetUtils::tessellate_faces(*temp_polyset);
```
kintel, line 70: Can't you just call `PolySetUtils::tessellate_faces()` on the initial polyset, then iterate over triangles?

folkertvanheusden: Then you loose context. What I mean with that you no longer know which triangles are part of the same mesh.

kintel: Something feels wrong.

One PolySet conceptually should map to one mesh. When you call tessellate_faces, you get the same geometry and color information back, just with the guarantee that each polygon is a triangle.

It's possible that this is a result of you creating one PovRay mesh per _polygon_ instead of per _PolySet_.
Take a look at https://www.povray.org/documentation/3.7.0/r3_4.html#r3_4_5_2_4
It describes that you can create indexed meshes where you specify a texture index for each triangle:

```
face_indices {
    2,
    <0, 1, 2>, 0,
    <3, 4, 5>, 1,

..etc.
```

PR #5461, opened 2024-11-29
Sorry,  reopening Fileformat mapper

gsohler:
Found yet more duplicate export code which can be cancelled


src/gui/MainWindow.cc
```diff
@@ -286,13 +251,21 @@ void removeExportActions(QToolBar *toolbar, QAction *action) {
...
-  if (export2D) {
-    toolbar->insertAction(action, export2D);
+  const std::string format2d = QString::fromStdString(Settings::Settings::toolbarExport2D.value()).toStdString();
+  const std::string format3d = QString::fromStdString(Settings::Settings::toolbarExport3D.value()).toStdString();
+
+  for(auto format : {format2d, format3d}) {
+    for(const FileFormat &ff: fileformat::all()) {
+      const FileFormatInfo & info = fileformat::info(ff);
+      int l = format.size();
+      if (info.description.size() >= l && info.description.substr(0,l) == format){
```
kintel, line 261: This is the reason `fileformat::fromIdentifier()` exists. The `toolbarExport2D` and `-3D` should really be mappings from descriptions to identifiers, similarly to how `SettingsEntryEnum Settings::localSlicerFileFormat` is set up.

As an added bonus, once we get rid of the `STL` description, we can make the Preferences combobox use ASCII STL vs. Binary STL too, and we can remove the extra checkbox.

src/gui/MainWindow.h
```diff
@@ -310,6 +310,7 @@ private slots:
 
   QList<double> getTranslation() const;
   QList<double> getRotation() const;
+  std::unordered_map<QAction*, FileFormat>  export_map;
```
kintel, line 313: I'm wondering if this map should be reversed, to allow us to look up an action from a file format.

src/gui/MainWindow.cc
gsohler: sorry, could not see a way how to use fileformat::fromIdentifier for that.
identifier differs in case from e.g. "OBJ"  and even in length e.g. "binstl" vs "STL"

no way i could get "binstl" from "STL"

kintel: It will be a bigger change. Let’s merge this first and I’ll open another PR

kintel (approved):

PR #5465, opened 2024-12-01
More export action refactoring

### LINKED ISSUE #5466: Export of STL-Files does not work anymore on macOS 15.1.1 since version 2024-12-01
**Describe the error**
Since version 2024-12-01 git 1ad142a44 it is not possible to export STL-files anymore. The STL-Button in the top bar of the editor does not react anymore, the menu „export STL“ does not work anymore, no matter if the code was rendered before or not, no reaction at all.

This is really a serious bug, which should be solved asap.

**To reproduce the error**
1. start OpenSCAD and open a `.scad` file
2. click „render“
3. click „export STL“
–> no reaction at all.

**Environment and version information**
 - macOS Sequoia 15.1.1
 - MacBook Pro 16’ (2021) with Apple M1 Max, 64GB
 - OpenSCAD version 2024-12-01 (git 1ad142a44)

kintel:
This also moves from having ASCII STL export a separate advanced preference setting into having separate menu entries for ASCII vs. binary STL.

Fixes #5466

@gsohler: FYI

### DIFF
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -256,9 +256,9 @@ void addExportActions(const MainWindow *mainWindow, QToolBar *toolbar, QAction *
                                         Settings::Settings::toolbarExport2D.value()}) {
     FileFormat format;
     fileformat::fromIdentifier(identifier, format);
-    const auto it=mainWindow->export_map.find(format);	      
+    const auto it=mainWindow->export_map.find(format);
     // FIXME: Allow turning off the toolbar entry?
-   	if (it != mainWindow->export_map.end()) {
+    if (it != mainWindow->export_map.end()) {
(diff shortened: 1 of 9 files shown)


gsohler:
Thank you!
Actually i am Interested, how the Error happened ...

Marius Kintel ***@***.***> schrieb am So., 1. Dez. 2024,
15:55:

kintel:
In Qt6, signalmapper no longer sends the mapped(int) signal. You have to use mappedInt(int)

AlverezYari:
Any idea when this will make it out to to Homebrew snapshot?

kintel:
It will make it to our download page (https://openscad.org/downloads.html) tonight as 2024.12.02. Then I expect Homebrewt to pick it up less than 24h later.

Issue #5380, opened 2024-10-17
Exporting binary STL to stdout produces corrupt data


tomasf:
## Description
When exporting binary STL to stdout, the triangle count field of the resulting data is set to zero.

The reason seems to be that OpenSCAD first writes the header and a placeholder triangle count before writing the triangle data with a function that returns the number of triangles. Then, seek is used to step back to the triangle count offset and write the actual count. That works fine for files, but you can't seek like that in a stream like stdout.

https://github.com/openscad/openscad/blob/ef6c55e069901244b92c788e391dd2a29930b065/src/io/export_stl.cc#L309-L313

## Reproduce
`openscad -o - --export-format binstl whatever.scad > corrupted.stl`

## System
 - OS: macOS
 - System: MacBook Pro M1 Max
 - OpenSCAD Version: 2024.8

PR #5470, opened 2024-12-03
Resolved Issues With Corrupted STL Files When Exporting to stdout

IamLRBA:
@tomasf
@kintel

## Intro

As per issue [#5380](https://github.com/openscad/openscad/issues/5380), this PR includes changes that are aimed at improving the `export_stl` function to address issues with exporting binary STL files to non-seekable output streams (e.g., `stdout`).  I made some resaerch into the issue and from the description and external sources learned that the problem revolved around attempting to write a binary STL to `stdout` while relying on seek operations to update the triangle count. Since seeking isn't supported on a non-seekable stream like `stdout`, the previous approach led to corrupted STL files. With that in mind, my main updates include:

### Buffered Writing for Binary STL
- I introduced a `std::ostringstream` buffer to temporarily store the binary STL data.
- The buffer allows updating the triangle count at the correct offset (80 bytes) without relying on seek operations on the `output` stream.

### Header and Placeholder Management
- I made sure it writes a placeholder for the triangle count immediately after the header.
- I made sure too that it updates the actual triangle count in the buffer after calculating it.

### Logging for Exceeded Triangle Count
- I added a check to log an error if the triangle count exceeds the maximum allowable value for a 32-bit unsigned integer (4,294,967,295).

### ASCII Mode Handling
- I ensured ASCII STL data is directly written to the output stream, maintaining the existing behavior.

### Compatibility with `stdout`
- I made it ensure that binary STL files written to `stdout` or other non-seekable streams are consistent and valid by fully preparing the output in the buffer before flushing it to the stream.

Please have a review and provide further guidance on what I need to do.
Thank You!


IamLRBA:
> Could take a look at what it would take to add automated tests for stdio export?

@kintel I did some digging, and in order to add automated tests for the `stdout` export functionality, the best suggestion would be setting up a framework where the program's output can be captured and verified. For OpenSCAD, where binary and ASCII STL exports to `stdout` are involved, the key steps i could think of might include:

1. Identifying the Test Framework.
2. Capturing and Verifying `stdout` Output: I think that any automated tests to solve the issue should - Redirect `stdout` to a temporary file, execute the export command for binary and ASCII formats and verify the content of the captured output.

3. Adding Test Cases for Different Scenarios: such as Binary STL export to stdout, ASCII STL export to stdout, large models with many triangles to ensure no memory issues and Edge cases like zero triangles or maximum triangle counts.

That's all I can think of, though I am handicapped as most likely the test script would have to be written in Python or Bash as the best options or another supported scripting language. And I am not yet well acquainted with those languages.

kintel:
FYI: Added tests, mostly as an example of how this is done in OpenSCAD, in case you plan on picking up more issues.

kintel (approved):

Issue #5467, opened 2024-12-02
Make running `--animate` on multiple cores ?


hzeller:
I like using openscad in batch mode to create animations using the [`--animate` option](https://github.com/hzeller/hot-snipper/blob/master/Makefile#L61-L66)

Running this only create a single thread building all the frames, which can take a while. Machines typically have many cores these days so running that in parallel would be neat.

I see two possible solutions; the first might be a bit more involved, but the second should be fairly trivial:

  1. Provide an option `--animation_threads` of sorts, that tells openscad to run multiple renderings in parallel for different frames. Since each frame is entirely state-less, it can be parallelized like that and would be the 'neatest', but might be a bit involved as it could uncover some subtle multi-threading issues (e.g. are all backend-libraries thread-safe ?)
  2. Provide a 'range' option. Next to, say, `--animate=100` , the user can provide a range of frames to be produced, e.g. `--animate_frame_range=50-99` to tell openscad to only render the given range. Then the user can invoke multiple independent openscad instances with different ranges (`--animate_frame_range=0-49` and `--animate_frame_range=50-99`) in scripts or Makefiles. This should be fairly  easy to implement (Alternatively, consider some other sharding functionality, but range seems to be simplest).
  
With the second option in place, we could even simulate the first solution by spawning multiple processes from within openscad itself providing different range options (though implementing platform-independent spawning might be annoying as not every operating system has a neat `posix_spawn()` or even `fork()`).

Anything I am missing ? I might consider test-implementing (2).

kintel:
The second options is safe, but won't benefit from any caching, which could limit its usefulness, depending on what kind of designs are being animated.

The first option it tricky for various reasons, but yeah, since we essentially parse-eval-export on unique data structures, we "only" need to be thread compatible everywhere + thread-safe caches.

A third option would be to work on optimizing re-evaluations, as we currently parse the source code from scratch for every frame.

hzeller:
To see where time is spent, created a quick flame-graph (runing on a linux `perf` output, using [`pprof`](https://github.com/google/pprof) `-http=:9000` with the [`perf_data_converter`](https://github.com/google/perf_data_converter) installed), it seems the most cycles are used at `VertexArray::createVertex` and in particular using the shared pointers and all their associated costs (could this be put in an arena instead ?).
Is this part used in the re-evaluations that you think might be worth optimizing ?

Also `encodeLZ77` seems to also use significant time (generating PNGs ?)

[...]

PR #5472, opened 2024-12-03
Allow to do --animate work in shards.

hzeller:
Introduce new `--animate_sharding=<num_shards>,<shard>` flag that allows to output a subset of animation frames given simple modulo sharding: Divide the work in `num_shards`, and only output frames that match `frame# % num_shards == shard`.
So an invocation only outputs `frames/num_shards` reducing wall-time for that process.

This allows an easy parallelization on a machine with multiple cores, or even distribute the work to multiple machines.

A typical use could be like this: run two instances and divide the work into two shards; have one output all the even frames and the other the odd frames:

```bash
   openscad -q --animate=42 --animate_sharding=2,0 foo.scad &
   openscad -q --animate=42 --animate_sharding=2,1 foo.scad &
   wait
```

(of course, one would use probably more shards closer to the number of cores on the machine).

Issues: #5467

### DIFF
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -241,6 +241,57 @@ void localization_init() {
   }
 }
 
+struct AnimateArgs {
+  unsigned frames = 0;
+  unsigned num_shards = 1;
+  unsigned shard = 0;
+};
+
+struct CommandLine
+{
+  const bool is_stdin;
+  const std::string& filename;
+  const bool is_stdout;
+  std::string output_file;
+  const fs::path& original_path;
+  const std::string& parameterFile;
+  const std::string& setName;
+  const ViewOptions& viewOptions;
+  const Camera& camera;
+  const boost::optional<FileFormat> export_format;
+  const AnimateArgs animate;
+  const std::vector<std::string> summaryOptions;
+  const std::string summaryFile;
+};
+
+AnimateArgs get_animate(const po::variables_map& vm) {
+  AnimateArgs animate;
+  if (vm.count("animate")) {
+    animate.frames = vm["animate"].as<unsigned>();
+  }
+  if (vm.count("animate_sharding")) {
+    std::vector<std::string> strs;
+    boost::split(strs, vm["animate_sharding"].as<std::string>(),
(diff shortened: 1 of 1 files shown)


hzeller:
(if accepted, I probably also should that to the upcoming documentation ... where do I find that ?)

hzeller:
example use https://github.com/hzeller/hot-snipper/blob/master/Makefile#L66-L73

kintel:
I think the only (and somewhat sparse) command-line docs are at https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/Using_OpenSCAD_in_a_command_line_environment

src/openscad.cc
```diff
@@ -572,13 +606,16 @@ int cmdline(const CommandLine& cmd)
...
-  if (cmd.animate_frames == 0) {
+  if (cmd.animate.frames == 0) {
     render_variables.time = 0;
     return do_export(cmd, render_variables, export_format, root_file);
   } else {
     // export the requested number of animated frames
-    for (unsigned frame = 0; frame < cmd.animate_frames; ++frame) {
-      render_variables.time = frame * (1.0 / cmd.animate_frames);
+    for (unsigned frame = 0; frame < cmd.animate.frames; ++frame) {
+      if (frame % cmd.animate.num_shards != cmd.animate.shard) {
```
kintel, line 615: Interesting approach to sharding. Intuitively, I'd think the most appropriate sharding approach would be to evaluate consecutive frames on the same node, as that may yield more coherent geometry and potentially more cache hits.
Did you consider the different approaches?

hzeller: No particular reason. In absence of other considerations (and I didn't know if caches made a big impact), I just defaulted to the default mod sharding as it usually allows for the nicest distribution with the least likelyhood to have some straggler shards.

I've played with range sharding now and it tends to be slightly better (only measured on my laptop, so results are probably more noisy). Maybe that is indeed the right approach, and it might also be easier to understand for the casual user. Will change it to range sharding.

hzeller:
(I messed up my local branch, will send a separate pull request)

PR #5473, opened 2024-12-04
Allow to do --animate work in shards.

hzeller:
Introduce new `--animate_sharding=<shard>/<num_shards>` flag that
allows to output a subset of animation frames.
e.g. `--animate_sharding=2/5` only outputs the second 1/5 of all shards.

This allows an easy parallelization on a machine with multiple cores,
or even distribute the work to multiple machines.

A typical use could be like this: run two instances and divide
the work into two shards that output the first and second half
of the frames.

```bash
   openscad -q --animate=42 --animate_sharding=1/2 foo.scad &
   openscad -q --animate=42 --animate_sharding=2/2 foo.scad &
   wait
```
    
(of course, one would use probably more shards closer to the
number of cores on the machine).

Issues: #5467

### DIFF
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -241,6 +241,57 @@ void localization_init() {
   }
 }
 
+struct AnimateArgs {
+  unsigned frames = 0;
+  unsigned num_shards = 1;
+  unsigned shard = 1;
+};
+
+struct CommandLine
+{
+  const bool is_stdin;
+  const std::string& filename;
+  const bool is_stdout;
+  std::string output_file;
+  const fs::path& original_path;
+  const std::string& parameterFile;
+  const std::string& setName;
+  const ViewOptions& viewOptions;
+  const Camera& camera;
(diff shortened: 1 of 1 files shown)


hzeller:
(redo of #5472 as I messed up my local branch there)

hzeller:
Difference to the previous PR

  * Sharding is now by range, not mod-sharding. Might make it easier to re-use caches, and might be easier to understand in general
  * shard numbers are not zero-based anymore but 1 based, which makes for a more natural way of describing it and less headache when creating a shell-script loop. Also written with slash; `2/5` looks like second of 5.

kintel (approved):

kintel:
@t-paul LGTM, but since you've worked with animation a lot more than I, do you have any comments on this?

kintel:
Side note: We should really test this, but I don't think we even have tests for the existing animate option. Out of scope for this PR, but Something to consider.

PR #5476, opened 2024-12-05
Import parts from PartCAD

EMPanisset:
Enable importing other OpenSCAD parts as well as CadQuery, build123d or any other part available through PartCAD (either in the [public PartCAD repository](https://partcad.org/repository) or in your own private repositories).
No BOSL or BOSL2 in PartCAD yet, but it is likely a few hours of effort to add them there.


t-paul:
How does that interact with the ongoing (well unfortunately a bit stalled due to lack of spare time) effort to include Python support #4841 by @gsohler. I don't think we want to have 2 different ways of calling into python.

Also why do we need to even link to python for importing files?

clairbee:
I would not say they have much in common.
Even the fact that they both use Python (thus you can probably benefit from some of this code)
is a temporary coincidence: PartCAD client will get refactored soon and
OpenSCAD will begin talking JSON RPC to a subprocess instead of instantiating a Python interpreter.

If you are invoking a local python script, a script you placed and versioned yourself, then it might make sense to access it directly without using PartCAD to skip any overhead.
However if you are trying to use products of other people's work, or simply reusable components (that have their own versioning), they you should probably use packaged modules.

All code gets packaged and maintained as a package when there is certain amount of code and modularity becomes a necessity. PartCAD is a package manager for code-CAD.

EMPanisset:
We may however want to unify the way the Python Interpreter is initialized. This is the common part between PartCAD and the implementation referred by @t-paul.
For instance Py_InitializeFromConfig is called by the PartCAD we introduced and also by the implementation referred by @t-paul. After interpreter is initialized the implementations can proceed independently with loading their own modules and accessing the embed Python Interpreter ( I believe we can even make it possible the implementations get independent [Pyhton SubInterpreters](https://tonybaloney.github.io/posts/sub-interpreter-web-workers.html) with their own GIL -Global Interpreter Lock).
I can make the PythonInterpreter class available there for common use.

gsohler:
I appreciate the Python interpreter class. Escpecially the versatile deleter class has great value.
The pr @t-paul is refering to, is little outdated.(but showing target direction)

Latest version is in github.com/gsohler/openscad , branch python
This codebase also uses python subinterpreters in separate threads and naturally needs GIL-locks
(in CSG-Worker, in CGAL-Worker).
I am happy to work together using a common Python Interpreter class.
( i believe, that python for partcad is needed at all to be ableto  python import the partcad module, right ?), interface is STL ?

kintel:
This sounds like the start of a long discussion, so running the risk of making it longer: If this is a direction in which we want to go, I wonder what the best import format would be. STL sounds like the least common denominator, but rarely the best solution. e.g. PartCAD objects designed in OpenSCAD could benefit from being imported as OpenSCAD, parts with colors or multiple materials could use one of the formats supporting colors etc.. Allowing users to choose could be fun, but requires people to know what the options are.
An alternative approach to to importing directly coula also be to look into whether this would fit with `olman` (https://github.com/openscad/openscad-library-manager).

partcad-admin:
Well, now this sounds like a start of two conversations, and they don't need to be long... :)
1) We can start iterating on the format of data received by OpenSCAD. It doesn't have to be STL. We can agree on something more native to OpenSCAD and we can try converting to it directly (for example, from CadQuery to OpenSCAD without any intermediate state like STL). However we will support a very long list of formats down the road and we won't always have a direct conversion mechanism into OpenSCAD's internal representation. So there will be some kind of a default intermediate representation for all corner cases. From mesh formats, we have 3MF and OBJ. We may add others. Please, help us understand what's the target representation and what is the preferred intermediate representation, and we'll start working towards that incrementally. This should not block the proposed integration if that's sufficient quality for current PartCAD users. It's better than not being able to import at all. Nothing changes for those who don't use PartCAD.
2) I'm not sure what olman is as there doesn't seem to be much documentation or functionality right now. But, judging by its name it sounds like a package manager for importing openscad into openscad. It can certainly be a more lightweight and simpler alternative to PartCAD if you only need OpenSCAD models. There is nothing wrong with having a solution like that. But it's not the same thing as PartCAD which is a package manager for all CAD models, code-CAD or not. And even if it would be the same thing, why not to have both? So why would we not continue the olman discussions in parallel with the deployment of PartCAD?

clairbee:
I thought of another aspect which is not directly related to this PR but very much related to olman. The aspect of exporting. How would we want OpenSCAD files to be packaged for external consumption. Do we want them to be packaged as PartCAD packages or olman packages? That's a more interesting question even though it's not really related to importing at all. I don't know what the olman features are, but in theory it can become the default way to packages OpenSCAD, and PartCAD will build a bridge to import olman packages using "/pub/repo/olman/<olman-package-name>". This means a lot less users to support for PartCAD which I personally like a lot.

EMPanisset:
I agree with @partcad-admin on separating the conversations and focus this PR on the minimum requirements to make it possible import parts from PartCAD to OpenSCAD.
Currently we are using PartCAD capabilities to import the part and convert it to an intermediary STL format (importable by OpenSCAD). Although it does not have to be this way (and it is certainly not ideal for all the cases), it seems to be a good start. We will improve on that incrementally to make it more efficient for different formats we want to support.
Thanks @gsohler for your comments and suggestions. I will take a closer look at the latest python branch you mentioned (in github.com/gsohler/openscad , branch python) which makes use of Python SubInterpreters and then I can come up with the common PythonInterpreter class design supporting them too.

alexanderilyin:
@t-paul @gsohler @kintel I'm helping with some build failures on the PR and while I'm on this step:

> *Note that this process can take several hours, and tens of gigabytes of disk space*

Are there by any chance read-only [ccache(1) - Remote storage backends]() I can use?

kintel:
@alexanderilyin Did you check out https://github.com/openscad/openscad/blob/master/.circleci/config.yml ?

It uses the docker image openscad/mxe-x86_64-gui:latest, which should have most/all dependencies taken care of

alexanderilyin:
Yep, I created Dev Container from `openscad/mxe-x86_64-gui:latest` but then I started to follow build steps [Building for Windows](https://github.com/openscad/openscad#building-for-windows), so probably I could have skipped script with building deps, but it's running for last 4+ hours and fingers crossed it will finish building deps by the morning:

image

Probably it's my bad and I should have taken closer look on the steps used in `[.circleci/config.yml](https://github.com/openscad/openscad/blob/master/.circleci/config.yml), but thanks for the input.

alexanderilyin:
Last large piece was qt:

```
[done]        qtdeclarative           x86_64-w64-mingw32.static.posix                535188 KiB     35m5.489s
```

After that I run this:

```
root@529a57dc5c1b:/workspaces/openscad# time ./scripts/release-common.sh mingw64

-- Checking for module 'tbb'
--   Package tbb was not found in the pkg-config search path.
Perhaps you should add the directory containing `tbb.pc'
to the PKG_CONFIG_PATH environment variable
Package 'tbb', required by 'world', not found
CMake Error at /root/openscad_deps/mxe/usr/x86_64-pc-linux-gnu/share/cmake-3.27/Modules/FindPkgConfig.cmake:607 (message):
  A required package was not found
Call Stack (most recent call first):
  /root/openscad_deps/mxe/usr/x86_64-pc-linux-gnu/share/cmake-3.27/Modules/FindPkgConfig.cmake:829 (_pkg_check_modules_internal)
  CMakeLists.txt:938 (pkg_check_modules)


-- Configuring incomplete, errors occurred!

real    0m7.899s
user    0m0.995s
sys     0m0.710s
```

So probably I should not follow `Building for Windows` steps and use CircleCI as source of truth for building env / commands.

kintel:
As long as you're building your own docker images, you can alway layer on top of the ones we use for CircleCI. As with all docker interactions, you have to be careful and manage your images appropriately. The few times I needed to debug locally, I also bit the bullet and built everything from source.

alexanderilyin:
@kintel thanks for the input. Turns out I was having Layer 8 problem. Reproduce is actually takes seconds if done right. My problem was mostly caused by `./build/` contents after run with wrong params. Here is the snippet which works like a charm after [`To debug the jobs more interactively, you can go the manual route...`](https://github.com/openscad/openscad#running-ci-workflows-locally)

```bash
export CIRCLE_BRANCH=$(git branch --show-current)
export CIRCLE_BUILD_NUM=1597463007
yq '.jobs.openscad-mxe-64bit.steps[].run | select(. != null) | .command' .circleci/config.yml | bash -euxo pipefail -

```

PR #5508, opened 2024-12-20
Update mingw-x-build-dependencies.sh

gsohler:
Shall we sync the clipper2.mk files ?


kintel:
When using clipper2 from msys, we should also pass -DUSE_BUILTIN_CLIPPER=OFF to make in the appropriate build script.

gsohler:
yes, i can try that, things might become much easier then. thank you for pointing out

PR #5509, opened 2024-12-20
Fix libsvg build with external Clipper2

cjmayo:
There's an identical include in src/core/OffsetNode.cc so should be OK for everyone - let's see...


kintel (approved):
Thanks, this slipped through the cracks..

PR #5526, opened 2024-12-23
3mf color import / export

ochafik:
Follow up to https://github.com/openscad/openscad/pull/5185

I was initially planning to rely on assimp for all color import/export beyond OFF format, but I still have 3 patches to upstream for it to work (see https://github.com/openscad/openscad/pull/5180), so updating our 3mf import/export code is the path of least resistance atm.

Note that this PR does NOT include multimaterial support for PrusaSlicer & BambuStudio, which needs a non-standard "triangle" attribute (cf. playground's [materialize](https://github.com/openscad/openscad-playground/compare/main...ochafik:openscad-playground:color-assimp3#diff-4aa3f4c63c5170faa373ef7161c9e870676ee4f9789a26d9b0c84889203af4cd) code).

TODO:
- [ ] Tests still need a bit of love.

### DIFF
--- a/src/core/parsersettings.cc
+++ b/src/core/parsersettings.cc
@@ -87,10 +87,18 @@ inline fs::path find_valid_path_(const fs::path& sourcepath,
                           const std::vector<std::string> *openfilenames)
 {
   if (localpath.is_absolute()) {
-    if (check_valid(localpath, openfilenames)) return fs::canonical(localpath);
+    if (check_valid(localpath, openfilenames)) {
+#ifndef __EMSCRIPTEN__
+      return fs::canonical(localpath);
+#else
+      return localpath;
+#endif
+    }
   } else {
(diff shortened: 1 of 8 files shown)


t-paul:
Ah, nice, let me push my v1 changes, now that the advent calendar concluded. This also has major fixes on reading (nested objects + transformation matrices) that likely need to be added to v2 too.

kintel:
Small note: PolySetBuilder is usually the wrong solution as it doesn't support preserving distinct vertices with the same position (for manifold topology corner cases).
That said, it's probably fine to merge this first, then do a green refactoring to port off PolySetBuilder later on, then add in the corner case tests.

t-paul:
Added the changes from here to #5527.

ochafik:
Superseded by https://github.com/openscad/openscad/pull/5527

PR #5529, opened 2024-12-24
Remove fast-csg

kintel:
This removes fast-csg. The primary reason is that it adds too much cognitive load to refactoring tasks, keeping the special case code paths alive and tested.
fast-csg has already been disabled for almost 3 months with no issues.

@ochafik Let me know if you have any objections!


ochafik:
@kintel good to go, thanks!!

image

The hybrid was full of flaws but it had a nice life, kept me company during harsh cold lockdown winters, got us all [excited to do more](https://ochafik.com/jekyll/update/2022/02/09/openscad-fast-csg-contibution.html) with OpenSCAD. I’m happy we moved on to exciting Manifold territory, doesn’t look like we’ll have to look back 🥳.

RIP, friend ⚙️🤖🩵

kintel:
Thanks @ochafik for not only spearheading the fast-csg stuff, but also pushing the boundaries further by integrating Manifold, keeping the WASM playground alive and more!

PR #5534, opened 2024-12-27
Update icon-nightly.scad (moon color)

UBaer21:
more yellowish for better contrast on gray window background

### DIFF
--- a/resources/icons/icon-nightly.scad
+++ b/resources/icons/icon-nightly.scad
@@ -1,8 +1,8 @@
 RES=100;
 Logo(50, RES);
-colorValueR=246/255;
-colorValueG=241/255;
-colorValueB=200/255;
+colorValueR=255/255;
+colorValueG=250/255;
+colorValueB=100/255;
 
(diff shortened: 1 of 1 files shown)


kintel:
IMO, it doesn't look very good unless we do something about the stars too

UBaer21:
Making them the same color will have no contrast to the moon
image

it also seems that the exported logo that is used has no transparency on the surface.
image

here a "SlateGrey" variant
image

kintel:
The original was rather quick&dirty. I wonder if it would look better with the stars were yellow, and embedded into the surface of the sphere rather than subtracted. ..or perhaps slightly embossed?
Transparency rendering is currently not great, so we may need to play around to make the rendering look decent

UBaer21:
image

here in gold on surface

UBaer21:
closing and changing to
image

jordanbrown0:
I never knew about the moon and stars, because the largest I ever see this icon is about 30x30 pixels.  They are quite cool, but are they really useful at icon sizes?

PR #5538, opened 2024-12-27
update icon-nightly.scad colors

UBaer21:
moon and stars gold for better contrast.

TODO before merging:
* [ ] Render a high resolution PNG
* [ ] Go through all existing renderings, validate that they're all in use and remove any unused files
* [ ] Update all existing renderings


t-paul:
What, noooo. Note to self... find some time to integrate it.

PR #5552, opened 2024-12-30
Catch failure to parse font

### LINKED ISSUE #5536: text(..., font=":charset=junk") crashes OpenSCAD
**Describe the bug**
Setting "charset" in a font specification crashes OpenSCAD.

**To Reproduce**
```
text("hello", font=":charset=foo");
```
F5, OpenSCAD exits.
Note that the colon is a critical part of the specification.

**Expected behavior**
Doesn't exit :-)

I believe that fontconfig's usual behavior would be to ignore an unknown specifications.

I'm not sure from the fontconfig docs exactly what this property does, but exiting isn't it.

 - OS: Windows 11
 - System: 13th Gen Intel(R) Core(TM) i7-13700   2.10 GHz
 - OpenSCAD Version: 2024.12.26, development snapshot (and in 2021.01)

fontconfig version: 2.14.2

**Additional context**
Also fails for `textmetrics()` and `fontmetrics()`.

kintel:
Also adds charset= test

Fixes #5536


kintel:
Meh, of course it won't pick up the same font across platforms. Perhaps this test case won't fly..

jordanbrown0:
Hmm.  Yuck.  But at least you can retain the negative test.  You could also have a positive test, that it successfully parses a valid entry; set the font to `Liberation Sans:charset=41`.  That won't test that it actually matches based on that code point, but it'll test the parsing.

One could imagine creating a sterile environment, with only a controlled set of fonts available, and controlling the set of fonts so that particular code points are available only in particular fonts, and using that to test.  But that seems like way too much work for the benefit; our job is to test our interface to fontconfig, not to test fontconfig itself.

Issue #3972, opened 2021-11-10
rotate_extrude misaligned for angle=360


UBaer21:
the corner is on -x  while a circle is on +x 
when the angle is reduced it is on +x
```
  rotate_extrude($fn=5)square();//180° wrong
  
  rotate_extrude($fn=6,angle=359.9)square();
  circle(1,$fn=5);
```
![image](https://user-images.githubusercontent.com/93376232/141101035-7478e063-16dc-444b-85bd-3c862b0b99e8.png)

resulting in issues when you rotate_extrude something around a cylinder.

t-paul:
I think that's intentional for backward compatibility, see discussion in https://github.com/openscad/openscad/pull/1478

UBaer21:
how can this be compatible to anything if All Objects center their point on +x  (sphere, circle, cylinder)?
And as soon an angle <360 is used it is jumping 180° into the right position.
As $fa is limited to 72 (not or 120) we end up with 5-gon elements which are not matching.

t-paul:
It probably should depend on the `angle` parameter having a value or not. The compatibility would be for scripts written for older OpenSCAD versions that don't have the `angle` parameter at all.

UBaer21:
But it doesn't make a difference if angle has a value or if the value is 360 it is always wrong.
Only when angle gets <360 and the loop opens it has it start point correctly on [+X , 0].
So older scripts may have corrected this by an additional rotation of 180 - but this would mean we can never correct the problem and new scripts will also need to correct for this. 
So at least the polyhedron need to be turned by 180 when angle == 360 (or 0).

thehans:
Duplicate of #1886

thehans:
Also see issue #3142 for backwards compatibility discussion.

Issue #1886, opened 2016-12-07
rotate_extrude and cylinders with $fn have sides which don't align


neiljp:
I couldn't find a reference to a bug similar to this. I'm running a nightly debian build 20161201T231538.git5192388-1734.1, after suggestion to do so to resolve a dxf/svg import issue via irc.

Having both a cylinder and doing a rotate_extrude on a 2D shape, each with $fn set to finite values, leads to edges (sides) which don't align properly due to different rotational offsets, as illustrated by the following:
```
coin_diameter=40;
coin_thickness=2;
ring_height=1;
slope_inset=1.5; // non-zero values add slope inside ring (eating into text_padding)
sides=7; // or 0 for default (~round)
side_rotation=90;
width_to_fit_text=35;

rotate_extrude_extra_angle=180/sides; // Necessary to ensure sides>0 has edges which align
// =========================================================================
rotate([0,0,side_rotation])
difference()
{
  cylinder(r=coin_diameter/2,h=ring_height+coin_thickness,$fn=sides);
  translate([0,0,coin_thickness]) 
    cylinder(r=width_to_fit_text/2,h=ring_height+1,$fn=sides);
}

if (slope_inset>0)
{
  overlap=1;
  rotate([0,0,side_rotation+rotate_extrude_extra_angle])
  rotate_extrude($fn=sides)
  polygon(points=[[width_to_fit_text/2-slope_inset,coin_thickness],
                  [width_to_fit_text/2,ring_height+coin_thickness],
                  [overlap+width_to_fit_text/2,ring_height+coin_thickness], // into ring
                  [width_to_fit_text/2,coin_thickness]]);
}
```

thehans:
Yes this is a particular inconsistency in the way that rotate_extrude works.  

In early code, before rotate_extrude had an angle parameter, it was implemented such that it started drawing the points from the negative x-axis, instead of positive x.  IMO, this behavior was a mistake from the start, but has been purposely kept for legacy support (so people's older scripts don't break).  So its actually rotated a complete 180 degrees, regardless of the number of sides;  and rotate_extrude_extra_angle=180 would also have worked in your example.

However, if an angle parameter other than 360 is given, the newer angle functionality starts the first point at the more logical positive x-axis, (where angle=0 by convention for cartesian system).
So you can potentially do a kludge like:  rotate_extrude($fa=359.999/sides, angle=359.999)  
and it would appear how you want without any rotate_extrude_extra_angle

The related code is here in case you are curious:
[...]

PR #5553, opened 2024-12-31
Start rotate_extrude on -X only for no angle specified.

jordanbrown0:
Start on +X when an angle is specified, including 360.

Ref #1886, #3972.

Existing behavior is that when the angle parameter is not specified or is bad (undef, NaN, et cetera) it defaults to 360, and that when the angle parameter is 360 (whether by default or explicitly) the extrusion starts along the -X axis.  All *other* values for angle, and circles and cylinders, have the extrusion start along +X.  (The difference is in whether an odd-number-of-sides shape has a point on the left and flat on the right, or vice versa.)

The behavior of rotate_extrude with no angle specified is, unfortunately, pretty much nailed down by backward compatibility concerns.  I say "unfortunately", because it requires workaround to have a rotate_extrude align with, say, a cylinder.  However, rotate_extrude *with* an angle supplied is more recent, and the explicit angle=360 case is very much a special case; a caller who wants 360 degrees is more likely to omit the parameter entirely.

This change keeps the unspecified-angle case starting on -X, but changes the angle=360 case so that it starts on +X, consistent with other angles and with circles and cylinders.

One subtle effect:  when exporting to CSG, current behavior emits an explicit "angle=360" for both the unspecified-angle case and the angle=360 case.  That would mean that CSG couldn't distinguish the two cases.  This change drops the angle parameter from the CSG output when it wasn't specified in the OpenSCAD input, so that CSG executed as OpenSCAD input behaves as expected.


jordanbrown0:
My local test runs show no differences in the images created.  I changed the CSG-dump tests to remove the explicit "angle=" in those cases where the OpenSCAD input did not specify an angle.

I wrote a specialty test program to examine the models resulting for several interesting cases.

Here's the old output for various angles and whether the 2D shape is in +X or -X:
Rlh5rmC7F03nEKVU

Here's the new output:
PMtcfSrh8A9gkNrT

Note that they differ only for explicit 360 and -360.

jordanbrown0:
Still narrower case:  explicit angles greater than 360 or less than -360 are clamped to ±360.  In the current release, their extrusion is then started on -X.  With this change they are still clamped to ±360, but are considered to be "explicit angle" cases and so start on +X.

The only cases that start on -X are
* No angle specified.
* Angle specified cannot be converted to a finite number.

jordanbrown0:
Here's the test program mentioned above (but with an additional case for angle>360):
```
cases = [
    [ undef, 1 ],
    [ undef, -1 ],
    [ 360, 1 ],
    [ 360, -1],
    [ -360, 1 ],
    [ -360, -1],
    [ 300, 1 ],
    [ 300, -1 ],
    [ -300, 1 ],
    [ -300, -1 ],
    [ 1360, 1 ],
    [ 1360, -1],
];

for (i = [ 0: len(cases)-1 ])
    translate([i*4,0,0]) {
        angle = cases[i][0];
        center = cases[i][1];
        rotate_extrude(angle=angle)
            translate([center,0])
                square(center=true);
        color("black") linear_extrude(height=0.001) {
            translate([0,3]) text(str(angle), size=0.5, halign="center");
            translate([0,2]) text(center < 0 ? "-X" : "+X", size=0.5, halign="center");
        }
    }
```

jordanbrown0:
Another possibility for CSG would be to emit `angle=undef`, but that seems more likely to cause problems for hypothetical non-OpenSCAD consumers.

jordanbrown0:
It would be nice to deprecate the behavior for no-angle-specified, so that we could eventually flip it around and have everything consistent.  However, I'm not really sure how to do that without having the warnings be unreasonably intrusive - for every person who is depending on the start-on-negative-X behavior, there are probably ten who aren't.  Maybe the answer would be to warn only on no angle supplied *and* $fn supplied, which is the people most likely to be depending on the behavior.

kintel:
I generally agree that this is a good thing, but the one thing I don't like is that dumping the file and reading it back yields different results. It would be really nice to be able to continue to rely on that behavior, even if it's just for internal consistency.

I'll think about it for a bit.

jordanbrown0:
> the one thing I don't like is that dumping the file and reading it back yields different results.

Hmm?  I have the generated CSG mirroring the input OpenSCAD.  If you don't specify an angle in the input OpenSCAD, it doesn't appear in the generated CSG, and so the results when you execute that generated CSG as OpenSCAD should be the same as for the original.

jordanbrown0:
A more general approach would be to add, to both OpenSCAD and the generated CSG, a "start" parameter that gives the angle of the start of the extrusion.  The default "start" would be zero, except if the arc-angle is not specified then the default "start" would be 180.  Such a mechanism would allow the CSG to not depend on the "magic" behavior of a missing "angle" parameter.

It would add a small amount of convenience, in that you could start your extrusion at whatever angle you needed, instead of having it always start at zero and needing to rotate it into place.  But that rotation is a very small cost.

jordanbrown0:
That is, `rotate_extrude()` would turn into `rotate_extrude(angle=360, start=180)`, while `rotate_extrude(angle=360)` would turn into `rotate_extrude(angle=360, start=0)`.

kintel:
Actually, specifying a start angle is not a bad idea. I still need to weight the pros/cons.
It's the special-case stuff that's bothering me the most; but I think it would be find if we could deprecate some legacy variants, and offer special case handling of that, and eventually kill it.
..basically aim for the future to be clean and accept temporary mess :)

jordanbrown0:
Yes, special cases are bad.  For the time being at least, we're stuck with the special case where not specifying "angle" causes the extrusion to start on -X.  The question here is whether that special case should extend to angle=360.

And yes, we should always be moving towards cleaner answers, even at a modest cost in compatibility.

kintel:
Some thoughts, structured as alternatives. Will let it simmer for a bit, but feel free to comment / edit the pros/cons:

**A: Simply change the default start angle to 0, as a potentially breaking change**
   **+** Cleanest solution
   **-** Breaks existing behavior 
   **-** Bad for people relying on existing behavior, and no trivial way to detect (would need to add a rotation() to reinstate existing behavior).

**B: Change the default start angle only when manually specifying angle=. Same idea as **A**, but only breaks for a subset of usages.**
   **-** Breaks existing behavior
   **+** ..but # of people effected is likely extremely small.
   **-** Bad for CSG export, as we cannot re-import without changing how CSG export works.
   **-** Uses conditional defaults, which is suboptimal

**C: Introduce start=, defaulting to 0**
   **+** Equally clean solution as **A**
   **-** Same negatives as *A*
   **+** .. but offers a trivial fix (start=180) for people affected.

**D: Introduce start= but default it to 180 if no parameters or angle=360**
   **-** Same negatives as *C*
   **+** Same positives as *C*
   **-** .. but uses conditional defaults, which is suboptimal

**E: Introduce start= but default it to 180 if no parameters.**
   **-** Same negatives as *D*
   **+** ..but # of people effected is likely extremely small.
   **+** Same positives as *D*
   **+** Slightly more structured than **D** as we'll only have a single special case (no parameters)

**F: Introduce start= and deprecate and warn on parameterless instantiation, but offer existing behavior until next release.**
   **+** Carves out a path for a clean future situation 
   **-** Bad for people who don't care about starting angle, as it would cause lots of unnecessary warnings

**G: Deprecate rotate_extrude() and start fresh**
   **+** Good for new sweep-related innovations
   **-** Hard to pull off before next release

jordanbrown0:
> A: Simply change the default start angle to 0, as a potentially breaking change

Agree.

> B: Change the default start angle only when manually specifying angle=. Same idea as A, but only breaks for a subset of usages.

This is the proposed behavior.  Note that it only changes the behavior when the angle is explicitly set to 360 or is explicitly set but clamped to 360.

Note that changing how CSG export works is only a couple of lines, and the proposed change does it.  However, it's ugly in that CSG normally fully populates all parameters.

> C: Introduce start=, defaulting to 0

Agree.  Also adds the "start" feature.  (Which is a very minor one, since it's equivalent to `rotate(start) ...`.)

> D: Introduce start= but default it to 180 if no parameters or angle=360

So start=0 is the way to get a full extrusion started on +X, which is the point of this work.
You say "same negatives as C", which recursively means "same negatives as A", which is "breaks existing behavior", but this option does *not* break existing behavior.  The only negative is that it leaves the unclean behavior that angle=359 and angle=360 start in different places.

> E: Introduce start= but default it to 180 if no parameters.

This is the same as B, but adds "start" for additional flexibility and to make the CSG picture clearer.

> F: Introduce start= and deprecate and warn on parameterless instantiation, but offer existing behavior until next release.

Lots and lots of unnecessary warnings.  Maybe warn on parameterless invocation with $fn set.  But even then, there's a lot of people who set $fn globally.  Maybe only warn when $fn is set to an odd number.

> G: Deprecate rotate_extrude() and start fresh

Even with general sweep, I think you'll want a simple rotational extrusion and a simple linear extrusion.  Those are common enough that they need the simplest possible invocation.  **Maybe** if you could say `extrude(circle(10)) ...` and `extrude(zline(10))` or something like that.

jordanbrown0:
H:  Introduce "start" and default to 180 if no parameters, as in E, but warn if no parameters and $fn odd, in preparation for a future transition to always defaulting to 0.

kintel:
Thanks for comments, I edited my overview to address the issues you pointed out and inserted your suggested **H** as **F** and renumbered the remaining items.

kintel:
My current evaluation:
* Try to avoid issuing warnings, hoping to clean up later. While ideal in the future, I've been thinking a bit about this: I don't have a good sense of what $fn typically ends up being, but my fear is that even odd $fn could cause a lot of false positive warnings.
* **H** might be a good long-term solution, but not for a few years
* My current vote would be **E: Introduce start= but default it to 180 if no parameters.**. My main concerns is that conditional defaults could make things less structured, but I haven't really thought it through.

jordanbrown0:
Note that any $fn proposal is only for an explicitly-specified $fn, not for the "effective" $fn that's calculated from $fa and $fs.  Many people will explicitly set $fn to 50 or 64 or 100 or 500, trying to get a high-precision circle.  How many will explicitly set it to an odd number while trying to generate a circle (vs a regular polygon)?\

Of course, even somebody who is trying to generate a triangle may not care about the alignment.

kintel:
You might be right :) If you otherwise agree, implementing **F: Introduce "start=" and default to 180 if no parameters, but warn if no parameters and $fn odd**, and evaluate the warning as we go, could be a good solution.

jordanbrown0:
Will do.

jordanbrown0:
That push adds `start` and the deprecation warning discussed.

Note:  I'm seeing two test failures:
```
The following tests FAILED:
        645 - previewtest_issue1105d (Failed)
        1609 - previewmanifoldtest_issue1105d (Failed)
```

but in looking at the images, it looks like the "expected" images have small flaws that this new version fixes.  (I don't immediately know *why* it fixes them, but I did clean up the math some.)

Assuming that the CI tests show the same failures, it would be good if somebody else would take a look at the images and opine.

kintel:
I see the same failure locally. The old goldens look like they have a convexity-related artifact. I say regenerate.

jordanbrown0:
The flaws are just convexity artifacts.  I don't know why they moved or disappeared, but it may be relevant that the old implementation did "full" extrusions clockwise, while I do all extrusions counter-clockwise.  The right fix is to bump the convexity up to 4 and fix the images.

tests/regression/dumptest-examples/example007-expected.csg
```diff
@@ -1,7 +1,7 @@
 multmatrix([[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 1, -10], [0, 0, 0, 1]]) {
 	group() {
 		difference() {
-			rotate_extrude(angle = 360, convexity = 3, $fn = 0, $fa = 12, $fs = 2) {
+			rotate_extrude(convexity = 3, $fn = 0, $fa = 12, $fs = 2) {
```
kintel, line 4: I think this needs updating (ctest -C Examples). Might as well run ctest -C All

jordanbrown0: Yeah, oops.  When I did the initial work (where I dropped angle= from the CSG output), I fixed the tests by grepping for rotate_extrude.  When I added start=, I fixed the tests by running them and fixing them... which didn't include the examples.  Running -C All now.

kintel (approved):

PR #5560, opened 2025-01-04
Centralize checking for children of modules that don't accept children.

jordanbrown0:
I got annoyed by how many times I saw the exact same "check for children and warn if there are any" sequence, so I centralized it.  The mechanism is kind of cool:  since the registration mechanism passes the builtin_* function to the constructor, it was easy to just have an alternate constructor that wanted a builtin_function that didn't have a children argument.

### DIFF
--- a/src/core/ImportNode.cc
+++ b/src/core/ImportNode.cc
@@ -56,13 +56,8 @@ namespace fs = std::filesystem;
 using namespace boost::assign; // bring 'operator+=()' into scope
 
 
-static std::shared_ptr<AbstractNode> do_import(const ModuleInstantiation *inst, Arguments arguments, const Children& children, ImportType type)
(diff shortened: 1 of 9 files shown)


kintel (approved):
Nice, LGTM!

jordanbrown0:
I'm confused by why it thinks I want to merge @kintel's fix.

jordanbrown0:
I hit it with the "rebase" and "push -f" rocks and now it seems right.

kintel:
Probably because I merged my font fix PR just before you opened this one, so your master branch was probably off by one commit.

kintel:
I might merge #5559 first too, which will have a tiny impact on this PR; a couple of noChildren() calls will be unnecessary.

PR #5562, opened 2025-01-04
Consistent quoting of variable names (via central function).  Fixes #2833.

### LINKED ISSUE #2833: Format warnings - delimit variable names etc.
From [forum](http://forum.openscad.org/New-Release-of-OpenSCAD-soon-please-test-tp25648p25721.html). I tend to agree. 
The file name should also be delimited as it could be called something confusing too, perhaps just a `:<space> ` before it assuming it is always at EOL.

> Regarding syntax checking, the following code outputs an odd waring

```
function somefunc(size) = 2 * size;
a = somefunc(size = 1, type = "foo");
```

> $ openscad-nightly -o out.stl bug-1.scad
> WARNING: variable type not specified as parameter, in file bug-1.scad, 
> line 3 Current top level object is empty.

> Formatting could perhaps be improved to more easily tell variable names from the actual warning text.
> Like wrapping the variable name in quotes, angle brackets, something like that.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/70812356-format-warnings-delimit-variable-names-etc?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jordanbrown0:
Add a new function `qVariable(name)` that returns the name appropriately wrapped in quote markup.  The current implementation of `qVariable()` wraps the name in double quotes, but could be changed to single quotes, angle brackets, et cetera.

Similar efforts might be appropriate for module names, function names, file names, enumerated constants (like halign/valign).  One step at a time.  (For one thing, it's less obvious what the right quoting is for modules and functions.  Should they get quotes, or parentheses (`foo()`), or words (`module foo`), or something else?)

Note that although this addresses the top-line comment in #2833, it doesn't address secondary comments (quoting file names, "UI-ERROR".  If this PR is accepted, I'll split those into a new issue.

### DIFF
--- a/src/core/control.cc
+++ b/src/core/control.cc
@@ -191,6 +191,7 @@ static std::shared_ptr<AbstractNode> builtin_assign(const ModuleInstantiation *i
       LOG(message_group::Warning, inst->location(), context->documentRoot(), "Assignment without variable name %1$s", argument->toEchoStringNoThrow());
     } else {
       if (assignContext->lookup_local_variable(*argument.name)) {
+        // TODO Should maybe quote the entire assignment with a new quoteExpr() or quoteStmt().
         LOG(message_group::Warning, inst->location(), context->documentRoot(), "Duplicate variable assignment %1$s = %2$s", quoteVar(*argument.name), argument->toEchoStringNoThrow());
       }
       assignContext->set_variable(*argument.name, std::move(argument.value));
--- a/src/core/control.cc
+++ b/src/core/control.cc
@@ -191,7 +191,7 @@ static std::shared_ptr<AbstractNode> builtin_assign(const ModuleInstantiation *i
       LOG(message_group::Warning, inst->location(), context->documentRoot(), "Assignment without variable name %1$s", argument->toEchoStringNoThrow());
     } else {
       if (assignContext->lookup_local_variable(*argument.name)) {
-        LOG(message_group::Warning, inst->location(), context->documentRoot(), "Duplicate variable assignment %1$s = %2$s", *argument.name, argument->toEchoStringNoThrow());
+        LOG(message_group::Warning, inst->location(), context->documentRoot(), "Duplicate variable assignment %1$s = %2$s", quoteVar(*argument.name), argument->toEchoStringNoThrow());
       }
       assignContext->set_variable(*argument.name, std::move(argument.value));
     }
--- a/src/core/Context.cc
+++ b/src/core/Context.cc
@@ -88,7 +88,7 @@ const Value& Context::lookup_variable(const std::string& name, const Location& l
 {
   boost::optional<const Value&> result = try_lookup_variable(name);
   if (!result) {
-    LOG(message_group::Warning, loc, documentRoot(), "Ignoring unknown variable '%1$s'", name);
+    LOG(message_group::Warning, loc, documentRoot(), "Ignoring unknown variable %1$s", quoteVar(name));
     return Value::undefined;
   }
   return *result;
--- a/src/core/EvaluationSession.cc
+++ b/src/core/EvaluationSession.cc
@@ -67,7 +67,7 @@ const Value& EvaluationSession::lookup_special_variable(const std::string& name,
 {
   boost::optional<const Value&> result = try_lookup_special_variable(name);
   if (!result) {
-    LOG(message_group::Warning, loc, documentRoot(), "Ignoring unknown variable '%1$s'", name);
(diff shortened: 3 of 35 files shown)


t-paul:
Nice! I just wonder if the choice of function name is a bit unfortunate as it *very* much looks like Qt related which it is not and actually must not be. Maybe shortening the other part? Something like `quoteVar`?

jordanbrown0:
I'm open to `quoteVar`.

I did think a bit about Qt, but all of Qt's stuff starts with an upper-case Q.  OTOH, there's a fair number of OpenSCAD-defined identifiers that are Qt-related that start with a lower-case q.

I'll change it.

t-paul:
Most, not all... `qDebug() << "hello";`

But we are not using that specific one in OpenSCAD code at all, I think.

jordanbrown0:
That push renames it to quoteVar.

kintel:
AFAIU, `std::quoted()` does the same as the custom `quoteVar()` in this PR. Could we just use that, now that we can require C++14 ?

jordanbrown0:
It's certainly related, but the intent of `quoteVar()` is not simply to add quotes but also to set a particular style.  Callers deliberately cannot control how quoting is done.  This allows us to quote variable names as `"name"`, or `'name'`, or `<name>`, as we like, and change if we like, without changing the callers.

It sets the stage for `quoteMod()`, `quoteFunc()`, and `quoteFile()`, which might or might not use double quotes.  In particular, we might decide that the appropriate style for modules and functions is `name()`.

These functions might use std::quoted in their implementation, but that's very different from having the callers use it directly.

jordanbrown0:
If we just wanted to add double quotes, mostly we could just put `\"` into the messages.  And in fact I started out that way, but I rapidly said "why am I embedding this style choice in every single caller?".

jordanbrown0:
Or we could even use curly quotes, or guillemets.  `std::quoted()` does not appear to allow for the left delimiter being different from the right.

src/core/control.cc
```diff
@@ -191,7 +191,7 @@ static std::shared_ptr<AbstractNode> builtin_assign(const ModuleInstantiation *i
       LOG(message_group::Warning, inst->location(), context->documentRoot(), "Assignment without variable name %1$s", argument->toEchoStringNoThrow());
     } else {
       if (assignContext->lookup_local_variable(*argument.name)) {
-        LOG(message_group::Warning, inst->location(), context->documentRoot(), "Duplicate variable assignment %1$s = %2$s", *argument.name, argument->toEchoStringNoThrow());
+        LOG(message_group::Warning, inst->location(), context->documentRoot(), "Duplicate variable assignment %1$s = %2$s", quoteVar(*argument.name), argument->toEchoStringNoThrow());
```
kintel, line 194: Not sure we want to quote just the variable here; it's the expression that's interesting.

jordanbrown0:
The quote marks used by `quoteVar()` should really be internationalized, but I haven't yet tried to understand how OpenSCAD i18n works.

src/core/control.cc
jordanbrown0: I debated that one, and would accept either way, but was persuaded by the original-issue case where the variable name is "type", which could in theory yield something like:
```
Duplicate variable assignment type = "number"
```

jordanbrown0: or did you mean something that would put quotes around the entire assignment, like
```
Duplicate variable assignment "name = value"
```
?

jordanbrown0: (Which would need a `quoteExpr()` function.)

kintel: Yeah, I meant that we should quote the expression. But as you point put, the value could also be quoted, which could be confusing when using double quotes, so perhaps that's indeed a use-case for quoteExpression :) ..or leave unquoted for now and revisit later.

jordanbrown0: I'd rather not dive into `quoteExpr()` today.  We could leave the variable name quoted as I have it, or revert the change for that particular case.  (I think there might have been two of them.)  Your call.

jordanbrown0: BTW, I think the answer for `quoteExpr()` might be to use something like `>>>...<<<` around the expression, like
```
Duplicate variable assignment >>> name = ... some expression ... <<<
```

kintel: I'm fine with leaving it,  but perhaps add a TODO comment there so we don't lose track

kintel (approved):

PR #5563, opened 2025-01-04
Add parameter list for local application start. Unify dark mode switch.

t-paul:
* Add configurable parameter list for local application start
* Add (optional) custom temp folder, e.g. PrusaSlicer switched to Flatpak recently which can't access `/tmp` by default - it's possible to `flatpak --user override com.prusa3d.PrusaSlicer --filesystem=/tmp` but using some other folder might be still useful
* Get rid of dark mode checks
  * Only icon theme is switched in `openscad_gui.cc`
  * Background colors are handled by blending transparent color overlay, see `UIUtils.cc`
* Icon themes are generated to two separate `*.qrc` files


kintel:
Let me know when you're ready, and I can port the parameter handling to macOS and test it.

t-paul:
Ok, I think that needs a wrap up. I would like to see multiple apps configurable but that might be another big change, so better keeping that as a future step. I'm not sure about the macOS handling.

Automatically generate `open` `-a` `<executable>` at the beginning, expecting the user to put everything else, including the `--args`  manually?

src/gui/Settings.cc
```diff
@@ -154,10 +185,12 @@ SettingsEntryEnum Settings::octoPrintFileFormat(
     createFileFormatItems({FileFormat::ASCII_STL, FileFormat::BINARY_STL, FileFormat::_3MF, FileFormat::OFF}),
     fileformat::info(FileFormat::ASCII_STL).description);
 
-SettingsEntryString Settings::localSlicerExecutable("printing", "localSlicerExecutable", "");
-SettingsEntryEnum Settings::localSlicerFileFormat(
-    "printing", "localSlicerFileFormat", createFileFormatItems(fileformat::all3D()),
+SettingsEntryString Settings::localAppExecutable("printing", "localAppExecutable", "");
+SettingsEntryString Settings::localAppTempDir("printing", "localAppTempDir", "");
+SettingsEntryEnum Settings::localAppFileFormat(
+    "printing", "localAppFileFormat", createFileFormatItems(fileformat::all3D()),
```
kintel, line 191: Since local apps don't inherently only deal with 3D file formats, should we rather enable all formats here? (`fileformat::all()`)

t-paul: Yes, that would make sense, even more so once we can get multiple configurable at some point.

src/gui/ExternalToolInterface.cc
```diff
@@ -86,18 +90,66 @@ bool OctoPrintService::process(const std::string& displayName, std::function<boo
...
+      args.append(QString::fromStdString(exportedFilename_));
+      break;
+    case Settings::LocalAppParameterType::dir:
+      args.append(info.absoluteDir().path());
+      break;
+    case Settings::LocalAppParameterType::extension:
+      args.append(info.suffix());
+      break;
+    case Settings::LocalAppParameterType::source:
+      args.append(QString::fromStdString(sourceFilename_));
```
kintel, line 139: Since we have the option to only supply the source file, there isn't technically a need to render before sending this to an external app.
This might be out of scope for this PR though.

t-paul: Yes, I know. It's never safe to assume we have a source file on disk. We could leave the 2 source options out, I was not really sure if it's a good idea to include.

kintel: I say leave it for now - we can let it simmer in public for a bit

src/gui/InitConfigurator.cc
```diff
@@ -1,13 +1,17 @@
 
 #include "gui/InitConfigurator.h"
+#include <qchar.h>
+#include <qlistwidget.h>
```
kintel, line 4: These should be CamelCased and lose the .h

t-paul: Ah, missed them. VSCode insists on adding those automatically. I need to find out if there's a way to disable that (or point to the public ones).

t-paul:
```suggestion
#include <QListWidget>
```

src/gui/Settings.cc
t-paul: As discussed, I'll update in some later patch...

PR #5572, opened 2025-01-07
Parameter aliases in linear_extrude, rotate_extrude.  Fixes #1609.

### LINKED ISSUE #1609: Parameter name consistency
Is there a reason to choose height in `linear_extrude(height=10)` and h in `cylinder(h=10)` instead of both h (or height), or accepting both?

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/32326176-parameter-name-consistency?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jordanbrown0:
linear_extrude allows "h" as an alias for "height". rotate_extrude allows "a" as an alias for "angle". Provides a general-purpose alias-capable lookup function in Parameters. Cleans up linear_extrude parameter processing.

### DIFF
--- a/src/core/Parameters.cc
+++ b/src/core/Parameters.cc
@@ -67,6 +67,30 @@ const Value& Parameters::get(const std::string& name) const
   return *value;
 }
 
+const Value& Parameters::get(const std::initializer_list<std::string> names) const
+{
+  std::string match;
+  for (std::string name: names) {
+    boost::optional<const Value&> value = lookup(name);
+    if (value && value->isDefined()) {
+      if (match.empty()) {
+        match = name;
+      } else {
+        LOG(message_group::Error, loc, documentRoot(),
+            "Specified both %1$s and %2$s", quoteVar(match), quoteVar(name));
+      }
+    }
+  }
+
+  for (std::string name : names) {
+    boost::optional<const Value&> value = lookup(name);
+    if (value && value->isDefined()) {
+      return *value;
+    }
+  }
+  return Value::undefined;
+}
+
 double Parameters::get(const std::string& name, double default_value) const
 {
   boost::optional<const Value&> value = lookup(name);
--- a/src/core/Parameters.cc
+++ b/src/core/Parameters.cc
@@ -69,26 +69,23 @@ const Value& Parameters::get(const std::string& name) const
 
 const Value& Parameters::get(const std::initializer_list<std::string> names) const
 {
-  std::string match;
+  std::string matchName;
+  boost::optional<const Value&> matchValue;
+
   for (std::string name: names) {
     boost::optional<const Value&> value = lookup(name);
     if (value && value->isDefined()) {
-      if (match.empty()) {
-        match = name;
+      if (!matchValue) {
+        matchName = name;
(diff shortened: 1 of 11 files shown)


jordanbrown0:
Sigh.  Thought I ran a full test.  Fixed tests to match new behavior.

src/core/Parameters.cc
```diff
@@ -67,6 +67,30 @@ const Value& Parameters::get(const std::string& name) const
...
+      }
+    }
+  }
+
+  for (std::string name : names) {
+    boost::optional<const Value&> value = lookup(name);
+    if (value && value->isDefined()) {
+      return *value;
+    }
+  }
```
kintel, line 90: Instead of iterating and looking up twice, can't you just keep the boost::optional around from earlier and return the value stored in that?

kintel (commented):
We should also add tests of the new features: Validate that h=/height=, and a=/angle= are equivalent, and validate that specifying both prints an error.

src/core/Parameters.cc
jordanbrown0: That does seem desirable, but my C++ is weak enough that I don't understand the rules for keeping around Value& values.  I'll try something and see if the compiler lets me.  (But I think I already tried that, and failed.)

kintel: I would try smth. like this (not tested):
```
  std::string match_name;
  boost::optional<const Value&> match_value;
  for (const std::string& name : names) {
    boost::optional<const Value&> value = lookup(name);
    if (value && value->isDefined()) {
      if (!match_value) {
        match_name = name;
        match_value = value;
      } else {
        LOG(message_group::Error, loc, documentRoot(),
            "Specified both %1$s and %2$s", quoteVar(match_name), quoteVar(name));
      }
    }
  }

  return match_value ? *match_value : Value::undefined;
```

kintel: ..and we should probably port from boost::optional to std::optional at some point

jordanbrown0: Wind and fire mean that I haven't had power since Tuesday evening. Not in danger, but may be offline for another couple of days.

kintel: Holy crap - you live in that part of CA. Stay safe, nobody is loosing sleep over late PRs :)

jordanbrown0:
I think the tests cover what needs to be covered, though there is an argument that they should be structured differently.  I avoided needing to regenerate any new result-images by changing existing tests to use various combinations of height and h, angle and a; if they generate the old result then it's correct.  But at the same time this "overloading" means that there aren't individual tests targeting individual behaviors.

src/core/Parameters.cc
jordanbrown0: Done.  (Not the std::optional part, because that requires more extensive changes.)

kintel (approved):

PR #5575, opened 2025-01-08
Fix manifold project nullptr

### LINKED ISSUE #5574: crash in projection function
**Describe the bug**
An example STL import file causes a client crash when **projection** is called on the imported object.

**To Reproduce**
1. Download the sample STL file from https://www.printables.com/model/493676-zardoz?lang=en
2. Execute the code below in the client (adjust the value of importfile to the actual filepath of the sample STL file).

**Expected behavior**
Expected to see a 2D object which is the "shadow" of the imported STL file.

```
importfile = "zardoz_PD.stl";
projection()
import(importfile);
```

**Screenshots**
n/a

 - OS: Windows 11
 - System: Intel PC 64-bit
 - OpenSCAD Version: 2025.01.06 (git 92a13b4f0) from openscad website.

n/a

**Additional context**

pca006132:
Fixes #5574.


kintel:
Thanks for picking this up! During review, I found and fixed another instance of the same lack of nullptr check.
I'll go through and look for other corner cases while at it.

PS! I'm surprised that we didn't actually add a test for this yet, or perhaps it got refactored away at some point..

kintel:
Will merge this for now. We have other related issues, which boils down to that internally, we sometimes use an empty Geometry and sometimes a nullptr to hold the result of a failed operation. This needs cleanup.

PR #5577, opened 2025-01-09
Add export dialog for 3MF.

t-paul:
ToDo:
- [ ] Check behaviour when mixing parts with and without color
- [ ] Implement for V1? (precision option is only available in V2, everything else should be possible in V1 too)


gsohler:
You Could extend the Customizer parser to fill in the fields in
export-3mf export form automatically with default values ?

As hybrid geometries and LazyUnion are gone now, 
you could also add meta data into the openscad script to
define mappings between openscad module()s and 3mf parts 
users are waiting for long time already.

(
this would be  analogue to
 export({ "part1": part1solid, "part2": part2solid } , "result.3mf")
)

t-paul:
The existing comment parser? No
Using dedicated [meta data](https://github.com/openscad/openscad/wiki/Meta-Data-Use-Cases)? Yes, that's on my wish list for ages

kintel:
Nice, adding export cmd-line options is great. We should really do some automated testing of this, but that might be quite a bit of work to do right..

kintel (approved):

PR #5584, opened 2025-01-12
RFC: New general-purpose extrusion mechanism - 2025 refresh

scrameta:
Refresh of https://github.com/openscad/openscad/pull/2796

- Adds a skin() module that takes a list of 2D polygons and extrudes a path between them.
- The number of vertices need not match on the polygons (interpolate=true/false)
- Polygons are aligned based on a line drawn from the centre at an optionally specified angle (align_angle=x degrees)
- The number of segments of triangles on each quad may be specified (x/sides = segments per side) (segments=x)
- For loops can be used to generate the list of 2D polygons, specify union=false in the for to disable the implicit union (or enable experimental feature LazyUnion)
- Hidden behind an experimental feature
- Added 2.5D polygon support to Polygon2d, which could in future be used for other modules.

Reviews/tips to make this suitable for merging very welcome.


scrameta:
I tested only with the 3 examples from the original PR: zigzag, horn and arch.
Tested those works with the manifold backend as well as CGAL.
Also tested they work with Lazy Union enabled. Note that 'union=false' is not needed in that case for horn and arch, though it still works with it.

scrameta:
After some discussion on IRC, working on adding some tests.

scrameta:
Added some tests:

1. linear_extrude clone built with extrude
2. rotate_extrude clone built with extrude
3. edge_cases to tests a few obvious edge cases (different number vertexes etc)

Working on manifold ok, on CGAL running into some assertion errors:

Expression : itl != it->second.end()
File       : /opt/homebrew/include/CGAL/Nef_3/SNC_external_structure.h
Line       : 1078
Explanation: 
ERROR: CGAL error in CGALUtils::applyUnion3D: CGAL ERROR: assertion violation!
Expr: itl != it->second.end()
/opt/homebrew/include/CGAL/Nef_3/SNC_external_structure.h
Line: 1078

Trying to debug though lldb arm64 on the Mac unfortunately does not seem to stop on CGAL assertions! Will try on my linux box in a few days.

scrameta:
The assertion errors were due to:
i) Degenerate triangles from scaling to 0 in my clone of linear_extrude test using extrude. Limited it to slightly over 0.
ii) Overlaps due to quad->triangle conversion that didn't prioritise the shorted diagonal.

Still getting a couple of test failures. One seems real to investigate (probably an edge case I should protect against). The other is a csg file serialisation precision issue.

scrameta:
The remaining issue is a rotation of >90 degrees, showing up as a problem when extrusion triangles overlap.

scrameta:
e.g.
extrude_segments-tests-expected
This has a 90 degree rotation which can't be seen without enough segments.

scrameta:
Added interpolation support, so that different numbers of vertices can be handled. Which will make this usable on a great deal more cases: different polygon shapes, rounded versions of polygons with offset etc.

Also adding support for alignment since there is no guarantee that the first vertex of a polygon is a specific angle. They don't all start with vertices[0] being at the bottom left for example.

jordanbrown0:
I would reserve the "align" parameter name for a hypothetical future "more than center" alignment feature, one that would allow, e.g., align to the front+top+left corner.  See #265 and several others.

scrameta:
Changed the parameter to 'interpolate'.
Also try to align polygons using a new parameter 'align_angle' which tries to find for example the bottom corner of a different polygon so that extrude can work between different shapes. 
Not quite there yet but some examples are working. e.g. can extrude from a hollow square to a rounded hollow square:
$fn=32;

module hollow_square_rounded(sz,thickness)
{
    difference()
    {
        offset(r=thickness) //offset(r=-(thickness))
        square(sz,center=true);
        offset(r=thickness) offset(r=-(thickness*1.5))
        square(sz,center=true);
    }
}
module hollow_square(sz,thickness)
{
    difference()
    {
        square(sz,center=true);
       offset(r=-thickness)
       square(sz,center=true);
    }
}

extrude(interpolate=true,align_angle=90,segments=20)
{
        hollow_square(10,2);
        translate([0,0,20])
        hollow_square_rounded(20,2);
}

scrameta:
Screenshot_2025-01-21_23-24-38

scrameta:
Renamed to skin.

e.g. 
`
skin()
{
  square(5);
  translate([0,0,10]) square(5);
}
`

gsohler:
would you mind merging "skin"/extrude_refresh somewhere else 1st ?

scrameta:
> would you mind merging "skin"/extrude_refresh somewhere else 1st ?

What do you mean?

scrameta:
Anyway no-one seems interested so guess I’ll close the PR soon so as not to clog things up.

gsohler:
nope, i am interested ... but this is another "openscad" - pythonscad.org

scrameta:
Ah sure yeah I can try to raise it for pythonscad. Out tonight but can take a look later this week.

kintel:
There is general interest, but very limited time, so it will take some time to merge PRs like this. Joining the discussion on IRC is likely the easiest way to make progress.

scrameta:
Raised a similar PR for pythonscad:
https://github.com/pythonscad/pythonscad/pull/1

scrameta:
There are a couple of csg render tests failing, but it looks like a union bug rather than this change.

gsohler:
BTW: the new PolySetRenderer introduced by kintel needs the same tweak for 2D shapes to show up correctly

kintel:
Ideally, we should refactor the PolySet and Polygon2d rendering out of CGALRenderer, but there is a bit of work remaining in order to pull that off..

PR #5588, opened 2025-01-14
Converting all manual connect() invocations to use the new style

### LINKED ISSUE #5547: Convert old Qt signal connection scheme to the typesafe new one
Qt5 introduced a more modern way of connecting signals to functions: https://wiki.qt.io/New_Signal_Slot_Syntax

The new way offers the benefit of being able to perform checks at compile-time for whether the signal and slot exists and that their parameters are correct.

TODO: Convert all manual `connect()` invocations to use the new style.

akashmahesh2805:
WIP to fix #5547

I have converted all the connect() invocations from the old syntax to the new syntax : 

Old syntax
:
connect(
    sender, SIGNAL( valueChanged( QString, QString ) ),
    receiver, SLOT( updateValue( QString ) )
);

New syntax:
connect(
    sender, &Sender::valueChanged,
    receiver, &Receiver::updateValue
);

### DIFF
--- a/src/gui/parameter/ParameterSlider.cc
+++ b/src/gui/parameter/ParameterSlider.cc
@@ -55,10 +55,20 @@ ParameterSlider::ParameterSlider(QWidget *parent, NumberParameter *parameter, De
   //connect(slider, SIGNAL(sliderPressed()), this, SLOT(onSliderPressed()));
   connect(slider, &QSlider::sliderReleased, this, &ParameterSlider::onSliderReleased);
   connect(slider, &QSlider::sliderMoved, this, &ParameterSlider::onSliderMoved);
+  
+  #if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)
   connect(slider, &QSlider::valueChanged, this, &ParameterSlider::onSliderChanged);
+  #else
+  connect(slider, static_cast<void(QSlider::*)(int)>(&QSlider::valueChanged), this, &ParameterSlider::onSliderChanged);
+  #endif
 
-  connect(doubleSpinBox, &QDoubleSpinBox::valueChanged, this, &ParameterSlider::onSpinBoxChanged);
-  connect(doubleSpinBox, &QDoubleSpinBox::editingFinished, this, &ParameterSlider::onSpinBoxEditingFinished);
+  #if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)
+    connect(doubleSpinBox, &QDoubleSpinBox::valueChanged, this, &ParameterSlider::onSpinBoxChanged);
+    connect(doubleSpinBox, &QDoubleSpinBox::editingFinished, this, &ParameterSlider::onSpinBoxEditingFinished);
+  #else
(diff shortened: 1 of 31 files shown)


akashmahesh2805:
So for all the connect() lines in which the method needs parameters, I need to handle it differently based on the Qt version, right?

Example : 
connect(this->e_tval, SIGNAL(textChanged(QString)), this, SLOT(updatedAnimTval()));

Will become: 
#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)
  connect(this->e_tval, &QLineEdit::textChangedQString, this, &Animate::updatedAnimTval);
#else
  connect(this->e_tval, static_cast<void (QLineEdit::*)(QString)>(&QLineEdit::textChanged), this, &Animate::updatedAnimTval);
#endif

Please correct me if I understood this wrong.

kintel:
@akashmahesh2805 It has nothing to do with whether the signals need parameters, but whether the same signals name is overloaded with multiple variants. In Qt6, all signals have been redesigned to have unique names, but in Qt5, this isn't always the case.

From your build failures, this seems to be the case for e.g. `QDoubleSpinBox::valueChanged`. The new, unique `textChanged` signal was introduced in Qt-5.14 (https://doc.qt.io/qt-5/qdoublespinbox.html#textChanged), so any builds against older versions of Qt would need a cast to help find the correct function pointer.

kintel:
You only need to use the static_cast trick for signals which are ambiguous in older Qt versions. If you revert your latest commits and read the compile errors, you should be able to see which signals are affected.

src/gui/parameter/ParameterSlider.cc
```diff
@@ -53,13 +53,21 @@ ParameterSlider::ParameterSlider(QWidget *parent, NumberParameter *parameter, De
...
-  connect(doubleSpinBox, SIGNAL(editingFinished()), this, SLOT(onSpinBoxEditingFinished()));
-
+  connect(slider, &QSlider::sliderReleased, this, &ParameterSlider::onSliderReleased);
+  connect(slider, &QSlider::sliderMoved, this, &ParameterSlider::onSliderMoved);
+  
+  #if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)
+  connect(slider, &QSlider::valueChanged, this, &ParameterSlider::onSliderChanged);
+  #else
+  connect(slider, static_cast<void(QSlider::*)(int)>(&QSlider::valueChanged), this, &ParameterSlider::onSliderChanged);
+  #endif
```
kintel, line 63: Is this needed for QSlider::valueChanged()?

akashmahesh2805: Corrected that, is there more?

kintel: Not that I know of, but check the compile errors.

akashmahesh2805: I don't understand why this isn't working. The compile errors just seem to show that there is ambiguity as both QDoubleSpinBox and QSpinBox have overloaded valueChanged signals, which I have already fixed. Should I close this PR or..?

kintel:
Oh, the signal is overloaded. I think you may need to use QOverload or smth. - check the Qt docs for spin box; it's described there, I've just never used it before..

kintel:
Let's leave it open in case someone comes along and fixes the last bit

kintel:
Superseded by #5726

PR #5590, opened 2025-01-15
Add constructors for SettingsEntryEnum::Item to allow existing initia…

### LINKED ISSUE #5589: Combo boxes in Preferences are all blank
**Describe the bug**
All combo boxes in the preferences window are blank.

**To Reproduce**
Open the Preferences window in the app. Try to change any of the settings that are set by a combo box. Note that all the combo box lines are blank.

**Expected behavior**
Combo box popups should have text.

N/A

**Screenshots**
combo2
combo1

 - OS: MacOS 15.1.1 
 - System: Mac (Apple Silicon)
 - OpenSCAD Version: current head of master (first noticed in daily build, then confirmed by building from source)

N/A

**Additional context**
It appears that commit a71071c73 broke this. It added a "name" field to the SettingsEntryEnum::Item structure, but did not take into account that many of these structures are initialized using direct structure assignment, which depends on the order of data members in the struct and is inherently unsafe. (See Settings.cc, line 149 and onwards for the initializations in question).

For example, in this line:

```
SettingsEntryEnum<std::string> Settings::tabKeyFunction("editor", "tabKeyFunction", {{"Indent", _("Indent")}, {"InsertTab", _("Insert Tab")}}, "Indent");
```

`{"InsertTab", _("Insert Tab")}` is the initializer for one SettingsEntryEnum::Item structure. Before the change, it would have assigned "InsertTab" to .value and "Insert Tab" to .description, after the change it assigns "Insert Tab" to .name and leaves .description set to the empty string.

This means that when `InitConfigurator::initComboBox()` later does this:
```
    comboBox->addItem(QString::fromStdString(item.description), QString::fromStdString(item.name));
```
it adds the item with an empty string as its contents.

kintel:
…lizers to work correctly.

Fixes #5589

### DIFF
--- a/src/core/Settings.h
+++ b/src/core/Settings.h
@@ -179,6 +179,8 @@ class SettingsEntryEnum : public SettingsEntry<enum_type>
 {
 public:
   struct Item {
+    Item(enum_type value, std::string description) : value(value), description(std::move(description)) {}  
+    Item(enum_type value, std::string name, std::string description) : value(value), name(std::move(name)), description(std::move(description)) {}  
     enum_type value;
     std::string name;
     std::string description;


t-paul:
I think I'll add the names so we only have the 3 field constructor. Making the name officially optional can cause issues with radio button settings as it's required there. It's probably not worth tossing template specialization in for this case where the enum value type is `std::string`.

t-paul:
See #5592.

t-paul:
Superseded by #5592.

PR #5621, opened 2025-01-25
Drop examples.json; Collect examples from resource and user directory.

t-paul:
ToDo:
- [x] Test on Linux
- [x] Test on Windows
- [x] Test on macOS

Examples are read from both resource folder and user directory (shown in library info window).
image

An optional file `example-dir.json` can add some extra info and overwrite the name:
```
{
	"sort": 1000,
	"name": "Awesome Examples",
	"tooltip": "Hi there!"
}
```


kintel:
It wasn't obvious that you need to  create an extra level of directories under the user examples directory. If not too much work, perhaps consider supporting files and example-dir.json directly in the examples directory itself.

t-paul:
Right now the 2 hierarchies are merged, so in both cases 1 level directories (with the optional config file inside) + files in those directories.

How would that work with just files? Put them on top-level? Or have a dedicated and separate user sub-folder instead of merging? @raster, do you have a suggestion for that?

kintel:
Didn't really think about the how :) I imagined just listing top-level files by default, and if a top-level json file is there, make it into a folder in the list view.

t-paul:
Uh, oh, I see, allow both depending on the JSON file :smile:.

raster:
I was looking at the way the Arduino IDE looks at it and inside the libraries folder are all the libraries, and each library can have an"examples" folder, but it looks like they also need a "libraries.properties" folder, as that gets parsed by the IDE at launch. (Just having the "examples" folder isn't enough to get it added.)

t-paul:
Merged as there's some discussion on adding support for grabbing examples from libraries too by someone on IRC. We can extend the current behavior later in case that change would not happen.

PR #5637, opened 2025-02-01
Fix Console output on save error

cjmayo:
"File/Save As..." to a read-only directory can cause invalid characters to be written to the Console.

---

I get a lot of  `�` and a few other characters written to the console - Qt 6.8.1 and Qt 6.8.2 on GNOME on Linux.

It kind of makes sense, and to let Qt do the conversion to std:string. But there are a number of other `toLocal8Bit()` used with `LOG()` and I haven't seen the invalid characters before - not that I have checked carefully. Maybe it still makes sense to replace them all?

### DIFF
--- a/src/gui/TabManager.cc
+++ b/src/gui/TabManager.cc
@@ -618,8 +618,8 @@ bool TabManager::shouldClose()
 
 void TabManager::saveError(const QIODevice& file, const std::string& msg, const QString& filepath)
 {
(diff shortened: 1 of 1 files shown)


kintel:
Thanks, I think this is the correct fix.

In general, I believe filename output to non-utf-8-capable terminals (like MS Windows) needs to be converted to the local code page. ..but if the output goes to the GUI, like in this case, we don't need any conversions.

PR #5644, opened 2025-02-04
also reset camera translation in viewAngle* to make one axis disappear

gsohler:
it was noticable that even when i pressed "view from Top" stil Z axis was visible in perspective projection.
This was due to camera translation not reset.
This PR will make 3d View look more perfect.

BTW: messing with the axes instead is not a good idea as they would not be precise anymore to read dimension

### DIFF
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -2922,42 +2922,49 @@ void MainWindow::editorContentChanged()
 void MainWindow::viewAngleTop()
 {
   qglview->cam.object_rot << 90, 0, 0;
+  qglview->cam.object_trans << 0, 0, 0;
   this->qglview->update();
 }
 
 void MainWindow::viewAngleBottom()
 {
   qglview->cam.object_rot << 270, 0, 0;
+  qglview->cam.object_trans << 0, 0, 0;
   this->qglview->update();
 }
 
 void MainWindow::viewAngleLeft()
 {
   qglview->cam.object_rot << 0, 0, 90;
+  qglview->cam.object_trans << 0, 0, 0;
   this->qglview->update();
 }
 
 void MainWindow::viewAngleRight()
 {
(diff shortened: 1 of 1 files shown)


kintel:
On second thought: This makes it harder to work on models in isometric view if they're not centered in the origin.

jordanbrown0:
Two ideas:

* Clicking the view-from-direction icon (or choosing the menu item, or the keyboard shortcut) once changes the view angle.  Clicking it again, when the view angle is already "correct", changes the view origin.  Perhaps the two would have to be in quick succession.  (Maybe three times would zoom to fit.)
* Clicking (et cetera) would change the view angle.  Clicking (et cetera) with a modifier would also change the view origin.  (But what modifier?  Shift is obvious, but Ctrl+0 is already "view from diagonal, without translating" and Ctrl+Shift+0 is already "center the origin in the viewport, without changing angle or zoom".)

Q; Clicking "+Z" / Top would center X and Y, but what happens to Z?  (And similarly for the other five directions.)  Suggestion:  the coordinate for the axis you're aligning on would remain unchanged.  Extending that to diagonal views, the distance from the view origin would remain unchanged.  (This might already be the definition, but it should be explicit.)

kintel:
Yeah, I think the challenge is to support multiple different UI workflows using limited button real-estate.
When working on small, origin-centered objects, it does make sense to point the camera towards the origin (and possibly also zoom to fit). However, when working on larger assemblies, I would think that the currently selected camera origin is often preferred to be retained. It's also possible that preferences changes depending on whether perspective or orthogonal projection is used.

Anyway, that is to say: I completely misjudged the complexity of this from the original issue, so it's a bit premature to solve this using a PR without coming up with a good solution first. Sorry about jumping the gun.

kintel:
Closing as we misjudged the complexity of this issue

PR #5650, opened 2025-02-07
Removed relative include paths, and fix stragglers

kintel:


kintel:
@hzeller FYI

kintel:
Note: This has a small chance of breaking PRs in progress, but the impact should be minimal.

hzeller:
Nice!

t-paul:
How do we make sure to not regress from the cleaned state?

hzeller:
At least since the include paths are removed from the CMake file, it won't compile otherwise.
Having said that, the compiler probably will still allow to `-I.` so includes from the same directory will still accidentally work.

We could have a script that to run in the CI to make sure include-paths are always relative to project root.

kintel:
Getting some sort of clang-tidy script running as a GitHub action would be cool. The main challenge is tweaking it so that it doesn't get in the way until we can establish a situation with no issues.

hzeller:
I can make a simple one that outputs at least the overall findings and can cache the results between runs as a starting point.
Then, we could use that data to output before/after in a pull request.

I can work on that a little if nobody is.

kintel:
Yeah, that would be awesome.

t-paul:
Yes, that would be great. If it could easily be run locally too, that would be perfect.

hzeller:
You can already run `scripts/run-clang-tidy-cached.cc` . First time is slow as clang-tidy is slow, but it will cache results and in later runs only reprocess changed sourcces.

I'll probably also add a script 'scripts/check-potential-problems.sh' or so that can test quickly if we have include files with full relative path to project root (and that then can also be extended with whatever quick check we come up with). That then can be run locally and also in the CI.

hzeller:
Started running the `clang-tidy` and preserving cache between runs in #5653

PR #5652, opened 2025-02-08
scripts/uni-get-dependencies.sh: add support for Debian-like distros using /etc/os-release

peap:
[Pop!_OS](https://pop.system76.com/) is a popular distribution[^1] based on Ubuntu.  As for Ubuntu itself, the Debian dependencies are required to build OpenSCAD, so add it to the list of supported distros in the script.

Example /etc/issue content on a machine running Pop!_OS:

    Pop!_OS 22.04 LTS \n \l

[^1]: Top 10 in January 2025, per https://distrowatch.com/table.php?distribution=popos

### DIFF
--- a/scripts/uni-get-dependencies.sh
+++ b/scripts/uni-get-dependencies.sh
@@ -157,8 +157,6 @@ if [ -e /etc/issue ]; then
   get_debian_deps
  elif [ "`grep -i linux.mint /etc/issue`" ]; then
   get_debian_deps
- elif [ "`grep -i pop\!_os /etc/issue`" ]; then
-  get_debian_deps
  elif [ "`grep -i suse /etc/issue`" ]; then
   get_opensuse_deps
  elif [ "`grep -i fedora.release.2[2-9] /etc/issue`" ]; then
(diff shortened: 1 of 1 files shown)


t-paul:
Could you please verify if `/etc/os-release` exists and maybe has a simpler check. It would be nice slowly moving to this file instead of `/etc/issue`.

peap:
It does, yes:

```
$ cat /etc/os-release
NAME="Pop!_OS"
VERSION="22.04 LTS"
ID=pop
ID_LIKE="ubuntu debian"
PRETTY_NAME="Pop!_OS 22.04 LTS"
VERSION_ID="22.04"
HOME_URL="https://pop.system76.com"
SUPPORT_URL="https://support.system76.com"
BUG_REPORT_URL="https://github.com/pop-os/pop/issues"
PRIVACY_POLICY_URL="https://system76.com/privacy"
VERSION_CODENAME=jammy
UBUNTU_CODENAME=jammy
LOGO=distributor-logo-pop-os
```

t-paul:
Nice! If you don't mind changing that, the spec at https://www.freedesktop.org/software/systemd/man/latest/os-release.html has a snippet under "Example 3. Reading os-release in sh" that should work here as is. Maybe going last to keep the old style stuff for now.

peap:
Thanks for the pointer!  Took a stab at it.

t-paul (approved):
Nice, thank you! Looks good. I don't think the CI checks here use the script, I'll let them run through anyway before merging.

PR #5653, opened 2025-02-08
Start a CI running the clang-tidy

hzeller:
We can use the cached data later to compare if there are regressions in PRs.

Context: https://github.com/openscad/openscad/pull/5650#issuecomment-2645923795


hzeller:
The first run takes a while, but then results are cached, so later only changed files will be looked at.

hzeller:
This now reports the current status of `clang-tidy`-ness

https://github.com/openscad/openscad/actions/runs/13219537553/job/36902843512?pr=5653#step:7:13

hzeller:
Alright, this shall be a starting point.
I can experiment more once we it is merged to master and it is possible to see if some of the before/after mechanism works.

.github/workflows/code-tidy.yml
```diff
@@ -0,0 +1,73 @@
...
+    - name: Create Cache Timestamp
+      id: cache_timestamp
+      uses: nanzm/get-time-action@v2.0
+      with:
+        format: 'YYYY-MM-DD-HH-mm-ss'
+
+    - name: Retrieve cached results
+      uses: actions/cache@v4
+      with:
+        path: /home/runner/.cache/clang-tidy
```
kintel, line 41: The name of the directory immediately below this one contains a hash of the clang-tidy version used. Does that mean we may need to clean out this from time to time as well? If so, perhaps add a TODO comment to that tune.

.github/workflows/code-tidy.yml
```diff
@@ -0,0 +1,73 @@
...
+      run: cmake -S . -B build -GNinja
+
+    - name: Build autogenerated files
+      run: ninja -C build CMakeFiles/OpenSCAD_autogen
+
+    - name: Create Cache Timestamp
+      id: cache_timestamp
+      uses: nanzm/get-time-action@v2.0
+      with:
+        format: 'YYYY-MM-DD-HH-mm-ss'
```
kintel, line 36: Is the intent to get a unique build id? Timestamp is probably fine, but there might exist a build ID variable we could use. Fine for now though, just thought I'd ask..

kintel (approved):
LGTM, just two minor comments

.github/workflows/code-tidy.yml
hzeller: I deliberately choose a timestamp to have 'latest first' to be found by the cache. Subsequent runs will then find results that are most likely contain all relevant files.

The matching in finding the cache goes alphabetically if there is nothing found with an exact match, so that should work with that intention.

Having said that, there is room for improvement: if we are already in a pull request and have written the cache within that before, maybe we should re-use that very cache. Added a TODO.

.github/workflows/code-tidy.yml
hzeller: Yes, things will accumulate, albeit probably pretty slow. Added a comment.

PR #5654, opened 2025-02-08
Include some missing system headers

hzeller:
At least instances where we have more than one file affected. One commit per header.

All done essentially semi-automatic by taking the output of `scripts/run-clang-tidy-cached.cc`, then using an awk-script to find files with relevant reports to then be using [`insert-header.cc`](https://github.com/hzeller/dev-tools/blob/main/insert-header.cc) to add the missing header.

So typically something like

```bash
../dev-tools/insert-header.cc '<ostream>' $(awk -F: '/^src.*no header providing.*std::ostream/ {print $1}' OpenSCAD_clang-tidy.out)
```

### DIFF
--- a/src/core/AST.cc
+++ b/src/core/AST.cc
@@ -1,4 +1,5 @@
 #include "core/AST.h"
+#include <filesystem>
 #include <ostream>
 #include <memory>
 #include <sstream>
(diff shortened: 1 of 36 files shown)


hzeller:
mmh, what is going on in the Linux tests ? Do I have a typo somewhere in the header names ?

hzeller:
Ah, looks like MOC just didn't like the `<filesystem>` header first.

PR #5663, opened 2025-02-11
Include core/ModuleInstantiation.h in files using symbols from it.

hzeller:


kintel:
[misc-include-cleaner]   -15

Issue #4943, opened 2024-01-20
center()


yahbluez:
I miss the possibility to center or place an imported object.

It should be possibly to have a center() module that centers his childs.

center() import();
Now the imported object should be centered.

If center could take parameters to set one or more axes that would be nice to.

jordanbrown0:
Written the way you describe, it's not as easy as it seems, because in general during a preview you can't figure out anything at all about the child object.  You have to actually *do* the boolean operations, and preview doesn't actually do them.  (It does black magic instead.)  PR#4478 has mechanisms that can be used to implement such a module, at the expense of having to render the child.

There could also be centering (or, more generally, alignment) options on the import.  The import can see the data it's importing, so can know how to center it.

yahbluez:
Thank you

> There could also be centering (or, more generally, alignment) options on the import. The import can see the data it's importing, so can know how to center it.

How can i do that?

For example importing an STL and get the size xyz of a surrounding block and the position of one corner.

jordanbrown0:
*You* can't.  But OpenSCAD could more or less easily be modified to allow it.

jordanbrown0:
A hypothetical `center()` module could be like `resize()` and render its children - at a performance cost.  In addition to the performance cost, the obvious implementation would lose color, transparency, et cetera.

So it's not impossible to implement, but it wouldn't be pretty.

yahbluez:
I do not care about CPU time a function needs that implements a needed functionality.
The prerelease versions are so much faster than the old stable.
If i look how much cpu time is needed to do fancy stuff with imported STL files a center() will not add some real time to that.

I wonder if their is a way to motivate the openscad developers to add some useful things to openscad like the center() or an export() that could be called from inside an openscad script.

Using bash/python wrappers for that increases the workload a lot.

gsohler:
most primitive functions of openscad have somehow a center option,
but especially import() has not.
it would actually be very easy to have a center option in import(), and i
believe this would satisfy yahbluez needs ....

jordanbrown0:
Let's not discuss it here because it's off-topic for this issue, but `export()` per se is very unlikely to happen for security reasons.  If you'd like to discuss it, please take it to the mailing list or e-mail me privately at openscad(at)jordan(dot)maileater(dot)net.

wolfwood:
the center parameter to import is now working for all (except nef3) geometry file types.

[...]

PR #5669, opened 2025-02-14
support centering an stl when importing

wolfwood:
`import()` already has a `center` parameter, but it is ignored for most file types. this implements centering for stls. includes a test.

mostly addresses the motivation for #4943


wolfwood:
closing in favor of #5671

Issue #5670, opened 2025-02-14
import of SVG with viewbox attribute overrides centering


wolfwood:
via a reddit post https://www.reddit.com/r/openscad/comments/1in51qq/comment/mcaw2im/ by @garblesnarky

`import("test.svg", center=true);`

with [test.svg](https://github.com/user-attachments/assets/83782e33-94f8-4fba-9515-9d17521f398a)

doesn't result in the imported file living at the origin. the user determined a viewbox attribute is the reason.

I was assuming this was a bug, but from the svg centering code it could be intentional? I'm looking for clarification so I can add an appropriate testcase, and a fix (or not).

wolfwood:
this is where the viewbox is used to translate each point.
https://github.com/openscad/openscad/blob/master/src/io/import_svg.cc#L203-L204

wolfwood:
it looks like originally SVG centering worked as expected. this commit https://github.com/openscad/openscad/commit/d5b3ae2c158fc3b1dd5f9ba16ca61993b4f7b8fb added scaling based on the viewbox. it doesn't explicitly mention centering behavior.

there's no test of the interaction between viewbox and center. removing `-viewbox.x()` and `-viewbox.y()` from the centering here https://github.com/openscad/openscad/blob/master/src/io/import_svg.cc#L203-L204 still passes all tests.

t-paul:
Right, the "auto-center" of the test suite hides placement issues when there is not another fixed object added.

I suppose the common expectation would be to still scale based on the viewbox, but center on actual geometry? The current calculation for the center case seems to be wrong in any case, so we should fix that.

wolfwood:
that matches my expectation. I updated my PR for this, #5683, to fix the calculation.

stefano2734:
Not so important for next OpenScad Release, when more than a year untested with actual dev versions?

PR #5671, opened 2025-02-14
support centering when importing

wolfwood:
implement the `import()` center parameter for all file types.
 
enhanced version of #5669. ~~currently treats #5670 as a bug and~~ mostly addresses #4943.

currently only has test cases for stl ~~and svg~~.


wolfwood:
also need to consider interaction between origin and center for dxg. could also support origin for all file types as well?

t-paul:
As the parameter exists, I think it makes sense to support this for as many formats as possible.
 
I'm not sure this is a good idea for 3MF as that can contain multiple objects. Right now those are unioned to a single geometry, but that hopefully will change at some point. In that case there would be no simple way of keeping the same result.

wolfwood:
I am happy to drop 3mf if you wish. However, I don't think that having a set of objects actually prevents centering the group.

even if the items are independent, all we need to do is find the min and max value for each axis across all the vertices, either by iterating each object bounding box, or each individual vertex.

then translate all objects by the same amount (max - min / 2 on each axis), or modify each vertex.

src/core/ImportNode.cc
```diff
@@ -163,37 +192,37 @@ std::unique_ptr<const Geometry> ImportNode::createGeometry() const
...
   case ImportType::DXF: {
     DxfData dd(this->fn, this->fs, this->fa, this->filename, this->layer.value_or(""), this->origin_x, this->origin_y, this->scale);
-    g = dd.toPolygon2d();
+    g = optionally_center(dd.toPolygon2d(), this->center);
     break;
   }
 #ifdef ENABLE_CGAL
   case ImportType::NEF3: {
-    g = import_nef3(this->filename, loc);
+    g = optionally_center(import_nef3(this->filename, loc), this->center);
```
kintel, line 225: Since all the cases do exactly the same: Combined with my earlier suggestion, you could pull optionally_center() out of the switch and just do `if (this->center) g = center(g)`, or possibly also move center() into the Geometry class.

wolfwood: I wasn't sure if centering should stay in the import or become generally available.

I would love to implement it on Geometry, but the issue I was having is that the 2D transform on Polygon2d can't be called once it becomes a Geometry. I could implement `center_on_origin()` as an abstract Geometry method for each of PolySet, Polygon2d and  CGAL_Nef_polyhedron, but it would be nice if I could only do it on Geometry alone.

src/core/ImportNode.cc
```diff
@@ -153,6 +153,35 @@ static std::shared_ptr<AbstractNode> do_import(const ModuleInstantiation *inst,
 static std::shared_ptr<AbstractNode> builtin_import(const ModuleInstantiation *inst, Arguments arguments)
 { return do_import(inst, std::move(arguments), ImportType::UNKNOWN); }
 
+template<typename T>
+static std::unique_ptr<T> optionally_center(std::unique_ptr<T> g, bool center) {
+  if (center) {
+    auto bbox = g->getBoundingBox();
+    auto center = bbox.center();
+
+    if constexpr (std::is_same_v<T, Polygon2d>) {
```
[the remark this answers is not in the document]
wolfwood: good idea, I missed that method. But when I try the issue is that the dimension needs to be known at compile time, since it changes the signature of `g->transform()`.

it seemed like it should be easier to share code on each side of the if , but the type of the affine matrix is determined by the dimension an then everything else depends on the matrix.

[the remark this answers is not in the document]
kintel: oh. right. That's not great API design on Geometry's part. Good opportunity for refactoring, but this PR probably isn't the best time for it..

src/core/ImportNode.cc
kintel: Yeah, it follows the other case. Let's leave it like this, and if we refactor Geometry in the future, code can be unified a bit..

wolfwood: well, I could do a 3D implementation in Geometry, and then override it in Polygon2d. not really any more duplication than there is now. would that be preferrable?

kintel: Nah, let's just leave it for now. We're considering allowing 3D transformations on 2D geometry, in which case we need to rethink some of the APIs, so we can defer all this until after that's sorted.

wolfwood:
I decided to leave svg centering untouched in this commit. that way it doesn't depend on a decision about #5670.

wolfwood:
rebased and added 2 tests, an stl test for `center=false` and an obj test. 

It looks like there isn't a data directory for 3mf or OFF, openscad doesn't have an export for nef3, and AMF is deprecated (and my stl test file export fails as non-manifold, although other formats don't have that issue). 

one thing I noticed is that the import is cached, so toggling the centering doesn't take effect until I exit and reopen openscad. not sure if there is an easy way to invalidate the cache when the properties change.

I have also written but not committed a dxf test. 7 test get get created from the file automatically (can I make this happen for the other tests?). all pass except the csgrendertest_dxf-import-centered variant, which does not do the centering. very strange. I've committed that test and associated files here instead: https://github.com/wolfwood/openscad/tree/feature/center-dxf-test

t-paul:
Cache issue happens due to ImportNode::ToString() only handling `center` for SVG currently.

https://github.com/openscad/openscad/blob/8c9a3728b5dadcdce44048663e26934fd3fef88b/src/core/ImportNode.cc#L223-L226

kintel:
Some notes:
* There's a 3mf file here: https://github.com/openscad/openscad/blob/master/tests/data/scad/3D/features/import.3mf. Ideally, it should move to a different folder. Not sure why it's there.
* We should (also ideally) add an off test file too.
* Regarding the non-manifold; that sounds like a real issue; perhaps double-check your STL. Using ascii STL is usually better for test purposes, as it's easier to manually inspect the contents.
* I say don't worry about AMF.
* Centering nef3 might be annoying as you'd need to evaluate to concrete polygons first, which may fail etc., so perhaps there isn't any need to support that. It's mostly used for testing anyway and not really meant as and en user feature.

wolfwood:
I picked a better test file, no manifoldness issues.

3mf and OBJ tests added. I made an AMF test and it has the same issue as the dxf, csgrendertest fails, the other 6 succeed.

wolfwood:
also fixed the caching issue and dropped nef3 centering support. stl test file is now ascii. added OFF test.

so there will be full test coverage, if I bring in the dxf test and add or drop the amf test, once the csgrendertest issue is resolved (with svg and nef3 excluded and the svg test living in its own PR).

wolfwood:
sorry,forgot to do an unfiltered test run. have updated the failing dumptests to reflect the center param. a side effect is that many of the timestamps are now negative? not sure if they are used for comparison, since the tests are passing...

wolfwood:
after switching branches and rebuilding, the csgrendertest failures disappeared. I wish I could pinpoint what the cause was, but I am no longer able to reproduce.

dxf and amf tests have therefore been added here. I believe that the feature, and the tests, are complete. if you would like 3mf or amf to be removed let me know.

kintel (commented):
Looks great! Only thing I don't quite understand is the rendered DXF import; you added an origin reference but I'm not seeing it..

wolfwood:
the reference square is smaller than far-out-shape.dxf, so it gets covered if centering works. its mostly to defeat the auto-centering of the test so that failures can be detected, because things get very small and the middle is empty.

kintel (approved):

t-paul (approved):
Very nice! I still have some concerns regarding 3MF, but maybe this can be solved by a meta-data import reporting the file structure. So I guess that's a go from my side. :+1: :rocket: :grinning:

wolfwood:
thank you both for being so welcoming!

Issue #3689, opened 2021-02-24
Transparency leaks through on wayland


waweic:
Hi, there's a thing I just noticed, I didn't investigate it further. I'm using Openscad in Sway and whenever there's a transparent object, I can see my wallpaper in the background. Is this expected behaviour? I do find this mildly irritating / funny.

![Screenshot of described behaviour](https://user-images.githubusercontent.com/14358706/108980175-af638a00-768b-11eb-8b10-d6ea1d4c2d71.png)

t-paul:
No, that's not supposed to happen and it does not when using Gnome on Wayland, so I'm not sure this is an OpenSCAD issue.

nobodyinperson:
Also happens for me on Manjaro XFCE with X, not Wayland.

ednisley:
Still happens with Manjaro XFCE with X as of #bb54b97, although it took me a while to recognize the Veil Nebula image recolored from the desktop background:
![OpenSCAD transparency desktop showthrough](https://user-images.githubusercontent.com/599600/128226908-014aeea1-39d9-44d3-8f15-c0cd0d5a4af7.png)
[Hubble Veil Nebula](https://www.nasa.gov/image-feature/goddard/2021/hubble-revisits-the-veil-nebula)

sephthir:
Also happening on Fedora 35 + Wayland, installed via `dnf` (in order to support dark theming...), it's picking up whatever window is behind it (in this case, this thread). This can be more than distracting, as it can make it really hard to tell what's going on if the window behind it has a lot going on.

![OpenSCAD transparent bleed](https://user-images.githubusercontent.com/4183785/185721543-30fcc0d5-2e88-48a8-8d2e-12ae76b02ee5.png)

pipatron:
And definitely happening for me on Gnome+Wayland in debian bookworm in 2023.12.01.nightly (git 31b56c7)

flaviut:
Related issue in another app with a proposed solution from a wayland developer: https://github.com/swaywm/sway/issues/6549

See also: [FreeCAD](https://github.com/FreeCAD/FreeCAD/issues/6177), [OpenSCAD](https://github.com/openscad/openscad/issues/3689), [Cura](https://github.com/Ultimaker/Cura/issues/8466)

frysztak:
This issue is driving me insane, so I did some digging.

[...]

PR #5677, opened 2025-02-18
Fix transparency issues on Wayland

frysztak:
Fixes https://github.com/openscad/openscad/issues/3689

Tested on GNOME Wayland and Qt6

### DIFF
--- a/src/glview/GLView.cc
+++ b/src/glview/GLView.cc
@@ -219,6 +219,24 @@ void GLView::paintGL()
   }
   glDisable(GL_LIGHTING);
   if (showaxes) GLView::showSmallaxes(axescolor);
+
+  // Workaround for inconsistent QT behavior related to handling custom OpenGL widgets that
+  // leave non opaque alpha values in final output.
+  // On wayland that can cause window to become transparent or blurry trail effect in the
+  // parts that contain partially transparent objects.
+  //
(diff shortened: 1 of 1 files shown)


src/glview/GLView.cc
```diff
@@ -219,6 +219,23 @@ void GLView::paintGL()
...
+  // compositing stack at QT and desktop level would interpret transparent pixels.
+  //
+  // Solves https://github.com/openscad/openscad/issues/3689.
+  //
+  // Originally developed by @karliss for FreeCAD (https://github.com/FreeCAD/FreeCAD/pull/19499).
+  glPushAttrib(GL_COLOR_BUFFER_BIT);
+  glColorMask(false, false, false, true);
+  glClearColor(0, 0, 0, 1);
+  glClear(GL_COLOR_BUFFER_BIT);
+  glPopAttrib();
```
kintel, line 238: I don't think you need to push/pop here. We're trying to move off immediate-mode OpenGL, so that we can move 
 to OpenGL > 3 or GLES, so it would be good to avoid introducing new code using it.

frysztak: without push/pop preview is completely black

frysztak: that seems to work!

kintel (approved):

kintel:
Thanks for fixing this!

PR #5681, opened 2025-02-18
Upload clang-tidy results to artifacts

kintel:


kintel:
Ideally, we should report the findings inline or smth., but this is a minimal effort to expose the findings.

kintel:
One challenge, if we introduce a regression, it's still pretty hard to locate it, as we only diff the summary, but cannot easily locate the actual report entries in question.

kintel:
Example: #5679 introduces some include-cleaner regressions

PR #5683, opened 2025-02-19
fix the interaction of svg viewbox with svg_import centering

wolfwood:
~~pending a determination on #5670 either I will replace the expected file so this passes or~~ I will remove the viewbox coordinates from the centering code ~~(or just use the centering function from #5671).~~


wolfwood:
I have removed the viewbox coordinates from the scaling and centering adjustment loop. all tests pass, including the new one. although it doesn't involves scaling, it had a viewbox and wants the geometry centered.

kintel:
@wolfwood This fell off the radar. Still interested in landing this?

wolfwood:
I'm happy to update it. not an svg viewbox user myself, but centering the contents rather than the view box makes sense to me.

kintel:
Agreed, not too familiar with viewbox, but I believe the intent of center= is to consider the geometry itself.

wolfwood:
@kintel i think the svg testing was less thorough before. now, the image comparison of all the other svg tests pass, but the svg files change because effectively the origin is different.

the minimum x and y values as specified in the viewbox change, and then all the coordinates change by a corresponding amount. 

example:
```diff
diff --git a/tests/regression/export-svg-fill-only/size-percent-expected.svg b/tests/regression/export-svg-fill-only/size-percent-expected.svg
index e04e78de9..b7e968e5c 100644
--- a/tests/regression/export-svg-fill-only/size-percent-expected.svg
+++ b/tests/regression/export-svg-fill-only/size-percent-expected.svg
@@ -1,11 +1,11 @@
 <?xml version="1.0" standalone="no"?>
 <!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
-<svg width="170mm" height="111mm" viewBox="0 -333 170 111" xmlns="http://www.w3.org/2000/svg" version="1.1">
+<svg width="170mm" height="111mm" viewBox="0 -222 170 111" xmlns="http://www.w3.org/2000/svg" version="1.1">
 <title>OpenSCAD Model</title>
 <path d="
-M 110,-332.076 L 10.0004,-332.076 L 10.0004,-232.076 L 110,-232.076 z
-M 170,-227 L 5,-227 L 5,-332 L 0,-332 L 0,-222 L 170,-222
+M 110,-221.076 L 10.0004,-221.076 L 10.0004,-121.076 L 110,-121.076 z
+M 170,-116 L 5,-116 L 5,-221 L 0,-221 L 0,-111 L 170,-111
  z
-M 170,-322.068 L 120,-322.068 L 120,-232.068 L 170,-232.068 z
+M 170,-211.068 L 120,-211.068 L 120,-121.068 L 170,-121.068 z
 " stroke="none" fill="blue" stroke-width="0.35"/>
 </svg>
```

the current behavior always has the minimum x as 0 but frequently with the new code is it offset, positive or negative. the y is always a negative value equal to twice the viewbox y size, or more.

~~I am not sure if I should rework the code to always have 0,0 as the min coordinate value? i believe the svgs are equivalent either way.~~

wolfwood:
the files for comparison:
a
b

wolfwood:
when importing these two to openscad (without my patch), the location differs.

b sits at 0, ~445
a at 0, ~665

kintel:
The SVG issue seems related to #6161 - I'll ping that issue and see if there is a common solution.

wolfwood:
it seems to me that this is a real consequence of the viewboxes being different, but then the y axis inversion is doubling the effect of a non-zero minimum value, instead of cancelling out?

I need to read up on svg viewboxes / viewports. and probably find the issue where this was introduced to review the expected behavior.

PR #5691, opened 2025-02-22
union, difference, intersection and translate added

gsohler:
and object oriented calls
this will enable to even do some useful things


t-paul (approved):
It would be nice to reduce the code duplication, e.g. between `python_oo_csg_sub` and `python_csg_sub`, but let's keep going for now and have a look a that once most of the features are in.

PR #5693, opened 2025-02-23
Python pr5

gsohler:
now have minimal set to do useful design.
node cloning and python gil locks
enhanced python acceptance

### DIFF
--- a/src/core/node_clone.cc
+++ b/src/core/node_clone.cc
@@ -1,0 +1,112 @@
+/*
+ *  OpenSCAD (www.openscad.org)
+ *  Copyright (C) 2009-2011 Clifford Wolf <clifford@clifford.at> and
+ *                          Marius Kintel <marius@kintel.net>
+ *
+ *  This program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  As a special exception, you have permission to link this program
+ *  with the CGAL library and distribute executables, as long as you
+ *  follow the requirements of the GNU GPL in regard to all of the
+ *  software in the executable aside from CGAL.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program; if not, write to the Free Software
+ *  Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ *
+ */
+
+#include "src/geometry/linalg.h"
+#include "src/geometry/GeometryUtils.h"
+#include "src/core/primitives.h"
+#include "src/core/TransformNode.h"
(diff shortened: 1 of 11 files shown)


src/core/node_clone.cc
```diff
@@ -0,0 +1,112 @@
...
+#include "src/core/SurfaceNode.h"
+#include "src/core/TextNode.h"
+#include "src/core/OffsetNode.h"
+#include "src/core/ProjectionNode.h"
+#include "src/core/ImportNode.h"
+
+std::vector<ModuleInstantiation *> modinsts_list;
+
+#define NodeCloneFunc(T) std::shared_ptr<T> clone_what(const T *node) {\
+       	ModuleInstantiation *inst = new ModuleInstantiation(node->modinst->name() ,\
```
kintel, line 47: @gsohler Reviewing some old code: This is a naked pointer which is (I believe) never deleted. Nodes generally keep a const* to the ModuleInstantiation around for read-only purposes. The ModuleInstantiation is owned by the instantiating scope and we promise that the scope objects outlive the node tree. Could we instead just copy the pointer here?

gsohler:
Yep, the ModInstantiation concept  works out well in OpenSCAD,
I believe that all the memory referred in ModuleInstanciation actually
belongs to SourceFile andSourceFile goes out of scope only after recompile,
which is late enough.
However this concept, does not work out in PythonSCAD as there is no
SourceFile and I cannot use a pointer to a ModuleInstantiation, which goes
out of scope after the current function returns.

Here is our actual  solution in place:

https://github.com/pythonscad/pythonscad/blob/master/src/genlang/genlang.h#L16

every allocated ModInst  pointer is tracked and memory is freed
right before re-execution. (last bit is freed when program terminates, but
its not accumulated)
I think if you decided to make ModInstattion a shared pointer this would be
the nicest solution, but it's quite some effort.

kintel:
Not sure I understand: Every time you call e.g. `python_sphere()`, you'll allocate a new `ModuleInstantiation` and never free it. That does sound a bit like accumulation of memory to me. ..or perhaps I'm not finding the delete code?..

gsohler:
Nope, I *do* delete it, here:

https://github.com/pythonscad/pythonscad/blob/master/src/python/pyopenscad.cc#L961-#L964

Not sure if openscad got that "*feature". Neither I know, if that's an
acceptable  method to you or if know better options...
lets discuss

kintel:
OpenSCAD itself doesn't include the code to delete the Python-originating module instantiations. Doing this using a global variable which accumulates instantiations does feel a bit crude. I'm not familiar with the design of the Python stuff, so I don't really have a good suggestion. The typical way would be to build some object which represents a "python program", then keeping everything related to that program there. ..and if you open multiple programs, find a way of isolating them from each other.

gsohler:
Yeah, I suspected that you would not like the solution with the global
pointer accumulation,
this is why we don't yet have a solution yet.
In Python there is no need to do fancy stuff with ModuleInstantiation. I
just allocate the memory because the system requires a ModuleInstantiation
Memory Object.
A More pleasant way than the global pointer allocation could be to host 4
global ModuleInstantiation Variables like

mi_none, mi_highlight, mi_background and mi_root and supply pointers to
these appropriately.
We won't' face memory waste nor a global garbage collector, *just* 4 global
variables, which could even be hidden in a namespace.,

BTW Even Python has its own native "smart pointers" which automatically
free when they are not referenced anymore. Only issue is, that
it's a different concept than std::shared<ptr>  and combining both words
transparently could mean a big win for the future.

PR #5697, opened 2025-02-24
separate Python Target

gsohler:
helps me to pr the mxe build tweaks next


CMakeLists.txt
```diff
@@ -470,11 +470,37 @@ if(ENABLE_PYTHON)
     message(WARNING "Nettle not found, disabling python support.")
     set(ENABLE_PYTHON OFF CACHE BOOL "" FORCE)
   else()
-    message(STATUS "Python ${Python_VERSION} enabled, using Nettle ${Nettle_VERSION}")
-    target_include_directories(OpenSCAD PRIVATE ${Python_INCLUDE_DIRS})
-    target_link_libraries(OpenSCAD PRIVATE ${Python_LIBRARIES})
+    SET(PYTHON_SOURCES
```
vincele, line 473: Naive question: why not putting (almost) all of this in a `src/python/CMakeLists.txt` file ?

gsohler:
I did Not know and realize that this IS possible.it could make the File
much clearer.
In any Case, i believe there are different priorities on both Sides.


Vincent Legoll ***@***.***> schrieb am So., 13. Juli 2025,
15:15:

vincele:
Yeah, moving it outside will probably require a bit more work to properly reference the externalized thing, but would be nice to have, eventually...

PR #5701, opened 2025-02-25
Squashed some clang-tidy warnings, slightly adjusted clang-tidy config

kintel:
TODO:
* [x] Look into how to best silence `misc-include-cleaner` for 3rd party libraries; LibraryInfo.cc in particular -> This is a bit involved; deferring to later
* [x] We should explain the inclusions/exclusions in .clang-tidy for future reference

```
      [misc-const-correctness]   -29
        [misc-include-cleaner]  -397
     [misc-unused-alias-decls]    +1
[misc-use-anonymous-namespace]    -1
          [modernize-use-auto]    -1
       [modernize-use-emplace]    -1
```


kintel:
@hzeller Do you happen to have a good strategy for dealing with symbols from 3rd party libs? See `LibraryInfo.cc` and `openscad_gui.cc` in particular.

hzeller:
These are a annoying indeed. What these 3rd party headers seem to do is to provide public headers that then include private headers. The includes would work if marked with [`// IWYU pragma: export`](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md#iwyu-pragma-export) to be properly recognized by `clang-tidy` as an indirection and then things are good.

For Qt in particular, that is annoying as a public `QFooBar` header typically just has one private `#include "qfoobar.h"` without that annotation. Such IWYU addition would be neat to get upstreamed. 

Ideally we could tell `clang-tidy` that headers that match a particular pattern (e.g. `"Q*"`) are meant to be all export to silence that issue, but I am not aware that this is possible. Maybe something to file a feature request with clang-tidy.

A short-term workaround I could imagine is to switch off `misc-include-cleaner` for certain files (but then we also miss legitimate issues), or filter out certain symbols (`grep -v 'no header providing "Q'`  essentially) (might need to hack something into `run-clang-tidy-cached.cc`)

.clang-tidy
```diff
@@ -16,6 +18,10 @@ Checks: '
     -modernize-avoid-c-arrays,
     -modernize-use-nodiscard,
     -modernize-use-trailing-return-type,
-  performance-*
+  performance-*,
+    -performance-avoid-endl
```
hzeller, line 22: Another report I find worthwhile disabling is the `misc-use-anonymous-namespace`. It suggests to convert

```c++
static void foo() {}
```

to

```c++
namespace {
void foo() {} 
}
```

Which I think it is a counterproductive suggestion, bad for readability reasons. If I see `foo()` and want to verify that it is a local symbol I first would have to go through half the file on average to find an opening anonymous namespace. `static` makes that immeidately clear.

I think c++ briefly deprecated the use of `static` for this use-case in c++11 (which is why this report exist), but it was un-deprecated later.

What _is_ much more useful is `misc-use-internal-linkage`, which appeared in clang-tidy 19 or 20. It finds places that  should be static/in anonymous namespace but isn't.

kintel: I believe the case against `static` is that someone can `extern` a function which was intended to be local to the compilation unit. We somehow do that in a few places in OpenSCAD for unknown reasons.

hzeller: if it is `static`, it is equivalent to being in an anonymous namespace, i.e. it can not be referred to by `extern` somewhere else.

hzeller (commented):
clang-tidy syntax bikeshedding -- feel free to ignore :)

kintel:
Re bikeshedding: The reason I want list syntax is that I want to add per-check comments explaining why we enabled/disabled them, as this may be a moving target

.clang-tidy
```diff
@@ -1,21 +1,27 @@
...
+  - '-misc-no-recursion'
+  - '-misc-non-private-member-variables-in-classes'
+  - 'modernize-*'
+  - '-modernize-avoid-c-arrays'
+  - '-modernize-use-nodiscard'
+  - '-modernize-use-trailing-return-type'
+  - 'performance-*'
+  - '-performance-avoid-endl'
+
+CheckOptions:
```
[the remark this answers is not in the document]
kintel: Thx. key: value looks ok now, but good to know in case lines get too long in the future.

.clang-tidy
kintel: Right, I managed to confuse myself again :) I'll do another global run and look at the `misc-use-anonymous-namespace` check results.

kintel:
@hzeller btw., in terms of clang-tidy checks. There are still a handful of disabled groups of checks; things like `cppcoreguidelines`, `google`, `readability`. Just curious if you considered adding from them or if it ends up offering diminishing returns for longer runtime or smth.?

hzeller:
I think last time I looked the `cppcoreguidlines` were not too useful. I regularly do use the `readabilty-*` and `google-*` ones (but with a few rules I disable in these). The `clang-diagnostic-*` can also be useful.

Here are two projects where I mostly selected various `.clang-tidy` rules over time (some of the disabled rules only show up in later clang-tidy 19 or so)

https://github.com/chipsalliance/verible/blob/master/.clang-tidy
https://github.com/hzeller/bant/blob/main/.clang-tidy

hzeller:
I have some plans to filter the run given a set of filenames and emit json that can be used in the github API to add the comments, but probably a few weeks out as spare-time permits.

I was playing with https://github.com/marketplace/actions/clang-tidy-review before I think, but it didn't work for me at the time. But might be worthwhile playing with.

kintel:
ok, I'll try to give that linter tool a spin on some toy project first. If we're lucky, the authors built something somewhat composable.

PR #5717, opened 2025-02-28
Adding Handles(matrices)  to objects

gsohler:
Handles makes it easy to align objects together

Here is an example:

from openscad import *

c=cube([10,10,10])

\# translate the origin with an offset, so top_center sits on top of the cube
c.top_center=translate(c.origin , [5,5,10])

 \# This one even points to the right side
c.right_center=translate(rotate(c.origin,[0,90,0]),[10,5,5])
 
   \# Create something to align with
cyl = cylinder(d=3,h=5,fn=10)

   \# This placecs cyl onto the top and right side of the cube 
   ]#    obj       source handle  dest handle

c |= cyl.align(c.top_center,cyl.origin)
c |= cyl.align(c.right_center,cyl.origin)

c.show()

   \#of course, any data can be stored and retrieved  with objects , not only handles
   \# objects with  handles  are transformed,  their handles make same transform

image


amatulic:
This looks similar to the [anchor and attachment features](https://github.com/BelfrySCAD/BOSL2/wiki/Tutorial-Attachments) offered by BOSL2. If you scroll down to the "[positioning children](https://github.com/BelfrySCAD/BOSL2/wiki/Tutorial-Attachments#positioning-children)" section, it shows examples similar to what you describe.

t-paul (approved):

PR #5725, opened 2025-03-02
Provide GLSL vertex shader to OpenCSG to fix z-artifacts (#4595)

floriankirsch:
I have just released OpenCSG 1.8.0, which contains a change to address https://github.com/floriankirsch/OpenCSG/issues/8, which is the twin-bug of https://github.com/openscad/openscad/issues/4595

It is now possible to pass a GLSL vertex shader to OpenCSG. If the same vertex transformation is performed as in the later shader to shade the geometry, then the z-artifacts do not happen anymore. This is what is implemented in this commit.

The OpenCSG submodule is updated to a pre-release only. But effectively there are no software changes missing, only the documentation is not up-to-date.


kintel:
It somehow didn't make a difference for be; re-enabled test to see how this plays out on the CI.

kintel:
Things seem to work OK using SW rendering, but I get this locally when rendering a sphere subtracted from a cube:
view-options-tests-expected

kintel:
I found the issue: OpenCSG's GLSL shader requires OpenGL 3.0 (https://github.com/floriankirsch/OpenCSG/blob/2762e5e9d84dbf28ed4315ceb406ccbeece0198b/src/renderSCS.cpp#L308-L320), while macOS is stuck at 2.1, since Apple doesn't implement the compatibility profile for newer versions of OpenGL.

It can be fixed by doing smth. like this:
```
        static const char mergeFragmentProgram2D[] =
            "#version 120\n"
            "uniform sampler2D texture0;\n"
            "uniform vec4 color;\n"
            "uniform float viewWidth;\n"
            "uniform float viewHeight;\n"
            "void main() {\n"
            "    vec2 texCoord = gl_FragCoord.xy / vec2(viewWidth, viewHeight);\n"
            "    vec4 temp = texture2D(texture0, texCoord);\n"
            "    temp = temp - color;\n"
            "    if (dot(temp, temp) > 0.000001)\n"
            "        discard;\n"
            "    gl_FragColor = color;\n"
            "}\n";
```

..and populate the new uniforms with these values:
```
            const int width = OpenGL::canvasPos[2] - OpenGL::canvasPos[0];
            const int height = OpenGL::canvasPos[3] - OpenGL::canvasPos[1];
```

floriankirsch:
Thanks for the diagnosis. This obviously will require yet another fix in OpenCSG.

An alternative solution to yours may be to fallback to the texture rectangle path, that probably already works with OpenGL 2.0. To test this, after https://github.com/floriankirsch/OpenCSG/blob/2762e5e9d84dbf28ed4315ceb406ccbeece0198b/src/frameBufferObject.cpp#L73 one may set the target to GL_TEXTURE_RECTANGLE_ARB unconditionally. 

But the ARB texture rectangle is optional only. It is very well possible that MacOS does not support it. So I think your fix must be implemented in any case.

kintel:
FYI: Updating OpenCSG to 9c5ebf5afaad3197c894b20f58ca7954929c52ea fixes the remaining issue on macOS.

floriankirsch:
This is good to know, thank you. In particular because I do not have a Mac at my possession.

I must do internal (partly unrelated) testing myself. You can expect the next official release of OpenCSG with this fix probably the next weekend. Of course you can update / merge my PR whenever you want :-)

kintel:
Sounds good. I'll try to merge this, then update to the next OpenCSG release when it lands.

kintel:
Humm, the new failures on macOS are puzzling. Everything seems to work well on native GPU, but fails on the SW renderer. Do you any pattern that could explain that? You should be able to download the test report from https://github.com/openscad/openscad/actions/runs/13669151792?pr=5725

kintel:
@floriankirsch More debugging reveals this:
* SCSChannelManagerAlphaOnly yields bad results on the macOS SW renderer. This ChannelManager was never used in the past since the ARB shader version was always preferred if available.
* This behavior change happened in https://github.com/floriankirsch/OpenCSG/commit/23ce133eba89e48b10ab6a127c4dce66b9019e3f
* If I revert the change in question (i.e. pass false to `needMoreThan255Primitives` in `getChannelManager()`), everything works again.

I haven't yet attempted to debug `SCSChannelManagerAlphaOnly` to understand why it fails..

floriankirsch:
See the comment in https://github.com/floriankirsch/OpenCSG/blob/e9f1dc81c64b7069ea6ae45cbd01b7ac7fd045d1/src/renderSCS.cpp#L99 - I think it is the same issue. Basically I think that is is not possible to implement the alpha test with GL_EQUAL and a float in a robust way. OpenGL even specifies the way how the float reference value is converted to a fixed-point int, but not all drivers adhere to it. 

This was originally the reason why the ARB shader path was developped: There, GL_LESS can be used for the alpha test. 

I must think about this a day. Likely I will revert that change. This issue, and also the issue that I wanted to solve (the failure of Nouveau drivers to implement position invariance in the ARB shader path) both are driver problems IMHO. But getting a fix is more likely from the Nouveau direction.

In any case, if you force the Goldfeather algorithm, both these problems should not occur.

kintel:
Testing with goldfeather, just to prove that a happy path can be achieved that way. Whether OpenSCAD should _always_ prefer Goldfeather for headless previews, if we should prefer SCS, or make it an option remains to be discussed. I'm not too happy about exposing details like that as options though.

If you have any gut feeling around this, I'd love to hear it :)

floriankirsch:
I cannot really interpret the Goldfeather run; it looks quite red for me :-(

Anyway, my current plan would be to revert https://github.com/floriankirsch/OpenCSG/commit/23ce133eba89e48b10ab6a127c4dce66b9019e3f for now and do the next release as is. Reason: the completely missing geometry is somehow worse than the z-artifacts that would be reintroduced by the revert.

kintel:
Yeah, for some reason, the z artifacts are still there with Goldfeather.
The additional failures are because Goldfeather deals with flipped faces differently than SCS. Also not sure why - I can observe the same effect in the OpenSCAD GUI. It's a rare case, so not very important to deal with right now.

All this is a bit hard to test, since even on a Mac, the result depends on which renderer is used, and they may behave slightly different (Apple Silicon, nVidia, SW on Apple Silicon, SW on x86). GitHub only tests the last variant.

I'll run some tests with this branch as it, but with https://github.com/floriankirsch/OpenCSG/commit/23ce133eba89e48b10ab6a127c4dce66b9019e3f  reverted and see if that gives us good results on the 4 variants mentioned.

kintel:
OK, bumped everything to the correct version and removed my Goldfeather hack. For some reason, the 3 tests in question still fail with z artifacts. No idea what's going on; I'm 90% sure I saw this working sometime last week...
I guess I'll have to painfully bisect or try to reason about it.

Let me know if you have any ideas.

floriankirsch:
I do not really believe that this is a recent regression. Or at least I do not see where it would come from.

From the list of test runs at https://github.com/openscad/openscad/actions the only run for #4595 on MacOS that suceeded was
https://github.com/openscad/openscad/actions/runs/13619815782 - but in that run these 3 tests were still disabled. You then enabled the tests and updated the reference images due to the different edges. Do you have these images in mind?

The actual bug is puzzling. I have never seen an artifact like this.

kintel:
It's weird. I managed to chase the issue to our own fragment shader, which is the one we use to render objects with depth func GL_EQUAL: Whenever I call the GLSL function `dFdy()`, it messes up the depth values written, causing horizontal artifacts due to the depth test randomly failing for some scanlines. If I axis-align the camera, it works again.

As this is only triggered on the Apple SW renderer, I'm tempted to call this an Apple SW renderer bug, remove the tests again and move on.

floriankirsch:
Maybe you can emulate dFdy() as described in https://stackoverflow.com/questions/22442304/glsl-es-dfdx-dfdy-analog.

kintel:
> Maybe you can emulate dFdy() as described in https://stackoverflow.com/questions/22442304/glsl-es-dfdx-dfdy-analog.

That workaround only works in extreme cases where axis aligned quads are rendered. For other geometry, quality isn't good enough. Will disable tests for now, hopefully Apple's SW renderer catches up in the future.

kintel:
All green - thanks a lot for pulling through!

damienmarchal:
Hi, 

I just give it a try on a recent checkout of openscad and still have the z-fighting artifact as in: 
image

Do I have to compile openscad with something special to have it fixed or update version of gl ?
Currently it is using: 
```raw
Extended renderer info (GLX_MESA_query_renderer):
    Vendor: Intel (0x8086)
    Device: Mesa Intel(R) UHD Graphics (CML GT2) (0x9bc4)
    Version: 23.0.4
    Accelerated: yes
    Video memory: 31719MB
    Unified memory: yes
    Preferred profile: core (0x1)
    Max core profile version: 4.6
    Max compat profile version: 4.6
    Max GLES1 profile version: 1.1
    Max GLES[23] profile version: 3.2
```

damienmarchal:
So it makes sense it fails on my system because a git submodule update tells me that the reference to openCSG/313018fbf997f484f66cb4a320bbd2abf79a4fc1 is not there and thus I'm not surprised that in OpenCSGRenderer::draw  the #if OPENCSG_VERSION >= 0x0180 fails.

PR #5726, opened 2025-03-03
Upgraded old string-based connects to new functor-based syntax

### LINKED ISSUE #5547: Convert old Qt signal connection scheme to the typesafe new one

scat25:
I replaced all the old string-based connects with the new functor-based syntax.
I made some small changes in some Interfaces to achieve this, see the changed header files. Overloaded functions were resolved with QOverload where applicable.

The code compiles and I can run OpenSCAD.

This fixes #5547


src/gui/MainWindow.cc
```diff
@@ -563,67 +561,67 @@ MainWindow::MainWindow(const QStringList& filenames) :
...
   const QString cs = Preferences::inst()->getValue("3dview/colorscheme").toString();
   this->setColorScheme(cs);
 
   //find and replace panel
-  connect(this->findTypeComboBox, SIGNAL(currentIndexChanged(int)), this, SLOT(selectFindType(int)));
-  connect(this->findInputField, SIGNAL(textChanged(QString)), this, SLOT(findString(QString)));
-  connect(this->findInputField, SIGNAL(returnPressed()), this->findNextButton, SLOT(animateClick()));
+  connect(this->findTypeComboBox, QOverload<int>::of(&QComboBox::currentIndexChanged), this, &MainWindow::selectFindType);
+  connect(this->findInputField, &QWordSearchField::textChanged, this, &MainWindow::findString);
+  connect(this->findInputField, &QWordSearchField::returnPressed, this->findNextButton, [this]{this->findNextButton->animateClick();});
```
kintel, line 609: Out of curiosity: Why did you end up using a lambda for this slot (and the one below)?

kintel (commented):
One minor question

src/gui/MainWindow.cc
scat25: The compiler complained about missmatching parameters. animateClick has a parameter with a default value. A lambda was the easy solution

kintel: Oh, I see - they removed the argument in Qt6, but we need it for Qt5 builds

Issue #5738, opened 2025-03-13
Manifold render test failure: example017


kintel:
`rendermanifoldtest-different_example017` is failing:

<img width="1394" alt="Image" src="https://github.com/user-attachments/assets/c647a5b9-5491-48b1-ac28-fcbc4895b62d" />

This seems to be an issue where Manifold is not preserving face attributes for faces which lie in the same plane.

Minimal example:
```
translate([0, 0, 24])
  linear_extrude(height = 6)
    difference() {
     translate([0,20]) square(30, center=true);
     translate([0, 16]) square([6, 10], center=true);
    }

color([0.7, 0.7, 1]) 
translate([3, 32, 0])
  rotate([90, 0, -90])
    linear_extrude(height = 6) 
      polygon([
        [27, 0],
        [27, 23],
        [21, 23],
        [21, 30 -0],
        [11, 30 -0],
        [11, 23],
        [0, 23],
        [0, 0],
      ]);
```

Before/after screenshots. The expectation is that Manifold should preserve the face attributes for the "peg" which is fitted into the slot, but instead it simplifies the entire top surface to a single color:

<img width="1295" alt="Image" src="https://github.com/user-attachments/assets/e596b9bb-f4e7-4ae1-a259-862e50f61e41" />
<img width="1295" alt="Image" src="https://github.com/user-attachments/assets/4a4c01c3-8b6c-4d73-a1b0-dbe2acbd3dbc" />

kintel:
TODO: Bump Manifold to latest master and re-test. If still failing, open an issue upstream.

kintel:
With manifold master - more simplification applied, but same underlying issue:

<img width="1295" alt="Image" src="https://github.com/user-attachments/assets/8ec82b74-2ba1-4104-9581-c895d237aa64" />

elalish:
Interesting - looks like we need to reproduce this as our own `TEST` case. I am working on this, so hopefully fixed once I can repro.

kintel:
@elalish Did you get around to reproducing this? If not, is there anything I can do to assist?

elalish:
Ugh, alas no, I've been making pretty spotty progress lately - trying to organize more time. It would be very helpful if you could make a PR to add a TEST that reproduces the problem (and causes the CI to fail). You can use our test function `RelatedGL` to catch this kind of error.

elalish:
Okay, I've got a repro of this now - very weird. I'll take a look.

elalish:
[...]

PR #5740, opened 2025-03-14
Bump manifold to latest master

kintel:


kintel:
@pca006132 Latest manifold works. It didn't have any positive effect on #5738, but that was expected.
I'll try to run some performance measurements, but otherwise this looks good to merge.

PR #5750, opened 2025-03-16
Ranges where start is "after" end (whichever way that is) yield nothing.  Fixes #5721.

### LINKED ISSUE #5721: DEPRECATED: Using ranges of the form [begin:end] with begin value greater than the end value is deprecated...
I'm getting this a lot these days, and it's super annoying.  I'm either looking for a easy fix, or asking humbly for it to be un-depriciated if we can't find one.

Here's my use case:

I'd like to iterate over a vector that may have any non-negative number of elements (including 0).

I want to write:

`for (a=[0:len(my_vector)-1])`, and that used to work, and now it doesn't when my_vector=[].

My current work around is to do this:

```
for (a=[0:len(my_vector)])
    if a!=len(my_vector)
```
But that's super ugly to put everywhere.

This also applies to list comprehensions.

I think iterating over a 0-length vector should produce a 0-length iteration (no-op), and doing a list comprehension over a 0 length vector should result in `[]`.  Makes sense to me.  Thoughts?

Version: 2025.02.25.snap

jordanbrown0:
Changes implicit-step ranges to be the same as explicit-step ranges:
* Start "after" end yields nothing.
* Warn on an all-constants range that is "backwards".

Note:  this will fail several tests.  I have the test fixes in hand, but I wanted people to have a chance to see the test failures.

### DIFF
--- a/src/core/Expression.cc
+++ b/src/core/Expression.cc
@@ -234,53 +234,33 @@ Range::Range(Expression *begin, Expression *step, Expression *end, const Locatio
 {
 }
 
-/**
- * This is separated because both PRINT_DEPRECATION and PRINT use
- * quite a lot of stack space and the method using it evaluate()
- * is called often when recursive functions are evaluated.
- * noinline is required, as we here specifically optimize for stack usage
- * during normal operating, not runtime during error handling.
- */
-static void NOINLINE print_range_depr(const Location& loc, const std::shared_ptr<const Context>& context){
-  LOG(message_group::Deprecated, loc, context->documentRoot(), "Using ranges of the form [begin:end] with begin value greater than the end value is deprecated");
-}
-
 static void NOINLINE print_range_err(const std::string& begin, const std::string& step, const Location& loc, const std::shared_ptr<const Context>& context){
   LOG(message_group::Warning, loc, context->documentRoot(), "begin %1$s than the end, but step %2$s", begin, step);
 }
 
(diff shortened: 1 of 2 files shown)


src/core/Expression.cc
```diff
@@ -234,53 +234,33 @@ Range::Range(Expression *begin, Expression *step, Expression *end, const Locatio
 {
 }
 
-/**
- * This is separated because both PRINT_DEPRECATION and PRINT use
- * quite a lot of stack space and the method using it evaluate()
- * is called often when recursive functions are evaluated.
- * noinline is required, as we here specifically optimize for stack usage
- * during normal operating, not runtime during error handling.
- */
```
t-paul, line 243: We should leave that comment for the other print method.

jordanbrown0: Someone who is more C++ savvy than I am should consider whether that comment is still accurate, given that Range::evaluate() and LOG() are their own functions.  In C, there would be no stack concern.

kintel: I don't quite understand that comment, and also not sure how those functions use "quite a lot of stack space", neither now or when the comment was written. Not sure it's worth taking any action on it now anyway.

kintel (approved):

jordanbrown0:
Updated documentation.
https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/General#Ranges

PR #5756, opened 2025-03-18
Manifold export is always predictible

kintel:
We don't need to set the cmd-line option


kintel:
Closing - turns out "preductible" is limited to run-to-run predictability on the same machine

PR #5763, opened 2025-03-20
Allow font features to be enabled via the fontconfig specification.

t-paul:
Example: "Linux Biolinum O:style=Regular:fontfeatures=+zero"

Multiple features need to be separated by semicolon: "Linux Biolinum O:style=Regular:fontfeatures=+smcp;+onum"

See also: https://en.wikipedia.org/wiki/List_of_typographic_features

Screenshot from 2025-03-20 19-03-01


src/FontCache.cc
```diff
@@ -408,36 +410,50 @@ FT_Face FontCache::find_face_fontconfig(const std::string& font) const
...
-  if (FcPatternGet(match, FC_INDEX, 0, &font_index)) {
+  if (FcPatternGet(match, FC_INDEX, 0, &font_index) != FcResultMatch) {
     return nullptr;
   }
 
-  FT_Face face;
-  const FT_Error error = FT_New_Face(this->library, (const char *) file_value.u.s, font_index.u.i, &face);
+  FcValue font_features;
+  std::string font_features_str;
+  if (FcPatternGet(match, FC_FONT_FEATURES, 0, &font_features) == FcResultMatch) {
```
jordanbrown0, line 419: Maybe FcPatternGetString().

src/FontCache.cc
```diff
@@ -408,36 +410,50 @@ FT_Face FontCache::find_face_fontconfig(const std::string& font) const
...
-    if (!charmap_set) charmap_set = try_charmap(face, TT_PLATFORM_ISO, TT_ISO_ID_7BIT_ASCII);
-    if (!charmap_set) LOG(message_group::Font_Warning, "Could not select a char map for font '%1$s/%2$s'", face->family_name, face->style_name);
+    if (!charmap_set) charmap_set = try_charmap(face->face_, TT_PLATFORM_MICROSOFT, TT_MS_ID_UNICODE_CS);
+    if (!charmap_set) charmap_set = try_charmap(face->face_, TT_PLATFORM_ISO, TT_ISO_ID_10646);
+    if (!charmap_set) charmap_set = try_charmap(face->face_, TT_PLATFORM_APPLE_UNICODE, -1);
+    if (!charmap_set) charmap_set = try_charmap(face->face_, TT_PLATFORM_MICROSOFT, TT_MS_ID_SYMBOL_CS);
+    if (!charmap_set) charmap_set = try_charmap(face->face_, TT_PLATFORM_MACINTOSH, TT_MAC_ID_ROMAN);
+    if (!charmap_set) charmap_set = try_charmap(face->face_, TT_PLATFORM_ISO, TT_ISO_ID_8859_1);
+    if (!charmap_set) charmap_set = try_charmap(face->face_, TT_PLATFORM_ISO, TT_ISO_ID_7BIT_ASCII);
+    if (!charmap_set) LOG(message_group::Font_Warning, "Could not select a char map for font '%1$s/%2$s'", face->face_->family_name, face->face_->style_name);
```
jordanbrown0, line 453: Maybe pass `face` down to `try_charmap()` and let it pull out `face_`.

src/FontCache.cc
```diff
@@ -337,13 +337,15 @@ void FontCache::check_cleanup()
       pos = it;
     }
   }
-  FT_Done_Face((*pos).second.first);
+  const FontFace *face = (*pos).second.first;
   this->cache.erase(pos);
+  FT_Done_Face(face->face_);
+  delete face;
```
jordanbrown0, line 343: Maybe use smart pointers?

jordanbrown0:
Not really a comprehensive review, just a few ideas.  Basic idea looks sound, nothing critical.

We should find out what other consumers use for a separator for this string.

src/FontCache.cc
t-paul: The benefit seems superficial, but why not. They took the time to provide those :grinning:.

src/FontCache.cc
t-paul: Done.

src/FontCache.cc
```diff
@@ -337,13 +337,15 @@ void FontCache::check_cleanup()
       pos = it;
     }
   }
-  FT_Done_Face((*pos).second.first);
+  const FontFace *face = (*pos).second.first;
   this->cache.erase(pos);
+  FT_Done_Face(face->face_);
```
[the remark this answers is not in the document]
t-paul: Yeah, has to now, due to std:shared_ptr().

src/FontCache.cc
t-paul: I considered it, but did not want to do all the typing. In the end it was not a big change at all. Probably safer this way.

src/FontCache.cc
jordanbrown0: Type safety is good.

PR #5781, opened 2025-03-27
Bump manifold. Fixes #5766

### LINKED ISSUE #5766: Manifold: projection(cut=true) crashes on empty objects
Empty objects can happen in various ways, e.g. importing an empty or non-existing file, creating a zero-volume boolean.

Examples:

```
projection(cut=true) import("notfound.stl");
```

```
projection(cut=true) {
  intersection() {
    cube(center=true);
    translate([2,0,0]) cube(center=true);
  }
}
```

Cmd-line:
```
openscad file.scad -o out.png --backend=manifold
```

kintel:


kintel:
The new manifold triggers another issue, see https://github.com/elalish/manifold/pull/1194#discussion_r2017757687

pca006132:
It triggers CGAL assert error:

```
Expression : 
File       : /nix/store/94a41wjpnccb9v826cghdpjjj63n7j23-cgal-5.6.2/include/CGAL/Convex_decomposition_3/SM_walls.h
Line       : 440
Explanation: wrong handle
ERROR: Nef minkoswki operation failed: CGAL ERROR: assertion violation!
File: /nix/store/94a41wjpnccb9v826cghdpjjj63n7j23-cgal-5.6.2/include/CGAL/Convex_decomposition_3/SM_walls.h
Line: 440
Explanation: wrong handle
```

I think it is fundamentally another issue, probably because CGAL Nef polyhedra input requirement is violated by manifold's output?

kintel:
It's possible - I was hoping it rang a bell on your end :) I'll dig a bit deeper and see what I find.

pca006132:
I checked the differences, it seems that using the fastacos can cause larger error in the normal computation. At most ~8°. Replacing it with the acos in cosmopolitan fixes the issue.

The reason we want to replace the acos function is to get deterministic results across machines.

kintel:
CMake on macOS got updated to 4.0.0 - I suspect that may have triggered the build issue

PR #5790, opened 2025-03-31
Update mimalloc to 1.7.9

pludi:
Update `mimalloc` to 1.7.9 in order to allow building with current CMake releases that don't support certain features from pre-3.5 versions anymore.


t-paul (approved):
Thanks!

PR #5794, opened 2025-04-02
Temporarily downgrade CMake to 3.31.6 on macOS, until we're ready for CMake 4

kintel:


kintel:
Note: This fails since we build lib3mf with Homebrew, which will pull CMake 4. To mitigate that, we need to patch lib3mf in our homebrew tap to support CMake 4.

kintel:
lib3mf updated in homebrew in https://github.com/openscad/homebrew-tap/pull/10

PR #5798, opened 2025-04-02
Add format() function based on the fmt library.

t-paul:
```openscad
echo(format("hallo {name:*>8}! {r} {n} {3}",
    name = "welt",
    n = 42,
    r = [1:0.5:6],
    [2,2.4, "blah"]
));

// ECHO: "hallo ****welt! [1:0.5:6] 42 [2, 2.4, blah]"
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -343,6 +343,11 @@ if (MSVC)
   target_include_directories(OpenSCAD SYSTEM PRIVATE ${HARFBUZZ_INCLUDE_DIRS}/harfbuzz)
   target_link_libraries(OpenSCAD PRIVATE harfbuzz::harfbuzz)
 
+  find_package(fmt REQUIRED)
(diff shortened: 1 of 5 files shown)


gsohler:
suppose exceptions are cought on a higher level(when there is an argument count mismatch)

jordanbrown0:
Does this support specifying the number of decimal places?  Does it support specifying the width of a field, notably getting right justification on numbers?

Where I ran into trouble trying to do this was that I couldn't get it to work when the width or number of decimal places was itself a variable.

t-paul:
@jordanbrown0 I had not. It seems our lack of integers makes this especially difficult. I can get this relatively easy to work when pretending to add integers for the dynamic spec parameters. I don't have a good solution yet covering everything. With some hacking it seems possible.

jordanbrown0:
Yeah, the library has a lot of flexibility for handling the values being formatted, but not a lot for handling those formatting parameters.

Issue #872, opened 2014-07-13
Square bracket ( "[" ) issue on some keyboard layouts


tehcereal:
There is an issue when I try to type square brackets in the editor my shortcut for the character is altgr+F and openscad treats it like I pressed Ctrl+Alt+F (Find and replace) menu. The ] character works as its should. If I press the [ shortcut (altgr+f) when the menu is open it just highlights the text box. I'm using Windows 8.1 and OpenSCAD 2014.03 and my keyboard layout is Croatian but i think the problem occurs on every layout that has the [ character on altgr+f combination.

AakaFosfor:
I've solved this issue by using an external editor - actual stable version of OpenSCAD is hooked-up on the opened file and automatically re-displays when I save the file in my editor.

kuzfix:
Ctrl+Alt+F is standard shortcut for "[" on a lot of keyboard layouts (Slovene in my case). It would be nice if the developers either didn't use it for anything or make it possible to customize the shortcuts. Write now I still use an external editor the way AakaFosfor describes since this shortcut clash renders the built-in editor practically unusable.

kintel:
@kuzfix Which OS are you on?

kuzfix:
Hello @kintel. I am currently using Windows 10, but I had the same issue with some other programs as well on windows XP, 7 and 8.1. 
The Right alt (altgr) on Slovene keyboard acts as Ctrl+Alt and in combination with some characters gives square brackets and similar. Therefore any program that uses Ctrl+Alt+ any of those characters for a shortcut will not allow me to use any of those characters.

If it is any help, here is a short list of the most problematic shortcuts:
Ctrl+Alt+f   -   "["
Ctrl+Alt+g   -   "]"
Ctrl+Alt+v   -   "@"
Ctrl+Alt+b   -   "{"
Ctrl+Alt+n   -   "}"
Ctrl+Alt+q   -   "\"
Ctrl+Alt+w   -   "|"
Ctrl+Alt+e   -   "€"
Ctrl+Alt+1   -   "~"
Ctrl+Alt+2   -   "ˇ"
Ctrl+Alt+3   -   "^"
Ctrl+Alt+5   -   "°"

Here is a link to a picture of a Slovene keyboard: 
http://www.shrani.si/f/1G/lI/4uw1nCMk/g80-3000-lsmsl.jpg

Whatever symbol is in the lower right corner of a key will get typed if you press Ctrl+Alt+"that key".

[...]

PR #5801, opened 2025-04-03
Fix QKeySequence for Find and Replace

### LINKED ISSUE #5760: Ctrl+Shift+F doesn't work in Windows client
**Describe the bug**
The menu shows **Ctrl+Shift+F** as a shortcut for the **Edit > Find and Replace** function, but when I try this key combination it does nothing.

**To Reproduce**
1. Open a .scad file in the client.
2. Hold down Ctrl and Shift and press F.
3. Nothing happens.

**Expected behavior**
I had hoped for the same behavior as selecting **Edit > Find and Replace**  from the menu.

 - OS: Windows 11
 - System: Intel PC 64-bit
 - OpenSCAD Version 2025.03.19 (git 2c9f43c00) downloaded from openscad.org website.

Sparsh-N:
Fixes #5760 

Seems like the only issue is not adding each key action, which is how it's used elsewhere in the file. I've built it locally with MSYS2 to confirm the changes work, also on Windows 11.

### DIFF
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -506,7 +506,7 @@ MainWindow::MainWindow(const QStringList& filenames) :
   connect(this->editActionFind, &QAction::triggered, this, &MainWindow::showFind);
   connect(this->editActionFindAndReplace, &QAction::triggered, this, &MainWindow::showFindAndReplace);
 #ifdef Q_OS_WIN
-  this->editActionFindAndReplace->setShortcut(QKeySequence(Qt::CTRL, Qt::SHIFT, Qt::Key_F));
+  this->editActionFindAndReplace->setShortcut(QKeySequence(Qt::CTRL + Qt::SHIFT + Qt::Key_F));
 #endif
   connect(this->editActionFindNext, &QAction::triggered, this, &MainWindow::findNext);
   connect(this->editActionFindPrevious, &QAction::triggered, this, &MainWindow::findPrev);


kintel:
The inverse of this exact change was made in #4913 - it could be a Qt5 vs. Qt6 issue. Could you make sure to also test on Qt6, as that seems to not have compiled earlier.

Sparsh-N:
I was able to build it without the QKeySequence, and it seems like Ctrl+Alt+F works (although this is implemented in ui_MainWindow.h, line 1455). Would it be better to check for Q_OS_WIN here instead, and rebind the shortcut? The alternative would be modifying the MainWindow.cc case to also check QT version, although I think the former option would be simpler. Thanks!

kintel:
ui_MainWindow.h is generate from MainWindow.ui, so it might not be easy to add an OS check there..

kintel:
FYI: The original issue was related to AltGr: #872

Sparsh-N:
I'll rebase the commits soon, but from what I can tell, the option to use tr() works for both Qt5 and Qt6:

https://doc.qt.io/qt-6/qkeysequence.html
https://doc.qt.io/qt-5/qkeysequence.html

In both examples, tr("Ctrl+A") is also provided as an option alternative to using QtKey.

An alternative would be to check the Qt Version and set the shortcut accordingly.

    #ifdef Q_OS_WIN
        #ifdef QT_VERSION < QT_VERSION_CHECK(6, 0, 0) // Windows, Qt 5
            this->editActionFindAndReplace->setShortcut(QKeySequence(Qt::CTRL + Qt::SHIFT + Qt::Key_F));
        #endif
        #ifdef QT_VERSION >= QT_VERSION_CHECK(6, 0, 0) // Windows, Qt 6
            this->editActionFindAndReplace->setShortcut(QKeySequence(Qt::CTRL, Qt::SHIFT, Qt::Key_F));
        #endif
    #endif

However, I haven't been able to build it locally with MSys using Qt6 (many errors regarding deprecated methods also show up). I've tried referencing the github workflow which uses Ninja, but that also didn't build. I've been able to build with Qt5 and confirm it works.

kintel:
It’s a bit concerning that the Qt6 version doesn’t build locally, as we’re hoping to deprecate Qt5 soon

Sparsh-N:
Just built locally again, and the shortcut works with Qt6. I could look into not using tr() for the call but was wondering if this would bring any benefit?

image

kintel:
Nice. In terms of `tr()` - why not use just the string? I don't see a need to translate the keybindings per language. We also don't use tr() for translations in OpenSCAD :)

Sparsh-N:
I built it without the tr() call as well, and it also worked fine. I also built it using the if/else condition, so the options for the fix boil down to:


    #ifdef Q_OS_WIN
        #if QT_VERSION < QT_VERSION_CHECK(6, 0, 0) // Qt 5
            this->editActionFindAndReplace->setShortcut(QKeySequence(Qt::CTRL + Qt::SHIFT + Qt::Key_F));
        #endif
        #if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0) // Qt 6
            this->editActionFindAndReplace->setShortcut(QKeySequence(Qt::CTRL | Qt::SHIFT | Qt::Key_F)); // Qt6 doc uses | delimiters to append keys, not ,
        #endif
    #endif

OR

    #ifdef Q_OS_WIN
        this->editActionFindAndReplace->setShortcut(QKeySequence("Ctrl+Shift+F"));
    #endif

For consistency, the first option might look closer as I haven't seen QKeySequence directly with the shortcut string used to bind a shortcut. However, given both would work, is there a preference over either option? Both work across Qt5/Qt6 build on windows.

kintel:
I say the latter one is more readable

Sparsh-N:
Removed the translation, it should be good for merging now :)

kintel (approved):

PR #5813, opened 2025-04-10
Add the pt-BR translation

afmenez:
Add the pt-BR translation.  Still needs improvement, but I believe it is already usable.


kintel (approved):

afmenez:
The macOS build break doesn't seem to be related to my patch...

PR #5814, opened 2025-04-10
Improve the spanish translation

afmenez:
I am not fluent in Spanish, but was able to improve it a little. Removed automatic translations that were obviously wrong. Added a few easy translations. It is nor usable yet, but closer to.


kintel (approved):

afmenez:
The macOS build break doesn't seem to be related to my patch...

PR #5823, opened 2025-04-13
Fix checking whether Dock widgets are floating

cjmayo:
The QDockWidget::topLevelChanged() signal and its topLevel parameter are reporting changes of the floating property.

Fortunately QWidget::isTopLevel() is now deprecated.


kintel:
@damienmarchal Since you looked into this in more detail than me: Those this look good to you?

kintel (approved):

PR #5825, opened 2025-04-13
Bugfix: Don't generate nullptr geometry on errors

kintel:
This fixes a crash happening when a non-manifold geometry is not repairable.


jordanbrown0:
What is the goal and user visible effect?

kintel:
> What is the goal and user visible effect?

Updated description

PR #5829, opened 2025-04-17
Import backup files -  Improve the User experience

gsohler:
Programs and algorithms are not perfect and, sometimes sh*t happens.

But we can easily offer  to restore the valued models n one click rather having to google the location of the backup directory once again.

Right timed for Easter: Help the folks to find  their treasure ...


src/gui/MainWindow.ui
```diff
@@ -271,6 +271,7 @@
     <addaction name="menuPython"/>
     <addaction name="separator"/>
     <addaction name="fileShowLibraryFolder"/>
+    <addaction name="fileShowTreasureTrove"/>
```
kintel, line 274: Perhaps name this smth. like "Open File from Backup" and put it higher up in the file menu?

src/gui/MainWindow.cc
```diff
@@ -1747,6 +1748,36 @@ void MainWindow::actionShowLibraryFolder()
...
+  QString dirname (PlatformUtils::backupPath().c_str());
+  QStringList extensions = {};
+  extensions << "scad" << "csg";
+#ifdef ENABLE_PYTHON
+    extensions << "py";
+#endif
+  extensions.replaceInStrings(QRegularExpression("^"), "*.");
+  const auto filter = QString("OpenSCAD Backups (%1)").arg(extensions.join(" "));
+
+  // Bug in QT: dirname is ignored for getOpenFileName
```
kintel, line 1762: We use getOpenFileName() with dirname in multiple other places. Could you clarify which bug this is, tag the appropriate QTBUG, and protect the workaround with a QT_VERSION_CHECK macro as needed?
If we need to also change the other instances of getOpenFileName(), that probably deserves a GitHub issue.

src/gui/MainWindow.cc
```diff
@@ -1747,6 +1748,36 @@ void MainWindow::actionShowLibraryFolder()
...
+  std::ostringstream stream; // compile restored design
+  stream << "/*\n"; // put into comments in case it crashes (again)			     
+  std::ifstream file( filename.toStdString());
+  if ( file ) {
+    stream << file.rdbuf();
+    file.close();
+  }
+
+  stream << "\n*/\n";		    
+  activeEditor->setText(stream.str().c_str());
```
kintel, line 1778: I think it's easier to keep this within Qt: Use a QString-backed QTextStream. That way you don't have to convert to/from std::string

src/gui/MainWindow.cc
gsohler: Thank you for the hint on looking ino the other examples of getOpenFileName().
There are some, but only the one UIUtils:::openFile makes use of a predefined directory.
This one *and* mine only show some collection of *latest* files but not *all* filtered files from the specified directory.

image
This is not strictly from backups directory , and far from complete.

Is the function working correct in your place instead ?
I am using Qt version: 5.15.10 here according to the Lib Info

kintel: Looks good here. Qt-6.8.2 on macOS:
Screenshot 2025-04-18 at 17 01 03

gsohler: great ...
but I cannot push a PR, which does not even work in my place ...
Apparently you are not aware of an "incorrect setting" ...
how to proceed ?

kintel: Does it work with Qt6 on your system? If so, perhaps enable the workaround only for Qt5. That way it will disappear once we finally port everything to Qt6

src/gui/MainWindow.cc
```diff
@@ -1759,23 +1759,30 @@ void MainWindow::actionTreasureTrove()
...
+  QTextStream stream(&result); // compile restored design
   stream << "/*\n"; // put into comments in case it crashes (again)			     
   std::ifstream file( filename.toStdString());
   if ( file ) {
-    stream << file.rdbuf();
+    std::string file_content;
+    std::getline(file, file_content, '\0');
+    stream << file_content.c_str();
     file.close();
   }
```
t-paul, line 1782: Qt file reading (without error checking):
```c++
QFile file("data.txt");
QTextStream in(&file);
QString text;
in >> text;
file.close();
```

src/gui/MainWindow.cc
```diff
@@ -1747,6 +1748,43 @@ void MainWindow::actionShowLibraryFolder()
...
+void MainWindow::actionShowBackupFiles()
+{
+  QString dirname (PlatformUtils::backupPath().c_str());
+  QStringList extensions = {};
+  extensions << "scad" << "csg";
+#ifdef ENABLE_PYTHON
+    extensions << "py";
+#endif
+  extensions.replaceInStrings(QRegularExpression("^"), "*.");
+  const auto filter = QString("OpenSCAD Backups (%1)").arg(extensions.join(" "));
```
t-paul, line 1760: We should not copy whole functions. If it's needed from outside, move `fileOpenFilter()` from anonymous namespace to UIUtils so it's accessible here. Or as alternative extract the getBackupFilename() part and move that to UIUtils.

src/gui/MainWindow.cc
```diff
@@ -1747,6 +1748,43 @@ void MainWindow::actionShowLibraryFolder()
...
+  QString filename = QFileDialog::getOpenFileName(this, "Open File", dirname, filter);
+#endif
+
+  if(filename.size() == 0) return; // dont proceed when cancelled
+  
+  tabManager->actionNew();
+
+  QString result;
+  QTextStream stream(&result); // compile restored design
+  stream << "/*\n"; // put into comments in case it crashes (again)
```
t-paul, line 1775: Is the load operation directly causing a compile again somehow, so that the crash could directly happen again? It would be better to suppress that if needed.

Adding `/*` `*/` only works cleanly if there are no other block comments in the file and removing them seems cumbersome as default.

src/gui/MainWindow.cc
t-paul: Scratch that, we obviously have a "load file from path" function already at `TabManager::refreshDocument()`. So it would make even more sense to extract that core number of lines and use it for both cases.

src/gui/MainWindow.cc
gsohler: Most of the time people are interested in backup files, when openscad crashed before and it was apparently because of a bug triggered by that code. 
So make sure to see the code, but not hit the bug again

t-paul: Obviously, but if you trigger the bug again only by explicitly and manually hitting F5 or F6 again, then *don't do that*. It's more annoying to remove the comments as people may not expect those and if the file has other multi-line comments, it will not comment out the whole file but give some random errors.

gsohler: This is why in latest version there is an 
assert(0); 
in 1st line, which will make sure that file is not executed again
The User has the option to 

* immediately see/have the code again
* not having issues due to nested comments
* has the possibility to bisect the error location by moving  the assert statement around

jordanbrown0:
Don't muck with the user's program.  Just don't.

gsohler:
But then you need an extra switch to enable/disable automatical execution of the scad program,
which makes use yet more complicated. Without a broadcast to all people, you will have to give a lot of support and explain,
, why their code is not executed automatically anymore, until they realize.
I feel , the assert is by most the simplest solution.
Personally I dont see a strong demand to get that merged,  I just felt, it would be a great addition to openscad so i offered.
(my Fork has to already for few days)

jordanbrown0:
That "extra switch" already exists, in the form of the "Automatic Reload and Preview" switch, and I believe it defaults to "off".

(However, in experimenting to confirm, it seems to be squirrelly; a few times it automatically previewed even with the switch off.)

gsohler:
I did not know about the "Automatic Reload and Preview" switch, so i never used it until now.
When I found it, i found it turned on, so I believe default is "on"  instead. 
When I turned it on, it worked as expected. It did not even autoreload nor auto-render when changed by external editor.
So how do you think, I have to change this PR to make it " usable"   ?

jordanbrown0:
I just blew away my OpenSCAD config (from the Windows Registry) and restarted OpenSCAD, and that switch defaulted to off.

I haven't been following this PR in detail; I was reacting only to the idea of editing the user's file to try to keep them from hitting the problem again.  Don't do that.

kintel:
In terms of Automatic compilation on open of a backup file: Why not temporarily disable the automatic compile for this particular case?
I haven't read that code in a while, but we do something similar for the feature of changing numbers in the editor with `alt-arrows`; we compile and preview automatically on each keypress, even if the Automatic Reload and Compile is off.

jordanbrown0:
I don't object to that special case.  (But I should note that I don't use auto-reload-and-preview.)

PR #5830, opened 2025-04-17
Feature/write function

Dante-U:
# Add `write` Function to Write Text to Files

## Description
This pull request introduces a new built-in function `write(filename, text, append = true)` to OpenSCAD, enabling users to write text to files during script execution. The function is implemented in `control.cc` to align with `echo`, ensuring compatibility with the CSG tree and support for child geometries. It addresses the need for file output in OpenSCAD scripts, useful for logging, data export, or generating external files.

### Features
- **Function Signature**: `write(filename, text, append = true)`
  - `filename`: String specifying the output file path.
  - `text`: String to write to the file.
  - `append`: Boolean (default `true`) to append or overwrite the file.
- Logs errors for invalid arguments or file access issues.
- Supports child geometries, similar to `echo`.

### Changes
- **src/core/control.cc**: Added `builtin_write` and registered it in `register_builtin_control`.
- **src/core/parser.y**: Added `TOK_WRITE` to `module_id` to parse `write` as a built-in module.
- **src/core/lexer.l**: Added `write` tokenization (`TOK_WRITE`).
- **src/core/Builtins.cc**: Added `write` to `keywordList` as a reserved keyword.
- **doc/documentation.md**: Documented `write` function.

### Example Usage
```openscad
write("output.txt", "Hello from OpenSCAD!");
write("output.txt", "Another line.", true); // Appends
write("log.txt", "Overwritten.", false);    // Overwrites

### DIFF
--- a/tests/data/scad/functions/write-test.scad
+++ b/tests/data/scad/functions/write-test.scad
@@ -1,0 +1,3 @@
+write("test_eureka.txt", "Test line 1");
+write("test_eureka.txt", "Test line 2", true);
+write("test_log.txt", "Test overwrite", false);


t-paul:
OpenSCAD so far has intentionally no direct access to the file system (and even more generally to the host system in general) from the user defined scripts. People are running the OpenSCAD binaries on servers and in the current state this is pretty safe.

There is currently an effort to additionally bring python scripting to OpenSCAD. This would (after being enabled intentionally by the user running the application) allow basically anything that python can do, including network, file system, database or API interactions.

jordanbrown0:
Note that in addition to server deployments, it's currently fairly safe to download an OpenSCAD program and run it without inspecting it for malicious elements.

jordanbrown0:
Not planned, for the security reasons discussed above.

PR #5836, opened 2025-04-19
build: Adding configuration for working within a Development Container.

pauldoo:
This adds initial configuration for working within a [Development Container](https://containers.dev/).  I have tested this on Linux using both VSCode and the devcontainer CLI.  I suspect not all workflows are supported, but my commit message lists the workflows that I've verified do work.

The out-of-the-box experience with VSCode (using the DevContainer extension) is automatic setup of the dev container and the C++ language support "just working".  For example building, executing the tests, launching (with debugger attached), all work from the VSCode UI.

I have not added information to the documentation (e.g. the README) on using the development container.  Would be happy to do so, but would first like to hear initial thoughts from OpenSCAD maintainers on adding this.

This initial configuration includes support for:
* Building of OpenSCAD
* Running the tests
* Running the app exposing the GUI to the host system (assumes Wayland).
* Debugging
* VSCode configuration to include C++ extension pack.


pauldoo:
I have tested this on Bluefin and Ubuntu 24.04 and on both the devcontainer works OOTB.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,40 @@
...
+        "ms-vscode.cpptools-extension-pack"
+      ]
+    }
+  },
+
+  // https://universal-blue.discourse.group/t/devcontainer-for-developing-an-application-with-a-gui/4634/4
+  // Configuration to allow OpenSCAD running inside the container to connect to host display.
+  "runArgs": [
+    "--device=/dev/dri",
+    "--env=WAYLAND_DISPLAY=${localEnv:WAYLAND_DISPLAY}",
```
kintel, line 29: Should we also forward the X11 DISPLAY here?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,40 @@
...
+    "--network=host"
+    // "--env=XDG_RUNTIME_DIR=${localEnv:XDG_RUNTIME_DIR}",
+    // "--env=DBUS_SESSION_BUS_ADDRESS=unix:path=${localEnv:XDG_RUNTIME_DIR}/bus",
```
kintel, line 32: I suggest either removing these or add comments explaining what they're for and when to enable them

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,40 @@
...
+  ],
+  "mounts": [
+    "source=${localEnv:XAUTHORITY},target=/home/vscode/.Xauthority,type=bind,consistency=cached"
+    // "source=${localEnv:XDG_RUNTIME_DIR},target=${localEnv:XDG_RUNTIME_DIR},type=bind,consistency=cached",
+    // "source=${localEnv:XDG_RUNTIME_DIR}/${localEnv:WAYLAND_DISPLAY},target=${localEnv:XDG_RUNTIME_DIR}/${localEnv:WAYLAND_DISPLAY},type=bind",
+    // "source=${localEnv:XDG_RUNTIME_DIR}/bus,target=${localEnv:XDG_RUNTIME_DIR}/bus,type=bind"
```
kintel, line 38: Same here: Remove commented out code or add comments explaining their purpose.

kintel (commented):
Added a couple of comments. If you're able to verify on X11 as well that would be awesome. Otherwise just adding DISPLAY might be a good start.

pauldoo:
Thanks for the comments. I'm afraid I don't have a system using Xorg/X11 to test forwarding of that.  I have XWayland of course, but I'm not certain how faithful a test that would be (I could accidentally still be using Wayland and not realise).

The commented out code includes snippets to share other XDG directories. I'll remove them as you suggest as the forum thread I link has them if anyone wants them.

I'll update the code, rebase, and update the PR.

pauldoo:
I explored the configuration a bit more deeply and have a better understanding of all the parts now (see comments in code).

I confirm that OpenSCAD and xeyes (an X-only application) both work when running the devcontainer from VSCode, or when running from the devcontainer CLI and the noted lines in config are uncommented.

I note that when running from the devcontainer CLI there are warnings from OpenSCAD about pulseaudio failing (so maybe sound is broken).  These do not appear when running from VSCode, so I assume VSCode is configuring more bits and pieces than I understand.  I don't think sound is important, and additionally I expect most developers will use an IDE like VSCode anyway, so have not looked deeper

Lastly, I have not tested from a host which is X-only.  All the machines I have easy access to are Wayland.

kintel (approved):

PR #5843, opened 2025-04-22
add nix flake for compiling

AaronVerDow:
Simple change that makes it easy to install required dependencies for compiling in Nix. This will not make anything ready for packaging, but it does enable incremental builds which is quicker for development.

I put this in `scripts/` since that's where the other dependency scripts are stored.


scripts/nix/flake.nix
```diff
@@ -0,0 +1,47 @@
...
+            flex
+            pkg-config
+            gettext
+            cmake
+            python3
+            ghostscript
+            eigen
+            boost
+            glew
+            opencsg
```
kintel, line 24: If OpenCSG is >= 1.6, you don't need glew.

scripts/nix/flake.nix
```diff
@@ -0,0 +1,47 @@
...
+            fontconfig
+            qscintilla
+            cairo
+            tbb
+            libGLU
+            libGL
+            wayland
+            wayland-protocols
+            libxml2
+            libsForQt5.full
```
kintel, line 43: Ideally, you'd probably want to develop against Qt6

scripts/nix/flake.nix
```diff
@@ -0,0 +1,47 @@
...
+            pkg-config
+            gettext
+            cmake
+            python3
+            ghostscript
+            eigen
+            boost
+            glew
+            opencsg
+            cgal
```
kintel, line 25: Note that manifold will soon be the default geometry backend, so it should probably be installed via nix if possible, or we don't have to rely on building it from a submodule.

AaronVerDow:
Added all requested changes, works as expected.

kintel:
@hzeller @pca006132 you guys are more familiar with Nix than I am. Does this look reasonable to you?

pca006132:
I think if the lock file is ignored, maybe we should just use shell.nix.

Also, not sure if using qt6 is a good idea, iirc qscintilla is still qt5 on nixos.

kintel:
> Also, not sure if using qt6 is a good idea, iirc qscintilla is still qt5 on nixos.

Oh, that would be a reason to revert back to Qt5 for this platform.

AaronVerDow:
I was still building on QT5, testing again with USE_QT6 set now.

It appears QT6 may work: https://github.com/NixOS/nixpkgs/pull/386982 The qscintilla package still has qt5 in its name in some contexts, but qtbase is accepted as build input so it may still work with qt6.

kintel:
I think Qt6 is going to be a bit more future proof than Qt5, so if you get it to work, that would be great, but happy to merge this with Qt5 for now.
Let me know which one you prefer.

AaronVerDow:
I think I can get QT6 working. I found the missing piece: `qt6Packages.qscintilla` instead of `qscintilla`

I have QT6 building on NixOS now, but it needs more testing. I should have a new version in a few days; feel free to close this pull request in the meantime if needed.

AaronVerDow:
Qt6 on NixOS works. I've confirmed local, incremental builds and linked a working Qt6 package.  Qt5 packages are left in as comments in case anyone needs them.

kintel (approved):

Issue #5131, opened 2024-05-11
Selection of line thickness for SVG/PDF export


Virindi-AC2:
I often export vector images of projections for the purpose of printing paper templates. While the thickness of exported lines seems to be a good default, it would be nice to be able to make them thinner in certain situations for more accuracy.

"Line width" seems like it would be a good option to add to the PDF export dialog, and perhaps future export dialogs.

jordanbrown0:
I'd think the default should be "hairline".

AaronVerDow:
Where should this option be defined?

I'm thinking of adding new special variables like $fn so the option could be defined through the CLI and within scripts.

AaronVerDow:
Per IRC discussion this will be an export option.

t-paul:
Unfortunately SVG has no export options yet, so that's going to be a bit more work. I started a while ago adding this, but that only went so far as an initial GUI file, so that's probably not worth much. I'll attach it here anyway, just in case...

[ExportSvgDialog.ui.txt](https://github.com/user-attachments/files/19822933/ExportSvgDialog.ui.txt)

jordanbrown0:
If we ever do
[OEP7: Mixed Dimension Geometry Support](https://github.com/openscad/openscad/wiki/OEP7%3A-Mixed-Dimension-Geometry-Support)
then line widths would be a good component.

t-paul:
> I'm digging into how 3MF handles colors to see if multi-colored output based on SCAD code would be possible. I'm still pretty new to the code base, so I may leave multi-color mode as a future improvement.

At this point, the 2D subsystem does not support retaining color across CSG operations.

AaronVerDow:
Thanks for the tip on 2D not supporting color.

I have this working via CLI using `parse_hex_color`, but I would like to leverage webcolors as well. `builtin_color` from `core/ColorNode.cc` seems to have the logic I want, but it's not for anything but the `color()` scad command.

Should I stick to hex, expose `builtin_color`, or is there some solution I'm not aware of?

t-paul:
Exposing `builtin_color` directly is probably not useful due to the specialized module invocation parameters. But extracting all the string->color code and the `webcolors` list itself and moving it into `ColorUtil.cc` seems like a viable option. It looks like `pyfunctions.cc` is poking directly into the `webcolors` list too. I hope that would easily translate to the new general string->color mapping too.

AaronVerDow:
SVG options still need to be done, I plan on starting on that next.

PR #5845, opened 2025-04-22
PDF fill and stroke export options

AaronVerDow:
This adds the following export options for PDFs:

```
  - fill (bool): true/<false>
  - fill-color (string): "#000000"
  - stroke (bool): <true>/false
  - stroke-color (string): "#000000"
  - stroke-width (double): 0.000000 : <0.350000> : 999.000000
```

This partially resolves: https://github.com/openscad/openscad/issues/5131

# Examples:
_test.scad_
```
$fn=200;
circle(d=148);
```
## Default
```
./openscad test.scad -o test.pdf
```
default

## Fill only
```
./openscad test.scad -o test.pdf -O export-pdf/fill=true -O export-pdf/stroke=false
```
fill_only

# Fill and stroke with colors
```
openscad test.scad -o test.pdf -O export-pdf/fill=true -O export-pdf/fill-color=#00FFFF -O export-pdf/stroke-color=#FF00FF -O export-pdf/stroke-width=3
```

circle

# Issues
* The GUI has not been updated yet.
* `stroke-width` is in mm, but I don't know where to specify that.
* I would like to utilize `builtin_color` so webcolors can be used, but I'm unsure how to proceed.
  * Should I expose or move that function?


AaronVerDow:
I've moved webcolors into `ColorUtil.cc`, which seemed to be the best place to avoid circular dependencies.  The functionality is complete except for the UI changes. 

I still need to test the Python functions to ensure they work the same way.

I'm closing out this pull request so I can cut a cleaner version that separates the webcolor move from the new PDF functionality.

src/io/export_pdf.cc
```diff
@@ -290,10 +291,22 @@ void export_pdf(const std::shared_ptr<const Geometry>& geom, std::ostream& outpu
...
+
+  if (options->fill) {
+    Color4f fillColor = OpenSCAD::parse_color(options->fillColor);
+    cairo_set_source_rgba(cr, fillColor[0], fillColor[1], fillColor[2], 1.0);
+    draw_geom(geom, cr);
+    cairo_fill(cr);
+  }
+
+  if (options->stroke) {
+    Color4f strokeColor = OpenSCAD::parse_color(options->strokeColor);
```
AaronVerDow, line 304: This is where I plan on using `getColor`

PR #5850, opened 2025-04-23
improve color utils

AaronVerDow:
# Changes

* Create new `core/ColorUtil.cc` functions `parse_color` and `getColor`. These both accept a string, check it against webcolors, and then attempt to parse it as hex.
  * `parse_color` will return false if nothing is found, allowing for functions to handle the error.
  * `getColor` will always return a color and will revert to the default while throwing a warning message
* All code has been updated to use these new functions, reducing duplicate logic and standardizing color processing.
  * Export 3MF can now have colors specified by name via the command line.
  * Pyfunction will now apply alpha channel to colors specified by name and hex.

This change will reduce the amount of duplicate code while implementing additional color support for PDF and SVG outputs: https://github.com/openscad/openscad/pull/5845

# Examples

_test.scad_
```
color("blue")
sphere(d=10);
```

Confirm color is still rendered in GUI: 
blue

Export with new settings:
```
./build/openscad test.scad -o test.3mf -O export-3mf/color-mode=selected-only -O export-3mf/color=red
```

red

__Pyfunctions are untested.__

### DIFF
--- a/src/core/ColorUtil.cc
+++ b/src/core/ColorUtil.cc
@@ -1,6 +1,9 @@
 #include <string>
+#include <unordered_map>
 #include "geometry/linalg.h"
 #include "core/ColorUtil.h"
+#include "utils/printutils.h"
+#include <boost/algorithm/string/case_conv.hpp>
 
 namespace OpenSCAD {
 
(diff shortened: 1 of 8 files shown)


AaronVerDow:
Confirmed I have tests to fix:

```
The following tests FAILED:
	2277 - render-3mf-manifold_issue904 (Failed)
	2278 - render-3mf-manifold_issue1105 (Failed)
	2279 - render-3mf-manifold_issue1105b (Failed)
	2280 - render-3mf-manifold_issue1105d (Failed)
	2281 - render-3mf-manifold_issue1215c (Failed)
	2282 - render-3mf-manifold_internal-cavity (Failed)
	2283 - render-3mf-manifold_internal-cavity-polyhedron (Failed)
	2284 - render-3mf-manifold_rotate_extrude-hole (Failed)
	2285 - render-3mf-manifold_polyhedron-tests (Failed)
	2436 - 3mfcolorpngtest_color-cubes (Failed)
	2437 - 3mfcolorpngtest_color-tests3 (Failed)
	2438 - 3mfcolorpngtest_linear_extrude-parameter-tests (Failed)
	2439 - 3mfcolorpngtest_resize-tests (Failed)
```

AaronVerDow:
Tests pass locally:

```
100% tests passed, 0 tests failed out of 2544
```

gsohler:
not many tests for python yet.
I am excited to use the new ColorUtil class.
do you have a function to convert a Color4d into an hexstring ?

AaronVerDow:
> not many tests for python yet. I am excited to use the new ColorUtil class. do you have a function to convert a Color4d into an hexstring ?

Not yet, though I will need that for SVG export. Do you know of any specific places that could use that functionality now?

src/core/ColorUtil.cc
```diff
@@ -100,4 +103,194 @@ std::optional<Color4f> parse_hex_color(const std::string& hex) {
   return rgba;
 }
 
-} // namespace
\ No newline at end of file
+std::optional<Color4f> parse_web_color(const std::string& col) {
+  std::string colorname = boost::algorithm::to_lower_copy(col);
```
kintel, line 107: The reason for test failures is that previously, the results of to_lower() was used to also print errors/warnings. I think your approach is more correct, you just have to update tests accordingly

src/core/ColorUtil.cc
```diff
@@ -100,4 +103,194 @@ std::optional<Color4f> parse_hex_color(const std::string& hex) {
...
+std::optional<Color4f> parse_color(const std::string& col) {
+  const auto webcolor = parse_web_color(col);
+  if (webcolor) {
+    return webcolor;
+  }
+
+  const auto hexcolor = parse_hex_color(col);
+  if (hexcolor) {
+    return hexcolor;
+  }
```
kintel, line 123: Nit: Since hexcolor is cheap to test (as it has specific length requirements and needs to start with a `#`, perhaps check it first.

src/core/ColorUtil.h
```diff
@@ -24,8 +25,16 @@ void rgbtohsv(float r, float g, float b, float& h, float& s, float& v);
...
+
 Color4f getContrastColor(const Color4f& col);
 
 Color4f getColorHSV(const Color4f& col);
 
-} // namespace
\ No newline at end of file
+extern std::unordered_map<std::string, Color4f> webcolors;
+
+} // namespace
```
kintel, line 40: `} // namespace OpenSCAD`

src/core/ColorUtil.h
```diff
@@ -24,8 +25,16 @@ void rgbtohsv(float r, float g, float b, float& h, float& s, float& v);
 
 std::optional<Color4f> parse_hex_color(const std::string& hex);
 
+std::optional<Color4f> parse_web_color(const std::string& col);
+
+std::optional<Color4f> parse_color(const std::string& col);
+
+Color4f getColor(const std::string& col, const Color4f& defaultcolor = Color4f(0.0f, 0.0f, 0.0f));
```
kintel, line 32: getColor() is never called without providing a default, so I'd skip the default value.
However, since the only function of this extra parameter is to convert from an optional<Color4f> to a Color4f, why not simply expose the optional? That way clients are responsible for dealing with it, as well as providing a contextually correct error message.

..which means you can simply kill getColor() and call parse_color() everywhere; one less function to deal with.

src/core/ColorUtil.h
```diff
@@ -24,8 +25,16 @@ void rgbtohsv(float r, float g, float b, float& h, float& s, float& v);
 
 std::optional<Color4f> parse_hex_color(const std::string& hex);
 
+std::optional<Color4f> parse_web_color(const std::string& col);
```
kintel, line 28: parse_web_color(),  parse_hex_color(), and rgbtohsv() are implementation details and can be removed from the header.

src/core/ColorUtil.cc
```diff
@@ -100,4 +103,194 @@ std::optional<Color4f> parse_hex_color(const std::string& hex) {
...
+    LOG(message_group::Warning, "Unable to parse color \"%1$s\", reverting to default color.", col);
+    LOG("Please see https://en.wikipedia.org/wiki/Web_colors");
+  } 
+
+  return parsed.value_or(defaultcolor);
+}
+
+// Colors extracted from https://drafts.csswg.org/css-color/ on 2015-08-02
+// CSS Color Module Level 4 - Editor’s Draft, 29 May 2015
+std::unordered_map<std::string, Color4f> webcolors{
```
kintel, line 142: Once you separate the public from the private API, you can move the private functions, as well as this map, into an anonymous namespace

kintel (commented):
Added some comments. A bit nitpicky, but I think that with such low-level APIs we should take any opportunity to proactively clean them up as that will serve us in the future.

src/core/ColorUtil.cc
```diff
@@ -100,4 +103,194 @@ std::optional<Color4f> parse_hex_color(const std::string& hex) {
   return rgba;
```
jordanbrown0, line 103: As long as you're nearby, a few lines up on old line 96 there is a comment that says
```
// convert the hex character(s) from base 16 to base 10
```
The part about base 10 is not at all correct; there is no base 10 anywhere in the picture.  It's converting to integer (a binary format) and thence to floating point (also a binary format).

jordanbrown0: Best might be "to a fraction between 0 and 1 inclusive".

src/core/ColorUtil.h
AaronVerDow: I am planning to use `getColor` for PDF and SVG exports and there I will be using the default. Here's an example from a closed pull request: https://github.com/openscad/openscad/pull/5845/files#r2059363559 (The function is called `parse_color` in that version.)

AaronVerDow:
> Added some comments. A bit nitpicky, but I think that with such low-level APIs we should take any opportunity to proactively clean them up as that will serve us in the future.

I'm down with some cleanup.

src/core/ColorUtil.h
kintel: Ah, that makes more sense, as a common place to emit the warning. I'd still kill the default value though, as black isn't necessarily a universal color fallback.

src/core/ColorUtil.cc
AaronVerDow: The logic to test this is already in `parse_hex_color`. It accepts four different string lengths, and testing is necessary to choose how it's parsed. I don't see any value in duplicating that logic here.

src/core/ColorUtil.cc
AaronVerDow: Not sure if I moved this correctly.

kintel: By "anonymous namespace" I meant create a block like this:
```
namespace {

std::unordered_map<std::string, Color4f> webcolors{ [ your stuff here ] };

std::optional<Color4f> parse_web_color(const std::string& col) { ...}

etc.

}  // namespace
```

That way, all of those implementations will be taken off the global namespace and only be available within this file. This is a very common way of hiding implementation details in C++

src/core/ColorUtil.h
```diff
@@ -29,3 +29,15 @@ Color4f getContrastColor(const Color4f& col);
...
+
+void rgbtohsv(float r, float g, float b, float& h, float& s, float& v);
+
+std::optional<Color4f> parse_hex_color(const std::string& hex);
+
+std::optional<Color4f> parse_web_color(const std::string& col);
+
+extern std::unordered_map<std::string, Color4f> webcolors;
+
+} // namespace
```
kintel, line 43: Almost: Remove this from the header, then wrap the corresponding implementations in an anonymous namespace in the .cc file.

AaronVerDow: Sorry, this is the largest C++ project I've worked on.

kintel (approved):

kintel:
Thanks for hanging in there. There are lots of refactoring opportunities in the OpenSCAD codebase, and we tend to address that on demand. One tip for the future if you're motivated to continue contributing: Sometimes it's worth it to simply clean up existing code first without adding any new features, and open a separate PR for that. That makes it much easier to review PRs quickly and makes PRs containing real changes nice and small.

PR #5858, opened 2025-04-26
add color to PDF export tests

AaronVerDow:
This allows the comparison of color PDFs in integration testing.

The original script converted each PDF to PNG through a grayscale Ghostscript device, which stripped out all color data.

This change updates the Ghostscript device so that color PNG files can be exported, and updates the expected files. The expected files look identical because the current ones are in black and white, but the additional color channels must be included for comparison.

This change will be used while adding additional PDF fill and stroke export options.


kintel:
What's with the font change of the expected images?

AaronVerDow:
> What's with the font change of the expected images?

My best guess is the original image was old. The [source code](https://github.com/openscad/openscad/blob/4a33ca0d23fce40389745d788fe9e5efd12dea03/src/io/export_pdf.cc#L24) calls for the font Liberation Sans. The 1 looked wrong for a sans-serif font, but this font does have a line beneath the 1: https://fonts.adobe.com/fonts/liberation-sans

I think the font is correct now, but it does look worse due to Ghostscript. Left is the original PDF, right is the Ghostscript output:

pdf_vs_png

kintel:
Thx. Looks like the wrong font is used when exporting to PDF on macOS. I'll have a quick look at that.

kintel:
Huh, looks like antialiasing made it worse : (

Two things:
1. I figured out the macOS issue: Cairo is using Liberation Sans for export, but pulls it from the system, and Liberation Sans isn't a common system font. Upon installing it, I get the correct result locally.
2. I'm not convinced that those annotations are super important in the context of testing, so if we can simply disable them for the tests, we could do that and move on.

kintel (approved):

PR #5859, opened 2025-04-26
Improve scalability of Launching Screen.

jordanbrown0:
Adds style sheet entries for labels, pushbuttons, and check boxes that are defined in terms of points, rather than left unspecified.  These seem to yield about the same size (at least on my Windows 11 system), but scale with Settings>System>Display>Scale.

Remove the size hints for the dialog box, which were allowing it to be the wrong size for its contents.  Qt seems to do the right thing by default.

Notes:
- You don't get good results from dynamically changing the scale; you need to restart the application.  It looks like the text resizes but the boxes containing it don't.
- This change doesn't change the behavior of the setting at Settings > Accessibility > Text Size.  I think that may be intended to change the point size of the text.
- Further research and experimentation is needed on both of those points, but this seems like an improvement.

Tested only on Windows 11.

Here's the original and new at 100%:
image
image

Here's both of them, resized down as far as it will let us:
image
image

Here they are at 200%:
image
image

And then resized down as far as it will let us:
image
image

Here's the effect from starting at 100% and dynamically switching to 200% without restarting the app.
image
image

### DIFF
--- a/src/gui/LaunchingScreen.ui
+++ b/src/gui/LaunchingScreen.ui
@@ -2,19 +2,12 @@
 <ui version="4.0">
  <class>LaunchingScreen</class>
  <widget class="QDialog" name="LaunchingScreen">
-  <property name="geometry">
-   <rect>
-    <x>0</x>
-    <y>0</y>
-    <width>620</width>
-    <height>418</height>
-   </rect>
-  </property>
-  <property name="minimumSize">
-   <size>
-    <width>600</width>
-    <height>0</height>
-   </size>
+  <property name="styleSheet">
+   <string notr="true">
+    QLabel { font-size: 9pt; }
+    QPushButton { font-size: 9pt; }
+    QCheckBox { font-size: 9pt; }
+   </string>
   </property>
   <property name="windowTitle">
    <string>Welcome to OpenSCAD</string>


jordanbrown0:
Edited to note that the 200% shot is the reason for the change:  all of the text is enlarged, and nothing is cut off.

jordanbrown0:
That new push addresses much more of the UI.  Consider the Preferences dialog:

image
image

I suspect that there are several more places where it's appropriate to remove the size specifications, so that the default doesn't require the horizontal scrolling that you see in the "new" screen shot there.

It would then probably be appropriate to remove the Customizer font size setting that I added a while back, since it should be covered by the global preference.  I'd keep the Editor and Console settings, though, because they are a different *kind* of text and the user might have different preferences.  Also the user may have stronger font-family preferences there.

jordanbrown0:
Removing the size settings for Preferences does not work well.  The resulting window is much smaller.  I think it uses the initial tab (3D View) to derive the window size, and the contents of that tab are much smaller than the contents of several of the other tabs.  It seems worth looking at, but since the simple fix of removing the specification doesn't work... not today.

t-paul:
Nice, being able to scale the UI based on the external settings is a good step. Even better that it can rely mostly on Qt default behavior. I guess the worry is that the designer will put in those section on the next edit again. We may need to add some scripting to maybe at least fail the build if they come back.

I wonder how complicated it would be to have an additional setting to scale just the OpenSCAD UI.

It maybe not super critical but it might be an additional lifeline for people who need an extra big UI. I love how Blender can do that in a very smooth fashion (no surprise there, all their own code) but Thunderbird has a similar feature too, not as general as in Blender but probably good enough for the real world.

jordanbrown0:
And alas, looking at LaunchingScreen.ui, qtcreator *does* try to restore the "geometry" setting that I took out.  I'm looking more to see if there's a way to tell it not to use a fixed size.

rcolyer:
I gave it a go on Linux for a second platform check.  The differences are much more modest for me, but everywhere there is a difference, I categorize this PR as "Jordan did it right here".  Looks great to me.

Master branch splash screen:
master_splash
This PR splash screen (default sizes itself correctly):
pr5859_splash
Master branch preferences:
master_preferences
This PR preferences (very subtle improvements in the preferences layout margins, and also the main window bottom text has properly improved margins):
pr5859_preferences

jordanbrown0:
The horizontal scrolling that you show in Preferences is wrong too, but I've spend much of today trying to figure out how to fix it, without much luck.  The core problem is that QScrollArea doesn't seem to have a way to scroll its child vertically but adopt the child's width, at least as invoked from a `*.ui` file.

jordanbrown0:
WRT the question about how qtcreator interacts with removing the "geometry" stanza... it appears that it *intends* to support that, but looks to me like it has a bug.  If you explicitly tell it "revert to default", it will generate a `.ui` file without the stanza, but if you then load that file again, it will helpfully supply the explicit values again and consider them a customization, not default.

I asked on Reddit, and got surprise (from one user) that it didn't work.

I'd say "review the `.ui` files when people change them, and ensure that these stanzas don't get reintroduced".

PR #5863, opened 2025-04-27
PDF fill and stroke export options

AaronVerDow:
This adds the following export options for PDFs:

```
  - fill (bool): true/<false>
  - fill-color (string): "black"
  - stroke (bool): <true>/false
  - stroke-color (string): "black"
  - stroke-width (double): 0.000000 : <0.350000> : 999.000000
```

These can also be set via the GUI:
export_dialog

This partially resolves: https://github.com/openscad/openscad/issues/5131

New integration tests have been added as well.

# Examples:
_test.scad_
```
$fn=200;
circle(d=148);
```
## Default
```
./openscad test.scad -o test.pdf
```
default

## Fill only
```
./openscad test.scad -o test.pdf -O export-pdf/fill=true -O export-pdf/stroke=false
```
fill_only

# Fill and stroke with colors
```
openscad test.scad -o test.pdf -O export-pdf/fill=true -O export-pdf/fill-color=#00FFFF -O export-pdf/stroke-color=#FF00FF -O export-pdf/stroke-width=3
```

circle


src/io/export_pdf.cc
```diff
@@ -290,10 +291,22 @@ void export_pdf(const std::shared_ptr<const Geometry>& geom, std::ostream& outpu
...
+    draw_geom(geom, cr);
+    cairo_fill(cr);
+  }
+
+  if (options->stroke) {
+    Color4f strokeColor = OpenSCAD::getColor(options->strokeColor, black);
+    cairo_set_source_rgba(cr, strokeColor[0], strokeColor[1], strokeColor[2], strokeColor[3]);
+    cairo_set_line_width(cr, mm_to_points(options->strokeWidth));
+    draw_geom(geom, cr);
+    cairo_stroke(cr);
```
t-paul, line 308: If I'm reading the docs correctly, it should be possible to use `cairo_stroke_preserve()` which preserves the path, so a following fill should work without the need to redraw the geometry.

t-paul (commented):
Nice, with the earlier color changes this ends up pretty concise.

src/io/export_pdf.cc
AaronVerDow: Good catch, updated in next revision.

t-paul: Ah, cool, there's also a fill-preserve, I did not check that. Even better this way :+1:.

t-paul (approved):

AaronVerDow:
Removed an unrelated NixOS build fix.

t-paul:
Uh, sorry for the delay, that somehow slipped through.

Issue #5709, opened 2025-02-26
3D Navigation style


erjoalgo:
I am used to the blender navigation style and find it difficult to switch back and forth between openscad and the blender 3D navigation. For example, in blender orbiting and panning is done via the middle click and middle click + shift. But in the openscad 3D viewer, this would be left click, right click.

It would be useful to support adjusting the navigation style. Other users have expressed the same desire:

[https://www.reddit.com/r/openscad/comments/kzpurw/blender_style_view_navigation_with_the_mouse/](url)

> Is there a way in OpenSCAD to change the view navigation style to the style that Blender uses (similar to how FreeCAD lets you set this)? Did a brief search of this, and I'm not sure if this is supported. I use Blender as well as FreeCAD, and it's been really helpful in FreeCAD to be able to use the same keys and mouse buttons to move around. I'm using version 2019.05 (Ubuntu).

**Is your feature request related to a problem? Please describe.**
N/A

**Describe the solution you'd like**
A "navigation style" dropdown menu option, where users can select among a few different navigation styles. Each navigation style would likely only require re-binding the orbiting and panning mouse bindings.

Freecad implements this feature, here's how it looks like as an example:

![Image](https://github.com/user-attachments/assets/e3744fbe-fe11-4d3d-946d-6955f2bdfc46)

**Describe alternatives you've considered**
Alternatively, exposing the key bindings used for orbiting and panning as a menu option would also suffice.

**Additional context**

harrybraviner:
You can at least switch the orbiting and panning mouse bindings with this button:

![Image](https://github.com/user-attachments/assets/c5cf8e06-4088-4cf8-b6ed-3b519fc718ce)

It sounds like matching Blender would require enabling the middle button as a pan and/or rotate control?

I would find this feature useful, as I regularly switch between OpenSCAD and Fusion, and it would help my brain to be able to use a common navigation style for each. If we added a page to Preferences that enabled selection of the actions performed by `[modifier] + [mouse button]`, with a drop-down menu to set them to defaults (OpenSCAD, Blender, Fusion), would other people also find that useful? I've been poking around at the code and would be happy to implement this.

t-paul:
I suppose the ideal solution would be unifying the mouse input with the input control code for 3d mouse / gamepad / joystick. But that also might be one of those "too big to do in one step" cases.

So starting with a configurable mouse behavior might be a good first step towards more options for users who are accustomed to other mouse setups.

harrybraviner:
[...]

PR #5874, opened 2025-05-01
User-specified mouse behaviour in 3D navigation

harrybraviner:
Allows the user to specify what 3D navigation action they want when left/middle/right click-and-drag-ing. A new preferences menu is added to achieved this. Defaults of OpenSCAD, Blender, and Fusion settings are available, and the user can customize behaviour.

Note that the 'swap mouse buttons' checkbox has been removed because it is now redundant - the user should set this behaviour in the "Mouse" preferences page.

This PR addresses #5709


src/core/Settings.h
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include "gui/input/MouseConfig.h"
```
t-paul, line 3: The include looks unused, I think we can drop that here (also getting rid of the core->gui dependency).

src/gui/QGLView.cc
```diff
@@ -314,43 +314,77 @@ void QGLView::mouseMoveEvent(QMouseEvent *event)
...
   if (mouse_drag_active) {
     mouse_drag_moved = true;
-    auto button_compare = this->mouseSwapButtons?Qt::RightButton : Qt::LeftButton;
-    if (event->buttons() & button_compare
-#ifdef Q_OS_MACOS
-        && !(event->modifiers() & Qt::MetaModifier)
-#endif
-        ) {
-      // Left button rotates in xz, Shift-left rotates in xy
-      // On Mac, Ctrl-Left is handled as right button on other platforms
```
t-paul, line 324: What's the replacement for that macOS special case? While we do have full configuration now, it would not be nice forcing all macOS users to use a custom setting for keeping the old behavior. Does this maybe need a special case when defining the `presetSettings`?

t-paul (changes requested):
Looks great! There are just a couple of minor points to go through. My local build just finished, so I'll give it some testing time now...

src/core/Settings.cc
```diff
@@ -1,5 +1,7 @@
 #include "core/Settings.h"
 
+#include "gui/input/MouseConfig.h"
```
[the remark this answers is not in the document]
harrybraviner: I've moved it. Thanks for pointing this out - I didn't actually know there was a non-GUI mode!

src/gui/QGLView.cc
harrybraviner: This was a little trickier to resolve. I think the right call is to add the option to define Ctrl+Shift+{Left, Middle, Right} click behaviour. I've now done so. (Additionally, Ctrl+Shift+Middle click is used for zoom in Fusion, so this is necessary to properly mimic that behaviour.)

I considered having the OpenSCAD preset modified between Mac OS and other OSs at compile time, but decided against it. Instead I've added this as a different preset, which is selected by default on a Mac. This has the benefit that a non-Mac user can also select this behaviour (e.g. maybe some Windows/Linux users out there have an injury that makes it harder to operate other buttons on their mouse, and would find this useful).

They key thing is: Mac users should now experience no sudden change of behaviour after this update.

harrybraviner:
@t-paul Ok, I think I've addressed your comments. Let me know if there's anything else that needs fixing. Thanks!

t-paul (approved):
Looks very nice! I did some comparison of the new default with an older build and that also seems to be good as far as my tests go on Linux. I'd say we go ahead and see what other platforms say. I assume we'll hear about it quite soon in case there are some hidden corner cases but with the configuration nicely listed in the config tables, it should be relatively easy to handle if something comes up.

PR #5912, opened 2025-05-17
Syncing some Scintilla Magic

### LINKED ISSUE #5905: Wrong comment type when editing python files
**Describe the bug**
When editing a python file and pressing C-d C-Style comments ('//') are used instead of python style comments ('#').

**To Reproduce**
Edit a python file.
Press C-d.

**Expected behavior**
"#" style comments should be used for python files.

Any python code.

 - OS: GNU guix
 - System: PC 64-bit
 - OpenSCAD Version: 2025.05.08-1.72c9919 (git 72c9919d63116f8e711f3566ae34e9eb63a2d6e6)

gsohler:
Fixes #5905


src/gui/ScintillaEditor.cc
```diff
@@ -132,7 +132,7 @@ const boost::property_tree::ptree& EditorColorScheme::propertyTree() const
   return pt;
 }
 
-ScintillaEditor::ScintillaEditor(QWidget *parent) : EditorInterface(parent)
+ScintillaEditor::ScintillaEditor(QWidget *parent, MainWindow &mainWindow) : EditorInterface(parent), mainWindow(mainWindow)
```
kintel, line 135: This feels too coupled. Could you make the connection between the MainWindows and the editor from outside the editor instead?

src/gui/MainWindow.h
```diff
@@ -431,6 +431,9 @@ public slots:
 signals:
   void highlightError(int);
   void unhighlightLastError();
+  #ifdef ENABLE_PYTHON
+  void pythonActiveChanged(bool pythonActive);
```
kintel, line 438: Who is triggering this signal?

gsohler: sorry that was a bad code split. i added now the function, which is triggering pythonActiveChanged.

PR #5914, opened 2025-05-18
Update manifold to 3.1.1

### LINKED ISSUE #5738: Manifold render test failure: example017

kintel:
Fixes #5738

TODO:
* [x] Look into AMF failures
* [x] CMake 4 on Windows is breaking stuff. See if we can work around it or if we need to wait for #5940 - seems to have fixed itself
* [x] Some tests fail only on Linux (e.g. empty import()): Fixed in Manifold 3.1.1 (https://github.com/elalish/manifold/pull/1261)


pca006132:
Seems to be fixed by https://github.com/elalish/manifold/pull/1261

pca006132:
Manifold 3.1.1 is failing three tests: `render-amf-manifold_bad-stl-pcbvicebar`, `render-amf-manifold_bad-stl-tardis`, `render-view-edges-manifold_render-preserve-colors`. For the last one it is just a different triangulation. For the first two it seems that they are unrelated to Manifold, there are errors with CGAL.

kintel:
Light investigation of `render-amf-manifold_bad-stl-pcbvicebar` shows that the same input `MeshGL64` yields different Manifold objects in 3.1.1 vs. 3.0.1 (different vertex and triangle count), and the manifold object from 3.1.1 fails to be converted to a CGAL polyhedron.
I haven't dug deeper, but I hope it's just a case of duplicate vertices which cannot be represented as a CGAL polyhedron.

We don't really need to convert to CGAL in this case, but that's how AMF export is implemented. It's deprecated, so I don't envision us changing it, so just disabling the test is likely the best approach.

pca006132:
Looking at it, it seems that there is nothing preventing Manifold from building with cmake4 (at least with 4.0.2). I have no idea why the CI failed a few months ago.

PR #5929, opened 2025-05-29
prevenrt segfault

gsohler:
Another attempt to suport OpenSCAD

### DIFF
--- a/src/glview/GLView.cc
+++ b/src/glview/GLView.cc
@@ -59,7 +59,7 @@ void GLView::setupShader() {
 }
 
 void GLView::teardownShader() {
-  if( edge_shader == nullptr) return; // in case GLX cannot be initialized 	
+  if( edge_shader == nullptr) return; // if OpenGL context was not initialized
   if (edge_shader->resource.shader_program) {
     glDeleteProgram(edge_shader->resource.shader_program);
   }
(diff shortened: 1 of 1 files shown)


src/glview/GLView.cc
```diff
@@ -59,6 +59,7 @@ void GLView::setupShader() {
 }
 
 void GLView::teardownShader() {
+  if( edge_shader == nullptr) return; // in case GLX cannot be initialized
```
kintel: This is independent of GLX, perhaps change it to "if no OpenGL context could be initialized"

gsohler: ahh ok, but how can I change my PR now ? its an in-github CR with no actual local branch

kintel: I think:
```
git fetch
git checkout gsohler-patch-6
<make change>
git push
```

t-paul: If open a new comment there's a document icon with +/- directly right to the Preview tab. There you can "suggest a change" which can be applied directly after posting the comment.

kintel:
fwiw, this fixes the crash issue, but it won't magically make rendering work, as the crash always happens when we exit due to a fatal error. In any case, good fix!

gsohler:
wow - its merged before checks are pass, 
will mark this special event in my calendar .

kintel:
> wow - its merged before checks are pass, will mark this special event in my calendar .

Yeah, only comment changes, so risk was low :)

PR #5930, opened 2025-05-29
Fix python shebangs

nomike:
The shebangs in a few of the python scripts are wrong.
They should refer to `python3` and not `python`, as that's usually python2 which is end of life and probably not what's intended.

And two of the scripts even refer directly to `/usr/bin/python`. All scripts should use `/usr/bin/env python3` so that the correct python binary is selected.

This is causing almost all tests to fail when building for GNU guix, which doesn't follow the traditional FHS, so `/usr/bin/` doesn't even exist.

Should be an easy one to merge though.

### DIFF
--- a/tests/export_import_pngtest.py
+++ b/tests/export_import_pngtest.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # Export-import test
 #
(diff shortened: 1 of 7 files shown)


nomike:
BTW: I have submitted this patch to pythonscad already where it was merged, I cherry-picked the commits there as they where to ensure there aren't any merge conflicts later.

t-paul (approved):

PR #5933, opened 2025-05-29
Fix search for HidAPI in CMake

nomike:
Currently the CMake code in regards to the HidAPI library is weird.

There is a flag "USE_HIDAPI", but if it's set to "ON" and HidAPI is not found, it will be ignored (similar to how `USE_PYTHON` is set to "OFF" if the requirement "Nettle" is not found (See #5897)).

If the flag "ALLOW_BUNDLED_HIDAPI" is set to "ON", the bundled HidAPI is used, but only if HidAPI is not found in the system using `find_package` first.
And despite the description of that flag stating that this is only supported on Windows, this would still happen on non-Windows platforms.

This pull request aims to clean this up.
The flag "ALLOW_BUNDLED_HIDAPI" is replaced with "USE_BUNDLED_HIDAPI". When it is set to "ON", bundled HidAPI is used, regardless of whether it is available system wide.
As this is only supported on Windows, a fatal error is thrown if that flag is set to true on non-windows platforms.

If the flag is not set, the system wide HidAPI library is looked up and if there is none with the required minimum version, the build will fail, notifying the user, so they can either address this or take the decision to set "USE_HIDAPI" to "OFF".

I don't know why bundled HidAPI is only supported on Winwodws (a comment explaining that, ideally with a link to some ticket or article going into more detail would be nice), so maybe a WARNING would be sufficient in case someone tries to use "ALLOW_BUNDLED_HIDAPI" on a non Windows platform.


nomike:
I have had a look as to why the build is failing and it looks like the patch is already doing what it should:

On the CircleCI build environment, HidAPI is not available, but the flag "USE_HIDAPI" defaults to "ON" and it doesn't seem to be disabled when calling cmake.

So you need to adapt your ci/cd configs, but whether you want to use `-DUSE_HIDAPI=OFF` or make sure HidAPI is available so you can test if that feature builds correctly is something you should decide I guess.

nomike:
I gave this some more thought and I think that this and some other flags should be a three-state variable:
It should be possible to set e.g. `USE_PYTHON` to either `ON`, `OFF` or `AUTO`.

If it's set to `OFF`, Python support will just not be integrated. If it's set to `ON`, a valid version of Python should be searched for. If it's found, it should be built in. If none is found, CMake should fail.

If it is set to `AUTO` though (the new default), it could be 

See #5933 for example. If the user explicitly set's `USE_HIDAPI` to `ON` and hidapi is not found, the user should be made aware of this and the build should fail. Instead the CMake script just set's `USE_HIDAPI` to `OFF` and goes on. It prints out message that hidapi is being disabled. but that's not even a warning. It's just a status message, which makes it feel even more odd to me.

I've just force pushed a new commit here which implements this for `ENABLE_HIDAPI`. I tried to make it generic so it can be applied to other flags as well.

So my suggestion is to change the type of "USE_HIDAPI" from boolean to string so it can be set to either "ON", "OFF" or :"AUTO". The default could just be "AUTO".

This would solve that issue.

nomike:
I just gave the same treatment to SpNav and CAIRO as those were facing the same issue.

CMakeLists.txt
```diff
@@ -8,6 +8,42 @@
...
+
+  # Normalize truthiness values
+  if(DEFINED ${VAR_NAME})
+    set(var_value ${${VAR_NAME}})
+
+    # Convert various truthiness values to standard ON/OFF
+    if(var_value MATCHES "^(1|[Tt][Rr][Uu][Ee]|[Yy][Ee][Ss])$")
+      set(${VAR_NAME} "ON" PARENT_SCOPE)
+    elseif(var_value MATCHES "^(0|[Ff][Aa][Ll][Ss][Ee]|[Nn][Oo])$")
+      set(${VAR_NAME} "OFF" PARENT_SCOPE)
```
t-paul: Why such complicated logic instead of just giving an error if none of the 3 allowed values are given?

nomike: Because I have seen a lot of people do "-DENABLE_HIDAPI=1" on the command line.
If we change now to only allow "ON", "OFF" and "AUTO" it would break their automation/scripts/etc..

It even says so in REAMDE.md:

> ### Compilation
> 
> First, run `cmake -B build -DEXPERIMENTAL=1` to generate a Makefile in the `build` folder.

If you search the whole source-tree for the regular expression ~/-D[^=]*=1/ you will find a number of occurrences where the examples use "1" as value for boolean options. This is perfectly accepted by CMake after all and I guess that some users might have followed that example and used "1" as well instead of "ON".

Of course the code would be cleaner with out that normalization and I'm happy to remove it if backwards compatibility is not a concern of yours.

But do keep in mind that this will deviate from existing behavior.

t-paul: So that needs to be fixed *once* and that's it. If it's giving an error, it's easy to spot. Yes, that could be quite a number of places and that's slightly annoying, but I think preferable to making the `cmake` setup much more complex than needed and keeping that forever.

If we need to allow "truthy" values, we should just rely on `cmake` to do the interpretation instead of implementing our own (e.g. like suggested in https://cmake.org/pipermail/cmake/2016-October/064342.html).

nomike: Nice finding. Thank you!

I've adapted the code. Does this look better?

t-paul (approved):

kintel:
One thing to be aware of with this type of changes is that there is now significant repetition in CMakeLists.txt. Ideally this repetition should be separated out so that the information is only given once, and that ON/OFF/AUTO is managed by some utility function. Not urgent, but the natural next step to do, before adding too many other changes, IMO.

PR #5934, opened 2025-05-30
Fix: Set CMP0167 policy in CMakeLists.txt

### LINKED ISSUE #5932: cmake warning: Policy CMP0167 is not set: The FindBoost module is removed.
CMake is issuing the following warnings:

```plaintext
CMake Warning (dev) at CMakeLists.txt:386 (find_package):
  Policy CMP0167 is not set: The FindBoost module is removed.  Run "cmake
  --help-policy CMP0167" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

This warning is for project developers.  Use -Wno-dev to suppress it.
```

I'm not too familiar with CMake yet, but that looks like something which should be addressed.

nomike:
Fixes #5932.

A similar warning is still printed during building, but this is coming from CGAL.
I've filed a [bug ticket](https://github.com/CGAL/cgal/issues/8914) there as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,9 @@ endif()
 if(POLICY CMP0077) # Let parent project set child project options (used to skip python lookup from manifold & force TBB)
   cmake_policy(SET CMP0079 NEW)
 endif()
+if(POLICY CMP0167) # Use upstream's FindBoost.cmake to find Boost libraries.
+  cmake_policy(SET CMP0167 NEW) # Use upstream's FindBoost.cmake to find Boost libraries.
+endif()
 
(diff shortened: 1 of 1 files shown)


t-paul (changes requested):
Current docs say minimum boost is v1.61 which is not compatible with this.

So we need to check if we can bump to boost >= v1.70 and update the README.

nomike:
I updated the minimum required boost version in CMakeLists.txt as well.

t-paul:
> I updated the minimum required boost version in CMakeLists.txt as well.

Ah, indeed, thank you!

t-paul (approved):

PR #5935, opened 2025-05-30
Fix search for Qt5Gamepad in CMake

nomike:
Similar to the PR's I submitted already, CMake is also silently ignoring if `-DENABLE_GAMEPAD=ON` and Qt5Gamepad is not found.

This patch addresses this and make the build fail properly if the library is missing.

This should prevent users from head-scratching while despite enabling gamepad support during build time, their gamepad still doesn't work in the built application.

I expect the CI/CD pipeline to fail again here, similarly to as it did with #5933.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,7 +73,8 @@ option(BUILD_SHARED_LIBS "Build shared library" OFF)
 include(CMakeDependentOption)
 cmake_dependent_option(APPLE_UNIX "Build OpenSCAD in Unix mode in MacOS X instead of an Apple Bundle" OFF "APPLE" OFF)
 cmake_dependent_option(ENABLE_QTDBUS "Enable DBus input driver for Qt." ON "NOT HEADLESS" OFF)
-cmake_dependent_option(ENABLE_GAMEPAD "Enable Qt5Gamepad input driver." ON "NOT HEADLESS;NOT USE_QT6" OFF)
+set(ENABLE_GAMEPAD "AUTO" CACHE STRING "Enable Qt5Gamepad input driver")
+set_property(CACHE ENABLE_GAMEPAD PROPERTY STRINGS "AUTO" "ON" "OFF")
 set(WASM_BUILD_TYPE "web" CACHE STRING 
   "WebAssembly build type. Support web (builds an OpenSCAD module) or node (standalone .js executable)")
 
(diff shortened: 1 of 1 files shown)


nomike:
I gave this the same trinary treatment as `ENABLE_HIDAPI` or `ENABLE_SPNAV` and `ENABLE_CAIRO` in the past.

As qt5gamepad is not available with qt6, a fatal error is thrown if `ENABLE_GAMEPAD` is set to "ON" and "USE_QT6" is enabled.

BTW: I found a project of someone who [ported qt5gamepad over to qt6](https://www.arnorehn.de/blog/2023/10/31/qtgamepad-ported-to-qt-6/). Maybe it's worth looking at that.

t-paul:
> I gave this the same trinary treatment as `ENABLE_HIDAPI` or `ENABLE_SPNAV` and `ENABLE_CAIRO` in the past.

Thanks, I think that's the best solution.

> As qt5gamepad is not available with qt6, a fatal error is thrown if `ENABLE_GAMEPAD` is set to "ON" and "USE_QT6" is enabled.

Seems fine, I suppose we'll find out in case it ever just happens to get included in Qt6. 

> BTW: I found a project of someone who [ported qt5gamepad over to qt6](https://www.arnorehn.de/blog/2023/10/31/qtgamepad-ported-to-qt-6/). Maybe it's worth looking at that.

Hmm, I'm not sure it's worth it to introduce another dependency for this as it never really worked well. At this point I would 
go with:

1) Let it die and remove once we have no Qt5 builds anymore
2) Move to official Qt6 module if that ever happens before that (which seems not too likely considering the "we want to rewrite it" is almost as old as the last OpenSCAD release)

Meanwhile we could try asking around if anyone is using it.

t-paul (approved):

nomike:
Why are we still maintaining the qt5 builds anyway?
The qt6 builds seem stable enough already. Are there other things besides that gamepad thingy which aren't working yet in the qt6 builds?

We have to have qt6 for MacOS, but are there reasons we have to have qt5?

t-paul:
> Why are we still maintaining the qt5 builds anyway?

I suppose that really boils down to me trying to have a life and also... that.

t-paul:
Oh, and bullseye. From my view the ideal (well, and also at least sightly realistic) timeline would be keeping Qt5 till we have a new release, maybe somewhere beginning of next year. After that, and after Ubuntu 26.04 LTS is released we drop support for Debian 11, Ubuntu 22.04 and also Qt5. «*insert dreaming emoji*»

Issue #5120, opened 2024-05-02
slicer integration and parameters


petaflot:
I find myself editing the slicing parameters for basically every model (rather, part of model) that I print. This is cumbersome and a cause of stupid errors (forgetting to set this or that parameter in the slicer).

AFAIK, the best slicer now is the arachne slicer ; some UIs allow for more than one slicer, so anyhow this should not be a tough barrier to overcome.

I would really like to be able to
* set slicer options in `.scad` files comments
* use openscad to call the slicer while passing the desired options (and possibly send to printer, ie. with octoprint)

pca006132:
I think this is much easier to implement in a slicer rather than openscad. If you implement this in a slicer, this is just letting the slicer read the `.scad` file comment for options, call openscad (via subprocess) to generate the model, and import the model into the slicer. And I still have no idea how to implement this in openscad without modifying the slicer code.

petaflot:
a "slicer" (actually GUIs for slicers) reads an stl file, displays it, and calls the slicer (possibly offering some configuration widgets). finally, it displays an new object that is the sum all all toolpaths.

Openscad generates the object, exports it as an stl file.

I fail to see how it would be difficult get Openscad to call a command-line tool for slicing, such as the [Cura engine](https://github.com/Ultimaker/CuraEngine). However, I fail to see why a particular, "random" slicer frontend would implement reading numerous arbitrary and more or less obscure filetypes from a very large variety of 3D editors.

OpenSCAD already had this feature to send the STL to a print service (commercial or a local octoprint instance) : either case, for a result pf any use the slicing parameters must be provided : specifying the slicing parameters inside the source file would only fill a gap that exists since the `F8` was added. Printing with "default" (or "random") parameters makes for absolutely useless parts unless one is printing silly figures, companion cubes, you-name-it that has no actual mechanical purpose.

I understand that for this to be cross-compatible between many slicers, it would involve a joint effort for the community, and better definition of some concepts : IMHO this would only help improve the global understanding and quality of parts produced.

gsohler:
Seems that 3MF can embed Slicer information

https://lib3mf.readthedocs.io/en/master/source/Cpp/lib3mf_SliceStackIterator.html

Remaining tasks are just

* how to use this function and what it can do
* openscad to export this information
* slicer to understand this information

tcurtin:
[...]

Issue #5089, opened 2024-04-13
Add / change Metadata for 3mf export


UBaer21:
**Is your feature request related to a problem? Please describe.**
3mf file contain only "<object id="1" name="OpenSCAD Model"  which gets confusing in object lists if multiple files are opened

**Describe the solution you'd like**
Allow to set a object name and to set metadata for the default fields like 

  |Title | A title for the 3MF document
  | Designer | A name for a designer of this document
  | Description | A description of the document
  | Copyright | A copyright associated with this document
  | LicenseTerms | License information associated with this document
  | Rating | An industry rating associated with this document
  | CreationDate | The date this documented was created by a source app
  | ModificationDate | The date this document was last modified
  | Application | The name of the source application that originally created this document

optional additional customizable tags - or adding a thumbnail (automatic) would be great

https://github.com/3MFConsortium/spec_core/blob/master/3MF%20Core%20Specification.md#341-metadata

I can see 4 options to obtain this data

1. creation / modification (if file exist) data is automatically entered 
2. a pop up dialog when exporting
3.  a metatag in the openSCAD script that can be parsed to create the 3mf metadata
4.  a saved set within openscad that is used / selected to fill the data.

tcurtin:
+1 to this.  Please please!  Extra credit for the ability to set coordinates of objects when opened in BambuLab.  But just being able name objects (even numerically number instances of modules) would be almost unspeakable helpful to my current workflow!!

UBaer21:
in Regards to BambuStudio (or Slic3r / Prusa) the coordinates are within the 3mf - these programs just handle them differently. If you import the first object they are used (because it is opened as project), if they are imported additionally they are centered - they are also dropped to bed (except if imported as parts then the object is still dropped to bed) - CURA has an option to deactivate "drop" and auto arrange (keep models apart).  So this is nothing OpenSCAD could change - you need to address this with BS/Slic3r

jordanbrown0:
This is strongly related to #4977 "Ability to set solid name", probably enough that they should be considered to be duplicates.

kintel:
We may want to downscope this to support whatever is possible without changing the OpenSCAD language, as that would be a much easier feature to add.

Laserology:
A separate labeling option using ``union()`` might be a good choice for this. Example:
```scad
union("Part1") {
    Code go here...
}
union("Part2") {
    Other code go here...
}
```
[...]

Issue #5277, opened 2024-09-01
Proposal:  Use names or comments in CSG output


EmperorArthur:
# Rationale / Problem to solve
CSG as an intermediary file format looses so much information that it is difficult/impossible to perform some tasks.

# Examples:
* `children()` should be linked objects in FreeCAD, instead of separate objects that are repeated.
  * https://github.com/KeithSloan/OpenSCAD_Alt_Import/issues/29
* FreeCAD importers use generic names for the CSG nodes, instead of module names.
  * https://github.com/KeithSloan/OpenSCAD_Alt_Import/issues/30

# Proposed Solution
The [wiki](https://github.com/openscad/openscad/wiki/CSG-File-Format) seems to suggest that CSG files support named objects.  Therefore parsers **should** support this.  If this is incorrect, then the wiki needs to be clarified.

* When generating a CSG tree, generate a unique object name based on the module, number of times called, and calls to `children()`

### Example
```
module lineup(num, space) {
for (i = [0 : num-1])
translate([ space*i, 0, 0 ]) children();
}
lineup(2, 65){sphere(30);}
lineup(2, 65){sphere(30);}
```
Would produce:
```
lineup_0_children = group() {
				sphere($fn = 0, $fa = 12, $fs = 2, r = 30);
			}
lineup_0 = group() {
		multmatrix([[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]) {
			lineup_0_children()
		}
		multmatrix([[1, 0, 0, 65], [0, 1, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]) {
			lineup_0_children()
		}
	}
lineup_1_children = group() {
				sphere($fn = 0, $fa = 12, $fs = 2, r = 30);
			}
lineup_1 = group() {
		multmatrix([[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]) {
			lineup_1_children()
		}
		multmatrix([[1, 0, 0, 65], [0, 1, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]) {
			lineup_1_children()
		}
	}
```

# Alternatives

## CSG Tags
This is not my **preferred** solution, but maintains compatibility with CSG parsers which may not support named objects.

* When generating a CSG tree, add machine readable comments which tags with the module name where possible.
* When generating a CSG tree, add machine readable comments with unique IDs where possible.

[...]

Issue #5126, opened 2024-05-08
Named groups, or other ways to identify components.


julianstirling:
**Is your feature request related to a problem? Please describe.**
We generate assembly documentation from OpenSCAD. Each component is rendered in OpenSCAD, and then we have OpenSCAD assemblies.

If we want to make a coloured 3D model to share we have a script that can read OpenSCAD CSG files. The script runs openscad on each colour group. The resulting STL is created read into python and converted to a GTLF mesh. This allows us to share colour 3D models generated in OpenSCAD.

The next stage of this would be to allow the user to toggle the visibility components/ sub assemblies in an online 3D viewer. For this we would need to have the information of which code corresponds to which components passed through to the CSG file.

Some sort of named group system would allow the main OpenSCAD editor to also display a tree view and to allow a user to toggle visibility of elements without needing to add `*` and `!` into their code.

**Describe the solution you'd like**

CSG already uses groups. It is possible to add these groups directly into OpenSCAD code. If `group` had an optional argument for a name that was passed through to the CSG file this would allow further processing in separate tools. It could also be used by the OpenSCAD editor to create a tree view.

For example, the following would be possible:
```C
group("Top assembly"){
    group("Lid"){
        /*Lots of code*/
    }
    group("Hinge"){
            /*Lots of code*/
    }
    group("Screws"){
            /*Lots of code*/
    }
}
group("Bottom assembly"){
    group("Case"){
        /*Lots of code*/
    }
    group("Gubbins"){
            /*Lots of code*/
    }
    group("Fixings"){
            /*Lots of code*/
    }
}
```

I feel this would unlock a number of further features to make it easier to design assemblies and complex components in OpenSCAD

**Describe alternatives you've considered**
I did try to echo at the start and end of each component as a hacky solution, but echos are not passed to the CSG.

jordanbrown0:
Duplicate of #4691.

PR #5942, opened 2025-05-31
Add a `name` parameter to `group(…) {}` that is used to name objects in exported `.3mf` files (when using `lazy-union`)

lgarron:
Partially addresses: https://github.com/openscad/openscad/issues/5089

This change adds an optional `name` parameter to the built-in `group(…)` module, which is tracked and set as the `name` attribute on objects in `.3mf` exports (when using `lazy-union`).

```scad
color("blue") group(name="octo_dome") {
    difference() {
      sphere(r=12, $fn=8);
      translate([-500, -500, -1000]) cube(1000);
    }
  }
color("red") group(name="flag")
    translate([0, 0, 12 * cos(360 / 16)]) {
      translate([0, -0.25, 8.5]) cube([5, 0.5, 3]);
      translate([-0.5, -0.50, 0]) cube([1, 1, 12]);
    }
```

This has the particularly useful benefit that these names are recognized and used as part names in common 3D printing slicers (e.g. Bambu Studio, PrusaSlicer, OrcaSlicer):

clipboard

It also allows post-processing scripts to identify and operate on these objects without making fragile assumptions about which pieces are exported in which order. (This is particularly useful if different customizer settings result in different sets of objects in the output.)

Note that:

- This is effectively a more minimal alternative to https://github.com/openscad/openscad/pull/4986 , which seems to have stalled.
- This change only has an effect with the `lazy-union` feature enabled, as otherwise OpenSCAD just creates a top-level wrapper object that obviates all the others.
- The `group(…)` built-in module was previously possible to use, it just didn't do anything useful. This adds an explicit use case to it.
- I looked at a few options, such as setting object/item IDs, but `lib3mf` doesn't seem to support this. Using the `name` field was the most natural option. Note that this allows names to appear for multiple models in an export, but I think that can be useful in some cases.
- It can be a bit unintuitive what modules can be applied to a `group(…)` that preserve the name, because braces can create an additional `group(…)` wrapper implicitly. However, I think it's reasonable to document and recommend to apply the `group(…)` call at the very top level for any names you want in a `.3mf output
- I'd really like a way to make the `.3mf` output more useful for 3D printing, such as the following. But I think that functionality can be designed separately, and just having the name transferred into the `.3mf` makes it much easier to write more robust post-processing scripts in the meantime.
  - Marking objects as different types e.g. "support blocker", "negative part", "modifier".
  - Coloring objects in a slicer-compatible way, perhaps even associating filament information from a config file.
  - Exporting multiple objects each consisting of (potentially) multiple parts, potentially across multiple plates.
  - Also see: https://github.com/openscad/openscad/issues/5120

I need help with the following:

- Is this a reasonable approach in the first place?
- Should this be placed behind an additional feature flag?
- How do I update the tests? I've tried adding one, but there isn't much documentation about adding a test and they take a long time to run.
- Should I update any documentation in this PR itself?

### DIFF
--- a/src/core/GroupModule.cc
+++ b/src/core/GroupModule.cc
@@ -37,8 +37,14 @@
 std::shared_ptr<AbstractNode> builtin_group(const ModuleInstantiation *inst, Arguments arguments,
                                             const Children& children)
 {
-  Parameters parameters = Parameters::parse(std::move(arguments), inst->location(), {});
-  return children.instantiate(std::make_shared<GroupNode>(inst));
+  Parameters parameters = Parameters::parse(std::move(arguments), inst->location(), {}, {"name"});
+
+  std::string group_name = "";
+  if (parameters["name"].type() == Value::Type::STRING) {
+    group_name = parameters["name"].toString();
+  }
+
+  return children.instantiate(std::make_shared<GroupNode>(inst, "", group_name));
 }
 
 void register_builtin_group()
@@ -46,5 +52,7 @@ void register_builtin_group()
   Builtins::init("group", new BuiltinModule(builtin_group),
                  {
                    "group",
+                   "group()",
+                   "group(name)",
                  });
 }
--- a/src/core/node.cc
+++ b/src/core/node.cc
@@ -135,6 +135,11 @@ std::string GroupNode::verbose_name() const
   return this->_name;
 }
 
+std::string GroupNode::model_name() const
+{
+  return this->_model_name;
+}
+
 std::string ListNode::name() const
 {
   return "list";
--- a/src/core/node.h
+++ b/src/core/node.h
@@ -48,6 +48,8 @@ public:
       the verbose name shall be overloaded. */
   virtual std::string verbose_name() const { return this->name(); }
 
+  virtual std::string model_name() const { return ""; }
+
   const std::vector<std::shared_ptr<AbstractNode>>& getChildren() const { return this->children; }
   int index() const { return this->idx; }
 
@@ -116,15 +118,17 @@ class GroupNode : public AbstractNode
 {
 public:
   VISITABLE();
-  GroupNode(const ModuleInstantiation *mi, std::string name = "")
-    : AbstractNode(mi), _name(std::move(name))
+  GroupNode(const ModuleInstantiation *mi, std::string name = "", std::string model_name = "")
+    : AbstractNode(mi), _name(std::move(name)), _model_name(std::move(model_name))
   {
   }
   std::string name() const override;
   std::string verbose_name() const override;
+  std::string model_name() const override;
 
 private:
   const std::string _name;
+  const std::string _model_name;
 };
 
 /*!
--- a/src/geometry/Geometry.h
+++ b/src/geometry/Geometry.h
@@ -41,6 +41,8 @@ public:
   [[nodiscard]] virtual size_t numFacets() const = 0;
   [[nodiscard]] unsigned int getConvexity() const { return convexity; }
   void setConvexity(int c) { this->convexity = c; }
+  [[nodiscard]] std::string getModelName() const { return model_name; }
+  void setModelName(std::string model_name) { this->model_name = model_name; }
   virtual void setColor(const Color4f& c) {}
 
   virtual void transform(const Transform3d& /*mat*/) { assert(!"transform not implemented!"); }
@@ -53,6 +55,7 @@ public:
 
 protected:
   int convexity{1};
+  std::string model_name{""};
 };
 
 /**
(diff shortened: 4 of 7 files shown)


coryrc:
From what I've read, lazy-union is not going to be kept. Could this be made without lazy-union?

As I noted in https://github.com/openscad/openscad/issues/350#issuecomment-3183117345 it would be nice to keep the root lazy-union.

I think most dev discussion happens on the email list.

lgarron:
> From what I've read, lazy-union is not going to be kept. Could this be made without lazy-union?
> 
> As I noted in [#350 (comment)](https://github.com/openscad/openscad/issues/350#issuecomment-3183117345) it would be nice to keep the root lazy-union.

I mean, as long as as the root isn't implicitly unioned. This code might work as-is with another approach, depending on the implementation.

`openscad` would become *extremely* less useful to me if only one output object was supported, though.

> I think most dev discussion happens on the email list.

Perhaps? I'm chugging along happily on a build with this PR's code, if I don't hear back hear I guess this PR will just hang around indefinitely. 🤷

lgarron:
For what it's worth, I have (manually) rebased this PR.

It would be nice to have some signal from maintainers about moving forward with some way to name pieces, whether in this PR, or one of:

- https://github.com/openscad/openscad/issues/6354
- https://github.com/openscad/openscad/issues/5277
- https://github.com/openscad/openscad/issues/5126

lgarron:
> OpenSCAD bot: this PR is stale because it has been open for 120 days with no activity. It will be closed if no further activity occurs within 60 days.

I mean, I'm happy to rebase it again, else it would be nice to have an explicit rejection.

jordanbrown0:
Lazy union is unlikely to survive in its current form, and so any development based on it is counterproductive.

lgarron:
> Lazy union is unlikely to survive in its current form, and so any development based on it is counterproductive.

Are there any medium-term options to ship something?

It is possible to do a lot of reliable post-processing if there is a way to identify individual objects by name in a `.3mf` — they can be grouped, refactored, and assigned additional metadata properties arbitrarily. Then OpenSCAD wouldn't need to support more advanced features any time soon — I could automate a lot of 3D printing workflows just based on the object names.

But I don't feel like I could write anything reliable unless I have *some* way to identify objects by name. It doesn't have to use `lazy-union` and I'm certainly fine using a feature flag. In fact, I have regularly used builds based on this branch just because of the sheer benefit that named objects provide. But these builds tend not to keep working indefinitely. It would be really useful to have something that can work on my and other people's machines without regular rebasing work, even if it's really basic and not guaranteed to stick around forever. And I'd of course be be glad to put some effort into writing code to move forward any approach that has a chance of being merged.

jordanbrown0:
> Are there any medium-term options to ship something?

As a strictly volunteer effort, any prediction is fraught with peril.  It all depends on who has time, energy, and interest.

As far as my personal priority list... I've got a Prusa INDX upgrade kit ordered, and I'd like to have multi-color design available conveniently in OpenSCAD by the time it arrives.  (Whenever that is!)

My proposal (which i can't say has significant traction, but I'm hopeful) is at https://github.com/openscad/openscad/wiki/OEP9:-Generating-Multiple-%E2%80%9CParts%E2%80%9D-in-One-Run .

Until that, my best suggestion is to wrap your OpenSCAD program in a shell script (or your favorite alternative), and call it multiple times with -D options to generate the various parts.

> not guaranteed to stick around forever

I don't think we'd do a new feature for this that we didn't think would be usable indefinitely.  It would probably be extensible, but the intent would always be that it remained compatible.  Too many people get upset when there are compatibility breaks.

lgarron:
> As far as my personal priority list... I've got a Prusa INDX upgrade kit ordered, and I'd like to have multi-color design available conveniently in OpenSCAD by the time it arrives. (Whenever that is!)

Same! Very excited. 😁

> My proposal (which i can't say has significant traction, but I'm hopeful) is at https://github.com/openscad/openscad/wiki/OEP9:-Generating-Multiple-%E2%80%9CParts%E2%80%9D-in-One-Run .

I mean, it looks like a pretty reasonable direction to me!

> Until that, my best suggestion is to wrap your OpenSCAD program in a shell script (or your favorite alternative), and call it multiple times with -D options to generate the various parts.

I've considered that, but I think that the need to regenerate every part [uncached](https://github.com/openscad/openscad/issues/1645) would be prohibitive for the use cases where keeping track of parts is also important. 😔

I guess I'll keep rebasing for now. Thanks for the reply!

jordanbrown0:
> I've considered [wrapping multiple runs], but I think that the need to regenerate every part https://github.com/openscad/openscad/issues/1645 would be prohibitive for the use cases where keeping track of parts is also important. 😔

Eh, maybe.  Manifold is really fast.  It isn't clear that caching really matters much these days.  (Minkowski excepted.)

Of course, you need to arrange that your wrapper and your OpenSCAD program cooperate so that the process is totally automated, so that you say "do it" and a complete set of output files appears.

(To be clear:  I don't consider that to be a **good** answer to the question... just one of the least bad currently available.)

PR #5959, opened 2025-06-06
Python pr10

gsohler:
objects.separate()  finds all separate pieces within a solid and creates a python array with the result

explode operator is demoed on the pythonscad homepage(camera case)


kintel:
I don't have the capacity to deal with PRs against the Python subsystem. As long as the Python stuff is experimental and isolated from the rest of OpenSCAD, I'm happy to rubber stamp PRs though, but let's hear what @t-paul has to say first

Issue #5732, opened 2025-03-08
Measurement tool per-axis


cherrydev:
**Is your feature request related to a problem? Please describe.**
The new measurement tool is useful under some circumstances, but I often want to (for example) measure the distance between two edges or vertices on a specific axis. Unless the edges or vertices are aligned on that axis, this is not currently possible to do.

**Describe the solution you'd like**
I would like the measurement tool to additionally output measurements per-axis. For example, right now it displays: `Distance is 17.8045`. More comprehensive output like the following would be useful:
```
Distance : 17.8045
𝛥-x: 11.0000
𝛥-y: 14.0000
𝛥-z: 0.0000
```

**Describe alternatives you've considered**
Other than exporting the model and using a different piece of CAD software to perform the measurement, I can't really think of an alternative.

**Additional context**
Other CAD software has this feature. For example, FreeCad shows the different axis deltas overlayed graphically on the model. KiCad (2d cad) shows a text popup as follows:

```
x: 10.249 mm
y: -2.769 mm
r: 10.616 mm
θ: 15.1°
```

gsohler:
That should be easy to implement.
Measurement tool does yet already display the 3D coordinates of the
selected point in the status bar.
It could already yet be an improvement if you subtract the values manually.

coryrc:
This was fixed by #6273 please close as complete.

PR #5960, opened 2025-06-07
Improve measurement tools

### LINKED ISSUE #5931: Measuring tool doesn't appear to function correctly on non-primitive shapes
**Describe the bug**

I can use the measuring tool on a cube (`cube(30, 30, 30)`) but not non-primitive shapes.

**To Reproduce**

Try code below and try to measure hollow area, which appears not possible currently.

**Expected behavior**

Able to measure inside of hollow area of cube.

```
		difference()
		{
			bat_len = 85;
			bat_wid = 21;
			bat_hig = 21 + 3;
			wall = 2;
			cube([bat_len + wall, bat_hig + wall, bat_wid + wall]);
			translate([wall, wall, wall])
			cube([bat_len - wall, bat_hig, bat_wid - wall]);
		}
```

 - OS: Arch Linux
 - System: x86-64
 - OpenSCAD Version: 2025.05.16.ai25460

mvirkkunen:
* Added depth sorting to edge selection so that edges closer to the camera are favored
* Improved selection rendering so that the selected edge does not get hidden inside the object

The "improved" rendering uses some fairly crusty GLU functions, but a lot of pretty old OpenGL stuff is used all over the place anyways.

N.b. the new functions in `GeometryUtils` are purposefully written to be a bit more flexible than the old ones. I have a follow-up PR that will improve edge cases (heh) in edge-to-edge distance calculation, and also adds X/Y/Z delta display to the measurement (#5732). I'll send it in if this one gets merged.

I think the improved rendering also fixes #5931.


kintel:
Re. "fairly crusty GLU functions". Ideally we want to move off anything immediate mode onto a clean (GLES3 compatible) subset of OpenGL. This is a somewhat longer term goal, but if you find a way of moving in that direction, it would be appreciated, but probably not in scope for this PR.

Will give this a review and test.

mvirkkunen:
Aww, too crusty for the emulation layer the WASM build uses it seems.

I guess I'll write a quick line/point renderer myself then but that'll have to wait until tomorrow (almost tempted to use the primitives that OpenSCAD already comes with - but that feels like a hack)

kintel (approved):

kintel:
Approved modulo making it build everywhere :)

kintel:
Using OpenSCAD primitives would feel natural, but some infrastructure probably needs to be built for that. This could be useful in a number of places, but it's a bit of an investment..

mvirkkunen:
Okay, attempt two! Ended up using OpenSCAD geometry anyways because we already have all the primitives we need. I have some ideas to use this for drawing something more complex for measurements purposes as well. I didn't add any caching because there aren't that many objects to draw, although it might still hit the global geometry cache for all I know.

I also made selections behind the model faintly visible because currently it's possible to click "through" the model to select a back edge for example. True hidden edge culling when clicking would be better than this, but that's a lot more complicated to implement.

src/glview/UIGeometryRenderer.cc
```diff
@@ -0,0 +1,99 @@
...
+  if (root->children.empty()) {
+    return;
+  }
+
+  Tree tree(root, "GUI");
+  GeometryEvaluator evaluator(tree);
+  std::shared_ptr<const Geometry> geometry = evaluator.evaluateGeometry(*tree.root(), true);
+
+  PolySetRenderer renderer(geometry);
+  renderer.prepare(shaderInfo);
```
kintel, line 65: While this is general, it feels slightly wasteful to go through an entire geometry evaluation per rendered frame.
I guess this depends a bit on the use-case; interactive selection while the object is static vs. moving the camera once the markup is rendered. My worry is mostly about performance behavior of the latter case.

mvirkkunen: I didn't make anything persist across frames because, at least for now, the amount of geometry is pretty small and only evaluated when the measurement tool is in use.

If you think that's still too much of a performance hit, I can add some caching.

kintel: It might pan out - I need to think about it and test it. I'm just a bit surprised about evaluating CSG geometry per frame for rendering annotations :)

PR #5971, opened 2025-06-12
Cmake parametrize appname

nomike:
This MR allows the name of the executable to be changed.
It's a first step in harmonizing the build scripts between OpenSCAD and PythonSCAD.

Per default, the executable is still called "openscad".

I introduced a variable which is set to "OpenSCAD" and two more convenience variables with the name in upper- and lower-case so it can be used further later on if there is a need.


t-paul:
The use in code is not yet centralized, so right now there are multiple places that need to be updated to use the configurable name:

https://github.com/openscad/openscad/blob/811ef8cd64ea5ca4e064a7ecb8a65e10f1baf0fd/src/python/pymod.cc#L95
https://github.com/openscad/openscad/blob/811ef8cd64ea5ca4e064a7ecb8a65e10f1baf0fd/src/python/pymod.cc#L119
https://github.com/openscad/openscad/blob/811ef8cd64ea5ca4e064a7ecb8a65e10f1baf0fd/src/openscad.cc#L815
https://github.com/openscad/openscad/blob/811ef8cd64ea5ca4e064a7ecb8a65e10f1baf0fd/src/python/pyopenscad.cc#L324

tests/test_cmdline_tool.py
```diff
@@ -311,7 +311,8 @@ def run_test(testname, cmd, args, redirect_stdin=False, redirect_stdout=False):
         args[0] = "-"
 
     try:
-        is_openscad = os.path.split(cmd)[1].lower().startswith("openscad")
+        is_openscad = cmd == os.environ.get('OPENSCAD_BINARY')
```
kintel, line 314: Not sure that OPENSCAD_BINARY is usually set when running tests

nomike:
Quite some time has passed since I opened this PR and I have learned a lout about building *SCAD since then.
I will close this PR for now as I'm anyway working on incremental improvements of the build system.
Maybe we can bring OpenSCAD and PythonSCAD together one day so that they no longer have diverging CMake configs.

But this will take time.

PR #5972, opened 2025-06-14
Add ability to build to a (wasm-based) node module

jesseditson:
I'm working on adding inline preview support to https://github.com/Leathong/openscad-support-vscode - since it runs in electron, the most straightforward way to do so would be to build a node.js module that handles compilation inside vscode.

The emscripten stuff was mostly there, just needed different linking options to generate something I could bundle.

This PR just adds a reusable `WASM_BUILD_TYPE` called `node-module` and some light usage instructions in the README.


xcode/OpenSCAD copy-Info.plist
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
t-paul: This looks like it's committed by mistake?

jesseditson:
Yes absolutely! Feel free to ignore this until the tests pass too, I should have a moment next week to peek at it!

PR #5984, opened 2025-06-19
MSVC based headless build support.

Sparsh-N:
Work in progress PR, will add ../.. to the Windows search path. Another focus is to start introducing some minor changes (package versions, etc.). Will update this PR until a working headless build can work without any errors (post build, ctest not working, etc.)

TODO:

- [x] Add vcpkg.json and instructions for how to install with manifest mode
- [x] Fix post build error (translation files) -> To be added in PR for GUI Build 
- [x] Fix issue with headless build not finding test library (ctest)

Please feel free to comment any suggestions/questions!


CMakeLists.txt
```diff
@@ -336,7 +336,7 @@ if (MSVC)
   message(STATUS "Eigen: ${Eigen3_VERSION}")
 
   set(Boost_USE_STATIC_LIBS TRUE)
-  find_package(Boost 1.70 REQUIRED COMPONENTS system regex program_options)
+  find_package(Boost REQUIRED COMPONENTS system regex program_options)
```
kintel: Why did you need to remove the version? This should represent the minimum Boost version required by OpenSCAD.

src/platform/PlatformUtils.cc
```diff
@@ -47,6 +47,7 @@ static std::string lookupResourcesPath()
     ".", // Release location
     RESOURCE_FOLDER("../share/openscad"), // MSYS2 location
     "..", // Dev location
+    "../..", // MSVC
```
kintel: Perhaps add that it's due to binaries are built in build/Release/ or build/Debug/

CMakeLists.txt
Sparsh-N: It's under the MSVC build case, which I'm assuming means it would reference the (WIP) vcpkg based dependency installation. Since the most recent version of boost on vcpkg is 1.88, I'd assume a minimum version wasn't needed. However, if building with MXE or MSYS2 also enters the `if MSVC()` case in CMakeLists, I could keep the 1.70 minimum version specification back in.

It builds without any error either way, so it might make more sense to leave it in.

kintel: It makes more sense to leave it in, as it represents the API contract between OpenSCAD and Boost, and if we update it in the future, it will fail early if the required API isn't available.

Sparsh-N (commented):
Should the MXE case minimum version also be updated? Since the versions differ we could also set a minimum boost version variable at the start and just reference it whenever it is called. (Unless MXE build is referencing a older boost version)

It's in line 387, I forgot to select the line but it sets 3.61 as the minimum on MXE.

kintel:
> Should the MXE case minimum version also be updated? Since the versions differ we could also set a minimum boost version variable at the start and just reference it whenever it is called. (Unless MXE build is referencing a older boost version)

Ideally, we shouldn't have branches in CMakeLists at all :)
But in the interim I'd say use the same version everywhere we can, but keep in mind that version checking behavior may differ between platforms, so every change needs to be verified to work before merging, which is probably why it's not fully in sync.

Sparsh-N:
Hi @vincele sounds like a good plan, I'll open that as a separate PR. That said, some of the changes in the `CMakeLists.txt` are specific to a if (MSVC) case, which also gets triggered during MSYS2 builds to my understanding. I'll see if the changes (just libintl > intl and Fontconfig package) dont cause any issues for the existing windows build process on MSYS2.

vincele:
> are specific to a if (MSVC) case

Yeah, I did not expand the diff view up, so missed that.

Sparsh-N:
Most of the main changes for a headless build work. I have one more minor change (all in CMakeList) which I will use this PR/branch for.

Anything GUI related will be a separate branch, I'll reference in the main GSoC issue #6123

Sparsh-N:
Not sure why it closed when updating the branch, but this commit should allow a user to build a headless version with MSVC without any issues. I'll add a PR for documentation changes because some changes to CMakeLists may take time (i.e. setusing c++ version 20 over 17, etc)

CMakeLists.txt
```diff
@@ -112,13 +112,21 @@ if(CLANG_TIDY)
   #  append ";--fix" to arguments above to apply automatic fix (if the given checks support it)
 endif(CLANG_TIDY)
 
-add_executable(OpenSCAD)
+if (MSVC)
+  add_executable(OpenSCAD WIN32)
+else ()
+  add_executable(OpenSCAD)
+endif()
```
kintel: Is this necessary?
There is a line further down which does the same:
```
  set_property(TARGET OpenSCAD PROPERTY WIN32_EXECUTABLE ON)
```

PR #5985, opened 2025-06-20
Fix(export): Handle no-arg color() correctly in 3mf and off exports

### LINKED ISSUE #5966: Color rendering: color() without arguments exports as black to 3mf
`color() cube()` exports as a black cube, while it renders as a default (yellow) colored cube.

Only tested with lib3mf v2 on macOS.

Needs:
* [ ] Add test case
* [ ] Test on other platforms
* [ ] Test on lib3mf v1
* [ ] Add alpha channel test
* [ ] Test other color-capable export formats

kintel:
Fixes #5966


kintel:
Abandoning this. At first glance it looks reasonable, but it's just the right solution as it resolves the color too early.

PR #4478, opened 2022-12-29
Objects, geometry as data, module references

jordanbrown0:
This PR incorporates module literal work by @kwikius , the object() function by @revarbat , and the render() function by @sublim9 .

This is NOT final.  It includes a number of alternative approaches to several questions.  The intent is to allow a few people to play with it and opine on how to address some open questions.

### Builds for testing this PR (2022-12-29 / [cffa44f](https://github.com/openscad/openscad/commit/cffa44f61a5fd4934897e35753bdd207ea32dcd6))

http://files.openscad.org/pr/PR4478/

### Features included

- object() function - creates objects from primitives and previous objects.
- module literals / module references - define a module in an expression and put it into a variable.
- object literals - a JavaScript-like syntax for creating objects.
- geometry literals / geometry-as-data - define a CSG tree in an expression, render it into polyhedra data.
- hybrid literals - an alternative syntax combining the features of objects and geometry into one syntax and one data type

### Details

- object() function creates objects from primitive data.  Arguments are any number of:
  - object - merged into result
  - `name=value` - adds name+value to result
  - `[ namestr, value ]` - adds name+value to result
  - `[ namestr ]` - removes name and its value from result
- has_key(object, keystr) -> boolean

- module literals / module references
  - new expression component `module (<params>) { <body> }` yields module reference.
  - module reference is just a value, can put in variable, put in array, pass as argument, return from function, et cetera.
  - call as
    - `m(args)`
    - `a[i](args)`
    - `o.m(args)`
    - `(expr)(args)`
    - or combinations, e.g. `a[i].m(args)` for an array of objects with a member that is a module reference.
    - but not as `f(fargs)(margs)`, not as a function returning a module reference
      - but but  there's nothing wrong with `(f(fargs))(margs)`.

- Object literals - new expression component
  - `{ name: value, ... }`
  - `{ "name": value, ... }` for non-identifier names
  - Note no computed-key syntax.

- geometry literals / geometry as data
  - new expression component `{{ ...statement-context OpenSCAD... }}`, e.g. `{{ cube(); }}`
  - value is opaque
  - add to model with:
    - `g;`
    - `a[i];`
    - `o.g;`
    - `(expr);`
    - or combinations, e.g. `a[i].m(args)` for an array of objects with a member that is a geometry value.
    - but not as `f(args);`, not as a function returning a geometry value
      - but but there's nothing wrong with `(f(args));`.
    - Also as `import(geometry)`
  - new function `render(geometry)`
    - returns an object
[...]

PR #6004, opened 2025-06-27
Add an struct type to OpenSCAD using the let() syntax

pkriens:
I like OpenSCAD but there is one thing that is almost impossible for me to overcome and that is the lack of a structure type. The use of arrays is very hard because you need then to define the index constants and that takes a toll on the namespace.

I noticed that OpenSCAD already had an object type, it only lacked the proper syntax and construction from the program text. To minimize the changes in the language, I used the syntax for `let` since it already defines a structure. If the let is used without a final expression, its value is a structure.
```
extend = let( width = 100, height=200, area = function() width*height );
```
As you can see, the changes to get to this are absolutely minimal because it stays so close to the existing language. However, I had to modify the ObjectType. The  current implementation stores the Value's twice. Once in a map and once in a values vector. This wreaked havoc in the garbage collection because the garbage collector assumed the ObjectType had one of its values but it was counted as twice .... hard problem to diagnose.

I now store the values once and use a function to lookup the index of the key and get the value at that position. More efficient as well.

Curious how I should start advocating for this. I am leaving the code here to get feedback of its implementation and if this could become part of OpenSCAD.

### DIFF
--- a/src/core/Expression.cc
+++ b/src/core/Expression.cc
@@ -726,8 +726,24 @@ const Expression *Let::evaluateStep(ContextHandle<Context>& targetContext) const
 
 Value Let::evaluate(const std::shared_ptr<const Context>& context) const
 {
-  ContextHandle<Context> letContext{Context::create<Context>(context)};
-  return evaluateStep(letContext)->evaluate(*letContext);
+    ContextHandle<Context> letContext { 
+        Context::create<Context>(context)
+    };
+
+    if ( !expr) {
+      ObjectType obj(context->session());
+
+      for (auto & assignment : this->arguments) {
+        auto && name = assignment->getName();
+        Value && value = assignment->getExpr()->evaluate(*letContext);
+        letContext->set_variable(name, value.clone());
+        obj.set(name, std::move(value));
+      }
+      return std::move(obj);
+
+    } else {
+        return evaluateStep(letContext)->evaluate(*letContext);
+    }
 }
 
 void Let::print(std::ostream& stream, const std::string&) const
--- a/src/core/Value.cc
+++ b/src/core/Value.cc
@@ -1268,15 +1268,18 @@ ObjectType::ObjectType(EvaluationSession *session) :
 
 const Value& ObjectType::get(const std::string& key) const
 {
-  auto result = ptr->map.find(key);
-  // NEEDSWORK it would be nice to have a "cause" for the undef, but Value::undef(...)
-  // does not appear compatible with Value&.
-  return result == ptr->map.end() ? Value::undefined : result->second;
+    auto it = std::find(ptr->keys.begin(), ptr->keys.end(), key);
+    if ( it != ptr->keys.end()){
+        size_t index = std::distance(ptr->keys.begin(), it);
+        if (index < ptr->values.size()) {
+            return ptr->values[index];
+        }
+    }
+    return Value::undefined;
 }
 
 void ObjectType::set(const std::string& key, Value&& value)
 {
-  ptr->map.emplace(key, value.clone());
   ptr->keys.emplace_back(key);
   ptr->values.emplace_back(std::move(value));
 }
--- a/src/core/Value.h
+++ b/src/core/Value.h
@@ -388,8 +388,6 @@ private:
 
 // The object type which ObjectType's shared_ptr points to.
 struct Value::ObjectType::ObjectObject {
-  using obj_t = std::unordered_map<std::string, Value>;
-  obj_t map;
   class EvaluationSession *evaluation_session = nullptr;
   std::vector<std::string> keys;
   std::vector<Value> values;
--- a/src/core/parser.y
+++ b/src/core/parser.y
@@ -342,7 +342,7 @@ expr
             {
               $$ = new TernaryOp($1, $3, $5, LOCD("ternary", @$));
             }
-        | TOK_LET '(' arguments ')' expr
+        | TOK_LET '(' arguments ')' expr_or_empty
             {
               $$ = FunctionCall::create("let", *$3, $5, LOCD("let", @$));
               delete $3;


t-paul:
Discussion about how to handle objects is in #4478. I'm not sure how it fits in there, but the whole topic needs some love and engagement to move forward.

kintel:
We also have two experimental features which uses the object API: [textmetrics and JSON import](https://github.com/openscad/openscad/blob/25725172edaf7a10006590fe10e15757eb311516/src/Feature.cc#L33-L34), both of which need some validation before we decide to include it in the next stable release.

@jordanbrown0 is the champion behind this, so he should definitely be involved.

I think the big question for objects in general is whether a reasonable scoped feature (like JSON import) can make it into the next release or whether we should wait for more fully fledged language support to make it available outside of experimental.

In any case, for a feature like this to be mergeable, it needs to be protected by a feature flag (or merged into one of the two flags mentioned above), but lets discuss it first. If we can get critical mass on IRC at the same time, that tends to be pretty efficient.

jordanbrown0:
Your let() looks basically the same as #4478's object().  There's some debate as to whether it's necessary given that #4478 has native syntax for constructing objects.

pkriens:
I think the advantage of my proposal is that

* it is an incredibly small code change, and
* does not include any new syntax, and
* fully supports member access in functions.

It’s a natural change.

I fully understand and support the goal to keep Openscad simple and
declarative but then there should also not be let’s. Let’s introduce
structures but then seem to yell pester you that those structures cannot be
assigned to a variable.

I saw some of the proposals but they seem major and have serious drawbacks
like not having member access. They feel like turning openscad into another
language? Or not?

Anyway, I think openscad is long overdue to have structures. Long overdue …

How do we move this along?

pkriens:
> it needs to be protected by a feature flag

Why would this be need? It won’t affect existing code nor users?

If this would be considers experimental because people have doubts this is
a good feature and we want to try it, it would probably have fall under an
experimental flag?

jordanbrown0:
I'd be perfectly happy starting out with a function like this (but probably not named "let", because that already has meaning in expression context).  Revar proposed object() several years ago, but the sense at the time seemed to be that we wanted a more fully-integrated feature, and that once we had that fully-integrated feature we wouldn't need the function any more.

kintel:
Experimental because the master branch will be promoted to release, and I don't want to have to back out a bunch of code if it turned out to not be release ready. We usually follow some process for promoting features from experimental, which includes thorough API review, proof of usage in the field, documentation, test coverage etc. - an experimental flag allows us to merge features before all of this is in place, and easily back it out if it turned out to be a wrong bet.

kintel:
In terms of this particular feature: If we decide to implement a simple let/object function, would it make sense to merge this with the JSON import, as they kind of solve very similar problems, and could be promoted to a stable feature using similar criteria?

jordanbrown0:
No objection to tying several related features to a single feature flag.  Even if we eventually decide that one is more ready than another, there's nothing that stops us from separating them later.

pkriens:
If #4478 brings me structs, fine. However, it seems like a very large PR. In my experience this generally means that the really uncontroversial stuff gets blocked by the more contentious. Imho it is better to break it up in separate cohesive features.

From that perspective, the lack of a struct in OpenSCAD is very hard to understand. Even C already 60 years ago had structs. All data declaration languages have scalars, vectors/lists, and structs. 

I am willing to do work to get structs/objects/records in. I've no idea how the decision process works here ... how can I get structs in OpenSCAD ASAP? Help :-)

t-paul:
> I am willing to do work to get structs/objects/records in. I've no idea how the decision process works here ... how can I get structs in OpenSCAD ASAP? Help :-)

I think the way to do that would be splitting out the feature from the big PR while showing the result is
1) compatible with later extension from the big PR (from user/language perspective, the internal coding is 99.7% irrelevant here!)
2) working without breaking existing scripts, seems like this is not a high risk for this specific feature
3) having test cases, showing it's doing what it claims to do

kintel:
#4478 is a bit big (which also makes it hard and time consuming to engage with), and [OEP 8](https://github.com/openscad/openscad/wiki/OEP8:-Objects-(dictionaries%3F),-Geometry-as-data,-and-Module-References) is a bit large in scope as well.

I'm off for most of the summer and have very limited opportunities for contribution, but in general I agree with t-paul@ - and would perhaps suggest to implement the large feature set as a number of layers, and iteratively improving it as an experimental feature.

Very vague example of how to approach iteration (where each iteration could be a merged PR growing an experimental feature):
* Object access and a simple way (object()/let(), import JSON, textmetrics) to create objects. This only misses a creation syntax, which is what this PR (vs. object()) is about.
* Helper functions like is_object()
* Object literals
* Iteration (for loops)
* Comprehensions

I'd say geometry as objects and modules as values feel a bit tangential, while informed by and informing the design of this, so good to keep in mind but perhaps manage as separate features.

I'm sure a lot more will come up as people start using the features.

As mentioned above I think the big question (which also drives our priorities atm.) is: Which features have a chance of making it into the next release? Whatever we decide, it will need some soaking time.

We don't have a formal process in place, but as t-paul@ hints at, the guidelines is going to be smth. like: Does it provably work, is there a possibility for breaking any existing designs, are all imaginable corner cases accounted for in automated tests, are power users engaging with the feature, is it documented, do we have example(s), are we going to be able to support this for the next decade without incompatible changes?

pkriens:
I made this proposal the absolute minimal because I felt there is an
attempt to keep OpenSCAD as simple as possible. The `let` syntax seems to
cover this perfectly.

I’d like to make a case for the absolute minimum package to get it in asap.
This such a huge missing feature.

This seems to additionally require:

- is_object()
- testcases
- documentation

I will add this.

There are many ways we could do iteration and comprehension but I’d prefer
to defer to when we have experience. The current PR seems to have more than
sufficient value to add to OpenSCAD anyway. Iteration, comprehension, and
reflection are nice to haves but in no way crucial. Most languages do not
have them over strict. And imho the syntax for these takes lots of
discussions.

I’m not sure why we need object literals, the let syntax is the object
literal?  If we need a literal for external import I’d think JSON is the
only valid option?

I understand that Jordan is responsible for #4478. Jordan indicated that he
wanted a rename of let to object but that this PR is similar to his
proposal. I’d not be terrible upset when we give it another name but
personally I really liked the elegance of the let syntax. Why introduce a
new keyword with the chance of collisions with existing code if not
necessary?

So Jordan, do you see anything would conflict with #4478’s object in the
future?

About the feature flag. If I look at the matrix a feature flag seems not
very useful, does it?

Application   Source
Old                 Old               Ok
Old                  New             Fails
New                Old               Ok
New                New             Ok

The fail Ok/new will not be prevented by the feature flag while the flag
would unnecessarily fail the new/new case. Is that really useful?

kintel:
The feature flag allows us to merge something into the master branch without committing to making it part of the upcoming release. We can skip this if there is unanimous support for the implementation and a core developer committing to taking it over the finish line before the next release.

jordanbrown0:
I'm happy to split #4478 apart (or have somebody else split it apart).

I brought in the geometry-value and module-reference pieces because they were syntactically very close to the object literal pieces, and I wanted to make sure that they all peacefully coexisted.  (Also, for a while my plan was that the geometry-value and object pieces might be the same thing, but eventually decided that I didn't like that idea.)  But now that I am confident that they *can* peacefully coexist, splitting them is fine.

---

Iteration is already implemented and merged.

```
x = fontmetrics();
for (k = x) echo(k);
```
>ECHO: "nominal"
>ECHO: "max"
>ECHO: "interline"
>ECHO: "font"

So is is_object().

---

With respect to `let()` versus some other name... yes, they're very similar.  Except that they do very different things.

Quick, what does this snippet do in the middle of an expression?
```
let(
    nu_curr = [
        if (i1) edge_corners[i1[0]][0],
        each curr,
        if (i2) edge_corners[i2[0]][0],
    ],
    nu_ecns = [
        if (!i1) ecns[0] else [
            for (ecn = edge_corners[i1[0]][1])
                if (ecn != ecns[0]) ecn
        ][0],
        if (!i2) ecns[1] else [
            for (ecn = edge_corners[i2[0]][1])
                if (ecn != ecns[1]) ecn
        ][0],
    ],
    rem = [
        for (i = idx(edge_corners))
            if (i != i1[0] && i != i2[0])
                edge_corners[i]
    ]
)
```

(No peeking...)

It's a trick question.  You don't know yet.  It might be an in-expression set of assignments, or it might be your new let() function.  You can't know until you see what comes next.

In fact, I'm pretty sure you can't tell even then, at least based on syntax.

What does this do, and why?
```
x = let(a=1) -1;
```
?

---

By the way, this proposal is exactly where PR#4337 started.  Additional things that went into that before it was largely abandoned in favor of #4478:
* Computed keys
* Non-identifier keys
* Unioning objects
* Copying an object excluding a particular entry

The major reason that #4337 was abandoned was that there seemed to be a significant feeling that we should have a syntactic representation for objects, and that a syntactic representation would make a function redundant.

I was perfectly happy for #4337 to go in.

jordanbrown0:
Oh and, to amplify a bit on what Marius said, using a feature flag warns people that the feature that they are using is not yet a committed part of the program; they should be prepared for changes or even outright removal.

pkriens:
* **let syntax** – The let syntax implies that you do not know the expression type until you reach the end. This maybe not the best solution but that is OpenSCAD today, and I noted on the list people really like the language. In your example, clearly `x = let(a=1) -1;` has the value `-1` while `x=let(a=1);` will be an object `{a=1}`. The fact that it is unambiguous syntax is also shown by the fact that the change to the grammar was absolutely minimal.
* **context** – A `l=let( a=1, b=function() a);` works as expected,`l.a()` returns 1. In # 4337, `object` is a function thus `o=object( a=1, b=function() a)` would give an error since the arguments have no access to the prior arguments. This is a prime advantage of the `let` syntax, it does not introduce new semantics but would give access to prior entries.
* **keyword** –You do not address the problem that `object` would take an often used word from the namespace likely breaking existing code.
* **Map functionality** – There is a distinct difference between a _struct_ and a _map_. A struct has fields known to the code, there is no need to iterate over keys because they are well known to the programmer. A map is more dynamic. The immutability of OpenSCAD raises a lot of issues for maps with a lot of hairy semantics. I'd love to help out but would prefer if we can treat this as a separate PR to not delay this PR?

I will add
* test cases
* feature
* propose documentation changes

Can a core committer become my mentor to bring this to completion as fast as possible? If @jordanbrown0 could do this, that would be great.

jordanbrown0:
I was mystified by how this could be unambiguous syntax.  It obviously is, because it's the same syntax as is used for assert() and echo().  The answer is that it's unambiguous syntax, but it's ambiguous **lexing**.  If the string "let" is lexed as TOK_LET, you get this interpretation; if it's lexed as TOK_ID, you get a function call, a subtraction, and a number.  TOK_LET matches first.  I had assumed that you'd implemented your new let() semantics as a function, but that's not what you did.

BTW, go look at the syntactic changes for #4337.  (Spoiler:  there aren't any.)

Context... yeah, it's a little terser than some other options, but it's equivalent to `let(a=1) object(a=a, b=function() a)`.  But also this is a way that it's subtly *not* a function.  You cannot say `cylinder(h=10, d1=10, d2=d1*2)`.  So this is a syntax that looks like a function, but isn't a function.

keyword... `object` is not a keyword and should not break anything.  It's just a function name, and user programs can override built-in function names.  Namespace is certainly a consideration, and an is ugly aspect of extending the built-in library (or extending any library), but the problems for built-in functions are all associated with cases where a program that's already using the name is changed to *also* use the new function.  (Same applies to modules.)  It would be nice if we could do something about the function / module namespaces and avoiding conflicts between the core and applications, but at the moment we can't.  (I have some hope around using objects to hold function and module references, reducing the namespace exposure, but we're not there yet.)

struct vs map... while I agree that some languages make that distinction, I don't think it's a necessary distinction for a dynamic-type language like OpenSCAD.  The model I've been using matches JavaScript's "object".  Iteration is already there, but the more relevant question is computed and non-identifier keys.  #4337 addresses those.  (So does #4478, of course.)

I'm not sure exactly how "core contributor" is defined, but I'm not one.

Net... if we want to move ahead with a function-oid object constructor, I'd move ahead with #4337 rather than making the already-kind-of-magic `let()` syntax be even more magic.

pkriens:
I disagree the let syntax is magic, it is a crucial component of the current language. Anybody using OpenSCAD understands that earlier entries are in scope. For using it to construct an object it seems to have only advantages. 

- no name conflict 
- prior entries are in scope ( sq = `let( w=10, h=40, area=w*h)`).
- parser support, when key is missing (you can do object( [], 2))
- absolutely minimal code change
- very little documentation needed

Maybe none are huge but there is absolutely no disadvantage and the small advantages add up imho. You objection that let is magic and we make it more I find not true, let is one of the cornerstones of writing OpenSCAD code. Would I design it differently if I was in charge, probably. But it is the let we have and that people learned to love.

However ...

I find _some_ way of creating structs in OpenSCAD way more important than spending time on the exact details. So if I am the only one arguing for the `let` syntax I will let it slide ...

PR #6010, opened 2025-07-03
Fix test 845

nomike:
This fixes test 845 - preview-cgal_issue3262.

The expected image was taken directly from OpenSCAD.


nomike:
Sorry. This was meant for PythonSCAD but I didn't realize that due to it being a fork github opened a PR for OpenSCAD.

PR #6012, opened 2025-07-04
object function from #4337

pkriens:
This is PR 4337 updated to the current master with additional functionality.

## Garbage Collection Issue in ObjectType

The ObjectType held two references to its values: in the map and in a vector. The ContextMemoryManager uses the smart ptr references to decide if a Context can be garbage collected. This is normally 2 (one ref for the ContextMemoryManager and one from the context) but if you placed a function in an ObjectType, the smart ptr use count would be three. At the end of a context session, the garbage collector must end up with zero contexts otherwise it core dumped.

The implementation I devised has the following qualities:

- values are ordered in insertion order
- a key that is overridden will stay in its original place
- a map is used to do fast retrieval
- deletes will have to rebuild the map, this is delayed until the first time the map is needed

## Diagnostics

When writing the tests, a lot of questions popped up. The code from #4337 was excellent but provided very little feedback where the problem was happening.

## Equality check

For testing, it was necessary to add an equality check. I made the decision that equality was same order keys, same value for each key. A case could be made to make this ordering independent.

## Testing

Tests were added. one test for each function and heavily testing of deletion and insertion ordering. The other tests verifies the warnings that are now generated.

## Sequenced commits

This PR contains 3 commits. 
- The first is the original PR, and
- The second fixes the garbage collection bug, and 
- The last adds the equality, diagnostics and testing.


jordanbrown0:
I'll try to look at this in the next day or two.  (But to be clear:  I don't have authorization to merge changes.)

Just to be clear:  is the GC problem that you discovered in the 4337 changes, or is it present in the current master?

pkriens:
The GC problem is in the current master. 

However, the problem only appears fatal when you store a function in an ObjectType because it has a reference to the current context. I think ObjectType is only used in the text- and fontmetrics and guess they do not store functions?

gsohler:
Ahh this nicht explain, why i faced so severe memory issues During osuse.
Also Handling Contests and they are Prone to Go Out of scope due to the
Architekture Chosen.
Does to current Master Access freed memory? If yes, how does IT Happen?

Peter Kriens ***@***.***> schrieb am Sa., 5. Juli 2025, 12:03:

pkriens:
@gsohler This was not causing memory problems. The runtime would do an assert and coredump if not all Context's were no longer referenced at the end of a session. the console version shows a message, in the GUI version (at least on the mac) it was just poef.

And as I said, although the issue is there, it was never triggered because nobody was putting functions in ObjectType objects.

jordanbrown0:
You are correct that existing mechanisms will never put function references into objects.

src/core/builtin_functions.cc
```diff
@@ -436,6 +436,59 @@ Value builtin_concat(Arguments arguments, const Location& /*loc*/)
...
+    Value value = argument->getExpr()->evaluate(context);
+    if (argument->getName().empty()) {
+      if (value.type() == Value::Type::VECTOR) {
+	const auto &vec = value.toVector();
+	for (const auto& keyval : vec) {
+	  if (keyval.type() != Value::Type::VECTOR || (keyval.toVector().size()!=1 && keyval.toVector().size()!=2)) {
+	    LOG(message_group::Warning, call->location(), context->documentRoot(), "un-named object() arguments must be another object, a list of [keystring,value] pairs to set, or a list of singleton [keystring] names to delete.");
+	  } else {
+	    if (keyval[0].type() != Value::Type::STRING) {
+	      LOG(message_group::Warning, call->location(), context->documentRoot(), "un-named object() arguments must be another object, a list of [keystring,value] pairs to set, or a list of singleton [keystring] names to delete.");
```
jordanbrown0: This case can be combined into the previous case, so that it's

* if the value isn't a vector
* or the vector has a length other than 1 or 2
* or the first element of the vector isn't a string

then fail.

jordanbrown0:
It is really unfortunate that `ContextMemoryManager` requires a `std::vector` of the values, because that's what drives a lot of this complexity.

Absent that, we'd just keep a `std::unordered_map` of `key:value`, plus a std::vector of the keys.  Add-entry would add to the vector and delete would delete from it, but otherwise it wouldn't require modifying.  I would think that that would be do-able by having CMM require an iterable object, where both vectors and maps are iterable, and leave *how* you iterate up to the individual class.  But my C++ isn't strong enough to know how to say that.  I would have expected that unordered_map and vector would be subclasses of some "iterable" superclass, but they aren't.

I have to suspect that there is a way to do this that I just don't know.  Maybe returning a lambda function?

src/core/Value.cc
```diff
@@ -1266,68 +1266,20 @@ ObjectType::ObjectType(EvaluationSession *session) :
...
-{
-  return ptr->keys;
-}
+const Value& ObjectType::get(const std::string& key) const              { return ptr->get(key); }
+bool ObjectType::set(const std::string& key, Value&& value)             { return ptr->set(key,value); }
+bool ObjectType::del(const std::string& key)                            { return ptr->del(key) != NOINDEX; }
+bool ObjectType::contains(const std::string& key) const                 { return ptr->find(key)!= NOINDEX; }
+bool ObjectType::empty() const                                          { return ptr->values.empty(); }
+const std::vector<std::string>& ObjectType::keys() const                { return ptr->keys; }
+const std::vector<Value>& ObjectType::values() const                    { return ptr->values; }
```
jordanbrown0: This makes sense, but note that `VectorType` (which `ObjectType` largely copies) doesn't do it; `VectorObject` has only `size()` and `empty()` and otherwise the logic is in `VectorType`.  I'm not saying that's the rightest approach, but it's the existing approach.

src/core/Value.h
```diff
@@ -389,13 +390,69 @@ class Value
...
-  std::vector<std::string> keys;
-  std::vector<Value> values;
+
+  // for consistency, it is important that the 
+  // order of the keys+values remains the insertion
+  // order. No such map exists in the C++ lib.
+  // However, for the garbage collection in ContextMemoryManager.cc
+  // it is paramount we only have 1 reference to a Value
+  // so we use a map for performance but maintain an index
+  // instead of a secondary reference.
```
jordanbrown0: Mention that CMM wants a `std::vector`.

jordanbrown0:
Along the same lines, it seems like if we're going to have a vector it should be a vector of {name,value} pairs, but CMM wants a vector of just the values.  (But best would be to have us keep the values in the map, and somehow have CMM iterate across the map.)

src/core/builtin_functions.cc
```diff
@@ -436,6 +436,59 @@ Value builtin_concat(Arguments arguments, const Location& /*loc*/)
...
+	}
+      } else if (value.type() != Value::Type::OBJECT) {
+	LOG(message_group::Warning, call->location(), context->documentRoot(), "un-named object() arguments must be another object, a list of [keystring,value] pairs to set, or a list of singleton [keystring] names to delete.");
+      } else {
+	// Argument is another object to copy from.
+	const auto obj = value.toObject();
+	for (const auto& key : obj.keys()) {
+	  result.set(key, obj.get(key).clone());
+	}
+      }
```
jordanbrown0: I'd prefer that this was
```
else if (value.type() == Value::Type::OBJECT) {
 ... stuff ...
} else {
    ... error ...
}
```
or maybe `switch (value.type()) { ... }`.

src/core/builtin_functions.cc
```diff
@@ -436,6 +436,59 @@ Value builtin_concat(Arguments arguments, const Location& /*loc*/)
...
+    } else {
+      // Argument is a named key=value to add or set.
+      result.set(argument->getName(), std::move(value));
+    }
+  }
+  return std::move(result);
+}
+
+Value builtin_has_key(Arguments arguments, const Location& loc)
```
jordanbrown0, line 543: I can't say that I'm totally happy with this name.

Maybe "contains"?  Maybe it would be polymorphic and mean something for other types?

C++ std::unordered_map calls this "contains".  Java calls it "containsKey" (with "contains" searching for a matching value rather than a matching key).  Python says "k in dict", as does JavaScript.

src/core/Value.cc
```diff
@@ -720,12 +720,32 @@ bool Value::isUncheckedUndef() const
...
+    for ( size_t i = 0; i< this->ptr->values.size(); i++){
+        auto a = this->ptr->keys[i] != other.ptr->keys[i];
+        if ( a ) {
+            return false;
+        }
+
+        auto b = this->ptr->values[i] != other.ptr->values[i];
+        if ( b.toBool() ) {
+            return false;
+        }
```
jordanbrown0, line 740: The temporary variables (a, b) don't hurt, but they don't add much value.

I don't have a strong feeling on whether two objects that have the same keys with the same values are equal.  My immediate reaction is that they are, that ordering is for presentation only and otherwise keys are considered unordered.

src/core/builtin_functions.cc
```diff
@@ -436,47 +436,121 @@ Value builtin_concat(Arguments arguments, const Location& /*loc*/)
...
-      }
-    } else {
-      // Argument is a named key=value to add or set.
-      result.set(argument->getName(), std::move(value));
+    ObjectType result(context->session());
+    int n = 0;
+    for (const auto& argument : call->arguments) {
+        Value value = argument->getExpr()->evaluate(context);
+        if (argument->getName().empty()) {
+            const auto error = builtin_object_unnamed(result,value, n);
```
jordanbrown0, line 530: Should this have `std::move(value)`?  (I don't fully understand when it should and should not be used, or why this case is different from the `result.set(` case below.)

src/core/builtin_functions.cc
```diff
@@ -436,47 +436,121 @@ Value builtin_concat(Arguments arguments, const Location& /*loc*/)
   return std::move(result);
 }
 
+static Value warning( const std::shared_ptr<const Context>& context, const FunctionCall *call, const std::string & msg) {
+    LOG(message_group::Warning, call->location(), context->documentRoot(), msg.c_str());
```
jordanbrown0: Fourth argument to LOG is a printf-style string, so must be tightly controlled.  Should be `"%s", msg.c_str()`.  (Yes, likely the string does not contain any `%`, but it's bad practice to even risk it.)

jordanbrown0: This new `warning()` function seems like a good addition that can reduce a lot of clutter, but unless it gets broad adoption it will just be an inconsistency.

src/core/builtin_functions.cc
```diff
@@ -436,47 +436,121 @@ Value builtin_concat(Arguments arguments, const Location& /*loc*/)
...
+    std::snprintf(&output[0], size+1, format.c_str(), std::forward<Args>(args)...);
+    return output;
+}
+
+#define OBJECT_HELP "In an unnamed list, entries must be [key,value] to set or [key] to delete. The key must be <string>."
+
+static std::string builtin_object_unnamed(ObjectType & result, Value & value, int arg_index) {
+
+    std::string prior_args = std::string();
+        for ( int i=0; i<arg_index; i++) prior_args += "_, ";
```
jordanbrown0: Should only do this when we have a problem.

Other cases, e.g. min() and max(), as implemented in min_max_arguments(), report the argument index rather than these "_, " placeholders.

src/core/builtin_functions.cc
```diff
@@ -436,47 +436,121 @@ Value builtin_concat(Arguments arguments, const Location& /*loc*/)
...
+    }
+    case Value::Type::VECTOR: {
+        std::string prior_entries;
+
+        const VectorType & vector = value.toVector();
+        for (const auto& member : vector) {
+
+            if (member.type() != Value::Type::VECTOR) {
+                return format(
+            "object( %s[%s<%s>] ) Entry type is not a list, it is <%s>. " OBJECT_HELP,
```
jordanbrown0: For i18n, I believe you must use `%1$s` form, where `1` is the printf argument index.

src/core/builtin_functions.cc
```diff
@@ -436,47 +436,121 @@ Value builtin_concat(Arguments arguments, const Location& /*loc*/)
...
+{
+    auto size = std::snprintf(nullptr, 0, format.c_str(), std::forward<Args>(args)...);
+    std::string output(size + 1, '\0');
+    std::snprintf(&output[0], size+1, format.c_str(), std::forward<Args>(args)...);
+    return output;
+}
+
+#define OBJECT_HELP "In an unnamed list, entries must be [key,value] to set or [key] to delete. The key must be <string>."
+
+static std::string builtin_object_unnamed(ObjectType & result, Value & value, int arg_index) {
```
jordanbrown0: I believe that a more usual pattern would be to return nothing here, and throw (via LOG) on error.

src/core/builtin_functions.cc
jordanbrown0: ... and I'd say that applies to both the argument count and the element-in-list count, so it would say something like "argument N, vector element M".

src/core/builtin_functions.cc
jordanbrown0: Note the existing `print_argConvert_positioned_warning() function`.

src/core/builtin_functions.cc
```diff
@@ -436,47 +436,121 @@ Value builtin_concat(Arguments arguments, const Location& /*loc*/)
...
+                    result.del(key);
+                } else {
+                    result.set(key,entry[1].clone());
+                }
+                break;
+            };
+
+            case 0: return format("object(%s[%s[]]) Entry is empty." OBJECT_HELP, 
+                                prior_args.c_str(), 
+                                prior_entries.c_str());
```
jordanbrown0: I wouldn't really feel the need to distinguish the 0 case from the >2 case.

src/core/builtin_functions.cc
```diff
@@ -436,47 +436,121 @@ Value builtin_concat(Arguments arguments, const Location& /*loc*/)
...
+
+            default: return format("object(%s[%s[/* %d entries */]]) Entry length is not 1 or 2 but %d. " OBJECT_HELP, 
+                    prior_args.c_str(), 
+                    prior_entries.c_str(),
+                    entry.size(), 
+                    entry.size());
```
jordanbrown0: Printing the same value (entry.size()) twice is a hint that one of them isn't necessary.

I'd think "argument N, element M, has X entries, must be 1 or 2".

tests/data/scad/experimental/object/object-tests.scad
```diff
@@ -0,0 +1,83 @@
+o1 = object(a=1,b=2,c=[3], f= function() 1);
+
+// printouts
+
+echo("basic format              ",                  o1);
+echo("basic format of a copy    ",                  object(o1));
+
+// type functions
+
+test( function() !is_undef(o1),                     "o1 must not be undef");
```
jordanbrown0, line 11: seems like
```
assert(!is_undef(o1));
```
would be enough, or maybe
```
assert(!is_undef(o1), "o1 must not be undef");
```
(but I don't think that actually adds much)

tests/data/scad/experimental/object/object-tests.scad
```diff
@@ -0,0 +1,83 @@
+o1 = object(a=1,b=2,c=[3], f= function() 1);
```
jordanbrown0, line 1: For completeness, you might add a member that is itself an object.

tests/data/scad/experimental/object/object-tests.scad
```diff
@@ -0,0 +1,83 @@
...
+// member access
+
+test( function() o1["a"]==1,                        "o1['a'] must be 1");
+test( function() o1["c"]==[3],                      "o1['c'] must be [3]");
+test( function() o1.a==1,                           "o1.a' must be 1");
+test( function() o1.c==[3],                         "o1.c must be [3]");
+test(function() is_function(o1["f"]),               "o1['f'] must be a function");
+test(function() is_function(o1.f),                  "o1.f must be a function");
+test( function() [for (i=o1) has_key(o1, i)] 
+    ==  [true, true, true, true],                   "all keys must be present");
```
jordanbrown0, line 27: Or
```
for (i=o1) assert(has_key(o1, i));
```

tests/data/scad/experimental/object/object-tests.scad
jordanbrown0: But I guess this style allows reporting multiple failures, where assert() would end the test after the first one.
Shrug.

tests/data/scad/experimental/object/object-tests.scad
```diff
@@ -0,0 +1,83 @@
...
+
+testEq( object(a=4, b=2, c=[3]), object(o1, [["f"], ["a",4]]),                      "!,a=4");
+testEq( object(a=1, b=4, c=[3]), object(o1, [["f"], ["b",4]]),                      "!f,b=4");
+testEq( object(a=1, b=2, c=[4]), object(o1, [["f"], ["c",[4]]]),                    "!f,c=[4]");
+testEq( object(a=1, b=2, c=[3]), object(o1, [["f"]]),                               "!f");
+testEq( object(a=4, b=2, c=[3],x=42), object(o1, [["f"], ["a",4]], x=42),           "!f,a=4,x=42");
+testEq( object(a=1, b=4, c=[3],x=42), object(o1, [["f"], ["b",4]], x=42),           "!f,b=4,x=42");
+testEq( object(a=1, b=2, c=[4],x=42), object(o1, [["f"], ["c",[4]]], x=42),         "!f,c=[4],x=42");
+testEq( object(a=1, b=2, c=[4],x=undef), object(o1, [["f"], ["c",[4]]], x=undef),   "!f,c=[4],x=undef");
+testEq( object(a=1, b=2, c=[3],x=undef), object(o1, [["f"]], [], [], [], x=undef),  "!f []... x=undef");
```
jordanbrown0, line 60: The messages for all of these seem a bit cryptic.  I think you're saying "we do not expect to find this at the end of the object".  Maybe "deleting f, replacing a", "deleting f, replacing b", et cetera.

tests/data/scad/experimental/object/object-tests.scad
```diff
@@ -0,0 +1,83 @@
...
+os = [ for ( i = [0:1000] ) object( v = i )];
+test(function() len(os) == 1001,                    "check size");
+test(function() os[1000].v == 1000,                 "check value");
+
+entries = [ for( i=[1:100000] ) [ str("_",i), i]];
+o = object(entries);
+test( function() o._99999 == 99999,                 "test 100k entries");
+
+// do random access to 100k entries 100k times
+access = [ for ( n = rands(0,100000,100000)) str("_",floor(n)) ];
```
jordanbrown0: Might set a seed (fourth arg to rands()) so that all runs test the same thing.

tests/data/scad/experimental/object/object-tests.scad
```diff
@@ -0,0 +1,83 @@
...
+for ( i=access ) {
+    x=o[ i ];
```
jordanbrown0: Might `assert(o[i] == i)`.

tests/data/scad/experimental/object/object-warning-tests.scad
```diff
@@ -0,0 +1,7 @@
+o1 = object(a=1,b=2,c=[3], f= function() 1);
+echo("warnings");
+echo( object(o1, [[true,5]]), "true as remove key");
```
jordanbrown0: That's a "set" key, not a "remove" key.

jordanbrown0:
Done reviewing.

Mostly style issues.  The biggest thing is wishing that CMM could directly traverse the std::unordered_map, instead of needing to construct a vector of the values.

src/core/builtin_functions.cc
jordanbrown0: Stale.

src/core/builtin_functions.cc
jordanbrown0: Stale.

src/core/builtin_functions.cc
jordanbrown0: Ah, now I understand the github process.  I could have deleted this, instead of submitting it and then resolving it.

src/core/builtin_functions.cc
jordanbrown0: This line, and many below, ends with a space.  I don't know whether OpenSCAD wants to avoid that, but at work we avoid it because some editors will eat that extra space and the result is spurious diffs.

jordanbrown0:
I pulled this over and built it and exercised it a little (nothing all that exotic), and so far it's fine.

src/core/builtin_functions.cc
pkriens: This is also my first encounter with std::move … I last used C++ in the nineties.

With the help of some ChatGPT:

* My IDE (Eclipse) is very good at showing errors/warnings/info. I keep the code 100% clean.
* `std::move` enables an optimization that allows a type to transfer ownership of resources via a move constructor or move assignment, avoiding deep copy.
* `std::move` is a cast to an rvalue reference; it does not move anything by itself.
* Since ObjectType is a kind of smart pointer wrapping a shared_ptr, copying is cheap, so the benefit of std::move may be negligible in this case.
* Modern compilers optimize return values with Named Return Value Optimization (NRVO). If NRVO applies, no copy or move occurs, making std::move on the return potentially unnecessary — and in fact, it can prevent NRVO in some compilers.
* It is hard to imagine that this code is used so intensely that any performance issues in copy/move constructors would be noticed. 
* 
Therefore, in the return case you mention, the cast is not necessary; the compiler will typically use NRVO or fall back to move construction. The call to `set` the move is mandatory because set is defined as `set( ... Value && v)` requiring the cast to an rvalue.

I propose to remove the the `std::move` in the return value.

src/core/builtin_functions.cc
pkriens: That is a good one! agree this is bad. Did not know LOG was printf style.

pkriens: I removed it. When I started I did not refactor out the builtin_object_unnamed function and the warning was used in many places. However, after the refactoring, it was only needed in one place.

So removed.

src/core/builtin_functions.cc
pkriens: Disagree, but I defer to you here.

src/core/builtin_functions.cc
pkriens: Can you elaborate?

src/core/builtin_functions.cc
pkriens: I can change this. However, I've fought my whole life to not use this pattern :-( 

Exceptions are imho for programmer errors and this is 100% input validation. In Rust for example, they bypass exceptions by always return a monad that is either a result value for ok or information when something went wrong.

Are you strong on this?

src/core/builtin_functions.cc
pkriens: That looks very interesting! I missed that. I don't think I can use immediately since I do not have expected values and I am a bit worried about printing out the whole value. They are often huge.

I am not sure what you want me to do here?

src/core/builtin_functions.cc
pkriens: I always detest it when you get an error that says it is error A OR error B. 

We know the different cases, why not report it?

src/core/builtin_functions.cc
pkriens: ok

tests/data/scad/experimental/object/object-tests.scad
pkriens: and assert is not reported in the echo test driver ...

tests/data/scad/experimental/object/object-tests.scad
pkriens: yes, good idea!

tests/data/scad/experimental/object/object-tests.scad
pkriens: Are you sure? Assume there is a pattern that fails. If we pick a seed we never ever will find it. None of the indexes is printed or referenced so the exact sequence is a dont care for the test. However, on the infinite small change that there is a pattern dependency, wouldn't it be nice to get a crash so we can look deeper?

tests/data/scad/experimental/object/object-tests.scad
pkriens: good idea.

tests/data/scad/experimental/object/object-warning-tests.scad
pkriens: oops

src/core/builtin_functions.cc
pkriens: Good catch. Using Eclipse and assumed it fixed this up on save.

src/core/Value.cc
pkriens: The reason I used this style is because the relation between keys/values/map is very tight and error prone. I wanted to be sure they were all tightly in one place to prevent mistakes. In VectorType, the data structure is trivial.

The data encapsulation is imho an important argument but I am the newcomer here :-)

src/core/Value.h
pkriens: ok

I saw your comments. I seriously considered changing CMM but in my current apprentice level I felt it would take a long time to get any change through. 
A lambda would be the correct solution imho, this lambda could then add the values to `valueQueue`. However, this is so much at the heart of the algorithm (which I still don't really get, and I spent a day in that code) that I lack the courage to touch it.

There is a way to iterate over map but you iterate over pair<std::string,Value>. Also notice that the unordered map is exactly that, so it would randomly order the elements. The map is ordered but by lexical order of the key. Neither is a good solution. So you would still need to implement extra logic for this ordering.

src/core/builtin_functions.cc
pkriens: I think it is good enough. `has_key` is shorter than `contains_key` and `contains` raises a lot of issues because what would it mean when you pas a vector and string? Neither type has a meaning for `has_key` but `contains` implies more. Contains a value? Contains an index? hairy.

But I think you picked it? :-) We can skip it if you want, you can discover it with the `object()` method, albeit inefficient.

src/core/Value.cc
pkriens: Problem is that the first expression is `bool` and the second is `Value`. Having them in one `if` expression caused my IDE to warn about a deleted object that was used. I admit a and b are not useful names, it was just a struggling to get rid of the warning.
I renamed the variables so they at least increase the readability as well.
```
    for ( size_t i = 0; i< this->ptr->values.size(); i++){
        auto key_the_same = this->ptr->keys[i] != other.ptr->keys[i];
        if ( key_the_same ) {
            return false;
        }

        auto value_the_same = this->ptr->values[i] != other.ptr->values[i];
        if ( value_the_same.toBool() ) {
            return false;
        }
    }
```

tests/data/scad/experimental/object/object-tests.scad
pkriens: I want to be sure I have 4 keys ... and found this more readable ... but arbitrary.

tests/data/scad/experimental/object/object-tests.scad
pkriens: The messages are more an identifier. In the rare case in the future when you get an error, it is usually hard to find what failed. The description of the test is rarely enough. By using a coded name, you have quick short  handle. Since your type of descriptions hold many long words, it is hard to see what differs and the handle loses its purpose.

pkriens:
@jordanbrown0 did a first pass. The issues not resolved I am not sure what to do with it. Maybe you can resolve them after reading my answer or comment what you want to happen.

Thanks for taking the time to review this. Hope we can get this in OpenSCAD soon.

src/core/builtin_functions.cc
```diff
@@ -436,6 +436,129 @@ Value builtin_concat(Arguments arguments, const Location& /*loc*/)
...
 }
 
+template <typename ...Args>
+std::string format(const std::string& format, Args && ...args)
+{
+    auto size = std::snprintf(nullptr, 0, format.c_str(), std::forward<Args>(args)...);
+    std::string output(size + 1, '\0');
+    std::snprintf(&output[0], size+1, format.c_str(), std::forward<Args>(args)...);
+    return output;
+}
```
kintel: We're using boost::format() in most other places. To reduce divergence, is that feasible instead of introducing this custom function?

src/core/builtin_functions.cc
```diff
@@ -436,6 +436,129 @@ Value builtin_concat(Arguments arguments, const Location& /*loc*/)
...
+
+#define OBJECT_HELP "In an unnamed list, entries must be [key,value] to set or [key] to delete. The key must be <string>."
```
kintel, line 440: Nit: Using "constexpr auto" instead of a #define is often better in modern C++; same cost, and some C++ tooling isn't always aware of the preprocessor.

src/core/builtin_functions.cc
kintel: It would be nice to use smth. like absl::StatusOr in general, but that's a larger change. I say keep it if the intent is to avoid LOG from lower-level functions like this. This is a good opportunity for future refactoring, but we probably want to align the methodology across the codebase.

src/core/builtin_functions.cc
kintel: value is read-only in this context. The correct change, IMO, is to modify the `value` parameter of `builtin_object_unnamed()` to `const Value &`.

src/core/builtin_functions.cc
```diff
@@ -436,6 +436,129 @@ Value builtin_concat(Arguments arguments, const Location& /*loc*/)
...
+    for (const auto& argument : call->arguments) {
+        Value value = argument->getExpr()->evaluate(context);
+        if (argument->getName().empty()) {
+            const auto error = builtin_object_unnamed(result,value, n);
+            if ( !error.empty()) {
+                LOG(message_group::Warning, call->location(), context->documentRoot(), error.c_str());
+                return Value::undef(error);
+            }
+        } else {
+           result.set(argument->getName(), std::move(value));
```
kintel, line 536: While `move()` makes sense here, the effect is nulled by the `set()` implementation:

`bool ObjectType::set(const std::string& key, Value&& value) { return ptr->set(key,value); }`
-> By common convention, Value should rather be passed by value here `Value value`, then moved: `ptr->set(key, std::move(value))`

`bool ObjectObject::set(const std::string & key, Value & value);`
-> Value should also be passed by value here: `Value value`, as it's always moved in the implementation.

Generally I think `&&` function arguments should be used sparingly and only in special situations, which tend to be pretty rare in practice.
This article explains why passing by value combined with move() is often the best solution: https://borjaportugal.com/2023/12/20/move-semantics/

src/core/builtin_functions.cc
kintel: No real preference for me on naming, except having such a function is nice. Perhaps naming becomes more clear if we decide to do smth. similar for lists, like JS' `Array.includes()`

src/core/builtin_functions.cc
```diff
@@ -436,6 +436,129 @@ Value builtin_concat(Arguments arguments, const Location& /*loc*/)
...
+    }
+}
+/**
+    The builtin_object function takes either a named or unnamed argument.
+    A named argument is assigned with name=value to the result object. An unnamed
+    argument is either an object or a vector. An object will have its
+    entries copied, potentially overwriting earlier entries. A vector
+    must be of vectors. It can contain as member vectors:
+
+        ["k"]       deletes key k
```
kintel, line 518: The delete semantics is a bit surprising/uncommon, but it kinds of fits so I'm ok keeping it. We can always change or minds later as long as we're still experimental.

src/core/Value.h
```diff
@@ -386,13 +390,70 @@ class Value
   Variant value;
 };
 
+static size_t NOINDEX = std::string::npos;
```
kintel: constexpr

tests/data/scad/experimental/object/object-tests.scad
```diff
@@ -0,0 +1,89 @@
...
+testEq( all, all,                                   "same object with all types must be equal");
+testEq( o1, object(o1),                             "copy must be equal");
+testNEq( o1, object(o1,[["f"]]),                    "copy must not be equal when f removed");
+testNEq( all, object(all,[["o"]]),                  "copy must not be equal when o removed for all types");
+
+// editing
+
+testEq( object(a=4, b=2, c=[3]), object(o1, [["f"], ["a",4]]),                      "!,a=4");
+testEq( object(a=1, b=4, c=[3]), object(o1, [["f"], ["b",4]]),                      "!f,b=4");
+testEq( object(a=1, b=2, c=[4]), object(o1, [["f"], ["c",[4]]]),                    "!f,c=[4]");
```
kintel, line 54: Since an empty key is allowed, we should add a test for it.

src/core/builtin_functions.cc
```diff
@@ -436,6 +436,129 @@ Value builtin_concat(Arguments arguments, const Location& /*loc*/)
...
+                    result.set(key,entry[1].clone());
+                }
+                break;
+            };
+
+            case 0: return format("object(%s[%s[]]) Entry is empty." OBJECT_HELP, 
+                                prior_args.c_str(), 
+                                prior_entries.c_str());
+
+            default: return format("object(%s[%s[...]]) Entry length is %d, must be 1 [key] or 2 [key,value]. " OBJECT_HELP,
```
kintel: Should we add a warning test for this as well?

kintel (commented):
Added a few comments, I think we're close to being able to merge this.
There are some refactoring opportunities in this part of the code; tighten up style, modernize C++, clean up interfaces, but let's refactor later when someone has motivation.

src/core/builtin_functions.cc
pkriens: ok, that is good to idea.

src/core/builtin_functions.cc
pkriens: I'd prefer to keep it. Have to do a runtime work to concatenate the format with the standard part. If you think it is important I will obey of course but as an old embedded programmer I hate to move things to runtime that can be handled in the compiler ...

kintel: Performance impact is minimal as it's only used in error handling, but bine for now, we can align such things later if we want to.

src/core/builtin_functions.cc
pkriens: Read the blog, was interesting. 

* copy constructor is deleted so when you pass by value you need either std::move or clone(). This seems to go against the blog?
* It creates a lot of mental housekeeping, is it worth it? Value is already kind of a smart ptr to a smart ptr?

Anyway, I now got:
```
    bool set(const std::string & key, Value  value) {
        size_t index = find(key);
        if ( index != NOINDEX) {
            // if contains key, keep at same position
            values[index] = std::move(value);
        } else {
            if (!map.empty()) {
                map[key]=keys.size(); // incremental update
            }
            keys.emplace_back(key);
            values.emplace_back(std::move(value));
        }
        return index == NOINDEX;
    }
```

src/core/Value.h
pkriens: done

tests/data/scad/experimental/object/object-tests.scad
pkriens: good idea

src/core/builtin_functions.cc
kintel: The Value class itself may not be following best practices. If you run into any issues, just drop it and fall back to "classic C++" - we may want to do some deep refactoring of the Value class when we have some time, to ease future maintenance.

src/core/builtin_functions.cc
pkriens: I read the blog and that did enlighten me. However, it feels that the deleted constructor in Value goes against the pattern to pass by value? It might be my long absence of C++ but the std::move feels like another worry in my code :-( And since Value is already a kind of smart ptr, the value seems extremely low?

Anyway, I've update the code and in ObjectObject::set to follow this pattern but would appreciate if you could take another look.

src/core/builtin_functions.cc
pkriens: Using boost::format now

src/core/builtin_functions.cc
pkriens: I started to use a Result class in Java a couple of years ago. It does clean up the code a lot.

pkriens:
@kintel thanks for the quick review! If there is anything more, let me know, I think I made all updates. @jordanbrown0 still had some comments but I do not think they were blocking and this is still experimental.

tests/data/scad/experimental/object/object-tests.scad
```diff
@@ -0,0 +1,98 @@
...
+module testEq( a, b, s) {
+    if ( a != b) {
+        echo("FAIL:", a, "==", b, s );
+    }
+}
+module testNEq( a, b, s) {
+    if ( a == b) {
+        echo("FAIL:", a, "==", b, s );
+    }
+}
```
kintel: Once concern about this: Test are passing silently, meaning that if we introduce a bug failing to call, or accidentally remove, a test, it will go undetected, IIUC.

While this is common practice for unit tests, it poses a bit of a risk for integration/regression tests like this.
Would it make sense to explicitly PASS every test?

Related to this, we could probably do a bit of test refactoring to build out common test tools and use them in a more coherent fashion across all such types of tests, but that's something for another time. The cleanup opportunities are pretty substantial :)

src/core/builtin_functions.cc
kintel: For classes like Value, which is used a lot, avoiding extra reference counts _may_ have a positive effect on performance, but I haven't really looked into it. Barring that, making copies of pointers usually has negligible cost, and the remaining value is really only consistency and living up to common expectations of C++ codebases.  That said, your changes look good, let's roll with that.

tests/data/scad/experimental/object/object-tests.scad
pkriens: The reason I dislike this (but this is not my project :-) is that it is hard to find the issue when there is a lot of noise in the output. In projects I am responsible for 0 warnings is an absolute requirement for this reason. Once you accept one warning as ok, you miss the other 763 ...
And how could we miss a test? You'd have to remove it from the data file and NOT from the regression echo file. If you want malice, you'd also remove it from the regression echo file? So imho you just made it harder to add new tests ... As I told many of colleagues: embrace the silence, only report the exceptions. 

Anyway, I added a pass message.

kintel: Thx! The reason I prefer the "PASS" strategy here: While the overall test file roughly corresponds an entire test run, each individual "test()" invocation in OpenSCAD corresponds to a single, named, unit test. When running a test suite on a piece of software, you generally don't want silence, you want a list of tests executed correctly. You may not want to present all of that to the user running tests, but the underlying logs will show what got executed. At least that's the case on every system I've worked on :)

src/core/Value.cc
jordanbrown0: It seems unfortunate that comparison operators between Values yield Value instead of yielding boolean.  But that's not today's project.

src/core/builtin_functions.cc
jordanbrown0: I'm not sure whether I originated the name or Revar did.

I'm OK with leaving it for now; we might want to reconsider before we un-experimental-ize it.

src/core/builtin_functions.cc
jordanbrown0: I don't speak C++ well enough to have an opinion here.

src/core/builtin_functions.cc
```diff
@@ -436,26 +437,11 @@ Value builtin_concat(Arguments arguments, const Location& /*loc*/)
...
-}
-
 #define OBJECT_HELP "In an unnamed list, entries must be [key,value] to set or [key] to delete. The key must be <string>."
 
-static std::string builtin_object_unnamed(ObjectType & result, Value & value, int arg_index) {
+static std::string builtin_object_unnamed(ObjectType & result, const Value & value, int arg_index) {
 
-    std::string prior_args = std::string();
-        for ( int i=0; i<arg_index; i++) prior_args += "_, ";
+    std::string prior_args = "Argument " + std::to_string(arg_index) + " ";
```
jordanbrown0, line 444: Seems odd to always construct this string, and then use it only on error.  Why not just "Argument %d" at appropriate points in the error messages?

src/core/builtin_functions.cc
```diff
@@ -466,18 +452,19 @@ static std::string builtin_object_unnamed(ObjectType & result, Value & value, in
         return "";
     }
     case Value::Type::VECTOR: {
-        std::string prior_entries;
 
         const VectorType & vector = value.toVector();
+        int element = 0;
         for (const auto& member : vector) {
+            std::string prior_entries = "Element " + std::to_string(element++) + " ";
```
jordanbrown0, line 459: As for prior_args above.

jordanbrown0 (approved):
Close enough.

I want to come back to the need for the std::vector, but that can be another day.

src/core/Value.cc
jordanbrown0: Shrug.

jordanbrown0:
Actually, hold a moment on that approval.  I didn't navigate github right and didn't look at the responses to my comments.

src/core/builtin_functions.cc
jordanbrown0: I suspect that the issue is still present for boost::format().  @kintel?

Consider a message that says "x is type %s, but must be type %s".  Now we want to translate that into Yoda-speak, where we would say "Must be type %s, but type %s x is".

The i18n way to do this is to use %n$s form:
* x is type %1$s, but must be type %2$s
* Must be type %2$s, but type %1$s x is

I suppose it isn't necessary on the English versions of the messages, since those have the luxury of requiring that the args be presented in the order that they need, but existing convention (look at LOG() calls) is to use the %n$s form.

jordanbrown0: Close enough.

src/core/builtin_functions.cc
jordanbrown0: I agree that "error A OR error B" is a problem - it's one of my pet peeves - but I would say that this is really one error case:  you must give either one or two elements, and you gave something else.

But it's not a big deal either way.

tests/data/scad/experimental/object/object-tests.scad
jordanbrown0: OK

jordanbrown0:
I think it's OK, though it might be best to change the %s references to %n$s references.

src/core/builtin_functions.cc
jordanbrown0: For some reason I missed this in my last pass over the comments.  (I think I skipped over "resolved" items.  Review UIs like this really need a two-step resolution process:  author says "I think this is resolved", reviewer says "I agree that it is resolved".)

I'd say that exceptions are for unusual conclusions to a sequence.  Some of them are programmer errors, some of them are user errors, some of them are not errors at all. (For instance, failing to open a file could be any of the three.)

In OpenSCAD with hardwarnings set, an error deep in the execution of the user's OpenSCAD-language program aborts its execution, and needs to blow out through an arbitrary stack of C++ calls.  It's much, much simpler if that stack does not have to explicitly check each call to see if it failed.  It seems sensible to start that blowing-out process at the point where the error is detected, not a level or two higher than that.

If Marius is OK with it then I'm OK with it, but I'd be very nervous about expanding a return-code-based error scheme across the program.

src/core/Value.h
jordanbrown0: I had missed that std::unordered_map iterators yield pairs.  (Looking at [cppreference](https://en.cppreference.com/w/cpp/container/unordered_map.html), I still don't find where it says that, but the example makes it clear that it does.)

So indeed the standard iterators wouldn't do the trick; you'd need a different iterator that just returns the value.

But the concept remains:  have a way to give CMM the list of values, without them having to be stored in a std::vector.

CMM doesn't need the values in any particular order.

Indeed, we want to preserve order, mostly for UI reasons, but the data structures are simpler if we don't have to have a std::vector of the values.  We could have a map of key:value, and a vector of keys.  Adding an entry is adding to both of them.  Modifying an entry is just changing the map.  Deleting an entry is deleting the entry from the map and from the vector.  It wouldn't invalidate the map when you delete an entry.  (But it's ugly that you'd need to linearly search the vector to find the entry.)

Close enough for now.

tests/data/scad/experimental/object/object-tests.scad
jordanbrown0: OK

jordanbrown0:
Sorry for the back-and-forth there; I don't have a lot of experience navigating github reviews and I kept missing stuff.  I think I've looked at everything now.  There are a couple of things that I'm a little concerned about, but they aren't *problems* and they can be cleaned up later if desired.

src/core/Value.h
jordanbrown0: I found it in cppreference.  In the "Member types" section, it says that iterator is an iterator to `value_type`, and `value_type` is std::pair<Key, T>.

src/core/builtin_functions.cc
pkriens: ok, I misunderstood. 
But then it is still not necessary? The translator will translate the format string. He can use the reordering when the need is there?

Are we having translations? Should the messages be translated?

src/core/builtin_functions.cc
pkriens: Because it is used in several places. First rule of software is never to repeat yourself.

pkriens:
Thanks for the comments @jordanbrown0.

So I hope we're good to go now?

Next process is then to get this out of experimental. How would that process go and what can I do to expedite? After the merge, would it be an idea if I wrote an explanation on the OpenSCAD list to get more feedback and build buy-in?

kintel:
In terms of moving out of experimental: The most important is soak time where people start using it and offer feedback. Posting on the mailing list is a great start. Writing docs and/or examples may help people get started.

Then we can start an overall GitHub ticket with a list of tasks/issues needing work, similar to what already exists for other features.

src/core/builtin_functions.cc
jordanbrown0: Yeah, but the other first rule is KISS.

src/core/builtin_functions.cc
jordanbrown0: Yes, OpenSCAD is translated.  I don't know exactly what that process is, but I know that individual contributors are not expected to arrange for translation of the messages that they add.

Indeed, I don't think the English actually needs to use the %n$s form, because it has the luxury of requiring that the call present the args in the right order.  However, if you look at existing practice, almost the only place where you see %s is in PRINTDB() calls; LOG() calls all use %n$s form.
```
$ grep %s src/core/* | wc -l
grep: src/core/customizer: Is a directory
21

$ grep %s src/core/* | grep -v PRINTDB | wc -l
grep: src/core/customizer: Is a directory
3

$ grep '%1$s' src/core/* | wc -l
grep: src/core/customizer: Is a directory
102
```

kintel:
@jordanbrown0 I believe you can just submit another review undoing the approval, if you still have outstanding questions

src/core/builtin_functions.cc
jordanbrown0: Anyhow, it works.

jordanbrown0:
There are a couple of things that I'd change (like the %s vs %n$s) but don't seem harmful.

jordanbrown0:
Net, go for it.

pkriens:
Thanks!

jordanbrown0:
Woo hoo!

pkriens:
> Woo hoo!

You wanna announce it on the list? You did the majority of work.

jordanbrown0:
Actually, Revar did the work on object().  But (with credit) I don't think he'll have any problem with me announcing it.

amatulic:
Can anybody point to where this feature is documented? How to use, examples, and so on? Is there anything about it on the wiki, like in a WIP page where textmetrics() is currently documented? There are a lot of older PRs related to this, some closed, some not. Is it basically Revar's first comment in #4337 ?

jordanbrown0:
I believe the most comprehensive writeup may be OEP8a1: https://github.com/openscad/openscad/wiki/OEP8a1:--The-object()-function .

I believe there is some stuff in the wikibooks manual, but I am not sure that it is accurate yet.

amatulic:
I found it in the wiki, and it seems comprehensive enough: https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/Objects

~~What seems to be missing is an `is_object()` type test, distinguished from `is_list()`. I guess one could hack it using something like `is_list(obj) && is_list(obj[0]) && is_string(obj[0][0])` or something similar assuming `is_list` returns true for an object (I haven't tested it).~~
Nevermind, I found it. https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/Type_Test_Functions#is_object

jordanbrown0:
Yeah, is_object() has been in for a couple of years.

PR #6022, opened 2025-07-14
Provide $this and automatic context for functions in objects

pkriens:
## Current Discussion / ToDo

- [ ] Decide on name of the reference, concern with `$this` is that it looks like a dynamically scoped variable which are already not so easy to explain in comparison to normal variables. Using just `this` the name would still some magic, but it would not point to some specific additional behavior.
    - [ ] `$this`
    - [ ] `this` as identifier
    - [ ] `this` as keyword
    - [ ] ...

## Problem
Using the new `object()` function I've tried to use it with _functions_ as values. This works fine but I found that any useful scenario ran into the problem that the function has no access to the object itself. First attempt to address this was to use functions and use their scope:
```
function Rect(w,h) = object( w=w, h=h, area=function() w * h );
r1 = Rect(10,10);
echo(r1.area()); // 100
```
This works but it falls apart when you set the `w` or `h` member to something else. 

```
r2 = object(r1, w=20);
echo(r2.area()); // 100 (wrong!)
```
The boilerplate to copy the parameters in the object is also annoying.

## Solution: $this & use object as scope

In [OEP8][1], future directions already discussed a `$this` reference.  When a member function gets executed, its context will have a magic variable `$this` that is the object where the function came from. This is dynamic so that if a function is copied to another object, its `$this` will come from the new object.
```
r1 = object( w=10, h=10, area = function()  $this.w * $this.h );
echo(r1.area()); // 100
r2 = object(r1, w=50);
echo(r2.area()); // 500 YES!
```
However ... the `$this` is in itself awkward to use. So as a proposal I decided to automatically add the members to the context as well. This worked extremely well.
```
r1 = object( w=10, h=10, area=function() w * h );
echo(r1.area()); // echo 100
r2 = object(r1, w=50);
echo(r2.area()); // echo 500
```

### Issues

* Copying the receiver values in the function's scope shadows the variables in a parent scope and they become inaccessible. This is similar to how OO languages handle their members but should be discussed because I do not see a way these shadowed globals can be accessed.
* When a function is copied from an object to a normal variable or function parameter, it will keep a reference to the object. If that function is executed from the variable, it will access its receiver. 
* There is no way to get the (optional) receiver from a function.
* I convinced myself that this implementation could not provide mutable state but this could require some smarter minds than me.

## Implementation

Added a field to `std::shared_ptr<std::vector<Value>> receiver`. The receiver is normally empty and then has no effect. However, if a `MemberLookup` is done (like `bar.f`) and the member is a function, the function is copied and it's receiver is set to the object it came from.

When a `Function` object is executed, we look if its `receiver` is not empty. If it isn't we set `$this` in the context before the parameters are set. (A function parameter $this will thus override the automatic `$this`.).

The `receiver` field is explicitly a `public: vector<Value>` and not a private `boost::opt` or `shared_ptr<Value>` because we need a pointer to a `vector<Value>` in the `ContextMemoryManager` to ensure that an optional receiver is garbage collected.

[1]: https://github.com/openscad/openscad/wiki/OEP8a:--Objects-(dictionaries%3F)

### DIFF
--- a/src/core/FunctionType.h
+++ b/src/core/FunctionType.h
@@ -3,9 +3,9 @@
 #include <utility>
 #include <memory>
 #include <ostream>
+#include <boost/optional.hpp>
 
 #include "core/Assignment.h"
-#include <boost/optional.hpp>
 
 class Context;
 class Expression;
--- a/src/core/Parameters.cc
+++ b/src/core/Parameters.cc
@@ -26,8 +26,6 @@
 
 #include "core/Parameters.h"
 
-#include "Context.h"
-
 #include <initializer_list>
 #include <cassert>
 #include <sstream>
@@ -39,6 +37,7 @@
 #include <vector>
 
 #include "core/AST.h"
+#include "core/Context.h"
 #include "core/EvaluationSession.h"
 #include "core/Expression.h"
 #include "utils/printutils.h"
@@ -257,7 +256,13 @@ Parameters Parameters::parse(Arguments arguments, const Location& loc,
 
   for (const auto& parameter : required_parameters) {
     // see builtin_functions.cc::builtin_object() for an explanation
-    if (parameter->getName() == "this" && defining_context->try_lookup_variable("this")) continue;
+    if (parameter->getName() == THIS_PARAMETER) {
+      auto const it = defining_context->lookup_local_variable(THIS_CONTEXT);
+      if (it) {
+        frame.set_variable(THIS_PARAMETER, it->clone());
+        continue;
+      }
+    }
 
     if (!frame.lookup_local_variable(parameter->getName())) {
       if (parameter->getExpr()) {
--- a/src/core/builtin_functions.cc
+++ b/src/core/builtin_functions.cc
@@ -563,27 +563,26 @@ Value builtin_object(const std::shared_ptr<const Context>& context, const Functi
     n++;
   }
   std::vector<Context *> contexts;
-  for (int i = 0; i < result.ptr->values.size(); i++) {
-    const Value& value = result.ptr->values[i];
+  for (Value& value : result.ptr->values) {
     if (value.type() == Value::Type::FUNCTION) {
       auto& function = value.toFunction();
-      auto it = function.findAssignmentByName("this");
+      auto it = function.findAssignmentByName(Parameters::THIS_PARAMETER);
       if (!it) continue;
 
       auto parent = function.getContext();
-      if (parent->lookup_local_variable("this")) {
+      if (parent->lookup_local_variable(Parameters::THIS_CONTEXT)) {
         // function is already a method, so take the parent context
         parent = parent->getParent();
       }
       ContextHandle<Context> ctx = Context::create<Context>(parent);
       contexts.push_back(ctx.operator->());
       Value method(FunctionType(*ctx, function.getExpr(), function.getParameters()));
-      result.ptr->values[i] = std::move(method);
+      value = std::move(method);
     }
   }
   Value object(result);
   for (Context *c : contexts) {
-    c->set_variable("this", object.clone());
+    c->set_variable(Parameters::THIS_CONTEXT, object.clone());
   }
   return std::move(object);
 }
--- a/tests/data/scad/experimental/object/object-tests.scad
+++ b/tests/data/scad/experimental/object/object-tests.scad
@@ -134,7 +134,17 @@ testEq( object( $fs = 42 ).$fs, 42, "using config names as field is ok");
 testEq( object( this = 42 ).this, 42, "also this");
 testEq( object( this = 42, f=function(this) this.this ).f(), 42, "but functions see the this object");
 
-
+{
+    this = 42;
+    fthis = function(this) this;
+    test( function() is_undef(fthis()), "global 'this' must not leak in functions when not a method");
+    test( function() !is_undef(object(f=fthis).f()), "must work as method");
+}
+{
+    p=object( f = function(child,this) child());
+    c=function(this) this;
+    test( function() is_undef(p.f( c )),"leaks from nested invocation" );
+}
 
 module test( f, s ) {
     if ( !f()) {
--- a/src/core/Parameters.h
+++ b/src/core/Parameters.h
@@ -66,6 +66,9 @@ public:
   const std::string& documentRoot() const { return frame.documentRoot(); }
   const Location& location() const { return loc; }
 
+  static constexpr auto THIS_PARAMETER = "this";
+  static constexpr auto THIS_CONTEXT = "#THIS";
+
 private:
   Location loc;
   ContextFrame frame;
--- a/tests/regression/echo/object-tests-expected.echo
+++ b/tests/regression/echo/object-tests-expected.echo
@@ -42,7 +42,6 @@ ECHO: "PASS:", "test 100k entries"
 ECHO: Wilma = { spouse = { }; }
 ECHO: Betty = { friends = [{ spouse = { }; }]; }
 ECHO: "PASS:", "optin gets this"
-WARNING: Ignoring unknown variable "this" in file object-tests.scad, line 99
 ECHO: "PASS:", "optout does not get this"
 ECHO: "PASS:", "check referring to other field"
 ECHO: "PASS:", "check referring from copied field"
@@ -57,3 +56,6 @@ ECHO: "PASS:", "recursively calling a function in an object"
 ECHO: "PASS:", "using config names as field is ok"
 ECHO: "PASS:", "also this"
 ECHO: "PASS:", "but functions see the this object"
+ECHO: "PASS:", "global 'this' must not leak in functions when not a method"
+ECHO: "PASS:", "must work as method"
+ECHO: "PASS:", "leaks from nested invocation"


pkriens:
@jordanbrown0 @kintel @reverbat Would love to have feedback on this.

jordanbrown0:
I think there'd be a lot of developer resistance to adding to Value itself; that multiplies up in memory cost pretty rapidly.  (Did you have to change the `static_assert()` that limits the size to 24 bytes?)  I don't know how big a shared_ptr is, but any addition multiplies up when you have a vector of a million doubles.

With respect to including the object's members in the context search... seems plausible.  My recent work has primarily been in languages where you must explicitly refer to the object (JavaScript, Python), and so requiring `$this.x` doesn't bother me, but I see the interest.  Not that I think it would be important, but note that explicit `$this` would let it work with function references that come out of arrays too:  `a = [ function () $this[1], 456 ]; echo(a[0]());` would yield 456.

Shadowing is an issue, but I think only a minor one.  If you need to preserve a value from the context, copy it into the object with a different name:
```
function newobj(x) = object(arg_x = x, x=123, f=function() x + arg_x);
```
or even into one of the contexts:
```
function newobj(x) = let(arg_x  = x) object(x=123, f=function() x + arg_x);
function newobj(x) = object(x=123, f=let(arg_x  = x) function() x + arg_x);
```
OpenSCAD already has shadowing; a local variable can shadow the function or module's argument list.

Keeping a reference to the object through a copy to a normal variable, or through any other sort of copy, seems ... confusing.  Theoretically, we could have any assignment clear the reference.  That's a small cost, but we'd have to do it on every assignment.

I started a couple of times to implement methods, and that was the stumbling block I kept running into:  where to keep the "this" reference between the MemberLookup and the subsequent function call.  Hmm.  I'd have to double-check the data structures, but unlike most languages where the works is largely done in RPN / postfix order, OpenSCAD does it in prefix order.  (That is, most language implementations would have A+B be "load A, load B, plus", but OpenSCAD has more like "plus(exprA, exprB)" and plus() is responsible for evaluating the expressions.)  This means that the "function call" operation could pass down something saying "hey, if you're a MemberLookup, tell me the container".  It could even peek down the tree to see if it's a MemberLookup, but I don't think that would work well because it would need to evaluate the object-expression to get the object, and you wouldn't want that evaluated twice.  Hmm.  It could specially recognize MemberLookup, and either bypass it and do the lookup on its own, or do the object evaluation and the pass the evaluated object to a variant on MemberLookup::evaluate().  Ugly, but maybe not too ugly.  Of course, it would hurt people's heads if they ever move OpenSCAD to a more conventional postfix evaluation engine.

The interaction with ContextMemoryManager is important.  This would be another reason to want a type-independent way to iterate over contained values, so that this could use an iterator that just returns the one and only child.

pkriens:
> I think there'd be a lot of developer resistance to adding to Value itself; that multiplies up in memory cost pretty rapidly. (Did you have to change the static_assert() that limits the size to 24 bytes?) I don't know how big a shared_ptr is, but any addition multiplies up when you have a vector of a million doubles.

A `FunctionType` is stored as a `FunctionPtr` in `Variant`, this mechanism is also used for `RangePtr`, which extend the generic mechanism `ValuePtr<T>`. `ValuePtr<T>` objects have the size of a `shared_ptr<T>` which is 16. So the overhead is only for the FunctionType objects, which is always going to be very low since they require source code.

It does beg the question why we do not use the `ValuePtr<ObjectType>` mechanism? It would clean up the code and remove an unnecessary indirection? And gets rid of the distinction between `ObjectObject`, `ObjectObjectDeleter` and `ObjectType`. 

Shall I make a PR to change this?

> Keeping a reference to the object through a copy to a normal variable, or through any other sort of copy, seems ... confusing. Theoretically, we could have any assignment clear the reference. That's a small cost, but we'd have to do it on every assignment.

I think I can optimize the overhead to be purely in the MemberLookup, FunctionCall case once. However, I wonder if it is worth the effort. It is not a feature that I'd design for but it can actually be convenient, it is close to Java lambdas where a MethodHandle can have a receiver object, like the .NET Delegate.

> I started a couple of times to implement methods, and that was the stumbling block I kept running into: where to keep the "this" reference between the MemberLookup and the subsequent function call.  ...

I had several false starts. I created a new type MethodType, subclass of FunctionType but it required too many changes. I tried having a static variable but coming from concurrent software that has a threshold that is a bit too high for me and you can always come up with convoluted scenarios where it doesn't work. I looked at AST construction time, since you can detect the MemberLook and FunctionCall you could do something very clever and fast but I don't think it is worth the effort. The FunctionCall is a very complex piece of code.

Since it doesn't increase the Variant, it is the best way I could come up with.

Thanks for the quick review!

jordanbrown0:
I am not sure, but I think that when you have a Value containing a FunctionType, and you copy it to another variable, you get a new Value but share the same FunctionType.  But maybe that's OK, because you only need to keep the "this" value for a very short time, from the MemberLookup to the parent function call.

I don't understand the relationship between the various types well enough to have an opinion on the class/object relationships.  Object copied Vector.

pkriens:
I spent a lot of time cursing the ContextMemoryManager ... basically everything I tried ran either into a problem that FunctionType has no access to Value.h (cyclic dependency) or whatever, I even started rewriting it. The issue with ObjectType <-> FunctionType was a huge PITA.

Then yesterday it hit me after struggling with ContextMemoryManager, the Context is the core object there. The Context is a map, why not create the map with the context there? A FunctionType takes a context and it is easy to insert a context with the object values. 

This then worked like a charm. I can now do all the work in object() ... meaning there is zero overhead in the function invocation. It also turned out that I did not need to touch ContextMemoryManager! Talk about a happy ending :-)

The only thing a bit awkward is that I need to exclude `$this` from the config variables in Context ... Open for suggestions there. 

The setting of the context is done in two passes. Since I change the FunctionType, I must make sure that what ends up in the Context is actually the updated function and not a previous value. Proud to say that I thought of this ahead of time. Second tripping wire is to make sure that we're not adding contexts to function that were already methods.

I add test cases and unless I get very negative feedback, I think it can move out of draft status. With this method I did not have a single crash in ContextMemoryManager ...

jordanbrown0:
I can't say that I really understand CMM, but the intent was that objects were the same as vectors for this purpose.

pkriens:
I am on the same level about CMM. It is a kind of mark and sweep GC but unfortunately the algorithms is highly integrated with the Value and Context classes. Key problem was that it can't handle `value <-> value` dependencies because it never checks its direct children. We never had the case that a value A contained a value that referenced A directly. E.g. a Vector that had a member that referenced itself is impossible.

In the previous attempt, creating this cycle felt hackish, usually indicating there is an issue.

Going through the Context very breaks this issue, now it becomes a common cycle that the CMM handles well. Huge advantage is that I now only need to touch the builtin_object function, no API changes. (Except a tiny change in Context to allow "$this" to come from the variables and not the session).

t-paul:
Why does it need to be `$this` instead of just `this` or `self`? E.g. in python it's even manually declared in the function arguments and has normal static scoping.

jordanbrown0:
I was thinking that it was special and system-y, and so was thinking of special variables.  The scope is kind of wrong, because things it calls would get the same $this, but that didn't seem catastrophic.

I don't think using a normal variable (but magically set) would really work...
```
method = function () this.a;
o = object(meth=method, a=123);
echo(o.meth());
```
But the "this" for method would be scoped to the top level, and it needs to be scoped inside the object.

It could be really magic, but that seems undesirable.

Python's scheme of treating it as a somewhat-visible parameter is clean from a namespace perspective, but seems like a hack if you're used to Java or C++ or JavaScript where it's magic.

pkriens:
> Why does it need to be $this instead of just this or self?

Slight preference for `$this`. The `$` gives it a bit of status that is _special_ (which it is) and it was part of the original discussion. And it does make it easier to exclude it from the names in the object. But no strong feelings either way.

t-paul:
Hmm, special yes, but also magic and different from all the other special variables which are already hard to explain at times. Are there any languages that use something else we could look at instead? Crazy idea... keep only the `$`? Seems a bit too Perl style though :grin:.

pkriens:
Hmm, just the $ would be fine with me ... another candidate would be to use _. 

But don't you think `$this` is quite predictable? The `this` is used in JavaScript, C++, C#, Swift, Kotlin, TypeScript, Dart, Scala, and Java and is well known for what it does. PHP even calls it `$this`! Python, Objective C, Rust call it `self`.
In OpenSCAD I'd expect it to be called `$this` due the conventions.

In the greater scheme of things I am not sure it matters a lot?

jordanbrown0:
I believe that C++, Java, and JavaScript all have a keyword `this` with these semantics.  (I think it's a keyword rather than a magic variable, but the difference is subtle.)

Python treats it as a semi-visible first argument to a method.  The caller doesn't specify it, but the method receives it as its first argument.  The name is conventionally `self` but is controlled by the method; it can use whatever name it likes.

Making it a keyword would reduce or avoid scope questions, but I think it would be the first instance of a keyword that looks like a variable.  (`PI` is, I believe, a variable defined above the top scope.)

jordanbrown0:
> ... with these semantics

Well, for JS, yes; it has the same "calling a function that is a member of an object" semantic.
For C++, Java, and Python there's more formality around a function being a class method.

t-paul:
I don't see an issue with `this` only with `$this` because the `$` otherwise means it's a **dynamically scoped** variable. PHP calls it `$this` because *all* it's variables use the `$`, which is not the case in OpenSCAD.

OpenSCAD has both statically *and* dynamically scoped variables where the latter are indicated by the `$`. Now having extra magic and again using the `$` for that makes things even more difficult to explain. Explaining a partially magic `this` is probably much easier.

pkriens:
Let's go for that then? Shall I make a new PR?

t-paul:
I'd say lets give it a bit workout first. For testing out some things the specific naming does not matter too much yet. I'll put the current state of discussion in the first post and we can update the code once it gets a thumbs up overall. I think @kintel should get a chance to look at this too.

pkriens:
So you will merge this or you want to wait for @kintel?

jordanbrown0:
Neither a special variable nor a normal variable really has the right scope.  A special variable has the wrong scope looking "down" - that is, things called by the method would have the method's $this in scope when they aren't logically part of the method.  A normal variable has the wrong scope looking "up" - the function may well have been defined in a context where there is nothing object-related above it to have lexically provided the variable.

Choosing between those two, I'm a little bit happier with a special variable, because it seems less wrong.  In particular, trying to use a normal variable means that you can't always tell whether a normal variable is defined in your scope.  Before this, you could statically analyze the program and determine all of the normal variables that are in scope at any point in the program.  With this, you wouldn't be able to tell whether a particular instance of "this" will have been defined.  Similarly, I think it's currently possible to upgrade our symbol-table processing to statically look up all normal variables, and with this you wouldn't.

Special variables, however, are not amenable to such static analysis, and $this would be nothing new there.

Having `this` be a keyword, shrug.  It wouldn't need to be very *much* of a keyword.  A static analyzer or symbol table improvement would need to recognize it, but otherwise it could behave like a variable.

t-paul:
I'm looking at the automatic object scope and shadowing and it does seem a bit problematic. While it's nice and short to type the references from the object scope, it does get even more confusing when defining the function outside of the object.

```
w = 10;
f = function() 3 * w;

o = object(w = 20, f=f);

echo(f = f(), o = o.f()); // ECHO: f = 30, o = 60
```

I wonder if handling things a bit similar to python would be better, meaning inject the object reference as first positional argument to the function call. It could not allow the shortcut notation but prevent the shadowing and making things a bit more specific.

The shortcut notation would also interfere with any optimization of resolving statically scoped variables before the evaluation as discussed in some other issue (like shown in https://craftinginterpreters.com/resolving-and-binding.html).

```
w = 10;
f = function(self) echo(w = self.w) 3 * w;

o = object(w = 20, f=f);

echo(f = f(), o = o.f());

// ECHO: w = undef
// ECHO: w = 20
// ECHO: f = 30, o = 30

echo(f = f(o));
// ECHO: w = 20
// ECHO: f = 30
```

jordanbrown0:
When and if we incorporate object-literal notation, a function defined inside the object could be statically associated with the members of the object, in one way or another.  Doesn't help a function defined outside the object, or with `object()`.

I don't really *like* the Python style, because I learned OO in Java and JavaScript where it's not necessary to declare "this", but I agree that it makes the scoping clearer.  (And I admit I have occasionally been reading  a Java function, where instance members are automatically in scope, and said "where the heck was **that** defined?", and having it explicitly be `this.` would be a big clue.)

pkriens:
I really dislike the idea of argument injection for several reasons. It just adds boiler plate code for imho no reason. If it is mandatory to declare for object functions, we create a whole slew of error cases to handle. It also makes it impossible to use existing functions. 
Last but not least, it will require work at the function invocation which I worked so hard to avoid ... I noted that function invocation is high in the top culprits when I ran a profiler. 

The scoping problem is real but this kind of scoping issues are extremely common in most languages. C++, Java, C# all have a default this. Since the scoping is fully defined in the object() invocation I'd prefer a warning when a field overrides a variable higher in scope for a function. Exactly the same problem happens if you're in a module and a parameter or local variable has the same name.

One of the crucial things in software for me is the mental overload, reading code is really hard. Many theories show that we can handle 3-5 'things' in our brain simultaneously. we recognize 456-555-1234 as a US telephone number without parsing the 3 components  of 3-4 digits and can relatively easy remember it because the pattern is recognized by our hardware. When you add 'cruft' to the syntax you might theoretically gain something on a rare problem but you lose the common case that the algorithm in the function becomes harder to be grasped by our brains because it needs to filter the cruft. XML is the most horrible offender here.  However, it is one of the reasons I never liked Python that you always have to use self. I was working with the brother of Guido van Rossum in the Netherlands in the '80s. Then we needed explicit 'self' or 'this' (I created CO2, an OO env in TurboC) because we had to use existing environments. Using the first parameter really was a hack to bolt OO on an existing language. All modern languages have an implicit this/self as far as I know and I think there are sound reasons for this.

Its been my quest in my active software life so forgive me if I sound zealous ... :-(

t-paul:
> C++, Java, C# all have a default this.

True, BUT: they also have the concept of classes which we don't have. In those languages you can only define member functions *in the object itself*. Even with partial classes in C#, you still have to give the specific class.

> Since the scoping is fully defined in the object() invocation I'd prefer a warning when a field overrides a variable higher in scope for a function.

I think that would be ok-ish if we are just talking about single file scripts. Considering libraries, it gets very problematic. If you have a complex object created inside a library and you are supposed to pass in functions to customize some behavior, you then have to inspect *all* the variable names in your function. Now lets imagine the customizing functions also come from a different library. In that case it would only be possible to solve the issue by renaming variables which is not good.

>  Last but not least, it will require work at the function invocation which I worked so hard to avoid ... I noted that function invocation is high in the top culprits when I ran a profiler.

I agree that would be great to keep, but discussing the logic and behavior based on an initial optimization seems a bit backward. If we can solve the open concerns somehow, I happily take the short variable names instead of an always needed self variable as I don't like that in python at all.

Even with the introduction of objects, I don't think OpenSCAD has enough properties to qualify as OO language and I'm not sure it even has to go there. It's main target is object description, not to be a general purpose programming language.

pkriens:
> Even with the introduction of objects, I don't think OpenSCAD has enough properties to qualify as OO language and I'm not sure it even has to go there. It's main target is object description, not to be a general purpose programming language.

Fully agree. I am not pushing OO because I like OO, I got my favorite languages and don't need another one for general work. However, I like the simplicity of OpenSCAD for its target application: 3D modeling. The main reason I am advocating for these features is that it allows me to _scope the global name space_.  I use BOSL2 and one of the problems is that all names are on the same level shared with all other libraries I use. This forces people to use unnatural and long names and this makes it hard to remember the names. A hierarchy that will be possible with these objects will make this a lot easier.

Having these objects will reduce the global scope significantly and thus conflicts. However, this will require that methods in objects do not look more complicated or very different from normal functions or people will not use them. 

> Considering libraries, it gets very problematic. If you have a complex object created inside a library and you are supposed to pass in functions to customize some behavior, you then have to inspect all the variable names in your function. Now lets imagine the customizing functions also come from a different library. In that case it would only be possible to solve the issue by renaming variables which is not good.

Agree, this would be very bad. A library that uses this model of callback functions should clearly not store the callback function in its own object, that would break all modularity. A library should wrap functions to callback in a list, function, or object to prevent them from becoming bound to the object.
```
scope = "global";
f = function() scope;
mylib = object(
    scope    = "mylib",
    f1 = f,
    f2 = [f],
    f3 = function() f(),
    f4 = object( f = f),
);
echo( val1=mylib.f1(), val2=mylib.f2[0](), val3 = mylib.f3(), val4 = mylib.f4.f() );
// ECHO: val1 = "mylib", val2 = "global", val3 = "global", val4 = "global"
```

jordanbrown0:
Q.  Is JavaScript OO?  Why or why not?

There are, I think, two and a half separable questions:
* Should a function called via an object implicitly get the object's members in its scope, or should it have to explicitly use a "this" or "self" object of some sort?
  * If it's explicit, should it be an explicit argument to the function?
* When a function calls a callback function that was provided to it, what object should it be in the context of?

We're talked a bit about the first two, but not the last one.

I agree that that last one must *not* be in the context of some library-internal object if you implicitly put an object's members in scope; that would be a namespace disaster.

In JS, the default seems to be that a callback function is called without any "this".  For some (but not all?) functions that take a callback function argument, it's possible to supply an additional argument that says what value to use for "this" when calling that callback function.  Ref, e.g., [Array.forEach()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach).  It would be easy enough to invoke callback functions without a "this":
```
let(f = this.callback) f();
```
would presumably protect against inappropriately setting the "this" context for the callback function.  And if the object association is "sticky" and attaches to f (which I would discourage), you could put it into its own temporary object with no other members.  (But there would still be one name that would need to be reserved.)

Note that JS has a couple of functions intended for calling functions with a specified "this" context, so that you can make it behave like it's called as an object method.  We could add those, but I'd really rather not.

pkriens:
I'd suggest we do make functions sticky for $this, exactly for the callback reason! It would be very unpredictable if you could not pass a method as a callback for exactly the reasons Torsten gave. 

Ecmascript very specifically introduced the `=>` operator for this purpose.

jordanbrown0:
`=>` is about the behavior of an anonymous function, and where its `this` comes from.  It isn't about whether retrieving a function from an object via `o.func` binds `o` to the reference.

I think `=>` creates an anonymous function that throws away the `this` that it would normally get, and instead uses the `this` from its own context, so that `this` behaves like a normal variable in that context.

But indeed there's a lot of stuff in this area that we need to think about.  I really hope that we don't need to create a separate function-literal syntax like JS did.  But you really do want both kinds of function reference:  ones that get their `this` from their creation context, and those that get `this` from their invocation context.

jordanbrown0:
> The main reason I am advocating for these features is that it allows me to scope the global name space.

I wouldn't put it quite that way - I would say that it lets us get library functions, modules, and variables *out* of the global namespace.  But yes, exactly.

And the same is true for some builtin functions (though maybe not modules).  If you have the data types have methods, you don't need to put all of those functions into the global namespace.  You can use `"foo".length()`, `["a","b","c"].join(",")`, `"a,b,c".split(",")`, et cetera.

pkriens:
To prevent binding to the object as context, you can wrap the function in another function/list, which seems fine to me. I thought we're not building a cool OO language here? So I think there is a chance to make the better the enemy of the good?

Anyway, analyzing the concerns I see two avenues:

a) make the binding to the object _explicit_, or
b) provide a way to prevent the binding from happening, _seal_ the function.

The problem with a) is that the only implementation I see is to mark the function as bindable with for example a new keyword or operator:
```
f = function() a;
o = object( a=42, f=method( function() a) );
o = object( a=42, f=method() a );
o = object( a=42, f=method( f );
o = object( a=42, f=method f ;
// or being bold
o = object( a=42, f => function() a) );
```
Issues
– a new keyword/operator
– since we use a standard builtin_function, it has to work for every expression but outside object it would have no meaning.


For b) we could provide a builtin function for _sealing_, a function that prohibits rebinding.
```
f = function() a;
o = object( a=42, f=f.seal() );
o = object( a=42, f=(function() a).seal() );
o = object( a=42, f=seal(f) );
o = object( a=42, f=seal(function() a) );
```
The `seal()` function would prevent binding the function to the object.

Issues
– a new function that will not be easy to explain to many
– can also be easily implemented by wrapping the function in a list or proxy function.

jordanbrown0:
This topic hurts my head at the best of times, and right now (in a lounge where I have cell service) is not the best of times.

However, I do have a few quick comments.

> method()

In addition to the undesirability of adding a new keyword, we will hopefully eventually have module references, and then we'd need to distinguish between function-methods and module-methods.

> o = object( a=42, f => function() a) );

I would *not* add magic syntax to `object()`, and I don't know what `=>` like this would mean in other contexts.

But maybe a JS-like alternative way to give a function literal:
```
o = object( a=42, f = () => a );
```
But:  I **really** dislike that syntax, because it is not at all self-explanatory.  Also, again, we need to consider module references.

I will try to spend some time offline organizing use cases and questions so that I can think about this in a more structured way.

pkriens:
@jordanbrown0 aren't you making it too complicated? We're not designing a super-duper OO language, nor is there any need to, there are lots of good languages out there. 

The issue is: is the _default_ binding?

So instead of discussing syntax, why not first define what the default should be? To bind or not to bind?

jordanbrown0:
> @jordanbrown0 aren't you making it too complicated? [...] So instead of discussing syntax, why not first define what the default should be?

Er, I must be confused.  Who was it that was suggesting the syntax alternatives that I was responding to? :-)

So it seems to me that we have about four use cases.  Let's look at them and their requirements.

Ignoring the this/$this/implicit choice...

1.  We have an `object()` call that creates an object that has a method and an instance variable, and the method accesses the instance variable.
```
o1 = object(method=function() echo(this.that) undef, that=123);
```
I hope that we can agree that this case is one of the most common, and that it should be as simple and obvious as possible.

2.  We have an `object()` call that creates an object starting with a clone of another object.  (This is how you would do inheritance.)
```
o2 = object(o1, that=321);
```
This case has the *potential* for being common and so needing to be obvious.  `o2.method()` needs to get a `this` that points at `o2`, not at `o1`.

3.  We have an `object()` call that creates an object using a function defined previously, perhaps retrieved from an object.
```
o3 = object(method=o1.method, that=456);
```
Since I don't think we should have object()  have any special semantics as to how its arguments are processed, I think this needs to be almost the same as case 1, but with the same note as case 2:  `o3.method()` needs to get a `this` that points at `o3`.

4.  We have a factory function that accepts a callback function and returns an object, and the callback function needs to **not** have a `this` that refers to that object.
```
// library
function libfactory(cb) = object(
    callback=cb,
    method=function() this.callback()
);
// user
userobj = object(
    libobj = libfactory(cb=function() this.that),
    method = function() this.libobj.method(),
    that = 654
);
```
(Or something like that; the details are one of the questions.)

When you call `userobj.method()`, and it calls `userobj.libobj.method()`, and that calls `userobj`'s callback function, `this` must either be `userobj` or `undef`, and must **not** be `libobj`.

Once we've handled cases 1-3, probably so that the exact snippets I quoted above would work "correctly", I don't think that the snippet I quoted for case 4 will work.  Because `libobj.method()` calls `this.callback()`, it seems like it **must** get a `this` that points at `libobj`, violating our requirement.

With respect to this requirement, I think that the `userobj` above should not have to dance to **avoid** getting a `this` that points at `libobj`.  Addressing that requirement should fall solely on `libobj`; it should not pollute `userobj`'s context.  (Whether it will **work** is another question; see below.)

So what is the idiom that `libobj` must use to call the callback function, **without** setting its `this`?

Maybe copy the callback function reference out of the object before calling it:
```
// library
function libfactory(cb) = object(
    callback=cb,
    method=function() let(f=this.callback) f()
);
```
But if the callback function wants a reference to `userobj`, that's harder.

I don't immediately see how `libobj` can achieve that requirement on its own.  Alone, it simply doesn't have a reference to `userobj` to pass down to the callback function, even if we gave it a way to.

Old-school JS has two ways to achieve this goal:
* `userobj` would use the callback function's context (probably the constructor's context) to store a copy of `this`, and the callback function would use that copy instead of `this`.  In JS that looks like so:
```
function userConstructor() {
    var userThis = this;
    this.libobj = libfactory(function () {
        return userThis.that;
    });
}
constructor.prototype.userMethod = function () {
    this.libobj.libMethod();
};
```
* `libfactory`, or perhaps `libobj.method()`, would accept an argument saying what object to use as `this` for the callback function.  This is what, e.g., `Array.forEach()` uses.  The library method then uses an indirect call mechanism - IIRC, `Object.call()` - to call the function and supply `this`.  It looks something like so:
```
// library
function libConstructor(cb) {
    this.callback=cb;
);
libConstructor.prototoype.libMethod = function(co) {
        Object.call(co, this.callback /* , args */)
}
// user
function userConstructor() {
    this.that = 789;
};
userConstructor.prototype.method = function () {
    this.libobj.libMethod(this);
}
```

Neither of those is particularly appealing.

The first (use the callback function's context) seems like it would run into trouble when you want to clone `userobj` to make a subclass-y thing, because we don't have (and don't plan to have) constructors and JS's prototype chain.  When you clone the object, the function would stiff have the original factory's context, that would point to the original object.

The second requires a new indirect-call mechanism, and I'd rather avoid that.

New-school JS has a replacement for `function()` that retains its context's `this`, something like:
```
function userConstructor() {
    this.libobj = libfactory(() => {
        return this.that;
    });
}
```
(For work-legacy reasons I don't have a lot of experience with new-school JS; if I got the syntax slightly wrong I hope you'll understand anyway.)

I would really rather not have to introduce a new syntax for this case, and especially not **that** syntax.

---

WRT "not designing a super-duper OO language", I agree... but where do we draw the line?  Many people seem to want at least what we've currently got in the way of objects (which one might call data-only objects, or something like that).  But that immediately leads them to wanting `this`, and the rabbit hole gets deeper.

And something to think about as we dig the rabbit hole deeper... if you do inheritance by cloning an existing object and modifying it, how do you call the original object's methods when you've overridden them?  Do you need a `super`?

coryrc:
You've got some goals and are trying to make *object* work for them. I think the goals are:

1. Replace hardcoded [1] and [CONSTANT] index into arrays by being able to name values by identifier
2. Be able to namespace functions, modules, and constants.
3. Be able to assert some piece of data as conforming to some standard (i.e. data validation in constructor, expect method X to exist, etc).

Is that correct?

coryrc:
I don't think prototype inheritance is a great fit for a language with immutable data types. I think either generic functions or class definitions are a better fit. Just because you *can* use *object*+lambdas to keep many things you want doesn't mean you *should*.

coryrc:
Copying from the discussion list:

I don't think OpenSCAD should have complex OO, which is what worries me about the `this` discussion and all the complexities of scoping and such you have brought up. Here's what I've been thinking about:

In the context of the 3 problems I stated (maps/structs, namespacing, data compartmentalization). Add namespaces. Define maps/structs/records/whatever-you-call-it as only containing Values (ints, lists, lambdas, other records, etc) and having one parent. Both of these things can be implemented simply (name mangling for namespaces, records can be a hash map + parent lookup on miss). Both of these have immediate benefits, both are easy to explain to people. They both have established ways of being optimized for speed.

That just leaves "how do I do OO?" We know from Java's history that long inheritance chains are bad and people rarely change out implementations and OO didn't make this any easier.
But then that isn't the real request. The real request is "How do I make it less likely to make mistakes and spend less time writing code to get my result?"

Go's philosophy is, in part, to notice people don't spend much time typing, so don't focus on minimizing key strokes, instead focus on readability and unit testability, even if that means you write a lot of redundant things over and over again.
Go has classes ("structs") which are allowed to have up to one parent and must be a concrete type.
It also has "interfaces" which is a list of method signatures and has its own methods. If your class has all the method signatures of the interface, you can cast it to the interface and call the interface's methods' on it.

The single-inheritance is a good choice. The only inheriting from concrete types is good. We learned from Java.

However, Go is very imperative. Further lessons are not helpful to a language with immutable values.

So, what do languages with immutable values usually have in common? Elixir/Erlang, Haskell, Ocaml, etc: Pattern matching.

And if you combine functional language, objects, and pattern matching, you get generic functions. The exact syntax details of the following are mostly not important.

```
namespace bosl2 { namespace threading { namespace MyClass {
  obj = struct(some_default=2);

  // Constructors can be just normal functions
  function new(something,some_default) =
    assert(!is_undef(something))
    let (
      args = [["something" something]],
      args = concat(args,is_undef(some_default) ? [] : [["some_default" some_default]])
    ) struct(obj, args);

  // Methods are functions and modules.
  function sum(o is obj) = o.something+o.some_default;
  module fat_cylinder(o is obj) { cylinder(d=o->sum()*2); }

  // Constructors can also be method functions
  function coerce(l is list) = new(l[0],l[1]);
} // namespace MyClass

// Modules/functions can be generic on anybody's class:
module nut(o is MyClass::obj, normal_arg) {
  if(normal_arg) {
    cylinder(d=MyClass::sum(o));
  } else {
    cylinder(d=o->sum()); // -> walks from object through its ancestors for the first matching method in each object's creation namespace
    o->fat_cylinder();
  }
}

module nut(o is list, a_different_arg_is_possible) {
  obj_o = MyClass::coerce(o);
  assert(!is_undef(obj_o));
  nut(o, normal_arg=a_different_arg_is_possible);
}
}}

// backward compatibility
module which_takes_complex_thing(existing, arg, list) {
  assert(on arg);
bosl2::threading::nut([existing, arg*2, list]);
}

```

Okay, so if you've got the basics of how this works, how does this actually help?

Let's consider the BOSL2 Threading library (surprise, I bet you wouldn't have guessed).

You have very long and repeated argument lists [<https://github.com/BelfrySCAD/BOSL2/blob/d252f781232a9262583806fe05bbb8529c108899/threading.scad#L384>](https://github.com/BelfrySCAD/BOSL2/blob/d252f781232a9262583806fe05bbb8529c108899/threading.scad#L384). And we could say, from Go, that's it's okay to type things repeatedly, but it's really easy to screw up. So how can we avoid this?

The builder pattern can help. In our hypothetical class hierarchy, threaded_nut_builder inherits from generic_threaded_nut_builder.

```
bosl2::threading::nut_builder::new(required, args, here)->optional_generic_arg(its_value)->reify();
```

From a user's standpoint, this used to be:

```
threaded_nut(required, args, here, optional_generic_arc=its_value);
```

From a user standpoint, it's hardly different. We might even have syntax sugar to simplify `::new` and `->reify()`.

But from a maintainer's standpoint, you no longer have to edit an ever-growing list of places for new arguments to `generic_threaded_nut` and `generic_threaded_rod`. threaded_nut_builder's `reify()` module only has the same dozen lines of computation before chucking it up to `generic_threaded_nut`. You have a private namespace for any logic you want to unit test. No semantics of the language are changed regarding types, handling missing args, etc. You can maintain 100% compatibility with the existing API while adding options/features to the new API. You can incrementally update the codebase a single module or function at a time, so it's easy to adopt. It's focused around answering the question "How do I make it less likely to make mistakes and spend less time writing code to get my result?" from real lessons.

pkriens:
@jordanbrown0 not sure what to say ... I took a lot of time to go through your text, but weren't all those cases no handled by my post already? 

Maybe I am too simplistic (people have accused me of that!) but isn't the question pretty basic? Sometimes we want to bind a function to its holding object (like in a method) and sometimes we don't (like in a callback). We clearly need both but the core issue is: what's the default?

And @coryrc: isnt this a completely different set of issues you're talking about? You propose new keywords (`namespace`) and syntax (`::`). This is a very limited PR that we're trying to close down, it has a long history being discussed in OEP8. (Of which I was not part but try to follow closely.) 
I think these issues you raise deserve a different issue and should not be raised on this PR.

nophead:
Why is it a problem to pass this to a callback? If it doesn't need it then it can ignore it but it might need access to it anyway. In my c experience I often had to store an argument with a callback pointer for exactly that reason.

pkriens:
@nophead the problem is that the callback may have its own this (or context) and you do not want a callback where the context is overridden by an unknown set of members. The members could then shadow access to your local information.
```
scope="global";

echo( object( f=function() scope, scope="local").f()); // ECHO: "local"

cb = object( cb=function() scope, scope = "callback").cb;
echo( object( f=cb, scope="local").f()); // ECHO: "local" !! unexpected

// proposal wrap
cb2 = object( cb=function() scope, scope = "callback").cb;
echo( object( f=function() cb2(), scope="local").f()); // ECHO: "callback"
```
Clearly you can make it part of the callback protocol to pass any object like `$this`.
```
cb2 = object( cb=function(caller) concat(caller.scope, scope), scope = "callback").cb;
echo( object( f=function() cb2($this), scope="local").f()); // ECHO: "localcallback"
```

Basically, we need to define if we want to bind the `$this` to a function when it is stored in an object automatically. 

– If we automatically bind: requires a feature to handle callbacks where we don't want binding: I propose wrapping in function or list.
– If we do not bind: need a specific syntax to bind a function to an object.

 It is a pretty straightforward problem but it involves syntax and that tend to be really hard in a large group.

jordanbrown0:
I was trying to walk through use cases, requirements, and previous art in a structured way.

jordanbrown0:
@nophead it's much more of a problem if we choose what you might call the "implicit `this`" option, where (like Java and C++, but not like JavaScript and Python) you can refer to instance variables without having to explicitly say "this".  Leaking the library's internal object would pollute the callback function's context.

pkriens:
Probably different style but I lose the context with so much text, sorry. And I went to each example but I think I discussed all of them already? 

The crux is that we need be able to call a function with and without the object's context. I.e. the _method_ and _callback_ cases. 

The key choice, and this is what raised by @t-paul is: what should the default be?

– If the default is the method case (as it is now) wrapping in a function can easily handle the callback case.
– If we make the default the callback case, we need some syntax for indicating the method case.

A secondary question is if the members should be in context when the function gets called. If it would expedite merging this PR then I am ok to stick with the `[$]this` only.

This is an experimental feature so I think it is important to get it out there and get wider feedback asap. We can always change it.

nophead:
Seems like an explicit this is what we should go with if both Python and JavaScript do that because OpenSCAD is script rather than program code.

jordanbrown0:
I think you did cover all of the bases, but that wasn't clear to me until I wrote down all of the use cases that I could think of in a list, and looked at how each would work.

> The crux is that we need be able to call a function with and without the object's context. I.e. the method and callback cases.

The callback case may want to have the function be called in the context of some particular object, for when higher-level function is itself a method call.  That is, if A.foo() calls B.bar() and passes in a callback that wants access to A.  (But absent that capability, I think the caller can work around it by storing the higher-level "this" in a variable that is then accessible to the callback function - one of the strategies used in old-school JavaScript.)

Combining two of the semantics that Cory suggested does neatly solve several of the problems, at the expense of what I consider a bit of ugliness:
* The method declaration explicitly specifies a "this" parameter as the first parameter.
* A call derived directly from an object - that is, `A.foo()`, implicitly adds that object as the first argument.
* A call not derived from an object, which would include `foo=A.foo; foo();`, would *not* implicitly add the object.  (It is not instantly clear to me whether `(A.foo)()` would implicitly add the object as an argument; my immediate thought is no.)

Advantages:
* The explicitly-declared "this" parameter has a very clear and normal scope, including being visible to callback functions that it defines.
* The two call cases allow for both the simple method call case and the case where you want to call an arbitrary function with an arbitrary object as its "this".  (But I'm not sure how important that latter point is, once you have the scope advantage described above.)

Costs:
* Need to explicitly declare `this` as a parameter.
* Need to be careful about how callback functions are defined (by the user caller) and called (by the library).  Do they get a "this", or not?

jordanbrown0:
> if both Python and JavaScript do that because OpenSCAD is script rather than program code.

I do think that JS and Python are better languages to copy than C++ or Java (or Go), but I don't think that the "script" versus "program" distinction is well-defined or useful.

I'd be happy to hear what you think the distinction is, but this doesn't seem like the right place; I suggest private e-mail.

coryrc:
> And @coryrc: isnt this a completely different set of issues you're talking about? You propose new keywords (`namespace`) and syntax (`::`). This is a very limited PR that we're trying to close down, it has a long history being discussed in OEP8.

I don't know where you mean "being discussed in OEP8"; there isn't an Issue for it, there's [this github wiki page](https://github.com/openscad/openscad/wiki/OEP8%3A-Objects-%28dictionaries%3F%29%2C-Geometry-as-data%2C-and-Module-References) but `this` is listed under "Futures" and, per the discussion going on here, is certainly far from already agreed (and I don't understand how it would be discussed there?).

My point is I don't think this should be added because there are other, more intuitive forms of "OO" appropriate to a language with immutable data than copying Javascript. I think this approach complicates things significantly while only slightly improving things.

coryrc:
Actually, my bigger point is I don't think this (pun intended) makes code better. In the mailing list I'm trying to come up with some examples of how what I'm proposing makes existing code better.

pkriens:
I feel this discussion goes in all directions except closure ...

@t-paul added the discussion section at the top, which is how we name the variable for the object the function came from. 

We now seem to be starting to discuss the foundations of OO ... and even proposing completely removing access to the underlying object, which I can't get my head around since it is useful in many scenarios for many people and fail to see the cost for people that want to ignore it.

To expedite this, I can live when there is at least some way for a function to find the 'current' object and ignore the rest but lets try to focus on closing this PR, even if it is not accepting it.

coryrc:
I think the callback issue proves this PR's implementation is fundamentally flawed.

You could have "this" added as an implicit statically-scoped variable on every call to object() and it would work as expected: anonymous functions are still anonymous functions, functions defined in an object scope are methods.

jordanbrown0:
>You could have "this" added as an implicit statically-scoped variable on every call to object() and it would work as expected: anonymous functions are still anonymous functions, functions defined in an object scope are methods.

You'd have to further explain this before I could discuss it.

coryrc:
@jordanbrown0 

> You'd have to further explain this before I could discuss it.

```
this = 5;
function five () this;
x = object(identity=function () this, f=five);
assert(x.identity()==x);
assert(x.f()==5);
```

That's the semantics of `this` in every language using `this` that I'm familiar with. All those languages either have mutable objects and/or have class definitions. If we had mutable data, then it would be equivalent to:

```
this = 5;
function five () this;
x = let( this = object() )
    this.identity = function () this;
    this.f = five;
    return this;
assert(x.identity()==x);
assert(x.f()==5);
```

jordanbrown0:
JavaScript won't let you assign to `this` - it's a keyword, not a variable - so the rough equivalent is something like:
```
function five() { return this; };
x = {
    identity: function () { return this; },
    f: five
};
console.log(x.identity());
console.log(x.f());
```

You're expecting `five` to have `this` be bound to the object that it's created as part of, the global object.

```
C:\...>node tmp.js
{ identity: [Function: identity], f: [Function: five] }
{ identity: [Function: identity], f: [Function: five] }
```

Nope.  `this` is bound by calling a function that was retrieved from an object.  None of the binding happens at object creation time.

And that's important for the scheme that we're suggesting, because in this program:
```
o1 = object(value = 5, f = function () this.value);
o2 = object(o1, value=6);
```
it's important that `o2.f()` have `this` bind to `o2`, not to `o1`.

coryrc:
> JavaScript won't let you assign to `this` - it's a keyword, not a variable - so the rough equivalent is something like:
> 
> ```
> function five() { return this; };
> x = {
>     identity: function () { return this; },
>     f: five
> };
> console.log(x.identity());
> console.log(x.f());
> ```
> 
> You're expecting `five` to have `this` be bound to the object that it's created as part of, the global object.

No, I'm expecting it not to. It's a function, not a method. Particularly because in this proposal, `this` is not a keyword.

jordanbrown0:
For JavaScript, your expectation is incorrect.  In that example, if you call `five()` in isolation, `this` binds to the global object.

And note that when you put a reference to `five()` into an object, and call it via the object, `this` binds to that object.

It is the call that binds `this`, not the definition of the function.

I can supply a more elaborate example where the function is defined as part of one object and then the reference is copied to a second object and called; it will bind to whichever object it's called through.  Or you can do the experiments yourself.

jordanbrown0:
It is the very fact that it is called via retrieval from an object that makes it a method, not where it is defined.  (Or `Function.call()` can do it explicitly.)

coryrc:
> For JavaScript, your expectation is incorrect. In that example, if you call `five()` in isolation, `this` binds to the global object.
> 
> And note that when you put a reference to `five()` into an object, and call it via the object, `this` binds to that object.
> 
> It is the call that binds `this`, not the definition of the function.

Then the `this` in your callback example:

```
// user
userobj = object(
    libobj = libfactory(cb=function() this.that),
    method = function() this.libobj.method(),
    that = 654
);
```

refers to the global `this` or, if this code is within a function-called-as-a-method, to that method's `this`. It would never refer to `userobj` nor `libobj`.

Which is not simple.

coryrc:
> It is the very fact that it is called via retrieval from an object that makes it a method, not where it is defined. (Or `Function.call()` can do it explicitly.)

That's creating "spooky action" and is not easy to debug or intuitive, hence all the innovations in the last 19 years on JavaScript.

jordanbrown0:
>refers to the global this or, if this code is within a function-called-as-a-method, to that method's this. It would never refer to userobj nor libobj.

Exactly what it would refer to, and how, is the entire subject of that sub-discussion.

If it does the most natural thing, the thing that JS would do, we can't know from that snippet what it would bind to, because we don't know what `libfactory()` does with the callback function.  If libfactory() stashes it in the `callback` member of its returned object, and calls it with `this.callback()`, it will indeed have `this` bound to libobj.  And yes, that's confusing - and disastrous from a namespace perspective if `this` is implied.  And that's exactly why we are having this conversation.

>That's creating "spooky action" and is not easy to debug or intuitive

And yet according to many or most surveys, JavaScript is the most heavily used programming language in the world, and so familiar to the most people...

If we were trying to create a first-class OO language, I'd agree with you.  But if we were trying to create a first-class OO language, I would strongly argue for *not* doing it, because the world already has too many of them.  We're not trying to create a first-class OO language.  We're trying to take our little domain-specific language and give it some OO-ish features, to allow our library authors to do things a bit more easily than they can today, help out with namespace issues, and so on.

pkriens:
I am a bit lost how I can get this decided upon ... 

I need this functionality in the snapshot so I can work with others to try this out in libraries. I am willing to adapt this PR asap to whatever changes are deemed necessary. I made my case in this PR and if that is not convincing then I go along with whatever is decided, even if it means closing this PR. 

Leaving a PR hanging without a decision is imho not a good strategy to make progress.

Puzzled how the process works in these cases?

t-paul:
I'm not able to follow 200 mails per day in detail lately, I'll try to filter the discussion in regard to just the "self-reference" whatever it will be named as that's the scope of the PR. I outlined my main concern somewhere at the top of this discussion that a fully automatic reference could case very surprising results.

I really don't like `self` in python and that you have to type it always, but it has the strong benefit of being very clear and explicit.

In general it would be ideal to collect a couple of small code examples for the various cases that need to be considered. I think the Wiki here on Github is generally editable, I'll try to find some time to start a page for that.

pkriens:
I am sorry, I sympathize with your plight ... have been in similar situations, you guys are doing a tremendous job, a lot of things seem to be going on.

The current Issues are:

1. name of `$this`
2. must `$this` be a:
   a. Parameter in the function to opt-in, or
   b. In the context of the function invocation.
3. should the members of the object be in the context.

Current PR (and my preferences) are 

* name -> `$this`
* `$this` in the function's context
* all members in context

Jason is torn on these decisions. I will go along with whatever he decides since it is more important for me to have access to $this some way than the syntactic sugar of the rest.  Once Jordan decides (or someone overrides) I will update the PR faster as lightning with the decisions and I hope it can be merged then.

### Some Examples

* _Binding_ is the process of letting a function have access to the members of the object that holds it,
* A _method_ is bound to its owning object (`$this`), 
* A _callback_ is a function that should not be bound to `$this` in the `object()` function.

**2.a** Binding by using `$this` as a parameter

      object( a=1, f = function($this) $this.a ); // method
      object( a=1, f = function() $this.a ); // FAIL, $this not defined
      object( a=1, f = function() ... ); // normal callback 
      object( a=1, f = [function($this) ... ]); // escaped callback for no rebinding to this $this

**2.b** Binding by placing `$this` in the function's context:

      object( a=1, f = function() $this.a ); // method
      object( a=1, f = [function() $this.b] ); // wrapped callback (b is from another $this!)
 
**3** Binding by placing members in context:

      object( a=1, f = function() a ); // method, a refers to $this.a == 1

pkriens:
@jordanbrown0 if I update the PR to remove members and `$this` in the function's invocation context and will look for a `this` function arguments and fill it in, will you then agree merging this? 
```
o = object( 
       b=42,  
       method = function(a=1, this) this.b,
       callback = function(a=1) a
);
```
I really want to progress on this.

jordanbrown0:
First, to be clear:  my agreement isn't sufficient or necessary.  I don't have the authority to merge PRs.  I also don't feel like it's my decision to make.

My inclination is toward the side of making `this` be a normal variable created by a call through a function reference that's directly retrieved from an object.  That is, `o.func()` sets it, but `f = o.func; f();` would not.

[ This meshes a bit with what you suggest, but I didn't realize it until a bit later in my thought chain.  Let me work my way there. This first bit is some stuff I thought about before deeply looking at your suggestion. ]

As I think about this a little more, I realized two things.  First, what I just described is a parameter to the function that's automatically supplied by the call.  Second, as a parameter... you should be able to explicitly set it in the call if you want to.

That latter, explicitly setting it in the call, answers an important question:   How do you make a function call against an object that the function reference didn't come from?  `func(this=obj)`.

* Unfortunately, it does **not** directly answer how you have a library not-set `this` when it calls a callback, because an explicit `this=undef` would override any `this` that the callback had based on lexical scoping.  Perhaps that could be a special case (though I don't like that answer), or perhaps we rely on the "assign to a temporary variable" answer for this relatively uncommon case.  (It's quite common in JavaScript, but historically JavaScript is very callback-heavy, and I don't expect to see that nearly as much here.)

Note that that latter implies that `this` is a valid argument to all functions and modules, whether or not they declare it as a parameter.

Supercall example:

```
sup = object(
    method = function () "sup"
);
mid = object(
    sup,
    method = function () str("mid", sup.method(this=this))
);
sub = object(
    mid,
    method = function () str("sub", mid.method(this=this))
);
```
`echo(sub.method())` should yield `"submidsup"`.

---

You've suggested that `this` be an explicit parameter to a method, by name.  (Versus my thought above where it was implicit in all functions.)  That's an interesting variation on Python style, where the this-ish argument is the first argument, with no specific name.  (Conventionally it's `self`, but it can be anything.)

Advantages:  makes where "this" comes from (in the method) be obvious.  The call can check for whether the function accepts a `this`, and use that to decide whether or not to pass one, answering the library-object-callback question.  If an explicit assignment of `this` in the call overrides any implicit assignment, that answers the call-against-specific-object question and thus the "super" question.

Disadvantages:  methods require boilerplate "this" declaration.  Not really compatible with ever implicitly including members in the called function's context.

I am unclear on whether an attempt to call a function that declares a "this" parameter, without implicitly or explicitly supplying one - that is, calling a function that expects to be a method, in a non-method-y-way, should be an error.  Existing argument/parameter behavior suggests "no".  Error-resistance suggests "yes".  I think I lean slightly toward "no".

Let's work through a few examples that combine this-as-explicit-parameter with the ability to override it in a call...

Simple method call:
```
o = object(
    value = 1,
    method = function (this) this.value
);

echo(o.method());    // 1
```

Library function with callback:
```
function libfunc(cb) = cb() + 1;

userobj = object(
    value = 2,
    method = function (this)
        libfunc(
            function () this.value
        )
);

echo(userobj.method());   // 3 (2+1)
```


Factory function with callback:
```
function factory(cb) =
    echo("factory")
    object(
        value = "lib",
        cb = cb,
        libmethod = function (this)
            echo("libmethod", this.value)
            this.cb()
    );

userobj = object(
    value = "user",
    usermethod = function (this)
        echo("usermethod", this.value)
        let (libobj = factory(function () 
            echo("callback", this.value)
            0
        ))
        libobj.libmethod()
);
userobj.usermethod();
```
should yield:
```
"usermethod", "user"
"factory"
"libmethod", "lib"
"callback", "user"
```

Note that the callback function **could** peek at the library object by explicitly having a `this` argument.  But if it really wants to peek, it doesn't have to try that hard; it could just directly look at libobj.

This does expose an awkward point:  `libobj` can't be a member of `userobj`, because it needs a `this` in context for the callback, and we don't have a `this` available when we're constructing the object.  In JS there wouldn't be a problem, because you'd create `libobj` from a constructor rather than using object-literal syntax, and the constructor **does** have `this`.

This would work, using the creation context (rather than "this") to get a reference to the object:
```
userobj = object(
    value = "user",
    usermethod = function (this)
        echo("usermethod", this.value)
        this.libobj.libmethod(),
    libobj = factory(function () 
            echo("callback", userobj.value)      // no "this"
            0
        )
);
```
... but as previously discussed, relying on the creation context does not play well when you copy pieces of objects around, notably if you ever tried to subclass `userobj` and have the subclass supply a different `value`.

I **think** that may be deep enough in the weeds that we can ignore it.  Or maybe five years from now somebody will be cursing us.

Supercall (basically the same as my example above, but with `this` as an explicit parameter):
```
sup = object(
    method = function (this) "sup"
);
mid = object(
    sup,
    method = function (this) str("mid", sup.method(this=this))
);
sub = object(
    mid,
    method = function (this) str("sub", mid.method(this=this))
);

echo(sub.method());     // "submidsup"
```

---

Net, I'm pretty happy with that.

* It does have the Python-esque explicit parameter, but because it's by name (and because our call semantics can cheat and peek at the function's parameters) there's a straightforward way to call a function reference that's an object member, without calling it as a method.
* Supercalls are reasonably clean.
* The callback-from-library-object case doesn't require a reassignment dance.
* Callbacks get `this` from their "parent".

pkriens:
I make a new PR with `this` as a named parameter in the function invocation and hope you'll finally agree with it then.

Name:			this
Opt-in			parameter in function
Context members		NO

Men, you need a lot of words ... 

	Peter

pkriens:
Yeah!

coryrc:
Your current proposal is forward compatible with generic functions.

A function is a method only when it has a `this` argument, is a member of an object, and is called as a member, right?

So it works like this:

```
y = function (a, this) this.x+1;
echo(y(this=object(x=2))); // 3
echo(y("b", object(x=4))); // 5

obj=object(x=10,y=y);
echo(obj.y()); // 11

z = obj.y;
echo(z()); // undef, warning can't add undef + int

assert(z==y); // true
```

Furthermore, "parent" objects can have private things like so:

```
function make_parent(private_value) =
  let( private_function=function () private_value )
    object(
      public_value = private_value*2,
      public_method = function (this, some_arg) private_function() * some_arg,
    );
obj = make_parent(3);
// obj has only two values: public_value and public_method
```

[since objects and closures are really the same thing](https://wiki.c2.com/?ClosuresAndObjectsAreEquivalent).

pkriens:
You can try out the PR ...

src/core/builtin_functions.cc
```diff
@@ -557,7 +562,30 @@ Value builtin_object(const std::shared_ptr<const Context>& context, const Functi
...
+      auto it = function.findAssignmentByName("this");
+      if (!it) continue;
+
+      auto parent = function.getContext();
+      if (parent->lookup_local_variable("this")) {
+        // function is already a method, so take the parent context
+        parent = parent->getParent();
+      }
+      ContextHandle<Context> ctx = Context::create<Context>(parent);
+      contexts.push_back(ctx.operator->());
```
coryrc, line 578: If you make `contexts` a std::vector<std::shared_ptr<Context>> then this can be `push_back(*ctx)`.

coryrc:
> You can try out the PR ...

So assigning functions with a `this` argument to objects is what creates a method. Methods are bound to that object until they get assigned to another object. `this` can be overridden for methods.

```
f1 = function() 1;
f2 = function(this) this.f1()+1;
o = object(f1=f1,f2=f2);
f3=o.f2;
echo(f1==o.f1); // true
echo(f2==o.f2); // false
echo(f3()); // 2
echo(f3(object(f1=function () 4))); // 5
```

src/core/builtin_functions.cc
coryrc, line 578: This can be `.push_back(*ctx)` if contexts is vector of shared_ptr<Context>

src/core/builtin_functions.cc
```diff
@@ -557,7 +562,30 @@ Value builtin_object(const std::shared_ptr<const Context>& context, const Functi
     }
     n++;
   }
-  return result;
+  std::vector<Context *> contexts;
+  for (int i = 0; i < result.ptr->values.size(); i++) {
```
coryrc: Could be for each loop

coryrc:
However, a small bug is found:

```
module a() {
  this = object(nope=10);
  f = function (this) this.nope;
  cube(f());
}
a(); // Size 10 cube
module b() {
  b = object(nope=10);
  f = function (b) b.nope;
  cube(f());
}
translate([-10,0,0]) b(); // Small cube (b.nope is undef)
```

src/core/Parameters.cc
```diff
@@ -254,6 +256,9 @@ Parameters Parameters::parse(Arguments arguments, const Location& loc,
     [](const std::shared_ptr<Assignment>& assignment) { return assignment->getName(); })};
 
   for (const auto& parameter : required_parameters) {
+    // see builtin_functions.cc::builtin_object() for an explanation
+    if (parameter->getName() == "this" && defining_context->try_lookup_variable("this")) continue;
```
coryrc: try_lookup_variable is recursive; at least I think you should use `lookup_local_variable`, but that still preserves the bug I just posted.

src/core/Parameters.cc
```diff
@@ -26,6 +26,8 @@
 
 #include "core/Parameters.h"
 
+#include "Context.h"
```
coryrc: should be "core/Context.h" and done with the others in the alphabetised list

src/core/FunctionType.h
```diff
@@ -29,6 +30,16 @@ class FunctionType
   [[nodiscard]] const std::shared_ptr<Expression>& getExpr() const { return expr; }
   [[nodiscard]] const std::shared_ptr<AssignmentList>& getParameters() const { return parameters; }
 
+  boost::optional<size_t> findAssignmentByName(const std::string& name) const
```
coryrc, line 33: I believe you are only using this as `hasAssignment` so that returning a bool would be better. If not, please use std::optional

src/core/builtin_functions.cc
```diff
@@ -557,7 +562,30 @@ Value builtin_object(const std::shared_ptr<const Context>& context, const Functi
...
+      }
+      ContextHandle<Context> ctx = Context::create<Context>(parent);
+      contexts.push_back(ctx.operator->());
+      Value method(FunctionType(*ctx, function.getExpr(), function.getParameters()));
+      result.ptr->values[i] = std::move(method);
+    }
+  }
+  Value object(result);
+  for (Context *c : contexts) {
+    c->set_variable("this", object.clone());
```
coryrc: Would be more efficient to use the same shared_ptr for all of them. Ideally builtin_object would also return the same shared_ptr otherwise you'll have at least 2x in memory

tests/data/scad/experimental/object/object-tests.scad
```diff
@@ -94,6 +94,48 @@ Betty = object(
...
+testEq( osc3.f(), 43, "access copied object");
+testEq( osc.f(), 42, "access original object");
+
+Y = object( f=function(n,this) n> 1 ? n*this.f(n-1) : 1).f;
+testEq( Y(5), 120, "recursively calling a function in an object");
+
+testEq( object( $fs = 42 ).$fs, 42, "using config names as field is ok");
+testEq( object( this = 42 ).this, 42, "also this");
+testEq( object( this = 42, f=function(this) this.this ).f(), 42, "but functions see the this object");
+
```
coryrc, line 136: Add negative test so it doesn't get `this` unless it's actually assigned to the object.

src/core/Parameters.cc
```diff
@@ -254,6 +256,9 @@ Parameters Parameters::parse(Arguments arguments, const Location& loc,
     [](const std::shared_ptr<Assignment>& assignment) { return assignment->getName(); })};
 
   for (const auto& parameter : required_parameters) {
+    // see builtin_functions.cc::builtin_object() for an explanation
```
coryrc, line 258: I'd rather it just say why...

tests/data/scad/experimental/object/object-tests.scad
```diff
@@ -94,6 +94,48 @@ Betty = object(
...
+
+// Tests to check the implicit scope for functions in objects
+osc = object( a=42, f=function(this) this.a );
+test( function() osc.f() == 42, "check referring to other field");
+fsc = osc.f;
+test( function() fsc() == 42,   "check referring from copied field");
+
+fsc1 = object(a=42, f=function(this) this.a).f;
+test( function() fsc1() == 42,  "function stored from orphaned object");
```
coryrc, line 112: It's a method now; no longer a "function"

nophead:
> However, a small bug is found:
> 
> ```
> module a() {
>   this = object(nope=10);
>   f = function (this) this.nope;
>   cube(f());
> }
> a(); // Size 10 cube
> module b() {
>   b = object(nope=10);
>   f = function (b) b.nope;
>   cube(f());
> }
> translate([-10,0,0]) b(); // Small cube (b.nope is undef)
> ```

You declared b as a parameter to f but didn't pass it so it is undefined.

coryrc:
> > However, a small bug is found:
> > ```
> > module a() {
> >   this = object(nope=10);
> >   f = function (this) this.nope;
> >   cube(f());
> > }
> > a(); // Size 10 cube
> You declared b as a parameter to f but didn't pass it so it is undefined.

I know, the problem is that it *isn't* undefined in the first example.

src/core/builtin_functions.cc
pkriens: yes, I changed it but is it important?

src/core/Parameters.cc
pkriens: But lookup gives a warning. I added a boolean to try_lookup to recurse or not to recurse.

src/core/Parameters.cc
pkriens: is there a list of rules abuot this?

src/core/FunctionType.h
pkriens: I see boost::optional used everywhere? I'd prefer to stick to the already included dependencies?

coryrc: https://github.com/openscad/openscad/pull/6144#discussion_r2311765790

src/core/Parameters.cc
coryrc: The one with *_local* in the middle doesn't and isn't recursive.

src/core/builtin_functions.cc
coryrc: Consistent with rest of the codebase and less prone to error when refactoring; so, yes.

src/core/builtin_functions.cc
pkriens: I don't think you looked how Value worked? A clone is a clone of the shared ptr, not of the value.

# Value in OpenSCAD

## What

The C++ Value class represents the different value types in the OpenSCAD language. It uses a std::variant to represent the different types. It supports the following types:

		UndefType, 
		bool, 
		double, 
		str_utf8_wrapper, 
		VectorType, 
		EmbeddedVectorType, 
		RangePtr, 
		FunctionPtr, 
		ObjectPtr

Value is an immutable _value_ object (as in contrast to reference). The Value object has no identity it only has _content_ (logically). Its content can never change, it is immutable. Two Value objects v1 and v2 are equal if their content is equal.

If v1==v2 (or other comparisons) is true or false at any time, it will remain so forever. This is identical to how most languages treat scalars like numbers, but this is also true for the more complex types. A Value that is, for example, an object will never change its members.

## Storage

In OpenSCAD, large arrays of arrays of doubles are common but double is represented as a Value. There is therefore a strong incentive to minimize the sizeof(Value). The Value object holds a std::variant which overlaps the memory of the supported types and tracks which type 'owns' the memory and should therefore be deleted when the Value object is deleted. Currently, the variant has a size of 24 bytes: an 8-byte index for the current type and up to 16 bytes for the payload.

The double and bool types are stored directly in the available 16 bytes. Complex types like Undefined, Vector, Range, Function, and Object use considerable memory to store their content. For this reason, Value stores a shared_ptr in the variant, which explains why 16 bytes are needed. A pointer or a double only needs 8 bytes, but a shared_ptr doubles this.

## Complex types

Complex Value objects implement different shared ptr schemes. When a complex Value gets copied or moved, this does not copy the memory of its resource but shares the resource. Reference counting is used to detect when the last Value object is deleted and then the resource is also deleted.

There are different schemes used for the various complex types:

	UndefType		unique_ptr<vector<string>>> !! currently not immutable
	str_utf8_wrapper	shared_ptr<str_utf8_t>
	VectorType		shared_ptr<VectorObject>
	RangePtr		ValuePtr<RangeType>
	FunctionPtr		ValuePtr<FunctionType>
	ObjectPtr		ValuePtr<ObjectType>

ValuePtr<T> is a template to abstract/unify this shared_ptr behavior. It holds a shared_ptr to an object of type T. RangeType,FunctionType,ObjectType use it currently.

The Value class has constructors for each of the variant types. This is achieved with a std::forward template. That is:

	Value v( new ObjectType() );

Will assign v.value an ObjectPtr < ValuePtr<ObjectType> < new ObjectType. The ValuePtr<T> template has a std::shared_ptr and implements identical semantics for its types.

Therefore, except for  double and bool, Value is more or less a shared ptr to an  immutable constant object on the heap that will get deleted when the last Value object referencing it is deleted. 

Value objects cannot be copied because the standard copy constructor in Value is deleted. Only the move constructor exists.

	Value v1( new ObjectType(nullptr) );
	Value v2( v1 );			// error, requires copy
	Value v3( std::move(v1));	// move(&v1,&v3)

There is a Value.clone() method that makes an explicit copy of the Value object.

	Value v1( new ObjectType(nullptr) );
	Value v2( v1.clone() ); 	// move( &v1.clone(), &v2)

The clone() method makes only a copy of the Value, increasing any reference counts for the complex types. A clone will always be identical to its source. Source == clone must always be true except for the UndefinedType.

## Usage

Functions that take Value should use the following prototypes:

– access( const Value & v ) – v should NOT be stored

	Value v(1);
	access( v );			// ok
	access( 1 );			// &move(&Value(1),&v)

  In access

	void access( const Value & v ) {
	  Value v1(v);		   // error, no copy constructor
	  Value v2(std::move(v));  // error, v const, cannot be moved
	  Value v3(v.clone());     // move(&v.clone(),&v3)
	}
  
- store( Value v ) – v is a local copy and should be moved to a storage

  Temporary object

	store( 1 ); 		   // move( &Value(1), &v)

  When handing over a Value

	Value v1(1);		   // move( &Value(1), &v1) 
	store( std::move(v1) );	   // move( &v1, &v )

  No confusion, trying to pass a Value cannot create a copy

	Value v1(1);		   // move( &Value(1), &v1) 
	store( v1 );		   // error, requires copy

  In store()

	void store( Value v) {
	  Value v1( v );		// error, requires copy
	  Value v2( std::move(v) ); 	// move( &v1, &v2 )
	  Value v3( v.clone() );    	// move( &v.clone(), &v2)
	}

When Value objects are stored in containers, like vector<Value>, the deleted copy constructor can cause issues. The use of the Value.clone() method is needed to make explicit copies.

  std::vector<Value> clone_vector(const std::vector<Value>& input) {
    std::vector<Value> result;
    for (const auto& v : input) {
        result.push_back(v.clone());
    }
    return result;
  }

Using Value.clone() as a temporary object will allow it to be efficiently moved.

src/core/FunctionType.h
pkriens: and the return of the index is much better because quite often you need to have the index.

src/core/Parameters.cc
pkriens: No, better to not have redundancy ... always point to one place

src/core/builtin_functions.cc
coryrc: Right in the middle it says:
"There is a Value.clone() method that makes an explicit copy of the Value object."

coryrc: Ah, you didn't mean `Value` when you wrote "A clone is a clone of the shared ptr, not of the value.", you mean it's not copying `result`. Which I see now, thanks.

src/core/Parameters.cc
coryrc: Don't know, but it's that way everywhere

src/core/FunctionType.h
coryrc: You don't now so YAGNI

src/core/builtin_functions.cc
pkriens: I see plenty of boost:: in the code and I see plenty of for(int i... in the code. I will follow a style book this feels arbitrary. Is there a document where these rules are mentioned or are these your preferences? And in this case, because the vector is updated, it is more documenting to see the index used instead of a & imho.

coryrc: I shared the link to Marius asking me to use std::optional, he's the one you need approval from. We also just fixed a couple (big) memory leaks from dumb pointers being used, so it's probably not a good idea to keep using them in easily-avoidable situations (particularly here where it makes the code less readable).

src/core/builtin_functions.cc
jordanbrown0: Drive-by:  Value.clone makes a copy of the Value object.  For anything other than numbers and booleans, the Value object has a shared pointer to the actual data.  That actual data is *not* copied by the clone operation; it is shared between the old and new copies of the Value.  (Actually, for undef I'm not sure what it does, other than that it's not a trivial copy of the Value.)

src/core/FunctionType.h
pkriens: I just wasted another hour because `std::optional` with templates don't compile, this was the reason I now recalled I went to boost because I tried them. I seem to be mistaken that etiquette obliged one to only propose a fix if one actually knows what one is talking about. You also had no qualms to comment on the use of the `clone()` method although you had no experience with the `Value` class; the complaint about the indexed for loop was code already approved by Marius so I am not sure it is your place to correct him. The hangup you had with the Context pointers which I used from my deep fear based on days of debugging the cycle garbage collection algorithm which you're undoubtedly not familiar with. That is 4 annoying mistakes in one review and then I do not count your pushing preferences that have as far as I know no formal ground in the OpenSCAD culture. So I'd appreciate if in the future you stick to the parts you actually have deep knowledge about.

Despite my mild criticism of your behavior, I do appreciate actual bugs so please keep up the good work!

coryrc: > I just wasted another hour because `std::optional` with templates don't compile, this was the reason I now recalled I went to boost because I tried them.

https://github.com/openscad/openscad/pull/6144/commits/64d73a3b29542ac1f87620a51ff341a9eccb0b16

There's no shame in asking for help when you can't figure things out!

tests/data/scad/experimental/object/object-tests.scad
```diff
@@ -134,7 +134,17 @@ testEq( object( $fs = 42 ).$fs, 42, "using config names as field is ok");
 testEq( object( this = 42 ).this, 42, "also this");
 testEq( object( this = 42, f=function(this) this.this ).f(), 42, "but functions see the this object");
 
-
+{
```
coryrc, line 137: This line doesn't create a new scope ([see parser.y](https://github.com/openscad/openscad/blob/ce6fcc5480eebe5a421c67aee03f39dc352d3327/src/core/parser.y#L187)) so I wouldn't use it to remain consistent with the remainder of the tests in this file and not be misleading (I have to remind myself it doesn't create new scope every time I see it).

pkriens: I suggest after this is in the code base you make a PR with these suggestions.

pkriens:
@kintel, @jordanbrown0  I would appreciate a review from one of you so this PR can finally go through.

pkriens:
@kintel can this be merged?

kintel:
This is on my list, but I haven't had time to prioritize it. If we can de-risk the impact  by making it experimental, it will make it easier to merge as we can then defer the discussion on whether it's solid enough to make it into the next release.

t-paul:
> This is on my list, but I haven't had time to prioritize it. If we can de-risk the impact by making it experimental, it will make it easier to merge as we can then defer the discussion on whether it's solid enough to make it into the next release.

@kintel I believe the feature is essentially behind the experimental flag already due to the `object()` function itself being marked as experimental. Without it there's not really any way of creating object that could trigger the code.

I have not found enough time to give it a longer testing, but I think merging would help getting it more visibility to find things that we might want to change before fully committing to it.

kintel:
Right, didn't think about that. I guess we can cross that bridge if we decide to include any of this in the next release, but good for now :)

kintel:
@pkriens Are you ok merging this with the caveat that we'll do a design review later, when we decide how to bring this out of experimental mode?

pkriens:
@t-paul thanks a lot of approving the merge. 

@kintel I am happy with any progress. My issue is that I want this in libraries, specifically in BOSL2, so unless it is in available in some officially released form I am OK even if people have to enable it specifically.

kintel:
Ok, will merge. It won't be available in an official release right now, but as an experimental feature in nightly builds only, so if any external libraries utilize this, they run the risk of having to track any changes to this feature, but that's business as usual for any experimental feature usage.

pkriens:
Yeah!!

PR #6023, opened 2025-07-15
ValueIterator (generic iterator for types to return to CMM)

jordanbrown0:
This is very much a draft; it is playing in C++ areas that I only dimly understand, and took a lot of trial and error.  Comments solicited.

The goal is to make it so that an aggregate type (ObjectType and VectorType, primarily, but perhaps others in the future) can implement a generic iterator and return it to ContextMemoryManager, so that they don't have to store their Values in a std::vector if they don't want to.

There are two class hierarchies that are more or less obvious:

Values (subclasses VectorValues and ObjectValues) is an object that has begin() and end() methods for traversing the aggregate.  This needs to be a separate class, not just part of VectorType/VectorObject and ObjectType/ObjectObject, because there may be multiple ways to traverse the aggregate.  (For instance, for Object you might traverse by keys, by values, or by {key, value} pairs.)  The begin() and end() methods return ValueIterator.

The iterators that do the work (VectorValueIterator and ObjectValueIterator) are subclasses of VirtualValueIterator.  They implement the standard iterator methods, but one runs on top of a vector and the other on top of an object.

And then there's the relationship between ValueIterator and VirtualValueIterator.  It appears (and I am too tired right now to double check a reference) that `for (v : aggregate)` expands into something that starts with `auto it = aggregate.begin();`.  That doesn't work so good, if aggregate.begin() is defined to return the superclass and hide the actual subclass involved, because the compiler can't know how much memory to allocate for `it`.  ValueIterator is a proxy that holds a pointer to the underlying per-aggregate iterator.  It's constant-size and known to the caller, while allowing the underlying iterator to be opaque.

A cleaner mechanism might be to use a lambda and a forEach() style mechanism, where the lambda is called once for each Value.

### DIFF
--- a/src/core/Value.cc
+++ b/src/core/Value.cc
@@ -635,6 +635,61 @@ void VectorType::VectorObjectDeleter::operator()(VectorObject *v)
   delete orig;
 }
 
+std::shared_ptr<Values>
+VectorType::embeddedValues() const
+{
+  class VectorValueIterator: public VirtualValueIterator  {
+  private:
+    VectorType::iterator it;
+  public:
+    VectorValueIterator(VectorType::iterator i) : it(i) {};
+
+    void inc() { ++it; };
+    const Value& operator*() const { return *it; }
+    bool operator==(const VirtualValueIterator& other) const { return it == dynamic_cast<const VectorValueIterator *>(&other)->it; }
+    bool operator!=(const VirtualValueIterator& other) const { return it != dynamic_cast<const VectorValueIterator *>(&other)->it; }
+  };
+
+  class VectorValues : public Values {
+  public:
+    VectorValues(const VectorType *v) : vector(v) { };
+
+    const ValueIterator begin() const { return ValueIterator(std::make_shared<VectorValueIterator>(vector->begin())); };
+    const ValueIterator end() const { return ValueIterator(std::make_shared<VectorValueIterator>(vector->end())); };
+  private:
+    const VectorType *vector;
+  };
+  return std::make_shared<VectorValues>(this);
+}
+
+std::shared_ptr<Values>
+ObjectType::embeddedValues() const
+{
+  class ObjectValueIterator: public VirtualValueIterator  {
+  public:
+    ObjectValueIterator(std::vector<std::string>::const_iterator i, const ObjectType *o) : it(i), obj(o) {};
+
+    void inc() { ++it; };
+    const Value& operator*() const { return obj->get(*it); }
+    bool operator==(const VirtualValueIterator& other) const { return it == dynamic_cast<const ObjectValueIterator *>(&other)->it; }
+    bool operator!=(const VirtualValueIterator& other) const { return it != dynamic_cast<const ObjectValueIterator *>(&other)->it; }
+  private:
+    std::vector<std::string>::const_iterator it;
+    const ObjectType *obj;
+  };
+
+  class ObjectValues : public Values {
+  public:
+    ObjectValues(const ObjectType *o) : obj(o) { };
+
+    const ValueIterator begin() const { return ValueIterator(std::make_shared<ObjectValueIterator>(obj->keys().begin(), obj)); };
+    const ValueIterator end() const { return ValueIterator(std::make_shared<ObjectValueIterator>(obj->keys().end(), obj)); };
+  private:
+    const ObjectType *obj;
+  };
+  return std::make_shared<ObjectValues>(this);
+}
+
 const VectorType& Value::toVector() const
 {
   static const VectorType empty(nullptr);
--- a/src/core/Value.cc
+++ b/src/core/Value.cc
@@ -644,18 +644,18 @@ VectorType::embeddedValues() const
   public:
     VectorValueIterator(VectorType::iterator i) : it(i) {};
 
-    void inc() { ++it; };
-    const Value& operator*() const { return *it; }
-    bool operator==(const VirtualValueIterator& other) const { return it == dynamic_cast<const VectorValueIterator *>(&other)->it; }
-    bool operator!=(const VirtualValueIterator& other) const { return it != dynamic_cast<const VectorValueIterator *>(&other)->it; }
+    VectorValueIterator& operator++() override { ++it; return *this; }
+    const Value& operator*() const override { return *it; }
+    bool operator==(const VirtualValueIterator& other) const override { return it == dynamic_cast<const VectorValueIterator *>(&other)->it; }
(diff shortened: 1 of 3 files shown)


jordanbrown0:
Looks like all of the tests are failing.  Haven't checked out why yet.  I tried it and smoke didn't come out!

src/core/Value.cc
```diff
@@ -635,6 +635,61 @@ void VectorType::VectorObjectDeleter::operator()(VectorObject *v)
   delete orig;
 }
 
+std::shared_ptr<Values>
+VectorType::embeddedValues() const
+{
+  class VectorValueIterator: public VirtualValueIterator  {
```
pca006132, line 641: I think these local classes causes visibility issues, and leads to linker issue in the CI.

src/core/Value.cc
```diff
@@ -635,6 +635,61 @@ void VectorType::VectorObjectDeleter::operator()(VectorObject *v)
...
+  private:
+    VectorType::iterator it;
+  public:
+    VectorValueIterator(VectorType::iterator i) : it(i) {};
+
+    void inc() { ++it; };
+    const Value& operator*() const { return *it; }
```
pca006132: In general when you override something, you want to mark it with `override`.

src/core/Value.h
```diff
@@ -461,3 +464,36 @@ std::ostream& operator<<(std::ostream& stream, const Value::ObjectType& u);
 using VectorType = Value::VectorType;
 using EmbeddedVectorType = Value::EmbeddedVectorType;
 using ObjectType = Value::ObjectType;
+
+class VirtualValueIterator
+{
+public:
+  // These need to know how the particular iterator works, and so must be implemented in
+  // a subclass.
+  virtual void inc();
```
pca006132: Mark these with `= 0` as they do not have an implementation yet. Same for the virtual methods in `Values`.

src/core/Value.h
```diff
@@ -461,3 +464,36 @@ std::ostream& operator<<(std::ostream& stream, const Value::ObjectType& u);
 using VectorType = Value::VectorType;
 using EmbeddedVectorType = Value::EmbeddedVectorType;
 using ObjectType = Value::ObjectType;
+
+class VirtualValueIterator
```
pca006132, line 468: When working with iterator-like stuff, you want to define `difference_type`, `value_type`, `pointer`, `reference` and `iterator_category` (see https://en.cppreference.com/w/cpp/iterator/iterator_traits.html). It is a bit annoying but worth the hassle.

src/core/Value.h
pca006132: And why `inc` instead of the typical `++`?

src/core/Value.h
```diff
@@ -461,3 +464,36 @@ std::ostream& operator<<(std::ostream& stream, const Value::ObjectType& u);
...
+  const Value& operator*() const { return (**vi); }
+  bool operator==(const ValueIterator& other) const { return *vi == *other.vi; }
+  bool operator!=(const ValueIterator& other) const { return *vi != *other.vi; }
+  const Value *operator->() const { return &**this; }
+};
+
+class Values
+{
+public:
+    [[nodiscard]] virtual const ValueIterator begin() const;
```
pca006132: Why `[[nodiscard]]`? Is it that important?

src/core/Value.h
```diff
@@ -461,3 +464,36 @@ std::ostream& operator<<(std::ostream& stream, const Value::ObjectType& u);
...
+  ValueIterator(std::shared_ptr<VirtualValueIterator> i) : vi(i) { };
+
+  ValueIterator& operator++() { vi->inc(); return *this; }
+  const Value& operator*() const { return (**vi); }
+  bool operator==(const ValueIterator& other) const { return *vi == *other.vi; }
+  bool operator!=(const ValueIterator& other) const { return *vi != *other.vi; }
+  const Value *operator->() const { return &**this; }
+};
+
+class Values
```
pca006132, line 494: Why would you need `Values`, instead of just returning a `std::pair<ValueIterator, ValueIterator>` from `embeddedValues`?

src/core/Value.cc
jordanbrown0: I suppose.  The whole point of local scoping is to cause visibility issues :-)  But why would they cause problems for CI?

src/core/Value.h
jordanbrown0: I couldn't get ++ to work.  I never figured out why.  Quite possibly prefix ++ versus suffix ++; I found later that on a different instance.

src/core/Value.h
jordanbrown0: Will look into it.  I was deliberately *not* doing that, because I was having enough trouble keeping the types straight without hiding them behind pseudonyms.

src/core/Value.h
jordanbrown0: Shrug.  Copied from VectorType::begin.

src/core/Value.h
jordanbrown0: Want the caller to be able to say
```
for (v : something) {
}
```
which implies returning an object that has `begin()` and `end()`.

And:
* As best I can tell it's got to be a concrete "leaf" class, because the `for()` wants to implicitly assign it to an invisible iterator variable.
* Don't want to just return the aggregate class itself (VectorType, ObjectType) because there may be more than one way to iterate across the aggregate.  For ObjectType you might want to iterate across keys, or across values, or across {key, value} pairs.

src/core/Value.cc
jordanbrown0: I'm not used to that from my experience in Java and JavaScript, but it makes sense.

src/core/Value.h
jordanbrown0: I think that was probably it, changed over and it's building OK now.

pkriens:
I'd not seen this PR ... excuses.

I think the iterator is not the optimal approach here. Notice that FunctionType might have for example to return the optional receiver, just 1 object. Seems awkward to make one object into an iterator? A lambda approach, as I did in #6029 seems to work better imho?

Currently our variant types  unnecessarily use different strategies for the shared_ptr approach. There is a template ValuePtr<T> which is a shared_ptr, used by FunctionType and RangeType. This template should imho also be used for the domain types VectorType, EmbeddedVectorType, ObjectType, and UndefinedType. The ValuePtr<T> template can then handle the use_count, the identity, and call back based method to access all referenced Value objects.

I suggest we merge #6029 over this one for now. When we moved all the compound types to ValuePtr<T> we could revisit.

But I defer to what you want.

jordanbrown0:
Closed in favor of #6029.

bitbasher:
Jordan .. that was brave work in a tough area of OO programming. now i am off to see what #6029 offers :)

PR #6024, opened 2025-07-15
update manifold to v3.2.0

pca006132:
https://github.com/elalish/manifold/releases/tag/v3.2.0, which contains bug fixes and performance improvement for large meshes.


pca006132:
ah crap, we probably need a patch release upstream for the missing include :)

pca006132:
checking if the simple patch can fix the CI error.

pca006132:
@t-paul this should be ready.

t-paul:
I'm getting a build error on Debian 11. What's the view for Manifold in that regard. With Debian 13 just around the corner I'm not too worried, it's just that up to now we supported Debian 11.

pca006132:
@t-paul just open an issue upstream and we will fix it :)

t-paul:
Ok, so lets keep it alive for a bit longer.

PR #6025, opened 2025-07-15
faster syntax highlighting

### LINKED ISSUE #5991: Syntax highlighting can be extremely slow for larger files
manifold rendering is much faster, but it seems to only be in effect on objects already loaded in openscad.
add this feature to objects on initial load.

when opening an already generated object, use manifold rendering when manifold rendering is selected.

I do have a work around.
1. load a simple file.scad.
2. generate the complex object code
3. reload.
The reload will use the fast rendering manifold option. 
where if you generate a complex object, then try to load it directly, it will take absolutely forever to render.

this works up to 200K objects, before some other function breaks the work around
WARNING: Normalized tree is growing past 200000 elements. Aborting normalization.

WARNING: CSG normalization resulted in an empty tree

**Additional context**

Image

pca006132:
Closes #5991.

This performs batch highlighting by using the `SCI_SETSTYLINGEX` event. Previously, we were using one event per token for highlighting, which triggers rendering after every single token (apparently, from gdb stepping) and was very slow for large files. With the batch API, this is much faster and can open the 33MB file in less than 10s (most of the time is probably not in highlighting).

### DIFF
--- a/src/gui/ScadLexer.cc
+++ b/src/gui/ScadLexer.cc
@@ -205,11 +205,16 @@ void Lex::lex_results(const std::string& input, int start, LexInterface *const o
   int isstyle = obj->getStyleAt(start - 1);
   if (isstyle == ecomment) results.state = rules_.state("COMMENT");
 
+  char* styles = new char[input.size()];
   lexertl::lookup(sm, results);
   while (results.id != eEOF) {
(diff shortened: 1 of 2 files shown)


t-paul (approved):
Huge improvement, very nice! :rocket:

PR #6026, opened 2025-07-15
Python: fix crash in union/difference/intersection

jstebbins:
Fixes https://github.com/openscad/openscad/issues/5906

### DIFF
--- a/src/python/pyfunctions.cc
+++ b/src/python/pyfunctions.cc
@@ -1382,8 +1382,10 @@ PyObject *python_csg_sub(PyObject *self, PyObject *args, PyObject *kwargs, OpenS
   for (i = 0; i < PyTuple_Size(args);i++) {
     obj = PyTuple_GetItem(args, i);
     PyObject *dict = nullptr;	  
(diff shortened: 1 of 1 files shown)


src/python/pyfunctions.cc
```diff
@@ -1382,8 +1382,10 @@ PyObject *python_csg_sub(PyObject *self, PyObject *args, PyObject *kwargs, OpenS
   for (i = 0; i < PyTuple_Size(args);i++) {
     obj = PyTuple_GetItem(args, i);
     PyObject *dict = nullptr;	  
-    if(i == 0) child = PyOpenSCADObjectToNodeMulti(obj, &dict);
```
t-paul, line 1385: What happened to the `if(i == 0)`?

gsohler:
the original plan was to derive the dictionary only from the first
argument, but that's nonsense.
Now all the dictionary is derived from all of them, with
1.st parameter : highest priority
last parmeter:  least priority

thus i == 0 is not needed anymore

t-paul (approved):

PR #6028, opened 2025-07-16
Fix LIB3MF detection in CMake

nomike:
As previously mentioned in https://github.com/openscad/openscad/issues/5877 there is a bug in one of the CMake scripts where a wrong variable name is used when trying to detect LIB3MF.

This patch fixes that detection.


nomike:
Sorry. wrong repo.

PR #6033, opened 2025-07-19
Python: Fix argument processing in offset

jstebbins:
'r' was a required argument when it should not be


gsohler:
Here is how it was fixed in pythonscad

https://github.com/pythonscad/pythonscad/commit/177288075d5569b5c00e4399dfd9719cfa5c3149
there is also the OO method which requires the same fix

jstebbins:
Yes, my fix addresses both methods.  The pythonscad fix also improves the error message.  I'm happy to add the improved error message if you like.

t-paul:
It guess keeping the messages in sync too would be nice.

PR #6034, opened 2025-07-20
Add support for per-polygon colour in WRL exports.

smartavionics:
Adds Color and ColorIndex fields to IndexedFaceSet.

RGB only, alpha is not supported in WRL file.


smartavionics:
I forgot to say that it has been tested using Kicad 3D model import. Here is a simple model of an APEM TL36 switch.

Screenshot_2025-07-20_13-08-14

t-paul (approved):
Nice, thank you!

PR #6042, opened 2025-07-24
Add explicit constData() assignment for a QByteArray

Sparsh-N:
Helps avoid compiler error and warnings. Currently, this operation would rely on implicit conversion of `fnameba` (type QByteArray) into type const char*. 

By calling `fnameba.constData()`, both parts of the ternary opn are the same, which is safer.

This PR is linked to #5984 , but merging this first could be of more general use and isnt specific to the windows/MSVC build that PR is working towards.


Sparsh-N:
On a somewhat related note, would replacing `editor->filepath.isEmpty()` with `fnameba.isEmpty()` be simpler? Considering that `fnameba` is set to `editor->filepath.toLocal8Bit()`, adding this change could also make it more readable and avoid redundancy. Any thoughts/suggestions appreciated.

t-paul (approved):

PR #6051, opened 2025-07-30
#6048 Bad output converting object to string

### LINKED ISSUE #6048: Bad output converting object to string
**Describe the bug**
In a rather convoluted case, brackets for a vector and braces for objects end up in the wrong places.

**To Reproduce**

```
Wilma = object(
    spouse=object()
);
echo( Wilma=Wilma);
 
Betty = object(
    friends=[Wilma]
);
echo( Betty  = Betty );
```
yields
```
ECHO: Wilma = { spouse = { }; }
ECHO: Betty = [{ spouse = { friends = { }; }]; }
```

Note that the Betty result isn't even valid; its braces and brackets are mismatched.    But note also that somehow `friends` has migrated inside `spouse`, when it's actually a member of the top-level object.

The object/vector structure *does* look valid; if you walk it and dump the components they appear to be correct.  It appears that just the string representation is wrong.

`str()` returns the same result.

**Expected behavior**
Output that correctly represents the data structure.

 - OS: Windows 11
 - System: Intel PC 64-bit
 - OpenSCAD Version: recent personal build, also 2025.07.11.

pkriens:
Value implements a stream and a string visitor. These visitors assume that a value that has descendants also uses the same visitor or the streams seem to be getting mixed up. In our case we defaulted to the standard version but then the arrays were formatted using the visitor.

Added the appropriate methods to the visitors, mostly copying the Vector methods and adapting them

Fixes #6048


t-paul:
Are there any object test yet? How much effort would it be to add the example from #6048 as a simple echo-test?

jordanbrown0:
There are object tests, though it looks like coverage of output formatting is weak.

However, my bet is that even fairly comprehensive tests wouldn't have found this issue. @pkriens would know better than me exactly what cases trigger this, but all simple and most complex cases work correctly.

jordanbrown0:
I'm not in a position right now to fully understand what's going on, but it seems very wrong that this formatting is duplicated. There should be exactly one function that does to-stream of a particular data type.  (And it seems like operator<<(stream, value) is the more obvious pattern, but maybe I'm missing something there.)

That is, it seems like maybe vectors are wrong and objects are right.

jordanbrown0:
I got >> and << backwards. Edited.

t-paul:
> There should be exactly one function that does to-stream of a particular data type.

Good point, but I think it makes sense to treat that as separate issue as right now the code changes here just follow the existing pattern we already have for the other types and do fix an actual issue.

jordanbrown0:
I found the actual problem.  It wasn't about a missing method, or at least not exactly.  (There *is* a missing method, and it looks like it'll trip an assert... if asserts are enabled.  But the fallback behavior if asserts are disabled is correct for most (all?) cases.)

Rather, it is that STR() does not work if the << method for the particular type itself uses STR().

The reason is that STR() puts its output stream in a thread_local, and when it's done it clears the stream.

thread_local is sort of static-for-this-thread.  So when somebody calls STR() on value A of type T, where A is an aggregate that includes another value B of type T, the inner calls muck with the stream being used by the outer calls, and hilarity ensues.  Text emitted by the outer call gets caught up by the inner call, returned as part of *its* return value, and cleared from the string.  Simple recursive cases of Object appear to work, because everything gets piled into one stream, and the string gets returned, the stream cleared, and then the string is re-added to the stream.  However, when you have a Vector in the middle, with its own tostream_visitor implementation, it gets more complicated because some things are going to Vector's stream and other things are going to STR's stream, and STR's stream is getting flushed out to a string at the wrong points relative to Vector's stream.

In particular, in ObjectA(Vector(ObjectB)), the initial output for ObjectA is instead returned by ObjectB and added to the final result by Vector.  Thus, if Object is going to emit `{...}` and Vector is going to emit `[...]`:
* ObjectA will put `{` into the STR stream
* Vector puts `[` into its stream
* ObjectB puts `{` into the STR stream
* ObjectB puts '}' into the STR stream
* ObjectB returns the STR stream content `{{}`, and clears the STR stream
* Vector puts `{{}` into its stream
* Vector puts `]` into its stream
* Vector returns its stream contents `[{{}]`
* ObjectA puts `[{{}]` into the STR stream
* ObjectA puts `}` into the STR stream
* ObjectA returns the STR stream contents `[[{{}]}`

Or something a lot like that.

This PR avoids this problem by not calling STR recursively.

STR should either properly handle being called recursively, or should guard against recursive calls.

I am not sure, but I think that the tostring_visitor / tostream_visitor infrastructure could be replaced by a much simpler scheme based entirely on operator<<.

PR #6053, opened 2025-07-30
Fix OpenCSG version detection

nomike:
When detecting the version of OpenCSG it is stored in a variable which is cached. Per default cache values are not overwritten as they are user-definable.

This means that when you run `cmake` for the first time, it will store the version number of OpenCSG and use the same one until you delete `CMakeCache.txt`, regardless of whether the actual version of OpenCSG changed or not (e.g. when you turn "-DUSE_BUILTIN_OPENCSG" on or off.

This patch removes the caching option, so it allways detects the version correctly.


nomike:
Sorry peeps. Wrong repo again. This is so annoying...

t-paul:
Well, working upstream would have benefits...

gsohler:
working downstream has other benefits ...
FYI
https://github.com/pythonscad/pythonscad/pull/64/files

nomike:
> Well, working upstream would have benefits...

I was anyway planning to submit a PR for this, as it affects OpenSCAD as well. I just didn't get to it yet as I wanted to take care of some other issues first.

PR #6084, opened 2025-08-11
Switch to clang-format and add formatting workflow

AaronVerDow:
This change removes `uncrustify` from the codebase and documentation in favor of `clang-format`, and adds a workflow for enforcement.

* Switch to `clang-format` for code formatting
* Add a GitHub workflow to enforce formatting
  * Workflow outputs a patch that can be used to resolve issues in the worst-case scenario
* Updated documentation
  * Workflow can be run locally using `act`
  * `beautify.sh` can be used as a git hook
* Added `hard_beautify.sh` that automatically beautifies code using the git workflow. This is the most accurate method of beautifying code, and can also be used as a git hook.

After this change, the code base will need to be reformatted to allow future pull requests.

https://github.com/openscad/openscad/issues/6055

Issue #6055, opened 2025-07-31
beautify workflow


AaronVerDow:
I would like to have beautify applied uniformly and enforced to make it easier to submit clean pull requests.

I created a quick GitHub worker that can run uncrustify, but there is the issue of matching uncrustify versions between the worker and the local development environment.

As a solution to this, I am looking at adding the uncrustify source as a submodule. It does not have any extra dependencies on top of what OpenSCAD already requires, and it compiles relatively quickly. This would allow a specific version to be picked for the project, and it would be applied uniformly across development environments and the GitHub worker.

Would the dev community be interested in this change?

This would be a new worker that rejects pull requests if beautify finds any errors along with a local script that can fix the errors using the exact same version of uncrustify.

AaronVerDow:
I'm also looking at getting the worker to upload a patch of the required changes as an artifact.

t-paul:
Hmm, I'm not a fan of adding submodules, but if that ends up being the easiest solution, I'm not totally against it either. So far we did not manage to automate this formatting step which means it starts drifting as soon as the latest run is through.

The other options I see:
* Use [OBS](https://build.opensuse.org/package/show/home:t-paul/OpenSCAD) to build defined packages for Ubuntu used as CI runner, just use normal apt installation. We use that for some other packages already.
* Setup a [docker image](https://github.com/openscad/docker-openscad) as runtime environment for the runs.

AaronVerDow:
I'm not familiar with OBS, but from what I'm seeing I don't see a clean path for local development. Installing a package from a custom source is a higher bar than I'd like to require.

I'm good with Docker, I just leaned away from it to avoid the extra local development requirements. However, with Docker, `act` can be used with the workflows.

I just had another idea: do not add a submodule, just have the script git clone a specific commit of uncrustify and compile that as needed. The commit version can be coded into the script to ensure it's the same for everyone.

t-paul:
Right, I was only considering the github action / CI part where we have (almost) full control over the environment. When trying to include the local development, things get a bit more difficult.

Another thing to consider is if maybe going with something else than `uncrustify` is an option. IIRC at the time the script was introduced, `clang-tidy` had some issues getting to the results expected. As that was quite some time ago, that problem may be gone, or maybe there are even other tools.

nomike:
I was just filing a ticket about exactly that topic :-).

[...]

PR #6085, opened 2025-08-11
Beautify with changes

AaronVerDow:
This is a copy of https://github.com/openscad/openscad/pull/6084 with the formatting changes applied.

https://github.com/openscad/openscad/issues/6055


AaronVerDow:
My version of clang-tools in NixOS is newer than the version used in the workflow, causing a difference. Applying the patch manually fixed the issue and the check passed.

kintel:
I'll add some exceptions using `// clang-format off` comments, especially for formatted matrix initializations and a few other places where clang-format didn't do well.

Will just push to your branch unless you have other preferences.

kintel:
One thing to mention: Different versions of clang-format may yield different formatting results. I had clang-format 19 installed, but the GitHub action seems to use clang-format 18. I updated to clang-format 20, which seems to give the same results as v18...

kintel:
@t-paul Reading through the diff is quite a bit of work, but I'm happy to accept this and deal with mishaps as we discover it later, as being able to auto-format the entire codebase has more value IMO. Any opinions on your end?

AaronVerDow:
> One thing to mention: Different versions of clang-format may yield different formatting results. I had clang-format 19 installed, but the GitHub action seems to use clang-format 18. I updated to clang-format 20, which seems to give the same results as v18...

I made this clearer in the docs and added some notes on the different versions.

kintel (approved):

kintel:
Thanks a lot for pulling this together!

Issue #4246, opened 2022-05-02
RFC - More tessellation control


zougloub:
`$fn`, `$fs`,  `$fa` are allowing to control the number of fragments in a shape; their goal is to control the deviation between the tessellated geometry and the ideal geometry, or to simplify rendering.

I would like to propose something which extends their functionality without changing the code too much, eg. using special values for them; it's a bit "dirty" but see, for example:

```C
if (fn == -1.0) {
  /*
    fs means max. error between polygon edge and ideal shape.
    eg. circle is tessellated using regular polygon with vertices on the circle;
    fs means max. error between polygon edge and circle;
    fa is unused.

    (partial implementation)
  */
  return int(ceil(M_PI / ( acos((r + fs) / r) )));
}
else if (0) {
  /*
   Leaving room for future $fn perversion here.
  */
}
else {
  /*
    Reuse good old way.
  */
  return Calc::get_fragments_from_r(r, fn, fs, fa);
}
```

t-paul:
Problems with that:
* Special case for *some* places, or even more tricky to apply to other things like fonts
* Introducing magic values is usually not good
* Interferes with automatic passing along of `$fX` values if other built-ins are supposed to keep there non-magic behavior

I'm leaving a strong vote against this kind of change. There should be better ways of giving more/finer control over tesselation.

zougloub:
I concur with your analysis @t-paul hence the RFC mention and title.

The quoted implementation of a new `get_fragments_from_r()` only handles `circle()` (and sphere/cylinder), and of course all shapes should be handled for each new convention.

I didn't formulate clearly enough that ensuing discussions would be about:
- The fact that there could be several definitions of “tessellation error”;
- Introduction of “max chord deviation considering segments on the original ideal shape” error control modality, which seems more relevant considering the goal of error control;
- Better ways than `fn/fa/fs` to control the error, vs. using magic negative values of `$fn` to encode different error modalities and up to 2 control values.

coryrc:
Why not proposing a new special `$fe` for "means max. error between polygon edge and ideal shape".

I'm going to write this PR soon, as this is exactly what I want. I want the error to be a fair bit smaller than the line width of my printer and I don't care at all what degrees or number that is.

coryrc:
> Why not proposing a new special $fe for "means max. error between polygon edge and ideal shape".

Heh, looking through the call stack, I see why now.

coryrc:
[...]

PR #6096, opened 2025-08-14
Allow the public to assign issues to themselves

coryrc:
Openscad has 893 open issues. 886 of them are not assigned.

I started work on #4246 but I can't mark myself as assigned since I am not a collaborator, but I am marked as "Contributor" so maybe there's an existing permission that allows that, but I don't think that's possible i.e. https://github.com/orgs/community/discussions/13692

Maintainers list was taken from
https://api.github.com/orgs/openscad/members?role=admin and not intended to be authoritative (and also everyone in this list can just edit assignments directly, no?).

I just tested this for OrcaSlicer at
https://github.com/coryrc/OrcaSlicer/issues/4 if you want to see it working.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -36,3 +36,5 @@ If applicable, add the content of the Help->Library Info window, or the output o
 
 **Additional context**
 Add any other context about the problem here.
+
+<!-- If you want to assign this issue to yourself, write /assign-me on its own line -->
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -18,3 +18,5 @@ A clear and concise description of any alternative solutions or features you've 
 
 **Additional context**
 Add any other context or screenshots about the feature request here.
+
+<!-- If you want to assign this issue to yourself, write /assign-me on its own line -->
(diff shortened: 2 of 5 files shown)


t-paul (changes requested):
We can give that a try. I guess this has some potential for spamming issues even with the limits, but lets go with the assumption that this will not happen.

To make it actually useful, I think it needs at least a short note in the README, otherwise it seems pretty hard to find out the feature even exists.

coryrc:
> We can give that a try. I guess this has some potential for spamming issues even with the limits, but lets go with the assumption that this will not happen.
> 
> To make it actually useful, I think it needs at least a short note in the README, otherwise it seems pretty hard to find out the feature even exists.

Assuming you mean the root readme, I'd have to make a new section. Opinions where? I was thinking under Documentation and call it `Tackling Issues`?

https://openscad.org/community.html is the only existing documentation AFAICT about filing bugs.

I did just add a commented line to the end of new issue templates.

t-paul:
Yes, I meant the README.md here in the repo as I assume most people who would want to create a PR have at least a quick look at that.

Maybe between "Getting the source code" and "Building for macOS" so it's not at the end where nobody scrolls to.

Title could be "Contributing code changes" or yes, "Tackling issues" could work too. And a note like "If you want to work on an existing issue and plan to contribute changes via a PR later, you can assign the issue to yourself by commenting `/assign-me` in that issue.".

Feel free to find some better words, English is not my native language.

coryrc:
Done.

coryrc:
Oops, I put the commit in the wrong PR; now it's here.

t-paul (approved):
Thanks for the updates. Let's see if it gets used.

PR #6098, opened 2025-08-16
Correct the message upon getting assigned an issue

coryrc:
Once this is fixed, I will send out an announcement email.

The existing message is incorrect. I'm going to send PR to the action's repo to fix it also.

Original message: https://github.com/takanome-dev/assign-issue-action/blob/5cd840e7e688c10cdc933155c3289514d32879fa/action.yml#L64

On my example PR: https://github.com/openscad/openscad/issues/4246#issuecomment-3193001559

### DIFF
--- a/.github/workflows/assign.yml
+++ b/.github/workflows/assign.yml
@@ -21,3 +21,17 @@ jobs:
           block_assignment: false
           reminder_days: 7
           max_assignments: 12
+          assigned_comment: |
+            ### 🎉 Issue Assignment Confirmation
+
+            Hey @{{ handle }}, thanks for taking on this issue!
+
+            This issue will be automatically unassigned if idle for **{{ total_days }} days**.
+
+            <details>
+            <summary>ℹ️ Important information</summary>
+
+            - To prevent automatic unassignment, a maintainer can add the **{{{ pin_label }}}** label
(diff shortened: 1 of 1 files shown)


t-paul:
You mean "idle" vs. "not closed"? What's the expectation?

Idle means, the issue gets updated with comments or anything and that will keep it alive.

Not closed means the issue would actually be closed, either because it was decided it's not going to be solved, or because it's solved in a PR that got merged. This would not allow extension of the time frame by just commenting on the issue.

t-paul:
Ah, I see in the upstream issue the plugin code indeed seems to check the updated timestamp. I'd prefer if we can get this fixed upstream instead of keeping our own overrides. Lets see how the upstream issues is moving.

coryrc:
> Ah, I see in the upstream issue the plugin code indeed seems to check the updated timestamp. I'd prefer if we can get this fixed upstream instead of keeping our own overrides. Lets see how the upstream issues is moving.

I also like this message better, this is already pinned to a particular release, and I'd like to send out an invitation for people to assign issues to themselves, but this message is overly confusing. If this is merged we can just move on and never touch this again, and start getting people to take ownership of the items needed to get the release out.

t-paul (approved):
Ok, I guess it's not that critical, so lets keep moving forward then... 🚀

PR #6099, opened 2025-08-16
Treat a malformed escape sequence as an error.  Fixes #331.

### LINKED ISSUE #331: Parser has issues with backslashes in strings [$5]
Escaped characters cause weird behavior:

a = "\n";
echo(a);

-> ECHO: "\n"

a = "\x";
echo(a);

-> ECHO: "x"

+an additional backslash on stdout
## 

<bountysource-plugin>

---
There is a **[$5 open bounty](https://app.bountysource.com/issues/271891-parser-has-issues-with-backslashes-in-strings?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jordanbrown0:
As described in #331, a malformed escape sequence - not a known backslash combination, or a malformed `\x` or `\u` - drops the backslash from the string, retains the rest, and emits the backslash on stdout(!).

This change causes a malformed escape sequence to be a parse error.

It also rearranges the string lexer rules a little bit to group them by type.

I have not run the test suite; I'd rather run it on the server.  If it passes, I think this is ready for merge.

It does *not* address some nearby ... surprising ... behaviors that I noticed:
* A newline in the middle of a string is ignored.  The string is continued on the next line.  This might have been an accident, but it allows for a continuation mechanism for strings.
* A carriage return in the middle of the string is included in the string.

Nor a not-nearby lexer issue that I noticed:
* bytes 0xc2 and 0xa0 are ignored in main-line (not-string, not-comment) text.  The comments imply that the UTF-8 sequence {0xc2, 0xa0}, U+00A0 NO-BREAK SPACE, is ignored, but the actual rule matches and ignores the two bytes independently.  Theoretically this will break hundreds of Unicode characters.  I can't think of any way that this can cause a misbehavior, because non-ASCII characters are not allowed in main-line text.


jordanbrown0:
OK, so a bunch of failures.

One of them, unicode-tests, exercises a bunch of bad cases of `\x` and `\u` and expects the backslash to be dropped (but does not check for it being emitted on stdout!).  I'd say for that one we just fix the test, either so that it doesn't trigger the errors or so that it expects the errors.  But to expect the errors would probably also require that they be non-fatal, and at the moment they are fatal.

The others all derive from [MCAD/fonts.scad line 77 et seq](https://github.com/openscad/MCAD/blob/bd0a7ba3f042bfbced5ca1894b236cea08904e26/fonts.scad#L77), which uses a variety of C escape sequences (\0\a\b\v\f\e) that are not defined in OpenSCAD, and again at 609.

Options there:
* Back off on making this an error case.
* Remove these invalid sequences from MCAD.
* Implement these C sequences.  (But would we ever implement \0?)

t-paul:
At first thought, I'd say fix MCAD. I don't see the C escapes making much sense so implementing them just for test cases feels wrong.

jordanbrown0:
To my surprise, making an undefined escape sequence be a warning and fixing the one test that explicitly exercised these cases made it come out clean.  (Except for MacOS, which is failing for some cmake reason that's pretty clearly not related to this change.)  The MCAD-based tests still generate the warnings, but apparently are not set to stop on first warning, and do not compare their text output.

We should probably still fix MCAD.  (Or support the full range of C escape sequences, to silence people who decide to do strange things.)

src/core/lexer.l
```diff
@@ -179,18 +179,27 @@ use[ \t\r\n]*"<"        { BEGIN(cond_use); LOCATION_COUNT_LINES(parserlloc, yyte
...
 \\\\                    { stringcontents += '\\'; }
 \\\"                    { stringcontents += '"'; }
-{UNICODE}               { /* parser_error_pos -= strlen(lexertext) - 1; */ stringcontents += lexertext; }
 \\x[0-7]{H}             { unsigned long i = strtoul(lexertext + 2, NULL, 16); stringcontents += (i == 0 ? ' ' : (unsigned char)(i & 0xff)); }
 \\u{H}{4}|\\U{H}{6}     { const auto c = strtoul(lexertext + 2, NULL, 16); stringcontents += str_utf8_wrapper(c).toString(); }
-[^\\\n\"]               { stringcontents += lexertext; }
-[\n\r]                  { LOCATION_ADD_LINES(parserlloc, yyleng); }
-\"                      { BEGIN(INITIAL); parserlval.text = strdup(stringcontents.c_str()); return TOK_STRING; }
-<<EOF>>                 { parsererror("Unterminated string"); return TOK_ERROR; }
+\\                      { LOG(message_group::Warning, LOCATION(parserlloc), "", "Undefined escape sequence"); }
```
kintel, line 194: Nit: Is there a way to print the bad escape sequence in the warning?

kintel:
macOS build issue is fixed by #6110

src/core/lexer.l
jordanbrown0: Hmm.  My initial answer is "yes", but on thinking about it it's not so clear.

`\xzzz` is bad, but all the lexer knows is that the upcoming characters didn't match any of the patterns.  The error message could look ahead a character, but that would say that `\x` is bad, which isn't exactly right; it isn't bad until we get to the `z`.

We could totally change how escape sequences are lexed, maybe.  We could match `\x..`, so that the lexer would match `\xzz` as one lexeme, but we could then reject the `zz` as an invalid hex value.  Similarly, we would match `\u....`, `\U......`, and `\.`, and check whether the wild-card portions are legal.

A compromise would be to have the fallback rule be to match `\.`, and then we could report the value  of the `.`.  UTF-8 would mean that we would  need to have more complicated rule that matches one, two, three, or four bytes depending on the value of the first one.  We would probably want to take care *not* to match CR or LF.  Filtering out other non-printables (including non-ASCII Unicode non-printables) would make it still trickier.

So I guess the answer is "yes, sort of", but it wouldn't be trivial to do right.

My initial response is "not worth the effort", but if you like I can take a stab at it.

kintel: Thanks. I say not worth the effort at this point

PR #6105, opened 2025-08-17
Add Pre commit config

nomike:
This patch adds a config for pre-commit.

[Pre-commit](https://pre-commit.com/) is a python tool which manages a local git hook which could perform various tests on files prior to committing them.

I've added a few common hooks for fixing trailing white-spaces, ensuring that every text-file ends with a newline, fixing yaml files and ensuring that there are no private keys committed to the code.

I also added a hook which runs clang-format to ensure files are styled correctly. It uses the projects .clang-format file of course.

`pre-commit` only scans files which have been modified, so existing files will be left untouched (unless you use the `--all-files`  option, which I would not recommend as there are plenty of error in existing files (mainly with trailing white-space characters and end of file newlines).

This is just a convenience feature though as it doesn't prevent people from pushing code which does not adhere to the style guides. One could always opt to not run `pre-commit install` and thus circumvent it's protections. Thus this will not replace github actions doing the same checks, so they are still necessary. It does tell you earlier though if something's wrong.


doc/hacking.md
```diff
@@ -1,3 +1,18 @@
+# pre-commit
```
t-paul: This section collides with the "Beautifying code" section. The two will need to be combined, otherwise it's just confusing to tell people 2 different way of handling the formatting stuff.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,38 @@
+# OpenSCAD (www.openscad.org)
+# PythonSCAD (www.pythonscad.org)
+# Copyright (C) 2009-2025 The OpenSCAD Developers.
+# Copyright (C) 2024-2025 The PythonnSCAD Developers.
```
t-paul: Typo.

doc/hacking.md
```diff
@@ -1,3 +1,18 @@
...
+
+OpenSCAD uses pre-commit which ensures that contributed code has consistent styling.
```
t-paul: At this point, I'd say "supports using pre-commit". We'll have to see how well this works across platforms and developers.

This probably also should have a link to the main website, so people don't have to search what that actually is.

doc/hacking.md
nomike: I've made "pre-commit" a subsection of "Beautifying code" and removed the previous instructions about manually setting up a hook.

I also fixed a sentence which was still mentioning uncrustify.

.pre-commit-config.yaml
nomike: Thx. Fixed.

doc/hacking.md
nomike: Fixed and addressed.

nomike:
@t-paul Thanks for the review.
I hope I addressed all your remarks well.

BTW: IMHO the `.uncrustify.cfg` should be removed.

It can always be recovered from the git history if someone wants to look up something.

t-paul:
I have it now enabled on my local repo, but did not have much chance to see it in action...

nomike:
Change something in a C file which would cause calng-format to reformat it, stage it and run `pre-commit run`.

PR #6108, opened 2025-08-18
Fully functional AI chat feature

locchh:
Dear OpenSCAD team,

I’m really impressed with your work on building an application that generates 3D objects from code. I think it would be very exciting to integrate AI to help engineers shift from traditional design to programming. This could greatly speed up and enhance the design process.

My work includes:

- Integration of OpenAI GPT models with proper API handling
- A modern chat UI with user/assistant message bubbles
- Persistent conversation history across sessions
- Robust error handling for network issues and API errors
- Settings integration for API key configuration

I’d love the opportunity to contribute my ideas to your project.  Here is demo
Screenshot from 2025-08-18 19-27-00

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1094,6 +1094,7 @@ set(GUI_SOURCES
   src/gui/Measurement.cc
   src/gui/Animate.cc
   src/gui/FontList.cc
+  src/gui/ChatWidget.cc
   src/gui/MouseSelector.cc
   src/gui/OctoPrint.cc
   src/gui/OpenCSGWarningDialog.cc
@@ -1168,6 +1169,7 @@ set(GUI_HEADERS
     src/gui/MouseSelector.h
     src/gui/Network.h
     src/gui/NetworkSignal.h
+    src/gui/ChatWidget.h
     src/gui/OctoPrint.h
     src/gui/OpenCSGWarningDialog.h
     src/gui/OpenSCADApp.h
--- a/doc/build.md
+++ b/doc/build.md
@@ -1,0 +1,75 @@
+# Building OpenSCAD from source (Ubuntu 20.04 example)
+
+This guide shows a clean, step-by-step process to build and run OpenSCAD from source in `openscad`.
+
+**Notes:**
+- On Ubuntu 20.04, we upgrade CMake and disable Manifold to avoid a oneTBB incompatibility.
+- The unified dependency script supports multiple distros; commands below focus on Ubuntu 20.04.
+
+## 1) Clone repository and submodules
+
+```bash
+git clone https://github.com/openscad/openscad.git
+cd openscad
+git submodule update --init --recursive
+```
+
+## 2) Install dependencies
+
+Use the unified installer (auto-detects distro, installs required packages):
+
+```bash
+sudo ./scripts/uni-get-dependencies.sh
+```
+
+Optional: verify versions
+
+```bash
+./scripts/check-dependencies.sh
+```
+
+Expected on Ubuntu 20.04: CGAL may be 5.0.2 (below requested 5.4). Build still works with Manifold disabled.
+
+## 3) Install modern CMake
+
+Ubuntu 20.04 ships CMake 3.16; submodules may require ≥ 3.18. Install via snap:
+
+```bash
+sudo snap install cmake --classic
+/snap/bin/cmake --version
+```
+
+**Tip**: If `/snap/bin` isn’t first in PATH, call CMake explicitly as `/snap/bin/cmake`.
+
+## 4) Configure the build
+
+Minimal, compatible configuration on Ubuntu 20.04 (Manifold OFF):
+
+```bash
+/snap/bin/cmake -S . -B build -DCMAKE_BUILD_TYPE=Release -DENABLE_MANIFOLD=OFF
+```
+
+Advanced (Manifold ON):
+- Prefer a newer OS (Ubuntu 22.04+) so the system oneTBB works, then:
+
+```bash
+/snap/bin/cmake -S . -B build -DCMAKE_BUILD_TYPE=Release
+```
+
+- Or try built-in oneTBB (may still fail on 20.04):
+
+```bash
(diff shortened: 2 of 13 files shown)


t-paul:
I would not mind seeing this feature added, even though my experiments so far with generating OpenSCAD code via LLM did not show very good results.

I see 2 points that would need to be addressed...
* Setting a fixed URL and configuration is not ideal, this should allow people to specify their own limits and give them the option to select a configuration that works for their use case
* Only supporting an external, paid service is (in my personal view) not acceptable. If we have this feature is should give at least the option to configure a local service under full control of the user. I guess the obvious candidate for that would be Ollama which is very easy to setup on the local network.

...and 2 more that are maybe nice to have:
* Allowing the user to set a system message
* Option to include a screenshot of the 3D view port with the message

jordanbrown0:
Please ensure that if the chat pane is hidden via the menu there is approximately zero impact on the UI.  (Maybe a block of preferences items.)

That's partially a knee-jerk reaction to all of the tools that have been adding an un-removable AI button to their primary UI, and partially that for reasons that I can't justify the whole thing makes me a little queasy so I want to be able to make it just go away.

kintel:
One thing to keep in mind: OpenSCAD is likely to outlive the APIs being used or the companies offering them. It would be good to somehow insulate us from upstream API changes. Ideally, but that's admittedly some work, we'd offer this via a downloadable plugin or smth.

t-paul:
Considering the OpenAI API is used as compat layer pretty much everywhere (Ollama, Azure, Github), and it's explicitly versioned in the URL, that might not be too big of a risk. Breaking the API would break a whole lot of applications.

That said, a middle ground could be having the message templates as external JSON files, but I think it would be ok to declare this out of scope for the initial PR.

nomike:
As we already have Python support in OpenSCAD it is likely quite easy to use an existing client like https://github.com/mazen160/llmquery for example.

This would give us access to a large number of LLMs at one and if an upstream API of an AI provider changes, all we have to do is to update the library as all the fine details are abstracted away from us.

But I can't really assess how feasible is to integrate this python application into OpenSCAD.
It could also be used externally. That way it would be more isolated at the expense of less tight integration of course.

In Ubuntu the library could easily be installed with `pipx`, which makes an executable available in the path.
And I can most likely add a package to GNU Guix without much of a hassle.

nomike:
I wasn't able to find any C or C++ based AI client APIs, but almost all AI stuff is written in Python, so that's not a surprise.

t-paul:
Sure, that would be possible, but it's certainly not needed. The API is just a little bit of JSON which can be easily handled by a shell script in 100 lines of code and some `curl` calls. We definitely will not do `pip install langchain` and drag in 2GB of cuda dependencies, just to call a REST endpoint.

Disclaimer: I do think langchain, llamaindex and all those *are* impressive work, but we don't need a CNC machine to drill a single hole into a wooden block either.

nomike:
> We definitely will not do pip install langchain and drag in 2GB of cuda dependencies, just to call a REST endpoint.

Good point. That's where my loosely coupled idea could help though. 

AI is quite fast paced, there are new models coming quite often, APIs change, etc..

I'm just concerned that if we hook OpenSCAD up to those APIs directly, we will constantly get tickets about "Please add support for Copilot GP5", "DeepSeek v3.2 support is broken", etc..

t-paul:
Ok, just for fun, I checked the OpenAI package, looks that's much leaner, but still installs 16 modules in a venv of ~36MB.

https://ollama.com/blog/openai-compatibility has the `curl` call as example and the PR shows that using Qt JSON and Qt Networking makes it not much more complicated.

t-paul:
Anyway, back to the PR. It looks like Qt6 did change the signals of `QNetworkReply`, so this needs some special treatment to work with Qt5 too.

I did a quick test changing the URL to a locally running Ollama (and different model obviously) and that works fine via the `http` endpoint.

It failed initially via SSL which means it's not picking up the system certificates. I guess it's not an issue with the PR itself, just a general Qt topic, but it would be nice if we could handle that too somehow.

This seems to fix the issue but I'm not sure that's the correct way of doing this. The documentation is not exactly clear.
```
QSslSocket::setDefaultCaCertificates(QSslConfiguration::systemCaCertificates());
```

kintel:
In terms of this PR, two things to point out:
1. If we make this an experimental feature, it can be merged early as we can always back it out.
2. I didn't test it, but according to the screenshot, this doesn't seem to integrate anything with OpenSCAD; it's just a chat window that happens to live in the OpenSCAD process instead of a web browser. I'd say that for this to be worth maintaining, we should integrate with OpenSCAD itself, e.g. things like auto-add open files to the context or some sort of RAG, interaction with the program (render, preview), interactive code suggestions/completion, refactoring of open files etc. If these are planned extensions, we can merge and iterate, but as a code drop I'd be hesitant.

locchh:
Thanks everyone for the constructive feedback and suggestions. After carefully listening to all opinions, I believe there are many improvements we can make to turn this feature into a major one. Regardless, I think we had a really exciting discussion.

t-paul:
@locchh I'd suggest we put this change on a branch and get it to a state for merging from there. It seems like a good base and the minimal requested changes do not seem to be a lot. What do you think about that?

locchh:
That sounds great. I’m happy to put it on a branch.

t-paul:
I can move it over from this PR, I'll comment here once it's created.

t-paul:
Hi @locchh, I put your changes into branch `ai-chat-window` associated with PR #6118. If you are interested to bring that into shape to merge, you can create PRs against this branch so the changes will appear in #6118.

t-paul:
Oh, another thing... I removed the additional `build.doc` and put that here for now: https://github.com/openscad/openscad/wiki/Building-OpenSCAD-on-Ubuntu-20.04 - having multiple descriptions in the repo is not useful, we could merge this with the readme at some point.

locchh:
Great @t-paul , thank you alot🙏

PR #6113, opened 2025-08-19
Fix compiler warnings

nomike:
This patch fixes all current warnings in OpenSCAD.
It also enables the "-Werror" compiler flag so we don't run into a state like this anymore in the future.

IMHO warnings should either be fixed. If they are intended, they should be acknowledged properly, as I did in `src/utils/StackCheck.h` for example.

I hope that looks good to you folks.


CMakeLists.txt
```diff
@@ -88,6 +88,7 @@ project(OpenSCAD
   HOMEPAGE_URL "https://openscad.org/"
   LANGUAGES C CXX
 )
+add_compile_options(-Werror)
```
kintel: `-Werror` isn't a standard compiler flag. Prefer `set(CMAKE_COMPILE_WARNING_AS_ERROR ON)`

Anyway, it might be a bit premature to enable that, as it's pretty challenging to kill all warning across OS'es, compilers and library versions.

nomike:
All test succeeded on my local machine.
[__amd-radeon-graphics-_radeonsi,_shcu_report.html.gz](https://github.com/user-attachments/files/21873610/__amd-radeon-graphics-_radeonsi._shcu_report.html.gz)

CMakeLists.txt
nomike: I agree. I've removed that last commit which added the flag for now.
Do the actual fixes look acceptable to you?

src/utils/StackCheck.h
```diff
@@ -3,6 +3,8 @@
 #include <cstdlib>
 #include "platform/PlatformUtils.h"
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdangling-pointer"
```
kintel, line 8: gcc specific - not sure it's worth putting in compiler-specific pragmas like this, since we'd have to repeat them for at least gcc, clang and msvc.

src/glview/fbo.h
```diff
@@ -22,9 +22,9 @@ class FBO
 private:
   void destroy();
 
-  bool use_ext_;
   int width_ = 0;
   int height_ = 0;
+  bool use_ext_;
```
kintel, line 27: Which warning made you do this?

src/glview/preview/OpenCSGRenderer.cc
```diff
@@ -286,8 +285,8 @@ void OpenCSGRenderer::createCSGVBOProducts(const CSGProducts& products, bool hig
     for (const auto& csgobj : product.subtractions) {
       if (csgobj.leaf->polyset) {
         const Color4f& c = csgobj.leaf->color;
-        const auto csgmode =
-          RendererUtils::getCsgMode(highlight_mode, background_mode, OpenSCADOperator::DIFFERENCE);
+        // const auto csgmode =
+        //   RendererUtils::getCsgMode(highlight_mode, background_mode, OpenSCADOperator::DIFFERENCE);
```
kintel: You can just remove this. ..and probably the entire `getCsgMode()` function.

src/gui/MainWindow.cc
```diff
@@ -2626,16 +2626,19 @@ void MainWindow::setSelection(int index)
 
   // highlight in the text editor all the text fragment of the hierarchy of object with same mode.
   for (const auto& element : nodesSameModule) {
-    if (element->index() != currentlySelectedObject) {
+    if (!currentlySelectedObject.has_value() ||
+        static_cast<int>(element->index()) != static_cast<int>(*currentlySelectedObject)) {
```
kintel: I think it would be better to simply change the return value of index() to int, as that's the actual underlying type anyway.
If that ends up in a rabbit hole, static_cast it here instead of messing with std::optional.

src/glview/fbo.h
nomike:
```plaintext
In file included from /home/nomike/coding/openscad/src/glview/fbo.cc:1:
/home/nomike/coding/openscad/src/glview/fbo.h: In constructor ‘FBO::FBO(int, int, bool)’:
/home/nomike/coding/openscad/src/glview/fbo.h:27:7: warning: ‘FBO::height_’ will be initialized after [-Wreorder]
   27 |   int height_ = 0;
      |       ^~~~~~~
/home/nomike/coding/openscad/src/glview/fbo.h:25:8: warning:   ‘bool FBO::use_ext_’ [-Wreorder]
   25 |   bool use_ext_;
      |        ^~~~~~~~
/home/nomike/coding/openscad/src/glview/fbo.cc:59:1: warning:   when initialized here [-Wreorder]
   59 | FBO::FBO(int width, int height, bool useEXT) : width_(width), height_(height), use_ext_(useEXT)
      | ^~~
```

src/utils/StackCheck.h
nomike: Yeah, I was thinking about adding a pragma for the M$ compiler as well.

It just would be very cool if we could enable `set(CMAKE_COMPILE_WARNING_AS_ERROR ON)`, because as the past has shown, warnings will get ignored quite often unfortunately.

src/glview/preview/OpenCSGRenderer.cc
nomike: Good catch, I will do that tomorrow and update this PR.

src/gui/MainWindow.cc
nomike: I will give this a shot tomorrow.

src/glview/fbo.h
kintel: ok, makes sense!

src/utils/StackCheck.h
kintel: Yeah, alternatively, run -Werror on a single platform just for Pull Requests, as that might be easier to manage.

t-paul: I would not mind having that ON as default if we have an option to disable on other platforms where we have no control over the build.
If possible it would be useful to disable on submodule builds as we don't control that code either.

src/glview/preview/OpenCSGRenderer.cc
nomike: I removed the commented code and the function.

src/gui/MainWindow.cc
nomike: I refactored this as you suggested.

src/utils/StackCheck.h
nomike: I've put an if clause around the Pragmas so they only are used on gcc.
I added additional pragmas for MSVC and clang.

Does this look okay now?

src/gui/MainWindow.cc
kintel: You changed the `index()` return type, but the (unnecessary) std::optional stuff is still here..

nomike: Oops. Fixing that now.

CMakeLists.txt
jordanbrown0: Is it possible to enable this option for particular platforms/compilers?  If so, we might do that incrementally as they are cleaned up.

nomike:
```cmake
if (CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
    add_compile_options(-Werror)
endif()
```

Something like this.

nomike:
I test-merged all of these changes into a branch in PythonSCAD, resolved some conflicts and did the same error corrections there as well.
PythonSCAD is now compiling without warnings as well (at least on Ubuntu) and all tests are green.

If you merge this MR as it is, I will just merge my branch to PythonSCAD/main.
If there are changes here, I will rebase my changes to PythonSCAD. Shouldn't be too tricky.

src/utils/StackCheck.h
kintel: I think this is fine for now. I'd prefer if we had a way of doing this without so many lines, but we can cross that bridge later.

PR #6118, opened 2025-08-21
Ai chat window

t-paul:
This is just a followup to #6108 by @locchh, putting the code into a dedicated branch on the OpenSCAD repo. See also discussion there.

Minimum changes for potential merge:
- [x] Add URL and Model selection to preferences 
- [x] Add system message to set the scene
- [x] Fix build issue due to Qt6 changes to network classes
- [ ] Add system message to preferences
- [ ] Show model selection in chat window title
- [ ] Add button to start a new chat
- [ ] Make it an optional feature, details tbd.
  - [ ] Make sure the window / menu option does not show up unless API is configured

Nice to have, could be pushed to later changes:
- [ ] Allow sending the current viewport as image
- [ ] Better formatting, QTextEdit seems to support markdown
- [ ] Make copying replies easier

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1094,6 +1094,7 @@ set(GUI_SOURCES
   src/gui/Measurement.cc
   src/gui/Animate.cc
   src/gui/FontList.cc
+  src/gui/ChatWidget.cc
   src/gui/MouseSelector.cc
   src/gui/OctoPrint.cc
   src/gui/OpenCSGWarningDialog.cc
@@ -1168,6 +1169,7 @@ set(GUI_HEADERS
     src/gui/MouseSelector.h
     src/gui/Network.h
     src/gui/NetworkSignal.h
+    src/gui/ChatWidget.h
     src/gui/OctoPrint.h
     src/gui/OpenCSGWarningDialog.h
     src/gui/OpenSCADApp.h
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1097,6 +1097,7 @@ set(GUI_SOURCES
   src/gui/ChatWidget.cc
   src/gui/MouseSelector.cc
   src/gui/OctoPrint.cc
+  src/gui/OpenAiApi.cc
   src/gui/OpenCSGWarningDialog.cc
   src/gui/OpenSCADApp.cc
   src/gui/Preferences.cc
@@ -1171,6 +1172,7 @@ set(GUI_HEADERS
     src/gui/NetworkSignal.h
     src/gui/ChatWidget.h
     src/gui/OctoPrint.h
+    src/gui/OpenAiApi.h
     src/gui/OpenCSGWarningDialog.h
     src/gui/OpenSCADApp.h
     src/gui/Preferences.h
--- a/src/core/Settings.cc
+++ b/src/core/Settings.cc
@@ -201,7 +201,6 @@ SettingsEntryEnum<std::string> Settings::printServiceFileFormat(
 
 SettingsEntryString Settings::octoPrintUrl("printing", "octoPrintUrl", "");
 SettingsEntryString Settings::octoPrintApiKey("printing", "octoPrintApiKey", "");
-SettingsEntryString Settings::openaiApiKey("api", "openaiApiKey", "");
 SettingsEntryEnum<std::string> Settings::octoPrintAction(
   "printing", "octoPrintAction",
   {{"upload", "upload", _("Upload only")},
@@ -366,6 +365,10 @@ SettingsEntryDouble Settings::axisDeadzone7("input", "axisDeadzone7", 0.0, 0.01,
 SettingsEntryDouble Settings::axisDeadzone8("input", "axisDeadzone8", 0.0, 0.01, 1.0, 0.10);
 
 SettingsEntryInt Settings::joystickNr("input", "joystickNr", 0, 9, 0);
+
+SettingsEntryString Settings::aiApiUrl("ai", "apiUrl", "");
+SettingsEntryString Settings::aiApiKey("ai", "apiKey", "");
+SettingsEntryString Settings::aiModel("ai", "model", "");
 
 SettingsEntryString SettingsPython::pythonTrustedFiles(SECTION_PYTHON, "trusted-files", "");
 SettingsEntryString SettingsPython::pythonVirtualEnv(SECTION_PYTHON, "virtual-env", "");
--- a/src/core/Settings.cc
+++ b/src/core/Settings.cc
@@ -201,6 +201,7 @@ SettingsEntryEnum<std::string> Settings::printServiceFileFormat(
 
 SettingsEntryString Settings::octoPrintUrl("printing", "octoPrintUrl", "");
 SettingsEntryString Settings::octoPrintApiKey("printing", "octoPrintApiKey", "");
+SettingsEntryString Settings::openaiApiKey("api", "openaiApiKey", "");
 SettingsEntryEnum<std::string> Settings::octoPrintAction(
   "printing", "octoPrintAction",
   {{"upload", "upload", _("Upload only")},
--- a/src/core/Settings.h
+++ b/src/core/Settings.h
@@ -410,6 +410,7 @@ public:
 
   static SettingsEntryString octoPrintUrl;
   static SettingsEntryString octoPrintApiKey;
+  static SettingsEntryString openaiApiKey;
   static SettingsEntryEnum<std::string> octoPrintFileFormat;
   static SettingsEntryEnum<std::string> octoPrintAction;
   static SettingsEntryString octoPrintSlicerEngine;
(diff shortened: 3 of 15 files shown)


kostnik:
Could you please add a CMake flag so people can completely disable this at build time, e.g. using gentoo USE flags?

t-paul:
Not sure about a compile option, those are usually for technical dependencies that are required for a feature and may not be available. At any rate, it's not going to show up or can do anything before the user actively configures it.

kostnik:
> Not sure about a compile option, those are usually for technical dependencies that are required for a feature and may not be available. At any rate, it's not going to show up or can do anything before the user actively configures it.

they can also be used to customize the build, e.g. to disable functionality that some people will never use, and it would be nice to have an option to save some disc space and RAM on this

jordanbrown0:
The vast majority of OpenSCAD users use one of the pre-built distributions, so the primary question is what is included in those.  What you can do at build time to trim down a home-made build is very much a secondary question.

kostnik:
sure, you are right

stiff:
QT6 has builtin [HTTP server](https://doc.qt.io/qt-6/qhttpserver.html), maybe it is easier to expose core functions for external agents and avoid prompt engineering and keeping AI providers API updates?

Something like:

- GET /files — List open documents
- POST /files - Create new document
- GET /files/{path} - Read source of open document
- PUT /files/{path} - Update source
- DELETE /files/{path} - (unsure), close document?
- GET /files/{path}/preview - render preview
- POST /files/{path}/camera - update camera pos
- GET /files/{path}/export?format=stl - get STL

And the plugins (MCP servers) will add support for AST navigation, partial editing, etc. Because as implemented in this PR it seems only suitable for small simple scenes.

t-paul:
> QT6 has builtin [HTTP server](https://doc.qt.io/qt-6/qhttpserver.html), maybe it is easier to expose core functions for external agents and avoid prompt engineering and keeping AI providers API updates?

Ideally the core feature would not depend on Qt, so it can run headless, e.g. via the docker container. We'll have to see if that option causes to big an effort compared to Qt.

If the core implementation is separated well enough, the features that could be exposed as MCP are identical to what the chat would use as tools directly. So the only extra stuff needed would be the HTTP server itself.

All this may also open up the LSP discussion again, the PR on that is stuck for a long time already.

stiff:
> Ideally the core feature would not depend on Qt, so it can run headless, e.g. via the docker container. We'll have to see if that option causes to big an effort compared to Qt.

There is very lightweight and simple `cpp-httplib`. If current code is not too heavily relying on Qt, then probably should be doable. And probably change slow API operations, like export, into two parts `POST export -> { id: '123' }`, `GET export?id=123 -> { status: 'pending' | status: 'done', path: '/tmp/exported.stl' }` since it's blocking

kintel:
I haven't been able to follow discussions, will try to carve out some time to follow up.

I did some experiments with HTTP hosting in OpenSCAD in the past. My main hurdle was related to who are going to talk to this server? e.g. if your client is in a browser, unencrypted connections to localhost are not very well supported, and adding SSL made things a lot less fun.

Not sure if this is relevant in this case, just mentioning it..

stiff:
cpp-httplib is specifically targeted at API clients (like MCP servers) browsers nowadays use HTTP/2, etc. But it has https support, not sure, why it might be needed at localhost, for remote non-interactive hosts just edit file and run CLI render/export.

t-paul:
Strictly needed: no. But considering some of the corporate overlords, e.g. Apple and Google decide to force https on everyone regardless of context, specifically on their mobile platforms, support for https can be useful.

jordanbrown0:
Don't even think about non-HTTPS network connections. Localhost, mostly yeah sure. (But on a shared system do you need to authenticate that you have reached the desired service, versus an imposter who grabbed the port first?)

How would this relate to the existing DBUS support?

If it's only for noninteractive use, what is the advantage over running new instances of the program? Are we really confident enough of the robustness of the program that we'd encourage running multiple projects (perhaps from mutually nontrusting users) in the same instance?

jordanbrown0:
And yes, doing HTTPS right is not at all fun.

t-paul:
> Don't even think about non-HTTPS network connections.

Why, going quite off-topic, but I don't see *any* improvements of serving our fully static HTML company homepage without cookies to people via HTTPS+Let's-Encrypt. It does not authenticate anything and transport-encryption for public information is not really giving anything extra.

Anyway. At this point it does not relate to the DBUS stuff at all, but it would certainly be an option to provide that stuff via HTTP either in addition or even replacing the (still experimental) DBUS interface.

> encourage running multiple projects (perhaps from mutually nontrusting users) in the same instance?

I don't get where this is coming from. I don't think that's suggested anywhere yet. Depending on restrictions we add regarding file access, it may become possible if multiple projects are in the same root folder and we would declare that as readable, but otherwise I think the focus is not on having multi-project and certainly not multi-user support.

stiff:
> Strictly needed: no. But considering some of the corporate overlords, e.g. Apple and Google decide to force https on everyone regardless of context, specifically on their mobile platforms, support for https can be useful.

If you listen something on loopback it's not even visible from outside that you launched non-https server (without malicious software running on your machine and reporting it).

> Don't even think about non-HTTPS network connections. Localhost, mostly yeah sure. (But on a shared system do you need to authenticate that you have reached the desired service, versus an imposter who grabbed the port first?)

HTTPS doesn't help in any way neither to auth (in simplistic implementations), nor to port grabbing.

> How would this relate to the existing DBUS support?

I see only inputs, like zoom and rotate, which is also might be good for interactivity, like "show me some deeply hidden bolt", and LLM guides your view to it, but don't think it's top priority.

UPD: Forgot to add that biggest limitation of DBUS is that it basically Linux-only, HTTP is more universal.

The use case: on one half of screen you launch OpenSCAD, on the other some agent app, and it connects to your other tools like memories, codebase etc, and say "design my PC case for 3d printing in OpenSCAD" and watch it appearing. At minimum it needs to read/write source of current scene, better it also need to access current tree (AST/nodes) + partial edits, but this API is not so obvious and may need trial and error

t-paul:
> UPD: Forgot to add that biggest limitation of DBUS is that it basically Linux-only, HTTP is more universal.

Technically not quite true due to Qt, but for practical purposes I certainly agree. Hence to comment to possibly replace the DBus stuff, as it's not yet an officially supported interface.

> better it also need to access current tree (AST/nodes) + partial edits, but this API is not so obvious and may need trial and error

Yeah, lets get *something* going first and build based on that. I suppose there are so many possibilities that it risks getting out of hand and stall completely. FWIW, the general topic is covered by the current [GSoC](https://summerofcode.withgoogle.com/programs/2026/projects/Bb5fzRQS) project, so even if this PR is going to be abandoned, we should see some results over the next couple of month.

jordanbrown0:
For local... it sort of sounded like we're talking about an OpenSCAD server, which might be interesting in server installations where there might be multiple users. If it's just a 1:1 relationship I kind of wonder about using even localhost, because that **allows** multiple clients. HTTP over a pipe seems more natural, forcing 1:1 with no rendezvous questions. (Though I don't know whether anybody easily supports HTTP over a pipe.)

On a shared system (like some server systems) there might be malicious clients or servers. Even on a system where in theory everything is owned by the same person, there is an argument for authentication for damage containment, so that a malicious process that can act as a client cannot interfere with your server process.

Truly naïve HTTPS, encryption only, is almost pointless. Less naïve HTTPS, the level typically deployed, does server authentication, proving that you have reached the server that you intend (and not an imposter), defeating both network intercepts and port stealing.

t-paul:
> For local... it sort of sounded like we're talking about an OpenSCAD server, which might be interesting in server installations where there might be multiple users.

Well, I hope we don't talk about that scenario :grin:. I agree that this is much more difficult handle correctly and I don't see that in scope any time soon(™️).

> If it's just a 1:1 relationship I kind of wonder about using even localhost, because that **allows** multiple clients. HTTP over a pipe seems more natural, forcing 1:1 with no rendezvous questions. (Though I don't know whether anybody easily supports HTTP over a pipe.)

Yeah, but that's possible with DBus too, and even may make sense for quite a number of use cases. Run different coding agents (not concurrently I hope), connect a joystick and a 3d-mouse, have Emacs control the viewport, ...

stiff:
Not sure what's the purpose of network server is, like submit source -> get .stl?

That could be done in 2 lines of python HTTP server calling OpenSCAD CLI. For me it seems that only real-time interaction makes sense as builtin API

jordanbrown0:
If it's only for communications with limited "partner" clients, HTTP (even localhost) doesn't seem like a good match.  It's too open, in both directions.  (Are you connected to the right server? Is this an expected client?) It seems like you would want an IPC scheme that at a minimum requires that the client and server belong to the same user, and in the general case arranges that one user can have multiple instances of servers and their clients. I would think of an orchestrator (perhaps embedded in a master client or server) that sets up the communications channel(s) and hands them to the various parties.

stiff:
Yes, but its todays state of IPC protocols, nothing comes close to widespread and thus simplicity of use of HTTP.

jordanbrown0:
This isn't really my area of interest, so I'm going to bow out now.

PR #6122, opened 2025-08-22
Fixes error with Hershey Draw call on windows

Sparsh-N:
DrawText resolves to DrawTextW. An alternative solution could be renaming the DrawText call itself to something similar but not a `windows.h` function.

This helps in getting a headless build working, part of #5984


Sparsh-N:
I think the alternative would work better. `DrawTextHershey()` could replace existing DrawText, and since theres only 2 references outside of hershey.cc, it isn't a major change either. Also makes the code more readable than this change currently.

kintel:
> I think the alternative would work better. `DrawTextHershey()` could replace existing DrawText

Yeah, that's my vote

kintel:
Just needs formatting; use clang-format through and IDE (e.g. clangd extension in VS Code) or run scripts/beautify.sh

Sparsh-N:
I've been trying to run the script, but it seems to have an issue whenever I run on windows. I tried using a online clang-format tool but it seems like even with that, it doesn't change much. I'll try seeing if running the script in WSL might work, as it had issues in Git Bash.

t-paul:
You can also use the patch from the action. See https://github.com/openscad/openscad/actions/runs/17186003448?pr=6122 at the bottom.

PR #6128, opened 2025-08-24
Windows/MSVC Build Script, add cairo to vcpkg dependency, and update CMakeLists

Sparsh-N:
Most of these changes were discussed during weekly meeting, although there are some other changes to the CMakeLists which are nessecary. I think these are the safest changes for now, as I'm not sure if removing the `if(not MSVC)` case for installing Lib3MF could cause issues for MSYS2 based builds and thereby the CI. For reference, the provided code works fine when I build when using vcpkg to install lib3mf.

Work towards headless build, #5984


CMakeLists.txt
```diff
@@ -223,7 +223,7 @@ else()
   set(SUFFIX_WITH_DASH "")
 endif()
 
-if (MXECROSS OR WIN32)
+if (MXECROSS OR WIN32 OR MSVC)
```
kintel: I think `MSVC` Is unnecessary: https://cmake.org/cmake/help/latest/variable/WIN32.html

kintel (approved):

PR #6131, opened 2025-08-26
Add namespace keyword and functionality

### LINKED ISSUE #522: Namespace
I have start coding namespaces.
Actually, I (successfully) use module as namespace and just allow namespace variables using.

Here is the design I try:
- parser.y no change in _expression_, use Expression with 'L' type (lookup)
- expr.cc: in 'L' evaluation, find variable inside a double array [ [ 'name', Value], ... ] that represents the namespace dictionary, and return the correspondent Value
- context.cc: virtualize lookup_variable (should be rename as lookup_name or something else)
- modulecontext.cc: inherit from lookup_variable and search for module name before variable. If found call evaluate_module for the found module
- module: add evaluate (like function) that gather assignments and return dictionary value (array of array with [name,value])

This allows things like:

``` javascript
module m() { t=5; }
mt = m.t;
echo(m.t); // outputs 5
echo(m); // ouputs [['t',5]]
```

To analyze:
- 'namespace' keyword ?
- real dictionary Value type ?

To do:
- module instantiation from namespace
- library namespace management (embedded no-namespace lib in namespace, and rename or remove (embed in global namespace) library namespace)

What do you think ?
Is the design correct ?
## 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/1053391-namespace?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

coryrc:
Adds a new keyword, `namespace`, and associated functionality:

```
namespace myName {
  x = function () 5;
  function y() = "name";
  module w() { cube($size); }
};
namespace other {
  $size = 10;
} import_specials;
// These all work!
myName::w();
echo(myName::x(), myName::y());
using myName;
w();
echo($size);
```

Open issues:
* Should `::` be the separator? I don't think it should be `.` because 1. we already have lots of environments so adding one more is normal 2. it avoids inadvertent shadowing which is annoying
* Update syntax highlighting

Resolved issues:
* Should namespaces be allowed to nest? Erlang does not. C++ does. Things can always reference other namespaces if desired. It's not really uglier to have `bosl2_threading::` vs `bosl2::threading::`
    * Uh, actually, with `using` they don't need to nest because they can pull from whichever they want.
* Does it work with `use`'d files across multiple previews? (i.e. is the caching breaking this?)
    * Namespaces don't and won't work inside use<>d files
* Should the builtins be accessible in a namespace? oooooo
    * They are! The namespaces "builtins" and "top_level" are automatically created.
* Convenience method to import some/all dynamic assignments into the top-level?
    * Specifier added `import_specials` only does all.

Possible changes for future PRs:
* `namespace name <file1.scad>;` semantically equivalent to `namespace name { include <file1.scad> };` but only parses file1.scad once
* `import_specials` allows specifying a list of variables to import, or maybe even an evaluator function to decide

Fixes #522


coryrc:
`using` is working for functions and modules. Some debug code present.

revarbat:
How in heck did I not know of this PR for a full year?

Anyways, some thoughts:
- I like `::` for the namespace separator.
- Nesting namespaces may be convenient, but if they aren't critically important.
- Preventing use of namespaces in `use<>`d files seems shortsighted, but `use<>` is kinda already broken.   `namespace name <file1.scad>;` provides a superior alternative.
- I love the idea of the "builtins" namespace.  I suppose it internally does `using builtins;`.
- import_specials seems... confusing.  For one thing, I expected that last echo to be `echo($other::size);` or `echo(other::$size);`

That last bit may need explaining. Lately, I've been fond of saying that openscad has four internal namespaces (not this PRs type of namespace): function names, module names, variable names. and $special names.  Because variables and $variables are handled vastly differently in the interpreter.

coryrc:
> How in heck did I not know of this PR for a full year?

Because the maintainers don't want to take the language in this direction. 


> - I love the idea of the "builtins" namespace.  I suppose it internally does `using builtins;`.

Indeed 


> - import_specials seems... confusing.  For one thing, I expected that last echo to be `echo($other::size);` or `echo(other::$size);`

I was also highlighting the interaction with `using`. Your latter case is the one, `other::$size`. Without `import_specials` though, it would be nil in the main context, because the scope is closed on the namespace when the main context is processed and the values of dynamic (special) variables are gone when the scope is closed. Import specials is the equivalent of this meta pseudocode if it existed at the end of a namespace definition:

For every special in this namespace's top-level: 
  main::$special = $special;

(My wording of it having to be at the end of the namespace is important, because as soon as the namespace is done being evaluated, the special values go away while lexical values stay)


> That last bit may need explaining. Lately, I've been fond of saying that openscad has four internal namespaces (not this PRs type of namespace): function names, module names, variable names. and $special names.  Because variables and $variables are handled vastly differently in the interpreter.

Importantly though, you can never have the same name as a special and non-special. The former will always start with $ the and latter never. I can see your POV though. In Common Lisp you can't tell from the variable name whether it has lexical or dynamic scope (norms are to to have asterisk as first and last character of dynamic variables though) so they both truly occupy the same namespace and you cannot have the same name in both, just like you can't with special/non-special in OpenSCAD and the opposite of how you can have the same function and variable name.

PR #6136, opened 2025-08-27
Fix a number of non-ASCII file name issues

### LINKED ISSUE #3384: Encoding Bug - Welcomescreen Open Files
Hello,
I think there is an encoding bug when you want to open a file.
If you use some special charakters (like Ä Ö Ü, etc) in the Filename, you can't open the file from the "Last used files" in the welcome screen, because the file was not found

image

image

When I open the File from the Windows explorer, there is no problem and when you open the file with the open file dialog, there is also no problem.

image

I think there is an encoding bug, because of the typical charakcter ;-)

best whises 
Fragle

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/92160701-encoding-bug-welcomescreen-open-files?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #4071: Can't save customizer json file when path contains unicode characters on windows.
With OpenSCAD version 2021.01 on windows 10 the customizer json file will not be saved when the path to the file contains unicode characters. Examples:

With 
C:\Users\Richard\Desktop\xx\Path\path2\test.scad it works as exprected

With
C:\Users\Richard\Desktop\xx\Path\path2\öäüÖÄÜß\test.scad
openscad tries to save the file as 
C:\Users\Richard\Desktop\xx\Path\path2\Ã¶Ã¤Ã¼Ã–Ã„ÃœÃŸ\test.json
which doesn't work, obviously.

When the filename is ÄÄtest.scad, openscad save the file as Ã„Ã„test.json.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/109923370-can-t-save-customizer-json-file-when-path-contains-unicode-characters-on-windows?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jordanbrown0:
[ Updated to summarize... ]

This fixes a number of issues with non-ASCII file names on Windows.  It does not fix all of them.  It is not expected to make anything **worse**.

Fixes:
* Fixes passing non-ASCII file names on the command line
* Fixes a number of places where they couldn't be opened, either to read or write.
* Fixes a number of cases where the name would get mangled in console or log output.

Does NOT fix:
* Non-ASCII **directory names** still seem to have  problems in several places.
* Several cases still turn file name on console to gibberish.  Note:  if you use Windows Command Prompt, it defaults to Code Page 437.  You might try `CHCP 65001` to set it to UTF-8 mode.
* include<> and use<> of non-ASCII names.  (Works in MSYS2 builds, but not yet in production builds.)
* Fonts with non-ASCII file names.
* Import of 3MF file with non-ASCII name.
* surface() with non-ASCII names.  (Works in MSYS2 builds, but not yet in production builds.)

These changes are not expected to affect Linux or MacOS, but it is possible that they will cause problems in non-UTF-8 locales.

For a list of tests run and their status, see tests/data/manual/UTF-8/file-names.txt.

Fixes #4071.  Fixes #3384.

Fixes some but not all of #113; it needs to be closed in favor of issues for particular problems.


jordanbrown0:
For anybody who is looking... at least for the MSYS2 build, a number of Unicode problems are related to converting to/from UTF-8 at inappropriate times.

Really, I believe, our strings should *always* be in UTF-8, except at the very periphery of the program where we interface with the system, and there and only there should they be converted to the local encoding.  (If we want to support non-UTF-8 locales at all, which I think is a good question.)

In particular, LOG() calls should *not* be doing conversions on their arguments.  Any conversions should be done down inside the LOG() infrastructure, where it writes to the outside world.  (If it does; in the GUI it writes to the console window, which is always UTF-8 capable.)

jordanbrown0:
Another, narrower, case to consider is environment variables containing non-ASCII characters.  I have not yet investigated how MSYS2/MINGW64 handle those.

jordanbrown0:
Removed the part about main() and argv having trouble, because I've done the work (I think) to make that work, inserting a wmain() that translates.

jordanbrown0:
Pending beauty tests and exercising the MXE build, this is plausible for merge.  It's certainly ready for review.

I'm pretty sure that it won't make anything worse in a pure UTF-8 environment or a pure ASCII environment, but it does seem to make many things better in an MSYS2 environment and hopefully an MXE environment.

It might make things worse in a non-UTF-8 non-ASCII environment, if anybody cares about those.  (I don't know that it does, but I also don't know that it doesn't.  I'm not in a position to test such an environment at the moment.)

I'm saying it fixes #113.  I'm not sure that it gets every case, but it certainly hits a lot of them.

jordanbrown0:
Might want some test cases, but since they would not be simple "run this program and check the results" cases - they'd have to use custom file names - I don't immediately know how to write them.  Point me at an example?

jordanbrown0:
Would want to test:
* CLI processes .SCAD file with non-ASCII name.
* CLI exports to {STL, PNG, ...?} with non-ASCII name.
* Same with non-ASCII directory name
* include<> of non-ASCII name
* use<> of non-ASCII name

Several of the fixes were to the GUI, but I don't think we have a way to automate tests of that.

kintel:
At some point, I started on a manual test script for the cmd-line parts of this, but it's pretty time consuming to pull off. This is as far as I got:
https://github.com/openscad/openscad/pull/1296/files#diff-b52973f29b16a30177324c01026e965d0081909dd9427450c84d7e3d2a33bce4

We could pull the test pieces off of that PR as the rest is dead code.

jordanbrown0:
I have wmain() in openscad.cc, under #ifdef WIN32.  It could instead be in PlatformUtils-win.cc.  Mixed feelings.  At a minimum, the normal main() would have to have a comment pointing at wmain() so that the reader knows to look there.

jordanbrown0:
Missed echo export.  Working on it, but about to run out of batteries and then probably offline for at least a day.

jordanbrown0:
Summary:

Works:
* Most things

Fails:
* surface()
* include<>
* use<>
* use&lt;font>
* non-8859-1-compatible (?) fonts from font directories

Not tested:
* MXE
* Running from CMD
* Running via file association
* Similar tests with parent directory having a non-ASCII character.

Tests extracted from PR#1296:

  Tests:
  * file loads
  * Windows title OK
  * Console load & echo OK

|   Test   |  2021.01 | master | MSYS2 | MXE | Description |
|--------|-----------|---------|---------|------|--------------|
|  1        |   -           | -           | -           | -      | **Main file** 
|  1a      |   -           | -           | -           | -      | -  *Pass on .exe command line* 
|   1a1   |                |             |              |  m   |  - - CMD
|   1a2   |                |    b       |    OK     | OK  | - - bash
|  1b      |   -           | -           | -           | -      | - *Pass on .com command line*
|   1b1   |               |              |             |  m   | - - CMD
|   1b2   |               |       b     |   OK     |  OK  |  - - bash
|   1c     |               |      b      |     x       |  OK  | - *double-click*
| 1d      | -             |  -           |  -          |  -     | - *Drag-and-drop*
| 1d1    |               |               |             |  OK  | - - on icon
| 1d2    |               |       d      |     OK   |  OK  | - - on mainwin
|  1e     |   -           | -            | -           | -      | - *Launch Screen*
| 1e1    |               |     b        |     OK   |  OK  | - - Open recent
| 1e2    |               |     b        |    OK    |   OK | - - Open
|  1f     |   -           | -             | -           | -      | - *mainwin*
| 1f1    |               |       a,d    |     OK    | OK  | - -  Open recent
| 1f2    |               |     a,d      |    OK     |  OK | - - File->Open
| 2       |               |     j          |      j       |        | **include<>**
| 3       |               |                |     k       |        | **use<>**
| 4       |               |                |              |        |  **Fonts**
| 4a     |               |      e        |      e      |   e   | - use<>
| 4b    |                |     f         |       f      |        | - from normal font directories
| 4c     |               |               |               |        | - Help/Font list
| 4d    |                |               |              |         | - Window/Font list
| 5      |                |               |              |         | **import() module / function**
| 5a    |                |     h        |      OK   |  OK  | - dxf
| 5b    |                |    h        |    OK      | OK   | - stl
| 5c    |                 |      h     |     OK     |  OK  |  - off
| 5d   |                 |    h       |   OK       |  OK   |  - JSON
| 5e   |                 |    h       |     OK     |  OK   |  - obj
| 6     |                 |             |               |          | **surface()**
| 6a   |                 |      c      |       c     |    c     | - dat
| 6b   |                |       c     |       c      |   c      | - png
| 7     |                |              |              |           | **GUI export**
| 7a   |                |     OK    |     OK     |  OK    | - dxf
| 7b   |                |      OK   |    OK      |  OK    | - svg
| 7c   |                |    OK     |      OK     |   OK   | - stl
| 7d   |                |     OK    |   OK       |   OK   | - off
| 7e   |                |      OK   |      OK    |   OK   | - amf
| 7f    |                |       g     |     OK    |  OK    | - csg
| 7g   |                |      OK   |     OK    |   OK   | - obj
| 7h   |                |    OK     |     OK    |   OK   | - pov
| 7i    |                |      OK   |     OK    |    OK  | - wrl
| 7j    |                |      OK   |     OK    |   OK   | - 3mf
| 7k   |                |      OK   |     OK    |  OK    | - pdf
| 7L   |                |       g     |    OK     |    OK  | - png
| 8     |                |              |              |           | **cmd-line export**
| 8a   |                |    g        |      OK   |   OK   | - dxf
| 8b   |                |     g       |       OK  |   OK  | - svg
| 8c   |                |      g       |      OK  |   OK   | - stl
| 8d  |                |     g        |      OK   |   OK   | - off
| 8e  |                |     g        |       OK  |   OK   | - amf
| 8f   |                |      g       |     OK    |   OK   | - echo
| 8g  |                |      g       |     OK    |  OK    | - ast
| 8h  |                |       g      |   OK      |   OK   | - term
| 8i   |                |         g    |  OK       |   OK   | - csg
| 8j   |                |       g      |  OK       |   OK   | - png
| 8k  |                |     g        |  OK       |   OK   | - 3mf
| 9    |                |      i        |    OK     |   OK   | **Save: utf8-☠.scad -> utf8-☠.scad**
| 10  |                |    i          |  OK       |   OK   | **Save As: utf8-☠.scad -> utf8-☠.scad**

Results
| Code | Description |
|-------|--------------|
| OK    |  OK
| a       | loaded OK, windows OK, but stdout "Loaded design" message corrupt
| b       |  failed to load, title corrupt 
| c       | failed to import, failure is in lodepng.cpp; it seems that fopen() can't handle UTF-8 paths.
| d       |  failed to load associated JSON file
| e       |  failed to load in FcConfigAppFontAddFile().
| f        | Limited, maybe only 8859-1 compatible names, rest silently ignored
| g       | Failed to write
| h       | Failed to import
| i        | Primary file OK, name of associated JSON file corrupted
| j        | Failed to read include file, fopen() failed
| k       | Silently fails(!); modules are just not found, because stat() fails.
| m     | Load OK, default CHCP 437 has corrupt console message, CHCP 65001 console message is OK for single-width characters, spaced wrong for double-width.
| x       | have not been able to test

All tests done with ☠U+2620 SKULL AND CROSSBONES.  It is possible that some tests may succeed with an 8859-1 compatible character (but so far I haven't noticed any).

kintel:
> Tests extracted from PR#1296:
> Results
> -	OK
> 
Are you saying all tests pass on this PR both in MSYS2 and MXE mode?

jordanbrown0:
No.  That "- OK" is all that's left from a legend listing failure codes in the table above.  The table is at the moment empty (and so has no failure codes).

Many of those tests *are* OK, but I need to methodically go through them.

jordanbrown0:
Updated the test matrix above.  As you can see, there are some serious gaps (notably include<> and use<>).  We could maybe claim "it works" with surface() and the font stuff not working, but include<> and use<> are pretty core.

Still, it's an enormous improvement over the previous situation.

I think I see how to fix include<> and use<>, but they involve more extensive changes than the others, and I think they merit their own PRs.  (They use C functions, and for whatever reason none of those work with non-ASCII.)  The font issues are, I believe, a fontconfig issue.  surface() is a problem in lodepng.

kintel:
Since all users are going to be using the MXE build, it would be nice to make sure we didn't regress anything there.

CMakeLists.txt
```diff
@@ -1387,6 +1387,10 @@ elseif(MSVC)
   )
 endif()
 
+if (WIN32)
```
kintel: I believe `-municode` is a MINGW specific flag. Should this be `if (MINGW)` ? I guess one question is what the corresponding MXE flag should be. Perhaps add a comment above this explaining why we need this?

tests/CMakeLists.txt
```diff
@@ -59,9 +60,9 @@ if(USE_IMAGE_COMPARE_PY)
   # how to utilize the venv
   find_path(VENV_BIN_PATH activate PATHS "${VENV_DIR}/bin" "${VENV_DIR}/Scripts" NO_DEFAULT_PATH NO_CACHE)
   if(WIN32)
-    set(IMAGE_COMPARE_EXE "${VENV_BIN_PATH}/python.exe")
+    set(IMAGE_COMPARE_EXE "${VENV_BIN_PATH}/python.exe -Xutf8=1")
```
kintel: Since `-Xutf8=1` is used as a Windows-specific fix, plus that you write that it's unnecessary for Python 3.15+, we should probably gate use of that flag on these two things.

jordanbrown0:
Yep, just downloaded and installed the PR's MXE build.  Haven't tested it yet.  I wanted to deal with MSYS2 issues first, because the build/test cycle is much shorter for local incremental builds.

tests/data/scad/misc/utf8-☠-2D.scad
```diff
@@ -0,0 +1,2 @@
+echo("Hello");
```
kintel: In the tests I did a long time ago, I simply echoed the filename, which helps testing the echo output, and the log makes it easier to walk through the manual test script. Did you opt out of that scheme for a reason?

jordanbrown0:
It's only a fraction of the cases, but note that I added a test case for the main file having a non-ASCII character and CLI exports to non-ASCII names, and that gets run using MXE.

src/openscad.cc
```diff
@@ -1196,3 +1196,22 @@ int main(int argc, char **argv)
 
   return rc;
 }
+
+#ifdef _WIN32
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+static std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>, wchar_t> convert_wchar;
```
kintel: This might not be the right time to address this, but the codecvt stuff was deprecated in C++17, so ideally we should try to avoid it. We still have some time though, if alternatives are annoying to establish.

src/openscad.cc
```diff
@@ -1196,3 +1196,22 @@ int main(int argc, char **argv)
...
+
+int wmain(int argc, wchar_t **argv)
+{
+  char *argv8[argc + 1];
+  for (int i = 0; i < argc; i++) {
+    argv8[i] = strdup(wchar_to_string(argv[i]).c_str());
+  }
+  argv8[argc] = NULL;
+
+  return (main(argc, argv8));
```
kintel: Should we free the strdup'ed memory after main() returns?

src/openscad_gui.cc
```diff
@@ -116,8 +116,8 @@ namespace {
 QString assemblePath(const std::filesystem::path& absoluteBaseDir, const std::string& fileName)
 {
   if (fileName.empty()) return "";
-  auto qsDir = QString::fromLocal8Bit(absoluteBaseDir.generic_string().c_str());
-  auto qsFile = QString::fromLocal8Bit(fileName.c_str());
+  auto qsDir = QString::fromStdString(absoluteBaseDir.generic_string());
```
kintel, line 120: I wonder if `genetic_u8string()` should be preferred considering all paths are now created using `u8path()`

CMakeLists.txt
jordanbrown0: Well... note that the CI builds aren't failing, and that the test works, suggesting that this works on MXE too.  I know that I thought about this a bit at the time, but I don't remember the thinking now.  But indeed I have no idea how it would fit into an MSVC build.

tests/CMakeLists.txt
jordanbrown0: 3.15 is still a year out, and presumably the option will continue to be allowed (if unnecessary) long after that.

We **can** drop it conditionally when running Python 3.15.
We **can** drop it unconditionally when we no longer support Python before 3.15.
We **must** drop it when Python drops it; we'll need to drop it conditionally if we are still supporting pre-3.15 releases.

My tentative bet is that Python won't drop support for the option until after we're ready to drop support for Python before 3.15.  If I'm wrong, we'll need to make it conditional when they drop it.  If I'm right, we can avoid making it conditional - we unconditionally add it now, and we unconditionally delete it later.

Do we even know what the Python version is?

As for Windows vs other platforms... I'm not completely sure, but I don't think these tests can work if Python is ever *not* in UTF-8 mode, on any platform.  They manipulate files with names that can't be represented in any non-Unicode encoding.  One of the key questions for this PR is how much we want to support (and test) non-UTF8 environments.

tests/data/scad/misc/utf8-☠-2D.scad
jordanbrown0: No.  I haven't looked at your stuff enough to have an opinion on it.  I'm not opposed to making it echo the file name, but here I'm not really concerned with the data **content**, just with file names.  There are existing tests for data content.

On the C++ side I didn't really try to understand what the "nowide" stuff really meant, but it seemed like it was way more pervasive than one would like, and way more pervasive than the approach that I had taken (before I saw yours).

On the testing side, I don't consider it really "done"; I just wanted to try a non-ASCII name to see if it worked, and the tests so far are little more than that initial experiment - a couple of non-ASCII names, exported into the various formats to test the "write" side.  I need to look at yours more to see if there's stuff there I should steal.

src/openscad.cc
jordanbrown0: Yeah, I looked into that a bit.  It's deprecated, but there's no replacement proposed and no schedule for actually removing it.  That suggests three alternatives:
* Just use it until there's a replacement.
* DIY the conversion.  I don't think we want to go there.
* Use the C wide/MBCS conversion functions, with the attendant conversions to C++ strings.

I decided that the existing deprecated C++ stuff was the simplest/easiest, and to extract it out into a separate function so that it can be easily replaced in the future, when and if that's necessary.

src/openscad.cc
jordanbrown0: Does anything interesting happen after main() returns?

src/openscad_gui.cc
jordanbrown0: I just left that as-is without looking at the question.

Looking at it a bit now... generic_u8string() returns std::string until C++20, and std::u8string after that.  Will that change be a compatibility hiccup?

There's a bigger question of whether, in C++20, we should use `std::string` or `std::u8string`, across the program.  My inclination (without understanding the intended distinction) is that `std::u8string` is appropriate for those very few cases where you need to deal with multiple encodings, and that everything is ... cleaner ... if the vast majority of the program does not make encoding assumptions, if encoding is only specified at the periphery.

src/openscad.cc
jordanbrown0: Generally, I'm happy never deallocating memory that's intended to be kept until the end of the program.   Deallocating during shutdown seems like complexity and an opportunity for bugs, for no real value.

CMakeLists.txt
kintel: I guess MXE technically uses Mingw under Linux. Do you know if MINGW is defined under MXE. Just trying to not break any of the WIP MSVC stuff..

jordanbrown0: Sold.  Moved it up into the existing if(MINGW) just above.  The MSVC fork there may (or may not) need to add something.

I get the impression that `wmain()` is an MS-ism, so for the moment I'm leaving it under `#if _WIN32` on the assumption that it will apply under MSVC too.  If MSVC can get the runtime to call `main()` with UTF-8 strings, it can be limited down to MINGW.

Pushed, pending confirmation that it still works with MXE.

tests/CMakeLists.txt
kintel: Python version: I would assume ${Python3_VERSION}.
I guess Python on non-Windows platforms defaults to utf8, making the setting a no-op.

tests/data/scad/misc/utf8-☠-2D.scad
kintel: The nowide stuff was a random experiment, but I abandoned it, partially because it was a bit invasive.

src/openscad.cc
kintel: Sounds good. Let's wait for the next C++ standard :)

src/openscad.cc
kintel: Just to silence memory leak checkers

src/openscad_gui.cc
kintel: Oh, didn't notice the u8string switch in C++20. Perhaps leave this for now then. Until I read more C++20 codebases, my instinct is to stick with std::string, and use std::u8string where the extra guarantees or disambiguation is valuable.

tests/CMakeLists.txt
jordanbrown0: For some reason I didn't find it earlier:  `Python_VERSION`.  But I think the analysis above is pretty compelling; if my bet is right we can avoid the complexity of making it conditional, and if it's wrong the only cost is needing to make it conditional later.

It looks like `-Xutf8=1` applies to all platforms.  [See the doc on UTF-8 mode.](https://docs.python.org/3/library/os.html#utf8-mode)

That might be a problem if somebody is trying to run the tests in a non-UTF-8 locale.  In that case, the question is whether there's any hope at all for them, since the file names can't be represented in a non-Unicode encoding.  (Setting `-Xutf8=1`  might save them, if the file system is UTF-8 but their session isn't, and if there's nothing else that pays attention to the locale's encoding.)

kintel:
Should we commit the test script (and results) too? ..to make incremental changes (and a future MSVC column) easier to manage?

tests/CMakeLists.txt
kintel: sg, let's leave it as is

src/openscad.cc
jordanbrown0: Hmm.  At work, we only run memory leak checkers on core dumps (perhaps dumped from running processes).  I don't think we even have a simple way to run one at exit time.

I bet that there are lots of places throughout libraries that leave memory allocated at exit time, but indeed this is novel in that there would be no remaining pointers to the allocated memory.  Sold, added, pushed.

Hmm:  I didn't check for the `strdup()` failing.  I could, but if `strdup()` is failing on startup we have far worse problems.  I can add checks if you like.  (Actually, if `strdup()` fails anywhere we have serious problems.  My heretical opinion is that nothing should ever check for memory allocation failures, because in practice they are not survivable.)

jordanbrown0:
If by "test script" you mean shell scripts that exercise the various cases, yes; I need to turn the manual testing that I've done into automated testing.  (Note there's already a number of test cases for these changes.)  In particular, I need to add automated import tests.

If you mean the list of things to test, for a human to follow - the list shown above - then yeah, sure.  But it needs a notation as to what's covered by the automated tests so that people don't feel the need to do them by hand.  Let's wait a bit on that until we see if there are more cases, so that I don't have to duplicate the list here and in the sources.

src/openscad.cc
```diff
@@ -1207,11 +1207,19 @@ std::string wchar_to_string(wchar_t *wcp) { return convert_wchar.to_bytes(wcp);
...
 
-  return (main(argc, argv8));
+  auto rc =  main(argc, argv8);
+
+  // Silence exit-time leak checkers.
+  for (int i = 0; i < argc; i++) {
+    free(argv8[i]);
+  }
+
+  return (rc);
```
kintel: Nit: If you want to make this easier to read, you can do this, and let the the destructors free the memory:
```
std::vector<std::string> args;
for (int i = 0; i < argc; i++) {
  args.push_back(wchar_to_string(argv[i]);
  argv8[i] = args.back().c_str();
  [...]
```

kintel:
> 
> If you mean the list of things to test, for a human to follow - the list shown above - then yeah, sure. But it needs a notation as to what's covered by the automated tests so that people don't feel the need to do them by hand. Let's wait a bit on that until we see if there are more cases, so that I don't have to duplicate the list here and in the sources.

I meant the human script :) Yeah, we can wait until it's in better shape.

src/openscad.cc
jordanbrown0: That's certainly more C++ish, but I'm not immediately convinced that it's easier to read.

With the exception of `wchar_to_string()` returning `std::string`, this is C, not C++.  The call into the function is Cish and the function that it calls is Cish; I don't think that making it more C++ish helps.

src/openscad.cc
jordanbrown0: Do you feel a need for me to check for the strdups failing?

(Note that your C++ish replacement does memory allocation too, without checking for failure... I'm sure that failure would cause something bad to happen but I don't know what.)

CMakeLists.txt
jordanbrown0: Tests passed; it still works with MXE.

jordanbrown0:
If you run openscad.exe, it emits UTF-8 that is then displayed properly.
If you run openscad.com, it runs openscad.exe and copies its output to its own stdout... and in the process emits the UTF-8 as 8859-1 (or probably Windows-1252), turning non-ASCII strings into mojibake.

I noticed this for file names in "Loaded design" messages, but it also applies to `echo()` output.

Ah, and I find this comment in winconsole.c:
```
TODO:

Fix printing of unicode on console.
```

src/openscad.cc
kintel: Yeah, not a big deal - I just dislike having to think about manual memory management :)

src/openscad.cc
kintel: We generally don't check for memory allocation errors. In C++ this usually boils down to someone throwing std::bad_alloc, but I believe we just rely on someone catching that and crashing appropriately. In your case, it will just result in a segmentation fault whenever someone accesses the data. Crash may be harder to find, but it's unlikely to happen.

kintel:
It's possible that the winconsole stuff may be better to scope as a separate PR, but you probably have a better sense of how to chop up your work into appropriate pieces. Things look good from my end, I think the main missing pieces is an MXE build sanity check and to check in the human test script.

src/openscad.cc
jordanbrown0: I thought I had a simpler answer than yours, that avoided heap allocation...
```
  char *argv8[argc + 1];
  std::string argvString[argc];

  for (int i = 0; i < argc; i++) {
    argvString[i] = wchar_to_string(argv[i]);
    argv8[i] = argvString[i].c_str();
  }
```

And now that I try it, I think I tried a variation earlier, but it (and, I believe, yours) runs into the thing that caused me to need the `strdup`:  `c_str()` returns `const char *`, and argv is (non-const) `char **`.  (I tend to think that argv should be `const char * const *`, but that horse left the barn a long time ago.)

jordanbrown0: Ah.  `data()` rather than `c_str()`.

jordanbrown0: I think the new version (which is what I quoted above, except using `data()` instead of `c_str()`) answers your concern, so I'm marking this resolved.  Un-resolve if you disagree.

kintel: oh, const. Using `data()` instead if `c_str()` should allow mutable access.

src/openscad.cc
jordanbrown0: No longer uses heap allocation.

src/openscad.cc
kintel: heh, I wasn't refreshing fast enough :)

jordanbrown0:
I think the winconsole UTF-8 thing belongs in this PR (and is already there), but I think the simplification that I'm playing with belongs in a separate PR.  (It happens that I'm building it in the same repo at the moment, but that's just for development convenience.)

kintel:
Btw., not sure if you're aware, but the MXE-built .zip file distribution ships with a runnable test framework. It probably haven't been run in a long time and may have bitrotted, but it was meant to be able to quickly run tests on Windows although the binaries were built on Linux.

jordanbrown0:
I can run the tests under MSYS2 against the MSYS2-built binary.  I haven't felt the need to run them against MXE-built binaries, because the CI build does that.

kintel:
> I can run the tests under MSYS2 against the MSYS2-built binary. I haven't felt the need to run them against MXE-built binaries, because the CI build does that.

Actually, it's the opposite: CI only runs MSYS2 tests.

jordanbrown0:
I have a change that fixes the console output under CMD, but I've just noticed that it changes the mode of that console and the change persists; it does not merely affect **this** process's output.  That seems rude and I think that change needs to be backed out.

jordanbrown0:
Tested most of the cases against an MXE build. Results were the same as for MSYS2.  There are still a couple of cases I want to exercise, and I want to add several more automated tests.

Several of the tests that I could include would fail today. That would accurately reflect today's imperfect status, and those tests would be useful as we improve the situation, but it think it would be inappropriate and distracting to have unrelated work see these currently-expected failures. Any recommendations  on how to proceed?

kintel:
I think merging with failing tests are fine. We're aiming for improvement, not one-shot perfection :)
The thing that tends to trip people up are regressions, which is why I wanted to see some MXE test.

kintel:
@jordanbrown0 Where are we at with this? I think I'm close to having MXE testing going in my other PR, if that's what you're waiting for..

jordanbrown0:
Distracted onto other things.  I'll see if I can write a few last tests, and check the manual-test checklist in somewhere.

Do we have any other manual-test checklists?

kintel:
Re manual tests, there is:
https://github.com/openscad/openscad/tree/master/tests/data/modulecache-tests
https://github.com/openscad/openscad/tree/master/tests/data/manual/issue214

..but they could be organized better.

kintel:
Btw., testing MXE builds using #6177 works now.

jordanbrown0:
Feh.  Some might remember my function `std::string wchar_to_string(wchar_t *)`.

I just stumbled across the already-existing (from PlatformUtils-win.cc) `std::string winapi_wstr_to_utf8(std::wstring)`, which is based on the Windows-specific `WideCharToMultiByte()`.

That looks plausible, and reusing something existing is a win over creating something new, but it's used ... let me count ... zero places.  A quick test indicates that it works as a replacement.  So does it *really* work?  Is it better to use a deprecated C++ mechanism, or a Windows-specific mechanism?

kintel:
I don't think it matter exactly which implementation we use, as long as it works..

jordanbrown0:
Neither the commit email notification nor the entry above include the full commit comment, so I'll make it explicit:  I flipped the coin, and it came down on the "use the existing function" side.

jordanbrown0:
Fixes some but not all of #113 - in particular, does not fix include<> with non-ASCII name.

jordanbrown0:
Fixes the very specific case in #4071 : saving Customizer data when the `.SCAD` file comes from a directory with a non-ASCII name.

That did cause me to discover a nearby case that it doesn't fix:  opening that `.SCAD` file from the command line.
```
$ b/openscad öäüÖÄÜß/test.scad

Failed to open file C:/Users/Jordan/OneDrive/linux.home/openscad/unicode/öäüÖÄÜßC:/msys64/test.scad: The filename, directory name, or volume label syntax is incorrect.
```

jordanbrown0:
I should take another pass looking through it, but although this wad doesn't fix all of the outstanding issues it fixes a bunch of them.  I would not claim that "the issue" is fixed, but it's not going to be a single particular fix; it's going to be a whole sequence of fixing individual cases.  Some are easy; others are hard.

I suggest that we merge this, close out all of the issues that it completely fixes, copy unfixed stuff from multi-case issues into new issues, and close out those multi-case issues.  Then we should try to be pretty aggressive about keeping each particular case in its own issue, so that they can be knocked down one at a time.

jordanbrown0:
Squashed out all of the intra-development commits.

kintel:
Agreed, no point in waiting until every corner case is dealt with. Post merge, we could edit the issues in question, or close them and create new issues for remaining topics. Let me know (by pressing "ready for review") when you're ready for review.

jordanbrown0:
Several of the previously "unsolved" issues were around the fact that `fopen()` doesn't handle UTF-8.

I tried just replacing `fopen()` with a function that translates its arguments to UTF-16 and calls `_wfopen()`, and it seems to fix `surface()` and `include<>` issues.  (The test run will give a hint whether it breaks anything.)

I'm not super-comfortable replacing libc functions, but I think it's more or less well-defined.  I'd be interested in other opinions.

Another possibility would be to `#define fopen myfopen`, or change the callers to call `myfopen()`, but that seems more intrusive on the components that are being pulled in more or less unmodified.

kintel:
Would it be possible to merge this _without_ dragging in some solution for fopen(), and do the remaining work separately?

Also, it appears Microsoft is slowly recognizing the error in their ways, and now supports UTF-8: https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page#set-a-process-code-page-to-utf-8

jordanbrown0:
Sure, easy enough to back out the fopen() changes.  I don't know enough about building Windows programs to know how to integrate the flag that that article discusses.  I tried `chcp 65001` and that didn't fix the problems, but that might only affect the console.

We should also check whether just setting that flag is enough to solve some or all problems.  I'd still want some of the changes here, as a form of cleanup, but we might be able to get rid of most of them.

jordanbrown0:
Backed out the fopen() changes.  Left in (via a new commit) a small change to surface() that matches the changes elsewhere.  I don't know whether that fixes any surface() issues on its own.

kintel:
Cool, let me know what you think is needed to merge, apart from a code review. 
I'm not particularly excited about doing my own manual verification, but now that we have the test checklist checked in, it feels a lot safer than before.

jordanbrown0:
Got back to looking at this a bit.  I need to methodically walk through the test cases, both on MSYS2 and MXE, and need to exercise the automated test case with MXE.

Switching to a MSYS2 UCRT64 build appears to fix the fopen/stat related issues, but that probably doesn't help MXE until we can switch it over to UCRT64.

It's quite plausible that UCRT64 would address a number of the other issues, without some of the changes made here, but they seem conservative and appropriate whether or not they are strictly necessary.

jordanbrown0:
> they seem conservative and appropriate

In particular, if we want all internal std::string processing to be in UTF-8 (and I think we do) then
* nothing "internal" should be converting QString to "local 8 bit"
* std::filesystem::u8path() explicitly converts a UTF-8 std::string to appropriate local path form (even if that's a no-op).

kintel:
Is there anything in UCRT64 which touches on the whole idea of mounting `C:\` to `/cygdrive/c/` or whatever? If so we should probably avoid it for release builds. I'm not familiar with how that is done..

jordanbrown0:
Note that UCRT64 is not an MSYS2-ism; it's a standard Windows runtime library.  The switch here is between using the old Windows runtime library MSVCRT and using the newer UCRT.

I *believe* that the mapping between the DOS-style file system layout and the UNIX-style file system layout is in the "operating system" layers of MSYS2, not in the applications themselves.  But maybe I misunderstand.  My understanding is that MSYS2 is intended to be usable to build "normal" Windows executables.

I agree that anything like that *does not* belong in the application.

Running the same binary from CMD:
```
C:\Users\Jordan\OneDrive\linux.home\openscad\master\b>.\openscad /c/users/jordan/onedrive/linux.home/openscad/master/b/x.scad

Failed to open file C:/c/users/jordan/onedrive/linux.home/openscad/master/b/x.scad: The system cannot find the path specified.
```
and from the MSYS2 shell:
```
$ ./openscad /c/users/jordan/onedrive/linux.home/openscad/master/b/x.scad

Loaded design 'C:/users/jordan/onedrive/linux.home/openscad/master/b/x.scad'.
```
But I'm not sure how that happens.  (I think that *maybe* something mucks with argv before the application sees it.)

I tried an `include<>` using a `/c/users/...` path, which works if I say `ls /c/users/...`, and it didn't work in either MSYS2 or CMD.  I tried it with a `C:/users/...`, and it worked in both.

kintel:
> Note that UCRT64 is not an MSYS2-ism; it's a standard Windows runtime library. The switch here is between using the old Windows runtime library MSVCRT and using the newer UCRT.
> 
Oh, I missed that detail. Yeah UCRT makes sense in that case.

jordanbrown0:
Tests (including new UTF-8 test) succeed on MXE build.  Now for manual tests - MXE, UCRT64.

jordanbrown0:
Updated the top comment to:

---

[ Updated to summarize... ]

This fixes a number of issues with non-ASCII file names on Windows.  It does not fix all of them.  It is not expected to make anything **worse**.

Fixes:
* Fixes passing non-ASCII file names on the command line
* Fixes a number of places where they couldn't be opened, either to read or write.
* Fixes a number of cases where the name would get mangled in console or log output.

Does NOT fix:
* Non-ASCII **directory names** still seem to have  problems in several places.
* Several cases still turn file name on console to gibberish.  Note:  if you use Windows Command Prompt, it defaults to Code Page 437.  You might try `CHCP 65001` to set it to UTF-8 mode.
* include<> and use<> of non-ASCII names.  (Works in MSYS2 builds, but not yet in production builds.)
* Fonts with non-ASCII file names.
* Import of 3MF file with non-ASCII name.
* surface() with non-ASCII names.  (Works in MSYS2 builds, but not yet in production builds.)

These changes are not expected to affect Linux or MacOS, but it is possible that they will cause problems in non-UTF-8 locales.

For a list of tests run and their status, see tests/data/manual/UTF-8/file-names.txt.

Fixes #4071.  Fixes #3384.

Fixes some but not all of #113; it needs to be closed in favor of issues for particular problems.

tests/data/scad/misc/utf8-import.scad
```diff
@@ -0,0 +1,17 @@
...
+// However, any interesting errors in these imports happen
+// during geometry processing, so we can't just be an
+// echo, csg, or ast test - we have to actually do the
+// geometry processing.
+import("../../dxf/☠-2D.dxf");
+import("../../stl/☠-3D.stl");
+import("../../off/☠-3D.off");
+import("../../obj/☠-3D.obj");
+
+// These fail because Windows fopen() can't handle UTF-8 paths.
```
kintel: ..but it only fails in MXE mode?

kintel:
Added expected files for the PDF tests you couldn't generate.

tests/data/scad/misc/utf8-import.scad
jordanbrown0: Correct, works for UCRT64.  Presumably works for Linux and MacOS.

jordanbrown0: Since we don't test MXE (yet), I'll go ahead and add those.

jordanbrown0: Note that for this test the actual resulting image isn't all that interesting; what's interesting is whether we get any error messages.  (Hmm.  I need to check whether the test actually detects error messages.)

kintel: We do run some tests with `--hardwarnings` to detect unexpected warnings, see e.g. `render-force-manifold-hardwarnings`. ..and `add_failing_test(hardwarnings` for detecting expected warnings.

kintel: ..but not necessary to solve that right now of you want to put this behind you for now :)

kintel (approved):

kintel:
Let me know when you're ready to merge.

tests/data/scad/misc/utf8-import.scad
jordanbrown0: Foo, no it doesn't.  And I don't immediately know how to test that.  This needs to be a test that does geometry processing, because that's when the errors would be detected... and I don't immediately see any examples of geometry-processing tests that have golden output to verify against.

It could indirectly test for errors by ensuring that all of the imported objects are visually separated in the resulting image, so if one was missing it'd be an image mismatch, but it would be better to be able to assert that there were no errors reported.

include<> and use<> are similarly fixed by UCRT64 (and do not work on MXE-MINGW64); I should add a test for them.

jordanbrown0:
Added include<> and use<> tests.

Need advice on testing for errors in accessing files in surface().

jordanbrown0:
Come to think of it, the same probably applies to detecting errors in accessing files in import().

kintel:
Should we write a new ticket for testing error output, or do you have a convenient place to put that?

jordanbrown0:
I don't immediately see a way to both do geometry processing **and** check for golden-master stderr/stdout content.  It does seem like we should do that - perhaps simply including an "expected" file that matches up with a geometry-processing test should trigger it - but I'd rather not take it on right now.  Inspection says that these tests are running without error, so I'd prefer to go ahead with this and leave that to another project.

(But I should note that my knowledge of the test infrastructure is very much monkey-see-monkey-do; it is entirely possible that I'm missing something.)

kintel:
Sounds good. Ideas for future reference:
* OpenSCAD can output two files at the same time: `openscad file.scad -o out.stl -o out.echo`. We could utilize that.
* If we can get OpenSCAD to actually fail (e.g. --hardwarnings), we could simply write an expected failure test, but I'm not sure how it maps here.

PR #6139, opened 2025-08-28
Rewrite uni get dependencies

nomike:
This pull request will add a fancy new get-dependencies script to OpenSCAD which is intended to replace the old `uni-get-dependencies.sh`.

As Python is standard on all Unix flavors nowadays, and as OpenSCAD uses it itself, I took the liberty to rewrite the script in Python.

I've separated the business logic from the package config.
It has support for even more Linux distributions than the old script and even supports macOS (although I'm not able to test this on anything that doesn't run in a docker container).

Overrides for specific OS versions could be specified in the config to cover situations where a package changes it's name between versions of the same distribution.

It has a "--list" and "--dry-run" option as well.

One of the design goals was to cater for PythonSCAD (and possible other forks) as well.

Therefore there is a file `pacages.json`, which lists packages common to all forks and there could be additional files for forks, which add additional packages (e.g. PythonSCAD also requires "curl" and "libfive").

These fork-specific overrides are defined in separate files, so PythonSCAD can manage it's own file without interfering with OpenSCAD in any way.

I didn't update the readme or development documentation yet, as I wanted to hear your feedback first.

The script definitely needs some more testing. I'm currently setting up some Docker containers to test it on various Linux distributions.

This might not look like an important part of OpenSCAD, but it is the first thing people come in contact with when they try to compile OpenSCAD from source, so I think it's vital to provide a good user experience here.

Let me know what you think.

### DIFF
--- a/scripts/deps/packages.json
+++ b/scripts/deps/packages.json
@@ -1,0 +1,151 @@
+{
+  "schema": 1,
+  "notes": "Base distro definitions (shared across profiles). Profiles add or extend packages without redefining managers.",
+  "distros": {
+    "debian": {
+      "manager": "apt",
+      "packages": [
+        "build-essential", "bison", "flex", "git", "curl", "cmake", "ninja-build",
+        "libffi-dev", "libboost-program-options-dev", "libboost-regex-dev", "libboost-system-dev",
+        "libmpfr-dev", "libglew-dev", "libcairo2-dev", "libharfbuzz-dev", "libeigen3-dev",
+        "libcgal-dev", "libopencsg-dev", "libgmp-dev", "imagemagick", "libfreetype-dev",
+        "libdouble-conversion-dev", "libxml2-dev", "gtk-doc-tools", "libglib2.0-dev", "gettext",
+        "xvfb", "pkg-config", "ragel", "libtbb-dev", "libgl1-mesa-dev", "libxi-dev", "libxmu-dev",
+        "libfontconfig-dev", "libzip-dev", "python3-venv",
+        "qtbase5-dev", "libqscintilla2-qt5-dev", "libqt5opengl5-dev", "libqt5svg5-dev",
+        "qtmultimedia5-dev", "libqt5multimedia5-plugins", "qt5-qmake"
+      ],
+      "version_overrides": {
+        "12": { "add": [], "remove": [] }
+      }
+    },
+    "ubuntu": {
+      "extends": "debian",
+      "version_overrides": { "24": { "add": [], "remove": [] } }
+    },
+    "fedora": {
+      "manager": "dnf",
+      "packages": [
+        "gcc", "gcc-c++", "qt5-qtbase-devel", "bison", "flex", "eigen3-devel", "harfbuzz-devel",
+        "fontconfig-devel", "freetype-devel", "boost-devel", "mpfr-devel", "gmp-devel", "glew-devel",
+        "CGAL-devel", "pkgconfig", "opencsg-devel", "git", "libXmu-devel", "curl", "ImageMagick",
+        "glib2-devel", "make", "xorg-x11-server-Xvfb", "gettext", "qscintilla-qt5-devel",
+        "mesa-dri-drivers", "libzip-devel", "ccache", "qt5-qtmultimedia-devel", "qt5-qtsvg-devel",
+        "double-conversion-devel", "tbb-devel", "libxml2-devel", "libffi-devel", "redhat-rpm-config",
+        "qtchooser"
+      ],
+      "version_overrides": {
+        "42": { "add": [], "remove": [] }
+      }
+    },
+    "arch": {
+      "manager": "pacman",
+      "packages": [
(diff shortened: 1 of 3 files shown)


t-paul:
This feels a bit over-engineered for what it's supposed to do. I guess the main concern is how to maintain as we don't have any means to test all the various distros. The simple script could be fixed by anyone just trying to compile OpenSCAD. With the python version that would only be as easy while it's about the package list.

Another point is: who is going to update and test if it works in all the build pipelines?

nomike:
IMHO the current script is quite messy and it's harder to adapt a shell script than to adapt a simple json file.
Unless a new distribution is added, that's all people ever need to touch.

PythonSCAD has a dependency on libfive and curl, OpenSCAD does not.
I thought about extending the current script to accommodate for that, but that would have made things even more messy.

And then someone comes up with Perl5SCAD (I do hope that this will never happen though....) or IDK what. And that will create total chaos.

I have seen in the past that Linux distributions change package names sometimes It's quite rare, but it happens. The current script can't handle that well.

As for the testing there are two questions coming up:

1. How can we know that the new script does the same as the old one?
2. How can we test if a new package is added to OpenSCAD whether this works on all supported plattforms?

Ad 1.:
I was anyway planning to test that thoroughly. What I will do is to modify the old script to not actually call the package management systems but just to print out the command it would run. I can then simulate various operating systems in containers, run the script and get a dictionary of {f'{os}+{version}' : package_install_command}. The new script already has functionality for specifying the OS version as command line parameters, so I don't even need containers to generate the same dictionary. And then it's just a diff.

The other part is proving whether the OS detection works well enough:
I'm sure we can find one person with a mac who can run the script once and can confirm that it detects the OS correctly.
I also have macOS in a VM running on my Ubuntu machine, so that's a second data point.

The BSDs I can test in VMs as well and for everything based on Linux I can just use Docker again.

Would such testing address your concerns in this regard?

Ad 2.:
Well, that's a problem we have already with the existing script. Plus every time a new package is added, a code change is required. which is even more risky (that's mostly the CloudOps person in me speaking).

And to your point about testing pipelines: The new script is meant to be a drop-in replacement. So if we can prove without reasonable doubt, that it functionally does the same and real-life test it on the most important platforms, that should be enough.

And due to it's extensible nature, the same script could probably be adapted to support Windows as well, where it could install dependencies using [winget](https://github.com/microsoft/winget-cli), [chocolatey](https://chocolatey.org/) or [cygwin](https://cygwin.org/).

If it's a hard "No" from you, I will probably not invest any more time in this, but if we're able to find a common ground on this, I would like to continue working on this.

Do you have any other concerns which I might be able to address?

nomike:
Thank you for your comment and feedback BTW :-).

FOSSBOSS:
> we don't have any means to test all the various distros. 
Outside of MacOS, everything else can be run in a VM for testing.

t-paul:
> Outside of MacOS, everything else can be run in a VM for testing.

Right, let me rephrase that.

The build team over the last 10 years is *me*. One person. Unpaid. Limited time. And I'd rather spend at least a couple of hours to actually improve OpenSCAD itself from now and then.

In the past people have contributed fixes to the shell script applying to the distro they care about.

Maybe at some point we could have some automation via our CI instance, but people prefer to work on forks. So it's again on my list. We'll see what the future brings, but right now the effective result is: "we don't have any means to test all the various distros".

t-paul:
Oh, as far as the PR goes, I'm going to try it out and probably merge in parallel, but having both version in the long run would be a bad idea.

nomike:
I have force-pushed some changes a few days ago with some improvements.
Does this look good? Do you have additional questions, suggestions or requests?

coryrc:
I can't speak for the people who actually have power to merge things, but I would personally get the test infrastructure merged first and running reliably, then add my change.

Personally I also don't like yaml that has any sort of operators in it; you're just defining a very ugly bespoke programming language. I'd rather you just use native Python structures if you were doing it at all.

nomike:
This PR is quite old already, but the topic might warrant to be revisited.

I've switched PythonSCAD to fully using this script and further enhanced it so it is now not only used for the Linux builds but also for macOS and Native Windows (MSYS2) builds as well.

I'm not perfectly happy with macOS though, but that's not the fault of this script.

There is this script which builds the dependencies for macOS from scratch and I had some issues with that due to non-working mirrors at some point. A handcrafted script which downloads source packages from hardcoded URLs and individually builds them is not the most reliable anyway.

So I've replaced that with `uni-get-dependencies.py` (which I renamed to `get-depenedencies.py` once Windows Support was added) which is using homebrew to install all dependencies.
Unfortunately GitHub no longer offers macOS 14 intel action runners and that makes PythonSCAD not support macOS 14.

But support for macOS 14 is running out soon and GitHub will even drop support for Intel macOS in September this year, so it's only a temporary issue. But I'm open to suggestions and ideas nonetheless.

Let me know, if you're still interested in this, in which case I can port the latest version to OpenSCAD and update this PR (or open a new one).

PR #6140, opened 2025-08-29
Add readme/instructions for MSVC based builds

Sparsh-N:
This PR adds the instructions for building with MSVC in a separate markdown file, but is referenced in the top level README as well. I could further mention how to build the GUI mode separately.


README.md
```diff
@@ -225,6 +225,9 @@ OpenSCAD for Windows is usually cross-compiled from Linux. If you wish to
 attempt an MSVC build on Windows, please see this site:
 https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/Building_on_Windows
 
+MSVC build support has been added to OpenSCAD. For instructions on how to build it,
+refer to [development Nix shell](doc/win-build.md)
```
kintel: Wrong link title

doc/win-build.md
```diff
@@ -0,0 +1,37 @@
+# Build using Visual Studio and vcpkg
+
+Follow these steps to build OpenSCAD with Microsoft Visual Studio:
+
+- Make sure you have git installed and in your PATH.
```
kintel: Perhaps link to a recommended way to install git

README.md
Sparsh-N: Updated the link in latest commit

doc/win-build.md
Sparsh-N: Added the default way, although I also added a section for adding the variable to path, as that might be a bit more confusing. I might want to specify where they would find the Path for these variables, thoughts?

kintel (approved):

PR #6142, opened 2025-08-29
Refactor: switch LocalScope usage from mix of stuff to smart pointers

coryrc:
The code is grabbing a pointer from a unique_ptr and then fobbing it off random places; that's definitely not the right way to use unique_ptr...

This standardizes on shared_ptr everywhere for usage of LocalScope.

### DIFF
--- a/src/core/parser.y
+++ b/src/core/parser.y
@@ -73,7 +73,7 @@ int lexerlex_destroy(void);
 int lexerlex(void);
 static void handle_assignment(const std::string token, Expression *expr, const Location loc);
 
-std::stack<LocalScope *> scope_stack;
+std::stack<std::shared_ptr<LocalScope>> scope_stack;
 SourceFile *rootfile;
 
 extern void lexerdestroy();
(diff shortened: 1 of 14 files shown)


kintel:
I think this should be fine. I believe the reason the LocalScope stuff wasn't resolved last cleanup was that I was uncertain if I needed to break up any potential reference cycles by using weak_ptr to LocalScope, and I didn't want to deal with that at the time.

src/core/parser.y
```diff
@@ -73,7 +73,7 @@ int lexerlex_destroy(void);
 int lexerlex(void);
 static void handle_assignment(const std::string token, Expression *expr, const Location loc);
 
-std::stack<LocalScope *> scope_stack;
+std::stack<std::shared_ptr<LocalScope>> scope_stack;
```
kintel, line 76: When we do the final pop() in `parse()`, I wonder if we should assert that this stack is empty, just in case a scope somehow leaks.

PR #6144, opened 2025-08-29
Refactor looking up function and module definitions from LocalScope to use methods

### LINKED ISSUE #6134: Refactor lookup_local_* methods to take environment as argument and don't share LocalScope internals all over
```
boost::optional<CallableFunction> ScopeContext::lookup_local_function(const std::string& name,
                                                                      const Location& loc) const
{
  const auto& search = scope->functions.find(name);
  if (search != scope->functions.end()) {
    return CallableFunction{CallableUserFunction{get_shared_ptr(), search->second.get()}};
  }
}

boost::optional<InstantiableModule> ScopeContext::lookup_local_module(const std::string& name,
                                                                      const Location& loc) const
{
  const auto& search = scope->modules.find(name);
  if (search != scope->modules.end()) {
    return InstantiableModule{get_shared_ptr(), search->second.get()};
  }
  return Context::lookup_local_module(name, loc);
}
```

How a scope stores environments should be an internal detail and not accessed like this.

Also LocalScope should have a method somewhat like `boost::optional<T> searchEnvs(const std::string& name, const Location& loc) const` instead of `scope->modules.find(name)` being used.

So, the two above lookup_local should be combined to something like:

```
boost::optional<T> ScopeContext::lookup_local(const std::string& name,
                                                                      const Location& loc) const
{
  auto found = scope<T>searchEnvs(name);
  if found {
    switch T {
      // do returns based on requested return
    }
  }
  return Context::lookup_local<T>(name, loc);
}
```

Obviously I'm not a C++ expert but this gets the idea. Normally I'd use an enum but I'm assuming C++ polymorphism covers this.

Making this change dedupes function and module lookups, which makes adding first-class modules cleaner.

coryrc:
It's generally not good design to let other classes mess with a class's (internal) fields. This abstracts out looking up functions and definitions that exist in a scope, which makes cleaner code.

This is a pure refactor and contains no behaviour changes.

Extracted from #6131

Fixes #6134
Also includes #6142

### DIFF
--- a/src/core/Children.h
+++ b/src/core/Children.h
@@ -14,7 +14,7 @@ class ScopeContext;
 class Children
 {
 public:
-  Children(const LocalScope *children_scope, std::shared_ptr<const Context> context)
+  Children(const std::shared_ptr<LocalScope> children_scope, std::shared_ptr<const Context> context)
     : children_scope(children_scope), context(std::move(context))
   {
   }
@@ -37,7 +37,7 @@ public:
   [[nodiscard]] const std::shared_ptr<const Context>& getContext() const { return context; }
 
 private:
-  const LocalScope *children_scope;
+  const std::shared_ptr<LocalScope> children_scope;
   std::shared_ptr<const Context> context;
 
   [[nodiscard]] ContextHandle<ScopeContext> scopeContext() const;
--- a/src/core/LocalScope.cc
+++ b/src/core/LocalScope.cc
@@ -42,6 +42,30 @@ void LocalScope::addAssignment(const std::shared_ptr<Assignment>& assignment)
   this->assignments.push_back(assignment);
 }
 
+template <typename T>
+boost::optional<T> LocalScope::lookup(const std::string& name) const
+{
+  return boost::none;
+}
+
+template <>
+boost::optional<UserFunction*> LocalScope::lookup(const std::string& name) const {
+  const auto& search = this->functions.find(name);
+  if (search != this->functions.end()) {
+    return search->second.get();
+  }
+  return boost::none;
+}
+
+template <>
+boost::optional<UserModule*> LocalScope::lookup(const std::string& name) const {
+  const auto& search = this->modules.find(name);
+  if (search != this->modules.end()) {
+    return search->second.get();
+  }
+  return boost::none;
+}
+
 void LocalScope::print(std::ostream& stream, const std::string& indent, const bool inlined) const
 {
   for (const auto& f : this->astFunctions) {
--- a/src/core/LocalScope.h
+++ b/src/core/LocalScope.h
@@ -40,12 +40,12 @@ public:
   AssignmentList assignments;
   std::vector<std::shared_ptr<ModuleInstantiation>> moduleInstantiations;
 
+private:
   // Modules and functions are stored twice; once for lookup and once for AST serialization
   // FIXME: Should we split this class into an ASTNode and a run-time support class?
   std::unordered_map<std::string, std::shared_ptr<UserFunction>> functions;
   std::unordered_map<std::string, std::shared_ptr<UserModule>> modules;
 
-private:
   // All below only used for printing:
   std::vector<std::pair<std::string, std::shared_ptr<UserModule>>> astModules;
   std::vector<std::pair<std::string, std::shared_ptr<UserFunction>>> astFunctions;
--- a/src/core/LocalScope.h
+++ b/src/core/LocalScope.h
@@ -8,6 +8,7 @@
 #include <memory>
 #include <string>
 #include <vector>
+#include <boost/optional.hpp>
 
 class AbstractNode;
 class Context;
@@ -28,14 +29,30 @@ public:
   void addAssignment(const std::shared_ptr<class Assignment>& assignment);
   bool hasChildren() const { return !(moduleInstantiations.empty()); }
 
+  /**
+   * @brief Search corresponding environment of name for type
+   *
+   * FYI can only find `function x()` not `x = function ()`
+   */
+  template <typename T>
+  boost::optional<T> lookup(const std::string& name) const;
+
   AssignmentList assignments;
   std::vector<std::shared_ptr<ModuleInstantiation>> moduleInstantiations;
 
   // Modules and functions are stored twice; once for lookup and once for AST serialization
   // FIXME: Should we split this class into an ASTNode and a run-time support class?
   std::unordered_map<std::string, std::shared_ptr<UserFunction>> functions;
-  std::vector<std::pair<std::string, std::shared_ptr<UserFunction>>> astFunctions;
-
   std::unordered_map<std::string, std::shared_ptr<UserModule>> modules;
+
+private:
+  // All below only used for printing:
   std::vector<std::pair<std::string, std::shared_ptr<UserModule>>> astModules;
+  std::vector<std::pair<std::string, std::shared_ptr<UserFunction>>> astFunctions;
 };
+
+template <>
+boost::optional<UserFunction*> LocalScope::lookup(const std::string& name) const;
+
+template <>
+boost::optional<UserModule*> LocalScope::lookup(const std::string& name) const;
--- a/src/core/ScopeContext.cc
+++ b/src/core/ScopeContext.cc
@@ -61,7 +61,7 @@ boost::optional<InstantiableModule> ScopeContext::lookup_local_module(const std:
 UserModuleContext::UserModuleContext(const std::shared_ptr<const Context>& parent,
                                      const UserModule *module, const Location& loc, Arguments arguments,
                                      Children children)
-  : ScopeContext(parent, &module->body), children(std::move(children))
+  : ScopeContext(parent, module->body), children(std::move(children))
 {
   set_variable("$children", Value(double(this->children.size())));
   set_variable("$parent_modules", Value(double(StaticModuleNameStack::size())));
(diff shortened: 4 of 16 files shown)


src/core/LocalScope.cc
```diff
@@ -42,6 +42,30 @@ void LocalScope::addAssignment(const std::shared_ptr<Assignment>& assignment)
   this->assignments.push_back(assignment);
 }
 
+template <typename T>
+boost::optional<T> LocalScope::lookup(const std::string& name) const
+{
+  return boost::none;
+}
```
kintel: A default specialization is not necessary - let's just make the compiler complain if someone attempts to go there

src/core/LocalScope.h
```diff
@@ -28,14 +29,30 @@ class LocalScope
...
   std::unordered_map<std::string, std::shared_ptr<UserModule>> modules;
+
+private:
+  // All below only used for printing:
   std::vector<std::pair<std::string, std::shared_ptr<UserModule>>> astModules;
+  std::vector<std::pair<std::string, std::shared_ptr<UserFunction>>> astFunctions;
 };
+
+template <>
+boost::optional<UserFunction*> LocalScope::lookup(const std::string& name) const;
```
kintel: We should probably prefer `std::optional`. I know there is still a bunch of boost::optional in the codebase (and we should clean that up), but new code uses std::optional.

src/core/LocalScope.cc
```diff
@@ -42,6 +42,30 @@ void LocalScope::addAssignment(const std::shared_ptr<Assignment>& assignment)
...
+
+template <>
+boost::optional<UserFunction*> LocalScope::lookup(const std::string& name) const {
+  const auto& search = this->functions.find(name);
+  if (search != this->functions.end()) {
+    return search->second.get();
+  }
+  return boost::none;
```
kintel: FYI: If you don't want to have to think about what exact value to use, you can just `return {};`. Just a matter of taste really.

src/core/ScopeContext.cc
```diff
@@ -41,19 +41,19 @@ void ScopeContext::init()
 boost::optional<CallableFunction> ScopeContext::lookup_local_function(const std::string& name,
                                                                       const Location& loc) const
 {
-  const auto& search = scope->functions.find(name);
-  if (search != scope->functions.end()) {
-    return CallableFunction{CallableUserFunction{get_shared_ptr(), search->second.get()}};
+  auto defined = scope->lookup<UserFunction*>(name);
```
kintel: const?

src/core/ScopeContext.cc
```diff
@@ -87,13 +87,17 @@ boost::optional<CallableFunction> FileContext::lookup_local_function(const std::
   for (const auto& m : source_file->usedlibs) {
     // usedmod is nullptr if the library wasn't be compiled (error or file-not-found)
     auto usedmod = SourceFileCache::instance()->lookup(m);
-    if (usedmod && usedmod->scope->functions.find(name) != usedmod->scope->functions.end()) {
-      ContextHandle<FileContext> context{Context::create<FileContext>(this->parent, usedmod)};
+    if (usedmod) {
+      if (auto defined = usedmod->scope->lookup<UserFunction*>(name)) {
```
kintel: const?

src/core/LocalScope.cc
coryrc: Done, I didn't think that would work

src/core/LocalScope.h
coryrc: Done

src/core/LocalScope.cc
coryrc: Thanks, done.

src/core/ScopeContext.cc
coryrc: done and the others

src/core/ScopeContext.cc
coryrc: done and the other

kintel:
Re. "Also includes https://github.com/openscad/openscad/pull/6142" - I generally prefer opening and merging isolated PRs, at it makes reviews easier and allows us to defer smaller topics needing discussion without holding back necessary cleanup.
But this is clean enough to merge now.
Could you close #6142 if you prefer this to superseded it?

coryrc:
> Re. "Also includes #6142" - I generally prefer opening and merging isolated PRs, at it makes reviews easier and allows us to defer smaller topics needing discussion without holding back necessary cleanup. But this is clean enough to merge now.

I wasn't sure it'd be wanted.

> Could you close #6142 if you prefer this to superseded it?

Done. I added the requested assert from that PR in here.

coryrc:
Also, re: #6142 coments I am unsure if there's ownership cycles; I didn't think about that and I could remove that commit from here. I just included it here because I thought it'd be annoying to rebase despite actually being orthogonal changes and I thought the original PR would go in quicker than this one.

kintel:
I think the pointer change makes sense. There is a slight risk of cycles, but let's cross that bridge later; we'd have to run/automate some leak checks - it's better to prove it's a problem than to hope it's not :)

coryrc:
Agreed; I was hoping for something simple which could be enabled which would solely print the number of shared_ptr which exist. It appears valgrind or asan might be it but also catch lots of other issues (which, probably good, but also not pinpointed on this exact problem).

kintel:
Intuitively I'd say let set up asan and msan, as a start as an option to easily run locally and as a nightly run on the master branch. Then we can look into enforcing it later once/if they become clean enough.

coryrc:
Aha, I have an easy ad-hoc way: increment and decrement global counter in LocalScope constructor and destructor, respectively. Then run preview a bunch of times and see if counter goes up. Do you want me to do that before merging this PR?

kintel:
> Aha, I have an easy ad-hoc way: increment and decrement global counter in LocalScope constructor and destructor, respectively. Then run preview a bunch of times and see if counter goes up. Do you want me to do that before merging this PR?

That feels a bit too ad-hoc to me. How about do it locally as a sanity check, but not commit it? ..then we can look into the asan route later?

coryrc:
Oh yeah, I wasn't going to commit.

coryrc:
Okay, nope, it's a bad change. I'm going to remove the commit. I added:

```
unsigned int ls_counter=0;

LocalScope::LocalScope() {
  if (++ls_counter%1000 == 0) {
    std::cerr << "Localscope+" << std::setw(4) << std::setfill('0') << ls_counter << std::endl;
  }
}
LocalScope::~LocalScope() {
  if (--ls_counter%10 == 0) {
    std::cerr << "Localscope-" << std::setw(4) << std::setfill('0') << ls_counter << std::endl;
  }
}
```

Hit F5 a few times and only got +, only got some - when quitting:

```
Localscope+1000
Localscope+2000
Localscope+3000
Localscope+4000
Localscope+5000
Localscope+6000
Localscope+7000
Localscope+8000
Localscope+9000
Localscope+10000
Localscope+11000
Localscope+12000
Localscope-12670
Localscope-12660
Localscope-12650
Localscope-12640
Localscope-12630
Localscope-12620
Localscope-12610
Localscope-12600
Localscope-12590
...
Localscope-10150
```

coryrc:
Uh-oh @kintel it's not my change, it's in HEAD. I put 
[localscope_counter.patch](https://github.com/user-attachments/files/22196662/localscope_counter.patch) on top of commit [50fd7a31e5e66d087be023ad640e97a778d48ea1](https://github.com/openscad/openscad/commit/5d6e37dd177d9d9329234cb5d3c0491ab0f23dcd) and it has the same effect: LocalScope count continues to grow every time I hit F5.

RSS grows about 25kB on every preview, which sounds in the ballpark for parsed data structures (my project includes bosl2).

coryrc:
So I'd like to get this PR merged; do you want me to remove the smart pointer commit or can we merge them all and deal with the memory leak from a more solid base of being able to put weak refs?

kintel:
> Uh-oh @kintel it's not my change, it's in HEAD. I put 
> 
Ouch. I'll take a look.

src/core/Children.h
```diff
@@ -37,7 +37,7 @@ class Children
   [[nodiscard]] const std::shared_ptr<const Context>& getContext() const { return context; }
 
 private:
-  const LocalScope *children_scope;
+  const std::shared_ptr<LocalScope> children_scope;
```
kintel: Re-reading this PR: The LocalScope is meant to be const here, no its pointer, so I believe that here (and other places) you'd want to use `shared_ptr<const ...>` rather than `const shared_ptr<...`

kintel:
> > Uh-oh @kintel it's not my change, it's in HEAD. I put
> 
> Ouch. I'll take a look.

I pushed a fix for that - it looks like it was a leak of SourceFile, plugged by using a shared_ptr :)

src/core/Children.h
coryrc: Aha! Though apparently it can also be a const shared_ptr in most places FWIW. Done.

coryrc:
> > > Uh-oh @kintel it's not my change, it's in HEAD. I put
> > 
> > 
> > Ouch. I'll take a look.
> 
> I pushed a fix for that - it looks like it was a leak of SourceFile, plugged by using a shared_ptr :)
> Let's fix the const issue before merging.
> 
> Btw., try to rerun your leak analysis again after my fix.

Const issue fixed, leak check now drops back down to ~2560 LocalScope objects in existence while spamming F5.

src/core/Children.h
kintel: Yeah, the main thing to think about is that the object itself should not be changeable in most places. ..and in some places, the pointer itself can be constant too.

kintel:
Sounds good!

kintel (approved):

PR #6157, opened 2025-09-03
SourceFileCache: Rename and reword away from compile/evaluate to parse/process

coryrc:
SourceFileCache does not compile or evaluate anything; it is parsing. Make this clear with the comments and method names.


kintel (approved):

kintel:
I really appreciate refactoring to clarify code. Especially green refactoring as separate PRs instead of lumping it in with a functional change!

PR #6164, opened 2025-09-05
Reduce header pollution

coryrc:
A problem all started when I couldn't dynamic_cast a BuiltinContext
and apparently header pollution can be the cause or trigger. This is
just a good start toward cleaning up.  Several
cc files had to be updated because of indirect includes that are now missing.

Though it turned out it was a typo and I was comparing the
value-to-be-assigned for nullptr and not the
value-from-dynamic-cast. C'est la vie.

### DIFF
--- a/src/core/Arguments.h
+++ b/src/core/Arguments.h
@@ -8,7 +8,9 @@
 #include <boost/optional.hpp>
 
 #include "core/Assignment.h"
-#include "core/Context.h"
+#include "core/Value.h"
+
+class EvaluationSession;
 
 struct Argument {
   boost::optional<std::string> name;
@@ -45,7 +47,7 @@ public:
   [[nodiscard]] Arguments clone() const;
 
   [[nodiscard]] EvaluationSession *session() const { return evaluation_session; }
-  [[nodiscard]] const std::string& documentRoot() const { return evaluation_session->documentRoot(); }
+  const std::string& documentRoot() const;
 
 private:
   EvaluationSession *evaluation_session;
--- a/src/core/Context.cc
+++ b/src/core/Context.cc
@@ -33,8 +33,50 @@
 #include <vector>
 
 #include "core/AST.h"
+#include "core/EvaluationSession.h"
 #include "core/function.h"
 #include "utils/printutils.h"
+
+// Just for explicit template instantiations:
+#include "core/BuiltinContext.h"
+#include "core/ScopeContext.h"
+
+template <typename T>
+ContextHandle<T>::ContextHandle(std::shared_ptr<T>&& context)
+  : ContextFrameHandle(context.get()), context(std::move(context))
+{
+  try {
+    this->context->init();
+  } catch (...) {
+    session->contextMemoryManager().addContext(std::move(this->context));
+    throw;
+  }
+}
+
+template <typename T>
+ContextHandle<T>::~ContextHandle()
+{
+  assert(!!session == !!context);
+  if (session) {
+    session->contextMemoryManager().addContext(std::move(this->context));
+  }
+}
+
+template <typename T>
+ContextHandle<T>& ContextHandle<T>::operator=(ContextHandle&& other) noexcept
+{
+  assert(session);
+  assert(context);
+  assert(other.context);
+  assert(other.session);
+
+  // session->contextMemoryManager().releaseContext();
+  session->contextMemoryManager().addContext(std::move(this->context));
+  other.release();
+  context = std::move(other.context);
+  ContextFrameHandle::operator=(context.get());
+  return *this;
+}
 
 Context::Context(EvaluationSession *session) : ContextFrame(session), parent(nullptr) {}
 
@@ -153,3 +195,10 @@ std::string Context::dump() const
   return s.str();
 }
 #endif  // ifdef DEBUG
+
+// Alternative choice is to move the implementation to the header.
+template class ContextHandle<FileContext>;
+template class ContextHandle<BuiltinContext>;
+template class ContextHandle<UserModuleContext>;
+template class ContextHandle<ScopeContext>;
+template class ContextHandle<Context>;
--- a/src/core/Arguments.cc
+++ b/src/core/Arguments.cc
@@ -28,7 +28,9 @@
 
 #include <ostream>
 #include <memory>
+#include "core/Context.h"
 #include "core/Expression.h"
+#include "core/EvaluationSession.h"
 
 Arguments::Arguments(const AssignmentList& argument_expressions,
                      const std::shared_ptr<const Context>& context)
@@ -49,6 +51,11 @@ Arguments Arguments::clone() const
     output.emplace_back(argument.name, argument.value.clone());
   }
   return output;
+}
+
+const std::string& Arguments::documentRoot() const
+{
+  return evaluation_session->documentRoot();
 }
 
 std::ostream& operator<<(std::ostream& stream, const Argument& argument)
--- a/src/core/BuiltinContext.h
+++ b/src/core/BuiltinContext.h
@@ -2,8 +2,9 @@
 
 #include <string>
 
-#include "core/AST.h"
 #include "core/Context.h"
+
+class Location;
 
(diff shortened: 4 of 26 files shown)


src/core/Arguments.h
```diff
@@ -8,7 +8,9 @@
 #include <boost/optional.hpp>
 
 #include "core/Assignment.h"
-#include "core/Context.h"
+#include "core/Value.h"
+
+class EvaluationSession;
```
kintel, line 13: While forward declarations has the potential for speeding up compilation, I'm uncertain how much value that really has these days vs. the value you get from the compiler knowing about all our types.

There is a bunch of existing code using a mixture of forward declaration and includes for similar situations; we haven't gotten around to modernizing that yet.

If you have any opinions on the matter I'd love to hear them.

src/core/Context.cc
```diff
@@ -153,3 +195,10 @@ std::string Context::dump() const
   return s.str();
 }
 #endif  // ifdef DEBUG
+
+// Alternative choice is to move the implementation to the header.
+template class ContextHandle<FileContext>;
+template class ContextHandle<BuiltinContext>;
+template class ContextHandle<UserModuleContext>;
+template class ContextHandle<ScopeContext>;
+template class ContextHandle<Context>;
```
kintel: I don't quite understand the motivation for moving this here and explicitly instantiating it.

If we want to limit the possible template parameters, we could use a `static_assert(std::is_base_of<>)` during construction (or a more complex `enable_if` SFINAE, but I strongly prefer static_assert).

kintel (commented):
FYI: We have a [clang-tidy checker](https://github.com/openscad/openscad/actions/runs/17486035508/job/49665112181?pr=6164), which checks some of the stuff you're cleaning up. It's currently just informationan, until we have a clean enough state to raise a flag. You PR fixes 4 known issues:

Screenshot 2025-09-05 at 09 39 01

jordanbrown0:
I'm not following this in detail, but one thing caught my eye:
>While forward declarations has the potential for speeding up compilation, I'm uncertain how much value that really has these days vs. the value you get from the compiler knowing about all our types.

If you can **avoid** fully declaring  something (a class, a function, et cetera) in all contexts, that has three benefits that I can think of:
* Makes it more explicit when you want to "import" something.  An explicit `#include <semiprivate.h>` can raise a yellow flag in a reviewer's head, where just quietly using a function/class/member that was not intended for wide use might not.  This may help avoid leaking implementation details inappropriately.
* Changing the implementation details doesn't force a recompilation of numerous files that are not actually affected by the change.
* ... and compilation performance might improve because it just doesn't have to look at as much.

But it's the first two that I would really look at. As Cory has noted, there's a lot of tendrils of implementation detail that leak throughout the program; maybe there would be fewer if those leaks were a little harder and more obvious.  I hate touching Value.h, because it forces recompiling the world.  That's less of an issue now that I have a new hot desktop, but if I'm on my laptop that's a long time.

Including the full definition allows inlining, but except for the hottest of functions inlining doesn't help much.

src/core/Arguments.h
coryrc: The bigger thing is you don't need to recompile everything including Arguments.h just because you changed some private field in EvaluationSession. (oops you know this)

coryrc: Knowing about the type in the header doesn't change the actual code in Arguments.cc, because it does know about the actual type.
And other code which doesn't use EvaluationSession but does use Arguments doesn't change because it was never generating code which calls EvaluationSession.
... right?
hmm, Jordan points out possible inlining, but I think the hot paths are going to prove to be long chains and not single function calls. Particularly a fully-evaluated namespace can have all lexical variable lookups saved instead of searching every single time.

```
namespace a {
  x = 7;
  module y() {
    module z() {
      module w() {
        cube(x);
      }
      w();
    }
    z();
  }
}
for (i=[0,1000]) y();
```

Every call to y() involves looking up x in the contexts for w, z, y, then namespace all 1000 times, which inlining won't help, because we could fix because namespace evaluation is done before evaluating top_level.

src/core/Context.cc
coryrc: Because it fails during linking if we don't; previously I believe Context.o happened to create all the needed templated implementations because the headers for Context.h included these usages of ContextHandle (via BuiltinContext.h and ScopeContext.h), but now that all those usages were moved to the respective .cc files, compiling Context.o can't see them anymore and doesn't know to create them.

coryrc:
> FYI: We have a [clang-tidy checker](https://github.com/openscad/openscad/actions/runs/17486035508/job/49665112181?pr=6164), which checks some of the stuff you're cleaning up. It's currently just informationan, until we have a clean enough state to raise a flag. You PR fixes 4 known issues:
> Screenshot 2025-09-05 at 09 39 01

Cool! I felt pretty good until I saw there are currently 3000 issues :rofl:

src/core/Context.cc
kintel: Yeah, that much is clear to me. I'm questioning the value of moving the implementation out of the header.
I very seldomly see this pattern used for built-time optimization as it makes it harder (IMO) to read the code.

coryrc: Oh, it wasn't in the header before. I could move it. I'm not a C++ expert.

coryrc: Oh, wait, I get it now. Umm, let me think.

coryrc: This way means only one set of templated class implementations exist instead of duplicate ones; having duplicates means dynamic_cast might not work (according to my research).

src/core/Arguments.h
kintel: Right, so really just compilation time optimization. It hasn't really been a pain point for me for while, but I'm not going to block the effort based on my experience, so let's run with this :)

kintel:
> 
> Cool! I felt pretty good until I saw there are currently 3000 issues 🤣

Yeah, this effort is low-priority, but the first step is seeing the issues, then deciding what's worth addressing.

src/core/Context.cc
kintel: I don't see how this changes anything about dynamic_cast..
Anyway, since the rest of this PR looks fine, perhaps move the `Context` out and we can merge the rest?

coryrc: Aha, so the reason I moved it to Context.cc is the implementation of ContextHandle requires you include EvaluationSession.h, which makes everything using Context or ContextHandle (since they share a header) depend on EvaluationSession.h, which also has a nest of includes (though I'm improving it as I write this).

coryrc: If ContextHandle was in its own file (along with ContextFrameHandle in its own file), then putting the implementation in the header would approximately result in these files having a dependency on EvaluationSession.h when it wouldn't as it currently stands:

```
src/core/Children.cc
src/core/Children.h
src/core/Context.h
src/core/control.cc
src/core/Expression.h
src/core/Namespace.cc
src/core/RenderVariables.cc
src/core/RenderVariables.h
src/core/ScopeContext.cc
src/core/UserModule.cc
```

(the headers might be able to have forward references)

coryrc: Which I guessed by doing 

```
rgrep -l "core/EvaluationSession" src/core | sort
rgrep -l "ContextHandle<" src/core | sort
```

and some set differencing.

kintel: I still think manually managing template instantiations is a step backwards. This is C++ after all, and generic implementations and ADL-based template instantiation is kind of the bread and butter of C++, so trying to optimize everything to minimize includes is going to be an uphill battle, so I'd say let's focus this effort on the low hanging fruits.

coryrc: Okay. I also figured out there's no reason for it to be in its own file because nothing uses a ContextHandle without also referring to a Context of some sort (duh, I guess).

coryrc:
Hmm... somehow Windows header includes are different? Oh... probably missing a library that causes something not to be conditionally included and a transitive dependency gets missed. I will fix.

coryrc:
Fixed now, tests all pass!

PR #6168, opened 2025-09-06
Fix docker working directory in wasm build script

BridgerB:
The `wasm-base-docker-run.sh` script currently mounts the source directory at `/src`
  inside the Docker container but doesn't set the working directory. This causes commands
  like `cmake` to run from `/home/ubuntu` (the default container working directory) instead
   of `/src` where the source code is mounted.

  This results in errors like:
  CMake Error: The source directory "/home/ubuntu" does not appear to contain
  CMakeLists.txt.

  **Fix:** Add `-w /src` to the `docker run` command to set the working directory inside
  the container to `/src` where the source code is mounted.

  **Testing:** This fix allows `make wasm` to work properly when building OpenSCAD WASM
  binaries from source.

### DIFF
--- a/scripts/wasm-base-docker-run.sh
+++ b/scripts/wasm-base-docker-run.sh
@@ -38,6 +38,7 @@ echo "
 
 docker run --rm -it \
   --platform=linux/amd64 \
+  -w /src \
   -v "$PWD:/src:rw" \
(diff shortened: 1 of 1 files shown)


t-paul (approved):
I guess this script does not get much use. Thanks for the fix!

Issue #6152, opened 2025-09-01
Visual Studio Build Setup Issues


kintel:
This is a collection of various issues/topics related to getting a Visual Studio build going by following https://github.com/openscad/openscad/blob/master/doc/win-build.md

It's meant as a dump of topics which hopefully can be solved relatively easily. Larger topics should be split out into separate tickets.

* [ ] [Dependencies] In general, can we install dependencies from a script or package manager too, to avoid having to click-install, add to PATH etc. for each tool? ..plus allow us to manage dependency versions? This includes git, cmake, flex/bison
* [ ] [Dependencies] Git installation instructions are unclear. The link kind of recommends GitHub Desktop, but that doesn’t seem to install cmd-line git, at least it’s not clear how to make it work. Perhaps a specific PATH needs to be set? Installing the alternative git from git-scm yields an installer with a number of options. Perhaps the default options are good, but we should probably make that explicit
* [ ] [Dependencies] CMake is required to use vcpkg - we should add that as a dependency.
* [ ] [vcpkg] Our doc says to add vcpkg to PATH, but vcpkg docs also says to set VCPKG_ROOT. Should we do that?
* [ ] [build script] `win-msvc-build.bat` doesn’t break on first error, but keeps going - makes it hard to spot issues
* [ ] [build script] Packages get installed into the openscad source folder; we should move this somewhere more global, so that it doesn’t get wiped accidentally and so that multiple clones can use the same cache.
* [ ] [Packages] Boost: Unnecessary to install all boost libs; just install the ones we need
* [ ] [Build] Fails with `Could NOT find FLEX (missing: FLEX_EXECUTABLE)`, despite it being in the PATH

kintel:
@Sparsh-N

kintel:
@jordanbrown0 If you have interest/motivation in getting a native Windows build env going without dealing with msys2 and its run-time quirks, this could eventually turn out to be a better option. Needs more testing and more work, but hoping to land a minimal stable state soon.

Sparsh-N:
I'll look into the other issues once I'm free this weekend, but looked at the readme again and it mentioned

```
If you're using a newer version of Ubuntu, you can install these libraries from aptitude. If you're using Mac, or an older Linux/BSD, there are build scripts that download and compile the libraries from source. Follow the instructions for the platform you're compiling on below.

A C++ compiler supporting C++17
[cmake (3.5 ->)](https://cmake.org/)
[Qt (5.12 ->)](https://qt.io/)
[...]

PR #6175, opened 2025-09-08
Set and use vcpkg_root in Windows/MSVC build

Sparsh-N:
#6152 notes that we can use VCPKG_ROOT instead of having the user manually set or change it. This PR modifies the build script, as well as the documentation.

edit - dont merge this just yet, I have to update the TBB_DIR path as well, but ideally that path shouldnt be specified (but thats for another issue).


kintel:
I marked as draft, just click "Ready for review" once you think it's ready

Sparsh-N:
I added the other parts where VCPKG_ROOT is needed, this PR should be ready for merging. I'll rebase the commits.

PR #6177, opened 2025-09-09
Test refactoring to allow stand-alone testing

kintel:
This allows configuring tests as a stand-alone project using an external OpenSCAD binary to test against:
```
mkdid build-tests
cmake ../tests -DOPENSCAD_BINPATH=<path to some OpenSCAD binary>
ctest
```

This is done by checking if `tests/CMakeListst.txt` is the top-level CMakeLists file (`if(PROJECT_IS_TOP_LEVEL)`), and managing the two variants accordingly.

Related changes:
* Split out a bunch of CMake helpers into separate files
* Allow specifying EXAMPLES_DIR; falls back to searching relative to the exe location
* Moved `enable_testing()` to tests/CMakeLists.txt


kintel:
@jordanbrown0 This works for me on macOS now, so it should be possible to test it on Windows:

```
<clone OpenSCAD>
mkdir build-tests
cd build-tests
cmake ../tests -DOPENSCAD_BINPATH=path/to/openscad-executable
ctest
```

Notes:
* Needs cmake and python in the PATH
* Windows path weirdness is completely untested
* I have no idea what implications the choice of shell will have

PR #6196, opened 2025-09-12
Fix uni-get-dependencies.sh

nomike:
I did experiments with building PythonSCAD using Docker.

My Dockerfile looks like this:

```Docker
FROM archlinux:latest

COPY . /opt/openscad

RUN rm -rf /opt/openscad/build

WORKDIR /opt/openscad

RUN scripts/uni-get-dependencies.sh

RUN cmake -B build

RUN cmake --build build -j$(nproc)

WORKDIR /opt/openscad/build

RUN ctest -j$(nproc) --output-on-failure

# Default command
CMD ["build/openscad"]
```

You can build it with `docker build -f /path/to/dockerfile /path/to/openscad/repository`.

I did tests with a few Linux distributions by changing the `FROM` line accordingly:

```Docker
FROM archlinux:latest
FROM debian:12
FROM fedora:42
FROM linuxmintd/mint22.2-amd64
FROM ubuntu:22.04
FROM ubuntu:25.04
```

With this approach, it's extremely low effort to test whether OpenSCAD builds on a particular flavor of Linux (would be nice it this worked for Windows, macOS and *BSD too but unfortunately it doesn't). As those Docker base images are quite minimalist, this seems to be a good approach to test for the completeness of `scripts/uni-get-dependencies.sh`.

There were a few build failures, mainly due to `cmake` not finding required libraries, but also some compile time issues due to missing libs.

I updated `scripts/uni-get-dependencies.sh` accordingly until all builds were successful and ported the changes over to OpenSCAD. During that porting I removed "curl" and "libsvg", as at the moment those are not required by OpenSCAD yet.

The result is this PR.

So far I have those Docker files locally on my machine only, but if you like, I'm happy to contribute them to OpenSCAD in a new PR.
Maybe in `/docker-files/ubuntu22.04/Dockerfile` or something like this, with a small documentation of how to use them properly in `/docker-files/README.md`.

Let me know what you think please.

### DIFF
--- a/scripts/uni-get-dependencies.sh
+++ b/scripts/uni-get-dependencies.sh
@@ -176,7 +176,7 @@ if [ -e /etc/issue ]; then
   get_mageia_deps
  elif [ "`grep -i qomo /etc/issue`" ]; then
   get_qomo_deps
- elif [ "`grep -i arch /etc/issue`" ]; then
+ elif test -r /etc/arch-release ; then
    get_arch_deps
  elif [ -e /etc/fedora-release ]; then
   if [ "`grep -i fedora.release /etc/fedora-release`" ]; then
--- a/scripts/uni-get-dependencies.sh
+++ b/scripts/uni-get-dependencies.sh
@@ -104,21 +104,22 @@ get_debian_deps()
 {
  apt-get update
  apt-get -y install \
-  build-essential bison flex git curl cmake ninja-build libffi-dev \
+  bison build-essential cmake curl flex gettext git gtk-doc-tools imagemagick \
   libboost-program-options-dev libboost-regex-dev libboost-system-dev \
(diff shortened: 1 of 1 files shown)


t-paul:
Just stashing the Docker files on github would only mean more files that bitrot.

The only way those remain being useful is when they are rebuilt regularly and new problems get fixed.

We do have the infrastructure at https://ci.openscad.org (running https://github.com/openscad/openscad-ci) which is currently building the OpenSCAD docker images once a week (and partially failing to build the python based library manager which nobody except me cares about unfortunately).

So what would help is getting those builds automated and also maintained in case of issues. If all that will just end up on my plate like all the other builds, I'm not interested.

t-paul (approved):

PR #6203, opened 2025-09-15
Add documentation of the Python API

damz:
Improve the developer experience of Python models by documenting the API surface of the `openscad` module in a `.pyi` file.

Screenshot from 2025-09-14 20-17-13

## How do I use this?

Place this file in the root of your project as `typings/openscad.pyi`. Pylance will detect it automatically.

It should be possible to place it into the virtualenv directory created by openscad, but I will leave this for a separate PR.

### DIFF
--- a/doc/openscad.pyi
+++ b/doc/openscad.pyi
@@ -1,0 +1,914 @@
+from typing import Union, Optional, overload
+
+OpenSCADObjects = Union["OpenSCADObject", list["OpenSCADObject"]]
+"""Type for functions that accept either a single OpenSCAD object or a list of objects."""
+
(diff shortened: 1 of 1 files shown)


gsohler:
Thats outstanding documentation of the current python functionality.
OpenSCAD can handle quite some operators which can make code more readable and shorter. 
Does pyi support describing them ?

damz:
> OpenSCAD can handle quite some operators which can make code more readable and shorter.
> Does pyi support describing them ?

That's a great point. It seems like if we add stubs for `__and__` / `__or__` / and `__sub__`  stubs they do show up in Pylance on hover, and the type of the result is correct.

PR #6205, opened 2025-09-15
docs: Document operators `&`, `|` and `-` between two objects

damz:
As suggested by @gsohler in #6203, document the Python operators between two objects:

* `&` returns the intersection of the two operands
* `|` returns the union of the two operands
* `-` returns the difference of the two operands

The code also seems to support `+`, `-` and `*` between an object and a vector, but I left that as a TODO, because if I understand the code correctly, it currently requires a _list of vectors_ on the RHS, not just a vector.

### DIFF
--- a/doc/openscad.pyi
+++ b/doc/openscad.pyi
@@ -292,6 +292,21 @@ class OpenSCADObject:
     def __getitem__(self, name): ...
     def __setitem__(self, name, value): ...
 
+    # Operators:
+
+    def __or__(self, other: OpenSCADObjects) -> "OpenSCADObject":
+        """Create a union of two objects"""
+        ...
+
+    def __and__(self, other: OpenSCADObjects) -> "OpenSCADObject":
+        """Create an intersection of two objects"""
+        ...
+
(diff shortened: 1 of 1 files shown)


t-paul (approved):

damz:
> The code also seems to support +, - and * between an object and a vector, but I left that as a TODO, because if I understand the code correctly, it currently requires a list of vectors on the RHS, not just a vector.

Can anyone confirm this? That sounds maybe unintentional?

gsohler:
python_vectors understands both a vector AND a list of vectors.
Only vector is yet  used, list of vectors is yet an unused feature as ^ operator is not yet backported.
see
image
example.  
Happy to create a PR when interested

gsohler:
hmm, github screwed the  animation. checkout top of 

https://pythonscad.org/examples.php

gsohler:
Having a list of vectors for + (plus) makes much sense to me - specifying a set of poiints where to translate to.

{197D5AB0-2C69-44D2-AC60-57FB4CD431FA}

BTW: * (multiplicate) equates to scale (hope its logical)

PR #6209, opened 2025-09-16
Allow list of vectors

gsohler:
This simple addition enables + and - operator to accept a complete list of vectors.
And yes: NodeCache checking has to be weakened, but my ctests are clean, lets check here


src/core/NodeCache.h
```diff
@@ -36,16 +36,38 @@ class NodeCache
 
   void insertStart(const size_t nodeidx, const long startindex)
   {
-    assert(this->cache.count(nodeidx) == 0 && "start index inserted twice");
-    this->cache.emplace(nodeidx, std::make_pair(startindex, -1L));
+#ifdef ENABLE_PYTHON
+    if (true) {
+      if (this->cache.count(nodeidx) ==
+          0)  // with python it can happen that nodes get dumped several times,
+              // but its understood that the dump will always be identical
```
kintel, line 43: I don't quite understand why. This sounds like a failure to do a cache lookup before inserting into the cache?

gsohler: My understanding is that in contrast  to OpenSCAD, Python+Openscad can reference IDENTICAL instead of EQUAL subtrees and therefore "disabled" this check.
in any case: the assert will fire with a list of vectors. Will you help me to track down the "correct"  issue ?

Issue #6198, opened 2025-09-13
Rotate Extrude error message could use improvement


bitbasher:
**Describe the bug**
doing
`rotate_extrude() circle(5);`

gives this error
> ERROR: all points for rotate_extrude() must have the same X coordinate sign (range is -5.00 -> 5.00)
which fails to inform on two counts
1. "all points" without context is not helping
points are not relevant .. shapes are. The use of "for" suggests the "points" are part of the extrude, when it is actually the position(s) or the the child shape(s) that is (are) causing a problem

3. "range is -5->5" suggests to me that the allowed, or perhaps desired, range should be between -5 and 5
but how does this range apply to the extrude? of course, it applies to the children
 
**Expected behavior**
a message like would be more informative

The children of rotate_extrude must all be on the same side of the X axis. Points in the children shape(s) range from -5 to 5.

**Environment and Version info (please complete the following information):**
 - OS: [e.g. Windows 10,
 - OpenSCAD Version - last weeks snapshot

bitbasher:
i found the rotate_extrude.cc file .. and the parameter processing code .. but i do not find where the special conditions for the parameters are checked .. if someone can show me where that happens i could have a stab at improving the messages

jordanbrown0:
I believe this is the place where that error is detected and reported:

https://github.com/openscad/openscad/blob/ee40fe50d23401e2f19b17ba5c0bcc9f13bacf54/src/geometry/rotate_extrude.cc#L94

In general, some of the parameters are validated in `src/core/*Node*` and `src/core/primitives.*`, but that's only generating the CSG tree and doesn't know what the underlying geometry will look like... and that's what's important here.

bitbasher:
thx Jordan .. some of the param validation is in rotate_extrude_node.cc (sic) but yes, the X coord range checking is in r_e.cc .. i will see what i can do with it

bitbasher:
i have been experimenting with the rotate_extrude() module and i conclude that this bug should be solved as is for the moment, but that there is a bigger bug in the current operation ( which will cause discussion so it should be a new issue .. if there is interest in making the change .. i will post about this on the mailing list so folks can discuss.

%rotate_extrude( 360, $fn=5 ) translate([4,4,0]) base_shape();
draws this
<img width="645" height="388" alt="Image" src="https://github.com/user-attachments/assets/0ff8e7e4-29a7-4eeb-859a-d761c37c7a2b" />

but this:
```
%rotate_extrude( $fn=5)
[...]

PR #6212, opened 2025-09-17
fixing issue #6198 - improve Rotate Extrude error message

bitbasher:
i propose the new message to fix issue #6198
"children of rotate_extrude() may not lie across the Y axis (child verticies' X coords in range [%1$.2f : %2$.2f] )"

i also changed the check on convexity param in RotateExtrudeNode.c to have both parts in one place using max() so the code is a bit better at self documenting.

I have also changed the deprecation message in RotateExtrudeNode.cc to make clear that wanting the start to be on -X axis will, in future, have to be given as start=180 explicitly .. which was grammatically unclear in the existing msg.

### DIFF
--- a/.github/workflows/macos-tests.yml
+++ b/.github/workflows/macos-tests.yml
@@ -17,7 +17,7 @@ jobs:
   build_and_test:
     strategy:
       matrix:
-        os: [macos-13, macos-latest]
+        os: [macos-15-intel, macos-latest]
         qt: [qt5, qt6]
         exclude:
           # macos-latest runs on arm64, which has a broken SW renderer


bitbasher:
please give feedback if my PR has not been done correctly .. thx

src/geometry/rotate_extrude.cc
```diff
@@ -93,7 +93,7 @@ std::unique_ptr<Geometry> rotatePolygon(const RotateExtrudeNode& node, const Pol
   if (max_x > 0 && min_x < 0) {
     LOG(
       message_group::Error,
-      "all points for rotate_extrude() must have the same X coordinate sign (range is %1$.2f -> %2$.2f)",
+      "children of rotate_extrude() may not lie across the Y axis (child verticies' X coords in range [%1$.2f : %2$.2f] )",
```
jordanbrown0: You're looking for "vertices".  ("vertexes" is also correct, but scanning the sources and the documentation the convention used here is "vertices".)

I can't say that I'm a fan of either the original or this replacement.  I haven't really got any better suggestions, though the thing that I find most awkward in both variants is the parenthesized part.  That's not how parentheses are used in English.  I would probably either eliminate that part entirely, or split it off into a separate sentence.

src/core/RotateExtrudeNode.cc
```diff
@@ -71,12 +72,9 @@ std::shared_ptr<AbstractNode> builtin_rotate_extrude(const ModuleInstantiation *
   bool hasStart = parameters["start"].getFiniteDouble(node->start);
   if (!hasAngle && !hasStart && (int)node->fn % 2 != 0) {
     LOG(message_group::Deprecated,
-        "In future releases, rotational extrusion without \"angle\" will start at zero, the +X axis.  "
-        "Set start=180 to explicitly start on the -X axis.");
+        "In future releases, with no given \"angle\" nor \"start\" params the extrusion will start at the +X axis and start=180 will have to be explicitly given to start on the -X axis.");
```
jordanbrown0: The only information that this seems to add is that the default starting position only applies when the start position is not explicitly specified... which seems like it can go without saying.  Other than that, it adds nine words, turns two medium sentences into one long sentence, and switches the second half from active voice to passive voice.

Net, I don't think this is an improvement.

src/core/RotateExtrudeNode.cc
```diff
@@ -56,7 +56,8 @@ std::shared_ptr<AbstractNode> builtin_rotate_extrude(const ModuleInstantiation *
   node->fs = parameters["$fs"].toDouble();
   node->fa = parameters["$fa"].toDouble();
 
-  node->convexity = static_cast<int>(parameters["convexity"].toDouble());
+  node->convexity = std::max( 2, static_cast<int>(parameters["convexity"].toDouble()) );
```
kintel: Should this be `std::min`?
The larger question is: Why did this not break any tests?

jordanbrown0: This combines this assignment with the
```
if (node->convexity <= 0) node->convexity = 2;
```
from further below.

Technically it's a small behavior change, in that formerly an explicit convexity=1 would be used, and with this max() it'll be bumped up to 2.

bitbasher:
i used std::max() because Jordan suggested that over just plain max()
and i used max as 2 the smallest value for convexity that rotate_extrude() should use .. so if any value less than two is given ..

the original code extracted the convexity param before all the stuff working on angle and start .. then checked for convexity being less than 2 after .. I find the code "reads" better when operations on things are all done in one place.

reply to your feedback Jordan ( and thanks for that)
error messages are only loosely connected to English grammar IMHO so using parens to mark off a hint to the reader about what might be wrong is okay by me .. if the project's standards are higher I will up my game 

changing to "children of ..." is to cover when several 2D shapes are children to the r_e() .. and "lies across the Y axis.." was the shortest wording i could find for "having vertices of child shapes in both -X and +X parts of the X-Y plane.

The "in future releases..." change .. i think is important to be clear that extrusion starts at -X only if  _both_  angle and start are _not_ given .. if start is explicitly given, and angle not .. the existing wording is ambiguous to me.

jordanbrown0:
I totally approve of moving the two convexity operations together.  I do not have a real feeling on replacing an explicit convexity=1 with 2.

I don't have any additional suggestions with respect to the "crossing the Y axis" message.

With respect to the start/angle message... the idea that with "start" explicitly specified we might *not* use it seems silly, so I wouldn't spend any words at all on it.

src/core/RotateExtrudeNode.cc
bitbasher: i noticed the convexity = 2 as a deviation from the normal default of 1 and thot it was down to an r_e() always being a torus

bitbasher:
shall i cancel this PR and redo it leaving out the changes to messages in RotateExtrudeNodes.cc as they don't really help anything?

t-paul:
You can force-push to the branch in your repo. This PR will pick up those changes. For your feature/bugfix branch, that's not a problem. Force-pushing is only bad for branches that are actively worked on by multiple people.

src/core/RotateExtrudeNode.cc
jordanbrown0: > an r_e() always being a torus

```
rotate_extrude() square(10);
```

bitbasher: still technically a torus with a zero sized hole in the middle .. doing View > Show Edges reveals the top and bottom to be a network of triangles but yes, in this case convexity = 1 would be enough .. but it would still be forced to 2 by the code in RotateExtrudeNode.cc

kintel: Duh, my mistake. This is correct :)

kintel:
> You can force-push to the branch in your repo. 
You don't even need to force-push; you can just keep adding commits, and when merging we will usually "squash" everything into a single commit, so that your local history of edits doesn't make it into the final repo.

bitbasher:
so to avoid messing anything up .. so i should
1. leave the change to use std::max() in RotateExtrudeNode.cc
2. but revert the change to the message about "in future releases..." 
3. also keep the text change in rotate_extrude.cc as i suggested it

i hope i have navigated your feedback correctly

src/geometry/rotate_extrude.cc
bitbasher: would 
"... ( Range of X coords for all children [ %1$.2f : %2$.2f ] )"
be a better wording for you Jordan?

jordanbrown0: Eh, sure.

bitbasher:
it looks like we need a mac with Mac OS 14 or better .. i have such a machine here if it can be hooked into the build somehow

kintel:
Unfortunately, we need cloud-hosted runners, as these builds are running continuously. We have some alternatives, but it's all a bit of work. I'll open a separate ticket for it, plus find a way of safely merging PRs which fail on the current macOS runner.

kintel:
Attempting to fix via #6214

Issue #23, opened 2011-09-06
Module /Function Parameter Passing By Reference


rbaumann:
Would it be possible to pass parameters into modules /functions by reference? Having such a mechanism allows writing modules that can significantly enhance re-usability of components. For example imagine the following:

c = 5;  // side c of the triangle
b = 4;  // side b of the triangle
a = 3;  // side a of the trinagle
r = 1;  // radius of round triangle corners
h = 3; // hight of triangle
p = [...]; // Coordinates of corner centers and triangle's center of gravity

// create a triangle with round corners and a 2 mm diameter centered hole in each corner
difference () {
      3D_TRIANGLE (s1=a,s2=b,s3=c,hi=h,p_arr=p);  // module drawing 3D-triangles with rounded corners
      translate (p[0] cylinder (r1=1,r2=1,h=5,center=true); // p[0] got co-ordinate of triangle A - corner
      translate (p[1] cylinder (r1=1,r2=1,h=5,center=true); // p[1] got co-ordinate of triangle B - corner
      translate (p[2] cylinder (r1=1,r2=1,h=5,center=true); // p[2] got co-ordinate of triangle C - corner
}

Issue #22, opened 2011-09-01
Reload and Compile is broken, Mac OS X-OpenSCAD 2011.08.06


mshearn:
Compile (F5) works and displays the model, as does Compile and Render (F6).  However, Reload and Compile (F4) does not.
Occurs in:
Mac OS X, OpenSCAD 2011.08.06

Issue #14, opened 2011-06-05
MCAD/involute_gears.scad problem in F5 mode.


knuds:
I use OpenSCAD 2011.04.27 on Ubuntu. 

It renders OK with F6 but F5 appear strange. 

See 
http://all-technology.com/openscad-F5.png
and 
http://all-technology.com/openscad-F6.png

The script I use is:
include <MCAD/involute_gears.scad>;

teeth=11;
dist=150;

translate([dist,dist,0]) rotate([0,90,90])  bevel_gear_pair(gear1_teeth = teeth, gear2_teeth = teeth);
translate([dist,-dist,0]) rotate([0,-90,90]) bevel_gear_pair(gear1_teeth = teeth, gear2_teeth = teeth);
translate([-dist,dist,0]) rotate([0,-90,-90])  bevel_gear_pair(gear1_teeth = teeth, gear2_teeth = teeth);
translate([-dist,-dist,0]) rotate([0,90,-90]) bevel_gear_pair(gear1_teeth = teeth, gear2_teeth = teeth);

PR #6213, opened 2025-09-17
More operators

gsohler:
Openening again, as i closed it inadvertently.
I am interested if the NodeCache Issue is on the Cache side or on the Tree assembly side.


kintel:
In terms of the NodeCache issue: Do you have a stack trace or smth. showing how `insertStart()` is called twice on the same node?

gsohler:
yes of course. I am happy to show a copy of my stack trace. i just learned
now , that irts useful

Loaded design '/home/gsohler/git/openscad/build/a.py'.
openscad: /home/gsohler/git/openscad/src/core/NodeCache.h:48: void
NodeCache::insertStart(size_t, long int): Assertion
`this->cache.count(nodeidx) == 0 && "start index inserted twice"' failed.
Aborted (core dumped)

#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=<optimized
out>) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=<optimized out>) at
./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=<optimized out>, ***@***.***=6) at
./nptl/pthread_kill.c:89
#3  0x00007e448384526e in __GI_raise ***@***.***=6) at
../sysdeps/posix/raise.c:26
#4  0x00007e44838288ff in __GI_abort () at ./stdlib/abort.c:79
#5  0x00007e448382881b in __assert_fail_base
    (fmt=0x7e44839d01e8 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n",
***@***.***=0x5f6ed62d8108 "this->cache.count(nodeidx) == 0
&& \"start index inserted twice\"", ***@***.***=0x5f6ed62d80d8
"/home/gsohler/git/openscad/src/core/NodeCache.h", ***@***.***=48,
***@***.***=0x5f6ed62d80a8 "void NodeCache::insertStart(size_t,
long int)")
    at ./assert/assert.c:94
#6  0x00007e448383b507 in __assert_fail
    (assertion=0x5f6ed62d8108 "this->cache.count(nodeidx) == 0 && \"start
index inserted twice\"", file=0x5f6ed62d80d8
"/home/gsohler/git/openscad/src/core/NodeCache.h", line=48,
function=0x5f6ed62d80a8 "void NodeCache::insertStart(size_t, long int)")
    at ./assert/assert.c:103
#7  0x00005f6ed5888aa6 in NodeCache::insertStart(unsigned long, long) ()
#8  0x00005f6ed5888056 in NodeDumper::visit(State&, AbstractNode const&) ()
#9  0x00005f6ed571d99f in NodeVisitor::visit(State&, AbstractPolyNode
const&) ()
#10 0x00005f6ed571dabf in NodeVisitor::visit(State&, LeafNode const&) ()
#11 0x00005f6ed5860bd2 in Response
BaseVisitable::acceptImpl<LeafNode>(State&, LeafNode const&, BaseVisitor&)
()
#12 0x00005f6ed5860785 in LeafNode::accept(State&, BaseVisitor&) const ()
#13 0x00005f6ed588ebd5 in NodeVisitor::traverse(AbstractNode const&, State
const&) ()
#14 0x00005f6ed588ecb2 in NodeVisitor::traverse(AbstractNode const&, State
const&) ()
#15 0x00005f6ed588ecb2 in NodeVisitor::traverse(AbstractNode const&, State
const&) ()
#16 0x00005f6ed58fb287 in Tree::getIdString[abi:cxx11](AbstractNode const&)
const ()
#17 0x00005f6ed5a69b61 in GeometryEvaluator::smartCacheGet(AbstractNode
const&, bool) ()
#18 0x00005f6ed5a6792f in GeometryEvaluator::evaluateGeometry(AbstractNode
const&, bool) ()
#19 0x00005f6ed5da218e in CGALWorker::work() ()
#20 0x00005f6ed5da2aa7 in QtPrivate::FunctorCall<QtPrivate::IndexesList<>,
QtPrivate::List<>, void, void (CGALWorker::*)()>::call(void
(CGALWorker::*)(), CGALWorker*, void**) ()
#21 0x00005f6ed5da2a26 in void QtPrivate::FunctionPointer<void
(CGALWorker::*)()>::call<QtPrivate::List<>, void>(void (CGALWorker::*)(),
CGALWorker*, void**) ()
#22 0x00005f6ed5da298d in QtPrivate::QSlotObject<void (CGALWorker::*)(),
QtPrivate::List<>, void>::impl(int, QtPrivate::QSlotObjectBase*, QObject*,
void**, bool*) ()
#23 0x00007e4484312a76 in ??? () at /lib/x86_64-linux-gnu/libQt5Core.so.5
#24 0x00007e44840d940d in QThread::started(QThread::QPrivateSignal) () at
/lib/x86_64-linux-gnu/libQt5Core.so.5
#25 0x00007e44840db65d in ??? () at /lib/x86_64-linux-gnu/libQt5Core.so.5
#26 0x00007e448389ca94 in start_thread (arg=<optimized out>) at
./nptl/pthread_create.c:447
#27 0x00007e4483929c3c in clone3 () at
../sysdeps/unix/sysv/linux/x86_64/clone3.S:78

==================================

This is my code:

from openscad import *
a=cube()+ [[10,0,0],[0,10,0]]
a.show()

CSG tree, once successful looks like


union() {

multmatrix([[1, 0, 0, 10], [0, 1, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]) {

cube(size = [1, 1, 1], center = false); <= SAME subnode(not only equal)

}

multmatrix([[1, 0, 0, 0], [0, 1, 0, 10], [0, 0, 1, 0], [0, 0, 0, 1]]) {

cube(size = [1, 1, 1], center = false); <= SAME subnode(not only equal)

}

}


The code which assembles a CSG tree with "SAME" subnodes is here:


https://github.com/openscad/openscad/blob/master/src/python/pyfunctions.cc#L1516-L1524


In case that's a problem: why and when ?

kintel:
I see what you mean. Until now, a node tree has had the requirement that each node has a unique ID.
I cannot tell from the top of my head if violating that will cause bad behavior, but there may be some features which requires this (e.g. picking in the view which maps node IDs back to the CSG tree and source code).

_If_ we decide that duplicate nodes are OK, I would suggest to solve this in `NodeDumper`. Instead of writing twice to the cache, call `cache->contains()` and prune the traversal if the cache already exists in `NodeDumper::visit()`.
Until we have a good idea, I think a safer way forward would be to duplicate the node and give it a unique index.

gsohler:
ok, got it.
But give me  chance to state some facts:

* NodeDump duplication is disabled in python mode for more  than 1.5 years,
and neither me nor my users have reported an issue about that
* ctest are pass as you can see

I cannot see a problem when picking  a node id to get the the CSG tree
source code:  Two regions on the screen will just see the same id and the
result code will still show correct.

When in your code there is a module with a cube inside. When you
instantiate the module twice, won't it be the same node twice in the CSG
tree ?
Personally it's hard to understand to duplicate memory just for the sake to
be "safer", but I see your view of "never change a running system"

On the Other hand, sometimes we intentionally want to clone nodes like here

a=cube()
b=a # cloning here
b += sphere() # because  we don't want "a" to be affected by this operation
Luckily this can hardly happen in a descriptive language :)

"Duplicating the node" means  duplicating one level or hierarchically ?
I will need to study your proposed solution now.

gsohler:
And: why is it a problem when caching the same thing twice when 2nd  item
is identical . You waste compute time at most.

kintel:
> And: why is it a problem when caching the same thing twice when 2nd item is identical . You waste compute time at most. 
>
Since the purpose of the cache is to avoid wasting compute, doing it on purpose in the cache is a bit of an antipattern.

kintel:
> When in your code there is a module with a cube inside. When you instantiate the module twice, won't it be the same node twice in the CSG tree ?

Yes, they are instantiated twice, since we don't know until instantiation time what exact parameters will be passed. Perhaps not so much for a cube, but for a sphere, it needs to know e.g. the `$fn` value.

I agree that there is room for optimization here, and that's untapped potential. I'm just a little hesitant to switching low-level behavior like this without a broader analysis.

jordanbrown0:
When I did the geometry value work in #4478, when inserting a geometry value into the model I started out by just splicing in the subtree (and so forming a DAG, when spliced in multiple times).  I don't remember the misbehavior, but I ended up needing to clone the tree to get unique IDs.

gsohler:
Suppose its easier to copy the nodes rather than to convince, that duplicate nodes are ok.
Do we need a full hierarchical copy or do we only need to clone the toplevel ?
Does OpenSCAD already have a function to clone (or how does openscad get unique ids for same objects ?)

kintel:
> Does OpenSCAD already have a function to clone (or how does openscad get unique ids for same objects ?)

We do have node cloning, thanks some some guy called Günther: https://github.com/openscad/openscad/blob/master/src/core/node_clone.cc ;)

gsohler:
I am impressed that you got my name correct even though you might not be
not local to my area .

And I am sorry: It's hard for me to keep an overview, which features are
merged.
My intent was to challenge first, if openscad has different cloning
strategies.
In any case, I have updated my PR with the clone() and now the additional
feature works as expected without an assert.
Thx for your attention

kintel:
Thanks, I think this solution is a better starting point.

..but I'm interested in exploring how duplicated nodes could be well-supported, as well as whether they could offer any optimization opportunities for the OpenSCAD language itself. But that's a larger issue which should be dealt with separately.

PR #6217, opened 2025-09-20
Remove QUIET flag from find_package calls to improve build diagnostics

nomike:
I'm running GNU Guix  on top of Ubuntu GNU/Linux and libraries are often provided by both.
When I build an application, it is important for me to know whether a library was picked up from the former or the latter.

In CMakeLists.txt we inconsistently do stuff like this:

```CMake
    find_package(manifold REQUIRED QUIET)
    message(STATUS "manifold: ${manifold_VERSION}")
```

This hides important information like library locations.
There is a message command afterwards which at least prints out the version, but if we were to ditch the `QUIET` flag, we would not even have to do this.

Thus my patch removes the `QUIET` flag from all `find_package` statements in our CMake files.

Even though this technically makes printing out the detected version as a status message redundant, I decided to keep it, to not break people's workflows if they parse for these messages or something.


t-paul:
I guess that contradicts https://github.com/openscad/openscad/pull/6178 a bit.

In general I'd like to see a trace of at least all the decisions made, but looking at the 2 PRs, we need to find some common ground.

kintel:
Yeah, I found the verbosity made it hard to spot what's going on. I'd say a verbose mode would be nice.

nomike:
There is the CMake option `-DCMAKE_FIND_DEBUG_MODE=ON`  which does exactly that. So I consider this PR as unnecessary and will close it.
And this way, things are documented if ever someone else searches for this issue.

t-paul:
Right, I guess we need to make it possible to find https://github.com/openscad/openscad/wiki/Developer-Notes ... and also update it to current state.

Issue #6215, opened 2025-09-20
Manifold tests are being disabled when ENABLE_MANIFOLD_TESTS == ON


nomike:
I'm trying to make sense of this [chunk of code in tests/CMakeLists.txt](https://github.com/openscad/openscad/blob/c8fbef05ba900e46892e9a44ea05f7d88e576e13/tests/CMakeLists.txt#L1376):

```CMake
  if (ENABLE_MANIFOLD_TESTS)
    set_tests_properties(
      # https://github.com/openscad/openscad/issues/5966
     render-off-manifold_issue5216
     render-3mf-manifold_issue5216

     PROPERTIES DISABLED TRUE
    )
  endif (ENABLE_MANIFOLD_TESTS)  
```

I'm trying to get the build more stable and figured out that although lib3mf is optional, the build fails due to failed tests when it is not detected or when it is turned off.

I already managed to disable a few lib3mf relates tests locally, but the test `render-off-manifold_issue5216` still fails when lib3mf is not present and should be disabled in that case.

The code referenced above however, disables those two manifold related tests when `ENABLE_MANIFOLD_TESTS` is set to true.

I assume this is a mere typo and either the if condition should be the other way around, or the "DISABLED" property of those tests should be set to FALSE instead of TRUE.

Can someone confirm this please?

If so, I will fix this in a local branch alongside with my other lib3mf related fixes and will file a PR.

kintel:
Looks like the off test should be pulled out of the lib3mf check.
While lib3mf is optional, it's strongly recommended to include it, and our official binaries should all have it. Hence, we very rarely run tests with lib3mf turned off.

nomike:
Well, the thing is that it has to stay optional as there are platforms which don't have it in their repositories which makes it hard to build OpenSCAD on them.

But anyway, IMHO we need to decide: 

If there is a flag which allows for lib3mf to be turned off, we need to make sure all tests which depend on it are disabled.

Or we remove the flag and mark the library as REQUIRED in CMake, with the consequence of making it harder to build on some systems (not Impossible though as users could of course just download the library and build it on their own).

Should I do the required refactoring or is this something you could take care of?

t-paul:
I do think lib3mf and others need to stay optional specificially for the reason mentioned, so it's at least possible to build a minimal version even though it would miss some of the import/export and other features but still provides the core which is the language and mesh calculation. So the default build option should not enforce all the optional stuff.

[...]

PR #6221, opened 2025-09-22
fix: improve CMake test configuration logic and conditional compilation

nomike:
- Add explicit FALSE settings for ENABLE_LIB3MF_TESTS and ENABLE_MANIFOLD_TESTS
- Fix inverted logic in test property conditions (NOT EXAMPLES_DIR, etc.)
- Standardize CMake conditional syntax and remove extra whitespace
- Update disabled test properties to match corrected feature availability

The conditionals seem to have been the wrong way around (as I mentioned in #6215 already).
I fixed those.
Additionally I added code to disable some lib3mf related tests if lib3mf tests are not enabled.

Last but not least I added code to disable lib3mf tests if lib3mf is not found and did the same for manifold.

With this patch, OpenSCAD successfully builds even when lib3mf is not available.

### DIFF
--- a/.github/workflows/linux-tests.yml
+++ b/.github/workflows/linux-tests.yml
@@ -28,7 +28,10 @@ jobs:
       with:
         submodules: 'recursive'
     - name: Install dependencies
-      run: ./scripts/github-ci-linux-get-dependencies.sh ${{ matrix.qt }}
+      env:
+        DEBIAN_FRONTEND: noninteractive
+        TZ: Etc/UTC
+      run: sudo ./scripts/uni-get-dependencies.sh ${{ matrix.qt }}
     - uses: actions/setup-python@v5
       with:
         python-version: '3.x'


nomike:
It looks like this needs some more polishing and I just saw that I accidentally left a few debug messages in there.
Stay tuned.

t-paul:
You can convert the PR to draft mode to make it explicit. As for the unrelated msys2 fail, @kintel is trying to fix that via  #6219.

tests/CMakeLists.txt
```diff
@@ -71,9 +71,13 @@ else() # !PROJECT_IS_TOP_LEVEL
   endif()
   if(LIB3MF_FOUND)
     set(ENABLE_LIB3MF_TESTS TRUE)
+  else()
+    set(ENABLE_LIB3MF_TESTS FALSE)
   endif()
```
kintel, line 89: Not quite sure I understand the motivation for doing this. In traditional programming languages, it's tempting to force variables to have a certain type with well-defined values. CMake is kind of built around a certain definition of truthy/falsey, and it feels a bit like you're trying to engineer yourself around CMake instead of embracing it :)

nomike:
I might need some help with those failing tests.
Are they because they should have been disabled or are these really some issues which remained undetected due to being wrongfully disabled?
I can have a look at the CMake files, but I'm not deep enough into OpenSCAD to actually debug those issues I'm afraid...

kintel:
> I might need some help with those failing tests.

Your goal is simply to build without lib3mf and run all applicable tests, right?

tests/CMakeLists.txt
nomike: Good point. Setting the variable to "OFF" is not necessary. I will remove this.

I also found a way to run github actions locally (https://github.com/nektos/act) which makes troubleshooting much more convenient.

Another question which popped up:
What's the deal with "github-ci-linux-get-dependencies.sh"? It looks like it solved issues which might have existed in the past (install lib3mf and libcgal) and also some workaround for a zlib issue.
I would like to try to replace this with `uni-get-dependencies.sh` to deduplicate the code.

Do you have any reservations against or concerns about this?

kintel: If we can remove that file, I'm all for it. I'm not familiar with any corner cases handled by that script though..

Yeah, running nektos/act on a number of variants could be interesting. Challenge is that there are hundreds of combinations of variants : /

nomike:
That's part of my goal.
Where I want to get at is:

If you run `cmake -B build .` it should either fail e.g, because some required library is missing, or it should not fail.

If it does not fail, the subsequent `cmake --build build` must not fail because of a missing library. If this does happen that means that the CMake scripts have a bug and didn't do their job correctly.

And lastly: If the config step of cmake green-lights the build, the tests should also be successful.

Right now if `lib3mf` is missing for example, cmake ignores this and just builds OpenSCAD without it. But there are tests which require lib3mf to run successfully.

IMHO there are two options:
1. cmake already fails when lib3mf is not available and it's just no longer possible to build OpenSCAD without it
2. if lib3mf is not available, certain tests should be disabled

Because right now, building without lib3mf is just not possible.

And it looks like we have the same issue with manifold.

kintel:
We do have provisions for lib3mf and manifold availability in tests, so failure to respect those are bugs which should be fixed. I can take a look.

We haven't really prioritized this, and we don't run automated tests for such variants, so bitrot will happen..

nomike:
Ahhhhh...I understand this now...

There are certain lib3mf related tests which are known to fail and which are not about to be fixed right now.
Because of that, those test are being disabled.

If you run 

```CMAKE
set_tests_properties(
    preview-cgal_3mf-import-centered
    PROPERTIES DISABLED TRUE
)
```
however and the test does not exist, cmake will fail.

The whole bunch of lib3mf related tests is only being added when ENABLE_LIB3MF_TESTS is true and thus they can only be disabled when ENABLE_LIB3MF_TESTS for cmake not to fail.

So that explains the if-condition which looked wrongfully inverted at first glance, but it actually makes sense.

kintel:
You got it :) This could all be better structured but it's tricky

nomike:
J've refactored the code a little and introduced a `set_tests_properties_safe()` wrapper function.

This way we can just disable all the tests which we know don't work and don't have to care whether those are even defined or not.

Makes the code much simpler and easier to understand and get's rid of all those nested if-blocks later on.

I will still have to rebase all my changes and clean 'em up a bit, but I'm eager to hear your opinion on this idea first.

kintel:
> J've refactored the code a little and introduced a `set_tests_properties_safe()` wrapper function.
> 
The refactoring makes sense. I'll comment on the file itself.

tests/CMakeLists.txt
```diff
@@ -22,6 +22,37 @@ set(CSD ${CMAKE_SOURCE_DIR}) # project top level
 set(CCBD ${CMAKE_CURRENT_BINARY_DIR}) # tests build dir
 set(CCSD ${CMAKE_CURRENT_SOURCE_DIR}) # tests source dir
 
+function(set_tests_properties_safe)
+    # Parse arguments: test names followed by PROPERTIES and key-value pairs
+    set(TEST_NAMES)
+    set(CURRENT_LIST TEST_NAMES)
+    set(PROPERTIES_KEYWORDS PROPERTIES LABELS ENVIRONMENT WORKING_DIRECTORY)
```
kintel: Since we only ever use this to disable test, why not just create a `disable_tests()` function instead? That way you don't need to deal with arguments parsing, which should make this function a lot slimmer.

nomike: I had that in the beginning, but then extended it to be more future proof.
But on the other hand, how long do we have this CMakeLists.txt file without the need of anything than disable tests.

So, I agree, it makes more sense to chose the simpler path.

Give me a moment.

nomike: This should look better now and I hope the build actions are successful.

scripts/uni-get-dependencies.sh
```diff
@@ -104,15 +104,20 @@ get_debian_deps()
...
+  libfreetype-dev libgl1-mesa-dev libglew-dev libglib2.0-dev libgmp-dev \
+  libharfbuzz-dev libmimalloc-dev libmpfr-dev libopencsg-dev \
+  libqt5gamepad5-dev libtbb-dev libxi-dev libxml2-dev libxmu-dev \
+  libzip-dev nettle-dev ninja-build nodejs pkg-config python3-dev \
+  python3-setuptools python3-venv ragel xvfb
+ if [ "$QT" = "6" ]; then
+  get_qt6_deps_debian
+ else
+  get_qt5_deps_debian
+ fi
```
kintel: I'd prefer not using environment variables to drive build variants; they are too easy to set and forget.
See e.g. this similar script which takes a simple cmd-line arg for Qt5 vs. Qt6 : https://github.com/openscad/openscad/blob/master/scripts/macosx-build-homebrew.sh

.github/workflows/linux-tests.yml
```diff
@@ -28,7 +28,11 @@ jobs:
       with:
         submodules: 'recursive'
     - name: Install dependencies
-      run: ./scripts/github-ci-linux-get-dependencies.sh ${{ matrix.qt }}
+      env:
+        DEBIAN_FRONTEND: noninteractive
+        TZ: Etc/UTC
```
kintel, line 33: What does timezone do to package installations? Shouldn't this be the same as the timezone of the running container?

nomike:
I've rebased this branch on the current master and cleaned up the commits and commit comments.
If the build is okay now, it should finally be ready to be merged.

nomike:
As `scripts/uni-get-dependencies.sh` has now feature parity with `github-ci-linux-get-dependencies.sh` on Debian/Ubuntu systems, I will test replacing it in the other actions it's still used (`codeql-analysis.yml`, `manual.yml`, `non-experimental.yml`, and `test-examples.yml`).

If that works, which I expect, we could get rid of that script and have one less thing to maintain.

nomike:
But that's something for another PR.

scripts/uni-get-dependencies.sh
```diff
@@ -150,6 +162,19 @@ unknown()
  echo "in README.md using your system's package manager."
 }
 
+printUsage()
+{
+  echo "Usage: $0 [qt6]"
+}
+
+# Qt6 is default
```
kintel: Qt5?

.github/workflows/linux-tests.yml
```diff
@@ -28,7 +28,11 @@ jobs:
...
+        QT: ${{ matrix.qt == 'qt6' && '6' || '5' }}
+      run: sudo ./scripts/uni-get-dependencies.sh $QT
```
kintel: Just pass ${{ matrix.qt }} ?

nomike: I will change that tomorrow. I need to get up in 4 hours. But thanks for spotting this.

scripts/uni-get-dependencies.sh
nomike: Well the thing is:
The example you showed me is a script for compiling for macos, where QT6 is the default.
Right now, `uni-get-dependencies.sh` only installs qt5 components.

I intentionally did it the other way around here, to not change the default behavior of that script. So QT6 support is an option, and it can be added, but by default it stays as it is.

If we'd change it to default to qt6 here, then we should also change it to default to qt6 in here:
https://github.com/openscad/openscad/blob/f992825e9923ea8cff24a1b77c14f7cf40b7a95e/CMakeLists.txt#L45

nomike: I don't mind either, but IMHO we should not force people to set the qt-version when running one of these scripts.
If you don't specify it anywhere, the build should still be successful.

kintel: Agreed, I was referring to that the code comment is wrong.

.github/workflows/linux-tests.yml
nomike: fixed

scripts/uni-get-dependencies.sh
nomike: Sure, I just fixed that.
Additionally I had a look at `macos-build-homebrew.sh`, where I originally copied the code from and figured that this `printUsage()` function wasn't actually used anywhere, so I just replaced it with a comment.

.github/workflows/linux-tests.yml
nomike: The timezone should not influence the package at all, after all, I could install the same .deb package on my machine in Vienna as someone else in Tokyo for example.

What I did was to try and run this worflow with act. Act spun up a Docker container it fetched from GitHub, ran `apt install` and at some point during the configure phase of "tzdata" it stopped and interactively asked for me to provide which region I'm in ("Africa", "South America", "Asia", etc.). Since this was in a Docker container and I was only seeing the logs, having no way to interact with it, the only chance I had was to cancel the workflow.

Setting the "TZ" env variable prevents that as the "tzdata" package (which is a dependency of one of the packages we need) doesn't ask questions. And "Etc/UTC" is the most generic region available.

I wonder though, why this was not an issue on GitHub itself...

scripts/uni-get-dependencies.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
nomike, line 1: I just had to change the shebang as`if [[ ]]` is not supported by sh.

t-paul: Please revert this back to `/bin/sh`. Forcing `bash` will make the script much less useful, especially for environments like docker where `bash` is not necessarily available.

Both string `[ "$var" = "1"]` and number `[ $var -eq 1 ]` compare are supported by the ancient `test` program, so why would we need `[[ ]]`?

t-paul:
Meh, looks like I commented on a resolved thread. So here is it again...

Please revert this back to /bin/sh. Forcing bash will make the script much less useful, especially for environments like docker where bash is not necessarily available.

Both string [ "$var" = "1"] and number [ $var -eq 1 ] compare are supported by the ancient test program, so why would we need [[ ]]?

nomike:
I haven't seen a docker container yet which does not have bash or at least busybox, but whatever, doesn't hur to go back to `sh`.

I've filed #6225 for this.

t-paul:
```
docker run -it --rm alpine:latest
/ # /bin/bash
/bin/sh: /bin/bash: not found
```

```
docker run -it --rm busybox:latest
/ # /bin/bash
sh: /bin/bash: not found 
```

PR #6222, opened 2025-09-22
python: Run python when called as `python` or `python3` in addition to `openscad-python`

damz:
This is arguably a PEBKAC problem, but I couldn't make the virtualenv created by OpenSCAD to work, until I realize that it only works when you call `bin/openscad-python`, and `bin/python` and `bin/python3` are broken.

Support calling the OpenSCAD binary as `python` or `python3` in addition to `openscad-python`.

### DIFF
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -824,7 +824,8 @@ int main(int argc, char **argv)
   // The original name as called, not resolving links and so on. This will
   // just forward everything to the python main.
   const auto applicationName = fs::path(argv[0]).filename().generic_string();
-  if (applicationName == "openscad-python") {
+  if (applicationName == "python" || applicationName == "python3" ||
+      applicationName.rfind("python3.", 0) == 0 || applicationName == "openscad-python") {
     return pythonRunArgs(argc, argv);
   }
 #endif


t-paul:
Can you elaborate why this is needed and what the specific scenario is? I suppose for compatibility with external tool it might be useful but it would be good to understand the use case as there are other places in the OpenSCAD code that rely on the binary name.

damz:
@t-paul The virtualenv created by OpenSCAD contains these:

```
$ ls -lha /home/damz/.openscad-venv/bin/ | grep python
lrwxrwxrwx. 1 damz damz   51 Sep 22 08:53 openscad-python -> /home/damz/Downloads/openscad/build/openscad-python
lrwxrwxrwx. 1 damz damz   15 Sep 22 08:53 python -> openscad-python
lrwxrwxrwx. 1 damz damz   15 Sep 22 08:53 python3 -> openscad-python
lrwxrwxrwx. 1 damz damz   15 Sep 22 08:53 python3.13 -> openscad-python
```

But only `openscad-python` works, the others launch the `openscad`  UI.

damz:
Updated to also support anything that starts with `"python3."`.

PR #6227, opened 2025-09-24
python: Fully destroy the intepreter on re-evaluation

damz:
I have seen a crash in the current code when importing a file that is not part of a package, like:

```
├── foo
│   ├── bar
│   │   └── baz.py
│   └── __init__.py
└── scad.py
```

scad.py:

```
import os
import sys

if os.getcwd() not in sys.path:
    sys.path.append(os.getcwd())

import foo.bar.baz
```

Would crash `openscad` on re-evaluation of the `scad.py` file.

The crashdump is in `initPython`, somewhere where it tries to empty the `sys.modules` dict:

```
#0  0x00007fa843b7e12e PyUnicode_AsEncodedString (libpython3.13.so.1.0 + 0x17e12e)
#1  0x0000000000d2f093 _Z10initPythonRKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEd (/home/damz/Downloads/openscad/build/openscad + 0x92f0>
#2  0x0000000000f0b40b _ZN10MainWindow13parseDocumentEP15EditorInterface (/home/damz/Downloads/openscad/build/openscad + 0xb0b40b)
```

## Proposed solution

Instead of trying to manually reinit the environment, which is messy and doesn't reverts all the side-effects that could have happened (like modifying `sys.path` above), tear down the interpreter with `Py_FinalizeEx` (available since Python 3.6, released in 2016).

Note that the patch is messy but trivial: it is literally just removing a bunch of code from `initPython`. Look at it with ["Hide whitespace" enabled](https://github.blog/news-insights/product-news/ignore-white-space-in-code-review/) in the diff view (or `git show -w` in the CLI).

### DIFF
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -2127,6 +2127,7 @@ std::shared_ptr<SourceFile> MainWindow::parseDocument(EditorInterface *editor)
     auto error = evaluatePython(fulltext_py, false);
     if (error.size() > 0) LOG(message_group::Error, Location::NONE, "", error.c_str());
     fulltext = "\n";
+    finishPython();
   }
 #endif  // ifdef ENABLE_PYTHON
 
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -605,6 +605,7 @@ int cmdline(const CommandLine& cmd)
     auto error = evaluatePython(fulltext_py, false);
     if (error.size() > 0) LOG(error.c_str());
     text = "\n";
+    finishPython();
   }
 #endif  // ifdef ENABLE_PYTHON
   text += "\n\x03\n" + commandline_commands;
--- a/src/python/pyopenscad.cc
+++ b/src/python/pyopenscad.cc
@@ -39,14 +39,14 @@ bool python_trusted;
 
 void PyObjectDeleter(PyObject *pObject) { Py_XDECREF(pObject); };
 
+PyThreadState *globalInterpreterThreadState = NULL;
+PyThreadState *subInterpreterThreadState = NULL;
+
 PyObjectUniquePtr pythonInitDict(nullptr, PyObjectDeleter);
 PyObjectUniquePtr pythonMainModule(nullptr, PyObjectDeleter);
-std::list<std::string> pythonInventory;
 bool pythonDryRun = false;
 std::shared_ptr<AbstractNode> python_result_node =
   nullptr; /* global result veriable containing the python created result */
-PyObject *python_result_obj = nullptr;
-bool pythonMainModuleInitialized = false;
 
 void PyOpenSCADObject_dealloc(PyOpenSCADObject *self)
 {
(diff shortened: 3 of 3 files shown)


gsohler:
would be awesome if you managed to really reinitizlize python without a segfault.
PS: most nasty libs are numpy, scipy and ctypes ...

damz:
What I can say is that this seems to work in basic tests. I haven't tried numpy at all.

damz:
Yeah, `numpy` fails on re-evaluation with:

```
RuntimeError: CPU dispatcher tracer already initlized
```

(which is numpy/numpy#28271)

damz:
The [C documentation for `PyDict_Next`](https://docs.python.org/3/c-api/dict.html#c.PyDict_Next) clearly states that you cannot mutate a dictionary while iterating it:

> The dictionary p should not be mutated during iteration. It is safe to modify the values of the keys as you iterate over the dictionary, but only so long as the set of keys does not change.

damz:
Sub-interpreters would be an option, but these are Python 3.12+ (released 2023). Any thoughts about adding that restriction to more recent versions of Python?

t-paul:
> Sub-interpreters would be an option, but these are Python 3.12+ (released 2023). Any thoughts about adding that restriction to more recent versions of Python?

Yes, fine by me. It will make some builds a bit more difficult and we probably need to disable python for older distributions. But if that means we get clean environments, that's worth it in my view.

coryrc:
I noticed some build environments are newer, but the appimage builder image (manually uploaded by Torsten to Docker, I'm not sure from where) has something older than Python 3.13 (because my draft PR failed only for appimage), so you'll need to update at least that. But, you know, perfect enemy of the good and all.

t-paul:
Yep, correct. This will need to build a newer Python for the AppImage build-env. https://github.com/openscad/openscad/wiki/Build-Matrix - but if this will allow clean restarts, it's very much worth the extra effort in my view.

damz:
I reworked this based on sub-interpreters.

It turns out that sub-interpreters have been there forever, it is sub-interpreters _with restrictions and independent GILs_ that are new to Python 3.12.

damz:
Marked as draft because this makes the situation with numpy worse. Now it doesn't load at all, not even once.

t-paul:
Right, a query for [Py_mod_multiple_interpreters](https://github.com/search?q=repo%3Anumpy%2Fnumpy%20Py_mod_multiple_interpreters&type=code) shows a lot of `Py_MOD_MULTIPLE_INTERPRETERS_NOT_SUPPORTED`.

So is actually forking a separate process the only option? I suppose not supporting numpy would be a huge drawback for the specific use in OpenSCAD. Looking at https://github.com/numpy/numpy/issues/24755 and https://github.com/numpy/numpy/issues/28271 there's ongoing work towards sub-interpreters but it also sounds like there's still some way to go.

MarcelKilgus:
Just randomly stumbled upon this issue. I have done the same thing for work one or two years ago and I can attest that it is simply impossible to initialize Python cleanly a second time in the same process (because of popular 3rd party libraries, Python itself is fine). Basically numpy people say nobody uses sub-interpreters, so no need to support them. And nobody uses sub-interpreters because numpy doesn't work in them, Catch 22.

With Python 3.14 sub-interpreters came into the focus once more as they are now available within Python itself (previously one had to use the C API). Maybe that finally moves the goalpost so far that numpy has to get its act together 🤷‍♂️

t-paul:
I did a simple test and forking and running python against 2 different numpy versions (one installed via Debian, the other via venv) and that worked. Now the question would be what other problems will that drag in. It certainly makes other things more difficult, but as the forked process has all the other stuff too, maybe it's just a matter of getting the result back into the main process?

gsohler:
I like the forking option as it removes the need to do
python-re-initialization . I believe we can even use the SAME memory for
parent and child and just need to ensure proper synchronization of data.
But I am not sure how that works with the Windows platform. Does Windows
API support "forking"  ?

PR #6229, opened 2025-09-24
Make GitHub workflow compatible with act

nomike:
It is possible to locally run github actions with [act](https://github.com/nektos/act).
Most of our actions fail though, as the `sudo` command is not available in the container act is using.

As `sudo` is not necessary in GitHub actions, this change removes it from all actions.

It also switches all remaining actions to the recently updated `uni-get-dependencies.sh` and removes the now no longer used `github-ci-linux-get-dependencies`


nomike:
Looks like sudo is a requirement on GitHub.
I've changed to code to gracefully install sudo when it is not yet there, which enables the workflows to run successfully in act and on GitHub.

However, Docker-Hub currently has an issue, that's why the circle-ci builds are failing and that's why I can't test-run the workflows with act right now.

image

See: https://www.dockerstatus.com/

nomike:
A merge commit snuck in here somehow. I removed it and rebased the PR on the current master.

It should be fit to be merged now, if you agree with it's contents of course.

kintel:
From a cursory glance it looks good to me, but @t-paul may be a bit more aware of nuances in this domain..

t-paul (approved):
I have not tried `act` yet, but looks good to me and reducing the number of scripts to maintain is nice.

t-paul:
Right, looks like the master branch build needs some more updates...

t-paul:
All checks on master branch are back to green now.

PR #6231, opened 2025-09-26
Examples docs

bitbasher:
i want this to work


bitbasher:
no .. how did this get here .. i DO NOT want this to work

README.md
```diff
@@ -159,7 +159,7 @@ Prerequisites:
 
 Install Dependencies:
 
-After building dependencies using one of the following options, follow the instructions in the *Compilation* section.
+There are two methods for building the dependencies: From Source OR Using Homebrew.
```
jordanbrown0, line 162: What's confusing about the original?  It says "using one of the following options", which doesn't give much margin for confusion.

src/gui/TabManager.cc
```diff
@@ -665,11 +665,39 @@ bool TabManager::saveAs(EditorInterface *edt, const QString& filepath)
...
 bool TabManager::saveACopy(EditorInterface *edt)
 {
   assert(edt != nullptr);
 
-  const auto dir = edt->filepath.isEmpty() ? _("Untitled.scad") : edt->filepath;
+  const auto dir;
+  if (edt->filepath.isEmpty()) dir = _("Untitled.scad");
+  else {
+    const QFileInfo info(edt->filepath);
+    info.setfile(info.absolutePath() % info.basename() % "_copy.scad");
```
jordanbrown0, line 689: It is best not to use string concatenation to form paths, because conventions vary from one platform to another.    (And does QFileInfo.absolutePath() including a trailing slash?)

Better is to use the library-provided mechanisms for assembling paths.

Perhaps:
```
info.setFile(info.absoluteDir(), info.basename() % "_copy.scad");
```

src/gui/TabManager.cc
```diff
@@ -665,11 +665,39 @@ bool TabManager::saveAs(EditorInterface *edt, const QString& filepath)
...
+ Otherwise append  "_copy" to the base file name.
+ The QFileINfo does not provide a method to update only the file
+ name so the info needs to be broken down and reassembled.
+ The new name is then tested for existance and the save is
+ aborted if the user declines the save.
+ The operation of the file dialog could delete the .scad suffix
+ so it is checked a last time before the save operation.
+
+ The name of the editor tab should NOT be changed
+ */
```
jordanbrown0, line 680: There seems to be a weak style pattern of using leading `*` on multi-line comments.
```
/*
 * some comment here
 */
```

At a minimum, indent the comment three spaces.

Please use complete sentences with periods.

"existance" is a spelling error, should be "existence".

`QFileINfo` should be `QFileInfo` with a lower-case N.  I personally would not say "The" in front of it; I would treat it as a proper noun.  Alternatively, I would write "The QFileInfo class ...".

src/gui/TabManager.cc
```diff
@@ -665,11 +665,39 @@ bool TabManager::saveAs(EditorInterface *edt, const QString& filepath)
...
 
-  const auto dir = edt->filepath.isEmpty() ? _("Untitled.scad") : edt->filepath;
+  const auto dir;
+  if (edt->filepath.isEmpty()) dir = _("Untitled.scad");
+  else {
+    const QFileInfo info(edt->filepath);
+    info.setfile(info.absolutePath() % info.basename() % "_copy.scad");
+    if (info.exists()) {
+      const auto text =
+        QString(_("%1 already exists.\nDo you want to replace it?")).arg(info.fileName());
```
jordanbrown0, line 692: It might be better to use `info.filePath()` so that it's absolutely clear what file you're about to overwrite.

src/gui/TabManager.cc
```diff
@@ -665,11 +665,39 @@ bool TabManager::saveAs(EditorInterface *edt, const QString& filepath)
...
+      if (QMessageBox::warning(par, par->windowTitle(), text, QMessageBox::Yes | QMessageBox::No,
+                               QMessageBox::No) != QMessageBox::Yes) {
+        return false;
+      }
+    }
+    dir = info.filepath();  // FIXME does this need tobe _(info.filepath()) ?
```
jordanbrown0, line 698: `_("whatever")` is a marker for internationalizing strings for translation, so no, it doesn't apply here.  I believe it almost never applies to anything other than a constant string.

src/gui/TabManager.cc
jordanbrown0: But:  is this prompt really required?  Isn't this about to give the user a Save-as style dialog with the _copy name, that will then lead to a "do you want to replace" dialog if the file exists?

examples/Old/README.md
```diff
@@ -0,0 +1,36 @@
...
+15 | |
+16 | |
+17 | |
+18 | |
+19 | |
+20 | |
+21 | |
+22 | |
+23 | |
+24 | |
```
jordanbrown0, line 36: This change doesn't seem to be part of the "Save a copy" fix and should probably be split away.

examples/README.md
```diff
@@ -0,0 +1,10 @@
+[![GitHub (master)](https://img.shields.io/github/checks-status/openscad/openscad/master.svg?logo=github&label=build&logoColor=black&colorA=f9d72c&style=plastic)](https://github.com/openscad/openscad/actions)
+[![CircleCI (master)](https://img.shields.io/circleci/project/github/openscad/openscad/master.svg?logo=circleci&logoColor=black&colorA=f9d72c&style=plastic)](https://circleci.com/gh/openscad/openscad/tree/master)
+[![Coverity Scan](https://img.shields.io/coverity/scan/2510.svg?colorA=f9d72c&logoColor=black&style=plastic)](https://scan.coverity.com/projects/2510)
+
+[![Visit our IRC channel](https://kiwiirc.com/buttons/irc.libera.chat/openscad.png)](https://kiwiirc.com/client/irc.libera.chat/#openscad)
+
+# What is OpenSCAD?
+<p><a href="https://opencollective.com/openscad/donate"><img align="right" src="https://opencollective.com/openscad/donate/button@2x.png?color=white" width="200"/></a>
+
+OpenSCAD is a software for creating solid 3D CAD objects, and we have a lot of examples that can help with learning how it works and how to use it.
```
jordanbrown0, line 10: This change doesn't seem to be part of the "Save a copy" fix and should probably be split away.

jordanbrown0:
And I didn't see that this was subsequently closed.

README.md
bitbasher: Hey Jordan, thanks for all the feedback, i appreciate you taking the time.

The original was unclear to me .. my fault for not taking enuf time to really absorb what was written, but i thot it could be clearer .. so i made it abundantly obvious .. i think it will help others in future

src/gui/TabManager.cc
bitbasher: i worked thru all of the misspelled method names in the process of learning Qt and working though all the type mismatches .. OpenSCAD really spoils one for that sort of coding issue

I hate the column of dots / stars along the side of a block comment because when you want to re-write the comment you have all those stars in the way .. I still have to do my cleanup pass on the code so i will get the periods in place and the spellings correct .. many updates to work thru before i can make a PR for this

src/gui/TabManager.cc
bitbasher: thanks .. noted

examples/Old/README.md
bitbasher: yep .. i should not be .. i made some mistakes applying changes and merging new PRs from the scad master into my fork .. it is a complete muddle so i have to clean it all up 
this is part of my organizing the examples .. i have a LOT of new ones that i want to add and i would like them all to be in version control .. i want to be able to refer to examples in GitHub from the documentation using some smart templates .. it is all being explored in my sandbox and may never see the light of day .. but it is fun, keeps me off the streets etc

examples/README.md
bitbasher: yep . it will be .. once i figure out how to get my GPG signature working again

src/gui/TabManager.cc
bitbasher: that line of code is a survivor from the original i think .. i will change it to filePath() and see if the added length of the path + name is less readable.
And yes .. the idea is that you are working on xyz.scad .. it has been saved but you go to make a copy .. the previous step made a new filename, xyz_copy.scad, for you .. but sees that file already exists and prompts the user about overwriting .. in case this should be xyz_copy2.scad to preserve the previous one.

src/gui/TabManager.cc
bitbasher: well i did a bunch of reading about how to form strings and assemble QOption flags .. the Best Practice is to us the overloaded '%' operator to concatenate QStrings .. one has to do #include <QtSringBuilder> 
line 689 is the result of about 30 tries to get info.setFile() to work the way i needed it to .. so it is not changing anymore
and a lot of that code was copy-pasted out of examples i found that i found online .. none of which were a good fit for what i was trying to do .. there was a lot of "learning experience" to get that my new TabManager.cc to compile but i got there in the end

src/gui/TabManager.cc
jordanbrown0: I wouldn't bother with incrementing numbers on the "save a copy" logic.  That seems like it's way into the weeds.

Just make it be equivalent to "Save as", adding "_copy" and not changing the current name.  If that file already exists, then (exactly as in "Save as") it should ask for confirmation after you step through the save-as dialog.

src/gui/TabManager.cc
jordanbrown0: >well i did a bunch of reading about how to form strings 

And if you were forming strings, that'd be a good way.  But you're not; you're forming paths, which are only sort of strings.  You're assuming that simply concatenating info.absolutePath() with info.basename() yields a correct path.  For Windows and UNIX, it will... if absolutePath() ends with a slash.  But does it?  Does the documentation promise that it will?

No, what you should do is to use the supplied path manipulation tools to extract the directory and the leaf name, then hack the leaf name as desired, then use the supplied path manipulation tools to glue them back together again.

examples/Old/README.md
jordanbrown0: Examples are good, but the point isn't quantity... the point is demonstrating particular points clearly.  Only rarely should there be two examples that demonstrate the same point.

PR #6232, opened 2025-09-26
Update pyfunctions.cc

gsohler:
added missing angle assignment


gsohler:
sorry, was deleteing too much when backporting from pythonscad. angles exist, but not scale ...

t-paul (approved):

PR #6237, opened 2025-09-29
Fix: Disable manifold tests when ENABLE_MANIFOLD is OFF

nomike:
These tests fail when ENABLE_MANIFOLD is disabled as they require the presence of the manifold library.
This patch conditionally disables them.

Should be quite straight forward to merge.


nomike:
The failing windows build is not related to this PR. It complains about not finding cmake.

But that means I can already cherry-pick my commit into PythonSCAD.

t-paul (approved):

PR #6249, opened 2025-10-04
reorganize tests/CMakeLists.txt

AaronVerDow:
This breaks up `tests/CMakeLists.txt` into multiple files so it should be easier to maintain. Everything was done using includes, so it will still function the same. 

Nothing should change for test output or results, besides ordering.


AaronVerDow:
I'm going to redo this from scratch; there were too many changes.

kintel:
It's tricky to get this right. I've tried a couple of times and given up after chasing the wrong abstraction. One of the challenges is how to define which test files exists and which tests should use them, combined  golden images being reused across tests. Ideally, the inputs could be organized as a big spreadsheet for better visibility, but expressing it in CMake is a bit annoying..

Issue #6250, opened 2025-10-05
macOS: lib3mf not installing correctly with Homebrew


kintel:
Since approx. Oct 2 2025, the macOS GitHub action no longer picks up lib3mf at build time on macOS.
It looks like it fails to install all files during Homebrew packaging; the executables and pkg-config files seem to be installed into the wrong location (see log below).
This sounds like either a CMake or Homebrew regression.

```
-- Installing: /private/tmp/lib3mf-20251004-6031-773ol/lib3mf-2.4.1/build/lib/pkgconfig/lib3mf.pc
-- Installing: /private/tmp/lib3mf-20251004-6031-773ol/lib3mf-2.4.1/build/lib/cmake/lib3mf/lib3mfConfig.cmake
-- Installing: /private/tmp/lib3mf-20251004-6031-773ol/lib3mf-2.4.1/build/lib/lib3mf.2.4.1.0.dylib
-- Installing: /private/tmp/lib3mf-20251004-6031-773ol/lib3mf-2.4.1/build/lib/lib3mf.2.dylib
-- Installing: /private/tmp/lib3mf-20251004-6031-773ol/lib3mf-2.4.1/build/lib/lib3mf.dylib
-- Installing: /opt/homebrew/Cellar/lib3mf/2.4.1/include/Bindings
-- Installing: /opt/homebrew/Cellar/lib3mf/2.4.1/include/Bindings/CppDynamic
-- Installing: /opt/homebrew/Cellar/lib3mf/2.4.1/include/Bindings/CppDynamic/lib3mf_dynamic.h
-- Installing: /opt/homebrew/Cellar/lib3mf/2.4.1/include/Bindings/CppDynamic/lib3mf_abi.hpp
-- Installing: /opt/homebrew/Cellar/lib3mf/2.4.1/include/Bindings/CppDynamic/lib3mf_dynamic.hpp
-- Installing: /opt/homebrew/Cellar/lib3mf/2.4.1/include/Bindings/CppDynamic/lib3mf_types.hpp
-- Installing: /opt/homebrew/Cellar/lib3mf/2.4.1/include/Bindings/Pascal
-- Installing: /opt/homebrew/Cellar/lib3mf/2.4.1/include/Bindings/Pascal/Unit_Lib3MF.pas
-- Installing: /opt/homebrew/Cellar/lib3mf/2.4.1/include/Bindings/NodeJS
[...]

PR #6251, opened 2025-10-05
Refactor Discretization Special Usage

coryrc:
First step for #4246 The discussion there explains the motivation and naming.

Currently every single function which uses any discretization specials ($fn, $fs, $fn) or calls anything using them must save and/or pass them along individually. This refactors the code toward having one class own the formulas and code to extract the control variables from the environment, thus making it simple to add a new discretization special.

This is almost a no-op; the Python version of the `text` function was not setting the defaults for unspecified variables like OpenSCAD does. I changed it to do so. If this is unwanted, it can trivially be removed, but it sure seems like it should be this way.

### DIFF
--- a/src/core/CurveDiscretizer.cc
+++ b/src/core/CurveDiscretizer.cc
@@ -1,4 +1,4 @@
-#include "core/TessellationControl.h"
+#include "core/CurveDiscretizer.h"
 
 #ifdef ENABLE_PYTHON
 #include "python/pyopenscad.h"
@@ -12,7 +12,7 @@
 
 #define F_MINIMUM 0.01
 
-TessellationControl::TessellationControl(const Parameters& parameters, const ModuleInstantiation *inst)
+CurveDiscretizer::CurveDiscretizer(const Parameters& parameters, const ModuleInstantiation *inst)
 {
   fn = parameters["$fn"].toDouble();
   fs = parameters["$fs"].toDouble();
@@ -74,7 +74,7 @@ void trySetVariable(PyObject *kwargs, const char *string_input, double *var_ptr)
   }
 }
 
-TessellationControl::TessellationControl(PyObject *kwargs)
+CurveDiscretizer::CurveDiscretizer(PyObject *kwargs)
 {
   setValuesFromPyMain();
   if (kwargs == nullptr || !PyDict_Check(kwargs)) {
@@ -86,7 +86,7 @@ TessellationControl::TessellationControl(PyObject *kwargs)
   trySetVariable(kwargs, "fa", &this->fa);
 }
 
-void TessellationControl::setValuesFromPyMain()
+void CurveDiscretizer::setValuesFromPyMain()
 {
   PyObject *mainModule = PyImport_AddModule("__main__");
   if (mainModule == nullptr) {
@@ -133,7 +133,7 @@ void TessellationControl::setValuesFromPyMain()
    Returns the number of subdivision of a whole circle, given radius and
    the three special variables $fn, $fs and $fa
  */
-std::optional<int> TessellationControl::circular_segments(double r, double angle_degrees) const
+std::optional<int> CurveDiscretizer::GetCircularSegmentCount(double r, double angle_degrees) const
 {
   // FIXME: It would be better to refuse to create an object. Let's do more strict error handling
   // in future versions of OpenSCAD
@@ -146,10 +146,10 @@ std::optional<int> TessellationControl::circular_segments(double r, double angle
                   1);
 }
 
-std::ostream& operator<<(std::ostream& stream, const std::shared_ptr<TessellationControl>& f)
+std::ostream& operator<<(std::ostream& stream, const std::shared_ptr<CurveDiscretizer>& f)
 {
   if (!f) {
-    return stream << "[null TessellationControl]";
+    return stream << "[null CurveDiscretizer]";
   }
   stream << "$fn = " << f->fn << ", $fa = " << f->fa << ", $fs = " << f->fs;
   return stream;
--- a/src/core/CurveDiscretizer.h
+++ b/src/core/CurveDiscretizer.h
@@ -11,51 +11,51 @@
 class Parameters;
 class ModuleInstantiation;
 
-class TessellationControl
+class CurveDiscretizer
 {
 public:
-  TessellationControl(const Parameters& parameters, const ModuleInstantiation *inst = nullptr);
+  CurveDiscretizer(const Parameters& parameters, const ModuleInstantiation *inst = nullptr);
 #ifdef ENABLE_PYTHON
   /**
    * Extract discretization values from environment if possible,
    * then override any which are explicitly set as keyword arguments.
    */
-  TessellationControl(PyObject *kwargs);
+  CurveDiscretizer(PyObject *kwargs);
 #endif
 
   /**
-   * Create a TessellationControl for a spot in dxfdim.cc that used a hardcoded value.
+   * Create a CurveDiscretizer for a spot in dxfdim.cc that used a hardcoded value.
    * These values of 36,0,0 go back to the first commit in Github.
    * Unknown why it is that.
    */
-  static std::shared_ptr<TessellationControl> DefaultForDxf()
+  static std::shared_ptr<CurveDiscretizer> DefaultForDxf()
   {
-    return std::make_shared<TessellationControl>(std::move(TessellationControl(36, 0, 0)));
+    return std::make_shared<CurveDiscretizer>(std::move(CurveDiscretizer(36, 0, 0)));
   }
 
   /**
    * Calculate segments for a circle or circular arc.
    */
-  std::optional<int> circular_segments(double r, double angle_degrees = 360.0) const;
+  std::optional<int> GetCircularSegmentCount(double r, double angle_degrees = 360.0) const;
 
   /**
    * @brief Calculate segments for a path.
    * Currently only uses $fn. Won't use $fs or $fa, but future variables could be used.
    */
-  unsigned long path_segments(unsigned long minimum) const
+  unsigned long GetPathSegmentCount(unsigned long minimum) const
   {
     unsigned long result = (fn > 3.0) ? static_cast<unsigned long>(fn) : 3;
     return std::max(result, minimum);
   }
 
-  friend std::ostream& operator<<(std::ostream& stream, const std::shared_ptr<TessellationControl>& f);
+  friend std::ostream& operator<<(std::ostream& stream, const std::shared_ptr<CurveDiscretizer>& f);
   bool IsFnSpecifiedAndOdd() { return static_cast<int>(fn) & 1; }
 
 private:
-  TessellationControl(double fn, double fs, double fa) : fn(fn), fs(fs), fa(fa) {}
+  CurveDiscretizer(double fn, double fs, double fa) : fn(fn), fs(fs), fa(fa) {}
   double fn, fs, fa;
 #ifdef ENABLE_PYTHON
   void setValuesFromPyMain();
 #endif
 };
-std::ostream& operator<<(std::ostream& stream, const std::shared_ptr<TessellationControl>& f);
+std::ostream& operator<<(std::ostream& stream, const std::shared_ptr<CurveDiscretizer>& f);
--- a/src/core/FreetypeRenderer.cc
+++ b/src/core/FreetypeRenderer.cc
@@ -41,7 +41,7 @@
 
 #include "FontCache.h"
 #include "core/DrawingCallback.h"
-#include "core/TessellationControl.h"
+#include "core/CurveDiscretizer.h"
 #include "utils/calc.h"
 
 #include FT_OUTLINE_H
@@ -251,7 +251,7 @@ void FreetypeRenderer::Params::detect_properties()
   hb_direction_t hbdirection = detect_direction(hbscript);
   set_direction(hb_direction_to_string(hbdirection));
 
-  auto segments = tessFIXME->circular_segments(size).value_or(3);
+  auto segments = discretizer->GetCircularSegmentCount(size).value_or(3);
   // The curved segments of most fonts are relatively short, so
   // by using a fraction of the number of full circle segments
   // the resolution will be better matching the detail level of
@@ -266,7 +266,7 @@ std::ostream& operator<<(std::ostream& stream, const FreetypeRenderer::Params& p
                 << ", spacing = " << params.spacing << ", font = \"" << params.font
                 << "\", direction = \"" << params.direction << "\", language = \"" << params.language
                 << (params.script.empty() ? "" : "\", script = \"") << params.script << "\", halign = \""
-                << params.halign << "\", valign = \"" << params.valign << "\", " << params.tessFIXME;
+                << params.halign << "\", valign = \"" << params.valign << "\", " << params.discretizer;
 }
 
 const FontFacePtr FreetypeRenderer::Params::get_font_face() const
@@ -309,7 +309,7 @@ void FreetypeRenderer::Params::set(Parameters& parameters)
   (void)parameters.valid("halign", Value::Type::STRING);
   (void)parameters.valid("valign", Value::Type::STRING);
 
-  tessFIXME = std::make_shared<TessellationControl>(parameters);
+  discretizer = std::make_shared<CurveDiscretizer>(parameters);
 
   set_size(parameters.get("size", 10.0));
   set_text(parameters.get("text", ""));
@@ -325,7 +325,7 @@ void FreetypeRenderer::Params::set(Parameters& parameters)
 #ifdef ENABLE_PYTHON
 void FreetypeRenderer::Params::set(PyObject *kwargs)
 {
-  tessFIXME = std::make_shared<TessellationControl>(kwargs);
+  discretizer = std::make_shared<CurveDiscretizer>(kwargs);
 
   set_size(10.0);
   set_text("");
--- a/src/core/FreetypeRenderer.cc
+++ b/src/core/FreetypeRenderer.cc
@@ -41,6 +41,7 @@
 
 #include "FontCache.h"
 #include "core/DrawingCallback.h"
+#include "core/TessellationControl.h"
 #include "utils/calc.h"
 
 #include FT_OUTLINE_H
@@ -250,13 +251,22 @@ void FreetypeRenderer::Params::detect_properties()
   hb_direction_t hbdirection = detect_direction(hbscript);
   set_direction(hb_direction_to_string(hbdirection));
 
-  auto segments = Calc::get_fragments_from_r(size, fn, fs, fa);
+  auto segments = tessFIXME->circular_segments(size).value_or(3);
   // The curved segments of most fonts are relatively short, so
   // by using a fraction of the number of full circle segments
   // the resolution will be better matching the detail level of
   // other objects.
   auto text_segments = std::max(segments / 8 + 1, 2);
   set_segments(text_segments);
+}
+
+std::ostream& operator<<(std::ostream& stream, const FreetypeRenderer::Params& params)
+{
+  return stream << "text = \"" << params.text << "\", size = " << params.size
+                << ", spacing = " << params.spacing << ", font = \"" << params.font
+                << "\", direction = \"" << params.direction << "\", language = \"" << params.language
+                << (params.script.empty() ? "" : "\", script = \"") << params.script << "\", halign = \""
+                << params.halign << "\", valign = \"" << params.valign << "\", " << params.tessFIXME;
 }
 
 const FontFacePtr FreetypeRenderer::Params::get_font_face() const
@@ -299,9 +309,7 @@ void FreetypeRenderer::Params::set(Parameters& parameters)
   (void)parameters.valid("halign", Value::Type::STRING);
   (void)parameters.valid("valign", Value::Type::STRING);
 
-  set_fn(parameters["$fn"].toDouble());
-  set_fa(parameters["$fa"].toDouble());
-  set_fs(parameters["$fs"].toDouble());
+  tessFIXME = std::make_shared<TessellationControl>(parameters);
 
   set_size(parameters.get("size", 10.0));
   set_text(parameters.get("text", ""));
@@ -313,6 +321,23 @@ void FreetypeRenderer::Params::set(Parameters& parameters)
   set_halign(parameters.get("halign", "default"));
   set_valign(parameters.get("valign", "default"));
 }
+
+#ifdef ENABLE_PYTHON
+void FreetypeRenderer::Params::set(PyObject *kwargs)
+{
+  tessFIXME = std::make_shared<TessellationControl>(kwargs);
+
+  set_size(10.0);
+  set_text("");
+  set_spacing(1.0);
+  set_font("");
+  set_direction("");
+  set_language("en");
+  set_script("");
+  set_halign("default");
+  set_valign("default");
+}
+#endif
 
 FreetypeRenderer::ShapeResults::ShapeResults(const FreetypeRenderer::Params& params)
 {
--- a/src/core/OffsetNode.h
+++ b/src/core/OffsetNode.h
@@ -1,25 +1,30 @@
 #pragma once
 
+#include <memory>
+#include <string>
+
 #include "core/ModuleInstantiation.h"
 #include "core/node.h"
-#include "core/TessellationControl.h"
 #include "clipper2/clipper.h"
 
-#include <string>
+class TessellationControl;
 
 class OffsetNode : public AbstractPolyNode
 {
 public:
   VISITABLE();
-  OffsetNode(const ModuleInstantiation *mi, TessellationControl&& tess)
-    : AbstractPolyNode(mi), tess(tess)
+  OffsetNode(const ModuleInstantiation *mi, std::shared_ptr<TessellationControl> tessFIXME)
+    : AbstractPolyNode(mi), tessFIXME(std::move(tessFIXME))
   {
   }
+
+  ~OffsetNode();
+
   std::string toString() const override;
   std::string name() const override { return "offset"; }
 
   bool chamfer{false};
-  TessellationControl tess;
+  std::shared_ptr<TessellationControl> tessFIXME;
   double delta{1};
   double miter_limit{1000000.0};  // currently fixed high value to disable chamfers with jtMiter
   Clipper2Lib::JoinType join_type{Clipper2Lib::JoinType::Round};
--- a/src/core/OffsetNode.h
+++ b/src/core/OffsetNode.h
@@ -7,14 +7,14 @@
 #include "core/node.h"
 #include "clipper2/clipper.h"
 
-class TessellationControl;
+class CurveDiscretizer;
 
 class OffsetNode : public AbstractPolyNode
 {
 public:
   VISITABLE();
-  OffsetNode(const ModuleInstantiation *mi, std::shared_ptr<TessellationControl> tessFIXME)
-    : AbstractPolyNode(mi), tessFIXME(std::move(tessFIXME))
+  OffsetNode(const ModuleInstantiation *mi, std::shared_ptr<CurveDiscretizer> discretizer)
+    : AbstractPolyNode(mi), discretizer(std::move(discretizer))
   {
   }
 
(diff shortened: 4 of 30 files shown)


coryrc:
AppImage build is failing due to older build of Python; the image used is `openscad/appimage-x86_64-base:latest`.

https://hub.docker.com/r/openscad/appimage-x86_64-base

coryrc:
The test failure is due to https://github.com/openscad/openscad/issues/6250

coryrc:
> The python specific parts should not be mixed with the core implementation.

Done.

src/core/CurveDiscretizer.cc
```diff
@@ -0,0 +1,78 @@
...
+#include <algorithm>
+
+#include "core/ModuleInstantiation.h"
+#include "core/Parameters.h"
+#include "geometry/Grid.h"
+#include "utils/printutils.h"
+
+#define F_MINIMUM 0.01
+
+CurveDiscretizer::CurveDiscretizer(const Parameters& parameters, const ModuleInstantiation *inst)
```
kintel: While in practice, the discretizer is created from a ModuleInstantiation, it not technically necessary to pass it when we only care about the location. We can test a location (`loc.isNone()`) where needed.

src/core/CurveDiscretizer.h
```diff
@@ -0,0 +1,56 @@
...
+{
+public:
+  CurveDiscretizer(const Parameters& parameters, const ModuleInstantiation *inst = nullptr);
+
+  /**
+   * @brief Provide a function which returns the value for any named special.
+   * Should work on "fn" and not "$fn".
+   * Will not hold a reference to valueLookup after object constructed.
+   */
+  CurveDiscretizer(std::function<std::optional<double>(const char *)> valueLookup);
```
kintel: It does feel a bit messy to get into Python-specific conventions here. Any way this could be dealt with or bridged by code in the python/ folder?

src/core/CurveDiscretizer.cc
```diff
@@ -0,0 +1,78 @@
...
+  if (fn > 0.0) {
+    result = std::ceil(fn >= 3 ? fn : 3) * std::fabs(angle_degrees) / 360.0;
+  } else {
+    result = std::ceil(std::max(std::min(360.0 / fa, r * 2 * M_PI / fs), 5.0)) *
+             std::fabs(angle_degrees) / 360.0;
+  }
+  return std::max(1, static_cast<int>(std::ceil(result)));
+}
+
+std::ostream& operator<<(std::ostream& stream, const std::shared_ptr<CurveDiscretizer>& f)
```
kintel: Not sure I'm a huge fan of passing shared_ptrs to stream operators. I assume you did it to make client code look nicer, but it just feels wrong.

src/core/CurveDiscretizer.h
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * Create a CurveDiscretizer for a spot in dxfdim.cc that used a hardcoded value.
+   * These values of 36,0,0 go back to the first commit in Github.
+   * Unknown why it is that.
+   */
+  static std::shared_ptr<CurveDiscretizer> DefaultForDxf()
+  {
+    return std::make_shared<CurveDiscretizer>(std::move(CurveDiscretizer(36, 0, 0)));
+  }
```
kintel: I don't think special magic values belong here, should find a way of cleaning that up now that we're doing some good refactoring.
Also, std::move() is superfluous here.

src/core/CurveDiscretizer.h
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * Calculate segments for a circle or circular arc.
+   */
+  std::optional<int> GetCircularSegmentCount(double r, double angle_degrees = 360.0) const;
```
kintel: Nit: We generally don't follow the UpperCamelCase naming scheme for functions and methods; prefer lowerCamelCase.

There are probably scattered violations - we'll get that cleaned up later.

src/core/FreetypeRenderer.cc
```diff
@@ -299,10 +306,7 @@ void FreetypeRenderer::Params::set(Parameters& parameters)
   (void)parameters.valid("halign", Value::Type::STRING);
   (void)parameters.valid("valign", Value::Type::STRING);
 
-  set_fn(parameters["$fn"].toDouble());
-  set_fa(parameters["$fa"].toDouble());
-  set_fs(parameters["$fs"].toDouble());
-
+  discretizer = std::make_shared<CurveDiscretizer>(parameters);
```
kintel: Generally, is it worth messing with shared_ptr on order to store 3 doubles? I can see that for small object s with lots of instances it can make a difference, but passing by value and move/copy might just be cleaner in most cases.

kintel (commented):
Just had a quick glance and left some leading comments.
I appreciate the cleanup, and think it's worth doing it right.

src/core/CurveDiscretizer.cc
coryrc: `inst` defaults to `nullptr` in the header, so that indeed works and is already used. Hmm, so would it be preferred to have two constructors then and not default `inst` to `nullptr`?

src/core/CurveDiscretizer.h
coryrc: I'm not sure how to do that; do you mean by having them configurable somewhere?
Will remove `std::move`.

src/core/CurveDiscretizer.h
coryrc: Aha, thanks. There are many styles and I wasn't sure :)

src/core/FreetypeRenderer.cc
coryrc: I was trying to avoid header pollution. Value types require including `core/CurveDiscretizer.h` in dozens of header files instead of none. I now forget why a reference didn't work; I think it had something to do with requiring an explicitly-defined destructor in every class which uses the reference.

src/core/CurveDiscretizer.cc
coryrc: I did it because I am not that familiar with C++ norms :-). I had two before, actually, this and via reference, and since I made everything use shared_ptr, I figured just keep this one because smart pointers are safer.

So it sounds preferred to instead change the prints from `<< discretizer <<` to `<< *discretizer <<` and update this to use, uh, probably a const reference?

src/core/CurveDiscretizer.h
kintel: I'm saying find a way of keeping DXF-specific stuff out of this class, and ideally clean it up/change it so it's not longer specific.

src/core/CurveDiscretizer.h
coryrc: Well, the point was to centralize the definitions and calculations of the discretization parameters, so that's somewhat inevitable?

I did move all Python types out, so technically this interface could replace `CurveDiscretizer(const Parameters& parameters, const ModuleInstantiation *inst)` with a helper function, except for the defaults being different between environments.

Or maybe you're thinking the other way, and the Python code should create a `Parameters`? I think that would be messier.

But I wanted to at least get all the "magic constants" into one place first.

src/core/FreetypeRenderer.cc
kintel: Just a copy if fine too. With references you need to care about ownership and lifetimes.
I wouldn't care too much about header pollution, at least I wouldn't complicate stuff as a compile-time optimization.

src/core/CurveDiscretizer.cc
kintel: You could add a location parameter and default it to NONE.

src/core/CurveDiscretizer.h
kintel: We can revisit this later; there might be a nicer way to manage this.

src/core/CurveDiscretizer.cc
kintel: yup, and if you combine this with my comment to perhaps not use so many shared pointers, you can ditch the `*` ;)

src/core/CurveDiscretizer.cc
coryrc: Oh, I misread your comment. I get it now.

src/core/CurveDiscretizer.h
coryrc: I think I'm getting it. I could have a constructor `CurveDiscretizer(double segmentsPerCircle)` and then the `36` moves back to `dxfdim.cc`, but we're still avoiding any other code manually parsing fn/fa/fs. How does this sound?

src/core/FreetypeRenderer.cc
coryrc: You would if you were iterating on this refactor on my laptop!

I'll change it to value type -- after I make all the other changes because it 10x+ the time for update/compile.

src/core/OffsetNode.h
```diff
@@ -1,21 +1,31 @@
...
 {
 public:
   VISITABLE();
-  OffsetNode(const ModuleInstantiation *mi) : AbstractPolyNode(mi) {}
+  OffsetNode(const ModuleInstantiation *mi, std::shared_ptr<CurveDiscretizer> discretizer)
+    : AbstractPolyNode(mi), discretizer(std::move(discretizer))
+  {
+  }
+
+  ~OffsetNode();
```
coryrc: I think this is an artifact of me testing with reference type and can be removed.

src/core/RotateExtrudeNode.cc
```diff
@@ -83,6 +80,10 @@ std::shared_ptr<AbstractNode> builtin_rotate_extrude(const ModuleInstantiation *
 
 }  // namespace
 
+// Needs the full definition of CurveDiscretizer
```
coryrc: This also can probably be removed.

src/core/CurveDiscretizer.cc
coryrc: Oh, yes, thanks, I would have been `sed`ing twice to add and eventually remove `*` after a facepalm...

src/core/CurveDiscretizer.h
coryrc: I'm more troubled by the mix of snake_case vs lowerCamelCase, but I tried to lean toward the latter

src/core/CurveDiscretizer.cc
coryrc: All fixed.

src/core/FreetypeRenderer.cc
coryrc: Now a value type everywhere.

src/core/CurveDiscretizer.cc
coryrc: It's now:

```
  CurveDiscretizer(const Parameters& parameters);

  CurveDiscretizer(const Parameters& parameters, const Location& loc);
```

Callers have all been updated.

src/core/CurveDiscretizer.h
```diff
@@ -0,0 +1,53 @@
...
+  /**
+   * Calculate segments for a circle or circular arc.
+   */
+  std::optional<int> getCircularSegmentCount(double r, double angle_degrees = 360.0) const;
+
+  /**
+   * @brief Calculate segments for a path.
+   * Currently only uses $fn. Won't use $fs or $fa, but future variables could be used.
+   */
+  unsigned long getPathSegmentCount(unsigned long minimum) const
```
kintel: One of the getters return int, the other unsigned long. Perhaps align the two?

src/core/CurveDiscretizer.h
```diff
@@ -0,0 +1,53 @@
...
+  {
+    unsigned long result = (fn > 3.0) ? static_cast<unsigned long>(fn) : 3;
+    return std::max(result, minimum);
```
kintel: would smth like `return std::max({fn, 3, minimum});` with appropriate casting work as well?

src/core/CurveDiscretizer.h
kintel: Then again, it's a bit misaligned that one of the Count functions has a minimum and the other doesn't.

src/core/CurveDiscretizer.cc
```diff
@@ -0,0 +1,86 @@
...
+#include "geometry/Grid.h"
+#include "utils/printutils.h"
+
+#define F_MINIMUM 0.01
+
+CurveDiscretizer::CurveDiscretizer(const Parameters& parameters, const Location& loc)
+{
+  fn = parameters["$fn"].toDouble();
+  fs = parameters["$fs"].toDouble();
+  fa = parameters["$fa"].toDouble();
```
kintel, line 21: Here's a nice C++ feature to save some repetition:
```
CurveDiscretizer::CurveDiscretizer(const Parameters& parameters, const Location& loc)
  : CurveDiscretizer(parameters) {
  // range checks here
}
```

src/geometry/rotate_extrude.cc
```diff
@@ -100,9 +101,8 @@ std::unique_ptr<Geometry> rotatePolygon(const RotateExtrudeNode& node, const Pol
 
   // # of sections. For closed rotations, # vertices is thus fragments*outline_size. For open
   // rotations # vertices is (fragments+1)*outline_size.
-  const auto num_sections = (unsigned int)std::ceil(fmax(
-    Calc::get_fragments_from_r(max_x - min_x, node.fn, node.fs, node.fa) * std::abs(node.angle) / 360,
-    1));
+  const int num_sections =
+    node.discretizer.getCircularSegmentCount(max_x - min_x, node.angle).value_or(5);
```
kintel: Where did `5` come from?

src/core/CurveDiscretizer.h
coryrc: I won't repeat myself in the other posts, but I was trying to minimize any change to behaviour, however unlikely. It seems you'd prefer I clean up some stuff, which I don't mind. Doing so now.

In this case it was an unsigned long due to https://github.com/openscad/openscad/commit/bd20fe6bf8122ce01f9f45ebe501845c8f77d5f9 but it definitely wasn't necessary because it was a fixed value of 20 :). I'm changing to match now. I will be keeping std::optional for the other method because it can fail.

src/core/CurveDiscretizer.cc
coryrc: The value is already changed to F_MINIMUM, so that won't work.

I considered other options but WET seemed better in this case.

Alternatively is using pointer Location and doing null check like before.

src/geometry/rotate_extrude.cc
coryrc: Well, this behaviour is maybe a little different. It'll only be used if $fn is invalid or max_x and min_x are the same. In those cases, we used to multiply 3 by the angle/360. 5 is used with $fa and $fs calcs, so I thought 5 was a better minimum for this edge case -- which probably isn't anything useful to begin with -- since angle can be also more than 360.

coryrc: Wait, how about we follow through on the FIXME from the original code and why I chose optional in the first place: refuse to create the geometry?

src/core/CurveDiscretizer.cc
kintel: Fair, my idea doesn't do much good in this case

src/core/CurveDiscretizer.h
kintel: sg, this is minor stuff. I just ask questions now so we don't have to ask them later :) 
Feel free to push back if you want to move on

src/core/CurveDiscretizer.h
```diff
@@ -35,12 +35,14 @@ class CurveDiscretizer
...
    */
-  unsigned long getPathSegmentCount(unsigned long minimum) const
+  int getPathSegmentCount() const
   {
-    unsigned long result = (fn > 3.0) ? static_cast<unsigned long>(fn) : 3;
-    return std::max(result, minimum);
+    // Prior to https://github.com/openscad/openscad/commit/f5816258db263408a7aa2feec1fafffe77644662
+    // fn was set to a fixed value of 20 where this is now used.
+    // The author decided it should never be smaller than this original value.
+    return std::max(static_cast<int>(fn), 20);
```
kintel: I think we have a small miscommunication: I think any magic constant, like `20` here, should strictly belong int the client code, unless it's a universally valid constant (like `3` used in some places).

src/geometry/rotate_extrude.cc
kintel: Ideally, but we just have to take care not to break any existing use-cases. ..and if we decide to break something on purpose, we should make it an isolated change and make sure to track that in release notes or smth.

src/core/CurveDiscretizer.h
coryrc: I'd rather any magic be contained to this class, but you previously were clear you didn't want that. I apologize. I'll use max() in the caller code

kintel: Yeah, I feel pretty strongly that core components like this shouldn't really leak application-specific magic constants. This would also allow us to later unit-test them pretty effectively.

..but as with everything else in OpenSCAD, it's all an effort vs. impact tradeoff.

src/core/CurveDiscretizer.h
coryrc: I'd rather clean up, so as long as you continue to engage and we make progress, I'm happy to collaborate on implementing them.

src/geometry/rotate_extrude.cc
coryrc: Good point. Given the context, are you alright with 5? I think 1 also makes some sense. Or to be exactly the same, `max(1,3*angle/360)`, but I hope not a lot of people set $fn="banana"...

src/libsvg/path.cc
```diff
@@ -40,6 +40,7 @@
 #include <boost/tokenizer.hpp>
 #include <boost/algorithm/string.hpp>
 
+#include "core/CurveDiscretizer.h"
```
kintel: Let's bring in @t-paul to comment on this in the context of libsvg not being dependent on OpenSCAD itself.

t-paul:
I think having `libsvg` compiling separate is useful for many reasons including unit tests which created this discussion, so we should try to keep it this way.

Most of the OpenSCAD specifics are already in https://github.com/openscad/openscad/blob/master/src/io/import_svg.cc intentionally.

The ideal solution is probably not something that can be done right now. That would be moving the conversion to polygons completely out of the library or to a separate part that can be parameterized.

For the time being, it should be easy to introduce a simple callback that is passed into the library that provides the needed calculation calls implemented by the caller.

kintel:
> I think having `libsvg` compiling separate

Perhaps we should update the build system to actually enforce this by building libsvg as a separate target. Out of scope for this PR, but could be done as part of the whole unit test discussion.

src/geometry/rotate_extrude.cc
kintel: The existing behavior is a bit subtle, as it differs depending on whether $fn was given or $fs/$fa. Setting it to e.g. `0/0` will trigger different behavior. This is not too unlikely (albeit probably considered a user error).
I think the big question will be: Should we introduce any changes in this PR or make it a clean re-implementation?

coryrc:
> I think having `libsvg` compiling separate is useful for many reasons including unit tests which created this discussion, so we should try to keep it this way.
> 
> Most of the OpenSCAD specifics are already in https://github.com/openscad/openscad/blob/master/src/io/import_svg.cc intentionally.
> 
> For the time being, it should be easy to introduce a simple callback that is passed into the library that provides the needed calculation calls implemented by the caller.

I can make that happen.

coryrc:
> > I think having `libsvg` compiling separate
> 
> Perhaps we should update the build system to actually enforce this by building libsvg as a separate target. Out of scope for this PR, but could be done as part of the whole unit test discussion.
> 

I'll try in unit test PR; if it's easy I'll do it.

src/libsvg/path.cc
coryrc: > For the time being, it should be easy to introduce a simple callback that is passed into the library that provides the needed calculation calls implemented by the caller.

Done

src/geometry/rotate_extrude.cc
coryrc: Well, it's truly different if $fn="banana", because it might be 5 instead of 3 or 1. So I made it exactly the same in that case, and someone can open a PR to take advantage of the optional to not generate geometry when given invalid inputs.

src/geometry/rotate_extrude.cc
```diff
@@ -100,9 +101,8 @@ std::unique_ptr<Geometry> rotatePolygon(const RotateExtrudeNode& node, const Pol
 
   // # of sections. For closed rotations, # vertices is thus fragments*outline_size. For open
   // rotations # vertices is (fragments+1)*outline_size.
-  const auto num_sections = (unsigned int)std::ceil(fmax(
-    Calc::get_fragments_from_r(max_x - min_x, node.fn, node.fs, node.fa) * std::abs(node.angle) / 360,
-    1));
+  const int num_sections = node.discretizer.getCircularSegmentCount(max_x - min_x, node.angle)
+                             .value_or(std::max(1, static_cast<int>(std::fabs(node.angle) / 360 * 3)));
```
kintel, line 105: At this point I feel we're just arbitrarily changing the formula for this case.
As discussed, we've kind of painted ourselves into a bit of a corner by accepting bad input and using a fallback instead of issuing an error. But that's out of scope to address here.
I'm probably being a bit too conservative wrt. changing this behavior - let's try to get this merged and do a follow-up ticket

coryrc: This should be the same formula as before. Followup filed: #6351

coryrc:
I'm going to stop adding to it now so we can get it in before someone touches something it is also changing...

src/core/CurveDiscretizer.cc
```diff
@@ -0,0 +1,400 @@
...
+      unsigned int edge_segments =
+        static_cast<unsigned int>(std::ceil((v1 - v0).norm() * max_scale / fs));
+      add_segmented_edge(o2, v0, v1, edge_segments);
+      v0 = v1;
+    }
+  }
+  return o2;
+}
+
+Outline2d CurveDiscretizer::splitOutline(const Outline2d& o, double twist, double scale_x,
```
kintel, line 356: I'm not convinced this function belongs here, but let's leave that for another refactoring job.

src/geometry/linear_extrude.cc
```diff
@@ -426,52 +272,20 @@ std::unique_ptr<Geometry> extrudePolygon(const LinearExtrudeNode& node, const Po
...
-        } else {
-          seg_poly.addOutline(
-            splitOutlineByFn(o, node.twist, node.scale_x, node.scale_y, node.fn, num_slices));
-        }
-      }
-    } else {  // $fs and $fa based segmentation
-      auto fa_segs = static_cast<unsigned int>(std::ceil(360.0 / node.fa));
+
+  // We skip segmentation if the user passed in segments=0
+  if (!(node.has_segments && node.segments == 0)) {
```
kintel, line 277: This logic looks different; !node.has_segments can go into this branch now. Why didn't tris trigger any failed tests?

coryrc: Ha, I had to go through that like 10 times to be sure.

Start at https://github.com/openscad/openscad/blob/b0a4274914607db2a03dcfb5cdd7b98a73553e63/src/geometry/linear_extrude.cc#L430

There is only that if/else-if between `seg_poly` and checking `is_segmented`:

```
  if (node.has_segments) {
    if (node.segments > 0 ) {
      ...
      is_segmented = true;
    }
  } else if (non_linear) {
    ...
    is_segmented = true;
  }
```

`has_segments` means the user provided a **numeric** parameter to the `segments` argument. -1, 0, 1e16, whatever. `node.segments` is then clamped on the low end to zero or is zero when has_segments is false.

Inside "if has_segments {", there is only a single child statement: `if (node.segments > 0) {`, and inside that `is_segmented` is set to true. So that means if has_segments is true and node.segments is zero, then we don't do any segmentation. Hence I changed the logic to `!(node.has_segments && node.segments == 0)`. All other paths through set is_segmented to true.

So to preserve the existing behaviour, inside my new if, we segment if node.segments > 0 (which can only be true if node.has_segments is true) or non-linear; the same cases as before.

So `node.discretizer.splitOutline` is only called in cases where we would have segmented before. So it takes the logic from both branches of the `if/else-if` in order:

1. If node.segments > 0, segment with it. (from the first branch)
2. If $fn is set, segment with it (first half of the else-if)
3. Segment with $fa/$fs (the rest of the else-if).

except now it's much easier to follow because we haven't mixed *whether* to do it with *how* to do it.

kintel: Ah, gotcha, I missed that segments was initialized to zero and safe to access if !has_segments.
This code is tricky to follow : /

kintel (approved):

coryrc:
@t-paul please approve, merging is blocked because you "requested changes" originally (and I obliged:) )

YoSTEALTH:
`fn` inside `cylinder()` (maybe others) isn't working anymore, its commonly used/needed feature, please fix. Thanks

coryrc:
> `fn` inside `cylinder()` (maybe others) isn't working anymore, its commonly used/needed feature, please fix. Thanks

@YoSTEALTH Please include a little more detail than that.

YoSTEALTH:
Using:
`cylinder(h=2, d=7.25 + 1/8 + 1/64, fn=20).rotate([0, 0, 9]).color("brown")`
Raises:
```
Traceback (most recent call last):
  File "<string>", line 186, in <module>
  File "<string>", line 104, in bottom_holder
TypeError: Error during parsing cylinder(h,r|r1+r2|d1+d2)
```
It was working before, removing `fn` will work without error.

edit: OpenSCAD version 2025.11.27.ai29484 (git 2f0df8c3)

jordanbrown0:
That's PythonSCAD, not OpenSCAD.  @gsohler should take a look at it.

gsohler:
IMHO It's extra effort to create hexagons e.g. if fn is not available to
sphere, cylinder and circle anymore.

previously you wrote

cylinder(r=5,h=5, fn=6).show()

now you have to work with global fn instead

fn=6
cylinder(r=5,h=5).show()
fn=0

I regret the decision of coryrc to remove this parameter along with his
Discretizer effort.

PythonSCAD did not merge this loss and i feel, openscad should do that
either


My little 2 cents

coryrc:
> Using: `cylinder(h=2, d=7.25 + 1/8 + 1/64, fn=20).rotate([0, 0, 9]).color("brown")` Raises:
> 
> ```
> Traceback (most recent call last):
>   File "<string>", line 186, in <module>
>   File "<string>", line 104, in bottom_holder
> TypeError: Error during parsing cylinder(h,r|r1+r2|d1+d2)
> ```
> 
> It was working before, removing `fn` will work without error.
> 
> edit: OpenSCAD version 2025.11.27.ai29484 (git [2f0df8c](https://github.com/openscad/openscad/commit/2f0df8c39e359a0a6e6f7fb77b796baeb1f7dd55))

Please file a new issue with full repro instructions. I tried your input and got:

```
ERROR: Traceback (most recent call last):
  File "<string>", line 1, in <module>
NameError: name 'cylinder' is not defined
```

so I'll need to more information.

coryrc:
> I regret the decision of coryrc to remove this parameter along with his Discretizer effort. PythonSCAD did not merge this loss and i feel, openscad should do that either

@gsohler I did not intend to remove the parameter and added significant code to keep it working.

Given there are zero tests (see https://github.com/openscad/openscad/issues/6377) you can't expect others to not break things.

gsohler:
not blaming anybody.

just suggesting to add back the fn parameter.

There are some tests which can aid as templates in pythonscad code and i
even mentioned them explicitly in IRC
Unfortunately relevant people appear to be deaf on that :(

YoSTEALTH:
Looks like changes are not intentional.

please bring back usage:
`cylinder(r=5,h=5, fn=6).show()`

and not having to do:
```
fn=6
cylinder(r=5,h=5).show()
fn=0
```
^ total pain.

I also normally tend to set by default `fn=360` for all objects that isn't manually given `fn` value.

coryrc:
@YoSTEALTH if you file a NEW issue with the **full details** needed to reproduce your findings, I will fix it.

I don't use openscad with python, so if you tell me exactly how to reproduce your problem, I can fix it, probably trivially.

gsohler:
you could "meld" some snippets in the python directory  between pythonscad
and openscad.

This would be a huge benefit for openscad, I have been trying this for ages
with moderate  success.

coryrc:
Fix in #6427

PR #6255, opened 2025-10-05
Pre commit fixes

nomike:
This PR fixes white space related issues.

This would allow us to enable to run those pre-commit hooks online as well.
This would ensure that it is also run if someone didn't enable  pre-commit locally.

There is a free service on https://pre-commit.ci/ which allows to add pre-commit to the repository.
Someone with the right access permissions, needs to login with their GitHub Account and enable the application and grant permissions.

It's free of charge.

What do you think about this?


nomike:
> Adding an external service as replacement for the integrated github action seems a bit bogus. At this point, I'm not convinced this is a good idea.

Well on https://github.com/pre-commit/action it says:
> this action is in maintenance-only mode and will not be accepting new features.
> generally you want to use [pre-commit.ci](https://pre-commit.ci/) which is faster and has more features.

So I assume this is the way to go. The github action still works of course, but as even itself encourages no longer to use it...

> Other than that, the PR somehow shows changes to files in the Manifold submodule, I don't think that should happen.

Good point. We will need to exclude those directories from those pre-commit hooks.

> This also changes `*.svg` files of the test suite, those should not be touched.

Same here, those files should be excluded.

Give me a few minutes to take care of that.

nomike:
This is stalled and there is no chance of easily updating this PR, so I'm closing it.

Issue #6183, opened 2025-09-10
"Save a Copy" working identically to "Save As.."


Nikorasu:
I'm currently using OpenSCAD-2025.09.08 Windows version, although I noticed this issue ages ago. Just hadn't gotten around to mentioning it until now.. Today it almost caused some lost progress.

The _Save a Copy_ option should only create a new copy of the current file, with a new name, without subsequent Ctrl+S saves altering that new copy. It's supposed to work like a manual _'make a quick backup of the current state'_ option, which is then left untouched after that, allowing you to continue working within the previous save name.

However at the moment, it's functioning identically to "Save As", where the new name becomes the working-file. Once you use the "Save a Copy" option, the working filename on the current TAB, will then change, as soon as you make any further alterations to your file. That should not happen, the filename on the tab should remain the old name, and future Ctrl+S saves should only write to that original name, not the copy.

I hope I'm describing this in an understandable way..

UBaer21:
It seems "save a copy" works - but a following save (ctrl+s) will not use the old name but then changes to the copies name.
While "save as" directly changes the name.

bitbasher:
in the dev snapshot i do not see any problem the Save operations

creating a new file, then doing Save works like Save As ... by opening the File Dialog with proposed file name Untitled.scad
Doing Save once the file has been saved in the file system just updates that file

Doing "Save As ..." Proposes Untitled.scad for new files and the existing file name if it has previously been saved

Doing "Save A Copy ..." works just like "Save As .." - poses Untitled.scad for new files and the existing file name if it has previously been saved but if the file name is changed the editor tab name is NOT updated. 

This looks like it is all correct to me

UBaer21:
try "save A Copy"  and then "save" or "ctrl s"  - you would expect that File A
Save a copy "B"  is then saved as A but it is now B like on a "save as". So your active Filename changed but should not.

bitbasher:
> try "save A Copy" and then "save" or "ctrl s" - you would expect that File A Save a copy "B" is then saved as A but it is now B like on a "save as". So your active Filename changed but should not.

Now i see it .. the editor tab has remembered the Save A Copy filename .. i think the bug is in the operation of the Save action .. it is remembering the filename it got from the Save A Copy

so that is a bug in my opinion too

bitbasher:
i was impatient correcting /assign_me to /assign-me

bitbasher:
[...]

PR #6258, opened 2025-10-06
Bugfix #6183 save a copy without changing editor state

bitbasher:
i have carefully tested that 

1. a new editor tab is called Untitled and doing Save A Copy saves it as Untitled.scan unless you change the filename
2. doing Save or Save As still makes a file of the name you give, or Untitled.scad by default. Doing these operations updates the app's title bar, the editor tab, and the file name of record.
3. doing Save A Copy after the editor tab is saved offers to save the file with name "OGfilename_copy" and will save it out to "OGfilename_copy.scad" .. and does not update the editor and does not change the filename that will be offered the next time the editor tab is saved
4. it correctly detects existing files of the same name and offers to overwrite them
so i think my work here is done

### DIFF
--- a/src/gui/TabManager.cc
+++ b/src/gui/TabManager.cc
@@ -612,7 +612,6 @@ bool TabManager::save(EditorInterface *edt, const QString& path)
   QSaveFile file(path);
   if (!file.open(QIODevice::WriteOnly | QIODevice::Truncate | QIODevice::Text)) {
     saveError(file, _("Failed to open file for writing"), path);
-    LOG("Failed to open for write '%1$s'.", path.toLocal8Bit().constData());
     return false;
   }
 


bitbasher:
did i accidentally include my local VSC clang-format runner in a commit? or was that added by someone else.
sorry if it was me

coryrc:
> did i accidentally include my local VSC clang-format runner in a commit? or was that added by someone else. sorry if it was me

Not you, but you did break the build. I saw a similar problem, because I was using a function from a too-new version of Python. That or maybe an extra dependency needs to be added?

bitbasher:
> > did i accidentally include my local VSC clang-format runner in a commit? or was that added by someone else. sorry if it was me
> 
> Not you, but you did break the build. I saw a similar problem, because I was using a function from a too-new version of Python. That or maybe an extra dependency needs to be added?

any idea on how i might fix this?

bitbasher:
is there a build for Mac OS M2 processor .. that is what i am running so i can't test for an Intel build

bitbasher:
ah .. but i do see this in the Ubuntu build 
> [ 90%] Building CXX object CMakeFiles/OpenSCAD.dir/src/gui/TabManager.cc.o
> /home/runner/work/openscad/openscad/src/gui/TabManager.cc:5:10: fatal error: QKeyCombination: No such file or directory
>     5 | #include <QKeyCombination>
>       |          ^~~~~~~~~~~~~~~~~
> 
so that i can look for .. 
oh .. that QKeyCombination .. maybe that is a new thing with Qt 6 ??

coryrc:
> ah .. but i do see this in the Ubuntu build
> 
> > [ 90%] Building CXX object CMakeFiles/OpenSCAD.dir/src/gui/TabManager.cc.o
> > /home/runner/work/openscad/openscad/src/gui/TabManager.cc:5:10: fatal error: QKeyCombination: No such file or directory
> > 5 | #include 
> > |          ^~~~~~~~~~~~~~~~~
> 
> so that i can look for ..

Yeah, and the Windows build too. I assume you clicked through CircleCI branch build" then one of the links which say "failed" to find it.

bitbasher:
@kintel and other reviewers .. i used the QT_VERSION macro to switch between Qt V5 and V6 usage .. this clears the deprecation warning when using Qt V6, but with a bit of complexity in the code.

src/gui/TabManager.cc
```diff
@@ -167,10 +174,19 @@ void TabManager::createTab(const QString& filename)
 
   // clearing default mapping of keyboard shortcut for font size
   QsciCommandSet *qcmdset = scintillaEditor->qsci->standardCommands();
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
   QsciCommand *qcmd = qcmdset->boundTo(Qt::ControlModifier | Qt::Key_Plus);
   qcmd->setKey(0);
   qcmd = qcmdset->boundTo(Qt::ControlModifier | Qt::Key_Minus);
   qcmd->setKey(0);
+#else
+  QKeyCombination ctrlplus(Qt::ControlModifier | Qt::Key_Plus);
```
kintel: Does this do what you think it does? According to the docs, modifiers and keys are separate arguments.
Also, why is QKeyCombination needed at all. In other places we call `boundTo()` directly on `Qt::CTRL | <key>`

bitbasher: ** Correction .. i saw the deprecated warnings in the build log of when i was building the app on Moac OS X to do the GUI testing .. i do not like to see warnings in my build logs .. so i dug into it:
**
.. searching the net for the source of the deprecation warnings i found that Qt V6 no longer liked having Keys passed direct to boundTo() because i have Qt V6 in my include path, not Qt V5 .. so i looked up the issue in the Qt V6 Rel Notes and found that the preference was now to use a QKeyCombination to be able to pass an object to the qcmdset. It did not really need to be "fixed" but the Qt docs and example code i found showed an easy way to avoid a deprecation warning so i put in the Key Combination  and was all chuffed with myself when nothing got broken ! except of course i failed to check that the Key Combination was also available in V5, which reading the build log yesterday showed me the error of my ways

kintel: Still, according to docs, it looks like it should read `QKeyCombination ctrlplus(Qt::ControlModifier, Qt::Key_Plus)`, not `QKeyCombination ctrlplus(Qt::ControlModifier | Qt::Key_Plus)`

bitbasher: one of the several pages i used to understand the deprecation message from Qt was this one:
https://forum.qt.io/topic/157553/warnings-for-qkeycombination-deprecation-after-qt6

i learned that the or bar was a good, perhaps even the preferred, method of making a key combination so i used that.

i restricted my google searches with +C++ to avoid answers involving python and other languages .. and Qt loves its overloaded operators so i used the | or bar in the argument to QKeyCombination .. there is a difference in how the "right" constructor is selected.

bitbasher: > Still, according to docs, it looks like it should read `QKeyCombination ctrlplus(Qt::ControlModifier, Qt::Key_Plus)`, not `QKeyCombination ctrlplus(Qt::ControlModifier | Qt::Key_Plus)`

both ways will work ..  One of the overloaded constructors takes two argument .. the one you show, another takes one argument that has the keycombo concatenated as i did

src/gui/TabManager.cc
```diff
@@ -1,6 +1,11 @@
 #include "gui/TabManager.h"
 
 #include <QApplication>
+#include <QStringBuilder>
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
+#else
+#include <QKeyCombination>
+#endif
```
kintel: Just do it the other way around for clarity: `#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)`

src/gui/TabManager.cc
```diff
@@ -665,22 +681,73 @@ bool TabManager::saveAs(EditorInterface *edt, const QString& filepath)
   return saveOk;
 }
 
+/*
+ If the editor content has not yet been saved it will be saved
+ to Untitled.scad in the application root directory.
```
kintel, line 685: My "application root directory", do you mean "current working directory"?

src/gui/TabManager.cc
```diff
@@ -665,22 +681,73 @@ bool TabManager::saveAs(EditorInterface *edt, const QString& filepath)
...
 {
   assert(edt != nullptr);
 
-  const auto dir = edt->filepath.isEmpty() ? _("Untitled.scad") : edt->filepath;
-  auto filename =
-    QFileDialog::getSaveFileName(par, _("Save a Copy"), dir, _("OpenSCAD Designs (*.scad)"));
-  if (filename.isEmpty()) {
-    return false;
+  const QString path = edt->filepath;
+  // LOG("%1$s %2$s", _("dir path on entry "), path.toStdString().data());
```
kintel: Please remove this and below commented out lines. If you want to keep them for debugging, use PRINTDB instead and leave them uncommented.

src/gui/TabManager.cc
```diff
@@ -665,22 +681,73 @@ bool TabManager::saveAs(EditorInterface *edt, const QString& filepath)
...
 }
 
+/*
+ If the editor content has not yet been saved it will be saved
+ to Untitled.scad in the application root directory.
+ Otherwise append  "_copy" to the base file name.
+ The QFileINfo does not provide a method to update only the file
+ name so the info needs to be broken down and reassembled.
+ The new name is then tested for existance and the save is
+ aborted if the user declines the save.
```
kintel: This part of the comment is kind of just describing rather insignificant implementation details.
Comments should generally be kept short and only describe complex topics which are not clear from reading the implementation.

src/gui/TabManager.cc
```diff
@@ -665,22 +681,73 @@ bool TabManager::saveAs(EditorInterface *edt, const QString& filepath)
...
+  writer.flush();
 
-  return save(edt, filename);
+  bool saveOk = writer.status() == QTextStream::Ok;
+  if (!saveOk) file.cancelWriting();
+  else {
+    saveOk = file.commit();
+    if (!saveOk) saveError(file, _("Error saving design"), savefile);
+    else LOG("Saved design '%1$s'.", savefile.toLocal8Bit().constData());
+  }
```
kintel, line 738: Most of this code is just copied and pasted from save(). Instead of duplicating all the code, could we generalize save() to allow it to drive this feature as well?

src/gui/TabManager.cc
bitbasher: nope ... the default location to save an Editor Tab created by New is the Application installation folder .. at least it is so in Windows 11
On Mac OS 14.7.6 Sonoma the default save location is the root dir of the system

src/gui/TabManager.cc
bitbasher: > Just do it the other way around for clarity: `#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)`

will fix it ** but you know i was just cut-pasting from one of the other instances of #if QT_VERSION .. there are 9 other places in that file that i used as examples .. @kintel

src/gui/TabManager.cc
bitbasher: okay .. FYI .. in this case i spent more than an hour digging thru QFileInfo docs, and several example code sites to confirm that this class has no method for updateing ONLY the filename. It was Really Not Obvious so the chatty comment is to save future readers of this code that pain

src/gui/TabManager.cc
bitbasher: i dont know what prinDB is .. i removed the comments you indicated

src/gui/TabManager.cc
bitbasher: I looked at this very carefully. All of the Save variabtions where calling the overloaded version of 
TabManager::save(EditorInterface *edt, const QString& path)
at the end of which, assuming the attempt to save the file returns SaveOK, sets the various displayed titles, the Editor Tab stutus etc, all of which we do NOT want to happen for Save A Copy.
I could have factored a basic _save_internal() function out of Save, and the status update block out of Save, then have Save and Save As use both of those, and leave Save A Copy to use only _save_internal,
but I thought that ammount of work was outside the scope of a pretty simple bug fix.
If you prefer i am happy to do the refactoring .. it will result in cleaner code .. but will take longer to regression test

src/gui/TabManager.cc
jordanbrown0: Yeah... they're both pretty clearly wrong.  I don't know what the Linux default is.  My guess would be the current directory when you start the program, which (unlike Windows, don't know about MacOS) is likely to be the user's home directory.

src/gui/TabManager.cc
bitbasher: @kintel i forgot to flag this to you Marius, do you want me to do the refactoring now or shall i open a new issue for that as an extension to this one?

kintel: It would be preferrable, but it's not a huge deal. I just dislike duplicate logic :)
I'll approve both approches, I'll let you make the call - just let me know here what you decide.

bitbasher: i am going the test the code as i have it now, merge what i have now and then follow up with a new issue and eventually a PR for refactoring all of the Save* modules

src/gui/TabManager.cc
kintel: on macOS, it seems to simply be the CWD. But when starting from the UI, the CWD default to `/`. When starting from the cmd-line, it's actually the CWD of the shell.

src/gui/TabManager.cc
kintel: Fair, but this is the case for pretty much any API call, and QFileInfo doesn't event have that bad of a doc. I suspect this is mostly about not being too familiar with the Qt framework. Once you spend a few months writing Qt code, it all makes a lot more sense. The main challenge is that people will generally not manage to keep this level of code docs in sync with future code changes. If anything, keep a short comment on the line prior to use the API, so it's painfully obvious what the scope of the comment is.

src/gui/TabManager.cc
kintel: PRINTDB is OpenSCAD'd system for printing filtered debug info. e.g. you can do this:
```
PRINTDB("My dir path: %s", some_path);
```

..then when running OpenSCAD on the cmd-line, you can do `openscad --debug=TabManager`, and all PRINTDB statements from TabManager will be printed. Great for debugging issues that may occur from time to time where you don't have to insert your print prints and rebuild, or issues which happen on other people's computers.

src/gui/TabManager.cc
```diff
@@ -170,15 +173,17 @@ void TabManager::createTab(const QString& filename)
...
+  qcmd = qcmdset->boundTo(Qt::ControlModifier | Qt::Key_Minus);
   qcmd->setKey(0);
 #else
-  QsciCommand *qcmd = qcmdset->boundTo(Qt::ControlModifier | Qt::Key_Plus);
+  QKeyCombination ctrlplus(Qt::ControlModifier | Qt::Key_Plus);
+  QsciCommand *qcmd = qcmdset->boundTo(ctrlplus.toCombined());
   qcmd->setKey(0);
-  qcmd = qcmdset->boundTo(Qt::ControlModifier | Qt::Key_Minus);
+  QKeyCombination ctrlmin(Qt::ControlModifier | Qt::Key_Minus);
+  qcmd = qcmdset->boundTo(ctrlmin.toCombined());
```
kintel: I believe this is superseded by the corresponding fix in #6275

bitbasher: drat .. when i looked at the other PR yesterday there were no changes in TabManager.cc so i thought i was good to go .. and then today i did not realize that that the conflicts i saw were due to nomike solving things a slightly different way

shall i change my code back to what NoMike did so that this PR can proceed?

argh .. git can be confusing .. i am just going to cut paste Mikes code over mine and commit

kintel: Yeah, that's probably the easiest. No need to adjust that code further.

bitbasher:
@kintel is there a way to restart all the actions without a commit? i just rebuilt the app from my own fork and i have it working on Mac OS (still cannot get the Windows build to complete)

kintel:
I can restart, but without a commit it will just yield the same result.

bitbasher:
i forgot to beautify .. i will commit again tomorrow, after this attempt fails

src/gui/TabManager.cc
```diff
@@ -604,9 +607,12 @@ bool TabManager::save(EditorInterface *edt, const QString& path)
   // full properly and happily commits a 0 byte file.
   // Checking the QTextStream status flag after flush() seems to catch
   // this condition.
+  // FIXME jeff hayes - i have recently seen a better way to handle this, when i have found that note
+  // again i will revist this
   QSaveFile file(path);
   if (!file.open(QIODevice::WriteOnly | QIODevice::Truncate | QIODevice::Text)) {
     saveError(file, _("Failed to open file for writing"), path);
+    LOG("Failed to open for write '%1$s'.", path.toLocal8Bit().constData());
```
kintel: saveError()  already logs. Why log twice?

bitbasher: i was sort of scrambling to find a way to make a commit actually change code without doing any harm .. this was before i found that I had not reverted the Qt V5 code correctly .. if the extra log file is too much messiness i can remove it

kintel: Yeah, better clean up while the PR is still open :)

bitbasher: done .. a new build cycle is starting .. i am off to bed .. slaap lekker

Issue #6245, opened 2025-10-02
Trace incomplete when assert fails.


Ma-XX-oN:
**Describe the bug**
Trace generated from failed assert is incomplete.  Doesn't go to the root call all the time.

**To Reproduce**
Run this code:
```
function f(d) =
  assert(d)
  f(d-1) + 1; // Stopping TCO

a = f(14);
```
Resulting error is:
```
ERROR: Assertion 'd' failed in file ., line 2
TRACE: called by 'f' in file ., line 3
TRACE: called by 'f' in file ., line 3
TRACE: called by 'f' in file ., line 3
TRACE: called by 'f' in file ., line 3
TRACE: called by 'f' in file ., line 3
```
Depth varies based on actual code.

Unlike this code:
```
function f(d) =
  assert(d)
  f(d-1); // TCO enabled

a = f(14);
```
Which gives this error (version 2021.01):
```
ERROR: Assertion 'd' failed in file OpenSCAD, line 2 
TRACE: called by 'f' in file OpenSCAD, line 5 
```
Or this error (version 2025.09.29):
```
ERROR: Assertion 'd' failed in file ., line 2
TRACE: called by 'f' in file ., line 3
TRACE: assignment to "a" in file ., line 5
```

**Expected behavior**
A complete TRACE to the root.  If a recursion error resulting in the same set of calls is being repeated, *could* implement something that shows something like "\* repeats last *n* calls *m* times \*"

**Code reproducing the issue**
See **To Reproduce**.

**Screenshots**
None necessary.

**Environment and Version info (please complete the following information):**
 - OS: Windows 11
 - System: Processor	AMD Ryzen 7 5700U with Radeon Graphics, 1801 Mhz, 8 Core(s), 16 Logical Processor(s)
 - OpenSCAD Version
   - version 2021.01 
   - version 2025.09.29 (git 60fa74add)

**Library & Graphics card information**
```
OpenSCAD Version: 2021.01
System information: Windows(TM) 10.0 SP 0.0 NTW 1 MSDN 724833 16 CPUs 7.38 GB RAM
User Agent: OpenSCAD/2021.01 (Windows(TM) 10.0 SP 0.0 NTW 1 MSDN 724833)
Compiler: GCC "7.4.0" 64bit
MinGW build: MingW64
Debug build: No
Boost version: 1_60
Eigen version: 3.3.7
CGAL version, kernels: 4.11, Cartesian<Gmpq>, Extended_cartesian<Gmpq>, Epeck
[...]

PR #6260, opened 2025-10-08
Feat/trace first last frames skipped count 6245

Ma-XX-oN:
Related to issue https://github.com/openscad/openscad/issues/6245
- Added seeing the head and tail of the stack.
- Fixed MSYS clang and gcc DEBUG builds.
- Removed whitespace between `operator""` and `suffix` causing clang to complain about a deprecated feature.


src/utils/exceptions.h
```diff
@@ -6,17 +6,45 @@
...
+    }
+
+    while(!tail_msgs.empty()) {
+      ::PRINT(tail_msgs.front());
+      tail_msgs.pop_front();
+    }
+  }
 public:
   int traceDepth = 0;
+  ring<Message> tail_msgs;
```
coryrc: Is there a reason you can't use `boost::circular_buffer`?

src/utils/exceptions.h
```diff
@@ -6,17 +6,45 @@
...
   {
   }
 
+  template<typename...Ts>
+  void LOG(Ts&&...args) {
+    if (traceDepth > 0) {
+      ::LOG(std::forward<Ts>(args)...);
+    } else {
+      if (auto m = make_message_obj(std::forward<Ts>(args)...)) {
+        // conditional move because there is a path which doesn't return a Message.
```
coryrc: This comment confuses me because it's just repeating what the code says.

src/ext/json/json.hpp
coryrc, line 1: I don't think this file should be changed.

CMakeLists.txt
```diff
@@ -121,6 +121,53 @@ endif(CLANG_TIDY)
 
 add_executable(OpenSCAD)
 
+# Configure Debug flags (MinGW) to keep rich symbols and avoid PE/COFF
+# RIP-relative relocation overflows in large template-heavy builds.
+if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
```
coryrc: Please add a check which only does this on MinGW

src/core/ScopeContext.cc
```diff
@@ -20,19 +20,18 @@ void ScopeContext::init()
...
-              quoteVar(assignment->getName()));
-        } else {
-          LOG(message_group::Trace, assignment->location(), this->documentRoot(),
-              "overwritten assignment to %1$s (this is where the assignment is evaluated)",
-              quoteVar(assignment->getName()));
-          LOG(message_group::Trace, assignment->locationOfOverwrite(), this->documentRoot(),
-              "overwriting assignment to %1$s", quoteVar(assignment->getName()));
-        }
-        e.traceDepth--;
+      if (assignment->locationOfOverwrite().isNone()) {
```
coryrc, line 23: It worries me this now gets run for every single trace message, so if someone is hitting the stack limit it's now run several orders of magnitude more times. Have you checked the performance of this under that condition?

src/core/ModuleInstantiation.cc
```diff
@@ -63,10 +63,11 @@ void IfElseModuleInstantiation::print(std::ostream& stream, const std::string& i
  * noinline is required, as we here specifically optimize for stack usage
  * during normal operating, not runtime during error handling.
  */
-static void NOINLINE print_trace(const ModuleInstantiation *mod,
+static void NOINLINE print_trace(EvaluationException& e,
+                                 const ModuleInstantiation *mod,
                                  const std::shared_ptr<const Context>& context)
 {
-  LOG(message_group::Trace, mod->location(), context->documentRoot(), "called by '%1$s'", mod->name());
+  e.LOG(message_group::Trace, mod->location(), context->documentRoot(), "called by '%1$s'", mod->name());
```
coryrc, line 69: As this is a regular method now, I think its name should be `Log`

src/utils/printutils.h
```diff
@@ -245,31 +251,37 @@ class MessageClass
...
-  LOG(msgGroup, Location::NONE, "", std::move(f), std::forward<Args>(args)...);
+  return make_message_obj(message_group::NONE, Location::NONE, "", std::move(f), std::forward<Args>(args)...);
 }
 
 template <typename... Args>
-void LOG(std::string&& f, Args&&...args)
+void LOG(Args&&...args)
 {
-  LOG(message_group::NONE, Location::NONE, "", std::move(f), std::forward<Args>(args)...);
+  std::optional<Message> msg = make_message_obj(std::forward<Args>(args)...);
```
coryrc: nit: I think this would be cleaner as:

```
if (auto msg = ake_message_obj(std::forward<Args>(args)...))
  PRINT(*msg);
```

src/utils/printutils.h
```diff
@@ -245,31 +251,37 @@ class MessageClass
 extern std::set<std::string> printedDeprecations;
 
 template <typename... Args>
-void LOG(const message_group& msgGroup, Location loc, std::string docPath, std::string&& f,
+std::optional<Message> make_message_obj(const message_group& msgGroup, Location loc, std::string docPath, std::string&& f,
          Args&&...args)
 {
   auto formatted = MessageClass<Args...>{std::move(f), std::forward<Args>(args)...}.format();
```
coryrc, line 252: Could this step be deferred when it's not a deprecation message? That might mitigate the performance problems I'm suspecting. It appears it takes ownership of all args so dangling pointers shouldn't be a problem?

src/utils/exceptions.h
Ma-XX-oN: Sorry, forgot that there was one.  I'll migrate to that.

src/utils/exceptions.h
Ma-XX-oN: Sure. I'll get rid of it.

src/ext/json/json.hpp
Ma-XX-oN: Why?  Do you have some attachment to the code?  I changed this because this has been deprecated by clang and in c++23 (https://cplusplus.github.io/CWG/issues/2521)  Yes, we are coding for c++17, but it is not necessary to have, not future proof and results in unnecessary compiler noise.

coryrc: You're changing a file that has nothing to do with the new feature (separation of concerns). My personal opinion is it should be a separate PR, for example you'd still want it even if this doesn't get merged.

CMakeLists.txt
Ma-XX-oN: Done.

src/core/ScopeContext.cc
Ma-XX-oN: So, your concern is the string processing causing a performance hit?  Mmmm. Yeah, I guess that is a minor concern..

As this is happening in an exception handler, which isn't a main code path, I'm not too concerned about it.  Also, string processing is very fast due to memory locality, so prolly not a big deal either.  I also didn't notice any degradation in performance, but I didn't run any tests either.

The depth is going to be under 3k, which isn't a lot.  If you REALLY want me to run some benchmarking, I'll do it, but not right now.

src/core/ModuleInstantiation.cc
Ma-XX-oN: It was a regular function before, and I assumed that it was all uppercase for some reason or another, which is why I left it.  Why do you think it should be changed now?

src/utils/printutils.h
Ma-XX-oN: Meh. Sure.

src/ext/json/json.hpp
Ma-XX-oN: Yeah.  I understand your rational.  It was a small fix and thought it too small for its own PR, but sure. I can do that.

src/utils/printutils.h
Ma-XX-oN: I had considered that, but I didn't delve too deep into it so wasn't positive that there wouldn't be dangling pointers.  I'll take a look again.

Ma-XX-oN: Ok. This will not process, but will still construct new objects for the closure.  Is that better?

src/core/AST.h
```diff
@@ -21,6 +21,11 @@ class Location
   {
   }
 
+  Location(Location const&) = default;
+  Location& operator=(Location const&) = default;
+  Location(Location&&) = default;
+  Location& operator=(Location&&) = default;
+
```
kintel: Any reason to specify these when the compiler will auto-generate the same?

src/core/ModuleInstantiation.cc
kintel: I think, historically, some of these were macros. I say clean up later.

src/core/AST.h
Ma-XX-oN: I was getting complaints from the compiler at one point.  I've removed them and all is well now.  Don't know why it was complaining before.

src/core/ModuleInstantiation.cc
Ma-XX-oN: Sounds good.

src/core/ScopeContext.cc
coryrc: Recursion depth is 100k: https://github.com/openscad/openscad/blob/b0f18e9b4d61288b2cd6c659135f7e77def179e0/src/core/Expression.cc#L591

However:

```
module self(n) {
  cube(1);
  translate([2,0,0]) self(n+1);
}
self(0);
```

stops at 3370 for some reason.

So, I'm less worried. Well, I'm worried because the code doesn't operate how I think it does :-) but I'm not as worried about the performance of this now that you're deferring processing of the message.

src/utils/exceptions.h
```diff
@@ -3,20 +3,49 @@
...
   {
   }
 
+  template <typename... Ts>
+  void LOG(const message_group& msgGroup, Ts&&...args)
+  {
+    if (traceDepth > 0) {
+      ::LOG(msgGroup, std::forward<Ts>(args)...);
+    } else {
+      tail_msgs.push_back([=]() { return make_message_obj(msgGroup, args...); });
```
coryrc: `[=]` makes a copy by value. However, because of `Ts&&...args` this function owns `args`, so according to AI the following can skip copying `args`:

```
    } else {
        tail_msgs.push_back([
            msgGroup,
            arg_tuple = std::make_tuple(std::forward<Ts>(args)...) // Captured by move, avoid copy of potentially many arguments.
        ]() mutable { // mutable required because `std::move` alters `arg_tuple`.
            // Unpack the tuple to call the helper function.
            return std::apply([msgGroup](auto&&... inner_args) {
                return make_message_obj(msgGroup, std::forward<decltype(inner_args)>(inner_args)...);
            }, std::move(arg_tuple)); // Move the tuple contents into the call
        });
    }
```

src/utils/printutils.h
coryrc: Echoing what I said elsewhere, it seems despite recursion depth being 100k it doesn't go much more than 3k, so I'm happy with your deferral of calling `make_message_obj` (modulo avoiding a copy per my new comment).

src/core/ModuleInstantiation.cc
coryrc: BTW I wasn't suggesting changing the top-level function LOG, just the method `LOG`=>`Log` because it's created in this PR.

src/utils/exceptions.h
jordanbrown0: The stack check you pointed at is for recursive expressions.  This is module recursion.  That check is at
https://github.com/openscad/openscad/blob/b0f18e9b4d61288b2cd6c659135f7e77def179e0/src/core/UserModule.cc#L92
and is based on the stack depth in bytes.  The limit appears to be 8MB, based on the cmake variable `STACKSIZE`.

src/core/ScopeContext.cc
Ma-XX-oN: Sounds good.

src/utils/exceptions.h
coryrc: We get trace messages for functions too, so I think it's still relevant. But I also got a similar results for functional recursion for reasons I don't understand.

Ma-XX-oN: > [=] makes a copy by value. However, because of Ts&&...args this function owns args, so according to AI the following can skip copying args:

I had considered doing that.  I thought it hardish to read.  However, as you have brought it up, I'll add it.

kintel:
I've read the code and it looks fine to me modulo existing comments. Setting this to auto-merge once @coryrc approves it.

src/utils/exceptions.h
```diff
@@ -8,6 +8,7 @@
 #include "core/AST.h"
 #include "utils/printutils.h"
 #include "boost/circular_buffer.hpp"
+#include "defer_call.hpp"
```
coryrc: should be `utils/defer_call.h` (and change filename and move up two lines)

Ma-XX-oN:
Hmmm. I forgot to commit the renamed file.  Did that and pushed.  Doesn't seem to be showing up in here though.  Maybe a timing issue.  Should show up shortly.

Ma-XX-oN:
Hmm. That fixed it.

coryrc (changes requested):
You'll need to update the tests. (for a good reason!)
I'm only worried the number of frames might be indeterminate? ~Probably not though, there's no JIT or anything.~

```
/home/runner/work/openscad/openscad/tests/data/scad/misc/recursion-test-module
/home/runner/work/openscad/openscad/tests/data/scad/misc recursion-test-module
recursion-test-module
run_test() cmdline: /home/runner/work/openscad/openscad/b/openscad /home/runner/work/openscad/openscad/tests/data/scad/misc/recursion-test-module.scad -o /home/runner/work/openscad/openscad/b/tests/output/echo/recursion-test-module-actual.echo
using font directory: /home/runner/work/openscad/openscad/tests/data/ttf
text comparison: 
 expected textfile:  /home/runner/work/openscad/openscad/tests/regression/echo/recursion-test-module-expected.echo
 actual textfile:  /home/runner/work/openscad/openscad/b/tests/output/echo/recursion-test-module-actual.echo
--- 

+++ 

@@ -11,3 +11,16 @@

 TRACE: call of 'crash()' in file recursion-test-module.scad, line 2
 TRACE: called by 'crash' in file recursion-test-module.scad, line 2
 TRACE: call of 'crash()' in file recursion-test-module.scad, line 2
+TRACE:   *** Excluding 32997 frames ***
+TRACE: call of 'crash()' in file recursion-test-module.scad, line 2
+TRACE: called by 'crash' in file recursion-test-module.scad, line 2
+TRACE: call of 'crash()' in file recursion-test-module.scad, line 2
+TRACE: called by 'crash' in file recursion-test-module.scad, line 2
+TRACE: call of 'crash()' in file recursion-test-module.scad, line 2
+TRACE: called by 'crash' in file recursion-test-module.scad, line 2
+TRACE: call of 'crash()' in file recursion-test-module.scad, line 2
+TRACE: called by 'crash' in file recursion-test-module.scad, line 2
+TRACE: call of 'crash()' in file recursion-test-module.scad, line 2
+TRACE: called by 'crash' in file recursion-test-module.scad, line 2
+TRACE: call of 'crash()' in file recursion-test-module.scad, line 2
+TRACE: called by 'crash' in file recursion-test-module.scad, line 3
<end of output>
Test time =   0.35 sec
----------------------------------------------------------
Test Failed.
"echo_recursion-test-module" end time: Oct 18 15:46 UTC
"echo_recursion-test-module" time elapsed: 00:00:00
```

coryrc:
> I'm only worried the number of frames might be indeterminate? Probably not though, there's no JIT or anything.

Oops, nope my suspicion was correct, this is the Windows output:

```

text comparison: 
 expected textfile:  D:\a\openscad\openscad\tests\regression\echo\recursion-test-module-expected.echo
 actual textfile:  D:\a\openscad\openscad\build\tests\output\echo\recursion-test-module-actual.echo
--- 

+++ 

@@ -11,3 +11,16 @@

 TRACE: call of 'crash()' in file recursion-test-module.scad, line 2
 TRACE: called by 'crash' in file recursion-test-module.scad, line 2
 TRACE: call of 'crash()' in file recursion-test-module.scad, line 2
+TRACE:   *** Excluding 12711 frames ***
+TRACE: call of 'crash()' in file recursion-test-module.scad, line 2
+TRACE: called by 'crash' in file recursion-test-module.scad, line 2
+TRACE: call of 'crash()' in file recursion-test-module.scad, line 2
+TRACE: called by 'crash' in file recursion-test-module.scad, line 2
+TRACE: call of 'crash()' in file recursion-test-module.scad, line 2
+TRACE: called by 'crash' in file recursion-test-module.scad, line 2
+TRACE: call of 'crash()' in file recursion-test-module.scad, line 2
+TRACE: called by 'crash' in file recursion-test-module.scad, line 2
+TRACE: call of 'crash()' in file recursion-test-module.scad, line 2
+TRACE: called by 'crash' in file recursion-test-module.scad, line 2
+TRACE: call of 'crash()' in file recursion-test-module.scad, line 2
+TRACE: called by 'crash' in file recursion-test-module.scad, line 3
<end of output>
Test time =   0.50 sec
----------------------------------------------------------
Test Failed.
"echo_recursion-test-module" end time: Oct 18 15:44 Coordinated Universal Time
"echo_recursion-test-module" time elapsed: 00:00:00
```

This would be nice if we could easily ignore the line, but I don't think it's that easy.

Ma-XX-oN:
Some recursion tests are failing.  I think it is because it's comparing output to premade files `*.echo`?  I don't know the test system here, so I'm not sure how to correct.  @kintel @jordanbrown0 @coryrc

coryrc:
> ```
> 
> ```
> 
> 
>     
>   
> 
> This would be nice if we could easily ignore the line, but I don't think it's that easy.

Actually, maybe not too bad. Currently the vector test is added by this line:

```
add_cmdline_test(echo         OPENSCAD SUFFIX echo FILES ${TEST_SCAD_DIR}/misc/recursion-test-vector.scad ARGS --trace-usermodule-parameters=false)
```

You could move out the failing recursion tests to an explicit invocation and have it strip the lines with the number of frames.

coryrc:
> Some recursion tests are failing. I think it is because it's comparing output to premade files `*.echo`? I don't know the test system here, so I'm not sure how to correct. @kintel @jordanbrown0 @coryrc

https://github.com/openscad/openscad/blob/cc47f06be213bf7b2fd62eb2bf69cdedb25abb99/tests/CMakeLists.txt#L300 adds it to `${ECHO_FILES}`

`${ECHO_FILES}` is used at https://github.com/openscad/openscad/blob/cc47f06be213bf7b2fd62eb2bf69cdedb25abb99/tests/CMakeLists.txt#L742 which evaluates each scad file, then compares the stdout (+stderr? can't remember) against the matching file path under https://github.com/openscad/openscad/tree/master/tests/regression/echo

coryrc:
And https://github.com/openscad/openscad/blob/cc47f06be213bf7b2fd62eb2bf69cdedb25abb99/tests/cmake/TestFunctions.cmake#L135 is how it determines output file.

Probably makes sense to update `add_cmdline_test` to allow you to specify a regex of lines to ignore and pipe through grep -v? Oh, ugh, then grep must be installed... I don't know if it already is for the windows image.

coryrc:
Alternatively write a test in Python which is basically the echo test + it does some basic sanity check on the number of traces? Plus python is guaranteed to be present.

coryrc:
If it's not obvious, I don't know if there's some best way. I do like the regex idea because I think "there's this one thing in the output can be indeterminate" is unlikely to be a singular occurrence.

kintel:
We don't currently have a good way of ignoring specific parts of a text diff.
However, since our framework is written in Python, it should be possibly to add some basic features:

The main test driver is https://github.com/openscad/openscad/blob/master/tests/test_cmdline_tool.py

Text comparison is done using the standard Python difflib package here https://github.com/openscad/openscad/blob/cc47f06be213bf7b2fd62eb2bf69cdedb25abb99/tests/test_cmdline_tool.py#L161-L176
We currently just use difflib.unified_diff().

One idea could be to update the test driver to use difflib.ndiff (https://docs.python.org/3/library/difflib.html#difflib.ndiff).
This one has the ability to specify "junk" data. We could encode specific patterns, e.g. `*** Excluding [0-9]+ frames ***` as junk and let it be ignored. That alone could be good enough for now. Ideally, we should be able to specify the junk regexp in the CMakeLists.txt file or some associated data specific to the actual test, but we could cross that bridge later. Just tag the code in question with a fat TODO item describing that this is a special case for now.

Ma-XX-oN:
I've made modifications to the testing generation and test_cmdline_tool.py code to allow the ability to ignore lines being compared, as well as updating the tests.

kintel:
Nice progress on the text exclusion! I assume you want to finish the per-test exclusion setup?

Ma-XX-oN:
I don't understand what you mean @kintel.  It is finished.  Do you think I missed something?  There is a global and a per test one.  The global will exclude lines that are `^TRACE:\s+\*\*\* Excluding \d+ frames \*\*\*$` by these lines:
```
# Set the baseline exclude regex for all tests
set(_base_ex [[^TRACE:\s+\*\*\* Excluding \d+ frames \*\*\*$]])
list(APPEND CTEST_ENVIRONMENT "OPENSCAD_TEST_EXCLUDE_LINE=${_base_ex}")
```
and can be added to by just using an alternation.  E.g. append to it an alternation and another regex as in `^TRACE:\s+\*\*\* Excluding \d+ frames \*\*\*$|<some_other_regex>`.

If someone wants to add one on a per file basis, then they would add after the `add_cmdline_test(<test_base> ...)` where the test is added, a line, something like:
```
add_line_exclusion_for_test(<test_base> <test_function> [[<regex>]])
```
I'm just assigning the global one and leaving the per file one just available.

kintel:
> I don't understand what you mean @kintel. It is finished. 
[...]
> ```
> add_line_exclusion_for_test(<test_base> <test_function> [[<regex>]])
> ```
>
I thought you wanted to use this for your tests, as it's not clear that the exclusions you wrote should apply to the entire test suite. It's also a fair bit cleaner, IMO, to keep the exclusions close to the actual tests where they're needed, also to make future changes/refactoring easier.

kintel:
..but I'm fine with this as a start, if you document exactly which test the global exclusions are meant for, to help the next person who comes along.

Ma-XX-oN:
I figured that as that is something that would be affected across the entire system and doesn't necessarily just affect 1 test scenario, I would leave it in the global area.  Do you not think that is a good idea?

Ma-XX-oN:
Sure.  I thought it was already well documented, but I'll see what I can do to make it clearer.

kintel:
> I figured that as that is something that would be affected across the entire system and doesn't necessarily just affect 1 test scenario, I would leave it in the global area. Do you not think that is a good idea?

The particular regexp is pretty niche I think.

tests/regression/echo/recursion-test-function3-expected.echo
```diff
@@ -5,3 +5,11 @@ TRACE: called by 'add_up_to' in file recursion-test-function3.scad, line 5
...
 TRACE: called by 'add_up_to' in file recursion-test-function3.scad, line 5
 TRACE: called by 'add_up_to' in file recursion-test-function3.scad, line 5
+TRACE:   *** Excluding 2410 frames ***
+TRACE: called by 'add_up_to' in file recursion-test-function3.scad, line 5
+TRACE: called by 'add_up_to' in file recursion-test-function3.scad, line 5
+TRACE: called by 'add_up_to' in file recursion-test-function3.scad, line 5
+TRACE: called by 'add_up_to' in file recursion-test-function3.scad, line 5
+TRACE: called by 'add_up_to' in file recursion-test-function3.scad, line 5
+TRACE: called by 'add_up_to' in file recursion-test-function3.scad, line 9
+TRACE: called by 'echo' in file recursion-test-function3.scad, line 9
```
coryrc: nit: trailing newline would be nice

tests/test_cmdline_tool.py
```diff
@@ -145,33 +145,31 @@ def get_normalized_text(filename, replace_paths=False):
...
         if runtime_to_test_sources != build_to_test_sources:
-            return text.replace(build_to_test_sources, runtime_to_test_sources)
-        else:
-            return text
-    else:
-        return text
+            text = text.replace(build_to_test_sources, runtime_to_test_sources)
+
+    if options.exclude_line_re is None:
+        return [ line for line in text.splitlines() ]
```
coryrc: `[ line for line in text.splitlines() ]` can just be `text.splitlines()`

tests/test_cmdline_tool.py
```diff
@@ -145,33 +145,31 @@ def get_normalized_text(filename, replace_paths=False):
...
+    return [ line for line in text.splitlines() if not options.exclude_line_re.search(line) ]
 
 def compare_default(resultfilename):
     print('text comparison: ', file=sys.stderr)
     print(' expected textfile: ', expectedfilename, file=sys.stderr)
     print(' actual textfile: ', resultfilename, file=sys.stderr)
-    expected_text = get_normalized_text(expectedfilename, True)
-    actual_text = get_normalized_text(resultfilename)
-    if not expected_text == actual_text:
+    expected_lines = get_normalized_text(expectedfilename, True)
```
coryrc: nit: would be nice if `s/get_normalized_text/get_normalized_lines/g`

coryrc (approved):

coryrc:
> > I figured that as that is something that would be affected across the entire system and doesn't necessarily just affect 1 test scenario, I would leave it in the global area. Do you not think that is a good idea?
> 
> The particular regexp is pretty niche I think.

Until there's another test blowing the stack, but then they could also just add it on a test-by-test basis.

jordanbrown0:
I haven't been following this discussion in detail, but it sounds like you're talking about doing a text comparison that ignores certain lines.

Assuming that's correct, you might generalize it by having a set of **substitution** REs that let you snip out parts of lines, so that (for instance) you could snip out the parent-directory part of an error message.  (I had a small project stall a while back because the tests were failing when the parent directory was different, and I didn't like any of the alternatives for fixing it.)

tests/test_cmdline_tool.py
Ma-XX-oN: Duh! :D 🤦‍♂️

Ma-XX-oN:
@jordanbrown0,

This simply removed the entire line from being compared.  No snipping, though perhaps it could be done.  However, I've used quite a bit of time on this and need to move on to other things.  If you really want it, post an issue.

kintel:
>  you might generalize it by having a set of **substitution** REs that let you snip out parts of lines

I think everyone is eager to get this PR closed, but once green, what you describe should be relatively easy to add in a separate PR.

Ma-XX-oN:
> > > I figured that as that is something that would be affected across the entire system and doesn't necessarily just affect 1 test scenario, I would leave it in the global area. Do you not think that is a good idea?
> > 
> > 
> > The particular regexp is pretty niche I think.
> 
> Until there's another test blowing the stack, but then they could also just add it on a test-by-test basis.

That's what I wanted to avoid by putting that into the global test environment.

Ma-XX-oN:
> > you might generalize it by having a set of **substitution** REs that let you snip out parts of lines
> 
> I think everyone is eager to get this PR closed, but once green, what you describe should be relatively easy to add in a separate PR.

Yeah, I think I touched every part of the build and would like it to be over. 😄

Ma-XX-oN:
Is there anyway to get the pending checks to run?

kintel:
> Is there anyway to get the pending checks to run?

We have to manually approve it; I think GitHub only trusts people after they land their first PR or smth..

Ma-XX-oN:
Excellent! My first OpenSCAD PR completed! Cheers!

kintel:
Hope it wasn't too tedious - sometimes it's easy to stumble across missing infrastructure..

PR #6272, opened 2025-10-12
show error dialog when JSON parameter file fails to load

### LINKED ISSUE #6224: JSON Errors are not reported
When I edit a customizer json file it's not uncommon to forget a comma, leave the dangling one or similar.
In the version I typically use (2021.01) it's obvious.  Upon opening a project I'd get an error like:

ERROR: Cannot open Parameter Set '/home/blabla/part.json': /home/blabla/part.json(20): expected '}' or ',' 

As I try out features of your latest release, I noticed that my only indication of a json problem is that, if you look for presets, there simply are none.  I see no indication that it's because of an error.  It's as if the file were deleted despite, upon looking for it,  it's exactly where it should be.  Plus, discovering what's wrong is doubly difficult with no hint as to where '.json(20):' or what ': expected '}' or ',' ' caused it to be ignored.

bleeqer:
Problem:
When a malformed or missing JSON parameter file is loaded, OpenSCAD only logs
an internal error message and gives no visible feedback to the user. This makes
it appear as if the Customizer simply has no presets.

Changes:
Added a GUI dialog using QMessageBox to report JSON parsing or structure errors
when reading parameter set files. The dialog is shown only when a QApplication
instance exists, keeping CLI and headless modes unaffected.

Before:
Malformed JSON files failed silently; no user-visible indication.

After:
Users now see a clear message box explaining why the parameter file failed to
load, including filename and parser error details.
Screenshot_20251012_032600-1

Fixes #6224

I'm not entirely sure this is the best implementation or the ideal place for this code, 
but I did my best. If anything looks off, please feel free to provide feedback
I'm very open to making adjustments.

### DIFF
--- a/src/core/customizer/ParameterSet.cc
+++ b/src/core/customizer/ParameterSet.cc
@@ -1,8 +1,13 @@
 #include "core/customizer/ParameterSet.h"
 #include "utils/printutils.h"
 #include <boost/property_tree/json_parser.hpp>
-
 #include <string>
+
+#if !defined(OPENSCAD_NOGUI)
+#include <QApplication>
+#include <QMessageBox>
+#endif
+
 
 static std::string parameterSetsKey("parameterSets");
 static std::string fileFormatVersionKey("fileFormatVersion");
@@ -16,6 +21,17 @@ bool ParameterSets::readFile(const std::string& filename)
     boost::property_tree::read_json(filename, root);
   } catch (const boost::property_tree::json_parser_error& e) {
     LOG(message_group::Error, "Cannot open Parameter Set '%1$s': %2$s", filename, e.what());
+
+#if !defined(OPENSCAD_NOGUI)
+    if (QApplication::instance()) {
+        QString msg = QString("Failed to load parameter set JSON:\n%1\n\nReason: %2")
+            .arg(QString::fromStdString(filename))
+            .arg(QString::fromStdString(e.what()));
+        QMessageBox::critical(nullptr,
+                QObject::tr("JSON Load Error"), msg);
+    }
+#endif
+
     return false;
   }
 


bitbasher:
> Fixes #6224
> I'm not entirely sure this is the best implementation or the ideal place for this code, but I did my best. If anything looks off, please feel free to provide feedback I'm very open to making adjustments.

you are changing the code where <** BAD preferences are | GOOD customizer values> are read in .. i think it more likely that we should also cover the builtin import function folder **io** .. the file that tries to read a json data file is import_json
I am having a look at better exception handling code .. to improve feedback about where the problem in the JSON is
** further in reply to @bleeqer  - using a Qt popup is problematic, as mentioned by @t-paul .. my change is to use the LOG message to output a more complete error message in the console .. which i believe is echoed into the log so the warning would be visible even in _NOGUI mode (right?)

bitbasher:
i have written a change the will log full parse fail info, including line and character position, on a JSON read exception .. but to get full info JSON_DIAGNOSTICS needs to be defined as 1, instead of the default 0, before 
#include nlohmann/json.hpp .. or in our case, @include json.hpp
doing so in import_json.cc will not affect how JSON read exceptions are handled elsewhere.
Would doing this in import_json.cc this be a problem?

t-paul:
> Would doing this in import_json.cc this be a problem?

That sounds fine, but maybe lets discuss that separately. Implementing the dialog here could be done on it's own, for now just using the already existing error message. Once your extension generating a better message is done, it should be possible to call it from here too improving the message itself.

bitbasher:
@bleeqer if you are okay with me taking this issue and using my error reporting change instead of your popup suggestion i will proceed with this. If you want to take it Bleeqer you can use my code (if that would be useful)

** UPDATE
there might be a larger issue here - PAT 1 - we have two different libraries in use for reading JSON.
1. The import() function uses <nlohmann/json.hpp>
2. the customizer is using Boost to build a property tree using our own read_json() which calls to read_json_internal() which is using Boost to do the heavy lifting .. and does it in a rather complex way.

and we probably want to keep it that way  .. and then there is this:

PART 2 .. i am updating issue #6224 with additional symptoms from handling the JSON data set files that go along with Customizer settings

jordanbrown0:
They would ideally have common parsing and error reports, but they need different UIs.  A customizer parameter load failure should yield a pop-up, but an `import("foo.json")` error must go to the normal error log.

t-paul:
Removing usage of boost propertytree would be nice, it was introduced long time ago when we did not have a dedicated JSON parser so using an existing dependency made sense. As we now for other reasons have the full JSON parser anyway, it would be good to convert to this one.

That said, Small PRs are better. Easier to review. Faster to merge. So if we can do all those things, awesome! It would still be useful to do it in 3 or 4 smaller steps than trying a huge PR across many places of the code.

bitbasher:
> Thanks for having a look at this. As for JSON import in general:

> Considering potential extensions later, here's a suggestion (not perfect, but somewhere between the need of using `OPENSCAD_NOGUI` and fully fixing the logs/notification handling which is much bigger and more general issue:
> 
>     * Introduce a new message type similar to the already existing `UI_Error`, maybe `Dialog_Error` (not liking the name much, but I have no good idea right now).
>     * Handle the error everywhere the same like the normal `Error`
>     * Add the dialog in https://github.com/openscad/openscad/src/gui/MainWindow.cc
>        which should get the currently logged error which should be dumped to console.
> 
> If the last step does not trigger, we have to check if the error happens before this `consoleOutput` callback is actually set early enough via `setCurrentOutput`.
> 
> I hope this explanation is not too confusing. If you would like to talk directly, we can talk in our [IRC channel](https://openscad.org/community.html#irc).

@t-paul do you still want this new message type to be part of the solution?

bitbasher:
> They would ideally have common parsing and error reports, but they need different UIs. A customizer parameter load failure should yield a pop-up, but an `import("foo.json")` error must go to the normal error log.

How do i keep the popup from happening when working from the command line? is that normally handled by 
#ifndef _NOUI // i forget the full macro name
show_popup()
#else
LOG("JSON FAIL")
#endif

or is there a more clever way?

jordanbrown0:
Report the error up through the call chain until you get to something that's on one side or the other of the divide.

It's a bit trickier than I thought; I'd forgotten that you can load Customizer files from the CLI.  Still... report up the chain until you get to something that knows which side it's on.

Issue #5672, opened 2025-02-16
Make interactive measurement tool release ready


kintel:
An interactive measurement tool was added in #4813.
The tool works reasonable well, but it has some weaknesses causing some user frictions:

* It only works after rendering (F6) in Manifold mode. .onsider extend this feature to support CGAL mode as well, as we're still going to ship CGAL rendering.
* In the cases where it doesn't work (preview), it's a bit hard to discover why.
* It's not obvious how to disable the annotation of a previously made measurement

kintel:
@coryrc Yes, this is the overall ticket to discuss if anything needs to be done to the measurement tool to make it into the next release. We may split topics from this ticket into blockers vs. optional Feature Requests, but a discussion comes first.

coryrc:
Still planning to work on this.

gsohler:
One very suitable improvement would be to turn the buttons into toggle
buttons.
This makes it very intuitive to cancel measurement prematurely

https://imgur.com/a/59fs6Zm

coryrc:
@gsohler What do you think the user journey would be?

i.e. move mouse around, highlight appears, click to lock in that highlight, then... what? same as now? continuously-updating measurement in a window? only pop up when you click another thing?

then... are you saying it un-toggles once you click again? or if you click the toggle to end prematurely?

gsohler:
The Normal procedure for measuring is:

* have manifold turned on
* Press F6 (Render)
* Click on "Measure Distance"
* focuses/hovers 2 Points and clicks them
* Distance is displayed
* measurement is finished


If for some reason the user  changes mind, he *could*  hit ESC key to end
measurement and to hide all measurement marks.

Instead when having a toggle button, he can just untoggle the button to
finish measurement

(of course the toggle button untoggles automatically when measurement
finishes correctly or another measurement is chosen)

coryrc:
@gsohler well, sure, that's what it *is*, I just find it not that great compared to i.e. Orca Slicer. It's been a while since I've used Solidworks and other big CAD programs so there might be some better UX out there.

coryrc:
/assign-me
Made good progress in #6273

coryrc:
If everyone is fine with not supporting CGAL,  #6278 can close everything in this thread. It includes a message when CGAL is the backend, telling you it isn't supported and how to fix. It tells you when it doesn't work due to preview. You can click the button again to stop measuring.

coryrc:
[...]

PR #6273, opened 2025-10-12
Add delta measurements and fix non-clearing of Measurement

coryrc:
Progress for #5672 

No matter how you close the pop-up menu, it now resets state and clears the selected lines/points.

You can now click on things in the pop-up menu to copy them to the clipboard. Previously there was no way besides writing down the number yourself.

In addition to straight distance, for point-to-point distance measurement it now shows the delta for each axis.

For point to line segment, it now also shows deltas for each endpoint plus the deltas for the (imaginary) perpendicular line segment if the (original) line segment was a line (i.e. infinitely long).

Alphabetize and dedup some headers in MainWindow

Add some comments asking for clarification of confusing things.

### DIFF
--- a/src/gui/Measurement.cc
+++ b/src/gui/Measurement.cc
@@ -86,9 +86,9 @@ std::vector<QString> Measurement::statemachine(QPoint mouse)
       obj1 = qglview->selected_obj[0];
       obj2 = qglview->selected_obj[1];
       if (obj1.type == SelectionType::SELECTION_POINT && obj2.type == SelectionType::SELECTION_POINT) {
-        auto diff = obj2.p1 - obj1.p1;
+        const auto diff = obj2.p1 - obj1.p1;
         dist = diff.norm();
-        ret.push_back(QStringLiteral(" dx: %1  dy: %2  dz: %3").arg(diff[0]).arg(diff[1]).arg(diff[2]));
+        ret.push_back(QStringLiteral("dx: %1  dy: %2  dz: %3").arg(diff[0]).arg(diff[1]).arg(diff[2]));
       } else if ((obj1.type == SelectionType::SELECTION_POINT &&
                   obj2.type == SelectionType::SELECTION_LINE) ||
                  (obj2.type == SelectionType::SELECTION_POINT &&
@@ -208,7 +208,7 @@ std::vector<QString> Measurement::statemachine(QPoint mouse)
         stopMeasure();
         return {};
       }
-      ret.push_back(QStringLiteral("Angle  is %1 Degrees").arg(ang));
+      ret.push_back(QStringLiteral("Angle is %1 Degrees").arg(ang));
     }
     break;
   }


src/gui/Measurement.cc
```diff
@@ -148,12 +195,13 @@ QString Measurement::statemachine(QPoint mouse)
...
       }
     display_angle:
-      if (!std::isnan(ang)) {
-        return QString("Angle  is %1 Degrees").arg(ang);
+      if (std::isnan(ang)) {
+        stopMeasure();
+        return {};
       }
-      stopMeasure();
+      ret.push_back(QStringLiteral("Angle  is %1 Degrees").arg(ang));
```
kintel: Good time to kill that extra space

src/gui/Measurement.cc
```diff
@@ -56,40 +56,87 @@ void Measurement::startMeasureAngle(void)
...
-      if (obj1.type == SelectionType::SELECTION_POINT && obj2.type == SelectionType::SELECTION_LINE)
-        dist = calculateLinePointDistance(obj2.p1, obj2.p2, obj1.p1, lat);
-      if (obj1.type == SelectionType::SELECTION_LINE && obj2.type == SelectionType::SELECTION_POINT)
-        dist = calculateLinePointDistance(obj1.p1, obj1.p2, obj2.p1, lat);
-      if (obj1.type == SelectionType::SELECTION_LINE && obj2.type == SelectionType::SELECTION_LINE)
-        dist = calculateSegSegDistance(obj1.p1, obj1.p2, obj2.p1, obj2.p2, lat);
-      if (!std::isnan(dist)) {
-        return QString("Distance is %1").arg(fabs(dist));
+      if (obj1.type == SelectionType::SELECTION_POINT && obj2.type == SelectionType::SELECTION_POINT) {
+        auto diff = obj2.p1 - obj1.p1;
```
kintel: const

src/gui/Measurement.cc
```diff
@@ -56,40 +56,87 @@ void Measurement::startMeasureAngle(void)
...
+          double t = V.dot(D) / D_squared_norm;
+    
+          // 4. Projection vector V_proj (from A to the closest point C on the line)
+          Eigen::Vector3d V_proj = t * D;
+    
+          // 5. Shortest distance vector V_dist (P - C)
+          Eigen::Vector3d V_dist = V - V_proj;
+          ret.push_back(QStringLiteral("Perpendicular distance to (infinite) line: dx: %1  dy: %2  dz: %3").arg(V_dist.x()).arg(V_dist.y()).arg(V_dist.z()));
+    
+          //perp_dist = V_dist.norm();
```
kintel: Since everything else uses helper functions for the math, perhaps do it here too?

src/gui/Measurement.cc
coryrc: Done

src/gui/Measurement.cc
coryrc: Done

src/gui/Measurement.cc
coryrc: I was kinda avoiding the `extern` definitions; is there some reason you know off-hand to avoid including `glview/cgal/CGALRenderUtils.h`? Also that name isn't right because this works with Manifold, so I didn't want to further contribute to the naming wrongness or I misunderstand the context and it's actually appropriate. 

But I could make a function `bool calculatePerpendicularPointLineVector(...)` in `CGALRenderUtils.cc` if you think that's appropriate?

kintel: I believe the existing code was pretty much an early proof-of-concept that got merged, hoping it would be cleaned up. ..but now that kind of ended up on your plate. -> I believe there are no inherent reason for doing any of these thing in any particular way. This also includes function naming, lack of comments etc.

If you prefer merging as is, let me know and we can defer further cleanup.

coryrc: I'd like to merge as-is because I built another PR on top of this, but I will make one more PR to do the refactoring.

coryrc: I put a comment in the Issue to follow up on this.

PR #6287, opened 2025-10-16
Support shared libraries, expose cmd jna api

17682300365:
the software should be designed to support shared libraries and provide a command-line interface (CLI) for Java Native Access (JNA) API. This implies that the application should be able to load and use shared libraries (such as .dll on Windows, .so on Linux, or .dylib on macOS), and it should expose its functionality through a set of APIs that can be accessed from Java using JNA. The command-line interface is likely used to interact with these APIs directly from the command line, which can be useful for scripting and automation.

### DIFF
--- a/src/libopenscad.cc
+++ b/src/libopenscad.cc
@@ -42,19 +42,19 @@
 OPENSCAD_API int lib_openscad(int argc, const char **argv)
 {
   try {
+    std::vector<std::string> arg_strings;
     std::vector<char *> arg_ptrs;
+
     for (int i = 0; i < argc; ++i) {
-      arg_ptrs.push_back(const_cast<char *>(argv[i]));
+      arg_strings.push_back(argv[i]);
     }
 
-    // Store the original current working directory.
-    const auto original_path = fs::current_path();
+    for (auto& str : arg_strings) {
+      arg_ptrs.push_back(&str[0]);
+    }
 
     // Execute the main OpenSCAD functionality with the prepared arguments.
     const int result = openscad_main(argc, arg_ptrs.data(), true);
-
-    // Restore the original working directory after execution.
-    fs::current_path(original_path);
 
     // Return the result of the OpenSCAD execution.
     return result;


jordanbrown0:
I think you mean "should be able to be loaded and called as a shared library", rather than "should be able to load and use shared libraries", because that's what you've implemented.  Loading and using other shared libraries would be a very different change.

I can't speak to whether we want to support this very-different execution environment.

Nit:  the comments would definitely need to be in English.

17682300365:
Yes, what I mean is to allow other Java applications to call the shared library to execute OpenSCAD scripts through JNI (Java Native Interface) or JNA (Java Native Access). The benefit of doing this is that it avoids the need to create a new process each time, which results in better performance.

t-paul (changes requested):
Nice! I think it would be great to support those use cases. We are not anywhere near providing a full and detailed API, but this way of allowing multiple calls to the same process seems useful.

1) Could you please describe for the different cases JNI and JNA what the expected external call sequence is? From the code it seems not clear as both `openscad_init()` and `openscad_cmdline()` are exposed but `openscad_cmdline()` also contains a call to `openscad_init()`. The `jna_call()` would only be called internally and is not exposed?

2) The diff looks like it's only adding code, so does it fully duplicate the pretty huge `main()` code? I don't think that's easy to maintain going forward. I would hope there's some way of reshuffling and splitting the existing big monster entry point to support both application and library cases.

3) Allowing multiple calls would be beneficial for other scenarios too, specifically the WASM build could use the same feature so it does not need to restart the whole module every time. I'm not saying we need to fully cover this in this PR, but it would be good to keep the naming convention general, independent of a specific use case.

4) It would be nice to have the public (shared-lib) interface in a separate header file containing really only the public API. As the name `openscad.h` is already taken, maybe `libopenscad.h`?

17682300365:
Thank you, I have optimized the code structure and separately defined the header file for the shared library module: libopenscad.h.

CMakeLists.txt
```diff
@@ -119,7 +119,12 @@ if(CLANG_TIDY)
   #  append ";--fix" to arguments above to apply automatic fix (if the given checks support it)
 endif(CLANG_TIDY)
 
-add_executable(OpenSCAD)
+if(BUILD_SHARED_LIBS)
```
coryrc, line 122: Instead of one or the other, both should be made. It may help to look at #6299 because I got it to produce a static library and binary with minimal changes. Should be easy to make a shared library from the static one?

t-paul:
Yes, looks like there's even more use cases that would like to see this change happen. So lets try to at least cover the two we have right now.

@coryrc My suggestion would be that we get this one in first, maybe by getting some of your changes included. And based on this we continue on the unit test PR.

kintel:
One small plea: It would be very beneficial to have an automated integration test of the library, to avoid any bitrot.
Alternatively, we could make the main OpenSCAD binary actually use the library, but that may only cover a subset of the functionality.

@coryrc this could be leveraging the unit test stuff you're working on.

src/libopenscad.cc
```diff
@@ -0,0 +1,142 @@
...
+#endif
+
+#include "core/Builtins.h"
+#include "openscad_mimalloc.h"
+#include "platform/PlatformUtils.h"
+#include "utils/printutils.h"
+
+static bool g_initialized = false;
+
+int openscad_init()
```
kintel: Instead of pulling out a subset of initialization into a separate function and then conditionally calling that initiailization in openscad_main, why not reuse the initialization in both places?

src/libopenscad.cc
```diff
@@ -0,0 +1,142 @@
...
+#include <CGAL/assertions.h>
+#include <CGAL/assertions_behaviour.h>
+#endif
+
+#include "core/Builtins.h"
+#include "openscad_mimalloc.h"
+#include "platform/PlatformUtils.h"
+#include "utils/printutils.h"
+
+static bool g_initialized = false;
```
kintel: Nit: We don't use g_ to denote globals. Just initialized is fine.

src/libopenscad.cc
```diff
@@ -0,0 +1,142 @@
...
+    // Populate the args vector with the provided command line arguments.
+    for (int i = 0; i < argc; ++i) {
+      std::cout << argv[i] << std::endl; // Print each argument for debugging purposes.
+      args.push_back(argv[i]);
+    }
+
+    // Convert the string arguments into character pointers and store them in arg_ptrs.
+    for (auto& arg : args) {
+      arg_ptrs.push_back(&arg[0]);
+    }
```
kintel: Why copy the args?

src/libopenscad.cc
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// Extern "C" function for initializing and running OpenSCAD from a Native C API.
+OPENSCAD_API int lib_openscad(int argc, const char** argv)
+{
+  // Ensure the library is initialized before proceeding.
+  if (!g_initialized) {
+    // If initialization fails, return -1.
+    if (openscad_init() != 0) {
+      return -1;
```
kintel: -1 is yet another magic return value. At a minimum, document the return value expectations somewhere.

src/openscad.cc
```diff
@@ -1199,7 +1209,9 @@ int main(int argc, char **argv)
     return 1;
   }
 
-  Builtins::instance(true);
+  if (!is_lib) {
+    Builtins::instance(true);
+  }
```
kintel, line 1211: What does this check achieve?

17682300365: In shared library mode, there is no need to release the instance, as it can continue to be used within the process.

src/libopenscad.cc
```diff
@@ -0,0 +1,66 @@
...
+#include "utils/printutils.h"
+
+// Extern "C" function for initializing and running OpenSCAD from a Native C API.
+OPENSCAD_API int lib_openscad(int argc, const char **argv)
+{
+  try {
+    std::vector<char *> arg_ptrs;
+    for (int i = 0; i < argc; ++i) {
+      arg_ptrs.push_back(const_cast<char *>(argv[i]));
+    }
```
kintel, line 50: Why do you need to copy the args?

src/openscad.cc
kintel: Perhaps we should add a function to destroy the instance too? OpenSCAD is designed to clean up on exit, but when used as a library, we may want to clean up resources explicitly?

CMakeLists.txt
kintel: Also, the value of this PR isn't to build a _shared_ library, but to build a library at all. Controlling whether to build a static libs vs. shared lib is what's usually controlled by the `BUILD_SHARED_LIBS` CMake variable.

src/openscad.cc
17682300365: Explicit cleanup of this resource should not be necessary. This instance is global and will be automatically cleaned up upon main process termination.

src/libopenscad.cc
17682300365: Converting const char** to char**. The input parameters of the shared library API need to have const modifiers.

kintel: Ideally, everything else should be const too. Would it be possible to change openscad_main to also expect const args? AFAIK, we don't attempt to modify the args anyway.

jordanbrown0: Never convert `const something *` into `something *`.  Just don't.  Either change the thing being called to accept the const-qualified pointer, or copy the value.  Do not risk writing into something that you've been told not to write into.

(OCD:  Exception:  if a particular component has sole responsibility for updating a particular kind of value, I'm OK with passing it around as `const` and then having **that** component strip away the `const`.)

jordanbrown0: Getting a little concrete there:  in an API context, a string value might well be a C/C++ constant string.  A constant string may be allocated in read-only memory.  Are you **absolutely sure** that nothing parsing a command line argument changes it, say by calling something like `strtok()` on it?  I'm not.  Either const-ify the value all the way down (and so preclude writing to it), or copy it so that it's clearly not const.

17682300365: Well, I copied the args here out of an abundance of caution regarding this concern.

src/libopenscad.cc
```diff
@@ -0,0 +1,66 @@
...
+
+    // Store the original current working directory.
+    const auto original_path = fs::current_path();
+
+    // Execute the main OpenSCAD functionality with the prepared arguments.
+    const int result = openscad_main(argc, arg_ptrs.data(), true);
+
+    // Restore the original working directory after execution.
+    fs::current_path(original_path);
```
jordanbrown0: Note that `fs::current_path()` is not thread-safe; it changes a process global.  If OpenSCAD calls it or chdir(), that's a problem for multithreaded callers.

src/libopenscad.cc
jordanbrown0: You copied an array of `const char *` pointers into a vector of `char *` pointers, changing the types of the pointers but not copying the strings.  It is the strings that the `const` applies to.

17682300365: Indeed, I'll make the change.

CMakeLists.txt
17682300365: Currently, with reference to several other submodules, it is controlled by BUILD_SHARED_LIBS.

kintel: The submodules only use this to switch between static vs dynamic libraries. I think we should still have an option for building OpenSCAD as a static library, which is currently missing.

src/openscad.cc
kintel: "Main process termination" really depends on the upstream use-case. e.g. for a web browser, is it when you close the page, or quit the browser?

I'm not necessarily saying we need to solve all these in this PR, but if unsolved it would be nice to leave a mention of it in the code.

src/libopenscad.cc
kintel: We should fix the downstream API. That may be out of scope for this PR, but please leave a comment explaining why jump through these hoops. I'm pretty sure nobody modifies argv, perhaps except the Python subsystem.

src/libopenscad.cc
kintel: Good call: We should define some rules of engagement. Thread safety and global process state are two.

Eventually we may need more; things like explicit cache clearing to avoid accumulating too much memory? Ability to provide a GL context for correct rendering? That belongs in a new ticket.

17682300365: Yes, I'm currently troubled by this thread safety issue as well. For now, I've worked around it using other methods, but long-term I feel it needs to be resolved.

t-paul: Yes, at some point we need to get rid of the CWD changes in OpenSCAD as this collides in other places too, like when multiple main windows are open. But that's probably for a separate PR.

coryrc:
I need a library so I can make unit tests. Is this still the way forward?

kintel:
> I need a library so I can make unit tests. Is this still the way forward?

I think your needs are slightly different; you need access to all the internal symbols, while this one only exports `main()`. It's possible that the two can be combined though, perhaps just live side-by-side as two separate libs sharing some common build infra?

jordanbrown0:
Since the internal-unit-test needs are different from this one, let's pull them out into a separate issue.

jordanbrown0:
See #6344 for internal testing.

src/libopenscad.cc
```diff
@@ -0,0 +1,66 @@
+/*
```
coryrc, line 1: nit: the copyright header isn't needed; please remove from all new files.

src/libopenscad.cc
```diff
@@ -0,0 +1,66 @@
...
+    const int result = openscad_main(argc, arg_ptrs.data(), true);
+
+    // Return the result of the OpenSCAD execution.
+    return result;
+  } catch (const std::exception& e) {
+    // Log any exception that occurs during command line execution and return 1 to indicate failure.
+    LOG("Command line execution failed: %1$s", e.what());
+    return 1;
+  }
+}
```
coryrc, line 66: nit: please add newline at EOF

CMakeLists.txt
coryrc: Please rebase off #6299. The changes I have there is to build library targets `OpenSCADLibInternal` and `svg` which expose all APIs. You can create a new target `OpenSCADLib` (or whatever) which exposes the `lib_openscad` you are trying to publish.

Issue #6285, opened 2025-10-15
3mf export made mandatory and the default method


coryrc:
**Is your feature request related to a problem? Please describe.**

#6165 STL export causes non-manifold errors. [It feels like it's slowly time to actively discourage people to export to STL, and drive them towards better export formats (both OBJ and 3MF should be reasonable alternatives I think).](https://github.com/openscad/openscad/issues/6165#issuecomment-3313169411) - Kintel.

**Describe the solution you'd like**

3mf made the default export method (F7).

**Describe alternatives you've considered**

Could also be OBJ. I don't know anything about OBJ.

**Additional context**

lib3mf is currently considered optional https://github.com/openscad/openscad/issues/6132

MichaelAtOz:
I vote: Too soon. I'm yet to gear-up to 3mf. (but I still have one PC with Windows-7).
Tho as there is Preferences/Export_features, if you mean set it default there, I could live with it I suppose...

jordanbrown0:
The preference applies only to the toolbar button and not to F7 so it's not a complete solution, even for opt-in.

#3982 more or less covers F7.

jordanbrown0:
> I vote: Too soon.

Unless we wait until absolutely nobody is using STL, there will always be a time when the default has changed and people who want the old behavior will have to select it as a preference.  That's just the way that such transitions work:  first you have to opt-in to the new thing, then you have to opt-in to the old thing, then (usually, eventually) the old thing goes away.

t-paul:
Yes, I think we'll want to keep the default assignment. Focus should be to get #3338 finished eventually. i did look at this not too long ago, but did not get too far yet.

coryrc:
@t-paul [I was asked by Marius to open this.](https://github.com/openscad/openscad/issues/6165#issuecomment-3404389888)

coryrc:
> I vote: Too soon. I'm yet to gear-up to 3mf. (but I still have one PC with Windows-7).

OBJ would still be compatible, right?

But isn't the Windows build built with lib3mf? Does that work on win7 still?

t-paul:
Hmm, that does not match what this issue asks for. "Driving people to use 3MF" (I totally agree with) can be done via different means, not necessarily by changing the keyboard defaults.

[...]

Issue #6132, opened 2025-08-26
Add lib3mf to scripts/check-dependencies.sh


coryrc:
The CI automatically installs it, appears Nix and Mac OS get it; it should be in check-dependencies.sh especially as some of the tests rely on it.

t-paul:
This script is only supposed to check the required core libraries. Lib3mf is optional.

t-paul:
Note that for now this will still need to check for both v1 and v2. It's probably going to be tricky to cover all cases.

PR #6288, opened 2025-10-16
Require lib3mf by default

kintel:
This PR makes lib3mf a required dependency by default.

This is configurable, to make lib3mf optional:
```
cmake -DCMAKE_REQUIRE_FIND_PACKAGE_Lib3MF=FALSE
```

Ref. #6285


kintel:
@nomike FYI - I know you hinted at wanting to do something similar across the board. This is more of a quickfix, but using a built-in CMake feature, so it doesn't add any custom magic.

coryrc:
@kintel can you resurrect #6132 to go along with this?

kintel:
I believe the scripts in question are meant to manage required dependencies - not sure where the lines is drawn between "required" and "opt-out" for Linux as the ecosystem is so fragmented. For Windows and macOS we have tighter control and can probably require those. Is there a path for showing warnings for opt-out deps?

coryrc:
> I believe the scripts in question are meant to manage required dependencies - not sure where the lines is drawn between "required" and "opt-out" for Linux as the ecosystem is so fragmented. For Windows and macOS we have tighter control and can probably require those. Is there a path for showing warnings for opt-out deps?

I can do that; I'd do it by ignoring if the user opt-out via the mechanism you have and adding a note offering that option when detecting its absence otherwise. 

My annoyance was partially that it's actually required to run the tests; they aren't skipped when 3mf isn't compiled in; but also that I wanted it.

Issue #3982, opened 2021-11-21
Suggestion: F7 does "preferred export". 


rewolff:
I often work with 3D models and F7-export-STL is very convenient. 

I now that I'm working with 2D stuff, I find myself hitting F7 and getting "top level object is not a 3D object". 

I think it would be useful if: 
* when the top level opbject is 2D, then F7 would do: "preferred 2D export". 
* that "preferred 2D export" and "preferred 3D export" would be configurable. 

I can imagine the frustration that those working with say AMF have that they don't have a hotkey like STL-users have. 
And similarly, I respect others who's 2D workflow doesn't use DXF.

t-paul:
Right, in general we need to get #3338 brushed up. But it seems useful to have this kind of "Smart-Export" action.

jordanbrown0:
The toolbar export button is controlled by a preference.  It seems like it would be straightforward and desirable to have F7 be controlled by that same preference.

jordanbrown0:
We now have two export buttons, currently one 2D and one 3D.  There's a request to make there be two 3D export buttons (perhaps allowing the two buttons to each be either 2D or 3D), see #5714.  Such a change would mean that the button's exporter selection would not be enough to set F7's export.  We'd need a separate selection.

It's not exactly the target market, but we might also consider a default 2D export.  Perhaps F7 would export to a specified 3D format if you have a 3D object, and to a specified 2D format if you have a 2D object.

jordanbrown0:
Marius just changed it so that F7 triggers the default 3D export.  The next step would be bringing 2D into it somehow.

PR #6290, opened 2025-10-17
Rebind F7 to trigger the currently selected 3D toolbar action

kintel:
Ref #6285

### DIFF
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -245,20 +245,6 @@ void removeExportActions(QToolBar *toolbar, QAction *action)
   }
 }
 
-void addExportActions(const MainWindow *mainWindow, QToolBar *toolbar, QAction *action)
-{
-  for (const std::string& identifier :
-       {Settings::Settings::toolbarExport3D.value(), Settings::Settings::toolbarExport2D.value()}) {
-    FileFormat format;
-    fileformat::fromIdentifier(identifier, format);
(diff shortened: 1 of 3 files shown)


jordanbrown0:
See also #3982.  That's maybe a bit more expansive:  it wants to allow having F7 do a 2D export.

But:  for keystroke and toolbar-button export, do we really need 2D and 3D variants?  The model is either 2D or 3D.  It never makes sense to do a 2D export of a 3D model, nor does it make sense to do a 3D export of a 2D model.

Why do we need two buttons?  Why not have one export button and one export keystroke, that exports 2D models using the selected 2D export and 3D models using the selected 3D export?

(For super extra credit, make the icon automatically change based on whether the model is 2D or 3D.)

Note that "export as image" applies to both types, but is not considered an export for these purposes; you cannot select PNG as your preferred export type for 2D.

kintel:
While I agree that this could be better configurable, I went for the baby steps approach - GUI programming sucks and I want out of there as quickly as possible ;)

jordanbrown0:
Looks good to me.

jordanbrown0:
But:  I see that you're connecting F7 to the onWindowShortcutExport3DActivated action in C++.

I don't know much about Qt, but can't you do that with an &lt;action> stanza in MainWindow.ui?

kintel:
It might be possible to do it in the .ui file, but this is associated with a "naked" QShortcut, and AFAIK those are not designable in Qt Designer.

PR #6300, opened 2025-10-19
Fix/cleanup compiler warnings openscad/openscad#6245

Ma-XX-oN:
Removing compiler noise to make a clean build.


t-paul:
I very much like the cleanup in general. Changing files in `src/ext` is an exception to that as it's external code. I think the better strategy there would be checking if it's maybe fixed upstream and can be updated this way. Or posting the fix in the upstream repo.

Ma-XX-oN:
Oh.  Didn't realise that `src/ext` was external.  I thought that was what `modules/submodules` was for.  I'll remove it and see if it can be fixed upstream.

Ma-XX-oN:
Hmmm. I don't actually see any attached .git directory.  Where is it from @t-paul?

t-paul:
https://github.com/openscad/openscad/blob/cc47f06be213bf7b2fd62eb2bf69cdedb25abb99/src/ext/json/json.hpp#L1-L7

t-paul:
Yeah, maybe we should at least put a README.md into the `src/ext` folder to make it more discoverable that this is imported code (by copying instead of git submodules).

src/utils/StackCheck.h
```diff
@@ -3,18 +3,14 @@
...
-#endif  // defined(__GNUC__)
+#endif  // defined(__GNUC__) && !defined(__clang__)
 #if defined(_MSC_VER)
 #pragma warning(push)
 #pragma warning(disable : 26486)  // Disable warning for dangling pointers
 #endif                            // defined(_MSC_VER)
-#if defined(__clang__)
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wdangling-pointer"
-#endif  // defined(__clang__)
```
kintel, line 17: I believe this is a warning since the flag was introduced in some later version of clang; it should probably check for clang version.

src/utils/StackCheck.h
```diff
@@ -3,18 +3,14 @@
 #include <cstdlib>
 #include "platform/PlatformUtils.h"
 
-#if defined(__GNUC__)
+#if defined(__GNUC__) && !defined(__clang__)
```
kintel: When is __GNUC__ and __clang__ ever defined at the same time?

src/utils/StackCheck.h
Ma-XX-oN: I have version 21.1.1 and I've attempted to find any version that accepted it and I can't find one.
Most recent documentation: https://clang.llvm.org/docs/DiagnosticsReference.html

I could do:
```
#if defined(__clang__)
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wdangling"
#endif  // defined(__clang__)
```
or some subset.  Is there a reason why StackCheck.h generates dangling pointers?

src/utils/StackCheck.h
Ma-XX-oN: From https://clang.llvm.org/docs/UsersManual.html
> The Clang driver and language features are intentionally designed to be as compatible with the GNU GCC compiler as reasonably possible, easing migration from GCC to Clang. In most cases, code “just works”. Clang also provides an alternative driver, clang-cl, that is designed to be compatible with the Visual C++ compiler, cl.exe

If we don't want clang to try and be a drop in replacement for gcc, then we should use this flag and set it to 0 in CMakeLists.txt.
> -fgnuc-version=
This flag controls the value of __GNUC__ and related macros. This flag does not enable or disable any GCC extensions implemented in Clang. Setting the version to zero causes Clang to leave __GNUC__ and other GNU-namespaced macros, such as __GXX_WEAK__, undefined.

src/utils/StackCheck.h
kintel: Right, perhaps it's just -Wdangling on clang.
There's a dangling pointer since we retain the address of a stack variable at the start of the program, to get a sense of how much stack we use.

kintel: Btw., I'm not in love with how these warnings are suppressed, as it's quite verbose. I don't have better ideas though.

Ma-XX-oN: > Right, perhaps it's just -Wdangling on clang. There's a dangling pointer since we retain the address of a stack variable at the start of the program, to get a sense of how much stack we use.

If you are pointing at something that is there, that shouldn't cause a dangling pointer as it is pointing as something.

Where is the dangling pointer reported?  I didn't see anything when I built with clang when there was no warning suppression.

> Btw., I'm not in love with how these warnings are suppressed, as it's quite verbose. I don't have better ideas though.

The only other way is to put the warning suppression in the CMakeLists.txt, but that would affect everything, not just the header, which isn't what is wanted.

kintel: This was contributed by @nomike to kill warnings surfacing in in his environment. I've never seen such warnings myself. Perhaps he can comment on this. If not, we can always assume the problem has gone away and fix it if it resurfaces.

Ma-XX-oN:
I've posted an issue with [nlohmann](https://github.com/nlohmann)/[json](https://github.com/nlohmann/json) here: https://github.com/nlohmann/json/issues/4956.  @t-paul, it's not fixed there.

Ma-XX-oN:
@t-paul, its now been fixed.

https://github.com/nlohmann/json/pull/4957

src/utils/StackCheck.h
Ma-XX-oN: @nomike, I don't see any warnings when I build in MSYS gnu or clang after I remove the pragmas for each.  I've not tried MSVC because I've not been able to get it to work for me yet.  What environment were you building in and what commit were you building?

kintel: According to the PR introducing this, it was run on GCC-14.2.0 on Ubuntu 25.04, but I'm not sure if that was the configuration on which the warning happened.

I'd say feel free to do this whichever way you feel is correct, and we can revisit this if it resurfaces, and next time document whether this is a warning only surfacing on specific systems.

Ma-XX-oN: Ok. Since I verified that there isn't any warning reported on clang and gcc on MSYS, I'm just going to remove those from the file.  I'll leave the MS one in.

PR #6303, opened 2025-10-19
Calendar-based release

flaviut:
I'm proposing that the OpenSCAD move from a feature-based release process to a calendar based release process, similar to that that Firefox, Linux, and Docker. This transition will be supported by the existing nightly release automation, which with relatively minor tweaks, can be repurposed as release automation.

# FAQ

## Why is this needed?

Cutting a new release is a big burden on the release manager, so it doesn't end up happening. A fully automated process allows this responsibility to be diffused out among all contributors, in the form of passing CI on master and a shared responsibility to only merge fixes during the feature freeze period.

## Why not tell users to just use nightlies?

Package managers and app stores have polices around only publishing official releases, not just git snapshots. We need to cut official releases on a regular basis to make the default user experience a good one. Advanced users will read forums and see out additional information, but as it is, casual users will receive a poor first impression of OpenSCAD from the old versions available in the package repos.

## Why every 3 months?

This is admittedly a rapid pace. The goal is to build the organizational muscle memory around the process, and to quickly work out kinks in the automation. Once this is put into practice, it can always be changed to be longer or shorter in response to user and developer feedback.

# Additional required work

Additional work that is needed for this transition:

- [x] replace `doc/openscad.1` with `doc/openscad.1.in`, and have cmake automatically set the version number
- [ ] build Snap & Flatpak as part of OpenSCAD's pipeline as well
- [ ] pipelines will detect if they are running on a tagged version. if they are, they will use the tag as the release version
- [ ] github action added to on the last of the month
  - run makereleasenotes, and commit changes
  - create a git tag
  - push both to github
- [ ] makereleasenotes to be converted to python, and updated to handle next.md:
  - copy next.md into appropriately named file given the current release version
  - create a new next.md file from template
- [ ] src_release_links to be removed, people should get sources directly from github

### DIFF
--- a/doc/release-checklist.md
+++ b/doc/release-checklist.md
@@ -53,3 +53,4 @@ reference, announce the new release:
     - Arch Linux: Kyle Keen <keenerd@gmail.com>
     - MacPorts: https://svn.macports.org/repository/macports/trunk/dports/science/openscad/Portfile
     - Homebrew: https://github.com/caskroom/homebrew-cask/blob/master/Casks/openscad.rb
+- Make & merge PRs to the Flatpak and Snap repos bumping the version number


doc/release-checklist.txt
```diff
@@ -1,91 +0,0 @@
...
-o Notify package managers
-  - Debian/Ubuntu: https://launchpad.net/~chrysn
-  - Ubuntu PPA: https://github.com/hyperair
-  - Fedora: Miro Hrončok <miro@hroncok.cz> or <mhroncok@redhat.com>
-  - OpenSUSE: Pavol Rusnak <prusnak@opensuse.org>
-  - Arch Linux: Kyle Keen <keenerd@gmail.com>
-  - MacPorts: https://svn.macports.org/repository/macports/trunk/dports/science/openscad/Portfile
-  - Homebrew: https://github.com/caskroom/homebrew-cask/blob/master/Casks/openscad.rb
-o Update dev version to release version in documentation
-  - https://en.wikibooks.org/wiki/OpenSCAD_User_Manual
```
flaviut, line 68: not needed, doesn't seem to have a link to the latest version (and if it did, I suggest setting it to the evergreen download page)

doc/release-checklist.txt
```diff
@@ -1,91 +0,0 @@
...
-o Sanity check; build a binary or two and manually run some tests
-
-o git push --tags master
-
-o Upload Source package
-  $ scp openscad-$VERSION.src.tar.gz openscad@files.openscad.org:www
-
-o Build binaries for all platforms and wait for upload
-
-o Announce:
```
flaviut, line 49: automated as part of CI

doc/release-checklist.txt
```diff
@@ -1,91 +0,0 @@
...
-    64-bit: run on a 64-bit machine or VM
-
-    $ ./scripts/release-common.sh -> openscad-$VERSION.x86-ARCH.tar.gz
-    (where ARCH will be detected and set to 32 or 64)
-    $ scp openscad-$VERSION.x86-ARCH.tar.gz openscad@files.openscad.org:www
-    o Update web page with download links
-
-Windows mingw cross-build:
-
-FIXME: Adapt scripts/builder.sh to build release binaries
```
flaviut, line 91: automated as part of CI

doc/release-checklist.txt
```diff
@@ -1,91 +0,0 @@
...
-    o In MCAD clone:
-      $ git fetch upstream
-      $ git merge upstream/master
-      $ git push
-    o In OpenSCAD:
-      $ cd libraries/MCAD
-      $ git pull
-      $ cd ../..
-      $ git commit -m "Updated MCAD" libraries/MCAD
-      $ git push
```
flaviut, line 19: doesn't seem to be receiving updates anymore

doc/release-checklist.txt
```diff
@@ -1,91 +0,0 @@
-OpenSCAD Release Checklist
---------------------------
-
-o Pre-release preparations
-  - Update translations:
-    - Send emails to translation managers, see .po files:
-      https://github.com/openscad/openscad/tree/master/locale
-  - Update resources/html/AboutDialog.html with new contributors etc.
```
flaviut, line 8: would suggest moving this to the PR template

doc/release-checklist.txt
```diff
@@ -1,91 +0,0 @@
-OpenSCAD Release Checklist
---------------------------
-
-o Pre-release preparations
-  - Update translations:
-    - Send emails to translation managers, see .po files:
-      https://github.com/openscad/openscad/tree/master/locale
```
flaviut, line 7: This becomes part of the scheduled feature freeze. Translation maintainers should set up calendar reminders etc on the appropriate months.

kintel:
The main caveat is that we expect OpenSCAD designs authored by one version to be usable by future versions, so this type of release schedule, while attractive, would require us to be extremely careful about what kind of language-level changes we merge to master. This is largely solved, technically, by the experimental feature system, and we've wanted to make it possible to continuously release from master for a while. I think the first checkpoint would be to make it _possible_ to release from master, then work on how to automate or regularize that.
For the next release plan, see https://github.com/orgs/openscad/projects/2 - once that's out we can discuss techniques and workflows for better release methodologies.

kintel:
I'm closing this as this is a larger topic and the needed discussion should probably be held somewhere else than in a pull request. Mailing list or IRC are good places to start.

t-paul:
I'm not sure the mailing list would be the best place as this is much more focused on the user side than development details. To actually make things move into the direction asked here, I could see smaller issues created and implemented for a much more automated release workflow. This could target getting RCs build in addition to the current fully automatic nightly build pipeline.

If anyone wants to help doing that (great!), we could start with some small issues and discuss on IRC. BUT, please only create issues with the intention of actually helping with the implementation. If the expectation is "someone else" should do it, that would be just me. I really don't need extra work handling a flood of issues for the build pipeline instead of using the scarce time to implement stuff. First tiny issue could be the first bullet point from the description above taking care of the manual page...

kintel:
t-paul worded it better than me :) I fully agree that iteratively establishing automation to enable us to approach calendar-based releases makes sense. Even it calendar-based release may be a longer-term opportunity, such automation could significantly help with and rerisk manual releases too.

PR #6311, opened 2025-10-22
Check for save/discard/cancel on quit.  Fixes #3971.

### LINKED ISSUE #3971: ctrl + q leaves changes unsaved
quitting with ctrl+q didn't pop up dialog  for unsaved changes it just closes reverting to last save.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/112040433-ctrl-q-leaves-changes-unsaved?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jordanbrown0:
When the user requests a quit, ask each MainWindow to close.  If any of them refuse, don't quit.

One can imagine a number of different behaviors here.

If you have two MainWindows, and one of them says "yes" and the other says "no", should you leave both MWs open, or should you close the one that said "yes"?  I started out with the more transactional "if any of them say no, don't close any", but when I actually tried that it seemed wrong:  if I said "discard" on a window and another window declined to close, the "discard" decision was ... discarded, and I'd be asked again on the next quit attempt.  This one closes any MW that agrees to close, so the next time you aren't asked again about that window (because it's gone).

There's a similar question around "cancel".  If you say "cancel", does that mean "stop trying to quit" or does it mean "don't close *this* window, but keep trying to close other windows".  I started out with continuing to close other windows, but that didn't seem very cancel-y; this iteration stops trying to quit the first time you hit "cancel".

---

One thing slightly confuses me:  I originally marked OpenSCADApp::quit as "override", but the compiler tells me it doesn't override.  OpenSCADApp is a subclass of QApplication, and QApplication has a quit(), so it sure looks like an override to me.  But it's happy without the "override" marker.

### DIFF
--- a/src/gui/OpenSCADApp.cc
+++ b/src/gui/OpenSCADApp.cc
@@ -93,3 +93,24 @@ void OpenSCADApp::setApplicationFont(const QString& family, uint size)
   )");
   scadApp->setStyleSheet(stylesheet.arg(family, QString::number(size)));
 }
+
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
+// https://doc.qt.io/qt-6/qtcore-changes-qt6.html#other-classes
+//
+//     In Qt 5, QCoreApplication::quit() was equivalent to calling QCoreApplication::exit().
+//     This just exited the main event loop.
+//
+//     In Qt 6, the method will instead try to close all top-level windows by posting a close
+//     event. The windows are free to cancel the shutdown process by ignoring the event.
+//
+// For Qt5, simulate the Qt6 behavior.
+void OpenSCADApp::quit()
+{
+  for (MainWindow *mw : scadApp->windowManager.getWindows()) {
+    if (!mw->close()) {
+      return;
+    }
+  }
+  QApplication::quit();
+}
+#endif
--- a/src/gui/OpenSCADApp.h
+++ b/src/gui/OpenSCADApp.h
@@ -18,6 +18,10 @@ public:
 
   bool notify(QObject *object, QEvent *event) override;
   void requestOpenFile(const QString& filename);
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
+  // See comments in OpenSCADApp.cc.
+  static void quit();
+#endif
 
 public slots:
   void showFontCacheDialog();


kintel:
> One thing slightly confuses me: I originally marked OpenSCADApp::quit as "override", but the compiler tells me it doesn't override. OpenSCADApp is a subclass of QApplication, and QApplication has a quit(), so it sure looks like an override to me. But it's happy without the "override" marker.

QCoreApplication::quit() is a static function.

src/gui/OpenSCADApp.cc
```diff
@@ -93,3 +93,13 @@ void OpenSCADApp::setApplicationFont(const QString& family, uint size)
   )");
   scadApp->setStyleSheet(stylesheet.arg(family, QString::number(size)));
 }
+
+void OpenSCADApp::quit()
+{
+  for (MainWindow *mw : scadApp->windowManager.getWindows()) {
+    if (!mw->close()) {
```
kintel, line 110: The weird part is that Ctrl-Q should already send a close event to all windows, and we handle this in the notification chain. `OpenSCADApp::notify()` even intercepts and forwards it, then `MainWindow::event()` also intercepts and forwards the event, which should eventually call `MainWindow::closeEvent()`.

While your fix seems to work, I think the big question is why (and where) did it not work before?

I haven't been able to reproduce the issue myself, so I'm only going off reports..

kintel: PS! I did some related work to cleanup the quit process in #5668

jordanbrown0: It may be relevant that while Qt6's QCoreApplication::quit() is defined to close all windows, Qt5's isn't.

https://doc.qt.io/archives/qt-5.15/qcoreapplication.html#quit
https://doc.qt.io/qt-6/qcoreapplication.html#quit

jordanbrown0: https://doc.qt.io/qt-6/qtcore-changes-qt6.html#other-classes
>In Qt 5, QCoreApplication::quit() was equivalent to calling QCoreApplication::exit(). This just exited the main event loop.

>In Qt 6, the method will instead try to close all top-level windows by posting a close event. The windows are free to cancel the shutdown process by ignoring the event.

jordanbrown0: So it seems like the first question is how close we are to switching to Qt6.

jordanbrown0: Built on MSYS2 with Qt6, it seems to work properly without my change.

kintel: Ahh, that makes a ton of sense; I haven't used Qt5 in a loong time, and it doesn't even build on macOS any longer.. Perhaps the right solution is to wrap your code in a compile-time check.

kintel: > So it seems like the first question is how close we are to switching to Qt6.

https://github.com/openscad/openscad/issues/4909

I believe teepee is slowly working through the MXE parts, which is the hardest part to do.

jordanbrown0:
Added the version check and some commentary.  Built and tested with both Qt5 and Qt6.

I did observe one ... interesting ... behavior.  If you have the editor window undocked, and you have two windows with changes (and thus two 3D views and two editor windows), when you [Discard] the first window, its 3D view window closes but its editor window does not close.  Cancelling or completing the quit closes that "orphan" editor.  I'm inclined to say that the fact that the editor window is left open is a bug, but that it is a separate bug.

jordanbrown0:
... and that's true in both Qt5 with this change and Qt6 without.

jordanbrown0:
I don't have any reason why this isn't ready to merge.

Issue #6315, opened 2025-10-23
openscad.desktop does not contain a StartupWMClass on Fedora workstation 42 install from the repo


nicvagn:
File generated on install:
```
[Desktop Entry]
Type=Application
Version=1.0
Name=OpenSCAD
Icon=openscad
Exec=openscad %f
MimeType=application/x-openscad;
Categories=Graphics;3DGraphics;Engineering;
Keywords=3d;solid;geometry;csg;model;stl;
```

Fixed file, icon works on Gnome

```
Type=Application
Version=1.0
Name=OpenSCAD
Icon=openscad
Exec=openscad %f
MimeType=application/x-openscad;
Categories=Graphics;3DGraphics;Engineering;
Keywords=3d;solid;geometry;csg;model;stl;
StartupWMClass=org.openscad.openscad
penscad.openscad
```

`StartupWMClass=org.openscad.openscad`
Was the only change. I do not know where the problem originates, upstream or in Fedora workstation 42

nicvagn:
```
if(CMAKE_INSTALL_COMPONENT STREQUAL "Unspecified" OR NOT CMAKE_INSTALL_COMPONENT)
  file(INSTALL DESTINATION "${CMAKE_INSTALL_PREFIX}/share/applications" TYPE FILE RENAME "openscad.desktop" FILES "/usr/local/src/openscad/resources/icons/openscad.desktop")
endif()

```

I am trying to install from src and:


``` /usr/local/share/applications/openscad.desktop
[Desktop Entry]
Type=Application
Version=1.0
Name=OpenSCAD
Comment=The Programmers Solid 3D CAD Modeller
Icon=openscad
Exec=openscad %f
MimeType=application/x-openscad;
Categories=Graphics;3DGraphics;Engineering;
Keywords=3d;solid;geometry;csg;model;stl;
```

The problem is there too

t-paul:
fixed via #6316

PR #6316, opened 2025-10-24
add correct StartupWMClass to set the correct icon on Gnome

nicvagn:
I believe this is the XDG way to do this, but I am not sure.

https://specifications.freedesktop.org/desktop-entry-spec/1.1/recognized-keys.html

It sets the runtime icon.

addresses: #6315


t-paul (approved):
I'm a bit surprised it works with all lowercase, we usually use `org.openscad.OpenSCAD` but if it works for you, lets go with that. Thank you!

PR #6330, opened 2025-10-29
reformat python tests scripts in pep8

AaronVerDow:
This is a formatting-only change for the Python testing scripts to help make future pull requests more readable.

These have been formatted with Python Black for PEP8.

### DIFF
--- a/tests/export_import_pngtest.py
+++ b/tests/export_import_pngtest.py
@@ -29,70 +29,87 @@
 import sys, os, re, subprocess, argparse
 from validatestl import validateSTL
 
+
 def failquit(*args):
-    if len(args)!=0: print(args, file=sys.stderr)
-    print('export_import_pngtest args:',str(sys.argv), file=sys.stderr)
-    print('exiting export_import_pngtest.py with failure', file=sys.stderr)
+    if len(args) != 0:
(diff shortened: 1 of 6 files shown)


kintel:
Thanks! This kind of cleanup really makes reviews easier.

t-paul:
Should that go into `.git-blame-ignore-revs`?
Is there a way to keep it clean using the `pre-commit` hooks?

kintel:
Possibly - I'm not a big fan of having to manually maintain such things. If a script could auto-detect whitespace-only change and sort it out, cool :)

AaronVerDow:
> Should that go into `.git-blame-ignore-revs`? Is there a way to keep it clean using the `pre-commit` hooks

A pre-commit hook is definitely possible in the future, for now I just need to get the SVG tests working though.

t-paul:
Sorry for not being specific. I did not mean a generic git pre-commit hook, but [pre-commit.com](https://pre-commit.com/) using https://github.com/openscad/openscad/blob/master/.pre-commit-config.yaml

t-paul:
> for now I just need to get the SVG tests working though.

Oh, also I meant my comment more like a general question / note for later. I'm not saying it's required right now or asking to change what you are working on.

Issue #6312, opened 2025-10-23
undocumented feature menu File > Open creates new edit tabs with relevant command per file type


bitbasher:
i did not know this was here .. File > Open and select a JSON file will create a new edit tab with this command in it:

```
data = import("D:/repositories/openscad-examples-docs/example-curves-fragments.json");
```
likewise
```
surface(invert=false, "D:/repositories/openscad-examples-docs/escaped-unicode-codes.png",1);
```
and
```
import( "D:/repositories/openscad-examples-docs/example-baseplate-convert.dxf");
```
and
```
import("D:/repositories/openscad-examples-docs/surface.stl");
```

so we should document all this in the[ User Interface section](https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/The_OpenSCAD_User_Interface) by making a new section for the Menus, File, Edit etc.

we should start with FIle > Open and add more menu items as people have time to write them up

jordanbrown0:
You can also drag and drop these file types into the editor window, with similar effect.

coryrc:
> /assign_me

(it's a hyphen, not underscore)

jordanbrown0:
#6341 is related.  I think it will provide a better mechanism for picking files and inserting import calls for them.

bitbasher:
i am back .. i will be able to work on this again

jordanbrown0:
Note that to trigger this undocumented behavior you have to explicitly type the full name of the file, including the extension.  You can't pick such a file from the file picker, because it won't show you anything but `.scad` and `.csg` files.

This is strongly related to #4791, which asks for the ability to edit arbitrary text-based files.

See also PR #6341, which adds the ability to create a template that pops a file picker to insert the file name into the resulting text.

Issue #4791, opened 2023-10-25
The editor should allow to load and edit any text file not only scad files.


yahbluez:
While openscad did not allow scripting actions,
it is useful to run scripts in other langauages
and there for sometimes minor changes to the scripts or data files is necessary.

Would be nice to stay in the same editor.

solution:
remove the fixed file ending and allow to load any kind of text file.

jordanbrown0:
The standard Windows user interface for this is to have the file-type pulldown (in the bottom right of an Open dialog, above [Open] and [Cancel]) include "All Files (*.*)", as in this snippet from Notepad:

![image](https://github.com/openscad/openscad/assets/31949071/cb7669c5-6f78-4e86-8e0a-d865c00ff251)

yahbluez:
![image](https://github.com/openscad/openscad/assets/47297504/07685b60-ce87-46c1-ad6d-58870eaa584f)

In Linux too, but the selector is fixed to openscad files, i can't click it and choose/edit others.

OpenSCAD  version 2023.10.12
Manjaro Linux

jordanbrown0:
Yes, sorry, I was not saying that OpenSCAD did what you want on Windows.  It is the same on Windows; it only offers *.scad and *.csg.  I was trying to point out how it *should* work.

jordanbrown0:
One complexity:  OpenSCAD has some [existing special handling](https://github.com/openscad/openscad/blob/f602162d26be3ed2a4aa6e8e64ae6c5c66fb73bc/src/gui/MainWindow.cc#L295) of a few file types (stl, obj, 3mf, off, dxf, svg, amf, dat, png, json).  You can't open them using the dialog, but if you drag-and-drop one of them on OpenSCAD, or open them via the command line, the effect is to insert an `import()` or `surface()` referring to the file.

yahbluez:
Yah, that work around is be used, but it would be much more useful to remove the limitation from the file open dialog. This limitations are hand made and not the default for file open dialogs.

jordanbrown0:
Maybe the special handling could be disabled when the file is loaded via the Open dialog.

Or maybe you can't edit those few file types.  (But not being able to edit json would be a nuisance.)

Or maybe we could just get rid of that special handling.

But unfortunately [XKCD 1172](https://xkcd.com/1172/) applies.

keithel:
[...]

PR #6341, opened 2025-11-06
File-picking template

### LINKED ISSUE #5172: Allow selecting STL files to import via a file dialog

jordanbrown0:
The idea is to have a way for a template to say "ask the user for a file of *this* type, and put its name *here* in the text inserted by the template", and then there would be templates for the `import()` module, the `import()` function, `surface()`, `include<>`, `use<>`, et cetera.  This would be tidier than having magic "open" semantics to insert these statements, would (I think) be more convenient, et cetera.  It would allow moving **away** from the magic "open" semantics, and thus allow "open" to open text-based files (like DXF, SVG, and ASCII STL) in the editor.

Here's a sample template JSON file using this mechanism:

```
{
    "key" : "surface(file)",
    "content" : "surface(\"^~^\");",
    "fileFilter": "Common (*.png *.dat);;Images (*.png);;Data (*.dat);;All (*.*)"
}
```

Fixes #5172.

### DIFF
--- a/.github/workflows/beautify.yml
+++ b/.github/workflows/beautify.yml
@@ -39,7 +39,7 @@ jobs:
 
     - name: Upload patch artifact
       if: steps.check.outcome == 'failure'
-      uses: actions/upload-artifact@v4
+      uses: actions/upload-artifact@v5
       with:
         name: beautify-patch
         path: ${{ runner.temp }}/beautify.patch
--- a/.github/workflows/code-tidy.yml
+++ b/.github/workflows/code-tidy.yml
@@ -80,7 +80,7 @@ jobs:
         cp OpenSCAD_clang-tidy.summary ${BASE_DIR}/${{ github.sha }}.summary
 
     - name: Upload clang-tidy Results
-      uses: actions/upload-artifact@v4
+      uses: actions/upload-artifact@v5
       with:
         name: clang-tidy Results
         path: |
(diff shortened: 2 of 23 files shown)


jordanbrown0:
Elaborating a bit on what the user sees...

User says "insert template", either with a right click or Alt+Ins, and gets the existing pop-up list of templates.  They pick (say) "Import shape".  It pops a file-open dialog, with the filter set to the file types of shape files (STL, 3MF, DXF, et cetera).  They select one of the files, and the template inserts an `import()` of that file.

jordanbrown0:
Related to #4791, which wants to edit arbitrary text-based files, and to #6312, which discusses the current behavior if you force File/Open to open a non-SCAD file.

PR #6348, opened 2025-11-09
Warn on offset() with both r and delta specified

jordanbrown0:
We currently silently ignore `delta` if both are specified.  We shouldn't ever do that - like we don't silently ignore `r` or `d` for round shapes when both are specified.

### DIFF
--- a/tests/data/scad/misc/errors-warnings.scad
+++ b/tests/data/scad/misc/errors-warnings.scad
@@ -99,3 +99,5 @@ cylinder();
 sphere()
 cylinder()
 cube();
+
+offset(r=1, delta=1);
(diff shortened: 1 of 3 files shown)


kintel (commented):
Should we add a corresponding test to https://github.com/openscad/openscad/blob/master/tests/data/scad/misc/errors-warnings.scad ?

kintel (approved):

jordanbrown0:
Should it also warn if you specify `r` and `chamfer`, given that `chamfer` isn't used in radial mode?

kintel:
Possibly - it's been a while since I played with offset() in detail..

PR #6356, opened 2025-11-12
std::wstring_convert and std::codecvt_utf8_utf16 is going to be depreciated. Use boost::nowide::narrow

nicvagn:
Boost is already used in OpenSCAD, and will silence the depreciation warning. If is a small change, and limited to the to_string function.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -16,6 +16,7 @@ on:
 jobs:
   build:
     strategy:
+      fail-fast: false
       matrix:
         qt: [ qt5, qt6 ]
     runs-on: windows-latest
@@ -33,15 +34,19 @@ jobs:
         if: ${{ github.event_name == 'workflow_dispatch' && inputs.debug_enabled }}
         with:
           detached: true
-      - name: Setup MSYS2 Environment
-        uses: msys2/setup-msys2@v2
-        with:
-          msystem: UCRT64
-          update: true
-      - name: Install MSYS2 Dependencies
+      - name: Calculate PACBOY_PACKAGES
+        shell: bash
         run: |
           pwd
           ./scripts/msys2-install-dependencies.sh ${{ matrix.qt }}
+      - name: Setup MSYS2 Environment
+        uses: msys2/setup-msys2@v2
+        id: msys2
+        with:
+          msystem: UCRT64
+          update: true
+          install: libxml2
+          pacboy: ${{ env.PACBOY_PACKAGES }}
       - name: Install Mesa
         uses: ssciwr/setup-mesa-dist-win@v2
         with:
--- a/scripts/msys2-install-dependencies.sh
+++ b/scripts/msys2-install-dependencies.sh
@@ -1,26 +1,29 @@
 #!/usr/bin/env bash
+set -e
 
 QT="$1"
 
-if [ -z $MSYSTEM ]; then
-  MSYSTEM_DEFAULT=UCRT64
-  # Possible values: (MSYS|UCRT64|CLANG64|CLANGARM64|CLANG32|MINGW64|MINGW32)
-  # For explanation of options see: https://www.msys2.org/docs/environments/
-  echo "MSYSTEM is unset or blank, defaulting to '${MSYSTEM_DEFAULT}'";
-  export MSYSTEM=$MSYSTEM_DEFAULT
-else
-  echo "MSYSTEM is set to '$MSYSTEM'";
-fi
-
 date "+### %Y-%m-%d %T msys2-install-dependencies started"
 
-pacman --query --explicit
+if [[ -z "${GITHUB_RUN_ID}" ]]; then
+    if [ -z $MSYSTEM ]; then
+	MSYSTEM_DEFAULT=UCRT64
+	# Possible values: (MSYS|UCRT64|CLANG64|CLANGARM64|CLANG32|MINGW64|MINGW32)
+	# For explanation of options see: https://www.msys2.org/docs/environments/
+	echo "MSYSTEM is unset or blank, defaulting to '${MSYSTEM_DEFAULT}'";
+	export MSYSTEM=$MSYSTEM_DEFAULT
+    else
+	echo "MSYSTEM is set to '$MSYSTEM'";
+    fi
 
-date "+### %Y-%m-%d %T install pactoys (for pacboy)"
-pacman --noconfirm --sync --needed pactoys libxml2
-# pacboy is a pacman wrapper for MSYS2 which handles the package prefixes automatically
-#            name:p means MINGW_PACKAGE_PREFIX-only
-#            name:  disables any translation for name
+    pacman --query --explicit
+
+    date "+### %Y-%m-%d %T install pactoys (for pacboy)"
+    pacman --noconfirm --sync --needed pactoys libxml2
+    # pacboy is a pacman wrapper for MSYS2 which handles the package prefixes automatically
+    #            name:p means MINGW_PACKAGE_PREFIX-only
+    #            name:  disables any translation for name
+fi
 
 if [[ "$QT" == "qt6" ]]; then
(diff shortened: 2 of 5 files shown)


kintel:
@jordanbrown0 You probably know the most about this domain by now

t-paul:
Boost is used, but `nowide` not yet, I think. Looks like it's available since boost 1.66, so that's fine. We probably need to update some dependencies to make it available.

nicvagn:
> Boost is used, but `nowide` not yet, I think. Looks like it's available since boost 1.66, so that's fine. We probably need to update some dependencies to make it available.

I think it is a warranted dependency. It would be verbose and error prone to roll your own converter.  On Linux/macOS, wchar_t is UTF-32, on Windows it’s UTF-16. :. there is no solution that covers all cases without a library or a branching code path.

t-paul:
> I think it is a warranted dependency.

I agree, I'm not advocating against the change. It's really just stating the fact that we need to watch out for potential build failures and update the scripts/ci-pipelines accordingly. Dependency changes are not always caught by test runners and/or package builds may need to declare this dependency (e.g. for *.deb, Snap, Flatpak, ...).

nicvagn:
It is a header only module, so no cmake changes should be required.  From my looking.

jordanbrown0:
Yeah, the C++ picture for UTF-8 is pathetic.

Note that for the particular case of Windows compatibility, we have a wide-to-UTF8 function:  https://github.com/openscad/openscad/blob/b0a4274914607db2a03dcfb5cdd7b98a73553e63/src/platform/PlatformUtils-win.cc#L30

jordanbrown0:
But I haven't looked at the HID stuff at all and don't know whether that applies.

kintel:
@jordanbrown0 Does this possibly interact with any of your changes in #6136 ?

jordanbrown0:
No.  The only real interaction is that we should really only have one wide-char-to-UTF-8-std::string function, and we have another in src/platform/PlatformUtils-win.cc winapi_wstr_to_utf8().  (Which is unused in master, but I added one call in my #6136 wad.)

It looks like this one is needed on all platforms (right?) and is more generic, so it should move to some more central place, get a better name, and replace winapi_wstr_to_utf8().  But there's no rush on that; they can coexist for the time being.

kintel:
> 
> It looks like this one is needed on all platforms (right?) 

I believe Windows is the only platform using wide chars, at least in the context of OpenSCAD.

jordanbrown0:
HIDAPI uses wchar_t.

https://github.com/openscad/openscad/blob/b0a4274914607db2a03dcfb5cdd7b98a73553e63/src/ext/hidapi/hidapi.h#L90

https://github.com/libusb/hidapi/blob/f18d2c0768bed300d13758dd9f52b57163c0498f/hidapi/hidapi.h#L165

I suspect that it's using them for strings that are retrieved from the device.

kintel:
> HIDAPI uses wchar_t.
> 
Oh, good point!

nicvagn:
I do not think that is valid. It is complaining about not finding cmake?
```
error: failed retrieving file 'mingw-w64-ucrt-x86_64-qscintilla-qt5-2.14.1-4-any.pkg.tar.zst.sig' ...
error: failed to commit transaction (unexpected error)
Errors occurred, no packages were upgraded.
```

and 

```
line 2: cmake: command not found
Process completed with exit code 127.
```
I think this is a issue with the windows workflow

kintel:
The Windows workflow has been a bit flaky lately - not sure why. Until we get it fixed, we can try rerunning it..

kintel (commented):
While at it, would you mind replacing winapi_wstr_to_utf8() in src/platform/PlatformUtils-win.cc with the narrow() function as well?

jordanbrown0:
> While at it, would you mind replacing winapi_wstr_to_utf8() in src/platform/PlatformUtils-win.cc with the narrow() function as well?

Sounds good to me.  As a sanity check, try using the command line to open a file with a non-ASCII name:

openscad ñämé.scad

You don't even have to create the file; just verify that the name ends up right in the window title.

t-paul:
The "`cmake` not found" just means the package download earlier. Usually re-running after a couple of hours works. They might get better via #6384.

nicvagn:
https://github.com/openscad/openscad/pull/6356/commits/6007cf00ea8217296b914ed8a1eb6b922f0867e3

is the major change. Hiding a bit of complexity.

The /opt/openscad/build/openscad ñämé.scad displays correctly for me, but I do not have Windows so I can not test properly.

jordanbrown0:
Come to think of it, the automated tests should be fine.  Thanks.

src/platform/PlatformUtils-win.cc
```diff
@@ -26,39 +27,16 @@
...
-  cbMultiByte = numbytes;
-
-  int result = WideCharToMultiByte(CodePage, dwFlags, lpWideCharStr, cchWideChar, lpMultiByteStr,
-                                   cbMultiByte, lpDefaultChar, lpUsedDefaultChar);
-
-  if (result != numbytes) {
-    DWORD errcode = GetLastError();
+  try {
+    return boost::nowide::narrow(wstr);
+  } catch (const std::exception& e) {
```
kintel: My understanding of `boost::nowide::narrow()` is that it does _not_ throw any exceptions. Have you made other experiences?

jordanbrown0: Sorry, didn't look too deeply at the details.

Yes... I don't know what the behavior is for an exception this early in startup, but since this is used exclusively for argv processing I think it's entirely reasonable for it to be fatal.

The docs say that illegal sequences are replaced with the replacement character, and do not mention any exceptions.


But also:  I would suggest eliminating winapi_wstr_to_utf8 and replacing it with a call to to_string(), though I would probably rename it and would certainly move it to some more appropriate file.  Since we're not 100% sure of the long-term plan here, let's funnel everything through one function that we control.

jordanbrown0 (changes requested):

src/platform/PlatformUtils-win.cc
nicvagn: > My understanding of `boost::nowide::narrow()` is that it does _not_ throw any exceptions. Have you made other experiences?

No, I have not. I do not have much C++ experience since uni, and it is a pattern I used without full understanding.  sorry.

src/platform/PlatformUtils-win.cc
```diff
@@ -26,40 +27,8 @@
...
-  if (result != numbytes) {
-    DWORD errcode = GetLastError();
-    LOG(message_group::Error, "Error converting w_char str to utf8 string");
-    LOG(message_group::Error, "error code %1$i", errcode);
-  }
-
-  return utf8_str;
-}
+// Use boost::nowide::narrow to convert wide strings
+std::string winapi_wstr_to_utf8(const std::wstring& wstr) { return boost::nowide::narrow(wstr); }
```
kintel: Nit: just kill winapi_wstr_to_utf8() and use narrow() directly.

nicvagn: Ok, I can do that.
As I am unfamiliar with the project, I want to touch as little as possible. 
I do not want to knock over the sandcastle!

nicvagn:
I am unfamiliar with the machinery of github review. I think I did what was asked, sorry.

src/platform/PlatformUtils-win.cc
```diff
@@ -246,7 +212,7 @@ int wmain(int argc, wchar_t **argv)
   std::string argvString[argc];
 
   for (int i = 0; i < argc; i++) {
-    argvString[i] = winapi_wstr_to_utf8(argv[i]);
+    argvString[i] = boost::nowide::narrow(argv[i]);
```
jordanbrown0, line 215: With the future of this conversion in doubt, I would have left this operation wrapped in our own function so that we could swap it out (e.g. to a future C++ standard function) in the future.  But that's a very small thing, and if it's OK with Marius it's certainly OK with me.

PR #6361, opened 2025-11-14
Discretization based on allowable error with $fe for circular arcs and document linear_extrude

### LINKED ISSUE #6366: Document why slicing non-uniformly-scaled untwisted linear_extrude is necessary

The code for this is over-complicated and creates unnecessary faces, slowing evaluation. It's also tech debt.

It was originally added in #3351 because [Non-uniform scaling scale.x != scale.y, No twist:
??? Currently this situation still defaults to 1 slice. And while the vertices follow linear paths, the edges between them can form curved surfaces if given enough slices.](https://github.com/openscad/openscad/pull/3351#issuecomment-643066866)

No one else talked about this and I think it was based on a misconception from a small view. Also the main diagonals are running vertically now instead of horizontally; I'm not sure that matters, but it could be a difference.

You can view the pictures below with `slices=1`, they appear "curved" just as much as when `slices=20`.

~~I want both untwisted uniform and non-uniform scaling to operate with slices=1.~~

Document why it needs to be this way. Below comments describe why.

**Additional context**

```
linear_extrude(20, scale=[2,0], slices=1) rotate([0,0,45]) square(60,center=true);
```

Image

With single slice, it appears curved.

Same for the taller one:

```
linear_extrude(60, scale=[2,0], slices=1) rotate([0,0,45]) square(20,center=true);
```

Image

```
linear_extrude(60, scale=[2,0], slices=20) rotate([0,0,45]) square(20,center=true);
```

Image

coryrc:
This introduces `$fe` to specify the most-allowable error when discretizing an arc. So circle, sphere, and cylinder. Not yet implemented for more advanced cases except linear_extrude when scaling is uniform without twist.

Progress for #4246

Fixes #6366

Added tests for the new functionality. Also added tests for the slices you can't see in an untwisted scaled linear_extrude by adding a test with view-edges enabled.


kintel:
Sorry for late review. I'm not to keen on merging this before the next release, until it's gotten more field testing. I wonder if this can be made experimental so that it gets distributed to dev snapshots until it's ready for release?

coryrc:
> Sorry for late review. I'm not to keen on merging this before the next release, until it's gotten more field testing. I wonder if this can be made experimental so that it gets distributed to dev snapshots until it's ready for release?

Are you fine with `$fe` in the namespace when not enabled (the only difference is it's `0.0` instead of `undef`) or must it be out entirely?

coryrc:
Hmm, actually `undef` is treated the same as `0.0` so I'm not sure it ever does need initializing...

jordanbrown0:
Guarding the "set" with a feature test doesn't seem onerous and, in conjunction with guarding the consumption, leaves the behavior 100% identical when the feature is disabled.

coryrc:
> Guarding the "set" with a feature test doesn't seem onerous and, in conjunction with guarding the consumption, leaves the behavior 100% identical when the feature is disabled.

Problem is it doesn't start being set when you turn on the feature via GUI, because Builtins are only initialized once.

But I'm going to leave it undefined (at least for the time being) because it's functionally equivalent to the value I had set.

coryrc:
I've made a copy of render, render-manifold, and dump regression test outputs and `add_cmdline_test` setup for them with this feature enabled, then restored all the existing test values.

Also just got a stick of 32GB for my laptop and it's a lot nicer to be compiling at 100% CPU instead of 12.5%.

coryrc:
I think I'm going to need to break out my test updates to a different PR.

kintel:
> I think I'm going to need to break out my test updates to a different PR.

I think a good starting point would be to write a few simple tests exercising this new feature, and run them with the feature enabled, and leave all the other tests alone.

I don't think it's strictly necessary to make copies of existing tests etc.; that's more interesting once we move this out of experimental. You may want to sanity-check it locally of course, just in case you missed something in the targeted tests.

coryrc:
> > I think I'm going to need to break out my test updates to a different PR.
> 
> I think a good starting point would be to write a few simple tests exercising this new feature, and run them with the feature enabled, and leave all the other tests alone.

Got it, I broke those changes out into https://github.com/openscad/openscad/pull/6559 so it's easier to audit the intended no-change.

and put some nice commenting/refactoring for linear_extrude into https://github.com/openscad/openscad/pull/6557

coryrc:
This should be ready for review now.

I used the Feature lock to prevent setting the internal value of `fe` to anything other than zero. All code assumes zero means to not apply discretization by error, so I believe that is sufficient change, rather than complicating the code which calculates segment count with Feature:: checks. The tests also confirm the no-change state when the feature isn't enabled (they're separately visible at #6559).

kintel:
Yeah, sounds good - the features is to guard user-visible features, not internally unused ones :) Will put this all on my review TODO list

src/Feature.cc
```diff
@@ -44,6 +44,9 @@ const Feature Feature::ExperimentalPredictibleOutput(
   "order)");
 const Feature Feature::ExperimentalVectorSwizzle(
   "vector-swizzle", "Enable vector swizzling (e.g. <code>vec4.zyx</code> to reverse a 3D vector).");
+const Feature Feature::ExperimentalDiscretizationByError(
+  "discretization-by-error",
+  "Specify the maximum error in $fe and shapes will be segmented appropraitely.");
```
kintel: appropriately

src/core/CurveDiscretizer.h
```diff
@@ -43,21 +43,37 @@ class CurveDiscretizer
...
+   * @param r_sqr Largest 2D delta from origin of all vertices, squared.
+   * @param h Height of extrusion.
+   * @param twist_degrees Total degrees the extrusion will twist while extruding.
    */
   std::optional<int> getHelixSlices(double r_sqr, double h, double twist_degrees) const;
 
+  /**
+   * Returns the number of slices for a linear_extrude with twist when scale is uniform.
+   * FIXME: Should return similar results to getHelixSlices when scale==1? Is this more efficient? Why
+   * does this exist and/or why does getHelixSlices exist separately otherwise?
```
kintel, line 56: It sounds like this is a (small) unresolved question. The way we usually deal with this is to create new GitHub issue "Move $fe out of experimental" and write a checklist of things we need to resolve. That way we can simply check off all the boxes and promote it into a core feature when done.

src/core/CurveDiscretizer.cc
```diff
@@ -102,9 +162,92 @@ static double helix_arc_length(double r_sqr, double height, double twist_degrees
...
+  double theta = 2 * (M_PI - acos(fe / r - 1));
+  int fe_slices = static_cast<int>(std::ceil(twist_degrees * M_DEG2RAD / theta));
+  // min_slices already handles the requirement to not exceed 120 degrees per step.
+  return std::max(fe_slices, min_slices);
+}
+
 std::optional<int> CurveDiscretizer::getHelixSlices(double r_sqr, double height,
                                                     double twist_degrees) const
 {
+  // std::cerr << "getHelixSlices("<<r_sqr<<","<<height<<","<<twist_degrees<<")\n";
```
kintel: If this is useful debug info, it can be turned into a PRINTDB(), otherwise, it can probably be removed.

src/core/CurveDiscretizer.cc
```diff
@@ -102,9 +162,92 @@ static double helix_arc_length(double r_sqr, double height, double twist_degrees
...
+  // mp_ds: (sqrt(2*cos(θ)+2)*r)/2
+  // err1: r-mp_d;
+  // assume(r>0);
+  // s3: solve(fe=err1,r);
+  // s4: trigsimp(s3[1]);
+  // s5: solve(s4,θ);
+  // s5_s: trigsimp(s5[1]);
+  // Which yields θ=acos((r^2-4*fe*r+2*fe^2)/r^2)
+  // Fortunately that graphs to the same thing as θ=2(pi-acos(fe/r-1)) over the domain of interest (r>fe,
+  // 0<theta<120deg). So use the simpler equation.
```
kintel, line 239: This is a bit of an intense comment. Fine to keep while we figure things out. Good think to add to a "move out of experimental" ticket perhaps?

src/core/Builtins.cc
```diff
@@ -113,6 +113,10 @@ std::string Builtins::isDeprecated(const std::string& name) const
 Builtins::Builtins()
 {
   this->assignments.emplace_back(new Assignment("$fn", std::make_shared<Literal>(0.0)));
+  // $fe doesn't need initializing because `undef` is treated identical to 0.0,
+  // but it could be used for feature detection when complete,
+  // but does have the problem that these are initialized only once in `openscad_main`,
+  // so enabling the feature wouldn't initialize the variable until the program is restarted.
```
kintel, line 119: Perhaps out of scope, but this is a good indicator that we shouldn't initialize stuff at program startup, but rather when a script is evaluated?

tests/data/scad/3D/features/sphere-tests.scad
```diff
@@ -11,3 +11,5 @@ translate([22, 11, 0]) sphere(5, $fa=40, $fs=0.3);
 translate([11, 22, 0]) sphere(5, $fn=0.1);
 translate([33,  0, 0]) sphere(d=10);
 translate([33, 11, 0]) sphere(r=1, d=10);
+translate([33, -11, 0]) sphere($fe=0.05, d=10, $fa=0.5, $fs=0.5); // Will be shiny smooth if $fe doesn't override $fa/$fs
+translate([22, 22, 0]) sphere($fe=0.001, d=10, $fn=6); // Will be shiny smooth if $fn doesn't override $fe
```
kintel, line 15: Another thing to add to a "move $fe out of experimental": finalize priority of $fe vs. $fa/$fs

src/core/CurveDiscretizer.cc
```diff
@@ -75,13 +98,50 @@ std::optional<int> CurveDiscretizer::getCircularSegmentCount(double r, double an
...
   } else {
-    result = std::ceil(std::max(std::min(360.0 / fa, r * 2 * M_PI / fs), 5.0)) *
-             std::fabs(angle_degrees) / 360.0;
+    if (std::isinf(fe) || std::isnan(fe)) return {};
+    // $fe measures "the most allowed error from the platonic circle to the discretized circle".
+    // Which is the distance along a radius from the circumference to the midpoint of an edge
+    // of the inscribed regular polygon we create when discretizing a circle.
+    // $fe respects the minimums for $fa and $fs, but ignores their dynamic values.
+    // aka specifying $fe means $fa/$fs are not used.
+    if (fe >= GRID_FINE) {
```
kintel, line 114: Out of scope, but: I feel GRID_FINE is an outdated concept and we should probably clean that up.

src/core/Builtins.cc
coryrc: https://github.com/openscad/openscad/issues/6442 and https://github.com/openscad/openscad/issues/6451 would both benefit from that

src/Feature.cc
coryrc: Done.

src/core/CurveDiscretizer.cc
coryrc: Is there a different epsilon you'd like to see? Obviously I'm just picking something close-but-not-entirely-zero.

src/core/CurveDiscretizer.cc
coryrc: It's a bit of literate programming. If anybody wanted to edit this in the future or do something similar, this can be dropped into Maxima as a head start.

src/core/CurveDiscretizer.cc
coryrc: Mostly not, just missed cleaning it up. Done.

src/core/CurveDiscretizer.h
coryrc: This is orthogonal to $fe, it came to mind while reviewing this code. If either I figure out why it's this way or someone else does, it'd be nice to document, or maybe even something could be simplified.

kintel: Gotcha, in that case we're good

src/core/CurveDiscretizer.cc
kintel: Yeah, but once someone changes the code without updating this comment it starts having negative value

kintel: Edit: keeping the math makes sense - it just felt like way too much text

src/core/CurveDiscretizer.cc
kintel: Not really, GRID_FINE was something used as CGAL preprocessing to limit small details, but ended up being a liability as it also occasionally messed with topology..

kintel:
Hmm, is the CircleCI build failing due to some caching issue? @flaviut what do you think? This PR was opened before the switch to Qt6, but it may be using caches which weren't wiped correctly?

flaviut:
I suspect the issue is not a caching issue, but I'll take a look tomorrow night. The reason I don't think it's a caching issue is that while Cmake does compile small test programs (which might hit the cache), for the Qt5 dependency it is not even getting to that stage, it is complaining about missing CMake configuration files.

kintel:
I haven't looked in detail either. It's kind of like it's caching CMakeCache.txt from a previous run, and didn't realize that it wants to pick up Qt6 now.

tests/data/scad/3D/features/sphere-tests.scad
coryrc: Oops, I forgot to reply to this. This is how I intend it to work. The point of the message was for future people (which was also me) to look at the test report and approve the change, or if they make a change and see it be smooth, then they know exactly where to look.

src/core/CurveDiscretizer.cc
coryrc: > Yeah, but once someone changes the code without updating this comment it starts having negative value

Don't let them?

coryrc:
This is missing [Flip default for USE_QT6 to ON for all platforms.](https://github.com/coryrc/openscad/commit/13de4908b3ee6aeff209e247235de86a9ebab57b)

Were the Qt5 libraries removed from the MXE image? [Yes](https://github.com/openscad/docker-openscad/commit/ff01762386969834952dbc3735e555c48692ef6a#diff-5abc92d745994fc4fcbf6ea031724643718900f066b0ea9989608548ac7fd07c)

so it's not surprising it is borked. Will update branch.

tests/data/scad/3D/features/sphere-tests.scad
kintel: Yeah, I think this is good. Gotta document it properly, but that's a bridge we can cross a bit later.

kintel:
> This is missing [Flip default for USE_QT6 to ON for all platforms.]

Ah, thanks, I missed that! Sorry for the noise

src/core/CurveDiscretizer.cc
kintel: Haha, it's guaranteed going to slip through the review net. otoh, perhaps the code doesn't need to change :) We'll do another round once we get initial feedback on the feature through dev snapshots.

coryrc:
> > This is missing [Flip default for USE_QT6 to ON for all platforms.]
> 
> Ah, thanks, I missed that! Sorry for the noise

No worries, you'll probably run into this again many times and now you'll know exactly what to look for!

PR #6407, opened 2025-11-29
linear_extrude more efficient copy, breakout functionality, more unit tests

coryrc:
Can use the internals in units tests.

Prove (well, "prove") the more-efficient replacement results in the same geometry with [the first commit](https://github.com/openscad/openscad/pull/6407/commits/a34d191368ff0e13ec1e858141450f952183cc57), you can look at Checks for it to see they all passed, including my temporary unit test comparing the two copy methods.

Since then, I removed the duplicate code and left a couple basic unit tests.


coryrc:
Huh, that's funny. It fails to compile the `PMP::volume()` call on ubuntu-24, which is using CGAL version 5.6-1build3, but it works both on ubuntu-22 with 5.4-1 and on my machine with 6.0.1-1. And there's so much templating it's not at all obvious what the problem is.

coryrc:
https://github.com/CGAL/cgal/commit/f34745c4d87c94f954fe1e6b7a3a1b3dbdf85e62 appears to have fixed it.

There's a workaround:

```
#include <CGAL/Gmpq.h>

// Define a pass-through exact() function for Gmpq
namespace CGAL {
    inline const CGAL::Gmpq& exact(const CGAL::Gmpq& d) {
        return d;
    }
}

// Now include the PMP header
#include <CGAL/Polygon_mesh_processing/measure.h>
```

Let's see if there's a VERSION header to scope this.

src/geometry/linear_extrude_test.cc
```diff
@@ -0,0 +1,179 @@
...
+
+Polygon2d makeCross(double s)
+{
+  Outline2d o;
+  o.vertices = {{0, 0}, {s / 2, s / 4},     {s, 0}, {s * 3 / 4, s / 2},
+                {s, s}, {s / 2, s * 3 / 4}, {0, s}, {s / 4, s / 2}};
+  return std::move(Polygon2d(o));
+}
+
+TEST_CASE("sgn_vdiff basic functionality")
```
kintel, line 82: It's hard to map these test names to test output on the CI. If there isn't a way to convince CTest to output test names with more info, perhaps we should establish a convention of naming all tests with something that allows us to quickly find them, e.g. `linear_extrude_test: sgn_vdiff basic functionality` or smth.?

coryrc: `rgrep "sgn_vdiff basic functionality"` finds it, I don't think it's necessary to pretend grep doesn't exist.

However it's also builtin if desired: https://catch2-temp.readthedocs.io/en/latest/command-line.html#filenames-as-tags

kintel: Yeah, but to find the tests related to this module, you have to read the code and grep for each test; there isn't a way to run, or even identify, just those tests without inspecting each one individually. I know catch2 has features for this, but ctest doesn't expose that.

coryrc: https://github.com/openscad/openscad/blob/80f2c761f3353cc68e29acb8f9e034ace0862e02/doc/testing.md?plain=1#L55-L62

kintel: Ah, right, I remember that now. I just have ctest on my fingertips.
I'll think about how to make it less confusing in the future :)

src/geometry/linear_extrude_test.cc
```diff
@@ -0,0 +1,182 @@
...
+}  // namespace LinearExtrudeInternals
+
+Polygon2d makeSquare(double s)
+{
+  Vector2d v1(0, 0);
+  Vector2d v2(s, s);
+
+  Outline2d o;
+  o.vertices = {v1, {v2[0], v1[1]}, v2, {v1[0], v2[1]}};
+  return std::move(Polygon2d(o));
```
kintel: nit: move is no-op in this case (object is already an rvalue)

src/geometry/linear_extrude_test.cc
```diff
@@ -0,0 +1,182 @@
...
+}
+
+Polygon2d makeCross(double s)
+{
+  Outline2d o;
+  o.vertices = {{0, 0}, {s / 2, s / 4},     {s, 0}, {s * 3 / 4, s / 2},
+                {s, s}, {s / 2, s * 3 / 4}, {0, s}, {s / 4, s / 2}};
+  return std::move(Polygon2d(o));
```
kintel: remove move

src/geometry/linear_extrude_test.cc
```diff
@@ -0,0 +1,182 @@
...
+    // This works too and has some repair mechanisms for bad inputs, which is not a problem here.
+    // auto nef_shape = *CGALUtils::createNefPolyhedronFromPolySet(*cps)->p3;
+    // CGAL::Surface_mesh<CGAL_Point_3> mesh;
+    // CGAL::convert_nef_polyhedron_to_polygon_mesh(
+    //   nef_shape,
+    //   mesh,
+    //   true /* triangulate_all_faces = true */
+    //   );
+    // namespace PMP = CGAL::Polygon_mesh_processing;
+    // PMP::orient(mesh);
```
kintel: We can probably kill this comment. You're using CGAL for volume calculation and there are many ways of arriving there on a large API surface like CGAL>

kintel:
Approving, but giving you some buffer in case you want to address my nits.

src/geometry/linear_extrude_test.cc
coryrc: Done

src/geometry/linear_extrude_test.cc
coryrc: Done

src/geometry/linear_extrude_test.cc
coryrc: Done

PR #6408, opened 2025-11-29
Basic full-precision support (affects AST only)

### LINKED ISSUE #6367: User-specified numbers displayed via Display AST are being rounded
**Describe the bug**
Numbers specified in scripts are being rounded to 6 significant [decimal] digits when being displayed via **Display AST** (the user-specified values are correctly stored internally though).  The numbers used internally should instead be displayed at full precision.

**To Reproduce**
1. Enter the following script:
```
a=1234567890.0123456789;
b=1234567890;
c=0.0123456789;
d=0.00123456789;
e=0.000000000123456789;
f=1234.56789;
```
2. Generate the AST by pressing **`F5`**.
3. Click on **Design** → **Display AST…**    The following is displayed:
```
//Parameter("")
a = 1.23457e+9;
//Parameter("")
b = 1.23457e+9;
//Parameter("")
c = 0.0123457;
//Parameter("")
d = 0.00123457;
//Parameter("")
e = 1.23457e-10;
//Parameter("")
f = 1234.57;
```

**Expected behavior**
The values displayed should not be rounded (they should reflect the exact values used internally).  The AST should look something like the following:
```
//Parameter("")
a = 1234567890.0123458;
//Parameter("")
b = 1234567890;
//Parameter("")
c = 0.0123456789;
//Parameter("")
d = 0.0012345678899999999;
//Parameter("")
e = 1.2345678900000001e-10;
//Parameter("")
f = 1234.56789;
```

[Values shown here were generated via C++ using `std::cout << std::setprecision (17) << N << std::endl;`]

See above.

 - OS: Ubuntu 22.04
 - System: x86_64
 - OpenSCAD Version 2025.11.16, built from Git [[9ed115a](https://github.com/openscad/openscad/commit/9ed115a534e7ff381da9f789b08b6740ac614ea2)]

**Additional context**
 
* This does not appear to be a regression as the same occurs with version 2021.01.
* I experimentally verified that the internal values are correct (not unduly-rounded).
* For 64-bit IEEE 754 (`double`) outputting decimal with 17 significant digits is sufficient to fully-represent the internal state (i.e. the values remain the same when round-tripped).

Alex-Kent:
This patch changes the number output precision for AST display (only) from 6 significant digits to 17 significant digits.  This precision is sufficient to losslessly represent any value stored in the IEEE 754 64-bit floating point format used by OpenSCAD to store the numeric values of numbers stored in Value objects.  The rationale for making this change is so that the AST output shown to the user accurately reflects the actual internal state of the AST (as discussed in issue #6367).

**Note:** This PR is meant primarily as a demonstration of the basic logic of how the precision is distributed through the various code paths.  This scope-limited version of the code should probably not be merged.  The [full-featured version [#6419]](https://github.com/openscad/openscad/pull/6419) of the code should be merged instead.

# Tests

The changes were tested both via the GUI and via the CLI.  The only change in behavior is how numbers in AST dumps are presented (all using 17 significant digits, the desired result); no other [read: unintended] changes were observed.

A [test script](https://github.com/user-attachments/files/23836255/ast-float-precision-6367.scad.txt) (rename to `ast-float-precision-6367.scad`) is supplied.  The results of running it are:

* [`ast-float-precision-6367.ast.2025.11.16.txt`](https://github.com/user-attachments/files/23836232/ast-float-precision-6367.ast.2025.11.16.txt)
  AST output without the patch applied.
* [`ast-float-precision-6367.ast.patch-6367.txt`](https://github.com/user-attachments/files/23836234/ast-float-precision-6367.ast.patch-6367.txt)
  AST output with the patch applied.
* [`ast-float-precision-6367.ast.diff.txt`](https://github.com/user-attachments/files/23836249/ast-float-precision-6367.ast.diff.txt)
  The difference in the AST outputs.
* [`ast-float-precision-6367.output.2025.11.16.txt`](https://github.com/user-attachments/files/23836250/ast-float-precision-6367.output.2025.11.16.txt)
  Output when running script without the patch applied.
* [`ast-float-precision-6367.output.patch-6367.txt`](https://github.com/user-attachments/files/23836251/ast-float-precision-6367.output.patch-6367.txt)
  Output when running script with the patch applied.
* [`ast-float-precision-6367.output.diff.txt`](https://github.com/user-attachments/files/23836252/ast-float-precision-6367.output.diff.txt)
  The difference in the outputs when running the script.
  (only difference is a trivial difference in the run time)

# Additional notes

This patch only affects AST output; a larger patch that allows the output precision to be set globally by the user (and changed at arbitrarily points within a running script) has been coded and can be found in #6419.

Resolves #6367


Alex-Kent:
Updated expected test results have been added.  These use full-precision (17 significant digit) output as there is no way (in this version of the code) to specify a numeric lower precision for the AST output.  The full-featured code includes the ability to specify a lower precision (e.g. 6 digits, as currently used by OpenSCAD) so with that these tests could be run with a lower output precision (i.e. the current default of 6).

jordanbrown0:
My instinctive reaction is that a program should not be allowed to control the form of the AST dump, and that allowing the user to control it wouldn't be worth the incremental UI complexity.

Alex-Kent:
> a program should not be allowed to control the form of the AST dump

In this PR the precision shown in the AST is always 17 digits (previously it was 6).

With the full version programs have no control over the format of the AST dump, that [i.e. the precision shown in the AST] is entirely up to the user, controlled either by a GUI preference or a command line option (for GUI and CLI invocation, respectively).

jordanbrown0:
> entirely up to the user

I think that level of control costs more in UI complexity (more than zero) than it gains in benefit (nearly zero).

Alex-Kent:
Full-featured version is in PR #6419.

kintel:
I finally managed to look at this. Sorry for the delay, too many things going on and too few people...
Two thoughts:
1. The intended purpose of AST output is to, ideally, output exactly what was fed into OpenSCAD. This is only really possible if we capture the actual input text in the AST, so it's not realistic to fix without significant refactoring. The big question is whether this PR helps. I kind of think it does, but I'm not aware of any existing AST-based workflows.
2. Another very common workflow is to export .csg files (which are literal CSG trees). Especially for debugging, it may be crucial that numeric precision is maintained. The caveat is that this format is also used internally to generate cache keys. More precision in cache keys should make the caches more robust, at the (hopefully) small expense in compute and storage of the keys. This PR doesn't influence .csg files, but perhaps it should?

PR #6412, opened 2025-11-30
CircleCI release builds

kintel:
Add release builds to CircleCI, triggered by tagging with a valid release tag (`openscad-YYYY.MM.DD-RRR`).

Adding a tag will trigger the release build process:
1. Determine OpenSCAD version
2. Build a tarball for the given version
3. Build release builds for all platform from the tarball (without git connectivity)
4. Upload artifacts

TODO:
* [x] Decide if we should let each job write artifacts or write them all in the `openscad-deploy-release` job
   * -> As a start it's better to let each job write artifacts to align with the current snapshot behavior
* [x] Decide what to do if one of the platforms fail for some reason (transient infrastructure failure vs. code error)
   * -> This is only important if we manage all binaries using a collection job like `openscad-deploy-release`. We don't use a collection job, but if we do, we should be able to use CircleCI's rerun mechanism.
* [x] Decide if we want to be able to re-trigger a single job and get the resulting binary collected.
   * -> May be important to deal with transient infrastructure issues/timeouts. We can use regular CircleCI retries.
* [x] Decide how/if to re-trigger the pipeline in case of unexpected errors (infra errors or code issues)
   * For full retrigger, we could delete and re-apply the tag. We could also use the API, but need to test that it works. Note: We can abandon the release in this case and increment the RC/patch number, albeit it feels suboptimal.
   * -> Let's cross this bridge later. A third option: Don't tag something releasable until we have a verified test build.

What to do from here is TBD. Alternatives include:
* Download release builds to files.openscad.org by slightly amending the corresponding scripts to look for release builds (CircleCI parameters and/or tag triggers)
* Automatically create a GitHub release and pushing tarball and binaries there.


flaviut:
> * [ ]  Decide what to do if one of the platforms fail for some reason (transient infrastructure failure vs. code error)
>   
>   * -> This is only important if we manage all binaries using a collection job like `openscad-deploy-release`
> 
> * [ ]  Decide if we want to be able to re-trigger a single job and get the resulting binary collected.
>   
>   * -> May be important to deal with transient infrastructure issues/timeouts.

For both of these situations, CircleCI provides a "retry pipeline from failed" in the UI whenever there is a failing job:

image


> * [ ]  Decide how/if to re-trigger the pipeline.
>   
>   * -> For full retrigger, we could delete and re-apply the tag. We could also use the API, but need to test that it works.

I would suggest this is not necessary and would be a bad idea. If there was a bug in the build process or OpenSCAD itself, the right course of action in my opinion would be to increment the patch or date, and cut a new release. It costs $0 after all to increment an integer.

If everything was completely pinned, then re-running the pipelines with the same code should produce the same result, so there is no reason to ever re-run a successful pipeline. This only case where I would see this coming up is if the release process did not have all dependencies completely pinned. In which case I'd consider that a bug, and deserving of a patch release.

kintel:
]> >   * -> For full retrigger, we could delete and re-apply the tag. We could also use the API, but need to test that it works.
> 
> I would suggest this is not necessary and would be a bad idea. If there was a bug in the build process or OpenSCAD itself, the right course of action in my opinion would be to increment the patch or date, and cut a new release. It costs $0 after all to increment an integer.
> 
I was mostly thinking about re-triggering failed pipelines, but there may be a reasonable case for retriggering a successful pipeline too, if something is missing. The cost of publishing a new RC release is not free, as it would typically involve writing a changelog and sending some email/social releases. Release don't necessarily have to be increased by 1, so we could get away with multiple increases between each RC launch, but it would feel a bit odd to me.

flaviut:
If I was doing it, I'd personally be writing the emails and change logs after the CI pipelines have succeeded and artifacts uploaded to the file server, in which case there's no extra work to be done.

I acknowledge I'm a lot more willing than most to do something that just "feels a bit odd" in pursuit of doing less work :)

kintel:
For RC releases, it matters less, as the number doesn't really have a meaning. For actual patchlevel releases, I don't 100% trust us to get it right at the first tag.

Issue #6385, opened 2025-11-21
New output/formatting suite:  print(), quote(), format()


jordanbrown0:
**Is your feature request related to a problem? Please describe.**
* `echo()` always emits "decorations" around its output:  `ECHO: ` and quotes around strings.  However, those strings are not themselves guaranteed to be legal OpenSCAD strings; any internal special characters (quote, newline, backslash, et cetera) are not quoted.  This combination makes it suboptimal for emitting text, but also suboptimal for emitting OpenSCAD syntax.
* `str()` has similar issues:  when presented with an aggregate data value (list, object) it emits something that looks like OpenSCAD syntax but does not have internal special characters quoted.  In addition, if the goal is to emit valid OpenSCAD source text, it fails on simple strings because it does not emit quotes around them.
* There is no way to control the formatting of output - especially to specify the number of digits to put after the decimal point of a number, but also to specify when and whether to transition to exponential notation, how to pad output, and how to truncate output.

**Describe the solution you'd like**
* The module `print()` would emit scalar arguments unadorned, except that multiple arguments would be separated by spaces.  (TBD what it does with lists, objects, functions, et cetera.)  It would thus be available to emit pure unadorned text, entirely under the program's control.
* The function `quote()` would convert its arguments into a string that is a valid OpenSCAD-language representation of their values.  (Perhaps it would also be restricted to being a valid JSON representation - in particular, by quoting the member names in objects - or perhaps that would be a separate `json()` function.)  It would be available, notably in conjunction with print(), to produce text representations of data values that could then be used as input to OpenSCAD.
* The function `format()` would accept a format string and a list of arguments, and combine them to form a string, analogous to C's `sprintf()`, Python's `str.format()` and the `{fmt}` library.  It could be used to control numeric presentation in `text()` calls, or construct formatted strings for `text()` or `print()`.

**Describe alternatives you've considered**
* Python-style formatted string literals are a possible replacement for `format()`.
* `str()` could gain number-formatting options, and could then be used (with a large amount of DIY) to achieve the same goals as `format()`.
* We could abandon backward compatibility and modify `echo()` and/or `str()` to address some of these issues.  Note that because of the semantics of `echo()` we could not control its behavior using arguments.

**Additional context**
* These three features are entirely separable.  However, they work well together and so it seems best to discuss them together.
* I do not yet propose specific behavior - in particular I do not attempt to specify the format-string language used by format().  The reason for this feature request is to spur discussion on the subject.

[...]

PR #6419, opened 2025-12-02
Adds floating point precision specification (including default precision) to `echo`, `str`, and AST output

### LINKED ISSUE #6367: User-specified numbers displayed via Display AST are being rounded

Alex-Kent:
This PR allows the precision used when displaying numbers to be set both by the user and via scripts.  This includes values output via `echo`, those constructed using `str`, as well as those in AST dumps.

# Rationale

While writing OpenSCAD scripts I found the need to examine the full internal values of variables (i.e. via viewing the AST), reported in issue #6367.  This resulted in the PR I submitted earlier (#6408).  Having coded that PR I realized that it would not take much additional code to extend the ability to select floating precision to code being evaluated as well, hence this PR.

# Usage

Precisions are specified as the number of significant digits to display.  Valid precisions are integers 0 through 17, inclusive, with 0 indicating that OpenSCAD's default precision (currently 6) should be used.  (17 significant digits is sufficient to losslessly represent an IEEE 754 64-bit float, such are used internally by OpenSCAD to store all numbers.)

## Within scripts

The precision used while evaluating a script can be set using the `$fp` variable at either the block level or at the expression level.  Briefly, the following syntaxes can be used (further examples can be found in [tests/data/scad/misc/full-precision-test.scad](https://github.com/Alex-Kent/openscad/blob/full-arbitrary-output-precision-support/tests/data/scad/misc/full-precision-test.scad)):

* Set at the top level of the script
```
$fp=10; // All code uses precision 10 unless overridden
```
* Within a block
```
union() {
  $fp=10; // All code in block uses precision 10 unless overridden
  …
}
```
* Using a `let` prefix
  This can be applied in many (but not all) situations to set the locally-used precision.
```
a = let($fp=2) str(1.2345);
echo(a=a);                    // ECHO: a = "1.2"
echo(str(a,a));               // ECHO: "1.21.2"

let($fp=3) echo(1.2345);      // ECHO: 1.23

function f(x) = str(x,"×",x,"=",x*x);
b = let($fp=3) f(1.2345);
echo(b=b);                    // ECHO: b = "1.23×1.23=1.52"
echo( let($fp=4) f(1.2345) ); // ECHO: "1.234×1.234=1.524"

function g(x) = let($fp=1) str(x,"×",x,"=",x*x);
c = g(1.2345);
echo(c=c);                    // ECHO: c = "1×1=2"
```
When evaluating, the innermost `$fp` value takes precedence.

Note that the precision used for outputting the AST can only be controlled by the user (see below), it cannot be set from within a script.

## When running via the GUI

* The default precision used when running scripts is set via **Edit** → **Preferences** → **Advanced** → **Displayed number precision** → **Global**.
  This is stored under `advanced/precisionGlobal` in the preferences file.

* The precision used when displaying the AST (via **Design** → **Display AST…**) is set via **Edit** → **Preferences** → **Advanced** → **Displayed number precision** → **AST**.
  This is stored under `advanced/precisionAST` in the preferences file.

## When running via the CLI

Both the default precision used when running scripts and the precision used when outputting the AST are set using a single command line option `-O advanced/numberOutputPrecision=N` where `N` is the desired precision (1 through 17, inclusive).

# Theory of operation

The overall purpose for this code is to convey user-specified precision (from GUI preferences, the command line, or `$fp`) to the number-to-string conversion function (`DoubleConvert`).

The precision is applied differently for AST dumps and for normal evaluation:

* For AST the desired precision passed directly to `ASTNode::dump()`.

  The specified precision will remain unchanged throughout the AST dump operation.

* When evaluating code the user-specified default precision (or 0 if none was specified) is assigned when evaluation starts to the variable `$fp` in the top-level `Context` (`BuiltinContext`) using `builtin_context->set_variable("$fp", precision);` (in `do_export(…)` and `MainWindow::instantiateRoot()`).  Note that `$fp=…;` is not simply appended to the script (as the `-D` command line option does) since doing so would prevent overriding the top-level `$fp` value from within a script.

  Within any context the active precision can be changed by setting the `$fp` variable (stored in the block's `Context`).

In both cases the precision propagates through the code using the same mechanisms.

There is no single mechanism that can be used throughout the code for propagating the precision and keeping track of the active precision.  Instead, depending on the situation, one of several methods of storage/propagation are used:

* **In a `$fp` variable in a block's `Context`**
  This only applies when a script is being run.  This is also the only way that a script can alter the active precision.  Being in a block's `Context` means that all function calls in the block (and in any child blocks, unless `$fp` is reassigned therein) see the same desired precision.
* **As an attribute on a stream being written to**
  This method is used when the program's logic flow is via an output to a stream (`stream << something;`).  The desired precision is applied to the stream as it's `precision` [a standard C++ stream attribute] (`stream.precision(precision);`); retrieval is done with `stream.precision()`.

  This is the method used when dumping the AST and is also the method used internally when calling `toString()` for complex types (vectors, objects, ranges, and function pointers).

  The advantage to passing the precision via the stream in these contexts is that 1) it uses a standard mechanism and 2) it requires no modification to the OpenSCAD code beyond simple set and get statements.  The downside though is that it cannot be extended in future to pass the full `Context` (rewriting to pass that using a different mechanism would be required).

* **Stored within a `Value` object**
  As each expression is evaluated the active precision is applied to each `Value` encountered.  When the evaluated code is actually run (i.e. when the `toString()` methods are called) the precision to use is retrieved from the `Value` object that is being output.

* **Explicitly passed via a method/function argument**
  This is only used for the penultimate step in generating a string from a number.  The `double` value to be converted and the precision to use are passed as arguments to the `DoubleConvert` function (which returns a `std::string`).  This function then calls the `double_conversion::DoubleToStringConverter.ToPrecision(…)` method  with the supplied precision (in lieu of `DC_PRECISION_REQUESTED`, the OpenSCAD default value [6]) to actually generate the string.

# Tests

Automated tests have been included for both normal and AST output.  The test script for both is [tests/data/scad/misc/full-precision-test.scad](https://github.com/Alex-Kent/openscad/blob/full-arbitrary-output-precision-support/tests/data/scad/misc/full-precision-test.scad).

# Additional comments

A similar mechanism could be used to pass not just the precision but additional contextual information as well (e.g. by passing the full `Context` instead of the precision).  This could be pretty much a drop-in replacement in most places, however in the `toString()` call the value is being passed via the output stream's precision attribute, a technique not amenable to passing arbitrary data.  Having access to the full `Context` throughout the code might be very useful in regards to implementing the `format`/`[s]printf` idea [#6385].

---

Supersedes #6408
Resolves #6367

### DIFF
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -399,7 +399,7 @@ int do_export(const CommandLine& cmd, const RenderVariables& render_variables, F
   ContextHandle<BuiltinContext> builtin_context{Context::create<BuiltinContext>(&session)};
   render_variables.applyToContext(builtin_context);
 
-  // Pick up optional numeric output precision used both when running scripts 
+  // Pick up optional numeric output precision used both when running scripts
   // and when displaying their AST.  This is specifiable from the CLI with
   //   -O advanced/numberOutputPrecision=N
   // where N is the desired precision (1 through 17, inclusive).


t-paul:
I'm very much against adding this. A dedicated much more useful formatting feature is very high on the list which makes this redundant while adding a huge maintenance burden.

Alex-Kent:
@t-paul
I did address this under **Additional comments**.  The issues discussed in this PR are applicable beyond just specifying output precision.  Unless I've missed something (which is distinctly possible) there's no single mechanism available for passing node context (e.g. formatting information) throughout the code.  At higher levels such can be passed via `Context` but when it comes time to actually execute the code that mechanism is no longer available, _cf._ **Theory of operation**.  Even if this specific implementation (i.e. setting solely the precision using `$fp`) doesn't warrant inclusion I feel that the issues encountered are worth examining and discussing.

For precision specifically, it probably would be better to either store the desired precision within the `Context` or have it be a parameter to a `format`/`[s]printf` call; doing so is beyond the scope of this PR.

As for the advanced formatting feature that is being discussed, the state propagation question is equally applicable.  It may be desirable to be able to have, say, `sprintf("%.1f", [1.23, 4.56])` return something like "[1.2, 4.6]"; doing so requires a mechanism for propagating the format parameters to the nested nodes (my code only covers precision, a full formatting implementation would require passing more).  Absent this there's no obvious way to apply formatting to arbitrary [read: complex/nested] structures.

It's also worth noting that `ASTNode::dump()` is implemented using a single call to `ASTNode::print(…)` with the actual walking of the tree being handled internally using recursive `toString()`, etc. calls with no `Context` being available.  As such examining the mechanisms used by those calls to pass and store higher-level state (e.g. formatting data) are apropos.  (As suggested above, perhaps this could be store within `Context`; in the case of `ASTNode::dump()` a `Context` could be provided to it that includes the desired output format.)

That's my understanding from reading the code though I may well have misunderstood it.

Thoughts?

t-paul:
That a lot of things I would not mix.
* Extending the AST: Makes sense, it's a technical thing that probably should not do rounding, although I'm wondering why all those *.0000000001 are needed, I thought DoubleConversion can cut that down in most cases while still providing the guarantee of restoring the exact same binary representation on reading the value back. I would not even mind if AST export allows for a setting to specify the precision as export parameter, although I'm not sure it's worth the effort
* Adding more context (not necessarily just looking at the class with that name, but also evaluation session) to the whole evaluation logic would be useful. Especially the broken console output would hugely benefit from that, but that's all internal, not related to the user level scripts
* Why do we need that $fp? What's the actual use case to put that huge cannon ball on that (to me so far) very tiny problem.
* Using $fp would define precision also for library calls, I'm not sure that's helpful.
* We do need actual formatting (not just for numbers) as that's often requested, so hopefully it's coming at some point. I feels it would be much more of general use to dust that PR off. Assuming that would already exist, what would be the remaining use cases for $fp that are not covered except the magic behavior which is not necessarily always good?

jordanbrown0:
I agree that general formatting is the more useful feature here, and that a global-ish parameter is of very limited value.

Do *any* other languages have global-ish controls on how numbers are formatted?

> I would not even mind if AST export allows for a setting to specify the precision as export parameter, although I'm not sure it's worth the effort

The cost that interests me is not the development effort; it appears that @Alex-Kent is willing to spend that.

The cost that interests me is the incremental increase in the complexity of the preferences UI.  It's a very small cost, but the journey to a thousand parameters is taken one parameter at a time.  I'd say that the value is lower than that cost.

(Which also applies to the proposed $fp.)

t-paul:
I'd say lets get the other AST-only PR going, I think that one is pretty uncontroversial. I'll try to find some time to have a closer look...

Alex-Kent:
> the other AST-only PR

Which one are you referring to?

t-paul:
Well, #6408, are there more that I missed?

Alex-Kent:
> incremental increase in the complexity of the preferences UI … the journey to a thousand parameters is taken one parameter at a time.

The ability to specify a default precision could be removed both from the GUI and the command line options with the default precision always being set to 6 (when running scripts) or 17 (when displaying the AST).  In the former case the user could override the default using a `$fp` assignment at the top level of their script.

> (Which also applies to the proposed $fp.)

I'm of the opinion that the user should be able to specify this in some fashion, for normal execution if not for the AST.

jordanbrown0:
> I'm of the opinion that the user should be able to specify this in some fashion, for normal execution if not for the AST.

Via a per-case formatting function, absolutely.

Globally, or special-variable-scope?  I'm very dubious.

I think that, absent explicit formatting, echo() and str() should produce reasonable presentations of numbers.  (Noting that 1.000000000000001 is **not** reasonable.)  I believe that it's currently using something roughly equivalent to `%.6g`.  That seems like a limit that's far too small; human-interesting values can easily exceed it.  1,000,001 triggers exponential notation (which might or might not be considered reasonable) and loses the "1", which I think is clearly wrong.  While I would not want it to precisely represent last-few-bit errors, I suspect that I would be happier with something around `%.14g`, and it's hard to see a downside.

Alex-Kent:
> it's currently using something roughly equivalent to %.6g

Correct.

> I would be happier with something around `%.14g`, and it's hard to see a downside.

The downside is if extra precision is present and the user cares about it (the latter being impossible to implicitly ascertain).

One possibility is to return the shortest possible string representation that can accurately represent the internal `double` value.  The `double-conversion` library (used internally to do `double` to `std::string` conversion) offers a [`ToShortest(…)`](https://github.com/google/double-conversion/blob/f1a2ac8318801624a0d6f2bf2970ef7e85bd80e9/double-conversion/double-to-string.h#L202-L249) method that does this [technically it outputs the fewest possible significant digits].  I've implemented that as default in my local code and am verifying that it works correctly/as-expected.  Currently a number of `ctest` checks are failing, however all of those that I've manually checked are cases where the internally value really does have extra precision, sometimes obviously, sometime not e.g. [`echo((11)*pow(10,-9)`](https://github.com/openscad/openscad/blob/93c839e762b982bbae76dc5fea5618c886caa48d/tests/data/scad/misc/echo-tests.scad#L29) ≠ `echo(11e-9)` (for the former `ToShortest` accurately returns `1.1000000000000001e-8` not `1.1e-8`).

Alex-Kent:
> I'm wondering why all those *.0000000001 are needed, I thought DoubleConversion can cut that down in most cases while still providing the guarantee of restoring the exact same binary representation on reading the value back.

Sometimes those trailing digits are needed to accurately represent the value, at other times a decimal form with fewer significant digits will yield the same `double` value.  The `double-conversion` library offers the [`ToShortest(…)`](https://github.com/google/double-conversion/blob/f1a2ac8318801624a0d6f2bf2970ef7e85bd80e9/double-conversion/double-to-string.h#L202-L249) method that does this, returning the shortest string [in terms of SDs] that fully-represents the binary `double` value.  (Currently with OpenSCAD conversion is done using [`ToPrecision(…)`](https://github.com/google/double-conversion/blob/f1a2ac8318801624a0d6f2bf2970ef7e85bd80e9/double-conversion/double-to-string.h#L337-L375) with a precision of 6.)

>  I would not even mind if AST export allows for a setting to specify the precision as export parameter, although I'm not sure it's worth the effort

For AST output this could just be hard-coded to 17 (the minimum guaranteeing round-tripability) or to the minimum required for a given `double` (i.e. using `ToShortest(…)`).  Note that there still needs to be a mechanism to convey this from the `ASTNode::dump(…)` call through to the `DoubleConvert(…)` calls that generate the output strings.  Additionally there likely needs to be a way to differentiate between AST and non-AST output; with this PR that's done entirely within the `ASTNode::dump(…)` method (by setting the output stream's precision; the rest of the output code is identical for both with no special-casing).  As far as effort goes, in this PR the only substantive change (aside from retrieving the user's preference) that's AST-specific is adding a single `stream << std::setprecision(precision);` line to `ASTNode::dump(…)`.

> • Adding more context (not necessarily just looking at the class with that name, but also evaluation session) to the whole evaluation logic would be useful. Especially the broken console output would hugely benefit from that, but that's all internal, not related to the user level scripts

Agreed.  For `echo(…)` in particular whether to output `ECHO: ` could be controlled by such context, obviating the need for a new `echo_without_prefix(…)` function.

> • Why do we need that $fp? What's the actual use case to put that huge cannon ball on that (to me so far) very tiny problem.

I chose that route for a number of reasons:

* It parallels the use of other `$`* variables to set state within a block (especially `$fn`).
* It is propagated through much of the logic with no changes to the code.  (Specifically throughout evaluation.)
* It exposes the precision to user code so that:
  * Code can determine the active precision.
  * Block-local changes to it can be made.
(For rendering labels and such this is particularly important.)
* Though it's beyond the scope of simple precision-setting, using a user-settable variable (which is just a `Value`) to specify the output formatting would allow for things like user-definable formatting to be defined (using a function reference, e.g. for an `object`) with minimal changes to the OpenSCAD code base.

As for the name chosen, `$fp` alludes to Floating Point or Float Precision as well as the kindred `$fn`.

> • Using $fp would define precision also for library calls, I'm not sure that's helpful.

How so?  Internally OpenSCAD uses IEEE-754 64-bit `double` values for all maths.  Lower precision is only used when converting a number to a string (i.e. for display).

> • We do need actual formatting (not just for numbers) as that's often requested, so hopefully it's coming at some point. I feels it would be much more of general use to dust that PR off.

Agreed.  Implementing actual formatting is highly desirable and would probably be quite straight-forward to implement.  I have some specific thoughts regarding this that I'll bring up on the other issue (#6385).

> Assuming that would already exist, what would be the remaining use cases for $fp that are not covered except the magic behavior which is not necessarily always good?

Mainly that it would allow some degree of user control over numeric output in the simple case (e.g. `echo(a=a)`); no need to write (and maintain) `echo(sprintf("a = %g", a))` and the like (especially an issue when there are multiple such calls scattered throughout a large code base).  Makes scripts more readable as well.

Additionally, by having `$fp` be settable within code users have the option to set the active precision only where needed; modules or libraries can likewise (if necessary) override the user's preference.

> magic behavior

For clarity, could you be explicit about what you're referring to?

> the other AST-only PR … ["Which one"] … Well, #6408, are there more that I missed?

Sorry, I was a bit overloaded when I wrote my query, conflated it with #6385.  The innermost portion of the code in #6408 (everything once `Value::ToString(…)` or `<< Value` is called) is essentially the same as in this PR.  This PR simply adds 1) a way to specify the default precisions to use, and 2) a way (`$fp`) for user code to access the arbitrary-precision option (via `str` or `echo`).  As mentioned above, using `$fp` for the latter involves trivial code changes.

Alex-Kent:
> Do _any_ other languages have global-ish controls on how numbers are formatted?

* Perl previously did (using `$#`, until 5.10).
  Perl also has a per-file-handle [`format`](https://perldoc.perl.org/perlform) option but that's more of a filter applied to the output stream (for generating reports and the like).
* C++ allows precision to be set on-the-fly for all streams (including `std::cout` and `std::cerr`).  Indeed this PR uses that mechanism to pass the precision within `toString(…)` (for complex types specifically, `Vector` and the like).
* Various Python extensions (e.g. `numpy`) allow setting global defaults for string formatting (e.g. [numpy.set_printoptions](https://numpy.org/doc/2.2/reference/generated/numpy.set_printoptions.html)).

jordanbrown0:
> I believe that it's currently using something roughly equivalent to `%.6g`. [...]  I suspect that I would be happier with something around `%.14g`, and it's hard to see a downside.

I meant a downside to `%.14g` as compared to `%.6g`.  I believe I would support unconditionally switching to `%.14g` as the output format for `echo()` and `str()`.

Going back to AST for a moment, are there any cases where a short input value (e.g. `1.1`) yields a long output value (e.g. `1.1000000000001`)?  A "shortest possible" strategy should always yield the input value, or shorter.  After all, that input value **was** the value that yielded that precise bit pattern.

Alex-Kent:
> I meant a downside to `%.14g` as compared to `%.6g`. I believe I would support unconditionally switching to `%.14g` as the output format for `echo()` and `str()`.

I'm opposed to such a change (specifically 6→14 SD) because it can give the user the false impression that a full representation of the value is being displayed when it may well have been rounded.  At least with `%.6g` it's very strongly implied that one's seeing a rounded value.  If anything switching to a "shortest" [lossless] representation would be the more sensible approach.  Note that any change to the output format (yours, mine, or otherwise) could cause compatibility issues with existing code which is a big part of why I think the output format should be user-controllable at the block level (with the default [for non-AST output] being the current behaviour).

> Going back to AST for a moment, are there any cases where a short input value (e.g. `1.1`) yields a long output value (e.g. `1.1000000000001`)?

For a correct "to shortest" function there should be no instances where that's the case.  Based on some testing it looks like Gnu `libc`'s `%g` `printf` option doesn't always return the shortest string (to be fair the `sprintf` man page is a bit hazy on its precise behaviour).  The `ToShortest` method in `double-conversion` though **does** behave sensibly for all values I've examined; importantly it never returns a string that can't be losslessly round-tripped.

I'm strongly inclined to the use of something akin to `ToShortest` for all simple numeric output with the rationale being that it always yields the most concise possible representation of a value without discarding any bits from it.  The only downsides that come to mind are the aforementioned backwards-compatibility issue and that it may generate more precision than the user actually cares about (though it's better to have more precision than less IMO, especially in this context).

MichaelAtOz:
Assuming something like #6385 gets up, you don't need to mess with backward comparability with `echo()`.
People do lots of things with echo output, such as Bill of Materials, which feed into other processes.
Changes to echo output would have wide ranging effects.

jordanbrown0:
> I'm strongly inclined to the use of something akin to ToShortest for all simple numeric output with the rationale being that it always yields the most concise possible representation of a value without discarding any bits from it.

Works for me.

> The only downsides that come to mind are the aforementioned backwards-compatibility issue

I think that anybody broken by such a change deserves what they get.  There are things you get to depend on, and things you don't get to depend on, and this is one of the things you don't get to depend on.

> and that it may generate more precision than the user actually cares about

I don't think we promise any particular formatting.  This just isn't that kind of language.  If one is really concerned, tie any generic-output changes to the availability of a format() function, so that if some user doesn't like the new formatting they have a way to get what they want.

> (though it's better to have more precision than less IMO, especially in this context).

I agree, out to about 50 bits (of the 53 available).  I think that representing last-couple-of-bits fuzz has negative value.  If I ever need to know the difference between X and X+epsilon, I know how to subtract.

MichaelAtOz:
> > The only downsides that come to mind are the aforementioned backwards-compatibility issue
> 
> I think that anybody broken by such a change deserves what they get. There are things you get to depend on, and things you don't get to depend on, and this is one of the things you don't get to depend on.
> 
> > and that it may generate more precision than the user actually cares about
> 
> I don't think we promise any particular formatting. This just isn't that kind of language. If one is really concerned, tie any generic-output changes to the availability of a format() function, so that if some user doesn't like the new formatting they have a way to get what they want.

Deserves what they get? You have gone mercinary.
I can only hope the project as a whole has more respect for what people have created over many years.

jordanbrown0:
I have a lot of respect for what people have created over time.  This is the kind of dependency that would indicate a disrespect for the realities of maintaining and evolving a tool like OpenSCAD.

Unreasonably depending on something that clearly isn't a precisely-intended behavior is bad faith on the part of the consumer.

I work on enterprise-class software.  Compatibility is a daily concern, and frequently dominates the conversation.  Those needs have led our organization to develop an extensive set of policies around compatibility - about what needs to be held compatible, what does not need to be held compatible, and what to do when reality demands a compatibility break.

Those policies can largely be summarized in about three points:
* Document precisely what the committed interfaces are.  Don't change committed interfaces incompatibly.
* But when you must change a committed interface, or even a well-known not-committed interface, provide lots of notice and provide a transition period if applicable.
* And the flip side is:  if a behavior is not documented, and especially if it is obscure or not clearly intended as a committed interface, the consumer must not depend on it and it can be changed without notice.

OpenSCAD does not have that level of formality.  However, the same principles apply.  There are things that are promised, and those must be maintained - and, at the same time, the reality is that there are times when a compatibility break is the best alternative.  And in return, the consumer must take care and pay attention to what they are depending on, and not depend on anything that isn't clearly intended to be maintained compatibly.

That said...

In this particular case, looking at the documentation, somebody wrote down something ... undesirable ... in the documentation for echo().  They wrote "Numeric values are rounded to 5 significant digits.".  (Which is wrong; they are rounded to six.)  Those who pay attention to what I've written about our documentation might recognize this as something I would specifically *not* document, for exactly the reason that we see today:  the possibility that somebody is depending on that statement.  (Thankfully, str() does not say anything about rounding.)  Document intent; do not document behavior.

Although changing that would theoretically be a compatibility break, I would go ahead with it anyway.  I think that (a) it is not reasonable to have depended on that statement, and (b) the product will be in a better place if we retract it, especially if we replace it with better alternatives.  Mistakes of the past should not bind us forever, especially in an environment where the person writing the documentation may not have understood the intent.

BTW, the word you're looking for is draconian, not mercenary.  Mercenary implies getting paid.  But yes, draconian, guilty as charged.

MichaelAtOz:
Yes sorry Draconian, ... or some other word if it is cauing others pain??

While I'm now retired, my whole career was with enterprise systems.

> precisely-intended behavior

OpenSCAD has few precisely documented things.

One is command-line `-o echo`. "specifying the extension .echo causes openscad to produce a text file containing error messages and the output of all echo() calls in filename as they would appear in the console window visible in the GUI"
While it does not define the `echo` format, it says it is a valid interface mechanism.

> And in return, the consumer must take care and pay attention to what they are depending on

You are saying the OpenSCAD consumer cannot depend on much.
The only means of output, echo(), should not be depended on, when it is the only way?

> there are times when a compatibility break is the best alternative. 'when reality demands'

The _best_ &  _reality_ is in the mind of the beholder. 
Then there are times when it is easy to not break. 
I would have the rule 'If it is easy not to break, don't break it' pretty high on the list.

Just to clarify my position, provided another feature is introduced allowing output format to be specified,
changing **default** behaviour of echo() is not adding much benefit, if any, it has sufficed for more than a decade.
The original proposal here was not to change the default, only change output format intentionally via $fp.

Yet breaking the default output will likely ruin anything that subsequently processes such output, alignment of fields is usually  important when processing text input.

I don't have anything that relies on echo() format, but I have seen over the years plenty of things which post process echo() output. Many examples of Bill of Materials. 

Some people actually use OpenSCAD in their businesses.
I'm pretty sure they would argue “Breaking changes need a business case”. I don't see one.

I'll ask the community how much it may hurt.

jordanbrown0:
> OpenSCAD has few precisely documented things.

Right.  Which means that you shouldn't depend on precision that isn't there.

> While it does not define the echo format, it says it is a valid interface mechanism.

Absolutely.

> > And in return, the consumer must take care and pay attention to what they are depending on

> You are saying the OpenSCAD consumer cannot depend on much.

Hmm?  There's a large manual full of stuff I think they can depend on.  It's not the greatest, but it does describe a large fraction of what you need to know.  But if you are doing something that it doesn't say, you should be applying some wisdom to decide whether it's something that you **should** depend on.  And even when it **does** say something, you should apply a little thought to whether it might change.

> The only means of output, echo(), should not be depended on, when it is the only way?

Did I even come close to saying that?

I said that one particular point case - the rounding and exponentiation behavior of numbers - is not something that you should depend on.  You absolutely can depend on the fact that echo() will do something sensible with numbers that you give it, and you can reasonably depend on it giving you **at least** as much precision as it gives you now.

You can depend on the generally-inconvenient `echo()` output format, with the "ECHO" prefix and quotes.  You can even (sigh) depend on the bug that the strings that it emits aren't valid string syntax since they don't properly quote special characters.

But I don't think it's reasonable to complain when printing 0.1234567 yields `0.1234567` instead of `0.123457`, or when printing `1234567` yields `1234567` instead of `1.23457e+6`.

> I would have the rule 'If it is easy not to break, don't break it' pretty high on the list.

Absolutely.  Somebody has complained that the output doesn't give enough precision, and they're right.  (I happen to disagree with them on how much is "enough", but six digits isn't it.)  The default behavior is not the greatest.  We can either fix that default behavior, and make things a tiny bit better, or we can say "we know it's not great, but we're stuck, here's a workaround".

> changing default behaviour of echo() is not adding much benefit, if any, it has sufficed for more than a decade.

Well, sure, but that can be said for absolutely everything.

> The original proposal here was not to change the default, only change output format intentionally via $fp.

Yes, which imposes yet a different kind of cost, an incremental increase of the complexity of the language, for very little benefit.  (And how many people will be confused by the fact that `echo($fp=9, 1.23456789);` does not do what they expect?)

And, while we're talking about microscopic compatibility breaks, what about programs that are already using $fp for something else?

> Yet breaking the default output will likely ruin anything that subsequently processes such output, alignment of fields is usually important when processing text input.

I think that's dramatically overstated.  I would say that with today's common tools, delimited-value processing is far more common than fixed-width-oriented processing.  (And generating fixed-width output in OpenSCAD is not impossible, but is less than totally fun.)

> I'm pretty sure they would argue “Breaking changes need a business case”. I don't see one.

The business case is very small, to be sure:  the fact that you can't print 1.234567 or 1234567 and get the right answer.

But the business case for maintaining "compatibility" in this case is also very very small:  the risk that somebody, somewhere, has fragile post-processing that can't handle this small change in the behavior of unusual numbers.

jordanbrown0:
I mentioned it in passing in the long note above, but I realized a non-trivial problem with the $fp idea.

You can't use `$fp` **in** the echo() that you want to format, because echo() already defines behavior for all name=value pairs.

You can say `circle(10, $fn=100)` and it will give you a circle with a hundred sides.  But if you say `echo(123, $fp=8)`, it must say `ECHO: <something>, $fp=8`.  **Maybe** it would be OK for it to say something other than "123", controlled by the $fp setting, but it must emit the "$fp=8" part because it's defined to emit all such name-value pairs.

So that would mean that you can set a $XX variable in the call that uses it... except for echo() and $fp.

str() in theory would have a similar issue, but its current behavior of just ignoring the name is one that I would be OK with changing.  `str(123, $fn=456)` currently yields `123456`.

jordanbrown0:
You say that $fp would control the number of significant figures.

Would it affect the point at which the output switches to exponential notation?  Today, it switches for numbers less than 1E-05 or greater or equal to 1E+06; would that change or would you just get a longer exponential-notation number?

bitbasher:
> Document intent; do not document behavior.
the context and style of a document surely follow from its intent @jordanbrown0 .. how can it not?

User guides document behaviour or they risk being irrelevant ( or less relevant given how most users ignore them)

Language References document have to document what exists or they fail the user. They may mention features removed in the march of time, but if used to describe _intent_ are complety useless to the poor sods who are thne left to experiment with test code and mock ups until they grow the own body of knowledge about how the syntax and built-in functions actually work

specifications describe intent and that is the only place for it.

Any other description of _things that exist_ fail in the goal of informing the reader if what is written does not match up with what a user is experiencing.

> Mistakes of the past should not bind us forever, especially in an environment where the person writing the documentation may not have understood the intent.

interesting to see that point enunciated 

> BTW, the word you're looking for is draconian, not mercenary. Mercenary implies getting paid. But yes, draconian, guilty as charged.

i can see you as a dragon Jordan .. singed keyboard and all

bitbasher:
and for he record .. i think $fp is a great idea as written .. having no control over the display precision of very large values, or very small fractions, has been a problem for me .. i had to write my own number formatting code  to see a number of 10-12 digits because i had no way to see what was being rounded off

Alex-Kent:
> You say that $fp would control the number of significant figures.
> Would it affect the point at which the output switches to exponential notation?

Depends on how the `$fp` value is interpreted (backend used [`printf` or `double-conversion`] and their other parameters).  The code in this PR feeds the value to what's essentially a variation on `%g` (`double-conversion`'s (potentially-lossy) `ToPrecision(…)`, i.e. what OpenSCAD currently uses).  My [unpublished] local code adds an `18` precision that switches to using `double-conversion`'s (lossless) `ToShortest(…)` method.

For `ToPrecision` (currently used by OpenSCAD) exponential notation is used less often as precision increases.  Generally `ToShortest` will output exponential notation less frequently than `ToPrecision` (using the current OpenSCAD settings).

The `ToShortest` method is pretty tweakable (particularly via `DC_DECIMAL_LOW_EXP`, `DC_DECIMAL_HIGH_EXP`, `DC_MAX_LEADING_ZEROES`, and `DC_MAX_TRAILING_ZEROES`); it wouldn't be hard to constrain the output to, say, 6 digits (when possible).  I wrote [some code](https://gist.github.com/Alex-Kent/649f980eae859af87e017cf3adc35bff) to experiment with the various options (e.g.`%e`, `%g`, `ToPrecision`, the various `ToShortest` settings) and their effects on round-trippability (double→string→double); it's not the prettiest code but it's fine for casual tinkering.

Alex-Kent:
> i had to write my own number formatting code to see a number of 10-12 digits because i had no way to see what was being rounded off

Indeed that's what led me to this rabbit hole.  I was writing code where the extra digits **do** matter.  Tracing my code using `echo` was misleading and of limited utility because the bits I cared about were being truncated.

Alex-Kent:
> Although changing that would theoretically be a compatibility break, I would go ahead with it anyway.

The #​1 priority should [must] be **don't break existing code**.  Apologies for shouting but there's a **lot** of extant OpenSCAD code out there and IMO it's not justifiable to potentially break it when there are backwards-compatible solutions that can be employed.

It's fine to improve things but said compatibility **must** be maintained unless there's an extraordinarily-good reason to do otherwise.  (An example of the latter is the AST output; its current behaviour [output not matching internal state] is arguably wholly incorrect and it should by default generate correct output even if that means breaking existing 3rd-party tooling.)

Alex-Kent:
> what about programs that are already using $fp for something else?

I think the documentation needs to be updated to explicitly state that all `$`* variables are reserved for OpenSCAD's use and should not be declared in user code.  (Users using existing ones is fine but they should not create new ones.)

> a non-trivial problem with the $fp idea.
> 
> You can't use `$fp` **in** the echo() that you want to format, because echo() already defines behavior for all name=value pairs.

`$fp` controls the output of string functions within the block it resides in.  For built-ins it does not directly modify behaviour when passed as an argument (indeed for functions it raises an warning) though it does apply to modules' child nodes.  This is partly because this PR makes the minimum level of changes needed to implement its functionality and partly because it intentionally does not change behavior or semantics beyond its remit.

Note that with this PR's code one can pass an explicit `$fp` parameter to a module (user or built-in) or user function and it will automatically be applied to the module or function even if not listed in the its argument list (this is existing behaviour, not something added by my code).

> echo() and … str() … current behavior of just ignoring the name is one that I would be OK with changing. `str(123, $fn=456)` currently yields `123456`.

`str` simply takes a list of values and return a concatenated string.  Much as with Unix `cat` there is zero hidden magic going on and there are no special cases.  its job is to concatenate verbatim, nothing more, nothing less.  `echo` does perform a bit of manipulation but it's limited and, importantly, consistent.  Neither should have their default behaviour altered.

jordanbrown0:
> I think the documentation needs to be updated to explicitly state that all $* variables are reserved for OpenSCAD's use and should not be declared in user code. (Users using existing ones is fine but they should not create new ones.)

That equine has long since departed the storage facility.

`$` variables are the **only** way to communicate with child modules.  A very large fraction of the things that BOSL2 does would not be possible without them.  Take a look at BOSL2 some time.  OpenSCAD defines something like six or eight `$` variables.  It looks like BOSL2 defines close to a hundred.

It's not clear what this rule would even mean.  It was OK that I used `$foo` yesterday, but if I type it today it's wrong?  What if I type it today into 2021.01 and *then* upgrade?  And of course from a compatibility position it is yesterday's use that is a concern

And, going back to `$fp` itself, I think somebody just said:
>The #​1 priority should [must] be don't break existing code.[...]

Note that we have people in this community who say "absolutely no changes that are even theoretically incompatible", and others who say "we should document absolutely all of the behavior; if we decide to change it later that's OK".  Those are incompatible positions.  In my opinion, neither of them is correct; the correct answer is in the middle somewhere.  Document those things that you plan to maintain forever.  Do *not* document things that might reasonably change (or if you do, mark them heavily, and consumers who care about compatibility must avoid them).  Do not change things that you've documented, absent a really good reason.  Feel free to change things that you have not documented.  Apply a little flexibility on both sides to account for inappropriately-documented behaviors and to account for incomplete documentation of behaviors that become reasonably used.  But if you do not draw a reasonably bright line between "committed" and "not committed" behavior, you are doomed.

Going back to proposed behavior...

**Every** current module and function that consumes a `$` variable allows you to set that `$` variable in the call to the module or function.  But you propose that `echo()` and `str()` should be an exception to that rule; they should consume the variable but should not allow setting it on the call.

**Every** function and module can accept **any** `$` variable as an argument.  But you propose to exclude `$fp` from being passed to functions.  (Even though it would make sense, to control any `str()` calls inside that function.)

PR #6423, opened 2025-12-04
fixing #6224 read JSON error reporting for Customiser

bitbasher:
… capture improved. it has been confirmed as working in the GUI (Windows only so far) and for the command line

the errors are logged to the Error DB and so are visible in that table as well as in the console

running from the command line i notices that customizer errors were not being logged .. turned out that was due to the reading of the JSON file happening before error reporting was set up. Opus figured out a way to log and preserve the error message so it could be shown in the console after the fact

### DIFF
--- a/src/core/customizer/ParameterSet.cc
+++ b/src/core/customizer/ParameterSet.cc
@@ -8,19 +8,27 @@ static std::string parameterSetsKey("parameterSets");
 static std::string fileFormatVersionKey("fileFormatVersion");
 static std::string fileFormatVersionValue("1");
 
-bool ParameterSets::readFile(const std::string& filename)
+bool ParameterSets::readFile(const std::string& filename, JsonErrorInfo& errorInfo)
 {
   boost::property_tree::ptree root;
 
   std::ifstream f(std::filesystem::u8path(filename));
   if (!f.good()) {
+    errorInfo.message = "Cannot open file for reading";
+    errorInfo.filename = filename;
+    errorInfo.line = 0;
     LOG(message_group::Error, "Cannot open Parameter Set '%1$s' for reading", filename);
     return false;
   }
   try {
     boost::property_tree::read_json(f, root);
   } catch (const boost::property_tree::json_parser_error& e) {
-    LOG(message_group::Error, "Cannot open Parameter Set '%1$s': %2$s", filename, e.what());
+    // Extract structured error information from the exception
+    errorInfo.message = e.message();  // Just the error message without file/line
+    errorInfo.filename = filename;    // Use our filename (e.filename() may be empty for streams)
+    errorInfo.line = e.line();
+    // Don't LOG here - the caller is responsible for reporting the error
+    // This allows for proper file/line info in the ErrorLog or console output
     return false;
   }
 
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -2092,6 +2092,9 @@ std::shared_ptr<SourceFile> MainWindow::parseDocument(EditorInterface *editor)
 {
   resetSuppressedMessages();
 
+  // Report any JSON parsing errors from the parameter set file
+  editor->parameterWidget->reportJsonErrors();
+
   auto document = editor->toPlainText();
   auto fulltext = std::string(document.toUtf8().constData()) + "\n\x03\n" + commandline_commands;
 
--- a/src/gui/parameter/ParameterWidget.cc
+++ b/src/gui/parameter/ParameterWidget.cc
@@ -42,6 +42,7 @@
 #include "gui/parameter/ParameterText.h"
 #include "gui/parameter/ParameterVector.h"
 #include "gui/Preferences.h"
+#include "utils/printutils.h"
 
 #include <filesystem>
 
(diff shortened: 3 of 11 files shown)


t-paul:
Please give the PR and also the commits dedicated titles/headlines that work in the commit history. https://github.com/openscad/openscad/commits/master/

bitbasher:
but this is the pr ? and the only commit is shown on the link for the source of the PR ?

bitbasher:
@t-paul this PR is based on one commit .. the noise is from all the commits that i pulled down to my fork to be sure my code changes were not causing conflict. i did not know about merge rebase, sorry

t-paul:
What I meant is really just the title of the commit and the PR, the link was there to illustrate what is shown in the full commit history.

Let my try with a picture of the github notification mails:

image

One thread says "Opus and i have fixed..." the main information may come like 3 screens further right. The other one has a very concise title making it directly clear what this is about.

Try running `git log --oneline` in the local repository folder. Ideally it gives a list of short titles one can go through a couple of years later and find commits about a specific topic.

In random unamed work related repos I've seen titles like:
* bugfix
* checkin
* fix issue 3934
* more bugfixes
* still not working

Trying to find anything in that history is impossible with those commit/PR titles.

Some actual older commits from this repo:

* Move projection implementation into ClipperUtils
* Workaround for Qt crash bug related to resizing/moving a bad geometry
* Added more corner case 2D tests
* Enable Manifold's triangulator by default
* Use prebuilt manifold on macOS

It does not need to be a perfect title, but it should spell out what it does in about 80 characters.

bitbasher:
> It does not need to be a perfect title, but it should spell out what it does in about 80 characters.

no i get it .. okay .. 80 characters full of pertinent info from now on

bitbasher:
do i now need to do the Update Branch before this can be reviewed? @t-paul  because i have done other work in the branch to spread my reporting fixes to the other places that read json and an update might mess things up

bitbasher:
What can I do to move this along

src/core/customizer/ParameterSet.cc
```diff
@@ -8,19 +8,27 @@ static std::string parameterSetsKey("parameterSets");
...
   try {
     boost::property_tree::read_json(f, root);
   } catch (const boost::property_tree::json_parser_error& e) {
-    LOG(message_group::Error, "Cannot open Parameter Set '%1$s': %2$s", filename, e.what());
+    // Extract structured error information from the exception
+    errorInfo.message = e.message();  // Just the error message without file/line
+    errorInfo.filename = filename;    // Use our filename (e.filename() may be empty for streams)
+    errorInfo.line = e.line();
+    // Don't LOG here - the caller is responsible for reporting the error
+    // This allows for proper file/line info in the ErrorLog or console output
```
kintel, line 31: Why are we not logging here, but we are logging a handful of lines above?

tests/bad-json/full-test-fix-session.txt
kintel, line 1: This file seems to be some AI chat log. Please turn this into a more human readable test script.

src/gui/parameter/ParameterWidget.cc
```diff
@@ -106,11 +109,11 @@ void ParameterWidget::saveFile(const QString& scadFile)
   }
 
   QString jsonFile = getJsonFile(scadFile);
-  if (jsonFile == this->invalidJsonFile) {
+  if (this->jsonError.hasError() && jsonFile.toStdString() == this->jsonError.filename) {
```
kintel, line 112: What does the second check do? Are there situations where these filenames may differ?

src/gui/MainWindow.cc
```diff
@@ -2092,6 +2092,9 @@ std::shared_ptr<SourceFile> MainWindow::parseDocument(EditorInterface *editor)
 {
   resetSuppressedMessages();
 
+  // Report any JSON parsing errors from the parameter set file
+  editor->parameterWidget->reportJsonErrors();
```
kintel, line 2096: Considering ParameterWidget collects errors on both read and write, are there better places to handle errors than in parseDocument()?

tests/bad-json/customizer-all-bad.scad
kintel, line 1: If you're simply checking for error vs. no error, you can probably use a minimal example file instead

kintel (commented):
Added some comments.

src/core/customizer/ParameterSet.cc
bitbasher: becuase the first check is some thing we can catch by checking in the file system, and there is no detail to report from the json file content.
But with the read_json() function one needs a try-catch arrangement .. I wanted all the json errors to be reported by the Parameter code, as it is now. The mandate was to provide better feedback on the nature and location of the problme in the json, so i capture that info in the catch block and then still return from the function.

src/gui/MainWindow.cc
bitbasher: possibly .. i wanted to limit the scope of the changes to just what was needed to get better reporting on bad json.
And the code is a bit tangled so .. i am happy to accept a pointer to a better place?

src/gui/parameter/ParameterWidget.cc
bitbasher: oops .. thanks .. that was to check that the error was really coming from the file name that got passed in as scadFile

tests/bad-json/customizer-all-bad.scad
bitbasher: oh well, that is true .. i built this file as part of learning about the customizer so i just grabbed it to build this test .. it is go/no go but it checks pretty much all of the widgets it forms a comprehensive smoke test .. if it would be better to use only a handful of the most used widgets i am happy to change it

tests/bad-json/full-test-fix-session.txt
bitbasher: this file should have been removed in my pre-PR cleanup

jordanbrown0:
Mentioning the issue number in the title doesn't actually link it to the PR.  So:  #6224.

PR #6427, opened 2025-12-06
Delete discretizer keyword args

### LINKED ISSUE #6426: Python fn/fa/fs not working as keyword args
**Describe the bug**

```
from openscad import *

c=cylinder(h=2, d=7.25 + 1/8 + 1/64, fn=20).rotate([0, 0, 9]).color("brown")
# c=cylinder(h=2, d=7.25 + 1/8 + 1/64).rotate([0, 0, 9]).color("brown")

show(c)
```

The uncommented line is giving an error about invalid args; the commented line won't.

**To Reproduce**
1. Enable Python in Experimental
2. Create a something.py file and put the above contents
3. F5

/assign-me

coryrc:
PyArg_ParseTupleAndKeywords doesn't allow keyword args it doesn't know about, so we delete them if found.

Manually verified the fix.

Not in love with the solution.

Fixes #6426

@YoSTEALTH

### DIFF
--- a/src/python/pyfunctions.cc
+++ b/src/python/pyfunctions.cc
@@ -98,6 +98,7 @@ PyObject *python_sphere(PyObject *self, PyObject *args, PyObject *kwargs)
 
   double vr = 1;
 
+  auto discretizer = CreateCurveDiscretizer(kwargs);
   if (!PyArg_ParseTupleAndKeywords(args, kwargs, "|Od", kwlist, &rp, &d)) {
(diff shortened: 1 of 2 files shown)


kintel (approved):

t-paul:
> Not in love with the solution.

Agreed, it would be much nicer to have something a bit like context managers. But lets go with the fix first. Thanks!

YoSTEALTH:
Just did a quick test and its working nicely. Good work @coryrc

PR #6432, opened 2025-12-09
Add optional auto preview on editor changes

### LINKED ISSUE #6431: Auto Preview while editing in internal editor (debounced)
I find the Number Scroll feature lovely.
It would be great to also get instant feedback when editing code.

Considered ex: `autosave=500ms` but saving seems unnecessary.
Instant or delayed (debounced) preview would be better.

/assign-me

gorbiz:
Fixes #6431

This adds an optional "auto preview on edit" behavior for the built-in editor.

- New setting: `editor/autoPreviewDelayMs` in `OpenSCAD.conf`
- When `autoPreviewDelayMs >= 0`, changes in the internal editor schedule a
  debounced Preview after the given delay.
- The delay can be configured from Preferences → Editor (“Automatic preview on
  code change” group).
- If the key is missing or empty, behavior stays as before (no auto preview
  on edit).

This mirrors the Customizer's "Automatic Preview" for normal editing, but with
a configurable debounce so complex models don't recompile on every keystroke.


src/gui/MainWindow.cc
```diff
@@ -3102,12 +3111,47 @@ void MainWindow::editorContentChanged()
...
     animateWidget->editorContentChanged();
 
     // removes the live selection feedbacks in both the 3d view and editor.
     clearAllSelectionIndicators();
   }
+
+  if (autoPreviewEnabled && autoPreviewTimer && docChanged) {
+    autoPreviewTimer->stop();
+    autoPreviewTimer->setInterval(autoPreviewDelayMs);
+    autoPreviewTimer->start(autoPreviewDelayMs);
```
kostnik: no need for `stop` and `setInterval` here:

https://doc.qt.io/qt-6/qtimer.html#start-2

```suggestion
    autoPreviewTimer->start(autoPreviewDelayMs);
```

gorbiz: Much better! :pray:

gorbiz (commented):
No need to `stop` & `setInterval` as `start` already restarts it.

PR #6436, opened 2025-12-11
Add --args to dump command line arguments.

### LINKED ISSUE #6435: Mechanism to have OpenSCAD dump its command-line arguments
Command-line quoting is difficult.  It's doubly difficult in an environment like Windows, where some of the processing is done by the shell-ish program (CMD, PowerShell) and some is done by the C runtime module.

One of the key things that you need to do to get quoting right is to analyze exactly what's happening at each level of quoting.  That's hard when one or more layers are opaque.  Shining light on as many of those layers as possible, so that it's possible to examine the command line at each step of its processing, should make it easier to find the right quoting.

Have a mechanism that causes command-line OpenSCAD to print its arguments on standard out, with markings showing each argv[] entry, et cetera.  Perhaps this mechanism would be a command-line option, e.g. `--args`, or perhaps it would be triggered by an environment variable.

On Windows, it should dump both the raw Windows command like (that comes back from GetCommandLineW()) and the resulting argv.  This would give you, I think, a complete picture of the command-line processing:  you know the input, it would show you the command line presented to the program, and it would show you the low-level processing that the program does on it.

Documentation and trial and error.  But PowerShell has defeated me.

jordanbrown0:
Fixes #6435.

The major reason that I'm creating a PR at this point is to get a CI-built Windows OpenSCAD to experiment with.

### DIFF
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -791,6 +791,17 @@ struct CommaSeparatedVector {
   }
 };
 
+void
+dumpArgs(int argc, char **argv)
+{
+  PlatformUtils::dumpArgs(argc, argv);
+
+  for (int i = 0; i < argc; i++) {
+    std::cout << "argv[" << i << "] = >" << argv[i] << "<\n";
+  }
+  std::cout << "\n";
+}
+
 // OpenSCAD
 int openscad_main(int argc, char **argv)
 {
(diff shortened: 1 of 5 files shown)


t-paul:
It's a bit sad it's needed but as people have had this type of issue repeatedly it's probably worth having. I'd prefer a more specific name of the command line switch itself though. Just `--args` sounds like it might be useful for other things at some point. Hoping this is not need all that often, a long name like `--dump-cmdline-args` might be ok?

PR #6447, opened 2025-12-20
chore: bump nix-shell for latest Nix release

flaviut:
Before this change:

```
❯ nix-shell ./scripts/nix/shell.nix
error:
       … while calling the 'derivationStrict' builtin
         at «nix-internal»/derivation-internal.nix:37:12:
           36|
           37|   strict = derivationStrict drvAttrs;
             |            ^
           38|

       … while evaluating derivation 'nix-shell'
         whose name attribute is located at /nix/store/sxx303my1hxx6zrbcrpwgjgvp9n1gd8v-source/pkgs/stdenv/generic/make-derivation.nix:541:13

       … while evaluating attribute 'buildInputs' of derivation 'nix-shell'
         at /nix/store/sxx303my1hxx6zrbcrpwgjgvp9n1gd8v-source/pkgs/stdenv/generic/make-derivation.nix:593:13:
          592|             depsHostHost = elemAt (elemAt dependencies 1) 0;
          593|             buildInputs = elemAt (elemAt dependencies 1) 1;
             |             ^
          594|             depsTargetTarget = elemAt (elemAt dependencies 2) 0;

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: qt6.full has been removed. Please use individual packages instead.
```

After this change:

```
❯ nix-shell ./scripts/nix/shell.nix

[nix-shell:~/dev/openscad]$ cmake -B build -DEXPERIMENTAL=1 -DUSE_QT6=on
...
[nix-shell:~/dev/openscad]$ cmake --build build
<success>
...

### DIFF
--- a/scripts/nix/flake.lock
+++ b/scripts/nix/flake.lock
@@ -49,22 +49,6 @@
         "type": "github"
       }
     },
-    "nixpkgs_2": {
-      "locked": {
-        "lastModified": 1761236834,
-        "narHash": "sha256-+pthv6hrL5VLW2UqPdISGuLiUZ6SnAXdd2DdUE+fV2Q=",
-        "owner": "nixos",
-        "repo": "nixpkgs",
-        "rev": "d5faa84122bc0a1fd5d378492efce4e289f8eac1",
-        "type": "github"
-      },
-      "original": {
-        "owner": "nixos",
-        "ref": "nixpkgs-unstable",
-        "repo": "nixpkgs",
-        "type": "github"
-      }
-    },
     "root": {
       "inputs": {
         "flake-parts": "flake-parts",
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by flaviut** [comment]
Even better, Nix has a way to do lockfiles, which will ensure everyone always gets the exact same version of dependencies. The current revision should work the same from now until the end of time.

I didn't start out doing it this way (with what's called "flakes") because flakes are a contentious topic in Nix-land and I didn't want to step on any toes. However they do solve the locking problem without any extra user input.

```
$ cd scripts/nix/
$ nix develop
develop$ cd ../../
develop$ rm -rf build
develop$ cmake -B build -DEXPERIMENTAL=1
develop$ nixGL ./build/openscad
```

The last step with the nixGL I think is optional because I'm on an unusual Asahi Linux system. But I made a note of it in the README anyway.

t-paul:
Sounds good to me, but I did not have any contact with Nix other than reading a bit about it. I believe @AaronVerDow has some additional insight as original author of the Nix setup.

PR #6448, opened 2025-12-21
chore: codegen version into manpage

flaviut:
testing process:

- created a VERSION.txt file
- ran the CMake build
- confirm that the output `build/doc/openscad.1` has the same version as
  VERSION.txt


kintel:
Thanks!

PR #6463, opened 2025-12-27
Move version info to .c file

flaviut:
This allows for better caching of objects, since the version.h include
will no longer change with each day/commit

image


CMakeLists.txt
```diff
@@ -29,6 +29,12 @@ if(POLICY CMP0167) # Since CMake 3.30
 endif()
 
 include("cmake/Modules/openscad_version.cmake")
+configure_file(
+  ${CMAKE_SOURCE_DIR}/src/version.c.in
+  ${CMAKE_BINARY_DIR}/version.c
+  @ONLY
+)
+set_source_files_properties(${CMAKE_BINARY_DIR}/version.c PROPERTIES LANGUAGE CXX)
```
kintel: Just make it version.cc instead of manually setting the LANGUAGE property?

flaviut: thank you! I'm not very familiar with CMake & C

kintel (approved):

PR #6464, opened 2025-12-27
Build appimage docker in pipeline

flaviut:
This moves the docker build for the appimage base container from a
separate Concourse CI to the main pipeline.

This new pipeline is more pure than before: there's no risk of using the
wrong base image, because the pipeline always build it's own base image.
The pipeline also now pins most of the custom build dependencies.

Performance in this pipeline is accomplished through caching. This makes
it easier to understand, because the developer can just pretend that
each step is executed every time the pipeline runs. The caching is
through the CircleCI docker layer cache, and through a multi-stage
build, where each stage only imports the dependencies that it actually
needs. Because each stage only imports the dependencies it actually
needs, we avoid unnecessary cache-busting.

This change (as well as some unrelated changes) [are currently running on my CircleCI org](https://app.circleci.com/pipelines/github/flaviut/openscad?branch=master).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -570,6 +570,18 @@ workflows:
           context: secret-context
           requires:
             - create-tarball
+      - publish-appimage-base-x86_64:
+          filters: *release-filters
+          context: secret-context
+      - publish-appimage-base-aarch64:
+          filters: *release-filters
+          context: secret-context
+      - publish-appimage-base-manifest:
+          filters: *release-filters
+          context: secret-context
+          requires:
+            - publish-appimage-base-x86_64
+            - publish-appimage-base-aarch64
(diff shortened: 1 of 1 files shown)


flaviut:
I missed the whole concept of needing to protect secrets from external users. hmm. thinking this through a little more.

flaviut:
nope! there's no problem there, since CCI only runs after PRs are merged. I did need to change things slightly to add a new workflow that runs on every commit -- I don't see a good reason to separate release & intermediate runs, and in fact I think it reduces reliability since we end up having less than 100% coverage.

The way I'd rather see it done is running both release & snapshot builds for each commit, and the only thing that happens that special on tagged releases is to update the website & Github releases. Build performance concerns should be addressed through caching instead.

PR #6466, opened 2025-12-27
CI: enable sccache

flaviut:
This avoids doing redundant work by caching compiler results. Because CI
does less work, we can run more redundant CI jobs and avoid the mental
overhead of manually de-duping jobs.

depends on https://github.com/openscad/openscad/pull/6463 to be effective.

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,25 +27,13 @@ executors:
 commands:
   # Provides OPENSCAD_VERSION and OPENSCAD_COMMIT
   establish_version:
-    parameters:
-      build_id:
-        type: string
-        default: ""
     steps:
       - run:
           name: Establish Version
           command: |
               source scripts/establish_version.sh
-              OPENSCAD_COMMIT=$(openscad_commit)
-              BASE_VERSION=$(openscad_version)
-              if [ -n "<< parameters.build_id >>" ]; then
-                  OPENSCAD_VERSION="${BASE_VERSION}.<< parameters.build_id >>"
-              else
-                  OPENSCAD_VERSION="${BASE_VERSION}"
-              fi
-              
-              echo "export OPENSCAD_COMMIT=$OPENSCAD_COMMIT" >> $BASH_ENV
-              echo "export OPENSCAD_VERSION=$OPENSCAD_VERSION" >> $BASH_ENV
+              echo "export OPENSCAD_COMMIT=$(openscad_commit)" >> $BASH_ENV
+              echo "export OPENSCAD_VERSION=$(openscad_version)" >> $BASH_ENV
   get_source:
     parameters:
       method:
@@ -114,8 +102,7 @@ jobs:
           method: << parameters.source_method >>
       - setup-sccache
       - restore-sccache-cache
-      - establish_version:
-          build_id: "ci${CIRCLE_BUILD_NUM}"
+      - establish_version
       - run:
           name: Build OpenSCAD Windows Application (64bit)
           no_output_timeout: 18000
@@ -168,8 +155,7 @@ jobs:
           method: << parameters.source_method >>
       - setup-sccache
       - restore-sccache-cache
-      - establish_version:
-          build_id: "ai"
+      - establish_version
       - run:
           name: Build OpenSCAD AppImage
           no_output_timeout: 18000
@@ -258,8 +244,7 @@ jobs:
                     npm init -y
                     npm i puppeteer
                     npx puppeteer browsers install chrome
-      - establish_version:
-          build_id: "wasm"
+      - establish_version
       - run:
           name: Build OpenSCAD WASM (<< parameters.openscad_platform >>)
           no_output_timeout: 18000


.circleci/config.yml
```diff
@@ -133,12 +166,15 @@ jobs:
     steps:
       - get_source:
           method: << parameters.source_method >>
+      - setup-sccache
+      - restore-sccache-cache
       - establish_version:
-          build_id: "ai${CIRCLE_BUILD_NUM}"
```
flaviut: build inputs should always be deterministic. if they are not, as this was, then the caching is broken.

t-paul: This is mainly used to help establish a reference in bug reports. Maybe using the first 4 or 5 characters of the git hash is fine, or we use something like YYMMDD of the last commit. The latter maybe not as safe, but easier to use from a screenshot. Both should be compatible with reproducible builds.

flaviut: this should already be the case: the `establish_version` step is defined as

```
              source scripts/establish_version.sh
              OPENSCAD_COMMIT=$(openscad_commit)
              BASE_VERSION=$(openscad_version)
              if [ -n "<< parameters.build_id >>" ]; then
                  OPENSCAD_VERSION="${BASE_VERSION}.<< parameters.build_id >>"
              else
                  OPENSCAD_VERSION="${BASE_VERSION}"
              fi
```

which in turn grabs it from the latest commit:

```
# Functions to retrieve version and commit information
#
# Usage:
#   source scripts/establish_version.sh
#   export OPENSCAD_VERSION=$(openscad_version)
#   export OPENSCAD_COMMIT=$(openscad_commit)

openscad_version() {
    local VERSION=""
    # /VERSION.txt is a file in the root folder containing a single version identifier
    if [ -f "$VERSION_FILE" ]; then
        # Read version identifier, removing any whitespace
        VERSION=$(tr -d '[:space:]' < "$VERSION_FILE")
        if [ -n "$VERSION" ]; then
            echo "Found VERSION.txt file, using VERSION=$VERSION" >&2
            echo "$VERSION"
            return 0
        fi
    fi

    # Default to YYYY.MM.DD of current day
    VERSION=$(date "+%Y.%m.%d")
    echo "No VERSION.txt file found, defaulting VERSION=$VERSION" >&2
    echo "$VERSION"
}

openscad_commit() {
    local COMMIT=""
    # /COMMIT.txt is a git commit
    if [ -f "$COMMIT_FILE" ]; then
        # Read git commit, removing any whitespace
        COMMIT=$(tr -d '[:space:]' < "$COMMIT_FILE")
        if [ -n "$COMMIT" ]; then
            echo "Found COMMIT.txt file, using COMMIT=$COMMIT" >&2
            echo "$COMMIT"
            return 0
        fi
    else
        # Try to get from git
        if command -v git >/dev/null 2>&1; then
            # Check if we are in a git repo
            if git -C "$ROOT_DIR" rev-parse --is-inside-work-tree >/dev/null 2>&1; then
                COMMIT=$(git -C "$ROOT_DIR" log -1 --pretty=format:"%h")
                echo "No COMMIT.txt file found, using git commit: $COMMIT" >&2
                echo "$COMMIT"
                return 0
            fi
        fi
    fi
}
```

.circleci/config.yml
```diff
@@ -133,12 +166,15 @@ jobs:
...
+          build_id: "ai"
```
kintel: We can probably just drop the build ID - using a constant doesn't really make much sense. Same in the other places.

.circleci/config.yml
kintel: @t-paul I believe all builds (including release builds) now have the git hash exposed in the About dialog and Help->Library Info: `OpenSCAD Version: 2025.12.26 (git cfdd775c4)`

kintel:
I'm not familiar with sccache, but some rather open-ended questions:
* How does sccache react to environment changes (e.g. new base image, different compiler version, updated libraries)?
* Cache key is currently `sccache-cache-stable-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ epoch }}`
   * If sccache already know about the compiler, is `{{ arch }}` redundant?
   * Why does the job name make a difference? `{{ .Environment.CIRCLE_JOB }}`

Update: The reason I ask is that we need to manually invalidate the library build cache on macOS as the cache doesn't have a way of knowing all the system parameters.

flaviut:
> How does sccache react to environment changes (e.g. 

> new base image,

Includes would change, and [sccache works post-preprocessor](https://github.com/mozilla/sccache/blob/main/docs/Caching.md#cc-compiler). The cache would be busted

> different compiler version,

sccache uses [the compiler binary's hash in its hash](https://github.com/mozilla/sccache/blob/main/docs/Caching.md#cc-compiler)

> updated libraries)?

I'm pretty sure the link stage doesn't get cached. But if it does, I'd expect the hashes to be different due to different headers, and if that wasn't the case, I'd expect the libraries to be the "Extra files that need to have their contents hashed" from the docs.

> Update: The reason I ask is that we need to manually invalidate the library build cache on macOS as the cache doesn't have a way of knowing all the system parameters.

I'm extremely unfamiliar with how macos works. But skimming the source code [I see a reference to `MACOSX_DEPLOYMENT_TARGET`](https://github.com/mozilla/sccache/blob/5b4a55786717ba155dcd137eb207bb6617b55d23/src/compiler/c.rs#L1434) as part of the cache key. Is this the kind of thing you're looking for?

---

> Cache key is currently sccache-cache-stable-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ epoch }} 

I tried a few things to get caching to work without any success, [until I found this blog post and took that from there](https://medium.com/@edouard.oger/rust-caching-on-circleci-using-sccache-c996344f0115). I still don't feel like I understand how CircleCI caching works, but this approach does work.

> If sccache already know about the compiler, is {{ arch }} redundant?

It reduces the amount of data stored in the cache, which makes it faster to save and restore. [The cache is fetched on job startup over the network.](https://circleci.com/docs/guides/optimize/caching/#cache-size)

> Why does the job name make a difference? {{ .Environment.CIRCLE_JOB }}

Caches [are shared between jobs](https://circleci.com/docs/guides/optimize/caching/#writing-to-the-cache-in-workflows).

flaviut:
Yup, confirmed that sccache doesn't handle link commands:

```
❯ SCCACHE_LOG=trace sccache gcc file1.o file2.o -o myprogram -lm
...
[2025-12-28T04:31:06Z DEBUG sccache::commands] Server sent UnhandledCompile
[2025-12-28T04:31:06Z TRACE sccache::commands] running command: Some(cd "/home/user" && "/home/user/etc/.local/bin/gcc" "file1.o" "file2.o" "-o" "myprogram" "-lm")
/usr/bin/ld: cannot find file1.o: No such file or directory
/usr/bin/ld: cannot find file2.o: No such file or directory
collect2: error: ld returned 1 exit status
```

Because there's no `-c` flag, it's failing to execute [this code path to enable caching](https://github.com/mozilla/sccache/blob/5b4a55786717ba155dcd137eb207bb6617b55d23/src/compiler/gcc.rs#L350C18-L354).

kintel:
> 
> macOS [...] as part of the cache key. Is this the kind of thing you're looking for?
> 
No, unrelated; we don't cache using sccache on macOS; we have a custom setup for building dependencies as we don't have Docker on macOS..

> > Cache key is currently sccache-cache-stable-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ epoch }}
> 
> I tried a few things to get caching to work without any success, [until I found this blog post and took that from there](https://medium.com/@edouard.oger/rust-caching-on-circleci-using-sccache-c996344f0115). I still don't feel like I understand how CircleCI caching works, but this approach does work.
> 
One possible failure mode:

* We store caches for every build by writing `{{ epoch }}` as part of the key, but we read caches without the epoch, meaning we'll pull the latest built cache with the same key prefix. This is fine.
* However, the CIRCLE_JOB in question is used for two purposes; release build and snapshot build, which have different build flags, and possibly different base images. What will happen if we write a cache for a snapshot build, then read the cache back during a release build?

flaviut:
It's necessary to think about the CircleCI cache & the sccache separately. CircleCI makes the cache available. sccache decides which values in the cache to use.

sccache also only operates on the preprocessed files, which means that all the includes have been copy-pasted in, and all the ifdefs have been applied. So, if the library includes have changed, then none of the caches will be used. If the ifdefs have changed, then the source files will be different and the caches will not be used.

sccache is in active use by the Mozilla build system, and their caches are simply a global S3 bucket shared between all jobs. They've definitely thought a lot about this and how to do it safely.

flaviut:
> we don't cache using sccache on macOS; we have a custom setup for building dependencies as we don't have Docker on macOS..

sccache is a separate and completely unrelated program from Docker, if that what you're suggesting. This PR doesn't include sccache for mac purely out of laziness on my part, since I expect that'll be a little harder than Linux, not because it can't work there.

kintel:
The macOS thing was a big of a digression - the main issue on macOS is that the system environment is not stable (we cannot pin all dependencies as we're dependent on the rolling Homebrew release + the upstream GitHub VM image). Both are probably solvable but it might need more mechanisms than just our own sccache. Let's shelve that until the Linux-based builds are working.

kintel:
OK, so my understanding is as follows (for Linux builds), highlighting my attempt at breaking the cache logic:
* If we run an `openscad-appimage` job with `openscad_build_type=snapshot`, it will use and create a new cache entry
* If we immediately afterwards run the same `openscad-appimage` job with `openscad_build_type=release`, it will load the recently generated cache, but since the CMake parameters are different, and different compiler options are used, the cache is blown and we have to rebuild.

The above scenario will always result in a correct build, but it will invalidate the cache every time we switch build type?

I'm not 100% sure if my analysis is correct, but if it is, we may want to include the job parameters as a cache key.

Edit: #6412 is probably needed to understand the context :)

kintel:
Actually, I just realized that `CIRCLE_JOB` is not going to be `openscad-appimage`, but `openscad-appimage-<n>` where `<n>` is the permutation index of the parameter combination used.

flaviut:
> If we immediately afterwards run the same openscad-appimage job with openscad_build_type=release, it will load the recently generated cache, but since the CMake parameters are different, and different compiler options are used, the cache is blown and we have to rebuild.

Yup exactly. But the sccache is keyed by compiler arguments and source files, so it doesn't necessarily mean the whole cache is useless. As it is now, changing the cmake flags leads to a total rebuild:

```
❯ cmake -B build -DCMAKE_C_COMPILER_LAUNCHER=sccache -DCMAKE_CXX_COMPILER_LAUNCHER=sccache
❯ cmake --build build -j --config RelWithDebInfo
❯ sccache --show-stats
Compile requests                     277
Compile requests executed            277
Cache hits                            37
Cache hits (C/C++)                    37
Cache misses                         240
Cache misses (C/C++)                 240
Cache hits rate                    13.36 %
Cache hits rate (C/C++)            13.36 %
Cache timeouts                         0
Cache read errors                      0
Forced recaches                        0
Cache write errors                     0
Cache errors                           0
Compilations                         240
Compilation failures                   0
Non-cacheable compilations             0
Non-cacheable calls                    0
Non-compilation calls                  0
Unsupported compiler calls             0
Average cache write                0.203 s
Average compiler                  15.290 s
Average cache read hit             0.152 s
Failed distributed compilations        0
Cache location                  s3, name: openscad-ci, prefix: /
Version (client)                0.12.0
❯ cmake -B build -DCMAKE_C_COMPILER_LAUNCHER=sccache -DCMAKE_CXX_COMPILER_LAUNCHER=sccache -DEXPERIMENTAL=1 -DSNAPSHOT=ON
❯ cmake --build build -j --config RelWithDebInfo
❯ sccache --show-stats
Compile requests                     520
Compile requests executed            520
Cache hits                            37
Cache hits (C/C++)                    37
Cache misses                         483
Cache misses (C/C++)                 483
Cache hits rate                     7.12 %
Cache hits rate (C/C++)             7.12 %
Cache timeouts                         0
Cache read errors                      0
Forced recaches                        0
Cache write errors                     0
Cache errors                           0
Compilations                         483
Compilation failures                   0
Non-cacheable compilations             0
Non-cacheable calls                    0
Non-compilation calls                  0
Unsupported compiler calls             0
Average cache write                0.203 s
Average compiler                  14.371 s
Average cache read hit             0.152 s
Failed distributed compilations        0
Cache location                  s3, name: openscad-ci, prefix: /
Version (client)                0.12.0
```

but with some care (in a different PR) I don't see why we wouldn't be able to get some cache hits on at least the stuff under `submodules/` that does not care about the experimental/snapshot status.

kintel:
Ah, of course - the cache has potentially multiple entries per source file

.circleci/config.yml
```diff
@@ -153,20 +175,23 @@ jobs:
...
+                  -DCMAKE_CXX_COMPILER_LAUNCHER=sccache \
                   -DUSE_BUILTIN_OPENCSG=ON \
                   -DMANIFOLD_PYBIND=OFF \
                   -DMANIFOLD_TEST=OFF \
                   -DENABLE_PYTHON=ON \
                   -DENABLE_TESTS=OFF \
                   $CMAKE_EXTRA_ARGS
-              make -j4
-              make install DESTDIR=../AppDir
+              cmake --build . -j 2  # using all 4 cores OOMs
```
kintel, line 186: It's a bit odd that the `large` Linux resource class made available only has 8 GB of RAM, while the macOS one they give us has 28 GB, which explains why the macOS builds are faster (more parallelization possible)..

PR #6471, opened 2025-12-29
print() module, format() function, quote() function

### LINKED ISSUE #6385: New output/formatting suite:  print(), quote(), format()

jordanbrown0:
Fixes #6385.  Or at least it will when done.

Implements print() and quote(), but only with one argument each, and print() only prints strings.  I'm not clear on whether you need more than that; it's simple and stupid an inflexible, but keeps the semantics simple.

Does not yet implement format().

### DIFF
--- a/src/core/builtin_functions.cc
+++ b/src/core/builtin_functions.cc
@@ -1478,12 +1478,12 @@ class Format
         return;
       }
       break;
-    default:
-      break;
+    default: break;
     }
     LOG(message_group::Warning, loc, args.documentRoot(), "format code '%1$s' not defined for '%2$s'",
         type, a.typeName());
   }
+
 public:
   Format(const std::string& fmt, const Arguments& args, int argStart, const Location& loc)
     : args(args), argStart(argStart), loc(loc), curArg(0)
@@ -1510,9 +1510,7 @@ public:
 
     result = stream.str();
   }
-  const std::string str() const {
-    return result;
-  };
+  const std::string str() const { return result; };
 };
 
 Value builtin_format(Arguments arguments, const Location& loc)


jordanbrown0:
Failures were all related to changing the format emitted by echo/str of an object.  That's not really part of this project, but it would be nice if quote() would emit a parsable form for objects.  See also #6472.  What I've got implemented at the moment emits `object()` calls.

Typical:
```
-ECHO: { array-mixed = ["one", 1, false]; array-string = ["one", "two", "three"]; array_number = [0, 1, 2, 3, 5]; bool = true; float = 3.5e-10; number = 2; object = { name = "The object name"; nested = { value = 42; }; number = 4; }; string = "hallo world!"; }
+ECHO: object([["array-mixed", ["one", 1, false]], ["array-string", ["one", "two", "three"]]], array_number=[0, 1, 2, 3, 5], bool=true, float=3.5e-10, number=2, object=object(name="The object name", nested=object(value=42), number=4), string="hallo world!")
```

jordanbrown0:
Oh, and that I made QuotedString (which already processes \r, \n, et cetera) emit non-ASCII as \uHHHH or \UHHHHHH, 
and so an AST dump with a non-ASCII character in it changed.

```
-echo([for(a = "a↑b😀") (a)]);
+echo([for(a = "a\u2191b\U01f600") (a)]);
```

jordanbrown0:
That last push added the beginnings of support for format().  It accepts `{{` and `}}` for quoted braces, `{n}` for a positional argument, `{name}` for a named argument, and combinations of `.name` and `[name-or-index]` to access members of a particular argument.

Here's a quick demo:
```
obj = object(first="Fred", last="Flintstone");
print(format("first: {0.first}, last: {0.last}", obj));
print(format("first: {rec.first}, last: {rec.last}", rec=obj));
```
yields
```
first: Fred, last: Flintstone
first: Fred, last: Flintstone
```

jordanbrown0:
That last push implements enough that it might fool some people.

```
print(format("basic: {} {}", 123, "abc"));
print(format("padding: {:10d} {:10s}", 123, "abc"));
print(format("left align: {:<10d} {:<10s}", 123, "abc"));
print(format("left align with fill: {:*<10d} {:*<10s}", 123, "abc"));
print(format("right align with fill: {:*>10d} {:*>10s}", 123, "abc"));
print(format("floating point: {0:10.4f} {0:10.4e} {0:10.4g} {0:10.4}", 123.456));
print(format("any arg order: {2} {1} {1} {0}", "last", "twice", "first"));
print(format("args by name: last={last}, first={first}", first="Fred", last="Flintstone"));
print(format("extract members from object: last={0.last}, first={0.first}", object(first="Fred", last="Flintstone")));
print(format("args by name, from objects: {name.first} {name.last} is married to {spouse.first} {spouse.last}.",
    name=object(first="Fred", last="Flintstone"),
    spouse=object(first="Wilma", last="Flintstone")));
print(format("character: {0:x} {0:c}", 65));
print(format("non-ASCII character: {0:x} {0:c}", 0x1f983));
print(format("variable width: {:{}d}", 123, 10));
```

yields:
```
basic: 123 abc
padding:        123        abc
left align: 123        abc       
left align with fill: 123******* abc*******
right align with fill: *******123 *******abc
floating point:   123.4560 1.2346e+02      123.5      123.5
any arg order: first twice twice last
args by name: last=Flintstone, first=Fred
extract members from object: last=Flintstone, first=Fred
args by name, from objects: Fred Flintstone is married to Wilma Flintstone.
character: 41 A
non-ASCII character: 1f983 🦃
variable width:        123
```

Known gaps:
* Booleans
* Precision doesn't work for strings
* Vectors, ranges, objects, and functions can be formatted, but just with width, fill; otherwise, formatting is basic to-string
* No ^ alignment
* Fill character must be ASCII
* "b" (binary) format code
* Not sure about "d" and default formatting for numbers.
* Probably lots more.

Implementation is very rough.

jordanbrown0:
One of the questions that I was uncertain about was how "d" and default formatting should work, given that Python and C++ versions of this function have distinct integer and float types to work with, while we have only numbers.

Right now my theory (not implemented) is that "d" (and "o" and "x") would round, and then would express the number using as many digits as required, never switching over to exponential notation.  (Note that for numbers larger than about 10^16, the final digits would be noise.)

Default formatting, on the other hand - `{}`, with no format code - would use "g" format, with some TBD default precision.  The Python default for "g" is 6 digits, which is to say that 123,456 is expressed as `123456` but 1,234,567 is expressed as `1.23457e+06`.  The Python default for an unspecified format seems to be equivalent to `{:.17}`, which in turn seems to be similar to (but not identical to) `{:.16g}` or `{:.17g}`.  I need to go reread the Python docs in detail.  But for discussion purposes, let's say that we make the default be "g" format, and the default precision be 17, which is at least close to the Python behavior.

jordanbrown0:
I see that in Python the default format for a floating point number is like g, but always emits at least one digit after the decimal point.  Thus in Python `"{}".format(1)` is `1`, but `"{}".format(1.)` is `1.0`.  We can't distinguish those two cases, so I think that we should **not** have it emit that `.0`.

jordanbrown0:
One question for this project... we've discussed introducing methods on built-in data types like "string".  As currently implemented, this project introduces new functions `quote()` and `format()`, but perhaps those should instead be methods on strings, avoiding introducing on the normal function namespace.  That is, perhaps they should be `"hello\nworld".quote()` and `"{}".format(123)`.

jordanbrown0:
* Q.  Should `print()` accept more than one argument?  If it does, exactly what should it do with them?  (Probably:  separate with a single space.)  My answer:  probably yes.
* Q.  Should `print()` accept non-strings as arguments, and if so what should it do with them?  (Probably:  toString(), but that leaves the question of how strings embedded in aggregate types should be processed.)  My answer:  uncertain, because of the behavior of aggregates.  Maybe limit it to scalars.
* Q.  Should quote() accept more than one argument?  What should it do with them?  (Probably:  separate with comma, maybe comma-space, so that it forms a syntactically-legal list of them, except for not having brackets.)  My answer:  don't really care.
* Q.  Should quote() accept named arguments, and dump them in name=value form?  My answer:  probably not; for one thing, it precludes named parameters to control the behavior.
* Q.  Should quote use REPR mode (Unicode as-is) or ASCII mode (Unicode encoded as \uHHHH or \UHHHHHH)?  My answer:  REPR mode; let's not perpetuate English-centric behavior.  (But:  should probably try to detect more non-ASCII control code points.)
* Q.  When dumping a string, should AST dump use REPR mode or ASCII mode?  My answer:  REPR mode, as above.
* Similarly for CSG dump.

PR #6473, opened 2025-12-29
CI: enable sccache

flaviut:
This avoids doing redundant work by caching compiler results. Because CI does less work, we can run more redundant CI jobs and avoid the mental overhead of manually de-duping jobs.

replacement for https://github.com/openscad/openscad/pull/6466

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,25 +27,13 @@ executors:
 commands:
   # Provides OPENSCAD_VERSION and OPENSCAD_COMMIT
   establish_version:
-    parameters:
-      build_id:
-        type: string
-        default: ""
     steps:
       - run:
           name: Establish Version
           command: |
               source scripts/establish_version.sh
-              OPENSCAD_COMMIT=$(openscad_commit)
-              BASE_VERSION=$(openscad_version)
-              if [ -n "<< parameters.build_id >>" ]; then
-                  OPENSCAD_VERSION="${BASE_VERSION}.<< parameters.build_id >>"
-              else
-                  OPENSCAD_VERSION="${BASE_VERSION}"
-              fi
-              
-              echo "export OPENSCAD_COMMIT=$OPENSCAD_COMMIT" >> $BASH_ENV
-              echo "export OPENSCAD_VERSION=$OPENSCAD_VERSION" >> $BASH_ENV
+              echo "export OPENSCAD_COMMIT=$(openscad_commit)" >> $BASH_ENV
+              echo "export OPENSCAD_VERSION=$(openscad_version)" >> $BASH_ENV
   get_source:
     parameters:
       method:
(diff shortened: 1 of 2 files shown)


kintel:
Somehow `CircleCI branch build (PR)` didn't get triggered on this PR. I don't quite understand why..

flaviut:
It's getting triggered in my CircleCI: https://app.circleci.com/pipelines/github/flaviut/openscad?branch=ci-enable-sccache

kintel:
> It's getting triggered in my CircleCI: https://app.circleci.com/pipelines/github/flaviut/openscad?branch=ci-enable-sccache

Huh, interesting - did you set up your own CircleCI connection to your fork?
It's supposed to run and report back here, like for other PRs, see e.g. https://github.com/openscad/openscad/pull/6471/checks

flaviut:
Yup, I set up my own CircleCI so I could test configuration changes without bothering y'all. It has my own Docker Hub credentials for example.

kintel:
We still want build failures to surface here so that we can block the PR on it. May not be critical right now, but something we should figure out.

t-paul:
I did some local testing with `sccache`, including the distributed build feature via an external server, and this looks nice so far :+1:. It lacks a bit of control over the build process (or I'm not able to find the needed configuration yet), but for the CI use case that's not relevant.

flaviut:
The main issue for the CI usecase is that sccache assumes that all clients are trustworthy. That's not the case for incoming PRs. So I'd suggest that

1. Distributed builds are a complete no-go for PR builds. Sandboxing properly is difficult and more trouble than it is worth.
2. PR builds could use a cache, but that cache must be segregated from the trusted cache running on master & release tags.

t-paul:
Like I said, distributed is not relevant here, that was just an additional note.

For the cache we maybe can add a configured secret as part of the key. That is not available to any PR build from external sources. It would be nice if we could find a way of actually validating that.

flaviut:
Reading too fast, my bad!

Yes: for a remote cache, there would need to be an authentication key to push/pull from the cache, and that'd need to be a CircleCI env config. Otherwise, internet bots would quickly notice and upload all sorts of unsavory and/or illegal content. Even then, it's impossible to prevent a motivated attacker leaking the cache secrets. If that was to happen, we'd just need to rely on banning the attacker and rolling the key. 

Cache secrets aren't enough though, because CI scripts can contain arbitrary code. It would be possible to for example guess the input hash for the next release's `version.cc` job, and to upload a fake malicious `version.o` file for it ahead of time in a PR. Then when the release build is triggered, there's a cache hit and we have malicious code in there without any sign of it in the source code. PR builds have to be fully separated.

t-paul:
Yes, and the separation might be possible by using a secured context variable (https://circleci.com/docs/guides/security/contexts/). Not accidentally leaking those can be a challenge though, maybe there's a more direct way of checking via some internal API or other predefined environment variable.

flaviut:
If you want to get very fancy, and you're willing to either host a separate authentication service or add some sort of custom auth plugin to your self-hosted S3 server, you can use the CircleCI OIDC key. [See this job](https://app.circleci.com/pipelines/github/flaviut/openscad/39/workflows/872d23c5-c624-4adf-86c3-e259e861235b/jobs/272), and the decoded result which the cache could assert against:

```
{
  "aud": "61a7ebbb-6ff4-4c1f-88ad-995259b6ba5e",
  "exp": 1767217968,
  "iat": 1767214368,
  "iss": "https://oidc.circleci.com/org/61a7ebbb-6ff4-4c1f-88ad-995259b6ba5e",
  "oidc.circleci.com/context-ids": [],
  "oidc.circleci.com/job-id": "c58f7ae2-352e-4b2d-892f-d4e0a348bf8d",
  "oidc.circleci.com/org-id": "61a7ebbb-6ff4-4c1f-88ad-995259b6ba5e",
  "oidc.circleci.com/pipeline-definition-id": "351af186-3e67-579c-bb82-98a349a0ad55",
  "oidc.circleci.com/pipeline-id": "08e57f5c-f175-4dad-9b45-d83eb63400f9",
  "oidc.circleci.com/project-id": "4a68d490-1f36-4956-b06c-e446fe950dd6",
  "oidc.circleci.com/ssh-rerun": false,
  "oidc.circleci.com/vcs-origin": "github.com/flaviut/openscad",
  "oidc.circleci.com/vcs-ref": "refs/heads/oidc-test",
  "oidc.circleci.com/workflow-id": "872d23c5-c624-4adf-86c3-e259e861235b",
  "sub": "org/61a7ebbb-6ff4-4c1f-88ad-995259b6ba5e/project/4a68d490-1f36-4956-b06c-e446fe950dd6/user/0cc6e35a-6fe3-4e31-beab-ef34fb9b7874/vcs-origin/github.com/flaviut/openscad/vcs-ref/refs/heads/oidc-test"
}
```

This would make the tokens *very* short lived, and completely delegate key management to circleci. For example, take a look at the key from that job, and put it into https://www.jwt.io/ along with the public key from https://oidc.circleci.com/org/61a7ebbb-6ff4-4c1f-88ad-995259b6ba5e/.well-known/jwks-pub.json

image

flaviut:
Rebased onto latest master. As far as I can see there are no remaining blockers here.

kintel (approved):
LGTM. Giving @t-paul a chance to comment before merging.

PR #6474, opened 2025-12-29
Build appimage docker in pipeline

flaviut:
This moves the docker build for the appimage base container from a separate Concourse CI to the main pipeline.

This new pipeline is more pure than before: there's no risk of using the wrong base image, because the pipeline always build it's own base image. The pipeline also now pins most of the custom build dependencies.

Performance in this pipeline is accomplished through caching. This makes it easier to understand, because the developer can just pretend that each step is executed every time the pipeline runs. The caching is through the CircleCI docker layer cache, and through a multi-stage build, where each stage only imports the dependencies that it actually needs. Because each stage only imports the dependencies it actually needs, we avoid unnecessary cache-busting.

replaces https://github.com/openscad/openscad/pull/6464

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -583,7 +583,7 @@ workflows:
             - openscad-tarball-release
       - openscad-appimage:
           # for openscad-ansible/.../circleci-download-artifacts.py
-          name: openscad-appimage-64bit
+          name: openscad-appimage-release
           filters: *release-filters
           openscad_build_type: release
           source_method: tarball


scripts/docker/appimage-base/Dockerfile
kintel, line 1: As @t-paul suggested, let's try to put dockerfiles in a separate folder, mimicking the folder structure in https://github.com/openscad/docker-openscad, to make it easier to manage/recognize where this fits in the bigger picture of docker files.

.circleci/config.yml
```diff
@@ -511,6 +598,19 @@ workflows:
...
+          context: secret-context
+          requires:
+            - openscad-appimage-release
+      - openscad-mxe:
+          filters: *release-filters
+          openscad_build_type: release
+          source_method: tarball
+          context: secret-context
+          requires:
+            - create-tarball
```
kintel: Duplicate of above

.circleci/config.yml
```diff
@@ -238,7 +313,7 @@ jobs:
                   -DOPENSCAD_VERSION="$OPENSCAD_VERSION" \
                   $CMAKE_EXTRA_ARGS
               cd ../build
-              make -j4 VERBOSE=1
+              cmake --build .
```
kintel, line 314: How do we know that cmake selects a parallelization for the available memory that doesn't cause OOM?

.circleci/config.yml
```diff
@@ -10,8 +10,18 @@ executors:
...
+  docker-x86_64:
+    machine:
+      image: ubuntu-2204:current
+      docker_layer_caching: true
+    resource_class: large
+  docker-aarch64:
+    machine:
+      image: ubuntu-2204:current
+      docker_layer_caching: true
+    resource_class: arm.medium
```
kintel: arm.large should be usable here, unless you downgraded to medium for a reason (large is really only interesting if there is parallelization)

.circleci/config.yml
```diff
@@ -10,8 +10,18 @@ executors:
...
     docker:
-      - image: openscad/appimage-x86_64-base:latest
+      - image: ${CIRCLE_PROJECT_USERNAME}/appimage-base:<< pipeline.git.revision >>
     resource_class: large
+  docker-x86_64:
+    machine:
+      image: ubuntu-2204:current
+      docker_layer_caching: true
+    resource_class: large
+  docker-aarch64:
```
kintel: We tend to use arm64 for this arch, but I'm not familiar with conventions used elsewhere..

.circleci/config.yml
```diff
@@ -10,8 +10,18 @@ executors:
     resource_class: large
   appimage-executor:
     docker:
-      - image: openscad/appimage-x86_64-base:latest
+      - image: ${CIRCLE_PROJECT_USERNAME}/appimage-base:<< pipeline.git.revision >>
```
kintel: Further down you use ${CIRCLE_SHA1} to get the revision. Should we use the same everywhere for clarity?

.circleci/config.yml
```diff
@@ -420,6 +495,22 @@ filters: &release-filters
...
+      - publish-appimage-base-x86_64:
+          context: secret-context
+      - publish-appimage-base-aarch64:
+          context: secret-context
+      - publish-appimage-base-manifest:
+          context: secret-context
+          requires:
+            - publish-appimage-base-x86_64
+            - publish-appimage-base-aarch64
+      - openscad-appimage:
```
kintel: Legacy name is still needed for us to identify this build when publishing snapshots on the website
```
          # Legacy name
          name: openscad-appimage-64bit
```

See https://github.com/openscad/openscad-ansible/blob/main/roles/webserver/files/docker-images/maintenance/bin/circleci-download-artifacts.py

kintel: Btw., why not sign this?

scripts/docker/appimage-base/Dockerfile
```diff
@@ -0,0 +1,197 @@
...
+
+# Install Lib3MF specific dependencies
+# uuid-dev, libssl-dev etc. are required for Lib3MF
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+        uuid-dev \
+        && rm -rf /var/lib/apt/lists/*
+
+# 1.8.1
+ARG LIB3MF_COMMIT=ff914b9fae76cd6a365952728f44f8d79a9f8f75
```
kintel, line 36: Can we bump this to V2 while at it, or are we waiting to resolve the IDL compiler source code issue? @t-paul

scripts/docker/appimage-base/Dockerfile
```diff
@@ -0,0 +1,197 @@
...
+# CGAL often needs GMP/MPFR (though sometimes header-only, full builds need them)
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+        libgmp-dev \
+        libboost-dev \
+        libmpfr-dev \
+        && rm -rf /var/lib/apt/lists/*
+
+# v5.6.3
+ARG CGAL_COMMIT=3654f780ae0c64675cabaef0e5ddaf904c48b4b7
```
kintel, line 72: 5.6.3 is getting old. Can we bump this, and perhaps use a tag instead of commit hash?

scripts/docker/appimage-base/Dockerfile
t-paul: I think the IDL stuff is only relevant for the upstream packaging, e.g. the Flatpak uses v2.4.1 with a [patch](https://github.com/3MFConsortium/lib3mf/pull/413) that was merged to [v2.5.0-alpha](https://github.com/3MFConsortium/lib3mf/releases/tag/v2.5.0-alpha) but is not released yet. Long story short it's worth trying if it "just builds".

scripts/docker/appimage-base/Dockerfile
t-paul: CGAL has a tendency to break build compatibility pretty quick, so that too is a matter of trying it out. Official Ubuntu 22.04 has an even older 5.4.

.circleci/config.yml
flaviut: We don't, but in practice, it works. For ["Build OpenSCAD AppImage"](https://github.com/openscad/openscad/pull/6474/changes/BASE..4fcc252933a157a9c610decf0795b716b48e84c2#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R172) I tried first with the default parallelization, and it failed with OOM, at which point I decreased it and it worked.

The default [is the number of CPUs the system has](https://ninja-build.org/manual.html#:~:text=based%20by%20default%20on%20the%20number%20of%20CPUs%20your%20system%20has).

.circleci/config.yml
flaviut: Signing is basically the maintainers putting their stamp of approval on the code. I would not want random PRs or random commits to be signed, especially since PRs could contain malicious code.

scripts/docker/appimage-base/Dockerfile
kintel: sg, we can do that separately from this PR

scripts/docker/appimage-base/Dockerfile
kintel: We're using CGAL 6.1 on macOS, so there's been a fair amount of testing on newer CGALs.

flaviut: I think a big part of the utility of this PR is the fact that such changes don't require multi-repo changes and can fit inside a single PR.

With that in mind, I'd like to put off incrementing any dependencies until a future PR. I'll revert the docker base image to 22.04 as well.

.circleci/config.yml
kintel: This is meant to be a distributable binary going out to users though (or the one you removed from "CircleCI snapshot build (master)" was).

Perhaps we need to write up a doc on how We build and distribute binaries :)

scripts/docker/appimage-base/Dockerfile
kintel: Agreed, let's aim for no functional changes in this PR :)

t-paul: Yeah, lets first do the move and have separate updates for the other stuff. I can look at the two upgrades (lib3mf and CGAL) once the move happened.

I missed the 22.04 to 24.04 change. As mentioned before, that's currently a no-go in my view as 22.04 has support till Jun 2027 and I'd suggest not dropping support for those distros yet. That's an inherent property of the AppImage functionality:

[Build on old systems, run on newer systems](https://docs.appimage.org/introduction/concepts.html#id4)
> It is considered best practice to develop and compile the application on the oldest still-supported Linux distribution that we can assume users to still use. For example, the oldest still-supported LTS release of Ubuntu is a good choice to develop applications against and build applications on.

Maybe we don't need to wait till 2027 but I would at least advocate to always support the last 2 LTS version, so we could consider dropping 22.04 support once 26.04 LTS is out.

t-paul: So CGAL 6.1 seems to work fine, same with lib3mf 2.4.1.

We may have a packaging issue though, the "runtime 2.4.0" indicates it's picking up my local install, not the lib from the build.

```
OpenSCAD Version: 2025.12.30.ai (git cfdd775c4)
User Agent: OpenSCAD/2025.12.30.ai (git cfdd775c4) (Linux x86_64; Debian GNU/Linux forky/sid)
Compiler: GCC "13.3.0" 64bit
CGAL version, kernels: 6.1, Cartesian<Gmpq>, Extended_cartesian<Gmpq>
Python Version: Python 3.12.3
Qt version: 5.15.13
QScintilla version: 2.14.1
freetype version: 2.13.2 (runtime: 2.13.3)
harfbuzz version: 8.3.0 (runtime: 12.2.0)
cairo version: 1.18.0
lib3mf version: 2.4.1 (runtime: 2.4.0)
```

.circleci/config.yml
t-paul: External PRs do not have access to secrets, so they are not able to do signing.

scripts/docker/appimage-base/Dockerfile
t-paul: And also:
```
Qt version: 6.4.2
```

.circleci/config.yml
kintel: right, so we should make this adhere to the PR workflow vs. snapshot workflow from master that we already have in this setup. @t-paul If you didn't start on the CI doc refresher yet I can take a stab at it.

scripts/docker/appimage-base/Dockerfile
kintel: Where does  Qt 6.4.2 come from? I feel that if we want to move to Qt6, it would be nice to select an LTS release (6.5, 6.8 or 6.10).

.circleci/config.yml
t-paul: It's only listing the platforms so far: https://github.com/openscad/openscad/wiki/CI-Setup-(new)

scripts/docker/appimage-base/Dockerfile
t-paul: Not sure yet. Base Ubuntu 22.04 is Qt6.2. I think I saw a "backports" repo somewhere but have not checked in detail.

.circleci/config.yml
flaviut: I don't have permission to view that repo. Can you also double-check the name for the release appimage build?

I'm applying this change. However, just to avoid any possible confusion between master & PR jobs, I would prefer for this to be `snapshot-appimage-x86_64`.

flaviut:
not yet ready for another round of review -- I still need to test and tweak a little bit next year. Happy new years y'all!

kintel:
One general comment: The config.yml is getting pretty large. I'm wondering if pulling out some tools into importable "orbs" would help. Out of scope for this PR; I'm just curious.

flaviut:
Works in Ubuntu 22.04 LTS:

image

What's left:

- [ ] We need to separate PR base image docker pushes from master docker pushes. It seems [like the only way to do this is by creating a new Docker hub account](https://github.com/orgs/community/discussions/153160#discussioncomment-12409658).

kintel:
Not sure I understand the remaining work: Do PRs need to push to docker hub, or can we just keep the image local for the next job to use? ..or does CircleCI mandate pulling job images from a remote registry?

flaviut:
CircleCI mandates pulling job images from a remote registry when using docker executors. So therefore PRs need to push to docker hub.

But now that you mention it, it is possible to manually run docker commands instead. This is probably the right way to do it, and never actually push anything. Combine the docker image build and the compilation into a single circleci job. I'll change this PR to do that instead, since that's much simpler infrastructure and much easier to test locally.

flaviut:
With https://github.com/flaviut/openscad/commit/f539ed097880f2e8039c2af1b8d9e88bb1d8e84c, it's working without any docker push. I'll add that commit to this PR once https://github.com/openscad/openscad/pull/6473 is resolved so I don't have to resolve merge conflicts again.

kintel:
FYI: The coverage failures seem to be unrelated to this PR. Feels like something environmental broke, but I have no idea what

.circleci/config.yml
```diff
@@ -183,9 +199,10 @@ jobs:
                   -DENABLE_PYTHON=ON \
                   -DENABLE_TESTS=OFF \
                   $CMAKE_EXTRA_ARGS
-              cmake --build . -j 2  # using all 4 cores OOMs
+              cmake --build . -j
```
flaviut, line 197: the non-docker machine actually has significantly more RAM than the docker machine, so this works fine

Image

.circleci/config.yml
```diff
@@ -444,51 +492,31 @@ jobs:
...
-  tags:
-    only: /^openscad-.*$/
-  branches:
-    ignore: /.*/
-
 workflows:
   version: 2
   "CircleCI snapshot build (master)":
+    when:
+      equal: [ master, << pipeline.git.branch >> ]
```
flaviut: equivalent to the filter above, and avoids duplicating this so much.

.circleci/config.yml
kintel: Yeah, I assume they charge more for the docker-builder machines, but I don't think we need to worry, at least not after all the optimizations :)

.circleci/config.yml
kintel: Actually, they're different. `release-filters` filter on specific tags.

I don't know if `when:  equal: [ master, << pipeline.git.branch >> ]` is identical to `filters:  branches:  only: - master`, but I'll look into it.

flaviut: alright, I'll revert (and maybe make a separate PR for this)

flaviut:
Pretty sure this is what's happening with the coverage failures: https://github.com/openscad/openscad/issues/6508

With CI, it's important to be incredibly obsessive about pinning all dependencies (and all your dependencies' dependencies) or stuff like this creeps in over time.

kintel:
This should be green now if you sync to master

.circleci/config.yml
```diff
@@ -500,7 +548,6 @@ workflows:
                       - coverity_scan
                       - /^(?i:continuous)$/
       - openscad-appimage:
-          name: openscad-appimage-pr
```
kintel, line 503: We should keep this name to avoid mixing it up with non-PR builds

.circleci/config.yml
```diff
@@ -539,7 +586,8 @@ workflows:
           requires:
             - openscad-tarball-release
       - openscad-appimage:
-          name: openscad-appimage-release
+          # Legacy name, for openscad-ansible/.../circleci-download-artifacts.py
+          name: openscad-appimage-64bit
```
kintel: This is not a legacy name, it's the actual job name for release builds used by `circleci-download-artifacts.py`

.circleci/config.yml
```diff
@@ -8,10 +8,16 @@ executors:
...
     resource_class: large
-  appimage-executor:
-    docker:
-      - image: openscad/appimage-x86_64-base:latest
+  docker-x86_64:
+    machine:
+      image: ubuntu-2204:current
+      docker_layer_caching: true
     resource_class: large
+  docker-arm64:
```
kintel: Unused? Do you plan to use this in the near future? If not, I'd say remove it to avoid potential confusion.

flaviut: happy to drop it for now.

.circleci/config.yml
kintel: We still need to put back `openscad-appimage-release`

flaviut: I thought you asked me to call this `openscad-appimage-64bit`? https://github.com/openscad/openscad/pull/6474#discussion_r2651421246

Anyway, updated, but https://github.com/openscad/openscad-ansible is a private repo and I can't double-check its correct.

kintel: Sorry for the confusion: The snapshot build is named `openscad-appimage-64bit`, and the release build is named `openscad-appimage-release`.

kintel: Added you to the repo

kintel:
@t-paul Any concerns from you?

PR #6482, opened 2026-01-01
Update Font List

t-paul:
* Fix font sample rendering
* Remove obsolete Font List Dialog


src/gui/FontList.cc
```diff
@@ -81,8 +81,10 @@ void FontItemDelegate::paint(QPainter *painter, const QStyleOptionViewItem& opti
   const auto fontName = FontList::colStr(idx, FontList::COL_FONT_NAME);
   const auto fontStyle = FontList::colStr(idx, FontList::COL_FONT_STYLE);
 
-  opt.font.setFamily(fontName);
-  opt.font.setStyleName(fontStyle);
+  auto font = QFont(fontName);
+  font.setStyleName(fontStyle);
+
+  opt.font = font;
```
jordanbrown0, line 87: Am I correct in gathering that the issue here is that the old way ended up sharing one QFont object across all entries in the list, and so updating that one object changed all of the entries, and the change is to have a separate QFont object for each entry?

jordanbrown0 (approved):
Assuming that my understanding of the issue is correct, looks good to me.

src/gui/FontList.cc
t-paul: No idea why that's needed, it should not be required as the Delegate is specifically there to be some sort of stamp. But Qt handling of fonts and style sheets is just an ugly inconsistent mess :(.

Issue #5485, opened 2024-12-09
Ever increasing memory when generating animation using Manifold backend (leak?)


hzeller:
When creating an animation with the Manifold backend (version 3.0), the memory use is ever increasing.

Let's have a simple animation
```openscad
difference() {
  sphere(r=10, $fn=128);
  rotate([$t*3600, $t*1234, $t*3210]) {
    translate([0, 0, -11]) cylinder(r=3, h=22);
  }
}
```

... then create some 3000 frames with it

```
openscad --animate=3000 --imgsize=2048,2048 --backend=Manifold --render=force some-animation.scad
```

The resident size of RAM is ever increasing; towards the end, the process uses about 10GiB of RAM, while when using the CGAL backend, the memory eventually settles on some (lower) steady-state.

This might be a leak somewhere (it could of course also be some sort of increasing cache without ever cleaned up; but in that case CGAL seems to be more efficient in this regard).

pca006132:
I recall kintel talked about some issue with cache size and wanted manifold to implement a mesh size estimation or something. Forgot how that went in openscad.

kintel:
Oh, right. I think we register 0 bytes cache size. That would be hard on memory..

kintel:
Manifold caching is a bit of a tricky topic, perhaps time to revisit it..

pca006132:
I'm actually quite curious about how the geometry cache works. Do we have examples in which there are many hits?

kintel:
Lots of hits would happen if there are repeated parts, but often one hit is enough to make a difference.
Consider a conceptual examples like this:
```
minkowski() {
  cube(20);
  sphere(2, $fn=64);
}
translate([10,10,-5]) cylinder(h=30, r=5);
```

If you don't change the minkowski object very often, but work on positioning the hole, you can shave off some time.
Especially in preview mode, where you could move the hole around using Alt-arrows in real-time despite the minkowski taking ~5 seconds to calculate.

pca006132:
I see. But as mentioned a long time ago, this kind of cache can also inhibit certain optimizations that manifold performs internally. I think we need to think about what should be cached and what should not, probably require some complicated heuristics.

pca006132:
Minkowski should definitely be cached, but for other boolean operations I am not so sure.

coryrc:
[...]

PR #6499, opened 2026-01-06
Estimate cache size for manifold objects

kintel:
Working towards a solution for #5485:
Estimate Manifold object size to avoid ever-increasing memory usage.

This fixes ever-growing memory caused by never flushing the cache, but this isn't necessarily fix all growing memory problems in the reported issue.

### DIFF
--- a/src/ext/lexertl/README.md
+++ b/src/ext/lexertl/README.md
@@ -1,3 +1,8 @@
+This was copied from https://github.com/BenHanson/lexertl14/tree/488602f9994b70fabc3e4869a48a50ef18a1e159 and its subdirectory include/lexertl too
+
+There was local modification afterward in OpenSCAD commit 304897b8147de6f24ec1d9bc9382a52d7d634043
+
+
 lexertl14: The Modular Lexical Analyser Generator
 =======
 
--- a/src/ext/lodepng/README.md
+++ b/src/ext/lodepng/README.md
@@ -1,0 +1,5 @@
+Version 20260102
+
+Seems the version gets updated with every/most commits in these files.
+
+These files are from https://github.com/lvandeve/lodepng/tree/c572511174ce6864574acda70cb0cbe9ec39e23b
--- a/src/ext/lodepng/lodepng.cpp
+++ b/src/ext/lodepng/lodepng.cpp
@@ -1,7 +1,7 @@
 /*
-LodePNG version 20230410
+LodePNG version 20260102
 
-Copyright (c) 2005-2023 Lode Vandevenne
+Copyright (c) 2005-2026 Lode Vandevenne
 
 This software is provided 'as-is', without any express or implied
 warranty. In no event will the authors be held liable for any damages
@@ -44,7 +44,7 @@ Rename this file to lodepng.cpp to use it for C++, or to lodepng.c to use it for
 #pragma warning( disable : 4996 ) /*VS does not like fopen, but fopen_s is not standard C so unusable here*/
 #endif /*_MSC_VER */
 
-const char* LODEPNG_VERSION_STRING = "20230410";
+const char* LODEPNG_VERSION_STRING = "20260102";
 
 /*
 This source file is divided into the following large parts. The code sections
@@ -297,12 +297,6 @@ static ucvector ucvector_init(unsigned char* buffer, size_t size) {
 #ifdef LODEPNG_COMPILE_PNG
 #ifdef LODEPNG_COMPILE_ANCILLARY_CHUNKS
 
-/*free string pointer and set it to NULL*/
-static void string_cleanup(char** out) {
-  lodepng_free(*out);
-  *out = NULL;
-}
-
 /*also appends null termination character*/
 static char* alloc_string_sized(const char* in, size_t insize) {
   char* out = (char*)lodepng_malloc(insize + 1);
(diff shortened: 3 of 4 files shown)


src/geometry/GeometryEvaluator.cc
```diff
@@ -330,6 +330,29 @@ std::vector<std::shared_ptr<const Polygon2d>> GeometryEvaluator::collectChildren
...
+      }
+    }
+
+    if (skipCache) return;
+  }
+#endif
+
   const std::string& key = this->tree.getIdString(node);
 
   if (CGALCache::acceptsGeometry(geom)) {
```
coryrc, line 335: nit: If `CGALCache` accepts Manifold, could you change the name?

src/geometry/GeometryEvaluator.cc
```diff
@@ -330,6 +330,29 @@ std::vector<std::shared_ptr<const Polygon2d>> GeometryEvaluator::collectChildren
 void GeometryEvaluator::smartCacheInsert(const AbstractNode& node,
                                          const std::shared_ptr<const Geometry>& geom)
 {
+#ifdef ENABLE_MANIFOLD
+  if (RenderSettings::inst()->backend3D == RenderBackend3D::ManifoldBackend) {
+    bool skipCache = false;
+    if (const auto opNode = dynamic_cast<const CsgOpNode *>(&node)) {
+      // Skip caching UNION to enable BatchUnion optimizations.
```
coryrc: Could you expand on why this happens? i.e.

```
// Attempting to cache a Manifold results in it being fully evaluated,
// because of the functions we use to estimate its cost.
// Doing this negates many performance advantages built into Manifold,
// so we should only do so if we think it's worthwhile
```

Also could we estimate cost without evaluating even if it's less accurate? ...  OpenSCAD could have a union with 100 things in it which might very well be worth caching.

kintel: I'll add more comments. The crux is that Manifold has internal optimization where it may reorder or cull geometry when looking at a CSG hierarchy, making evaluation faster. If we evaluate each node, we take away that possibility.
But I agree that some heuristic may be useful here, to cache intermediate geometry at some level. Currently, that happens at strategic points based on OpenSCAD node type.

I'll convert this PR to draft to indicate that it's WIP

coryrc: Wait, can you *copy* the node first? Keep the unevaluated one and only save the evaluated one? Or maybe that would usually be worse.

I would hope then the cached would be better than not-cache even if evaluated.

Or, wait, forget memory usage for a moment. The goal is to cache things which are expensive, right? Can we *time* this?

kintel: We can't deep-copy the Manifold CSG tree, and the evaluation method mutates the tree (collapses it into a single node).

Yeah, one of my ideas was to time it, but since we only evaluate the top union, we don't have timing info for any subtrees. Another idea is to keep a cache request count for each cache ID, and only trigger evaluation and caching for nodes being requested multiple times. ..but then we still run the risk of evaluating everything unless we do a top-down cache request count and a bottom-up evaluation.
..but even for a bottom-up evaluation, we could opt out of caching for very cheap nodes.

All in all, this can get pretty complex so I'm making baby steps first: Step 1: Avoid runaway memory usage.

kintel: I moved the selective caching to #6510, as that will take longer to sort out.

src/geometry/GeometryEvaluator.cc
kintel: Will deal with that separately.

coryrc:
I'm pretty sure this slowed things down 50x

kintel:
Hm, it should only slow things down if you exhaust the cache, forcing frequent recalculation. You can increase the memory size of the cache in Preferences. We may want to update the defaults.

coryrc:
> Hm, it should only slow things down if you exhaust the cache, forcing frequent recalculation. You can increase the memory size of the cache in Preferences. We may want to update the defaults.

First, I haven't been using Debug mode that long. Maybe that's the problem.

I took a complicated model and hit F6 with Manifold, and it's very slow the first time run. Very, very slow. It was not before.

kintel:
I'll do some casual testing.

We should probably set up some more structured performance testing very soon, perhaps prior to cutting an RC

coryrc:
> Hm, it should only slow things down if you exhaust the cache, forcing frequent recalculation. You can increase the memory size of the cache in Preferences. We may want to update the defaults.

I think the problem is you're forcing evaluation and removing most of Manifold's optimizations?

kintel:
This fix didn't change how often we evaluate, it only added a cost to caching, to plug runaway memory usage.

coryrc:
Hmm, okay. I'll try some bisecting some time.

kintel:
Some light testing shows that performance of large unions is equivalent, until the cache is full. After that recalculations occur for re-renders.

..when built using `-DCMAKE_BUILD_TYPE=Release`

coryrc:
It's probably the logging to the console in Debug mode. There's... a lot.

coryrc:
I have Manifold as the chosen backend, and with F6 I run and it spits out debug cache insertions for about 2 seconds (like `Geometry Cache inserted: multmatrix([[1,0,0,5],[0,1,0,12],[0,0,1, (1464 bytes)`), then sits burning CPU for about 60 seconds. I put it in the debugger, started the render, waited about 10 seconds, hit ctrl-c, and this popped up:

```
Thread 26 (Thread 0x7fffeaf5f6c0 (LWP 317769) "QThread"):
#0  mi_stat_free (page=0x2000c000218, block=0x2000c055800) at /opt/prg/openscad2/submodules/mimalloc/src/alloc.c:348
#1  0x000055555630b421 in mi_free (p=0x2000c055800) at /opt/prg/openscad2/submodules/mimalloc/src/alloc.c:542
#2  0x00005555555fc3a0 in gmp_free (ptr=0x2000c055800) at /opt/prg/openscad2/src/openscad_mimalloc.h:22
#3  0x0000555555a032f8 in CGAL::Gmpq_rep::~Gmpq_rep (this=0x2000c15c810, __in_chrg=<optimized out>) at /usr/include/CGAL/GMP/Gmpq_type.h:54
#4  0x0000555555a0532a in CGAL::Handle_for<CGAL::Gmpq_rep, std::allocator<CGAL::Gmpq_rep> >::RefCounted::~RefCounted (this=0x2000c15c810, __in_chrg=<optimized out>) at /usr/include/CGAL/Handle_for.h:41
#5  0x0000555555a038c5 in std::__new_allocator<CGAL::Handle_for<CGAL::Gmpq_rep, std::allocator<CGAL::Gmpq_rep> >::RefCounted>::destroy<CGAL::Handle_for<CGAL::Gmpq_rep, std::allocator<CGAL::Gmpq_rep> >::RefCounted> (this=0x555556ce17e3 <CGAL::Handle_for<CGAL::Gmpq_rep, std::allocator<CGAL::Gmpq_rep> >::allocator>, __p=0x2000c15c810) at /usr/include/c++/14/bits/new_allocator.h:198
#6  std::allocator_traits<std::allocator<CGAL::Handle_for<CGAL::Gmpq_rep, std::allocator<CGAL::Gmpq_rep> >::RefCounted> >::destroy<CGAL::Handle_for<CGAL::Gmpq_rep, std::allocator<CGAL::Gmpq_rep> >::RefCounted> (__a=<optimized out>, __p=0x2000c15c810) at /usr/include/c++/14/bits/alloc_traits.h:597
#7  CGAL::Handle_for<CGAL::Gmpq_rep, std::allocator<CGAL::Gmpq_rep> >::~Handle_for (this=0x7fffeaf4f1b8, __in_chrg=<optimized out>) at /usr/include/CGAL/Handle_for.h:167
#8  0x0000555555a03330 in CGAL::Gmpq::~Gmpq (this=0x7fffeaf4f1b8, __in_chrg=<optimized out>) at /usr/include/CGAL/GMP/Gmpq_type.h:63
#9  0x0000555555cfe4e4 in CGAL::Sphere_point<CGAL::Cartesian<CGAL::Gmpq> >::Sphere_point (this=0x7fffeaf4f200, x=..., y=..., z=...) at /usr/include/CGAL/Nef_S2/Sphere_point.h:88
#10 0x0000555555d1a987 in CGAL::Positive_halfsphere_geometry<CGAL::Cartesian<CGAL::Gmpq> >::xz_pi_half_rotate (this=0x7fffeaf4f9cc, p=...) at /usr/include/CGAL/Nef_S2/Sphere_geometry.h:67
#11 0x0000555555d06116 in CGAL::Positive_halfsphere_geometry<CGAL::Cartesian<CGAL::Gmpq> >::orientation (this=0x7fffeaf4f9cc, p1=..., p2=..., p3=...) at /usr/include/CGAL/Nef_S2/Sphere_geometry.h:103
#12 0x0000555555d28fbf in CGAL::Positive_halfsphere_geometry<CGAL::Cartesian<CGAL::Gmpq> >::orientation (this=0x7fffeaf4f9cc, s=..., p=...) at /usr/include/CGAL/Nef_S2/Sphere_geometry.h:111
```

So it seems to be using CGAL somehow.

This is the only change in `src/geometry/GeometryCache.cc` in the last year, which makes me suspect it, but I guess it could be something else.

coryrc:
This may just be a Debug thing I never noticed before, so I filed #6558 to discuss it.

PR #6505, opened 2026-01-08
Log Measurements, Fix Edge-Case Bugs

### LINKED ISSUE #5672: Make interactive measurement tool release ready

coryrc:
Here's what I'm currently getting after measuring point-to-point, point-to-line, line-to-non-skew-line (well, segments, of course, where the segments don't intersect but the (infinite) lines they lie on do), line-to-parallel-or-skew-line (again, actually segments):

```
Distance is 20
  Δ = [20, 0, 0]
  First selection point is at [-40, -7.0710678100585938, 30]
  Second selection point is at [-20, -7.0710678100585938, 30]
Distance is 20
  Point to Line Endpoint1 Δ = [27.07106876373291, 7.0710678100585938, 0]
  Point to Line Endpoint2 Δ = [20, 0, 0]
  Perpendicular to (infinite) line Δ = [-9.9999986513009382, 9.9999999999999076, 0]
  First selection point is at [-40, -7.0710678100585938, 30]
  Second selection line is at [-12.92893123626709, 0, 30] to [-20, -7.0710678100585938, 30]
Distance is 20
  First selection line is at [-40, -7.0710678100585938, 30] to [-40, 7.0710678100585938, 30]
  Second selection line is at [-12.92893123626709, 0, 30] to [-20, -7.0710678100585938, 30]
Distance is 20
  Perpendicular to (infinite) lines Δ = [20, 0, 0]
  First selection line is at [-40, -7.0710678100585938, 30] to [-40, 7.0710678100585938, 30]
  Second selection line is at [-20, -7.0710678100585938, 30] to [-20, 7.0710678100585938, 30]
```

In the context menu, if you select `Distance is 20` just `20` is copied to your clipboard, for `Point to Line Endpoint2 Δ = [20, 0, 0]` just `[20, 0, 0]`, etc.

~Something funny going on with segment lines TBD.~ I labeled the measurement wrong in this post originally, code is correct.

Also fixed a bunch of edge cases with nearly-parallel line segments. See added tests.

Fixes #5672

### DIFF
--- a/src/gui/Measurement.h
+++ b/src/gui/Measurement.h
@@ -47,7 +47,6 @@ public:
   struct Distance {
     double distance;
     int line_count;
-    QString codingError;
     std::optional<Eigen::Vector3d> ptDiff, toInfiniteLine, toEndpoint1, toEndpoint2;
   };
 
--- a/src/gui/Measurement.cc
+++ b/src/gui/Measurement.cc
@@ -106,9 +106,6 @@ Result Template<TView>::statemachine(QPoint mouse)
       obj1 = qglview->selected_obj[0];
       obj2 = qglview->selected_obj[1];
       Distance res = distMeasurement(obj1, obj2);
-      if (!res.codingError.isEmpty()) {
-        ret.addText(res.codingError);
-      }
 
       auto obj2s = QString::fromStdString(obj2.toString());
       ret.addText(QStringLiteral("Second selection %1 is at %2")
@@ -284,7 +281,8 @@ typename Template<TView>::Distance Template<TView>::distMeasurement(SelectedObje
       ret.toInfiniteLine = inf;
     }
   } else {
-    ret.codingError = "Only coded to handle lines and points; sorry";
+    assert("It should not have been possible to select something other than a point and a line" &&
+           false);
   }
   return ret;
 }


coryrc:
Continuing conversation with @jordanbrown0 

> ```
> Distance is 30.5666
>   [dx, dy, dz] = [-5.85786, 0, 30]
> ```
> 
> I don't see any advantage over
> 
> ```
> Distance: 30.5666 [-5.85786, 0, 30]
> ```
> 
> but I don't object so much as to fight about it.

The advantage is the context menu lets you copy `30.5666` or `[-5.85786, 0, 30]` by choosing which line.

> What is this measuring?

This time I labeled what the measurements are in my post.

> ```
> Distance is 19.1421
>   Point to Line Endpoint1: [dx, dy, dz] = [-12.9289, 14.1421, 0]
>   Point to Line Endpoint2: [dx, dy, dz] = [-20, 7.07107, 0]
>   Perpendicular distance to (infinite) line: [dx, dy, dz] = [13.5355, -13.5355, 0]
> ```
> 
> What is the first one measuring? I suspect it's the perpendicular distance, because the values match, but nothing here associates the two.

It's the shortest distance from the point to anywhere on the line **segment**, which is often *not* perpendicular distance.

> Why don't the "point to line endpoint" lines include the actual distance?

When I'm adjusting my code to line things up, I care about data I can put into `translate()`. Since each context line can only copy one piece of data, I'd rather have this. And if I cared about distance I'd measure to the endpoint!

>  I don't feel a need to specially discuss "perpendicular distance"; I think that's the most obvious distance.

See above.

> But also I have to wonder why the rule here isn't "measure to the thing you want to measure to". If you want the distance to an endpoint, measure to that endpoint, not to the line.

If I'm trying to line up something with a line, I can do one measurement with both endpoint deltas, which is nicer than trying to remember that two separate measurements line up with the same line.

> I would think of
> 
> ```
> Distance: 19.1421 [13.5355, -13.5355, 0]
>   to Endpoint1: 19.1613 [-12.9289, 14.1421, 0]
>   to Endpoint2: 21.2132 [-20, 7.07107, 0]
> ```
> 
> ... which is more information (or at least more-digested information), in less space with fewer words.
> ...
> You want exactly enough label to make it clear - given some knowledge and context - what the value is, and no more.

Problem is the deltas if you choose the Point as your second selection. The deltas are always output from the Point to the Line, so that's why I have the extra words.

jordanbrown0:
I don't really have anything more to say.  I would use this feature rarely enough that fine-tuning it wouldn't help my life appreciably.

src/gui/Measurement_test.cc
kintel, line 1: Was this meant to be a starting point of more tests?

src/gui/Measurement.h
```diff
@@ -22,11 +23,36 @@ class Measurement
   Measurement(void);
   void setView(QGLView *qglview);
 
+  struct Distance {
+    double distance;
+    int line_count;
+    QString codingError;
```
kintel: This is a bit of an odd solution.
We haven't implemented a structured convention for how to deal with internally generated error messages.
If you can generate the string at the call site, that's probably the best for now.
 
I've been using absl::StatusOr a lot, but not in OpenSCAD and it helps keep error handling clean, but requires a bunch of macros. std::expected solves this too, but only in C++23.
std::optional may be used too, if the error is generic.

src/gui/Measurement.h
```diff
@@ -40,4 +66,8 @@ class Measurement
 
 private:
   QGLView *qglview;
+  Distance distMeasurement(SelectedObject& obj1, SelectedObject& obj2);
+  friend class MeasurementTest;
```
kintel: Nit: Ideally, code shouldn't need to know about tests. Needing to use a friend is usually a sign that the component could be better designed.

src/gui/Measurement.h
coryrc: I'd use absl (used to work for the big G, so...)

src/gui/Measurement.h
coryrc: Well, turns out I didn't need it to write my tests, so I guess it was already "better designed" :)

It'll be removed in the next iteration.

src/gui/Measurement.h
kintel: Not sure I want to drag in absl though; it will quickly become a deep dependency. Ideally we'd use the C++23 feature if someone could bridge the gap in the interim,

src/gui/Measurement_test.cc
coryrc: Yes, I've added one because I forgot to plumb through the values once.

But if you think it's too complicated I'll delete it all.

src/gui/Measurement.h
coryrc: https://stackoverflow.com/questions/32126613/c-equivalent-of-rusts-resultt-e-type seems to have a summary of the situation. Also don't *need* to return it; could just LOG instead, it's basically an assertion at this point.

src/gui/Measurement_test.cc
kintel: If you don't intend to use it in this PR, I say remove it.

src/gui/Measurement.h
kintel: Don't have a strong opinion on LOG vs. returning an error marker like std::optional. I just don't like value structs with built-in error messages :)

src/gui/Measurement_test.cc
coryrc: I did add it, just not sure if you're going to like *how* I did.

src/gui/Measurement.h
coryrc: Are you saying you'd be happy with `std::optional<QString> codingError`?

kintel: I'm suggesting move `codingError` out of the struct and let anyone returning `Distance` either 1. return an optional and generate the error string at the call site, or 2. LOG internally.

coryrc: Okay, I'll do at least no. 2. Or... I'll just add an `assert`.

src/gui/Measurement_test.cc
kintel: oh, sorry, didn't notice. This is fine. I'm a lot less concerned about exactly how we write tests at this point. test beats no test :)

src/gui/Measurement.h
coryrc: `assert` added.

PR #6511, opened 2026-01-11
fix grey/darkgrey colors mixup

PlungeMaster:
values for grey and darkgrey were swapped


t-paul (changes requested):
Hmm, interesting. While I agree looking at the values, the CSS spec the colors come from specify exactly those values: https://drafts.csswg.org/css-color/

t-paul:
Closing, we'll follow the spec. Reason for the conflict is explained in https://en.wikipedia.org/wiki/X11_color_names#Clashes_between_web_and_X11_colors_in_the_CSS_color_scheme

PR #6518, opened 2026-01-13
Run ./scripts/translation-update.sh

### LINKED ISSUE #6514: Update translation files
Before cutting a Release Candidate: update translation files.

This is mostly a matter of running translation-update.sh.

See https://github.com/openscad/openscad/blob/master/doc/translation.txt

coryrc:
Fixes #6514 
AFAICT


t-paul:
Yes and no. Yes it does for the *current* time (which is good). It would make sense to run it again as one of the last actions before doing the RC.

PR #6527, opened 2026-01-14
Update Jump-To-Window menu

t-paul:
Minor updates to the CTRL-J popup menu:
* Add mnemonic and shortcut to Jump-To-Window menu.
* First make sure the window is visible before trying to raise it.

### DIFF
--- a/src/gui/ColorList.cc
+++ b/src/gui/ColorList.cc
@@ -135,7 +135,10 @@ ColorList::ColorList(QWidget *parent) : QWidget(parent), ui(new Ui_ColorListWidg
                    &ColorList::updateSort);
 }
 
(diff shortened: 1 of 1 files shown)


coryrc:
This breaks using Alt-E to go to the menu bar:

`QAction::event: Ambiguous shortcut overload: Alt+E`

coryrc:
(on Linux, probably why you didn't notice on mac?)

coryrc:
Also, only happens for Alt-E and Alt-V

coryrc:
And it depends on where your text cursor is; in the Editor I get that error, but using it from a text box in the Animate panel, it goes to the menu bar instead.

PR #6530, opened 2026-01-15
clang-format: Add AllowShortFunctionsOnASingleLine: InlineOnly

kintel:
How this was done:

* Lookup clang-format docs at https://clang.llvm.org/docs/ClangFormatStyleOptions.html
* Our format is based on the "Google" style. To inspect the Google style: `clang-format -style=Google -dump-config`
* Edit https://github.com/openscad/openscad/blob/master/.clang-format
* Run `./scripts/beautify.sh --all`, inspect the result and iterate as needed.

PR #6488, opened 2026-01-02
Add color list window.

t-paul:
A color list, similar to the font list.

* Select from all available color names
* Add [XKCD colors](https://xkcd.com/color/rgb/) (because we can :grinning:)
* Drag&drop names to editor
* Filter and sort settings are remembered (except the color name filter)

ToDo:
* Console extension to allow links, here to open the color list window

https://github.com/user-attachments/assets/260e609b-b4a2-4c66-9190-63d9874adb32

Closes #6138.

PR #6537, opened 2026-01-17
Add OSC disclaimer.

t-paul:


kintel (approved):

kintel:
The code format check is because the change in #6530 didn't cause #6488 to re-run checks. beautify will fix it.

t-paul:
Huh, I thought I pushed that update already...

t-paul:
Ah, I pushed that to the other branch, I'll just merge both then this should be back to green...

PR #6541, opened 2026-01-17
Appimage: Bump dependencies and update to Qt6

t-paul:
Updates:
* Update to Qt6, based on Ubuntu 22.04 which has 6.2.4
* Bump lib3MF to 2.4.1
* Bump CGAL to 6.1

Tested on Debian/Testing (Forky). Looks ok for core features, but has some issues:
* No sound due to gstreamer loader problems
* Probably no web access due to libssl problems


kintel:
> * No sound due to gstreamer loader problems
> * Probably no web access due to libssl problems

Do we need to get this sorted before RC, or release?

t-paul:
I'd say no. I'm not even sure it's actually fixable without providing a second version built against a more recent base platform. It should work fine on the older systems though.

PR #6543, opened 2026-01-18
Fix headless build by separating GUI tests.

t-paul:


kintel:
I wonder if we could set up some nightly/occasional build exercising a bunch of different build options to help us catch these things, e.g. something running this with a bunch of options: https://github.com/openscad/openscad/blob/master/.github/workflows/manual.yml

Issue #6362, opened 2025-11-15
(macOS) `window.geometry` corruption leading to unrecoverable crash loop


PhilipTrauner:
**Describe the bug**

sometimes going from multiple connected displays, to just the internal display, and then back to multiple displays seems to corrupt the `window.geometry` property list (`~/Library/Preferences/org.openscad.OpenSCAD.plist`) entry. 
in this corrupted state occurs, OpenSCAD will open normally, but crash soon after, regardless of user inputs having occurred or not. 
deleting the `window.geometry` property, followed by restarting `cfprefsd` reliably fixes the issue. 

alternatively:
```sh
rm ~/Library/Preferences/org.openscad.OpenSCAD.plist
killall -u "${USER}" cfprefsd
```

**To Reproduce**
i have not found a way to consistently reproduce the issue, but it appears to be related to display output changes as mentioned.
barring a reliable reproducer, i will attach the `window.geometry` value as soon as i encounter another crash.

**Environment and Version info (please complete the following information):**
 - OS: macOS 15.7.1 
 - OpenSCAD Version: 2025.10.25 (homebrew `openscad@snapshot`)

**Library & Graphics card information**
```
OpenSCAD Version: 2025.10.25 (git 5495ab4a7)
System information: macOS 15.7.1 arm64 MacBookPro18,3 10 CPUs 32.00 GB RAM 
User Agent: OpenSCAD/2025.10.25 (git 5495ab4a7) (macOS 15.7.1 arm64 MacBookPro18,3)
Compiler: Clang "17.0.0 (clang-1700.0.13.5)"
MinGW build: No
Debug build: No
Boost version: 1_88
Eigen version: 3.4.0
CGAL version, kernels: 6.1, Cartesian<Gmpq>, Extended_cartesian<Gmpq>
OpenCSG version: OpenCSG 1.8.1
Clipper2 version: 1.5.3
Manifold version: 3.2.1
Qt version: 6.5.5
QScintilla version: 2.14.1
InputDrivers: 
GLib version: 2.85.0
lodepng version: 20230410
libzip version: 1.11.4
fontconfig version: 2.16.0
freetype version: 2.13.3
harfbuzz version: 11.4.1
cairo version: 1.18.0
lib3mf version: 2.4.1
Features: roof, input-driver-dbus, lazy-union, vertex-object-renderers-indexing, textmetrics, import-function, object-function, predictible-output, vector-swizzle
[...]

PR #6546, opened 2026-01-18
Use QT_VERSION_MAJOR as a version for saveState/restoreState 

### LINKED ISSUE #6503: Reset Qt window settings on upgrade
When upgrading from OpenSCAD-2021.01 to a newer version (essentially upgrading from Qt5 to Qt6), old settings entries are incompatible, causing OpenSCAD to crash when Qt attempts to initialize Qt6 windows from Qt5 settings.
The current workarounds are to either:
* Remove `~/Library/Preferences/org.openscad.OpenSCAD.plist` (and make sure this file isn't cached)
* Run `OpenSCAD.app/Contents/MacOS/OpenSCAD  --reset-window-settings`

None of these are particularly user friendly; we should fix this. Ideas:
* Completely clear settings if old settings are detected (run the risk of wiping some carefully adjusted settings)
* Clear only certain settings (Window position) if possible
* To facilitate this, encode OpenSCAD version in settings, and assume that no version means 2021.01 or earlier

OS: Only observed on macOS, but it's possible that this has similar effects on other platforms when going from Qt5 to Qt6 without first clearing preferences.

Example issues: #6500 #6362 #6126 #6462

kintel:
to avoid trying to restore an incompatible state.
Fixes #6503

Theory:
* Upgrading from Qt5 to Qt6 may cause Qt not be unable to restore window state due to an unknown internal bug in Qt failing to associate saved state with existing widgets
* OpenSCAD made a number of changes in the last year related to dockable widgets and editor tabs. There's a chance that this triggers the above mentioned issue even if everything is Qt6.
* Introducing a version makes OpenSCAD ignore any window state saved from previous versions. This will cause people to lose window state once when they updated (or multiple times if they switch between old and new versions).

In the even that this doesn't fix the crash and we're unable to reproduce and troubleshoot further, the next idea is to detect the crash, and wipe he state on the subsequent launch so that it doesn't crash again.

### DIFF
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -669,7 +669,9 @@ MainWindow::MainWindow(const QStringList& filenames) : rubberBandManager(this)
     setGeometry(screen()->availableGeometry());
   }
 #endif
-  restoreState(windowState);
+  // Use QT_VERSION_MAJOR as the version number to invalidate incompatible states
+  // from older Qt versions (e.g. Qt5 -> Qt6 migration), which can cause a crash.
+  restoreState(windowState, QT_VERSION_MAJOR);
 
   if (windowState.size() == 0) {
     /*
@@ -3705,7 +3707,7 @@ void MainWindow::closeEvent(QCloseEvent *event)
 
     QSettingsCached settings;
     settings.setValue("window/geometry", saveGeometry());
-    settings.setValue("window/state", saveState());
+    settings.setValue("window/state", saveState(QT_VERSION_MAJOR));
     if (this->tempFile) {
       delete this->tempFile;
       this->tempFile = nullptr;


src/gui/MainWindow.cc
```diff
@@ -669,7 +669,9 @@ MainWindow::MainWindow(const QStringList& filenames) : rubberBandManager(this)
     setGeometry(screen()->availableGeometry());
   }
 #endif
-  restoreState(windowState);
+  // Use QT_VERSION_MAJOR as the version number to invalidate incompatible states
+  // from older Qt versions (e.g. Qt5 -> Qt6 migration), which can cause a crash.
+  restoreState(windowState, QT_VERSION_MAJOR);
```
coryrc, line 674: This will not load window state even if someone is already using Qt6, and also if someone is switching between dev versions with and without, the state will be invalidated regularly. But window state isn't the most critical, so this is fine? But also doesn't fix #6362

kintel: Yeah, switching versions will cause window state to not be restored, but I don't think we want to spend too much time dealing with that particular use-case.

I believe it does fix #6362, but I'm unable to reproduce that crash so we'll have to wait until someone complains :)

kintel: Added some more info in the description

coryrc: I don't think it will fix #6362 because the bug filer specifically said it was caused by modifying the number of displays and monitors, not an upgrade to OpenSCAD.

kintel: I'm starting to agree; that crash has the same stack trace as similar reports. I'm starting to wonder if the root cause is how we initialize dock windows. I still cannot reproduce it, so I'm fumbling around in the dark while reading Qt source code..

kintel:
Abandoning in favour of #6572

PR #6551, opened 2026-01-19
Use QT6 for all releases (including MXE)

### LINKED ISSUE #4909: Upgrade to Qt6

coryrc:
Fixes #4909

### DIFF
--- a/scripts/release-common.sh
+++ b/scripts/release-common.sh
@@ -61,7 +61,7 @@ if [ ! -f $OPENSCADDIR/src/openscad.cc ]; then
   exit 1
 fi
 
-CMAKE_CONFIG="-DCMAKE_BUILD_TYPE=Release -DENABLE_TESTS=OFF"
+CMAKE_CONFIG="-DCMAKE_BUILD_TYPE=Release -DENABLE_TESTS=OFF -DUSE_QT6=ON"
 
 if [ -n "${USE_SCCACHE}" ] && command -v sccache >/dev/null 2>&1; then
   CMAKE_CONFIG="$CMAKE_CONFIG -DCMAKE_C_COMPILER_LAUNCHER=sccache -DCMAKE_CXX_COMPILER_LAUNCHER=sccache"


scripts/release-common.sh
```diff
@@ -61,7 +61,7 @@ if [ ! -f $OPENSCADDIR/src/openscad.cc ]; then
   exit 1
 fi
 
-CMAKE_CONFIG="-DCMAKE_BUILD_TYPE=Release -DENABLE_TESTS=OFF"
+CMAKE_CONFIG="-DCMAKE_BUILD_TYPE=Release -DENABLE_TESTS=OFF -DUSE_QT6=ON"
```
kintel, line 64: I think we should rather switch this to default ON in the main CMakeLists.txt, and use `-DUSE_QT6=OFF` for any exceptions.

coryrc:
Not so easy :( but the right path:

```
CMake Error at CMakeLists.txt:826 (find_package):
  By not providing "FindQt6.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Qt6", but
  CMake did not find one.

  Could not find a package configuration file provided by "Qt6" with any of
  the following names:

    Qt6Config.cmake
    qt6-config.cmake

  Add the installation prefix of "Qt6" to CMAKE_PREFIX_PATH or set "Qt6_DIR"
  to a directory containing one of the above files.  If "Qt6" provides a
  separate development package or SDK, be sure it has been installed.
```

scripts/release-common.sh
t-paul: Yep, that gets my vote as well.

coryrc: Will do, pending new image with libraries available as right now it will fail again.

kintel:
This was superseded by the Qt6 work by @t-paul

PR #6556, opened 2026-01-23
Treat U+FEFF ZERO WIDTH NO-BREAK SPACE as whitespace.  Fixes #6555.

### LINKED ISSUE #6555: Files starting with U+FEFF are considered erroneous
**Describe the bug**
Some editors will put a U+FEFF ZERO WIDTH NO-BREAK SPACE / BYTE ORDER MARK at the top of a file that contains non-ASCII UTF-8 characters... or even if it doesn't.  We regard that as a syntax error, and shouldn't.

**To Reproduce**
1. Create an otherwise-valid file starting with U+FEFF (UTF-8 EF BB BF).
2. Process it from the command line, e.g. `openscad -o foo.stl foo.scad`
3. Observe the error.

**Expected behavior**
Ignore the U+FEFF, or treat it as a space.

**Additional context**
Something in the GUI editor path appears to strip U+FEFF, so you can't demonstrate it there.

jordanbrown0:
Also fix the handling of U+00A0 NO-BREAK SPACE, which was inappropriately matching other sequences.

Assuming that the tests pass, my only concern is that hypothetically maybe U+FEFF should be allowable as part of an identifier.  It is, after all, a **non-break** space; it shouldn't cause stuff adjacent to it to be split apart.  But that's pretty obscure even if we ever allow non-ASCII identifiers.


src/core/lexer.l
```diff
@@ -250,7 +250,18 @@ use[ \t\r\n]*"<"        { BEGIN(cond_use); LOCATION_COUNT_LINES(parserlloc, yyte
  files on the cmd-line as in the editor.
 */%}
 
-[\xc2\xa0]+
+"\xc2\xa0"
```
kintel: For completeness, perhaps mention that this is the UTF-8 encoding of U+00A0 NO-BREAK SPACE

jordanbrown0: The existing comment says pretty much that, only using lowercase "no-break space" instead of the more canonical upper case. I'll tweak it a little.

kintel: duh, sorry, I didn't scroll up :/

t-paul:
I don't think there's an issue with identifiers. If we get that at some point, I would suggest using the official ID_Start + ID_Continue codepoints and that should not include the BOM. Actually it does not even show up in https://www.unicode.org/Public/UCD/latest/ucd/PropList.txt so I assume it might not even be an official code point?

jordanbrown0:
I don't know exactly how `PropList.txt` fits in, but it's not the complete list of code points.  That's [`UnicodeData.txt`](https://www.unicode.org/Public/UCD/latest/ucd/UnicodeData.txt), where you'll find:
```
FEFF;ZERO WIDTH NO-BREAK SPACE;Cf;0;BN;;;;;N;BYTE ORDER MARK;;;;
```

You're right that ID_Start and ID_Continue are the right answer.  [`DerivedCortProperties.txt`](https://www.unicode.org/Public/UCD/latest/ucd/DerivedCoreProperties.txt) looks like it might have a complete list.  It does not include U+FEFF for ID_Start or ID_Continue:
```
FE76..FEFC    ; ID_Start # Lo [135] ARABIC FATHA ISOLATED FORM..ARABIC LIGATURE LAM WITH ALEF FINAL FORM
FF21..FF3A    ; ID_Start # L&  [26] FULLWIDTH LATIN CAPITAL LETTER A..FULLWIDTH LATIN CAPITAL LETTER Z
...
FE76..FEFC    ; ID_Continue # Lo [135] ARABIC FATHA ISOLATED FORM..ARABIC LIGATURE LAM WITH ALEF FINAL FORM
FF10..FF19    ; ID_Continue # Nd  [10] FULLWIDTH DIGIT ZERO..FULLWIDTH DIGIT NINE
```
I have not tried to wrap my mind around how XID_Start and XID_Continue relate to ID_Start and ID_Continue, but it isn't included there either.

In case anybody else was wondering, U+00A0 is not included either.

So, anyhow, we can drop that concern.

jordanbrown0:
The test failures are clearly relevant.  I'm mystified by them and looking into it.

The latin1 test failure seems inevitable if we want to continue towards UTF-8; the existing pattern causes \xA0 to be treated as a space even when it is buried in the middle of a UTF-8 sequence.

jordanbrown0:
I had changed the U+00A0 rule so that it would only gobble up a single U+00A0, rather than any sequence of {C2, A0}.  That had two effects:
* Because the `{UNICODE}+` rule could match multiple characters, it could gobble up sequences of U+00A0 and have the longer match, and win, where previously it would be a tie and the first rule would win.
* The `[\xc2\xa0]+` rule was also causing Latin-1 A0 (an illegal UTF-8 sequence) to be treated as a space, and that effect was lost and so its corresponding test failed.

I made the {UNICODE} rule match only a single character, so that rules that matched a single character could still win.
I added a specific rule for Latin-1 A0.
I added a bunch of comments.
I added a line to the UTF-8 NBSP test to test U+FEFF.

t-paul:
Ah, even better. The PropList only has the original classes and ID_Start and friends are derived like "lower-case-letters + upper-case-letters - ...".

Yeah, I think that "C2/A0" rule was just plain wrong.

jordanbrown0:
I'd rather not punch the "merge" button on my own PR, so if one of you would be so good...

jordanbrown0:
Oh, and a correction:

> the existing pattern causes \xA0 to be treated as a space even when it is buried in the middle of a UTF-8 sequence.

Not true; the `{UNICODE}` pattern, with or without the `+`, will gobble it up.  Still, the rule was unnecessarily opaque.

PR #6557, opened 2026-01-23
Linear Extrude Comments and Testing Fixes and Improvements

coryrc:
First off, the testing had `rotate_extrude` instead of `linear_extrude` in it.

I added a test for checking linear_extrude adds extra faces when it's requested to do so. Not sure how useful it is, but it says it does it and it'll be used for more testing later.

Finally, a couple comment cleanups in the tests.

Fixes https://github.com/openscad/openscad/issues/6366

Extracted from #6361


tests/CMakeLists.txt
```diff
@@ -1149,6 +1149,8 @@ add_cmdline_test(openscad-camvp-override      OPENSCAD FILES ${CAMERA_TEST_VP} S
 #
 # View Options tests
 #
+
+# Kinda weird these tests all set both --camera and --viewall
```
kintel: --camera sets the angle, --viewall set the zoom

tests/CMakeLists.txt
```diff
@@ -1178,6 +1180,20 @@ if(EXAMPLES_DIR)
...
+#
+
+# More about checking otherwise-invisible faces were created than whether the rendering
+# engine(s) are working correctly.
+# CGAL and Manifold both added just to ensure coverage during the transition.
+# Only one needs to exist long-term.
+
+set(WITH_EDGES_TESTS "${TEST_SCAD_DIR}/3D/features/linear_extrude_invisible-tests.scad")
+add_cmdline_test(preview-with-edges-cgal    OPENSCAD FILES ${WITH_EDGES_TESTS} SUFFIX png ARGS ${IMGSIZE} --viewall --view edges --backend=cgal)
+add_cmdline_test(render-with-edges-manifold OPENSCAD FILES ${WITH_EDGES_TESTS} SUFFIX png ARGS ${IMGSIZE} --viewall --view edges --render --backend=manifold)
```
kintel: Technically needs to be wrapped in `if (ENABLE_MANIFOLD_TESTS)`

Now that manifold is default, we can slowly start the process of making Manifold mandatory though, but this isn't the place to start :)

src/geometry/linear_extrude.cc
```diff
@@ -198,53 +198,85 @@ void add_slice_indices(PolygonIndices& indices, int slice_idx, int slice_stride,
...
+      // If you rotate 90 degrees ACW, the first vertex will be further away,
+      // but the second vertex starts and ends out further.
+
+      size_t slicesNonUniScale =
         (unsigned int)node.discretizer.getDiagonalSlices(max_delta_sqr, node.height[2]).value_or(1);
-      slicesTwist = (unsigned int)node.discretizer.getHelixSlices(max_r1_sqr, node.height[2], node.twist)
-                      .value_or(std::max(static_cast<int>(std::ceil(node.twist / 120.0)), 1));
+      size_t slicesTwist =
+        (unsigned int)node.discretizer.getHelixSlices(max_r1_sqr, node.height[2], node.twist)
+          .value_or(std::max(static_cast<int>(std::ceil(node.twist / 120.0)), 1));
```
kintel: Nit: mixing c and c++ cast. Prefer static_cast everywhere.

src/geometry/linear_extrude.cc
```diff
@@ -198,53 +198,85 @@ void add_slice_indices(PolygonIndices& indices, int slice_idx, int slice_stride,
...
-      slicesNonUniScale =
+      double max_delta_sqr = calc_max_delta_sqr(poly.outlines(), scale);
+
+      // Why would we not find the furthest *scaled* max_r1_sqr and use getHelixSlices on that?
+      // Because it scales non-uniformly and so you need a formula for the
+      // length of a non-uniformly scaled helix.
+      // And you would need to check every vertex, because the vertex that's furthest away
+      // before you start twisting may not be the furthest throughout the twist.
+      // Consider vertices at (3.99,2) and (4,-2), and a scale=(1,2).
+      // If you rotate 90 degrees ACW, the first vertex will be further away,
```
kintel: Nit: We use CCW elsewhere, let's stick with that.

tests/CMakeLists.txt
coryrc: Done

src/geometry/linear_extrude.cc
coryrc: Done

src/geometry/linear_extrude.cc
coryrc: Done

tests/CMakeLists.txt
coryrc: Done

PR #6559, opened 2026-01-23
Update circle&sphere tests for $fe

coryrc:
This is a separate PR so we can confirm no regression when $fe is added/enabled and also makes that PR (#6361) simpler to review.


kintel:
Wait, isn't this a clean subset of #6361? If yes, we can kill it as It's small enough to review as one.
I was worried about changing 200+ files, but I guess in experimental mode the changes are reasonable small..

coryrc:
already merged

PR #6561, opened 2026-01-24
Update cmake_minimum_required for resources to 3.10

cjmayo:
All previous versions are deprecated since CMake 3.31.

---

Saves a warning, /CMakeLists.txt is already at 3.13, even Debian oldoldstable has 3.18.


t-paul:
I believe we are good with 3.22.1 (https://github.com/openscad/openscad/wiki/Build-Matrix)

t-paul:
With tests using 3.21, maybe we should bump everything to that?

```
$ grep cmake_minimum_required $(find . -name CMakeLists.txt -a -not -path ./submodules/\*) | sort

./CMakeLists.txt:cmake_minimum_required(VERSION 3.13)
./resources/CMakeLists.txt:cmake_minimum_required(VERSION 3.5)
./tests/CMakeLists.txt:cmake_minimum_required(VERSION 3.21)
./winconsole/CMakeLists.txt:cmake_minimum_required(VERSION 3.5...3.20)
```

Submodules seem to be a mixed bag...

```
$ grep cmake_minimum_required $(find submodules -name CMakeLists.txt) | sort

submodules/Clipper2/CPP/BenchMark/CMakeLists.txt:cmake_minimum_required(VERSION 3.15)
submodules/Clipper2/CPP/CMakeLists.txt:cmake_minimum_required(VERSION 3.15)
submodules/manifold/CMakeLists.txt:cmake_minimum_required(VERSION 3.18)
submodules/mimalloc/CMakeLists.txt:cmake_minimum_required(VERSION 3.13)
submodules/mimalloc/test/CMakeLists.txt:cmake_minimum_required(VERSION 3.0)
submodules/OpenCSG/CMakeLists.txt:cmake_minimum_required(VERSION 3.11)
submodules/sanitizers-cmake/CMakeLists.txt:cmake_minimum_required(VERSION 2.8.12)
```

cjmayo:
I think we should leave submodules unmodified if at all possible, sanitizers-cmake has been updated to `2.8.12...3.10` so added a commit here to use that. (are we ever using mimalloc/test?).

Otherwise I guess it is just an effort/maintenance question - as long as they are all >=3.10 then looks like the issue will go away for a while.

t-paul:
> I think we should leave submodules unmodified if at all possible

Oh, sure, hence the separation. I listed those more for reference, with some luck they are maintained well enough so we do not need any action other the updating to newer versions and otherwise we would need to try changing upstream first.

> sanitizers-cmake has been updated to `2.8.12...3.10` so added a commit here to use that. (are we ever using mimalloc/test?).

I don't think those tests are run in any of our builds.

PR #6562, opened 2026-01-24
Flip default for USE_QT6 to ON for all platforms.

t-paul:


scripts/setenv-mingw-xbuild.sh
```diff
@@ -53,6 +53,9 @@ fi
 if [ ! $MXEQTSUBDIR ]; then
 	# default is qt5 see issue #252
```
kintel, line 54: This looks outdated; can we flip to qt6 as default here too?

.github/workflows/code-tidy.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     - name: Install dependencies
       run: |
         which sudo || apt install -y sudo
-        sudo ./scripts/uni-get-dependencies.sh
+        sudo ./scripts/uni-get-dependencies.sh qt6
```
kintel, line 24: We may want to switch this too, as this is what users will run blindly.
(this script is similar to macosx-build-homebrew.sh which already defaults to qt6)

t-paul: From what I can tell via issues, that one seems to be heavily used outside our own builds, so I'm not sure if we want to change that just yet. For now I mostly did it this way because it was the quickest path :-).

scripts/setenv-mingw-xbuild.sh
t-paul: I would like to leave that as 2nd step once the fully automatic MXE Qt6 build via CI is actually green.

PR #6566, opened 2026-01-25
Switch msys2-install-dependencies.sh to Qt6 by default.  Fixes #6565.

### LINKED ISSUE #6565: Switch default for msys2-install-dependencies.sh to Qt6
**Describe the bug**
With PR#6562 switching us to Qt6 by default, msys2-install-dependencies.sh should similarly default to Qt6.

**To Reproduce**
1. Clone.
2. Run scripts/msys2-install-dependencies.sh.
3. Run cmake.
4. See that it fails because Qt6 dependencies are not present.

**Expected behavior**
Successful build after documented setup.

 - OS: Windows 11
 - System: Intel PC 64-bit

jordanbrown0:
Before, the default was to install Qt5 components, and install Qt6 components instead if $1 is "qt6".
With this change, the default is to install Qt6 components, and install Qt5 components instead if $1 is "qt5".

Perhaps the default (and maybe only) behavior should be to install both, to be a superset of the old behavior, but it seemed simpler to track the build default... no reason to install Qt5 components for default builds.

Fixes #6565.

### DIFF
--- a/scripts/msys2-install-dependencies.sh
+++ b/scripts/msys2-install-dependencies.sh
@@ -25,10 +25,10 @@ if [[ -z "${GITHUB_RUN_ID}" ]]; then
     #            name:  disables any translation for name
 fi
 
-if [[ "$QT" == "qt6" ]]; then
-  QT_PACKAGES="qscintilla-qt6:p qt6-5compat:p qt6-multimedia:p qt6-svg:p"
-else
+if [[ "$QT" == "qt5" ]]; then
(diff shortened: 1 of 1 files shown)


jordanbrown0:
That failure is troubling, but on its face it's a problem retrieving packages, unrelated to **what** packages are being retrieved.

flaviut:
Ref #6565 (because GH doesn't seem to link it when it's in the title)

Not sure if this has been discussed elsewhere, but IMO this is a good idea :+1:

PR #6572, opened 2026-01-27
Refactor MainWindow constructor

### LINKED ISSUE #6571: QSettings::setValue: Empty key passed
**Describe the bug**

On Debian Linux, when starting the main GUI (i.e. clicking New on the first pop-up Window), it prints three errors:

```
QSettings::setValue: Empty key passed
QSettings::setValue: Empty key passed
QSettings::setValue: Empty key passed
```

**Screenshots**

With Debug mode:

```
SettingsWriter: SettingsWriter D: color-list/enable-xkcd-colors
SettingsWriter: SettingsWriter D: color-list/sort-type
SettingsWriter: SettingsWriter D: color-list/sort-ascending
QSettingsCached: QSettings::setValue():  = 'true'
QSettings::setValue: Empty key passed
QSettingsCached: QSettings::setValue():  = 'true'
QSettings::setValue: Empty key passed
QSettingsCached: QSettings::setValue():  = 'true'
QSettings::setValue: Empty key passed
QSettingsCached: QSettings::setValue():  = 'true'
QSettings::setValue: Empty key passed
QSettingsCached: QSettings::setValue(): view/hideEditorToolbar = 'false'
QSettingsCached: QSettings::setValue(): view/hide3DViewToolbar = 'false'
```

Hopefully will help narrow down where.

 - OS: Debian Sid
 - System: x64
 - OpenSCAD Version master `c9baefd0b`

**Additional context**

https://github.com/openscad/openscad/issues/6569#issuecomment-3801413104

kintel:
This fixes some issues and hopefully makes things cleaner:

* There is no need to save Dock visibility state, since Qt already does this. This also fixes issues with empty Dock config key (#6571).
* MainWindow constructor has been split into a number of helper functions each initializing a more well-defined subset of the app.
* There's a chance that this improves the root cause of #6503
* Silence "Already setting window visible" warning on macOS

Fixes #6571

### DIFF
--- a/src/gui/Dock.cc
+++ b/src/gui/Dock.cc
@@ -4,6 +4,29 @@
 #include <QWidget>
 #include "gui/QSettingsCached.h"
 
+namespace {
+
+QtMessageHandler originalHandler = nullptr;
+
+void silentMessageOutput(QtMsgType type, const QMessageLogContext& context, const QString& msg)
+{
+  if (type == QtWarningMsg && msg.contains("Already setting window visible")) {
+    return;
+  }
+  if (originalHandler) {
+    originalHandler(type, context, msg);
+  }
+}
+
+class ScopedMessageSilencer
+{
+public:
+  ScopedMessageSilencer() { originalHandler = qInstallMessageHandler(silentMessageOutput); }
+  ~ScopedMessageSilencer() { qInstallMessageHandler(originalHandler); }
+};
+
+}  // namespace
+
 Dock::Dock(QWidget *parent) : QDockWidget(parent)
 {
   connect(this, &QDockWidget::topLevelChanged, this, &Dock::onTopLevelStatusChanged);
@@ -57,6 +80,8 @@ void Dock::onTopLevelStatusChanged(bool isTopLevel)
   Qt::WindowFlags flags = (windowFlags() & ~Qt::WindowType_Mask) | Qt::Window;
   if (isTopLevel) {
     setWindowFlags(flags);
+    // show() rmits an innocuous "Already setting window visible!" warning on macOS
+    ScopedMessageSilencer silencer;
     show();
   }
   updateTitle();


src/gui/Dock.cc
```diff
@@ -7,7 +7,6 @@
 Dock::Dock(QWidget *parent) : QDockWidget(parent)
 {
   connect(this, &QDockWidget::topLevelChanged, this, &Dock::onTopLevelStatusChanged);
```
coryrc, line 32: Can this be initialized after restoreState() to be safe?

src/gui/MainWindow.h
```diff
@@ -126,14 +126,25 @@ class MainWindow : public QMainWindow, public Ui::MainWindow, public InputEventH
...
-  std::vector<std::tuple<Dock *, QString, QString>> docks;
+  std::vector<std::tuple<Dock *, QString>> docks;
 
   volatile bool isClosing = false;
   void consoleOutputRaw(const QString& msg);
   void clearAllSelectionIndicators();
   void setSelectionIndicatorStatus(EditorInterface *editor, int nodeIndex,
                                    EditorSelectionIndicatorStatus status);
 
+  void setupOutputWindows();
```
coryrc: I know this is draft, but could you add Doxygen comments for each of these?

src/gui/MainWindow.cc
```diff
@@ -275,3578 +275,3608 @@ MainWindow::MainWindow(const QStringList& filenames) : rubberBandManager(this)
...
-  connect(this->cgalworker, &CGALWorker::done, this, &MainWindow::actionRenderDone);
+    const QString toolTip(
+      "%1 &nbsp;<span style=\"color: gray; font-size: small; font-style: italic\">%2</span>");
+    action->setToolTip(toolTip.arg(action->toolTip(), shortCut));
+  }
+}
 
-  rootNode = nullptr;
+/**
+ * Update window settings that get overwritten by the restoreState()
```
coryrc, line 438: Why do this at all?

src/gui/MainWindow.cc
```diff
@@ -275,3578 +275,3608 @@ MainWindow::MainWindow(const QStringList& filenames) : rubberBandManager(this)
...
+           {colorListDock, _("C&olor List")}, {viewportControlDock, _("&Viewport-Control")}};
 
-  parameterDock->setWidget(newEditor->parameterWidget);
-  editActionUndo->setEnabled(newEditor->canUndo());
-
-  const QString name = getCurrentFileName();
-  setWindowTitle(name);
+  // Connect the menu "Windows/Navigation" to slot that process it by opening in a pop menu
+  // the navigationMenu.
+  connect(windowActionJumpTo, &QAction::triggered, this, &MainWindow::onNavigationOpenContextMenu);
```
coryrc, line 3629: Shouldn't we not connect anything until after restoreState is called to be safe?

Here and below.

src/gui/Dock.cc
kintel: Not sure. If restoreState() changes the toplevel flag, we'd need to iterate through all the docs afterwards and call onTopLevelStatusChanged() if necessary. This should be safe to call I think.

src/gui/MainWindow.cc
kintel: This is part of the `--reset-window-settings` workaround. Let's keep it until we're certain that it's not longer needed.

src/gui/MainWindow.cc
kintel: As mentioned above, if we have custom code reacting to state changes, we have to execute them at some point. If we don't listen for changes, we'd have to walk through the states ourselves, or re-trigger the signals manually.

jordanbrown0:
Appears to have broken Preferences/Buttons; see #6795.

PR #6574, opened 2026-01-27
Configurable error handling:  fatal errors, ordinary errors, warnings

### LINKED ISSUE #5004: Doesn't stop execution on errors

jordanbrown0:
Background:  OpenSCAD has the option to fail on warnings, but never fails on ordinary errors(!).

This change turns the existing checkbox into a three-way:  fail on fatal errors only, fail on ordinary and fatal errors, or fail on all three.  The default is, as before, to fail only on fatal errors.  (Which is probably wrong; it should probably default to failing on ordinary errors too, but this is compatible with the old default.)

The default tracks the existing setting of the hardwarnings flag, defaulting to either Fatal or Warning.

There is a small incompatible change:  previously, "stop on first warning" did not stop on ordinary errors.  Selecting Warnings, including via the existing setting of the hardwarnings flag, now stops on ordinary errors.  I suspect that people who want to stop on warnings will appreciate this change.

There is a new `--hardfail` option that takes `fatal`, `error`, or `warning`.
The `--hardwarnings` option is still available but is now hidden.

Also:  the `--trace-depth` CLI flag never worked, because the logic that consumed the value used the wrong name.  Fixed that.

Fixes #5004.

### DIFF
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -912,9 +912,8 @@ int openscad_main(int argc, char **argv)
     ("m,m", po::value<std::string>(), "make_cmd -runs make_cmd file if file is missing")
     ("quiet,q", "quiet mode (don't print anything *except* errors)")
     ("reset-window-settings", "Reset GUI settings for window placement and fonts.")
-    // Ideally, the fatal/error/warning selection would come from SettingsEntryBase.help().
     ("hardfail", po::value<std::string>(),
-      "=fatal/error/warning, stop on fatal errors, ordinary errors, or warnings")
+      ("stop on first " + std::get<1>(Settings::Settings::hardFailLevel.help())).c_str())
     ("trace-depth", po::value<unsigned int>(), "=n, maximum number of trace messages")
     ("trace-usermodule-parameters", po::value<std::string>(),
       "=true/false, configure the output of user module parameters in a trace")


jordanbrown0:
Here's the revised preferences UI:
image

jordanbrown0:
Argh... my inheritance/import scheme doesn't quite work.

It has the default for hardFailLevel be FATAL.  If enableHardwarnings is true, we set hardFailLevel to WARNING.  But then if you explicitly set hardFailLevel to FATAL, we'll see that that's the default and we won't write it out... and on the next restart you'll be back at WARNING.

jordanbrown0:
In Preferences.cc, there is
```
void Preferences::applyComboBox(QComboBox * /*comboBox*/, int val,
                                Settings::SettingsEntryEnum<OpenSCAD::HardFailLevel>& entry)
```
The body is identical to the immediately-preceding applyComboBox for a SettingsEntryEnum that yields std::string.  One might think that there's some way to have a single applyComboBox work for both types of SettingsEntryEnums, perhaps involving templates, but it's beyond me.  Not that duplicating these two lines is a big deal, but it's the principle of the thing.  Also, if more settings use this style (as, as best I can tell, they should) then they too would need a duplicate function here.

Advice solicited.

jordanbrown0:
Perhaps we should say that WARNING is the best practice, make that the default, and make anybody who wants ERROR or FATAL re-select it, completely ignoring enableHardwarnings.

jordanbrown0:
Or perhaps we should, through one mechanism or another, arrange that hardFailLevel is always written out, whether or not it's the same as the default.

jordanbrown0:
I'm going to mark this as ready for review - not because I think it's 100% ready, but because I need some input on how to proceed.

jordanbrown0:
I'd like to eliminate near-duplicate copies of the list of options (e.g. in the usage message, in the error message, handling CLI args, et cetera), and of course the list in the Settings declaration itself.)

However:
* It seems like the CLI argument ought to be in lowercase, but weak precedent is that the stored value should be capitalized.
* It seems like the stored value needs to be kept compatible, while the CLI argument might evolve over time as we find better words for a concept.  (The CLI needs to accept old aliases, but might have new preferred names for an old concept.)

Both of those would argue that tying the CLI value directly to the stored value is ill-advised.

jordanbrown0:
When I investigated this further, I found that Error cases are expected to throw after LOG() and the real problem was that there were a few cases where it didn't throw.  Having fixed those cases by changing them to Warning (which is optionally fatal), I no longer think that this flexibility is necessary or appropriate.

PR #6578, opened 2026-01-29
fix(python): fix segfault and reference counting bugs in CSG operators

nomike:
## Summary

Fixes two critical bugs in the Python integration that caused crashes when `None` values are used in CSG operations:

1. **Segfault in `PyOpenSCADObjectToNode`**: Function was casting input to `PyOpenSCADObject*` and accessing `->node` before verifying the object was actually a `PyOpenSCADType`. This caused a segfault when operations returned `None` or invalid types were passed.

2. **Reference counting error in CSG operators**: The `python_nb_sub` function (used by `|`, `&`, `-` operators) was returning `arg1` or `arg2` directly without incrementing their reference counts, violating Python C API contracts.

## Root Cause

When Python code uses operations like `cube(10) | None | None`:
- First `cube(10) | None` returns `None` (valid behavior)
- That `None` is passed to the union operator again in `None | None`
- The unsafe cast in `PyOpenSCADObjectToNode` tried to dereference `None` as if it were a `PyOpenSCADObject`, causing a segfault

### Commit 1: Fix reference counting in CSG operators  
- Added `Py_INCREF()` calls when returning `arg1` or `arg2` directly from binary operators
- Python C API binary operators must return new references
- Prevents premature garbage collection of returned objects

### Commit 2: Prevent segfault in PyOpenSCADObjectToNode
- Added `PyObject_IsInstance()` check before casting to `PyOpenSCADObject*`
- Returns `nullptr` for non-PyOpenSCADType objects
- Consistent with how `PyOpenSCADObjectToNodeMulti` handles invalid types

### DIFF
--- a/src/python/pyopenscad.cc
+++ b/src/python/pyopenscad.cc
@@ -115,6 +115,12 @@ void python_unlock(void)
 
 std::shared_ptr<AbstractNode> PyOpenSCADObjectToNode(PyObject *obj, PyObject **dict)
 {
+  // Verify obj is actually a PyOpenSCADType before casting to avoid segfault
+  if (!PyObject_IsInstance(obj, reinterpret_cast<PyObject *>(&PyOpenSCADType))) {
+    *dict = nullptr;
+    return nullptr;
+  }
+
   std::shared_ptr<AbstractNode> result = ((PyOpenSCADObject *)obj)->node;
   if (result.use_count() > 2) {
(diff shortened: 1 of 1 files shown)


nomike:
I created a regression test for this in PythonSCAD  but as OpenSCAD unfortunately doesn't support Python tests, I wasn't able to port that over easily.

PR #6583, opened 2026-01-29
Add collapse all

Shmerlard:
This PR just add a button for Collapsing every tab, really useful with larger projects.

2026-01-29-121117_hyprshot
2026-01-29-121122_hyprshot


kintel:
Perhaps someone who's a heavier user of Customizer than me can comment, but first impression:
* It would be nice to implement both collapse and expand at the same time
* It's not obvious (to me) that the `^` icon means "collapse all". I know there is a tooltip, but the UI is already a bit crowded.

t-paul:
> * I know there is a tooltip, but the UI is already a bit crowded.

Instead of adding the collapse button directly, make it a "..." menu?

Shmerlard:
I guess something like this?
image

PR #6585, opened 2026-01-29
FIX:fix console font update in MainWindow

dasbidyendu:
In MainWindow.cc, connected the consoleFontChanged signal to setConsoleFont to ensure the console updates its font correctly when preferences change.

### DIFF
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -342,7 +342,6 @@ MainWindow::MainWindow(const QStringList& filenames) : rubberBandManager(this)
           &MainWindow::onTabManagerEditorContentReloaded);
 
   connect(this->console, &Console::openWindowRequested, this, &MainWindow::showLink);
-  connect(GlobalPreferences::inst(), &Preferences::consoleFontChanged, this->console, &Console::setFont);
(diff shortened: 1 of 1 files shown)


src/gui/MainWindow.cc
```diff
@@ -343,6 +343,7 @@ MainWindow::MainWindow(const QStringList& filenames) : rubberBandManager(this)
 
   connect(this->console, &Console::openWindowRequested, this, &MainWindow::showLink);
   connect(GlobalPreferences::inst(), &Preferences::consoleFontChanged, this->console, &Console::setFont);
+  connect(GlobalPreferences::inst(), &Preferences::consoleFontChanged, this->console, &Console::setConsoleFont);
```
kintel: Should this rather replace the setFont connection above?

dasbidyendu: I thought perhaps setFont was related to setting the editor's font so i chose not to replace and just add the new line.

kintel: The editor font is set elsewhere. I think this bug is simply a typo in the slot name

dasbidyendu: Ohh , then no worries i will just replace the line

kintel:
Thx!

PR #6590, opened 2026-01-31
FIX: Skip compilation initially when editor is empty

### LINKED ISSUE #6580: Automatic Reload and Preview attempts to preview the initial, empty document
Launch OpenSCAD with Automatic Reload and Preview enabled.

It will immediately output in the Console window:
```
Geometries in cache: 0
Geometry cache size in bytes: 0
CGAL Polyhedrons in cache: 0
CGAL cache size in bytes: 0
Normalized tree has 1 elements!
Compile and preview finished.
Total rendering time: 0:00:00.254
```

..which feels a bit unnecessary.

dasbidyendu:
Only trigger actionRenderPreview when auto-reload is enabled and the active editor contains non-whitespace text. Previously the preview was requested whenever renderedEditor was null and auto-reload was on, which could cause unnecessary preview requests or attempts to render empty content. This adds a trimmed non-empty check on the active editor before requesting a preview.

Fixes #6580

### DIFF
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -3560,8 +3560,7 @@ void MainWindow::onTabManagerEditorChanged(EditorInterface *newEditor)
 
   // If there is no renderedEditor we request for a new preview if the
   // auto-reload is enabled.
-  if (renderedEditor == nullptr && designActionAutoReload->isChecked() &&
-      !MainWindow::isEmpty()) {
+  if (renderedEditor == nullptr && designActionAutoReload->isChecked() && !MainWindow::isEmpty()) {
     actionRenderPreview();
   }
 }


src/gui/MainWindow.cc
```diff
@@ -3560,7 +3560,8 @@ void MainWindow::onTabManagerEditorChanged(EditorInterface *newEditor)
 
   // If there is no renderedEditor we request for a new preview if the
   // auto-reload is enabled.
-  if (renderedEditor == nullptr && designActionAutoReload->isChecked()) {
+  if (renderedEditor == nullptr && designActionAutoReload->isChecked() &&
+      !activeEditor->toPlainText().trimmed().isEmpty()) {
```
kintel: You might want to simply call `isEmpty()` on this MainWindow instance.

dasbidyendu: Thanks! I double-checked MainWindow::isEmpty(), but it uses toPlainText().isEmpty() without trimming. I want to make sure we don't trigger auto-reloads on just spaces or newlines, so I think the explicit trimmed() check is safer here to avoid unnecessary renders.

kintel: I believe the intent of what you need and what isEmpty() tries to achieve is the same, so I'd still use isEmpty() here, and if needed change the implementation of isEmpty().

otoh, if this is meant to not render on startup, there will never be whitespace in the file :)

dasbidyendu: Well , that's a valid point , i was over guarding things. I will just use isEmpty()

kintel (approved):

kintel:
@dasbidyendu It looks like this may have caused a bug: #6598 - could you take a look?

Issue #6584, opened 2026-01-29
View pivot gets weird after double clicking


Shmerlard:
**Describe the bug**
Double click on object to center the view is a bit broken,
sometimes when i double click the pivot point gets totaly weird
it happens sometimes after panning the view

**To Reproduce**
1. i start with `QT_QPA_PLATFORM=xcb QT_QPA_PLATFORMTHEME=qt6ct ./build/openscad`
2. open a file
3. start to double click everywhene

**Expected behavior**
the origin of the view should be where i double click

**Screenshots**

https://github.com/user-attachments/assets/75bdbfe1-7824-447f-ba14-20bbe71ef63e

**Environment and Version info (please complete the following information):**
 - OS: archlinux
 - System: intel pc 64-bit
 - OpenSCAD Version  2026.01.29 (git 6fcd56761) (compiled from source)

Shmerlard:
solved by using the appimage version

crozone:
Please re-open. This is still an issue as of 2026.07.04 git 1f65580c on Windows.

Double-clicking moves the view translation origin significantly towards the current view's positive Z. It's almost as if the camera is zooming out, however the viewport control "Distance" does not change, the origin actually moves very far away from the object.

jordanbrown0:
I can't explain what it **is** doing, but I agree that it isn't right.

This is perhaps most easily demonstrated with a simple `cube(10);`, with the crosshairs enabled.  Double-click on various points on the cube.  One would think that that would cause the origin to be on the surface of the cube, but the crosshairs show that the origin is often or usually far off of the cube.

On Windows 11:  Works on 2026.01.19.  Fails on 2026.01.25.  This coincides with the Qt6 transition.

jordanbrown0:
This is similar to #5255 but I don't think it's the same thing.  This one involves much larger errors, and appears correlated to the Qt6 transition in early 2026, while that one is from 2024.

crozone:
It looks like Qt6 was enabled on all platforms by https://github.com/openscad/openscad/pull/6562 on 2026-01-25.

The final development snapshot before this change is 2026.01.19.

* 2026.01.19 (tested with dev snapshot "OpenSCAD-2026.01.19-x86-64.zip": The issue is not present. OpenSCAD is using the old style UI indicating it is on Qt5.
[...]

PR #6591, opened 2026-01-31
Fix double click object on wayland

Shmerlard:
fix for #6584 
so after some digging it turns out that the reason for this was wayland
when debugging the program using qt creator i've noticed it doesnt return the currect z, so clicking sometimes didnt work at all, and when it did it returned the wrong coordinates.
tested it on wayland and seems to workd now

https://github.com/user-attachments/assets/ebe18372-25ec-4114-8382-da51fc21187c

### DIFF
--- a/src/gui/QGLView.cc
+++ b/src/gui/QGLView.cc
@@ -197,45 +197,44 @@ void QGLView::paintGL()
   GLView::paintGL();
 
   if (mouse_has_pending_double_click) {
-      glMatrixMode(GL_PROJECTION);
-      glPushMatrix(); 
-      
-      glMatrixMode(GL_MODELVIEW);
-      glPushMatrix(); 
+    glMatrixMode(GL_PROJECTION);
+    glPushMatrix();
 
-      this->setupCamera();
+    glMatrixMode(GL_MODELVIEW);
+    glPushMatrix();
 
-      GLdouble matModelView[16];
-      GLdouble matProjection[16];
-      GLint viewport[4];
+    this->setupCamera();
 
-      glGetDoublev(GL_MODELVIEW_MATRIX, matModelView);
-      glGetDoublev(GL_PROJECTION_MATRIX, matProjection);
-      glGetIntegerv(GL_VIEWPORT, viewport);
+    GLdouble matModelView[16];
+    GLdouble matProjection[16];
+    GLint viewport[4];
 
-      glPopMatrix(); 
-      glMatrixMode(GL_PROJECTION);
-      glPopMatrix();
-      glMatrixMode(GL_MODELVIEW);
+    glGetDoublev(GL_MODELVIEW_MATRIX, matModelView);
+    glGetDoublev(GL_PROJECTION_MATRIX, matProjection);
+    glGetIntegerv(GL_VIEWPORT, viewport);
 
-      const double dpi = this->getDPI();
-      const double x = mouse_pending_double_click_pos.x() * dpi;
-      const double y = viewport[3] - (mouse_pending_double_click_pos.y() * dpi);
+    glPopMatrix();
+    glMatrixMode(GL_PROJECTION);
+    glPopMatrix();
+    glMatrixMode(GL_MODELVIEW);
 
-      GLfloat z = 1.0f;
-      glReadPixels((GLint)x, (GLint)y, 1, 1, GL_DEPTH_COMPONENT, GL_FLOAT, &z);
+    const double dpi = this->getDPI();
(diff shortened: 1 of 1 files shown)


src/gui/QGLView.cc
```diff
@@ -262,47 +262,44 @@ void QGLView::mouseDoubleClickEvent(QMouseEvent *event)
 {
   QOpenGLContext *oldContext = getGLContext();
   this->makeCurrent();
-  setupCamera();
 
-  int viewport[4];
-  GLdouble modelview[16];
-  GLdouble projection[16];
+  glMatrixMode(GL_MODELVIEW);
```
kintel: I assume that the point is to restore the matrices modified by setupCamera(). However, you only push/pop GL_MODELVIEW, and setupCamera also changes GL_PROJECTION.

src/gui/QGLView.cc
```diff
@@ -262,47 +262,44 @@ void QGLView::mouseDoubleClickEvent(QMouseEvent *event)
...
+
+  glGetDoublev(GL_MODELVIEW_MATRIX, matModelView);
+  glGetDoublev(GL_PROJECTION_MATRIX, matProjection);
   glGetIntegerv(GL_VIEWPORT, viewport);
-  glGetDoublev(GL_MODELVIEW_MATRIX, modelview);
-  glGetDoublev(GL_PROJECTION_MATRIX, projection);
+
+  glPopMatrix();
+
+  this->paintGL();
```
kintel: Why do we need to paint here? I though the idea was to simply read the existing z buffer?

src/gui/QGLView.cc
```diff
@@ -262,47 +262,44 @@ void QGLView::mouseDoubleClickEvent(QMouseEvent *event)
...
-  if (z == 1) {
-    setGLContext(oldContext);
-    return;  // outside object
-  }
+  GLfloat z = 1.0f;
+  glReadBuffer(GL_BACK);
+  glReadPixels((GLint)x, (GLint)y, 1, 1, GL_DEPTH_COMPONENT, GL_FLOAT, &z);
 
-  GLdouble px, py, pz;
+  if (z < 0.9999f) {  // if z is 1 we clicked on empty space
```
kintel: Didn't comparing with 1 work?

src/gui/QGLView.cc
```diff
@@ -262,47 +262,44 @@ void QGLView::mouseDoubleClickEvent(QMouseEvent *event)
...
-    return;
-  }
+  const double y = viewport[3] - (event->pos().y() * dpi);
 
-  if (z == 1) {
-    setGLContext(oldContext);
-    return;  // outside object
-  }
+  GLfloat z = 1.0f;
+  glReadBuffer(GL_BACK);
```
kintel: Since the front buffer is what was most recently displayed, should we read from that? ..or is this a side effect of manually calling paintGL?

src/gui/QGLView.cc
Shmerlard: yeah missed that

src/gui/QGLView.cc
Shmerlard: it does but just as safety measure
at first i didnt realize the bug i had so i though maybe changing this will solve

Shmerlard:
so instead of calling the paint function, (which does work but is unnecessary) ive created instead a more simple way to just do the calculations inside.
it does compile and run but seems like  354d82627e7cd35f2945c5e14500d6704083bf5b is breaking it.
if i rebase to 587202168679a476e4153202ce18cbace85d0854 or later it works, but gives a very weird bug where preview doesnt work currectly (as shown in the first image)


2026-02-01-171838_hyprshot
2026-02-01-171809_hyprshot

kintel:
Is the preview bug also in master, or just exposed when using this PR?

Shmerlard:
> Is the preview bug also in master, or just exposed when using this PR?

I just checked it, the preview bug happens in the master too, i downloaded the appimage  of c2aa013 , started it as usual (no environmental variables) and the preview bug is there.
weirdly enough the bug the double click isn't.

kintel:
Is this just an F9 vs. F12 issue? Try pressing F9 after preview.
We may have broken a setting..

Shmerlard:
after compiling c2aa013c again and running with `QT_QPA_PLATFORM=xcb` the preview error its still there,  and the camera bug returned so the appimage fixes this issue, (or maybe the X11 mode im running it on is causing it).
anyway this issue is somewhat solved

Shmerlard:
yes F9 worked

PR #6593, opened 2026-01-31
feat: tab owns language, language selector, language icons

nomike:
Some time ago we added some GUI related features to PythonSCAD in regards to Python handling.

This is my attempt of back-porting this to OpenSCAD.

What it does:

- the language (scad, py) is stored per tab now, no longer globally
- tabs have an icon depending on whether it's a python or an SCAD script
- there is a menu on the right of the status bar which allows the user to select whether it's SCAD or Python. It has an auto-detect option which uses the file extension. This is especially useful when you have a new file which isn't saved yet.


nomike:
image

image

image

nomike:
I want to port the session management feature I just implemented as well, but I thought it would be nice to submit this one first as it's a nice feature anyway and as the session management taps into that as well.

t-paul:
The tab specific handling seems like a good step forward, the amount of new `#ifdef`s in other places is a bit concerning though. Trying to maintain that in the future will become more and more difficult.

nomike:
I will have a look whether all these IFs are really necessary.

The thing is: If `-DENABLE_PYTHON=OFF` you don't want to have Python as an option in the language selection menu.

Actually you don't need the menu at all in this case, But what if other languages (e.g. lua, guile/scheme, etc.) get added at some point?

I will rebase the branch and clean it up a bit. There is a conflict due to the recent refactoring of the main window which I will solve as well.

t-paul:
> Actually you don't need the menu at all in this case, But what if other languages (e.g. lua, guile/scheme, etc.) get added at some point?

That's exactly the point, imagine all those places having one #ifdef block for each of the languages, maybe some behaving slightly differently.

One option is to use data structures combining combining all the information (e.g. file extensions, names, ...). That structure could then for example be assigned to the tab so no further switch are needed. If Python is disabled, it just needs to be prevented that structure is assigned to anything, but no further checks are needed.

For example saving does not need any switch/#ifdef but instead can just ask for `tab->lang->extension()` or Tab setup can configure additional things like `tab->lexer = tab->lang->create_lexer()` without needing to know what is enabled or not.

A very simple example is maybe `Feature.h/.cc` combining the name, description and activation state of experimental features. In release mode, the state is forced to be always false, so even though the code checks for the feature to be enabled, it never is.

nomike:
For now I managed to get rid of almost all of the `#ifdef ENABLE_PYTHON` blocks.
The only two left are in regards to the status-bar menu and are to prevent the python option to be shown when python is not enabled.

Is this better or do you prefer to refactor this a little more?

t-paul:
@kintel any objections to getting this in before the release? I don't think there's any bigger risk with the GUI changes.

kintel:
No conceptual concerns. I just haven't had bandwidth to review yet.

t-paul:
Rebased in #6666

resources/icons/filetype-python.svg
```diff
@@ -0,0 +1,24 @@
...
+              mime: contentType,
+              method,
+              contentType
+            }, '*');
+          }
+          this.removeEventListener('readystatechange', _);
+        }
+      })
+    }
+  }
```
t-paul, line 23: @nomike where did you get that icon from. Are you aware this contains a JavaScript payload?

nomike: I figured that out yesterday.
It was a bug in a firefox extension I'm using which injected that JavaScript.
Nothing malicious and it was apparently already fixed by the add-on authors, but nonetheless, the script shouldn't be in the files.

As this PR was rebased, I will have a look in the new one.

PR #6606, opened 2026-02-03
OpenGL clean up

ali1234:
For your consideration. This is not ready to be merged in any way, and some random fixes to other bugs may have slipped in by mistake.

 * Make a Shader class for shaders. It is RAII. I haven't quite decided how its API should be shaped yet.
 * Make the renderers that can use shaders *always* use them.
 * Move the code that checks VertexState type into a Shader draw function. This function should eventually end up as just a call to vertex_state->draw() at which point it can go.
 * ViewEdges is now the "default" shader, so rename it "Main".
 * Remove destructors that don't do anything.
 * Remove VertexStateFactory and VertexState subclasses. This huge amount of code can be replaced by just adding a property to the base VertexState.
 * Make the "main" shader do deferred rendering.
 * Remove CGALRenderer entirely as it is theoretically no longer needed.

Still TODO:

 * Make "main" shader also generate the "IDs" texture for mouse selection in the same pass, then remove MouseSelector shader.
 * Make fragid_color be an attribute rather than a uniform (allows shader->draw() to be removed?)

Other observations:

 * I probably broke NULLGL (if it even still works). If attempting a real clean up, I would just rip it out entirely and then put it back afterwards. I'm not sure how much of it is really needed vs just leaving files out of the build entirely.
 * Nothing actually extends Renderer except VBORenderer. Everything else extends VBORenderer, so not sure why they are separate classes. Perhaps something to do with NULLGL? If so, I think it can be done in a better way.
 * If we want to remove fixed function stuff eg glProjectionMatrix() we need a way to send eg the projection matrix into the OpenCSG vertex shader as a uniform. This is currently not possible because the vertex shader is passed into OpenCSG as source code and then becomes a black box. Can't set a uniform on a shader without a handle to it. Unless we can pull it back out of the GL environment, but that seems fragile.

### DIFF
--- a/shaders/Main.frag
+++ b/shaders/Main.frag
@@ -5,6 +5,8 @@ uniform int show_edges;
 varying vec4 color;
 varying vec3 vBC;
 varying float shading;
+varying vec3 normal;
+varying vec4 position;
 
 vec3 smoothstep3f(vec3 edge0, vec3 edge1, vec3 x) {
   vec3 t;
@@ -22,5 +24,7 @@ float edgeFactor() {
 
 void main(void) {
   vec4 color_edge = vec4((color.rgb + vec3(1))/2, 1.0);
-  gl_FragColor = mix(color_edge, vec4(color.rgb * shading, color.a), show_edges>0 ? edgeFactor() : 1.0);
+  gl_FragData[0] = position;
+  gl_FragData[1] = vec4(normal, 1);
+  gl_FragData[2] = mix(color_edge, vec4(color.rgb * shading, color.a), show_edges>0 ? edgeFactor() : 1.0);
 }
--- a/shaders/Main.vert
+++ b/shaders/Main.vert
@@ -6,13 +6,17 @@ attribute vec3 barycentric; // barycentric form of vertex coord
 varying vec3 vBC;           // varying barycentric coords
 varying float shading;      // Will be multiplied by color in fragment shader
 varying vec4 color;         // per-vertex colors
+varying vec3 normal;
+varying vec4 position;
 
 void main(void) {
   gl_Position = gl_ModelViewProjectionMatrix * gl_Vertex;
   vBC = barycentric;
-  vec3 normal, lightDir;
+  vec3 lightDir;
   normal = normalize(gl_NormalMatrix * gl_Normal);
+  normal = normal.z < 0 ? -normal : normal;
   lightDir = normalize(vec3(gl_LightSource[0].position));
   shading = 0.2 + abs(dot(normal, lightDir));
   color = gl_Color;
+  position = gl_ModelViewMatrix * gl_Vertex;
 }
--- a/shaders/Post.frag
+++ b/shaders/Post.frag
@@ -1,0 +1,59 @@
+#version 130
+out vec4 FragColor;
+
+in vec2 TexCoords;
+
+uniform int ssao;
+
+uniform sampler2D gPosition;
+uniform sampler2D gNormal;
+uniform sampler2D gAlbedo;
+uniform sampler2D texNoise;
+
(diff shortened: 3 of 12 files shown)


jordanbrown0:
I don't really know beans about this part of the program, but here's a thought.

Rendering, whether Manifold or CGAL, whether via F6 or render(), should yield a pile of faces that then fed into a pipeline that's equivalent to previewing a single polyhedron() with those faces.

Your statement that "renderer for CGAL does not use shaders at all" is worrisome.  It shouldn't be *possible* for it to be different, because it should be pouring its list of faces into the same pipeline that a preview of a polyhedron() uses.

Put still another way, neither the CGAL support nor the Manifold support should have **any** GL stuff in it.

jordanbrown0:
Caveats:
* Handwaving on color support.  It should probably be equivalent to N polyhedra, or to a single polyhedron that somehow magically has different-color faces.
* Wireframe mode would lead from the renderer into a different pipeline that would be different from the preview pipeline, at least at the top.  But it should be the same for both CGAL and Manifold.

ali1234:
> Put still another way, neither the CGAL support nor the Manifold support should have **any** GL stuff in it.

Note that the renderer as in F6 renderer does not have any GL stuff in it at all. However, in the codebase "Renderer" is the base class for the code that draws triangles on the screen, so for example the F5 "preview" drawing code is a subclass of Renderer called OpenCSGRenderer. This is confusing but there is not much can be done about it short of renaming everything. It's confusing for users too. To most people outside of OpenSCAD (developer or user) "rendering" means "drawing a picture" but in OpenSCAD UI it means "generating a mesh" - but in the source code it again means "drawing a picture".

In theory I agree that a mesh is a mesh and there should not be two ways of drawing one, but that isn't how the code works currently. There is the "old" renderer called CGALRenderer that works with the old CGAL backend, and the "new" renderer called PolySetRenderer that only works with Manifold output. In some cases the CGAL output can be converted to Manifold output which allows the new renderer to work with CGAL output, but not always for reasons I don't yet fully understand.

ali1234:
>     * Wireframe mode would lead from the renderer into a different pipeline that would be different from the preview pipeline, at least at the top.  But it should be the same for both CGAL and Manifold.

CGALRenderer draws wireframes or edges by iterating over the mesh a second time, generating line calls instead of triangle calls.

All other renderers generate the wireframe in the same shader pass as triangles by drawing the triangles a different colour when the generated fragment is very close to the edge of the triangle. This is entirely in the GLSL shader code running on the GPU and (after my patch) is simply a matter of setting a variable in the shader to turn wireframe on or off.

> color support

Not sure what you mean here. I haven't looked into color support much as it "just works". The `fragid_color` stuff has nothing to do with the visible colour of the objects. It is a special shader that draws everything with color = object ID, so that you can check the ID of an object at a specific x,y position in the image by looking at the "color" of the pixel. Currently this involves doing a whole separate draw pass, but visual colour and ID colour could both be out by the same shader writing to two different buffers. This would simplify a lot of things and outputting to multiple buffers is anyway required for the deferred shading that allows post processing effects like SSAO, which is what I am working towards.

jordanbrown0:
> In theory I agree that a mesh is a mesh and there should not be two ways of drawing one

I'm a bit surprised, but not shocked.  I'm just suggesting that a goal should be to to eliminate that divergence.  (And eliminate the difference from Preview too, though that's trickier.)

(Yes, I know that Preview is very very different from the other two.  But Preview of a single polyhedron shouldn't be.  Imagine turning the mesh resulting from the others into a single PolyhedronNode, and feeding it into the preview pipeline.)

>>color support

>Not sure what you mean here.

I said "just turn it into a single polyhedron", but that's not really enough because you need a polyhedron where different faces are different colors, and we don't have a way to say that.

PR #6612, opened 2026-02-04
fix(export): Ensure file name extension on export

LightDestory:
This PR attempts to fix a bug of the exportAction: when exporting STLs and 3MFs if the user do not type the full filename (TEST.3MF) the software will save an extensionless file.

image

I am unable to building locally, it could be useful to make Actions generate artifacts builds.

### DIFF
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -2442,17 +2442,8 @@ QString MainWindow::enforceExportSuffix(const QString& filename, const QString& 
   const QFileInfo info(filename);
   if (info.suffix().isEmpty()) {
     const auto suffixedName = filename + "." + normalizedSuffix;
-    const QFileInfo suffixedInfo(suffixedName);
-    if (suffixedInfo.exists()) {
-      const auto text =
-        QString(_("%1 already exists.\nDo you want to replace it?")).arg(suffixedInfo.fileName());
-      const auto overwrite = QMessageBox::warning(this, windowTitle(), text,
-                                                  QMessageBox::Yes | QMessageBox::No, QMessageBox::No);
-      if (overwrite != QMessageBox::Yes) return {};
-    }
     return suffixedName;
   }
-
   return filename;
 }
 


t-paul:
The question is if there's any way of preventing the confusing (and potentially double) overwrite check.

LightDestory:
@t-paul Sorry for the mistake, the double overwrite check has been removed.

t-paul:
Sorry, I guess I did not get the point across. The problem is not that you had the code there, it's a more general problem that's not solved by removing the check.

1) Imagine we have the files `x` and `x.stl` in the output folder already.
2) Now you save STL and in the file selector, you just give it `x`. Now the file selector (at least the native one on Linux) will give an overwrite warning which makes sense as the file `x` exists.
3) The *ensure-extension* now adds the `.stl` and the check (sort-of) correctly triggers again as we have `x.stl` file in the output folder.

For this scenario the first check is the offending one, not the one you have in the original PR code.

LightDestory:
Thanks for the explanation.
Assuming I'm using Windows, the first overwrite check is performed with `QFileDialog::getSaveFileName()`, which performs cross-platform magic behind the scenes.

It's a QT component... should I try to modify it?

Maybe I have underestimated the bug and it is out of my league. I am drafting this PR for now and looking for solutions, feel free to close it if it is not viable for a newbie like me

t-paul:
Yes, that's part of Qt and if you wonder why I have that explanation, that's because I stumbled over that myself earlier and had no idea how to solve it at the time. Maybe we can find good solution throwing 4 eyes on it :grinning:.

My wish would be that it's solved completely in the Qt selector, but if that's not possible if we can find a way to disable the internal check we could just have the one outside the getSaveFileName().

As we are now essentially Qt6 focused, it would be enough to find a solution that works well for that (while still compiling with Qt5).

t-paul:
This looks like the solution: https://doc.qt.io/qt-6/qfiledialog.html#defaultSuffix-prop

MichaelAtOz:
I just noticed that .OBJ export has the same missing extension problem.

From the code, it seems every generic export types would be the same.
So this should fix .OBJ too. Once finished.

Issue #6321, opened 2025-10-25
Minkowski still fails on corner-case Manifold results


revk:
**Describe the bug**

Minkowski still failing in Manifold when working in CGAL

**To Reproduce**

Render this in each engine

[FaikinT.zip](https://github.com/user-attachments/files/23141985/FaikinT.zip)

**Expected behavior**

Minkowski working :-)

**Code reproducing the issue**

This is the challenge, the above zip (scad) file is complicated.

**Screenshots**

Good (CGAL)

<img width="992" height="563" alt="Image" src="https://github.com/user-attachments/assets/3e44a2c3-c50c-479c-b0aa-3b5215c8af1a" />

Bad (manifold)

<img width="998" height="558" alt="Image" src="https://github.com/user-attachments/assets/93c52928-5042-46ae-a9b8-7395669a21bd" />

**Environment and Version info (please complete the following information):**

MacOS Version 15.6.1 (24G90)
OpenSCAD [OpenSCAD](https://www.openscad.org/) version 2025.10.25 (git f9829891a)

I'll try and make a more cut down example to narrow down the problem area and post an update.

revk:
OK I am commenting out "parts" from `module parts_top` to find which is causing issue.

- I commented out `C19` and no help.
- I commented out `R6` as well, and now a render just hangs, and does not complete at all.

Have I found a new bug?

- I comment out `R1` as well, and renders quickly again, still wrong.
- I comment out `C5` as well, and hangs again. But this time reloading and trying again rendered (still wrong).

So a tad weird.

Commenting out C19 R6 R1 C5 V4 J2 (not D1) V5 V3 C16 and D3 is another hang scenario.

Commenting all but D1 looks like it worked, messy, but the hole in D1 is there.

OK narrower case - all in `parts_top` commented out apart from D1 and J1 causes the hole for D1 to be missing. So that is the clash somehow.

<img width="421" height="540" alt="Image" src="https://github.com/user-attachments/assets/87e0ad63-361a-4bb6-8b2c-746b6fd1dfc9" />

<img width="995" height="592" alt="Image" src="https://github.com/user-attachments/assets/47008887-1b65-47d3-ba18-282effbcc789" />

kintel:
Did you intend to close this?
Note that: "hang" is a common symptom of #6297, as long as it eventually (buy may take hours) finishes.

revk:
No I did not intend to close it, D'Oh.

[...]

PR #6631, opened 2026-02-08
Optionally use Manifold's Minkowski sum implementation

ali1234:
Bumps the Manifold submodule version and adds a compile-time option to enable it (default off).


ali1234:
This seems to be slightly faster than the CGAL method within the Manifold backend.

OpenSCAD was built with the new variable set to YES and unset, then tested with this python script:

```python3
import subprocess


TEST = """
$fs = 0.2;
$fa = 0.2;

module A() {
    difference() {
        cube(10, center=true);
        sphere(6);
    }
}

module B() {
    sphere(1);
}



minkowski() {
    A();
    B();
}
"""

with open('test.scad', 'w') as f:
    f.write(TEST)


OPENSCAD = {
    "manifold": "cmake-build-releaseminkowski/openscad",
    "cgal": "cmake-build-release/openscad",
}

for backend in ('manifold', 'cgal'):
    for method, openscad in OPENSCAD.items():
        if backend == 'cgal' and method == 'manifold':
            continue
        print(f'backend={backend}, method={method}:')
        subprocess.check_output(['time', openscad, f'--backend={backend}', '--render', '-o', f'{backend}-{method}.png', 'test.scad'])
        print('---------------\n')
```

Result:

```
backend=manifold, method=manifold:
Geometries in cache: 3
Geometry cache size in bytes: 1624552
CGAL Polyhedrons in cache: 2
CGAL cache size in bytes: 22208250
Total rendering time: 0:03:34.837
   Top level object is a 3D object (manifold):
   Status:     NoError
   Genus:      5
   Vertices:    80709
   Facets:     161434
    Real: 0:03:35, User: 0:11:12, Sys: 0:01:49, CPU: 363%, Memory: 3,887.37MB
---------------
```
manifold-manifold

```
backend=manifold, method=cgal:
Geometries in cache: 3
Geometry cache size in bytes: 1624552
CGAL Polyhedrons in cache: 2
CGAL cache size in bytes: 22348500
Total rendering time: 0:04:30.142
   Top level object is a 3D object (manifold):
   Status:     NoError
   Genus:      5
   Vertices:    81270
   Facets:     162556
    Real: 0:04:30, User: 0:12:30, Sys: 0:05:57, CPU: 410%, Memory: 3,647.82MB
---------------
```
manifold-cgal

```
backend=cgal, method=cgal:
[edit: this ended up taking 2 hours 23 minutes]
```

ali1234:
This completely cuts out all the optimization in OpenSCAD that sometimes uses Manifold and sometimes uses CGAL depending on the operands, and instead just goes directly to using the implementation inside Manifold. 

In other words `manifold-applyops-minkowski.cc` is completely unused when this is enabled (or at least it should be if I've done it right).

The implementation seems to be part of Manifold proper now. It is documented here: https://manifoldcad.org/docs/html/classmanifold_1_1_manifold.html#ae174a8240802779eca5a0416af7ea78b

It was committed only a couple of weeks ago so you need a very recent version of Manifold.

Also note that Manifold got a matching `MinkowskiDifference` operation at the same time. If `MinkowskiSum() { X(); sphere(r)}` is equivalent to a 3D `offset(r)`, then `MinkowskiDifference(){X(); sphere(r);}` is equivalent to a 3D `offset(-r)`. In other words those two functions are everything needed to implement a 3D offset() that can be used to "round" or "bevel" objects the same way we commonly do in 2D with `offset()`. It won't directly allow `delta` and `chamfer` modes in 3D though.

kintel (approved):
Nice, will merge this as it's nicely encapsulated.
Will have to run some torture tests against it before enabling it by default

ali1234:
Note that it bumps Manifold to whatever was at the tip of their master branch yesterday. Minkowski isn't in a tagged release yet.

kintel:
Yeah, that's fine. Once they have a release we can move the tag

ali1234:
I tested the example from #6321 and it produces identical output with both minkowski implementations but it seems like those coincident unions got worse. Nightly is the same (ie before Manifold bump):

manifold-manifold

PR #6632, opened 2026-02-08
Quick stab at showing FPS on the status line

### LINKED ISSUE #6629: Report display FPS
When previewing[*] a complex model, rotate/pan/zoom can get slow.  Rather than just saying "it's a little laggy" / "it's laggy"  / "it's unusable", it would be nice to have a performance measurement.

Measure the time required to generate a frame.  Invert that to yield FPS, since that seems like the more common metric.  Either report the time for the most recent frame, or perhaps a decaying average of the frames generated.

Report actual time required.

**Additional context**
Note that measuring "true" frames per second - how many frames were generated during a given second - is meaningless, because frames are generated only when the user moves the camera.

jordanbrown0:
When and if done:  Fixes #6629.

Measures the time for GLView::paintGL(), which seems to be shared across all forms of rendering.  Inverts it to get FPS, then uses an exponential-decay smoother to reduce noise.

image

Known issues:
* Should reset smoothing on each new preview/render.
* Unrelated to animation FPS, needs clarity in presentation.
* value for smoothing parameter ("decay") is debatable.

More that I thought of after commit:
* Probably interacts very poorly with animation, no matter what you do.
* Measured FPS varies pretty widely even for a single model.  Maybe based on current viewport?
* FPS of rendered image is really high (~6000FPS).  Don't know if that's real or not.
* Currently uses the equivalent of `%.4g`, which is too many digits most of the time.  But having fewer than that pushed rendered-image FPS into exponential notation.

### DIFF
--- a/src/gui/QGLView.cc
+++ b/src/gui/QGLView.cc
@@ -215,13 +215,21 @@ void QGLView::resizeGL(int w, int h)
 
 void QGLView::paintGL()
 {
+  const std::chrono::steady_clock::time_point begin{std::chrono::steady_clock::now()};
   GLView::paintGL();
+  const std::chrono::steady_clock::time_point end{std::chrono::steady_clock::now()};
+  const std::chrono::microseconds us{std::chrono::duration_cast<std::chrono::microseconds>(end - begin)};
+  double d = us.count();
+  double rawFPS = 1000.0 * 1000.0 / d;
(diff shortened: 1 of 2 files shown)


jordanbrown0:
@ali1234 recommends against exponential-decay smoothing, and recommends other strategies.  It's good enough for my purposes and I'm not interested in spending time on something cleverer, but I wouldn't object if somebody else did.

PR #6638, opened 2026-02-09
output from File->Examples and File->Reload were printed in the termi…

kintel:
…nal instead of console


MichaelAtOz:
Tick on Windows.

PR #6639, opened 2026-02-09
Print time-to-first-frame on first frame

dasbidyendu:
Changed the rendering time message to 'Time to first image' and defer printing until the first rendered frame is actually displayed. Added QGLView::frameRendered signal (emitted from paintGL), MainWindow::onFrameRendered slot, and an isWaitingForFirstFrame flag that is set for preview compilations; connect the signal to the slot so renderStatistic.printRenderingTime() runs when the first frame finishes. Removed the immediate print call after compile to ensure timing reflects time-to-first-frame.

### DIFF
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -1056,10 +1056,18 @@ void MainWindow::compileCSG()
     this->thrownTogetherRenderer = std::make_shared<ThrownTogetherRenderer>(
       this->rootProduct, this->highlightsProducts, this->backgroundProducts);
     LOG("Compile and preview finished.");
-    renderStatistic.printRenderingTime();
+    renderStatistic.printRenderingTime();  // Removed: printRenderingTime()
     this->processEvents();
   } catch (const HardWarningException&) {
     exceptionCleanup();
+  }
+}
+
+void MainWindow::onFrameRendered()
+{
+  if (this->isWaitingForFirstFrame) {
+    this->renderStatistic.printRenderingTime();
+    this->isWaitingForFirstFrame = false;
   }
 }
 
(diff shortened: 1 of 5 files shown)


src/gui/MainWindow.cc
```diff
@@ -1056,13 +1056,21 @@ void MainWindow::compileCSG()
     this->thrownTogetherRenderer = std::make_shared<ThrownTogetherRenderer>(
       this->rootProduct, this->highlightsProducts, this->backgroundProducts);
     LOG("Compile and preview finished.");
-    renderStatistic.printRenderingTime();
+    renderStatistic.printRenderingTime();  // Removed: printRenderingTime()
```
kintel, line 1059: Did you intend to remove this line?

src/gui/QGLView.cc
```diff
@@ -208,6 +205,7 @@ void QGLView::paintGL()
                     .arg(size().rheight());
     statusLabel->setText(status);
   }
+  emit frameRendered();
```
kintel, line 208: Note: This doesn't actually emit when the frame is rendered, only when the frame is queued for rendering. It's an improvement over today, but I wonder if  `QOpenGLWidget::frameSwapped()` is a better signal

src/gui/MainWindow.cc
dasbidyendu: @kintel i don't know how that line came into the diff , my code is clean , maybe some copilot stuff from github desktop.
image
no i did not intend to remove that line.

src/gui/QGLView.cc
dasbidyendu: I will try and get back.

kintel (changes requested):

dasbidyendu:
Hey @kintel health has been an issue from a few weeks so I haven't got the time to work on it yet, will get back on it as soon as I recover.

PR #6642, opened 2026-02-09
Adding the ability to dynamically change the parameters [Ready for initial review]

### LINKED ISSUE #6592: Adding dynamic visibility and lockability of parameters using other parameters
**The Issue**
I've spent a lot of time creating gridfinity bins thanks to [this](https://github.com/ostat/gridfinity_extended_openscad) repo, but one problem I've always had was that the customizer was always too crowded, and its a big problem in bigger projects.
for instance if i have a "label settings" tab, and inside there is `label_enable = false` and `label_size = 5`
there is no reason to allow the change of label size if enable is false.

**Possible Solution**
The soltion Im suggesting is giving the user who writes the code the power to control the visibility or locked state of one variable based on the value of one variable. for instance by using special arguements we can control, for instance
```
label_enable = false; // &("label_size")
label_size = 1.0;
```
and it can also be extended where `%` controls the lock and `*` controls the visibility for instance.
of course these characters are just an example but it shows the idea.

Now i dont have a problem working on it myself, but i would like to know if its even feasible,
because i know that openscad is also used in makerworld and other ways that im not sure how they are implemented, so before i start is it something that the team is interested in?
are there any other things i need to know?
also I dont have much experience in working on big projects and i would like to try and improve my cpp skills
thanks in advance!

of course its not necessary but it would be very neat and i think a cool feature

**Additional context**
I've hard coded the value to make an example

https://github.com/user-attachments/assets/27f54e6f-8cdd-4960-a29b-d813d4ee741d

/assign-me

Shmerlard:
this PR introduces a native syntax for defining parameter attributes (description, group, range, etc.) directly in the code using an object-like block. This moves away from relying solely on magic comments and provides a foundation for more dynamic Customizer behavior.
fix #6592 

# Key changes:

- added '@' token and attribute block in the parser to detect `@{...}`
- added `locked` and `hidden` in ParameterObject 
- parameterObject now holds the expression for `locked` and `hidden` to be able to set during parameter change
- a parameter object now has dependencies variable and `expression` now has `collectDependencies` functionality
- the parameter widget now has a dependency map, a session and context
- he can update it when parameter has changed
- the `fromAssignment` function now checks if the assignment has `@{...}` block, if yes then it calls `fromObjectExpression` else it continues as before

example:
```
// Old way
/* [Settings] */
// The width of the box
width = 10; // [1:100]

// New way
label = "Box Width";
@{ group = "Settings"; 
	description = label; 
	range = [1:100];}
width = 10; 
```

there are 3 types of attributes
1. fully dynamic (locked and hidden) their expression and dependency is tracked by the parameter object and able to change by other variables
2. semi-dynamic (group desc) can be depend on other variables but only calculated at startup
3. static (range) can only be used as literal

i've currently implemented `hidden`, `locked`, `desc`, `group`, `min`, `max`, `step`, `range` and `slider` to disable the slider in numbers.
and in the future it will be able to use
`isColor` - to force a string variable to be generated as a color picker widget
`multiline` - to force a string to be generated as mulitline widget
`tabColor` - to change the color of the group

https://github.com/user-attachments/assets/ff49ffb6-9a95-4b81-96c5-0511f619292b

### DIFF
--- a/src/core/SourceFile.cc
+++ b/src/core/SourceFile.cc
@@ -198,6 +198,19 @@ std::shared_ptr<AbstractNode> SourceFile::instantiate(
   return node;
 }
 
+std::shared_ptr<const FileContext> SourceFile::instantiateVariablesOnly(
+  const std::shared_ptr<const Context>& context) const
+{
+  try {
+    ContextHandle<FileContext> file_context{Context::create<FileContext>(context, this)};
+    return *file_context;
+  } catch (HardWarningException& e) {
+    throw;
+  } catch (EvaluationException& e) {
+    return nullptr;
+  }
+}
+
 // please preferably use getFilename
 // if you compare filenames (which is the origin of this method),
 // please call getFilename first and use this method only as a fallback
--- a/src/core/SourceFile.cc
+++ b/src/core/SourceFile.cc
@@ -198,19 +198,6 @@ std::shared_ptr<AbstractNode> SourceFile::instantiate(
   return node;
 }
 
-std::shared_ptr<const FileContext> SourceFile::instantiateVariablesOnly(
-  const std::shared_ptr<const Context>& context) const
-{
-  try {
-    ContextHandle<FileContext> file_context{Context::create<FileContext>(context, this)};
-    return *file_context;
-  } catch (HardWarningException& e) {
-    throw;
-  } catch (EvaluationException& e) {
-    return nullptr;
-  }
-}
-
 // please preferably use getFilename
 // if you compare filenames (which is the origin of this method),
 // please call getFilename first and use this method only as a fallback
--- a/src/core/Expression.cc
+++ b/src/core/Expression.cc
@@ -1028,3 +1028,75 @@ void LcLet::print(std::ostream& stream, const std::string&) const
 {
   stream << "let(" << this->arguments << ") (" << *this->expr << ")";
 }
+
+ObjectExpression::ObjectExpression(AssignmentList members, const Location& loc)
+  : Expression(loc), members(std::move(members))
+{
+}
+
+Value ObjectExpression::evaluate(const std::shared_ptr<const Context>& context) const
+{
+  Value::ObjectType obj(context->session());
+  for (const auto& member : members) {
+    Value val = member->getExpr()->evaluate(context);
+    obj.set(member->getName(), std::move(val));
+  }
+  return Value(std::move(obj));
+}
+
+void ObjectExpression::print(std::ostream& stream, const std::string& indent) const
+{
+  stream << "@{ ";
+  for (const auto& member : members) {
+    stream << member->getName() << " = " << *member->getExpr() << "; ";
+  }
+  stream << "}";
+}
+
+void UnaryOp::collectDependencies(std::set<std::string>& deps) const
+{
+  this->expr->collectDependencies(deps);
+}
+
+void BinaryOp::collectDependencies(std::set<std::string>& deps) const
+{
+  this->left->collectDependencies(deps);
+  this->right->collectDependencies(deps);
+}
+
+void TernaryOp::collectDependencies(std::set<std::string>& deps) const
+{
+  this->cond->collectDependencies(deps);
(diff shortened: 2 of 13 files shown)


src/core/SourceFile.cc
```diff
@@ -198,6 +198,19 @@ std::shared_ptr<AbstractNode> SourceFile::instantiate(
...
+  try {
+    ContextHandle<FileContext> file_context{Context::create<FileContext>(context, this)};
+    return *file_context;
+  } catch (HardWarningException& e) {
+    throw;
+  } catch (EvaluationException& e) {
+    return nullptr;
+  }
+}
+
```
Shmerlard: when parsing the file we want to create the variables without calling functions like echo, if i used instantiate instead in the parsing stage then echo would have been called twice

Shmerlard:
This PR is not yet complete, but since i have changes in some core files i would like to address any issues with my approach before implementing some more, so if the general direction seems good id like to know so i can continue working on it

roel-v:
Hi, I'm very much interested in the behavior this patch adds and I hope it (or something like it) can be merged in the future. However I'm using a build to test this feature, and I ran into a case that didn't work as I expected:

dynamic_cube_special_include_controls.scad:
```

@{
    group = "Special-variable include test";
    description = "Imported cube size (mm)";
    min = 1;
    max = 100;
    step = 1;
    slider = false;
    hidden = is_undef($content_selected)
        ? true
        : $content_selected != "show_size";
}
imported_cube_size = 36;
```

dynamic_cube_special_include_minimal.scad:

```
@{
    group = "Special-variable include test";
    description = "Imported size control";
    range = [
        ["hide_size", "Hide imported size"],
        ["show_size", "Show imported size"]
    ];
}
$content_selected = "hide_size";

include <dynamic_cube_special_include_controls.scad>

cube(imported_cube_size, center = true);

```

Upon opening the second file, I expected I'd be able to hide/show the parameter from the included file by changing the parameter in the parent script. But that doesn't happen?

And a second issue, is there a way to hide/show groups completely? Like, when all children of a group are hidden, hide the group itself too?

Shmerlard:
> Hi, I'm very much interested in the behavior this patch adds and I hope it (or something like it) can be merged in the future. However I'm using a build to test this feature, and I ran into a case that didn't work as I expected:
> 
> dynamic_cube_special_include_controls.scad:
> 
> ```
> 
> @{
>     group = "Special-variable include test";
>     description = "Imported cube size (mm)";
>     min = 1;
>     max = 100;
>     step = 1;
>     slider = false;
>     hidden = is_undef($content_selected)
>         ? true
>         : $content_selected != "show_size";
> }
> imported_cube_size = 36;
> ```
> 
> dynamic_cube_special_include_minimal.scad:
> 
> ```
> @{
>     group = "Special-variable include test";
>     description = "Imported size control";
>     range = [
>         ["hide_size", "Hide imported size"],
>         ["show_size", "Show imported size"]
>     ];
> }
> $content_selected = "hide_size";
> 
> include <dynamic_cube_special_include_controls.scad>
> 
> cube(imported_cube_size, center = true);
> ```
> 
> Upon opening the second file, I expected I'd be able to hide/show the parameter from the included file by changing the parameter in the parent script. But that doesn't happen?
> 
> And a second issue, is there a way to hide/show groups completely? Like, when all children of a group are hidden, hide the group itself too?

thanks for the info, glad you liked it, there was a problem with the special variables that they were not save, it's should be fixed now,

regarding the other issue, currently there isnt but it would be very easy to implement, i just dont want to implement it yet because i really need a feedback from the maintainers if they even want something like this,
i also have many ideas for this can be a nice way to implement some missing features for the customizer like
1. tooltips instead of the current text situation
2. dynamic list like in godot or unity (although it's a really major change) 
3. edit group colors 
4. enable multiline in text parameters
5. add dashlines and seperators in groups

some are easier to implement some are harder, but all of them will be based on my approach of parameter object
but i don't want to do any of that until the base is approved
@jordanbrown0 
@t-paul

PR #6644, opened 2026-02-09
Partial application of sorting headers.

hzeller:
Due to improved includes over the time, there is less dependence on include sequence so now many header files can already be sorted.

This is fixing the include oder for many files (but not all yet). It reduces the work for the remaining files and also helps encouraging putting new headers into a sorted order.


hzeller:
mmh, I realize that this touches a lot of files at once, probably interfering with ongoing pull requests.

Let me know if I should break that up somehow.

kintel:
Ongoing PRs should be easy to clean up and/or beautify

kintel:
Btw., what's the end goal for order of system headers vs. local headers, and how/where is that enforced/analyzed?

hzeller:
Goal is to eventually switch on `SortIncludes` by default https://github.com/openscad/openscad/blob/master/.clang-format#L29

`clang-format` then will take care of maintaining at least 3 groups, each sorted in itself: c-system header, c++ system header and project header. The developer can also add separate groupings if neater for aesthetics which are then also kept sorted. 

So, `clang-format` will enforce/fix that on beautify runs.

src/gui/input/SpaceNavInputDriver.cc
```diff
@@ -30,15 +30,17 @@
...
 #include "gui/input/SpaceNavInputDriver.h"
-#include "gui/input/InputDriverEvent.h"
-#include "gui/input/InputDriverManager.h"
-#include "utils/printutils.h"
 
-#include <QThread>
 #include <spnav.h>
 #include <unistd.h>
+
+#include <QThread>
```
kintel, line 37: On X11 with `-DENABLE_SPNAV=ON` there seems to be a symbol clash between `<spnav.h>` and `<QThread>` (or any Qt header really):
```
/usr/include/x86_64-linux-gnu/qt6/QtCore/qtextstream.h:22:2: error: qtextstream.h must be included before any header file that defines Status
```

I assume this isn't uncommon, just checking if you happen to know a good strategy to deal with such in the context of IWYU and header sorting.

PR #6650, opened 2026-02-12
Notify parent when file reopened from disk

dasbidyendu:
Call parent->fileChangedOnDisk() when refreshDocument successfully reopens a file. Previously refreshDocument set file_opened but did not notify the parent, which could leave the UI or parent state out of sync after an on-disk reload.

### DIFF
--- a/src/gui/TabManager.cc
+++ b/src/gui/TabManager.cc
@@ -564,6 +564,9 @@ bool TabManager::refreshDocument()
       file_opened = true;
     }
   }
+  if (file_opened) {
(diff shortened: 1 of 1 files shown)


kintel (approved):
We should probably rethink the reload feature, but I think this will work for now.

dasbidyendu:
> We should probably rethink the reload feature, but I think this will work for now.

I think so too , have thought of some ideas. Should i share ?

kintel:
Please share. This is a bit of a tricky subject but opening a ticket with ideas is a good start

PR #6654, opened 2026-02-13
Add timer

### LINKED ISSUE #6653: Add timer builtins for script-level profiling

I've been working on a library and thought that benchmarking aspects of it might be handy. There is currently no way to measure execution time from within an OpenSCAD script. To profile which modules or functions are slow you have to resort to external tools, which can't see inside the script's evaluation and don't give per-call granularity.

A set of `timer_*` builtins for script-level profiling:

```
// Convenience wrapper — times a single function call and echoes the result
result = timer_run("my_fn", my_fn, arg1, arg2);
// Or manual control for more complex scenarios
id = timer_new("label", start=true);
// ... work ...
elapsed_us = timer_stop(id, undef);          // numeric microseconds

// These are alternative ways to stop
str       = timer_stop(id);                  // formatted string, e.g. "timer label 0:00.123"
timer_stop(id, output=true);                 // stop and echo
timer_stop(id, fmt_str="{n}: {sss}.{ddd}s"); // custom format
```

Timers are session-scoped (discarded at the end of each evaluation/render), support both monotonic wall-clock and CPU time, and can be named for readable output.  I've already implemented them and am sharing them if others think it's useful.

- **External profiling tools** — profile the whole process, not individual functions or code paths inside the script.
- **Manual echo timestamps** — OpenSCAD scripts have no access to the system clock, so this isn't possible without engine changes.  That's what these are for.

**Additional context**

The primary use case is library development: identifying which recursive functions, large list operations, or geometric constructions are the bottleneck before optimising them. `timer_run` covers the common one-shot case; the manual API covers lap timing, conditional timing, and accumulating results across multiple calls.

Ma-XX-oN:
Closes #6653

## Summary
- Add now builtin

[`now()` Design Doc](https://github.com/Ma-XX-oN/openscad/blob/add-timer/doc/spec/now_builtin.md)
[`now()` User API](https://github.com/Ma-XX-oN/openscad/blob/add-timer/doc/now_builtin.md)

### DIFF
--- a/doc/now_builtin.md
+++ b/doc/now_builtin.md
@@ -14,14 +14,16 @@ Author: _Adrian Hawryluk_ (a.k.a. [Ma-XX-oN](https://github.com/Ma-XX-oN))
 ## Signature
 
 ```text
-now() -> number
+now(unit="u") -> number
 ```
 
-Returns the current monotonic wall-clock reading in microseconds.  The
+Returns the current monotonic wall-clock reading in the requested unit.  The
 value is meaningful only when subtracted from another `now()` call within
 the same process — absolute values are not stable across runs.
 
-`now()` takes no arguments.
+The optional `unit` argument accepts `"u"` for microseconds (default),
+`"m"` for milliseconds, and `"s"` for seconds.  It may be supplied
+positionally (`now("m")`) or by name (`now(unit="m")`).
 
 ## What It Measures
 
@@ -42,6 +44,9 @@ result = expensive_computation();
 elapsed_us = now() - start;
 echo(str("elapsed: ", elapsed_us, " μs"));
 ```
+
+Use `now("m")` or `now("s")` when you want milliseconds or seconds
+without dividing manually.
 
 ## Evaluation Phases and now()
 
--- a/src/core/builtin_functions.cc
+++ b/src/core/builtin_functions.cc
@@ -1109,13 +1109,30 @@ Value builtin_import(Arguments arguments, const Location& loc)
 
 Value builtin_now(Arguments arguments, const Location& loc)
 {
-  if (!check_arguments("now", arguments, loc, 0)) {
+  Parameters parameters = Parameters::parse(std::move(arguments), loc, {}, {"unit"});
+  parameters.set_caller("now");
+  if (!parameters.valid("unit", Value::Type::STRING)) {
     return Value::undefined.clone();
   }
+
   const auto now_us = std::chrono::duration_cast<std::chrono::microseconds>(
                         std::chrono::steady_clock::now().time_since_epoch())
                         .count();
-  return Value(static_cast<double>(now_us));
+  const std::string unit = parameters.get("unit", "u");
+  if (unit == "u") {
+    return Value(static_cast<double>(now_us));
+  }
+  if (unit == "m") {
+    return Value(static_cast<double>(now_us) / 1000.0);
+  }
+  if (unit == "s") {
+    return Value(static_cast<double>(now_us) / 1000000.0);
+  }
+
+  LOG(message_group::Warning, loc, parameters.documentRoot(),
+      "now(..., unit=%1$s) unsupported unit: expected \"s\", \"m\", or \"u\"",
+      parameters.get("unit").toEchoStringNoThrow());
+  return Value::undefined.clone();
 }
 
 void register_builtin_functions()
@@ -1301,7 +1318,7 @@ void register_builtin_functions()
 
   Builtins::init("now", new BuiltinFunction(&builtin_now),
                  {
-                   "now() -> number",
+                   "now(unit=\"u\") -> number",
                  });
 
   Builtins::init("is_undef", new BuiltinFunction(&builtin_is_undef),
--- a/doc/specs/now_builtin.md
+++ b/doc/specs/now_builtin.md
@@ -4,16 +4,19 @@ Author: _Adrian Hawryluk_ (a.k.a. [Ma-XX-oN](https://github.com/Ma-XX-oN))
 
 ## Signature
 
-- `now() -> number`
+- `now(unit="u") -> number`
 
 ## Contract
 
-- Returns a numeric microsecond timestamp from a monotonic wall-clock source.
+- Returns a numeric timestamp from a monotonic wall-clock source.
 - Return type is OpenSCAD `number` (double).
 - Callers should use `now() - start` to compute elapsed time.
+- The default unit is microseconds (`"u"`).
 
 ## Argument rules
 
-- `now()` accepts zero arguments only.
-- Any positional or named argument is invalid and produces the standard
-  argument-count warning behavior.
+- `now()` accepts zero arguments or one optional `unit` argument.
+- `unit` accepts `"u"` for microseconds, `"m"` for milliseconds, and `"s"` for seconds.
+- `unit` may be supplied positionally (`now("m")`) or by name (`now(unit="m")`).
+- Unknown parameter names keep the standard parameter warning behavior.
+- Unsupported unit strings warn and return `undef`.
--- a/tests/data/scad/misc/now-tests.scad
+++ b/tests/data/scad/misc/now-tests.scad
@@ -1,6 +1,12 @@
 m0 = now();
 m1 = now();
-echo("monotonic_non_decreasing", m1 >= m0);
-
+u0 = now("u");
+u1 = now(unit="u");
+ms0 = now("m");
+s0 = now("s");
 bad_pos = now("CPU");
 bad_named = now(type="monotonic");
+bad_type = now(1);
+echo("monotonic_non_decreasing", m1 >= m0);
+echo("default_matches_u_scale", abs(m0 / u0 - 1) < 0.1 && abs(m1 / u1 - 1) < 0.1);
+echo("unit_scale_ratios", u0 / ms0 > 100 && u0 / ms0 < 10000 && ms0 / s0 > 100 && ms0 / s0 < 10000);
--- a/tests/regression/echo/now-tests-expected.echo
+++ b/tests/regression/echo/now-tests-expected.echo
@@ -1,3 +1,6 @@
-WARNING: now() number of parameters does not match: expected 0, found 1 in file now-tests.scad, line 5
-WARNING: now() number of parameters does not match: expected 0, found 1 in file now-tests.scad, line 6
+WARNING: now(..., unit="CPU") unsupported unit: expected "s", "m", or "u" in file now-tests.scad, line 7
+WARNING: variable "type" not specified as parameter in file now-tests.scad, line 8
+WARNING: now(..., unit=1) Invalid type: expected string, found number in file now-tests.scad, line 9
 ECHO: "monotonic_non_decreasing", true
+ECHO: "default_matches_u_scale", true
+ECHO: "unit_scale_ratios", true


jordanbrown0:
I'd suggest seconds rather than microseconds.  (Feel free to measure in microseconds, just divide by a million.)  In general, I'd recommend "primitive" units rather than "scaled" units.  (I do not understand why [SI](https://en.wikipedia.org/wiki/International_System_of_Units) considers [kg](https://en.wikipedia.org/wiki/Kilogram) to be the base unit for mass.)

I didn't look at the FunctionArgs stuff in any detail.  It looks like it might be interesting, but does not seem relevant to this PR.

doc/now_builtin.md
```diff
@@ -0,0 +1,170 @@
...
+value is meaningful only when subtracted from another `now()` call within
+the same process — absolute values are not stable across runs.
+
+`now()` takes no arguments.
+
+## What It Measures
+
+`now()` reads a monotonic wall-clock.  It is useful for benchmarking the
+cost of OpenSCAD expressions, function calls, and — when placed in a
+module argument expression — elapsed time across module instantiations
```
jordanbrown0, line 32: Nit:  there are a couple of different patterns for measuring module instantiation; an argument expression is only one.

doc/now_builtin.md
```diff
@@ -0,0 +1,170 @@
...
+elapsed_us = now() - start;
+echo(str("elapsed: ", elapsed_us, " μs"));
+```
+
+## Evaluation Phases and now()
+
+Understanding OpenSCAD's two-phase evaluation model is essential for using
+`now()` correctly.
+
+Inside any scope (file-level or module body), OpenSCAD processes statements
```
jordanbrown0: Scope is the top level, or anything that's a child of anything else.

Mostly, it's the top and anything inside braces.  The trivial child case of a single module invocation is a scope, but since it has no assignments it's not a very interesting one for these purposes.

Anonymous blocks (no module invocation, just braces) are not scopes.

```
scope 1
union() {
    scope 2
    if (whatever) {
        scope 3
    }
}
module foo() {
    scope 4
}
{
    not a scope
}
```

doc/now_builtin.md
```diff
@@ -0,0 +1,170 @@
...
+
+  ```openscad
+  start = now();                       // assignment — Phase 1
+  for (i = [0:999]) { x = sum(lst); } // module inst — Phase 2
+  elapsed = now() - start;             // assignment — Phase 1 (!)
+  ```
+
+  You _can_ place the second `now()` inside an `echo` (also Phase 2),
+  which runs after the loop, but the measurement includes Phase 1→Phase 2
+  overhead and any earlier module instantiations.  For precise
```
jordanbrown0: Good point about earlier module instantiations.  Also *later* assignments.  (I don't think the actual transition is significant.)

You can mitigate both of those by creating a scope for the purpose...

```
...
union() {    // isolate stuff to be benchmarked
    start = now();
    some_module();
    echo(now() - start);
}
...
```

Note that although unions can have significant rendering costs, they have no more **evaluation** cost than any other block.  Also, unioning a single child is basically free, and most contexts union their children anyway so you were going to pay any union cost eventually.

doc/now_builtin.md
```diff
@@ -0,0 +1,170 @@
...
+  expression-only timing, use a recursive function instead (see Common
+  Patterns).
+
+- **Expecting render times from assignments** — Module instantiation
+  (Phase 2) builds the CSG tree but does not perform the final render
+  (mesh generation, boolean evaluation).  A `now()` call in an
+  assignment (Phase 1) cannot even bracket instantiation time.  To
+  measure elapsed time across instantiations, place `now()` in a module
+  argument expression so it executes in Phase 2 (see
+  [Evaluation Phases and now()](#evaluation-phases-and-now)).
```
jordanbrown0, line 240: Might add a note that for rendering times, consult the console log.
We should probably explicitly time each phase and report both the individual and aggregate times.  But that's not this project.

src/core/builtin_functions.cc
```diff
@@ -1105,6 +1107,17 @@ Value builtin_import(Arguments arguments, const Location& loc)
...
 
+Value builtin_now(Arguments arguments, const Location& loc)
+{
+  if (!check_arguments("now", arguments, loc, 0)) {
+    return Value::undefined.clone();
+  }
+  const auto now_us = std::chrono::duration_cast<std::chrono::microseconds>(
+                        std::chrono::steady_clock::now().time_since_epoch())
+                        .count();
+  return Value(static_cast<double>(now_us));
```
jordanbrown0: Recommend returning the value in seconds.  Just divide the double by a million.

src/core/builtin_functions.cc
```diff
@@ -1286,6 +1299,11 @@ void register_builtin_functions()
                    "parent_module(number) -> string",
                  });
 
+  Builtins::init("now", new BuiltinFunction(&builtin_now),
+                 {
+                   "now() -> number",
```
jordanbrown0: We only have a couple of examples of specifying units, but asin, acos, atan, and atan2 explicitly say "degrees" rather than "number".  I would say "seconds" here.

doc/now_builtin.md
```diff
@@ -0,0 +1,170 @@
...
+## Quick Start
+
+```openscad
+start = now();
+result = expensive_computation();
+elapsed_us = now() - start;
+echo(str("elapsed: ", elapsed_us, " μs"));
+```
+
+## Evaluation Phases and now()
```
jordanbrown0, line 51: Probably needs a warning that evaluation order is subject to change without notice, and so this mechanism is useful as a diagnostic tool but may be unreliable across releases.

doc/specs/function_args.md
```diff
@@ -0,0 +1,198 @@
...
+- [ParamDef](#paramdef)
+- [Spec Definition](#spec-definition)
+- [Normalization APIs](#normalization-apis)
+- [Structural Rules](#structural-rules)
+- [Variadic Rules](#variadic-rules)
+- [Error Routing](#error-routing)
+- [Design Constraints](#design-constraints)
+
+This document defines `FunctionArgs::Spec`, the shared helper used to normalize
+function/builtin arguments into canonical form.
```
jordanbrown0, line 16: I know I said I didn't review this, but I did peek a little :-)

Note that functions and modules have very similar needs, and any new scheme should support both.

Note also [XKCD 927](https://xkcd.com/927/).  A new scheme should completely replace all existing schemes, so that we don't just end up with N+1 schemes.  (Except for the very few cases that need totally-custom handling.)

doc/specs/function_args.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Purpose
+
+`FunctionArgs::Spec` centralizes argument-shape handling so each builtin can:
+
+- declare its parameter shape once
+- accept mixed positional/named syntax consistently
+- receive canonical fixed values (and optional variadic values)
+- keep consistent structural diagnostics
+- disallows positional arguments after the first named argument.  This avoids
```
jordanbrown0, line 26: For better or worse, that's incompatible.  The behavior of mixed positional and named arguments is obscure but is well-defined.  *Maybe* we could make it a warning, but any replacement scheme should continue to support arbitrary mixing, as the current processing does.

src/core/FunctionArgs.cc
```diff
@@ -0,0 +1,189 @@
...
+      fail(STR(function_name_, "() internal error: fixed parameter index out of bounds"));
+    }
+    if (slot_sources[i]) {
+      continue;
+    }
+    if (params_[i].defaultValue()) {
+      result.params[i] = params_[i].defaultValue().get();
+    } else {
+      result.params[i] = &Value::undefined;
+    }
```
jordanbrown0, line 160: Caution:  existing callers will sometimes test for:
* was this parameter specified?  (Note that `param=undef` **does** specify the parameter.)
* was this parameter defined? (Excludes `param=undef`.)
* is this parameter string/number/bool/whatever?
* can this parameter be converted to string/number/boolean/whatever?

and those semantics all need to be preserved, or any change carefully evaluated.

src/core/FunctionArgs.cc
```diff
@@ -0,0 +1,189 @@
+#include "core/FunctionArgs.h"
```
jordanbrown0, line 1: Caution: Since this would go into almost every builtin function/module call, careful benchmarking is needed to ensure that it doesn't have a significant performance impact.

doc/now_builtin.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Common Patterns
+
+**Average cost per call over N iterations:**
+
+A bare `for` loop cannot interleave with `now()` calls in assignments at
+the same scope (see Common Mistakes).  Use a recursive helper instead:
+
+```openscad
+function _loop(fn, i, n) = i >= n ? undef : let(_ = fn(i)) _loop(fn, i+1, n);
```
jordanbrown0, line 157: Should benchmark to see what the lowest-overhead pattern is for this.  I suspect that a recursive function is *not* the best.

doc/now_builtin.md
```diff
@@ -0,0 +1,170 @@
...
+
+```openscad
+start = now();
+some_module();
+echo(str("elapsed to this point: ", now() - start, " μs"));
+```
+
+Note: this measures wall-clock time from the end of the assignment phase
+through `some_module()`'s instantiation — it includes any Phase 1→Phase 2
+transition overhead.
```
jordanbrown0: and any assignments after the "start=" and any module invocations before the "echo".

doc/now_builtin.md
jordanbrown0: Probably generalized to "this is a subtle tool; make sure that you fully understand its behavior and limitations".

Ma-XX-oN:
> I'd suggest seconds rather than microseconds. (Feel free to measure in microseconds, just divide by a million.) In general, I'd recommend "primitive" units rather than "scaled" units. (I do not understand why [SI](https://en.wikipedia.org/wiki/International_System_of_Units) considers [kg](https://en.wikipedia.org/wiki/Kilogram) to be the base unit for mass.)

The reason I left it as μs is because it is exact.  If I move to seconds, then it stops being exact.  It the same reason we use whole numbers for counting.

> I didn't look at the FunctionArgs stuff in any detail. It looks like it might be interesting, but does not seem relevant to this PR.

Yeah, it was relevant but not any more.  It's a standardised way to process parameters.  Also allows for variadic parameters.  I'll remove it.

doc/now_builtin.md
Ma-XX-oN: Maybe later

doc/specs/function_args.md
Ma-XX-oN: I agree, but there is no "standard" that I could decern except for chaos. 😉😄

doc/specs/function_args.md
Ma-XX-oN: Sure.  I'd settle for a warning.  But the current mechanism is too obscure.  I'd personally mark it as deprectated.

src/core/builtin_functions.cc
Ma-XX-oN: Maybe.  See [this comment](https://github.com/openscad/openscad/pull/6654/changes/BASE..8c5e5569a291d20f9c57a8a9d07b8f1b907440bf#r2845172354) first.

Ma-XX-oN:
>  I'd suggest seconds rather than microseconds. (Feel free to measure in microseconds, just divide by a million.) In general, I'd recommend "primitive" units rather than "scaled" units. (I do not understand why [SI](https://en.wikipedia.org/wiki/International_System_of_Units) considers [kg](https://en.wikipedia.org/wiki/Kilogram) to be the base unit for mass.)

The reason I left it as μs is because of the typical use case would be micro to millisecond range.  Those who want to deal with seconds can just multiply by a million.

> I didn't look at the FunctionArgs stuff in any detail. It looks like it might be interesting, but does not seem relevant to this PR.

Yeah, it was relevant but not any more. It's a standardised way to process parameters. Also allows for variadic parameters. I'll remove it.

src/core/builtin_functions.cc
Ma-XX-oN: Maybe.  See [this comment](https://github.com/openscad/openscad/pull/6654/changes/BASE..8c5e5569a291d20f9c57a8a9d07b8f1b907440bf#r2845216217) first.

jordanbrown0:
Hint:  when you go back and edit a comment, there's no e-mail notification, so somebody who is following along in e-mail won't see the update.  What I do is to make the change, but also add a new comment that describes the change, and then after the e-mail notification goes out I delete the new comment.  That leaves the history clean, but also keeps e-mail readers up-to-date.

You said, and edited away:
> The reason I left it as μs is because it is exact. If I move to seconds, then it stops being exact. It the same reason we use whole numbers for counting.

64-bit floating point inaccuracy is about one part in 10^16 - plus or minus a couple of nanoseconds, when you're measuring a year.  I wouldn't worry about it.  (Which might be why you edited it away.)

> The reason I left it as μs is because of the typical use case would be micro to millisecond range. Those who want to deal with seconds can just multiply by a million.

I think you're being optimistic to think that you can get repeatable measurements down in the microsecond or even millisecond range.  On my system, the time to process `cube()` varies plus or minus a millisecond between successive runs.  (That is, it's 3 to 5 milliseconds.)  The time to load BOSL2 and process its cuboid() varies by a hundredth of a second from run to run.  (It's about 160ms.)  I don't know where the noise is coming from, so I'm not *sure* that it affects the stuff you're likely to want to measure, but I think for any real benchmarking you're going to want to have run times up in the seconds to try to get away from it.

Ma-XX-oN:
> Hint: when you go back and edit a comment, there's no e-mail notification, so somebody who is following along in e-mail won't see the update. What I do is to make the change, but also add a new comment that describes the change, and then after the e-mail notification goes out I delete the new comment. That leaves the history clean, but also keeps e-mail readers up-to-date.
> 
> You said, and edited away:
> 
> > The reason I left it as μs is because it is exact. If I move to seconds, then it stops being exact. It the same reason we use whole numbers for counting.
> 
> 64-bit floating point inaccuracy is about one part in 10^16 - plus or minus a couple of nanoseconds, when you're measuring a year. I wouldn't worry about it. (Which might be why you edited it away.)
> 

Yes, that's why I edited it.  I had reevaluated my thoughts because realised the answer didn't make sense.  I was tired and wasn't thinking straight.

> > The reason I left it as μs is because of the typical use case would be micro to millisecond range. Those who want to deal with seconds can just multiply by a million.
> 
> I think you're being optimistic to think that you can get repeatable measurements down in the microsecond or even millisecond range. On my system, the time to process `cube()` varies plus or minus a millisecond between successive runs. (That is, it's 3 to 5 milliseconds.) The time to load BOSL2 and process its cuboid() varies by a hundredth of a second from run to run. (It's about 160ms.) I don't know where the noise is coming from, so I'm not _sure_ that it affects the stuff you're likely to want to measure, but I think for any real benchmarking you're going to want to have run times up in the seconds to try to get away from it.

You're right that run-to-run variance is significant — OS scheduling, cache state, etc. That's why the documented pattern is to run many iterations within a single evaluation and average. Within a single run, the measurements are much more stable. Microsecond resolution is useful for seeing per-call differences in those aggregated measurements.

jordanbrown0:
@rcolyer likes seconds too.

I know that I suggested it, but I'm starting to think that `now()` is not a good name.  If we ever support wall-clock time (e.g. for timestamping something), `now()` might imply wall-clock time.

@rcolyer suggests `time_elapsed()`.  I don't think the `time_` adds value, and like `elapsed()` better.

MichaelAtOz:
Seconds? If it has milli/microseconds as the decimal places does it matter.
If it is:
<0.1 e.g. `echo( (PI-3)/10 )` -> `ECHO: 0.0141593` (rounded in the last digit). That is milliseconds (ignoring the rounded digit)
<1 e.g. `echo(PI-3);` -> `ECHO: 0.141593` (rounded in the last digit). that is 141592 or 141593 microseconds.
(what's a millisecond between friends...)
\>=1 10's of microseconds, probably sufficient resolution.

jordanbrown0:
echo is always six digits of precision, no matter what the magnitude.

If you have 1,234,567 microseconds, and you print it as microseconds, you get 1.23457e+6.  If you print it as seconds, you get 1.23457.

jordanbrown0:
And while six digits may seem like a small number, if you're measuring something that runs for a second do you really expect to be able to get microsecond repeatability?

jordanbrown0:
Remember that that is only echo's precision.  The value is actually good to about 16 digits, so if you have a timer that's been running for a year (and so has a value around 31e6 seconds), and you grab two timestamps and subtract them, you have an accuracy just a bit worse than nanoseconds.

That is, again, independent of the units used for the timer.

jordanbrown0:
I was going to say that that was an argument for keeping T=0 relatively recent, to avoid wasting high-order digits, but practically it's not an issue.  Time since 1970 in microseconds is about a 16-digit number, so you would still have microsecond accuracy.

Ma-XX-oN:
> I know that I suggested it, but I'm starting to think that `now()` is not a good name. If we ever support wall-clock time (e.g. for timestamping something), `now()` might imply wall-clock time.
> 
> @rcolyer suggests `time_elapsed()`. I don't think the `time_` adds value, and like `elapsed()` better.

`elapsed()` is a duration; `now()` returns a monotonic clock reading. You call `now()` twice and subtract to get what's elapsed. Renaming it to `elapsed()` conflates the tool with its output and would be like renaming a `vector` as a "distance" — it mixes the input with a derived quantity.

> I think you're being optimistic to think that you can get repeatable measurements down in the microsecond or even millisecond range. On my system, the time to process `cube()` varies plus or minus a millisecond between successive runs. (That is, it's 3 to 5 milliseconds.) The time to load BOSL2 and process its cuboid() varies by a hundredth of a second from run to run. (It's about 160ms.) I don't know where the noise is coming from, so I'm not _sure_ that it affects the stuff you're likely to want to measure, but I think for any real benchmarking you're going to want to have run times up in the seconds to try to get away from it.

Benchmarking is done by running multiple executions for the same code fragment and then dividing by the number of executions.  So timing is still typically going to be in the μs-ms range.  Also, since the results are in the μs–ms range, seconds would just produce awkward small decimals like `0.001234` where `1234` is far more readable.  For long-duration reporting, users can convert (sec = μs / 1e6); for fine-grained profiling, direct μs is more readable.

The noise you're describing is mainly run-to-run variance. The loop-and-divide pattern runs within a single evaluation, which reduces those sources significantly.  This averages out the remaining within-run jitter across many iterations, which is exactly where the interesting differences are when profiling small code fragments.

jordanbrown0:
`elapsed()` could measure the time since the start of the run, which is a duration *and* a point in time.

If we continue with `now()`. we should probably pencil in what word we would use for the current wall-clock date/time.


As for units, I think that sticking with primitive units is simpler - that, for instance, converting from seconds to milliseconds or microseconds is mentally simpler than converting from microseconds to milliseconds.  But in the end this is very much a niche power tool, and the difference will be small compared to the work involved in getting useful results out of it at all, so... shrug.

Ma-XX-oN:
@jordanbrown0 ,

> If we continue with `now()`. we should probably pencil in what word we would use for the current wall-clock date/time.

How about `time()`?

> As for units, I think that sticking with primitive units is simpler

μs is the primitive unit.  It's what comes out of the base function.

> But in the end this is very much a niche power tool, and the difference will be small compared to the work involved in getting useful results out of it at all, so... shrug.

Agreed.  I could put a unit parameter.  "s" for seconds, "m" for ms, "u" for μs.  Would that be acceptable?

Ma-XX-oN:
> Agreed. I could put a unit parameter. "s" for seconds, "m" for ms, "u" for μs. Would that be acceptable?

Done @jordanbrown0

PR #6656, opened 2026-02-14
Add "em" parameter to text functions to set the size of an em-square.

### LINKED ISSUE #4304: text scaled wrong (resolution = 100 should be 72)

jordanbrown0:
Fixes #4304.

This PR takes a simpler approach than #4306 did.  #4306 would record how the size was specified, and would conditionally fix or not-fix the scaling error at the lower level.  Here we accept the same `em` parameter, but simply use it to set the `size` parameter, scaling it to compensate for the error at the lower level... and leaving lots of comments explaining the oddity.

This does not include the "underline" metric discussed in #4306; this topic is already complex enough without adding unrelated changes.  The "underline" metric can be covered in a separate PR.

### DIFF
--- a/src/core/FreetypeRenderer.cc
+++ b/src/core/FreetypeRenderer.cc
@@ -288,6 +288,17 @@ const FontFacePtr FreetypeRenderer::Params::get_font_face() const
     return nullptr;
   }
 
+  // 100 is the wrong value for resolution.  It should be 72.  100 yields a 100:72 scaling factor
+  // because the input size is in points and the output coordinates are in 1/resolution inches.
+  // OpenSCAD doesn't care about inches and points, but this means that the glyphs generated
+  // are about 1.39x as large as they should be, relative to the font size specified.
+  // This is described in issue #4304.
+  //
+  // We don't just change it to 72, because this 100:72 scaling factor has been present
+  // for many years and any model using text() depends on it.  We could fix it here and do
+  // something tricky above, but that has subtle compatibility effects.  Instead, we
+  // introduced a new "em" parameter that lets the user specify the em size of the font,
+  // and we set the size based on that to cancel out the error that we introduce here.
   FT_Error error = FT_Set_Char_Size(face->face_, 0, scale, 100, 100);
   if (error) {
     LOG(message_group::Warning, loc, documentPath, "Can't set font size for font %1$s", font);
@@ -303,6 +314,7 @@ FreetypeRenderer::Params::Params(Parameters& parameters)
   // the expected type.
 
   (void)parameters.valid("size", Value::Type::NUMBER);
+  (void)parameters.valid("em", Value::Type::NUMBER);
   (void)parameters.valid("text", Value::Type::STRING);
   (void)parameters.valid("spacing", Value::Type::NUMBER);
   (void)parameters.valid("font", Value::Type::STRING);
@@ -312,7 +324,28 @@ FreetypeRenderer::Params::Params(Parameters& parameters)
   (void)parameters.valid("halign", Value::Type::STRING);
   (void)parameters.valid("valign", Value::Type::STRING);
 
-  size = parameters.get("size", 10.0);
+  // There is a bug (#4304):  FreetypeRenderer::Params::get_font_face()
+  // will inappropriately apply a scaling factor of 100/72, so
+  // that a specified size of 72 would yield an em-size of 100.
+  // This went undetected for years because the error introduced
+  // was approximately the same as the typical ratio between the
+  // em size and the height of capital letters, and so "size"
+  // was explained as the height of capital letters.
+  // We can't possibly redefine "size" correctly, because there are years of inertia
+  // behind the bug.  Instead, we define a new parameter "em" that sets the em size correctly.
+  // We do not fix the bug down where it actually occurs, because that has subtle
+  // compatibility effects.  Instead, we just correct for it here.
+  const Value& emParam = parameters.get("em");
+  if (emParam.isDefined()) {
+    if (parameters.get("size").isDefined()) {
+      LOG(message_group::Warning, loc, documentPath, "%1$s: \"size\" ignored when \"em\" is set",
+          parameters.get_caller());
+    }
+    size = emParam.toDouble() * 72.0 / 100.0;
+  } else {
+    size = parameters.get("size", 10.0);
+  }
+
   set_text(parameters.get("text", ""));
   spacing = parameters.get("spacing", 1.0);
   set_font(parameters.get("font", ""));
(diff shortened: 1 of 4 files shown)


jordanbrown0:
If you look at #4304 you will find a very long discussion of the problem, that wandered off into irrelevant topics like units.

Here is a very simple summary.  In basically any application (other than OpenSCAD) that allows you to draw text at a specified size, and allows you to draw other things at specified sizes, and you ask it to create a Times New Roman X using an N-unit font and put it at the bottom left of an N-unit square, you will get something that looks a lot like this;

image

It does not matter what the units are.  It does not matter whether that's a 72-point font in a one-inch square, or a ten-mile font in a ten-mile square.  You'll get the same ratio between the size of the X and the size of the square.

Note that the X is considerably smaller than the square.  The reason is buried deep in how fonts are traditionally measured, and really isn't relevant.  What's relevant is that it's the standard:  if you ask for an N-unit X and an N-unit square, that's the ratio that you get.  This particular image is from Inkscape, but you will get basically the same result from any browser, from MS Word, from CorelDraw, et cetera.  (The exact ratio will depend on the font used.)

On the other hand, in OpenSCAD if you ask for an N-unit X and an N-unit square, you get:

image

Note that the X is very much larger, almost as large as the square.  (In fact, depending on the font it may be even closer to matching the square, which is how #4304 went undiscovered for so long.)

With this change, and specifying the font size using the new `em` parameter, you get this:
image

When I change the Inkscape image to have contrasting colors, scale the two squares to the same size, and superimpose the Inkscape image over the new OpenSCAD image I get a near-perfect match:

image

If you like the behavior of `size` - which is by accident really a pretty useful behavior - great, use it, it's unchanged.  But if you want the font size to mean the same thing that it does in other programs, relative to the size of other objects, `em` will do that.

ali1234:
My interpretation of the problem:

image

This is what you get when you ask for a font of size **n** units in other software vs OpenSCAD.

In other software the resulting text is approximately **n** units from the top of caps to the bottom of the descender. In fact it is the full [body](https://en.wikipedia.org/wiki/Body_height_(typography)) which is always exactly **n** units tall, and it may be larger than ascenders and descenders depending on the font. This is a typographical standard. Font size = height of **body** = 1em.

In OpenSCAD the resulting text is approximately **n** units from the [cap height](https://en.wikipedia.org/wiki/Cap_height) to the [baseline](https://en.wikipedia.org/wiki/Baseline_(typography)). However this is completely accidental. Nothing in particular is exactly **n** units tall, but the **body** ends up being exactly **n***100/72 units tall due to the way OpenSCAD (mis)uses the underlying font libraries.

The purpose of the new em parameter is purely to match the behavior of other software and the typographical standard they are all based on.

jordanbrown0:
@ali1234 Yes, that's a good summary.

jordanbrown0:
WRT the parameter name, note that on the design side, Font Forge calls this metric "Em Size":

image

jordanbrown0:
I think the problem to be solved is now well-understood.  The answer of offering a new parameter to control the font size in an industry-compatible way seems like the only practical answer.  The implementation here is simple, and regardless is behind-the-scenes so could be changed later if need be.

In my mind, the only potentially-open question is the name of the parameter.  There seem to be two common names for this measurement:  em, and body size or height. Those suggest that the words "em" or "body" would appear in the name.  Those words might, or might not, be followed by "size" or "height" to clarify what they mean.

That yields six possible names:  em, emsize, emheight, body, bodysize, bodyheight.

"em" versus "body" is an interesting decision.  There is an argument that "body" is less jargon-y.  However, that very fact is also an argument against it, because people who are not familiar with typography are likely to think that "body" means something other than what it means - they are likely to think that it means the cap-height, or the cap-height plus descenders, or, well, *anything* that's directly related to the shape of the font.  Somebody who is not familiar with typography seems likely to see "em" and say "I have no idea what that means", and seems more likely to actually read the documentation than they would if presented with "body".

For "em", I don't think that "size" or "height" adds much value.  An em is defined to be a height.

And "em" is shorter than any of the others.

So, net, I prefer "em" but would accept any of the other five.  My second choice would be "emsize", connecting it to "size".

kintel:
I agree with emsize or em.

jordanbrown0:
Last push resynced and fixed a typo in a comment.  Perhaps somebody could review and mark approved?

kintel:
@t-paul you had concerns re. naming, could you have a look again?

kintel:
Let's rather discuss any changes post-merge :)

PR #6658, opened 2026-02-14
Add underline position and thickness as font metrics.

jordanbrown0:
Adds an underline object with position and thickness to the return from fontmetrics(), e.g.:

```
echo(fontmetrics());
```
```
ECHO: {
    ...
    underline = { position = -2.48895; thickness = 1.0173; };
}
```

Sample use case:
```
size = 10;
text="Hello";
tm = textmetrics(text, size=size);
fm = fontmetrics(size=size);

text(text, size=size);
translate([tm.position.x, fm.underline.position])
    square([tm.size.x, fm.underline.thickness]);
```

image

Note that the new metric controls only the vertical position and thickness of the underline; the X position and width is derived from the metrics of the text to be underlined.

Application note:  if you want spacing around descenders, offset the text a little and subtract it from the underline.

### DIFF
--- a/src/core/FreetypeRenderer.cc
+++ b/src/core/FreetypeRenderer.cc
@@ -513,6 +513,19 @@ FreetypeRenderer::FontMetrics::FontMetrics(const FreetypeRenderer::Params& param
   family_name = face->face_->family_name;
   style_name = face->face_->style_name;
 
+  underline_thickness =
+    FT_MulFix(face->face_->underline_thickness, size_metrics->y_scale) / scale * params.size;
+  // FreeType defines underline position as the vertical center of the underline.
+  // That's inconvenient for our user, who probably wants to call square()
+  // to generate the underline and would need to translate it down to
+  // center it.  Rather than making our user do that, *we* define the
+  // underline position as being the position of the bottom of the underline,
+  // so that translate([0,position,0]) square(length, thickness) will do
+  // the right thing.
+  underline_position =
+    FT_MulFix(face->face_->underline_position, size_metrics->y_scale) / scale * params.size -
+    underline_thickness / 2;
+
   ok = true;
 }
 
--- a/src/core/FreetypeRenderer.h
+++ b/src/core/FreetypeRenderer.h
@@ -132,6 +132,8 @@ public:
     double max_ascent;
     double max_descent;
     double interline;
+    double underline_position;
+    double underline_thickness;
     std::string family_name;
     std::string style_name;
     FontMetrics(const FreetypeRenderer::Params& params);
--- a/src/core/builtin_functions.cc
+++ b/src/core/builtin_functions.cc
@@ -1019,20 +1019,20 @@ Value builtin_fontmetrics(Arguments arguments, const Location& loc)
   max.set("ascent", metrics.max_ascent);
   max.set("descent", metrics.max_descent);
 
-  ObjectType font(session);
-  font.set("family", metrics.family_name);
-  font.set("style", metrics.style_name);
-
   ObjectType underline(arguments.session());
   underline.set("position", metrics.underline_position);
   underline.set("thickness", metrics.underline_thickness);
+
+  ObjectType font(session);
+  font.set("family", metrics.family_name);
+  font.set("style", metrics.style_name);
 
   ObjectType font_metrics(session);
   font_metrics.set("nominal", nominal);
   font_metrics.set("max", max);
   font_metrics.set("interline", metrics.interline);
-  font_metrics.set("font", font);
   font_metrics.set("underline", underline);
+  font_metrics.set("font", font);
 
   return std::move(font_metrics);
 }
--- a/src/core/builtin_functions.cc
+++ b/src/core/builtin_functions.cc
@@ -1023,11 +1023,16 @@ Value builtin_fontmetrics(Arguments arguments, const Location& loc)
   font.set("family", metrics.family_name);
   font.set("style", metrics.style_name);
 
+  ObjectType underline(arguments.session());
+  underline.set("position", metrics.underline_position);
+  underline.set("thickness", metrics.underline_thickness);
+
   ObjectType font_metrics(session);
   font_metrics.set("nominal", nominal);
   font_metrics.set("max", max);
   font_metrics.set("interline", metrics.interline);
   font_metrics.set("font", font);
+  font_metrics.set("underline", underline);
 
   return std::move(font_metrics);
 }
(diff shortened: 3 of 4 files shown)


jordanbrown0:
The Ubuntu failure appears to be based on the metrics for the font having changed.  The version distributed with OpenSCAD is 2.00.1; here is a sample of the underline positioning:

image

Here is a sample of the underline positioning with the current 2.1.5, making no changes other than to swap the font files into place:

image

Except... that the font metrics in 2.00.1 appear to be identical to the metrics in 2.1.5, according to the history at
https://github.com/liberationfonts/liberation-fonts

HOWEVER, note https://github.com/liberationfonts/liberation-fonts/issues/27, which is suspiciously similar.

At least viewed through our lens (which mostly means FreeType), it looks like 2.1.5 is wrong.  The intent is that the metrics match Arial.  2.00.1 does; 2.1.5 doesn't.

<table>
<tr><th>Font</th><th colspan=2>nominal</th><th colspan=2>max</th><th>interline</th><th colspan=2>underline</th></tr>
<tr><th></th><th>ascent</th><th>descent</th><th>ascent</th><th>descent</th><th></th><th>position</th><th>thickness</th></tr>

<tr><td>Arial</td><td>12.5733</td><td>-2.9433</td><td>14.4382</td><td>-4.5098</td><td>15.9709</td><td>-2.48895</td><td>1.0173</td></tr>

<tr><td>LibSans 2.00.1</td><td>12.5733</td><td>-2.9433</td><td>13.6109</td><td>-4.2114</td><td>15.9709</td><td>-2.48895</td><td>1.0173</td></tr>

<tr><td>LibSans 2.1.5</td><td>12.5733</td><td>-2.9433</td><td>13.6109</td><td>-4.2114</td><td>15.9709</td><td>-1.47165</td><td>1.0173</td></tr>

</table>

jordanbrown0:
The mismatch between 2.00.1 and 2.1.5 matches the behavior predicted by
https://fontforge.org/en-US/bulletins/underline_position_2018_1/

jordanbrown0:
I'm not sure what to do with this PR.  All evidence is that the various installations are inconsistent, and that the tests are correct.  We can't have failing tests, though.

I suspect that the real answer is that we should ensure that OpenSCAD prefers the versions of the Liberation fonts that are delivered with it, over system-supplied versions, so that it gets consistent behavior on all installations.

jordanbrown0:
Previous push ("Tweak bug remplate") was an error; I made the change to the wrong copy and then pushed.  Undone.  (And, interestingly, the commit seems to have disappeared from the "conversation" log.)

kintel:
> 
> I suspect that the real answer is that we should ensure that OpenSCAD prefers the versions of the Liberation fonts that are delivered with it, over system-supplied versions, so that it gets consistent behavior on all installations.

This sounds like the correct way of doing it anyway, but not sure exactly how to pull that off..

jordanbrown0:
I **think** that it's a matter of handing files and directories to fontconfig in the right order.  Might need a fontconfig file too.

jordanbrown0:
#6675 confirms that the Ubuntu 24 environment used for testing has Liberation Sans Regular v2.1.5 when it fails its test:

```
-ECHO: { ... version = "Version 2.00.1"; }; }
+ECHO: { ... version = "Version 2.1.5"; }; }
```

jordanbrown0:
Looked a bit into Fontconfig.  It seems that there's no documented way to say "create a configuration, add my stuff to it, then add all of the standard stuff to it".  Using undocumented mechanisms, it looks like you could call `FcConfigCreate()` to create an empty configuration, then add your directory to that, then call `FcInitLoadOwnConfig()` on that configuration.

jordanbrown0:
Sent a message to the fontconfig mailing list.  We'll see what they say.

jordanbrown0:
I should note that even the undocumented sequence I described above might not do the trick; that might let you load config files in a desired order, but I haven't found anything that describes the search order used for font files.

Otzie2023:
I got this wrong, and the wrong part was load-bearing. Correcting it.

My claim that FreeType passes `post.underlinePosition` through untouched is
false. It subtracts half the thickness, converting the OpenType "top of the
stroke" convention into the "centre of the stem" that it documents. Measured
against both files with a small FreeType program:

| | `post.underlinePosition` | FreeType `underline_position` |
|---|---|---|
| 2.00.1 | -217 | -292 |
| 2.1.5 | -67 | -142 |

Which is exactly what you already reported in liberationfonts/liberation-fonts#27,
and I should have read that thread before posting.

So the `- underline_thickness / 2` in this PR is correct: FreeType has already
done the first half, this does the second, and the result is the bottom of the
stroke. There is nothing to fix in the formula.

That also voids my "same range in both versions" claim. It rested on reading
2.00.1 as encoding the bottom, and it doesn't — under one consistent
spec-conforming reading the two fonts genuinely place the underline 150 units
apart. And your Arial measurement settles which of the two is right: Arial and
2.00.1 both give -292, 2.1.5 gives -142. Metric compatibility with Arial is the
whole point of Liberation, so by Liberation's own standard 2.1.5 is the
regression. Your "I would not upgrade to it" was correct and my reasoning
against it was not.

What does still stand from the measurement is the glyph coverage. 2.1.5 adds 33
characters to Liberation Sans and removes none: the full set of subscript digits
U+2080-U+2089, three missing superscripts, U+2010 HYPHEN, U+2262, U+21D4, U+20BF.
Ascender, descender and unitsPerEm are unchanged across all twelve faces. So the
bundled font cannot render `text("H₂O")` today, and the choice is a trade-off
rather than a clear call - Arial-compatible underline against 33 glyphs.

One thought on where that trade-off could be resolved. The shift is exactly one
thickness in all twelve faces, which is the signature the FontForge bulletin
describes, so this looks like an unintended consequence of a rebuild rather than
a design decision upstream. liberationfonts#27 has been open since 2020 and does
not yet have the per-face numbers in it. Posting them there costs nothing and
puts the fix where it belongs, instead of OpenSCAD choosing between two
imperfect versions indefinitely. Happy to write that up if it's useful.

jordanbrown0:
> if OpenSCAD reliably gets its own copy of the font, it knows which convention
applies.

OpenSCAD supports fonts other than the ones it bundles, so it needs to be compatible with the spec.  It should not work around a broken font, even if it's a font that we bundle.

Otzie2023:
Adding the per-face numbers, since this thread has the diagnosis but not the
data. Comparing 2.00.1 with 2.1.5, `post.underlinePosition` moves by exactly
`underlineThickness` in every one of the twelve faces, while the thickness itself
is unchanged:

| face | 2.00.1 pos | 2.1.5 pos | thickness |
|---|---|---|---|
| Sans Regular, Italic | -217 | -67 | 150 |
| Sans Bold, BoldItalic | -217 | -2 | 215 |
| Serif Regular, Italic | -223 | -123 | 100 |
| Serif Bold, BoldItalic | -223 | -28 | 195 |
| Mono Regular, Italic | -477 | -393 | 84 |
| Mono Bold, BoldItalic | -477 | -272 | 205 |

Arial reports -217/150 for Regular, matching 2.00.1. Since metric compatibility
with the reference fonts is the stated goal, 2.1.5 no longer meets it here.

The uniform one-thickness offset matches the behaviour described in
https://fontforge.org/en-US/bulletins/underline_position_2018_1/, which suggests
this came in with a rebuild rather than as a deliberate change. unitsPerEm,
ascender and descender are unchanged, so nothing else appears to have moved.

Issue #6615, opened 2026-02-06
[FEATURE] Add "Close all" and "Close all, but this" options


SlimRG:
**Is your feature request related to a problem? Please describe.**

<img width="295" height="181" alt="Image" src="https://github.com/user-attachments/assets/28cbfd1e-404c-4161-b1f8-6f02daf6ea10" />

**Describe the solution you'd like**

<img width="233" height="76" alt="Image" src="https://github.com/user-attachments/assets/7c49aa50-5301-4896-a7cd-bfa0b4effc19" />

**Describe alternatives you've considered**
Integrate with VS Code)

PR #6659, opened 2026-02-14
added "close all but this tab" action, and the function to perform

Shmerlard:
solves #6615


Shmerlard:
> Strictly speaking it does not fully cover the request from #6615 but "Close all" can be done by just closing the window and even on MacOS where in theory the App could remain running without windows, we do not support his case anyway.

yes it didnt really made sense to close all unless openscad had some sort of welcome window like in vscode
the only problem that i can think of now is that if you have multiple unsaved tabs it asks you if you want to save the changes and it doesnt tell you which one. i might look into it later

PR #6660, opened 2026-02-14
chore: clickable link in local test results

flaviut:
When running, for example, `ctest -V -R '^render-manifold_surface_image$'`, the path printed would be a relative path. It takes a small amount of effort to look up the PWD and concatenate the strings in the browser.

With this change, there are two options when running this locally:

- select and copy-paste the absolute path into the browser to access results
- click the URL in compatible terminals

Incompatible terminals will just show the text by itself: https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#backward-compatibility

image

### DIFF
--- a/tests/test_pretty_print.py
+++ b/tests/test_pretty_print.py
@@ -459,7 +459,15 @@ def main():
     html_filename = os.path.join(builddir, 'Testing', 'Temporary', html_basename)
     debug('saving ' + html_filename + ' ' + str(len(html)) + ' bytes')
     trysave(html_filename, html)
-    print("report saved:\n", html_filename.replace(os.getcwd()+os.path.sep,''))
+    if os.environ.get("CI"):
+        # relative path because this ultimately this file will end up inside a zip archive
+        report_path = html_filename.replace(os.getcwd() + os.path.sep, "")
+        print("report saved:\n", report_path)
+    else:
+        report_path = os.path.abspath(html_filename)
+        report_uri = pathlib.Path(report_path).as_uri()
+        report_link = f"\033]8;;{report_uri}\033\\{report_path}\033]8;;\033\\"
+        print("report saved:\n", report_link)
 
     debug('test_pretty_print complete')
 


tests/test_pretty_print.py
```diff
@@ -459,7 +459,15 @@ def main():
     html_filename = os.path.join(builddir, 'Testing', 'Temporary', html_basename)
     debug('saving ' + html_filename + ' ' + str(len(html)) + ' bytes')
     trysave(html_filename, html)
-    print("report saved:\n", html_filename.replace(os.getcwd()+os.path.sep,''))
+    if os.environ.get("CI"):
```
t-paul, line 462: I wonder if it might be better to swap that check, so instead of checking for "is it a CI" using
```
import sys
print (sys.stdin.isatty())
```
to check if the script runs in a terminal. That said, I have not tested that actually works yet, but I would hope CIs do not run their actions in a pty.

flaviut: I hadn't considered that, but `sys.stdout.isatty()` always returns false. I think this is because this script is wrapped in ctest.

Testing `stdin` for `isatty()` just to write to `stdout` seems wrong to me. I can change it if you feel strongly about it, but my preference is the current implementation.

t-paul: It would not check `stdin` for writing to `stdout`. The check would mean "is the program running in an (user) interactive environment" without caring if the output goes directly to a terminal or a pipe. But it's probably not worth making a big thing out of it, so if you prefer to keep the `CI` check let me know, I'm fine merging as is too.

kintel:
Btw., just as a data point: The existing implementation work well for me (on iTerm on macOS), e.g.
```
report saved:
 Testing/Temporary/__apple-m1-pro_csfb_report.html
```

->
1. `open Testing/Temporary/__apple-m1-pro_csfb_report.html` opens the file in a brower
2. Cmd-click on the filename in the terminal also opens it in a browser

t-paul:
Well, lets see how it goes, hopefully not breaking the macOS case :grimacing:.

kintel:
> Well, lets see how it goes, hopefully not breaking the macOS case 😬.

All good here!

flaviut:
I didn't even think that maybe I needed to enable more smarts in my terminal.

Anyway, I'm glad that this doesn't break anyone's workflow, and thanks for merging. If it doesn't make sense to merge stuff like this in the future, that's fine with me, I have a good workflow with [git-pile](https://github.com/keith/git-pile) that means it's no trouble at all for me to just keep them on my personal fork.

PR #6661, opened 2026-02-14
Rollback header file sorting for spnav driver.

t-paul:
Temporary revert as spnav.h brings in X11 headers that pollutes the global namespace by #defining Status causing build failure with spnav enabled.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 .cache/
 .vs/
 .vscode/
+*.code-workspace
 .idea/
 .clangd/
 /*.scad
--- a/src/gui/input/SpaceNavInputDriver.cc
+++ b/src/gui/input/SpaceNavInputDriver.cc
@@ -30,16 +30,14 @@
  */
 
 #include "gui/input/SpaceNavInputDriver.h"
-
-#include <spnav.h>
-#include <unistd.h>
-
-#include <QThread>
-#include <string>
-
 #include "gui/input/InputDriverEvent.h"
 #include "gui/input/InputDriverManager.h"
 #include "utils/printutils.h"
+
+#include <QThread>
+#include <spnav.h>
+#include <unistd.h>
+#include <string>
 
 void SpaceNavInputDriver::run()
 {


kintel:
@hzeller FYI

hzeller:
Yep, seen that. Rollback on that file sounds good., then we can fix the issue separately.

thehans:
Should there be [IWYU pragmas](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md#iwyu-pragma-begin_keepend_keep) added to prevent this issue?

t-paul:
Yeah, that's just a band-aid for now as it breaks any build that enables the spnav driver.

We need to select a way on how to best handle those type of issues, preventing them from randomly coming back.

Felixoid:
Appreciate the patch, currently both https://aur.archlinux.org/packages/openscad-git and https://aur.archlinux.org/packages/openscad-qt6-git are broken

Tested it as the following `patch -d "openscad" -p0 < spnav.patch` in the build script, and it worked out

```
--- src/gui/input/SpaceNavInputDriver.cc
+++ src/gui/input/SpaceNavInputDriver.cc
@@ -30,17 +30,15 @@
  */

 #include "gui/input/SpaceNavInputDriver.h"
+#include "gui/input/InputDriverEvent.h"
+#include "gui/input/InputDriverManager.h"

+#include <QThread>
 #include <spnav.h>
 #include <unistd.h>
-
-#include <QThread>
 #include <string>

-#include "gui/input/InputDriverEvent.h"
-#include "gui/input/InputDriverManager.h"
-#include "utils/printutils.h"
-
 void SpaceNavInputDriver::run()
 {
   while (spnav_input()) {
```

kintel:
We should probably get a CI build going with spacenav enabled, so we can detect this before all the other folks..

t-paul:
Yeah, enabling most flags probably makes sense. Noticing was not an issue though, I noticed from maybe 50 failure mails from OBS about an hour after merge :grinning:.

kintel:
heh, fair. We'd have to block merges on it to move the needle I guess..

Issue #6005, opened 2025-06-28
Create Warnings with the assert function


Nexusnui:
**Is your feature request related to a problem? Please describe.**
I would like an option to issue warnings with the assert function, instead of a full error, that stops rendering.

**Describe the solution you'd like**
New parameter in the assert function, that enables a warning mode. The default should result in the current behaviour. This new parameter should probably be a boolean, unless there is a reason for additional modes.
Alternatively there could be a new function for warnings.

**Describe alternatives you've considered**
Use Echo to show warning, but this is not very visible in the console. Additionaly users of a library might not expect them and they do not appear in the Error-Log.

jimb15122:
This is another one that grok thinks already exists!

"Yes — OpenSCAD has a warn() function (since 2021.01) that works exactly like echo(), but outputs to the warning console (yellow text) instead of the regular log."

Use case: when parameters of minor importance conflict, possibly due to computed sizing calculations, but a logical outcome is still possible, it would be nice to let the user know in a noticeable way without aborting the entire construction.

eg:   
module hollow_cylinder(r=10, h=10, wall=2, center=false, rounded=false) 
// aka pipe with possibly rounded end edges

if (wall >= r) 
 warn(str("WARNING 'r<=wall' HOLLOWING IGNORED - hollow_cylinder(r=", r, ", h=", h, ", wall=", wall, ")"));

don't care if implemented as an echo replacement as shown, or as assert_warn()

Nexusnui:
I came up with these possible implementations:
- working like like echo:
`warn("This is a Warning");`
- like assert:
`warn(condition = r<=wall, message = "This is a warning");`
- extension of echo:
`echo("This is a Warning", is_warn=true);`
- extension of assert:
`assert(condition = r<=wall, message = "This is a warning", is_warn=true);`

In the extension cases is_warn should be false by default for backwards compatability.

jimb15122:
Another option, that requires no new syntax, would be to simply honor current color() for echo to set the text or bk color.   But this would affect existing echo output.  Color coding echo's could have real utility.  Perhaps a config setting to enable it.  or how about. #echo(). !!

Nexusnui:
But a warning also provides the file and line where the warning was issued and not simply a colored text.

Nexusnui:
You can also click on the warning to jump to the source.

jimb15122:
[...]

Issue #2884, opened 2019-03-20
Add error command


adrianVmariano:
The only existing mechanism for code or libraries to generate errors is assert.  It is currently impossible to generate warnings from user code.  

When assert is used to produce an error it looks like this:

    assert(big_complex_condition, "message for the user explaining the problem")

produces

ERROR: Assertion 'big_complex_condition': "message for user explaining the problem" failed in file test.scad, line 10

This is a rather cluttered output with the text of the condition, extra quote marks, and the word "failed".  It does not look the same as a normal error message.  I propose a new error command that might work like this in a function:

    x <= 0 ? error("x must be positive in function foo"): 
                  do_something(x,y);

and the output would not have extra quote marks, or the word "failed", but would look the same as native error messages:

ERROR: x must be positive in function foo in file test.scad, line10

Perhaps the error message display could give the function name automatically.  

In a module it would be:

    if (x>0) { make_some_geometry } 
    else error("x must be positive for module foo");

For warnings the error command could take a flag, so 

    error("text",warn=true)

would produce a warning.   I was thinking that it made sense for error to be unconditional to make the forms above simple, but it's not clear that unconditional warnings make sense since the code needs to continue.   If it does take a condition the test should be the reverse of assert, namely that when the condition is true, the error or warning occurs.   Perhaps an unconditional error statement and a conditional warning statement?

t-paul:
I think we should go with separate `error()` and `warning()` as suggested by @doug-moen in the forum discussion.

Sharma-Hrishabh:
Hey @t-paul , I want to work on this issue. It will help me in understanding the parser code.

t-paul:
Sure, we can discuss details here and/or on IRC. But note that language changes will require a lot of discussion, even for this one which is relatively simple. The reason for that is those are pretty much one-way changes. There's not much chance to change the calling conventions once the feature is released as that would break all scripts using the feature. While not perfect, we try hard to prevent that.

dearmash:
It's been a while since this issue was reported and PR proposed. I'm currently looking for this exact thing, but there's no good way I know of to +1 the issue besides just commenting on it.

jordanbrown0:
I would add `error()` and `warning()` modules and magic functions, with exactly the same arguments as echo().

(I say "magic" functions because they would want to be similar to the let, echo, and assert "functions", which are really not exactly functions because they take not only an argument list but also an expression afterward.)

dearmash:
[...]

PR #6665, opened 2026-02-15
Add warning capabilities to assert() with an abort parameter

AncientRoman:
This pr adds an optional abort parameter to assert() to control whether assert errors and throws or just warns. Abort defaults to true for backwards compatibility. Note that an assert warning won't generate a traceback (same as other warnings) because the current implementation of tracebacks requires throwing exceptions.

I had a need for this functionally in some libraries I was writing and it turns out this was discussed in #6005 last year.

This update includes regression tests for the new functionality and I updated the assert tooltip parameter names to match the documentation and actual parameter names. The old tooltip was `assert(boolean, string)`; the new is `assert(condition, message, abort = true)`.

Let me know if I'm missing anything or if you guys would prefer a different parameter name besides abort.
Also, the pre-commit hook ended up trimming some trailing space in tests/CMakeLists. Sorry that file's diff is a bit messy.


jordanbrown0:
So you want an assertion that only warns?

That seems like it's only a little different from an `if` and an `echo`; the only difference would be the prefix, the color coding, and the ability to automatically include the failing expression.

Note that warnings are fatal if you have "Stop on the first warning" set - as you really should.

I'd prefer that the sense of the flag be that `true` means to warn, so that not-present, `undef`, and `false` all mean the same thing.

AncientRoman:
Yeah pretty much. It would just be a way to create my own warnings. Openscad currently allows making errors and logs but not warnings. Like you said, you get the warning prefix, the color, the file and line number, and the little "compilation generated x warnings" message.

I'm currently using it for deprecation and changed/removed parameter warnings but I could see other use cases. The default cube() for example issues a warning instead of an error on bad arguments and libraries could do the same.

I'd be happy to add a `warn(...)` function similar to echo() instead of modifying assert(). Perhaps an `error(...)` function too? Although that makes assert a bit redundant so I don't know.
Or I could change abort to maybe warn? So the default would be `assert(condition, message, warn = false`).
Also, if this isn't something openscad is interested in, I'm happy to delete my pr. No worries.
Just let me know.

jordanbrown0:
Strongly related to the request in #2884 .

PR #6670, opened 2026-02-16
Run example tests concurrently with existing tests in PRs

### LINKED ISSUE #6627: Make Examples tests pass
Examples tests are currently failing, see e.g. https://github.com/openscad/openscad/actions/runs/21781289454

Get this sorted so the tests are green.

coryrc:
Fixes #6627

### DIFF
--- a/examples/Advanced/surface_image.scad
+++ b/examples/Advanced/surface_image.scad
@@ -5,13 +5,15 @@ echo(version=version());
 // The example takes 3 cuts from the height map and displays
 // those as 3 stacked layers.
 
-for (a = [1, 2, 3])
-    color([a/6 + 0.5, 0, 0]) translate([0,0,2*(a-1)])
-       linear_extrude(height = 2, convexity = 10)
-            projection(cut = true)
-                translate([0, 0, -30 * a])
-                    surface("surface_image.png", center = true);
-
+for (a = [1, 2, 3]) {
+    overlap = a==1 ? 0 : 0.1; // Overlapping ensures one mesh.
+    color([a/6 + 0.5, 0, 0])
+        translate([0,0,2*(a-1)-overlap])
+            linear_extrude(height = 2+overlap, convexity = 10)
+                projection(cut = true)
+                    translate([0, 0, -30 * a])
+                        surface("surface_image.png", center = true);
+}
 
 
 // Written in 2015 by Torsten Paul <Torsten.Paul@gmx.de>
--- a/tests/regression/dump-examples/surface_image-expected.csg
+++ b/tests/regression/dump-examples/surface_image-expected.csg
@@ -1,27 +1,33 @@
 group() {
 	color([0.666667, 0, 0, 1]) {
-		linear_extrude(height = 2, $fn = 0, $fa = 12, $fs = 2, convexity = 10) {
-			projection(cut = true, convexity = 0) {
-				multmatrix([[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 1, -30], [0, 0, 0, 1]]) {
-					surface(file = "surface_image.png", center = true, invert = false, timestamp = 1544039412);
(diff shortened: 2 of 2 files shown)


coryrc:
Oops, I didn't rerun *all* the example tests after making changes. Fixing.

Also, the "branch protection" will need to be updated with the new names for the test workflows, or I could create a dummy job with the same name as the old one to run successfully only when the tests pass.

.github/workflows/linux-tests.yml
```diff
@@ -18,39 +17,79 @@ jobs:
...
+          retention-days: 1
 
+  test:
+    needs: build
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu-22.04, ubuntu-24.04 ]
+        qt: [ qt5, qt6 ]
+        task: [ "test", "test_examples test coverage" ] # examples are faster than non-examples, so they get coverage
```
kintel: The coverage stuff has probably bitrotted a bit, but the intent is to run coverage over the whole test run to look for untested code. As such, it should probably run on the main test run, as that's where we cover corner cases.

examples/Advanced/surface_image.scad
```diff
@@ -6,8 +6,8 @@ echo(version=version());
 // those as 3 stacked layers.
 
 for (a = [1, 2, 3])
-    color([a/6 + 0.5, 0, 0])
-       linear_extrude(height = 2 * a, convexity = 10)
+    color([a/6 + 0.5, 0, 0]) translate([0,0,2*(a-1)])
+       linear_extrude(height = 2, convexity = 10)
```
kintel: I would make height 2.1, and subtract 0.1 in translate, to disambiguate between 3 disconnected slabs vs. one solid object

kintel:
> 
> Also, the "branch protection" will need to be updated with the new names for the test workflows, or I could create a dummy job with the same name as the old one to run successfully only when the tests pass.

Dummy job makes sense, and we can split it up later if needed.

.github/workflows/linux-tests.yml
coryrc: Coverage doesn't count the regression tests AFAICT -- how could it? It just does:

```
		echo "Generating code coverage report..."
		uv run --project "$REPO_ROOT/.github/ci/openscad-coverage" gcovr -r ../src CMakeFiles/OpenSCADLibInternal.dir $PARALLEL_GCOVR --html --html-details --sort uncovered-percent -o coverage.html
```

However, I now see the workspace upload and time to reinstall apt packages takes 3 minutes to run; and so do the example tests. So there's no reason to spin up separate VM.

kintel: > Coverage doesn't count the regression tests AFAICT -- how could it? It just does:
> 
We run CMake with `-DPROFILE=ON`, which instruments the code to collect coverage info and output into a database, which is later read by `gcovr`.

examples/Advanced/surface_image.scad
coryrc: Is that because this serves as an example and you think people should write their code this way?

.github/workflows/linux-tests.yml
coryrc: I added a comment to the script with that info, so I won't ask again :)

examples/Advanced/surface_image.scad
kintel: It's because coincident surfaces yield an undefined result, which may change between Manifold versions. In practice we always have to use some overlap to disambiguate, so it's also best practice when using OpenSCAD.

coryrc: Along with the required modifications, I added a comment:

`   overlap = a==1 ? 0 : 0.1; // Overlapping ensures one mesh.`

coryrc:
And currently we just no longer run throwntogether for example009, which solves the problem. Preview and render do get run.

kintel (approved):

PR #6675, opened 2026-02-17
Adds a font.version string to fontmetrics().  Fixes #6668.

### LINKED ISSUE #6668: Report font version in fontmetrics() and Font List window
Diagnosing font-version issues might be easier if there was a trivial way to extract the font version identifier.  In particular, it might have made diagnosing a test failure in #6658 easier.)

Report the font version.  It's claimed that it's available from [FT_Get_Sfnt_Name()](https://freetype.org/freetype2/docs/reference/ft2-sfnt_names.html), looking for name_id == TT_NAME_ID_VERSION_STRING.

Inspect the TTF file with Windows' font viewer.  (But:  that doesn't directly confirm that OpenSCAD is using **that** font file.)

**Additional context**
We can split this into fontmetrics() and Font List, separately, if we like.

jordanbrown0:
Well, sort of.  The FreeType documentation is somewhat opaque on encoding of these strings, so the interpretation here is very limited and conservative.

A quick look at Font List suggests that it gets its information from Fontconfig.  Fontconfig does have a "version" attribute, but (a) it's an integer, not a string, and (b) it's not present on any font on my Windows 11 system.

### DIFF
--- a/src/core/FreetypeRenderer.cc
+++ b/src/core/FreetypeRenderer.cc
@@ -47,6 +47,9 @@
 #include "utils/printutils.h"
 
 #include FT_OUTLINE_H
+#include FT_TRUETYPE_IDS_H
+#include FT_SFNT_NAMES_H
+
 // NOLINTNEXTLINE(bugprone-macro-parentheses)
 #define SCRIPT_UNTAG(tag) \
   ((uint8_t)((tag) >> 24)) % ((uint8_t)((tag) >> 16)) % ((uint8_t)((tag) >> 8)) % ((uint8_t)(tag))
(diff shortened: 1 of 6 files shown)


jordanbrown0:
Example result from fontmetrics(), with the actual "metric" stuff elided:
```
ECHO: {
    ...
    font = {
        family = "Liberation Sans";
        style = "Regular";
        version = "Version 2.00.1";
    };
}
```
Note that `font.version` is the string as reported; as reported it includes "`Version `".

jordanbrown0:
The failures are due to the fact that we deliver Liberation Sans Regular 2.00.1, but Ubuntu 24 delivers 2.1.5 and we end up picking the system copy of the font.

kintel:
Does it pick up the system font because it's newer, or because it has higher priority somehow?

Also, should we bump the version of the builtin, or do you prefer keeping it as is for testing?

jordanbrown0:
I think it's picking up the system font because it's got a newer version.

As best I can tell - ref the underline PR - 2.1.5 is incorrect; it has the underline in the wrong location.  I would not upgrade to it.

jordanbrown0:
Most recent change redacts the version number from the fontmetrics output, confirming that it *is* set, but not what it is.

jordanbrown0:
That works around #6774.

Otzie2023:
Re the underline objection to bumping the builtin: I posted measurements in
#6658. Short version — the stroke sits in exactly the same place in 2.00.1 and
2.1.5 for all twelve faces; only the meaning of `post.underlinePosition` changed,
and 2.1.5 is the one that follows the OpenType spec.

jordanbrown0:
I haven't looked at this in a long while, but I did look at it pretty carefully and found that the misbehavior I observed was pretty directly attributable to a since-fixed FontForge bug - that the older fonts were correct.

PR #6676, opened 2026-02-17
Include configured step for spinbox param export

### LINKED ISSUE #6597: `--export-format=param` should include spinbox settings
At the moment, `--export-format=param` only lists the default value for a spinbox field. It should also include additional settings, such as the step size. This will allow alternative customizer UIs to properly respect those settings.

This probably applies to other fields like the slider too, but I haven't tested those.

yawkat:
For number values that have a step but no min or max value, the step was not included in the `--export-format=param` output.

Fixes #6597


yawkat:
@t-paul PTAL

yawkat:
I suspect some customizer sites use it internally but don't publish the fact :(

PR #6681, opened 2026-02-20
Touchpad pinch-to-zoom and configurable scroll behavior

evan-goode:
Hello!

This PR is an attempt at improving the experience of 3D navigation on touchpads (issue https://github.com/openscad/openscad/issues/4297). With these changes, OpenSCAD can be configured to behave roughly the same as Blender's touchpad navigation.

First, I've added a handler for the pinch-to-zoom touchpad gesture. Rotating two fingers while pinching rotates about the azimuth similar to OpenSCAD's default left click+drag or Blender's pinch-to-zoom handler in "turntable" orbit mode. Pinching to zoom with shift held adjusts the FOV, like Shift+Scroll currently does.

Second, I've built onto https://github.com/openscad/openscad/pull/5874 and added controls to configure scroll behavior. On trackballs and touchpads, I personally prefer mapping Scroll to "Rotate in altitude/azimuth", Shift+Scroll to "Pan left/right & up/down", and Ctrl+Scroll to Zoom. That seems to match touchpad navigation in Blender.

It might be nice to configure touchpad scrolling distinctly from mouse wheel scrolling, but `qWheelEvent->device()->type()` always returns `QInputDevice::DeviceType::Touchpad` on my Wayland system, even for mouse scroll events, so I gave up on that.

I have only tested these changes on Wayland/Linux, as I don't have easy access to Windows or macOS systems with touchpads, so help testing would be much appreciated!

### DIFF
--- a/src/gui/QGLView.cc
+++ b/src/gui/QGLView.cc
@@ -44,7 +44,6 @@
 #include <QOpenGLWidget>
 #include <QSurfaceFormat>
 #include <QWidget>
-#include <QPinchGesture>
 #include <QNativeGestureEvent>
 #include <iostream>
 #include <QApplication>
@@ -109,7 +108,6 @@ void QGLView::init()
   this->statusLabel = nullptr;
 
   setMouseTracking(true);
-  grabGesture(Qt::PinchGesture);
 }
 
 void QGLView::resetView()


hoverboardhavoc:
I tested this on Mac Os 26.2 (25C56). For me, two finger drag to pan and pinch to zoom should be on by default on mac. I chose mouse preset "Open Scad (Mac Os)" and pinch-to-zoom and drag to pan did not work.

evan-goode:
Rebased and attempted to fix the Qt5 compilation error.

> I tested this on Mac Os 26.2 (25C56). For me, two finger drag to pan and pinch to zoom should be on by default on mac.

Thanks for testing. The maintainers can decide whether the default behavior for touchpad scroll should be changed to pan, on macOS or all platforms. Personally, I would prefer mapping touchpad scroll to pan on all platforms by default _if_ it's possible to distinguish touchpad scroll events from mouse wheel scroll events. Mouse wheel scroll events should probably still be zoom by default. Ideally touchpad scroll and mouse scroll behavior could then be customized separately. I tested your `NoScrollPhase` trick from https://github.com/openscad/openscad/pull/6731 and it also correctly discriminates touchpads from trackballs and mice on Wayland, but I'm not sure how "correct" or reliable it would be.

> I chose mouse preset "Open Scad (Mac Os)" and pinch-to-zoom and drag to pan did not work.

I didn't change any of the presets (yet). The pinch-to-zoom gesture should work always without changing any bindings. It seems the only real difference between your PR and mine is the `grabGesture` call. We're both using `QEvent::NativeGesture`. I took out the `grabGesture` since it didn't actually do anything on my system. Maybe pinch-to-zoom works for you now?

You can map scroll to pan in Edit -> Preferences (may be different on macOS) -> Mouse -> Change preset to "Custom" -> Change "Scroll" to "Pan left/right & up/down".

hoverboardhavoc:
I tested with a6f41c041f2c32078f950a487a72148b99d1aa9a and pinch to zoom works. Pan also works if you set "scroll" to pan left/right up/down. The trouble is, as a user, I'm not sure I understand what scrolling is in the context.

10a5er:
happy to test on aarch64 linux the moment it lands on nightly <3

PR #6683, opened 2026-02-21
Restore window state after show()

kintel:
Fixes an issue where restoring Window state was sometimes not working correctly on macOS


kintel:
Did not really do anything useful

PR #6685, opened 2026-02-22
Implement subtraction color for nefs and add compile switch to force PolySetRenderer

ali1234:


ali1234:
Implementing subtraction color for nef -> polyset prevents a regression in a hundred or so render tests when replacing CGALRenderer with PolySetRenderer. It also breaks six preview tests by making the result "more correct" than the tests expect, so those tests have been updated.

With the exception of render-preserve-colors.scad (which uses `color()`), previews are now identical between CGAL and Manifold in the updated tests, and *all* tests now pass when forcing everything through PolySetRenderer.

ali1234:
Found one more test that needs updating. For some reason it doesn't run locally.

kintel:
Just to clarify: You want to merge this with the renderer disabled for now, then deal with enabling the renderer later?

ali1234:
The first patch implements Nef colours in PolySetRenderer. It is unconditional. It does not change which renderer gets used. It only changes the output when PolySetRenderer was already used for Nefs.

The second patch optionally removes CGALRenderer from the build and forces PolySetRenderer to be used instead. This still passes all tests but does change some things, like how edges are rendered. I wouldn't be against setting it ON by default to get feedback from nightly users, but the choice is yours really.

kintel:
Thx, I say do one thing at the time, and I'll try to turn it on locally to give it some testing

tests/regression/preview-cgal/internal-cavity-expected.png
kintel, line 1: The right solution is to remove internal-cavity.scad from the `PREVIEW_DIFFERENT_EXPECTATIONS` list in tests/CMakeLists.txt, then remove the per-backend goldens `tests/regression/preview-cgal/internal-cavity-expected.png` and `tests/regression/preview-manifold/internal-cavity-expected.png`

Same for the 3 other files with preview changes

tests/regression/throwntogether/issue913-expected.png
kintel, line 1: Not sure I completely understand why this changed. The yellow object is considered a negative object, but why isn't it green?

tests/regression/preview-cgal/internal-cavity-expected.png
ali1234: Working on this but it is a bit more complicated:

 * render-preserve-colors can't be removed because it is still different.
 * Removing a test from `PREVIEW_DIFFERENT_EXPECTATIONS` changes the types of test that are run on it.
 * render-tests has z-fighting in the throwntogether test, which previously did not run.
 * Some of the tests don't have an expected case for throwntogether.
 * There is a note in the cmake: "we're missing coverage for ${PREVIEW_DIFFERENT_EXPECTATIONS} files for throwntogether"

tests/regression/throwntogether/issue913-expected.png
ali1234: I'm not sure but this is how Manifold has always displayed it. example016.scad has the same issue.

ali1234: This is https://github.com/openscad/openscad/issues/5218 - note that Manifold versions of issue913 and example016 are already disabled with a comment pointing to that bug. Somehow implementing nef colour makes this bug affect CGAL as well. This might provide a clue how to fix it...

tests/regression/preview-cgal/internal-cavity-expected.png
kintel: It's slightly more complex. Let me know if you'd rather I do it and push to your  branch

ali1234: No problem, I got it done. Just about to push.

ali1234:
Oops, I accidentally bumped Manifold.

ali1234:
Those last two tests are failing because of the z-fighting. Will disable them.

tests/regression/throwntogether/issue913-expected.png
kintel: Ah, good catch!

tests/CMakeLists.txt
```diff
@@ -627,13 +627,8 @@ list(APPEND PREVIEW_DIFFERENT_EXPECTATIONS
   ${TEST_SCAD_DIR}/3D/issues/issue1137.scad
 
   # render() in manifold mode preserves colors
-  ${TEST_SCAD_DIR}/3D/features/render-tests.scad
-  ${TEST_SCAD_DIR}/misc/internal-cavity.scad
   ${TEST_SCAD_DIR}/3D/features/render-preserve-colors.scad
```
kintel, line 630: This one can also be removed right?

ali1234: Do you mean render-preserve-colors.scad? No, that cannot be removed because it has a mixture of green subtraction faces and explicitly coloured faces. The latter does not work in CGAL so the previews are still different.

kintel: Duh, sorry, was not looking at the right results..

PR #6692, opened 2026-02-26
Update dxf

mconsidine:
Improving DXF export options with runtime selections.


mconsidine:
A local compilation that added R14 versions to the render-dxf tests passed those tests.

I'm thinking perhaps a decent dxf test would be to use the info output of ezdxf, rather than try to match a specific variant of R14 output?  But I'm not sure how to specify a test like that in tests/CMakeLists txt

And a future pull will have R14 fields further tweaked so that they pass the viewer.autodesk.com hurdle, which seem to finally have working.

Advice on any of the above is welcome.  If I get all that squared away then I'll look to contribute my version of DxfData.cc, which imports R10/R12/R14 that the exporter creates.

mconsidine:
I've added tests based on comparison to output from ezdxf to the update_dxf branch of my repo here https://github.com/mconsidine/openscad
A spot check of some of the resulting R14 files show they are accepted by viewer.autodesk.com.  "legacy" output predictably fail the ezdxf test - that has been the issue.  R10 and R12 output is enabled.  Even though openscad does not (yet) read those back, they are also accepted by viewer.autodesk.com, viacad and a couple of others in limited testing on linux.
In addition to a python script to deal with the ezdxf auditing, there is a set of "expected" results under the regressions folder.  The folder structures for the dxf tests are render-dxf/R14 (in the case of R14) and failed tests should have a line in the "actual" file that gives an idea what happened.  Otherwise a successful result just has "DXF audit passed."
For windows and mac I changed the .yml to make sure ezdxf was installed prior to ctest being run, but I have not tested that.  For linux, ezdxf was added to the prerequisites for debian in the script called by the github workflows.
As a finaly sanity check I cloned the branch noted above into a fresh folder on my machine, compiled, ran the scripted tests and did a couple of manual tests.  I didn't notice anything breaking that shouldn't have :)
Hope this helps.  If I need to reissue a PR or otherwise fix something, please advise.

mconsidine:
I'm at a loss as to what to do next.  I fixed the formatting issue on my repo, or so I thought.  And I don't know why the app image creation would fail, as I based this on the current master.  Am open to advice.
EDIT: there is a bash script called create_dxf_expected.sh that should be run before the build process and it takes as an argument the release,eg "R14" (without quotes).  It wasnt clear to me where that should be triggered in a build script so that it doesnt need to be done manually.

PR #6693, opened 2026-02-28
fix(gui): guard against null modinst and path in right-click context menu

nomike:
I found a bug in PythonSCAD which causes crashes. I fixed it but then saw, that the relevant code bits are the same in OpenSCAD.

I wasn't able to get OpenSCAD to crash in the same way as PythonSCAD, but nevertheless, IMHO this bugs should be fixed in OpenSCAD as well.

## Summary

Fixes a SIGSEGV crash when right-clicking the 3D view on certain objects.

## Problem

`MainWindow::rightClick()` iterates CSG tree nodes to build a context menu backtrace. Some nodes can have a null `modinst` pointer or a `Location::NONE` location. Accessing `step->modinst->location()` on such nodes, and then passing `location.filePath()` to `get_library_for_path()`, causes a crash.

**Backtrace:**
```
#0  std::filesystem::path::has_filename() — SIGSEGV
#1  path_contains_file(std::filesystem::path, std::filesystem::path)
#2  get_library_for_path(std::filesystem::path const&)
#3  MainWindow::rightClick(QPoint)
```

**Root cause:**
`Location::filePath()` dereferences `*path` (a `shared_ptr<fs::path>`) without a null check. The sibling method `fileName()` already has this guard (`path ? path->generic_string() : ""`), but `filePath()` does not. When `rightClick()` encounters a node with null `modinst` or an empty location, the unguarded dereference crashes.

## Fix

1. **`MainWindow::rightClick()`**: Skip nodes with null `modinst` or `Location::NONE` in the backtrace loop, preventing access to invalid data.
2. **`Location::filePath()` in `AST.h`**: Add a null guard matching the existing one in `fileName()` — return a static empty `fs::path` instead of dereferencing a null `shared_ptr`.

## Testing

Tested by right-clicking various objects in the 3D view that previously triggered the crash. The context menu now works correctly, skipping nodes without valid location information.


src/gui/MainWindow.cc
```diff
@@ -2066,7 +2066,13 @@ void MainWindow::rightClick(QPoint position)
       if (step->name() == "root") {
         continue;
       }
+      if (!step->modinst) {
+        continue;
+      }
       auto location = step->modinst->location();
+      if (location.isNone()) {
+        continue;
+      }
```
kintel: Instead of continuing here, shouldn't we rather insert an element indicating that this is an invocation without a source reference?

I'm not sure how invocations without source references could even happen. That kind of sounds like an issue somewhere else.

nomike: I've tried to address your remarks (see commit comment for details).

And yes, it's a bit strange that there are invocations without source. Perhaps this only happens with Python or in PythonSCAD. It might hint to another issue, it could also be that Python just behaves different in this regard.

That would explain why in PythonSCAD it's easy to make it crash but I wasn't able to reproduce the issue in OpenSCAD.

Regardless IMHO the guards this PR introduces are a good thing and once this is merged and I was able to port this to PythonSCAD, I will locally add some debug print statements and have a look what's actually going on.

kintel:
Thx, I assume it's a Python thing, otherwise having a fix for something that should never happens sounds like the wrong fix :) Let me know what you find, and if you find something interesting, it would be great to add appropriate comments to this code so that the next person reading it understands what it's for.

PR #6704, opened 2026-03-05
Fix/memleak

gsohler:
Even if a memory is just alocated once and its certain that it's returned the OS  on program termination, 
its still good practice  not to allocate memory without freeing it.
This will help to identify the real memory leaks when not polluted by lots of dummy one.
this PR addresses quite some of them.

### DIFF
--- a/src/core/builtin_functions.cc
+++ b/src/core/builtin_functions.cc
@@ -1375,7 +1375,7 @@ void register_builtin_functions()
                    "has_key(object, key) -> boolean",
                  });
 
-  static BuiltinFunction func_import(builtin_import, &Feature::ExperimentalObjectFunction);	
+  static BuiltinFunction func_import(builtin_import, &Feature::ExperimentalImportFunction);	
   Builtins::init("import", &func_import,
                  {
                    "import(file) -> object",


src/gui/ColorList.cc
```diff
@@ -142,6 +142,8 @@ ColorList::ColorList(QWidget *parent) : QWidget(parent), ui(new Ui_ColorListWidg
 ColorList::~ColorList()
 {
   delete ui;
+  for (const auto &item : itemList)
+    delete item;
```
kintel, line 146: This should be solved separately:
* The layout should be modified to take ownership over the item in `addItem()` to conform with other layouts.
* ColorList itself should probably inherit from Ui::ColorListWidget instead of manually manage the UI component, to align with how all other dialogs are managed in OpenSCAD

src/core/GroupModule.cc
```diff
@@ -43,7 +43,8 @@ std::shared_ptr<AbstractNode> builtin_group(const ModuleInstantiation *inst, Arg
 
 void register_builtin_group()
 {
-  Builtins::init("group", new BuiltinModule(builtin_group),
+  static BuiltinModule mod(builtin_group);
+  Builtins::init("group", &mod,
```
kintel, line 47: These are not really global modules, they get registered to `Builtins`, so IMO, they should ideally be deleted by `~Builtins` rather than turning them into static storage tied to the process.

Issue #1155, opened 2015-01-13
Feature Request: Export to SCAD polyhedron?


lordofhyphens:
I'very been wondering about being to export a compiled design to a set of polyhedron statements (one per separate object), with points and faces. I believe that it will help facilitate modifying existing STLs and provide a pseudo-debugging output. Additionally, I believe it would permit a more iterative work flow.

NateTG:
The STL files are already really close to that, and it's pretty easy to write an external utility that does that.

Can you give an example of what you want to do with a feature like that?

lordofhyphens:
I'd like to investigate iterative compilation--using multiple passes to build objects on top of objects (similar to how multiple LaTeX passes are used for cross-references in a manuscript). Currently, the way to put something "on top" of some other complex object is to step back and do a stack of math or guess/check repeatedly. An export to a polyhedron() statement would let us precompute the geometry and import it into another design that builds on it.

Doing everything with SCAD primitives that are readable, I think, will give be better reliability and transparency than import(). I also find myself wanting to do some manipulation of STLs I don't have the source for, and polyhedron() is a lot more readable than the raw STL format, especially if that STL is binary and not text.

t-paul:
I guess the geometry feedback part is discussed in #301 already.

For hand editing, AMF might be similar complex as a polyhedron (or maybe even easier as it's possible to process it with command line XML tools).

lordofhyphens:
I wasn't aware that import() allowed AMF files.

"Nothing unreal exists." - Kiri-kin-tha's First Law of Metaphysics.

t-paul:
The AMF import is waiting to be merged (probably after the release is out).

jmerc77:
Similar idea, but directly access points and faces from a module instead of needing to export then import. Wasn't sure if anyone knew they wanted this, though. Also thought it would be cool if it worked for 2d modules, too.

gsohler:
Such a Feature exists in my OpenSCAD Fork. The function is called 'mesh'.
Its only available on Python Side yet.it Returns a tuple with Point Array
and vertices, similar to polyhedron Input.
'mesh' can be Used together with polyhedron to create fancy translations.




jmerc77 ***@***.***> schrieb am Di., 29. Aug. 2023, 10:15:

t-paul:
I believe this is really just asking for an export format that produces a single file containing a single `polyhedron()` call that is a valid scad script.

AryaBhatt45:
[...]

PR #6706, opened 2026-03-06
Feature: Implement OpenSCAD Polyhedron Export (Issue #1155)

arybhatt4533:
Hi @t-paul,

I've been working on Issue #1155 to understand how OpenSCAD handles mesh data, as we discussed. I have implemented a new feature that lets users export their rendered 3D designs directly into a .scad file as a polyhedron(). I believe this is a good first step for my GSoC project on STEP export.

Here is what I've done:

The Logic: I wrote the export_scad_polyhedron function to pull the points and faces from the geometry.

Clean Code: I used createSortedPolySet so that the output stays the same every time you export it.

System Integration: * I added SCAD to the FileFormat list so the system recognizes it.
I updated is3D() so OpenSCAD knows this is a 3D format.
I added the "OpenSCAD Polyhedron (*.scad)" option to the actual Export menu.
Routing: I set up the switch case to make sure the right function is called when someone clicks export.
I am also focusing on my GSoC project proposal now and will share it soon. Please let me know if I should change anything in this PR.

### DIFF
--- a/src/io/export.cc
+++ b/src/io/export.cc
@@ -39,8 +39,6 @@
 #include "geometry/PolySet.h"
 #include "geometry/linalg.h"
 #include "glview/Camera.h"
-#include "glview/ColorMap.h"
-#include "glview/RenderSettings.h"
 #include "utils/printutils.h"
 
 #define QUOTE(x__) #x__


AryaBhatt45:
I noticed the CI failure regarding the missing PolySetGeometry.h header in the WASM and AppImage builds. I'm currently investigating if this is a path inconsistency or a missing submodule dependency in the build environment. I will push a fix shortly to get the checks back to green!

PR #6710, opened 2026-03-07
Fix #6705: Stabilize perspective projection on resize and dynamic clipping

### LINKED ISSUE #6705: preview window loses its "perspective"
Display incorrectly switches from orthogonal mode to perspective mode. Although this report uses the lastest nightly I have noticed this behaviour in nightlies for at least the last 3 weeks. This report provides a consistent reproducible example of the problem, but it is not the only situations in which the problem is encountered... this is just a presentation of one procedure I discovered that always exhibits the behaviour.

Launch OpenSCAD and load a file. (I reproduce the error by selecting a file from the splash intro window)

Image

Render the file (I reproduce the error with preview)

Image

Click the "View All" button

Image

Click one of the "left", "right", "back" front buttons.
The window is displayed in "perspective" mode although orthogonal mode is still selected.

Image

Workaround: click the orthogonal view button (even though it is already selected)

Image

OS: Mac Studio Tahoe 26.4
OpenSCAD: 2026.03.01 (git 447f1d69) downloaded from openSCAD website

arybhatt4533:
Overview
This PR addresses the issue where the perspective view would intermittently flatten or "pop" during window resizing and extreme zooming. I have modified the camera setup logic to ensure the projection matrix is synchronized with the viewport size and camera distance.

Key Changes

- Immediate Projection Update: Updated GLView::resizeGL() to call setupCamera() immediately after updating the aspect ratio. This prevents frames from being rendered with an outdated projection matrix during rapid resizing.
- Dynamic Clipping Planes: Replaced static near/far clipping values with dynamic calculations based on cam.zoomValue().
   $zNear = \max(0.1, 0.01 \times dist)$
   $zFar = \max(100.0, 1000.0 \times dist)$
   This ensures objects remain visible and depth precision is maintained even at high zoom levels.

- Syntax & Scope Fix: Refactored the switch block in setupCamera() to properly handle scope and prevent accidental fall-throughs.

### DIFF
--- a/src/core/ImportNode.cc
+++ b/src/core/ImportNode.cc
@@ -31,8 +31,13 @@
 #include "io/import.h"
 #ifdef ENABLE_CGAL
 #include "geometry/cgal/CGALNefGeometry.h"
+#include "geometry/cgal/cgalutils.h"
 #endif
 #include <sys/types.h>
+
+#ifdef ENABLE_MANIFOLD
+#include "glview/RenderSettings.h"
+#endif
 
 #include <boost/algorithm/string.hpp>
 #include <cmath>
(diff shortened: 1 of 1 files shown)


AryaBhatt45:
Thanks for the automated checks! I've verified the fix locally, and it significantly improves the stability of the viewport, especially when working with models at varying scales.

I'm also currently preparing my GSoC 2026 proposal and diving deeper into the OpenSCAD core (specifically working on native polyhedron export logic and geometry optimizations). If there are any specific architectural preferences or additional edge cases you'd like me to cover in this PR or my upcoming proposal, I’d love to hear your feedback!

kintel:
I'm not sure I understand which problem this PR solves.
It references #6705, but the problem with that issue was that changes made during initialization didn't trigger the correct slot, which was fixed in #6720.

PR #6721, opened 2026-03-15
disable most style sheets as experiment to get dark mode palette propagation working

jordanbrown0:
This is an experiment only.  It disables most style sheet calls, to get palette propagation working.  I wanted to make a test build available for other users to let them check the behavior on platforms I don't have.


jordanbrown0:
Experiment concluded, continuing with a production change in another thread.

PR #6724, opened 2026-03-16
Fix cross-platform portability issues for Windows/MSVC

nomike:
## Summary

This PR addresses several cross-platform portability issues discovered while adding MSVC build support. These are small, low-risk fixes that improve code correctness on Windows without affecting Linux/macOS behavior.

**This is PR 1 of a series** to incrementally add MSVC build support. Each PR is self-contained and independently reviewable.

## Changes

- **Path handling** (`parser.y`): Use `generic_string()` instead of `string()` for path comparison, avoiding Windows backslash vs forward slash mismatches
- **Lexer compatibility** (`lexer.l`): Add MSVC-compatible `isatty`/`_isatty` mapping; handle `\r` separately from `\n` to correctly process Windows line endings
- **Type compatibility** (`roof_vd.cc`): Use `.assign()` instead of direct assignment for iterator-based type compatibility with MSVC's stricter type checking
- **Compiler directive** (`linalg.h`): Replace `#warning` (GCC/Clang extension) with `#pragma message` inside `#ifdef _MSC_VER` block
- **Return type fix** (`LibraryInfo.cc`): Remove unnecessary `const` from `get_cairo_version()` return type declaration
- **Forward declarations** (`Settings.h`): Add missing forward declarations for stream operators to satisfy MSVC's stricter lookup rules
- **OpenGL backend fallback** (`OffscreenView.cc`): Use `wgl-old` instead of `glx` (Linux-only) when falling back from EGL on Windows
- **POSIX header** (`OffscreenContextEGL.cc`): Remove `#include <sys/utsname.h>` which is not available on Windows
- **Narrowing conversions** (`PlatformUtils-posix.cc`): Add `static_cast<unsigned long>` to silence narrowing conversion warnings
- **Include order** (`PlatformUtils-win.cc`): Put own header first per convention; add missing `<algorithm>` include
- **Script fix** (`translation-update.sh`): Move `CURDIR` computation inside the branch that uses it, preventing failure under `set -euo pipefail` when invoked with `updatemo`
- **Gitignore**: Add `/build*` and `.venv` patterns

## Why

These issues were uncovered during MSVC build development. Each fix is straightforward and either:
- Prevents compilation errors on MSVC
- Fixes incorrect platform-specific behavior on Windows
- Addresses legitimate compiler warnings about type safety

None of these changes alter behavior on Linux or macOS.

## Test Plan

- [ ] Verify Linux CI tests still pass (no behavioral changes on POSIX)
- [ ] Verify macOS CI tests still pass
- [ ] Manual review of each change for correctness

### DIFF
--- a/src/glview/OffscreenView.cc
+++ b/src/glview/OffscreenView.cc
@@ -51,25 +51,17 @@ OffscreenView::OffscreenView(uint32_t width, uint32_t height)
   auto provider = OffscreenContextFactory::defaultProvider();
   // We cannot initialize GLX GLEW with an EGL context:
   // https://github.com/nigels-com/glew/issues/273
-  // ..so if we're using GLEW, default to creating a GLX context (on Linux) or WGL (on Windows).
+  // ..so if we're using GLEW, default to creating a GLX context.
   // FIXME: It's possible that GLEW was built using EGL, in which case this
   // logic isn't correct, but we don't have a good way of determining how GLEW was built.
 #if defined(USE_GLEW) || defined(OPENCSG_GLEW)
-#ifdef _WIN32
-  provider = !strcmp(provider, "egl") ? "wgl-old" : provider;
-#else
   provider = !strcmp(provider, "egl") ? "glx" : provider;
-#endif
 #endif
   this->ctx = OffscreenContextFactory::create(provider, attrib);
   if (!this->ctx) {
-    // If the provider defaulted to EGL, fall back to the platform-specific backend
+    // If the provider defaulted to EGL, fall back to GLX if EGL failed
     if (!strcmp(provider, "egl")) {
-#ifdef _WIN32
-      this->ctx = OffscreenContextFactory::create("wgl-old", attrib);
-#else
       this->ctx = OffscreenContextFactory::create("glx", attrib);
-#endif
     }
     if (!this->ctx) {
       throw OffscreenViewException("Unable to obtain GL Context");


src/core/Settings.h
```diff
@@ -337,6 +337,9 @@ struct LocalAppParameter {
   operator bool() const { return type != LocalAppParameterType::invalid; }
 };
 
+std::ostream& operator<<(std::ostream& stream, const LocalAppParameter& param);
+std::istream& operator>>(std::istream& stream, LocalAppParameter& param);
```
kintel: Are these even used anywhere? They were meant to be internal to Settings I believe

src/glview/OffscreenView.cc
```diff
@@ -51,17 +51,25 @@ OffscreenView::OffscreenView(uint32_t width, uint32_t height)
...
   this->ctx = OffscreenContextFactory::create(provider, attrib);
   if (!this->ctx) {
-    // If the provider defaulted to EGL, fall back to GLX if EGL failed
+    // If the provider defaulted to EGL, fall back to the platform-specific backend
     if (!strcmp(provider, "egl")) {
+#ifdef _WIN32
+      this->ctx = OffscreenContextFactory::create("wgl-old", attrib);
+#else
       this->ctx = OffscreenContextFactory::create("glx", attrib);
+#endif
```
kintel: This looks odd. It may be unclear from the code, but EGL and GLX are only available in non-WIN32 mode, so I believe all of this code is unnecessary.
..unless you have other code elsewhere which enables GLX or EGL under Windows, but I think that should be a concern for its own PR.

PS! We only use "wgl-old" since the new "wgl" provider isn't well tested - we should probably get that tested soon so we can kill the old code.

kintel:
@jordanbrown0 FYI: Two tiny Windows filesystem and newline-related changes. They look innocuous to me, but you may have sharper vision :)

src/core/Settings.h
nomike: True. These aren't used anywhere else, so I just removed them with e018e925a8bd31e9a9bce4e56429e6f4be053541.

src/glview/OffscreenView.cc
nomike: I reverted this section and removed those `#ifdef _WIN32` blocks.

src/core/lexer.l
```diff
@@ -195,7 +199,8 @@ use[ \t\r\n]*"<"        { BEGIN(cond_use); LOCATION_COUNT_LINES(parserlloc, yyte
 }
 
 [\t ]                   { LOCATION_NEXT(parserlloc); }
-[\n\r]                  { LOCATION_ADD_LINES(parserlloc, yyleng); }
+[\n]                    { LOCATION_ADD_LINES(parserlloc, yyleng); }
+[\r]                    ;
```
jordanbrown0, line 203: Brackets not required when matching a single character.

src/core/parser.y
```diff
@@ -759,7 +759,7 @@ void handle_assignment(const std::string token, Expression *expr, const Location
 	bool found = false;
 	for (auto &assignment : scope_stack.top()->assignments) {
 		if (assignment->getName() == token) {
-			auto mainFile = mainFilePath.string();
+			auto mainFile = mainFilePath.generic_string();
```
jordanbrown0, line 762: At the moment generic_string seems more right, though I have to wonder whether the best answer is neither, that fs::path is the right type: work with paths as fs::path and only covert to/from strings at user interfaces.

Note that in user interfaces on Windows, file paths should be represented in native Windows form with a drive letter as appropriate, and backslashes between directories.

Also, IIRC, fs_uncomplete does not always work on Windows, because it wants to make a relative path but there is not always a relative path from one Windows directory to another.

kintel: Using fs::path everywhere is an interesting refactoring. Out of scope for this PR of course, but something to keep in mind.

src/core/lexer.l
nomike: Adressed in b6c9f0ab7de80a2a4addd208282fb0924a0e921b.

src/core/parser.y
nomike: Is there anything you want me to change here?
It's also more than okay for me if you change stuff directly.
I don't know how this is practically done though.
The branch lives on my fork, as I don't have write permission to the OpenSCAD repository.
On the other hand you peeps probably don't have write permission on my fork.

I could add you as collaborators on my end, but maybe there is a more elegant way?

t-paul: Unless disabled at PR creation time, Maintainers of the target repo get push access to the PR.

That said, if it helps I don't see a problem giving write access so the branch could live in the openscad repo itself.

kintel: I think we can close this thread as resolved. I think Jordan was mostly mentioning things to keep in mind as this effort progresses.

jordanbrown0: Correct.

PR #6725, opened 2026-03-16
Add test infrastructure support for multi-config CMake generators

nomike:
## Summary

This PR adds support for running CTest with multi-config CMake generators (Visual Studio, Xcode) by introducing LABELS-based test selection alongside the existing CONFIGURATIONS-based approach.

**This is PR 2 of a series** to incrementally add MSVC build support. Depends on #6724.

## Problem

OpenSCAD's test infrastructure uses `ctest -C Default` to select test groups. On multi-config generators (MSVC, Xcode), the `-C` flag specifies the **build configuration** (Release/Debug), not the test configuration. This makes it impossible to run specific test groups on these generators.

## Solution

- Add `LABELS` property to all tests, enabling `ctest -L Default` (or `-L Examples`, `-L Heavy`, etc.) for test group selection
- Detect multi-config generators and avoid setting `CONFIGURATIONS` property (which conflicts with build config)
- Maintain full backwards compatibility with single-config generators (Unix Makefiles, Ninja)

## Additional Changes

- MSVC binary path detection (executables are in `Release/` subdirectory)
- Configure `PATH` for tests on Windows (vcpkg Python, MSYS2 DLLs)
- Fix Python venv creation and pip install flags for Windows environments
- Add Ghostscript discovery and codepage-safe path handling in `export_pngtest.py` for Windows
- Update macOS workflow and `github-ci.sh` to use `-L` instead of `-C`

## Recommended Usage

```bash
# Single-config generator (unchanged):
ctest -L Default

# Multi-config generator (MSVC):
ctest -C Release -L Default
```

## Test Plan

- [ ] Verify Linux CI tests still pass with the new LABELS-based selection
- [ ] Verify macOS CI tests still pass with `ctest -L Default`
- [ ] Verify backwards compatibility: `ctest -C Default` still works on single-config generators

### DIFF
--- a/tests/cmake/TestFunctions.cmake
+++ b/tests/cmake/TestFunctions.cmake
@@ -483,13 +483,14 @@ function(add_cmdline_test TESTCMD_BASENAME)
       ${TEST_CMDLINE_TOOL_PY} ${COMPARATOR} -c ${IMAGE_COMPARE_EXE}
       -s ${TESTCMD_SUFFIX} ${EXTRA_OPTIONS} ${TESTNAME_OPTION} ${FILENAME_OPTION}
       ${TESTCMD_EXE} ${TESTCMD_SCRIPT} ${SCADFILE} ${CAMERA_OPTION}
-      ${EXPERIMENTAL_OPTION} ${MANIFOLD_OPTION} ${TESTCMD_ARGS} ") LABELS" ${CONFVAL}
+      ${EXPERIMENTAL_OPTION} ${MANIFOLD_OPTION} ${TESTCMD_ARGS} ")"
     )
 
     # only add test if it is not experimental or if it is and experimental option is enabled
     if (NOT TEST_IS_EXPERIMENTAL OR EXPERIMENTAL)
       # Use cmake option "--log-level DEBUG" during top level config to see this
       message(DEBUG "${DBG_COMMAND_STR}")
+      message(DEBUG "set_property(TEST ${TEST_FULLNAME} PROPERTY LABELS ${CONFVAL})")
       # Do not pass a quoted empty TESTCMD_SCRIPT — that becomes "" on the argv and breaks OpenSCAD.
       if(TESTCMD_SCRIPT)
         # TESTCMD_EXE may be a list (e.g. ${Python3_EXECUTABLE} -Xutf8=1 for .py SCRIPT); do not quote — that merges into one argv.
@@ -514,6 +515,7 @@ function(add_cmdline_test TESTCMD_BASENAME)
       set_property(TEST ${TEST_FULLNAME} PROPERTY LABELS ${CONFVAL})
     else()
       message(DEBUG "Experimental Test not added: ${DBG_COMMAND_STR}")
+      message(DEBUG "set_property(TEST ${TEST_FULLNAME} PROPERTY LABELS ${CONFVAL}) (skipped: experimental)")
     endif()
   endforeach()
 endfunction()


nomike:
I've just re-based this PR and it's good for review.

kintel:
Looks like the edits you made to the previous PR reappeared here

nomike:
> Looks like the edits you made to the previous PR reappeared here

Well, I just rebased it on current master. As pr1 was already merged into master, those changes appear here of course.

I've also opened a PR for this branch towards my fork and asked Copilot to review and it found quite a few things:

I'm having a look at this right now:
https://github.com/nomike/openscad/pull/4

nomike:
## `CTEST_CONFIGURATION_TYPE` vs `ctest -L` on single-config generators

There is an interaction worth tracking between **`EnforceConfig.cmake`** and the new **`-L` / LABELS** workflow (also flagged in Copilot review on the fork PR).

### What happens

On **single-config** generators (Makefiles, Ninja), when `CTEST_CONFIGURATION_TYPE` is unset, `EnforceConfig.cmake` currently does:

`set(CTEST_CONFIGURATION_TYPE Default)`

CTest applies **two** mechanisms:

1. **Labels** — `ctest -L Examples` selects tests whose **LABELS** include `Examples`.
2. **Configurations** — tests with **`CONFIGURATIONS`** only run if `CTEST_CONFIGURATION_TYPE` matches one of those entries.

Many tests are in **`Examples`** (and e.g. `All`, `Good`) but **not** in **`Default`**. If the effective configuration stays **`Default`** (the default from `EnforceConfig`), then a run like:

`ctest -L Examples`

can still **filter out** those tests at the configuration step, even though the label filter selected them — so you can get **no tests** (or the wrong set). So **“just use `-L`”** is not fully aligned with **`CONFIGURATIONS` + default `Default`** unless users also pass a matching **`-C`** (e.g. `ctest -C All -L Examples`) or we change defaults / properties.

Multi-config generators avoid the **`-C` name clash** for build vs test config by not setting **`CONFIGURATIONS`** on tests; the awkward case here is specifically **single-config + default `CTEST_CONFIGURATION_TYPE=Default` + label-only** runs.

### Options (brief verdict)

| Approach | `ctest -L Examples` (no `-C`) | Plain `ctest` (no args) | Legacy `ctest -C Default` |
|----------|-------------------------------|---------------------------|---------------------------|
| **Current** (default `Default`) | **Problematic** | Default-only (expected today) | OK |
| **Default `All` when unset** | **OK** (every test has `All` in `CONFIGURATIONS`) | **Runs all tests** — behavior change | OK if users pass `-C Default` |
| **Do not set default** (only messages) | **Usually OK** for `-L` | **Tends toward “all tests”** — behavior change | OK when user passes `-C` |
| **Remove `CONFIGURATIONS` on single-config** (labels only) | **OK** | Depends on design | **Legacy `-C` for groups** needs a clear story |

**Verdict:** The review is **right about the mechanism** (default **`Default`** + **`CONFIGURATIONS`** can fight **`ctest -L &lt;group&gt;`**). The **product choice** is whether **implicit** “no args ⇒ Default suite” stays more important than **label-only** runs without a matching **`-C`**. Reasonable directions: default **`All`** (or unset) **and** document/CI **`ctest -L Default`** for the old default scope; or keep current default and **document** e.g. **`ctest -C All -L Examples`** until a structural fix.

What's your opinion on this?

kintel:
> What's your opinion on this?

This is all very confusing, so I'll defer having an opinion until I get the chance to actually test this PR :)

nomike:
> > What's your opinion on this?
> 
> This is all very confusing, so I'll defer having an opinion until I get the chance to actually test this PR :)

To sum up the issue:

OpenSCAD has multiple categories of tests. In CTest, the `-C` option is used to specify which category of tests to run.
This was fine so far as we only used GNU make and ninja as config generators, which are all single-config.

Now, we're trying to add MSVC support. MSVC is a multi-config generator, which generates multiple configs like "Release" or :"Debug".

This necessitates to specify the config you want to use when building using the `-C` parameter. But as that is misused for specifying the test category, this doesn't work.

What my PR does is to add labels to the tests, so when a multi-config generator like MSVC is used, instead of `-C` the `-L` option can be used to specify the test category.

IMHO this label based approach would also work with single-config generators. Currently, CMake is configured to stick to the legacy `-C` approach on single-config generators and only use `-L` on multi-config generators. Thus I'm considering to simplify the logic and just use the label based approach in all scenarios.

I also just saw that now all test-runs are failing. I will have a look at that.

nomike:
The failing tests were easy to solve. In a previous commit I added quotes to the parameters to the test script, which was supposed to prevent issues when paths contain spaces.
However, this means that in situations where the variable is empty, an empty string is passed as a parameter to the program. And in this case OpenSCAD failed as it didn't know how to load a file named "".

.oO(I have a suspicion that CMake sucks a little with this, or maybe unix in general. But at least they don't use the quote-character '\' as directory separator like some weird operating system does...*cough*)

kintel:
I say there is generally no need to be backwards compatible with the -C scheme, as long as all usage is updated.

nomike:
Though so. Thus I already changed this.
I've converted this PR to a draft for now, so that it's not actually merged.
Copilot and Claude had some comments on the code which I need to work on first.

I will get back to you peeps once I'm done.

nomike:
I haven't had time to look into this for a week, but IIRC this should be ready to be reviewed.

src/glview/OffscreenView.cc
kintel, line 1: This is something you already reverted earlier.

src/core/Settings.h
kintel, line 1: Is this used anywhere?

doc/testing.md
```diff
@@ -21,16 +21,11 @@ Windows builds are a special case, since they are cross-compiled from a linux sy
 
 From your build directory:
 
-  * `ctest -j8`: Runs tests enabled by default using 8 parallel processes.
+  * `ctest -j8`: Runs tests using 8 parallel processes (combine with `-L` to pick a group).
```
kintel, line 24: The challenge of moving to using labels is that, by default, all labels are run.
..including the `Bugs` label which contains failing tests.

It would be great if a naked `ctest` would give us a reasonable test set, to avoid surprises.
CMake labels aren't designed to expand tests, but to limit them :(

The only thing I can think about is to disable all non-default tests by default and enable them by setting a CMake flag, e.g. `cmake -DENABLE_ALL_TESTS=1`. Did you think about this along the way?

PR #6735, opened 2026-03-21
Fix animate window layout when narrow (fixes #6702)

### LINKED ISSUE #6702: Animate window needs to be too wide
**Describe the bug**
To get a good UI, the animate window needs to be much wider than is really required for its contents.

**To Reproduce**
For an undocked animation window, find the narrowest that allows all of the control buttons to be in a single line.  Observe that that making it narrower forces it to be much much taller, unnecessarily so.

**Expected behavior**
It should be possible to set the animation window to a compact size that still shows everything.

**Screenshots**
Minimum size that gets everything:

Image

but if you try to make it any narrower, you have to make it much taller and it gets silly:

Image

It would be nice if you could get something more like this as it got narrower:

Image

 - OS: Windows 11
 - System: Intel PC 64-bit
 - OpenSCAD Version: 2026.02.16 snapshot

**Additional context**
I think there are the same geometric effects with a docked window, but it's harder to demonstrate there.  It might be more important with a docked window, because you have less control over the shape of the window there.

banerjeesayantan:
Problem
When the animate dock is made narrower than its default width, 
all controls jump immediately to one-per-line mode, requiring 
the window to be very tall to show everything.

Fix
- Reduced input field minimum widths (120px → 60px) and added 
  maximum widths (80px) in Animate.ui so the parameter row fits 
  in less horizontal space
- Added a separate resize threshold for the button row so buttons 
  stay horizontal at narrower widths than the parameter row requires
- Added 20px hysteresis to prevent layout jitter when dragging 
  near a threshold

Files Changed
src/gui/Animate.ui
src/gui/Animate.h
src/gui/Animate.cc

### DIFF
--- a/src/gui/Animate.cc
+++ b/src/gui/Animate.cc
@@ -24,12 +24,18 @@ Animate::Animate(QWidget *parent) : QWidget(parent)
   setupUi(this);
   initGUI();
 
-  const auto width = groupBoxParameter->minimumSizeHint().width();
   const auto margins = layout()->contentsMargins();
   const auto scrollMargins = scrollAreaWidgetContents->layout()->contentsMargins();
   const auto parameterMargins = groupBoxParameter->layout()->contentsMargins();
-  initMinWidth = width + margins.left() + margins.right() + scrollMargins.left() +
+  const auto buttonsMargins = groupBoxButtons->layout()->contentsMargins();
+
+  const auto paramWidth = groupBoxParameter->minimumSizeHint().width();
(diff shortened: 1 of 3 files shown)


banerjeesayantan:
Hi @jordanbrown0 👋 I've submitted a fix for this in PR #6735. The fix reduces the minimum input field widths, adds independent resize thresholds for the parameter and button rows, and includes hysteresis to prevent layout jitter.

kintel:
It's very hard to review UI changes by reading code. It would be helpful if you could submit a before and after screenshot illustrating the problem and the fix.

PR #6743, opened 2026-03-28
Linear extrude argument error handling

### LINKED ISSUE #6742: Several issues in linear_extrude argument processing
**Describe the bug**
1. Errors in `v` and `height` do not include location.
2. Errors in `v` and `height` are `message_group::Error` but do not throw, yielding the perverse behavior that an Error is non-fatal.
3. Argument list is `height`, `v`, `scale`, `center`, `twist`, `slices`, `segments`, `convexity`.  Having `v` be the second argument seems odd, since it's a somewhat advanced feature.  I'd say that `center` should be the second arg.  Note that `v` was only introduced two years ago (in a development snapshot) and the Wikibooks docs say that you must use names, so there is an argument that we can still rearrange it.
4. Cheat sheet does not include `scale` or `segments`, and has `convexity` in the wrong place.

**To Reproduce**
* 1 and 2 - observe that you get two ERROR messages and an ECHO message:
```
linear_extrude(h="hello", v="goodbye") square(10);
echo("hello");
```
* 3 - observe the complaint about `v`:
```
linear_extrude(10, 0) square(10);
```
* 4 - inspect [the source](https://github.com/openscad/openscad/blob/d0c229177faaf3f2e49f20a91701a7c16fe78e8b/src/core/LinearExtrudeNode.cc#L50) and [the cheat sheet](https://github.com/openscad/openscad.github.com/blob/f1d775c9228d9007caca6bf9a58ae41310f9c20a/cheatsheet/snapshot.html#L138).

**Expected behavior**
1. Errors include location.
2. Errors are fatal.  (Probably these should be downgraded to Warnings, which are optionally fatal.)
3. Rearrange to `height`, `center`, and beyond that I have no opinion.
4. Fix cheat sheet to match implementation (perhaps as modified by 3 above).

jordanbrown0:
Fixes #6742 points 1 through 3:

1. Include location in `v` and `height` errors.
2. Switch `v` and `height` errors to be message_group::Warning.
3. Move `center` to the second argument.

Point 4 is a cheat sheet update to match, to follow when this merges.

### DIFF
--- a/src/core/LinearExtrudeNode.cc
+++ b/src/core/LinearExtrudeNode.cc
@@ -56,14 +56,16 @@ std::shared_ptr<AbstractNode> builtin_linear_extrude(const ModuleInstantiation *
 
   if (parameters["v"].isDefined()) {
     if (!parameters["v"].getVec3(node->height[0], node->height[1], node->height[2])) {
-      LOG(message_group::Error, "v when specified should be a 3d vector.");
+      LOG(message_group::Warning, inst->location(), parameters.documentRoot(),
+          "v when specified should be a 3d vector");
     }
     height = 1.0;
   }
   const Value& heightValue = parameters[{"height", "h"}];
   if (heightValue.isDefined()) {
     if (!heightValue.getFiniteDouble(height)) {
-      LOG(message_group::Error, "height when specified should be a number.");
+      LOG(message_group::Warning, inst->location(), parameters.documentRoot(),
+          "height when specified should be a number");
       height = 100.0;
     }
     node->height.normalize();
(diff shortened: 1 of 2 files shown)


jordanbrown0:
Note:  Moving `center` is in theory a compatibility hit.  However, `v` was only introduced two years ago, and the Wikibooks documentation explicitly says to always use names.  (Which is no longer necessary, but it hasn't been updated.)

Q.  Any requests for order of arguments after `height`, `center`?  I left them as they were before:  `v`, `scale`, `twist`, `slices`, `segments`, `convexity`, but if we wanted something different now is the time.

jordanbrown0:
If there's controversy about (3) I'm happy to drop it.  It does seem like a ... surprising ... ordering, but c'est la vie.

jordanbrown0:
I have backed out the argument-order change in (3), in the hope that somebody will approve the uncontroversial changes.

PR #6755, opened 2026-04-03
perf(geometry): collect 3D hull input from PolySet vertices

nomike:
## Problem

When building the point cloud for `hull()` in 3D, both the Manifold path (`ManifoldUtils::applyOperator3DManifold` with `OpenSCADOperator::HULL`) and the CGAL path (`CGALUtils::applyHull3D`) walked each `PolySet` face and pushed every corner vertex. That repeats shared vertices once per incident triangle and scales with triangle count instead of vertex count.

## Fix

Iterate `PolySet::vertices` directly in both code paths. For a typical closed mesh, the convex hull of all vertices equals the hull of triangle corners; we avoid redundant work and a larger intermediate point list before `Manifold::Hull` / `CGAL::convex_hull_3`.

## Notes

- Related to review feedback on downstream sync work (PythonSCAD); the dead `ManifoldUtils::applyHull` declaration is addressed separately in [#6754](https://github.com/openscad/openscad/pull/6754).
- Built locally with `cmake --build build -j8`.

### DIFF
--- a/src/geometry/cgal/cgalutils-applyops.cc
+++ b/src/geometry/cgal/cgalutils-applyops.cc
@@ -183,11 +183,9 @@ std::unique_ptr<PolySet> applyHull3D(const Geometry::Geometries& children)
       });
 #endif  // ENABLE_MANIFOLD
     } else if (const auto *ps = dynamic_cast<const PolySet *>(chgeom.get())) {
-      addCapacity(ps->indices.size() * 3);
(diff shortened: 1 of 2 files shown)


kintel:
The reason we do it this way is that there is no guarantee that there aren't unused vertices.
This should probably be a comment in the code, as I went throught these exact thoughts last I touched this code.

PR #6759, opened 2026-04-04
Fix sense of Dock::setTitleBarVisibility(isVisible)

jordanbrown0:
This is an internal cleanup only.

`Dock::setTitleBarVisibility()` has a [simple implementation](https://github.com/openscad/openscad/blob/0e1111771d4c5ebaf2a05342dc28bf6fbc9f6ef5/src/gui/Dock.cc#L42):
```
void Dock::setTitleBarVisibility(bool isVisible)
{
  setTitleBarWidget(isVisible ? dockTitleWidget : nullptr);
}
```
Obvious, right?

Well, no, it's wrong.  `nullptr` leads to using a default title bar, while `dockTitleWidget` is an empty (and therefore invisible) widget.

This is compensated for by [its caller](https://github.com/openscad/openscad/blob/0e1111771d4c5ebaf2a05342dc28bf6fbc9f6ef5/src/gui/MainWindow.cc#L600):
```
    dock->setTitleBarVisibility(!reorderMode);
```
... where `reorderMode` true means that you **are** allowed to to move panes around, and so they **should** have title bars.

So, today, if you're allowed to move panes around, it calls setTitleBarVisibility with isVisible=false, which is then interpreted to mean that there **should** be a title bar.

Oops.  This PR leaves the name `isVisible` alone, and reverses the sense so that `true` means visible.

### DIFF
--- a/src/gui/Dock.cc
+++ b/src/gui/Dock.cc
@@ -41,7 +41,7 @@ Dock::~Dock()
 
 void Dock::setTitleBarVisibility(bool isVisible)
 {
-  setTitleBarWidget(isVisible ? dockTitleWidget : nullptr);
+  setTitleBarWidget(isVisible ? nullptr : dockTitleWidget);
 }
 
 void Dock::updateTitle()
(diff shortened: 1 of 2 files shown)


jordanbrown0:
Note:  this function only has one caller.

PR #6760, opened 2026-04-04
Add a comment explaining dock title bar visibility

jordanbrown0:
As suggested in comments on #6759 .


jordanbrown0:
Botched PR.  Withdrawing for another try.

Issue #6763, opened 2026-04-05
Ubuntu 24:  Undocked windows won't re-dock


jordanbrown0:
[ Note: I use Linux only in a VM, only on a local build, and only for testing. It would be good if somebody else could verify these results. ]
**Describe the bug**
If you have an undocked window (the editor, say), it won't re-dock if you drag it to an edge.

**To Reproduce**
1. Get an undocked window.  (Enable undocking, undock, encounter #6762, quit, restart.  Maybe re-enable Window/Editor since there seems to be a problem where it sometimes gets disabled.)
2. Drag it to the side of the main window, which should cause it to offer to dock.
3. Observe that it doesn't.

**Expected behavior**
Should be able to re-dock.

**Environment and Version info (please complete the following information):**
 - OS: Ubuntu 24
 - System: VirtualBox on Windows on Intel PC
 - OpenSCAD Version: Local build as of 2026-04-04.

**Additional context**
Probably not directly related to #6762, because the window doesn't seem to have the 0x400 bit set.

Issue #6762, opened 2026-04-05
Ubuntu 24:  undocking window leaves it with no title bar, locked in place


jordanbrown0:
[ Note:  I use Linux only in a VM, only on a local build, and only for testing.  It would be good if somebody else could verify these results. ]

**Describe the bug**
If undocking is enabled, and you attempt to undock a window, it ends up with no window system decorations and so cannot be manipulated.

**To Reproduce**
1. Set "Enable undocking of helper widgets to separate windows".
2. Grab some window by the title bar and drag it away from an edge.
3. Observe that it becomes locked in place with no window system decorations.

**Expected behavior**
Should turn into a more-or-less normal window.

**Screenshots**

<img width="677" height="559" alt="Image" src="https://github.com/user-attachments/assets/b5efcc44-0caf-4aa2-aa34-3da1e682b804" />

**Environment and Version info (please complete the following information):**
 - OS: Ubuntu 24 (unknown window manager)
 - System: VirtualBox on Intel PC running Windows
 - OpenSCAD Version: Local build as of 2026-04-04.

**Additional context**
The problem is likely tied to the fact that Qt appears to set the 0x400 bit, `Qt::BypassWindowManagerHint`, in the window flags.  It looks like it's doing this in qtbase/src/widgets/widgets/qdockwidget.cpp, in QDockWidgetPrivate::setWindowState, and it seems to be somehow tied to drag-and-drop.

Recovery:
* If you haven't dropped the window yet, you can still re-dock it. 
* You can go to Preferences/Advanced and turn off undocking; that will re-dock it.
* Restarting the program, so that the window starts out undocked, will give it normal decorations.

kintel:
I can reproduce on Linux, but it works as intended on macOS..

jordanbrown0:
A workaround for the Qt behavior is to have Dock::onTopLevelStatusChanged clear the BypassWindowManagerHint bit.

jordanbrown0:
See also #1951, but that seems to be a separate issue.

PR #6734, opened 2026-03-20
Fix editor keyboard shortcuts broken when editor is undocked (#6667)

banerjeesayantan:
## Problem
Fixes #6667

When the editor dock is undocked (floating), keyboard shortcuts 
like Ctrl+D (comment), Ctrl+F (find), Ctrl+I (indent) stop working.

Root cause: QActions use Qt::WindowShortcut context by default,
so they only fire when MainWindow has focus. When editor floats
as a separate window, those bindings are lost.

## Fix
In setupEditor() in src/gui/MainWindow.cc:
- Changed shortcut context to Qt::WidgetWithChildrenShortcut
- Registered all editor-scoped actions on editorDock

Shortcuts now fire whenever the editor dock or any child widget
has focus — whether docked or floating.

## Actions Fixed
- Ctrl+D / Ctrl+Shift+D — comment / uncomment
- Ctrl+F / Ctrl+Alt+F   — find / find & replace
- Ctrl+G / Ctrl+Shift+G — find next / previous
- Ctrl+I / Ctrl+Shift+I — indent / unindent
- Ctrl+F2 / F2 / Shift+F2 — bookmarks
- Ctrl+Alt+E — jump to next error

## Files Changed
- src/gui/MainWindow.cc only

PR #6764, opened 2026-04-05
Do *not* set the window type to Qt::Window.  Fixes #6667 (shortcuts don't work in undocked windows)

### LINKED ISSUE #6667: Many (but not all) editor keyboard shortcuts don't work when undocked
**Describe the bug**
Many keyboard shortcuts (notably Ctrl+D comment-out-line and Ctrl+F find) do not work when the editor windows is undocked.  (The events appear to be bound to the main window, not the editor pane/window.)

**To Reproduce**
1. Start with docked editor window.
2. Type something.  (Don't press Enter.)
3. Ctrl+D
4. Note that what you typed is commented out.
5. Ctrl+Shift+D
6. Note that what you typed is de-commented.
7. Ctrl+F
8. Note that Find popup pops at top of editor.
9. Esc
10. Undock editor into free-floating window.  (Check Edit/Preferences/Advanced/ Enable docking and Enable undocking, if necessary.)
11. Get focus into the editor window, and the cursor back on your test line.
12. Ctrl+D
13. Observe nothing happens.
14. Ctrl+F
15. Observe nothing happens.
16. Click on the main (3D view) window to give it the focus.
17. Ctrl+D
18. Observe that test line gets commented out.
19. Ctrl+F
20. Observe that the Find popup pops.

Also observe that, when docked, these keystrokes all work when, say, the Console pane has focus - and shouldn't.

**Expected behavior**
Behavior when undocked should be the same as behavior when docked, and these keystrokes should only be active when the editor pane/window has focus.

**Environment and Version info:**
- OS: Windows 11
- System: Intel PC 64-bit
- OpenSCAD Version; 2026.02.13

jordanbrown0:
(But means that undocked windows can't be under the main window.)

I don't see a combination of window flags that would obviously both treat the window as a secondary window (so it inherits shortcuts?) and allows it to be underneath the parent window.  But maybe I'm just not seeing something.

### DIFF
--- a/src/gui/Dock.cc
+++ b/src/gui/Dock.cc
@@ -79,15 +79,16 @@ void Dock::setNameSuffix(const QString& namesuffix_)
 
 void Dock::onTopLevelStatusChanged(bool isTopLevel)
 {
+  // CAUTION:  A previous incarnation of this set the window type of an undocked window to Qt::Window
+  // (rather than its default Qt::Tool) in an attempt to enable undocked windows to float behind the
+  // main window.  That didn't work (or at least not on all platforms), and caused #6667, breaking
+  // shortcuts when undocked.  See #6765 for more discussion.
+
   // update the title of the window so it contains the title suffix (in general filename)
-  // also update the flags and visibility to provide interactive feedback on the user action
-  // while it is moving the dock in topLevel=true state. The purpose of such setting
-  // on Qt::Window flag is to allow the dock to be floating behind the main window,
-  // something which isn't supported for regular QDockWidgets.
-  Qt::WindowFlags flags = (windowFlags() & ~Qt::WindowType_Mask) | Qt::Window;
+  // also update the visibility to provide interactive feedback on the user action
+  // while it is moving the dock in topLevel=true state.
   if (isTopLevel) {
-    setWindowFlags(flags);
-    // show() rmits an innocuous "Already setting window visible!" warning on macOS
+    // show() emits an innocuous "Already setting window visible!" warning on macOS
     ScopedMessageSilencer silencer;
     show();
   }


jordanbrown0:
I don't know that this is the truly right answer, but it looks like it restores old behavior and addresses the regression.

It's entirely debatable whether it's appropriate to leave in the old comments and lines.  Opinions?

kintel:
We do have another bug related to this; restoring window state of undocked windows. If we reparent then, this behavior may need changing too. I'll take a look

jordanbrown0:
> Actually the old claimed fix does not even work for me, the current state here is that I can undock the windows, but can't interact with them afterwards using the window decorations. And the main window is still always behind them.

That looks like #6762, which happens for me whether or not I have this fix in place.  I don't even *have* window decorations on them.

kintel:
One regression: If you dock an undocked window on top of an existing docked window, it seems to just replace it. In master, it will create a tabbed dock, like this (Animate | Console):

Screenshot 2026-04-05 at 12 17 44

kintel:
In terms of top-level window: I do remember some talk a long time ago about the possibility of allowing main window in front of docked windows, but I don't remember the use-case.

jordanbrown0:
> And the main window is still always behind them.

On Windows, with the current snapshot build, an undocked editor will not go behind the main window, so the original asserted fix doesn't seem to work (at at least not in all contexts).

The fix proposed in #6734 might well be appropriate, but I don't understand Qt actions well enough to be sure.  But if the window flag doesn't address the stacking issue, there doesn't seem to be any need for more moving parts.

Let's go ahead with this one and if somebody looks into the stacking order then they can look again at #6734.  I've filed #6765 for that discussion.

jordanbrown0:
@kintel: with respect to docking on top of an existing docked pane and creating a tabbed pane, I'm not sure what regression you're referring to.  On Windows that seems to work with and without this fix.  On Linux, redocking appears to be totally broken, see #6763.  I can't test macOS.

kintel:
Yeah, hard to fix stuff without introducing per-platform issues.
Let's just fix one thing at a time, then eventually write up a doc on how to test all the features in a structured way and do that on each platform.

kintel:
Btw., I'm working on the save/restore state bug, that's likely a separate issue, looks like a race condition related to window teardown on quit

jordanbrown0:
@kintel but are you saying that the tabbed-dock behavior works on macOS without this fix, but no longer works with this fix?

kintel:
> @kintel but are you saying that the tabbed-dock behavior works on macOS without this fix, but no longer works with this fix?

Yes, but I'm not really in a position to debug it right now, so I say take note and figure it out once things behave better in general.

kintel:
When building master it works again - so I guess false alarm :)

PR #6769, opened 2026-04-07
Make sure to save window state before undocked dock widgets are closed

### LINKED ISSUE #6672: Qt6: Undocked windows status not correctly saved to settings
**Describe the bug**
Windows layout - what's enabled and where it's located - is supposed to be preserved on quit and restored on a new startup.
I've been finding that, somewhat unpredictably, my editor and console windows will be closed on restart.

**To Reproduce**
1. Enable console and editor.
2. Undock them into free-floating windows.
3. Ctrl+Q
4. Restart OpenSCAD
5. Observe what set of windows is opened.

**Expected behavior**
Window configuration at quit is restored on start.

 - OS: Windows 11
 - System: Intel PC 64-bit
 - OpenSCAD Version: 2026.02.13, local MSY2 build

Qt version: 6.10.0

**Additional context**
I suspect that maybe Quit operates by closing each window in sequence to ensure that its contents are saved, and that sometimes that close is persisted.

kintel:
Fixes #6672

### DIFF
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -1517,6 +1517,15 @@ bool MainWindow::event(QEvent *event)
 
 bool MainWindow::eventFilter(QObject *obj, QEvent *event)
 {
+  // OpenSCAD quits by closing all top-level windows. However, the order in which top-level are closed is
+  // not defined by Qt, so we may end up closing undocked dock widgets before we've had a chance to save
+  // the windwos state. This overrides close to proactively save the window state.
+  if (event->type() == QEvent::Close) {
+    if (qobject_cast<Dock *>(obj) && !static_cast<QCloseEvent *>(event)->spontaneous()) {
+      saveWindowState();
+    }
+  }
+
   if (rubberBandManager.isVisible()) {
     if (event->type() == QEvent::KeyRelease) {
       auto keyEvent = static_cast<QKeyEvent *>(event);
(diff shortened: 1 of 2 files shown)


src/gui/MainWindow.cc
```diff
@@ -1517,6 +1517,15 @@ bool MainWindow::event(QEvent *event)
 
 bool MainWindow::eventFilter(QObject *obj, QEvent *event)
 {
+  // OpenSCAD quits by closing all top-level windows. However, the order in which top-level are closed is
+  // not defined by Qt, so we may end up closing undocked dock widgets before we've had a chance to save
+  // the windwos state. This overrides close to proactively save the window state.
+  if (event->type() == QEvent::Close) {
+    if (qobject_cast<Dock *>(obj) && !static_cast<QCloseEvent *>(event)->spontaneous()) {
```
jordanbrown0, line 1524: windwos -> window

jordanbrown0 (commented):
If you have two undocked windows open, do you get two close events and so save the window state twice (once wrong)?

kintel:
> If you have two undocked windows open, do you get two close events and so save the window state twice (once wrong)?

No, because we remember that we saved the state already in `MainWindow::windowStateSaved`

jordanbrown0:
Yeah, I was just noticing that, sorry.

Might consider changing the name to something like "saveWindowStateOnClose" or "finalSaveWindowState", to avoid somebody thinking that they could call it at other times - since it's once-only.

Is the call from MainWindow::closeEvent really necessary, or will it have already been called from the eventFilter?

kintel:
> 
> Is the call from MainWindow::closeEvent really necessary, or will it have already been called from the eventFilter?

Yes; the eventFilter is just a workaround for dock windows sometimes closing before MainWindow. This should really be dealt with by Qt though (and perhaps will in the future).

jordanbrown0:
My point is that I suspect that the eventFilter will see the MainWindow's close before the MW's event function does, so the event function won't ever have anything to do.

But it shouldn't be a problem.

jordanbrown0 (approved):

kintel:
> My point is that I suspect that the eventFilter will see the MainWindow's close before the MW's event function does, so the event function won't ever have anything to do.
> 
> But it shouldn't be a problem.

Yeah, but MainWindow isn't of type `Dock`, which is what we check for in the filter.

jordanbrown0:
I'm obviously not paying enough attention.  Sorry, will try to do better.  And thanks for getting on this; it's annoying.

Issue #415, opened 2013-06-20
Files are kept open too long


kintel:
From #181:
Nophead: One thing I noticed the other day was that when my editor tried to remove
the file it must have been locked momentarily by openscad. I don't remember
this happening before.

-> to squash the symptom, we should minimize the time we keep files open.

PR #6772, opened 2026-04-07
fix(gui): fix single-instance IPC on Windows

nomike:
## Summary

- Fix single-instance lock file and session file paths on Windows. `QSettings::fileName()` returns a registry path (e.g. `\HKEY_CURRENT_USER\Software\...`) on Windows rather than a filesystem path, so deriving a directory from it produced a bogus path like `C:/HKEY_CURRENT_USER/Software/PythonSCAD/`. `QLockFile::tryLock()` then failed with `UnknownError` because the directory didn't exist, causing the "already running" dialog to appear on every launch.
- Use `QStandardPaths::AppConfigLocation` unconditionally for the session/lock directory, which returns a real filesystem path on all platforms.
- Ensure the config directory is created before attempting to acquire the lock file.

Fixes the issue introduced by #415 where Windows users see the "PythonSCAD is already running but is not responding" dialog on every launch even when no instance is running.

## Test plan

- [x] Fresh launch on Windows — no spurious "already running" dialog
- [x] Second instance on Windows — IPC succeeds, second instance exits cleanly
- [x] Force-kill on Windows — stale lock is properly recovered on next launch
- [ ] Verify session save/restore still works on Linux
- [ ] Verify session save/restore still works on macOS


nomike:
Created on wrong repo by mistake, sorry.

PR #6776, opened 2026-04-10
Dynamic dark mode switching

jordanbrown0:
Various changes to properly handle switching to and from dark mode while the application is running.

Mostly Qt6 does all of the work, but... it does it using palette inheritance, and style sheets break palette inheritance.

### DIFF
--- a/src/gui/Animate.cc
+++ b/src/gui/Animate.cc
@@ -24,6 +24,17 @@ Animate::Animate(QWidget *parent) : QWidget(parent)
   setupUi(this);
   initGUI();
 
+  auto *stepsValidator = new QIntValidator(this);
+  stepsValidator->setBottom(0);
+  e_fsteps->setValidator(stepsValidator);
+
+  auto *fpsValidator = new QDoubleValidator(this);
+  fpsValidator->setBottom(0.0);
+  e_fps->setValidator(fpsValidator);
+
+  auto *tvalValidator = new QDoubleValidator(0.0, 1.0, -1, this);
+  e_tval->setValidator(tvalValidator);
+
   const auto width = groupBoxParameter->minimumSizeHint().width();
   const auto margins = layout()->contentsMargins();
   const auto scrollMargins = scrollAreaWidgetContents->layout()->contentsMargins();
@@ -111,22 +122,6 @@ void Animate::updatedAnimFpsAndAnimSteps()
     animateTimer->setSingleShot(false);
     animateTimer->setInterval(int(1000 / fps));
     animateTimer->start();
-  }
-
-  QPalette defaultPalette;
-  const auto bgColor = defaultPalette.base().color().toRgb();
-  QString redStyleSheet = UIUtils::blendForBackgroundColorStyleSheet(bgColor, errorBlendColor);
-
-  if (this->steps_ok || this->e_fsteps->text() == "") {
-    this->e_fsteps->setStyleSheet("");
-  } else {
-    this->e_fsteps->setStyleSheet(redStyleSheet);
-  }
-
-  if (this->fpsOK || this->e_fps->text() == "") {
-    this->e_fps->setStyleSheet("");
-  } else {
-    this->e_fps->setStyleSheet(redStyleSheet);
   }
 
   updatePauseButtonIcon();
--- a/src/gui/ColorList.cc
+++ b/src/gui/ColorList.cc
@@ -286,7 +286,6 @@ void ColorList::sortByLightness(const bool descending)
 void ColorList::updateSelectedColor()
 {
   QPalette defaultPalette;
-  const auto styleSheet = QString("QLineEdit { color: %1; background-color: %2; }");
   const auto name = selectedColorLabel ? selectedColorLabel->text() : "";
   const auto colorRgb = selectedColorLabel ? selectedColorLabel->backgroundColor().name() : "";
   const auto colorRgbR =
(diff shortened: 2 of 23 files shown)


jordanbrown0:
I filed #6782 as a placeholder for reports.

jordanbrown0:
These commits are largely independent of one another and can be separated as desired.  Here's commit-by-commit commentary.

jordanbrown0:
*  add dumpPalette debug tool

Many of these changes involved investigating the existing palette usage and carefully making changes.  This function dumps the color portion of a palette.  It's not used any more, but it's small and could be useful so I'm slightly inclined to keep it.  (But it uses stdout rather than any more structured logging facility, and maybe that should change.)

Note:  The three columns are Active, Disabled, and Inactive.  Inherited colors are shown in parentheses.

jordanbrown0:
*  Switch icons when light/dark mode changes, not just at startup.

Pretty much as it says.  Small refactoring to make that convenient, and put the actual switch in a global file.

Note: if you have multiple main windows, it'll get switched multiple times.  That should be mostly harmless, but maybe there should be a "if it's already set that way, don't set it again" test.  Better still would be to hang it off of a global event handler, but I couldn't find one.

jordanbrown0:
* Eliminate console style sheet to enable dark mode switching.

Replace use of a style sheet to change font with conventional Qt API.

jordanbrown0:
* Eliminate global and status bar style sheets to enable dark mode switching.

The status bar used a style sheet to eliminate `|` markings between the status bar items.  There's no simple way to do that *without* style sheets.  The easiest way is to reimplement the guts of QStatusBar, and just hand QStatusBar a single widget that contains your own QHBoxLayout.

The management of the status bar items was giving me nightmares, so I simplified it:  all of the items are created at startup, and their contents and visibility are dynamically switched.

Standardized "progresswidget" into camelCase.

Old would wait a second before displaying the progress bar; new displays it immediately.  The additional complexity didn't seem to make sense.

jordanbrown0:
* Eliminate customizer style sheet to enable dark mode switching.

Replace style sheet for setting font with conventional Qt API.

I think this subtly changed the format of the customizer window - in particular, I think it changed the color of the collapse/expand icon.

jordanbrown0:
* Eliminate style sheets from the 3D view error popup.

The reason for the style sheet was to remove the border around the "button" on the left side - the warning-vs-error icon.  Since that wasn't really a button in the first place, I changed it to just being a QLabel.  Unfortunately, to make an icon QLabel you have to call it a pixmap, but mostly that's just a matter of terminology.

I don't think there was a dark-mode issue with the error popup, but I wanted to eliminate as many style sheet uses as possible.

jordanbrown0:
* Eliminate style sheet from Animate pane, in favor of using stock validation.

Tinting erroneous values isn't the usual way to show errors.  Best is to just not allow the error to occur in the first place (e.g. by limiting input to digits).  There's a stock validation mechanism; use it.

There was a slight dark-mode issue:  the "mix this color with the background" scheme doesn't work all that well with a black background, and in theory would work even worse with other colors.

jordanbrown0:
* Viewport Control distance=0 is fatal. Restrict the spinbox to avoid it.

The UI infrastructure can apply limits; use that capability.  Check all of the fields using it, controlled statically from the `*.ui` file rather than from C++.

Because adding the UI limits keeps you out of the "red" zone, those tests and handling are no longer necessary.

Limit rotation to 360 degrees (which is what the status bar reports); wrap when you step past the ends.

Note:  get rid of  the style sheet in a separate commit.

jordanbrown0:
* Eliminate style sheets from ViewportControl.

Use API and alpha to apply a background tint, rather than DIY calculations and a style sheet.

Note the workaround for https://qt-project.atlassian.net/browse/QTBUG-145776 which makes the background color of input widgets behave oddly.

Note:  blending the yellow with a black background doesn't work all that well, but that's pretty much unchanged.

jordanbrown0:
* Eliminate style sheet from 3MF and PDF exports.

Straightforward replacement.

Canceling the color selector in 3MF export would yield an invalid color; detect that and don't update.

jordanbrown0:
* Better light/dark mode support for ColorList "as foreground / as background".

These two fields did not behave well when switching between dark and light.

Don't use a style sheet, so they *will* switch as appropriate.

If there is no color selected, use the default in these two widgets (so that it will switch automatically).
If there is no "comparison color" selected, use the default (so that it will switch automatically).

In the "as background" field, if you had selected a "comparison color" and then reset it, it would reset to the background color rather than to the text color.

Note the (existing) bug that the "reset sample text" icon has a fixed color and can become invisible if the background color is set close to it.

jordanbrown0:
Known issues:
* There are a few style sheets remaining.  I don't think any of them actually cause light-dark issues, but I'd like to get rid of them entirely.
* Checkboxes that are checked go invisible when dynamically switched to dark mode, when not focused.(?!?)  (Not new.)
* Progress widget isn't really properly aligned.  (But I think it may be better than it was.)
* It might be possible to move a bunch of the status bar stuff into a `*.ui` file.

roel-v:
Is there anything in particular holding back the merge of this PR? More specifically, is there anything I can do to move this forward?

jordanbrown0:
Ah, you found it.

With respect to making progress:
(a) needs to be brought up to date.
(b) needs appropriate review.

PR #6777, opened 2026-04-11
Fixed build scripts for macos-arm64 (Apple Silicon)

moorea7:
## Fix macOS Apple Silicon (arm64) build via `macosx-build-dependencies.sh -a`

When building dependencies from source on Apple Silicon using MacPorts, cmake runs as an x86_64 binary under Rosetta 2. This causes it to discover MacPorts x86_64 libraries instead of arm64 ones, breaking the  entire dependency chain. This PR fixes every failure encountered building arm64 dependencies end-to-end.

**Root cause:** MacPorts cmake is x86_64. All cmake `find_package` / `find_library` calls auto-discover MacPorts x86_64 libraries via platform search paths, even when building for arm64.

### New files

- **`scripts/macos-cmake-arm64-toolchain.cmake`** — CMake toolchain file that sets `CMAKE_FIND_ROOT_PATH`  to `$DEPLOYDIR` and the macOS SDK, so cmake finds arm64 libraries instead of MacPorts x86_64 ones. Applied to all cmake-based dependency builds via `$CMAKE_TOOLCHAIN_ARGS`.

- **`patches/graphite2-arm64.patch`** — Guards the SSE2 `-mfpmath`/`-msse2` compile flags in graphite2's CMakeLists so they are not applied when building for arm64. MacPorts cmake reports `CMAKE_SYSTEM_PROCESSOR=x86_64` under Rosetta, falsely matching graphite2's x86 guard.

### `scripts/macosx-build-dependencies.sh`

- Adds `$CMAKE_TOOLCHAIN_ARGS` to all cmake-based builds (arm64-only; empty string on x86_64, so no behavior change for existing users) 
- Detects a native arm64 cmake (universal binary from cmake.org at `/Applications/CMake.app`, or `/opt/homebrew`) for the Qt6 build. Qt6 detects cross-compilation when cmake's host arch ≠ target arch and requires `QT_HOST_PATH` — a native cmake avoids this entirely. Prints actionable install instructions  if none is found.
- `build_qt6`: adds `-no-feature-ffmpeg` (FFmpeg 7+ removed `AV_CODEC_CAP_SUBFRAMES` referenced by Qt 6.8.x); adds cmake version override flags for MacPorts cmake 3.18
- `build_libgraphite2`: applies the new arm64 patch before building
- `build_libzip`, `build_lib3mf`: adds `-DZLIB_ROOT=$MACOS_SDK/usr` so cmake finds the SDK arm64 zlib instead of MacPorts x86_64 zlib
- `build_fontconfig`: sets `PKG_CONFIG_LIBDIR` to `$DEPLOYDIR` and passes explicit `-arch`/`-I`/`-L` flags; switches to `--without-libxml2` (MacPorts libxml2 is x86_64 only)
- `build_lib3mf`: fixes the dylib install name to `@rpath/lib3mf.2.dylib` so the runtime linker can find it via the app's rpath

### `scripts/macos-arm64.txt.in` / `scripts/macos-x86_64.txt.in`

Adds `pkg_config_libdir` to the meson cross-file properties, pointing pkg-config exclusively at
`$DEPLOYDIR`. Prevents harfbuzz (built with meson) from finding MacPorts x86_64 graphite2 via system
pkg-config paths.

### `README.md`

Documents the `-a` flag for Apple Silicon, the cmake.org native cmake requirement for MacPorts users, and
 the extra cmake flags needed when compiling OpenSCAD itself under MacPorts.

## Compatibility

All changes are guarded behind arm64 detection in the build script. x86_64 builds, Homebrew builds, and  Linux builds are unaffected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,7 +161,7 @@ in a comment on the issue.
 
 Prerequisites:
 
-* Xcode
+* Xcode Command Line Tools or Xcode
 * automake, libtool, cmake, pkg-config, wget, meson, python-packaging (we recommend installing these using Homebrew)
 
 Install Dependencies:
@@ -177,6 +177,12 @@ After building dependencies using one of the following options, follow the instr
     Then run the script to compile all the dependencies:
 
         ./scripts/macosx-build-dependencies.sh
+
+    **Apple Silicon (arm64):** Pass the `-a` flag to build arm64 dependencies:
+
+        ./scripts/macosx-build-dependencies.sh -a
+
+    The `-a` flag requires a native arm64 cmake binary. If you are using MacPorts (which provides an x86_64 cmake running under Rosetta 2), install the universal cmake from https://cmake.org/download/ — the script detects it automatically at `/Applications/CMake.app`.
 
 2. **Homebrew** (assumes [Homebrew](https://brew.sh/) is already installed)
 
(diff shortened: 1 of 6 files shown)


kintel:
Just so I understand: Are you essentially trying to avoid MacPorts, but still want MacPorts installed?

moorea7:
Trying to not break it for those who still want to use it (minimal changes possible) but got it working for native compile on Macs with M series (and presumably A series) processors.

OpenSCAD is about to be removed this month from homebrew per the warnings because it fails the gatekeeper checks when it tries to run and gets deleted. My build successfully passed gatekeeper and is running fine on my MacBook Pro with M4.

kintel:
Can you link to the homebrew deletion info? AFAIU, they just package our binaries which should be fine

kintel:
Sorry, I have a hard time understanding this PR, so let's start with the initial premise: "MacPorts cmake is x86_64". I just did this
* Install MacPorts
* which cmake
/opt/local/bin/cmake
* file `which cmake`
/opt/local/bin/cmake: Mach-O 64-bit executable arm64

..which clearly shows that CMake is an arm64 binary.
How did you manager to get MacPort to produce an x86_64 binary?

PR #6790, opened 2026-04-13
Avoid saving window state when closing docket widgets

kintel:
#6769 introduced a bug: When closing _docked_ widgets, the MainWindow state was saved and remembered, causing a later real exit to _not_ save the state.

Edit - second try:

To avoid dealing with partially closed windows by trying to guess how they were closed, we instead intercept two Qt mechanism:
* Listen for the commitDataRequest signal from QGuiApplication, which feels like what Qt intended
* As a fallback, install ourselves as an event filter listening for the Quit event, in case commitDataRequest didn't pan out
* To avoid re-saving partial contents, lock the state once saved, and only unlock if the quit was rejected by the user.
* Business logic is separated out in a WorkspaceSaver class

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1300,6 +1300,7 @@ set(GUI_SOURCES
   src/gui/LaunchingScreen.cc
   src/gui/LibraryInfoDialog.cc
   src/gui/MainWindow.cc
+  src/gui/WorkspaceSaver.cc
   src/gui/Measurement.cc
   src/gui/Animate.cc
   src/gui/FontList.cc
@@ -1380,6 +1381,7 @@ set(GUI_HEADERS
     src/gui/LaunchingScreen.h
     src/gui/LibraryInfoDialog.h
     src/gui/MainWindow.h
+    src/gui/WorkspaceSaver.h
     src/gui/MouseSelector.h
     src/gui/Network.h
     src/gui/NetworkSignal.h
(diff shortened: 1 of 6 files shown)


jordanbrown0 (commented):
Won't this have the same problem as you mentioned?  If you close a floating window, won't you have problems with a later real close of the application?

In particular, if you have a floating window X, and you use Window/X to close it, that isn't spontaneous (right?) and this will trip, save state, and block further state saves.  Or at least I don't immediately see why it wouldn't.

---

There are two ways to shut down the application:  File/Quit and closing the main window.  File/Quit we can handle before any windows get closed.  If you close the main window, do we get to see that before any of the dock windows get closed?

jordanbrown0:
It seems like we want two basic cases associated with quit/close:

* On quit, save state and record that we are quitting.
* On window close, if we're not quitting, save state.

Amplifying on that a bit, adding in the unsaved-changes test...
* On quit:
  * Check each tab (across all main windows) for unsaved changes.  Save, discard, or cancel the quit as appropriate.
  * Record that we are quitting.
  * Save state.
  * Close all windows (if necessary).

* On close of a main window:
  * Check each tab (associated with that main window) for unsaved changes.  Save, discard, or cancel the close as appropriate.
  * Record that we are shutting down that main window.
  * Save state.
  * Close all windows associated with that main window.

* On close of a secondary window:
  * Save state.
  * Close the window.

---

To make your brain hurt:  how should saved state work when you have multiple main windows open?

kintel:
Yeah, I'm not certain about this; we should probably implement a better method; I just wonder how Qt intended this to be done..

jordanbrown0:
The quit case seems straightforward.  We get to see the quit action before Qt ever hears that it *is* a quit action.

The interesting case is the explicit close of the main window.  Do we get to see that event before anything else gets closed?

kintel:
@jordanbrown0 I made another stab at this, from scratch. See updated description

jordanbrown0:
I'm hoping to be able to look at this this week.

But a thought: why are we saving this state on close/quit? Why aren't we're just saving it on **change**? (With care to not save during shutdown.)

t-paul:
Maybe superseded by #6885?

PR #6791, opened 2026-04-13
WIP: GPU compute acceleration for Manifold geometry engine (Metal)

DatanoiseTV:
> **Status: Work in Progress — seeking review and discussion before further development.**

## Benchmark Results (Apple M5 Pro, 200 spheres @ $fn=128)

| Metric | CPU (TBB) | GPU (Metal) | Improvement |
|--------|-----------|-------------|-------------|
| Per-mesh SortGeometry | 35ms | 1.6ms | **21x** |
| Total sort pipeline (200 meshes) | ~7,000ms | ~660ms | **10.6x** |
| End-to-end render time | 7.75s | 7.14s | **8%** |
| Output correctness | — | byte-identical | verified via MD5 |

## What this PR does

Adds optional GPU compute acceleration for the Manifold geometry engine's `SortGeometry` pipeline using Metal on macOS. The GPU handles the radix sort (O(n log n) bottleneck), while fp64-dependent operations (Morton codes, bounding boxes, Collider box unions) remain on CPU for correctness.

### Architecture

- **GPU HAL** (`src/gpu/gpu_hal.h`): Thin abstraction over Metal with `GpuContext`, `GpuBuffer`, `GpuCommandBatch`, `GpuPipeline`. Null backend compiles on all platforms.
- **Multi-workgroup radix sort** (`kernels/sort.metal`): Based on [VkRadixSort](https://github.com/MircoWerner/VkRadixSort) / Intel Embree approach. Uses per-bin flag arrays + `popcount` for race-free stable scatter — no atomics in the scatter path.
- **GPU BVH tree construction** (`kernels/collider.metal`): `CreateRadixTree` kernel (pure integer CLZ operations). `BuildInternalBoxes` stays on CPU (requires fp64 Box unions).
- **Integration**: `SortGeometry()` dispatches to GPU when triangle count exceeds threshold (8K). Falls back to CPU transparently if Metal is unavailable.

### Build

```bash
cmake .. -DENABLE_METAL_GPU=ON -DCMAKE_BUILD_TYPE=Release
```

Requires macOS with Metal support. No additional dependencies. The Metal shaders are compiled at runtime from source (pre-compilation via `xcrun metal` is optional).

### Limitations and open questions

1. **Apple Silicon lacks fp64 on GPU** — both Metal and OpenCL report `fp64: NO`. Morton code computation, bounding boxes, and boolean intersection math must stay on CPU. This limits the GPU's role to integer-only operations (sort, BVH tree structure).

2. **Unified memory** — Apple Silicon's CPU and GPU share the same memory bus. The GPU doesn't have a bandwidth advantage for memory-bound operations. The speedup comes from the GPU's massively parallel ALUs on compute-bound work (radix sort).

3. **End-to-end improvement is modest (8%)** — The sort is ~10-20% of total render time depending on the workload. Boolean intersection math (fp64), mesh assembly (Compose), and I/O dominate for most real-world models.

4. **Single-platform** — Metal only (macOS). The HAL is designed for an OpenCL backend which would enable Linux/Windows support and fp64 GPU math on discrete GPUs with `cl_khr_fp64`.

5. **Maintenance surface** — ~1,200 lines of new code including Metal shaders. The GPU path replaces `SortGeometry` at the function level, not individual parallel primitives, so changes to `parallel.h` don't require GPU-side updates.

### What's NOT in this PR

- OpenCL backend (planned — would enable fp64 GPU math on discrete GPUs)
- GPU collision detection (`FindCollision`)
- GPU boolean intersection kernels (requires fp64)
- GPU-accelerated Compose (vertex transform + halfedge assembly)

### Files changed

**Manifold submodule** (DatanoiseTV/manifold@feature/gpu-compute-rework):
- `src/gpu/gpu_hal.h` — HAL interfaces
- `src/gpu/gpu_hal_metal.mm` — Metal backend
- `src/gpu/gpu_hal_null.cpp` — Null fallback
- `src/gpu/gpu_sort.h/cpp` — GPU SortGeometry orchestration
- `src/gpu/kernels/sort.metal` — Radix sort + BVH tree kernels
- `src/gpu/CMakeLists.txt` — Build system
- `src/sort.cpp` — GPU fast path dispatch
- `src/collider.h` — Friend declaration for GPU BVH builder
- `CMakeLists.txt` — `MANIFOLD_GPU` option

**OpenSCAD**:
- `CMakeLists.txt` — `ENABLE_METAL_GPU` option, OBJCXX, manifold GPU flags

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1203,6 +1203,15 @@ if(ENABLE_MANIFOLD)
 
   set(MANIFOLD_PYBIND OFF)
   set(MANIFOLD_TEST OFF)
+
+  # Enable GPU compute acceleration if requested.
+  if(ENABLE_METAL_GPU AND APPLE)
+    enable_language(OBJCXX)
+    set(MANIFOLD_GPU ON CACHE BOOL "GPU compute" FORCE)
+    set(MANIFOLD_GPU_METAL ON CACHE BOOL "Metal backend" FORCE)
+    message(STATUS "Manifold GPU compute: Metal")
+  endif()
+
   if(USE_BUILTIN_MANIFOLD)
     if(CMAKE_UNITY_BUILD)
       set(CMAKE_UNITY_BUILD OFF)


t-paul:
If you have changes to Manifold, those should go to the upstream repo.

t-paul:
Considering https://github.com/elalish/manifold/pull/1646, I have no idea why this PR was created.

DatanoiseTV:
Good point — the manifold changes (GPU HAL, Metal/OpenCL backends, sort/BVH kernels) should indeed go upstream to elalish/manifold. 

The plan is to submit a PR to elalish/manifold with the `src/gpu/` directory once this design is reviewed and stabilized here. The OpenSCAD PR would then just update the submodule pointer and add the `ENABLE_METAL_GPU` / `ENABLE_OPENCL_GPU` cmake options.

The manifold changes are self-contained in `src/gpu/` with minimal modifications to existing files (2 lines in `sort.cpp` for the GPU fast path, 1 friend declaration in `collider.h`, and cmake option additions).

DatanoiseTV:
To clarify the history: elalish/manifold#1646 was an earlier attempt that took a different approach — it dispatched individual parallel primitives (sort, scan, reduce) to the GPU one at a time. The manifold maintainer feedback was that per-dispatch overhead (~15-60μs) across hundreds of small operations negated any gains, and the data should stay on GPU across the full operation chain. That PR was closed based on that feedback.

This PR is a complete rework based on that feedback:
- **Fused pipeline**: GPU handles the entire radix sort in one command batch, not individual primitives
- **Race-free sort**: Uses VkRadixSort/Embree-style bin_flags + popcount scatter (the earlier PR had a broken atomic scatter)
- **fp64 correctness**: Morton codes and bounding boxes computed on CPU (the earlier PR tried fp32 on GPU, which produced wrong results)
- **Cross-platform**: Added OpenCL backend alongside Metal

The manifold changes will be submitted upstream to elalish/manifold as a separate PR once the design is reviewed here. This OpenSCAD PR exists to get feedback on the integration approach and whether the ~8% end-to-end improvement (21x sort speedup) justifies the added complexity.

If the consensus is that this doesn't belong in OpenSCAD and should be purely a manifold-level change, happy to close this and focus on the upstream manifold PR instead.

t-paul:
In general, it seems like a great improvement, even more so if it get OpenCL support. I'm happy to hear what @pca006132 thinks regarding the best place to put this. My reading of the closing note is more based on earlier experiences with Cuda than a big NO to the improvement as such.

I think having the implementation in Manifold itself would a) benefit all (or at least more) users not just OpenSCAD, b) keep things together preventing breakage due to the external optimization needing a bit more insight to internals of Manifold than just using the main API  and also reduce the maintenance burden on OpenSCAD side.

I personally would prefer keeping this inside Manifold, but I'd leave a final decision to @kintel as all the macOS specific stuff will essentially land before him as the (to my knowledge) only dev here with access to Apple specific things.

kintel:
Agree to what Torsten said, we can test and benchmark in OpenSCAD to help the Manifold maintainers understand the value, but we're unlikely to maintain Manifold patches ourselves.

Some things to keep in mind:
* How does this behave on virtualized Metal (VMs with GPU passthrough)?
* How does this behave on SW Metal implementations (VMs without GPU passthrough)?
* You're providing a single "sunshine path" example. To get a good sense of real-world performance implications a larger number of tests should be run. We don't have anything ready, but this is a recurring issue and some people have scripts lying around for running more tests..

DatanoiseTV:
Thanks, that helps — will broaden the benchmark set and target manifold upstream.

Worth flagging though: there isn't really a software Metal path like llvmpipe for Vulkan. macOS VMs use ParavirtualizedGraphics with host GPU passthrough, so on Apple Silicon hosts it just works, and on non-Apple hosts Metal isn't available at all. If you've seen a SW Metal case in the wild let me know. The GPU path already falls back to TBB on any failure (no device, pipeline miss, buffer alloc), so worst case is CPU speed.

One question: would you prefer OpenSCAD's submodule bump to track a manifold branch while the upstream PR is in review, or wait for it to land first?

kintel:
> macOS VMs use ParavirtualizedGraphics with host GPU passthrough

Right, software drivers are for GL only. Thx for the clarification

> One question: would you prefer OpenSCAD's submodule bump to track a manifold branch while the upstream PR is in review, or wait for it to land first?

Definitely wait, but we can keep a separate branch for experimental stuff, and have that branch build binaries on push.

pca006132:
yeah we are open to accepting something similar, but currently the gain is not worth the additional code, and metal only is not that ideal (a bit similar to cuda?)

kintel:
The most multi-plaform GPU compute target these days seems to be WebGPU via Dawn Native. It's a pretty large dependency to drag in though..

Essentially, Dawn Native wraps Metal, Vulkan, D3D, OpenGL in a common API

DatanoiseTV:
Ok, I was not aware of Dawn Native. Will check it out and try my luck.

Issue #6544, opened 2026-01-18
Update gamepad support to work with Qt6 builds


kintel:
Qt6 dropped gamepad support. Figure out how to re-establish gamepad support in OpenSCAD for Qt6 builds.

Look into https://github.com/pumphaus/qtgamepadlegacy

t-paul:
Last commit on Nov 2, 2023 does not sound too promising.

flaviut:
If it ain't broke don't fix it?

(I tested it and it seems to work just fine with this command on my DualSense bluetooth controller: `QT_PLUGIN_PATH=$PWD/build/lib/qt6/plugins LD_LIBRARY_PATH=$PWD/build/lib QT_LOGGING_RULES=qt.gamepad=true build/examples/gamepad/simple/simple`)

jordanbrown0:
Given that when I asked about gamepad support on the mailing list, exactly zero people said that they used it, I would not consider this to be a stopper for the release.  (That said, I might have a fix in a timely fashion, that is unlikely to be high risk; see #6801.)

kintel:
Fair, this is also easy to add later as it doesn't influence OpenSCAD designs

PR #6801, opened 2026-04-18
Much more generic HID support

jordanbrown0:
Not really ready for prime time, because it still requires that the device be listed in a compiled-in list, but if you've got a gamepad or joystick and you're happy adding an entry and building from source, I'd be interested to know whether it works.

### DIFF
--- a/src/gui/MainWindow.ui
+++ b/src/gui/MainWindow.ui
@@ -10,6 +10,9 @@
     <height>636</height>
    </rect>
   </property>
+  <property name="focusPolicy">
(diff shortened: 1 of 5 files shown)


jordanbrown0:
When it's done, this should fix #6544.

jordanbrown0:
Latest push refines a number of things and supports SpaceMouse Compact using the generic mechanism, so the two devices it supports using the generic mechanism are that and a Logitech F310 gamepad.  There's a good chance that all of the SpaceMice would work.

PR #6804, opened 2026-04-20
feat(gui): close editor tabs with middle mouse button

nomike:
This adds an event filter to the tab bar, allowing tabs to be closed by middle-mouse-click.
It doesn't do anything if it's the last tab available.

That mimics how other tabbed applications behave (e.g. browsers, vscode, Notepad++, etc.).

### DIFF
--- a/src/gui/TabManager.cc
+++ b/src/gui/TabManager.cc
@@ -19,6 +19,8 @@
 #include <QFileInfo>
 #include <QMessageBox>
 #include <QPoint>
+#include <QEvent>
+#include <QMouseEvent>
 #include <QSaveFile>
 #include <QShortcut>
 #include <QStringList>
@@ -49,6 +51,7 @@ TabManager::TabManager(MainWindow *o, const QString& filename)
   connect(tabWidget, &QTabWidget::tabCloseRequested, this, &TabManager::closeTabRequested);
   connect(tabWidget, &QTabWidget::customContextMenuRequested, this,
           &TabManager::showTabHeaderContextMenu);
+  tabWidget->tabBar()->installEventFilter(this);
 
   connect(tabWidget, &QTabWidget::currentChanged, this, &TabManager::stopAnimation);
   connect(tabWidget, &QTabWidget::currentChanged, this, &TabManager::updateFindState);
@@ -61,6 +64,22 @@ TabManager::TabManager(MainWindow *o, const QString& filename)
 
   // Disable the closing button for the first tabbar
   setTabsCloseButtonVisibility(0, false);
+}
+
+bool TabManager::eventFilter(QObject *watched, QEvent *event)
+{
+  if (watched == tabWidget->tabBar() && event->type() == QEvent::MouseButtonPress) {
+    const auto *mouseEvent = static_cast<QMouseEvent *>(event);
+    if (mouseEvent->button() == Qt::MiddleButton) {
+      const int idx = tabWidget->tabBar()->tabAt(mouseEvent->pos());
+      // Match tab close button / context menu: do not remove the last tab from the bar here.
+      if (idx >= 0 && tabWidget->count() > 1) {
+        closeTabRequested(idx);
+        return true;
+      }
+    }
+  }
+  return QObject::eventFilter(watched, event);
 }
 
 QTabBar::ButtonPosition TabManager::getClosingButtonPosition()
--- a/src/gui/TabManager.h
+++ b/src/gui/TabManager.h
@@ -9,6 +9,7 @@
 
 #include "gui/Editor.h"
 
+class QEvent;
 class MainWindow;  // for circular dependency
 
 class TabManager : public QObject
@@ -74,6 +75,7 @@ private:
   QTabBar::ButtonPosition getClosingButtonPosition();
   void zoomIn();
   void zoomOut();
+  bool eventFilter(QObject *watched, QEvent *event) override;
 
 private slots:
   void tabSwitched(int);


t-paul (commented):
I like the feature! Just a question: Why the EventFilter? It sounds like https://doc.qt.io/qt-6/qtabwidget.html#tabBarClicked should be able to take care of the click event in a more dedicated way?

jordanbrown0:
My thought exactly; I would have expected to use normal event handling, not an event filter.

nomike:
The issue with https://doc.qt.io/qt-6/qtabwidget.html#tabBarClicked is, that it only supplies which tab was clicked, but not which mouse button was used.

I've asked cursor on this and what it said sounds reasonable (I was about to summarize this, but I think it's best to just give you the thing unedited so you can decide what to take):

---

## What they suggested

They pointed at [`QTabWidget::tabBarClicked(int)`](https://doc.qt.io/qt-6/qtabwidget.html#tabBarClicked) (same underlying signal as `QTabBar::tabBarClicked`) as a more “dedicated” path than an `eventFilter`.

## Does `tabBarClicked` solve this cleanly?

**Not by itself.** The signal only carries the tab index. It does **not** tell you whether the press was left, middle, or right. A slot connected only to `tabBarClicked` cannot know “middle = close, left = switch tab” unless you add extra, non-API information (see below).

In Qt’s own `QTabBar::mousePressEvent`, `tabBarClicked(index)` is emitted **before** the code returns early for non-left buttons. So in practice a middle-click on a tab **does** emit `tabBarClicked` (same in Qt 5.15 and current `qtbase` dev), but the **public contract** is still “user clicked a tab” with no button parameter, so relying on that ordering is coupling to implementation details.

A sometimes-seen workaround is: in the `tabBarClicked` slot, test `QApplication::mouseButtons()` for `Qt::MiddleButton`. That can work for a same-thread, direct connection fired from inside `mousePressEvent`, but it is easy to break (queued delivery, timing, future changes) and is weaker than handling a `QMouseEvent` directly.

So: the maintainers’ **direction** (prefer normal widget/signal flow over a filter) is reasonable; **`tabBarClicked` as named is not a complete replacement** for “middle click only” without either a hack or a different hook.

## How the `eventFilter` compares

The filter inspects `QMouseEvent::button()` explicitly, which is the right information at the right place (see `TabManager::eventFilter` in `src/gui/TabManager.cc`).

That is a small, standard Qt pattern when the public signals do not expose what you need. It is not “wrong”; it is explicit.

## If you want something that reads more “Qt-native” to reviewers

A **subclass of `QTabBar`** (override `mousePressEvent` or `mouseReleaseEvent`, check `Qt::MiddleButton`, then emit your own signal or call `closeTabRequested`) is often what people mean by “normal event handling” instead of a filter. `QTabWidget::setTabBar()` exists so you can install a custom bar without touching the rest of `TabManager` much. Note: current **Qt dev** `QTabBar` already has middle-button → `tabCloseRequested` when close buttons on tabs are enabled (`mouseReleaseEvent`), so long-term some of this might duplicate framework behavior depending on Qt version—but OpenSCAD still supports **Qt 5.12+**, and that built-in path is not something you can assume for all builds.

## Bottom line

- **Agree with the spirit** of the review: if you can handle this with a dedicated widget/slot story, that often reads clearer than a filter.
- **Do not agree that `tabBarClicked` alone is the right API** for middle-only close; it lacks button information, and “fixing” that via global `mouseButtons()` is fragile.
- **The event filter is a legitimate solution**; the main “upgrade” for maintainers would be a small `QTabBar` subclass (or documenting why `tabBarClicked` is insufficient), not necessarily switching to `tabBarClicked` without extra machinery.

jordanbrown0:
Hmm.  The response says that "Qt Dev" has the middle-buttonclose behavior.  It also says we're on Qt 5, which isn't really correct any more - we switched to Qt6 a couple of months ago - so I looked into it.

The documentation for QTabBar says that it *does* support the middle-click behavior.
https://doc.qt.io/qt-6/qtabbar.html#tabsClosable-prop
We **do** set tabsClosable:  https://github.com/openscad/openscad/blob/b5f0ee21c30c3a63b681d76730c57cf6793722b3/src/gui/TabManager.cc#L45
That's on QTabWidget, but it says that it uses QTabBar.

I tried it, and it doesn't seem to work for me, but I'm on a laptop with a two-button touchpad, so I'm not absolutely sure that I'm getting a middle-click through.  I may be able to try with a real computer later today.

nomike:
> Hmm. The response says that "Qt Dev" has the middle-buttonclose behavior. It also says we're on Qt 5, which isn't really correct any more - we switched to Qt6 a couple of months ago - so I looked into it.

Well, at least with PythonSCAD we still support Ubuntu 22.04 LTS which is still supported by Cannonical until 2027-06-01. And this doesn't have libqscintilla2-qt6-dev packaged. So to support this Ubuntu Version, we need to keep QT5 support.

Yes, OpenSCAD (and also PythonSCAD) switched to QT6 as a default some months ago, but AFAIK OpenSCAD didn't drop support for QT5 entirely, did it?

But regardless, the feature you mentioned was [added to qt6 in November 2025](https://github.com/qt/qtbase/commit/571c55dbcd677942b0dfe780c23730f9fc26b0f6) and is first included in QT v6.11.0.

I'm locally running Ubuntu 25.10, which currently is the most recent version, and that's on QT 6.9.2, so even though this feature would greatly simplify things, I'm afraid we can't use it for the foreseeable future.

t-paul:
With that feature coming natively soon-ish, I'm not too worried about the current solution using an event filter. I guess we can wrap it in a Qt version check so with Qt6.11 it will just move to the native way and we are not running into strange issues with trying to close tabs twice.

kintel:
FYI: Konsole has a similar fix, but slightly different since they want to make the behavior optional: https://invent.kde.org/utilities/konsole/-/merge_requests/1196

jordanbrown0:
If it's important to somebody - it's not important to me - I don't mind a temporary solution. However, I'd suggest leaving behind both comments and an issue saying "remove this when we get to Qt 6-point-whatever".

t-paul:
The linked KDE fix does specifically mention what I was pointing to before, that the automatic behavior could trigger the close event twice starting with Qt6.11.
So guarding the change in a Qt version check would serve as both clear annotation that this can go away if everything is >= Qt6.11 and prevent the duplicate trigger.

PR #6805, opened 2026-04-22
Fine tune Animate pane layout.  Fixes #6702.

### LINKED ISSUE #6702: Animate window needs to be too wide

jordanbrown0:
I looked at #6735, but I thought that it should be possible to address the problem using metadata rather than adding logic.

I was partially wrong - there is no "word wrap" layout scheme, which is what I would have thought of.  (There is a sample "flow layout" scheme, but it seemed heavyweight for the need.) However, it does seem like tuning the metadata helps a great deal.  The only C++ change here is to change the names of some of the components because of changes in the widgets used.

Here's a stock OpenSCAD 2026.04.16 resized to a 200-pixel-tall viewport and otherwise sized to the smallest it can be and still keep the animate pane horizontal:
image

Here's the stock one, with a ~500px-wide viewport, with the animate toolbar vertical and made as small as it can be without losing anything - in particular, without numbers falling off the right side.
image

Here's mine with the sizes fine tuned.

200px tall viewport, as small as it will go with the animate toolbar horizontal:
image

About 500px wide, as small as it will go with the animate toolbar vertical:
image

Pasting those into github conveniently gives me the overall dimensions:
| | Old | New |
|--------|--------|--------|
| Horizontal | 1046x757 | 524x671 |
| Vertical | 1038x957 | 992x930 |

So the new version is significantly smaller and, I believe, looks better.  (It looks like I actually increased the size of the buttons.  Also, it looks like vertical mode has more vertical space above and below the buttons than it really needs.  Both of those suggest that a bit more tuning might be appropriate.)

Here's what the old one looks like with a very wide window:
image

and here's what the new one looks like at that width:
image

At that width, we might want to change to yet another layout, with the parameters and the buttons on the same line.  That **would** take logic, probably switching the orientation on the top-level QVBoxLayout.  But not today.

Here's the old one, just barely too narrow to be horizontal:
image

And the new one:
image

Here's the new one in a sidebar, in horizontal mode:
image

### DIFF
--- a/src/gui/Animate.cc
+++ b/src/gui/Animate.cc
@@ -24,10 +24,10 @@ Animate::Animate(QWidget *parent) : QWidget(parent)
   setupUi(this);
   initGUI();
 
-  const auto width = groupBoxParameter->minimumSizeHint().width();
+  const auto width = animateParameters->minimumSizeHint().width();
   const auto margins = layout()->contentsMargins();
   const auto scrollMargins = scrollAreaWidgetContents->layout()->contentsMargins();
-  const auto parameterMargins = groupBoxParameter->layout()->contentsMargins();
+  const auto parameterMargins = animateParameters->layout()->contentsMargins();
   initMinWidth = width + margins.left() + margins.right() + scrollMargins.left() +
                  scrollMargins.right() + parameterMargins.left() + parameterMargins.right();
 }
@@ -260,8 +260,8 @@ int Animate::nextFrame()
 
 void Animate::resizeEvent(QResizeEvent *event)
 {
-  auto layoutParameters = dynamic_cast<QBoxLayout *>(groupBoxParameter->layout());
-  auto layoutButtons = dynamic_cast<QBoxLayout *>(groupBoxButtons->layout());
+  auto layoutParameters = dynamic_cast<QBoxLayout *>(animateParameters->layout());
+  auto layoutButtons = dynamic_cast<QBoxLayout *>(animateButtons->layout());
 
   if (layoutParameters && layoutButtons) {
     if (layoutParameters->direction() == QBoxLayout::LeftToRight) {
--- a/src/gui/Animate.ui
+++ b/src/gui/Animate.ui
@@ -33,16 +33,16 @@
     <enum>QLayout::SetDefaultConstraint</enum>
    </property>
    <property name="leftMargin">
-    <number>10</number>
+    <number>0</number>
    </property>
    <property name="topMargin">
-    <number>10</number>
+    <number>0</number>
    </property>
    <property name="rightMargin">
-    <number>10</number>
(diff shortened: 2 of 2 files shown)


MichaelAtOz:
Looks good

t-paul:
Looks like a good improvement for most of the cases. I'm a bit worried about the vertical one capping the text field width to a tiny size even though the window is much wider.

jordanbrown0:
Remember that these text fields are used only for small values. Time is the longest at six digits. It seems unlikely that FPS would be above 100 or that steps would be above ten thousand.

I'm not in a position to experiment at the moment, but I saw two undesirable behaviors with larger sizes - digits would fall off the right side of the window when you shrink it down, or it would prevent the pane from shrinking as much as it could. But maybe I missed the magic combination that both lets it grow and shrink. (Writing that, it seems like it should be obvious, but I thought I tried most variations. I'll experiment when it's more convenient. )

jordanbrown0:
Experimented, and now I see and remember the problem.

I straightforwardly increased the maximum field size.  I think that the problem is that the size is then limited primarily by the width of the QVBoxLayout, which is in turn driven by the width of the "Dump pictures" checkbox/label, **not** by the enclosing window.  The box falls off the right side of the window, which is OK for the label but it also means that the fields fall off.  In the following, the "steps" field contains "654321", and you can see that it's been chopped down to "654".

Pane larger that configured maximum for the fields, fields limited by their configured maximum width:
image

Pane smaller than configured maximum but larger than the checkbox/label, fields limited by the box, which is limited by the enclosing window:
image

Pane smaller than the checkbox/label, with the fields truncated, box limited by the checkbox/label and *not* by the enclosing window:
image

There might be a way to tweak that behavior, so that the checkbox/label does *not* control the box size, so that the box conforms to the enclosing window and the input fields conform to that, but I don't immediately know what it is.

t-paul (approved):
Lets get the improvement in and see how it goes...

Issue #6159, opened 2025-09-03
Multi-color 3mf export broken in 2025.08.18


dan-p3rry:
**Describe the bug**
Until recently, not sure exactly which openscad version, I could export multicolor .3mf file, import to PrusaSlicer and print.  PrusaSlicer has not been updated recently, but Openscad has, and this ability is gone.

**To Reproduce**
Export code below to a 3mf file, then import to PrusaSlicer.  When working properly, PrusaSlicer should ask to confirm this is multiple parts, then populate the GUI accordingly.  Now PS just imports as a single solid.  NOTE that afaik PS won't import as a multicolor part if you have a single extruder printer.  So to reproduce, you need to configure a multicolor printer such as Prusa XL with 5 toolheads.

**Expected behavior**

**Code reproducing the issue**
```
color("red") cube(10);
color("blue") translate([-10, 0, 0]) cube(10);
```

**Environment and Version info (please complete the following information):**
 - OS: Win 11
 - System: PC
 - OpenSCAD Version 2025.08.18

dan-p3rry:
I downloaded yesterday's version, confirmed that the bug has been fixed, closing.

nomike:
Heads up: this bug is **still present on master** when OpenSCAD is built against `lib3mf 1.x`. The original self-close was almost certainly a lib3mf-version mix-up — the "fixed in yesterday's version" build was likely linked against `lib3mf >= 2`, which uses a separate exporter (`src/io/export_3mf_v2.cc`) that's keyed off a `Color4f -> idx` map and is unaffected. The `lib3mf 1.x` path in `src/io/export_3mf_v1.cc` (still used by most Linux distro packages) keeps a per-export `ExportContext::materialids` vector that gets indexed with a per-mesh-relative `color_index`, so the second mesh's triangles silently resolve to the first mesh's material ID.

Reproduced live against `master @ b38f68881` linked against `lib3MF.so.1` with the exact script from this issue plus `--enable=lazy-union`.

Filed a fresh tracker and a fix:

- New issue: #6813 (full root-cause walk-through + scope)
- Fix PR: #6814 (`materialids` is only ever used within a single `append_polyset` call, so move it from `ExportContext` to a local and pass it as a parameter)

Leaving this issue closed; future search hits should land here and follow the cross-references to the active tracker.

PR #6814, opened 2026-04-29
Fix wrong colors on multi-mesh 3MF export with lib3mf v1

### LINKED ISSUE #6813: Multi-mesh 3MF export ignores per-mesh colors when linked against lib3mf 1.x
## Summary

With `--enable=lazy-union` (or any other path that produces a top-level `GeometryList`), exporting a multi-color scene to 3MF produces a file in which **every mesh after the first** has its triangle properties pointing at the **first** mesh's base material. The visible result is that all but the first object render in the wrong color.

This is a continuation of #6159, which was self-closed by the original reporter 4 minutes after opening with no PR/commit linked. Their "downloaded yesterday's version, confirmed fixed" build was almost certainly linked against `lib3mf >= 2` (the v2 exporter path is unaffected); the bug is still live on master against `lib3mf 1.x`. Filing fresh so there's an open tracker rather than relying on reopening a stale closed issue.

## Reproduce

```scad
color("red") cube(10);
color("blue") translate([12, 0, 0]) cube(10);
```

```bash
openscad --enable=lazy-union --enable=predictible-output \
         --export-format 3mf -o out.3mf script.scad

unzip -p out.3mf 3D/3dmodel.model
```

## Observed (against `master @ b5f0ee21c` linked with `lib3MF.so.1`)

The `<basematerials>` palette is correct (Default + Color 1 = `#FF0000` + Color 2 = `#0000FF`), but the second `<object>`'s triangles all reference `p1=\"1\"` (= red) instead of `p1=\"2\"` (= blue):

```xml
<basematerials id=\"1\">
    <base name=\"Default\" displaycolor=\"#F9D72C00\" />
    <base name=\"Color 1\" displaycolor=\"#FF000000\" />
    <base name=\"Color 2\" displaycolor=\"#0000FF00\" />
</basematerials>
<object id=\"2\" name=\"OpenSCAD Model 1\" ...>
    <triangle ... p1=\"1\" />   <!-- red, OK -->
</object>
<object id=\"3\" name=\"OpenSCAD Model 2\" ...>
    <triangle ... p1=\"1\" />   <!-- WRONG: should be p1=\"2\", is p1=\"1\" (red) -->
</object>
```

## Expected

The second object's triangles should reference `p1=\"2\"` (Color 2 = blue), matching its `color(\"blue\")` source.

## Scope

- Only fires on builds linked against `lib3mf 1.x` (i.e. [`src/io/export_3mf_v1.cc`](https://github.com/openscad/openscad/blob/master/src/io/export_3mf_v1.cc)).
- Does **not** affect builds against `lib3mf >= 2` ([`src/io/export_3mf_v2.cc`](https://github.com/openscad/openscad/blob/master/src/io/export_3mf_v2.cc)).
- Triggered whenever `append_3mf` receives a `GeometryList` and visits each child via `append_polyset` — most easily through `--enable=lazy-union`. Without lazy-union the top level usually collapses into a single mesh and the bug stays dormant.

## Root cause

`ExportContext::materialids` is a `std::vector<DWORD>` member of the per-export context (see [`src/io/export_3mf_v1.cc:71`](https://github.com/openscad/openscad/blob/master/src/io/export_3mf_v1.cc#L71)). `append_polyset` `push_back`s each new mesh's per-color material IDs into it without ever clearing:

```cpp
ctx.materialids.reserve(sorted_ps->colors.size());
for (size_t i = 0; i < sorted_ps->colors.size(); i++) {
  ctx.materialids.push_back(materialFunc(materials + i, sorted_ps->colors[i]));
}
```

`handle_triangle_color` then indexes it with `color_index`:

```cpp
if (lib3mf_propertyhandler_setbasematerial(propertyhandler, triangle_index, ctx.basematerialid,
                                           ctx.materialids[color_index]) != LIB3MF_OK) {
```

But `color_index` is **per-mesh-relative** — an index into *this* PolySet's `->colors` array. After the first mesh seeds `ctx.materialids = [first_id]`, the second mesh's triangles (whose own PolySet has only one color, so all triangles carry `color_index=0`) look up `ctx.materialids[0]` and resolve to the **first** mesh's material ID instead of the newly-registered second one.

The lib3mf v2 path is unaffected because it keys per-color material registration off `ctx.colors` (a `Color4f -> idx` map) instead of a vector indexed by `color_index`.

## Fix

`materialids` is only ever pushed to and read from within a single `append_polyset` call. Move it from `ExportContext` to a local in `append_polyset` and pass it as a parameter to `handle_triangle_color`. Each mesh now starts with an empty `materialids` so its own `color_index=0` resolves to its own first material ID.

A PR will follow shortly.

## Refs

- #6159 (the prior, prematurely-closed report — same root cause, same reproducer).

nomike:
Closes #6813. Continuation of #6159 (which was self-closed by the original reporter 4 minutes after opening with no fix landed; their "fixed in yesterday's version" build was almost certainly linked against `lib3mf >= 2`, which is unaffected — see #6813 for the full reasoning).

## Summary

With `--enable=lazy-union` (or any other path that produces a top-level `GeometryList`), exporting a multi-color scene to 3MF produces a file in which **every mesh after the first** has its triangle properties pointing at the **first** mesh's base material. The visible result is that all but the first object render in the wrong color.

The bug only fires on builds linked against `lib3mf 1.x`. The v2 path is unaffected.

## Root cause

`ExportContext::materialids` is a `std::vector<DWORD>` member of the per-export context. `append_polyset` `push_back`s each new mesh's per-color material IDs into it without ever clearing, while `handle_triangle_color` then indexes it with the per-mesh-relative `color_index`:

```cpp
struct ExportContext {
  ...
  std::vector<DWORD> materialids;   // shared across every mesh
  ...
};

bool append_polyset(...) {
  ...
  for (size_t i = 0; i < sorted_ps->colors.size(); i++) {
    ctx.materialids.push_back(materialFunc(materials + i, sorted_ps->colors[i]));
  }
  ...
}

bool handle_triangle_color(...) {
  ...
  lib3mf_propertyhandler_setbasematerial(propertyhandler, triangle_index,
                                         ctx.basematerialid,
                                         ctx.materialids[color_index]);
  ...
}
```

After the first mesh seeds `ctx.materialids = [first_id]`, the second mesh's triangles -- whose own PolySet has only one color, so all triangles carry `color_index = 0` -- look up `ctx.materialids[0]` and resolve to the **first** mesh's material ID instead of their own newly-registered one.

The `lib3mf v2` path (`export_3mf_v2.cc`) is unaffected because it keys per-color material registration off `ctx.colors` (a `Color4f -> idx` map) instead of a vector indexed by `color_index`.

## Fix

`materialids` is only ever pushed to and read from within a single `append_polyset` call, so move it from `ExportContext` to a local in `append_polyset` and pass it as a parameter to `handle_triangle_color`. Each mesh now starts with an empty `materialids` so its own `color_index = 0` resolves to its own first material ID.

Diff is small and surgical (~14/-6):

- `ExportContext::materialids` field removed.
- `append_polyset` declares a local `std::vector<DWORD> materialids` per call (with an in-code comment explaining the per-mesh scoping).
- `handle_triangle_color` takes `const std::vector<DWORD>& materialids` as a parameter.

## Test plan

Reproducer (the exact script from #6159):

```scad
color("red") cube(10);
color("blue") translate([12, 0, 0]) cube(10);
```

```bash
openscad --enable=lazy-union --enable=predictible-output \
         --export-format 3mf -o out.3mf script.scad
unzip -p out.3mf 3D/3dmodel.model
```

- [x] Verified the bug exists on `master @ b38f68881` linked against `lib3MF.so.1` (second `<object>`'s triangles all reference `p1="1"` = red).
- [x] Verified the fix produces correct output: second object's triangles now reference `p1="2"` (= Color 2 = blue), matching its `color("blue")` source.
- [x] Three-mesh stress (red, blue, red again): each object correctly references its own slot (`p1=1`, `p1=2`, `p1=3`), no cross-mesh aliasing.
- [x] Single-color scene through the same code path is unaffected (no triangle-level properties when no `color()` is set).
- [x] No changes to `src/io/export_3mf_v2.cc` (the lib3mf-v2 path was already correct).

## Notes

- This commit was developed and validated in PythonSCAD, where it has been merged on a branch that triggers the same code path through the in-script `export(dict, ...)` form. Since the PythonSCAD and OpenSCAD `export_3mf_v1.cc` files only diverge in PythonSCAD-specific `ENABLE_PYTHON` / `Export3mfPartInfo` plumbing -- not in the `materialids` logic -- the cherry-pick was clean.
- A regression test (e.g. a multi-mesh `lazy-union` 3MF golden in `tests/regression/dump3mftest/`) is intentionally not included to keep this PR minimal. Happy to add one in a follow-up commit on this branch if you'd like.

### DIFF
--- a/src/io/export_3mf_v1.cc
+++ b/src/io/export_3mf_v1.cc
@@ -68,7 +68,6 @@ struct ExportContext {
   int modelcount = 0;
   Color4f defaultColor;
   DWORD defaultColorId = 0;
-  std::vector<DWORD> materialids;
   const ExportInfo& info;
   const std::shared_ptr<const Export3mfOptions> options;
 };
@@ -117,7 +116,8 @@ int count_mesh_objects(PLib3MFModel *& model)
 }
 
 bool handle_triangle_color(PLib3MFPropertyHandler *propertyhandler, const std::unique_ptr<PolySet>& ps,
-                           int triangle_index, int color_index, ExportContext& ctx)
+                           int triangle_index, int color_index, const std::vector<DWORD>& materialids,
+                           ExportContext& ctx)
 {
   if (color_index < 0) {
     return true;
@@ -134,7 +134,7 @@ bool handle_triangle_color(PLib3MFPropertyHandler *propertyhandler, const std::u
 
   if (ctx.basematerial) {
     if (lib3mf_propertyhandler_setbasematerial(propertyhandler, triangle_index, ctx.basematerialid,
-                                               ctx.materialids[color_index]) != LIB3MF_OK) {
+                                               materialids[color_index]) != LIB3MF_OK) {
       export_3mf_error("Can't set triangle base material.", ctx.model);
       return false;
     }
@@ -155,6 +155,14 @@ bool handle_triangle_color(PLib3MFPropertyHandler *propertyhandler, const std::u
  */
 bool append_polyset(const std::shared_ptr<const PolySet>& ps, ExportContext& ctx)
 {
+  // Per-mesh local: each PolySet's `color_indices` are indices into *this*
+  // mesh's `colors`, so the lib3mf-side material IDs registered for those
+  // colors must also be looked up per-mesh. Keeping this in `ctx` (as was
+  // done previously) caused the second mesh's `color_index=0` to resolve
+  // to the FIRST mesh's material ID, silently painting all subsequent
+  // meshes with the first one's color. See #6813.
+  std::vector<DWORD> materialids;
+
   PLib3MFModelMeshObject *mesh = nullptr;
   if (lib3mf_model_addmeshobject(ctx.model, &mesh) != LIB3MF_OK) {
     export_3mf_error("Can't add mesh to 3MF model.", ctx.model);
@@ -239,9 +247,9 @@ bool append_polyset(const std::shared_ptr<const PolySet>& ps, ExportContext& ctx
       }
     }
 
-    ctx.materialids.reserve(sorted_ps->colors.size());
+    materialids.reserve(sorted_ps->colors.size());
     for (size_t i = 0; i < sorted_ps->colors.size(); i++) {
-      ctx.materialids.push_back(materialFunc(materials + i, sorted_ps->colors[i]));
+      materialids.push_back(materialFunc(materials + i, sorted_ps->colors[i]));
     }
   }
 
@@ -253,7 +261,7 @@ bool append_polyset(const std::shared_ptr<const PolySet>& ps, ExportContext& ctx
 
   for (size_t i = 0; i < sorted_ps->color_indices.size(); ++i) {
     const int32_t idx = sorted_ps->color_indices[i];
-    if (!handle_triangle_color(propertyhandler, sorted_ps, i, idx, ctx)) {
+    if (!handle_triangle_color(propertyhandler, sorted_ps, i, idx, materialids, ctx)) {
       return false;
     }
   }


t-paul:
Superseded by #6857

PR #6819, opened 2026-05-05
Fix lib3mf write/seek callback ABI mismatch (wasm)

### LINKED ISSUE #6818: 3MF export aborts with "function signature mismatch" on the WASM build (lib3mf C ABI / OpenSCAD callback type mismatch)
**Affected:** `openscad-wasm` builds — confirmed on `OpenSCAD-2026.04.26-WebAssembly-web` and the prior `wasm30346` snapshot. Desktop x86_64 builds appear unaffected.

## Reproduction

1. Build OpenSCAD WASM via `openscad-wasm` (`make`), or use any current `WebAssembly-web.zip` snapshot.
2. Run any compile that produces real geometry, e.g. `cube([10,10,10]);`, into 3MF: `openscad input.scad -o out.3mf`.
3. The WASM aborts before producing output.

Console / log:

```
Could not initialize localization (application path is '/').
ECHO: …
[empty out.3mf written]
thrown: { status: null, message: "function signature mismatch" }
```

## Root cause

`src/io/export_3mf_v2.cc` declares the lib3mf write/seek callbacks with C++/STL types and casts them to the lib3mf typedefs at the call site:

```cpp
uint32_t lib3mf_write_callback(const char *data, uint32_t bytes, std::ostream *stream)
{
  stream->write(data, bytes);
  return !(*stream);
}

uint32_t lib3mf_seek_callback(uint64_t pos, std::ostream *stream)
{
  stream->seekp(pos);
  return !(*stream);
}

…
writer->WriteToCallback(
    (Lib3MF::WriteCallback)lib3mf_write_callback,
    (Lib3MF::SeekCallback)lib3mf_seek_callback,
    &output);
```

But lib3mf's typedefs (verified in v2.3.2 *and* v2.4.1, both at `Autogenerated/Bindings/Cpp/lib3mf_types.hpp`):

```cpp
typedef void (*WriteCallback)(Lib3MF_uint64, Lib3MF_uint64, Lib3MF_pvoid);
typedef void (*SeekCallback) (Lib3MF_uint64, Lib3MF_pvoid);
```

Three real ABI mismatches per call:

| | OpenSCAD's type | lib3mf typedef | wasm32 effect |
|---|---|---|---|
| Return | `uint32_t` (i32) | `void` | extra return register / value ignored |
| Write arg 1 | `const char *` (i32) | `Lib3MF_uint64` (i64) | size mismatch |
| Write arg 2 | `uint32_t` (i32) | `Lib3MF_uint64` (i64) | size mismatch |
| Seek arg 1 | `uint64_t` (i64) | `Lib3MF_uint64` (i64) | match |
| Both arg 3 / 2 | `std::ostream *` (i32) | `Lib3MF_pvoid` (i32) | match |

The C-style casts at the `WriteToCallback()` site silence the C++ compiler. On x86_64 desktop the indirect-call ABI is permissive: extra return registers are unused, the upper 32 bits of `Lib3MF_uint64` hold the kernel-zero pointer canonicalisation, and the call lands. WASM enforces strict signature matching on every indirect-call through the function table, so it correctly aborts with `"function signature mismatch"`.

## Patch

```cpp
// src/io/export_3mf_v2.cc

void lib3mf_write_callback(Lib3MF_uint64 nByteData,
                           Lib3MF_uint64 nNumBytes,
                           Lib3MF_pvoid pUserData)
{
  auto *stream = static_cast<std::ostream *>(pUserData);
  const char *data = reinterpret_cast<const char *>(static_cast<uintptr_t>(nByteData));
  stream->write(data, static_cast<std::streamsize>(nNumBytes));
}

void lib3mf_seek_callback(Lib3MF_uint64 pos, Lib3MF_pvoid pUserData)
{
  auto *stream = static_cast<std::ostream *>(pUserData);
  stream->seekp(static_cast<std::streamoff>(pos));
}

…

// no cast needed once the signatures match the typedefs:
writer->WriteToCallback(lib3mf_write_callback,
                        lib3mf_seek_callback, &output);
```

## Proof of fix

I built `openscad-wasm` against `openscad@master` with this patch applied (lib3mf v2.3.2, the same version openscad-wasm currently links). 3MF export from a multi-part assembly produces a valid file that Bambu Studio opens without complaint. Same compile path on the unpatched 2026.04.26 snapshot fails as above.

[IPR if useful](https://github.com/openscad/openscad/pull/6819)

blue-urban-sky:
## Summary

Fixes #6818.

`src/io/export_3mf_v2.cc` declared the lib3mf write/seek callbacks with C++/STL types and reinterpreted them via a C-style cast at `WriteToCallback()`. lib3mf's typedefs (verified in v2.3.2 *and* v2.4.1, both at `Autogenerated/Bindings/Cpp/lib3mf_types.hpp`):

```cpp
typedef void (*WriteCallback)(Lib3MF_uint64, Lib3MF_uint64, Lib3MF_pvoid);
typedef void (*SeekCallback) (Lib3MF_uint64, Lib3MF_pvoid);
```

OpenSCAD's previous declarations:

```cpp
uint32_t lib3mf_write_callback(const char *data, uint32_t bytes, std::ostream *stream);
uint32_t lib3mf_seek_callback (uint64_t pos, std::ostream *stream);
```

Three real ABI mismatches per call: return type, write-arg-1 (`const char *` i32 vs `Lib3MF_uint64` i64), write-arg-2 (`uint32_t` i32 vs `Lib3MF_uint64` i64). The C-style casts at the `WriteToCallback()` site silenced the C++ compiler. On x86_64 desktop the indirect-call ABI is permissive enough that the call lands; WebAssembly enforces strict signature matching on every indirect-call through the function table, so it correctly aborts with `"function signature mismatch"` and writes an empty `out.3mf`.

## Patch

- Match the lib3mf typedefs exactly: `void(Lib3MF_uint64, Lib3MF_uint64, Lib3MF_pvoid)` for write, `void(Lib3MF_uint64, Lib3MF_pvoid)` for seek.
- Demarshal `pUserData` → `std::ostream *` and the data uint64 → `const char *` inside the body.
- Drop the `(Lib3MF::WriteCallback)` / `(Lib3MF::SeekCallback)` casts at the call site since the function pointers now match the typedef directly.

This is also a strict improvement on x86_64: the original code is technically UB (calling a function pointer through an incompatible signature is undefined per the C++ standard, even when it happens to work on a given target ABI).

## Proof of fix

Built openscad-wasm against this branch (lib3mf v2.3.2, the version openscad-wasm currently links). 3MF export from a multi-part assembly produces a valid file that PrusaSlicer / Bambu Studio / OrcaSlicer all open without complaint. The unpatched 2026.04.26 wasm snapshot fails reliably on the same input. See #6818 for the full diagnosis and reproduction steps.

## Test plan

- [x] WASM build of this branch produces a valid 3MF.
- [x] Reproduction case from #6818 (any geometry exported to `.3mf`) now succeeds on wasm.
- [ ] No regression on x86_64 desktop (please verify on your CI — I don't have that environment set up).


t-paul (approved):
Thanks!.

PR #6820, opened 2026-05-07
wasm: optional -flto and -msimd128 build flags (ENABLE_WASM_OPT)

blue-urban-sky:
## Summary

Adds an opt-in CMake option `-DENABLE_WASM_OPT=ON` that enables `-flto` and `-msimd128` on the Emscripten build.

- **`-msimd128`** — WebAssembly fixed-width 128-bit SIMD. Universally supported in browsers since 2022 (Safari 16.4, Chrome 91, Firefox 89). Manifold's vector math and CGAL kernel arithmetic both benefit.
- **`-flto`** — cross-translation-unit inlining at the wasm-opt stage. Typical 5-15% runtime speedup on geometry-heavy code at the cost of a longer link step. Applied to both compile and link as emscripten requires for whole-program optimisation.

Off by default so CI builds don't accidentally pick up the longer link time. Native desktop is unaffected — the change is inside the existing `if (CMAKE_SYSTEM_NAME STREQUAL Emscripten)` block.

## Why optional

The `option(... OFF)` default keeps the upstream WASM build behaviour identical for anyone who isn't passing `-DENABLE_WASM_OPT=ON`. Downstream WASM bundlers (openscad-wasm, openscad-playground) can opt in when they want the extra perf.

## Test plan

- [x] `emcmake cmake -B build-web -DWASM_BUILD_TYPE=web -DCMAKE_BUILD_TYPE=Release -DEXPERIMENTAL=1` → builds as before (option default off).
- [x] `emcmake cmake -B build-web -DWASM_BUILD_TYPE=web -DCMAKE_BUILD_TYPE=Release -DEXPERIMENTAL=1 -DENABLE_WASM_OPT=ON` → builds with both flags on. Smoke-tested running BOSL2-using SCAD source through the resulting `openscad.{js,wasm}` against a multi-part 3MF export.
- [ ] Native desktop unaffected (no diff outside the Emscripten block).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -314,6 +314,25 @@ if (CMAKE_SYSTEM_NAME STREQUAL Emscripten)
     # Alternatively, can pass -gsource-map -sLOAD_SOURCE_MAP but then variable debugging isn't possible
     $<$<CONFIG:Debug>:-g -O0>
   )
+
+  # Optional WebAssembly performance flags. Off by default so CI builds
+  # don't accidentally pick up the longer link time. Enable with
+  # `emcmake cmake -DENABLE_WASM_OPT=ON ...`.
+  option(ENABLE_WASM_OPT
+    "Enable -flto and -msimd128 on the Emscripten build. Adds 5-15% runtime speedup on geometry-heavy code at the cost of a longer link step."
+    OFF)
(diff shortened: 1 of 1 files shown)


t-paul:
Are there some specific numbers for comparison on at least one of the browsers?

Issue #357, opened 2013-05-09
Example: Example using resize() [$5]


kintel:
Note: Examples should be functional and not purely a technical demo of the features.
##

MichaelAtOz:
My hull example also uses resize.

MichaelAtOz:
Another, perhaps under Advanced??

```
// public domain.
//
// Make a line proportional to the zoom level.
// Need to do a preview (F5) after you zoom the display to adjust the size.
// from/to are [x,y,z], real=true means the line will show in render (F6)
module line(from, to, real=false) {
	wide=($vpd==undef)	// Covers older version which do not support $vpd
			? 0.75
			: $vpd/300;
	echo($vpd, wide=wide);
	//
	// sub module, only visible within the line() module
	module _line(f, t) {
		hull() { // two cubes at each end of the line, hull() joins them
			translate(f)
				cube(wide);
			translate(t)
				cube(wide);
		}
	}
	color("black")
		if (real)
			_line(from, to);
		else
			%_line(from, to); // % is just for preview, not render
}

cube(10); // for perspective
// draw lines
line([10,10,0], [50,10,0]);
line([0,10,10], [0,50,10]);
```

bronzenose:
The essential difference between resize() and scale() is that for resize() you specify a literal final size; this can always be achieved with scale() instead if you know the size of the children, which, in scad, we generally do.  The only situation I know of where we **cannot** know the size of the children is when we use text().  This example draws a nameplate (as for an office or an ID badge) by laying out a field of inspirational words as a background grid and fitting a greeting and the name exactly into two foreground rectangles.

I think this effect (fitting the text to a grid) would be impossible without resize() so this is an effective example of the functionality.  I'm not sure if a nameplate counts as "functional" rather than a "technical demo" but nameplates seem like popular things to draw as there are (today) [396 things on thingiverse containing the word "nameplate"](https://www.thingiverse.com/search?q=nameplate&dwh=95bca1e5262d8a).

![hello badge requires resize](https://user-images.githubusercontent.com/2424042/47236210-40294d00-d390-11e8-9919-424c66623b5a.png)

```
cX=12;
mmX=15;
mmY=5;
mmYMargin = 0.1*mmY;
mmXMargin = mmYMargin;
mmZInspirational=1;
mmZPlate = 2;
mmZName = 3;
rgstrInspirational = ["OpenScad", "fresh", "damp", "push", "moon", "reaction",
[...]

PR #6831, opened 2026-05-14
Add example for resize() function

simonho234:
This adds an example demonstrating the resize() function in the Basics examples directory, addressing issue #357.


jordanbrown0:
Please split these two changes into two PRs.

With regard to the resize example, it's a good start but the `newsize` cases where the dimensions are non-zero are the easy ones.  What's really needed is something that demonstrates the `auto` cases and cases where `newsize` dimensions are zero.  It's those combinations that need to be explored; repeating the same basic operation on multiple shapes doesn't seem to add much.

PR #6838, opened 2026-05-22
Fixes state saving of docking windows.

t-paul:
Proposed as diff via IRC. Related to #6769.

### DIFF
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -1516,15 +1516,6 @@ bool MainWindow::event(QEvent *event)
 
 bool MainWindow::eventFilter(QObject *obj, QEvent *event)
 {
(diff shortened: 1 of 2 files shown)


jordanbrown0:
Quit currently quits the whole application, even if you have multiple main windows open.  This seems to change things so that Quit will only close the current window.

ISTM that the right answer is going to involve saving state immediately on any window manipulation (much like state is saved immediately on any preference change), and ensuring that during the quit and main-window-close sequences we **don't** save state.

jordanbrown0:
But also:  I don't know what saving window state means if you have more than one main window open.  Should it remember how many main windows you have open, and remember the pane state for each?

t-paul:
Right now we probably only have the option: Save = Position of the last closed window. Reopen App -> always only 1 windows comes back.

t-paul:
superseded by #6885.

PR #6840, opened 2026-05-26
fix(text): work around fontconfig 2.18.0 app-font scoring regression

nomike:
A few days ago we at PythonSCAD started seeing some font related tests fail on macOS and I did a thorough investigation into this.
I identified the root cause and developed a fix, which I'm here back-porting to OpeSCAD.

Should be relatively straight forward to merge.

## TLDNR

Fontconfig previously prioritized application supplied fonts more than system fonts. With version 2.18.0 they "fixed" this, now giving system fonts more preference. This is usually what you want in a user interface as it creates a more coherent user experience.

In OpenSCAD however, if a user adds a font to their design, they usually want that font to be used and not some system font which fonconfig thinks is similar enough.

## Problem

fontconfig 2.18.0 (shipped in Homebrew on 2025-05-22) changed priority
scoring so fonts registered via `FcConfigAppFontAddFile` — what OpenSCAD
uses for `use <font.ttf>` directives — can lose to system fonts even on
exact family name matches.

On macOS with a Japanese locale, `FcDefaultSubstitute` injects `lang=ja`
into the match pattern. This caused fontconfig 2.18.0 to rank Hiragino
Sans above the explicitly-requested app fonts (Amiri, EvenOddTT,
MarVoSym), breaking 16 `text-font-*` regression tests.

As fontconfig get's updated on other platforms as well, I expect this to become an issue on Windows and macOS as well.

## Fix

**`fonts/10-liberation.conf`** — change alias bindings from `"strong"` to
`"weak"`. The strong bindings were causing Liberation Sans to outrank any
other requested family in fontconfig's scoring.

**`src/FontCache.cc`** — in `find_face_fontconfig`, capture the requested
family/style from the raw parsed pattern *before* `FcConfigSubstitute` /
`FcDefaultSubstitute` run. Then scan the app font set directly:

1. Prefer an exact `family + style` match
2. Fall back to `family + "Regular"` when no style was requested
3. Fall back to any font with the matching family
4. Only call `FcFontMatch` (global scoring) when the family isn't in the
   app set at all

This bypasses the 2.18.0 scoring change entirely for fonts the user
explicitly registered via `use <>`.

## Testing

All 118 `text-font-*` ctest cases pass locally with fontconfig 2.18.0.

## Test plan

- [ ] All `text-font-*` tests pass on macOS (fontconfig 2.18.0)
- [ ] All `text-font-*` tests pass on Linux (fontconfig 2.17.x)
- [ ] No regression in `text-font-charset`, `text-font-composition`, `text-font-spacing` tests

### DIFF
--- a/fonts/10-liberation.conf
+++ b/fonts/10-liberation.conf
@@ -1,19 +1,19 @@
 <?xml version="1.0"?>
 <!DOCTYPE fontconfig SYSTEM "fonts.dtd">
 <fontconfig>
-  <alias binding="strong">
+  <alias binding="weak">
     <family>sans-serif</family>
     <prefer>
       <family>Liberation Sans</family>
     </prefer>
   </alias>
-  <alias binding="strong">
+  <alias binding="weak">
     <family>serif</family>
     <prefer>
       <family>Liberation Serif</family>
     </prefer>
   </alias>
-  <alias binding="strong">
+  <alias binding="weak">
     <family>monospace</family>
     <prefer>
       <family>Liberation Mono</family>
--- a/src/FontCache.cc
+++ b/src/FontCache.cc
@@ -240,6 +240,7 @@ void FontCache::registerProgressHandler(InitHandlerFunc *handler, void *userdata
 
 void FontCache::register_font_file(const std::string& path)
 {
+  PRINTDB("register_font_file: '%s'", path);
   if (!FcConfigAppFontAddFile(this->config, reinterpret_cast<const FcChar8 *>(path.c_str()))) {
     LOG(message_group::Warning, Location::NONE, "", "Can't register font '%1$s'", path);
   }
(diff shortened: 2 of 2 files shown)


jordanbrown0:
This relates to the question of whether a system installation of Liberation Sans wins over a copy of Liberation Sans delivered by the application.  I never really found a satisfactory way to get the application installation to win.  (I had a few clues, but got distracted and haven't gotten back to it.)

I'm interested in looking at this, but I'm traveling at the moment and don't really have the time and resources to look at it until I get home on June 5.

t-paul:
This seems to be already fixed via https://gitlab.freedesktop.org/fontconfig/fontconfig/-/work_items/520

Considering the list of affected apps mentioned in that ticket, it's likely a bugfix version will happen...

t-paul:
https://gitlab.freedesktop.org/fontconfig/fontconfig/-/releases/2.18.1

t-paul:
Fixed by fontconfig 2.18.1, verified for both macOS and msys2 in #6846.

PR #6844, opened 2026-05-30
feat(release): Windows code signing via Certum SimplySign (local script)

### LINKED ISSUE #688: Optimization parameter for union() to assume that all children are non-intersecting
Quite often when using OpenSCAD for printing, I create a module for creating a tightly packed "plate" of parts. While this works extremely well (esp. packing odd shapes), sometimes the CGAL step takes an enormous amount of time. For example, there's a part I'm looking at right now that took a few seconds to CGAL-render a single copy, but it's been over an hour to CGAL-render a plate of 24 (and still going). While it would be an option to use another package (Blender?) to perform the plating, I cannot understate just how useful it has been to do the math in OpenSCAD (especially for highly parametric parts).

To work around this problem, I would like there to be some sort of optimization flag I could set to effectively say "assume that all children are non-intersecting", either passed to union() to wrap up the output of my plating modules or as some sort of global flag (akin to global $fn) that affects the (to my knowledge) root's implicit union. If the parts are indeed intersecting and bad geometry is generated, then presumably the slicer would catch the problem (though at least visual inspection would have caught the intersection).

In any case, thank you for the excellent tool!

 -Austin

P.S. If alternatively it would be inexpensive to have union() always perform a first-pass set of intersection checks between children during CGAL rendering, then that would be superb as it would mean that output would still always be correct, and that situations such as a module generating plating would silently perform enormously faster.
## 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1432275-optimization-parameter-for-union-to-assume-that-all-children-are-non-intersecting?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52063&utm_medium=issues&utm_source=github).
</bountysource-plugin>

nomike:
## Summary

- Replaces the jsign/CRYPTOCERTUM approach from #688 (which targets the physical cryptoCertum smartcard, not the SimplySign **cloud** HSM) with a proven, tested approach using `signtool` via SimplySign Desktop.
- Adds a local PowerShell signing script for maintainer use per release.
- Modifies `publish-release.yml` to pause after builds and wait for manual signing before publishing.
- Rewrites `doc/windows-code-signing.md` to match reality.

## What changed

**`scripts/sign-windows-release.ps1`** (new)
- Downloads NSIS installer (`.exe`) and MSIX from a GitHub pre-release
- Signs both with `signtool` using the Certum certificate (via SimplySign Desktop's Windows cert store integration)
- Verifies each signature (`Status: Valid`)
- Re-uploads signed files to the release
- Works with the pinless certificate — no PIN prompt

**`.github/workflows/publish-release.yml`**
- On automatic release event: after all builds succeed, appends signing instructions to the release body and leaves the release as pre-release (instead of publishing immediately)
- On manual `workflow_dispatch` (triggered after signing): publishes as before

**`doc/windows-code-signing.md`** (complete rewrite)
- Correct signing flow: `signtool` + SimplySign Desktop (not jsign)
- One-time setup checklist, per-release procedure, what gets signed and why
- Notes on future TOTP-based automation on a self-hosted runner

## Why not jsign CRYPTOCERTUM?

Per [jsign's own docs](https://ebourg.github.io/jsign/): `CRYPTOCERTUM` "communicates directly with the card" — it's for the **physical** cryptoCertum smartcard. The SimplySign **cloud** HSM works differently: SimplySign Desktop runs as a tray app, authenticates to the cloud via a TOTP token from the mobile app, and exposes the certificate in the Windows cert store. `signtool` then uses it normally.

## Test plan

- [x] Local signing verified on the maintainer's Windows machine: `signtool verify` returned `sha256 RFC3161 — Successfully verified`
- [x] Certificate: `CN=Open Source Developer Michael Postmann` issued by `Certum Code Signing 2021 CA`, valid until 2027-05-30
- [x] Pinless certificate: signing proceeds immediately, no PIN prompt
- [x] All pre-commit hooks pass (markdownlint, YAML lint, zizmor)

Closes #688


nomike:
Opened against wrong repository by mistake. The actual PR is at https://github.com/pythonscad/pythonscad

Issue #3111, opened 2019-10-30
Allow User/All-Users selection in Windows Installer


t-paul:
Follow up on discussion in #3110:

> Some installers provide a checkbox to choose an all user install or a current user install, based on that they can do a few things:
>
> * If it's an alluser install they'll elevate to admin, install to the C:\program files directory and create an alluser start shortcut.
> * If it's a currentuser install they'll install to %appdata%\local and create a current user start menu. Apps that install this way can avoid requesting admin access.
>
> OpenSCAD's installer is a bit inconsistent currently because it acts like an alluser install except for the shortcut. It requests admin, installs to programfiles, and creates a system-wide uninstaller, but then it puts the shortcut in the currentuser profile. Making the shortcut system-wide by default would make OpenSCAD's entire install process system-wide by default.
>
> I think to support both flows the installer would have to:
>
> * Show a checkbox for all/current user.
> * Provide default install locations based on all/current user.
> * Provide a silent switch to select the install type.
> * Store the type of installation somewhere so the uninstaller can handle it.

Rotzbua:
Additional topics:

- remove legacy code: 32bit, before win10
- migrate gui to MUI2

Issue #6021, opened 2025-07-13
Windows:  do not create "Uninstall" entry on Start Menu; create only a single OpenSCAD item


jordanbrown0:
**Is your feature request related to a problem? Please describe.**
OpenSCAD (like many other applications) creates three nodes in the Windows Start Menu:  a parent, the actual OpenSCAD entry, and an "uninstall" entry.  The "uninstall" entry seems unnecessary - the system has an "add or remove programs" feature that handles uninstalling any program.  It triggers the need for the parent to contain it and the actual entry, so there are three entries cluttering the menu instead of only one.

**Describe the solution you'd like**
Drop the "uninstall" entry and the parent.  Create only the primary OpenSCAD entry, at the top level.

**Describe alternatives you've considered**
Manually rearrange the menu.  "Pin to start page".

PR #6845, opened 2026-05-30
chore: windows installer

Rotzbua:
## Changes

- remove scripts `x64` and `LogicLib` which are now included with NSIS
- remove 32-bit script
- abort if not 64-bit
- update help link to available documentation
- remove wrong uninstall entry

## Reference

mentioned in #3111


t-paul:
Installer Build:
https://output.circle-artifacts.com/output/job/5a241ba3-523e-407c-9546-223fb5d4f14a/artifacts/0/MXE/OpenSCAD-2026.05.30-x86-64_PR6845-Installer.exe

jordanbrown0:
Tried installing and uninstalling this on Windows 11.  Seems to mostly work.

Two hiccups - not sure whether they are new:
* Didn't immediately show up in the Start Menu all-apps list.
* If installed with an alternate Start Menu folder (e.g. OpenSCAD-6845), uninstall didn't remove the shortcut.

Also, while you're here, maybe you might look at #6021.

Also also, not new:  when you install multiple instances of OpenSCAD nightly to different folders, only the most recently installed shows up in the system uninstall list.  Maybe there is some sort of install ID that should be unique per installation folder.

Rotzbua:
Seems the script is currently not used. The installer is created by the cmake plugin CPack NSIS Generator.

So the changes are not necessary and the script should be removed. I will create a new PR to remove the old.

Machine and user scope installation is not supported by cmake, so a custom script is still will be required. I think a rewrite from scratch is easier for this goal.

PR #6846, opened 2026-05-30
Add autocomplete feature for user variables.

ddayjohn:


t-paul:
Nice, thank you! I'll give it a try.

Just 2 initial points:
* For the formatting check to go green, have a look at https://github.com/openscad/openscad/blob/master/doc/hacking.md#beautifying-code - It's either possible manually via the script or by using the `pre-commit` application. If there are questions, we can discuss on IRC.
* Please have a look through the changes and remove the commented out debug code (or if there's anything you really want to be included anyway, add a comment why it's there) and remove the extra `.qtcreator/CMakeLists.txt.user`, maybe we should add `/.qtcreator` to our `.gitignore`.

t-paul (approved):
Looks good, lets see how the different strategies work out. I guess long term, we should settle for one solution, but I don't see any harm in working with both for now.

The test failures should hopefully resolve themselves soon, assumption is they are caused by a regression in fontconfig 2.18.0 that was released for both macOS/homebrew and msys2. Both are in the process of releasing an update already, I'll re-trigger the actions once they have the updates published.

ddayjohn:
> Looks good, lets see how the different strategies work out. I guess long term, we should settle for one solution, but I don't see any harm in working with both for now.
> 
> The test failures should hopefully resolve themselves soon, assumption is they are caused by a regression in fontconfig 2.18.0 that was released for both macOS/homebrew and msys2. Both are in the process of releasing an update already, I'll re-trigger the actions once they have the updates published.

Thank you so much 
The reason I kept the Regex approach is that it can run in real-time within the 'onTextChanged' function instead of waiting for compilation to complete (maybe triggered via a timeout event), which could be useful in the future. I’ve already tested it on my machine, and the Regex is definitely fast enough to handle it.What do you think?

t-paul:
Yes, both have some properties that make them useful, running a compilation (actually the parsing part only) in the background should be possible and that could extend to easily pick up functions from libraries too. On the other hand, the regex version does not rely on everything being without syntax errors.

I tried both versions with a big (552K) library file and that worked fine with a quick test. Thanks for adding this feature!

ddayjohn:
Glad to hear that! Thanks for the quick review and merge. I'll definitely recommend OpenSCAD to my mechanical engineering friends and encourage them to give it a try too... and maybe brag to them a bit that my code is in it! 😉

t-paul:
:grinning: feel free to add yourself to https://github.com/openscad/openscad/blob/a88d5f1b0ac30af9e2f67f12e7ff841e4faa4cca/resources/html/AboutDialog.html#L132 - that makes it easier to see.

PR #6847, opened 2026-05-31
chore: windows installer

Rotzbua:
### Change

Remove old windows nsis script. Was replaced by cmake nsis package.

### Reference

- https://github.com/openscad/openscad/pull/6845#issuecomment-4586667232
- mentioned in #3111


t-paul (approved):
Oh my, thanks! Looking through history, that happened quite some time ago via #4067. Not using Windows myself, I totally forgot about that one.

PR #6849, opened 2026-06-01
COPY - fix(text): work around fontconfig 2.18.0 app-font scoring regression

rcolyer:
This is a duplicate of #6840 by @nomike to make the broken CI tests rerun.

fontconfig 2.18.0 changed priority scoring so fonts registered via FcConfigAppFontAddFile (the app font set) can lose to system fonts even on exact family name matches.  On macOS with a Japanese locale this meant FcFontMatch returned Hiragino Sans instead of the explicitly requested Amiri/EvenOddTT/MarVoSym fonts from use<> directives.

Two changes:

1. fonts/10-liberation.conf: change alias bindings from "strong" to "weak" so the Liberation fallback aliases don't override the explicitly requested font family in FcFontMatch scoring.

2. FontCache::find_face_fontconfig: before calling FcFontMatch, capture the requested family/style from the raw parsed pattern and scan the app font set directly.  Prefer an exact family+style match, fall back to family+"Regular" when no style is requested, then to any family match.  Only call FcFontMatch when the family is not present in the app font set at all.

Fixes all 16 text-font-* test failures introduced by the Homebrew fontconfig 2.17.1 → 2.18.0 upgrade.

### DIFF
--- a/fonts/10-liberation.conf
+++ b/fonts/10-liberation.conf
@@ -1,19 +1,19 @@
 <?xml version="1.0"?>
 <!DOCTYPE fontconfig SYSTEM "fonts.dtd">
 <fontconfig>
-  <alias binding="strong">
+  <alias binding="weak">
     <family>sans-serif</family>
     <prefer>
       <family>Liberation Sans</family>
     </prefer>
   </alias>
-  <alias binding="strong">
+  <alias binding="weak">
     <family>serif</family>
     <prefer>
       <family>Liberation Serif</family>
     </prefer>
   </alias>
-  <alias binding="strong">
+  <alias binding="weak">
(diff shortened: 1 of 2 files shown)


t-paul (changes requested):
Upstream is already fixed: https://gitlab.freedesktop.org/fontconfig/fontconfig/-/work_items/520

Let's give it a bit to see if this gets released sometime soon-ish.

t-paul:
https://gitlab.freedesktop.org/fontconfig/fontconfig/-/releases/2.18.1

t-paul:
Fixed by fontconfig 2.18.1, verified for both macOS and msys2 in #6846.

PR #6865, opened 2026-06-10
Add HTTPClient (Boost.Beast) and OpenSSL link

dasbidyendu:
Introduce an asynchronous HTTPClient using Boost.Beast for HTTP/HTTPS GET, POST and streaming POST responses (chunked/SSE). Adds src/core/HTTPClient.h/.cc implementing an IO context with background thread, SSL support (OpenSSL), platform certificate loading (Win32 custom loader / set_default_verify_paths otherwise), and a templated Session for plain/SSL connections. Update CMakeLists to find OpenSSL, link OpenSSL::SSL and OpenSSL::Crypto (and crypt32 on Windows), and add src/core/HTTPClient.cc to core sources. Client exposes callback-based APIs for response, chunk, error and completion handling.


t-paul:
MacOS build issue should be fixed via #6868 (caused by Homebrew update from 5.1.15 -> 6.0.0).

t-paul (approved):

PR #6878, opened 2026-06-23
Shortcuts GUI (2)

t-paul:
This is #3338 rebased with some minor updates.


roel-v:
@t-paul You mention in  #6922 that this PR is 'far from perfect' (although mergeable), what would still need to change on this PR for it to be better? Or if it's more of a wish list, what (from the top of your head) could be improved here?

t-paul:
I have some recent updates improving the config file handling, using a proxy model for live filtering and a couple more mostly smaller changes. I need to run a couple of more tests myself to see if that's then good to go.

PR #6881, opened 2026-06-23
Add AI Tool Calling to OpenSCAD

dasbidyendu:
This PR adds tool calling support and a multi-turn agent loop to the AI Assistant, letting the AI interact directly with OpenSCAD's editor and preview window.

Key Changes

- Tool Calling Support: Added JSON schemas for three new tools: get_editor_code, set_editor_code, and trigger_preview.
- Autonomous Agent Loop: Added a recursive loop to run these tools safely on the Qt main thread, parsing the stream deltas and feeding results back to the LLM.
- Stream Engagement: Refined the default system prompt to relax the strict tool-only constraint, so the AI can explain its thoughts in real-time while executing tools.
- Testing

Open the AI Chat dock and ask the assistant to make a change (e.g., "Inspect my code, set it to draw a sphere, and show the preview"). It should read the code, update the editor, and trigger the F5 preview in real-time.


t-paul (approved):

t-paul:
@dasbidyendu main branch CI highlighted an issue I missed. For headless build the `src/core/*` code can't have a dependency to Qt, so we need to factor that out into a callback or so.

dasbidyendu:
> @dasbidyendu main branch CI highlighted an issue I missed. For headless build the `src/core/*` code can't have a dependency to Qt, so we need to factor that out into a callback or so.

ohh , let me fix that

PR #6885, opened 2026-06-26
Change window close sequencing yet again

### LINKED ISSUE #6817: Window arrangement is not reliably saved
Hi, I'm Luigi. I speak Italian. I apologize if I make any mistakes translating from Italian to English.

I'd like to report a problem with the latest version of OpenScad Snapshots. Every time I set up my workspace with my windows, I get the same error as in the second screenshot.

Every time I close and reopen the program, the settings don't stay.

Version: OpenSCAD 2026.04.26 
Windows 11: 64bit

Screenshot 01 :
Image

Screenshot 02 : 

Image

t-paul:
Change window close sequencing yet again (fixes #6817) (hopefully?).

- Always rely on our own list of top-level windows (Qt5 and Qt6). This removes the need of messing with close events in the event filter.
- Don't use Qt::closeAllWindows() as that closes all top-levels including docks that are floating.
- Manually close floating docks of the current main window when close is confirmed.
- Move logic to closeEvent() instead of misusing the destructor.

### DIFF
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -615,6 +615,57 @@ MainWindow::~MainWindow()
   delete this->cgalworker;
 }
 
+void MainWindow::closeEvent(QCloseEvent *event)
+{
+  if (!tabManager->shouldClose()) {
+    event->ignore();
+    return;
+  }
+  event->accept();
+
+  // Only save when this is the last MainWindow (or there are none left,
+  // which covers the edge case where closeEvent fires after another
+  // MainWindow has already been destroyed).
+  if (scadApp->windowManager.getWindows().size() == 1) {
+    saveWindowState();
+  }
+
+  isClosing = true;
+  progress_report_fin();
+
+  if (this->tempFile) {
+    delete this->tempFile;
+    this->tempFile = nullptr;
+  }
+
+  // Log to stdout from now on
+  clearCurrentOutput();
+  // Disable invokeMethod calls for consoleOutput during shutdown,
+  // otherwise will segfault if echos are in progress.
+  hideCurrentOutput();
+  for (auto& [dock, title] : docks) {
+    if (dock->isFloating()) {
+      dock->close();
+    }
+  }
+
+  scadApp->windowManager.remove(this);
+  if (scadApp->windowManager.getWindows().empty()) {
+    // Quit application even in case some other windows like
+    // Preferences are still open.
+    QApplication::quit();
+  }
+}
+
+void MainWindow::saveWindowState()
+{
+  QSettingsCached settings;
+  settings.setValue("window/geometry", saveGeometry());
+  auto windowState = saveState();
+  UIUtils::dumpSaveState(windowState);
+  settings.setValue("window/state", windowState);
+}
+
 void MainWindow::showProgress()
 {
   updateStatusBar(qobject_cast<ProgressWidget *>(sender()));
@@ -3276,57 +3327,6 @@ void MainWindow::on_helpActionLibraryInfo_triggered()
     this->libraryInfoDialog = dialog;
   }
   this->libraryInfoDialog->show();
-}
-
-void MainWindow::saveWindowState()
-{
-  QSettingsCached settings;
-  settings.setValue("window/geometry", saveGeometry());
-  auto windowState = saveState();
-  UIUtils::dumpSaveState(windowState);
-  settings.setValue("window/state", windowState);
-}
-
-void MainWindow::closeEvent(QCloseEvent *event)
-{
-  if (!tabManager->shouldClose()) {
-    event->ignore();
-    return;
-  }
-  event->accept();
-
-  // Only save when this is the last MainWindow (or there are none left,
-  // which covers the edge case where closeEvent fires after another
-  // MainWindow has already been destroyed).
-  if (scadApp->windowManager.getWindows().size() == 1) {
-    saveWindowState();
-  }
-
-  isClosing = true;
-  progress_report_fin();
-
-  if (this->tempFile) {
-    delete this->tempFile;
-    this->tempFile = nullptr;
-  }
-
-  // Log to stdout from now on
-  clearCurrentOutput();
-  // Disable invokeMethod calls for consoleOutput during shutdown,
-  // otherwise will segfault if echos are in progress.
-  hideCurrentOutput();
-  for (auto& [dock, title] : docks) {
-    if (dock->isFloating()) {
-      dock->close();
-    }
-  }
-
-  scadApp->windowManager.remove(this);
-  if (scadApp->windowManager.getWindows().empty()) {
-    // Quit application even in case some other windows like
-    // Preferences are still open.
-    QApplication::quit();
-  }
 }
 
 void MainWindow::on_editActionPreferences_triggered()
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -3993,13 +3993,7 @@ void MainWindow::setupMenusAndActions()
 void MainWindow::restoreWindowState()
 {
   const QSettingsCached settings;
-  // fetch window states to be restored after restoreState() call
-  const bool isEditorToolbarVisible = !settings.value("view/hideEditorToolbar").toBool();
-  const bool is3DViewToolbarVisible = !settings.value("view/hide3DViewToolbar").toBool();
-
-  // make sure it looks nice..
   const auto windowState = settings.value("window/state", QByteArray()).toByteArray();
-  // Log to stdout
   clearCurrentOutput();
   UIUtils::dumpSaveState(windowState);
   setCurrentOutput();
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -623,9 +623,7 @@ void MainWindow::closeEvent(QCloseEvent *event)
   }
   event->accept();
 
-  // Only save when this is the last MainWindow (or there are none left,
-  // which covers the edge case where closeEvent fires after another
-  // MainWindow has already been destroyed).
+  // Only save when this is the last MainWindow.
   if (scadApp->windowManager.getWindows().size() == 1) {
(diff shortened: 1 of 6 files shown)


jordanbrown0:
Can't review until Sunday, but should be able to then.

src/gui/MainWindow.cc
```diff
@@ -310,6 +310,61 @@ MainWindow::MainWindow(const QStringList& filenames) : rubberBandManager(this)
...
+    // Preferences are still open.
+    scadApp->closeApp();
+  }
+}
+
+void MainWindow::saveWindowStateOnClose()
+{
+  // Only save when this is the last MainWindow (or there are none left,
+  // which covers the edge case where closeEvent fires after another
+  // MainWindow has already been destroyed).
```
jordanbrown0: Might add a note about the lack of support for saving the state of multiple main windows.

jordanbrown0: I'm not following the "none left" case.  If there are none left, how could we be getting a close event?

src/gui/MainWindow.cc
```diff
@@ -310,6 +310,61 @@ MainWindow::MainWindow(const QStringList& filenames) : rubberBandManager(this)
...
+  if (scadApp->windowManager.getWindows().size() > 1) return;
```
jordanbrown0: Maybe this "policy" decision should be left to the caller, which makes this function responsible for saving state when called without making any decisions on whether saving state is appropriate.  (Changing its name to "saveWindowState" would make sense if you do that.)

src/gui/MainWindow.cc
```diff
@@ -310,6 +310,61 @@ MainWindow::MainWindow(const QStringList& filenames) : rubberBandManager(this)
...
+{
+  if (!tabManager->shouldClose()) {
+    event->ignore();
+    return;
+  }
+  event->accept();
+
+  isClosing = true;
+  saveWindowStateOnClose();
+  progress_report_fin();
```
jordanbrown0: Probably not this change:  It seems like there are **three** distinct stop-logging mechanisms:  progress_report_fin(), clearCurrentOutput(), and hideCurrentOutput().  That seems like two too many.

src/gui/MainWindow.cc
jordanbrown0, line 1056: Probably not this change:  This test seems like it should be reversed; it would be simpler as

```
 if (isClosing) return;

progress_report_prep(this->rootNode, report_func, this);
```

And again in MainWindow::cgalRender().

src/gui/MainWindow.cc
```diff
@@ -310,6 +310,61 @@ MainWindow::MainWindow(const QStringList& filenames) : rubberBandManager(this)
   openRemainingFiles(filenames);
 }
 
+MainWindow::~MainWindow()
+{
+  delete this->cgalworker;
+}
+
+void MainWindow::closeEvent(QCloseEvent *event)
+{
```
jordanbrown0: Maybe move these down to their original locations so that the differences are shown, instead of looking like a big removal and a big addition.

src/gui/MainWindow.cc
```diff
@@ -310,6 +310,61 @@ MainWindow::MainWindow(const QStringList& filenames) : rubberBandManager(this)
...
+  if (!tabManager->shouldClose()) {
+    event->ignore();
+    return;
+  }
+  event->accept();
```
jordanbrown0: I like that you moved this all up to the top of the function.

src/gui/OpenSCADApp.cc
```diff
@@ -124,8 +125,10 @@ void OpenSCADApp::setApplicationFont(const QString& family, uint size)
 //     In Qt 6, the method will instead try to close all top-level windows by posting a close
 //     event. The windows are free to cancel the shutdown process by ignoring the event.
 //
-// For Qt5, simulate the Qt6 behavior.
-void OpenSCADApp::quit()
+// But this implements a slightly modified Qt6 behavior as that triggers
+// deletion of all widget windows before calling MainWindow::closeEvent.
+// We really just call close on our own list of main windows.
+void OpenSCADApp::closeApp()
 {
```
jordanbrown0, line 132: So if you say to quit with File/Quit, it will call OpenSCADApp::closeApp(), which will close each of the main windows, and closing the last main window will call OpenSCADApp::closeApp() recursively, which will realize that there are no main windows left to close and will actually quit.

That seems twisted.

It seems like there should perhaps be a "quit event" function that gets called on File/Quit, that closes each of the main windows, and when that list drops to zero that close processing would call the "really quit" function, which would just quit.  That still seems a bit mysterious, since the "quit event" function itself wouldn't directly do the quit.  A comment would be needed.

src/gui/MainWindow.cc
```diff
@@ -3993,13 +3992,7 @@ void MainWindow::setupMenusAndActions()
 void MainWindow::restoreWindowState()
 {
   const QSettingsCached settings;
-  // fetch window states to be restored after restoreState() call
-  const bool isEditorToolbarVisible = !settings.value("view/hideEditorToolbar").toBool();
-  const bool is3DViewToolbarVisible = !settings.value("view/hide3DViewToolbar").toBool();
```
jordanbrown0, line 3998: These are saved in the saved window state, and were duplicated in separate settings?

It seems like they are now set in these settings, but never read.

jordanbrown0 (changes requested):
Seems like a reasonable direction, but I have a couple of concerns.

OpenSCADApp::closeApp() appears to be called recursively.

view/hideEditorToolbar and view/hide3DViewToolbar appear to be set but never read.

src/gui/MainWindow.cc
t-paul: No idea, just cleaning up unused variables.

jordanbrown0: Hmm.  The related stuff was removed in f65ead20663df608ab4c4c654ee21717b7900a43 on March 1.  It looks like before that these settings were persistent and after that they are not.  But maybe they are tied up in the overall failure to save state.

jordanbrown0: Checked with your PR; those settings are not persistent.  But at least a quick test says that pane rearrangement/open/close is persistent.  Maybe @kintel knows some of the thought behind removing it?

src/gui/MainWindow.cc
t-paul: Yes, not changed here.

src/gui/MainWindow.cc
t-paul: Hmm, yeah, maybe I'll post it as separate commits. Let me check that...

src/gui/MainWindow.cc
t-paul: done.

src/gui/MainWindow.cc
t-paul: agreed, but unrelated to the change.

src/gui/OpenSCADApp.cc
t-paul: Right, that was in the destructor earlier. That was also not ideal I suppose. Changed to call the base function which should be ok I hope.

t-paul:
Force pushed with small updates. Select only the first commit for the main changes to the close sequencing.

src/gui/MainWindow.cc
```diff
@@ -613,14 +613,59 @@ void MainWindow::updateReorderMode(bool reorderMode)
...
+  if (!tabManager->shouldClose()) {
+    event->ignore();
+    return;
+  }
+  event->accept();
+
+  // Only save when this is the last MainWindow (or there are none left,
+  // which covers the edge case where closeEvent fires after another
+  // MainWindow has already been destroyed).
+  if (scadApp->windowManager.getWindows().size() == 1) {
```
jordanbrown0: The comment doesn't seem to agree with the test. I think you need to delete the parenthesized bit.

src/gui/MainWindow.cc
```diff
@@ -613,14 +613,59 @@ void MainWindow::updateReorderMode(bool reorderMode)
...
+
+  // Log to stdout from now on
+  clearCurrentOutput();
+  // Disable invokeMethod calls for consoleOutput during shutdown,
+  // otherwise will segfault if echos are in progress.
+  hideCurrentOutput();
+  for (auto& [dock, title] : docks) {
+    if (dock->isFloating()) {
+      dock->close();
+    }
```
jordanbrown0, line 652: Might add a comment here, or at least separate it from the hideCurrentOutput() call.

I assume that the reason for this is that closing a MainWindow will close its docked child windows, but not its floating child windows.

src/gui/OpenSCADApp.cc
```diff
@@ -134,4 +137,3 @@ void OpenSCADApp::quit()
   }
   QApplication::quit();
```
jordanbrown0, line 141: Note that this is never reached. If all of the main windows closed properly, the last of them will have called QApplication::quit(). If any of them refused to close , we would have returned from the loop above.

I don't mind this being here as belt-and-suspenders, but it should perhaps get a comment.

jordanbrown0 (approved):
A couple of comment suggestions, but AFAICT the logic looks sound.

I find the path through the close logic and quit logic a bit unsatisfying - in particular, that we don't straightforwardly quit at the bottom of the "close all windows" loop - but I don't see a better way to handle the "close of last window" case.

src/gui/MainWindow.cc
t-paul: Indeed, done.

src/gui/MainWindow.cc
t-paul: done.

src/gui/OpenSCADApp.cc
t-paul: Good point. I'll leave it in, comment added.

jordanbrown0:
I'm on an hour of free WiFi from the middle of the Atlantic.  Haven't decided whether to pay for more.  If I don't, no time to review until Friday or so.

PR #6889, opened 2026-07-02
Treat non-strings as not-equal in string search() calls.

### LINKED ISSUE #5017: search() builtin in newer versions: Inconsistent typing in vector to be searched leads to C++ exception "std::get: wrong index for variant"
**Describe the bug**
When the search() builtin is used to search a vector, which contains an inconsistent index type (e.g. `[[0,0], ["a",1]]`), a C++ exception is thrown with no indication of where the error happened (see code section). This only happens in newer 2024 versions. In older versions, the invalidly typed index element is simply ignored.

**To Reproduce**
Attempt to run the code in a 2024 version.

**Expected behavior**
The invalidly typed vector element should either be ignored (this happens in older versions), or an error should be thrown accurately indicating that the vector contains an element with the wrong type.

```
echo(search(
    "ab",
    [[0,0], ["a",1],["b",2],["c",3]]
));
```
Output in 2022.03.07: `ECHO: [1, 2]`
Output in 2024.02.22: `ERROR: Compilation aborted by exception: std::get: wrong index for variant`

 - OS: Windows 10
 - System: AMD PC x86 64-bit
 - OpenSCAD Version: 2024.02.22 from openscad.org

jordanbrown0:
Previous version would silently ignore bad values.  This warns and ignores (and so will abort if stop-on-warn is set).  Discuss.

Fixes #5017.

### DIFF
--- a/src/core/Value.cc
+++ b/src/core/Value.cc
@@ -369,6 +369,7 @@ bool Value::getPositiveInt(unsigned int& v) const
 
 const str_utf8_wrapper& Value::toStrUtf8Wrapper() const
 {
+  // Caution:  throws if the value is not STRING.
   return std::get<str_utf8_wrapper>(this->value);
 }
 
--- a/src/core/builtin_functions.cc
+++ b/src/core/builtin_functions.cc
@@ -743,8 +743,13 @@ static VectorType search(const str_utf8_wrapper& find, const VectorType& table,
             j, (index_col_num + 1), table[j].toEchoStringNoThrow());
         return {session};
       }
-      if (!ft.empty() &&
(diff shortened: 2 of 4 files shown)


jordanbrown0:
There is an argument that search() should be equivalent to walking through the array doing ordinary comparisons.  Under that standard... OpenSCAD silently returns false for string==number comparisons, so there is a good argument that this case should just be a silent failure to match.

jordanbrown0:
In addition, the complementary case
```
echo(search(
    123,
    [[0,0], ["a",1],[123,2],[456,3]]
));
```
silently treats the type mismatch as failing to match that entry.

t-paul (approved):

PR #6891, opened 2026-07-03
Fix XML escaping in 3MF metadata export (lib3mf v1)

### LINKED ISSUE #6360: Generating 3MFs from STDIN generates broken XML
**Describe the bug**
When you pipe into stdin the filename internally is stored as <stdin> this is written to the title of the metadata of the outputted 3dmodel.model in a 3mf file without escaping.

`	<metadata name="Title"><stdin></metadata>`

**To Reproduce**
1. In a terminal/console run `echo "cube(1);" | openscad - -o test.3mf`
2. Unzip that 3mf
3. Inspect 3d/3dmodel.model

**Expected behavior**
`<metadata name="Title">&lt;stdin&gt;</metadata>`

`echo "cube(1);" | openscad - -o test.3mf`

 - Docker: openscad/openscad:bookworm
 - OpenSCAD version 2025.11.101

Narayan-ops:
Fixes #6360

## Problem

When exporting to 3MF using lib3mf v1.x, metadata values (title, designer, description, copyright, license terms, rating) were written directly into the XML without escaping special characters. This produces invalid XML when a value contains `&`, `<`, `>`, `"`, or `'` — most visibly with the default `<stdin>` title when piping SCAD input via stdin:

```
echo "cube(1);" | openscad - -o test.3mf
```

produced:

```
<metadata name="Title"><stdin></metadata>
```

instead of:

```
<metadata name="Title">&lt;stdin&gt;</metadata>
```

## Fix

Adds a `quote_xml()` helper in `src/io/export_3mf_v1.cc` that escapes all five reserved XML characters. It's applied in `add_meta_data()` before values are passed to `lib3mf_model_addmetadatautf8()`, so it covers every metadata field routed through that function (title, designer, description, copyright, license terms, rating) rather than just the title case from the original report.

The lib3mf v2 exporter is untouched, since v2 already escapes correctly.

## Testing

Since MSYS2 (my usual build environment) only ships lib3mf 2.4.1, and this bug is specific to the v1.x code path, I couldn't test the fix directly in my normal setup - the v1 exporter file wasn't even being compiled. To get an environment with actual lib3mf 1.x, I built and tested in a Debian bookworm Docker container (matching `openscad/openscad:bookworm`, the environment the original bug report used), where `apt` provides lib3mf 1.8.1.

Verified:

* The original repro case (`echo "cube(1);" | openscad - -o test.3mf`) now produces correctly escaped `&lt;stdin&gt;`
* All five XML special characters (`&`, `<`, `>`, `"`, `'`) escape correctly when set via `-O export-3mf/meta-data-designer=...`
* Normal (non-stdin) 3MF export still works with no regressions

Also added a regression test (`export-3mf-stdio` in `tests/CMakeLists.txt`) that pipes a SCAD file through stdin and exports 3MF via stdout, along with the expected escaped XML fixture, so this is covered by CI going forward (CI presumably builds against whichever lib3mf version is current in its environment - happy to adjust if needed).

### DIFF
--- a/src/io/export_3mf_v1.cc
+++ b/src/io/export_3mf_v1.cc
@@ -116,6 +116,25 @@ int count_mesh_objects(PLib3MFModel *& model)
   return count;
 }
 
+std::string quote_xml(const std::string& value)
+{
+  std::string quoted;
+  quoted.reserve(value.size());
+
+  for (const char c : value) {
+    switch (c) {
+    case '&':  quoted += "&amp;"; break;
+    case '<':  quoted += "&lt;"; break;
+    case '>':  quoted += "&gt;"; break;
+    case '"':  quoted += "&quot;"; break;
+    case '\'': quoted += "&apos;"; break;
+    default:   quoted += c; break;
+    }
+  }
+
+  return quoted;
+}
+
 bool handle_triangle_color(PLib3MFPropertyHandler *propertyhandler, const std::unique_ptr<PolySet>& ps,
                            int triangle_index, int color_index, std::vector<DWORD>& materialMap,
                            ExportContext& ctx)
@@ -343,7 +362,8 @@ void add_meta_data(PLib3MFModelMeshObject *& model, const std::string& name, con
     return;
   }
 
-  lib3mf_model_addmetadatautf8(model, name.c_str(), v.c_str());
+  const std::string quoted = quote_xml(v);
+  lib3mf_model_addmetadatautf8(model, name.c_str(), quoted.c_str());
 }
 
 }  // namespace
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -972,6 +972,7 @@ add_cmdline_test(export-obj              EXPERIMENTAL OPENSCAD SUFFIX obj FILES 
 if (ENABLE_LIB3MF_TESTS)
 add_cmdline_test(export-3mf              EXPERIMENTAL OPENSCAD SUFFIX 3mf FILES ${EXPORT_3MF_TEST_FILES} ARGS --enable=predictible-output)
 add_cmdline_test(export-3mf              EXPERIMENTAL OPENSCAD SUFFIX 3mf FILES ${EXPORT_3MF_LAZY_UNION_TEST_FILES} ARGS --enable=predictible-output --enable=lazy-union)
+add_cmdline_test(export-3mf-stdio        EXPERIMENTAL OPENSCAD SUFFIX 3mf FILES ${EXPORT_3MF_TEST_FILES} STDIO ARGS --enable=predictible-output --export-format 3mf)
 endif()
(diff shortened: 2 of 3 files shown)


Narayan-ops:
The macOS CI failure (`export-3mf-stdio_3mf-export`) doesn't appear to be related to the XML escaping fix - it looks like a pre-existing, platform-specific issue with 3MF export via stdout on macOS.

**What's happening:** OpenSCAD renders successfully (visible in the log — correct geometry, no errors), but the resulting file written to stdout isn't a valid ZIP (`zipfile.BadZipFile: File is not a zip file` when the test harness tries to read it back).

**Likely cause:** `export_3mf_v1.cc` writes via lib3mf's `lib3mf_writer_writetocallback()`, which uses both a write *and* a seek callback — the ZIP writer needs to seek backward to patch header fields after writing entry contents. `exportFileStdOut()` binds this to `std::cout`. Seeking on `std::cout` when redirected to a file works fine on Linux and Windows(msys2) (both libstdc++), but macOS uses libc++ by default, which may not support seeking on `std::cout` the same way in this context.

This is the first regression test to exercise 3MF export via stdout on macOS in CI (`export-3mf-stdio` is new in this PR), so it's likely exposing a latent issue rather than something introduced by this change - the same code path exporting to a regular file works correctly on all platforms, including macOS.

I don't want to scope-creep this PR into fixing stdout seek behavior on libc++, so I wanted to flag it and get direction: would you prefer I exclude this specific test on macOS (e.g. `EXCLUDE_FROM_ALL` or an `APPLE` guard in `tests/CMakeLists.txt`) for now, or is this worth a separate issue for someone more familiar with the stdout export path to dig into?

t-paul (approved):
Looks good. Thanks for also adding a test case! I agree with the macOS fail being related to STDIO, not the code change itself as macOS is using 2.4.1 currently.
Lets go with the fix, we can decide on how to handle the failure separately.

jordanbrown0:
Seeking stdout is just bad. It can work for files, but can't work for pipes, and there are realistic use cases for pipes. If the library must seek, the caller needs to write to a temporary file and then copy to stdout. It could do that only if stdout isn't a regular file, if we wanted to optimize.

t-paul:
I don't think seek on `stdout` works anywhere. The test case is using `stdin` writing to a file.

jordanbrown0:
@Narayan-ops :  could you please file an issue on the export-3MF-to-stdout issue, citing the test failure?

Narayan-ops:
Filed as #6895

jordanbrown0:
Thanks!

jordanbrown0:
> I don't think seek on stdout works anywhere. The test case is using stdin writing to a file.

Reading the input from stdin triggers the original quote issue.  I was talking about the export-to-stdout issue that suggests that for some reason stdout isn't seekable on macOS, even if it's a file.  Let's take that to the new issue.

PR #6896, opened 2026-07-04
Re-enable default Release build type in CMake

momiga:
The fallback that defaults CMAKE_BUILD_TYPE to Release when unset was commented out in ea8ed6b28 and never restored.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -376,14 +376,15 @@ endif()
 set(OPENSCAD_LIB_OUTPUT_DIR "${CMAKE_CURRENT_BINARY_DIR}/objects")
 file(MAKE_DIRECTORY ${OPENSCAD_LIB_OUTPUT_DIR})
 
-# Default to Release build
-#if(NOT CMAKE_BUILD_TYPE)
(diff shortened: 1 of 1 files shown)


t-paul:
Can you explain why it should go back? I believe it was disabled because this breaks Windows setups that use both Release and Debug setup in parallel.

momiga:
I'm not sure about the context in which the change was introduced. I figured it might have been for local testing and was accidentally left in, since it was commented out with no explanation. If parallel Release and Debug caused a problem, it's not really clear to me why this would fix it.

As it is, running CMake without specifying a build type causes it to compile without `-O3` which produces a build that renders about 3x slower than a release build, while also not having debug symbols, so doesn't seem particularly useful compared to explicitly choosing either Release or Debug. My understanding is usually projects that use CMake pick a default release type (most often Release) since CMake doesn't define a default.

jordanbrown0:
I don't know exactly what controls what, but note that full-on DEBUG builds don't work on MSYS2 - they exhaust some Windows symbol table. (Or at least they did the last time that I tried one.)

PR #6902, opened 2026-07-11
Fix minimum version lookup in check-dependencies.sh

### LINKED ISSUE #6302: check-dependencies.sh doesn't check `make` minimum version correctly
**Describe the bug**

The script parses `README.md` with a regex which allows prefixes (so `make` is finding `cmake`)

```
check-dependencies.sh: get_minversion_from_readme make
check-dependencies.sh: make
check-dependencies.sh: get_minversion_from_readme README.md found
check-dependencies.sh: [cmake (3.5 ->)](https://cmake.org/)
```

**To Reproduce**

`scripts/check-dependencies.sh debug`

**Expected behavior**
Output error because no minimum found for `make`

Narayan-ops:
## Summary

Fix `scripts/check-dependencies.sh` dependency minimum version lookup, which had two related issues:

1. The README parser matched dependency names loosely, so `make` could incorrectly match the `cmake` entry.
2. `cmake`'s minimum version was sourced from the README, which can drift from the actual build requirement.

This PR fixes README matching to use exact dependency labels and changes `cmake`'s minimum version to be read directly from `cmake_minimum_required()` in `CMakeLists.txt`, the build-system source of truth.

For dependencies without a defined minimum version, such as `make`, the script now reports `unknown` instead of inferring a value.

## Changes

- Match README dependency entries by exact dependency label instead of substring.
- Read the `cmake` minimum version from `cmake_minimum_required()` in `CMakeLists.txt`.
- Report `unknown / NotOK` when no minimum version is defined for `make`, instead of using an incorrect match or fallback.

## Verification

Ran:

- `bash -n scripts/check-dependencies.sh`
- `scripts/check-dependencies.sh debug`
- `shellcheck scripts/check-dependencies.sh`

Confirmed:

- `cmake` minimum: `3.13`
- `make` minimum: `unknown / NotOK`

ShellCheck warnings in the touched area were reviewed; remaining warnings are pre-existing script-wide cleanup items outside the scope of this issue.

Fixes #6302

### DIFF
--- a/scripts/check-dependencies.sh
+++ b/scripts/check-dependencies.sh
@@ -400,12 +400,12 @@ get_minversion_from_readme()
   debug get_minversion_from_readme $*
 
   # Extract dependency name
-  if [ ! $1 ]; then return; fi
+  if [ ! "$1" ]; then return; fi
   depname=$1
 
   debug $depname
   local grv_tmp=
-  for READFILE in README.md ../README.md "`dirname "$0"`/../README.md"
+  for READFILE in README.md ../README.md "$(dirname "$0")/../README.md"
   do
     if [ ! -e "$READFILE" ]
     then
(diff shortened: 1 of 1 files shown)


scripts/check-dependencies.sh
```diff
@@ -440,19 +449,62 @@ get_minversion_from_readme()
...
 
+  if [ "$fmvdep" = "cmake" ]; then
+    get_minversion_from_cmake
+    fmvtmp=$get_minversion_from_cmake_result
+  fi
+
   # items not included in README.md
   if [ $fmvdep = "git" ]; then fmvtmp=1.5 ; fi
   if [ $fmvdep = "curl" ]; then fmvtmp=6 ; fi
-  if [ $fmvdep = "make" ]; then fmvtmp=3 ; fi
```
t-paul, line 455: Why remove this? It's not very likely this needs to be changed or separately documented anytime soon, so having that here seems fine. Now the report shows "Not ok" when having version 4.4.1 installed which is more confusing than helpful as that version works.

t-paul (changes requested):

scripts/check-dependencies.sh
Narayan-ops: Restored the make=3 fallback in 47f3d08. My original reasoning was that a missing README entry should mean "unknown" rather than guessing - but that doesn't apply to make, since 3 was always an intentional hardcoded minimum, not something inferred from the README. Removing it just broke the OK/NotOK check for valid installs like 4.4.1.

t-paul (approved):

PR #6904, opened 2026-07-11
perf(manifold): N-ary BatchBoolean for union/intersection/difference

tonuonu:
## Summary

`applyOperator3DManifold` currently folds the pairwise operators over the child list
(`geom = geom OP child`), issuing N−1 separate `manifold::Manifold::Boolean` calls and
re-processing the growing accumulator at each step. Manifold provides an N-ary
`BatchBoolean` that does a balanced (and internally parallel) reduction, and for `Subtract`
computes `first − union(rest)`.

This PR collects the non-empty operands and evaluates union / intersection / difference in a
**single `BatchBoolean` call**. Minkowski keeps the pairwise fold.

## Relationship to the `kintel-manifold-ops` experiment

This was previously prototyped in the `kintel-manifold-ops` branch
("Experiment with BatchBoolean()", 5678e527b, 2024). That experiment operated on raw
`getManifold()` results, so it **dropped the `ManifoldGeometry` metadata**
(`originalIDs_` / `originalIDToColor_` / `subtractedIDs_`) and the empty-geometry
short-circuits — i.e. it would have broken per-object colors, part-IDs and empty-result
handling. This PR is the completed version: a new `ManifoldGeometry::boolOp` merges that
metadata **exactly** as the equivalent `binOp()` fold, and the empty / single-child edge
cases are preserved.

## Correctness (validated, 5 machines)

Batched output is geometrically identical to the pairwise fold — **volume / surface-area /
bbox match to FP precision** across union / intersection / difference / nested / real models;
a colored model renders **byte-identical** PNG; empty-result and single-child edge cases match
the previous behaviour. The result triangle *count* can differ by a couple (same solid, different
triangulation), so invariants — not tri counts — are the oracle. CircleCI is green.

## Performance (interleaved A/B, min-of-7; speedup = sequential / batched)

Honestly **modest and mostly union-shaped** — Manifold's sequential path is already
well-optimized (spatial culling), so the batched form mainly helps N-ary union (and many-cutter
difference). No regression on any machine, x86 or ARM:

| CPU | arch | cores | union | difference | intersection |
|---|---|---|---:|---:|---:|
| Intel Core i9-9880H | x86-64 | 8c/16t | 1.0–1.08× | ~1.0× | ~1.0× |
| Intel Xeon W-2235 | x86-64 | 6c/12t | 1.08× | 1.06× | 1.00× |
| AMD Ryzen Threadripper 1900X | x86-64 | 8c/16t | 1.06× | 1.01× | 1.03× |
| Broadcom BCM2712 — Cortex-A76 (RPi5) | aarch64 | 4c | **1.13×** | 1.03× | 1.01× |
| Apple M4 Pro | aarch64 | 10P+4E | 1.06–1.12× | 1.04–1.05× | 1.01× |

Peak RSS is also **≤** the sequential fold (union of 120 high-poly spheres: ~900 MB batched vs
~1080 MB sequential — `BatchBoolean` builds a deferred CsgNode tree rather than materializing every
operand at once).

## Honest note

The win is modest. If this stalled at `kintel-manifold-ops` because the speedup didn't justify
touching this path, that's a fair call — the value here is mainly (a) completing it *correctly*
(metadata + edge cases) and (b) the cross-arch data showing it's a consistent, no-downside win.
Feedback welcome. (A single-binary A/B toggle + benchmark harness used for the numbers above lives
on the `bench-toggle` branch of the fork, if reviewers want to reproduce.)

### DIFF
--- a/src/geometry/manifold/ManifoldGeometry.cc
+++ b/src/geometry/manifold/ManifoldGeometry.cc
@@ -288,6 +288,54 @@ ManifoldGeometry ManifoldGeometry::binOp(const ManifoldGeometry& lhs, const Mani
   return {mani, originalIDs, originalIDToColor, subtractedIDs};
 }
 
+ManifoldGeometry ManifoldGeometry::boolOp(
+  manifold::OpType opType,
+  const std::vector<std::shared_ptr<const ManifoldGeometry>>& operands)
+{
+  std::vector<manifold::Manifold> manifolds;
+  manifolds.reserve(operands.size());
+  for (const auto& g : operands) manifolds.push_back(g->manifold_);
+
+  // Manifold's BatchBoolean performs a balanced (and internally parallel) reduction. For Subtract
+  // it computes operands[0] - union(operands[1..]). Both are geometrically identical to folding the
+  // pairwise operators, but avoid re-processing the growing accumulator on every step.
+  auto mani = manifold::Manifold::BatchBoolean(manifolds, opType);
+
+  // Merge the id / color / subtracted-id metadata exactly as the equivalent binOp() fold would.
+  std::set<uint32_t> originalIDs;
+  std::map<uint32_t, Color4f> originalIDToColor;
+  std::set<uint32_t> subtractedIDs;
+
+  if (opType == manifold::OpType::Subtract) {
+    const auto& first = *operands.front();
+    originalIDs = first.originalIDs_;
+    originalIDToColor = first.originalIDToColor_;
+    subtractedIDs = first.subtractedIDs_;
+    for (size_t i = 1; i < operands.size(); ++i) {
+      const auto& rhs = *operands[i];
+      originalIDs.insert(rhs.originalIDs_.begin(), rhs.originalIDs_.end());
+      // Mark rhs ids as subtracted, unless they're mapped to a color (matching binOp()).
+      for (const auto id : rhs.originalIDs_) {
+        auto it = rhs.originalIDToColor_.find(id);
+        if (it != rhs.originalIDToColor_.end()) {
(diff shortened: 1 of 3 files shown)


tonuonu:
@kintel — this completes the BatchBoolean idea from your `kintel-manifold-ops` branch (5678e527b). The piece that experiment left out was the `ManifoldGeometry` metadata (`originalIDs_` / `originalIDToColor_` / `subtractedIDs_`) and the empty/edge-case handling — so it would have broken per-object colors and part-IDs. I've added those (colors render byte-identical, edge cases match `binOp()` exactly) and benchmarked across 5 machines — Intel i9/Xeon, AMD Threadripper, Cortex-A76 (RPi5), and Apple M4 Pro.

It's a **consistent but modest** win — union ~1.06–1.13× (largest on the low-core ARM), difference/intersection ~1.0×, no regression on any arch, peak RSS ≤ sequential, CI green. Completely understand if the modest speedup is why it stalled two years ago; mainly wanted to offer the finished + cross-arch-validated version. Worth landing?

tonuonu:
Friendly ping — this completes the `kintel-manifold-ops` BatchBoolean experiment (5678e527b) with the `ManifoldGeometry` metadata (colors / part-IDs / subtracted-IDs) and empty-geometry edge cases that the prototype left out, validated on 5 machines (Intel i9/Xeon, AMD Threadripper, Cortex-A76, Apple M4 Pro) — geometrically identical output, colors byte-identical, no regression, peak RSS ≤ sequential, CI green.

Completely understand if the modest win (union ~1.06–1.13×) means it's not a priority — happy to adjust it or close it if it's not wanted. Just didn't want it to quietly fall through the cracks.

Issue #2682, opened 2019-01-06
macOS file association doesn't work without the app running


dschanoeh:
Using the current development snapshot (2018.11.25), the following behavior is observed on macOS 10.14.2:

**Steps to reproduce:**
1. Ensure that the setting to show the welcome screen is active
2. Quit the app
3. Double-click on a .scad-file

**Expected behavior:** The application loads and opens the file
**Actual behavior:** The application loads and presents the welcome screen

If the application is already running, files are opened correctly when they are double-clicked.
If the welcome dialog isn't shown, the files are also opened correctly.

kintel:
This feature has been there pretty much since the start, see https://github.com/openscad/openscad/blob/master/src/EventFilter.h#L8

..but it may have broken at some point.
Do you see the same behavior on older versions of OpenSCAD?

kintel:
One observation:
* If the welcome screen is already visible, double-click on an .scad file in Finder works.

QuackingPlums:
So is this an issue or is it intended behaviour?
Right now, I'm disabling 'Show welcome screen' in the preferences because it's annoying that OpenSCAD doesn't follow normal OS conventions.

kintel:
It's an issue - we should follow normal conventions.
This was written quite some time ago. Perhaps there are better platform abstractions in Qt now that we can use to implement this properly?

PR #6906, opened 2026-07-12
Fix macOS launch crash when opening scad file via double-click

### LINKED ISSUE #5891: double-click on file to open crashes on start
**Describe the bug**

On macOS when the OpenSCAD is closed I double-click on a scad file which opens OpenSCAD - which immediately crashes.

When I open OpenSCAD, then double-click on a file it opens the file as expected.

**To Reproduce**
1. Close OpenSCAD
2. Double-click on any `.scad` file
3. macOS shows the crash report

https://gist.github.com/tcurdt/d77f86db94732ec63ca3cb19c9271fdc

**Expected behavior**

It should just start OpenSCAD and open the file.
This used to work just fine. Last version I had installed that worked was 2024.05.24 (git 8a890a9e1)

So this must have been introduced between 8a890a9e1 and 72c9919d6.

NA

**Screenshots**

NA

 - OS: macOS 15.3.2
 - System: M1 pro
 - OpenSCAD Version 2025.05.09 (git 72c9919d6) (dev build)

NA

**Additional context**

`open foo.scad` unsurprisingly gives the same result.

`/Applications/OpenSCAD.app/Contents/MacOS/OpenSCAD foo.scad` surprisingly works fine.

whosawhatsis:
### Problem Description
On macOS, double-clicking a `.scad` file to launch OpenSCAD (or using `open -a OpenSCAD file.scad` from terminal) causes it to crash on launch if the welcome screen is enabled or settings initialization happens early.

During startup, `GlobalPreferences::inst()` is queried to read default configurations. This constructs the `Preferences` singleton instance. Inside its constructor, `setup3DPrintPage()` is called, which calls `PrintService::getPrintServices()`. If remote print services are enabled, this triggers a synchronous network request using a nested `QEventLoop::exec()`. 

Spinning a nested event loop during early static local initialization allows the pending `QFileOpenEvent` (from LaunchServices/Finder) to be processed immediately on the main thread. The event filter catches it and attempts to construct a `MainWindow`, which again queries `GlobalPreferences::inst()`. Since `GlobalPreferences::inst()` has not completed its static initialization, the C++ runtime detects recursive static variable initialization re-entrancy on the same thread and aborts the application immediately via `__cxa_guard_acquire` -> `abort()`.

### Solution
- Defer `setup3DPrintPage()` (and therefore the synchronous network fetch of print services) from the `Preferences` constructor to `Preferences::showEvent(...)` so it is initialized lazily only when the settings window is actually displayed to the user.
- Add a private boolean `printPageSetupDone{false}` to track when it has been set up.
- Call `setup3DPrintPage()` inside `Preferences::on_checkBoxEnableRemotePrintServices_toggled(...)` so that toggling the setting updates the default print service combobox dynamically.

Closes #5891
Also possibly related to #2682

### DIFF
--- a/src/gui/Preferences.cc
+++ b/src/gui/Preferences.cc
@@ -163,7 +163,6 @@ Preferences::Preferences(QWidget *parent) : QMainWindow(parent)
   init();
   AxisConfig->init();
   setupFeaturesPage();
-  setup3DPrintPage();
   updateGUI();
 }
 
@@ -1102,6 +1101,7 @@ void Preferences::on_checkBoxEnableRemotePrintServices_toggled(bool checked)
 {
   S::enableRemotePrintServices.setValue(checked);
   writeSettings();
+  setup3DPrintPage();
 }
 
 void Preferences::on_comboBoxDefaultPrintService_activated(int)
@@ -1808,6 +1808,10 @@ void Preferences::keyPressEvent(QKeyEvent *e)
 
 void Preferences::showEvent(QShowEvent *e)
 {
+  if (!this->printPageSetupDone) {
+    setup3DPrintPage();
+    this->printPageSetupDone = true;
+  }
   QMainWindow::showEvent(e);
   hidePasswords();
 }
--- a/src/gui/Preferences.cc
+++ b/src/gui/Preferences.cc
@@ -623,6 +623,7 @@ void Preferences::setup3DPrintPage()
   if (it != services.end()) {
     comboBoxDefaultPrintService->setCurrentText(it->second);
   }
+  this->printPageSetupDone = true;
 }
 
 void Preferences::on_colorSchemeChooser_itemSelectionChanged()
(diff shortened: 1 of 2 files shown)


src/gui/Preferences.cc
```diff
@@ -1102,6 +1101,7 @@ void Preferences::on_checkBoxEnableRemotePrintServices_toggled(bool checked)
 {
   S::enableRemotePrintServices.setValue(checked);
   writeSettings();
+  setup3DPrintPage();
```
t-paul, line 1105: Shouldn't this also set the `printPageSetupDone` flag?

t-paul (commented):
May thanks for looking into this, I'm not sure if @kintel is around currently to test it out. If not I'm fine to just merge it anyway, just one question regarding the checkbox toggle handler...

whosawhatsis:
Good catch! I've moved the flag setting directly into `setup3DPrintPage()` so it is cleanly encapsulated and updated no matter where the function is called.

t-paul (approved):
Hmm, I was wondering if the re-init when toggling the checkbox can cause issues. After looking at what the page setup does, I think it might be even better to re-initialize when the remote service enable changes.

Lets go with that and see if it fixes the startup crash.

PR #6908, opened 2026-07-14
Fixed macOS by building openssl for both architectures

tcurdt:
Using the the local openssl is not a great idea for universal build - at least not when OpenSCAD still wants to support x86 (while Apple/homebrew is fading out support).

It might be easier to drop x86 support but for now this pulls in openssl and should bring the universal build back to life.

....at least I got it working on my machine. I can only *hope* circle CI will agree :)


t-paul:
Pushed to #6924, due to very limited budget for macOS builds, only dedicated branches (ending with `-macos`) are auto-built for macOS.

scripts/macosx-build-dependencies.sh
```diff
@@ -569,9 +572,9 @@ build_fontconfig()
   # Build each arch separately
   for arch in ${ARCHS[*]}; do
     sed -e "s,@MAC_OSX_VERSION_MIN@,$MAC_OSX_VERSION_MIN,g" -e "s,@DEPLOYDIR@,$DEPLOYDIR,g" $OPENSCADDIR/scripts/macos-$arch.txt.in > macos-$arch.txt
-    meson setup --prefix $DEPLOYDIR --cross-file macos-$arch.txt -Dtests=disabled -Dnls=disabled build-$arch
-    meson compile -C build-$arch
-    DESTDIR=install/ meson install -C build-$arch
+    PKG_CONFIG_LIBDIR="$DEPLOYDIR/lib/pkgconfig" PKG_CONFIG_PATH="$DEPLOYDIR/lib/pkgconfig" meson setup --prefix $DEPLOYDIR --cross-file macos-$arch.txt -Dtests=disabled -Dnls=disabled build-$arch
+    PKG_CONFIG_LIBDIR="$DEPLOYDIR/lib/pkgconfig" PKG_CONFIG_PATH="$DEPLOYDIR/lib/pkgconfig" meson compile -C build-$arch
+    PKG_CONFIG_LIBDIR="$DEPLOYDIR/lib/pkgconfig" PKG_CONFIG_PATH="$DEPLOYDIR/lib/pkgconfig" DESTDIR=install/ meson install -C build-$arch
```
kintel: Why do you need the PKG_CONFIG_* variables? Meson is supposed to honor the --prefix

CMakeLists.txt
```diff
@@ -615,6 +615,14 @@ target_link_libraries(OpenSCADLibInternal PUBLIC LibXml2::LibXml2)
...
 
 if(NOT CMAKE_SYSTEM_NAME STREQUAL "Emscripten")
+  if(APPLE AND NOT "$ENV{OPENSCAD_LIBRARIES}" STREQUAL "")
+    if(EXISTS "$ENV{OPENSCAD_LIBRARIES}/lib/libssl.dylib" AND EXISTS "$ENV{OPENSCAD_LIBRARIES}/lib/libcrypto.dylib")
+      set(OPENSSL_ROOT_DIR "$ENV{OPENSCAD_LIBRARIES}")
+      set(OPENSSL_INCLUDE_DIR "$ENV{OPENSCAD_LIBRARIES}/include")
+      set(OPENSSL_SSL_LIBRARY "$ENV{OPENSCAD_LIBRARIES}/lib/libssl.dylib")
+      set(OPENSSL_CRYPTO_LIBRARY "$ENV{OPENSCAD_LIBRARIES}/lib/libcrypto.dylib")
+    endif()
+  endif()
```
kintel: We should get FindPackage working instead of this hack. Do you know why that was unsuccessful for you?

scripts/release-common.sh
```diff
@@ -320,6 +320,35 @@ case $OS in
...
+          install_name_tool -delete_rpath "$EXTERNAL_LIB_DIR" "$item" 2>/dev/null || true
+
+          while IFS= read -r dep; do
+            [ -n "$dep" ] || continue
+            install_name_tool -change "$dep" "@rpath/$(basename "$dep")" "$item"
+          done < <(otool -L "$item" | awk -v prefix="$EXTERNAL_LIB_DIR/" '$1 ~ "^" prefix { print $1 }')
+        done
+
         echo "Binary created: OpenSCAD.app"
         cd $OPENSCADDIR
```
kintel: This feels like a reimplementation of similar code in macosx-build-dependencies.

scripts/macosx-build-dependencies.sh
```diff
@@ -727,6 +730,69 @@ build_harfbuzz()
...
+  if (( ${#ARCHS[@]} > 1 )); then
+    SSL_DYLIBS=()
+    CRYPTO_DYLIBS=()
+    SSL_STATIC_LIBS=()
+    CRYPTO_STATIC_LIBS=()
+    for arch in ${ARCHS[*]}; do
+      SSL_DYLIBS+=("${OPENSSL_DIR}-${arch}/install/$DEPLOYDIR/lib/libssl.3.dylib")
+      CRYPTO_DYLIBS+=("${OPENSSL_DIR}-${arch}/install/$DEPLOYDIR/lib/libcrypto.3.dylib")
+      SSL_STATIC_LIBS+=("${OPENSSL_DIR}-${arch}/install/$DEPLOYDIR/lib/libssl.a")
+      CRYPTO_STATIC_LIBS+=("${OPENSSL_DIR}-${arch}/install/$DEPLOYDIR/lib/libcrypto.a")
```
kintel: We can probably just ignore static libs - we're never linking against those for other libs either

scripts/macosx-build-dependencies.sh
tcurdt: Indeed. Worth checking without.

CMakeLists.txt
tcurdt: Sorry, it's been too long. And I am not much a `CMake` person. But I agree that this feels like a hack.

scripts/release-common.sh
tcurdt: I might have missed that.

scripts/macosx-build-dependencies.sh
tcurdt: ok

kintel:
Happy to take a look at solving my own comments if you're not super confident in this domain :)

kintel:
I believe the current state represents the minimal solution to OpenSSL. Will verify in a full snapshot build.

kintel:
It builds and runs now. I haven't tested the actual feature using OpenSSL, but please give the snapshot a spin

tcurdt:
Thanks for finishing this up and merging 🙏

tcurdt:
OpenSCAD-2026.08.27 seems to work for me 🥳

PR #6911, opened 2026-07-16
locale: complete and fix French translation

Surfoo:


locale/fr.po
t-paul: While not changed in this PR, there seem to be some broken translations like this one which seem to have accepted without checking them earlier.

t-paul (commented):
Nice, thank you. Would you mind looking through the old translation that have a `#, fuzzy` marker? It seems some of those are mismatched or obsolete like using `"%s"` instead of the `"%1"` used in the original and probably have wrong translations too.

locale/fr.po
Surfoo: you are right, I will update the translation and remove fuzzy strings

t-paul (approved):
Many thanks for also cleaning up the earlier messages 🎉. That was much more than I saw initially.

If you like, feel free to add yourself to the header of the translation file and also the About Dialog.

https://github.com/openscad/openscad/blob/583f9c0213d321ad563f9bf2d23152ec885eafc8/resources/html/AboutDialog.html#L233-L236

Surfoo:
Thanks for the suggestion, but it’s not useful for me :)
However, when will a new version be released? I think it would be useful for better integration into Linux distributions, and for improving OpenSCAD’s communication and visibility.

t-paul:
I do hope a new release will happen soon-ish as there was good progress towards a release-candidate, but lately things seem to have halted pretty much as far as I can tell. I'm not going to be part of the release work so I can't give any definitive answer.

PR #6913, opened 2026-07-18
Make layout for Viewport Control pane more compact

### LINKED ISSUE #6824: Viewport-Control pane could be a lot more compact
Viewport-Control seems to have two modes:  a wide mode and a tall mode.  Both modes are significantly larger than they really need to be.

Minimal size.

Current minimum-size "tall" layout:

Image

Note that the left half of each entry field is always blank, and the right half of each title line is always blank.  It could be about half as wide.  Alternatively, "Width", "Height", "Distance", and "FOV" could be on the same line as their input fields, and "Translation" and "Rotation" could be on the same lines as their first input fields.

In "wide" mode:

Image

Width and Height aren't bad.  They're capped at 8192 (which might eventually be a problem), but they only have about 1-2 digits worth of space wasted.

On the other hand, Translation, Rotation, Distance, and FOV have space for about 18 digits.  They could be less than half as wide without sacrificing anything practical.  (Rotation and FOV could be much smaller, since they are practically limited to three-digit values, but making them smaller than Translation doesn't help.)

freastrid-a11y:
In 'tall' and 'wide' mode of the Viewport Control the input fields have been made a lot smaller, there was a lot of wasted space.

The dashes have also been removed in 'Viewport-Control' and 'Error-Log'

Fixes #6824

### DIFF
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -3772,11 +3772,11 @@ void MainWindow::setupDocks()
     {editorDock, _("&Editor")},
     {consoleDock, _("&Console")},
     {parameterDock, _("C&ustomizer")},
-    {errorLogDock, _("Error-&Log")},
+    {errorLogDock, _("Error &Log")},
     {animateDock, _("&Animate")},
     {fontListDock, _("&Font List")},
     {colorListDock, _("C&olor List")},
-    {viewportControlDock, _("&Viewport-Control")},
+    {viewportControlDock, _("&Viewport Control")},
     {aiDock,_("&AI Chat")}
   };
   // clang-format off
--- a/src/gui/ViewportControl.cc
+++ b/src/gui/ViewportControl.cc
@@ -30,10 +30,25 @@ ViewportControl::ViewportControl(QWidget *parent) : QWidget(parent)
 
 void ViewportControl::initGUI()
 {
+  QFontMetrics metrics(this->font());
+#if QT_VERSION >= QT_VERSION_CHECK(5, 11, 0)
+  int charWidth = metrics.horizontalAdvance("0");
+#else
+  int charWidth = metrics.width("0");
+#endif
+
+  int dynamicMinWidth = (charWidth * 8) + 35;
+  int dynamicMaxWidth = (charWidth * 12) + 35;
+
   auto spinDoubleBoxes = this->groupBoxAbsoluteCamera->findChildren<QDoubleSpinBox *>();
   for (auto spinDoubleBox : spinDoubleBoxes) {
     spinDoubleBox->setMinimum(-DBL_MAX);
     spinDoubleBox->setMaximum(+DBL_MAX);
+
+    spinDoubleBox->setSizePolicy(QSizePolicy::Preferred, QSizePolicy::Fixed);
+    spinDoubleBox->setMinimumWidth(dynamicMinWidth);
+    spinDoubleBox->setMaximumWidth(dynamicMaxWidth);
+
     connect(spinDoubleBox, QOverload<double>::of(&QDoubleSpinBox::valueChanged), this,
             &ViewportControl::updateCamera);
   }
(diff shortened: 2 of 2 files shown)


freastrid-a11y:
image
image
image

jordanbrown0:
I don't have time right now to do a full review or try it out, but looks like it's heading in the right direction.  It looks like the windows are still at their original widths.  Can you resize them snugly around the contents? What are the largest values that will with in the Translation fields?

t-paul:
Agreed looks good, and moving in smaller steps is fine.

On Linux (my Debian/Testing with Gnome), there's 2 minor things we maybe should look at before merge:

1) It's showing different spinner styles for Aspect and Details which is strange.
2) The Details spinners are a bit too tiny to show values >= 1000. Maybe there's a way to allow them to increase size a little bit if there's room?
 


Minimum in tall mode:

freastrid-a11y:
> I don't have time right now to do a full review or try it out, but looks like it's heading in the right direction. It looks like the windows are still at their original widths. Can you resize them snugly around the contents? What are the largest values that will with in the Translation fields?

The original screenshots were just showing OpenSCAD remembering the old, wider default window size. I've manually resized the panel to fit snugly around the new compact layout, and I've attached some updated screenshots here

image
image

Regarding the Translation fields: I tested them, and they comfortably fit an 8-digit number (e.g., -99999.9) before any text clipping occurs. Let me know if they should be a bit larger perhaps

freastrid-a11y:
> Agreed looks good, and moving in smaller steps is fine.
> 
> On Linux (my Debian/Testing with Gnome), there's 2 minor things we maybe should look at before merge:
> 
>     1. It's showing different spinner styles for Aspect and Details which is strange.
> 
>     2. The Details spinners are a bit too tiny to show values >= 1000. Maybe there's a way to allow them to increase size a little bit if there's room?
> 
> 
> 
> 
> Minimum in tall mode:
> 

I have now set a minimumWidth of 85px and a maximumWidth of 120px for all the entry fields across both Aspect Ratio and Details. This should give them enough room to easily fit values >= 1000 without triggering the compact spinner style, while keeping everything uniformly sized.

I am testing this on Ubuntu so it looks perfectly consistent on my end now, but please let me know if this resolves the spinner style differences on your Debian/Gnome setup

t-paul:
Qt Widget Layouts are just annoying. It looks like the spinner decides based on the size if it shows the big or small buttons while totally ignoring that showing the numbers would be so much more helpful. It may also depend a lot on the screen resolution + font size settings. I think the difference comes from Qt insisting on defining things in pixel instead of something that acknowledges DPI.

https://github.com/user-attachments/assets/54edf487-f9e2-400b-94b7-47bf582db6dd

freastrid-a11y:
> Qt Widget Layouts are just annoying. It looks like the spinner decides based on the size if it shows the big or small buttons while totally ignoring that showing the numbers would be so much more helpful. It may also depend a lot on the screen resolution + font size settings. I think the difference comes from Qt insisting on defining things in pixel instead of something that acknowledges DPI.
> Screencast.From.2026-07-18.22-30-46.mp4

Ah, I see.. I tried to look for a fix and came upon `QFontMetrics`. Could that work, or should this just be left as is?
https://doc.qt.io/qt-6/qfontmetrics.html

t-paul:
So far I did not really win my layout fights with Qt, so I can't say for sure. It might be worth a try, the +/- button behavior could still make it a bit more difficult than one would hope.

freastrid-a11y:
> So far I did not really win my layout fights with Qt, so I can't say for sure. It might be worth a try, the +/- button behavior could still make it a bit more difficult than one would hope.

Instead of using hardcoded pixels, the UI now calculates the minimumWidth and maximumWidth dynamically based on the width of the user's actual font (specifically, it sizes it to comfortably fit 8 to 12 characters plus the spin arrows).

I haven't tested it on a Debian/Gnome setup though since it turned out to be a bit of hassle running a virtual environment

t-paul (approved):
Looks nice, thank you! I'm getting about 5 digits if space permits and the layout switch works fine too. Lets go with that, it's a good improvement.

PR #6914, opened 2026-07-18
Add remove() module as syntactic sugar for creating differences.

t-paul:
Proposal for a `remove()` node which provides a way of writing differences as flat list. This is done by rewriting the CSG tree so in the final tree, the `remove()` node is not present anymore. Consecutive positive nodes are collected into a `union()` which is then placed as first element of a `difference()` that includes all following negative nodes.

This came out of a discussion with a teacher working with a blind student finding that a `difference()` clause is difficult to navigate when not having the option to see the whole structure on a screen.

Thoughts? Questions?

<details><summary>Source code from the screenshot</summary>

```openscad
$fa = 2; $fs = 0.2;

module test() {

cube(1);
color("green") cube(10, center = true);
#remove() cylinder(h = 20, d = 6, center = true);
%remove() rotate([0, 90, 0]) cylinder(h = 20, d = 6, center = true);
#remove() rotate([90, 0, 0]) cylinder(h = 20, d = 6, center = true);

color("cyan") sphere(5);
color("blue") cylinder(h = 15, d = 2, center = true);
remove() cube([20, 20, 1], center = true);

color("red") translate([0, 20, 0]) cylinder(h = 3, r = 2);

}

test();
```
</details>

The generated CSG tree looks like this:

```openscad
group() {
  difference() {
    union() {
      cube(size = [1, 1, 1], center = false);
      color([0, 0.501961, 0, 1]) {
        cube(size = [10, 10, 10], center = true);
      }
    }
#    cylinder($fn = 0, $fa = 2, $fs = 0.2, h = 20, r1 = 3, r2 = 3, center = true);
%    multmatrix([[0, 0, 1, 0], [0, 1, 0, 0], [-1, 0, 0, 0], [0, 0, 0, 1]]) {
      cylinder($fn = 0, $fa = 2, $fs = 0.2, h = 20, r1 = 3, r2 = 3, center = true);
    }
#    multmatrix([[1, 0, 0, 0], [0, 0, -1, 0], [0, 1, 0, 0], [0, 0, 0, 1]]) {
      cylinder($fn = 0, $fa = 2, $fs = 0.2, h = 20, r1 = 3, r2 = 3, center = true);
    }
  }
  difference() {
    union() {
      color([0, 1, 1, 1]) {
        sphere($fn = 0, $fa = 2, $fs = 0.2, r = 5);
      }
      color([0, 0, 1, 1]) {
        cylinder($fn = 0, $fa = 2, $fs = 0.2, h = 15, r1 = 1, r2 = 1, center = true);
      }
    }
    cube(size = [20, 20, 1], center = true);
  }
  color([1, 0, 0, 1]) {
    multmatrix([[1, 0, 0, 0], [0, 1, 0, 20], [0, 0, 1, 0], [0, 0, 0, 1]]) {
      cylinder($fn = 0, $fa = 2, $fs = 0.2, h = 3, r1 = 2, r2 = 2, center = false);
    }
  }
}

```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1015,6 +1015,8 @@ set(CORE_SOURCES
   src/core/ContextMemoryManager.cc
   src/core/CsgOpNode.cc
   src/core/CurveDiscretizer.cc
+  src/core/RemoveNode.cc
+  src/core/RemoveNodeRewriter.cc
   src/core/DrawingCallback.cc
   src/core/EvaluationSession.cc
   src/core/Expression.cc
--- a/src/core/CSGTreeEvaluator.h
+++ b/src/core/CSGTreeEvaluator.h
@@ -49,6 +49,7 @@ public:
   }
 
 private:
+  using ChildList = std::list<std::shared_ptr<const AbstractNode>>;
   void addToParent(const State& state, const AbstractNode& node);
   void applyToChildren(State& state, const AbstractNode& node, OpenSCADOperator op);
   std::shared_ptr<CSGNode> evaluateCSGNodeFromGeometry(State& state,
@@ -57,10 +58,8 @@ private:
                                                        const AbstractNode& node);
   void applyBackgroundAndHighlight(State& state, const AbstractNode& node);
 
-  using ChildList = std::list<std::shared_ptr<const AbstractNode>>;
-  std::map<int, ChildList> visitedchildren;
-
 protected:
+  std::map<int, ChildList> visitedchildren;
   const Tree& tree;
   GeometryEvaluator *geomevaluator;
   std::shared_ptr<CSGNode> rootNode;
--- a/src/core/CsgOpNode.cc
+++ b/src/core/CsgOpNode.cc
@@ -25,6 +25,7 @@
  */
 
 #include "core/CsgOpNode.h"
+#include "core/RemoveNode.h"
 
 #include <cassert>
 #include <memory>
@@ -59,6 +60,13 @@ static std::shared_ptr<AbstractNode> builtin_intersection(const ModuleInstantiat
   return children.instantiate(std::make_shared<CsgOpNode>(inst, OpenSCADOperator::INTERSECTION));
 }
 
+static std::shared_ptr<AbstractNode> builtin_remove(const ModuleInstantiation *inst, Arguments arguments,
+                                                    const Children& children)
+{
+  Parameters parameters = Parameters::parse(std::move(arguments), inst->location(), {});
+  return children.instantiate(std::make_shared<RemoveNode>(inst));
+}
+
 std::string CsgOpNode::toString() const
 {
   return this->name() + "()";
@@ -90,5 +98,10 @@ void register_builtin_csgops()
   Builtins::init("intersection", new BuiltinModule(builtin_intersection),
                  {
                    "intersection()",
+                 });
+
+  Builtins::init("remove", new BuiltinModule(builtin_remove),
+                 {
+                   "remove()",
                  });
 }
--- a/src/core/LocalScope.cc
+++ b/src/core/LocalScope.cc
@@ -10,6 +10,7 @@
 
 #include "core/Assignment.h"
 #include "core/ModuleInstantiation.h"
+#include "core/RemoveNodeRewriter.h"
 #include "core/UserModule.h"
 #include "core/function.h"
 #include "core/node.h"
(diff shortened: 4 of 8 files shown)


jordanbrown0:
Note that this is similar to BOSL2's diff(), except that BOSL2 allows positive-ness and negative-ness to be buried deeper in the hierarchy. Perhaps BOSL2's diff() is a better answer, under the general rule that stuff that can be done in a library should be. 

I can't say that I'm thrilled by rewriting the CSG tree, nor by imposing sequencing on a union's children to implicitly separate them into differences and unions.

I think I would prefer if the CSG tree was constructed normally, and the magic was done at CSG processing time, so that nodes that support it would "peek" at their children to see how to process them. (But that's less compatible for CSG consumers.)

Does this apply to all union-equivalent constructs?  E.g. `translate(){...}`? To the second-and-later children of difference(), but inverted? Or only to the top levels of modules?

What about user modules and children()? It could and perhaps should apply to children() with no args and to children(range), but couldn't apply to `for(i =[0:$children-1]) children(i);`.

Your example has the same number of nodes before and after rewriting. (4x remove vs 2x difference plus 2x union.) The only case where it has fewer nodes before rewriting is when there's more than one positive child and exactly one negative child.

Comparing the original with the generated CSG isn't really fair, because generated CSG is always verbose.  Here's a fairer comparison:
```
$fa = 2; $fs = 0.2;

module test() {

  difference() {
    union() {
      cube(1);
      color("green") cube(10, center = true);
    }
    #cylinder(h = 20, d = 6, center = true);
    %rotate([0, 90, 0]) cylinder(h = 20, d = 6, center = true);
    #rotate([90, 0, 0]) cylinder(h = 20, d = 6, center = true);
  }

  difference() {
    union() {
      color("cyan") sphere(5);
      color("blue") cylinder(h = 15, d = 2, center = true);
    }
    cube([20, 20, 1], center = true);
  }

  color("red") translate([0, 20, 0]) cylinder(h = 3, r = 2);

}

test();
```
This has the same number of words, but more lines and more hierarchy.

I wonder whether A-B+C-D should group as (A-B)+(C-D) as you have done, or as ((A-B)+C)-D as ordinary arithmetic would group it.

Net: I don't like that it's kind of alien to OpenSCAD design. However, I have to admit that a sequence of add-this, subtract-that is pretty human-natural.

t-paul:
> Comparing the original with the generated CSG isn't really fair, because generated CSG is always verbose.

That was not meant as comparison but as explanation of the rewrite. It's just the CSG that is produced by the example.

The use case it is trying to cover (on request as noted above) is making it more accessible for people who can't *see* the source code, literally as in being blind and using a screen reader or using a braille terminal.

jordanbrown0:
Sure, but the point is to make the source be in some sense simpler, right? So we should look at how much simpler it is - or, rather, in what ways it's simpler.

If I've got my head wrapped properly around the proposal, it is that a sequence of non-remove nodes followed by a sequence of remove nodes is rewritten into a difference of a union of the non-remove nodes, and the remove nodes.  That is,
```
A();
B();
...
remove() C();
remove() D();
...
```
is rewritten as

```
difference() {
  union() {
    A();
    B();
    ...
  }
  C();
  D();
  ...
}
```
... right?

* For this case, the number of words is the same.
* The "traditional" representation probably has four more lines, because the union, difference, and close braces are probably on separate lines, while the remove is probably on the same line as the child shape.
* To understand either, it seems like you must get all of A,B,C,D into your head at pretty much the same time.
* With the remove scheme, there are no mechanical tools to help you understand the boundaries.  No brace matching, no block collapsing. Hypothetically, you could tell a screen reader "read this block", and it could do so, eliding collapsed blocks.  That can't work here, at least not without the screen reader understanding semantics.

To make the **system** be understandable, there needs to be a consistent and comprehensible model for what the remove module does and how it interacts with other modules. Making one piece simpler, but making the big picture more complex, is not usually a win. If you just drop a remove() into the middle of a program, it should be easy (or at least straightforward and unambiguous) to understand what it will do. I'm not saying that there can't be such a model here, but it needs to be written out.  Your example is for the top level of a module, which is an implied union. Does the same work for other implied (or explicit) unions? What about blocks that are not implied unions - difference, intersection, and user module children?

Even in the "simple" case, consider:
```
A();
B();
...
remove() C();
remove() D();
...
```
Add a remove entry between A and B:
```
A();
remove() A1();
B();
...
remove() C();
remove() D();
...
```
Is the resulting change - that A no longer has C and D removed from it - an obvious result?  Or do the other way: when you delete the remove() A1(), is it obvious that A will have C and D removed?  Adding and deleting remove entries is changing the relationship between **other** nodes, which doesn't seem obvious. In the traditional representation you can't normally produce such a side effect without messing with braces.  (Difference and user modules with positional semantics excepted.)

PR #6917, opened 2026-07-19
Give all Menu items a Shortcut

### LINKED ISSUE #6876: Give all Menu items a Shortcut
While discussing hotkeys in #6875 , @jordanbrown0 pointed out that it would be best for all menu items to have shortcuts. When moving between programs, it would be very advantageous to my workflow if I can at least access the menus via keyboard easily.

Per discussion in #6875 : https://github.com/openscad/openscad/issues/6875#issuecomment-4770782652
> One easy thing that we should do is to ensure that every menu item has an & shortcut, e.g. `&Animate`, so that you can get to it via, e.g., Alt+W, A.  Ideally we should arrange that every menu item in all menus has an & shortcut, so that all menu items are only two keystrokes away.  That's much easier than full-powered hot keys, since it only needs to be unique within that menu.

Nothing really, just kinda bouncing off of y'all's ideas

**Additional context**

/assign-me

freastrid-a11y:
Also replaced the shortcut for some menu items,
who were sharing the same shortcut (under the same drop-down)

Fixes #6876

### DIFF
--- a/src/gui/MainWindow.ui
+++ b/src/gui/MainWindow.ui
@@ -730,7 +730,7 @@
   </action>
   <action name="fileActionNewWindow">
    <property name="text">
-    <string>New Window</string>
+    <string>New &amp;Window</string>
    </property>
   </action>
   <action name="fileActionOpen">
@@ -746,7 +746,7 @@
   </action>
   <action name="fileActionOpenWindow">
    <property name="text">
-    <string>Open in New Window</string>
+    <string>Open in New Windo&amp;w</string>
    </property>
   </action>
   <action name="fileActionSave">
@@ -770,12 +770,12 @@
   </action>
   <action name="fileActionSaveACopy">
    <property name="text">
-    <string>Save a Copy...</string>
+    <string>Save a C&amp;opy...</string>
    </property>
   </action>
   <action name="fileActionSaveAll">
    <property name="text">
-    <string>Save All</string>
+    <string>S&amp;ave All</string>
    </property>
   </action>
   <action name="fileActionReload">
@@ -898,7 +898,7 @@
   </action>
   <action name="editActionNextTab">
    <property name="text">
-    <string>Show Next Tab</string>
+    <string>Show &amp;Next Tab</string>
    </property>
    <property name="shortcut">
     <string>Ctrl+Tab</string>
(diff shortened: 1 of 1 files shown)


t-paul:
I believe the mnemonics need to be unique per menu, or is there a way to jump to the 2nd one somehow? I have not seen a way of making mnemonics case sensitive.

For `File` those 3 have duplicate letters:
* Open in New Windo«w»
* Save a C«o»py
* S«a»ve All

MichaelAtOz:
> is there a way to jump to the 2nd one somehow? I have not seen a way of making mnemonics case sensitive.

On Windows, repeating the character goes to the next menu item.
You can see this in OpenSCAD with:
Alt-F (file menu opens), 
x (export menu opens, with duplicate letters - s & o), 
s (goes to STL-ascii, but doesn't 'actuate', need to press enter to make it happen)
s (then goes to STL-binary, ditto, ditto)

Microsoft: "Access keys aren't case sensitive."

Note sub-menus can be used to allow different set of mnemonics, Edit/Copy-Viewport (set of) could be a candidate.

I'd suggest that changing existing mnemonics would need a very good reason, anyone using them will have muscle memory, transition would be painful.

freastrid-a11y:
> > is there a way to jump to the 2nd one somehow? I have not seen a way of making mnemonics case sensitive.
> 
> On Windows, repeating the character goes to the next menu item. You can see this in OpenSCAD with: Alt-F (file menu opens), x (export menu opens, with duplicate letters - s & o), s (goes to STL-ascii, but doesn't 'actuate', need to press enter to make it happen) s (then goes to STL-binary, ditto, ditto)
> 
> Microsoft: "Access keys aren't case sensitive."
> 
> Note sub-menus can be used to allow different set of mnemonics, Edit/Copy-Viewport (set of) could be a candidate.
> 
> I'd suggest that changing existing mnemonics would need a very good reason, anyone using them will have muscle memory, transition would be painful.

Okay seeing that repeating characters goes to the next menu item, I won't be changing any existing mnemonics

freastrid-a11y:
Now all menu items have a shortcut, some menu items share the same shortcut but they are still accessible as discussed. I have kept all existing mnemonics as they were

freastrid-a11y:
image
image
image
image
image
image

t-paul (approved):
Aha, nice, thank you! Lets go with that then.

PR #6925, opened 2026-07-21
Add chat export/import and context window management

dasbidyendu:
Implement sliding window context management to limit conversation history based on configurable context_limit parameter (default 10 turns). Add OpenSCAD reference documentation to system prompt including 3D/2D primitives, transforms, and available tools.

Add chat history export/import in JSON format and copy-as-markdown functionality via new menu button. Exclude context management parameters (system_prompt, default_prompt, context_limit) from API payloads.

### DIFF
--- a/src/core/AIService.cc
+++ b/src/core/AIService.cc
@@ -150,6 +150,13 @@ void AIService::chatCompletionStream(std::vector<ChatMessage>& history, ChunkCal
     sys_prompt = config.parameters["system_prompt"].get<std::string>();
   }
 
+  int context_limit = 10;
+  if (config.parameters.contains("context_limit") &&
+      config.parameters["context_limit"].is_number_integer()) {
+    context_limit = config.parameters["context_limit"].get<int>();
+    if (context_limit < 1) context_limit = 1;
+  }
+
   bool already_has_system = false;
   if (!history.empty() && history[0].role == "system") {
     already_has_system = true;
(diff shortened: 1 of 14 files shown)


tests/data/scad/issues/issue2342.scad
t-paul, line 1: Can you please remove this and the SVG files from the PR. I believe those appeared because of an earlier change and need to be either removed once, or added to the `.gitignore`.

src/gui/ai/ChatWidget.cc
```diff
@@ -267,6 +267,14 @@ void ChatWidget::onSendPressed()
     return;
   }
 
+  if (prompt.length() > 100000) {
+    addMessage(prompt, true);
+    addMessage(
+      tr("Error: Prompt size exceeds the limit of 100,000 characters. Please shorten your prompt."),
```
t-paul, line 298: Ah, good point, translations. We don't use the Qt native translation, this is instead mapped to gettext via `qtgettext.h` and `printutils.h`.

The incantation for marking a string as translatable is `_("...")` like here:
https://github.com/openscad/openscad/blob/587c2ee85af991d9589596b751ee4d19c1327eeb/src/gui/MainWindow.cc#L865

src/core/AIService.cc
```diff
@@ -158,11 +174,61 @@ void AIService::chatCompletionStream(std::vector<ChatMessage>& history, ChunkCal
...
+    "**Import/Use**: use <file.scad> (functions/modules only), include <file.scad> (full), "
+    "import(\"file.stl\")\n\n"
+    "### Available Tools\n"
+    "- **get_editor_code()**: Returns the current contents of the active editor tab.\n"
+    "- **set_editor_code({\"code\": \"...\"})**: Proposes code changes for user review in a "
+    "side-by-side diff dialog. "
+    "Always use this instead of pasting code in chat.\n"
+    "- **trigger_preview()**: Renders the current editor code in the 3D viewport. "
+    "Automatically postponed if code changes are pending review.";
+  ai_history.push_back({"system", ref_doc});
```
t-paul: This creates a 2nd system prompt. Looks like not all chat templates like that, e.g.:
```
{%- if message.role == "system" %}
  {%- if not loop.first %}
    {{- raise_exception('System message must be at the beginning.') }}
  {%- endif %}
  ...
```

dasbidyendu: Ohh , I'll take a look thank you for pointing out.

tests/data/scad/issues/issue2342.scad
t-paul: We can't just ignore the whole folder. I'll fix that up after merge...

t-paul (approved):

PR #6926, opened 2026-07-21
Fix autoreload when included files contain parse errors

### LINKED ISSUE #6870: Parse error in include file blocks automatic reload-and-rerun
**Describe the bug**
If you have a parse error (but not an execution warning) in an include&lt;> file, a subsequent change to the include file will not trigger a new run.

**To Reproduce**
1. main.scad:

```
include <lib.scad>
```
2. lib.scad:

```
echo("lib");
```
3.  Load main.scad.
4. F5.
5. Observe "lib" message.
6. With an external editor, change `"lib"` to `"foo"` and save.
7. Observe automatic re-run with "foo" message.
8. With an external editor, remove the close parenthesis from lib.scad and save.
9. Observe automatic re-run and parse error.
10. With an external editor, restore the close parenthesis and save.
11.  Observe no automatic re-run.
12. F5.
13. Observe "foo" message.
14. With an external editor, change `"foo"` to `"lib"` and save.
15. Observe automatic re-run with "lib" message.

**Expected behavior**
Re-run on change.

 - OS: Windows 11
 - System: Intel PC 64-bit
 - OpenSCAD Version: 2026.06.08, openscad.org

ishu599:
Fixes #6870

Root cause

When parsing failed due to a syntax error in an included file, MainWindow::parsedFile was reset to nullptr. Since the auto-reload mechanism depends on parsedFile->includesChanged() to detect changes in included files, further modifications to those files were no longer detected. As a result, fixing and saving an included file did not trigger automatic recompilation.

Changes
Preserve and consistently update MainWindow::parsedFile with the parsed SourceFile returned by parseDocument().
Delay updating SourceFileCache::includes_mtime until parsing completes successfully, preventing dependency timestamps from being committed after failed parses.
Result

Auto Reload now correctly detects changes to included files after a parse error. Once the syntax error is fixed and the included file is saved, OpenSCAD automatically recompiles without requiring a manual reload or changes to the top-level file.

### DIFF
--- a/src/core/SourceFile.cc
+++ b/src/core/SourceFile.cc
@@ -116,13 +116,13 @@ time_t SourceFile::includesChanged() const
 
 time_t SourceFile::include_modified(const std::string& filename) const
 {
-  struct stat st;
+struct stat st;
 
   if (StatCache::stat(filename, st) == 0) {
-    return st.st_mtime;
-  }
+      return st.st_mtime;
+    }
 
-  return 0;
+return 0;
 }
 
 /*!
--- a/src/core/SourceFile.cc
+++ b/src/core/SourceFile.cc
@@ -116,13 +116,13 @@ time_t SourceFile::includesChanged() const
 
 time_t SourceFile::include_modified(const std::string& filename) const
 {
-struct stat st;
+  struct stat st;
 
   if (StatCache::stat(filename, st) == 0) {
-      return st.st_mtime;
-    }
+    return st.st_mtime;
+  }
 
-return 0;
+  return 0;
 }
 
 /*!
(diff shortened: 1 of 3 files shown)


t-paul (changes requested):
Can you check why a number of use<> test cases fail. This looks related to the change.

The formatting changes will likely trigger failures too, have a look at `doc/hacking.md` for some notes how to run the beautify script, or use `pre-commit` to automate this step.

ishu599:
@t-paul i have made the changes, now the pr is passing the use cases, also i ran the beautify script.
please check and review.

ishu599:
I have updated the code to pass the use tests. And used beautify also. 
Now passing all tests. 

Can you approve from your end?

t-paul:
Don't worry about the beautify fail, we can handle that when merging. I'll have a look at the changes.

t-paul:
What's the reasoning for removing all the comments? Are those not correct or relevant anymore? As proven by the issue itself, the topic is a bit more complex than it's obvious at the first glance, so having a description of the logic via comments seems useful here.

PR #6931, opened 2026-07-26
Catch evaluation exceptions on the command line

chotchki:
While fuzzing OpenSCAD I ran across crashes from the search function and I commented on bug #5017 since that bug is what I hit in my fuzzing. However that bug is a symptom of the missing exception handling for the command line vs the GUI execution. In it I mentioned I would open a separate pull request and this is it to fix the command line crash abort. The following lines are Claude's write up. The system crash reporter is will happily scream repeatedly about this.

-------

An exception escaping evaluation aborts the headless process instead of being reported. The GUI already handles this: MainWindow wraps its compile in catch(HardWarningException) / catch(std::exception) / catch(...), routing the latter two to UnknownExceptionCleanup, which logs "Compilation aborted by exception: ...". openscad_main's equivalent try block has only the first clause, so anything else reaches the default terminate handler.

Reproducing with issue #5017, which throws out of search():
```console
  $ printf 'echo(search("a", [[1, 2]]));\n' > crash.scad
  $ openscad -o out.stl crash.scad; echo "exit=$?"
  libc++abi: terminating due to uncaught exception of type std::bad_variant_access
  exit=134
```
The same file in the GUI prints "ERROR: Compilation aborted by exception: bad_variant_access" and carries on. On macOS the abort also raises the system crash reporter, so a scripted build looks like a hard failure of the tool.

This adds the two missing clauses, with the same messages the GUI uses so the two paths report identically. HardWarningException stays first, since it derives from std::exception via EvaluationException/std::runtime_error and must keep matching its own clause.

Independent of the search() fix in #5017: it makes any escaped evaluation exception a diagnosed error and rc=1 rather than SIGABRT.

### DIFF
--- a/src/openscad.cc
+++ b/src/openscad.cc
@@ -1181,6 +1181,12 @@ int openscad_main(int argc, char **argv)
       }
     } catch (const HardWarningException&) {
       rc = 1;
+    } catch (const std::exception& ex) {
+      LOG(message_group::Error, "Compilation aborted by exception: %1$s", ex.what());
+      rc = 1;
+    } catch (...) {
+      LOG(message_group::Error, "Compilation aborted by unknown exception");
+      rc = 1;
     }
 
     if (deps_output_file) {


xypwn:
IMO something like

```
Internal compiler error: <error message>
Please open an issue on <github link>
```

would be better, since it provides more context and OpenSCAD is also used by non-technical people and they may not know what an exception means.

EDIT: As Jordan Brown mentioned, though, it would probably be better to still pass down the exception for diagnostics/debugging (so maybe error message plus throw?)

jordanbrown0:
I think this is a mistake. The goal, especially in the CLI, is not "never crash".  The goal is "don't have bugs". When we have a bug, we want to preserve as much forensic data as possible, so that it will be easier to diagnose the bug and fix it.

xypwn:
@jordanbrown0 see my edit

jordanbrown0:
Core dumps are the gold standard in post mortem forensic analysis, and the closer the core dump is to the failure the easier it is to analyze it.

When we have a bug, an unexpected error, what error message are we going to add that tells the user something useful?

jordanbrown0:
I suppose that there is some value in saying "We don't know what the heck is going on. You might have made a mistake, but if so we didn't handle it.  Please report a bug."

But if the cost of that value is even a small reduction in forensic data collected, I think it would be a mistake.

t-paul:
I agree with @jordanbrown0 on this. For command line, some additional info might be ok, but completely destroying the information about what went wrong is not a good idea.

A bigger topic would be looking at one of those "error reporting" libraries that actually keep the information. That was discussed before and is probably needed separately by platform, but that would cover both cases 1) keeping the user in the loop and 2) providing enough information to look at the issue.

t-paul (changes requested):
Giving additional information might be useful, although it's not 100% safe to assume logging still works in this context and does not just throw a different exception masking the root cause.

I would not like to see a change accepted that just exits with an error status in this case.

chotchki:
I'm happy to make more changes, I just want to also point to that I took the approach to handling the command line from how the GUI does it here: https://github.com/openscad/openscad/blob/04834adb1ba3c9a6e4160747c91cbadbce9c37b3/src/gui/OpenSCADApp.cc#L47

would logging a stacktrace be better? I'm thinking boost::stacktrace to handle the platform differences.

t-paul:
Having the GUI behave differently seems to make sense. Instead of just the window disappearing, the dialog gives at least some indication of an issue, even though it's eating most of the information currently.

Of course even better is to display the actual issue in a way that could be reported as bug with meaningful extra data. I have not looked at boost::stacktrace yet, but as we use boost already, that sounds promising. I don't know how deep that rabbit hole is and if one of those more high-level libraries could be integrated easily. It might be worth looking around what options there are at this point - and also how platform support looks like. Adding new libraries can be quite a challenge across all the different build environments.

jordanbrown0:
Stack back traces are better than messages (which are often useless).  Core dumps are better.  Exercising the problem in a debugger is best, but of course requires being able to replicate it.

Unfortunately, Windows doesn't have core dumps.

jordanbrown0:
GUI crashes also risk losing data from unsaved editor sessions, which aren't an issue for the CLI.

PR #6935, opened 2026-07-27
ci: cleanup test build matrix

nomike:
- No longer test qt5 builds on Ubuntu 24.04 as we've switched to qt6 being the default.
- Test macos on arm64 instead of x64 as it is usually much faster and also more likely to be the architecture used by users.


nomike:
Sorry for the noise.

PR #6939, opened 2026-08-03
Fix #3732: only create backups after document changes

### LINKED ISSUE #3732: Only backup untitled.scad when changed
I noticed very high CPU use, traced it to SpiderOakONE, my online backup.
It was saving a file repeatedly, `unsaved-backup-XEhSCvfw.scad`.
I had pasted the animation from the [forum](http://forum.openscad.org/Fit-an-rectangle-into-another-rectangle-td32366.html).
Each step of the animation was saving a backup of the (unchanged) unsaved `untitled.scad`.
The animation had been running for hours...luckily it does de-duplication so was not actually backing up the 1k, but it did keep over 4000 versions of metadata   :|

OpenSCAD should flag that it has been backed-up, and not repeat the backup and unflag it if it is subsequently edited. This will also speed-up animation of unsaved files.

lugoge:
## Summary

Fixes #3732.

Backups were triggered whenever `isContentModified()` returned true.
For unsaved files this state remains true until a save occurs, which can
cause repeated backup creation during repeated compilations.

This change tracks whether the document has changed since the last backup
and only creates a backup when new edits have occurred.

## Testing

Tested manually:

1. Created a new untitled document.
2. Modified the contents.
3. Triggered preview/compile.
4. Verified that a backup file was created.
5. Triggered repeated previews without modifying the document.
6. Verified that no additional backup files were created.
7. Modified the document again.
8. Verified that the next preview created a new backup.

## Notes

The original dirty-state (`isContentModified()`) behavior remains unchanged.
The change only affects backup creation logic.

Before:
Screenshot 2026-08-02 at 10 37 10 PM

After:
Screenshot 2026-08-02 at 10 38 11 PM

### DIFF
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -737,7 +737,9 @@ void MainWindow::compile(bool reload, bool forcedone)
       if (GlobalPreferences::inst()->getValue("advanced/consoleAutoClear").toBool()) {
         this->console->clear();
       }
-      if (activeEditor->isContentModified()) saveBackup();
+      if (activeEditor->isContentModified() && backupPending) {
+        saveBackup();
+      }
       parseTopLevelDocument();
       didcompile = true;
     }
@@ -1243,7 +1245,8 @@ void MainWindow::saveBackup()
     LOG(message_group::UI_Warning, "Failed to create backup file");
     return;
   }
-  return writeBackup(this->tempFile);
+  writeBackup(this->tempFile);
+  backupPending = false;
 }
 
 void MainWindow::on_fileActionSave_triggered()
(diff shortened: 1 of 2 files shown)


src/gui/MainWindow.cc
```diff
@@ -785,7 +737,9 @@ void MainWindow::compile(bool reload, bool forcedone)
       if (GlobalPreferences::inst()->getValue("advanced/consoleAutoClear").toBool()) {
         this->console->clear();
       }
-      if (activeEditor->isContentModified()) saveBackup();
+      if (activeEditor->isContentModified() && backupPending) {
+        saveBackup();
+      }
```
t-paul, line 742: I believe the `backupPending` state is tracked one level too high. A MainWindow can have multiple tabs with unsaved content. Running sequential previews in different tabs without additional changes in-between would skip the 2nd preview that should happen.

t-paul (changes requested):

PR #2792, opened 2019-02-16
Optimize away group nodes with <=1 children from node idStrings

thehans:
Note: I had a PR already for this, from my master, but recreated it on its own branch now.  I rebased my fork's master(force push) to match openscad which auto-closed the previous PR #2766 since it has no changes now.

This change reduces duplicate cache entries which can be caused by empty or single child groups.

I also changed idPrefix setting to be a preprocessor define, and reverted the cache debug PRINTB/PRINTDB statements to their previous 40character limit.

Here is some sample code
```
module expensive(test, $fn=40) {
    if (test) 
        echo(test);
    difference()  {
        sphere(10);
        translate([5,0,0]) sphere(10);
    }
}

module test2() {
    expensive();
}

expensive();
expensive("hi");
test2();
if(1) test2();
```

Before
```CGAL Cache insert: difference(){sphere($fn=40,$fa=12,$fs=2,r=10);multmatrix([[1,0,0,5],[0,1,0,0],[0,0,1,0],[0,0,0,1]]){sphere($fn=40,$fa=12,$fs=2,r=10);}} (2163808 bytes)
CGAL Cache hit: difference(){sphere($fn=40,$fa=12,$fs=2,r=10);multmatrix([[1,0,0,5],[0,1,0,0],[0,0,1,0],[0,0,0,1]]){sphere($fn=40,$fa=12,$fs=2,r=10);}} (2163808 bytes)
CGAL Cache hit: difference(){sphere($fn=40,$fa=12,$fs=2,r=10);multmatrix([[1,0,0,5],[0,1,0,0],[0,0,1,0],[0,0,0,1]]){sphere($fn=40,$fa=12,$fs=2,r=10);}} (2163808 bytes)
CGAL Cache insert: group(){group();difference(){sphere($fn=40,$fa=12,$fs=2,r=10);multmatrix([[1,0,0,5],[0,1,0,0],[0,0,1,0],[0,0,0,1]]){sphere($fn=40,$fa=12,$fs=2,r=10);}}} (2163808 bytes)
CGAL Cache hit: group(){group();difference(){sphere($fn=40,$fa=12,$fs=2,r=10);multmatrix([[1,0,0,5],[0,1,0,0],[0,0,1,0],[0,0,0,1]]){sphere($fn=40,$fa=12,$fs=2,r=10);}}} (2163808 bytes)
[...]

PR #6941, opened 2026-08-06
Revert the cache-string changes from #2792, because they cause false matches.

### LINKED ISSUE #6456: Bad interaction between caching, difference(), render(), and group()
**Describe the bug**
A slightly convoluted combination of difference(), render(), and group() appears to yield bad cache matching.

**To Reproduce**
1. Flush caches.
2. Preview or render the program below.
3. Observe two cubes, when there should only be one.

**Expected behavior**
Only one cube; the second difference should evaluate to nothing.

```
render() {
    difference() {
        cube(10);
    }
}
translate([20,0,0]) {
    render() {
        difference() {
            group();
            cube(10);
        }
    }
}
```

 - OS: Windows 11
 - System: Intel PC 64-bit
 - OpenSCAD Version 2021.01, 2025.10.21

**Additional context**
Based on the behavior, I suspect that cache matching is ignoring the empty group().

Changing the group() to union() fixes the problem, but is not really an answer since in the original the group() is derived from a module call, if(), et cetera.

jordanbrown0:
Revert the cache-string changes from #2792, because they cause false matches.
Fixes #6456.

### DIFF
--- a/src/core/NodeDumper.cc
+++ b/src/core/NodeDumper.cc
@@ -12,46 +12,6 @@
 #include "core/State.h"
 #include "core/node.h"
 
-void GroupNodeChecker::incChildCount(int groupNodeIndex)
-{
-  auto search = this->groupChildCounts.find(groupNodeIndex);
-  // if no entry then given node wasn't a group node
-  if (search != this->groupChildCounts.end()) {
-    ++(search->second);
-  }
-}
-
-int GroupNodeChecker::getChildCount(int groupNodeIndex) const
-{
-  auto search = this->groupChildCounts.find(groupNodeIndex);
-  if (search != this->groupChildCounts.end()) {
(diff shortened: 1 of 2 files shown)


t-paul:
Is there a way to guard that with a test case? I guess caching is always tricky to automatically test. But considering how complicated the logic is, it would be pretty useful to prevent the same cycle in the future.

jordanbrown0:
The repro case could be used as a test, but it would only test for this particular erroneous interaction with difference.

We could have tests that dump their cache strings, and check that they match the expected results.  Maybe we should do that for all tests that render.  (But then an intentional change to cache strings would be a huge pain in the neck.)

jordanbrown0:
Any further comments?  Do you want me to try to come up with a test case before merging this?

PR #6944, opened 2026-08-09
Fix string line continuation regression

### LINKED ISSUE #6942: REGRESSION: String continuation escape generates a warning
**Description of bug**
In the 2021.01 stable release all snapshot builds up to about mid-2025, the following code executes without any warnings or errors, and correctly outputs the entire string to the console:
````
sample = "Lorem ipsum dolor sit amet, consectetur adipiscing \
elit, sed do eiusmod tempor incididunt ut labore et dolore magna \
aliqua. Ut enim ad minim veniam, quis nostrud exercitation \
ullamco laboris nisi ut aliquip ex ea commodo consequat.";
echo(sample);
````
In the snapshot versions I get the console message:
> WARNING: Undefined escape sequence in file ., line 1
> Execution aborted

**To reproduce**
1. Paste the code above into the editor in the 2021.01 release and execute it. It runs as expected.
2. Paste the code above into the editor in a recent snapshot release. It generates a warning. If the option to terminate on the first warning is enabled (as is often the case), execution is terminated.

**Expected behavior**
The string continuation escape character `\`, when placed at the end of a line in a quoted string, should concatenate the string on the next line. This is the expected behavior in other languages as well as OpenSCAD 2021.01.

It is useful to be able to code long strings this way for the purpose of readability when showing demonstration code in documentation, to avoid horizontal scrolling.

**Environment and version info**
 - OS: Windows 11
 - System: Intel PC 64-bit
 - OpenSCAD Version: 2026.07.11 snapshot installed from the OpenSCAD respository

tianrking:
## Summary

- recognize backslash followed by LF or CRLF as an explicit string line continuation
- omit the continuation marker from the resulting string while advancing the lexer source location
- add an echo regression test for the behavior reported in #6942

## Root cause

The generic undefined-escape rule added for malformed escape diagnostics also matched a lone backslash immediately before a physical newline. That turned the previously supported line-continuation syntax into a warning, and `--hardwarnings` then aborted parsing. The new, more-specific lexer rule consumes the backslash and newline together before the fallback rule can report it.

## Testing

- `./scripts/beautify.sh --check` with clang-format 18.1.8
- repository pre-commit hooks on all changed files
- headless/NULLGL Release build with CGAL 6.1.1: `cmake --build build-codex --target openscad -j 4`
- `ctest -R 'echo_(issue6942|string-test|escape-test)' --output-on-failure` (3/3 passed)
- `--hardwarnings` regression with both LF and CRLF input (both exit 0 and emit `ECHO: "A continued string"`)
- baseline verification without the new lexer rule (exit 1 with `Undefined escape sequence`)

Fixes #6942


amatulic:
I'm amazed how quickly my report got attention. Thanks!
I recall this bug appeared in the snapshot builds sometime after mid 2025. Unfortunately, since I deleted all my past snapshots after downloading 2026.07.11 (because it's my wedding anniversary), I couldn't verify on old builds, but I do have a record of string continuation working in August 2025 using an older 2025 build.

jordanbrown0 (approved):
Works for me.

PR #6948, opened 2026-08-11
BUGFIX: Cache OpenCSG VBOs to prevent per-paint rebuilds

### LINKED ISSUE #6947: OpenCSG preview rebuilds VBOs on every paint for unchanged geometry
## Bug

When previewing a moderately heavy model, `OpenCSGRenderer::prepare()` rebuilds and re-uploads all Vertex Buffer Objects on every paint event (camera drag, window resize, animation tick), instead of reusing cached buffers when the geometry has not changed.

## Evidence

`sample <pid> 3` against a stock nightly build displaying a ~10k-polygon user model shows 1720 of 1963 main-thread samples inside `OpenCSGRenderer::prepare → VBOBuilder::create…`:

```
QGLView::paintGL()
  GLView::paintGL()
    OpenCSGRenderer::prepare(...)          1720
      VBOBuilder::create...                 776
        ... std::unordered_map lookups      ~85
```

## Root cause

`OpenCSGRenderer` is destroyed and recreated on every compilation (including per-frame during animation). Since `vertex_state_containers_` is stored directly on the renderer instance, VBOs are discarded with the old instance and rebuilt from scratch on the first paint of the new one.

The guard in `OpenCSGRenderer::prepare()`:
```cpp
if (vertex_state_containers_.empty()) {
    createCSGVBOProducts(...);
}
```
…correctly prevents rebuilding within a single renderer lifetime, but a new renderer always starts empty.

## Environment

- macOS arm64, Qt 6.11, OpenSCAD nightly (master at 889883ab0)
- Same build was near 0% CPU on small models — scales with geometry size as expected for a per-paint rebuild

## Fix available

PR incoming from `whosawhatsis/openscad` branch `feature/opencsg-vbo-rebuild-perf`: introduces a thread-safe, size-limited global VBO cache that survives renderer recreation and is cleared on GL context init.

whosawhatsis:
## Summary

Fixes #6947.

When previewing a moderately heavy model, `OpenCSGRenderer::prepare()` was rebuilding and re-uploading all Vertex Buffer Objects on **every paint event** (camera drag, window resize, animation tick). Profiling showed 1720 of 1963 main-thread samples inside `VBOBuilder::create…` during steady-state preview of an unchanged model.

## Root Cause

`OpenCSGRenderer` is destroyed and recreated on every compilation (including per-frame during animation). Since `vertex_state_containers_` was stored directly on the renderer instance, VBOs were discarded with the old instance and rebuilt from scratch on the first paint of the new one.

### `src/glview/preview/OpenCSGRenderer.h`
- Changed `vertex_state_containers_` from `std::unique_ptr` to `std::shared_ptr` to allow cache sharing.
- Added `static void clearCache()`.

### `src/glview/preview/OpenCSGRenderer.cc`
- Introduced `OpenCSGVBOCacheKey` — a structural key representing all inputs to VBO creation (PolySet pointers, transform matrices, colors, highlight/background modes, shader info).
- Introduced `OpenCSGRendererCache` — a thread-safe, LRU-evicting cache (max 100 entries) storing `shared_ptr<OpenCSGVBOProduct>` keyed by the above.
- `createCSGVBOProducts()` now checks the cache before building; on hit it reuses the existing VBO product, skipping all CPU mesh traversal and GPU buffer uploads.
- `clearCache()` empties the cache.

### `src/glview/GLView.cc`
- Calls `OpenCSGRenderer::clearCache()` inside `initializeGL()` to purge cached VBOs when the OpenGL context is (re-)initialized, preventing cross-context buffer name conflicts.

## Testing

- All Catch2 unit tests pass (65 assertions / 4 test cases).
- Full `ctest -j8` suite: 2721 passed, 9 failed (all pre-existing cairo/fontconfig PDF/SVG layout differences on macOS — unrelated to this change).
- Manually verified smooth camera interaction on a heavy model that previously caused visible lag.

This change has no user-visible syntax or settings changes. It is a backend performance optimization that transparently caches vertex buffer objects during OpenCSG preview rendering, eliminating redundant GPU uploads when the underlying geometry has not changed.


whosawhatsis:
Extended this with the other two renderers, since it already handles the OpenCSG preview (`a0fec87f9`, on top of a merge of current master).

`PolySetRenderer` and `ThrownTogetherRenderer` had the same staleness: per-vertex colors are written into the buffers when they are built, `prepare()` builds them only once, and neither discarded them on a scheme change — so the rendered (F6) and thrown-together views kept the colors they were first drawn under. `CGALRenderer` has always cleared its containers, so those two were the only gaps.

One trap worth flagging for review: `PolySetRenderer::setColorScheme()` also refreshes `colormap_` with the 2D face and edge colors, and the view calls it with the scheme already in force. Copying the early return that is safe in `OpenCSGRenderer::setColorScheme()` — which has no other work to do — skips that refresh and leaves **every 2D render uncolored**; it failed 375 regression tests before I caught it. So here the colormap is always refreshed and only the buffer discard is conditional.

Two GUI tests added beside the existing `checkChangingColorSchemeRecolorsPreparedPreview`, sampling the model's center pixel before and after a scheme change. Both were confirmed failing for the right reason before the change.

Note this fixes F6 only for geometry that carries no baked colors. A `difference()` rendered through Manifold still keeps its old colors after the rebuild, because `ManifoldGeometry::toPolySet()` resolves the scheme at build time and stores the result in the cached `PolySet` — that is the other half of #7053, in #7054. Measured:

| model | recolors with this PR alone |
| :--- | :--- |
| `cube(100, center = true)` | yes |
| `difference() { cube(100); cylinder(...); }` | no — needs #7054 |

The F6 test here therefore uses the plain cube, which genuinely passes; the `difference()` case belongs with #7054.

Verification: `ctest -j5` 2692/2695 on macOS (the three known SVG-arc failures, which fail on master too), full GUI suite 0 failures, `beautify.sh --check` clean with clang-format 18.1.8.

PR #6950, opened 2026-08-12
Preserve 16-bit PNG depth precision in surface() and fix invert Z-offset (#6949)

### LINKED ISSUE #6949: surface(invert=true) places 8-bit PNG surface ~99.6 units too low
`SurfaceNode::convert_image` (`src/core/SurfaceNode.cc`) computes the height as:

```cpp
double z = 100.0 / 255 * (invert ? 1 - pixel : pixel);
```

`pixel` is a luminance value in `[0, 255]`, so the inverted branch mirrors it about `1.0` instead of about `255.0`. The shape of the surface is correct, but every height value is offset by `-100.0 / 255 * 254 ≈ -99.6`: a white pixel that should yield `z = 0` gives `z ≈ -99.6`, and a black pixel that should yield `z = 100` gives `z ≈ 0.39`.

### Repro

```openscad
surface("smiley.png", invert = true);
```

The resulting mesh appears entirely below the XY plane. Because the offset is uniform, it is not obvious when `surface()` is rendered in isolation, but it misplaces the surface relative to any other geometry combined with it. `invert = false` is unaffected.

### Expected Behavior

`255.0 - pixel`, matching the documented `[0, 100]` height range used by the non-inverted path.

whosawhatsis:
This PR improves `surface()` heightmap PNG import in two ways:

1. **16-bit PNG depth precision:** Inspects PNG bit depth via `lodepng_inspect` and decodes 16-bit per channel PNGs natively without truncation, preserving all 65,536 fine height levels across the `[0, 100]` height range. 8-bit PNG inputs continue to import with 8-bit scaling with bit-identical backwards compatibility.
2. **Fix `invert = true` Z-offset (Fixes #6949):** Corrects `SurfaceNode::convert_image` height calculation from `1 - pixel` to `255.0 - pixel`. Previously, luminance was mirrored about `1.0` instead of `255.0`, shifting inverted 8-bit surfaces downward by ~99.6 units into negative Z space. Both 8-bit and 16-bit inverted PNGs now position correctly within `[0, 100]`.

## Testing

- **Catch2 Unit Tests (`src/core/surface_node_test.cc`):** 17 assertions covering 8-bit backwards parity, sub-8-bit 16-bit precision preservation, and 8-bit vs 16-bit `invert` height alignment (`./build/OpenSCADUnitTests "[SurfaceNode]"`).
- **CTest Suite:** `ctest -R surface` 40/40 tests passed across CGAL, Manifold, preview, and CSG renderers.
- **Style:** Verified with `./scripts/beautify.sh --check` using pinned `clang-format` 18.1.8.

User documentation for the wiki is provided in the section below.

=== `surface()` 16-bit PNG Heightmap Support ===

When importing PNG images into the `surface()` module, OpenSCAD automatically detects 16-bit per channel PNG images (e.g., 16-bit grayscale or 16-bit RGBA heightmaps) and preserves their full 16-bit depth precision (65,536 height levels). 8-bit PNG images continue to import with 8-bit scaling (256 height levels) with bit-identical backwards compatibility.

`invert = true` previously placed the resulting surface about 99.6 units below the XY plane. It is now positioned in the same 0–100 height range as `invert = false`. The shape of the surface is unchanged; only its position along Z is corrected. Models that compensated for the old offset with an explicit `translate()` will need that compensation removed.


t-paul:
The change looks a bit too noisy for what it's doing. As far as I can tell lodepng can decode to 16-bit regardless of the input, so it should be possible to have a much smaller change with a single code path covering the RGB(A?) cases for all resolutions. I don't think there's a big memory concern as the generated geometry is much more heavy, so the limitations are likely driven by the pixel count. 
It might be interesting to have a (new) dedicated grayscale path that keeps the gray values 1:1 with only the 0:100 scale at the end.
The test cases look useful, but I'm not too fond on generating and saving the images in the test case itself.

t-paul (approved):
Thanks for the updates.

PR #6961, opened 2026-08-18
Fix HTML test report parsing a stale ctest log, and failing the run on an unknown output type

### LINKED ISSUE #6959: test_pretty_print.py can parse a stale log: findlogfile() takes an arbitrary LastTest.log* match
## Summary

`findlogfile()` in `tests/test_pretty_print.py` picks the first filesystem-order match of `LastTest.log*`:

```python
logfilename = next(pathlib.Path(logpath).glob('LastTest.log*'), None)
```

The comment above it explains why the temp file has to be considered — ctest may not have flushed `LastTest.log` by the time this post-test hook runs, so the current run may only exist in `LastTest.log.tmp<random>`. But when **both** files exist, the glob's order decides, not recency.

An interrupted or killed ctest run leaves its `LastTest.log.tmp<random>` in `Testing/Temporary/` permanently. From then on, the HTML report can describe that old run rather than the one just executed.

## How it shows up

In my build directory a 5 MB `LastTest.log.tmp2fa58` from a run killed two days earlier was being parsed instead of the current log. Every subsequent `ctest` invocation ended with:

```
Problem running command: .../test_pretty_print.py --builddir=...
Problem executing post-test command(s).
Errors while running CTest
```

...on runs where every test passed. The report was describing tests from the stale log, including one whose output type the report cannot render (see the companion issue), so the generator raised and — because it runs as `CTEST_CUSTOM_POST_TEST` (`tests/CTestCustom.template`) — ctest exited non-zero.

Deleting the stale temp file makes the same tree run clean, which isolates the cause.

## Suggested fix

Take the newest matching file rather than the first. That keeps the in-progress case working, because the temp file is the newest exactly when it is the live one.

## Reproduction

```bash
cd build
touch -t 202401010000 Testing/Temporary/LastTest.log.tmpstale   # any stale temp log
ctest -R <any test>
```

Note that the visible failure needs a stale log containing something the report cannot render; a stale log alone silently produces a report about the wrong run.

## Environment

macOS 15 (arm64), CMake/CTest 4.4.0, Python 3.14.6. Nothing here looks platform-specific — the glob order and the leftover temp file are the only ingredients.

A PR follows with a fix and a test.

### LINKED ISSUE #6960: test_pretty_print.py raises on an unknown output type, which fails ctest from a post-test hook
## Summary

`to_html()` in `tests/test_pretty_print.py` matches an allowlist of output types and raises on anything else:

```python
if test.type in ('txt', 'ast', 'csg', 'term', 'echo', 'stl', '3mf', 'off', 'obj', 'pov', 'dxf', 'svg', ''):
    ...
elif test.type == 'png':
    ...
else:
    raise TypeError(f"Unknown test type '{test.type}' in test {test.fullname}")
```

The script runs as `CTEST_CUSTOM_POST_TEST` (`tests/CTestCustom.template`), so an unrecognised type does not merely lose the HTML report — it makes **ctest exit non-zero on a run whose tests all passed**.

The allowlist is already out of date. `tests/CMakeLists.txt` registers

```cmake
add_test_cmdline(export-param ... SUFFIX json ... --export-format param)
```

so `json` is a type the suite can produce and the report cannot render. Any future export format added to the suite has the same effect until someone remembers to edit this list.

## Why CI does not see it

The report only iterates tests it believes failed, and `export-param_issue6597` passes, so a normal run never reaches the raise. It is reachable when a test of that type genuinely fails — or, as I hit it, when the report is parsing a *stale* log that contains one (see the companion issue about `findlogfile()`).

So this is latent fragility rather than something breaking your CI today. It is worth fixing on its own terms: a reporting step should not be able to turn a green test run red.

## Suggested fix

`png` is the only type needing special handling; everything else is rendered as a text log. Inverting the condition removes the failure mode without an allowlist that goes stale.

## Environment

macOS 15 (arm64), CMake/CTest 4.4.0, Python 3.14.6.

A PR follows, together with the `findlogfile()` fix.

whosawhatsis:
Fixes #6959. Fixes #6960.

Two independent defects in `tests/test_pretty_print.py`, which runs as `CTEST_CUSTOM_POST_TEST`. Between them they can make `ctest` exit non-zero on a run whose tests all passed, while reporting on tests from an entirely different run.

## 1. `findlogfile()` could parse a stale log (#6959)

It globbed `LastTest.log*` and took the first filesystem-order match. The existing comment explains why the temp file has to be considered — ctest may not have flushed `LastTest.log` when this hook runs — but with both present the glob's order decided, and an interrupted run leaves its `LastTest.log.tmp<random>` behind permanently.

In my build directory a 5 MB temp file from a run killed two days earlier was being parsed instead of the current log, so every `ctest` invocation reported on that old run.

Now takes the newest match. The temp file still wins while a run is genuinely in progress, because then it is the newest.

## 2. An unknown output type raised, failing the run (#6960)

The type dispatch matched an allowlist and raised `TypeError` on anything else. Since this is a post-test hook, that turns a green run red rather than merely losing the report. The allowlist is already missing `json`, which `add_test_cmdline(export-param ... SUFFIX json ...)` produces.

`png` is the only type that needs special handling; everything else renders as a text log, so the condition is inverted. That removes the failure mode permanently instead of adding one more extension to a list that goes stale whenever a format is added.

## Testing

New `tests/test_pretty_print_logfile.py`, registered as `pretty_print_logfile`. It calls `findlogfile()` directly and needs no OpenSCAD binary, covering three states:

- a finished run — the real log is newest and must win
- a run in progress — `LastTest.log` not yet flushed, so the temp file must win
- several stale temp files from several interrupted runs

It fails on the first case before this change, picking the stale temp file.

Also verified against the original failure with the 5 MB stale log back in place: `ctest -R export-param` now writes its report and exits 0, where before it ended in "Problem executing post-test command(s)". `export-param`, `export-pdf`, `pretty_print_logfile` and the surrounding tests pass (11/11 locally).

## User-facing documentation

None required — this changes only the test harness. No language, CLI, GUI or file-format behaviour changes, so there is nothing to add to the wiki.

### DIFF
--- a/tests/test_pretty_print.py
+++ b/tests/test_pretty_print.py
@@ -339,13 +339,17 @@ def to_html(project_name, startdate, tests, enddate, sysinfo, sysid, imgcomparer
 
     templates = Templates()
     for test in report_tests:
-        # relative-output tests have no "type"
-        if test.type in ('txt', 'ast', 'csg', 'term', 'echo', 'stl', '3mf', 'off', 'obj', 'pov', 'dxf', 'svg', ''):
+        # Every type but png renders as text, including the empty type relative-output tests
+        # carry. This used to be an allowlist of known extensions, which meant adding an export
+        # format to the suite made the report generator raise -- and because it runs as a ctest
+        # post-test step, that turned an otherwise green run into a failing one. 'json' arrived
+        # with the export-param tests and did exactly that.
+        if test.type != 'png':
             text_test_count += 1
             templates.add('text_template', 'text_tests',
                           test_name=test.fullname,
                           test_log=html.escape(test.fulltestlog))
-        elif test.type == 'png':
+        else:  # png
             image_test_count += 1
             alttxt = 'OpenSCAD test image'
 
@@ -373,8 +377,7 @@ def to_html(project_name, startdate, tests, enddate, sysinfo, sysid, imgcomparer
                           actual=actual_img,
                           expected=expected_img,
                           mask=mask_img)
-        else:
-            raise TypeError(f"Unknown test type '{test.type}' in test {test.fullname}")
+
 
     for mf in sorted(makefiles.keys()):
         mfname = mf.strip().lstrip(os.path.sep)
--- a/tests/test_pretty_print.py
+++ b/tests/test_pretty_print.py
@@ -205,13 +205,16 @@ def png_encode64(fname, width=512, data=None, alt=''):
 
 def findlogfile(builddir):
     logpath = os.path.join(builddir, 'Testing', 'Temporary')
-    # The ctest command may not finish before the LastTest.log.tmp* is flushed,
-    # so read that file if it exists.
-    logfilename = next(pathlib.Path(logpath).glob('LastTest.log*'), None)
-    if not os.path.isfile(logfilename):
-        print('can\'t find and/or open logfile', logfilename)
+    # ctest may not have flushed LastTest.log by the time this post-test hook runs, in which case
+    # the current run is in LastTest.log.tmp<random> instead. Take the newest of the two rather
+    # than the first the glob happens to yield: an interrupted run leaves its temp file behind
+    # forever, and picking that describes a run from days ago -- or fails outright on an output
+    # type it contains.
+    candidates = [path for path in pathlib.Path(logpath).glob('LastTest.log*') if path.is_file()]
+    if not candidates:
+        print('can\'t find and/or open logfile in', logpath)
         sys.exit()
-    return str(logfilename)
+    return str(max(candidates, key=lambda path: path.stat().st_mtime))
 
 # --- Templating ---
 
--- a/tests/test_pretty_print.py
+++ b/tests/test_pretty_print.py
@@ -203,17 +203,45 @@ def png_encode64(fname, width=512, data=None, alt=''):
     return tag
 
 
+RUN_MARKER_NAME = '.pretty_print_run_marker'
+
+
+def mark_run(builddir):
+    # Touched by CTestCustom.template's CTEST_CUSTOM_PRE_TEST, right before ctest starts running
+    # tests, so findlogfile() can tell "the log this run wrote" from "whichever log is newest".
+    logpath = os.path.join(builddir, 'Testing', 'Temporary')
+    os.makedirs(logpath, exist_ok=True)
+    pathlib.Path(os.path.join(logpath, RUN_MARKER_NAME)).touch()
+
+
 def findlogfile(builddir):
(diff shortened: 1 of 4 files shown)


tests/test_pretty_print.py
```diff
@@ -370,8 +377,7 @@ def to_html(project_name, startdate, tests, enddate, sysinfo, sysid, imgcomparer
                           actual=actual_img,
                           expected=expected_img,
                           mask=mask_img)
-        else:
-            raise TypeError(f"Unknown test type '{test.type}' in test {test.fullname}")
+
```
t-paul: I don't think that's a good idea. This check and exception gives a dedicated indication something is missing in case new extensions are added. Silently doing *something* is not a good strategy for testing.

tests/test_pretty_print_logfile.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env python3
```
t-paul, line 1: Having tests for tests is confusing. Maybe it would be easier to just error out in case multiple files match the original glob pattern and error out with a clear message in the original script.

whosawhatsis:
You're right on both counts, pushed a fix:

1. Restored the allowlist + `TypeError`. New `test.type` values only come from hand-written
   `SUFFIX` arguments in `tests/CMakeLists.txt`, so updating the list is a bounded, deliberate
   step that belongs in the same commit that adds the type — not an open-ended maintenance
   burden. Added `json` to the list, which is what actually needed to happen for #6960.

2. Replaced "pick the newest `LastTest.log*`" with an explicit marker. `findlogfile()` still
   can't know which log belongs to *this* `ctest` invocation from the file content alone — ctest
   writes one shared log per run, not one per test — but mtime-only was never quite right either:
   two `ctest` processes touching the same builddir around the same time could both leave a
   fresh-looking file, and newest-mtime has no way to prefer "the one this run wrote" over "the
   one the other run wrote." So `CTestCustom.template` now touches a marker file via a new
   `CTEST_CUSTOM_PRE_TEST` hook right before ctest starts testing, and `findlogfile()` only
   considers candidates written at or after it. A run that finds nothing fresher than its own
   marker exits with a clear error instead of silently falling back to a stale log.

Kept `test_pretty_print_logfile.py` for now since the marker logic is exactly the kind of
`branch × edge-case` matrix that's easy to get subtly wrong (concurrent runs, missing marker,
nothing-fresh) — happy to fold it into the main script's tests directory differently if you'd
rather.

t-paul (approved):
Ok, lets go with this. I'm not sure testing tests is the best solution, but it's a well isolated case so we'll see how this goes. Just waiting for the macOS test to go green, it looks like that fail was unrelated to the changes here.

PR #6963, opened 2026-08-19
Export animations as GIF, APNG or MJPEG/AVI

whosawhatsis:
Adds animation export to a single container file — animated GIF, APNG, or Motion-JPEG AVI — instead of only a folder of numbered PNGs. Addresses #891, which asked for this and was closed as not-planned.

**User documentation for the openscad.org wiki is in the last section of this description.**

## What changes

`--animate N` already existed in the headless CLI and wrote `out00000.png`, `out00001.png`, …. The format is now chosen by the output file's extension:

```
openscad --animate 60 -o spin.gif  spin.scad     # animated GIF
openscad --animate 60 -o spin.apng spin.scad     # animated PNG
openscad --animate 60 -o spin.avi  spin.scad     # Motion JPEG in AVI
openscad --animate 60 -o spin.png  spin.scad     # unchanged: numbered stills
```

The GUI's **Dump Pictures** checkbox picks the format the same way, from the extension in the save dialog. `--animate_fps` sets the rate (default 30, range 1–100).

## Approach, and why no ffmpeg

The goal was no new build dependency and nothing spawned at runtime, so each container is written directly:

- **APNG** — an ordinary PNG plus `acTL`/`fcTL`/`fdAT` chunks. Frames are independently-compressed zlib streams, so the already-vendored lodepng encodes each frame and the encoder lifts its `IDAT` payload out. No new dependency at all.
- **GIF** — vendored `gif.h` (Charlie Tangora), a ~700-line single header, **public domain**, supplying median-cut palette and Floyd–Steinberg dithering.
- **AVI** — Motion JPEG, using vendored `stb_image_write` v1.16 (`nothings/stb`, **dual MIT / public domain**) for the per-frame JPEG, wrapped in a hand-written AVI container.

Licence text for both vendored files was read in full before vendoring; both are unmodified.

`VideoEncoder` (`src/io/VideoEncoder.h`) is the common interface, deliberately free of Qt so the same encoders serve the GUI dump and the headless CLI, and so the unit tests build in `HEADLESS`/`NULLGL` configurations.

## Points a reviewer may want to weigh in on

- **GIF cannot represent an arbitrary frame rate.** Delays are whole centiseconds, so only 10, 20, 25 and 50 fps are exact. The default of 30 fps becomes 33.3 fps. Rather than silently rounding, this warns and states what the file will actually play at. APNG and AVI store the requested rate exactly.
- **APNG viewer support is patchy in a misleading way** — a viewer without support typically shows frame 0 while still reporting the correct duration, so a good file looks broken. Documented in the user text rather than worked around.
- Motion JPEG is lossy and per-frame, which makes AVI files large but easy for other tools to ingest; it is offered as an interchange format, not a finished one.

## Testing

- `tests/test_animation_export.py` (registered as `animation_export`) — end-to-end for each container: frame counts parsed out of the real file structure, plus the still-image path.
- Catch2 unit tests for the encoders (`src/io/video_encoder_test.cc`), which run headless.
- GIF structural validity is checked separately (`GIF output is a structurally valid animated GIF`).
- GUI flow verified by hand, driving the real app (`ENABLE_GUI_TESTS` is off in this configuration).
- Full `ctest` on macOS, Linux and the Windows/MSYS2 build; `./scripts/beautify.sh --check` clean with clang-format 18.

_Ready to paste into the wiki as-is._

### Animation export

OpenSCAD can write an animation as a single file rather than a folder of numbered images. Three
container formats are available, chosen by the output file's extension:

| Extension | Format | Notes |
| :--- | :--- | :--- |
| `.gif` | Animated GIF | Limited to 256 colours per frame, dithered. Widely viewable, loops forever. |
| `.apng` | Animated PNG | Lossless. Viewable in any current browser. Larger files than GIF. |
| `.avi` | Motion JPEG in AVI | Lightly compressed. Intended as an intermediate to hand to a video editor or converter, not as a finished file. |

None of these require an ffmpeg installation, and no feature flag needs to be enabled.

To export one, pass `--animate` with the number of frames and give the output file the extension
you want:

    openscad --animate 60 -o spin.gif spin.scad

`$t` runs from 0 up to (but not including) 1 across the frames, exactly as it does in the
animation panel, so the example above renders 60 frames at `$t` = 0, 1/60, 2/60 … 59/60.

Frame rate defaults to 30 per second and is set with `--animate_fps`:

    openscad --animate 60 --animate_fps 24 -o spin.apng spin.scad

Valid rates are 1 to 100.

**GIF cannot store an arbitrary frame rate.** Its delays are whole hundredths of a second, so only
rates that divide 100 evenly — **10, 20, 25 and 50** — are exact. Everything else is rounded to the
nearest available delay, and OpenSCAD prints a warning saying what the file will actually play at.
30 fps is the case worth knowing about: the nearest delay is 3/100 s, so a 30 fps GIF plays at
33.3 fps, about 11% fast. `.apng` and `.avi` store the rate you asked for exactly, so use one of
those if the timing matters.

Image size is set as usual with `--imgsize`, and camera and view options behave exactly as they do
for a single `.png` export:

    openscad --animate 60 --imgsize=800,600 --camera=0,0,0,55,0,25,140 -o spin.avi spin.scad

Notes and limitations:

- `--animate` is required. Asking for `.gif`, `.apng` or `.avi` without it is an error.
- These formats cannot be written to standard output; give a real filename.
- Exporting to `.png` is unchanged: it still writes one numbered file per frame
  (`spin00000.png`, `spin00001.png`, …).
- `--animate_sharding` splits a render across several processes. Each shard writes its own
  complete file containing only its own frames, so the shards will need joining afterwards.
- Motion JPEG is a lossy format and every frame is compressed independently, which is what makes
  the files large but easy for other tools to read. Use `.apng` if you need the frames to be
  exactly what OpenSCAD rendered.

### Viewing an animated PNG

Animated PNG is well supported in web browsers — Chrome, Firefox and Safari all play them — but
support elsewhere is patchy, and the common failure is misleading rather than obvious. A viewer
without APNG support generally shows the first frame and nothing else, while still reporting the
correct duration and frame count, so a perfectly good file looks like a broken export.

Known cases at the time of writing: **VLC** does not play APNG correctly, and on macOS **Preview,
Quick Look and Finder** show only the first frame. If an `.apng` looks like a still image, open it
in a browser before concluding anything is wrong with it.

Export to `.avi` instead if you need a file that plays in an ordinary media player.

### Recording an animation from the GUI

Open the animation panel (**Window → Animate**), set the number of steps and the frame rate, then
tick **Dump Pictures**. OpenSCAD asks where to save the recording, and the extension you choose
decides the format:

- `spin.gif`, `spin.apng` or `spin.avi` record the whole animation into that one file.
- `spin.png` — or any other image extension — writes one image per frame instead, numbered:
  `spin00000.png`, `spin00001.png`, and so on.

Recording begins on the next frame and stops automatically after one complete loop of the
animation, at which point the checkbox unticks itself. Untick it earlier to stop and keep what has
been recorded so far. Cancelling the save dialog leaves the checkbox unticked and records nothing.

The frame rate in the animation panel is the frame rate stored in the recording. The recording is
made at the current size of the 3D view, so resize the window before you start rather than during:
a single-file recording cannot change size partway through, and resizing the view mid-recording
stops it.


stefano2734:
Really obsolete Gif with 256 colors?
gif was a format with many cve numbers in last years and a security problem.

There are enough animated formats with more colors and  true color support.

animated webp and animated avif are modern formats with support by security problems.

whosawhatsis:
@stefano2734 Gif was included because it has the widest compatibility. I considered adding in ffmpeg to support every possible export format, but decided that APNG and MJPEG were good inputs to a transcoder (more convenient than an image sequence) that were cheap to implement, and gave the project a more reasonable scope.

stefano2734:
> @stefano2734 Gif was included because it has the widest compatibility. I considered adding in ffmpeg to support every possible export format, but decided that APNG and MJPEG were good inputs to a transcoder (more convenient than an image sequence) that were cheap to implement, and gave the project a more reasonable scope.

Thank you for your excellent work. 
APNG and MJPEG are ok for me in relation function summary to necessary security updates. 
256 colors is really poor limitation in gif-Format in year 2026 with HDR coming wide in displays.
Gif is a permanent risk like other formats, but others have more possibilities.
https://www.heise.de/en/news/Gimp-Unpatched-vulnerability-allows-code-injection-with-GIFs-11260781.html
There is a responsibility by developer to dry out this older formats with less possibilities to reduce attack area for users.

whosawhatsis:
I appreciate your point about security vulnerabilities being found related to GIF files. However, I'm not sure it applies here. GIF-related security vulnerabilities are tied to programs that are consumers (decoders/parsers) of untrusted files. Since OpenSCAD is purely an exporter, it doesn't expose its users to those decoder risks. While I understand the goal of moving the broader ecosystem away from older formats, my priority is giving users universal compatibility where GIF is still required, alongside APNG and MJPEG for higher-fidelity needs.

PR #6965, opened 2026-08-19
Depth map export (metric/visual PNG, PFM, camera sidecar, viewport preview)

whosawhatsis:
## Summary

Adds depth-map export to OpenSCAD.

## Motivation

An AI agent driving OpenSCAD cannot visually inspect a rendered model the way a person looking at
a screen can — it needs the scene's geometry conveyed as structured data it can actually reason
over. A depth map is a direct, cheap way to give an agent (or any headless pipeline) that
feedback: per-pixel distance data derived straight from the existing render path, no separate
tooling required.

The `visual` profile also has an image-generation use case distinct from that: MiDaS/ControlNet-
style depth conditioning, where a depth image is used as a structural control input to steer image
generation. That's a different consumer with different expectations (bit depth, direction, scale),
which is why the export supports two profiles rather than one — see *Design decisions* below.

Together these motivated depth-map export:

- `--export-format=depthmap` — a greyscale PNG depth image, either `metric` (16-bit, linear
  millimetres from the camera, default) or `visual` (8-bit, normalized to the model's own depth
  extent — MiDaS/ControlNet-shaped output). Selected with `-O depthmap/profile=metric|visual`.
- `-O depthmap/range=near,far` (or a single `range=far` shorthand) to pin the encoded range
  explicitly, so exports at different zoom levels or of different models stay comparable.
- `-O depthmap/camera=path.json` — a JSON sidecar carrying the camera's modelview/projection
  matrices, clip planes, FOV and image size, so a depth map can be unprojected back to 3D points.
- `--export-format=pfm` — lossless 32-bit float depth (Portable Float Map), for consumers that
  want exact distance with no bit-depth ceiling.
- `View > Shade by Depth` in the GUI, and `--view=depth` on the command line, for a live preview
  of the same depth data (fixed-function fog, no readback).

Depth is always measured from the eye, in millimetres, and is consistent between perspective and
orthographic projections (see *Design decisions* below).

## Design decisions

**Depth buffer readback, not fog, for the export path.** The viewport preview uses fixed-function
`GL_LINEAR` fog — cheap, and fog distance is already eye-space linear. The export path instead
reads the depth buffer back (`GL_FLOAT`) and linearizes it, because the GL spec permits
implementations to *approximate* the fog factor, which is fine for a preview but not for a
`metric` profile where the pixel value is a millimetre measurement whose accuracy shouldn't vary
by GPU/driver.

**Two encoding profiles, not one flag with modifiers.** `metric` and `visual` consumers disagree
on direction, scale, *and* bit depth simultaneously, so a single `--depth-invert`-style flag would
leave two of the three wrong for whichever family isn't the default. Considered and rejected: a
log/inverse-depth encoding (no reader needs it — MiDaS-trained models are documented as robust to
linear depth from render engines, and a bounded CAD scene has none of the unbounded-range problem
inverse depth exists to solve).

**Rejected: a single color+depth file, à la iPhone Portrait-mode HEIF.** Before settling on the
JSON camera sidecar (and on depth/color always being separate export calls), we discussed whether
OpenSCAD should instead emit one file carrying both the color image and its depth map, the way
HEIF embeds a depth map alongside the photo. This was ruled out because **PNG has no suitable
payload type for it** — its only generic extension point is a `tEXt`/`iTXt` ancillary chunk, and
base64-encoding an entire second image (itself potentially 16-bit) into a text chunk is not a
reasonable use of it: it bloats the file, is opaque to every other PNG tool, and no consumer
anywhere expects to unpack it. Unlike HEIF (a real ISOBMFF multi-track container) or JPEG (which
at least has an MPO convention some tooling understands), PNG was never designed to carry a second
image payload. So color, depth, and the camera sidecar ship as separate files by necessity, not by
preference — there was no candidate single-file format on the table that actually supports it.

**Camera matrices are column-major, right-handed, depth-from-eye.** Chosen to match OpenGL's own
convention (so most graphics/linear-algebra libraries can consume the sidecar directly) rather than
inventing a private convention; the sidecar states its own conventions explicitly so nothing has to
be inferred.

## Testing

- `OpenSCADUnitTests "[Depthmap]"` — 101 assertions across 23 cases: pure-function encoding tests
  (`encode_depthmap`, both profiles, clamping, degenerate/zero-extent cases), `depth_range_for_bounds`
  (bbox pinning, camera-inside-model, degenerate models), `linearize_depth` (both projections,
  far-plane-to-background), and the range/sidecar/PFM option parsing. Full unit suite: 179
  assertions in 28 cases, all passing.
- `ctest -R "png|depthmap|surface"` — 50/50 passing, covering the colour-PNG and `surface()` paths
  this change shares code with.
- Manual end-to-end verification against the built binary, since a 16-bit PNG's byte layout isn't
  exercised by the pure-function tests: `metric` emits bitdepth 16 / colortype 0, `visual` emits
  bitdepth 8 / colortype 2, PFM emits a `Pf` / little-endian (`-1.0`) header, and the camera sidecar
  contains the documented keys including its stated `matrixOrder` / `handedness` / `depthOrigin` /
  `depthUnits` conventions.
- No end-to-end regression test is registered for the depth PNGs themselves — pixel-comparison
  regression tests are acutely GPU/driver-sensitive for a 16-bit depth encoding and would be flaky
  rather than useful; the deterministic logic is covered by the unit tests above instead.

*(Written for users rather than reviewers, and kept as the last section so it can be pasted into
the wiki as-is if this lands.)*

### Depth map export

OpenSCAD can export the depth of the rendered scene as a PNG image instead of its colour. Each pixel
holds how far the surface at that point is from the camera, which is what tools that consume depth
maps — 3D reconstruction and point-cloud software, and image-generation models that take a depth
image as a structural control — expect as input.

Depth map export is command-line only; there is no GUI equivalent. It does **not** require any
experimental feature flag.

Because a depth map is still a PNG, the file extension cannot select it — `-o out.png` on its own
means an ordinary colour render. Ask for it explicitly:

    openscad --export-format=depthmap -o depth.png model.scad

The image is rendered with the same camera, size and projection options as an ordinary PNG export,
so `--imgsize`, `--camera`, `--projection` and `--render` all behave as they do there.

#### Profiles

Two encodings are available, because the tools that read depth maps disagree about what a pixel
should mean. Select one with `-O depthmap/profile=...`:

`metric` (the default) — a 16-bit greyscale PNG in which the pixel value is the distance in
millimetres, one unit per millimetre. Near surfaces are dark, distant ones bright, and pixels with
no geometry hold the maximum value 65535. This is the convention used by depth cameras and by
robotics and point-cloud tooling.

    openscad --export-format=depthmap -o depth.png model.scad

`visual` — an 8-bit greyscale PNG stretched across the depth range of the model itself, with the
nearest surface white, the farthest black, and the background black. This is the form expected by
image-generation depth controls and by general image-editing software.

    openscad --export-format=depthmap -O depthmap/profile=visual -o depth.png model.scad

Whichever profile you choose, OpenSCAD prints the depth range it found, in millimetres, along with
what it measured from, when it writes the file. A PNG cannot carry that information inside the
image, so keep the printed line if you need to convert pixel values back to distances.

#### Pinning the depth range

By default the range a depth map is encoded across is derived automatically: `metric` from the
camera, `visual` from whatever depths are actually visible in that one image. Pin it explicitly with
`-O depthmap/range=near,far` (both in millimetres from the camera) when you need two exports — of the
same model at different zoom levels, or of different models — to use the same scale and be directly
comparable:

    openscad --export-format=depthmap -O depthmap/range=50,150 -o depth.png model.scad

A single number is shorthand for a range starting at 0 — `range=150` means the same as `range=0,150`
— for the common case of "everything past this distance is background":

    openscad --export-format=depthmap -O depthmap/range=150 -o depth.png model.scad

Geometry outside the range is clamped to the nearest end rather than lost or rejected, and OpenSCAD
prints a warning if anything was actually clamped. An invalid range (non-numeric, more than one
comma, or with the near value not less than the far value — including a single number that is zero
or negative, since near is implicitly 0) is rejected with an error rather than silently ignored.

#### Camera parameters sidecar

A depth map alone cannot be turned back into a 3D point cloud — that also needs the camera's
projection and view matrices. Write them alongside the image as a JSON file with
`-O depthmap/camera=path.json`:

    openscad --export-format=depthmap -O depthmap/camera=depth.json -o depth.png model.scad

The file contains the `modelview` and `projection` matrices (16 numbers each, **column-major**, the
order OpenGL itself uses — most graphics and linear-algebra libraries can consume this directly, but
transpose if yours assumes row-major), the near and far clip distances, field of view, whether the
projection is orthographic, and the image dimensions. It also states the conventions this feature
uses so nothing has to be guessed: a **right-handed** coordinate system, and depth measured from the
**eye** in millimetres (`depthOrigin: "eye"`, `depthUnits: "mm"`) — true for both perspective and
orthographic projections.

The camera sidecar is a separate file, not embedded PNG metadata — PNG has no suitable place to
carry a second image or a rich metadata block, so the JSON file travels alongside the depth image
rather than inside it. Match filenames if you need to keep them paired.

#### Float depth export (PFM)

For applications that want depth as exact floating-point distance rather than an 8- or 16-bit
encoding, export a [Portable Float Map](http://www.pauldebevec.com/Research/HDR/PFM/) instead of a
PNG:

    openscad --export-format=pfm -o depth.pfm model.scad

Each pixel is a 32-bit float, the distance from the camera in millimetres, with no rounding, no
65535mm ceiling, and no separate scale factor to keep track of. Background (no-geometry) pixels are
written as `+Inf`, which most PFM readers treat as "no data." PFM does not carry a projection or
camera convention of its own; use the camera sidecar above alongside it if you need to unproject the
result. `-O depthmap/range` does not apply here — a float export has no bit-depth ceiling to pin a
range against.

#### Previewing depth in the viewport

The **View > Shade by Depth** menu item shades the model in the 3D view by how far each surface is
from the camera, near surfaces bright and distant ones dark, instead of by lighting. It is a way to
see the shape of your depth data while you work — the setting is remembered between sessions, and it
affects only how the model is drawn, never the exported geometry.

The same shading is available from the command line for an ordinary colour render:

    openscad --view=depth -o preview.png model.scad

**The viewport is indicative, not an exact preview of the exported file.** The two are scaled
differently on purpose. The viewport spreads its shading across the model's bounding box, so a
surface keeps the same shade as you rotate the model; the exported `visual` profile stretches its
range to fit only the depths actually visible in that one image, which gives more contrast but would
flicker if it were used live. Expect the exported image to look higher-contrast than the viewport.
Shapes and the near/far ordering match; exact greys do not.

The background reads as pure black while depth shading is on - the far end of the shading, exactly
like a depth map's no-geometry pixels - rather than the normal colour scheme. Axes and crosshairs
keep their normal colours, since they are UI overlays and not part of the model. If the camera is
moved inside the model, anything behind it is drawn at the nearest shade rather than disappearing.

**`%` ("ghost"/background) geometry is depth-shaded when previewing, but is not exported at all once
`--render` is used** - the same as OpenSCAD's other export formats treat it, so the same file can
produce two different depth maps depending on that flag. `color(alpha)` transparency has no effect on
depth in any mode: a translucent surface blocks depth exactly like an opaque one, so only the
frontmost surface at each pixel is ever recorded.

#### Things to watch out for

- **Distances are measured from the camera, and the camera moves.** The origin is the camera in both
  projections, so a model exports the same distances whether you render it in perspective or
  orthographic. But the camera itself moves, so the same model exported at two different zoom levels
  produces different numbers. For values that are directly comparable between exports, either set the
  camera explicitly with `--camera`, or pin the scale itself with `-O depthmap/range=near,far` (see
  *Pinning the depth range* above) — the range is the more direct fix when the camera position is
  otherwise expected to vary.
- **The `metric` profile cannot represent distances beyond 65535mm.** Anything further away is
  written as 65534, so a model that large loses depth detail at the back without any warning. The
  `visual` profile is unaffected, because it rescales to the model.
- **In the `visual` profile the background is the same black as the farthest surface**, so a surface
  at the very back of the model cannot be told apart from empty space. Use `metric` if you need to
  distinguish them.
- **Depth is taken from the frontmost surface only.** Nothing behind a visible face is recorded; a
  hollow or nested model exports only its outermost visible skin.
- **Color and depth are always separate files.** There is no combined color+depth export (unlike,
  say, an iPhone Portrait-mode HEIF) — render the color image and the depth map as two separate
  commands against the same camera if you need both.

This feature is experimental and does not require enabling any flag.


jordanbrown0:
I admit that I didn't read all of that.  However, I have to ask a fundamental question.

Humans need 2D presentations of the 3D object, because we have to perceive them through 2D screens - and even if we had a 3D display, we *still* get a primarily 2D image.  We have to look at a 2D object, rotate it, et cetera, to reconstruct the 3D object in our heads.  However, an AI doesn't have either limitation.  Why can't it directly consume 3D objects?

jordanbrown0:
But a depth map would be cool for generating [single image random dot stereograms](https://en.wikipedia.org/wiki/Autostereogram#Random-dot).

whosawhatsis:
@jordanbrown0, two issues I see with that:

1: Yes, an AI could be trained to directly accept 3D models instead of images. However, current LLMs do not have this training. Therefore, we need ways to transmit the information that will give them information that they can understand. A depthmap is one candidate, but I pulled this back to a draft because I'm working on a larger suite of shaders that supersets this, several of which are intended to provide data to AI models that they might otherwise have trouble discerning from a regular image.

2: Handing a 3D model to the AI would require a full CGAL/Manifold render, which (at least for some models) may be significantly slower. The model should be able to take advantage of the speed of OpenCSG.

Issue #3520, opened 2020-12-08
Feature request: Use multiple processes (at least on Mac)


whosawhatsis:
On windows, it is easy to launch multiple instances that each have their own process. The Mac app architecture assumes that there will never be more than one instance of an app open at once, and that the app will handle multiple windows and processes as necessary. Creating a second instance can only be launched as a terminal process (using the full path to the executable file inside the application package), and that terminal window has to stay open.

Using only a single process means that you can't have two preview/render processes running at the same time, and can't make use of multiple processor cores. It would be much more useful if each window was its own process (though it may still make sense for tabs to share a process, if that's necessary for them to share cache data).

whosawhatsis:
You're right, I forgot about the -n option. Still, it would be much nicer if you didn't have to launch multiple instances of the program (which is not how Mac apps are intended to be used) for each document to have its own process for previewing/rendering.

monroewilliams:
A better way to improve performance here would be to improve the way the OpenSCAD app handles threading/task management. Forking off multiple processes seems like it would complicate caching and cause other issues, what this really needs is for the render to happen with proper threading (or preferably dispatch queues, which better manage thread pools and relative priority) and have the render operations for different windows not block each other.

jordanbrown0:
MT makes a program much more complicated, and puts all of your eggs in one basket - if one thread hits a bug and crashes, you lose all of your windows/tabs/whatever.

monroewilliams:
Threading/MP is nontrivial to get right, but it's not _that_ hard, and it's necessary for good performance. I'm running on a machine with 24 cores, and I'm waiting ridiculously long times for a render while OpenSCAD pegs a single one of my CPUs. There's no way spawning multiple processes is going to speed up a single render, that needs actual multithreading.

jordanbrown0:
In general, it's *very* hard to get right.

But in limited cases - which this likely is - yes, it's only non-trivial.

kintel:
Closing as not planned

PR #6977, opened 2026-08-25
Process isolation: per-window compute workers, streamed preview products, and structured diagnostics

whosawhatsis:
*User documentation for the wiki is in the last section.*

## The problem

A single typo deep inside a looped structure can generate thousands of errors. Today that turns a
preview that should take under a minute into one that takes hours — not because the geometry is
hard, but because of the cost of pushing each repeated warning to the console.

And it cannot be stopped. Cancellation does not work in that state, so the only way out is to kill
OpenSCAD, which takes the user's unsaved edits with it. The user pays for one mistyped character
with their work.

That single scenario is what this PR is built around, and it is why these changes arrive together
rather than separately. One typo triggers all three failures at once:

- the console-output delay that makes the preview take hours,
- the cancellation that does not respond, and
- the loss of unsaved work when killing the process is the only remaining option.

Fixing any one of them alone still leaves the user stuck. This PR addresses all three, and displays
the repeated errors in a form that makes the underlying mistake far easier to find.

### Why "Stop on the first warning" is not the answer either

OpenSCAD already has a switch for this: `--hardwarnings`, exposed in Preferences as **Stop on the
first warning**. It does prevent the multi-hour case, by throwing `HardWarningException` from
`PRINT()` the moment any warning is emitted. But it is a blunt instrument, and it costs two things
that matter.

**It destroys the diagnostic information you needed.** Stopping at the first warning tells you that
something is wrong at one location. The bundled form tells you the same warning fired ten thousand
times, from which loop, and at which source line -- which is usually what identifies the mistake.
Trading a storm for a single line is trading away the evidence.

**It makes a common editing pattern impossible.** A frequent way to test a change is to leave the
existing default in the file and add an overriding re-definition below it:

```openscad
wall = 2;      // the real value
wall = 3.5;    // temporary, just trying something
```

That is exactly what `parser.y` warns about -- *"wall was assigned on line 1 but was overwritten"* --
and it is a `message_group::Warning`, so **Stop on the first warning aborts the render**. The user
has to go back and comment out the original instead of simply shadowing it, every time, for a
throwaway experiment. Anyone who works this way has to keep the option off, which means the people
most exposed to warning storms are the ones for whom the existing mitigation is unusable.

Structured diagnostics removes the trade: the storm stops costing hours *and* the diagnostic detail
survives, with no change to how you edit.

### Why a second window is not an escape hatch

It is worth being precise about who this hits, because the obvious workaround does not work.

OpenSCAD has no single-instance mechanism on any platform, and `File -> New Window` calls
`new MainWindow(...)` in the current process everywhere. So two windows opened that way share one
process on Windows and Linux exactly as they do on macOS: one window's runaway computation blocks
the others, and killing the application to escape it discards unsaved edits in **all** of them, not
just the offending one.

The platforms differ only in whether a *separate* process is reachable at all. On Windows and Linux
a user who knows to launch `openscad` again from a shell or file manager gets one. On macOS the
bundle is single-instance by OS convention, so a second document is routed into the running process
via `QFileOpenEvent` and there is no escape hatch short of `open -n`.

That makes this worth most to macOS users, but the group it helps is larger than one platform: it is
everyone who reaches for New Window, which is the natural thing to reach for.

## Summary

The mechanism is an opt-in, restart-required **Process Isolation** experimental feature, together
with the two pieces that make it worth having.

The GUI stays a single application process that owns the menu bar, windows, tabs, editors and
viewports. Each window gains one persistent private compute process that does that window's parsing,
evaluation, preview and render work.

This keeps one window's computation from blocking other windows, limits a compute crash to that
window's worker, and lets **Stop** escalate from cooperative cancellation to terminating and
replacing a wedged worker without discarding the GUI's unsaved editor contents. It does not
parallelize or speed up a single render. Tabs in one window share that window's worker and caches;
different windows do not share caches.

Three experimental flags are added, all disabled by default:

| Flag | What it does |
| :--- | :--- |
| `process-isolation` | Runs each window's computation in its own process. Read once at startup, so a session is never a mix of isolated and legacy windows. |
| `streaming-preview` | Starts decoding preview geometry as the worker produces it, rather than waiting for the whole preview. Requires `process-isolation`; no effect without it. |
| `structured-diagnostics` | Streams and collapses worker diagnostics as structured frames instead of rendered stderr lines, and retains the unabridged record. |

Everything above is inert unless the corresponding flag is enabled, and with all three off the GUI
behaves as it does today.

### Concurrent preparation across windows

`OpenCSGRenderer::prepare()` is split so that the expensive middle — building each leaf's vertex
array in host memory, roughly 85% of the phase and entirely GL-free — runs on a worker **thread**,
with only the two small GL halves (`beginPrepare`/`finishPrepare`) on the GUI thread holding the
context. Previously the whole phase sat on the GUI thread and pumped the event loop once per leaf,
which made two windows preparing at once take twice as long as one.

This lives in `MainWindow::actionPreviewDone()`, which is reached only from
`ComputeWorker::previewDone`, so it is part of the isolated path and does not change behaviour with
the feature off. Cancellation and progress cross the thread boundary as `std::atomic<bool>` /
`std::atomic<size_t>`: the worker thread touches no widgets, and a 50 ms GUI-side timer reads the
counter into the progress bar.

## How the pieces fit

**Transport.** Worker results come back over a framed channel on the response stream rather than
through temporary files, with the payload framing in `src/io/ipc_channel.{cc,h}` and the geometry
payload format in `src/io/ipc_geometry.{cc,h}`. Geometry crosses as a binary payload rather than
being re-serialized to text and re-parsed, which is what keeps the hand-back cheap enough that
isolation is not a regression. Payloads are emitted before the terminating control line, so a reader
that has seen `done` has already seen them.

**Streaming.** With `streaming-preview` on, the worker emits each preview leaf as its geometry is
evaluated and the GUI decodes leaves as they arrive, instead of both sides waiting for the whole
request. The flag is read per preview: one already running keeps the setting it started with.

**Diagnostics.** With `structured-diagnostics` on, diagnostics travel as request-aware structured
frames rather than newline-delimited rendered stderr. The parent aggregates, collapses same-line
diagnostic families, and prepares rate-limited console summaries while the worker returns to model
evaluation, and a lossless ordered record is retained for full viewing. The previous stderr path is
what the flag falls back to.

**Worker lifecycle.** Workers are started per window and are cooperatively cancellable, with forced
termination and automatic respawn as the escalation. Worker crashes are recovered without taking
down the GUI process, and stopping a worker does not generate a crash report.

## Testing

New tests, all written before the implementation they cover:

- `tests/test_compute_worker.py`, `_cache.py`, `_parity.py`, `_python.py`, `_channel.py`,
  `_streaming.py` — worker protocol, cross-request geometry cache, export parity across the binary
  hand-back, the Python engine in the worker, channel framing, and streaming.
- `src/io/ipc_geometry_test.cc`, `src/io/ipc_channel_test.cc` — payload layout and framing, so CI
  catches what a round trip alone cannot.
- `tests/test_off_precision.py` — OFF export precision, since geometry precision has to survive the
  worker transport.
- `src/guitests/TestMainWindow.cc` — the GUI-side behaviour: per-window workers, crash recovery,
  cancellation, concurrent isolated windows, Customizer discovery and transport, dependency
  watching, progress display.
- `src/utils/ipc_transport_bench_test.cc` — transport cost benchmark on pre-rendered real models.

The GUI suite runs in both modes: isolated by default, and `OPENSCAD_GUI_TEST_LEGACY=1` runs the
same tests against the in-process path users get with the feature disabled.

Results on the commit this PR opens with:

| Where | Result |
| :--- | :--- |
| GitHub Actions, Linux (ubuntu-22.04 + 24.04, Qt5/Qt6) | pass |
| GitHub Actions, Windows (Qt5 + Qt6) | pass |
| GitHub Actions, macOS 15 Intel (Qt6) | pass |
| macOS local `ctest` | 2736/2739 — the three failures are pre-existing SVG arc-tessellation differences unrelated to this change |
| macOS GUI suite, isolated | 42 passed, 0 failed |
| macOS GUI suite, `OPENSCAD_GUI_TEST_LEGACY=1` | 20 passed, 0 failed |
| Linux local `ctest` (Debian, real GPU) | 2739/2740 |
| Linux GUI suite, isolated / legacy | 41 passed, 0 failed / 19 passed, 0 failed |
| Windows local `ctest` (MSYS2 UCRT64, Mesa software renderer) | 2735/2737 — the two failures are PDF export, which this change does not touch, and are a local code-integrity policy blocking the converter |

**One gap worth stating rather than hiding:** this branch has *not* been verified under `NULLGL`.
`src/core/AIService.cc` on master includes `<QApplication>` unconditionally and is in the
unconditional source list, so `-DNULLGL=1` does not build on master either. That is an upstream
breakage rather than something this PR introduces, but it does mean the usual "does it still link
without GL" check could not be run. Happy to re-run it once NULLGL builds again.

## Reviewer notes

- `Feature::ExperimentalProcessIsolation` is read once at startup and cached in
  `MainWindow::setProcessIsolation()`, deliberately: allowing it to change mid-session would mean a
  window set with two different execution models and two different cache ownership rules.
- Qt 5 and Qt 6 are both supported; worker startup errors have a Qt5-specific path.
- Caches are per worker, so a forced worker replacement discards that window's cached geometry. This
  is a deliberate trade — it is the cost of being able to kill a wedged computation at all.
- `emit` cannot be used as a function name anywhere a header can reach Qt, which is why the payload
  sink API is spelled the way it is.

### Process Isolation (experimental)

Process Isolation gives each OpenSCAD window its own private helper process, which does that
window's computing: parsing your script, evaluating it, generating geometry for preview (F5) and
render (F6), and running Python when the Python engine is enabled. The main application keeps the
menu bar, windows, tabs, editors and 3D views, so your system still shows one OpenSCAD application
and one Dock or taskbar icon.

The point is containment: when computation is stopped or goes wrong, it costs you that window's
computation rather than your session.

The case this is built for: a single typo deep inside a loop can produce thousands of errors, which
can stretch a preview from under a minute to hours — and previously could not be interrupted, so the
only way out was to force-quit OpenSCAD and lose whatever you had not saved. With Process Isolation
that preview can simply be stopped, and your unsaved text stays where it is. Enabling **Structured
Diagnostics** as well addresses the other half, the delay itself.

#### Enabling it

Process Isolation is **experimental and disabled by default.** Enable **Process Isolation** on the
**Features** page in Preferences, then **restart OpenSCAD**. Enabling or disabling it takes effect
only at the next application start, and applies to every window in that run — you cannot mix
isolated and non-isolated windows in one session. With the feature disabled, OpenSCAD behaves
exactly as before.

There is no command-line change: this affects the GUI only, and scripts, exports and command-line
options behave identically either way.

#### What changes for you

- **Stop actually stops.** A long or runaway preview or render can be interrupted, including cases
  that previously ignored Stop entirely.
- **A failure costs one window's computation, not your work.** If the helper process dies — from a
  bug that would previously have taken the whole application down — the window survives with your
  unsaved text intact and a fresh helper starts automatically.
- **Windows no longer wait for each other.** One window computing does not freeze the others.

#### The progress bars, and why a preview shows two

The progress panel tells you which kind of operation is running before you read any numbers:

- **A render (F6) shows one bar spanning the full width.**
- **A preview (F5) shows two bars side by side.**

That difference is not cosmetic — the two operations genuinely have a different number of stages.

**The left bar (and the single render bar) is work in the window's helper process:** parsing your
script, evaluating it, and generating geometry. On a complex model this is nearly all of the time.
Because it happens in a separate process, it can be stopped outright even when it is wedged.

**The right bar is work the main application has to do itself.** When a preview finishes, OpenSCAD
still has to hand each part of the model to your graphics card. That step cannot be delegated to
the helper process, because the connection to the graphics card lives in the main application — so
it gets its own bar, and it counts parts of the model rather than elapsed time.

A render has no second bar because it produces no preview to hand over: F6 gives you finished
geometry, which is displayed directly. That is the whole reason F5 is split and F6 is not.

So a typical preview looks like: the left bar advances while the model is computed and then stops;
the right bar fills as the result is prepared for display; the preview appears. **A pause between
the two is normal** — that is the hand-over between the two stages.

**Either bar may move far more slowly than the other, and that is normal.** The two bars measure
different kinds of work, not equal halves of one job, so their speeds are unrelated and depend
entirely on the model. A model built from a few enormous meshes computes slowly and then displays
almost instantly, so the left bar crawls and the right one flicks across. A model assembled from
hundreds or thousands of small parts can compute quickly and then spend much longer being handed
to the graphics card, so the left bar races and the right one crawls. Neither pattern indicates a
problem.

If you used OpenSCAD before this change, that second kind of model is the one that used to look
like a hang: the old single progress bar covered only the computing stage and was dismissed as
soon as it finished, while normalization and display preparation carried on untracked. The
application appeared to freeze with a completed progress bar, or none at all. That work is now
visible as the right-hand bar, and it can be stopped.

Each bar has a tooltip describing what it measures. The panel only appears if an operation takes
more than about a second, so quick previews finish without any bar being shown.

#### Caches, and what a restart costs

Each window has one helper process and one private set of geometry caches. **Tabs in the same
window share those caches; separate windows do not.** Two windows showing the same model compute it
twice.

Because stopping an unresponsive computation replaces the helper process, that window's cached
geometry is discarded with it, and the next preview or render there starts cold.

#### Current limitations

- **One computation at a time per window.** A window that is already computing will not start a
  second preview or render until the first finishes or is stopped. Different windows *can* compute
  at the same time — that is one of the things Process Isolation buys you.
- **Automatic Reload and Preview still takes an application-wide lock.** While one window is
  auto-reloading, another window's auto-reload will not start. Manual F5 and F6 are unaffected.
- **It does not make an individual preview or render faster.** A single computation does not use
  more processor cores than before.
- If the helper process cannot be started at all, the window reports that; repeated failures report
  that the compute worker has stopped. Both normally indicate a broken or partial installation.

### Streaming Preview (experimental)

Streaming Preview is a companion to Process Isolation. Normally the helper process finishes the
entire preview before handing anything back. With Streaming Preview enabled, it hands back each part
of the model as soon as that part is computed, and the main application starts preparing each one for
display immediately instead of waiting for the whole result.

On a model made of many parts, this overlaps the two stages of a preview that would otherwise happen
one after the other, so the preview appears sooner. On a model made of one or two very large parts
there is nothing to overlap, and it makes no difference.

#### Enabling it

Streaming Preview is **experimental and disabled by default.** Enable **Streaming Preview** on the
**Features** page in Preferences. It **requires Process Isolation** and does nothing at all without
it.

Unlike Process Isolation, this one does not need a restart — but it is read when a preview starts, so
a preview already running keeps the setting it began with. The next F5 picks up the change.

#### What to expect

The two progress bars behave as described above, except that they now overlap: the right-hand bar
begins filling while the left-hand bar is still advancing, rather than starting only when it
finishes. The finished preview is identical either way.

### Structured Diagnostics (experimental)

OpenSCAD can spend a surprising amount of time on warnings, errors and `echo()` output. The case
that hurts is a single typo deep inside a loop: one mistyped character produces the same complaint
thousands of times, and the cost of printing them can turn a preview that should take under a minute
into one that takes hours. The model is not what is slow — the console is. Meanwhile the one message
you needed to read is buried under thousands of copies of itself.

Structured Diagnostics changes how the window's helper process reports these. Instead of sending
finished text one line at a time, it sends the facts and lets the main application do the formatting,
grouping and display on its own schedule. Repeated complaints about the same line are collapsed into
a single entry, and the helper process gets back to computing your model instead of waiting on the
console.

Nothing is thrown away: the complete unabridged record is retained, so a summarized console does not
mean lost output.

#### Enabling it

Structured Diagnostics is **experimental and disabled by default.** Enable **Structured
Diagnostics** on the **Features** page in Preferences. It applies to the output of the window's
helper process, so it is only meaningful with **Process Isolation** enabled.

#### What changes for you

- A model that produces the same warning many times shows one collapsed entry instead of thousands
  of lines.
- The console stays responsive during a diagnostic storm, and the model keeps computing while the
  console catches up.
- With the feature disabled, diagnostics appear exactly as they do today, including the existing
  behaviour of suppressing output after five consecutive identical messages.


jordanbrown0:
### Timing of the change
Without regard to the quality or lack thereof of this proposed, I VERY STRONGLY feel that it is not suitable for merging at this time.  A change of this size and complexity is high risk, and that's just not compatible with trying to get a release out this year.

**After** the release... it should certainly be considered.

### The problems identified

First... I have not examined the proposal in detail, nor have I experimented with it.  The comments here are higher-level.

Although separate processes are certainly technologically advantageous, I worry that (a) this may not be the simplest way to  approach the problem, and (b)  the underlying problem may not be fixed.

The proposal lists three problems that combine to cause the overall problem:
* Messages are slow
* Killing the process loses unsaved changes.
* Cancel doesn't work during evaluation

Agreed, making messages be fast would not solve the problem.  No matter how fast messages are, you can generate more of them then you can wait for - or your processing could just be longer than you can wait for, even without messages.  Truly infinite loops are somewhat difficult to create in OpenSCAD, but very long loops are easy and infinite loops are possible.

Killing the process does not generally cause unsaved changes to be lost.  Every preview or render attempt saves a backup copy of the current file before starting to process it.  That backup remains available.  (Possible gap:  if you have *two* files A and B, and A is unsaved and hasn't yet been backed up, and B has a fatal flaw, and you preview/render B and so trigger the failure case described here, I suspect that changes in A will be lost.)

If cancel worked, then the problem would be completely solved.  You would click the cancel button,, and the processing would be canceled, and you would be happy.

### Approaches to addressing the problems

Now... is moving functionality to a second process the simplest way to make cancel work?  And will it even work to make cancel work?

Messages are slower than they should be, and should be fixed.  But that won't solve the problem.

It **would** be better if there was a direct "open backup" user interface, or even an automatic "recover unsaved changes" mechanism, rather than requiring the user to go find the backup directory.

Which brings us to the interesting one, that the cancel button doesn't work.

The fundamental problem with cancel is that the GUI buttons run in the same thread that the program evaluation runs in.  While the program is being evaluated, the GUI is not responsive.  The evaluation needs to be moved into a separate thread.  (Which could be, but need not be, in a separate process.)

Next, even if the GUI cancel button was handled, and set the wasCanceled flag, neither program evaluation nor preview processing actually checks that flag.  But they could, and the check could be cheap enough that it could be done very frequently.

Next, it is plausible that even if the GUI and the evaluation were split into separate threads, a flood of messages might overwhelm the GUI thread so that it spends all of its time processing messages and updating the message logs, and never get to checking the cancel flag.  This too seems addressable, by ensuring that UI components are higher priority than message handling, or by arranging that the expensive parts of message processing are in the evaluation thread rather than the GUI thread.  (Details would depend on exactly why message processing is slow.)

That last points out a potential problem:  simply moving evaluation to a separate process and returning structured message data does not automatically make cancellation work.  If the main GUI thread is busy reading and processing a flood of structured messages (or preview data!) from the pipe, to the exclusion of processing UI actions, cancellation won't work.  I suspect that you've already experimented and demonstrated that you avoid this problem, but the very changes that would fix it - moving that message and preview data processing to a separate thread - would likely address the hard parts of the high-level problem.

### Performance

Another question is the performance cost of pushing potentially hundreds of kilobytes of program to the worker process, and potentially megabytes of preview data back.  I don't have a feel for how expensive that will be, but it's not free.

### Compatibility

Linux and macOS, as UNIX derivatives, have similar multi-process capabilities, and it's a good bet that similar approaches will work on the two of them.  Windows, however, is a different  beast and I wouldn't be surprised if there are subtle differences that will cause a multi-process approach to encounter different problems there.

### Complexity

Even if everything works, a multi-process approach is more complex.  When the GUI process dies, what causes the worker process to die?  Are the worker processes recreated for each run?  If so, what's the performance impact?  If not, what pushes across any preference changes?  Are there deadlock cases in the protocol?  

### Net

Net, it seems like much simpler changes - primarily, moving evaluation to a separate thread - would address the high-level problem.

### Value of process isolation

Process isolation, both between the GUI and the evaluation thread, between the GUI and the render thread, and between multiple GUI windows, would remain helpful when the application crashes.  But it's pretty rare for evaluation or rendering to crash the application, and the automatic backup mechanism remains as a defense.

### Off-topic:  stop-on-first-warning versus forensic data

While I agree that "stop on first warning" is not the answer to the application hang, I disagree that it's forensically inferior.  A warning tells you the line that the problem was detected on, which is often **not** the line the problem originated on.

Consider this trivial example:
```
function add(a, b) = a + b;

echo(add(1, 2));
echo(add("aaa", "bbb"));
```

Without stop-on-first-warning, the output is:

```
ECHO: 3
WARNING: undefined operation (string + string) in file ., line 1
ECHO: undef
```
Nothing points at the actual problem, which is on line 4.  Contrast with the output with stop-on-first-warning set:
```
ECHO: 3
WARNING: undefined operation (string + string) in file ., line 1
TRACE: called by 'add' in file ., line 4
TRACE: called by 'echo' in file ., line 4
Execution aborted
```
That points at the exact call stack that triggered the problem.

"Stop after a hundred messages" would yield approximately the same forensic information that not-stopping yields,  Yes, a thousand messages may tell you something that a single message doesn't tell you, but the last nine hundred aren't adding much over the first hundred.

### Off-topic:  throw-away experiment case
With respect to the "throw away experiment" case... yes, sort of.  I've never found having to comment out the original setting to be an onerous requirement - not nearly as bad as the problems associated with running past warnings - and the Customizer provides a very simple way to try experimental values for top-level parameters, and to revert them without editing program source.

whosawhatsis:
The other issue, which was sort of glossed-over in my original message, is that by running everything in a single-threaded single process, you can't work on two files simultaneously. Running a preview on one files locks up the interface, so that you can't work on other files while waiting for it. Separating these operations into multiple thread/processes solves this problem. This also means that you can run multiple preview/render operations simultaneously, and not only will they not block each other from starting as they do now, but they process on separate cores, so they benefit from parallelization that has always been sorely lacking in OpenSCAD.

I explored using threads instead of processes for this, but abandoned the idea because there are parts of the code (CGAL, in particular) that are notoriously not thread-safe. @jordanbrown0, you yourself said that this approach was "the clear winner" over threads when I suggested this feature in 2020: https://github.com/openscad/openscad/issues/3520#issuecomment-794421293

I know that this is a big change, and I don't expect the PR to just be waved-through without significant testing. I've been using this for a couple of weeks now though, and it's a huge improvement. I wanted to open it up to comment/testing by others. Note that all of the features added are gated behind experimental feature flags, so they're easy to turn off it there's a problem.

jordanbrown0:
[ Caution:  when you edit a previous Github comment, there's no e-mail notification.  I happened to notice your "I know that this is a big change" comment when I went to write a comment, but otherwise might not have noticed it.  When you want to edit - to fix a mistake, or to add a bit that you'd rather not put in a separate comment, or even to back out something you said, I recommend a three-step process:  (1) add a **new** comment explaining the change, and wait for the e-mail notification to go out (almost immediate), (2) make the change, and (3) delete the new comment added above. ]

Context is everything.

Splitting separate MainWindows into separate processes - the context in #3520 - is more or less easy.  Turn File/New Window into a spawn of the application.  Done.  (Mostly.)  No cross-process coordination.  They just work.  (I say "mostly" because it would be good to automatically handle preference changes well, rather than having one process overwrite the other's changes.)

But what you're talking about here is a **lot** more complex.

whosawhatsis:
I tried exactly that first, and the experience (at least as a Mac user) was horrible. MacOS just isn't built for having multiple instances of the same app running concurrently. This is the same reason I have only actually used the `open -n` option a handful of times in the years since that previous thread where it was pointed out.

jordanbrown0:
Well, OK, but still:  that approach is relatively simple, and is the context for my comment in #3520.  Multiplexing multiple GUI instances (with a single thread?  multiple threads?  multiple processes?) with multiple worker processes is enormously more complex; my comment in #3520 does not apply.

whosawhatsis:
The specific complexity questions you raised have concrete answers in the branch, so let me just give
them:

**Are workers recreated per run?** No. One persistent worker process per `MainWindow`, created in
`MainWindow::setupCoreSubsystems()` and living as long as the window. So there's no per-run process
startup cost, and — importantly — the worker keeps `GeometryCache`/`CGALCache` warm across runs, which
is what makes a repeat render cheap.

**When the GUI process dies, what kills the worker?** Nothing has to. The worker's main loop is
`std::getline(std::cin, command)` on the pipe from the GUI. If the GUI goes away, the pipe closes,
`getline` hits EOF, the loop ends and the process exits. No reaper, no heartbeat, no PID-watching. On
a clean shutdown the destructor sends `quit`, then escalates `terminate()` → `kill()` on 1s timeouts.

**What pushes across preference changes?** Nothing needs pushing, because the worker holds no
persistent settings state. Every request carries everything it depends on: the feature-flag set
(worker does `Feature::enable_all(false)` and then enables exactly what the request names), cache
sizes, color scheme, camera, and the diagnostics-collapse setting. A preference change is therefore
in effect on the next run by construction. The one deliberate exception is that per-request settings
are snapshotted at dispatch, so toggling a flag while a preview is in flight doesn't change the run
that's already going.

**Deadlock cases?** The protocol is deliberately one-directional: requests in on stdin, responses out
on stdout, and the worker never asks the GUI a question, so there's no request/reply cycle to deadlock
on. The one thing that does need to reach a busy worker — cancellation — deliberately does *not* go
through the pipe: it's a flag file that the existing progress callback checks during evaluation. That
sidesteps the whole class of "can't cancel because the worker isn't reading stdin" bugs, which is the
original hang this PR is about. Backstop: if the request hasn't ended 1s after cancel, the GUI calls
`terminate()`.

Crash handling is in the same vein: a worker that dies is respawned with backoff, up to 3 consecutive
failures, and requests still buffered GUI-side (not yet written to the dead process) are replayed to
the replacement, while ones already in flight are reported as failed. That's why the split between
`pendingCount` and `activeRequests` exists.

**On Windows being a different beast** — it is, and you're right to expect subtle differences, but
they're the kind that show up loudly rather than lurking. The ones this branch actually hit, and what
they cost:

- stdout defaults to text mode, which rewrites every 0x0A byte. Ordinary doubles contain those, so a
  binary geometry payload arrives corrupted and every field after the first one decodes as garbage.
  Fix is one `_setmode(_fileno(stdout), _O_BINARY)` in the worker; the control-line readers already
  tolerate the lost CRLF.
- a process's current directory is an open handle on Windows. `do_export()` chdir()s into the
  document's directory and the persistent worker never left, so it pinned whichever document
  directory it last rendered from — and nobody could then rename or delete that directory. This is
  what failed `compute_worker_cache` on Windows CI, where the document sits in a temporary directory
  the test harness then tries to clean up. Fix is a scope guard restoring the worker's own path on
  every exit from the request loop, including the error and cancellation paths.
- the two ends can reach the same file by different routes and spell the separators differently, so
  payload keys have to be compared as paths and folded to '/' rather than compared as strings.

All three are in the branch now, with tests. And this is covered by CI, not just by my say-so: the
existing `windows.yml` matrix (Qt5 + Qt6, Mesa software renderer) runs on every push to the branch,
and I also run the suite locally in a Windows VM.

None of that makes it *simple* — you're right that it isn't, and I'm not claiming otherwise. But the
complexity is bounded and it's concentrated in one file (`ComputeWorker.cc`) plus the worker loop in
`openscad.cc`, rather than smeared across the geometry code the way thread-safety work would be. That
is the real argument for processes over threads here: making CGAL and the evaluation path thread-safe
touches code that is much harder to reason about, and much easier to get subtly wrong, than a pipe
protocol is — and unlike a pipe protocol, you can't put it behind a flag and turn it off.

And the whole thing is behind experimental feature flags — with them off, `MainWindow` builds a
`GeometryWorker` exactly as it does today.

jordanbrown0:
Thanks for the details.  The questions I asked were really more intended as examples of non-trivial problems than as an exhaustive list of them.

Again, I haven't looked at the implementation... I was just scared by the "72 files changed"; that didn't seem limited to a couple of files, and the interactions with logging and preview/rendering processing seemed like they would need to be pretty intrusive.

Note that if the worker has a cache, it needs a flush mechanism.

Anyhow, that's probably all I have to say until next week.

whosawhatsis:
Btw, I agree that this shouldn't block getting a release out. While it's a big usability improvement, just getting people access to Manifold and lazy union without needing to download a nightly is definitely a bigger deal.

t-paul:
I think the feature itself would be great. However, I don't know if anyone will review 169 commits with hand-crafted IPC and additional random stuff in-between (OFF precision? Customizer changes, for testing? Thousands of lines of python code).

I'd really would like to see this feature happening but I'm not sure there's a realistic chance for this huge PR. I'm happy to discuss how to split into smaller steps. I can only speak for myself, but I'm not going to review or approve a PR in the current form.

whosawhatsis:
@t-paul Fair criticism. I have some ideas about how to refactor this, and there is definitely some cruft left over from intermediate stages of development that can be cut (for example, results were originally handed back from the worker process via OFF files, which necessitated the precision increase, but that is no longer used).

The plan is to close this and re-land as three PRs, fresh commits against current master:

1. **Core** — feature flag, `CGALWorker` → `GeometryWorker` backed by a child process, the IPC channel, the geometry codec, Customizer transport, cancellation and crash recovery.
2. **Structured diagnostics** — diagnostic frames instead of rendered stderr lines. This is where the `printutils` change belongs; it can't go first, since on master there's no parent process to aggregate them.
3. **Streaming preview products** — the `CsgInfo` / `OpenCSGRenderer` work.

Cut entirely: the OFF precision change. I'll raise it separately if you think it's worth having on its own.

The `test_pretty_print*.py` changes in here are just an older copy of #6961, which I'd already split out and which landed earlier today — they disappear on a rebase. Same treatment, applied to the rest.

Replacing the hand-crafted IPC with **Boost.Asio**, which is already in the tree via `HTTPClient.cc` — no new dependency, header-only for this, and it drops the framing from ~280 lines to ~30. Spiked and passing on all three platforms. One wrinkle worth flagging: `BOOST_ASIO_HAS_LOCAL_SOCKETS` is defined on Windows, but AF_UNIX there fails at runtime under Asio's IOCP backend (WSAEOPNOTSUPP), so Windows uses `asio::windows::stream_handle` over a named pipe. Same framing either way — the platform difference is one typedef.

The Customizer changes aren't for testing — they're the parameter transport plus a fix for values being lost across the process boundary. I'll explain that properly in PR 1 rather than here.

Does that split work, or would you draw the boundaries differently?

t-paul:
As there already is this one as a baseline/prototype, a good first step would be separating out all (hopefully series of small) refactorings in preparation for the main feature. Main point would be that those don't need any feature flags and could be merged directly. Basically shuffling code around to prepare the boundaries where the separate process would hook into in the next step.

Boost.Asio sounds like a reasonable option, having it already available is definitely a plus as adding libraries is often an annoying step. An earlier multi-process render attempt (#3193) was using boost-serialization.

The part I'm not yet have a clear opinion is if the process separation should be a GUI only feature or if it would be useful to have it available in the core. I suppose there's not many use cases that would need it in the core.

I'm seeing the possibility for going even smaller then putting the core in one PR. Mainly due to the platform dependent nature of this, it may make sense to at least split the first (core) step into 4 parts with preparation first and then separate steps for the 3 main platforms. I don't think there's anyone having a simple way of testing on all the platforms, so it might help to get things tested and merged by platform.

whosawhatsis:
Agreed on all three — refactorings first, no flags, directly mergeable. I'll open that as a separate PR and keep it to moving the boundaries into place, with no behavior change.

On GUI-only vs core: I think it has to be at least partly in core, but for a build reason rather than a use-case one. The worker end lives in `openscad.cc`, which compiles in `HEADLESS` builds where Qt isn't linked at all, so the channel and the geometry codec have to sit in `CORE_SOURCES` regardless of who calls them. Everything above that — the worker's lifecycle, cancellation, the console wiring — is GUI-only and stays there.

On splitting by platform: worth doing for the transport specifically, which is where all the platform-dependent code ends up (one file, one `#ifdef`). I should say I do have all three platforms available — macOS, Linux, and a Windows VM — and I've been running each spike on all of them.

That's already earned its keep. Asio's `accept()` fails with EINVAL for AF_UNIX on macOS while working for TCP in the identical pattern, and working on Linux; it's not a version thing, 1.74 and 1.90 both fail. Raw `::accept()` on Asio's own fd works, so it's in the wrapper. The fix is to not use `accept` at all: `socketpair()` on POSIX, and on Windows a named pipe connected to itself, with the child inheriting one end. Same shape both sides, no listener, no race, and I've verified it end to end on all three.

Thanks for the #3193 pointer — useful context, though it looks like a different problem (parallelising CGAL booleans rather than separating the GUI), and boost-serialization would mean linking a compiled Boost component where Asio needs none.

PR #6979, opened 2026-08-28
Bumped Qt to 6.8.4 for macOS

kintel:
This fixes a build issue under macOS 26.6.2


stefano2734:
QT 6.8.4 is a commercial update, 6.8.3 ist here last to non-commercial.
6.8.4 and later are not  in non-commercial packages like pyside6.
Perhaps an update to 6.9.2 or 6.10.3 can help here.
https://wiki.qt.io/Qt_6.8_Known_Issues
https://wiki.qt.io/Qt_6.9_Known_Issues
https://wiki.qt.io/Qt_6.10_Known_Issues
https://wiki.qt.io/Qt_6.11_Known_Issues (no known issues here)

PR #6985, opened 2026-08-28
Bumped the load-bearing boost library to 1.92.0 for macOS

kintel:


stefano2734:
I see eigen 3.4.0 in changed file.
Here is update to 3.4.1 no problem.
https://gitlab.com/libeigen/eigen/-/releases
Actual is eigen 5.0.1
See PR for eigen 5
https://github.com/openscad/openscad/pull/6219

and more updates to 3rd party updates are possible like

Double conversion 3.3.1 —> 3.4.0
https://github.com/google/double-conversion/releases

PR #6992, opened 2026-08-31
Rename CGALWorker to GeometryWorker

whosawhatsis:
First of a short series of preparatory refactorings for process isolation, as suggested in #6977. These carry no feature flags and no behaviour change, so they can be reviewed and merged on their own merits.

`CGALWorker` has not been CGAL-specific for some time — it evaluates the tree through whichever backend is selected, and Manifold is the default — so the name now points at the wrong one. Renaming it to `GeometryWorker` also gives the boundary a backend-neutral name before anything else hooks into it.

Also initialises three members that were left indeterminate: `GeometryWorker`'s `tree` pointer, and `MainWindow`'s `compileErrors` and `compileWarnings`. Each is assigned before use today, so this changes nothing; it just stops that being something a reader has to verify.

### Scope

```
R  src/gui/CGALWorker.cc -> src/gui/GeometryWorker.cc
R  src/gui/CGALWorker.h  -> src/gui/GeometryWorker.h
M  CMakeLists.txt        (two source-list entries)
M  src/gui/MainWindow.cc (references)
M  src/gui/MainWindow.h  (references, two member initialisers)
```

`grep -rn CGALWorker src/ CMakeLists.txt` is clean.

### Testing

- `ctest -j8` on macOS: 2687/2690, the 3 failures being the known SVG arc tessellation cases (`export-svg_spec-paths-arcs01` and its fill-stroke/fill-only variants), which fail on master too.
- **No test file was modified.** That was the acceptance criterion I held this to: a refactoring that needs a test's expectations changed is not behaviour-free, and would belong in the feature PR instead.

### User documentation

No user-visible surface — this is an internal rename, so there is nothing to add to the wiki.

### DIFF
--- a/src/gui/MainWindow.h
+++ b/src/gui/MainWindow.h
@@ -44,7 +44,7 @@ Q_IMPORT_PLUGIN(QSvgPlugin)
 #endif
 
 class BuiltinContext;
-class CGALWorker;
+class GeometryWorker;
 class CSGNode;
 class CSGProducts;
 class FontListDialog;
@@ -121,8 +121,8 @@ public:
 
   Measurement::Measurement meas;
 
-  int compileErrors;
-  int compileWarnings;
+  int compileErrors = 0;
+  int compileWarnings = 0;
 
   MainWindow(const QStringList& filenames);
   ~MainWindow() override;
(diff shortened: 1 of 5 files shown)


src/gui/MainWindow.h
```diff
@@ -458,7 +458,7 @@ public slots:
   bool procevents{false};
   QTemporaryFile *tempFile{nullptr};
   ProgressWidget *progresswidget{nullptr};
-  CGALWorker *cgalworker;
+  GeometryWorker *cgalworker;
```
kintel: perhaps rename the member name too

kintel (approved):

PR #6993, opened 2026-08-31
import() function:  Derive type from extension.  Allow explicitly specifying.  Reject unknown types.

### LINKED ISSUE #6030: import(file) function should reject non-JSON files
**Describe the bug**
There was a bit of uncertainty about whether you could use the import() function on a DXF file, and an experiment showed that it "worked"... and returned the number 999.  That's all.

It turns out that
* import() unconditionally tries to import JSON.
* DXF files have a general format of a numeric code that describes an entry, followed by the entry on the next line.  (999 is "comment".)
* A simple integer is a complete and valid JSON value.
* Apparently, the JSON reader doesn't mind if there's more in the file, after the first JSON value; it's just ignored.

Net, a DXF file will be read as the code that's on the first line.

Although this behavior is in a sense sensible, it isn't very useful and it could be very confusing.  It would probably be best if this attempt had failed.

**To Reproduce**
```
echo(import("foo.dxf"));
```

**Expected behavior**
An error, probably either because the file extension isn't recognized or because there's more text in the file after the first JSON value.

jordanbrown0:
Allows for future expansion into importing additional file types.

Adds a `type` parameter to the import() function.

The type of the file is the `type` specified, if any.  If none is specified, it's the file extension, if any.

If there is no type specified - neither explicitly nor as part of the file name - or if the type is unknown, it's a warning and the function returns `undef`.

JSON (type="json", extension ".json") remains the only type supported.

Fixes #6030.

### DIFF
--- a/src/core/builtin_functions.cc
+++ b/src/core/builtin_functions.cc
@@ -55,6 +55,7 @@
 #include "version.h"
 // hash double
 #include "geometry/linalg.h"
+#include "handle_dep.h"
 
 #if defined __WIN32__ || defined _MSC_VER
 #include <process.h>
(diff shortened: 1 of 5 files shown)


jordanbrown0:
See the expected-output file for examples of the various error messages.

jordanbrown0:
That last commit or two added some tests for the various file-type cases:
* Unknown extension, no explicit type
* Unknown extension, explicit known type
* Known extension, explicit unknown type
* No extension
* No extension, explicit known type.

t-paul (approved):

jordanbrown0:
Documentation updated: https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/Importing_Geometry#import()_function:_Importing_data

jordanbrown0:
Maybe the file type should be a MIME type, so that we are using somebody else's namespace rather than inventing our own.

(For simplicity, maybe `type` should accept either the file extension or the MIME type.)

PR #6995, opened 2026-08-31
De-duplicate the preview view-mode selection in MainWindow

whosawhatsis:
`csgReloadRender()` and `csgRender()` each contained the same nine lines choosing between thrown-together and preview view modes — identical down to the comment and the `ENABLE_OPENCSG` fallback. This moves them into `MainWindow::selectPreviewViewMode()`.

One naming note: the duplicated comment read "Go to non-CGAL view mode", and I have not carried that into the method name. The choice is between the two non-rendered view modes; which backend produced the geometry has nothing to do with it, and with Manifold as the default the "CGAL" framing has stopped being accurate.

Net -20/+18 lines.

### Testing

- `ctest -j8` on macOS: 2687/2690, the 3 failures being the known SVG arc tessellation cases (`export-svg_spec-paths-arcs01` and its fill-stroke/fill-only variants), which fail on master too.
- No test file was modified.

### User documentation

No user-visible surface — internal refactoring, nothing to add to the wiki.

---

Rebased onto current master and applied jordanbrown0's review suggestion (moving the `ENABLE_OPENCSG` `#ifdef` outside `selectPreviewViewMode`'s branch test). This is also preparatory groundwork for re-implementing #6977 (declined on size) as a reviewable stack of smaller PRs.


jordanbrown0:
Looks basically fine.  Two comments:

* I don't think the word "select" is appropriate.  I would use that word for making a decision or saving a setting, not for acting on that decision.  I would consider just `preview()`.  But I guess there are other parts of the source that think of it as "preview mode" vs "rendered mode", so I guess I don't insist.

* The snippet you deduplicated is:
```
  // Go to non-CGAL view mode
  if (viewActionThrownTogether->isChecked()) {
    viewModeThrownTogether();
  } else {
#ifdef ENABLE_OPENCSG
    viewModePreview();
#else
    viewModeThrownTogether();
#endif
  }
```

Note that with ENABLE_OPENCSG undefined, you always end up in Thrown Together, and so the original test doesn't do anything useful.  I would refactor to:
```
  // Go to non-CGAL view mode
#ifdef ENABLE_OPENCSG
  if (viewActionThrownTogether->isChecked()) {
    viewModeThrownTogether();
  } else {
    viewModePreview();
  }
#else
  viewModeThrownTogether();
#endif
```

whosawhatsis:
Thanks @jordanbrown0. Applied the #ifdef restructure as 7162699cc. Kept the name selectPreviewViewMode to match the existing preview-vs-rendered vocabulary in this file (viewModePreview, previewModeGroup) rather than the old comment's CGAL framing. Rebased onto current master; mergeable is clean now.

t-paul (approved):
Yeah, naming is difficult, I suppose we can look at that separately.

PR #7000, opened 2026-08-31
Normalise parameter set names to NFC

Otzie2023:
# Normalise parameter set names to NFC

Follow-up to #6968, closing the loose end named in its description.

Identifiers reaching the AST are NFC, because the lexer normalises them. Names
that enter OpenSCAD by another route do not go through the lexer and carry
whatever normalisation their author used. Where those two meet, a mismatch is
silent.

## Where the two meet

`ParameterObjects::importValues()` looks parameter set entries up by the name
from the AST:

    auto it = values.find(parameter->name());
    if (it == values.end()) {
      parameter->reset();

A key in any other normalisation misses, and the parameter quietly falls back to
the design default. `openscad.cc` compares the set name from the file against the
one given with `-P`, and on a miss applies no set at all — also quietly. Both
together mean

    openscad -p sets.json -P "Groß" -o part.stl

can export the default dimensions with no warning and a zero exit code, which is
a bad failure mode for a tool whose output gets manufactured.

This cannot happen with a file OpenSCAD wrote itself: those keys come from the
AST and are therefore NFC. It needs a parameter set produced elsewhere — a
generator script, a hand edit, text taken from a macOS source, where the
filesystem hands out NFD — or a `-P` argument pasted from such a source.
Canonically equivalent spellings are indistinguishable on screen, so there is
nothing for the user to see.

## What changed

`normalize_utf8_nfc()` in `core/str_utf8_wrapper.h`, next to the other UTF-8
helpers, using the `g_utf8_normalize()` glib already provides. ASCII returns
immediately without conversion or allocation, which covers every existing
parameter set. A string containing a NUL byte is passed through untouched:
`g_utf8_normalize()` returns a NUL terminated buffer, so normalising one would
truncate it and could collapse two distinct names into a single lookup key.

`ParameterSet::setName()` normalises, rather than each of its callers, so a set
name is NFC whether it came from a file or from the customizer. Parameter names
are map keys with no such choke point, so `ParameterSets::readFile()` normalises
them where they are read. `openscad.cc` normalises the `-P` argument before
comparing.

## Why NFC and not NFKC

Same reason as in #6968: NFKC would fold `µ` into `μ` and `mm²` into `mm2`,
rewriting names the user chose. NFC only unifies spellings that are canonically
the same character.

## Unrelated one-liner

`UnicodeIdentifier.cc` uses `std::hex` and `std::uppercase` but only includes
`<iomanip>`; those come from `<ios>`, which include-cleaner flags. Folded in here
rather than opened as a separate PR. Say the word and I will split it out.

## Tests

`src/core/str_utf8_wrapper_test.cc` covers composition, the canonical singletons,
the ASCII fast path, malformed UTF-8, and the NUL case.

`tests/data/scad/customizer/unicode-identifiers-parameters.scad` with a matching
`.json` whose keys are spelled in NFD is an end-to-end check: the echo output
only shows the set's values if the keys were normalised on read. Without the
change it reports the design defaults instead, so the test does fail when the
normalisation is removed.


t-paul (approved):
Aha, no free lunch then 😀. Thanks for the update!

Issue #6990, opened 2026-08-30
Crash on closing main window: null deref in MainWindow::formatIdentifierToAction during MainWindow destruction (macOS)


infosat2000:
## Summary

On macOS, OpenSCAD reliably segfaults when the main window is closed. The crash happens
during `MainWindow` destruction: the editor `Dock` emits `visibilityChanged` while it is
being torn down, which re-enters `MainWindow::onEditorDockVisibilityChanged()` on an
already-destroyed `MainWindow`, and `formatIdentifierToAction()` then dereferences a null
container pointer.

Reproduced 5/5 times across two snapshots (2026.08.29 and 2026.08.30), including once
with the editor pane already hidden.

## Version / environment

- OpenSCAD 2026.08.29 and 2026.08.30 (both affected)
- macOS 15.7.7 (24G720), x86_64
- MacBookPro14,3 — Intel Core i7, Radeon Pro 560 / Intel HD 630
- Qt 6.8.4 (bundled)
- Installed via Homebrew cask `openscad@snapshot`, and via the official
  `OpenSCAD-2026.08.30.dmg` from files.openscad.org

## Steps to reproduce

1. Launch OpenSCAD (editor pane visible — `view.hideEditor = 0`)
2. Close the main window (Cmd+W, or the red close button)
3. Segmentation fault

Hiding the editor pane first (View → Hide Editor) does **not** avoid the crash — the
`visibilityChanged` signal still reaches the slot during teardown. So this is not simply a
matter of the dock being visible at destruction time.

## Backtrace

```
Exception Type:  EXC_BAD_ACCESS (SIGSEGV)
Exception Codes: KERN_INVALID_ADDRESS at 0x0000000000000000

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   MainWindow::formatIdentifierToAction(std::string const&) const + 229
1   MainWindow::addExportActions(QToolBar*, QAction*) const + 181
2   MainWindow::onEditorDockVisibilityChanged(bool) + 139
3   void doActivate<false>(QObject*, int, void**) + 1590        [QtCore]
4   QDockWidget::event(QEvent*) + 766                           [QtWidgets]
5   QApplicationPrivate::notify_helper(QObject*, QEvent*) + 308
6   QApplication::notify(QObject*, QEvent*) + 517
7   OpenSCADApp::notify(QObject*, QEvent*) + 32
8   QCoreApplication::notifyInternal2(QObject*, QEvent*) + 155
[...]

PR #7005, opened 2026-09-02
Fix use-after-free when a MainWindow is destroyed

danielpunkass:
### Problem

Closing a window can segfault during teardown. Two crash reports with this
stack (macOS 26/27, arm64, Qt 6):

```
MainWindow::~MainWindow
  QWidget::~QWidget -> QObjectPrivate::deleteChildren
    Dock::~Dock -> QDockWidget::~QDockWidget
      QWidget::setParent -> QWidgetPrivate::hide_helper
        visibilityChanged -> MainWindow::onEditorDockVisibilityChanged
          MainWindow::updateExportActions
            MainWindow::formatIdentifierToAction
              exportMap.find()          <-- EXC_BAD_ACCESS, already destroyed
```

The docks are children of the `MainWindow`, so `~QWidget` deletes them, and
deleting a `QDockWidget` reparents and hides it — which emits
`visibilityChanged` and pushes hide events through the window's event filter.
Both re-enter `MainWindow` after its members are gone: C++ destroys members
immediately after the `~MainWindow` body, whereas Qt only severs connections
in `~QObject`, which runs last.

Three dead objects get touched on that path:

- `exportMap` — `onEditorDockVisibilityChanged` → `updateExportActions` →
  `formatIdentifierToAction` (the actual fault)
- `activeEditor` — `onEditorDockVisibilityChanged` dereferences it directly
- `rubberBandManager` — read by `MainWindow::eventFilter`

### Fix

Disconnect the docks and remove the event filter in the destructor body,
while the object is still whole.

### Verification

Reduced to a standalone ~90-line Qt program (no OpenSCAD code) that mirrors
the topology: a `QDockWidget` child, a value member read from the
`visibilityChanged` slot, an event filter on the dock, and `WA_DeleteOnClose`.
Built with AddressSanitizer, it faults with a frame-for-frame match of the
report above; compiling in the two lines from this PR makes it pass.

```
ERROR: AddressSanitizer: SEGV on unknown address 0x000000000018
  #0  std::__hash_table<...>::find<int>(int const&)
  #2  Window::onEditorDockVisibilityChanged(bool)
  #9  QDockWidget::event(QEvent*)
  #13 QWidgetPrivate::hide_helper()
  #17 QDockWidget::~QDockWidget()
  #18 QObjectPrivate::deleteChildren()
  #19 QWidget::~QWidget()
  #20 Window::~Window()
  #23 QObject::event(QEvent*)      [DeferredDelete]
```

### DIFF
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -612,6 +612,17 @@ void MainWindow::updateReorderMode(bool reorderMode)
 
 MainWindow::~MainWindow()
 {
+  // The QWidget destructor deletes our child docks, and deleting a QDockWidget
+  // reparents and hides it, which emits Dock::visibilityChanged and sends hide
+  // events through our event filter. Qt only severs connections in ~QObject,
+  // which runs last, so those slots would re-enter this MainWindow after its
+  // own members (exportMap, activeEditor, rubberBandManager, ...) have already
+  // been destroyed. Sever the links here, while the object is still whole.
+  for (auto& [dock, title] : docks) {
+    dock->disconnect(this);
+    dock->removeEventFilter(this);
+  }
+
   delete this->geometryWorker;
 }
 


danielpunkass:
Closing — opened prematurely, will re-open once reviewed on my side.

Comment from Claude (model: claude-opus-5)

kintel:
The mentioned issue looks very much like #6990

kintel:
..but your fix looks reasonable at first glance

danielpunkass:
Thanks, it does look like that! I closed the PR because while I was testing it, I ran into a different crashing bug that I mistook for this one, so I didn't think the explanation generated above was accurate. I'll reopen this as a proposed fix for #6990 and look into the separate issue separately. I'll also scan the issues for duplicates before I file an issue or submit another PR.

kintel (approved):

jordanbrown0 (approved):

jordanbrown0:
@kintel I disabled your auto-merge in the hope of being able to force past the failing macOS check, but no luck so I re-enabled it.

PR #7006, opened 2026-09-02
Update German translation

Otzie2023:
The German catalogue was last revised in December 2021 and had drifted to
63 % translated. Since `msgfmt` drops fuzzy entries, 254 of 702 strings were
reaching the user interface in English.

| | before | after |
|---|---|---|
| translated | 448 | 655 |
| fuzzy | 77 | 0 |
| untranslated | 177 | 47 |

`msgfmt -c` reports no problems. Only `locale/de.po` changes.

## The fuzzy entries were wrong, not approximate

All 77 were msgmerge guesses matched on string similarity, and none of them
was salvageable:

| msgid | previous msgstr |
|---|---|
| `Page Size` | `Schrittgröße` (step size) |
| `Backend` | `Hinten` (at the back) |
| `Distance` | `Weiterführende Beispiele` (advanced examples) |
| `Copy Color Name` | `Font` |
| `Button 19` | `Knopf 1` |

Harmless while the flag is set, since msgfmt skips them, but anyone running
`msgfmt --use-fuzzy` would get an unusable interface. They are now either
translated or empty.

## 38 strings left untranslated on purpose

These are not display text, and each now carries a translator comment saying
so:

- Values from the `_settings_value` property of combo box entries: `a3`…`a6`,
  `letter`, `legal`, `tabloid`, `micron`, `millimeter`, `centimeter`, `meter`,
  `inch`, `foot`, `model`, `none`, `selected-only`, `landscape`, `auto`. The
  visible label sits in the adjacent `text` property and is translated;
  translating the value would change what gets written to the settings.
- The internal names built in `Settings.cc`: `axis-%d` and `axis-inverted-%d`.
  The display strings `Axis %d` and `Axis %d (inverted)` are separate and are
  translated.
- Qt object names that ended up in the template: `ResetSampleText`,
  `ViewportControlWidget`.

An empty msgstr makes gettext return the original, which is what these need.
That these strings are marked translatable at all looks like a bug in the
`.ui` files and in `Settings.cc` rather than something translators should be
working around, but that is a separate matter.

## Other changes

`Show Scale Markers` was translated as `Axen&einteilung anzeigen`. Five other
entries spell `Achsen` correctly; fixed.

The new export entries for OBJ and POV would both have taken `O` as their
accelerator, which `OFF` already uses. They are now `OB&J` and `POV e&xportieren`,
and no letter repeats in that submenu.

`Ctrl+N` style entries are left untouched: those are QKeySequence strings, not
labels, and Qt renders them as `Strg+N` on a German system by itself. The mouse
configuration labels, which are display text, do use `Strg` and `Umschalt`.

## Scope

Only gaps and wrong entries were touched. Existing translations that read
oddly — `Font` for `Font`, `Trim` for `Trim`, and a number of compounds written
as two words — were left alone; that is a separate pass and would have buried
this diff.

The 47 remaining untranslated strings are the HTML tooltips of the PDF export
dialog plus the marked internal values above.

## A note for a follow-up

A handful of the strings translated here no longer match the source. The
template was generated on 2026-01-13, and accelerator markers have been added
to `MainWindow.ui` since, so for example the template has `Preview` while the
source now says `P&review`. Those translations sit unused until the template is
regenerated, at which point msgmerge will pick them up.

I counted 31 of 105 menu entries in `MainWindow.ui` that are currently absent
from `locale/openscad.pot` and therefore untranslatable in all fourteen
languages. I will open a separate issue about that rather than fold a
regeneration into this PR, since regenerating touches every `.po` file.

(I am a native German speaker, so I was able to check everything)

### DIFF
--- a/locale/de.po
+++ b/locale/de.po
@@ -9,16 +9,16 @@ msgstr ""
 "Project-Id-Version: OpenSCAD 2014.01.05\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2026-01-13 22:53+0100\n"
-"PO-Revision-Date: 2021-12-04 20:17+0100\n"
+"PO-Revision-Date: 2026-09-02 14:00+0200\n"
 "Last-Translator: Torsten Paul <Torsten.Paul@gmx.de>\n"
 "Language-Team: German\n"
 "Language: de\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
 "X-Generator: Poedit 3.0\n"
 "X-Poedit-SourceCharset: UTF-8\n"
-"Plural-Forms: nplurals=2; plural=(n != 1);\n"
 
 #: build/OpenSCADLibInternal_autogen/include/ui_AboutDialog.h:102
 #: src/gui/AboutDialog.h:17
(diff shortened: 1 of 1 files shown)


t-paul (approved):
Thank you!

PR #7011, opened 2026-09-03
Fix VBO deletion landing in the wrong GL context

### LINKED ISSUE #7010: Crash on repaint after rendering in a second window: vertex buffers deleted in the wrong GL context
## Summary

With the same file open in two windows, rendering in one window destroys the
other window's vertex buffers. The other window then segfaults on its next
repaint, inside the driver's client-array vertex path.

## Environment

- OpenSCAD 2026.09.02, built from master (c55e8eff7)
- macOS 27.0 (26A5425a), MacBook Pro M4 (Mac16,6), arm64
- Qt 6.8.4

## Steps to reproduce

Reproduces reliably here (crashes within seconds of launch):

1. Open a `.scad` file, then open the same file again so it is showing in two
   separate windows.
2. Save in the first window, to trigger a render.
3. Switch to the second window (Cmd-`) and save there too.
4. Switch back to the first window. It crashes on the repaint.

## Crash

```
Exception: EXC_BAD_ACCESS (SIGSEGV), KERN_INVALID_ADDRESS at 0x0000000c

0  GLEngine   gleRunVertexSubmitImmediate
1  GLEngine   gleDrawArraysOrElements_ExecCore
3  GLEngine   glDrawArrays_IMM_Exec
4  OpenSCAD   VertexState::draw()
5  OpenSCAD   OpenCSGRenderer::draw(bool, ShaderUtils::ShaderInfo const*)
6  OpenSCAD   GLView::paintGL()
7  OpenSCAD   QGLView::paintGL()
8  QtOpenGLWidgets  QOpenGLWidgetPrivate::render()
```

## Analysis

Each `QGLView` gets its own unshared `QOpenGLContext` — `Qt::AA_ShareOpenGLContexts`
is not set anywhere — so every window has a separate object name space and both
start handing out buffer name 1.

`MainWindow::instantiateRoot()`, `MainWindow::cgalRender()` and
`MainWindow::onTabManagerAboutToCloseEditor()` drop the last reference to the
renderers during compilation. That happens outside `paintGL()` and without
making any context current, so the `glDeleteBuffers()` in
`~VertexStateContainer` runs in whichever context painted last — the *other*
window's. Its identically-numbered buffers are destroyed behind its back.

On that window's next repaint, `glBindBuffer()` names a buffer that no longer
exists, so nothing is bound, and the buffer offsets passed to
`glVertexPointer()` (VBOBuilder.cc) are dereferenced as client-side pointers,
faulting near address zero. The faulting address is the byte offset of whichever
attribute the driver reads first — 0x0 and 0xc in the two reports I have.

Setting `Qt::AA_ShareOpenGLContexts` stops the crash, which confirms the
diagnosis: with one shared name space, a delete always reaches the buffer it was
meant for.

With a single window the same call deletes into no current context at all, where
`glDeleteBuffers()` is a silent no-op, so the buffers leak until the context goes
away.

## Related

- #6947 / #6948 touch the same VBO lifecycle from the performance side. No
  overlap in the fix, but caching buffers across renderer recreation interacts
  with this: such a cache would need to be per-context.
- #5647 may be the same defect on Windows, reported against the editor-tab close
  path, which is one of the three sites above. Unconfirmed — no stack trace in
  that issue.

A fix is prepared: #7011.

---

*Analysis and reporting text by Claude Code.*

danielpunkass:
Fixes #7010.

### Problem

With the same file open in two windows, OpenSCAD segfaults in the driver's
client-array vertex path:

```
Exception: EXC_BAD_ACCESS (SIGSEGV), KERN_INVALID_ADDRESS at 0x0000000c

0  GLEngine   gleRunVertexSubmitImmediate
3  GLEngine   glDrawArrays_IMM_Exec
4  OpenSCAD   VertexState::draw()
5  OpenSCAD   OpenCSGRenderer::draw(bool, ShaderUtils::ShaderInfo const*)
6  OpenSCAD   GLView::paintGL()
```

Reproduced reliably on macOS 27, arm64, Qt 6.8.4:

1. Open a `.scad` file, then open the same file again so it shows in two
   separate windows.
2. Save in the first window, to trigger a render.
3. Switch to the second window and save there too.
4. Switch back to the first. It crashes on the repaint.

### Cause

Each `QGLView` gets its own unshared `QOpenGLContext`, so every window has a
separate object name space and both start handing out buffer name 1.

`MainWindow::instantiateRoot()`, `MainWindow::cgalRender()` and
`MainWindow::onTabManagerAboutToCloseEditor()` drop the last reference to the
renderers during compilation. That happens outside `paintGL()` and without
making any context current, so the `glDeleteBuffers()` in
`~VertexStateContainer` runs in whichever context painted last: the *other*
window's. Its identically-numbered buffers are destroyed behind its back.

On that window's next repaint `glBindBuffer()` names a buffer that no longer
exists, so nothing is bound, and the buffer offsets handed to
`glVertexPointer()` are dereferenced as client-side pointers, faulting near
address zero. The faulting address is the byte offset of whichever attribute
the driver reads first — 0x0 and 0xc in the two reports I have.

With a single window the same call deletes into no current context at all,
where `glDeleteBuffers()` is a silent no-op and the buffers leak until the
context goes away.

### Fix

- Set `Qt::AA_ShareOpenGLContexts`, so all views share one object name space
  and a delete always reaches the buffer it was meant for. This alone stops
  the crash.
- Add `QGLView::withCurrentContext()` and release the renderers through it, so
  their buffers are deleted in the context that created them rather than being
  leaked when no context is current. It saves and restores the previously
  current context, following the existing idiom in
  `QGLView::mouseDoubleClickEvent()`.

The second change addresses the leak by construction; I have not measured it.
The crash fix is the part demonstrated below.

### Verification

Reproduced reliably before the change, and does not reproduce after it —
tested with five windows on the same file, saving and switching between them
repeatedly, rotating the preview in between to confirm the buffers stay valid
across repaints.

All three sites where renderers are released were exercised without
regression: preview (F5) and save, render (F6), and closing an editor tab with
a preview showing.

### Related

- #6947 / #6948 touch the same VBO lifecycle from the performance side. No
  file overlap with this change, but caching buffers across renderer
  recreation interacts with it — such a cache would need to be per-context.
- #5647 may be the same defect on Windows, reported against the editor-tab
  close path, which is one of the three sites above. Unconfirmed, as that
  issue has no stack trace.

### DIFF
--- a/src/gui/MainWindow.cc
+++ b/src/gui/MainWindow.cc
@@ -959,12 +959,16 @@ void MainWindow::instantiateRoot()
 {
   // Go on and instantiate root_node, then call the continuation slot
 
-  // Invalidate renderers before we kill the CSG tree
-  this->qglview->setRenderer(nullptr);
+  // Invalidate renderers before we kill the CSG tree. Renderers own vertex
+  // buffers and delete them from their destructors, so drop them with our own
+  // GL context current.
+  this->qglview->withCurrentContext([this]() {
+    this->qglview->setRenderer(nullptr);
 #ifdef ENABLE_OPENCSG
-  this->previewRenderer = nullptr;
+    this->previewRenderer = nullptr;
 #endif
-  this->thrownTogetherRenderer = nullptr;
+    this->thrownTogetherRenderer = nullptr;
+  });
 
   // Remove previous CSG tree
   this->absoluteRootNode.reset();
@@ -1993,8 +1997,10 @@ void MainWindow::cgalRender()
     return;
   }
 
-  this->qglview->setRenderer(nullptr);
-  this->geomRenderer = nullptr;
+  this->qglview->withCurrentContext([this]() {
+    this->qglview->setRenderer(nullptr);
+    this->geomRenderer = nullptr;
+  });
   rootGeom.reset();
 
   LOG("Rendering Polygon Mesh using %1$s...",
@@ -3150,12 +3156,15 @@ void MainWindow::onTabManagerAboutToCloseEditor(EditorInterface *closingEditor)
   if (closingEditor == renderedEditor) {
     renderedEditor = nullptr;
 
-    // Invalidate renderers before we kill the CSG tree
-    this->qglview->setRenderer(nullptr);
+    // Invalidate renderers before we kill the CSG tree, with our own GL
+    // context current so their buffers are deleted where they were created.
+    this->qglview->withCurrentContext([this]() {
+      this->qglview->setRenderer(nullptr);
 #ifdef ENABLE_OPENCSG
-    this->previewRenderer = nullptr;
+      this->previewRenderer = nullptr;
 #endif
-    this->thrownTogetherRenderer = nullptr;
+      this->thrownTogetherRenderer = nullptr;
+    });
 
     // Remove previous CSG tree
     this->absoluteRootNode.reset();
--- a/src/gui/QGLView.cc
+++ b/src/gui/QGLView.cc
@@ -114,6 +114,20 @@ void QGLView::resetView()
   cam.resetView();
 }
 
+void QGLView::withCurrentContext(const std::function<void()>& fn)
+{
+  // See the discussion at QGLView::mouseDoubleClickEvent() for why the
+  // previous context is saved and restored rather than simply dropped.
+  QOpenGLContext *oldContext = getGLContext();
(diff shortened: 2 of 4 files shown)


kintel:
This looks pretty solid. Thanks!

PR #7014, opened 2026-09-03
Warn when adjusting parameter limits

### LINKED ISSUE #7013: Parameter limits are getting ignored
**Describe the bug**
Parameter limits are seemingly getting ignored, if their value is below or above the declared minimum and maximum. This seems to be a regression from last official release 2021.01.

**To Reproduce**
```
A = 0; // [5:10]
B = 10; // [0:5]
```

For `A` and `B` sliders with the range `[0:10]` are generated.

**Expected behavior**
The parameters should get limited to their specified limits, or a warning should be shown in the error console.

 - affected OpenSCAD versions: 2026.03.11, 2026.08.30
 - unaffected version: 2021.01

hasselmm:
The parameter parser silently adjusts parameter limits (minimum, maximum) to the parameter's specified value if that value is outside the specified limit. To the user this is surpising and looks like the limits are getting ignored.

To avoid this confusion, this change produces a warning that's shown in the console and error log.

Fixes #7013


src/core/customizer/ParameterObject.cc
```diff
@@ -424,11 +430,25 @@ static NumericLimits parseNumericLimits(const Expression *parameter, const std::
...
     }
   }
+
+  // output.minimum/maximum get widened multiple times when processing vectors,
+  // therefore the original values need to be cached for accurate warnings
+  const boost::optional<double> declaredMinimum = output.minimum;
+  const boost::optional<double> declaredMaximum = output.maximum;
+
   for (double value : values) {
     if (output.minimum && value < output.minimum) {
```
jordanbrown0: I think your comment about needing to cache the original value applies here too.  That is, you need
```
if (declaredMinimum && value < *declaredMinimum) {
```

Otherwise, if you have
```
x = [1,2,3];    // [ 4 : 9 ]
```
I think you will only get one warning.

src/core/customizer/ParameterObject.cc
```diff
@@ -424,11 +430,25 @@ static NumericLimits parseNumericLimits(const Expression *parameter, const std::
...
+      LOG(message_group::Warning, location, "",
+          "Parameter '%1$s': value %2$s is below declared minimum %3$s, adjusting minimum",
+          name, formatValue(value), formatValue(*declaredMinimum));
+
       output.minimum = value;
     }
     if (output.maximum && value > output.maximum) {
```
jordanbrown0: Similarly.

jordanbrown0 (changes requested):
Because it looks like this will be the first case where it produces a Warning (and thus possibly throws), please confirm that you have tested all three places that call fromSourceFile():
* `openscad -o foo.param foo.scad`
* `openscad -p foo.json -P myparamset foo.scad`
* GUI

Test with both stop-on-warning (--hardwarnings) set, and not set.  Confirm that with it set processing stops on the first failure, and that with it not-set processing continues and you get each warning message exactly once.

Please add automated test cases.  I think you can test the `-o foo.param` case and the `-p foo.json` case.  (You can't test the GUI case.) Test both the hardwarnings case and the not-hardwarnings case.

That probably means about 5 test cases:
* not-hardwarnings, `-o foo.param`, check resulting params file.  Possible example:
https://github.com/openscad/openscad/blob/a4c283c4012566cd087ded36fae24881f3f042d7/tests/CMakeLists.txt#L994
* not-hardwarnings, `-o foo.param`, check resulting error output.  Possible example: https://github.com/openscad/openscad/blob/a4c283c4012566cd087ded36fae24881f3f042d7/tests/CMakeLists.txt#L318
* not-hardwarnings, `-p file.json -P setname`, check resulting error output as above.
* hardwarnings, `-o foo.param`, check resulting error output.  Possible example: https://github.com/openscad/openscad/blob/a4c283c4012566cd087ded36fae24881f3f042d7/tests/CMakeLists.txt#L1077
* hardwarnings, `-p file.json -P setname`, check resulting error output as above.

Note that these can all probably be simple "expected output" tests, where you supply an expected output file and a mismatch is a failure.  It will probably need one `.scad` file, one `.json` file, one expected `.param` file, and four expected `.echo` files.

I assume that the hardwarnings param case would not generate a param output file and so there's no need to check one.

Make sure you test with multiple warnings present.

src/core/customizer/ParameterObject.cc
hasselmm: Fixed in 34c7516b874327a91bf83bfba91106be625c4ede.

src/core/customizer/ParameterObject.cc
hasselmm: Fixed in 34c7516b874327a91bf83bfba91106be625c4ede.

hasselmm:
> Please add automated test cases. I think you can test the `-o foo.param` case and the `-p foo.json` case. (You can't test the GUI case.) Test both the hardwarnings case and the not-hardwarnings case.
> 
> That probably means about 5 test cases:

Currently trying to add the tests. Those for which you could provide examples were easy to add:
>     * not-hardwarnings, `-o foo.param`, check resulting params file.  Possible example:
>     * not-hardwarnings, `-o foo.param`, check resulting error output.  Possible example:
>     * hardwarnings, `-o foo.param`, check resulting error output.  Possible example:

Still struggling with these, as the errors are unconditionally sent to stderr, not to any .echo file:
>     * not-hardwarnings, `-p file.json -P setname`, check resulting error output as above.
>     * hardwarnings, `-p file.json -P setname`, check resulting error output as above.

Checking right now if this stderr behavior looks intentional, or if this can be changed.

jordanbrown0:
Might be able to catch the messages with -o.

hasselmm:
Please recheck.

src/core/customizer/ParameterObject.cc
```diff
@@ -424,19 +431,34 @@ static NumericLimits parseNumericLimits(const Expression *parameter, const std::
...
   }
+
+  // output.minimum/maximum get widened multiple times when processing vectors,
+  // therefore the original values need to be cached for accurate warnings
+  const boost::optional<double> declaredMinimum = output.minimum;
+  const boost::optional<double> declaredMaximum = output.maximum;
+
   for (double value : values) {
-    if (output.minimum && value < output.minimum) {
+    if (output.minimum && value < *declaredMinimum) {
```
jordanbrown0: It's a nit, but I would use declaredMinimum on both sides of the &&.  It can't happen with the function as currently built, but if output.minimum started out unspecified and later got set to a real value, this "if" would fail - it would see that output.minimum was set, and would then try to use the value of declaredMinimum... which isn't set.

src/core/customizer/ParameterObject.cc
```diff
@@ -424,19 +431,34 @@ static NumericLimits parseNumericLimits(const Expression *parameter, const std::
...
+      LOG(message_group::Warning, location, docPath,
+          "Parameter '%1$s': value %2$s is below the declared minimum %3$s, adjusting minimum value",
+          name, formatValue(value), formatValue(*declaredMinimum));
+
       output.minimum = value;
     }
-    if (output.maximum && value > output.maximum) {
+    if (output.maximum && value > *declaredMaximum) {
```
jordanbrown0: As above.

src/core/customizer/ParameterObject.cc
```diff
@@ -501,7 +523,7 @@ std::unique_ptr<ParameterObject> ParameterObject::fromAssignment(const Assignmen
 
     if (expression->isDouble()) {
       double value = expression->toDouble();
-      NumericLimits limits = parseNumericLimits(parameter, {value});
+      auto limits = parseNumericLimits(name, parameter, assignment->location(), docPath, {value});
```
jordanbrown0, line 521: I don't insist, but I wouldn't use "auto" here.  I'm a strong-typing person.  This function expects a particular type here.  Specifying the type makes it clear what's expected, and checks that the right hand side **is** what's expected (or is at least conversion-compatible).  Even for a human, with "auto" the human who wants to know what the type is has to go look up parseNumericLimits to see what it returns.

src/core/customizer/ParameterObject.cc
```diff
@@ -522,7 +544,7 @@ std::unique_ptr<ParameterObject> ParameterObject::fromAssignment(const Assignmen
       value.push_back(item->toDouble());
     }
 
-    NumericLimits limits = parseNumericLimits(parameter, value);
+    auto limits = parseNumericLimits(name, parameter, assignment->location(), docPath, value);
```
jordanbrown0, line 542: As above.

src/core/customizer/ParameterObject.cc
```diff
@@ -531,9 +553,15 @@ std::unique_ptr<ParameterObject> ParameterObject::fromAssignment(const Assignmen
...
+  // Warnings logged while parsing parameters (see parseNumericLimits) need a
+  // stable docPath: an empty one makes Location::toRelativeString() fall back
+  // to the current working directory (see fs_uncomplete()), which is neither
+  // stable across invocations nor meaningful to the person reading the warning.
+  const auto docPath = fs::path(sourceFile->getFullpath()).parent_path().generic_string();
+
   ParameterObjects output;
   for (const auto& assignment : sourceFile->scope->assignments) {
-    std::unique_ptr<ParameterObject> parameter = ParameterObject::fromAssignment(assignment.get());
+    auto parameter = ParameterObject::fromAssignment(assignment.get(), docPath);
```
jordanbrown0, line 559: As above with respect to "auto".

tests/regression/export-param/issue7013-expected.json
```diff
@@ -0,0 +1 @@
+{"parameters":[{"group":"Parameters","initial":0.0,"max":10.0,"min":0.0,"name":"SmallScalar","step":1.0,"type":"number"},{"group":"Parameters","initial":[5.0,4.0,3.0,2.0],"max":10.0,"min":2.0,"name":"SmallVectorFwd","step":1.0,"type":"number"},{"group":"Parameters","initial":[2.0,3.0,4.0,5.0],"max":10.0,"min":4.0,"name":"SmallVectorRev","step":1.0,"type":"number"},{"group":"Parameters","initial":6.0,"max":6.0,"min":0.0,"name":"LargeScalar","step":1.0,"type":"number"},{"group":"Parameters","initial":[5.0,6.0,7.0,8.0],"max":8.0,"min":0.0,"name":"LargeVectorFwd","step":1.0,"type":"number"},{"group":"Parameters","initial":[8.0,7.0,6.0,5.0],"max":6.0,"min":0.0,"name":"LargeVectorRev","step":1.0,"type":"number"},{"group":"Parameters","initial":[0.0,1.0,2.0,3.0],"max":3.0,"min":0.0,"name":"MixedVectorFwd","step":1.0,"type":"number"},{"group":"Parameters","initial":[3.0,2.0,1.0,0.0],"max":3.0,"min":0.0,"name":"MixedVectorRev","step":1.0,"type":"number"}],"title":"issue7013"}
```
jordanbrown0: The two "rev" cases have wrong answers.  They both pick up the last of the out-of-range values, overwriting the earlier-found further-out-of-range values.

So my earlier suggestion was too simplistic.  The logic needs to be "If smaller than the original declared minimum, warn.  If smaller than the current adjusted minimum, re-adjust.".  (And similarly for maximum.)

tests/data/scad/issues/issue7013.scad
```diff
@@ -0,0 +1,12 @@
+// from https://github.com/openscad/openscad/issues/7013
+
```
jordanbrown0, line 2: Add positive tests please:  scalar that is within the limits, vector with all entries within the limits.

tests/CMakeLists.txt
```diff
@@ -912,6 +913,11 @@ list(APPEND EXPORT_3MF_LAZY_UNION_TEST_FILES ${TEST_SCAD_DIR}/experimental/lazyu
 
 list(APPEND EXPORT_POV_TEST_FILES ${TEST_SCAD_DIR}/pov/pov-export.scad)
 
+list(APPEND EXPORT_PARAM_TEST_FILES
+${TEST_SCAD_DIR}/issues/issue6597.scad
+${TEST_SCAD_DIR}/issues/issue7013.scad
```
jordanbrown0: Indent.

jordanbrown0 (commented):
Thanks.  Very close.  One actual bug (vector can end up with the wrong limit), a couple of coding style suggestions, a request for a positive test, and a tiny formatting cleanup.

jordanbrown0 (changes requested):
That last should have been a "request changes" comment.

src/core/customizer/ParameterObject.cc
hasselmm: You are right, fixed in 6d274fa146b93c656c944c6ea2fd8ff99e4623e5.

src/core/customizer/ParameterObject.cc
hasselmm: Fixed in 6d274fa146b93c656c944c6ea2fd8ff99e4623e5.

src/core/customizer/ParameterObject.cc
hasselmm: Actually `parseNumericLimits()` reports in its very name what type it returns.

Still the actual reason for me breaking my "don't participate in `auto` discussions" rule is the beautify script. It  produces this (IMHO) unpleasant text soup when keeping `NumericLimits` in place:

```C++
    if (expression->isDouble()) {
      double value = expression->toDouble();
      NumericLimits limits =
        parseNumericLimits(name, parameter, assignment->location(), docPath, {value});
      return std::make_unique<NumberParameter>(name, description, group, value, limits.minimum,
                                               limits.maximum, limits.step);
    }
```

src/core/customizer/ParameterObject.cc
hasselmm: As above.

src/core/customizer/ParameterObject.cc
hasselmm: As above.

tests/data/scad/issues/issue7013.scad
hasselmm: Fixed.

tests/CMakeLists.txt
hasselmm: Fixed.

tests/regression/export-param/issue7013-expected.json
hasselmm: Good catch. Would have seen, should have seen that if I had pretty printed the file. Fixed.

hasselmm:
Should be ready for another review round.

src/core/customizer/ParameterObject.cc
jordanbrown0: I would say that the function says what it **does**, not what it **returns**.

And yes, there is always a tension between being terse and hitting line length limits, and I'm not always a fan of the styles people choose to wrap lines.  I particularly dislike the "indent to the open paren" style, because it often spends huge amounts of horizontal space.

But like I said, I don't insist.

src/core/customizer/ParameterObject.cc
```diff
@@ -32,6 +32,11 @@ bool set_enum_value(json& o, const std::string& name, const EnumParameter::EnumI
   }
 }
 
+std::string formatValue(double value)
+{
+  return Value(value).toEchoStringNoThrow();
+}
+
```
jordanbrown0: This isn't necessary.

Use a format string of `%n$g`, or use `STR(number)`.

That is:
```
      LOG(message_group::Warning, location, docPath,
          "Parameter '%1$s': value %2$g is below the declared minimum %3$g, adjusting minimum value",
          name, formatValue(value), formatValue(*declaredMinimum));
```

or

```
      LOG(message_group::Warning, location, docPath,
          "Parameter '%1$s': value %2$s is below the declared minimum %3$s, adjusting minimum value",
          name, STR(value), STR(*declaredMinimum));
```

I mildly prefer the `%g` solution, because it's standard-ish, but there are no other examples of `%g` in OpenSCAD.  There are several examples of `STR(number)`.

jordanbrown0 (changes requested):
Sorry for the delay and the game of fetch-a-rock.  Last round, I think.

src/core/customizer/ParameterObject.cc
hasselmm: Puh, slightly embarrassing, that I didn't think of the "g" specifier myself. It works perfectly. Thank you.

src/core/customizer/ParameterObject.cc
```diff
@@ -424,19 +426,34 @@ static NumericLimits parseNumericLimits(const Expression *parameter, const std::
...
+  const boost::optional<double> declaredMinimum = output.minimum;
+  const boost::optional<double> declaredMaximum = output.maximum;
+
   for (double value : values) {
-    if (output.minimum && value < output.minimum) {
-      output.minimum = value;
+    if (declaredMinimum && value < declaredMinimum) {
+      LOG(message_group::Warning, location, docPath,
+          "Parameter '%1$s': value %2$g is below the declared minimum %3$g, adjusting minimum value",
+          name, value, *declaredMinimum);
```
jordanbrown0: Drat.  Sorry, one more that I shouldn't have missed since I added the feature.

Use `quoteVar(name)`, e.g.
```
      LOG(message_group::Warning, location, docPath,
          "Parameter %1$s: value %2$g is below the declared minimum %3$g, adjusting minimum value",
          quoteVar(name), value, *declaredMinimum);
```

It provides centralized standard markup for quoting a variable name - whatever style we want that to be.

src/core/customizer/ParameterObject.cc
```diff
@@ -424,19 +426,34 @@ static NumericLimits parseNumericLimits(const Expression *parameter, const std::
...
+
+      if (value < output.minimum) output.minimum = value;
     }
-    if (output.maximum && value > output.maximum) {
-      output.maximum = value;
+    if (declaredMaximum && value > declaredMaximum) {
+      LOG(message_group::Warning, location, docPath,
+          "Parameter '%1$s': value %2$g is above the declared maximum %3$g, adjusting maximum value",
+          name, value, *declaredMaximum);
```
jordanbrown0: And here, of course.

jordanbrown0 (commented):
Sigh.  One more, entirely not your fault.  I should have caught it much earlier.

src/core/customizer/ParameterObject.cc
hasselmm: No problem, it's rather me who feels stupid for not noticing these subtleties myself. As this turns more and more into micro-commits: Shall we keep this PR in micro-commits or shall I rather squash them a bit?

jordanbrown0:
Yes, we should squash them.  You can squash them, or I can when I punch the "merge" button.  (But in that latter case I'm not sure whether the resulting commit has your name or mine.)

jordanbrown0:
Also merge with the latest, then rebase.

hasselmm:
Okay, it's rebased to master and squashed into meaning full commits:
- introduce the warnings, but with invalid file names
- route proper document root to parseNumericLimits()
- now finally the tests, squashed with the first commit they'd have broken git bisect
- unrelated change to file list indent

jordanbrown0 (approved):
I probably would have mushed the first three together, but this is fine with me.

Issue #7016, opened 2026-09-03
Mac tests fail because homebrew has dropped support for installing on x86_64


danielpunkass:
Recent PRs have triggered the Mac Experimental Tests, which fail after reaching the "Install Homebrew Packages" phase, presumably because of issues related to the deprecation of Intel support by Homebrew:

https://github.com/openscad/openscad/actions/runs/33710032281/job/100510299627

Often the failure comes in the form of a 90m timeout, which is slowing down the CI process for anything Mac-related. I'm not sure what the best workaround is, but perhaps to avoid the laborious failure, it would be best to introduce some kind of stop-gap, even if it fails, so that the tests fail faster than 90m.

Issue #5201, opened 2024-07-01
macOS software renderer is broken on Apple Silicon


kintel:
Apple's software renderer seems to be broken on Apple Silicon. Rendered results are sometimes garbled. Other projects are reporting the same issue.
For more details, see https://github.com/actions/runner-images/issues/9712.

Possible alternative if the upstream issue isn't resolved is to move towards linking with Mesa on macOS, if possible.

kintel:
Expected solution: Once GitHub offers nested virtualization on M4, we may be able to use pass-through OpenGL HW drivers.

PR #7017, opened 2026-09-03
Stop using the GitHub Mac Intel image until Homebrew incompatibility is sorted

danielpunkass:
This PR simply changes the build machine from macOS15-Intel to the default arm64. The resulting CI fails some tests, but it gets farther along than the Intel-based build machine does, which fails during the preliminary Homebrew install. Short-term workaround for #7016

### DIFF
--- a/.github/workflows/macos-tests.yml
+++ b/.github/workflows/macos-tests.yml
@@ -17,7 +17,7 @@ jobs:
   build_and_test:
     strategy:
       matrix:
-        os: [macos-15-intel, macos-latest]
+        os: [macos-15, macos-latest]
         qt: [qt5, qt6]
         exclude:
           # macos-latest runs on arm64, which has a broken SW renderer


kintel:
See #5201 for bug report.
..and my old attempt at #5204

kintel:
It's a very slippery slope to start disabling tests : (
Some ideas:
* Start looking into using Mesa instead of Apple's OpenGL for CI purposes
* If the failing tests are somewhat stable, we could disable them as a last resort, but perhaps behind a CI flag, and with some fat README + a follow-up ticket to re-enable once runners are better behaved or we have e.g. Mesa support
* Look for alternative runners outside of GitHub. I belive any M3+ runner will give us virtualized Metal, which should be better

danielpunkass:
I agree it is a bad idea to disable tests, but just to be clear the main purpose of this PR is to change the failing tests from a 90 minute timeout failure to a ~10 minute test failure. This PR actually causes more tests (any at all) to be run than currently are :)

.github/workflows/macos-tests.yml
```diff
@@ -17,7 +17,7 @@ jobs:
   build_and_test:
     strategy:
       matrix:
-        os: [macos-15-intel, macos-latest]
+        os: [macos-15, macos-latest]
```
kintel, line 20: Perhaps just keep macos-latest, but remove it from the exclude block below

kintel:
Since Homebrew hasn't actually removed Intel support: Do you know what's actually going on? i.e. is this likely to be a transient or permanent issue?

danielpunkass:
> Since Homebrew hasn't actually removed Intel support: Do you know what's actually going on? i.e. is this likely to be a transient or permanent issue?

Unfortunately I don't really know that much about Homebrew, I just followed up on the error messages I was seeing and it seemed to be correlated to the builds timing out after 90m. I suppose it's possible the warnings include some kind of artificial delay that might be overridable to buy some time?

.github/workflows/macos-tests.yml
danielpunkass: Yeah, that might be a better goal. I ended up auditing that scenario in the hopes it would make the test errors disappear, but it didn't.

PR #7020, opened 2026-09-03
Rename Preferences "Features" tab to "Experimental".

jordanbrown0:
Like the title says, as @kintel suggested.

Note:  the word "Experimental" is, in the font I end up with, a little larger than the icon.  That causes the icons to spread out a little.

Before:
image

After:
image

I think this is better, but if others don't like the slight layout change that's OK.


kintel:
I say yes. @t-paul Any preference?

t-paul (approved):
Fine with me too.

PR #7028, opened 2026-09-06
Update French translation

jordanbrown0:
Note:  translation by a non-French-speaker using Google Translate.  Right now this is just an experiment with the weblate workflow.


jordanbrown0:
I did **not** try to add shortcut markings to the translated menu items, because I didn't have all of the items visible at once.

Issue #6940, opened 2026-08-05
Add a cross-platform API to retrieve the current viewport state


coffeecanary:
### Is your feature request related to a problem? Please describe.

This follows #6741. D-Bus can query some viewport state on Linux, but OpenSCAD documents that interface as Linux-only. On Windows and macOS, external tools cannot retrieve the viewport of a running OpenSCAD GUI session.

This limits AI-assisted and other automated workflows where a user wants a tool to inspect the exact view currently showing an artifact.

### Describe the solution you'd like

A small cross-platform CLI or local API that retrieves the active GUI session’s current viewport state, ideally:
- Translation
- Rotation
- Distance
- FOV
- Projection
- Active document or window identifier

As an interim improvement, a single command that copies the complete viewport state would reduce the current manual overhead. Viewport image capture would also be useful, but complete camera state alone would address the primary need.

### Describe alternatives you've considered

- $vpr, $vpt, $vpd, and $vpf only expose the viewport of the process evaluating the file. A separate CLI invocation receives a new session’s values.
- --camera can reproduce a known view but cannot query the running GUI.
- D-Bus provides partial access on Linux but is unavailable for native Windows and macOS sessions.
- Copying each viewport value from the Edit menu works but interrupts rapid iteration and is error-prone.
- Screen capture or GUI automation is less reliable and does not provide reusable camera coordinates.

### Additional context

My current workflow uses OpenSCAD Nightly on Windows with VS Code and an AI coding assistant. The assistant can edit models, render PNGs through the CLI, inspect those renders, and validate STL/3MF exports. The remaining manual step is communicating a viewport where I notice an artifact.

A read-only command such as openscad --current-viewport returning JSON would close that gap while leaving rendering and analysis to external tools.

jordanbrown0:
There appears to be a D-Bus port to Windows.  The stock builds don't enable it, but my MSYS2 build claims to have it enabled and lets me turn it on.  (Note:  there are three levels of enabling:  (1) whether it's compiled in, (2) whether it's enabled in Edit/Preferences/Features, and (3) whether it's enabled in Edit/Preferences/Axes.  My MSYS2 build has all three.)

I don't know beans about D-Bus and so don't know how to exercise it.  There are hints around that the Windows implementation includes only point-to-point connections and does not include a daemon that allows full functionality.

I would look into that before implementing something new.

coffeecanary:
[...]

PR #7035, opened 2026-09-07
dbus input driver: report viewport state in user-facing values

atirna:
extends the existing D-Bus input driver so the viewport state it reports is actually usable from outside, following the direction in #6940.

**getters now speak the user-facing dialect.** getRotation()/getTranslation() used to return Camera::object_rot/object_trans, the internal camera coordinates. the values Viewport Control, $vpr/$vpt and the Copy viewport values menu entries show are Camera::getVpr()/getVpt(), a different representation. so a getter and the ui could disagree about the same viewport, which is exactly the confusion reported in the issue (Viewport Control [33.30, 0, 148.20] vs dbus [56.7, 0, 211.8]). the getters now return the same values every other interface exposes.

**rotateTo/translateTo now take the same values the getters return.** translateTo had a worse problem: MainWindow::onTranslateEvent scaled X and Z by 0.001 * zoomValue() before applying, so translateTo(10, 20, 30) at distance 140 landed at vpt [1.4, 20, 4.2] and never round-tripped. that scaling is right for relative device motion but wrong for an absolute target, so absolute events now go through Camera::setVpt()/setVpr() (via new QGLView::setVpt/setVpr, which emit cameraChanged like the other view mutators so Viewport Control refreshes). relative translate/rotate behavior is unchanged.

**missing data items**: getDistance() (vpd), getFov() (vpf) and getProjection() ("PERSPECTIVE"/"ORTHOGONAL") through the same chain, plus getTranslation's out arg is now actually named translation in the introspection xml.

**registration failures are honored.** the driver ignored the return of registerObject()/registerService() and reported open either way, so a second OpenSCAD process showed "DBusInputDriver open" in Axes while being unreachable through the well-known name. open() now returns false and logs the reason. multi-session identity itself is deliberately not addressed here, that design question is still open in the issue.

Related to #6940. This covers viewport state; window identity and packaging remain open.

## Testing

built on macOS (qt 6.11, ENABLE_QTDBUS on) and exercised the live service over the session bus with dbus-send:

- getRotation/getTranslation match the Viewport Control panel for the same viewport (before: internal coords)
- translateTo(10, 20, 30) followed by getTranslation returns [10, 20, 30] exactly (before: X and Z scaled by 0.001*distance)
- rotateTo(30, 0, 150) followed by getRotation returns [30, 0, 150]
- getDistance/getFov/getProjection return live vpd/vpf/projection
- a second instance fails registerService and reports not-open instead of a bogus open

this is gui/session-bus behavior so there is no unit test seam for it in this repo (tests are cmdline .scad based and dont reach MainWindow/QGLView). the mechanical half (xml generates a matching adaptor/proxy, clang-format clean) is covered by the Beautify check.

## Scope

- src/gui/MainWindow.cc: getters use getVpt/getVpr, absolute input events set vpt/vpr directly, three new getter forwards
- src/gui/QGLView.{h,cc}: setVpt/setVpr mirroring zoomFov
- src/gui/input/{DBusInputDriver,InputDriverManager}.{h,cc}: new getters through the existing chain, registration failures honored
- org.openscad.OpenSCAD.xml: getDistance/getFov/getProjection, arg name fix, docstrings on the two getters whose meaning changed


src/gui/input/InputDriverManager.cc
```diff
@@ -202,6 +202,33 @@ QList<double> InputDriverManager::getRotation() const
...
 }
 
+double InputDriverManager::getDistance() const
+{
+  const MainWindow *window = currentWindow;
+  if (window) {
+    return window->getDistance();
+  }
+  return 0.0;
+}
```
jordanbrown0, line 208: Seems a bit verbose.  Why not just
```
return currentWindow ? currentWindow->getDistance() : 0.0;
```
?

And similarly below.

src/gui/MainWindow.cc
```diff
@@ -462,14 +462,25 @@ void MainWindow::onTranslateEvent(InputEventTranslate *event)
...
 void MainWindow::onRotateEvent(InputEventRotate *event)
 {
-  qglview->rotate(event->x, event->y, event->z, event->relative);
+  if (event->relative) {
+    qglview->rotate(event->x, event->y, event->z, event->relative);
+  } else {
+    // Absolute rotations are in the user-facing vpr format used by $vpr,
+    // Viewport Control and the Copy/Paste viewport values.
+    qglview->setVpr(event->x, event->y, event->z);
+  }
```
jordanbrown0: There's a similar compatibility concern here.

src/gui/MainWindow.cc
```diff
@@ -462,14 +462,25 @@ void MainWindow::onTranslateEvent(InputEventTranslate *event)
...
   if (event->viewPortRelative) {
     qglview->translate(event->x, event->y, event->z, event->relative, true);
-  } else {
+  } else if (event->relative) {
     qglview->translate(zoomFactor * event->x, event->y, zoomFactor * event->z, event->relative, false);
+  } else {
+    // Absolute translations are in the user-facing vpt format used by $vpt,
+    // Viewport Control and the Copy/Paste viewport values, so they must not be
+    // scaled by the zoom factor.
+    qglview->setVpt(event->x, event->y, event->z);
```
jordanbrown0: I am a little worried about the change here and below in onRotateEvent.  There are rumors of people using DBus as a way to implement an add-on input device, and if they are not setting event->relative and are depending on the current behavior, they could break.

This also in theory changes the behavior of other input devices (SpaceNav, gamepads); they need to be inspected for impact and preferably tested.  (I have a couple and can do some testing.)

I haven't wrapped my head around the semantics enough to truly understand what's right.  If the current behavior is truly wrong, then it should get fixed and consumers should get fixed to match.

If the compatibility hit is just too large, the workaround is not too onerous:  deprecate the current methods, leave them unchanged, and add new "right" methods called something like `translateTo2`.

src/gui/MainWindow.cc
```diff
@@ -1451,20 +1462,29 @@ void MainWindow::on_editActionCopyVPF_triggered()
...
-  ret.append(qglview->cam.object_rot.z());
-  return ret;
+  const auto vpr = qglview->cam.getVpr();
+  return {vpr.x(), vpr.y(), vpr.z()};
+}
+
+double MainWindow::getDistance() const
+{
+  return qglview->cam.zoomValue();
+}
```
jordanbrown0, line 1478: It's annoying that half of the program calls this value "zoom value" and the other half calls it "distance".  We should really change one side or the other.  (And particularly silly is that `Camera`, which is what exports the zoomValue() function, internally calls it "distance".)

But that's not this project.

src/gui/MainWindow.cc
jordanbrown0, line 471: I wondered why isn't this just a call to qglview->translate() with viewPortRelative=false and relative=false, possibly with the values negated.  And it probably could be.  However, there's something to be said for making the vpt/vpr model directly visible, with the difference hidden at the bottom.

src/gui/MainWindow.cc
```diff
@@ -462,14 +462,25 @@ void MainWindow::onTranslateEvent(InputEventTranslate *event)
 
   if (event->viewPortRelative) {
     qglview->translate(event->x, event->y, event->z, event->relative, true);
-  } else {
+  } else if (event->relative) {
     qglview->translate(zoomFactor * event->x, event->y, zoomFactor * event->z, event->relative, false);
```
jordanbrown0, line 466: Do we understand why zoomFactor is involved here, or why it is applied to X and Z but not Y?  I think this is the path that most input device cases come through, so it seems like it must be meaningful.

jordanbrown0 (commented):
I am a little worried about the potential compatibility hits here.

It is not clear to me how many real DBus consumers there are, or what their usage looks like.  It might be good to get feedback from @MichaelPFrey and @Lenbok, who seem to be DBus users.

If the compatibility hit is too large, the workaround does not seem onerous:  deprecate the existing methods and introduce new methods with new names - probably with "2" appended.


In theory this is a compatibility hit for InputEventTranslate with viewPortRelative=false and relative=false, and for InputEventRotate with relative=false.  We need to inspect other input event drivers to see if they can ever use those variants.

atirna:
the ternary is in already, 1c503ab6 covers the three getters this PR adds (`getDistance`, `getFov`, `getProjection`). left `getTranslation`/`getRotation` in the long form since theyre pre-existing and not part of this change, can add if needed.

**on the compatibility hit**, i went and checked which drivers can actually construct the affected variants. it comes out  to only the dbus driver.

every `InputEventTranslate` / `InputEventRotate` construction in the tree:

| site | relative | viewPortRelative | branch |
|---|---|---|---|
| `InputEventMapper.cc:126` | true (default) | false | unchanged |
| `InputEventMapper.cc:135` | true | true | unchanged |
| `InputEventMapper.cc:144` (rotate) | true (default) | n/a | unchanged |
| `DBusInputDriver.cc:119` | true | false | unchanged |
| `DBusInputDriver.cc:124` | **false** | false | **changed** |
| `DBusInputDriver.cc:104` (rotate) | true | n/a | unchanged |
| `DBusInputDriver.cc:109` (rotate) | **false** | n/a | **changed** |

both ctors default `relative = true` (`InputDriverEvent.h`), and none of SpaceNav, Joystick, HidApi or QGamepad construct these events at all, they all go through `InputEventMapper`:

```
SpaceNavInputDriver      direct events=0
JoystickInputDriver      direct events=0
HidApiInputDriver        direct events=0
QGamepadInputDriver      direct events=0
```

so the only consumers that can reach the changed paths are dbus callers, which is the api this PR is deliberately changing. that doesnt tell us how many real dbus consumers exist, so @MichaelPFrey @Lenbok input is still worth having, but it does mean no other input driver is affected.

happy to go the deprecate-and-add-`*2` route instead if they'd rather not break it.

**on the zoomFactor question** (`MainWindow.cc:466`), that line is master's which PR doesnt touch. we only split the old single `else` into `else if (event->relative)` with the original line verbatim, plus a new `else` for the absolute case.

for what its worth i couldnt find a justification for the asymmetry either. `QGLView::translate` (`src/gui/QGLView.cc:522`) builds one rotation matrix and applies it to all three components identically, so nothing downstream treats Y specially. that line predates the 2022 `src/gui` move. out of scope here.

Lenbok:
This seems fine in terms of not impacting my existing use at all, which to date has only been setters rather than getters.

src/gui/MainWindow.cc
atirna: we only split the old single `else` into `else if (event->relative)` carrying the original line verbatim, plus a new `else` for the absolute case.

i couldnt find a justification for the asymmetry either. `QGLView::translate` (`src/gui/QGLView.cc:522`) builds one rotation matrix and applies it to all three components identically, so nothing downstream treats Y specially. it predates the 2022 `src/gui` move. looks like its own bug, haven't touched here

src/gui/input/InputDriverManager.cc
atirna: done in 1c503ab6, all three getters this PR adds use the ternary now. left `getTranslation`/`getRotation` long-form since theyre pre-existing.

coffeecanary:
Built and tested PR 7035 at commit `1c503ab61dca4abd52d64f1f01111cd92a9cab74` using the separate MSYS2 build at `srcdir/builddir-pr-7035/openscad.exe`.

### Viewport Getters

Live D-Bus introspection exposed all expected methods, including the new getters:

- `getDistance()`
- `getFov()`
- `getProjection()`

The initial viewport returned:

```text
Translation: [0, 0, 0]
Rotation:    [55, 0, 25]
Distance:    140
FOV:         22.5
Projection:  PERSPECTIVE
```

The absolute setter/getter round trip succeeded exactly:

```text
translateTo(10, 20, 30) -> getTranslation() = [10, 20, 30]
rotateTo(30, 0, 150)    -> getRotation()    = [30, 0, 150]
```

The viewport was then restored to translation `[0, 0, 0]` and rotation `[55, 0, 25]`, and both restored values were confirmed through the getters.

A manually rotated and translated viewport was also compared directly with **Window > Viewport Control**:

| Value | Viewport Control | D-Bus getter |
| --- | --- | --- |
| Translation | `[10.80, -3.59, 15.70]` | `[10.7996, -3.58939, 15.7013]` |
| Rotation | `[247.70, 0.00, 355.60]` | `[247.7, 0, 355.6]` |
| Distance | `140.00` | `140` |
| FOV | `22.50` | `22.5` |
| Projection | Perspective | `PERSPECTIVE` |

The translation difference is only display rounding; all values agree at the precision shown in the UI. This verifies on Windows that the PR fixes the old absolute translation scaling problem, uses the user-facing viewport representation, and supplies the missing distance, FOV, and projection getters.

### Process Registration

Two independent PR-build processes were then started. A second process left at the startup chooser was not a valid test: the source shows that the chooser completes before `MainWindow` is created, the D-Bus driver is registered, and `InputDriverManager::init()` runs.

The second process was therefore launched with a SCAD file so it initialized fully. The first process retained ownership of `org.openscad.OpenSCAD`, and calls through that name continued to return the first viewport. The second process connected as `:1.29` but did not leave a callable object: a direct getter call returned `org.freedesktop.DBus.Error.UnknownObject`. This confirms that the PR unregisters the second application object after failing to acquire the well-known service name.

In the initialized second process, **Edit > Preferences > Axes** is expected to show:

```text
DBusInputDriver not open
```

The console-captured error was:

```text
ERROR: DBusInputDriver: could not register service:
```

Qt supplied no detail after the colon. The registration failure and object cleanup are verified on Windows; the missing error detail is a remaining diagnostic limitation.

No message appeared in the GUI Console or Error Log, which is consistent with the startup order. `MainWindow` uses a scoped output handler during construction and clears it before `InputDriverManager::init()` opens the D-Bus driver. With no GUI output handler active at registration time, `LOG(...)` writes the error to standard error instead of either dock.

PR #7018, opened 2026-09-03
Translation update (Template only)

t-paul:

PR #7036, opened 2026-09-07
Update German translation 

Otzie2023:
The template regeneration in #7018 exposed the strings that had been missing from the catalogue, and msgmerge matched the old translations to the new msgids by similarity. That left 65 fuzzy and 88 untranslated entries, 153 of 774 reaching the interface in English.

Confirm or replace all 65 fuzzy entries. Most were correct translations whose msgid had only gained an accelerator marker. Two were actively wrong and would have broken a shortcut had anyone accepted them: Ctrl+Alt+Up and Ctrl+Alt+Down had both been matched to "Ctrl+Alt+F". Both are now left untranslated, together with the other key sequences and object names, each carrying a translator comment.

Fix errors inherited from the old catalogue that were invisible until now because the msgids did not match: "Datei öffen" and "In neuem Fenster öffenen" for "open", "Nächsten Reitern anzeigen" for "Show Next Tab", "Alles Speichern" for "Save All", and several compounds written as two words.

Translate the 88 remaining strings: the AI chat widget, the AI preferences page, the network settings, and the tooltips of the PDF export dialog.

Assign accelerators per menu rather than per string. Every menu is now free of duplicates, which the previous catalogue was not: Redo and Find Next both used W, Cut and Use Selection for Find both used A, Copy and Find Previous both used K, About and Offline Cheat Sheet both used O. The Edit menu has 31 entries and German offers fewer distinct initial letters than that, so nine of them carry no accelerator; none of them collide.

Result: 755 translated, 19 untranslated, the latter all key sequences, object names and fixed values that must stay as they are. msgfmt -c reports no problems.

### DIFF
--- a/locale/de.po
+++ b/locale/de.po
@@ -9,7 +9,7 @@ msgstr ""
 "Project-Id-Version: OpenSCAD 2014.01.05\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2026-09-06 23:38+0200\n"
-"PO-Revision-Date: 2026-09-02 14:00+0200\n"
+"PO-Revision-Date: 2026-09-07 12:00+0200\n"
 "Last-Translator: Torsten Paul <Torsten.Paul@gmx.de>\n"
 "Language-Team: German\n"
 "Language: de\n"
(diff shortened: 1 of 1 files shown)


jordanbrown0:
Key sequences should probably stay the same, but translating them is allowed.

Things that must be constant should be marked as not translatable.  Note that for things in .ui files you must run a build before the changes are reflected in the .h files that the localization stuff consumes.

t-paul (approved):
Dankeschön!

PR #7041, opened 2026-09-12
WIP: Editor: fold (...) blocks and whole function definitions

### LINKED ISSUE #7040: Editor: fold (...) blocks and whole function definitions
**Is your feature request related to a problem?**
Code folding in the built-in editor currently only understands `{}` and `[]`. Two common constructs have no way to collapse:

- Any multi-line call or argument list that only uses plain parentheses — e.g. a multi-line `let(...)` definition, `object(...)` call (see OEP8a, the experimental object() builtin), or any other function/module call whose arguments span several lines.
- A `function name(...) = expr`; definition. Since function bodies are pure expressions with no braces, even a long, deeply-nested definition (e.g. one built around `let(...)`) has no single fold point that collapses the whole thing — only its own internal `([` sub-expressions fold individually.

- Treat `(` and `)` the same as the existing `{}` and `[]` bracket-depth counting in `ScadLexer2::fold()`.
- Additionally fold the entire span of a `function ... = ... ;` definition as one region, from the function keyword to its terminating semicolon. Since a function body cannot itself contain a semicolon the first semicolon following the keyword unambiguously terminates it.

I've also considered a more general `#region`/`#endregion`-style comment marker (as in VS Code), but that's out of scope here. This issue is only about the two gaps above, which are purely syntax-driven.

**Additional context**
No test infrastructure currently exercises ScadLexer2; a data-driven QtTest suite (`src/guitests/TestScadLexerFolding`) is included in the linked PR.

hasselmm:
Two small, stacked changes to `ScadLexer2::fold()` (`src/gui/ScadLexer.cc`), plus tests.

This fixes #7040.

## 1. Fold parenthesized blocks

`fold()` only counted `{}` and `[]`. Adds `()` to the same counter, under the same `OtherText` style guard, so a multi-line call or argument list now folds like any other bracketed block. Single-line parenthesized expressions still net to zero by end-of-line, so this doesn't add spurious fold markers to ordinary one-line calls.

## 2. Fold whole function definitions

Function bodies are pure expressions and can never contain a bare semicolon, so the first semicolon following a function keyword is always that definition's own terminator. This is tracked with a small piece of extra state layered on top of the bracket counter, opened when the function keyword is confirmed and closed on that terminating semicolon.

Known limitation: this relies on a single `fold()` call seeing both the function keyword and its terminating semicolon. That holds for a full re-fold (file open, "Fold All", or the `recolor()` that already runs after every parse/evaluate in `MainWindow::parseDocument()`), but if Scintilla ever restyles a narrow range starting strictly inside a very long multi-line function body, this could theoretically leave the fold level unbalanced until the next full reflow. Given how often `recolor()` already runs in normal use, I think this is an acceptable trade-off, but flagging it for review rather than burying it.

## Tests

`src/guitests/TestScadLexerFolding` — data-driven test. Since there's no headless lexer test infrastructure and `fold()` needs a live `QSciScintilla` widget.


hasselmm:
Hopefully this comment links this pull request to issue #7040.
This fixes #7040.

jordanbrown0:
Mentioning an issue or PR (#xxx) gives you a weblink and causes a "mentioned this" in the target, and that happened.

But if you want them to really truly be linked, so that merging the PR automatically closes the issue, there are only a few key patterns that work.

https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue

The key point there is that for keyword-based linking, the pattern must be in the PR description or a commit (not, as here, a comment).

You can also explicitly use the "Development" item in the right sidebar.

hasselmm:
Adding to the issue description did the trick.

t-paul:
I wonder if there are additional ways to close the `function`, right now it mostly jumps straight to the end while the closing `;` is not there yet.

hasselmm:
> I wonder if there are additional ways to close the `function`, right now it mostly jumps straight to the end while the closing `;` is not there yet.

Yes, and I don't like it. Still we currently have exactly the same behavior for curly braces and brackets. They also expand until end of file if not closed correctly. The folding parser operates at symbol level, not at semantic level. It doesn't evaluate the grammar and therefore fails on semantic errors.

Still I am experimenting right now with a recovery mechanism for `function`, as unfinished functions are trivial to discover: If you see a keyword at same folding level as the initial `function` keyword, then you are dealing with an unfinished function. Problem right now:
- The code's asymmetry when handling this kind of errors — with this fallback increments and decrements of `currentLevel` are unbalanced, which actually is correct, but appears like code smell in such parsers.
- WIth my current code the rest breaks the next folding section. I'd have to backtrack to cleanly recover, but would like to avoid this complexity.

t-paul:
If you have an idea that does not make everything massively complicated, that would be great. Otherwise, I would not see it as a showstopper.

It looks like the extended folding could be useful and in the (I believe) unlikely case someone reports a very bad use case breaking things, we can back-out GUI changes much easier compared to things affecting script processing.

hasselmm:
Just a short update: I have this now. Just have to add some final regression test  for editing behavior, but it's really late now. Therefore will stop here for today.

With unfinished function the fold extends to the next function's end, instead of end of the file:




After adding the missing semicolon folding gets repaired:

PR #7044, opened 2026-09-15
Clean up color-index handling in VBOBuilder and ManifoldGeometry

whosawhatsis:
## Summary

De-duplicates the two nearly-identical lambdas in `ManifoldGeometry::toPolySet()`
(`getFaceFrontColorIndex` / `getFaceBackColorIndex`) into a single
`getCachedFaceColorIndex(cachedIndex, renderColor)` helper.

This is one of the small preparatory refactorings requested by @t-paul in
#6977 to shrink the eventual process-isolation PR — see that discussion for
context.

Note: an earlier revision of this PR also changed `VBOBuilder::create_surface`'s
local `color_index` variable's type (first to `int32_t`, then to
`std::optional<size_t>` per @jordanbrown0's review). That change has been
reverted here — `src/glview/VBOBuilder.cc` now matches `master` exactly — because
it turns out this index is about to need to carry more than a present/absent
distinction (a compute-worker design in progress will give it two additional
sentinel states), which `std::optional<size_t>` cannot express. A purpose-built
type for that will be proposed as its own, separate PR.

## Testing

- `ctest -j8`: only the 3 known pre-existing macOS SVG-arc-tessellation failures.
- `./scripts/beautify.sh --check` (clang-format 18.1.8): clean.

User documentation for the wiki is in this section — there is none needed here.
This is an internal refactoring only: it de-duplicates two lambdas in geometry
export code. There is no change to behavior, output, or any user-facing
interface.


jordanbrown0:
My opinion is that the specified-size types (e.g. int32_t) should be used **only** if there is some reason why the values must be that particular size - typically, to conform to a file format or data protocol.  Abstractly, they tell the reader and the compiler that those values **must** be that particular size - which isn't true in most cases.  Concretely, they may unnecessarily force the compiler to use non-native mechanisms to force the specified semantics, and may in some cases (but not here) unduly restrict the range of the values.

I agree that relying on wrapping -1 around to SIZE_MAX is bad.  The minimum-change solution is to use the signed variant of size_t, ssize_t, for the purpose.

The better answer is to use a custom type, say `color_index_t`, for the purpose, rather than locking down the type across all of its uses.

But the next question is how to represent "missing".  A sentinel value (e.g. -1) has several problems.  First, it can force what is fundamentally a counting number, an unsigned type, (how many things?) to be a signed type, with all of the type-unsafety that implies.  Second, even if you use an unsigned sentinel value (e.g. SIZE_MAX), it's easy to involve the sentinel value in some operation where it is inappropriate.  ISTM that `std::optional<size_t>` is the right underlying type here.  Then you don't need any casts at all, and if you ever try to actually *do* anything with a missing value, it'll fail.

whosawhatsis:
You're right about the sentinel, and I tried `std::optional<size_t>` first — but working through it
surfaced a constraint that I think makes this a design decision worth settling before anybody writes
more code, rather than something to iterate on inside this PR.

"Missing" is not one state here, and is about to be three. A color index that is not an index into
`colors` can mean:

- **no specific color** — today's `-1`, on CGAL and imported faces, which are deliberately exported
  colorless;
- **use the viewer's default face color**;
- **use the viewer's cutout (subtracted-face) color**.

Upstream needs only the first today, so `std::optional<size_t>` fits — `nullopt` and "no specific
color" coincide. That stops being true in the process-isolation work I have in flight. There, geometry
is computed in a separate worker process, and the worker must NOT resolve Manifold's implicit face
colors against its own color scheme: it has no business knowing which scheme the window is displaying,
and if it baked them in, changing the color scheme in the GUI would force a round trip to the worker
to recompute geometry that has not changed. So the worker tags those faces and the window resolves
them against its own scheme at draw time:

```cpp
// PolySet.h
static constexpr int32_t COLOR_INDEX_CUTOUT  = -2;
static constexpr int32_t COLOR_INDEX_DEFAULT = -3;
```

`-1` keeps its existing meaning precisely so that behavior does not change — resolving `-1` would
colorize the CGAL and import faces that have always come out colorless. `create_surface()` then has to
tell the three apart, which `std::optional<size_t>` cannot express: every negative collapses into one
`nullopt`.

So of your two suggestions, the custom type is the one that survives what is coming, and I would
rather we settle on it now than land something we both know gets rewritten. Concretely: a
`color_index_t` wrapping the stored `int32_t`, with named states and an `std::optional<size_t> index()`
accessor for "is this an actual index into `colors`" — your `optional` where it genuinely means one
thing, and no casts at the call sites. It touches every consumer (`VBOBuilder`, `export_off`,
`export_pov`, both 3MF exporters, `ManifoldGeometry`), so it wants to be its own PR rather than a
rider on a cleanup.

So I have **dropped the `VBOBuilder` hunk from this PR** — that line is now exactly as master has it.
What remains here is the `getCachedFaceColorIndex` de-duplication in `ManifoldGeometry`, which is
independent of all of the above and can be reviewed on its own.

`color_index_t` will follow as its own PR, designed once, ahead of the isolation PR rather than after.
I will link it here when it is up.

If you would rather the type work happen right here instead, I am glad to do that — I would just like
the shape agreed before I write it. And if your idea of the named states differs from mine, now is the
moment to say so: I would rather build to your design than to mine and rework it.

On the `int32_t`s in `ManifoldGeometry::toPolySet()`: those write straight into
`PolySet::color_indices`, which is `std::vector<int32_t>`, so the width there is the container's rather
than a choice — which I take to be the case your comment carves out. If `color_index_t` happens, they
become that type.

whosawhatsis:
As promised: #7048 has the `color_index_t` type. It keeps `PolySet::color_indices` layout-identical (a single `int32_t` under the hood, `static_assert`-pinned), names the sentinels, and exposes `std::optional<size_t> index()` as the only way to get an actual index — so your `optional` is there where it genuinely means one thing, and no call site casts. Happy to take it in a different direction if you would rather the states be spelled differently.

PR #7047, opened 2026-09-16
Remove MainWindow's own event filters before its members are destroyed

whosawhatsis:
*User documentation for the wiki is in the last section.*

## Summary

Closing a window reads freed memory.

`MainWindow` installs itself as an event filter on itself (`MainWindow.cc`, `installEventFilter(this)`)
and on the find panel, and neither filter is removed when the window is destroyed. Destroying a member
such as `rubberBandManager` reparents the widget it owns; reparenting delivers events; those events
reach `MainWindow::eventFilter()`, which calls `rubberBandManager.isVisible()` — on the member that is
at that moment being destroyed.

Under AddressSanitizer this is a `heap-use-after-free`. In an ordinary build the freed block is usually
still intact, so the read silently returns stale data, which is why it has gone unnoticed.

```
ERROR: AddressSanitizer: heap-use-after-free
  READ  RubberBandManager::isVisible()             RubberBandManager.cc:25
        MainWindow::eventFilter(QObject*, QEvent*) MainWindow.cc:1619
        QCoreApplicationPrivate::sendThroughObjectEventFilters
        QObjectPrivate::setParent_helper
        QObject::~QObject
  FREED ~RubberBandManager()                       RubberBandManager.h:13
        MainWindow::~MainWindow()                  MainWindow.cc:627
  ALLOC QRubberBand ctor, RubberBandManager::RubberBandManager(MainWindow*) RubberBandManager.cc:6
```

## The fix

`~MainWindow` already severs the dock connections and dock event filters for exactly this reason, with
a comment describing the hazard. That fix simply missed the two filters the window installs on itself:

```cpp
removeEventFilter(this);
if (find_panel) find_panel->removeEventFilter(this);
```

Guarding inside `eventFilter()` instead would leave the re-entrancy in place and require a
"being destroyed" flag; removing the filters ends it at the source and matches the existing code.

## How this was found

Running the GUI test suite under AddressSanitizer while investigating an unrelated test-infrastructure
issue. It is independent of that work: it reproduces on master as soon as a test closes a window.

## Testing

- New `TestMainWindow::checkClosingWindowDoesNotUseFreedMembers()`: opens a second window, closes it,
  and checks the window count returns to its previous value. It passes in an ordinary build — the
  failure it exists to catch is the sanitizer report — so it was confirmed failing first: before the
  fix ASan reported the use-after-free with both the allocation and the free traced into this test's
  own lines; after the fix, `--run-all-gui-tests` is ASan-clean with 0 failures.
- ASan configuration: `-DENABLE_GUI_TESTS=1 -DCMAKE_BUILD_TYPE=RelWithDebInfo`,
  `-fsanitize=address -fno-omit-frame-pointer -g`, run with
  `ASAN_OPTIONS=detect_stack_use_after_return=1:detect_leaks=0`.
- `ctest -j8` on macOS: 2758/2761, the only failures being the 3 known SVG arc tessellation tests that
  also fail on master.

Closing a window no longer reads memory belonging to parts of that window that have already been torn
down. The effect was not reliably visible — usually nothing appeared to go wrong, since the freed
memory was typically still intact — but it could produce a crash on window close. No setting or
workflow changes.


jordanbrown0:
Is there a way to remove **all** event filters from the MainWindow, without needing to know what each of them is and remove them individually?

t-paul:
I don't think this is the correct solution for this case. The `RubberBandManager` is a `QObject`, so I believe this should be created via `new` and handed to Qt which handles the life cycle.

whosawhatsis:
Not with Qt's public API, unfortunately. There is no `removeEventFilters()`, and no way to
enumerate either direction of the relationship: the filters installed *on* an object live in
`QObjectPrivate::eventFilters`, and an object that has installed itself elsewhere keeps no list of
where. So "remove all of mine" can only be spelled out target by target.

Qt does clean this up automatically — but in `~QObject`, which for `MainWindow` runs *after* its
members are destroyed. That gap is exactly where this bug lives: `~MainWindow` destroys
`rubberBandManager`, the reparenting that causes delivers events, and the filter is still installed
at that moment because the base-class destructor has not run yet.

`MainWindow` installs itself in three places — on itself, on each dock, and on the find panel. The
destructor already removed the dock filters for this same reason; this PR adds the two that were
missed, so the set is now complete.

If you would rather it not depend on that set staying complete, the small version is to route the
installs through a helper that records its targets and removes them in its own destructor, with the
helper declared last in `MainWindow` so it is destroyed first:

```cpp
class EventFilterGuard {
public:
  explicit EventFilterGuard(QObject *filter) : filter_(filter) {}
  void install(QObject *target) { target->installEventFilter(filter_); targets_.append(target); }
  ~EventFilterGuard() { for (auto& t : targets_) if (t) t->removeEventFilter(filter_); }
private:
  QObject *filter_;
  QList<QPointer<QObject>> targets_;
};
```

That makes a future `installEventFilter(this)` unable to reintroduce this class of bug, which the
current fix does not. I left it out to keep the bugfix minimal and reviewable, but I am happy to add
it here or as a follow-up — whichever you prefer.

jordanbrown0:
> Not with Qt's public API, unfortunately. 

Sigh.

Yeah, writing our own install/remove/removeAll seems like the conservative thing to do, to make it harder to reintroduce.

I would have an explicit removeAll that we can call at the right point, rather than relying on C++ teardown order.  (But that's at least partially because I'm not a C++ expert and so would rather not rely on subtle behavior.)

For extra credit, file an enhancement request with Qt.

PR #7048, opened 2026-09-17
Introduce color_index_t for PolySet color indices

whosawhatsis:
## Summary

Introduces a small `color_index_t` type for the raw `int32_t` values stored
in `PolySet::color_indices`, replacing the ad-hoc `>= 0` check + cast pattern
repeated at every consumer with a single `.index()` call that returns
`std::optional<size_t>`.

This is one of the preparatory refactorings requested by @t-paul in #6977 to
shrink the eventual process-isolation PR. It's a companion to #7044: that PR
originally changed `VBOBuilder`'s local `color_index` to
`std::optional<size_t>` per @jordanbrown0's review, but that type can only
express "present" or "absent" — the process-isolation design needs a color
index that is *sometimes present, sometimes absent, and when absent, one of
two different reasons why* (see below), so a purpose-built type was needed
instead. #7044 has been trimmed back to just its uncontested de-duplication;
this PR is the follow-up.

## Why three states, not one

Upcoming work (a separate compute-worker process for geometry evaluation)
introduces two new tags a face's color index can carry, distinct from
"no specific color":
- a face whose color comes from the viewer's **cutout** (subtracted-face)
  color
- a face whose color is the viewer's **default** face color

A worker process can't resolve either of those against a color scheme
itself — it has no window and no scheme of its own — so it tags the face
with one of these two markers instead, and the *window* resolves the actual
color against whichever scheme the user currently has selected. That lets
the user change color schemes without a round trip back to the worker.
`std::optional<size_t>` collapses all of "no color" / "cutout" / "default"
into a single `nullopt` and can't distinguish them; a plain `int32_t` with
magic negative numbers works but pushes casts and bounds checks into every
call site. `color_index_t` gives the three states names and keeps every call
site simpler than before.

## What changed

- New `src/geometry/ColorIndex.h`: `color_index_t` wraps a raw `int32_t` with
  an implicit constructor (so the many existing `-1` / `colors.size()` /
  plain-int call sites need no ceremony), three named sentinels
  (`kNoColor = -1`, `kCutout = -2`, `kDefault = -3`), and
  `std::optional<size_t> index()` as the only way to obtain an actual index
  into `PolySet::colors` — every sentinel returns `nullopt`, so a missing
  value can never accidentally be used as an array index.
  `static_assert(sizeof(color_index_t) == sizeof(int32_t))` keeps it
  layout-compatible with the old type.
- `PolySet::color_indices` retyped from `std::vector<int32_t>` to
  `std::vector<color_index_t>` — no change to serialization or memory
  layout, since the wrapper is a single `int32_t` under the hood.
- Every consumer updated: `VBOBuilder`, `export_off.cc`, `export_pov.cc`,
  `export_wrl.cc`, `export_3mf_v1.cc`, `export_3mf_v2.cc`, `export.cc`,
  `PolySetBuilder`, `PolySetUtils`, `manifoldutils.cc`. Every `>= 0` check
  plus `static_cast<size_t>` became a single `.index()` call; no casts
  remain anywhere in this diff.
- One small behavior refinement fell out of giving sentinels names:
  `PolySetBuilder::appendPolySet`'s color-index remap used to collapse
  *any* negative source value to plain `-1`; it now preserves whichever
  sentinel it saw, which matters once the cutout/default tags exist upstream
  of it.

## Testing

- New Catch2 test `src/geometry/ColorIndex_test.cc`, written before the
  type existed: sentinel identity, that every sentinel's `.index()` is
  `nullopt`, that a real index round-trips through `.index()`. Added a
  `TEST_SOURCES` glob for `src/geometry/*_test.cc` (none existed before).
- `./OpenSCADUnitTests`: all tests pass, 346 assertions / 57 test cases.
- `ctest -j8`: only the 3 known pre-existing macOS SVG-arc-tessellation
  failures (`export-svg*arcs01`), nothing else.
- `./scripts/beautify.sh --check` (clang-format 18.1.8): clean.

User documentation for the wiki is in this section — there is none needed
here. This is an internal refactoring of how `PolySet` stores per-face color
indices; there is no change to `.scad` language behavior, exported output,
or any user-facing interface.


whosawhatsis:
Closing in favour of #7055, which is the same change with the two unused sentinel values (`kCutout`, `kDefault`) removed and rebased onto current master.

Those two were added here for a follow-up that needs them, but nothing in this PR produced or consumed them — so this PR was adding API with no caller, which isn't something to ask a reviewer to accept on trust. They now arrive alongside the code that emits and resolves them.

No review had happened here, so nothing is lost by moving.

PR #7049, opened 2026-09-18
De-duplicate the compile-lock release in MainWindow

whosawhatsis:
A compile request — preview, render, export, validity check — runs with the GUI locked so that a reload or an update triggered elsewhere cannot interfere with it. Three paths end such a request, and all three repeat the same two-line release:

```c++
GuiLocker::unlock();
```

This moves that pair into `MainWindow::unlockAfterRequest()` and calls it from `compileEnded()`, `exceptionCleanup()` and `UnknownExceptionCleanup()`.

**No behavior change** — the member does exactly what the two lines it replaces did.

Every use of `GuiLocker` at a request entry point is left exactly as it is on master, so the lock a site takes is still visible at that site. (An earlier revision of this PR also wrapped `GuiLocker::isLocked()` and `GuiLocker::lock()`; that was dropped after review — see the discussion below.)

### Why

The duplication is real on master today: three copies of a release policy that have to be kept in step, and the auto-reload restart is easy to forget in a fourth one.

It is also preparation for the process-isolation re-implementation (#6977), in the series of small, flag-free refactorings requested there — a window whose geometry is computed in its own process has to release whichever lock it took, which becomes a change inside this one member rather than an edit at three sites.

### Testing

`ctest -j8` on macOS: 2686/2689, the three failures being the known SVG arc tessellation cases (`export-svg_spec-paths-arcs01`, `-fill-stroke`, `-fill-only`), which fail on master too. `./scripts/beautify.sh --check` clean with clang-format 18.1.8. No test file was modified.

### User documentation

None — this is an internal refactoring with no user-visible surface.


jordanbrown0:
I'm not sure that this is an improvement.

It does remove a bit of duplication.  However, on its face it splits responsibility for locking between `MainWindow` functions and `GuiLocker` functions.  For instance, in the snippet:

```
  if (GuiLocker::isLocked()) return;
  GuiLocker::lock();
```
it is obvious that the two lines interact.

One can at least sort of connect the dots between these:
```
  if (GuiLocker::isLocked()) return;
  const GuiLocker lock;
```

But with your change:
```
  if (isBusy()) return;
  const GuiLocker lock;
```
there's no immediate way to connect those two statements to see that they are playing with the same lock.

There's also one remaining GuiLocker::lock()/GuiLocker::unlock() pair (in `MainWindow::rightClick()`).  Why is it special?  Why shouldn't it be replaced?

whosawhatsis:
`rightClick()` is special because it isn't a compile request. It takes the lock for the lifetime of the open selection menu and releases it in the `aboutToHide` lambda, so routing it through `unlockAfterRequest()` would be actively wrong — that member restarts the auto-reload timer, which a closing context menu has no business doing. Same reason the two `const GuiLocker lock;` scopes were left alone: the destructor already does the right thing and there is nothing to replace.

On the main point, you've convinced me. `isBusy()` and `lockForNewRequest()` were one-line renames of a `GuiLocker` call that buy nothing today and cost exactly the legibility you describe — `if (isBusy()) return; const GuiLocker lock;` really does hide that the two statements are about the same lock.

I've cut the PR down to the one piece that stands on its own: `unlockAfterRequest()`, replacing the three copies of unlock-then-restart-auto-reload in `compileEnded()`, `exceptionCleanup()` and `UnknownExceptionCleanup()`. Every `GuiLocker::isLocked()` / `GuiLocker::lock()` call site is now exactly as it is on master. Also rebased onto current master; title and body updated to match.

When #6977 needs a per-window "am I busy?", that's the PR to argue it in, with the use case visible rather than as anticipation.

jordanbrown0:
With respect to rightClick(), I think you're right, but that it's bad design.

For a compile, what happens is:
* Timer has been ticking periodically.
* User requests compile.
* Compile locks.
* Compile stops timer.
* Compile runs.
* Timer cannot fire here, because it's stopped.
* Compile completes.
* Compile unlocks.
* Compile starts timer.

On the other hand, for the context menu, what happens is:
* Timer has been ticking periodically.
* User requests context menu.
* Context menu locks.
* [ Context menu does *not* stop timer ]
* Context menu displays.
* If the timer fires here, it checks the lock and does nothing.
* Context menu is dismissed.
* Context menu unlocks.
* Timer eventually fires again.

Why do we need two different approaches for these two similar situations?  It seems like either:
* The timer should continue to run, just do nothing if locked, or
* The timer should stop while locked, and be restarted when unlocked.

Also:  this lock isn't a boolean lock; it's a counter that allows nested locks.  An unlock call does not necessarily mean that the counter has dropped to zero and so, if the timer is stopped while locked, does not necessarily mean that the timer should be restarted on an unlock call.

ISTM that the simplest resolution is to eliminate the timer stop/restart entirely.  Just let the timer run; if it fires while locked then it's ignored.  Alternatively, but more complex, would be to stop the timer on lock and restart it when the lock count drops to zero.

There's a subtle difference between the two approaches in how long it is from an unlock to the first timer tick.  If the timer is allowed to tick during a locked period, the next tick could be anywhere from immediate to the timer interval.  If the timer is stopped during a locked period and restarted when unlocked, the next tick will be after the full timer interval.  With an interval that is small by human standards, it doesn't really matter.

whosawhatsis:
Agreed, and I think your first option is the right one — with the consequence that it mostly deletes this PR rather than refining it.

Tracing it: the timer's only slot is `checkAutoReload()`, which does nothing but call `actionReloadRenderPreview()`, whose first line is `if (GuiLocker::isLocked()) return;`. So a tick during a locked period is already a no-op. The `autoReloadTimer->stop()` in `prepareCompile()` and `actionReloadRenderPreview()` buys nothing that the lock check isn't already buying, and the three `start()` calls exist only to undo those two `stop()` calls. Nothing else governs the timer except `on_designActionAutoReload_toggled()`, which is where it belongs.

So the stops and starts can all go, and the only difference is the one you name: the phase of the next tick after an unlock, somewhere between immediate and one interval instead of exactly one interval. At the polling period in use that isn't observable.

Your point about the counter is the reason not to take the second option. `guiLocked` is an `unsigned int`, `GuiLocker::unlock()` just decrements it, and the context-menu path takes a lock that outlives the `exec()` — so "restart the timer on unlock" would have to become "restart it when the count reaches zero", which means teaching `GuiLocker` about a timer it has no business knowing about, or routing every unlock through the window. That is more machinery than the thing it manages.

Once the timer handling is gone, `compileEnded()`, `exceptionCleanup()` and `UnknownExceptionCleanup()` are each left with a bare `GuiLocker::unlock()`, and there is no longer any duplicated policy for `unlockAfterRequest()` to hold. I'd rather close this PR and open a smaller one that just removes the timer stop/start, than keep a wrapper around a single call. Let me know if you'd prefer it as a force-push to this branch instead.

On the #6977 motivation I gave for this PR: it doesn't survive either. The lock work the re-land actually needs is making `guiLocked` per-window rather than a static counter, and that is the same edit whether release is centralized or not. Removing the timer stop/start is the part that helps there — `autoReloadTimer` is already per-window while the lock it's sequenced against is global, and that mismatch is worth deleting before, not after, the two scopes get separated.

jordanbrown0:
> Nothing else governs the timer except on_designActionAutoReload_toggled(), which is where it belongs.

And even that isn't truly needed; the timer could run all the time and checkAutoReload() could check the enable flag.  Whether that's simpler or not, shrug.

I guess I would very slightly prefer a new PR, just because this PR's title would no longer be accurate.  Do start that one off with a comment pointing at this one for history, and add a comment here pointing at that one as a replacement.  Regardless, the commit's comment needs to reflect the change actually made.

PR #7050, opened 2026-09-18
Move CsgInfo::compile_products() into a source file

whosawhatsis:
`CsgInfo` is header-only, so every translation unit that includes `CsgInfo.h` also parses `GeometryEvaluator.h`, `CSGTreeEvaluator.h`, `CSGTreeNormalizer.h`, `RenderSettings.h` and `printutils.h` in order to compile a function it does not call.

The header now declares `compile_products()` and includes only what the class definition needs. The body moves **verbatim** to `src/glview/CsgInfo.cc`, added to `CORE_SOURCES` next to the other `glview` sources that build headless.

**No behavior change** — it is byte-for-byte the same code in a different file. All of its dependencies (`CSGNode.cc`, `CSGTreeEvaluator.cc`, `GeometryEvaluator.cc`, `CSGTreeNormalizer.cc`) were already in `CORE_SOURCES`, so nothing new is linked anywhere.

### Why

Preparation for the process-isolation re-implementation (#6977), in the series of small, flag-free refactorings requested there: a product list has to be built and serialized outside the GUI, which needs a compiled `CsgInfo` on the core side rather than a header that only GUI translation units happen to instantiate.

### Testing

`ctest -j8` on macOS: 2686/2689, the three failures being the known SVG arc tessellation cases (`export-svg_spec-paths-arcs01`, `-fill-stroke`, `-fill-only`), which fail on master too. `./scripts/beautify.sh --check` clean with clang-format 18.1.8. No test file was modified.

### User documentation

None — this is an internal refactoring with no user-visible surface.


jordanbrown0:
I'm generally in favor of moving stuff out of header files, for exactly the reasons you give.

However, the only reference to CsgInfo.h is in export_png.cc, so I have to wonder what attracted your attention.  What tangible gain is there to this change?  My guess is that you have a planned change that will make more use of CsgInfo.h, but for the record it would be good to have your justification.

The MSYS2/qt5 test run failed for a reason unrelated to this change - a package installation timed out.  I've restarted it in the hope that it will succeed.

jordanbrown0 (approved):
As I mentioned in the comment, I'd like to know the back story for the change, but it's OK with me.

whosawhatsis:
Your guess is right — it's preparation, not a standalone win. Today `export_png.cc` is indeed the only includer, so the compile-time argument alone wouldn't be worth a PR.

The reason is #6977 (process-isolated rendering). There, the product list has to be built and serialized on the core side, outside the GUI process, so `CsgInfo` needs to be a compiled symbol that the core links rather than a header that happens to get instantiated in whichever GUI translation unit includes it. Once that lands there are several more includers, and the header-only form would pull `GeometryEvaluator.h` / `CSGTreeEvaluator.h` / `CSGTreeNormalizer.h` into each of them.

This is one of the small, flag-free refactorings that were asked for in #6977 so that the eventual functional change is reviewable on its own. Happy to have it stated in the record: the tangible gain today is near zero; the gain is that #6977 doesn't have to move code and change behavior in the same diff.

Thanks for restarting the MSYS2/qt5 job.

